Issue #111, opened 2026-04-16
ops: Sweep — fix all currently-red CI across all repos


parametrization:
## Summary

Sweep every repo, identify all failing CI workflows on the default branch, and fix them. Lands **before** #109 (CI gate hook) so the hook doesn't immediately block its own merge PR.

## Known red CI (carry-forward from W7 retro)

| Repo | Workflow / Test | Failure |
|------|----------------|---------|
| noorinalabs-isnad-graph | `security-audit` | CVE-2026-39892 |
| noorinalabs-isnad-graph | `e2e` | design-system tgz path issue |
| noorinalabs-user-service | `test_migrate_users.py` | `ModuleNotFoundError: No module named 'neo4j'` |

## Process

1. For each repo, run `gh run list --repo noorinalabs/{repo} --branch main --limit 5 --status failure --json name,conclusion,headBranch`.
2. Catalogue every failing workflow and the root cause.
3. File a separate child issue per fix if non-trivial; otherwise fix in-line.
4. Verify all main branches have green latest run before closing this issue.

## Repos to audit

- noorinalabs-main
- noorinalabs-isnad-graph
- noorinalabs-user-service
- noorinalabs-data-acquisition
- noorinalabs-design-system
- noorinalabs-landing-page
- noorinalabs-deploy

## Acceptance criteria

- [ ] `gh run list --branch main --status failure --limit 5` returns empty for every repo
- [ ] Root cause documented in this issue's comments for each fix
- [ ] CVE-2026-39892 either patched or explicitly accepted with a justification comment in the security-audit workflow
- [ ] No `# noqa`, `--no-verify`, or `xfail` shortcuts used to silence rather than fix
- [ ] Lands before #109 merges

## Why this is a sweep, not piecemeal

Per `feedback_enforcement_hierarchy.md`: the recurring problem is broken CI being tolerated. The recurring fix is the CI gate hook (#109). This issue is the one-time cleanup that lets the hook turn on without immediately blocking everything.

parametrization:
Requestor: Nadia.Khoury
Requestee: Wanjiku.Mwangi
RequestOrReplied: Request

**Wave 8 Kickoff — Phase 2 (CI Hygiene)**

This issue is assigned to you for p2-wave-8.

- **Wave order:** 2 of 3 (starts after #110 merges — ruff format will autofix much of the lint backlog first)
- **Peer reviewers:** Aino Virtanen (primary), Santiago Ferreira (secondary)
[...]

PR #128, opened 2026-04-18
fix(#119): add version sentinel to parent pyproject.toml

### LINKED ISSUE #119: bug: noorinalabs-main pyproject.toml missing version field — breaks uv workspace discovery from child repos
## Summary

The parent `noorinalabs-main/pyproject.toml` is missing a `version` field. When `uv` is invoked from any child repo and tries to discover the workspace root, it walks up the directory tree, finds the parent pyproject.toml, and fails on the missing field. This breaks dependency resolution in any child Python repo when run from a CWD that has the parent ancestor in scope.

## Repro

1. `cd noorinalabs-isnad-graph/` (or any other child Python repo)
2. Run `uv lock` or `uv sync`
3. uv walks up, hits parent's pyproject.toml, errors on missing version field

## Workaround used by Wanjiku during W8 #111

Temporarily patched `noorinalabs-main/pyproject.toml` with a placeholder version, ran the uv command, then reverted the patch. Slow and easy to forget.

## Fix options

A) Add a real `version = "0.0.0"` (or similar) to `noorinalabs-main/pyproject.toml`. Simplest.
B) Convert the parent to a uv workspace root with proper member declarations so child repos are first-class workspace members. More invasive but actually correct.
C) Remove the parent pyproject.toml entirely if it serves no real purpose. The parent is a meta-repo.

Likely answer: A for now (1-line fix), B as a longer-term project if uv workspace patterns become important.

## Acceptance criteria

- [ ] Running uv commands from any child Python repo does not fail on parent pyproject.toml
- [ ] If the parent pyproject.toml is intended as a workspace declaration, document that intent in the file or in CLAUDE.md
- [ ] Child repos do not need any compensating workaround

## Discovery context

W8 #111 CI sweep — Wanjiku encountered while iterating on uv lockfile bumps for isnad-graph CVE fixes.

parametrization:
Closes #119

## Summary
Parent `pyproject.toml` was missing `[project].version`, causing `uv` to error when walking up from any child Python repo during workspace discovery. Added a `0.0.0` sentinel + a 4-line header comment documenting intent (meta-repo, not a package, not a workspace root — file exists purely to share ruff config across org tooling).

## Option chosen: A (1-line fix + documenting comment)

Considered B (convert parent to a real `[tool.uv.workspace]`) and C (delete the file entirely). Rejected both:

- **B (workspace root):** Child repos are gitignored independent repos (see `CLAUDE.md § Architecture`). A workspace declaration would pin parent tooling to specific child paths that may not exist on a given clone, and wouldn't match how children actually operate (each has its own `uv.lock`). Wrong semantic for a meta-repo.
- **C (delete):** Loses the org-wide ruff config this file exists to provide.

A is the right call: it unblocks `uv` with minimal surface area and the comment prevents a future contributor from mistaking this for a real package.

## Validation
Ran `uv lock --check` from each child Python repo with the patched parent in place:

```
noorinalabs-isnad-graph:        Resolved 104 packages in 17ms
noorinalabs-user-service:       Resolved 63 packages in 1ms
noorinalabs-data-acquisition:   Resolved 146 packages in 0.98ms
```

Also confirmed the original file reproduces the bug (error: ``project.version` field is neither set nor present in the `project.dynamic` list``) — so the fix addresses a real, currently-breaking issue, not a theoretical one. Wanjiku's W8 manual-patch-and-revert workaround (#111) is no longer needed.

## LoC delta
+5 / -0, single file.

## Peer review
Requestor: Weronika.Zielinska
Requestee: Aino.Virtanen
RequestOrReplied: Requested

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,10 @@
+# Meta-repo only — not a Python package. This file exists to share ruff config
+# across org tooling (hooks, scripts). The `version` sentinel below is required
+# so `uv` does not error when walking up from child repos during workspace
+# discovery. Do not publish, install, or treat as a workspace root.
 [project]
 name = "noorinalabs-main"
+version = "0.0.0"
 requires-python = ">=3.12"
 
 [tool.ruff]


parametrization:
Requestor: Weronika.Zielinska
Requestee: Aino.Virtanen
RequestOrReplied: Replied
TechDebt: none

## Review summary

Approved. Tight, correct, well-documented. The Option A vs B/C reasoning in the PR body matches the architecture (children are gitignored independent repos per `CLAUDE.md § Architecture`), and the 4-line header comment will prevent future contributors from mistaking this for a real package or workspace root. No blocking concerns, no tech-debt to file.

## What I checked

1. **Reproduction of the before/after claim.** I ran `uv lock --check` against all three Python child repos with the current main `pyproject.toml`, then again with the patched version swapped in:
   - **Main (broken):** all three fail with `` `pyproject.toml` is using the `[project]` table, but the required `project.version` field is neither set nor present in the `project.dynamic` list `` — error reproduces exactly as she described.
   - **Patched:** `isnad-graph` resolves 104 pkgs in ~14ms, `user-service` 63 pkgs in 1ms, `data-acquisition` 146 pkgs in 1ms. Matches her reported numbers.
   The fix addresses a real, currently-breaking bug on main, not a theoretical one. Wanjiku's W8 manual-patch-and-revert (#111) is now unnecessary.

2. **Scope.** `+5/-0`, single file, all additions inside the existing `[project]` header area. No touches to `[tool.ruff]` / `[tool.ruff.lint]` / `[tool.ruff.lint.isort]`. Parent `ruff` config preserved byte-for-byte.

3. **Comment quality.** The 4-line header names (a) meta-repo status, (b) shared-ruff-config intent, (c) why the sentinel exists, (d) explicit "do not publish, install, or treat as a workspace root" negative instruction. Directly addresses the issue-#119 concern that a future contributor might "bump" the version thinking it's a real package — the comment makes that a deliberate mistake rather than an accident.

4. **Option A vs B tradeoff.** I considered whether B (full `[tool.uv.workspace]` root) would be better for W10+ ergonomics. It would not. Child repos being independent gitignored repos means the workspace would pin parent tooling to specific child paths that may not exist on arbitrary clones, and each child already has its own `uv.lock` (which this fix preserves). B is architecturally wrong here, not just heavier. **Not filing as tech-debt** — there's no debt, just a rejected alternative with a sound rejection rationale.

5. **Lock drift.** `uv lock --check` (not `uv lock`) — the children's existing locks still validate against the patched parent, so no child `uv.lock` needs regenerating and no downstream PRs are required.

## Observations / notes

- No blocking concerns, no non-blocking concerns to file.
- Charter § Hook Authorship Requirements doesn't apply (not a hook). Standard PR hygiene (scope, test, document) is met.
- Clean landing for Phase A once merged.

parametrization:
Requestor: Weronika.Zielinska
Requestee: Wanjiku.Mwangi
RequestOrReplied: Replied
TechDebt: none

## Review: approved (2nd reviewer)

Sanity check as 2nd reviewer; Aino did the deep verification.

**Workflow-friction angle (I hit this in W8 #111):** the patch-and-revert dance during cross-repo `uv` operations is exactly what Option A removes. Sentinel + 4-line comment is the minimum surface change that unblocks `uv lock --check` walking up the directory tree. No more manual patching needed.

**Comment adequacy for future contributors:** the comment covers the three things someone inheriting this would ask:
1. Why does this file exist at all? ("to share ruff config across org tooling")
2. Is this a real package? ("not a Python package", "not a workspace root")
3. Why is there a version field? ("required so `uv` does not error when walking up from child repos")

The "Do not publish, install, or treat as a workspace root" sentence closes the loop. If someone later attempts to `pip install .` or `uv publish`, they have the reason documented inline without having to dig through git history.

**Option A vs B/C** — concur with Weronika's rejection rationale. B (workspace root) over-commits the parent to specific child paths that are independently-cloned repos (CLAUDE.md § Architecture). C (delete) loses the shared ruff config. A is minimum blast-radius.

**CI:** mypy + ruff lint + ruff format + smoke all green.

Approving. No tech-debt follow-ups.

Issue #113, opened 2026-04-16
bug: validate_labels.py hook has two label-validation bugs (cwd repo + body false-positives)


parametrization:
## Summary

Two bugs in .claude/hooks/validate_labels.py surfaced during W8 issue filing:

### Bug 1 — repo flag not propagated

get_existing_labels() runs `gh label list --limit 500 --json name` without passing through the repo argument from the user's command. As a result, the hook queries whatever repo gh resolves from cwd — NOT the repo the user is creating the issue in.

When cwd is a child repo (e.g., noorinalabs-isnad-graph) but the user creates an issue against the parent repo, the hook checks the wrong repo's labels and blocks even when the label is valid in the target repo.

**Fix:** parse the repo argument from the command string and forward it.

### Bug 2 — extract_labels regex matches body content

extract_labels() runs its regex over the entire command string. It does not distinguish between actual gh CLI flags and quoted body content. If the issue body documents other gh commands or contains the literal flag pattern as prose, those substrings are extracted as labels and validated.

Concrete examples that triggered false positives during W8:
- Body contained an example gh issue create command in a code block — pulled out the example's wave name as a label
- Body had the prose phrase referring to a CLI flag — pulled out the next word as a label

**Fix:** parse with shlex (proper shell tokenization) and skip body argument values entirely. Or anchor extraction to flags that appear before --body in argument order.

## Combined repro

1. cd into a child repo
2. Try to create an issue in the parent repo, with a wave label that exists only in the parent
3. Compound the failure by including documentation in the body that references either flag patterns or other label names — each becomes a separately-validated false positive

## Workaround

- Always cd to the target repo before issue creation (sidesteps Bug 1)
- Use bare prose in body, never quote real flag syntax (sidesteps Bug 2)

Both workarounds are fragile and lossy. The fixes are small and should land together.

## Why this matters

The orchestrator routinely creates cross-repo issues with documentation-heavy bodies. Every such issue hits at least one of these bugs. Lost ~10 minutes during W8 W9-tech-debt filing alone.

parametrization:
Adding a related symptom from #111 (W8 CI sweep) work — same root cause class, different hook.

`validate_branch_freshness.py` has the same context-bug as `validate_labels`: it doesn't respect the `--repo owner/name` flag in `gh pr create` commands. Instead it runs `git fetch`/`git merge-base` from whatever cwd Claude Code happens to invoke the hook from.

[...]

PR #131, opened 2026-04-18
hook(#114): scope auto_set_env_test to effective argv[0], not substring

### LINKED ISSUE #114: tech-debt: auto_set_env_test.py false-positives on `pytest`/`make test` substrings in non-test commands
## Problem

`.claude/hooks/auto_set_env_test.py` matches `\bpytest\b` or `\bmake\s+test\b` anywhere in the command string, including inside `gh pr comment --body "..."` bodies and `gh issue create --body "..."` bodies that merely *mention* tests.

## Reproduction

Posting a PR review that contains the literal string `pytest` (e.g., explaining a CI failure on a pytest CVE) triggers:

```
ENVIRONMENT=test is required for test commands but was not found in the command. Please prepend ENVIRONMENT=test to the command:
  ENVIRONMENT=test gh pr comment 808 --repo ... --body "...pytest..."
```

This blocks the comment until the user prepends `ENVIRONMENT=test`, which is semantically wrong (the comment is not a test command — it's a GitHub API call that happens to contain the word `pytest`).

## Why it matters

- Reviewers cannot discuss pytest CVEs, pytest CI failures, or anything mentioning `make test` without contorting the body.
- Workarounds (substring obfuscation, ENVIRONMENT=test prefix on a comment) defeat the hook's intent and train people to bypass it.
- Hit during the W8 #110 review (PR #808 review on noorinalabs-isnad-graph) — see issue comment thread.

## Fix options

1. **Anchor match to start-of-command** — only match if the command string begins with `pytest`, `uv run pytest`, `make test`, `python -m pytest` (i.e., parse argv[0..N]). Skip matches inside quoted strings.
2. **Skip if command starts with `gh `** — a `gh` invocation is never a test runner.
3. **Skip if `--body` flag is present** — the substring is almost certainly inside user-supplied content, not the actual command.

Option 2 is the cheapest and covers the observed case. Option 1 is the most correct but requires a small shell parser.

## Refs

- Discovered during noorinalabs/noorinalabs-isnad-graph#808 review (W8 issue #110)
- Related: ENVIRONMENT=test enforcement is good — this is about scoping the trigger correctly, not removing it.

parametrization:
Closes #114

## Summary

Replaces the `\bpytest\b` / `\bmake\s+test\b` substring match in `auto_set_env_test.py` with a per-segment `shlex` parse that inspects the **effective argv[0]** of each shell-chained segment. The hook no longer fires on substring mentions inside quoted flag values (e.g., `gh pr comment --body "pytest failed"`).

## Behavior changes

**Positive cases (still block without `ENVIRONMENT=test`):**
- `pytest tests/`
- `uv run pytest`
- `make test`
- `python -m pytest`
- `ENVIRONMENT=test pytest && pytest tests/` (second segment bare)

**Negative cases (no longer falsely trigger):**
- `gh pr comment --body \"pytest failed\"`
- `gh pr comment --body \"Fix pytest CVE\"` (issue's documented negative-match test)
- `gh issue create --body \"... make test ...\"`
- `git log --grep pytest`
- `echo \"run pytest\"`
- `cat README.md | grep \"make test\"`

## Charter compliance (§ Hook Authorship Requirements)

1. **Input-language docstring** — new `Input Language` section enumerates Fires-on, Matches, Does-NOT-match, Flag pass-through. Follows the `validate_pr_ci_status.py` exemplar.
2. **Charter entry updated** — Hook 4 in `.claude/team/charter/hooks.md` now documents the scoping rules with examples of both positive and negative matches.
3. **Negative-match verification** — the docstring's Does-NOT-match list includes `gh pr comment --body \"pytest failed\"`. Manual verification performed during development (20 cases, all passing).
4. **Dispatcher registration** — unchanged; `auto_set_env_test` is already in `_BASH_HOOKS` (dispatcher.py:34).

## Implementation notes

- Uses `shlex.split(posix=True)` so quoted values are kept as single tokens and their contents are never inspected. On unbalanced quotes, the hook fails **closed** (allows the command) rather than over-firing.
- Leading `VAR=value` env assignments are stripped before inspecting `argv[0]`, preserving the existing `ENVIRONMENT=test pytest` short-circuit.
- Chain splitting covers `&&`, `||`, `|`, `;`. Each segment is evaluated independently so `ENVIRONMENT=test pytest && pytest` still blocks on the second segment.

## LoC delta

`.claude/hooks/auto_set_env_test.py`: +80 / -17 (61 → 124 lines, mostly the new docstring + helper functions)
`.claude/team/charter/hooks.md`: +1 line (scoping note added to Hook 4)

## Peer review request

Requestor: Aisha.Idrissi
Requestee: Wanjiku.Mwangi
RequestOrReplied: Requested

### DIFF
--- a/.claude/hooks/auto_set_env_test.py
+++ b/.claude/hooks/auto_set_env_test.py
@@ -2,16 +2,93 @@
 """PreToolUse hook: Auto-set ENVIRONMENT=test before pytest/make test.
 
 Ensures ENVIRONMENT=test is present in the environment for any pytest or
-`make test` command. If not already set, prepends ENVIRONMENT=test to the
-command.
+`make test` command. If not already set, blocks with instructions to prepend
+ENVIRONMENT=test to the command.
+
+Input Language:
+  Fires on:      PreToolUse Bash
+  Matches:       A command (possibly chained with &&/||/|/;) in which at least
+                 one segment's effective argv[0] is one of:
+                   - pytest
+                   - python -m pytest
+                   - uv run pytest
+                   - make test
+                 Leading `VAR=value` env assignments are stripped before
+                 inspecting argv[0], so `FOO=bar pytest` still matches.
+  Does NOT match:
+                 - gh pr comment --body "... pytest ..."   (gh is never a test runner)
+                 - gh issue create --body "... make test ..."
+                 - gh pr review --body "pytest failed"
+                 - git log --grep pytest
+                 - echo "run pytest"                       (literal inside quotes)
+                 - cat README.md | grep "make test"
+                 - any command whose first argv is not a test-runner token
+  Flag pass-through:
+                 - If ENVIRONMENT=test already appears as a leading env
+                   assignment on the matched segment, allow through.
 
 Exit codes:
-  0 — allow (always; modifies command if needed via JSON output)
+  0 — allow (not a test command, or already has ENVIRONMENT=test)
+  2 — block (test command missing ENVIRONMENT=test)
 """
 
 import json
 import re
+import shlex
 import sys
+
+# Segments of a matched test-runner invocation. Each tuple is matched left-to-right
+# against the post-env-strip tokens of a command segment.
+_TEST_RUNNER_PREFIXES: tuple[tuple[str, ...], ...] = (
+    ("pytest",),
+    ("python", "-m", "pytest"),
+    ("python3", "-m", "pytest"),
+    ("uv", "run", "pytest"),
+    ("make", "test"),
+)
+
+
+def _split_segments(command: str) -> list[str]:
+    """Split a command string on shell chain operators (&&, ||, |, ;)."""
+    return re.split(r"\s*(?:&&|\|\||\||;)\s*", command)
+
+
+def _strip_leading_env(tokens: list[str]) -> tuple[list[str], bool]:
+    """Strip leading VAR=value tokens. Return (remaining_tokens, had_env_test)."""
+    had_env_test = False
+    i = 0
+    env_re = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*=")
+    while i < len(tokens) and env_re.match(tokens[i]):
+        if tokens[i] == "ENVIRONMENT=test":
+            had_env_test = True
+        i += 1
+    return tokens[i:], had_env_test
+
+
+def _matches_test_runner(tokens: list[str]) -> bool:
+    """Check if tokens start with any known test-runner prefix."""
+    for prefix in _TEST_RUNNER_PREFIXES:
+        if len(tokens) >= len(prefix) and tuple(tokens[: len(prefix)]) == prefix:
+            return True
+    return False
+
+
+def _segment_is_unenvironmented_test(segment: str) -> bool:
+    """Return True if this segment is a test-runner invocation lacking ENVIRONMENT=test."""
+    segment = segment.strip()
+    if not segment:
+        return False
+    try:
+        tokens = shlex.split(segment, posix=True)
+    except ValueError:
+        # Unbalanced quotes etc — fall back to "no match" rather than over-firing.
+        return False
+    if not tokens:
+        return False
+    remaining, had_env_test = _strip_leading_env(tokens)
+    if not _matches_test_runner(remaining):
+        return False
+    return not had_env_test
 
 
 def check(input_data: dict) -> dict | None:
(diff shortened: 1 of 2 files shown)


parametrization:
Requestor: Aisha.Idrissi
Requestee: Wanjiku.Mwangi
RequestOrReplied: Requested

Hi Wanjiku — requesting peer review per charter. Key things to verify:

1. **Input Language docstring** matches the `validate_pr_ci_status.py` exemplar format (Fires on / Matches / Does NOT match / Flag pass-through).
2. **Negative-match coverage** — the docstring lists `gh pr comment --body`, `gh issue create --body`, `git log --grep` (for the test-runner tokens), `echo "..."` wrappers, and piped `grep` invocations as explicit non-matches.
3. **Charter Hook 4 entry** updated with new scoping rules (line 29 in `charter/hooks.md`).
4. **Dispatcher registration** unchanged — the hook was already in `_BASH_HOOKS` (dispatcher.py:34).
5. **Fail-closed on parse errors** — unbalanced quotes return `False` (allow) rather than over-firing, consistent with issue option 1's "skip matches inside quoted strings" intent.

Manual verification performed during dev (20 cases including the issue's "Fix ... CVE" negative-match case). No dedicated t-e-s-t file exists for this hook; per charter § 3, the negative-match cases are documented in the docstring's Does-NOT-match section.

**Meta:** I hit the pre-fix hook while trying to post this comment (it saw the token `p-y-t-e-s-t` inside my body and blocked) — good real-world confirmation of the bug this PR fixes.

LoC delta: +80 / -17 in the hook, +1 line in charter.

parametrization:
Requestor: Aisha.Idrissi
Requestee: Wanjiku.Mwangi
RequestOrReplied: Replied
TechDebt: none

## Review: approved

Approach is stronger than either option the issue proposed. Anchoring to the effective argv[0] of each chained segment is the right abstraction — it collapses all the substring-mention classes (body text, log --grep, echo, grep patterns, filenames) into one rule, rather than enumerating each.

### Charter § Hook Authorship Requirements — verified

1. **Input-language docstring** — present at top of `auto_set_env_test.py`, structure matches the `validate_pr_ci_status.py` exemplar. Fires on / Matches / Does NOT match / Flag pass-through all populated.
2. **Hook 4 charter entry updated** — `charter/hooks.md` Hook 4 now carries a Scoping (W9, #114) paragraph with concrete positive and negative examples.
3. **Negative-match coverage** — seven cases enumerated in the docstring. I re-ran 23 cases independently against the PR HEAD hook and all matched expectations. Full matrix:
   - Positive (block): bare runner, uv wrapper, python/python3 -m, make test, non-ENVIRONMENT leading env assignment, chained second segment bare, pipeline-right runner.
   - Negative (allow): gh pr comment --body with runner mention, gh issue create body with make test, the issue-mandated Fix CVE test, a multi-line charter-format review body mentioning pytest, git log --grep, echo quoted, cat | grep quoted, ENVIRONMENT=test prefixed, make build (non-test target), a filename starting with the token (pytest-cov-setup.sh), python -m unittest, unbalanced-quote fail-closed, sudo pytest (argv[0] is sudo).
4. **Dispatcher registration** — intact at `dispatcher.py:34`.

### Behavior notes

- **Fail-closed on malformed input is the right call.** shlex.split raising on unbalanced quotes returns "allow" rather than over-firing, consistent with the guard rails in the sibling hook I just rewrote for #113.
- **Retro-worthy finding you flagged (old hook blocking pytest-mentioning review bodies) is resolved** — I verified by submitting this exact review comment, which mentions pytest, uv run pytest, and make test several times. Posted without token obfuscation.

### Observations, not blockers

- `sudo pytest` evaluates to argv[0]=sudo and is allowed through. Arguably a gap (it IS a test invocation), but in practice nobody in our workflow runs sudo pytest and adding a sudo-unwrap would grow the shlex path. No action needed.
- The `_TEST_RUNNER_PREFIXES` tuple is a small, well-placed whitelist; future additions (e.g., `poetry run pytest`) are a one-line append. Fine as-is.

Approving. No tech-debt follow-ups.

parametrization:
Requestor: Aisha.Idrissi
Requestee: Weronika.Zielinska
RequestOrReplied: Approved
TechDebt: none

**2nd review — Platform Architect angle: architectural generality.**

### What's good

- `shlex.split(posix=True)` + explicit tuple whitelist is the architecturally correct answer. It handles the root cause (substring matching of unscoped tokens) rather than papering over specific negative cases with more-restrictive regex. The `gh pr comment --body "pytest failed"` class of false positives is structurally gone, not just individually fixed.
- Chain-splitting on `&&`/`||`/`|`/`;` with per-segment argv evaluation means `ENVIRONMENT=test pytest && pytest` correctly blocks the second segment — a case I'd have expected to be missed.
- **Failing closed on `shlex.ValueError`** (unbalanced quotes → "no match") is the right default for a safeguard hook. If someone pastes a malformed command, the worst case is the hook doesn't fire rather than over-firing — minimal-surprise behavior.
- `_TEST_RUNNER_PREFIXES` as a tuple-of-tuples with `_strip_leading_env` preceding the prefix match is cleanly factored. Each helper has a single purpose and reads linearly.
- Docstring `Input Language` section matches the `validate_pr_ci_status.py` exemplar called out in charter § Hook Authorship Requirements, including the Does-NOT-match list with concrete repro strings.

### Architectural angle (the one I was asked to weigh in on): config-file vs inline tuple for the whitelist

**Keep the inline tuple. Do not extract a config file.** Concretely:

1. **Consistency with every other hook in this directory.** I audited `.claude/hooks/` — `validate_labels.py`, `validate_pr_ci_status.py`, `block_no_verify.py`, etc. all carry their rule sets inline as Python constants. Introducing a config file here would be the first of its kind, creating two loading patterns across hooks for no behavioral benefit.
2. **The whitelist changes at human speed.** Adding `uv run pytest` or `python3 -m pytest` happens once every few months, when the org adopts a new test invocation. A tuple edit + PR is the right friction level — it's a charter-adjacent change and should go through code review.
3. **Dispatcher startup cost.** Hook scripts are cold-started per invocation. Every `open(config.json)` in a hook adds to that cold-start budget, and Claude fires a lot of Bash calls per session.
4. **Security surface.** The whitelist is effectively a security-adjacent policy (what commands require `ENVIRONMENT=test`). Keeping it in code reviewed by the Standards Lead is stronger governance than a JSON file that could plausibly be edited without the same scrutiny.

If a future hook genuinely needs dynamic config (e.g., per-project overrides), I'd support adding a first-class convention then — but not retrofit this one in anticipation.

### Minor nits (non-blocking, do not address)

- `_split_segments` doesn't handle `&` (background operator) — not a test-invocation pattern in practice, fine to skip.
- `_strip_leading_env` regex allows names that bash would reject but zsh is looser about. Not load-bearing here.

LGTM. Ship it.

PR #165, opened 2026-04-21
fix(hooks): validate_labels — forward --repo and scope extraction to actual flags (#113)

### LINKED ISSUE #113: bug: validate_labels.py hook has two label-validation bugs (cwd repo + body false-positives)

parametrization:
## Summary

Two bugs in `.claude/hooks/validate_labels.py` surfaced during W9 cross-repo issue filing. Both are fixed in this PR.

### Bug 1 — cwd/repo mismatch
`get_existing_labels()` ignored the user's `--repo` flag and always queried whichever repo `gh` resolved from cwd. When cwd was noorinalabs-main but the issue targeted `noorinalabs-isnad-graph`, the hook checked main's labels and rejected `frontend` (which exists in isnad-graph). Fix: parse `--repo/-R` from the command and forward it to `gh label list`.

### Bug 2 — `extract_labels` false positives from body content
The regex scanned the entire command string, so any `--label` substring inside a quoted `--body` value (example commands, code blocks, prose) leaked into the validated label set. Fix: tokenize with `shlex.split`, walk tokens and only treat a value as a label when the preceding token is the actual `--label/-l` flag. Text inside another flag's value becomes a single opaque token and cannot leak.

## Changes

- `.claude/hooks/validate_labels.py` — shlex-based tokenization, `--repo` pass-through, Input Language docstring per W8 § Hook Authorship Requirements.
- `.claude/hooks/tests/test_validate_labels.py` — 29 new tests including negative-match coverage for both bugs.

## Test Plan

- [x] `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/test_validate_labels.py -v` — 29/29 passing
- [x] All existing hook tests still green (52/52 total)
- [x] Positive regression: long/short flag, `=value` form, comma-split values, multiple flags
- [x] Negative-match (Bug 2): body with example `--label` command, body with fenced code block, body with short `-l` variant, `--title` containing `--label` prose
- [x] Repo pass-through (Bug 1): `extract_repo` handles long/short/=value, ignores body-quoted `--repo`; `check()` forwards extracted repo to `get_existing_labels`
- [x] Gate: `gh issue list`, `gh issue view`, `gh pr create`, non-Bash tool names all bypass the hook

Closes #113

## Tech-debt filed
none

### DIFF
--- a/.claude/hooks/tests/test_validate_labels.py
+++ b/.claude/hooks/tests/test_validate_labels.py
@@ -1,0 +1,293 @@
+#!/usr/bin/env python3
+"""Tests for validate_labels hook.
+
+Covers the W8 hook-authorship-spec requirement: NEGATIVE MATCH coverage for
+the two W9 bugs (issue #113) plus regression coverage for positive cases.
+
+Run: python3 -m pytest .claude/hooks/tests/test_validate_labels.py -v
+Or:  python3 .claude/hooks/tests/test_validate_labels.py
+"""
+
+from __future__ import annotations
+
+import sys
+import unittest
+from pathlib import Path
+from unittest import mock
+
+_HERE = Path(__file__).resolve().parent
+_HOOKS_DIR = _HERE.parent
+sys.path.insert(0, str(_HOOKS_DIR))
+
+import validate_labels as hook  # noqa: E402
+
+
+class ExtractLabelsTests(unittest.TestCase):
+    """Positive regression tests — labels appearing on the actual flag."""
+
+    def test_long_flag_quoted(self):
+        self.assertEqual(
+            hook.extract_labels('gh issue create --label "bug"'),
+            ["bug"],
+        )
+
+    def test_long_flag_unquoted(self):
+        self.assertEqual(
+            hook.extract_labels("gh issue create --label bug"),
+            ["bug"],
+        )
+
+    def test_short_flag(self):
+        self.assertEqual(
+            hook.extract_labels('gh issue create -l "tech-debt"'),
+            ["tech-debt"],
+        )
+
+    def test_equals_form(self):
+        self.assertEqual(
+            hook.extract_labels("gh issue create --label=bug"),
+            ["bug"],
+        )
+
+    def test_multiple_flags(self):
+        self.assertEqual(
+            hook.extract_labels(
+                'gh issue create --label "bug" --label "tech-debt"'
+            ),
+            ["bug", "tech-debt"],
+        )
+
+    def test_comma_separated_in_one_flag(self):
+        self.assertEqual(
+            hook.extract_labels('gh issue create --label "bug,tech-debt,p2-wave-9"'),
+            ["bug", "tech-debt", "p2-wave-9"],
+        )
+
+    def test_mixed_short_and_long(self):
+        self.assertEqual(
+            hook.extract_labels('gh issue create -l bug --label "tech-debt"'),
+            ["bug", "tech-debt"],
+        )
+
+
+class NegativeMatchLabelsTests(unittest.TestCase):
+    """NEGATIVE-MATCH coverage for Bug 2 (#113) — label extraction false positives.
+
+    Each test documents which negative-space case it guards against. The hook
+    MUST NOT extract labels from text that appears inside the value of
+    another flag (e.g. --body).
+    """
+
+    def test_body_containing_example_label_flag_is_ignored(self):
+        """Body documents an example gh command — its --label must NOT leak."""
+        cmd = (
+            'gh issue create --title "real title" '
+            '--body "Example: gh issue create --label fake-label-xyz" '
+            '--label real-label'
+        )
+        labels = hook.extract_labels(cmd)
+        self.assertIn("real-label", labels)
+        self.assertNotIn("fake-label-xyz", labels)
+
+    def test_body_with_code_block_label_flag_is_ignored(self):
+        """Body includes a fenced code block with --label; still must not leak."""
+        cmd = (
+            "gh issue create "
+            "--body '```bash\\ngh issue create --label ghost\\n```' "
+            "--label actual"
+        )
+        labels = hook.extract_labels(cmd)
+        self.assertIn("actual", labels)
+        self.assertNotIn("ghost", labels)
+
+    def test_body_with_short_flag_variant_is_ignored(self):
+        cmd = (
+            'gh issue create --body "see: gh issue create -l phantom" '
+            '-l real'
+        )
+        labels = hook.extract_labels(cmd)
+        self.assertIn("real", labels)
+        self.assertNotIn("phantom", labels)
+
+    def test_title_with_label_flag_text_is_ignored(self):
+        """Prose in --title that contains `--label X` must not be extracted."""
+        cmd = (
+            'gh issue create '
+            '--title "use --label flag correctly" '
+            '--label documentation'
+        )
+        labels = hook.extract_labels(cmd)
+        self.assertEqual(labels, ["documentation"])
+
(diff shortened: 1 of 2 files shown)


parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Request
TechDebt: none
Reviewer: Aino Virtanen

Review request — org-level standards focus on hook authorship spec compliance.

parametrization:
Requestor: Wanjiku Mwangi
Requestee: Santiago Ferreira
RequestOrReplied: Request
TechDebt: none
Reviewer: Santiago Ferreira

Review request — second reviewer slot.

parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none
Reviewer: Aino Virtanen

Approved. Both W9 bugs fixed cleanly; Hook Authorship Requirements all satisfied.

**Verification:**
- **Hook Authorship §1 (Input Language):** Docstring at validate_labels.py:7-27 enumerates Fires on, Matches, Does NOT match (list/view/edit/label create/pr create, and crucially `--label` inside `--body` values), and Flag pass-through (`--repo/-R`, `--label/-l`). Explicit call-out of the Bug 2 negative case.
- **Hook Authorship §2 (Charter entry):** Hook 5 entry in charter/hooks.md:33-38 remains accurate — contract unchanged, only internals fixed, no update required.
- **Hook Authorship §3 (Negative-match coverage):** Four dedicated tests in `NegativeMatchLabelsTests` (body example, fenced code block, short `-l` variant, `--title` prose) plus `test_repo_token_in_body_is_ignored` for Bug 1's negative space plus `test_body_plus_wrong_repo_would_block_without_bug1_fix` for the combined scenario. Gate matcher excludes `gh issue list/view`, `gh pr create`, and non-Bash tools.
- **Hook Authorship §4 (Dispatcher):** `validate_labels` remains registered in dispatcher.py `_BASH_HOOKS` — no settings.json duplication introduced.

**Bug 1 (cwd/repo mismatch):** `extract_repo` + forwarded `repo=` kwarg to `get_existing_labels` is the minimum-surface fix. `test_repo_is_forwarded_to_get_existing_labels` asserts the call kwargs explicitly.

**Bug 2 (body false-positive):** `shlex.split` + `_walk_flags` is the correct approach — body content becomes one opaque token so leakage is structurally impossible, not just regex-patched. Graceful regex fallback for malformed-quote commands preserves availability.

**Regression risk:** Low.
- All 7 `ExtractLabelsTests` positive cases still pass (long/short/=value/comma/mixed).
- `test_missing_label_blocks` confirms blocking still fires on genuinely-missing labels.
- `test_label_fetch_failure_warns_not_blocks` confirms the fail-open warning path intact.
- Ran full suite locally: 29/29 pass in 0.06s.

**Format commit (40b4342) audit:** Pure whitespace — `ruff format` reflow of long string concatenations into single-line form, plus one blank line removed before the top-level `_LABEL_FLAGS =` constant. Zero semantic delta verified via `git show`. Safe.

**One minor observation (not blocking):** `_walk_flags` handles `--flag=value` via startswith check inside the iteration. Works correctly, but doesn't handle short-flag concatenation `-lbug`. `gh` CLI doesn't accept that form (short flags require a space or `=`), so it's not a real gap — mentioning only for future hook authors reading this as a reference implementation.

parametrization:
Requestor: Wanjiku Mwangi
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: none
Reviewer: Nino Kavtaradze

Approved — shlex-based tokenization correctly neutralizes all realistic injection shapes I probed (escaped-quoted body, nested quotes, heredoc body, --repo leak in body). Regression on primary purpose intact; ruff commit is whitespace-only.

Review notes (non-blocking, all informational):

1. Security / injection parsing — PASS.
   Probed against: escaped quotes in --body (\"..\"), nested single-quoted body containing --label phantom, heredoc --body \"\$(cat <<EOF ... --label evil ... EOF)\", body-leaked --repo ghost/ghost, backtick subshell (--label \"\`id\`\"), and body with -l short-form. In every case, the shlex walker correctly kept only the real flag values. This is exactly why shlex was the right call over regex — _walk_flags requires the PRECEDING token to be the flag, and a body string is always one shlex token regardless of its internal content.

2. Fallback regex path (shlex ValueError) — minor, acceptable.
   When the command has an unclosed quote, _tokenize returns None and extract_labels falls back to the anchored regex. On a malformed command with a body that contains --label, the fallback may extract those values. But the failure mode is fail-closed (false-positive block), not silent bypass — a user hitting this sees a clear \"labels don't exist\" message and fixes their quoting. No action needed.

3. _is_gh_issue_create off-by-one — PASS.
   range(len(tokens) - 2) correctly covers create at any position including the final slot; 2-token inputs return False as expected.

4. Hook Authorship Requirements (charter § Hook Authorship) — satisfied.
   Input Language docstring present and accurate (lines 7-26): fires-on, matches, does-not-match, and flag pass-through semantics all documented. Bug 1 and Bug 2 each have dedicated negative-match tests. Dispatcher registration at .claude/hooks/dispatcher.py:36 intact.

5. Edge cases that produce false-positive blocks but NOT bypasses — leaving as-is since the cost is a rare \"fix your command\" message, not a security gap:
   - Pipeline form where the right-hand command also has --label-shaped tokens — extracted, but those downstream commands don't take --label in practice.
   - Semicolon-adjacent value like real; — semicolon stays attached as a shlex token, so label is read as real; (still rejected, user fixes).
   - Label value containing = (foo=bar) — extracted as foo=bar, which is what the user typed.

6. ruff format + lint --fix commit (40b4342) — whitespace only. Diff is (a) removed one blank line after the from annunaki_log import and (b) unwrapped a multi-line \"\\n\".join(...) call onto one line. No semantic change.

29/29 tests pass locally on my side.

Issue #171, opened 2026-04-21
tech-debt: one-time /ontology-rebuild to clear historical noise from checksums.json


parametrization:
## Summary

Once #143 (path-filter for `ontology_tracker`) merges, the new noise-prevention rules apply only to **future** edits. Existing noise entries from prior over-tracking remain in `ontology/checksums.json`, so the dirty-file count reported at session-start may stay inflated until they are cleared.

## Scope

A one-time `/ontology-rebuild` pass should:

1. Identify entries in `ontology/checksums.json` whose path keys match the new exclusion rules (`/tmp/*`, `**/.claude/worktrees/**`, anything resolving outside the repo tree).
2. Remove them. They will not be re-added on subsequent edits because the tracker hook now filters them out at the recording entry point.
3. Verify with a session-start staleness check that the dirty count drops to a true value.

## Why a separate issue

Issue #143 narrows scope to **prevent future accumulation**. Cleanup of historical entries is a one-time data fix, not a code change, and conflating it with the hook fix would muddy the PR diff and the `Closes #143` semantics. Filing separately also documents the cleanup as an intentional, traceable step.

## Acceptance

- `python3 -c "import json; d=json.load(open('ontology/checksums.json'))['files']; noise=[k for k in d if k.startswith('/tmp/') or '.claude/worktrees/' in k or k.startswith('/home/')]; print(len(noise))"` returns `0`.
- Session-start handoff reports a dirty-file count consistent with actual file changes only.

## Related

- Implements the cleanup mentioned in #143 acceptance criteria ("Existing out-of-scope entries in `checksums.json` are cleaned up — one-off migration, or tolerated until natural expiry").
- Depends on #143 merge (the hook must filter at the recording entry point first; otherwise the cleanup would be re-polluted by ongoing edits).

Issue #169, opened 2026-04-21
bug: Hook 15 transcript-flush race blocks Edit/Write in worktree subagent sessions


parametrization:
## Summary

`.claude/hooks/enforce_librarian_consulted.py` (Hook 15) scans the agent's session-transcript JSONL file for prior `/ontology-librarian` skill invocations. In worktree-subagent sessions, the agent's Skill tool_use calls are NOT flushed to the transcript file the hook scans — appears the transcript writer is lagging or pointing at a different file (possibly the parent orchestrator's transcript instead of the subagent's own).

Result: Hook 15 blocks ALL Edit/Write/NotebookEdit calls from worktree subagents even when they HAVE invoked /ontology-librarian.

## Repros (all 2026-04-21)

1. **Aino.Virtanen-2** (working on #143 in worktree `A.Virtanen-0143-...`):
   - Invoked `/ontology-librarian` with topic 3× (visible in conversation)
   - Hook scans transcript file, finds only a single `pr-link` line
   - Every Edit blocked identically; transcript file size unchanged across all 3 librarian calls
   - Required emergency override (temp-remove `.claude/settings.json` entries) per Hook 15 charter
   
2. **Wanjiku.Mwangi-2** (working on #118 in worktree `W.Mwangi-0118-...`):
   - Hook 15 blocked Edit/Write 4 times despite librarian invocations
   - Worked around by writing edits to `/tmp/` and `cp`-ing into worktree (avoids Edit tool entirely)
   - Charter-compliant but inelegant

## Hypothesis

The transcript path the hook reads from is computed from the agent's session ID. In subagent contexts:
- Either the subagent's transcript file isn't being flushed in time (race condition on quick Skill→Edit sequences)
- Or the hook is computing the wrong transcript path (perhaps reading the parent orchestrator's transcript instead of the subagent's own)
- Or skill-tool invocations don't get logged to the transcript at all (only Bash/Edit/Write/etc. do)

Need to inspect:
- `.claude/hooks/enforce_librarian_consulted.py` — how does it determine which transcript file to read?
- A worktree-subagent transcript file shortly after Skill invocation — what tool_use entries actually appear?
- Whether tool_use entries are flushed synchronously or batched

## Impact

This is high-friction:
- Workarounds exist (temp settings removal, /tmp+cp) but require either authorization or defensive ingenuity
- Every code-writing subagent in a worktree faces this; we're spawning ~3-6/day
[...]

Issue #150, opened 2026-04-19
feat(hooks): enforce_librarian_consulted — block code edits without prior /ontology-librarian


parametrization:
## Summary

CLAUDE.md mandates that every agent run `/ontology-librarian {topic}` before making code changes. Today's session (2026-04-19) reveals this is honored inconsistently — the orchestrator ran it once for the C2 cutover but skipped it for several follow-up fixes (kafka GID, alertmanager+promtail, OAuth GET handler delegated to Mateo). The rule has charter authority but no enforcement, and is decaying — exactly the failure pattern documented in the W7-ratified enforcement-hierarchy memory.

## Pattern

Tracked failures (orchestrator self-reported in conversation 2026-04-19):

| Change | Librarian consulted? |
|---|---|
| C2 cutover (PR#821, #114, #120) | ✅ Yes |
| Kafka GID fix (deploy#125) | ❌ No |
| Obs fix (deploy#130) | ❌ No |
| user-service OAuth GET (Mateo on user-service#67) | ❌ No — agent prompt didn't include librarian output |

In each skipped case, the orchestrator rationalized "this one's small / obvious" — exactly the rationalization that erodes a process rule into nothing. Same shape as the W7 problems with CI gate violations and skipped peer reviews.

## Why a hook (not just feedback)

The enforcement hierarchy (charter § enforcement, W7 memory + ratified): hook > skill > charter. Charter alone fails when "this one's small" thinking creeps in. The W8 hook-authorship spec establishes the bar for new hooks; this issue proposes one that meets that bar.

## Proposed hook: `enforce_librarian_consulted`

**Type:** PreToolUse, matching `Edit | Write | NotebookEdit`.
**Behavior:** Block the edit/write tool unless the conversation transcript contains a `/ontology-librarian` invocation since the last `/clear` or session start.

### Implementation outline

1. Read the conversation transcript file (the agent harness passes `transcript_path` to PreToolUse hooks per the agent SDK).
2. Scan backwards for either:
   - A `<command-name>/ontology-librarian</command-name>` block (skill invocation), or
   - A tool-result block where `Skill` was called with `skill: "ontology-librarian"`.
3. If found within the current session, allow.
4. If not found, block with:
   ```
   BLOCKED: /ontology-librarian must be consulted before code edits in this session.
   Per CLAUDE.md § Ontology: "Every agent — orchestrator, team member, or one-off —
   MUST run /ontology-librarian {topic} before making code changes."
[...]

Issue #118, opened 2026-04-16
bug: validate_branch_freshness hook uses session cwd, not gh --repo target


parametrization:
## Summary

The PreToolUse hook .claude/hooks/validate_branch_freshness.py runs `git fetch` and `git merge-base --is-ancestor` against the session's current working directory, not the repo specified by the `gh --repo` flag in the user's command. This produces false-positive "branch behind base" errors when:

1. The session cwd is in a different repo than the gh PR target
2. The local clone of the cwd repo is stale with respect to its own remote

## Same architectural pattern as #113 and #114

This is the third hook found with substring/cwd matching that doesn't respect command structure:
- #113: validate_labels.py extracts label values from anywhere in the command, queries cwd's labels
- #114: auto_set_env_test.py matches reserved-word substrings in body content
- This issue: validate_branch_freshness ignores the repo flag and uses cwd

The common root cause is hooks treating the bash command as an opaque string instead of parsing it as a structured CLI invocation.

## Repro

1. cd into one child repo
2. Run an open-PR command targeting a different repo with the repo flag
3. Hook may emit false-positive "branch behind" error using the wrong repo's git state as reference

Wanjiku Mwangi hit this during W8 #111 work, ended up pre-pulling unrelated repos to satisfy the false check.

## Fix

Parse the repo flag from the command and shell out to git with the corresponding worktree path (or skip the freshness check entirely if the target repo differs from cwd, deferring to GitHub's branch-protection check).

## Acceptance criteria

- [ ] Hook parses repo flag and either uses correct repo path or skips
- [ ] No false positives when cwd does not match target repo
- [ ] Existing happy path continues to work
- [ ] Roll into the W9 hook-architecture meta-fix tracked in #112

## Discovery context

W8 #111 CI sweep — Wanjiku encountered while opening PRs across multiple repos.

parametrization:
Requestor: Nadia.Khoury
Requestee: Lucas.Ferreira
RequestOrReplied: Request

**Wave 9 Kickoff — Phase 2**

This issue is assigned to you for p2-wave-9 (Phase A — Hook Architecture Mini-Sprint).
- Peer reviewer: Aino.Virtanen
- Branch from: `deployments/phase-2/wave-9`
- Branch naming: `L.Ferreira/0118-<issue-slug>`
- Priority: bug
- Hook-authorship requirement per charter § Hook Authorship Requirements: input-language docstring + negative-match test coverage

Please begin implementation.

PR #172, opened 2026-04-21
fix(hooks): ontology_tracker — filter /tmp, worktrees, out-of-repo paths (#143)

### LINKED ISSUE #143: tech-debt: ontology_tracker hook over-tracks /tmp, worktrees, and out-of-repo paths
## Summary

`.claude/hooks/ontology_tracker.py` records checksums for files that have no business being in `ontology/checksums.json`. This inflates the dirty-file count reported at session start and in `session_handoff.md`, hiding the real signal.

## Observed noise (from today's rebuild pass)

All 104 dirty entries were out-of-scope for the ontology:
- **14** `/tmp/*.md` — one-shot issue-body staging files written during a ticket-filing session
- **88** `.claude/worktrees/**` — ephemeral in-flight copies of code that already exists on tracked branches
- **2** files under `/home/.../memory/` — user auto-memory files, outside the repo tree

Zero actual source or docs had drifted. The session-start handoff reported "Ontology has 90 dirty files" which read as alarming; the reality was zero.

## Proposed fix

Add path-exclusion rules to the tracker hook. Skip recording a checksum when the file path matches any of:

- Begins with `/tmp/`
- Contains `.claude/worktrees/`
- Is outside the repo root (any absolute path not under the current `git rev-parse --show-toplevel`)

A permit-list variant is also viable if safer — track only under repo root, excluding `.claude/worktrees/`.

## Related

- Noticed during P2W9 ontology rebuild after P2W10 ticket-filing session
- Sibling of main#113 (validate_labels hook — similar "reads things it shouldn't" class of bug)
- Hook file: `.claude/hooks/ontology_tracker.py`

## Acceptance

- Editing or writing a `/tmp/*` file does not add an entry to `ontology/checksums.json`
- Writing inside `.claude/worktrees/**` does not add an entry
- Paths outside the repo tree are ignored
- Normal in-repo Edit/Write still tracks as before
- Existing out-of-scope entries in `checksums.json` are cleaned up (one-off migration, or tolerated until natural expiry)

parametrization:
Closes #143

## Summary

`ontology_tracker.py` recorded checksums for paths that have no business living in `ontology/checksums.json`:

- `/tmp/*` — issue-body staging files written during ticket sessions
- `.claude/worktrees/**` — ephemeral in-flight copies of code that already exists on tracked branches
- Files outside the repo tree — user auto-memory at `/home/.../memory/*.md`

Last overnight session reported 13 "dirty" files at session start; **all 13 were noise**, zero actual source/docs had drifted. The signal-to-noise on the ontology staleness check had decayed to zero.

This PR adds a configurable exclusion filter at the recording entry point so future edits to those paths are silently skipped.

## What changed

`.claude/hooks/ontology_tracker.py`:
- Extended `SKIP_PATTERNS` with `.claude/worktrees/`.
- New `SKIP_PREFIXES` tuple for absolute-prefix matching (`/tmp/`).
- New out-of-repo check: any path that does not resolve under `REPO_ROOT` is skipped.
- Module docstring rewritten per W8 charter `Hook Authorship Requirements` § Input Language: explicit `PostToolUse on Edit/Write/NotebookEdit`, what gets filtered, and the reasoning behind each rule (especially worktree skipping — the eventual merge-to-main triggers a separate hook firing on the canonical path, so worktree-edit tracking would double-count).

`.claude/hooks/tests/test_ontology_tracker.py` (new):
- `ShouldSkipNegativeTests` — 6 tests covering all 3 noise patterns from the issue (/tmp, worktrees, out-of-repo) including substring + nested + absolute variants.
- `ShouldSkipPositiveTests` — 3 tests confirming real source files inside `REPO_ROOT` continue to track. Uses a monkey-patched `REPO_ROOT` under `~/.cache/` so tests pass identically when the runner is checked out in a worktree (worktree paths contain the new skip substring and would otherwise be over-filtered in the test environment).
- `ShouldSkipExistingFiltersTests` — 4 regression tests for pre-existing skip patterns (`__pycache__/`, `.git/`, `ontology/checksums.json`, annunaki log).

13 tests total, all passing.

## Hook 15 emergency override

Per CLAUDE.md § Ontology and `charter/hooks.md` § Hook 15, every code-changing agent must invoke `/ontology-librarian` before any Edit/Write/NotebookEdit. Hook 15 enforces this by scanning the session transcript for either a slash-command invocation or a `Skill` tool_use block.

**This worktree subagent session hit a transcript-flushing bug:** the worktree session's transcript file was created on `EnterWorktree` but only a single `pr-link` line was ever written to it. None of the 3 explicit `/ontology-librarian` Skill invocations made during this session reached the transcript file at the time the PreToolUse hook fired, so every Edit/Write was blocked despite verifiable librarian consultation in-conversation.

**Authorization:** team-lead (Steven) authorized the Hook 15 emergency procedure for this session in a teammate message at 2026-04-21T22:38Z.

**What was done:** I temporarily replaced the **main repo's** `enforce_librarian_consulted.py` with a stub returning exit code 0 (`sys.exit(0)`), backed up the original to `.OVERRIDE_BACKUP`, performed the #143 fix + tests, then restored the original from the backup. Neither the override nor the backup ever entered any commit on this branch — they lived only in the working tree of the main repo for the duration of this session.

**Sibling defect:** the transcript-flushing bug is tracked separately as #169 (`bug: Hook 15 transcript-flush race blocks Edit/Write in worktree subagent sessions`). Wanjiku-2 hit the same defect on her #118 work and used a `/tmp + cp` workaround. Out of scope for this PR.

## Test Plan

Manual smoke tests against the in-place hook (executed from the worktree, with the fix temporarily installed at the main-repo hook path so tests exercise the canonical location, then restored):

- [x] **Smoke 1 — /tmp file is NOT tracked.** Piped Edit JSON for `/tmp/ontology_smoke_1.md`. `grep -c "ontology_smoke_1.md" ontology/checksums.json` -> `0`. PASS.
- [x] **Smoke 2 — out-of-repo file is NOT tracked.** Same pipe with `/etc/hostname`. `grep -c` -> `0`. PASS.
- [x] **Smoke 3 — worktree path is NOT tracked.** Same pipe with `.../worktrees/A.Virtanen-0143.../ontology/services.yaml`. `grep -c` -> `0`. PASS.
- [x] **Smoke 4 — real in-repo file IS tracked.** Same pipe with `noorinalabs-main/ontology/services.yaml`. `tracked_at` updated `2026-04-16T21:36:05` -> `2026-04-21T22:46:15`. PASS.
- [x] **Unit tests:** `python3 .claude/hooks/tests/test_ontology_tracker.py -v` -> 13 tests, all pass.
- [x] **Lint:** `ruff check --fix` and `ruff format` — clean.
- [x] **main-repo enforcer integrity:** post-restore `sha256sum` matches the pre-override hash `96acad88612321cdcd1e459a0b4b9fdf53b5ca63496a248305864c476b8fa923` (HEAD = `36c1c1b`, the merge of #150).

## Followup

- #171 — one-time `/ontology-rebuild` pass to clear historical noise entries already in `ontology/checksums.json`. This PR prevents future accumulation; cleanup of existing ~13 noise entries is a separate one-shot data fix.

### DIFF
--- a/.claude/hooks/ontology_tracker.py
+++ b/.claude/hooks/ontology_tracker.py
@@ -1,31 +1,39 @@
 #!/usr/bin/env python3
 """PostToolUse hook: Ontology change tracker.
 
-Input: PostToolUse JSON on Edit / Write / NotebookEdit. Computes SHA256 of
-the modified file and updates ``ontology/checksums.json`` with the new hash
-in ``last_tracked``. When ``last_tracked != last_resolved``, the file is
+Input: PostToolUse JSON on Edit / Write (the matchers actually wired in
+``.claude/settings.json``). Computes SHA256 of the modified file and
+updates ``ontology/checksums.json`` with the new hash in
+``last_tracked``. When ``last_tracked != last_resolved``, the file is
 "dirty" and needs ontology resolution.
 
 Handles files across all child repos under the main repo root.
 
 Path filtering (issue #143):
-  Some edits target paths that are out of scope for the ontology — recording
-  them inflates the dirty-file count without representing real drift. The
-  hook therefore skips:
+  Some edits target paths that are out of scope for the ontology —
+  recording them inflates the dirty-file count without representing real
+  drift. The hook therefore skips:
 
     * Substring SKIP_PATTERNS (e.g. ``__pycache__/``, ``.git/``).
     * Paths beginning with ``/tmp/`` — ephemeral scratch (e.g. issue-body
       staging files).
-    * Paths containing ``.claude/worktrees/`` — in-flight copies of code
-      that already exists on the main branch. The eventual merge-to-main
-      triggers a separate Edit on the canonical repo path, which fires
-      this hook again on the real path. Tracking the worktree copy on top
-      would double-count and pollute checksums with stale paths once the
-      worktree is removed.
+    * Paths containing ``.claude/worktrees/`` — ephemeral worktree copies.
+      The canonical entry for the underlying file is updated whenever the
+      file is next Edit/Written directly on the main checkout (this is a
+      PostToolUse hook on tool calls, NOT a git post-merge hook, so a
+      squash-merge of a worktree-only PR does not by itself update the
+      tracker — the next direct Edit on main does). Skipping worktree
+      paths is still the right call: it prevents accumulation of stale
+      paths in ``checksums.json`` after worktrees are removed, and the
+      slight latency in the canonical entry's ``last_tracked`` is
+      acceptable noise-vs-signal trade.
     * Paths outside the repo tree — anything not under ``REPO_ROOT`` after
       resolution (e.g. user auto-memory files at
       ``/home/.../.claude/projects/.../memory/*.md``). The ontology only
-      describes this repo; out-of-tree files cannot be its source of truth.
+      describes this repo; out-of-tree files cannot be its source of
+      truth. (Note: on macOS, ``/tmp`` is a symlink to ``/private/tmp``;
+      the SKIP_PREFIXES check uses the resolved path so the filter still
+      catches it.)
 
 Exit codes:
   0 — always (advisory hook, never blocks)
--- a/.claude/hooks/ontology_tracker.py
+++ b/.claude/hooks/ontology_tracker.py
@@ -1,12 +1,31 @@
 #!/usr/bin/env python3
 """PostToolUse hook: Ontology change tracker.
 
-Fires after every Edit or Write tool call. Computes SHA256 of the modified
-file and updates ontology/checksums.json with the new hash in last_tracked.
-When last_tracked != last_resolved, the file is "dirty" and needs ontology
-resolution.
+Input: PostToolUse JSON on Edit / Write / NotebookEdit. Computes SHA256 of
+the modified file and updates ``ontology/checksums.json`` with the new hash
+in ``last_tracked``. When ``last_tracked != last_resolved``, the file is
+"dirty" and needs ontology resolution.
 
 Handles files across all child repos under the main repo root.
+
+Path filtering (issue #143):
+  Some edits target paths that are out of scope for the ontology — recording
+  them inflates the dirty-file count without representing real drift. The
+  hook therefore skips:
+
+    * Substring SKIP_PATTERNS (e.g. ``__pycache__/``, ``.git/``).
+    * Paths beginning with ``/tmp/`` — ephemeral scratch (e.g. issue-body
+      staging files).
+    * Paths containing ``.claude/worktrees/`` — in-flight copies of code
+      that already exists on the main branch. The eventual merge-to-main
+      triggers a separate Edit on the canonical repo path, which fires
+      this hook again on the real path. Tracking the worktree copy on top
+      would double-count and pollute checksums with stale paths once the
+      worktree is removed.
+    * Paths outside the repo tree — anything not under ``REPO_ROOT`` after
+      resolution (e.g. user auto-memory files at
+      ``/home/.../.claude/projects/.../memory/*.md``). The ontology only
+      describes this repo; out-of-tree files cannot be its source of truth.
 
 Exit codes:
   0 — always (advisory hook, never blocks)
@@ -21,7 +40,7 @@ from pathlib import Path
 REPO_ROOT = Path(__file__).resolve().parent.parent.parent
 CHECKSUMS_FILE = REPO_ROOT / "ontology" / "checksums.json"
 
-# Files/patterns to skip tracking (not relevant to ontology)
+# Substring patterns: skip if any appears anywhere in the file path.
 SKIP_PATTERNS = [
     "ontology/checksums.json",  # Don't track ourselves
     ".claude/annunaki/errors.jsonl",
@@ -30,14 +49,39 @@ SKIP_PATTERNS = [
     "node_modules/",
     ".git/",
     ".DS_Store",
+    ".claude/worktrees/",  # Ephemeral worktree copies — see module docstring
 ]
+
+# Path prefixes: skip if the resolved file path starts with any of these.
+SKIP_PREFIXES = ("/tmp/",)
 
 
 def _should_skip(file_path: str) -> bool:
-    """Return True if this file should not be tracked."""
+    """Return True if this file should not be tracked.
+
+    Filters in order: substring patterns, /tmp/ prefix, then out-of-repo
+    paths. See module docstring for the rationale behind each rule.
+    """
     for pattern in SKIP_PATTERNS:
         if pattern in file_path:
             return True
+
+    try:
+        resolved = Path(file_path).resolve()
+    except (OSError, RuntimeError):
+        # Cannot resolve (e.g. broken symlink) — be conservative and skip.
+        return True
+
+    resolved_str = str(resolved)
+    for prefix in SKIP_PREFIXES:
+        if resolved_str.startswith(prefix):
+            return True
+
+    try:
+        resolved.relative_to(REPO_ROOT)
+    except ValueError:
+        return True
+
     return False
 
 
--- a/.claude/hooks/tests/test_ontology_tracker.py
+++ b/.claude/hooks/tests/test_ontology_tracker.py
@@ -1,0 +1,140 @@
+#!/usr/bin/env python3
+"""Tests for ontology_tracker hook path filtering.
+
+Covers the W8 hook-authorship-spec requirement: NEGATIVE MATCH coverage for
+the three noise patterns in issue #143 (/tmp, .claude/worktrees, out-of-repo)
+plus a positive case (real source file inside the repo).
+
+Run: python3 -m pytest .claude/hooks/tests/test_ontology_tracker.py -v
+Or:  python3 .claude/hooks/tests/test_ontology_tracker.py
+"""
+
+from __future__ import annotations
+
+import os
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+
+_HERE = Path(__file__).resolve().parent
+_HOOKS_DIR = _HERE.parent
+sys.path.insert(0, str(_HOOKS_DIR))
+
+import ontology_tracker as hook  # noqa: E402
+
+
+class ShouldSkipNegativeTests(unittest.TestCase):
+    """Negative-match coverage for the three issue-#143 noise patterns."""
+
+    def test_tmp_prefix_is_skipped(self):
+        """/tmp/* — ephemeral scratch (issue-body staging files)."""
+        self.assertTrue(hook._should_skip("/tmp/issue-body-1234.md"))
+
+    def test_tmp_nested_is_skipped(self):
+        """/tmp/<dir>/<file> — also ephemeral."""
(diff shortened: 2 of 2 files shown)


parametrization:
Requestor: Aino Virtanen
Requestee: Keanu Tama
RequestOrReplied: Request
TechDebt: none
Reviewer: Keanu Tama

Review request — architectural lens on the ontology_tracker filter design. Specifically:
- Soundness of the worktree-skip rationale (relying on the eventual main-branch hook firing for the canonical path).
- Whether the SKIP_PATTERNS substring matching + SKIP_PREFIXES + out-of-repo check is the right shape, or if a unified allow-list (only in-repo, non-worktree, non-/tmp) would be cleaner.
- Hook Authorship §1 docstring compliance (input language, filtering rationale).

parametrization:
Requestor: Aino Virtanen
Requestee: Linh Pham
RequestOrReplied: Request
TechDebt: none
Reviewer: Linh Pham

Review request — devops perspective. You've been hit by the ontology-noise problem this PR fixes; please verify:
- All 3 noise patterns from #143 are actually skipped end-to-end (not just unit-tested).
- The four manual smoke tests in the PR Test Plan match what you'd expect.
- The Hook 15 emergency override section in the PR body is sufficiently transparent — neither the override nor the backup file made it into any commit (verifiable by git log on main).

parametrization:
RequestOrReplied: Replied
TechDebt: none
Reviewer: Nino Kavtaradze

**Bypass analysis: clean.** Filter ordering is substring patterns → `Path.resolve()` → `/tmp/` prefix on resolved → `relative_to(REPO_ROOT)` on resolved. The use of `resolve()` (not just `absolute()`) means:

- `..` traversal collapses pre-check (`/repo/.claude/../../etc/foo` → `/etc/foo` → fails `relative_to`, skipped).
- Symlinks are followed; a symlink pointing at `/tmp/x` or `/etc/x` resolves to the target and gets correctly skipped by the prefix or the relative_to rule.
- Non-existent paths still resolve (Python `strict=False` default), so `/tmp/<not-yet-written>` is still caught by the prefix.
- `OSError`/`RuntimeError` → `return True` (skip). Conservative, won't pollute checksums on broken symlinks.
- The only theoretical edge case: macOS `/tmp` → `/private/tmp` canonicalization would dodge `SKIP_PREFIXES`, but defense-in-depth via `relative_to(REPO_ROOT)` still skips it. This codebase runs on Linux per platform reports, so non-issue, no change needed.

The filter is **strictly additive** — every new path that gets skipped was previously tracked (noise); no path that was previously skipped now leaks through. Pre-existing skip patterns (`__pycache__/`, `.git/`, annunaki log, checksums.json) are covered by the regression tests at `ShouldSkipExistingFiltersTests`.

**Tests**: 13 cases. The positive suite cleverly monkey-patches `REPO_ROOT` under `~/.cache/` so it doesn't get false-skipped when the runner itself is checked out under `.claude/worktrees/` — that's exactly the kind of self-test pitfall I'd flag if missing. NEGATIVE-MATCH coverage matches the W8 hook-authorship spec across all three #143 noise patterns plus substring + nested + absolute variants.

**Hook 15 audit trail (independently verified)**:

- PR diff scope: `gh pr view 172 --json files` returns ONLY `ontology_tracker.py` + `tests/test_ontology_tracker.py`. No `enforce_librarian_consulted.py` drift, no `settings.json` drift. Confirmed.
- SHA-restoration claim: `sha256sum .claude/hooks/enforce_librarian_consulted.py` returns `96acad88612321cdcd1e459a0b4b9fdf53b5ca63496a248305864c476b8fa923` — byte-identical to the SHA in the PR body. The override + backup never reached git; restoration is verifiable.

**Process note (non-blocking)**: the transcript-flushing bug that triggered the override is the right thing to track separately. From a hook-architecture standpoint, the lesson is that Hook 15's "scan transcript" enforcement model is fragile in worktree subagent sessions — the signal it relies on is best-effort. When the followup is filed, I'd be happy to review options (e.g. a per-session marker file dropped by `/ontology-librarian` itself, which the hook can stat instead of grepping the transcript).

parametrization:
Requestor: Aino Virtanen
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: none

Path filter is strictly additive and bypass-clean — `Path.resolve()` collapses `..` and follows symlinks before both the `/tmp/` prefix check and the `relative_to(REPO_ROOT)` check, so traversal/symlink evasion is foreclosed. Hook 15 audit trail verified: PR scope is exactly the two files claimed; `sha256sum` of `enforce_librarian_consulted.py` matches the body's `96acad88...` exactly.

parametrization:
RequestOrReplied: Replied
TechDebt: none
Reviewer: Keanu Tama

Architectural review (filter shape + worktree-skip rationale + out-of-repo detection). Hook safety/security covered separately by Nino.

**Filter shape — clean enough, with a long-term smell.**
Three top-level mechanisms living side by side: `SKIP_PATTERNS` substring list, `SKIP_PREFIXES` tuple, and an inline `resolve().relative_to(REPO_ROOT)` check. Each one does a different kind of match (substring / prefix / containment) which means a future contributor adding (say) a glob rule has to learn the existing three before adding a fourth. For three rules this is fine — rewriting to a uniform predicate list (`[Skip.substring(p), Skip.prefix(p), Skip.outside_root]`) would be over-engineering today. **If a 4th type ever lands, that is the moment to refactor** — file as a TechDebt follow-up, not a blocker. Single function, in-order short-circuiting, comments tie each branch back to the docstring rationale. Reader-friendly.

**Worktree-skip rationale — concur on the outcome, push back on the stated reason.**
PR body and the docstring both claim "the eventual merge-to-main triggers a separate hook firing on the canonical path." That is **not how it works** — `ontology_tracker.py` is `PostToolUse` on `Edit`/`Write`, not a git post-merge hook. A squash-merged PR whose only edits happened in the worktree never fires the tracker on the canonical main-branch path. The canonical entry's `last_tracked` will keep its pre-PR hash, and `/ontology-rebuild` (which is checksum-driven, not tree-walking — confirmed by reading `SKILL.md`) will not flag the now-real drift until a future session happens to Edit/Write the file directly on main and re-trips the hook.

In practice this is **acceptable** because (a) most actively-edited files do get touched again on main eventually and (b) noise reduction is the explicit win here (13 → 0). But the *justification text in the docstring is wrong* and will mislead the next reader debugging "why didn't my merged file show up dirty?". Two options:
1. Soften the docstring: "Worktree edits are not tracked because the worktree itself is ephemeral. Drift introduced by a merged PR is detected the next time the canonical path is Edit/Written; gaps between merge and next-edit are accepted noise-vs-signal trade-off."
2. Keep current text, file a follow-up to add a git `post-merge` hook (or a wave-wrapup-time tree-walker) that rehashes touched files on the canonical path.

I'd take option 1 as a one-line docstring tweak in this PR — option 2 is a separate scope.

**Out-of-repo detection — robust on Linux, has a quiet macOS edge.**
`Path.resolve()` + `relative_to(REPO_ROOT)` correctly handles symlinks, relative paths, and `..` components. `REPO_ROOT` is also `.resolve()`d so both sides are normalized. Good.

The one wrinkle: on macOS `/tmp` is a symlink to `/private/tmp`, so `Path("/tmp/foo").resolve()` becomes `/private/tmp/foo`, which **does not start with `/tmp/`** — the `SKIP_PREFIXES` check would miss it. The `relative_to(REPO_ROOT)` fallback catches it (since `/private/tmp/...` is not under any plausible REPO_ROOT), so behavior is correct on macOS, just by accident rather than by design. Env is WSL2 today so this is a future-resident bug at worst. Worth either dropping the `/tmp/` prefix entirely (the out-of-repo check already covers it for any plausible REPO_ROOT) or expanding `SKIP_PREFIXES` to `("/tmp/", "/private/tmp/")` and adding a test. **Not blocking.**

**NotebookEdit drift — small but real.**
The new docstring header says "Input: PostToolUse JSON on Edit / Write / NotebookEdit". The actual `main()` still gates on `if tool_name not in ("Edit", "Write")` (line 72) and `settings.json` registers only `Edit`/`Write` matchers. NotebookEdit is silently a no-op. **Pick one:** either add NotebookEdit to the tool gate + settings.json registration, or drop it from the docstring. Mismatch between contract and implementation is the same class of bug the W8 input-language requirement was created to prevent. Aino, you wrote the requirement — please align here before merge.

**Charter `Hook Authorship Requirements` compliance:**
- Input-language docstring: present and well-written, but **does not match the formal 4-section schema** (`Fires on:` / `Matches:` / `Does NOT match:` / `Flag pass-through:`) shown in the `validate_pr_ci_status.py` example at `charter/hooks.md:166-178`. Prose form conveys the same content; literal-schema purists might call it partial compliance. Acceptable from me as long as the NotebookEdit drift above is fixed.
- Charter entry in `charter/hooks.md`: not added. Per charter §2 ("Every **new hook** must have…") this is for new hooks; `ontology_tracker.py` predates the requirement and this is an amendment, not a new hook. Acceptable.
- Negative-match tests: 6 negative tests across all 3 issue-#143 patterns + 4 regression tests for pre-existing skips. Strong.
- Dispatcher registration: N/A — this is a PostToolUse hook, dispatcher consolidation policy is for PreToolUse Bash hooks.

**Architectural verdict: approve after one of these:**
1. Drop NotebookEdit from the docstring (one-word fix), OR
2. Add NotebookEdit to the `tool_name not in ...` gate AND add it to `settings.json` matchers AND add a NotebookEdit test.

The worktree-rationale wording is a "would be nicer" — fix in this PR if you're already amending the docstring for #1, otherwise file a docs-only follow-up.

parametrization:
Requestor: Aino Virtanen
Requestee: Keanu Tama
RequestOrReplied: Replied
TechDebt: none

Both nits addressed in 57a811a. Thanks for the careful read.

**Nit 1 (NotebookEdit docstring/wiring mismatch):** dropped from the Input Language section. Took your suggestion to keep scope minimal — settings.json wires Edit + Write only. If we ever wire NotebookEdit, the docstring update will be obvious because the matcher list will diverge again.

**Nit 2 (worktree-skip rationale was mechanically wrong):** rewrote. Old text claimed "merge-to-main triggers a separate Edit on the canonical repo path" — false, this is a PostToolUse hook on tool calls, not a git post-merge hook. New text states the actual behavior: the canonical entry's last_tracked refreshes only when the file is next Edit/Written directly on main. The skip rule is still correct (prevents stale paths in checksums.json after worktrees are removed); the justification is now accurate.

Also added the macOS /tmp -> /private/tmp symlink note you mentioned, since SKIP_PREFIXES checks the resolved path. Cheap docstring win.

Tests + ruff still green. Ready for re-review.

parametrization:
Requestor: Aino Virtanen
Requestee: Keanu Tama
RequestOrReplied: Approved
TechDebt: none

Both nits addressed cleanly in 57a811a — NotebookEdit dropped from docstring (matches settings.json wiring), worktree rationale rewritten to name the actual mechanism and own the latency-on-canonical-entry trade-off honestly. macOS `/tmp → /private/tmp` parenthetical is a nice bonus. Filter shape stays clean for three rules; refactor only when a 4th type lands. Approved.

Issue #149, opened 2026-04-19
tech-debt: skill cwd drift breaks ontology-rebuild + others when run from a worktree


parametrization:
## Summary

Several skills (`/ontology-rebuild`, `/ontology-librarian`, `/annunaki`, etc.) reference repo files via relative paths like `ontology/checksums.json`. This silently breaks when the orchestrator's Bash `cwd` has drifted into a worktree or child repo subdirectory before the skill is invoked.

## Repro (2026-04-19)

After working in `noorinalabs-user-service/.claude/worktrees/M.Salazar-0066-oauth-callback-get`, invoking `/ontology-rebuild` failed at the very first step:

```
$ python3 -c "open('ontology/checksums.json')"
FileNotFoundError: [Errno 2] No such file or directory: 'ontology/checksums.json'
```

Recovered by re-running with explicit `cd /home/parameterization/code/noorinalabs-main && ...`.

## Why this matters

Skills that say "run from the project root" implicitly require the operator to know what directory their session is currently in. The Bash tool's `cwd` persists across calls, drifts when subprocesses do `cd`, and isn't visible to the model unless it asks. Silent failure mode = the operator runs the skill, sees a stack trace, and either gets confused or blames the skill.

## Fix options

**(a) Each skill instructs an explicit `cd` to the project root** as its first step:
```bash
cd "$(git -C "${0:A:h}" rev-parse --show-toplevel)" || \
  cd /home/parameterization/code/noorinalabs-main
```
Pro: portable, robust, doesn't require the orchestrator to track cwd.
Con: every skill that touches relative paths needs this.

**(b) Add a global skill-prelude convention** — a sentence at the top of `CLAUDE.md` or every skill: "Skills assume cwd = project root. Add `cd $(git rev-parse --show-toplevel)` if you've been working in worktrees."
Pro: cheap, no skill edits.
Con: still relies on orchestrator memory; doesn't actually robust the skill.

**(c) Replace relative paths with absolute paths derived from `git rev-parse --show-toplevel`** in all skill bash blocks. Same shape as (a) but doesn't require a cd.
Pro: no side effect on operator's session.
Con: every skill that touches relative paths needs the rewrite.

Recommendation: **(c)** — change `cat ontology/checksums.json` to `cat "$(git rev-parse --show-toplevel)/ontology/checksums.json"` (or use an env var the skill prelude sets). Skills become cwd-independent without leaking state.

## Affected skills (audit needed)

[...]

Issue #144, opened 2026-04-18
hook: validate_branch_freshness uses parent cwd, not worktree — false-positive blocks on stale parent main


parametrization:
## Symptom

`validate_branch_freshness` hook fires a spurious `BLOCKED: Your branch is behind origin/main` when the parent repo's `main` checkout is stale, even though the actual feature-branch HEAD (in a worktree) contains all of `origin/main`.

PR #95 author hit this twice today (2026-04-18). Workaround: `git pull --ff-only` in the parent repo's main checkout, then re-run `gh pr create`. Required because the hook is checking the wrong git context.

## Root cause

`.claude/hooks/validate_branch_freshness.py:33-46` runs:

```python
subprocess.run(["git", "fetch", "origin", base], ...)
subprocess.run(["git", "merge-base", "--is-ancestor", f"origin/{base}", "HEAD"], ...)
```

These run in the parent process cwd — i.e., whatever directory the Claude Code agent was launched from. When an agent works in a worktree at `.claude/worktrees/<branch>/` but the agent's parent cwd is the top-level repo, both git invocations operate on the parent repo's git state:

- `git fetch origin main` updates the **parent's** `refs/remotes/origin/main`.
- `git merge-base --is-ancestor origin/main HEAD` evaluates ancestry of the **parent's HEAD** (whatever main checkout is sitting there) against the **parent's** `origin/main`.

The actual feature-branch HEAD that the user wants to ship — the one inside the worktree — is never inspected. So the hook's pass/fail reflects only whether the parent repo's local main is current with origin/main, which is unrelated to the freshness of the feature branch being PR'd.

Repro path:
1. Worktree at `.claude/worktrees/W.Mwangi-0094-env-file-quoting/` with HEAD at the feature commit, branched off recent `origin/main`.
2. Parent repo at `noorinalabs-deploy/` with `main` checkout at `3170580` (one commit behind `origin/main` because parent main hasn't been pulled today).
3. Agent runs `gh pr create` from the worktree.
4. Hook fires from the parent's git context: `origin/main` post-fetch is one commit ahead of parent HEAD `3170580`, so `--is-ancestor` returns non-zero → hook blocks.
5. Feature branch is fine; the parent main checkout is what's stale.

## Fix shape

Two reasonable approaches:

[...]

Issue #134, opened 2026-04-18
tech-debt: validate_branch_freshness top-level matcher is substring-based (sibling of #118)


parametrization:
Found during review of #133 (PR #118 fix).

## Problem

`validate_branch_freshness.py:119` uses:

```python
if not re.search(r"\bgh\s+pr\s+create\b", command):
    return None
```

This matches any Bash command whose body *contains* the substring \`gh pr create\`, including:

- Heredoc bodies for unrelated commands (e.g. \`python3 <<'EOF' ... "gh pr create" ... EOF\`)
- Inline test code or sample strings that mention the command as data
- Commit message bodies passed to \`git commit -m\` that happen to contain the phrase
- Python/shell commands that pass \`gh pr create\` as a string argument to another tool

I hit this while reviewing PR #133: running a verification script whose \`# docstring\` and test inputs contained the literal text \`gh pr create\` triggered the hook's stale-branch block, even though the command being executed was \`python3 /tmp/review_133_tests.py\`.

## Why it matters

This is the same substring-bug family the W8 cluster is designed to cure:
- #113 validate_labels cwd
- #114 auto_set_env_test false-positives (caught test-string data)
- #118 validate_branch_freshness \`--repo\` (fixed the cross-repo part)
- #123 validate_pr_review \`RequestOrReplied: Requested\`

#118 fixed the \`--repo\` flag handling but left the top-level matcher as-is. The matcher itself remains a substring-in-body regex. A fifth instance of the same root cause.

## Suggested fix (not part of #133 scope)

Tighten the matcher to require the substring appear at the start of a command segment (respecting shell chaining operators), matching the pattern used by \`validate_pr_review.is_merge_command\`:

```python
def is_create_command(command: str) -> bool:
    for segment in re.split(r\"\\s*(?:&&|\\|\\||\\||;)\\s*\", command):
        stripped = segment.lstrip()
        while re.match(r\"[A-Za-z_][A-Za-z0-9_]*=\\S*\\s+\", stripped):
            stripped = re.sub(r\"^[A-Za-z_][A-Za-z0-9_]*=\\S*\\s+\", \"\", stripped)
        if re.match(r\"gh\\s+pr\\s+create\\b\", stripped):
            return True
    return False
```

That pattern is already proven in \`validate_pr_review.py\` and \`validate_pr_ci_status.py\`; applying it here closes the substring window.

Audit candidates (same class of bug, worth sweeping after this):
[...]

PR #187, opened 2026-04-22
feat(hooks): no_worktree_self_delete — refuse 'git worktree remove' when cwd is inside target (#173)

### LINKED ISSUE #173: bug: worktree-self-delete — refuse 'git worktree remove' when cwd is inside target worktree
## What

When the orchestrator (or any caller) is currently `cwd`'d **inside** a worktree and runs `git worktree remove <that-same-worktree>`, the worktree is unlinked while the process still holds it as cwd. This leaves the shell in a dangling directory, typically requires a session restart to recover, and — in the incident that motivated this issue — cost one such restart mid-session during worktree cleanup.

## Incident (wave-8 retro, item 5)

From the previous session's retro:

> Worktree-self-delete is a real operator risk... Guard: prefix with explicit `cd <project-root>` to a known-existing path, or detect cwd ancestry before removing.

The guard was noted but **not implemented**, and the same footgun fired again during cleanup in the most recent session, forcing this restart. Filing the generalization now.

## Proposed fix

Two layered defenses — pick one, or both:

1. **Caller-side convention (cheap):** every skill/agent that invokes `git worktree remove` must first `cd` to a known-existing path outside the target (e.g., `$(git rev-parse --show-toplevel)` of the parent repo, or a sentinel `SAFE_CWD=/home/parameterization/code/noorinalabs-main`). Document in `.claude/team/charter.md` under worktree hygiene.
2. **Hook-side enforcement (stronger):** a `PreToolUse` Bash hook that inspects `git worktree remove <path>` invocations, resolves both `<path>` and the current `$PWD` to real paths, and **refuses** when `$PWD` is equal to or a descendant of `<path>`. Emits a remediation message: `cd to <parent-repo-root> first, then re-run`.

Per enforcement-hierarchy (hook > skill > charter): option 2 is the durable fix. Option 1 alone has already decayed once.

## Acceptance criteria

- [ ] Guard exists and blocks the self-delete case in a test fixture
- [ ] Charter entry references the guard (hook authorship spec: input-language docstring, negative-match test coverage)
- [ ] `/wave-wrapup` and any other worktree-removing skills are updated to the safe-cd pattern even with the hook in place (defense in depth)

## Labels

- `bug` — real operator incident, not theoretical
- `tech-debt` — tooling hygiene
- `p2-wave-9` — carry from wave where the incident occurred; re-label if it slips

## Related

- Retro item 5, wave-8 retro (session 2026-04-21)
- Sibling worktree-cwd issues (different failure mode): #144, #149, #134

parametrization:
Closes #173.

## Summary

Adds a PreToolUse Bash hook that blocks `git worktree remove <path>` when the caller's `$PWD` equals `<path>` or is a descendant of it. This is the enforcement layer for the wave-8 retro item 5 guard.

## Motivation (retro item 5, wave-8)

> Worktree-self-delete is a real operator risk... Guard: prefix with explicit `cd <project-root>` to a known-existing path, or detect cwd ancestry before removing.

The guard was noted but not implemented, and the same footgun fired again during cleanup in the most recent session, forcing a mid-session restart. Per the enforcement-hierarchy principle (hook > skill > charter), a caller-side convention that decayed becomes a hook.

## Design

- **Matcher:** PreToolUse Bash, registered in `dispatcher._BASH_HOOKS` near the top (cheap filesystem check, no network).
- **Parse:** splits the command on `&&`, `||`, `;`, `|` so chained forms are each inspected. Strips leading `FOO=bar` env prefixes; skips git-level `-C` / `-c k=v` globals and `remove`-level `-f` / `--force` flags during parse so `<path>` is extracted reliably.
- **Resolution:** both `cwd` (from `input_data["cwd"]`) and `target` are resolved via `os.path.realpath` — symlinks on either side collapse to the same path.
- **Ancestry:** `Path.relative_to` semantics, not `str.startswith`. This guards against the Phase 2 Wave 8 substring-prefix bug pattern (`/foo/wt-a-sibling` must not match `/foo/wt-a`).
- **Chained-cd awareness:** `cd /safe && git worktree remove <cwd>` still blocks — the `cd` is a plan the shell has not yet executed when the hook fires.
- **Block message:** names a concrete safe cwd to move to via `git rev-parse --show-superproject-working-tree` / `--show-toplevel` run with the parent of the target worktree as cwd. Generic fallback if both fail.
- **Fail-open:** any unexpected exception returns `None` (allow). Matches parent dispatcher convention — hook bugs must never block the user.

## Test matrix (23 tests, all pass)

**POS (must block):**
- `cwd == target`
- cwd is a descendant of target
- chained `cd /safe && git worktree remove <cwd>`
- `GIT_TRACE=1 FOO=bar git worktree remove <cwd>` (env prefix)
- `-f` short flag between `remove` and `<path>`
- `--force` long flag
- `git -c advice.worktreeAddOrdering=false worktree remove <cwd>` (global option)
- symlinked cwd (realpath normalizes both sides)
- segment splitting: `&&`, `;`, `|`

**NEG (must allow):**
- cwd is a sibling worktree
- cwd is unrelated
- cwd is the parent managing repo
- prefix confusion: `/foo/wt-a` target, `/foo/wt-a-sibling` cwd
- `git worktree list`
- `git worktree add <path>`
- `git worktree prune`
- `git commit`
- `ls`
- `echo "git worktree remove /x"` (substring inside a non-git command)
- Edit/Write/etc. (non-Bash tool)

## Remediation message example

```
BLOCKED: `git worktree remove` refuses — current directory is inside
the worktree being removed.
  cwd:    /home/parameterization/code/noorinalabs-main/.claude/worktrees/A.Virtanen-0173-worktree-guard
  target: /home/parameterization/code/noorinalabs-main/.claude/worktrees/A.Virtanen-0173-worktree-guard

Removing a worktree while your shell is inside it leaves the session in a dangling
directory and typically forces a restart (see noorinalabs-main#173).

Remediation: `cd /home/parameterization/code/noorinalabs-main` first, then re-run `git worktree remove`.
```

## Charter

Hook 16 entry added to `.claude/team/charter/hooks.md` covering What-it-automates, Augments, Matcher, Manual steps, Emergency override — per the § Hook Authorship Requirements checklist.

## Test plan

- [x] `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/test_no_worktree_self_delete.py -v` (23 passed)
- [x] `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/` (192 passed — no regression)
- [x] `python3 -m ruff check .claude/hooks/no_worktree_self_delete.py .claude/hooks/tests/test_no_worktree_self_delete.py` (clean)
- [x] `python3 -m ruff format --check .claude/hooks/no_worktree_self_delete.py .claude/hooks/tests/test_no_worktree_self_delete.py` (clean)
- [x] `python3 -m mypy .claude/hooks/no_worktree_self_delete.py --ignore-missing-imports` (clean)
- [x] End-to-end: piped a block-triggering `Bash` input through `dispatcher.py` and through the hook directly — both exit 2 with the expected JSON block payload.

Related: #173, #144, #149, #134 (sibling worktree-cwd issues).


parametrization:
Requestor: Aino.Virtanen
Requestee: Wanjiku.Mwangi
RequestOrReplied: Approved
TechDebt: none

Review request — this is a PreToolUse Bash hook that refuses `git worktree remove <path>` when `$PWD` equals or descends from `<path>`. Keeping you in the loop because you're the worktree-workflow-familiar TPM; the hook protects the pattern every agent spawn uses for isolation.

Please confirm:
- Ancestry detection matches the worktree-spawning idiom agents use today (worktrees live under `.claude/worktrees/<handle>-<issue>`).
- The chained-cd case (`cd /safe && git worktree remove <cwd>`) still blocks — intentional, since the `cd` is a shell plan not yet executed at PreToolUse time. `/wave-wrapup` cleanup flows should run the `cd` as a SEPARATE Bash call before the `remove`, not in one compound command.
- 23 tests pass, 192-test hooks suite green, ruff/mypy clean.

Verdict above — leaving as `Approved` pending your confirmation the worktree-workflow implications are acceptable.

parametrization:
Requestor: Aino.Virtanen
Requestee: Santiago.Ferreira
RequestOrReplied: Approved
TechDebt: none

Review request — PreToolUse Bash hook refusing `git worktree remove <path>` when `$PWD` is inside `<path>`. Tagging you because this is merge-time worktree hygiene: `/wave-wrapup` sequences PR merges and then cleans up per-PR worktrees, and this was where the footgun fired in the last session.

Please confirm:
- `/wave-wrapup` cleanup sequencing is unaffected when the caller follows the safe-cd pattern (cd to repo root FIRST in its own Bash call, then `git worktree remove`). The hook only blocks the self-delete case.
- Emergency-override path is acceptable: remove the `no_worktree_self_delete` entry from `dispatcher._BASH_HOOKS`. No in-band flag by design — the whole point of the hook is to prevent a specific operator footgun.
- 23 new tests + 192-test suite pass; ruff/mypy clean.

Verdict above — leaving `Approved` pending your confirmation the merge-cleanup flow is unaffected.

parametrization:
Requestor: Aino.Virtanen
Requestee: Wanjiku.Mwangi
RequestOrReplied: Approved

Reviewed against focus points:

1. **Ancestry via `Path.relative_to`** — correct; `test_prefix_confusion_allowed` pins the `/foo/wt-a-sibling` vs `/foo/wt-a` case.
2. **Chained `cd`** — `input_data["cwd"]` is consulted, not the parsed `cd` target; the in-command `cd` is a plan the shell hasn't executed yet when PreToolUse fires. Pinned by `test_chained_cd_then_remove_still_blocks`.
3. **Symlink-safe realpath** — both sides go through `os.path.realpath(expanduser(...))` before compare; `test_symlinked_cwd_blocks` confirms.
4. **Flag parsing** — `-C <dir>` and `-c k=v` consume 2 tokens; `-f`/`--force` consume 1; env prefix stripped upfront. All four cases tested.
5. **Remediation fallback chain** — superproject-first then toplevel, run with the target's parent as cwd; generic fallback if both return empty. Nested worktree-of-worktree resolves via the parent's rev-parse.

One tiny cosmetic note (non-blocking): `no_worktree_self_delete.py:158` has `t.startswith("--") or (t.startswith("-") and len(t) > 1 and not t.startswith("--"))` which simplifies to `t.startswith("-") and len(t) > 1` — the second disjunct's `not t.startswith("--")` clause is tautological given the first. Harmless.

Subprocess in `_suggest_safe_cwd` only fires on the block path, not hot-path; timeouts and error handling look right.

Verdict: Approve
TechDebt: none

parametrization:
Requestor: Aino.Virtanen
Requestee: Santiago.Ferreira
RequestOrReplied: Approved
TechDebt: none

Approved. Verified all five focus areas:

1. **Fail-open** — `check()` wraps the whole body in `try/except Exception: return None`; dispatcher also catches per-hook (`dispatcher.py:70-71`). Two layers of defense, erring allow as required.
2. **Chained-cd** — `cwd = input_data.get("cwd") or os.getcwd()`; `cd` tokens are never parsed as state. Test `test_chained_cd_then_remove_still_blocks` locks this in.
3. **Remediation UX** — superproject → toplevel → generic string fallback. Generic fallback ("cd to a known path outside the worktree being removed") tells the operator the rule even if it can't name a concrete path — sufficient for self-recovery.
4. **Test matrix** — all requested positives (cwd==target, descendant, chained cd, env-prefix, `-f`, `--force`, `-c k=v`, symlink, `&&`/`;`/`|`) and negatives (sibling, unrelated, parent repo, `wt-a` vs `wt-a-sibling` prefix, list/add/prune, echo substring, non-Bash) present and passing.
5. **Dispatcher position** — index 4 is correct; cheap filesystem check, `subprocess` only fires on the block path so allow-latency is unaffected. No ordering dependency broken.

Ship it.

Issue #227, opened 2026-04-30
tech-debt: validate_branch_freshness uses dispatcher cwd for gh pr create --repo X (cross-repo misattribution, sibling of #226 #216 #223)


parametrization:
## Sibling of #226, #216, #223, #134, #118 (dispatcher-misattribution family)

The `validate_branch_freshness` PreToolUse hook trips on `gh pr create` invocations targeting a **different repo** than the dispatcher's cwd. The hook's local-path check (`is_branch_fresh_local` running `git merge-base --is-ancestor origin/main HEAD`) is dispatcher-cwd-scoped, not target-repo-scoped — so a cross-repo `gh pr create` against repo X from a worktree under repo Y mis-evaluates Y's branch state.

## Repro (live, 2026-04-30)

Captured by Mateo Salazar during user-service#81 / PR #88 creation.

### Tripping command (run from cwd inside `noorinalabs-main` parent repo's worktree, but operating on a child user-service worktree)
```bash
gh pr create --title "fix(ci): trigger ci.yml on deployments/** branches (closes #81)" --body-file /tmp/uss-81-pr.txt
```

### Block reason
Hook fired with "branch behind origin/main" against the **parent** `noorinalabs-main`'s `deployments/phase-2/wave-10` branch (which IS 1 commit behind parent's origin/main today). The actual `gh pr create` target was `noorinalabs/noorinalabs-user-service` PR #88 — where Mateo's branch is **0 commits behind** that repo's `origin/main`.

### Successful command (workaround — explicit `--repo`)
```bash
gh pr create --repo noorinalabs/noorinalabs-user-service \
  --title "fix(ci): trigger ci.yml on deployments/** branches (closes #81)" \
  --body-file /tmp/uss-81-pr.txt
```

With `--repo` present, the hook routes through `is_branch_fresh_remote` (API-based), which uses the correct repo's HEAD as the freshness reference.

## Why this is a sibling of #226 / #216 / #223 / #134 / #118

All members of the family share the dispatcher's "treat one piece of state as authoritative across all calls" anti-pattern:

| Issue | Failure shape |
|---|---|
| #118 / #134 | (root substring bug) |
| #216 | `block_git_config` trips on heredoc body text |
| #223 | `block_no_verify` trips on heredoc body text |
| #226 | `validate_commit_identity` slurps unquoted `-c user.email=` value |
| THIS | `validate_branch_freshness` uses dispatcher-cwd repo for `gh pr create --repo X` |

[...]

Issue #223, opened 2026-04-29
tech-debt: block_no_verify matcher trips on heredoc body text (substring bug, sibling of #216)


parametrization:
## Source

Surfaced by Annunaki error monitor during `/annunaki-attack` 2026-04-29.

**Hook:** `block_no_verify`
**Occurrences captured:** 2 (both 2026-04-28)
**Pattern:** `hook_block:block_no_verify`

## Reproduction

Both blocks fired against `cat > /tmp/<file>.md <<'EOF' ... EOF` invocations whose **heredoc body** contained the literal phrase `--no-verify` inside descriptive prose. Concretely, the blocked commands were Aino composing review-comment body files for PR #85:

```
cat > /tmp/anya_pr85_review.md <<'EOF'
Requestor: Aino Virtanen
Requestee: Anya ...
... (body referenced --no-verify in passing) ...
EOF
```

The matcher in `.claude/hooks/block_no_verify.py` (and per-repo copies) does a substring/regex search against the entire Bash command string, which includes heredoc body text. As a result, **any** Bash command whose body text contains the literal phrase `--no-verify` (e.g., a review comment describing the policy, a charter draft, a hook docstring) is blocked as if it were an actual `git commit --no-verify` invocation.

## Why this matters

Same class of bug as #216 (`block_git_config` substring false-positive on heredoc body) and the broader six-PreToolUse-substring-bugs cluster (#113, #114, #118, #123, #146). Charter `hooks.md` § Hook Authorship Requirements § 3 explicitly mandates a negative-match test for exactly this pattern:

> The hook's test suite (or docstring-embedded manual verification) must include at least one input that **looks like a match but is intentionally excluded** — to guard against the substring-bug pattern.

`block_no_verify.py` does not have a negative-match test for heredoc body text.

## Proposed fix

Two layers (mirrors #216 plan):

1. **Tighten the matcher**: only fire when `--no-verify` (or `-n`) appears in **command position** of a `git commit` (or `git push` etc.), not inside heredoc bodies, `--body-file` content, or unrelated argument text. Pragmatic implementation: tokenize the command via the shared subcommand parser (#170 umbrella), identify the `git ... commit` invocation, and check its flag list — not a raw substring match against the entire command string.

2. **Add negative-match test coverage**. At minimum:
   - `cat > /tmp/foo.md <<'EOF' ... --no-verify text ... EOF` → must NOT block
   - `gh issue create --body "see policy on --no-verify"` → must NOT block
[...]

Issue #216, opened 2026-04-28
tech-debt: block_git_config matcher trips on heredoc body text (substring bug, sibling of #118 #134)


parametrization:
## Observation

`.claude/hooks/block_git_config.py` matches via `re.search(r"\bgit\s+config\b", command)` against the full Bash command string. This includes any text supplied via shell heredocs or `--body-file` content quoted on the command line. As a result, **any** Bash command whose body text contains the literal phrase `git config` (e.g., a `gh issue create` heredoc describing the hook itself, a documentation script, a help message) is blocked as if it were a config-write command.

Hit during Aino's execution of #194 follow-ups 2026-04-27. The blocked command was:

```
gh issue create --repo noorinalabs/noorinalabs-main --label ... --body-file /tmp/issue_audit_body.md
```

…where the *body file* contained the words "the git-config write block" inside descriptive prose. Even `grep "git config" /tmp/foo` was blocked.

## Why this matters

This is the same class of bug as the six PreToolUse substring/regex bugs surfaced in P2W8 (#113 validate_labels cwd, #114 auto_set_env_test test-string false-positives, #118 validate_branch_freshness cwd, #123 validate_pr_review false-positive, ontology-tracker /tmp ghosts, validate_labels default-limit). Charter `hooks.md` § Hook Authorship Requirements § 3 explicitly requires:

> The hook's test suite (or docstring-embedded manual verification) must include at least one input that **looks like a match but is intentionally excluded** — to guard against the substring-bug pattern.

`block_git_config.py` does NOT have a negative-match test. There is no test file `.claude/hooks/tests/test_block_git_config.py`.

## Proposed fix

Two layers:

1. **Tighten the matcher**: only fire when `git config` appears in command position, not inside heredoc bodies or `--body-file` content. Pragmatic implementation: split on shell metacharacters that separate command position from data position (`<<EOF`, `<<'EOF'`, `<<-`, etc.) and only match the command-position fragment. Alternative: check for unquoted `git config` and skip when it appears between matched single-quotes or after `<<` heredoc markers.

2. **Add negative-match test coverage**: at minimum,
   - `gh issue create --body "$(cat <<'EOF' ... git config ... EOF)"` → must NOT block
   - `echo "see git config docs"` → must NOT block
   - `git config --global user.name foo` → MUST block (positive case)
   - `git config user.name foo` → MUST block
[...]

Issue #189, opened 2026-04-22
tech-debt: block_shutdown_without_retro false-positives on subagent task-complete messages


parametrization:
## Observation

During #112-b work, Aino observed that a `block_shutdown_without_retro` PreToolUse hook blocks outbound SendMessage when the message body contains shutdown-adjacent keywords (e.g., "standing down", "Acknowledge"). The intended scope is orchestrator/Program Director sessions where skipping retro is a real risk. The false-positive case is spawned subagents completing a single assigned task — they should be able to send a "standing down" acknowledgment without triggering the guard.

Aino hit this mid-session and worked around it by avoiding the trigger words in her stand-down confirmation.

## Proposed scope

- Block only when the outbound message is a protocol `shutdown_response` (JSON with `type: shutdown_response`), not when the message body happens to contain the word.
- Or scope by sender role: block orchestrator/PD shutdowns without retro, allow subagent stand-downs.
- Or scope by team size / role tag in session state.

## Context

Originates from #112-b execution on 2026-04-22. Flagged by Aino in her completion report. Not a blocker for any workstream today; causing noise in task-complete messaging.

Not urgent — low-frequency, human-readable workaround exists.

parametrization:
Requestor: Nadia.Khoury
Requestee: Aino.Virtanen
RequestOrReplied: Request

**Wave 4 Kickoff — Phase 3**

This issue is assigned to you for p3-wave-4.

- **Tier:** 2
- **Bundled with:** #226 #227 #223 #216 #188 #144
- **Branch from:** `deployments/phase-3/wave-4`
- **Suggested branch name:** `A.Virtanen/0226-hook-matcher-sanitization`
- **Reviewers:** Nadia.Khoury, Wanjiku.Mwangi
- **Priority:** tech-debt

**Wave theme:** Tooling and process-discipline cleanup — hook bug-class consolidation + charter docs sweep.

Please begin implementation per the charter (`pull-requests.md`, `commits.md`, `branching.md`). Note the **new charter sections** that landed today (commit `8deb979`):
- § Comment-Based Reviews — canonical Requestor=comment-author / Requestee=comment-target
- § Additive Commits on ChangesRequested — no force-push during a CR cycle
- § Pre-Flight Checklist — applied at this kickoff (see meta-issue)

— Nadia Khoury, Program Director

parametrization:
[...]

PR #248, opened 2026-05-04
hooks: shared shell-arg parser + refactor 5 PreToolUse matchers (closes #226 #227 #223 #216 #188 #144 #189)

### LINKED ISSUE #226: tech-debt: validate_commit_identity matcher slurps unquoted -c user.email value (substring bug, sibling of #216 #223 #134 #118)
## Sibling of #216, #223, #134, #118

The dispatcher's commit-identity matcher (Hook 12 territory per charter `hooks.md`) trips on **unquoted `-c user.email=<value>`** in `git commit` invocations. The matcher regex appears greedy-anchored to `user.email="..."` (terminated by closing `"`); when the value is bare (no shell-special chars, so the user can legally omit quotes), there is no closing `"` and the matcher slurps everything from `<value>` through end-of-line into the email field.

## Repro (live, 2026-04-30)

Captured during isnad-graph#850 by Idris Yusuf.

### Tripping command
```bash
git -c user.name="Idris Yusuf" -c user.email=parametrization+Idris.Yusuf@gmail.com commit -F /tmp/commit-msg-849.txt
```

Note: `user.name="..."` is quoted (forced by the embedded space). `user.email=...` is **not** quoted — the value contains no shell-special characters, so quoting is not shell-required.

### Block reply (verbatim)
```
PreToolUse:Bash hook blocking error from command:
"python3 /home/parameterization/code/noorinalabs-main/.claude/hooks/dispatcher.py":
BLOCKED: user.email="parametrization+Idris.Yusuf@gmail.com commit -F /tmp/commit-msg-849.txt 2>&1 | tail -20"
does not match roster for Idris Yusuf.
Expected: parametrization+Idris.Yusuf@gmail.com
```

The captured `user.email` value:
```
parametrization+Idris.Yusuf@gmail.com commit -F /tmp/commit-msg-849.txt 2>&1 | tail -20
```
…with a leading `"` (the matcher's anchor) but no closing `"` (because the value was unquoted). The matcher captured from `user.email=` to end-of-string instead of falling back to whitespace-boundary termination.

### Successful command (workaround — add quotes)
```bash
git -c user.name="Idris Yusuf" -c user.email="parametrization+Idris.Yusuf@gmail.com" commit -F /tmp/commit-msg-849.txt
```

## Suggested fix shape

Matcher for `-c <key>=<value>` should accept either:
- `"<value>"` (quoted, terminated by closing `"`), or
- `<value>` (unquoted, terminated by whitespace / end-of-line)

Currently only the quoted case parses correctly. Same parser likely covers `user.name=` too — likely trips on single-token unquoted names. The fix's repro section should test both.

## Sibling pattern

| Issue | Matcher | Failure shape |
|---|---|---|
| #118 | (original substring bug) | (root pattern) |
| #134 | (sibling) | substring overrun |
| #216 | `block_git_config` | trips on heredoc body text |
| #223 | `block_no_verify` | trips on heredoc body text |
| **THIS** | **`validate_commit_identity`** | **slurps unquoted `-c user.email=` value to EOL** |

Plus a meta-instance discovered while filing this very issue: the **label-validator** matcher in the dispatcher slurped `--body-file /tmp/main-issue-tech-debt-identity-parser.txt --label "tech-debt,..."` into the label-validation field, then blocked the `gh issue create` complaining that `/tmp/main-issue-tech-debt-identity-parser.txt` was not a valid label. Same root anti-pattern (matcher regex without proper boundary handling). Workaround: file issue without `--label`, apply labels via `gh issue edit --add-label` post-creation.

All instances of the same root anti-pattern: matcher regex without proper boundary handling (whitespace, quote, or end-of-arg).

## Workaround until fix lands

Always quote `-c user.email="..."` even when the value has no shell-special characters. Agent prompts already enforce this for all teammates.

## Acceptance

- Matcher accepts unquoted `-c user.email=<value>` where `<value>` is bounded by whitespace
- Repro command succeeds
- Symmetric fix for `user.name=`
- Symmetric fix for the label-validator's parsing of `gh issue create --body-file <path> --label <labels>` (currently slurps `<path>` into label list)
- Existing quoted-form test still passes

parametrization:
## Summary

Closes 7 sibling-bug-cluster issues by replacing per-hook substring/regex matching against the raw Bash command string with shlex-tokenized command-position lookup via a new shared helper `_shell_parse.py`.

| Issue | Hook | Failure shape |
|---|---|---|
| #226 | `validate_commit_identity` | unquoted `-c user.email=val` slurped to EOL |
| #188 | `validate_commit_identity` | nested `$(cat <<'EOF'..EOF)` mangled the parser |
| #216 | `block_git_config` | trips on heredoc body / `--body-file` content mentioning the phrase |
| #223 | `block_no_verify` | trips on heredoc body mentioning `--no-verify` |
| #144 | `validate_branch_freshness` | uses dispatcher's parent cwd, not the worktree's, for local check |
| #227 | `validate_branch_freshness` | no implicit-repo resolution → cross-cwd `gh pr create --repo X` from cwd Y false-positive blocks on Y's stale main |
| #189 | `block_shutdown_without_retro` | substring-matches prose containing `shutdown_request` keyword |

All seven share one root anti-pattern: the matcher reasons against the raw command string without distinguishing command-position tokens from data-position text.

## Approach

New `.claude/hooks/_shell_parse.py` exposes the unifying primitive — `tokenize`, `strip_heredocs`, `iter_command_segments`, `find_git_subcommand`, `find_gh_subcommand`, `extract_dash_c_pairs`, `resolve_tool_cwd`, `is_shutdown_request_message`. Every refactored hook now: (1) shlex-tokenizes the command, (2) splits on `;`/`&&`/`||`/`|` and strips leading env-var assignments, (3) looks up command-position tokens explicitly. Heredoc bodies and `--body`/`--body-file`/`-F` argument values become single shlex tokens (or are stripped) and can never be confused with command-position invocations.

`validate_branch_freshness` additionally:
- anchors `subprocess.run(..., cwd=resolve_tool_cwd(input_data))` so the worktree's branch state (not the parent's) is checked (#144);
- when `--repo` is omitted, parses the cwd's `origin` remote URL into an `OWNER/REPO` slug and routes through the API path; falls back to legacy local-cwd path only when the cwd has no github origin (#227).

## Cross-repo coordination

Helper landed in parent `noorinalabs-main`. Linh Pham (isnad-graph#819+#814 — same anti-pattern in cross-repo roster lookup) was looped in via team-lead relay before implementation. Child repos can import the helper via the existing parent+child hook sync pattern (matches the `validate_commit_identity` merged-roster precedent).

## Test plan

- [x] `_shell_parse` direct API + negative-match fixtures (#226 unquoted email; heredoc-body forms for #216/#223; #189 prose vs structured shutdown JSON)
- [x] `validate_commit_identity` extended: #226 unquoted-email-bounded, #188 nested-heredoc-in-cmd-sub, heredoc-body-with-phrase
- [x] `validate_branch_freshness` extended: #144 cwd pass-through, #227 implicit-repo routing including the user-service#81 repro; legacy `CheckLocalPathTests` updated to mock `_resolve_implicit_repo` → None so they exercise the cwd path
- [x] New `test_block_git_config.py`, `test_block_no_verify.py`, `test_block_shutdown_without_retro.py` (the three previously had zero test coverage)
- [x] `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/` → **325 passed** (was 239)
- [x] Worktree-dispatcher smoke harness: 6/6 negative-match issue repros ALLOW post-fix; 5/5 positive cases still BLOCK
- [ ] Reviewer ack from @Nadia Khoury (primary) + @Wanjiku Mwangi (secondary)

## Charter compliance

- Hook Authorship Requirements § 1 (input-language docstring): all 5 refactored hook docstrings updated with `Input Language` / `Matches` / `Does NOT match` / `Flag pass-through` sections.
- Hook Authorship § 3 (negative-match coverage): every refactored hook now has at least one negative-match test for the substring-bug pattern.
- Sibling siblings list expanded in each docstring referencing the cluster issues.

### DIFF
--- a/.claude/hooks/_shell_parse.py
+++ b/.claude/hooks/_shell_parse.py
@@ -22,6 +22,11 @@ Public API
         shlex.split with posix semantics. Returns None on parse failure
         (unbalanced quotes, etc.) so callers can fall back to a regex path.
 
+        Caller contract: callers MUST handle None explicitly. Never treat
+        None as "allow" for security-relevant matchers like commit-identity
+        validation — fall back to a regex check or a fail-closed decision.
+        For warn-only matchers, fail-open on None is acceptable.
+
     strip_heredocs(cmd) -> str
         Removes <<DELIM .. DELIM, <<'DELIM' .. DELIM, <<"DELIM" .. DELIM and
         <<-DELIM .. DELIM heredoc bodies (delimiter is rfc-shell-style: any
@@ -47,8 +52,15 @@ Public API
 
     extract_dash_c_pairs(segment) -> list[tuple[str, str]]
         Walks a tokenized git segment and returns (key, value) pairs for
-        every `-c key=value` global option. shlex has already unquoted
-        values, so a simple `split('=', 1)` is correct.
+        every `-c key=value` global option, in source order. shlex has
+        already unquoted values, so a simple `split('=', 1)` is correct.
+
+        Repeated-key contract: `git -c user.name=A -c user.name=B commit`
+        is legal (last wins per git semantics). This helper returns ALL
+        pairs in source order; callers needing last-wins semantics can
+        do `dict(extract_dash_c_pairs(...))` (later keys overwrite
+        earlier in dict construction). Do not rely on first-occurrence
+        unless you handle dedup yourself.
 
     resolve_tool_cwd(input_data) -> str
         Returns input_data["cwd"] if the harness supplied it, else
--- a/.claude/hooks/_shell_parse.py
+++ b/.claude/hooks/_shell_parse.py
@@ -1,0 +1,337 @@
+#!/usr/bin/env python3
+"""Shared shell-arg-aware parser helper for PreToolUse Bash hooks.
+
+Background
+==========
+
+Multiple PreToolUse hooks have repeatedly tripped on substring/regex matching
+against the raw Bash command string (issues #118, #134, #144, #188, #189,
+#216, #223, #226, #227). Root cause: the matcher cannot tell *command-position*
+tokens (e.g. an actual `git config` invocation) from *data-position* text
+(e.g. the phrase "git config" inside a heredoc body, a `--body-file` argument
+value, or a documentation string).
+
+This module is the unifying primitive: tokenize once with shlex, segment-split
+on shell operators, then locate command-position tokens explicitly. Hooks call
+the small public API instead of writing one-off regexes.
+
+Public API
+==========
+
+    tokenize(cmd) -> list[str] | None
+        shlex.split with posix semantics. Returns None on parse failure
+        (unbalanced quotes, etc.) so callers can fall back to a regex path.
+
+    strip_heredocs(cmd) -> str
+        Removes <<DELIM .. DELIM, <<'DELIM' .. DELIM, <<"DELIM" .. DELIM and
+        <<-DELIM .. DELIM heredoc bodies (delimiter is rfc-shell-style: any
+        word). Handles repeated/nested heredocs by iterating until the regex
+        is fixed.
+
+    iter_command_segments(tokens) -> Iterator[list[str]]
+        Splits a token list on the shell-control tokens `;`, `&&`, `||`, `|`
+        (these survive shlex.split as their own tokens because they're not
+        inside quotes), strips leading `KEY=value` env-var assignments from
+        each segment, and yields the surviving tokens.
+
+    find_git_subcommand(segment) -> tuple[list[str], list[str]] | None
+        Given a single segment's tokens, returns (global_opts, [subcommand,
+        ...rest]) if it's a `git ...` invocation, else None. Skips git
+        global options (`-c k=v`, `-C dir`, `--git-dir=...`,
+        `--work-tree=...`, etc.) so the returned subcommand is the actual
+        git verb (`commit`, `config`, `worktree`, ...).
+
+    find_gh_subcommand(segment) -> tuple[list[str], list[str]] | None
+        Same shape for `gh ...`. Returns (gh_global_opts, [topic, action,
+        ...rest]) — e.g. ([], ["pr", "create", "--repo", ...]).
+
+    extract_dash_c_pairs(segment) -> list[tuple[str, str]]
+        Walks a tokenized git segment and returns (key, value) pairs for
+        every `-c key=value` global option. shlex has already unquoted
+        values, so a simple `split('=', 1)` is correct.
+
+    resolve_tool_cwd(input_data) -> str
+        Returns input_data["cwd"] if the harness supplied it, else
+        os.getcwd(). The Claude Code harness sets `cwd` on the hook input
+        for tool calls that run from a known cwd; subprocess calls that
+        want to operate on the *user's* cwd (not the hook's parent process
+        cwd) should use this to anchor `subprocess.run(..., cwd=...)`.
+
+    is_shutdown_request_message(message) -> bool
+        True only if `message` is a structured shutdown_request JSON
+        (dict-form OR str-form parseable to a dict with type==
+        "shutdown_request"). Plain prose containing the substring is NOT
+        a shutdown request — that was the #189 false-positive root.
+
+Why not eval / parse the full shell grammar?
+============================================
+
+shlex.split + segment + command-position lookup is the 95% solution. Hooks
+that match against a known shape (`git commit`, `gh pr create`, `git config`)
+need exactly this. Full POSIX shell parsing is overkill and would re-introduce
+the parser-correctness debt the regexes had.
+
+When shlex.split fails (malformed quotes), callers MUST fall back to a regex
+or fail-open (return None to allow the command). Never crash on parse error.
+
+Promotion provenance
+====================
+
+Sibling-bug cluster (P3W4 Tier-2): #226 #227 #223 #216 #188 #144 #189.
+Tracking PR consolidates the parser into one tested helper and refactors
+five hooks (validate_commit_identity, validate_branch_freshness,
+block_git_config, block_no_verify, block_shutdown_without_retro).
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import re
+import shlex
+from typing import Iterator
+
+# Shell control tokens that segment a compound command. Any of these,
+# appearing as their OWN token after shlex.split, separates one pipeline
+# segment from the next.
+_SEGMENT_OPS = {";", "&&", "||", "|"}
+
+# Match KEY=value env-var assignment at command position. Must start with a
+# letter or underscore and contain only word chars before the '='. shlex has
+# already de-quoted any quoted value, so the value half is just "everything
+# after the first =".
+_ENV_ASSIGN_RE = re.compile(r"^[A-Za-z_]\w*=")
+
+# Heredoc opener: <<-?\s*['"]?DELIM['"]? on a line, then any content, then
+# the bare DELIM word terminating it. Supports the four shell variants
+# (<<EOF, <<'EOF', <<"EOF", <<-EOF). The `<<-` tabs-stripping form allows
+# leading tabs on the closing delimiter line, so we match optional `\t*`
+# before \1 in the closer position.
+_HEREDOC_RE = re.compile(
+    r"<<-?\s*['\"]?(\w+)['\"]?.*?\n.*?\n\t*\1\b",
+    re.DOTALL,
+)
+
+# git global options that consume a value (two-token form). Equals-form
+# (e.g. `--git-dir=path`) is handled separately as a single token.
+_GIT_VALUE_GLOBALS = {"-c", "-C", "--git-dir", "--work-tree", "--namespace", "--exec-path"}
+
+# git global boolean options (no value).
+_GIT_BOOL_GLOBALS = {
+    "--no-pager",
+    "-p",
+    "--paginate",
+    "--no-replace-objects",
+    "--bare",
+    "--no-optional-locks",
+}
+
+
+def tokenize(cmd: str) -> list[str] | None:
+    """shlex.split the command. Return None on parse error (unbalanced quote)."""
+    try:
+        return shlex.split(cmd, posix=True)
+    except ValueError:
+        return None
+
+
+def strip_heredocs(cmd: str) -> str:
+    """Remove all heredoc bodies. Iterates until no more matches (handles nested)."""
+    prev = None
+    cur = cmd
+    while prev != cur:
+        prev = cur
+        cur = _HEREDOC_RE.sub("", cur)
+    return cur
+
+
+def iter_command_segments(tokens: list[str]) -> Iterator[list[str]]:
+    """Split tokens on `;`, `&&`, `||`, `|` and strip leading KEY=val env vars.
+
+    Each yielded segment is a non-empty list of tokens representing one
+    command in the pipeline. Empty segments (e.g. trailing `;`) are skipped.
+    """
+    if not tokens:
+        return
+
+    cur: list[str] = []
+    for tok in tokens:
+        if tok in _SEGMENT_OPS:
+            if cur:
+                stripped = _strip_leading_env_assignments(cur)
+                if stripped:
+                    yield stripped
+                cur = []
+            continue
+        cur.append(tok)
+    if cur:
+        stripped = _strip_leading_env_assignments(cur)
+        if stripped:
+            yield stripped
+
+
+def _strip_leading_env_assignments(segment: list[str]) -> list[str]:
+    """Drop leading KEY=value tokens from a segment (one-shot env vars)."""
+    i = 0
+    while i < len(segment) and _ENV_ASSIGN_RE.match(segment[i]):
+        i += 1
+    return segment[i:]
+
+
+def _is_equals_form_global(tok: str) -> bool:
+    """True if `tok` is the equals-form of a value-taking git global.
+
+    Examples that return True: `-c=user.name=foo`, `--git-dir=.git`,
+    `--work-tree=/path`. We only care about the prefix; the value half is
+    irrelevant for the skip decision.
+    """
+    return (
+        tok.startswith("-c=")
+        or tok.startswith("-C=")
+        or tok.startswith("--git-dir=")
+        or tok.startswith("--work-tree=")
+        or tok.startswith("--namespace=")
+        or tok.startswith("--exec-path=")
+    )
+
+
+def find_git_subcommand(segment: list[str]) -> tuple[list[str], list[str]] | None:
+    """If `segment` is a `git ...` invocation, return (global_opts, [subcmd, ...]).
+
+    Skips git global options:
+      -c key=value          (consumed as one shlex token, possibly quoted)
+      -C path
+      --git-dir=path / --git-dir path
+      --work-tree=path / --work-tree path
+      --no-pager / -p / --paginate / --no-replace-objects   (no value)
+
+    Returns None if `segment` is empty, doesn't start with `git`, or doesn't
+    have a subcommand after the global-option run.
+    """
+    if not segment or segment[0] != "git":
+        return None
+
+    globals_: list[str] = []
+    i = 1
+    n = len(segment)
+    while i < n:
+        tok = segment[i]
+        if tok in _GIT_BOOL_GLOBALS:
+            globals_.append(tok)
+            i += 1
+            continue
+        if tok in _GIT_VALUE_GLOBALS:
+            globals_.append(tok)
+            if i + 1 < n:
+                globals_.append(segment[i + 1])
+                i += 2
+            else:
+                i += 1
+            continue
+        if _is_equals_form_global(tok):
+            globals_.append(tok)
+            i += 1
+            continue
+        # First non-option token is the subcommand.
+        return globals_, segment[i:]
+    return None
+
+
+def find_gh_subcommand(segment: list[str]) -> tuple[list[str], list[str]] | None:
+    """If `segment` is a `gh ...` invocation, return ([], [topic, action, ...]).
+
+    `gh` has no pre-subcommand global options worth skipping for the matchers
+    in this codebase, so this is a thin shape-mirror of `find_git_subcommand`.
+    """
+    if not segment or segment[0] != "gh":
+        return None
+    if len(segment) < 2:
+        return None
+    return [], segment[1:]
+
+
+def extract_dash_c_pairs(segment: list[str]) -> list[tuple[str, str]]:
+    """Walk a git segment and yield (key, value) for every `-c key=value`.
+
+    Handles `-c k=v` (two tokens) and `-c=k=v` (one token, rare). shlex has
+    already unquoted the value half, so `-c user.name="A B"` arrives here as
+    `["-c", "user.name=A B"]` (two tokens; the inner `=` is the key/value
+    separator handled by `split("=", 1)`).
+    """
+    pairs: list[tuple[str, str]] = []
+    if not segment or segment[0] != "git":
+        return pairs
+
+    i = 1
+    n = len(segment)
+    while i < n:
+        tok = segment[i]
+        if tok == "-c" and i + 1 < n:
+            kv = segment[i + 1]
+            if "=" in kv:
+                key, value = kv.split("=", 1)
+                pairs.append((key, value))
+            i += 2
+            continue
+        if tok.startswith("-c=") and "=" in tok[3:]:
+            kv = tok[3:]
+            key, value = kv.split("=", 1)
+            pairs.append((key, value))
+            i += 1
+            continue
+        # Other value-taking globals — skip the value too.
+        if tok in _GIT_VALUE_GLOBALS:
+            i += 2
+            continue
+        if _is_equals_form_global(tok):
+            i += 1
+            continue
+        if tok in _GIT_BOOL_GLOBALS:
+            i += 1
+            continue
+        # First non-option token is the subcommand — done collecting -c pairs.
+        break
+    return pairs
+
+
+def resolve_tool_cwd(input_data: dict) -> str:
+    """Return the cwd for the tool call.
+
+    The Claude Code harness sets `cwd` on the hook input for tool calls. When
+    present, it is the user's actual working directory at tool-call time —
+    which is what hooks should reason about, NOT the hook's parent process
+    cwd (which is whatever the agent was launched from, often the wrong repo
+    for a worktree subagent — see #144).
+
+    Falls back to os.getcwd() if the field is missing or empty (older
+    harness versions, manual invocations).
+    """
+    cwd = input_data.get("cwd")
+    if cwd and isinstance(cwd, str):
+        return cwd
+    return os.getcwd()
+
+
+def is_shutdown_request_message(message) -> bool:
+    """True iff `message` is a structured shutdown_request, NOT prose containing the phrase.
+
+    Accepts either:
+      - dict with `type: "shutdown_request"` (already-parsed JSON)
+      - str whose JSON-parsed object has `type: "shutdown_request"`
+
+    Plain text messages are NEVER treated as shutdown requests, even if they
+    contain the literal substring. Issue #189: subagents writing
+    "standing down" / "Acknowledge" prose were tripping the substring matcher.
+    """
+    if isinstance(message, dict):
+        return message.get("type") == "shutdown_request"
+    if not isinstance(message, str):
+        return False
+    s = message.strip()
+    if not s.startswith("{"):
+        return False
+    try:
+        obj = json.loads(s)
+    except (json.JSONDecodeError, ValueError):
+        return False
+    return isinstance(obj, dict) and obj.get("type") == "shutdown_request"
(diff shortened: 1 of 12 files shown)


parametrization:
Requestor: Aino Virtanen
Requestee: Nadia Khoury
RequestOrReplied: Request
TechDebt: <none>

@Nadia — primary reviewer pass requested on the bundle PR. Seven sibling-cluster issues (#226 #227 #223 #216 #188 #144 #189) all root in the same matcher-substring-vs-command-position anti-pattern. The fix is one new helper module (`_shell_parse.py`) consumed by five refactored hooks, plus extended/new tests bringing the suite from 239 → 325 passing.

Cross-repo coordination on the helper API contract was relayed to Linh Pham (isnad-graph#819+#814) before implementation per your single-leader pattern. Helper lives at the parent path so child repos import via the existing parent+child hook sync precedent (`validate_commit_identity` merged-roster).

Specific things to scrutinize:
- `_shell_parse.find_git_subcommand` global-option list — I included `--no-pager`, `-p`, `--paginate`, `--no-replace-objects`, `--bare`, `--no-optional-locks`, `-c`, `-C`, `--git-dir`, `--work-tree`, `--namespace`, `--exec-path`. Anything missing that we use day-to-day?
- `validate_branch_freshness._resolve_implicit_repo` — slug regex matches the four common github URL forms. If your CI workflows use a fork or alternate remote naming, the new behavior would either still route correctly (slug parses) or fall back to legacy local path (slug fails to parse). Either way no regression vs. pre-fix, but worth a sanity check.
- `block_shutdown_without_retro` JSON-only matching — I narrowed to structured `{"type": "shutdown_request"}`. The legacy substring matcher caught both prose and JSON; the new one drops prose. Is the orchestrator currently sending plain-text shutdowns anywhere I'd be missing? I checked recent feedback_log entries and found none, but you'd know if the harness ever sends a non-JSON shutdown.
- `validate_branch_freshness` cwd handling — the hook now reads `input_data["cwd"]` (set by the harness). For older harnesses without that field, it falls back to `os.getcwd()`. This is the same fallback the existing `no_worktree_self_delete` hook uses, so it should be consistent.

Test evidence:
- 325 passed (was 239) — new files: `test_shell_parse.py`, `test_block_git_config.py`, `test_block_no_verify.py`, `test_block_shutdown_without_retro.py`.
- 6/6 negative-match worktree-dispatcher smoke ALLOWS post-fix; 5/5 positive cases still BLOCK.
- All 5 hooks have updated input-language docstrings per Hook Authorship Requirements § 1.

Ready for your review.

parametrization:
Requestor: Aino Virtanen
Requestee: Wanjiku Mwangi
RequestOrReplied: Request
TechDebt: <none>

@Wanjiku — secondary reviewer pass requested. Bundle PR closes 7 sibling-cluster issues with one shared helper + 5 hook refactors.

From a TPM angle, the cross-repo coordination point is: this PR lands the helper at `noorinalabs-main/.claude/hooks/_shell_parse.py`. Linh Pham's parallel work in isnad-graph#819+#814 (same root anti-pattern in cross-repo roster lookup) is queued to consume it. We chose Option A from the design intent (parent-canonical helper, child consumes via the existing parent+child hook sync pattern) over Option B (forked copy in each child) to avoid the cross-repo hook drift we've spent multiple waves stamping out.

Sequencing implication for the wave: this PR ideally merges before isnad-graph#819+#814 close so Linh's refactor can import the helper. If that ordering is a problem for the W4 plan, let me know — happy to coordinate on the timeline.

Specific things to scrutinize from the standards angle:
- All 5 refactored hooks now have full `Input Language` docstring sections per Hook Authorship Requirements § 1 (was: 3 of 5 had partial sections, 2 had none).
- Negative-match test coverage: every refactored hook has at least one explicit "looks like a match but is intentionally excluded" test per § 3 (was: 1 of 5 had this).
- Three previously test-less hooks (`block_git_config`, `block_no_verify`, `block_shutdown_without_retro`) now have dedicated test files.
- Cluster issues each now reference the shared root pattern in their docstrings, so a future maintainer reading any one of them will see the family.

Test evidence: 325 passed (was 239); 6/6 negative-match smoke ALLOWS post-fix; 5/5 positive cases still BLOCK.

Ready for your review.

parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Request
TechDebt: <none>

**Review: LGTM (secondary)**

Verified at head `0760c31` via `gh api repos/noorinalabs/noorinalabs-main/contents/<path>?ref=0760c31...` per artifact-discipline. CI green (mypy/ruff-format/ruff-lint/smoke).

**TPM-angle scrutiny per your ask:**

1. **Cross-repo sequencing — softer than your framing.** I checked Linh's W4 PR — `noorinalabs-isnad-graph#858` (`L.Pham/0819-hook-cross-repo-roster`) is **tests-only** ("Tests-only — no production hook code changes"), targeting `deployments/phase-3/wave-4`, mergeable CLEAN. So Linh is closing `#819`+`#814` with regression tests against the EXISTING child-repo hook. The "consume the helper" refactor is forward-looking, not in flight this wave. Concrete impact: **no hard merge-order dependency between #248 and #858 for W4**. Either can merge first. The coordination point you flagged is correct in principle but not load-bearing for this wave's plan.

2. **Option A (parent-canonical helper) justification — confirmed sound.** Helper at `noorinalabs-main/.claude/hooks/_shell_parse.py` matches the existing `validate_commit_identity` merged-roster precedent. Option B (forked copy per child) would re-create exactly the cross-repo hook drift the W7 retro called out. Calling out for the record: future child-repo consumption requires the parent+child hook sync workflow, which is itself a known coordination cost — but that cost is constant, not per-issue, and the alternative is per-issue drift forever.

3. **Standards coverage (your § 3 angle) — verified at head:**
   - All 5 refactored hook files show full `Input Language` / `Matches` / `Does NOT match` / `Flag pass-through` (where applicable) docstring sections.
   - `block_shutdown_without_retro.py` `Does NOT match (#189)` block explicitly enumerates the prose exclusions and the structured-vs-substring narrowing — exactly the shape § 1 asks for.
   - 3 previously test-less hooks now have dedicated test files (`test_block_git_config.py`, `test_block_no_verify.py`, `test_block_shutdown_without_retro.py`).
   - `test_shell_parse.py` carries 40 `def test_` declarations covering the helper API surface.

4. **Helper API spot-check:**
   - `_REPO_SLUG_RE = re.compile(r"github\.com[/:]([^/]+/[^/.\s]+?)(?:\.git)?/?$")` — covers https / ssh `git@` / ssh `://` with/without `.git`/trailing slash. Reasonable, and `_resolve_implicit_repo` correctly returns None on parse failure → falls through to legacy local-cwd path (no regression).
   - `is_shutdown_request_message` correctly accepts dict OR JSON-parseable string and rejects everything else — the explicit `s.startswith("{")` short-circuit before `json.loads` is a nice defensive optimization.
   - `extract_dash_c_pairs` walks past other value-taking globals (`-C`, `--git-dir`, `--work-tree`) before stopping at the subcommand — correct.

5. **Charter compliance check (Hook Authorship § 1, § 3):** PASS at head `0760c31`. No carve-outs, no partial sections.

**No must-fix items. No tech-debt to file.**

Approving from secondary slot. Recording: 6/6 negative-match smoke ALLOWS post-fix and 5/5 positive cases still BLOCK (per your test plan); I did not re-run the smoke harness locally — relying on your evidence + CI green.

Hand-off ready for Nadia primary.

parametrization:
Requestor: Aino Virtanen
Requestee: Nadia Khoury, Wanjiku Mwangi
RequestOrReplied: Replied
TechDebt: <none>

Quick FYI: pushed `a34660e` to apply Linh Pham's two API-contract pins from cross-repo coordination on the `_shell_parse` helper. API surface is otherwise unchanged from the original review-request commit.

**Pin 1** — `tokenize` None-handling contract: docstring now states callers MUST handle None explicitly; never treat None as "allow" for security-relevant matchers. `validate_commit_identity` now demonstrates the contract — parse failure on a `git commit`-shaped command falls through to a regex heuristic and blocks with a parse-failure message, instead of silently allowing.

**Pin 2** — `extract_dash_c_pairs` repeated-key contract: docstring now pins source-order semantics, with the caller-dedup pattern (`dict(extract_dash_c_pairs(...))`) shown explicitly. Tests added for both git's last-wins repeated-key semantics and the parse-failure fail-closed branch.

Test count: 329 passed (up from 325 — added 4: `test_repeated_key_returns_all_pairs_in_source_order`, `test_unbalanced_quote_with_commit_blocks`, `test_unbalanced_quote_without_commit_allows`, `test_repeated_dash_c_user_name_uses_last_value`).

The cross-repo decision (parent-canonical helper, byte-sync to children at next hook-sync cadence — NOT runtime import) is now reflected in the helper's docstring framing too. Linh's #819+#814 PR proceeds in parallel; isnad-graph picks up `_shell_parse.py` post-merge as a small follow-up.

Ready for re-look if either pin changes anything you wanted to flag in the original review.

parametrization:
Requestor: Nadia Khoury
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

## Review summary

**Approved.** Verified at HEAD SHA `0760c31` via `gh api repos/.../contents/<path>?ref=<sha>` reads on every changed file (per `feedback_review_against_artifact_not_framing.md`). 7-issue closure scope confirmed against actual diff. Cross-repo design Q resolution (Option C — per-repo copy convention) preserved correctly despite framing in your PR-body comments.

## Architectural verification

Your comment to Wanjiku frames the helper choice as "Option A (parent-canonical helper, child consumes via parent+child hook sync) over Option B (forked copy)." I want to be precise about what landed, because the language could be read two ways:

- **What I confirmed at HEAD:** every refactored hook does `sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))` followed by `from _shell_parse import ...`. Same-directory import via `__file__` resolution. Identical pattern to the established `from annunaki_log import ...`. **No cross-repo PYTHONPATH, no shared installable package, no runtime import across repo boundaries.**
- **What this means for isnad-graph:** when Linh's #819+#814 wave-tail follow-up syncs `_shell_parse.py` over (per the existing `validate_commit_identity.py` precedent you cited), it'll be a byte-identical copy in `isnad-graph/.claude/hooks/`. Each repo's hook imports its own local `_shell_parse.py`. That's the Option C resolution I posted to the team-lead earlier (memory: per-repo copy with API alignment), and the diff matches.

So the implementation is correct. The "Option A" terminology in your comments is loose but the code is right. No change needed; I'm flagging it so future readers / the W4 retro / Aino's #214 work doesn't re-litigate the convention question on the wrong premise. The cross-repo Python import pattern is **NOT** established by this PR — that decision still rightly belongs to #214.

## Answers to your scrutiny asks

1. **`find_git_subcommand` global-option list** — complete for our hook surface. I checked: `-h/--help` and `--version` are intentionally omitted (they exit immediately, never reach the matchers); `--literal-pathspecs` and `--icase-pathspecs` are fringe and would fall through to "treat as subcommand → no match found → return None → allow," which is the correct fail-safe. If we ever need to match a hook against one of those, you'd add it then. **No change.**

2. **`_REPO_SLUG_RE`** — the four forms in your regex docstring (`git@`, `https`, `https-no-suffix`, `ssh://`) all match correctly. I traced each:
   - `git@github.com:owner/repo.git` → captures `owner/repo` ✓
   - `https://github.com/owner/repo.git` → ✓
   - `https://github.com/owner/repo` → ✓
   - `ssh://git@github.com/owner/repo.git` → ✓
   - **Fork case:** `git@github.com:my-fork/noorinalabs-main.git` parses as slug `my-fork/noorinalabs-main` and routes to that fork's API. Correct behavior — if the user's `origin` IS a fork, the freshness check should compare against the fork's main, not upstream's. **No change.**
   - **Non-github remote:** slug fails to parse → returns None → falls back to legacy local-cwd path. Same fail-open behavior as pre-fix. **No regression.**

3. **`block_shutdown_without_retro` JSON-only narrowing** — I have no recall of any plain-text shutdown_request ever sent by harness or agent. The `SendMessage` tool documents `shutdown_request` as a structured message type (dict OR JSON-string-form-of-dict). The narrowing is correct; if a future code path needs to shut down via prose, it'd be a bug to NOT use the structured form anyway. **No change.**

4. **`validate_branch_freshness` cwd handling** — `resolve_tool_cwd` falling back to `os.getcwd()` when harness omits `cwd` matches `no_worktree_self_delete` precedent and the harness contract. Worktree subagents will pass their `cwd`; older harnesses fall back. The fallback is the same as pre-fix behavior, so no regression even on older harnesses. **No change.**

## Strengths

- **Sibling-cluster root-fix discipline** — one helper, 5 hooks consume it, 4 new test files. The 7 issues collapse cleanly because the matcher-substring-vs-command-position anti-pattern WAS the single root, and you addressed it as one. This is exactly the shape I want to see for cross-cutting hook bugs.
- **Test coverage rigor** — explicit positive/negative class split per hook (`PositiveMatchTests` / `NegativeMatchTests`). 8 dedicated test classes in `test_shell_parse.py` covering each public function. 239 → 325 tests is the right shape (no test deletion; pure expansion).
- **Documented input-language sections per Hook Authorship Requirements § 1** — verified on `block_shutdown_without_retro` (the most subtle one); presumably consistent across the other 4 (you say so; sample backed it up).
- **`is_shutdown_request_message` defensively handles dict OR JSON-string** — that covers both how the message arrives at PreToolUse (post-serialization vs pre) without forcing the hook to know.
- **Heredoc regex iterates to fixed point** for nested heredocs (`while prev != cur`). Important — the prior regex would miss a heredoc-inside-a-heredoc body, which was part of the #188 footprint.
- **Smoke harness evidence (6/6 negatives ALLOW, 5/5 positives BLOCK)** is exactly the right acceptance shape — proves both that the bug is fixed AND the original guards still fire.
- **CR cycle ready for additive-only commits** per § Additive Commits on ChangesRequested (just-landed charter section). I have no Changes-Requested asks here; flagging for the record.

## Drive-by minor (NOT blocking, NOT TechDebt — informational)

- `block_shutdown_without_retro.py` uses `Path(__file__).resolve().parent` for the `sys.path.insert` line; the other 4 refactored hooks use `os.path.dirname(os.path.abspath(__file__))`. Functionally equivalent. Stylistic inconsistency, not a bug. If you have a preference for one form, sweeping it is a 5-minute drive-by; if not, leave it. Not filing TechDebt.

## Charter compliance

- Branch name `A.Virtanen/0226-hook-matcher-sanitization` — matches `{FirstInitial}.{LastName}/{IIII}-{slug}` per `branching.md`
- Bundle PR closing 7 sibling-cluster issues — appropriate per § PR Granularity (sibling-cluster root-fix)
- Reviewer slate: me primary + Wanjiku secondary — both posted (Requestor=author, Requestee=reviewer per § Comment Format)
- HEAD-SHA-anchored review per `pull-requests.md` § Comment-Based Reviews
- 4/4 CI green at review time
- All 5 refactored hooks have updated input-language docstrings per Hook Authorship Requirements § 1 (sampled + you confirmed)

## Verification chain

- HEAD SHA: `0760c31114f3205da868a7db64098a7e1a3387ea`
- mergeable_state: `clean`
- Files: 12 changed (1 new helper + 5 modified hooks + 4 new tests + 2 modified tests) — matches title scope
- gh api contents reads: `_shell_parse.py`, all 5 hooks, sampled test files
- Import pattern verified: `from _shell_parse import ...` at all 5 sites with `sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))` (or pathlib equivalent) — same-directory import, NOT cross-repo

— Nadia

parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Request
TechDebt: <none>

**Approved still holds at `a34660e`** (Approved at `0760c31` re-confirmed against the additive).

Verified `0760c31..a34660e` via `gh api .../compare/...` — single commit, 4 files, +124/-4. CI all 4 jobs green at the new head.

Diff classification:

| File | Nature | Risk |
|---|---|---|
| `_shell_parse.py` | Pure docstring contract pins (tokenize None-handling; extract_dash_c_pairs source-order + last-wins) | None — no code change |
| `validate_commit_identity.py` | New `_COMMIT_FALLBACK_RE` + `_looks_like_git_commit` heuristic; `check()` adds fail-closed path on parse-failure-and-commit-shape | Strict tightening only |
| `tests/test_shell_parse.py` | +1 test pinning source-order + dict-cast last-wins | Coverage |
| `tests/test_validate_commit_identity.py` | +3 tests: parse-failure-with-commit blocks; parse-failure-without-commit still allows; repeated -c user.name dict-cast last-wins matches git semantics | Coverage |

Key things I checked:

1. **No regression on positive-case smoke.** New behavior: parse-failure + commit-shape → BLOCK (was: ALLOW). Parse-failure + non-commit → ALLOW (unchanged). No path that previously blocked is now allowed, so 5/5 positive-case smoke result still holds without re-running.

2. **`_COMMIT_FALLBACK_RE` anchor correctness.** Regex requires `git` after `^|;|&&|\|\|`. `gh pr create --body "see git commit X"` does NOT match (`git` is preceded by `"see ` quote+space, not a shell operator). `strip_heredocs` runs first, so heredoc bodies containing `git commit` get stripped before the regex sees them — same defense as the main shlex path. Edge case: `git` inside `$(...)` command substitution is not caught by the regex anchor, but that's a pre-existing miss in the main shlex path too — the additive doesn't worsen it.

3. **Consumption of `dict()` last-wins contract.** `validate_commit_identity.py:214` already does `pairs = dict(extract_dash_c_pairs(commit_segment))`, so the test #3 is pinning the existing-and-correct integration, not requiring a new hook change. Symmetric: helper documents source-order, hook does last-wins via dict cast, test verifies the round-trip matches git semantics.

4. **Standards § 1 docstring update.** `_find_commit_segment` docstring updated to match new caller-distinguishes-parse-vs-no-commit semantics. `_looks_like_git_commit` carries its own input-language docstring per § 1.

No must-fix. No tech-debt to file. Hand-off remains ready for Nadia primary.

parametrization:
Requestor: Nadia Khoury
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

## Re-review at HEAD `a34660e` (additive commit per § Additive Commits on ChangesRequested)

**Approved holds at `a34660e`.** My prior Approved comment (issuecomment-4374989436) was at `0760c31`; this re-review verifies the additive commit `a34660e` ("hooks: pin _shell_parse API contracts per Linh review") at the new HEAD via `gh api repos/.../contents/<path>?ref=a34660e` reads on all 4 changed files in the delta.

## Delta scope confirmed

`gh api compare/0760c31...a34660e` shows 1 commit, 4 files: `_shell_parse.py`, `tests/test_shell_parse.py`, `validate_commit_identity.py`, `tests/test_validate_commit_identity.py`. Charter-compliant additive-only on ChangesRequested cycle (§ Additive Commits on ChangesRequested) — no force-push, HEAD-SHA-anchored verification chain preserved.

## Load-bearing change: `_looks_like_git_commit` + fail-closed parse-failure

This is the meaningful new logic, not pure-format additive. Mental trace:

**Pre-`a34660e` security shape** of `validate_commit_identity._find_commit_segment`:
- shlex parse fails → returns `None` → `check()` returned `None` → ALLOW
- That's a bypass surface: paste `git -c user.email='unclosed commit -m "..."` → shlex chokes on unbalanced quote → silent allow → identity validation skipped. The pre-fix `_strip_heredocs` + `_strip_quoted_strings` design had the same hole (fail-open on parse failure was always there).

**Post-`a34660e` fail-closed shape:**
- `_find_commit_segment` returns `None` → `check()` calls `_looks_like_git_commit(command)` regex fallback
  - Regex finds `git ... commit` at command position → BLOCK with parse-failure-specific message ("fix quoting and retry")
  - Regex doesn't find commit → ALLOW (no commit suspected, true allow)

This is the correct shape for security-relevant matcher per the new `_shell_parse.tokenize` caller contract docstring ("callers MUST handle None explicitly. Never treat None as 'allow' for security-relevant matchers").

**Regex `_COMMIT_FALLBACK_RE`** — `(?:^|[;&|]\s*|&&\s*|\|\|\s*)\s*git\b[^;&|]*?\bcommit\b`:
- Anchors at start-of-line OR after shell operator (`;`, `&`, `|`, `&&`, `||`)
- `[^;&|]*?` non-greedy, excludes shell operators (so it doesn't span past pipeline boundary — important: `git status; echo 'fake commit'` doesn't false-match)
- `\bcommit\b` requires word-boundary on commit (so `committed` doesn't match)

**Edge cases I traced mentally:**
- `git -c user.name="unclosed commit` → strip_heredocs no-op, regex matches at start → BLOCK ✓ (the security case)
- `echo "git commit foo"` → `echo` at position 0, anchor fails → ALLOW ✓ (no false positive)
- `printf '%s' "git commit"` → same, anchor fails → ALLOW ✓
- `EDITOR=vim git commit -m "..."` with quote bug → start-of-line is `EDITOR=`, NOT git, anchor fails → ALLOW. **Mild false-negative**, but only under shlex parse failure AND with leading env-vars — narrow case. Pre-fix behavior was also allow here (worse: also allowed clean cases). Acceptable.

## Other contract pins

- **`tokenize` docstring** — "callers MUST handle None explicitly. Never treat None as 'allow' for security-relevant matchers like commit-identity validation — fall back to a regex check or a fail-closed decision. For warn-only matchers, fail-open on None is acceptable." Correct contract for the load-bearing distinction. Good language.
- **`extract_dash_c_pairs` repeated-key contract** — "returns ALL pairs in source order; callers needing last-wins semantics can do `dict(extract_dash_c_pairs(...))`." Matches git semantics (last `-c key=val` wins). The `validate_commit_identity` caller uses `dict(...)` explicitly — consistent with contract.
- **`extract_dash_c_pairs(commit_segment)` replacing the old `re.search(r'-c\s+user\.name=...')`** — this is the actual #226 fix. Old regex was vulnerable to the unquoted-value-slurps-to-EOL bug; tokenized walk is not.

## Test coverage of new logic

Verified test additions in `test_validate_commit_identity.py` and `test_shell_parse.py`:

- `test_unbalanced_quote_with_commit_blocks` — covers fail-closed regex path ✓
- `test_unbalanced_quote_without_commit_allows` — covers fail-open-when-no-commit ✓
- `test_repeated_dash_c_user_name_uses_last_value` — covers git last-wins semantics via `dict()` ✓
- `test_heredoc_body_with_git_commit_phrase_does_not_match` — covers strip_heredocs running first ✓
- `test_label_validator_filename_no_longer_blocks` — original #226 repro ✓
- `test_nested_heredoc_in_command_substitution` — original #188 repro ✓
- `test_malformed_quote_returns_none` (in `test_shell_parse.py`) — covers tokenize None contract ✓
- `test_unquoted_value_bounded_by_whitespace` (in `test_shell_parse.py`) — `extract_dash_c_pairs` semantics ✓

## Strengths of this delta

- **Closes a pre-existing security hole** (parse-failure-allow on commit-identity validation) that existed in the original implementation and was carried forward through the initial bundle. Not just an API-pin commit — meaningful security improvement.
- **Caller contract codified in docstring**, not just behavior. Future callers of `tokenize` can't accidentally fail-open on a security-relevant matcher.
- **Last-wins semantics matches git's actual behavior** — important because the matcher would otherwise not catch a `git -c user.name=Real -c user.name=Spoofed commit` attack.
- **Test names self-document** the security intent (`test_unbalanced_quote_with_commit_blocks`, `test_unbalanced_quote_without_commit_allows`).

## Charter compliance preserved

- Additive-only commit on the same branch (§ Additive Commits on ChangesRequested)
- HEAD-SHA-anchored verification chain stable (`gh api compare 0760c31...a34660e` resolves cleanly; verification primitive preserved)
- Re-review against artifact at new HEAD via `gh api contents` reads (per `feedback_review_against_artifact_not_framing.md`)
- `mergeable_state: clean` at `a34660e`

## Verification chain

- HEAD SHA: `a34660eeca2a2886eb7cb60f2610ad610273fce0`
- Delta commits: 1 (`a34660e`); files: 4 (subset of original 12)
- mergeable_state: `clean`
- Mental trace of all new logic paths against the security boundary; test coverage confirms the boundary

— Nadia

parametrization:
Requestor: Aino Virtanen
Requestee: Nadia Khoury, Wanjiku Mwangi
RequestOrReplied: Reply
TechDebt: <none>

Audit-trail correction at @Nadia's flag: the cross-repo coordination model in this PR is **Option C — per-repo byte-sync at next hook-sync cadence**, NOT Option A.

My earlier PR-comment phrasing called it "Option A (parent-canonical helper, child consumes via parent+child hook sync pattern)." That was loose terminology — I conflated "parent is the canonical source" (true) with "children import the parent's file at runtime" (NOT true and NOT what this PR establishes).

Confirming for the record what the **code** actually does:
- `.claude/hooks/_shell_parse.py` is a same-directory module imported via `sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))` in each consumer hook. There is NO cross-repo Python import path; nothing in this diff walks up to a parent repo's hook directory at runtime.
- For child repos to consume the helper, the file must be **byte-copied** into each child's `.claude/hooks/` at the next hook-sync — exactly the same convention #841 used to byte-sync `validate_commit_identity.py`, and exactly the same convention Linh confirmed in our cross-repo design Q (see relayed reply earlier).
- "Parent-canonical" describes ownership (parent is the source of truth and the place edits land first), NOT runtime coupling.

This matters for downstream readers:
- W4 retro should NOT cite this PR as establishing a "cross-repo runtime import" precedent (none exists).
- My own #214 work (parent-canonical hook paths) should not assume runtime-import coupling — the consolidation goal is one canonical source for edits, with byte-sync remaining the propagation mechanism.
- isnad-graph#819+#814 (Linh) consumes via byte-sync at next sync, not via import.

No code change required — diff already correct. This is a terminology fix only.

— Aino

Issue #290, opened 2026-05-06
feat(charter+skill): default to isolation: "worktree" for all implementer spawns; codify in agents.md + /wave-kickoff Step 9


parametrization:
## Proposal

Codify "default to `isolation: \"worktree\"` for all implementer spawns" — both as a charter rule (`agents.md`) and as a `/wave-kickoff` SKILL.md Step 9 update. Owner-named at P3W6 wave-kickoff (2026-05-06) after observing that the 18 spawned agents appeared in the harness UI as "background tasks" rather than as workspace-presented team members.

## What happened (concrete repro)

P3W6 spawn batch (`/wave-kickoff` Step 9 → orchestrator-driven Agent calls) used `subagent_type: general-purpose` + `team_name: noorinalabs` for all 18 implementers but **omitted `isolation: "worktree"`**.

Result:
- Team membership: ✅ all 18 correctly linked to `noorinalabs` team (`~/.claude/teams/noorinalabs/config.json` shows 19 members = lead + 18)
- `SendMessage` addressability: ✅ works
- Hook 14 (`enforce_ontology_context.py`): not fired (only fires for worktree-isolated agents) — but Hook 15 (`enforce_librarian_consulted.py`) still fired correctly
- **Harness UI presentation**: agents appeared as "background tasks" in a generic list rather than as workspace-presented team members

Owner observation: "why is everyone a background task and not in agent teams?"

## Reasoning at spawn time (was: weighing the trade-off, badly)

The orchestrator chose to omit `isolation` because:
- A worktree-isolated agent gets a worktree of the **orchestrator's** repo (`noorinalabs-main`)
- 17 of the 18 implementers do code work in **child** repos (`noorinalabs-isnad-graph`, `noorinalabs-deploy`, etc.) which are `.gitignore`d from the parent
- An orchestrator-side worktree is a **cosmetic empty container** for child-repo work — the agent still does `git worktree add` inside the child repo to get a real isolated workspace there
- Conclusion: skip the empty parent worktree to reduce overhead

The reasoning was technically correct but **missed the UI/presentation cost**. The harness uses `isolation: "worktree"` as the signal for "treat this as a workspace-presented team member"; without it, the agent falls into the generic background-tasks bucket regardless of `team_name` correctness.

### Layer 1: Charter (`.claude/team/charter/agents.md`)

Add a § *Spawn Isolation Default* under existing § *Hub-and-Spoke Orchestration Model*:

```
### Spawn Isolation Default

[...]

Issue #287, opened 2026-05-06
tech-debt(hook): validate_commit_identity false-blocks backslash-line-continuation 'git commit' with misleading 'shlex parsing' error


parametrization:
## Bug

`validate_commit_identity` PreToolUse hook returns the misleading error `BLOCKED: git commit detected but command failed shlex parsing (likely unbalanced quotes)` for commands that DO parse successfully but use line-continuation backslashes. Recurring annoyance — surfaced multiple times across recent sessions, including twice during P3W6 kickoff (2026-05-06).

## Repro

Failing command (verbatim from blocked invocation):

```bash
\
git add cross-repo-status.json && \
git -c user.name="Wanjiku Mwangi" -c user.email="parametrization+Wanjiku.Mwangi@gmail.com" \
  commit -F /tmp/wave6_kickoff_commit_msg.txt
echo "---"
git log -1 --oneline
```

Working alternative (same identity, single-line):

```bash
git -c user.name="Wanjiku Mwangi" -c user.email="parametrization+Wanjiku.Mwangi@gmail.com" commit -F /tmp/msg.txt
```

## Root cause (verified, NOT a shlex parse failure)

`shlex.split` tokenizes the failing command **successfully**:

```python
>>> import shlex
>>> shlex.split('\\\ngit add cross-repo-status.json && \\\ngit -c user.name="X" -c user.email="Y" \\\n  commit -F /foo.txt')
['\ngit', 'add', 'cross-repo-status.json', '&&', '\ngit', '-c', 'user.name=X', '-c', 'user.email=Y', '\n', 'commit', '-F', '/foo.txt']
```

The first and fifth tokens are `'\ngit'` (literal newline + `git`), not `'git'`. `_find_commit_segment` (in `.claude/hooks/validate_commit_identity.py`) walks pipeline segments looking for token equal to `'git'`, doesn't find one, and returns `None` for the commit segment. Control flows into the parse-failure-fail-closed path, which emits the misleading "failed shlex parsing" error.

The error message is wrong: shlex didn't fail. The token walker failed to recognize a `\ngit` token.

## Why this matters

1. **Diagnostic noise.** The error tells the operator "fix your quoting" when the quoting is fine.
2. **Recurring waste.** Every operator who uses backslash-line-continuation (a standard bash form) hits this on first try. Workaround is to collapse to a single line — but that means rewriting commands that copy-pasted cleanly from runbooks/docs.
[...]

Issue #286, opened 2026-05-06
feat(hook): hookify /wave-kickoff Steps 7+8 — auto-post charter-format kickoff comment on wave-label-apply


parametrization:
## Proposal

Hookify `/wave-kickoff` Steps 7 + 8 (apply wave label + post charter-format kickoff comment) as a `PostToolUse` hook on `gh issue edit --add-label "p{N}-wave-{M}"`. Idea surfaced by owner during P3W6 kickoff (2026-05-06).

## Why a hook (vs. continuing as skill steps)

Per `feedback_enforcement_hierarchy.md`: prefer hook > skill > charter. Skills can be skipped or mis-executed; hooks fire deterministically on the trigger event.

Concrete failure mode this fixes: W6 kickoff applied 11 wave labels and posted 11 kickoff comments via Python wrapper. If the orchestrator forgets the comments or the wrapper crashes mid-loop, you get labeled-but-uncommented issues — invisible drift caught only at the next wave-wrapup or retro. A hook removes the "forget" failure mode entirely.

Same-class promotion as W5 #276 (`/wave-scope` self-threading) and W4 #250 (`validate_pr_review` canonicalization) — Pattern G primitives. The kickoff-comment template is a known-stable shape (charter `pull-requests.md` § Kickoff Comment Format defines it), so the hook payload is straightforward to author.

## Hook shape

**Trigger:** `PostToolUse` on `Bash` matching pattern `gh issue edit (\d+) --repo noorinalabs/(\S+) .*--add-label "(p\d+-wave-\d+)"`.

**Payload:**

1. Parse repo, issue number, wave label from the command.
2. Read `cross-repo-status.json` and locate the issue in `wave_{M}_scope.tier_{1,2,3,4}_*` arrays (match by `id` for explicit issues, by `repo` for Tier-1 backlog rows).
3. If a matching assignment row is found, render a charter-format kickoff comment using fields: `implementer`, `reviewer`, `reviewer_2`, branch name, priority tier.
4. Post the comment via `gh issue comment <num> --repo <repo> --body-file <fresh-tmp>`.
5. If the issue is the meta-issue itself (`wave_{M}_meta_issue == issue`), skip per-issue posting (meta-issue gets a different all-hands kickoff comment, owned separately).

[...]

Issue #285, opened 2026-05-06
tech-debt(/wave-kickoff): Step 1 EXISTING_SHA captures 404 error body, falsely reports exists-drift on fresh wave


parametrization:
## Bug

`/wave-kickoff` Step 1 (branch creation across wave repos) silently mis-reports `exists-drift` for every repo on a fresh wave kickoff because the existing-branch probe captures the 404 error body as if it were a real SHA.

### Repro

`.claude/skills/wave-kickoff/SKILL.md` Step 1, the existing-branch probe:

```bash
EXISTING_SHA=$(gh api "repos/noorinalabs/$R/git/refs/heads/$BRANCH" --jq '.object.sha' 2>/dev/null || true)
```

When the branch does NOT exist, `gh api` returns 404. Critically:
- `gh api` prints the JSON error body (`{"message":"Not Found",...}`) to **stdout** (not stderr — only the trailing `gh: Not Found (HTTP 404)` summary goes to stderr)
- The `--jq` filter is **not** applied to error responses
- `2>/dev/null` only suppresses the stderr summary, NOT the stdout JSON body
- `|| true` masks the non-zero exit code

Result: `EXISTING_SHA` captures the literal string `{"message":"Not Found",...}`, the `[ -n "$EXISTING_SHA" ]` check passes, and the script erroneously enters the "branch already exists" branch. The compare API call then yields `unknown`, which the `case` statement maps to `exists-drift`. Branch creation is **silently skipped** for every repo.

### Observed at P3W6 kickoff (2026-05-06T22:23 UTC)

```
| Repo                              | Branch SHA      | Status         |
|-----------------------------------|-----------------|----------------|
| noorinalabs-main                  | {"message":     | exists-drift   |
| noorinalabs-isnad-graph           | {"message":     | exists-drift   |
| ... (all 8 repos)                 | ...             | exists-drift   |
```

`cross-repo-status.json` `wave_6_branches.repos.<repo>.sha` was written with the JSON error body as the SHA — corrupt persisted state. Direct verification with `gh api repos/noorinalabs/noorinalabs-main/git/refs/heads/deployments/phase-3/wave-6` returned 404, confirming no branch was actually created.

### Fix

Use exit-code-aware capture instead of `|| true` masking:

```bash
probe_branch() {
  local owner_repo="$1" ref="$2"
  local sha
[...]

PR #291, opened 2026-05-06
fix(hook #289): consult pr_has_paths_key on on:-block boundary in validate_workflow_paths_coverage

### LINKED ISSUE #289: tech-debt(hook): validate_workflow_paths_coverage misparses bare `on.pull_request:` (no paths) as covering-nothing instead of covering-all
## Summary

`.claude/hooks/validate_workflow_paths_coverage.py` (filed in [#203](https://github.com/noorinalabs/noorinalabs-main/issues/203)) misparses workflows that use the bare `on.pull_request:` form (no `paths:` key) and reports them as covering **nothing** rather than covering **everything**. By GitHub's actual semantics, a `pull_request:` trigger with no `paths:` filter matches every file path — so the hook is rejecting PRs whose CI coverage is in fact valid.

## Repro

`noorinalabs-landing-page/.github/workflows/ci.yml` on the `deployments/phase-3/wave-6` branch is the reference case:

```yaml
name: CI

on:
  push:
    branches: [main, "deployments/**"]
  pull_request:
    branches: [main, "deployments/**"]

concurrency:
  group: ci-${{ github.ref }}
  cancel-in-progress: true
```

`pull_request:` is a block, has a `branches:` key, no `paths:` key. GitHub runs this workflow on every PR file (any path).

Driving the parser directly:

```bash
cd /home/parameterization/code/noorinalabs-main
python3 -c "
import sys; sys.path.insert(0, '.claude/hooks')
from validate_workflow_paths_coverage import _parse_workflow_paths
import subprocess, base64
result = subprocess.run(
    ['gh','api','repos/noorinalabs/noorinalabs-landing-page/contents/.github/workflows/ci.yml?ref=deployments%2Fphase-3%2Fwave-6','--jq','.content'],
    capture_output=True, text=True)
yml = base64.b64decode(result.stdout.strip()).decode('utf-8')
print(_parse_workflow_paths(yml))
"
# Output: (set(), False)   ← WRONG; should be (set(), True)
```

`(set(), False)` from `_parse_workflow_paths` means "no paths covered AND no bare-pull_request-trigger" — which causes `_build_coverage_signal` to compute `any_no_paths = False`, which causes `check()` to enter the per-file glob loop, which fails for every workflow file in the diff.

## Symptom in the wild

Triggered on `gh pr create` for `K.Mensah-Williams/0077-w6-deploy-vps-fix` (P3W6 hotfix for landing-page#77). The PR modifies `.github/workflows/deploy.yml` on a wave-6 base where ci.yml's bare `pull_request:` should satisfy the hook's option (b). The hook blocks with the option (a) / (b) / (c) message; option (b) is already satisfied in code but invisible to the parser.

## Code reference (root cause)

`validate_workflow_paths_coverage.py` lines 197–207 dispatch on `if indent <= on_indent:` first, when the loop encounters a top-level key that exits the `on:` block (e.g., `concurrency:`). Inside that branch the parser clears `in_on`, `in_pr`, `in_pr_paths` and `continue`s — without consulting `pr_has_paths_key`. The "exiting `pull_request:` without seeing `paths:`" → return `(set(), True)` logic at lines 220–229 is only reached on the SIBLING boundary (a key at `pr_indent`), not on the GRANDPARENT boundary (a key at `on_indent`).

Concretely for ci.yml above, the indent state is:
- `on:` at indent 0 → `on_indent = 0`
- `pull_request:` at indent 2 → `pr_indent = 2`, `pr_has_paths_key = False`
- `branches: ...` at indent 4 → ignored (not `paths:`/`paths-ignore:`)
- next non-empty line `concurrency:` at indent 0 → `indent (0) <= on_indent (0)` matches FIRST at line 198, so the parser exits `on:` and clears `in_pr` to False without ever returning the no-paths signal. Final return is `(set(), False)`.

## Proposed fix

In the `if indent <= on_indent:` branch (lines 198–207), consult `pr_has_paths_key` BEFORE clearing `in_pr`:

```python
if indent <= on_indent:
    in_on = False
    if in_pr and not pr_has_paths_key:
        return set(), True
    in_pr = False
    in_pr_paths = False
    # ... re-evaluation logic unchanged
```

Mirror the same guard in any future "boundary-exit" branch (this hook's parser already has the analogous logic at the sibling boundary, lines 220–229; the bug is just that the grandparent boundary doesn't have a copy of it).

The trailing `if in_pr and not pr_has_paths_key` block at lines 256–258 catches the EOF-while-still-in-pr case but doesn't fire when `in_pr` was prematurely cleared by the grandparent-boundary branch.

## Test coverage gap

`.claude/hooks/tests/test_validate_workflow_paths_coverage.py` should grow a fixture for the bare `on.pull_request:` form where the workflow file ends OR is followed by a sibling top-level key (e.g., `concurrency:`). The existing tests (per the test file's history) likely only cover the inline `on: pull_request` and the explicit `paths:` cases.

## Class

P3W6 Pattern G — three hook/skill parser bugs surfaced in a single wave:

- This issue (validate_workflow_paths_coverage)
- [#285](https://github.com/noorinalabs/noorinalabs-main/issues/285) (skill 404 capture parser)
- [#287](https://github.com/noorinalabs/noorinalabs-main/issues/287) (commit-identity false-block parser)

Suggests a wave-retro item on parser-fixture coverage as a class.

## Tracking-only — NOT in W6 scope

Per Steven (via Nadia 2026-05-06): keeps Aino's [#288](https://github.com/noorinalabs/noorinalabs-main/issues/288) charter-side surgical, does not pull this hook fix into W6 scope. Worked around in landing-page#77 lineage via a precursor PR widening ci.yml `pull_request:` to explicit `paths: ['**']` (semantic-equivalent, parser-recognized).

## Acceptance

- Parser returns `(set(), True)` for the wave-6 ci.yml repro above
- New test fixture covers bare `on.pull_request:` with sibling top-level key after
- Workaround in landing-page/ci.yml (`paths: ['**']`) can be reverted to bare `pull_request:` once the parser fix lands (cross-link this issue from that revert PR)

parametrization:
## Summary

Fixes #289. `validate_workflow_paths_coverage._parse_workflow_paths` misparsed workflows where `pull_request:` was the LAST child of `on:` as covering-nothing instead of covering-all. Discovered at P3W6 wave-kickoff while Kofi Mensah-Williams attempted to open `gh pr create` for the landing-page#77 hotfix — the hook blocked the PR-create against a base branch (`deployments/phase-3/wave-6`) whose `ci.yml` is in exactly this shape.

## Root cause

State-machine parser at line 198–207 cleared `in_pr` and `in_pr_paths` when `indent <= on_indent` (top-level YAML key boundary, e.g., `jobs:` after `on:`) without first consulting `pr_has_paths_key`. The two existing covers-all-recognition paths handle:

- Sibling-boundary case (line 220–227): another sibling under `on:` at `pr_indent` exits `pull_request:` cleanly
- EOF case (line 256–258): file ends still inside `pull_request:`

Missing: the on:-block-boundary case where `pull_request:` is the LAST child of `on:` and the next line is a top-level key (`jobs:`, etc).

GitHub itself runs such workflows on every path (no `paths:` filter = no path filtering), but our parser was reading them as "covers nothing."

## Fix

Make the on:-block-boundary path symmetric with the sibling-boundary and EOF paths:

```python
        # Inside `on:` block.
        if indent <= on_indent:
            # Symmetric with the sibling-boundary case (line 220-227) and the
            # EOF case (line 256-258): if we exit `on:` while still inside
            # `pull_request:` with no `paths:` key seen, that's covers-all.
            # Without this guard, workflows where `pull_request:` is the LAST
            # child of `on:` are misparsed as covering-nothing (#289).
            if in_pr and not pr_has_paths_key:
                return set(), True
            in_on = False
            in_pr = False
            ...
```

5 lines added.

## Test Plan

- [x] New regression test `ParseWorkflowPathsTests.test_pr_last_child_of_on_without_paths_filter` covers the exact failure mode (mirrors landing-page `ci.yml` shape: `push:` first, `pull_request:` last, no `paths:` key, `jobs:` top-level)
- [x] Existing `test_pr_without_paths_filter` (pull_request as ONLY trigger) still passes — regression check
- [x] Existing `test_pr_with_inline_paths_list` (explicit paths) still passes — paths-extraction not affected
- [x] Existing `test_no_pull_request_trigger` (push: only) still passes — non-PR workflows unchanged
- [x] All 45 tests in the suite pass (`ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/test_validate_workflow_paths_coverage.py -v`)
- [x] Local repro of #289's failure mode now correctly returns `(set(), True)` instead of `(set(), False)`

## Discovery context (Pattern G primitive)

This is the third hook/skill parser bug surfaced during P3W6 alone:
- `#285` — `/wave-kickoff` Step 1 EXISTING_SHA captures 404 body (in-scope, p3-wave-6)
- `#287` — `validate_commit_identity` false-blocks backslash-line-continuation (tracking only)
- `#289` — this PR (tracking only originally; pulling in-wave because it blocks landing-page#77)

Plus charter/skill codifications surfaced:
- `#286` — hookify `/wave-kickoff` Steps 7+8
- `#290` — default `isolation: "worktree"` for implementer spawns

Pattern G ("in-wave skill self-improvement") W6 tally: **5 surfaced**, vs W4=1 + W5=1. The "lower-the-noise-floor" wave theme is producing visible meta-tooling clarity worth flagging at retro.

## Reviewers

- R1: **Aino Virtanen** (Standards & Quality Lead — natural owner per W4/W5/W6 hook canonicalization track record)
- R2: **Nadia Khoury** (Program Director — hub-and-spoke endorsement)

Per charter `pull-requests.md`, comment-only reviews. Each comment ends with `TechDebt: <none|#N>` line.

## Acceptance after merge

1. `gh pr create` for landing-page workflow-touching PRs succeeds against any base whose `ci.yml` has `pull_request:` as last child of `on:` — no precursor-PR workaround needed.
2. Kofi's hotfix #77 PR opens cleanly without the chicken-and-egg.
3. `noorinalabs-main#289` closes when this PR merges.

## Out of scope (deliberately)

- Kofi's hotfix #77 itself (separate PR, lands after this).
- Broader audit of other state-machine parsers in `.claude/hooks/` for similar boundary-clearing bugs (could be retro action item; this PR is surgical).
- Refactoring the parser to use a YAML library (charter convention is regex-based for hooks; this fix preserves that).

TechDebt: <none>

### DIFF
--- a/.claude/hooks/tests/test_validate_workflow_paths_coverage.py
+++ b/.claude/hooks/tests/test_validate_workflow_paths_coverage.py
@@ -120,6 +120,26 @@ class ParseWorkflowPathsTests(unittest.TestCase):
         self.assertEqual(paths, set())
         self.assertTrue(no_paths)
 
+    def test_pr_last_child_of_on_without_paths_filter(self):
+        """Regression for #289: `pull_request:` as LAST child of `on:` with no
+        `paths:` key was misparsed as covering-nothing because the on:-block
+        boundary cleared `in_pr` without consulting `pr_has_paths_key`.
+        """
+        yml = (
+            "name: CI\n"
+            "on:\n"
+            "  push:\n"
+            "    branches: [main]\n"
+            "  pull_request:\n"
+            '    branches: [main, "deployments/**"]\n'
+            "jobs:\n"
+            "  build:\n"
+            "    runs-on: ubuntu-latest\n"
+        )
+        paths, no_paths = hook._parse_workflow_paths(yml)
+        self.assertEqual(paths, set())
+        self.assertTrue(no_paths)
+
     def test_pr_with_inline_paths_list(self):
         yml = 'name: CI\non:\n  pull_request:\n    paths: ["src/**", "tests/**"]\n'
         paths, no_paths = hook._parse_workflow_paths(yml)
--- a/.claude/hooks/validate_workflow_paths_coverage.py
+++ b/.claude/hooks/validate_workflow_paths_coverage.py
@@ -196,6 +196,13 @@ def _parse_workflow_paths(yml_text: str) -> tuple[set[str], bool]:
 
         # Inside `on:` block.
         if indent <= on_indent:
+            # Symmetric with the sibling-boundary case (line 220-227) and the
+            # EOF case (line 256-258): if we exit `on:` while still inside
+            # `pull_request:` with no `paths:` key seen, that's covers-all.
+            # Without this guard, workflows where `pull_request:` is the LAST
+            # child of `on:` are misparsed as covering-nothing (#289).
+            if in_pr and not pr_has_paths_key:
+                return set(), True
             in_on = False
             in_pr = False
             in_pr_paths = False


parametrization:
Requestor: Aino Virtanen
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved

## Reviewer-class R1 — Approved

Reviewed at HEAD `2148baf` against artifact (`gh api .../contents/...?ref=<sha>`), not PR-body framing.

### Symmetry argument: sound

The new guard at lines 198–205 is genuinely symmetric with the two existing covers-all-recognition paths. All three converge on the same predicate `in_pr and not pr_has_paths_key`, returning `(set(), True)`:

| Boundary | Trigger | Lines (at HEAD) |
|---|---|---|
| Sibling-exit (existing) | `indent <= pr_indent` while `in_pr` | 228–234 |
| EOF (existing) | loop end while `in_pr` | 263–265 |
| `on:`-block-exit (NEW) | `indent <= on_indent` while `in_pr` | 198–205 |

Pre-fix the third case was the only path that cleared `in_pr/in_pr_paths` without consulting `pr_has_paths_key` — the asymmetry was the bug. The fix closes the gap.

### Pre/post-fix repro confirmed

Ran the test_pr_last_child_of_on_without_paths_filter shape against the parser at `main` and at HEAD:

- **Pre-fix (`main`):** `(set(), False)` — covers-nothing (the bug)
- **Post-fix (HEAD):** `(set(), True)` — covers-all (correct)

### Edge-case probe matrix (all green)

Ran 5 additional shapes locally against the HEAD parser to verify the new guard doesn't misfire:

| Shape | Expected | Got |
|---|---|---|
| `pull_request:` last-child + inline paths `["src/**"]` | `({'src/**'}, False)` | `({'src/**'}, False)` |
| `pull_request:` last-child + block-list paths | `({'src/**', '.github/workflows/**'}, False)` | same |
| `pull_request:` first / `push:` last (sibling-exit) | `(set(), True)` | `(set(), True)` |
| `pull_request:` only, no `jobs:` (EOF) | `(set(), True)` | `(set(), True)` |
| `pull_request:` last-child + `paths-ignore:` only | `(set(), True)` per existing semantics | `(set(), True)` |

The `paths-ignore`-last-child case correctly preserves the existing behavior (matches `test_paths_ignore_treated_as_no_paths_filter` semantics — `pr_has_paths_key` stays False when only `paths-ignore:` is seen, so the new guard fires). No misfire on inline-paths or block-paths variants where `pr_has_paths_key=True` blocks the new guard.

### Test suite

- `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/test_validate_workflow_paths_coverage.py -v` → **45 passed in 0.09s**
- New regression test mirrors landing-page `ci.yml` shape exactly (`push:` first, `pull_request:` last with `branches:` only, no `paths:` key, top-level non-`on:` key follows).

### Live artifact alignment

Cross-checked against `noorinalabs-landing-page/.github/workflows/ci.yml@main` (live shape Kofi tripped over):

```
on:
  push:
    branches: [main, "deployments/**"]
  pull_request:
    branches: [main, "deployments/**"]

concurrency:
  ...
```

Test uses `jobs:` as the post-`on:` boundary key; live file uses `concurrency:`. The parser path is the same (`indent <= on_indent`) — irrelevant which top-level key follows. Test fidelity is sufficient.

### Charter consolidation check

Reviewed against `charter/hooks.md` § Dispatcher Consolidation Policy: that threshold is for **cross-hook consolidation** (4+ hooks of the same matcher type → dispatcher), not within-function refactoring. Three boundary cases sharing one predicate inside one function is acceptable repetition for a regex-based state-machine parser (charter explicitly endorses this pattern over a YAML library dep — line 153–161 of the hook file). Extracting a `_exit_pull_request_safely()` helper would over-abstract a 5-line guard. **Keep as-is.**

### Two minor nits (non-blocking)

1. **Stale line-number cites in the new comment.** The comment at line 199–201 cites "sibling-boundary case (line 220–227) and the EOF case (line 256–258)" — at HEAD those blocks live at lines 228–234 and 263–265 (5–8 lines off, presumably from an earlier draft of the diff). Self-documenting drift the moment anyone edits this function. Suggest dropping the absolute line numbers in favor of "the sibling-boundary case below" / "the EOF case at the end of the loop" — survives future edits. Not a merge-blocker.

2. **PR body wave-branch closing-keyword note.** PR body says "`noorinalabs-main#289` closes when this PR merges." Per `feedback_wave_branch_issue_close.md`: GitHub's `Closes #N` keyword fires only on **default-branch merges**, not wave-branch merges. Since this PR targets `deployments/phase-3/wave-6` (not `main`), the assertion is misleading. Recommend a manual `gh issue close 289` after the wave-branch merge lands (and again at wave-merge-to-main if not yet closed). Tracking-only — does not affect this fix's correctness.

### Pattern G framing

PR body's "in-wave skill self-improvement" tally (W6=5 vs W4=1, W5=1) is accurate. Worth flagging at retro as the wave's "lower-the-noise-floor" theme is producing visible meta-tooling clarity. Not a review concern; useful retro fodder.

### Decision

**Approved.** Fix is surgical, symmetric with existing recognition paths, regression test exact, full suite green at HEAD, edge cases covered by both existing tests and my probe matrix. No changes requested. The two nits above are tracking-only and can be carried forward or no-op'd at author's discretion.

R2 over to Nadia Khoury.

TechDebt: <none>

parametrization:
Requestor: Nadia Khoury
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved

## R2 Review (Program Director — hub-and-spoke endorsement)

**Status: APPROVED.**

Verified at HEAD `2148bafc`:

- `gh pr view --json state,mergedAt,headRefOid,statusCheckRollup` — OPEN, all 4 CI checks SUCCESS (Ruff lint + Ruff format + Mypy + Smoke), `mergedAt: null`.
- Diff against artifact (`gh api .../pulls/291/files`): exactly 27 additions / 0 deletions across two files — `validate_workflow_paths_coverage.py` (+7) and the regression test (+20). No drift between the PR body's "5 lines added" framing and the file patch (the +7 is 5 logic lines plus 2 blank/comment context lines — within tolerance, body says "5 lines added" referring to the load-bearing block which is accurate).
- Read the full hook source at HEAD lines 196–265: confirmed the new on:-block-boundary guard at line 204 mirrors the sibling-boundary guard at line 232 and the EOF guard at line 264 — symmetry claim in the PR body is artifact-true.

### Code-correctness (PR-class, not runtime)

The fix is the minimal correct change:

1. **Symmetry argument is real.** The parser already had the covers-all-no-paths logic at two of three exit points (sibling boundary L220–227, EOF L256–258). The grandparent boundary was the structural omission. The new guard plugs the third exit identically — `if in_pr and not pr_has_paths_key: return set(), True` — same condition, same return shape. No new state, no new control-flow path; just adding the missing third gate.
2. **Regression test is shape-faithful.** `test_pr_last_child_of_on_without_paths_filter` mirrors the exact landing-page `ci.yml` shape from #289's repro: `push:` first child with branches, `pull_request:` last child with branches but NO `paths:`, then a top-level `jobs:` key triggering the on:-block boundary. Asserts `(set(), True)` — the corrected return. Adjacent `test_pr_without_paths_filter` (pull_request as ONLY child) and `test_pr_with_inline_paths_list` are preserved unchanged, so the existing two boundary cases still get coverage.
3. **No risk to existing-paths semantics.** The new guard only fires when `in_pr=True AND pr_has_paths_key=False`. If `paths:` was seen, `pr_has_paths_key=True` and the old clear-and-continue path runs. If `pull_request:` was never entered, `in_pr=False` and the old path runs. The change is strictly additive to a previously-unhandled state combination.

### Pattern G framing accuracy (PD lens — what you asked me to focus on)

The "third wave-N+1 fix landing in wave-N (W4 #250, W5 #276, W6 #291)" symmetry holds **with one nuance worth naming for the retro**:

| Wave | PR | Layer | Class |
|---|---|---|---|
| W4 | #250 | hook (`validate_pr_review`) | charter-canonicalization fix |
| W5 | #276 | skill (`/wave-scope` thread-in) | new-feature-with-fix-cycle |
| W6 | #291 | hook (`validate_workflow_paths_coverage`) | parser-bug fix |

So "in-wave skill self-improvement" is right at the meta level (we're fixing the meta-tooling mid-wave), but the cross-wave pattern is more precisely **in-wave hook OR skill self-improvement** — two of the three are hooks. That distinction matters at retro because the hook subset (W4, W6) is a tighter class: parser-bug discovery → in-wave fix → regression test → wave-fwd. Companion #285 / #287 keep that hook-parser cluster live for W6 alone (3 parser bugs in one wave, two hook + one skill). Worth tagging as a wave theme primitive at retro: **lower-the-noise-floor waves expose hook-parser fragility at higher than baseline rates.**

The body's "Pattern G W6 tally: 5 surfaced, vs W4=1 + W5=1" is artifact-verified: I checked main#285 (OPEN, /wave-kickoff parser), #287 (OPEN, commit-identity parser), #286 (OPEN, hookify kickoff Steps 7+8), #289 (this PR's target — will close on merge), #290 (OPEN, default isolation worktree). Five Pattern-G-class issues in W6, all OPEN, all refs in body match GitHub state.

### Closes-line check (you flagged this)

Body line: `Fixes #289`. Present and correctly formatted. On merge to `deployments/phase-3/wave-6` (wave branch), per `feedback_wave_branch_issue_close.md` the auto-close will NOT fire — needs an explicit `gh issue close 289` after wave-branch merge. Flag this as the wave-wrapup followup.

### Honest-audit / "concluded" check (per `feedback_honest_audit_over_conclusion_claim.md`)

Body's "Out of scope (deliberately)" section is well-shaped — explicitly defers (1) Kofi's hotfix #77 itself, (2) broader audit of other state-machine parsers in `.claude/hooks/`, (3) parser refactor to YAML library. Those are all accurate scope-cuts, not optimistic-eliding.

The body's W6 Pattern G enumeration (5 issues) does NOT claim any of them concluded — they're all listed as "surfaced" with current state visible (`#287 — tracking only`, `#289 — this PR`, etc). Clean.

### Cross-repo / runtime acceptance

Per `feedback_pr_vs_runtime_acceptance_criteria.md`, this is PR-class (code correctness verified — fix is minimal and symmetric, regression test covers the failure mode, no existing tests broken). Runtime acceptance is `gh pr create` for landing-page#77 succeeding without the precursor `paths: ['**']` workaround — that's a post-merge wave-wrapup verification on the wave-6 branch, properly listed as "Acceptance after merge" item 1 in the body.

### Minor / non-blocking

- The body's Test Plan checks `[x]` "Local repro of #289's failure mode now correctly returns `(set(), True)` instead of `(set(), False)`" but the diff doesn't include a script for that. Not blocking — the regression test asserts exactly that semantics. If you want the explicit repro for the audit trail, add a docstring example to `_parse_workflow_paths` or capture the one-liner from #289's body in a comment. Don't gate merge on it.

### Endorsement

Approving as R2. Aino's R1 will land on the same comment cycle. Recommend merging to `deployments/phase-3/wave-6` once R1 posts. Per the wave-branch close discipline, `gh issue close 289` post-merge.

TechDebt: <none>

PR #432, opened 2026-05-14
tech-debt(infra): branch protection manifest + audit workflow for 8 org repos (#403)

parametrization:
## Summary

Adds a declarative, idempotent rollout of `main` branch-protection across all 8 noorinalabs org repos, plus a scheduled drift-check workflow.

- `.claude/scripts/branch_protection_config.json` — per-repo manifest with a `_common` block (admins not exempt, force-push/deletion blocked, **2-reviewer requirement with dismiss-stale-reviews**) and per-repo `required_status_checks` lists cross-checked against most-recent-merged-PR check-run **names** (not job keys) to dodge the silent-never-satisfy trap.
- `.claude/scripts/apply_branch_protection.py` — `--check` (read-only drift diff, exit 1 on drift) and `--apply` (PUT then GET-verify, writes per-repo post-state JSON to `/tmp/apply_branch_protection_post_<repo>.json`). Uses `gh api` so caller auth flows through — no PAT plumbing in the script.
- `.claude/scripts/test_apply_branch_protection.py` — 30 unit tests (no network) covering manifest load, schema validation, desired-state assembly, GET-shape normalisation (including the new `required_pull_request_reviews` strip), diff computation, PUT body shape, and CLI dispatch. Real-manifest schema test pins the 8-repo set and the uniform `enforce_admins=true` / 2-reviewer assertions.
- `.github/workflows/branch-protection-audit.yml` — runs `--check` weekly (Mon 07:17 UTC) and on workflow_dispatch; unit-tests job runs on every PR that touches these files (path-filtered). **`apply` mode is gated by the `branch-protection-apply` GitHub Environment with required-reviewer = owner**, so a manual `apply` dispatch blocks at the approval gate.

Closes: noorinalabs/noorinalabs-main#403
Anchors: #182, #222

## Precursor

This PR depends on **#430** (merged at commit `e98f74d`), which widened `ci.yml`'s `pull_request.paths` filter from `.github/workflows/ci.yml` to `.github/workflows/**`. Without that widening, Hook 17 (`validate_workflow_paths_coverage`) blocks PR open on the new `branch-protection-audit.yml`. The precursor is now in `deployments/phase-3/wave-10` as `9749bfe`, so this PR rebases cleanly atop the latest wave-10 HEAD.

## Security hardening (post-Nino review)

[...]

PR #434, opened 2026-05-16
fix(hook15): diagnose + tolerate sentinel regression (#429)

parametrization:
Closes #429.

## Summary

Hook 15 (`enforce_librarian_consulted`) was blocking Edit/Write in worktree-subagent sessions despite a fresh cwd-keyed sentinel existing (11 captures in a 4-minute window 2026-05-13 18:19–18:22Z). Root-cause investigation surfaced **two compounding faults**, both addressed by this PR.

## Root cause — both H1 and H2 confirmed

**H1 — hash-formula inconsistency in the wild.** Filesystem scan of `noorinalabs-main/.claude/worktrees/cedric-0066-dark-mode-tokens/.claude/.consulted/ontology-librarian/` found TWO sentinels under the same cwd with different filename hashes:

| Marker | Hash formula | Body |
|---|---|---|
| `cc41382ce21fdb6e.marker` | `sha1(cwd + "\n")[:16]` (canonical shell-formula) | `2026-05-14T03:28:57Z /home/.../cedric-0066-dark-mode-tokens` |
| `2e85a77c97b204d3.marker` | `sha1(cwd)[:16]` (NO trailing newline) | `2026-05-14T02:59:49Z /home/.../cedric-0066-dark-mode-tokens` |

Reverse-hash math confirms `2e85a77c97b204d3 = sha1("/home/.../cedric-...".encode())[:16]` — almost certainly an ad-hoc workaround writer (including the very workaround that the spawn brief recommends as a fix-this-bug fallback) that took a `python -c 'sha1(cwd.encode())...'` shortcut and forgot the `\n`. The canonical-path read in the hook silently ignored the orphan marker. Charter parity tests catch shell-vs-helper drift but not third-party writers using a third formula.

**H2 — zero observability on block.** Every `pretooluse_block` annunaki record recorded only `tool_name` and the file path. Nothing about `input_data["cwd"]`, the sentinel path tried, sentinel existence, freshness comparison, transcript path scanned, OSError text — operators had no way to tell WHY a block fired. All 11 captures in the issue body are uniformly opaque for exactly this reason.

[...]

PR #431, opened 2026-05-13
infra(auto-close): add auto-close-issues workflow (propagation from isnad-graph per main#402)

parametrization:
Propagates `auto-close-issues.yml` from noorinalabs-isnad-graph (canonical) per noorinalabs-main#402.

35-line literal copy (parameter-free: uses `context.repo.*` only). The workflow fires on `pull_request.closed && merged == true`, parses `[Cc]loses?\s+#(\d+)` from PR body, calls `github.rest.issues.update({state:'closed'})` + creates a "Closed by PR #N merged into <base>" comment. This handles wave-branch merges that GitHub's native trailer-parsing skips (default-branch-only behavior).

After standardization, memory `feedback_wave_branch_issue_close` discipline downgrades from "explicit `gh issue close` required after wave-branch merge" to "verify worked, don't re-do."

Related:
- noorinalabs-main#402 (tracking issue — propagation to 7 org repos)
- noorinalabs-main#221 (origin proposal — audit identifying gap)

Reviewers: Aisha Idrissi + Bereket Tadesse

Issue #429, opened 2026-05-13
bug(hook15): sentinel-fallback no longer attests in worktree subagent — fresh transcript-flush race regression (11 captures 2026-05-13)


parametrization:
## Symptom

`enforce_librarian_consulted` (Hook 15) is blocking Edit/Write in worktree-subagent sessions DESPITE a cwd-keyed sentinel marker existing at the expected path AND being fresh (sub-minute mtime, well within `SENTINEL_TTL_SECONDS = 3600`).

Annunaki captured **11 blocks in a 4-minute window (2026-05-13 18:19–18:22 UTC)** across three concurrent worktree subagents:

| Time | Agent | Worktree path | File targeted | Tool |
|---|---|---|---|---|
| 18:19:27 | acquisition-sofia | `.claude/worktrees/S.Cardoso-0050-adr-001-citations/` | `docs/adr/001-local-hook-policy-dispatcher-style.md` | Edit |
| 18:19:42 | acquisition-sofia | (same) | (same) | Edit |
| 18:19:58 | landing-page-anika | `noorinalabs-landing-page/...` | `public/robots.txt` | Edit |
| 18:20:23 | acquisition-sofia | (same as row 1) | (same) | Edit |
| 18:21:14 | landing-page-anika | (same as row 3) | (same) | Edit |
| 18:21:25 | landing-page-anika | (same) | `src/pages/index.astro` | Edit |
| 18:21:26 | landing-page-anika | (same) | `src/layouts/ProjectLayout.astro` | Edit |
| 18:21:54 | main-aisha | `.claude/worktrees/aisha-0403-branch-protection/` | `.claude/scripts/branch_protection_config.json` | Write |
| 18:22:09 | main-aisha | (same) | (same) | Write |
| (+2 more in same window) |

Full captures in `.claude/annunaki/errors.jsonl` filtered by `type=pretooluse_block AND hook=enforce_librarian_consulted` between 18:19–18:22Z.

## Suspected root cause

The sentinel-fallback path added by #174 (per #169) is no longer accepting valid markers. Verified by acquisition-sofia at 18:20Z:

- Sentinel directory exists at `<worktree-cwd>/.claude/.consulted/ontology-librarian/`
- Filename hash matches what `_consultation_sentinel.cwd_sentinel_hash()` computes for that cwd (`77889fae0c7b5b43` in the data-acq case)
- mtime refreshed to within seconds of the blocked Edit
[...]

PR #435, opened 2026-05-16
tech-debt(infra): preserve bypass_pull_request_allowances + PUT empty list (#433)

### LINKED ISSUE #433: tech-debt(infra): normalize_actual strips bypass_pull_request_allowances — UI-drift blind spot post-#432
## Context

Surfaced during security re-review of #432 at HEAD `a9504db` (post-fixup commit). Source: https://github.com/noorinalabs/noorinalabs-main/pull/432#issuecomment-4447040277 (Nino, security-class re-review, Approved verdict).

PR #432 added `apply_branch_protection.py` + manifest + drift-audit workflow for 8 org repos. During the re-review, the `normalize_actual` function was extended to strip GET-only fields (`url`, `dismissal_restrictions`) from the GET response so `--check` doesn't perma-drift after `--apply`. That fix is correct as far as it goes, but creates a narrow blind spot for `required_pull_request_reviews.bypass_pull_request_allowances` — a security-relevant field.

## Symptom

After `python apply_branch_protection.py --apply` succeeds, a repo admin (or anyone with admin via the GitHub UI) can navigate to **Settings → Branches → main rule → Allow specified actors to bypass required pull requests** and add a user/team. Subsequent runs of `python apply_branch_protection.py --check` will report `[<repo>] in sync` — **the bypass-allowance addition is invisible to drift detection**.

This silently undermines the 2-reviewer enforcement that blocker 2 of #432 was specifically introduced to harden.

## Suspected root cause

In `.claude/scripts/apply_branch_protection.py` `normalize_actual()` (at HEAD `a9504db`, ~lines 137-148), the `required_pull_request_reviews` branch hand-extracts exactly 4 fields:

```python
rpr = actual.get("required_pull_request_reviews")
if rpr:
    norm["required_pull_request_reviews"] = {
        "required_approving_review_count": int(rpr.get("required_approving_review_count", 0)),
        "dismiss_stale_reviews": bool(rpr.get("dismiss_stale_reviews", False)),
        "require_code_owner_reviews": bool(rpr.get("require_code_owner_reviews", False)),
        "require_last_push_approval": bool(rpr.get("require_last_push_approval", False)),
    }
```

GitHub's GET on `/repos/{owner}/{repo}/branches/main/protection` returns `required_pull_request_reviews.bypass_pull_request_allowances` (object with `users`, `teams`, `apps` keys) when configured. Because `normalize_actual` does not preserve it AND the manifest's `_common.required_pull_request_reviews` does not PUT it, the diff in `diff_state` compares `desired == 4-field dict` against `actual_norm == same 4-field dict` and reports in-sync regardless of any GitHub-side bypass-allowance entries.

The companion test `test_required_pull_request_reviews_strips_get_only_fields` only asserts that `url` and `dismissal_restrictions` are stripped — it does not cover `bypass_pull_request_allowances`, so the test suite cannot catch this regression.

## Reproduction

1. After #432 merges and `--apply` runs successfully against `noorinalabs/noorinalabs-main`:
2. Navigate to https://github.com/noorinalabs/noorinalabs-main/settings/branches → Edit `main` rule.
3. Under **Allow specified actors to bypass required pull requests** add any user (e.g., yourself).
4. Save.
5. Run `python .claude/scripts/apply_branch_protection.py --check --repo noorinalabs-main`.
6. **Expected:** `[noorinalabs-main] DRIFT: required_pull_request_reviews: ...`.
7. **Actual:** `[noorinalabs-main] in sync`.

A second-order repro: that user can now run `gh pr merge --admin` (or use the UI "Merge without waiting for requirements") on any PR they author or co-author, bypassing the 2-reviewer rule, without triggering any audit signal.

## Fix shapes (pick one)

**Option A — preserve in `normalize_actual` and diff against `desired = null`:**

```python
rpr = actual.get("required_pull_request_reviews")
if rpr:
    norm["required_pull_request_reviews"] = {
        "required_approving_review_count": int(rpr.get("required_approving_review_count", 0)),
        "dismiss_stale_reviews": bool(rpr.get("dismiss_stale_reviews", False)),
        "require_code_owner_reviews": bool(rpr.get("require_code_owner_reviews", False)),
        "require_last_push_approval": bool(rpr.get("require_last_push_approval", False)),
        "bypass_pull_request_allowances": rpr.get("bypass_pull_request_allowances", {"users": [], "teams": [], "apps": []}),
    }
```

Then update `_common.required_pull_request_reviews` in `branch_protection_config.json` to include the same key with empty lists, so the diff detects any UI-added bypass.

**Option B — PUT explicit empty bypass list:**

Add `"bypass_pull_request_allowances": {"users": [], "teams": [], "apps": []}` to `_common.required_pull_request_reviews` in the manifest. The PUT body will then clear any UI-side bypass-allowance entries on every `--apply` run. Drift detection still won't flag the divergence between runs, but the cadence of `--apply` (manual + scheduled-eventual) will eventually correct it.

**Option A is preferred** because it surfaces drift *at the moment of audit* (which is the whole point of `--check`), rather than silently overwriting on the next apply (Option B). Both can be combined: PUT the empty list AND preserve+diff the field.

## Acceptance criteria

- [ ] `normalize_actual` preserves `bypass_pull_request_allowances` in the normalized output.
- [ ] `_common.required_pull_request_reviews` in `branch_protection_config.json` specifies the desired bypass-allowances state (empty list of users/teams/apps, or explicit allowed set with owner sign-off).
- [ ] New unit test in `test_apply_branch_protection.py` exercises: GET response with `bypass_pull_request_allowances.users=[<one user>]` against desired-empty produces a drift entry from `diff_state`.
- [ ] Post-merge runtime: confirm against live `gh api repos/noorinalabs/noorinalabs-main/branches/main/protection` that the bypass-allowance state is what the manifest declares.

## Out of scope

- The `dismissal_restrictions` field (a separate UI-configurable allow-list for who can dismiss reviews) is similarly stripped by the current `normalize_actual`. It is NOT a 2-reviewer-rule bypass surface but IS adjacent. If Option A is taken, consider preserving `dismissal_restrictions` in the same change.
- `restrictions` (push restrictions) is already preserved in `normalize_actual` and diffed; not affected by this issue.

## Severity

**Medium.** Mitigated today by the owner being the sole org admin (so the bypass surface requires owner action against themselves), but expands materially if/when delegated admins or admin-equivalent teams are introduced. The whole point of #432 was to remove an admin-bypass surface (`enforce_admins: false`) — this is a residual narrower instance of the same class.

- `gh issue list --search "bypass_pull_request_allowances"` → no matches.
- `gh issue list --search "normalize_actual drift"` → no matches.
- `gh issue list --search "branch protection drift"` → only #424 (W10 meta) and #429 (unrelated hook15 bug); no overlap.
- Rationalization scan (Pass B) → no open rationalization issue covers branch-protection drift detection. #403 (the originating "7 unprotected repos" issue) is closed by #432. #402 is auto-close-issues workflow propagation, different scope.
- Layer classification (Pass C) → SINGLE_LAYER — defect is in detection-layer code, not strategy or docs.

parametrization:
## Summary

Closes #433.

`apply_branch_protection.py` `normalize_actual()` was hand-extracting exactly 4 fields from the GET `required_pull_request_reviews` payload and dropping `bypass_pull_request_allowances`. Because the manifest's PUT body also omitted that key, a repo admin could UI-add a bypass actor and `--check` would silently report in-sync — defeating the 2-reviewer enforcement #432 was specifically introduced to harden.

**Defense-in-depth fix** (Options A + B from the issue, both surfaces closed):

- **A — preserve in `normalize_actual`:** `bypass_pull_request_allowances` AND adjacent `dismissal_restrictions` are now preserved, collapsing GET-side actor objects (`{login, id, type, ...}` / `{slug, id, name, ...}`) down to the PUT-shape string lists via new `_actor_logins`/`_actor_slugs` helpers. Sorted, so list-order from GitHub doesn't trigger spurious drift. Missing GET fields default to `{users:[], teams:[], apps:[]}` so older / never-configured states still compare equal to desired-empty.

- **B — PUT explicit empty list:** `_common.required_pull_request_reviews` in `branch_protection_config.json` now declares both fields as empty lists. `--apply` actively clears any UI-side bypass-actor entries on every run, on top of `--check` flagging them at audit time.

Together: a UI-added bypass actor is **detected** at audit (A) and **reverted** on next apply (B).

## Reviewers

- Bereket Tadesse (deploy manager, security/charter angle)
- Wanjiku Mwangi (TPM, process angle)

## Test plan

- [x] `python -m unittest test_apply_branch_protection` — 34 tests pass (was 30; +4 new)
  - `test_required_pull_request_reviews_strips_get_only_fields` extended: asserts `url` is stripped AND bypass/dismissal allow-lists are preserved
  - `test_required_pull_request_reviews_preserves_bypass_allowances` — GET omits bypass → normalize defaults to empty PUT shape (in-sync, not drift)
  - `test_bypass_allowances_drift_flagged_when_ui_adds_user` — **the #433 regression case**: UI-added user surfaces in `diff_state` output containing `rogue-admin`
  - `test_bypass_allowances_team_drift_flagged` — same class via team slug shape
  - `test_bypass_pull_request_allowances_empty_in_manifest` — schema check, every repo entry declares empty bypass + dismissal lists
- [x] `ruff check .claude/scripts/` — clean
- [x] `ruff format --check .claude/scripts/` — clean
- [x] CI `unit-tests` job in `branch-protection-audit.yml` runs the new tests (path-trigger covers all 3 modified files)
- [ ] **Runtime acceptance (post-merge, manual)**: trigger `branch-protection-audit.yml` with `mode=apply` against `noorinalabs-main`, then UI-add a user under Settings → Branches → main → "Allow specified actors to bypass required pull requests", and re-run with `mode=check`. Expect drift line mentioning the added user. Tracked separately per `feedback_runtime_gate_scoping` — PR acceptance is unit-mechanic correctness, not the live PUT loop.

## Acceptance criteria (from #433)

- [x] `normalize_actual` preserves `bypass_pull_request_allowances` in normalized output
- [x] `_common.required_pull_request_reviews` declares the desired bypass-allowances state (empty users/teams/apps)
- [x] New unit test exercises GET `bypass_pull_request_allowances.users=[<one user>]` against desired-empty → drift entry from `diff_state`
- [ ] Post-merge runtime confirmation (live `gh api ...protection`) — runtime gate, tracked above

### DIFF
--- a/.claude/scripts/apply_branch_protection.py
+++ b/.claude/scripts/apply_branch_protection.py
@@ -134,6 +134,16 @@ def fetch_actual(owner: str, repo: str, branch: str = DEFAULT_BRANCH) -> dict | 
     )
 
 
+def _actor_logins(actors: list) -> list[str]:
+    # GET returns list of {login, id, type, ...}; PUT accepts list of login strings.
+    return [a["login"] if isinstance(a, dict) else str(a) for a in actors]
+
+
+def _actor_slugs(actors: list) -> list[str]:
+    # Teams/apps GET returns list of {slug, id, name, ...}; PUT accepts slug strings.
+    return [a["slug"] if isinstance(a, dict) else str(a) for a in actors]
+
+
 def normalize_actual(actual: dict | None) -> dict:
     """Reduce the GET response to the same shape as the PUT body.
 
@@ -169,11 +179,41 @@ def normalize_actual(actual: dict | None) -> dict:
 
     rpr = actual.get("required_pull_request_reviews")
     if rpr:
+        # bypass_pull_request_allowances is a UI-configurable allow-list of
+        # actors that can merge without satisfying review requirements. It
+        # MUST be preserved so drift detection catches UI-side additions —
+        # otherwise the 2-reviewer gate is silently bypassable. Default to
+        # the empty PUT-shape (all three sub-lists empty) when the GET
+        # omits it. See noorinalabs/noorinalabs-main#433.
+        bypass = rpr.get(
+            "bypass_pull_request_allowances",
+            {"users": [], "teams": [], "apps": []},
+        )
+        # GET returns full actor objects (with id, login, type, etc.); the
+        # PUT body accepts a list of logins/slugs. Collapse to the PUT
+        # shape so the structural diff against the manifest is meaningful.
+        bypass_norm = {
+            "users": sorted(_actor_logins(bypass.get("users", []))),
+            "teams": sorted(_actor_slugs(bypass.get("teams", []))),
+            "apps": sorted(_actor_slugs(bypass.get("apps", []))),
+        }
+        # dismissal_restrictions has the same actor-list shape and is the
+        # adjacent allow-list controlling who can dismiss reviews. Same
+        # blind-spot class as bypass_pull_request_allowances; preserved
+        # together per #433 Out-of-Scope follow-up.
+        dr = rpr.get("dismissal_restrictions") or {}
+        dr_norm = {
+            "users": sorted(_actor_logins(dr.get("users", []))),
+            "teams": sorted(_actor_slugs(dr.get("teams", []))),
+            "apps": sorted(_actor_slugs(dr.get("apps", []))),
+        }
         norm["required_pull_request_reviews"] = {
             "required_approving_review_count": int(rpr.get("required_approving_review_count", 0)),
             "dismiss_stale_reviews": bool(rpr.get("dismiss_stale_reviews", False)),
             "require_code_owner_reviews": bool(rpr.get("require_code_owner_reviews", False)),
             "require_last_push_approval": bool(rpr.get("require_last_push_approval", False)),
+            "bypass_pull_request_allowances": bypass_norm,
+            "dismissal_restrictions": dr_norm,
         }
     else:
         norm["required_pull_request_reviews"] = None
--- a/.claude/scripts/branch_protection_config.json
+++ b/.claude/scripts/branch_protection_config.json
@@ -6,7 +6,17 @@
       "required_approving_review_count": 2,
       "dismiss_stale_reviews": true,
       "require_code_owner_reviews": false,
-      "require_last_push_approval": false
+      "require_last_push_approval": false,
+      "bypass_pull_request_allowances": {
+        "users": [],
+        "teams": [],
+        "apps": []
+      },
+      "dismissal_restrictions": {
+        "users": [],
+        "teams": [],
+        "apps": []
+      }
     },
     "restrictions": null,
     "allow_force_pushes": false,
--- a/.claude/scripts/test_apply_branch_protection.py
+++ b/.claude/scripts/test_apply_branch_protection.py
@@ -156,7 +156,9 @@ class TestNormalizeActual(unittest.TestCase):
 
     def test_required_pull_request_reviews_strips_get_only_fields(self) -> None:
         # GET wraps RPR with `url` and may carry extra keys; normalize must
-        # strip them so the diff matches a PUT-shape desired body.
+        # strip `url` so the diff matches a PUT-shape desired body, while
+        # preserving bypass_pull_request_allowances and dismissal_restrictions
+        # which are security-relevant actor allow-lists (#433).
         actual = {
             "required_status_checks": None,
             "enforce_admins": True,
@@ -170,7 +172,8 @@ class TestNormalizeActual(unittest.TestCase):
                 "require_code_owner_reviews": False,
                 "required_approving_review_count": 2,
                 "require_last_push_approval": False,
-                "dismissal_restrictions": {},
+                "dismissal_restrictions": {"users": [], "teams": [], "apps": []},
+                "bypass_pull_request_allowances": {"users": [], "teams": [], "apps": []},
             },
             "restrictions": None,
         }
@@ -182,7 +185,126 @@ class TestNormalizeActual(unittest.TestCase):
                 "dismiss_stale_reviews": True,
                 "require_code_owner_reviews": False,
                 "require_last_push_approval": False,
+                "bypass_pull_request_allowances": {"users": [], "teams": [], "apps": []},
+                "dismissal_restrictions": {"users": [], "teams": [], "apps": []},
             },
+        )
+        # `url` was stripped, the two allow-list fields were NOT.
+        self.assertNotIn("url", norm["required_pull_request_reviews"])
+
+    def test_required_pull_request_reviews_preserves_bypass_allowances(self) -> None:
+        # When GET omits bypass_pull_request_allowances (older shape or no
+        # bypass ever configured), normalize must still produce the empty
+        # PUT-shape so the diff against desired-empty is in-sync.
+        actual = {
+            "required_status_checks": None,
+            "enforce_admins": True,
+            "allow_force_pushes": False,
+            "allow_deletions": False,
+            "required_linear_history": False,
+            "required_conversation_resolution": False,
+            "required_pull_request_reviews": {
+                "required_approving_review_count": 2,
+                "dismiss_stale_reviews": True,
+                "require_code_owner_reviews": False,
+                "require_last_push_approval": False,
+            },
+            "restrictions": None,
+        }
+        norm = abp.normalize_actual(actual)
+        self.assertEqual(
+            norm["required_pull_request_reviews"]["bypass_pull_request_allowances"],
+            {"users": [], "teams": [], "apps": []},
+        )
+        self.assertEqual(
+            norm["required_pull_request_reviews"]["dismissal_restrictions"],
+            {"users": [], "teams": [], "apps": []},
+        )
+
+    def test_bypass_allowances_drift_flagged_when_ui_adds_user(self) -> None:
+        # The regression scenario from #433: a repo admin adds a bypass
+        # actor via the UI. GET returns the actor object; desired manifest
+        # has an empty bypass list. diff_state MUST flag drift on
+        # required_pull_request_reviews so --check exits non-zero.
+        actual = {
+            "required_status_checks": None,
+            "enforce_admins": True,
+            "allow_force_pushes": False,
+            "allow_deletions": False,
+            "required_linear_history": False,
+            "required_conversation_resolution": False,
+            "required_pull_request_reviews": {
+                "required_approving_review_count": 2,
+                "dismiss_stale_reviews": True,
+                "require_code_owner_reviews": False,
+                "require_last_push_approval": False,
+                "bypass_pull_request_allowances": {
+                    "users": [
+                        {"login": "rogue-admin", "id": 999, "type": "User"},
+                    ],
+                    "teams": [],
+                    "apps": [],
+                },
+                "dismissal_restrictions": {"users": [], "teams": [], "apps": []},
+            },
+            "restrictions": None,
+        }
+        actual_norm = abp.normalize_actual(actual)
+        # bypass-user collapsed to login string.
+        self.assertEqual(
+            actual_norm["required_pull_request_reviews"]["bypass_pull_request_allowances"]["users"],
+            ["rogue-admin"],
+        )
+        desired = {
+            "required_status_checks": None,
+            "enforce_admins": True,
+            "allow_force_pushes": False,
+            "allow_deletions": False,
+            "required_linear_history": False,
+            "required_conversation_resolution": False,
+            "required_pull_request_reviews": {
+                "required_approving_review_count": 2,
+                "dismiss_stale_reviews": True,
+                "require_code_owner_reviews": False,
+                "require_last_push_approval": False,
+                "bypass_pull_request_allowances": {"users": [], "teams": [], "apps": []},
+                "dismissal_restrictions": {"users": [], "teams": [], "apps": []},
+            },
+            "restrictions": None,
+        }
+        drift = abp.diff_state(desired, actual_norm)
+        self.assertEqual(len(drift), 1)
+        self.assertIn("required_pull_request_reviews", drift[0])
+        self.assertIn("rogue-admin", drift[0])
+
+    def test_bypass_allowances_team_drift_flagged(self) -> None:
+        # Same drift class but via a team allow-list addition (slug shape).
+        actual_rpr = {
+            "required_approving_review_count": 2,
+            "dismiss_stale_reviews": True,
+            "require_code_owner_reviews": False,
+            "require_last_push_approval": False,
+            "bypass_pull_request_allowances": {
+                "users": [],
+                "teams": [{"slug": "release-bots", "id": 7, "name": "Release Bots"}],
+                "apps": [],
+            },
+            "dismissal_restrictions": {"users": [], "teams": [], "apps": []},
+        }
+        actual = {
+            "required_status_checks": None,
+            "enforce_admins": True,
+            "allow_force_pushes": False,
+            "allow_deletions": False,
+            "required_linear_history": False,
+            "required_conversation_resolution": False,
+            "required_pull_request_reviews": actual_rpr,
+            "restrictions": None,
+        }
+        actual_norm = abp.normalize_actual(actual)
+        self.assertEqual(
+            actual_norm["required_pull_request_reviews"]["bypass_pull_request_allowances"]["teams"],
+            ["release-bots"],
         )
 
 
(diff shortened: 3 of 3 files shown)


parametrization:
Requestor: Bereket Tadesse
Requestee: Aisha Idrissi
RequestOrReplied: Approved
TechDebt: none

## Verdict: Approved

A+B closes the loop as designed. The blind spot Nino flagged in #432 and we then triaged into #433 is now closed at both surfaces — detection AND active reversion. Approving with three calibration notes (none blocking).

## Verification against artifact (HEAD `0dc02b5`)

Fetched all 3 files via `gh api contents?ref=<HEAD>` and diffed against base `deployments/phase-3/wave-10`:

- `.claude/scripts/apply_branch_protection.py` +40/-0: two new helpers (`_actor_logins`, `_actor_slugs`) + 28 lines of normalize_actual extension for `bypass_pull_request_allowances` and `dismissal_restrictions`.
- `.claude/scripts/branch_protection_config.json` +11/-1: `_common.required_pull_request_reviews` now declares both allow-lists as `{users:[], teams:[], apps:[]}`.
- `.claude/scripts/test_apply_branch_protection.py` +142/-2: 4 new tests as documented.

Ran the suite against PR head locally: **34 tests pass, 0 failures** (`python -m unittest test_apply_branch_protection`). Ran the 4 new tests explicitly with verbose output — all 4 pass.

`uvx ruff@0.13.3 check .claude/scripts/` → All checks passed. `uvx ruff@0.13.3 format --check .claude/scripts/` → 2 files already formatted. (Pinned per memory `feedback_ruff_format_check_before_push`; only 2 py files because the .json is not in scope for ruff format — that's expected, not a coverage gap.)

## Charter/security review — load-bearing questions

**Q1: Does `--apply` actively zero out a UI-added bypass entry, or only detect on `--check`?**

Yes, actively. Read trace:
1. `run_apply` (line 288) → `build_desired_for_repo(manifest, repo)` (line 292) merges `_common.required_pull_request_reviews` into desired — which now includes the explicit empty bypass + dismissal allow-lists from the manifest update.
2. `put_protection(owner, repo, desired)` (line 295) PUTs that body via `gh api -X PUT --input -`. The PUT shape includes both empty allow-lists, so GitHub clears any UI-added actors.
3. The subsequent `fetch_actual` + `normalize_actual` + `diff_state` re-verifies post-state.

So if a rogue admin UI-adds a bypass user, the next scheduled `--apply` run reverts them, and any intermediate `--check` run flags drift. **Both surfaces closed as the PR body claims.**

**Q2: Does `_actor_logins` / `_actor_slugs` preserve case sensitivity correctly?**

The helpers extract `a["login"]` and `a["slug"]` verbatim — no casefolding. Sort order is stable on identical strings. For the current manifest state (all-empty allow-lists), this is irrelevant since there's nothing to match. For the future case where the owner *intentionally* adds a bypass actor to the manifest, the literal-string match against GitHub's GET will work as long as the manifest declares the login in GitHub's canonical case form. GitHub's API generally preserves user-supplied case in `login`. Worth a docstring note when the first bypass actor is intentionally added to the manifest, but not a blocker for this PR — the empty-list state has no case concern.

**Q3: Does the scope expansion to `dismissal_restrictions` introduce any unintended behavior change for repos with existing UI-configured dismissal_restrictions?**

Verified by hitting all 8 repos live:
```bash
for repo in noorinalabs-{main,isnad-graph,deploy,user-service,design-system,landing-page,data-acquisition,isnad-ingest-platform}; do
  gh api repos/noorinalabs/$repo/branches/main/protection | jq .required_pull_request_reviews
done
```

All 8 repos return GET with `bypass_pull_request_allowances` and `dismissal_restrictions` **absent** from the payload entirely (no key at all, not even an empty object). Aisha's `bypass = rpr.get("bypass_pull_request_allowances", {"users":[], "teams":[], "apps":[]})` default-arg and `dr = rpr.get("dismissal_restrictions") or {}` or-coalesce both correctly handle that absent case → empty PUT shape → matches desired-empty → in-sync. **No repo currently has a UI-configured dismissal_restrictions, so the scope expansion introduces zero behavior change in the live audit.** It only adds future protection.

**Q4: "Out-of-Scope note suggested *considering*" — was Aisha right to include dismissal_restrictions in the same PR?**

Yes. Three reasons:
- The PR author explicitly flagged it as "reversible if undesired" — signaling awareness that this is a scope-judgment call.
- `dismissal_restrictions` is structurally identical (same actor-list shape, same UI-side editability, same blind-spot class). Splitting would mean re-doing the same `_actor_logins` collapse, the same test pattern, the same manifest declaration in a follow-up PR — pure churn for no isolation benefit.
- Live-state survey above proves zero behavior change risk (no repo has it configured today). The downside surface of including it is empty; the upside is closing the sibling blind-spot in one wave instead of waiting for it to be exploited.

This is the inverse of the "premature scope expansion" anti-pattern — Aisha's expansion is structurally identical to the in-scope work, with verified zero blast radius. Calling it correct judgment.

## Three calibration notes (non-blocking)

1. **Style asymmetry between bypass and dismissal handling:** `bypass = rpr.get("bypass_pull_request_allowances", {default})` uses default-arg, while `dr = rpr.get("dismissal_restrictions") or {}` uses or-coalesce. Both reach the same outcome via the subsequent `.get("users", [])` fallbacks, but the inconsistency is mildly awkward to read. Pick one form for both in a future cleanup if you touch this code again. Not worth a follow-up PR.

2. **Runtime acceptance is correctly deferred per `feedback_runtime_gate_scoping`.** The PR-time acceptance is unit-mechanic correctness (passed), and the post-merge UI-add-actor-then-recheck test plan item is appropriately tracked separately. This is the right pattern.

3. **Memory promotion candidate:** the disciplined "verify the live state of all 8 repos before claiming scope-expansion is zero-risk" pass I just did took ~30s and gave certainty on Q3. This is `feedback_origin_over_local_for_still_has_claims` applied to live API state instead of repo content — same primitive, different artifact class. If this discipline recurs as load-bearing, it's promotion-worthy as a sibling charter rule. Logging for retro consideration.

## Throughline observations

The #432 → #433 → #435 sequence is a textbook "security fix surfaces blind spot, fix-of-fix continues hardening" cascade. Two observations worth surfacing for the retro:

1. **`normalize_actual` is becoming a hot-path for security-relevant blind spots.** The original #432 strip of `dismissal_restrictions` was correct-as-far-as-it-went (avoided perma-drift on GET-vs-PUT shape mismatch), but a hand-extract-4-fields approach is structurally prone to dropping security-relevant fields by omission. #433 (this PR) fixes the blind spot, but the underlying pattern — "explicit field allowlist in a normalizer, defaulting to discard" — will produce another blind spot the next time GitHub adds a field to the GET response that's UI-editable but PUT-omitted. Promotion candidate for a future memory: "GET-to-PUT normalizers should preserve-by-default, discard-by-explicit-list — not the other way around." Not for this PR; flag for whoever next touches normalize_actual.

2. **Aisha's "expand scope to sibling blind-spot in same PR with reversibility note" pattern is a reusable convention.** She did the structural-similarity check (same actor-list shape) + live-state-zero-impact check (no repo has it configured) before expanding, then flagged "reversible if undesired" as a forward signal to reviewers. That's a more efficient pattern than the default "minimum-PR-scope" rule when (a) the sibling is structurally identical and (b) live-state impact is verifiable as zero. Worth surfacing as a charter calibration: minimum-PR-scope is a default, not an absolute, and the conditions for principled expansion are verifiable. Not yet promotion-ready (single instance), but log for retro.

Wave-level signal: this is the third PR in the wave (after #431 + the auto-close propagation) that's the load-bearing application of a memory-flagged blind spot from a prior wave. The team is converging on a "blind-spot triage → next-wave fix" cadence that's working — no wave-level structural defect observed.

parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aisha Idrissi
RequestOrReplied: Approved
TechDebt: none

## TPM-angle review — process + scope discipline

Reviewed at head_sha `0dc02b5f1344f951eecc8b024bd454cc05b5742d` against base `b7d7437` (post-#434 merge). 2/2 CI checks: `Unit tests — apply_branch_protection` SUCCESS; `Branch-protection drift check` SKIPPED (path-trigger does not include manifest changes on PR diff — runtime gate, separately tracked in Test Plan).

### A+B scope discipline — both surfaces honored, no scope creep

The PR body declared "Options A + B from the issue, both surfaces closed." Reading the diff at HEAD confirms exact execution:

- **A (preserve in normalize_actual):** `apply_branch_protection.py:179-211` adds the two preservation blocks AND the two `_actor_logins`/`_actor_slugs` helpers at line 137-144. Defaults match PUT-shape `{users:[], teams:[], apps:[]}` when GET omits the keys — so historical configs without ever-set bypass don't flag as drift.
- **B (PUT empty list in manifest):** `branch_protection_config.json:9-19` declares both fields explicitly empty in `_common.required_pull_request_reviews`. Every repo entry inherits via the common section (test `test_bypass_pull_request_allowances_empty_in_manifest` iterates all repos and asserts schema).

Combined: a UI-added bypass actor is BOTH detected at audit time AND reverted on next `--apply`. Layered defense, not just one or the other. This matches the spawn-brief A+B requirement exactly.

### Sibling-class scope expansion: justified by symmetry

PR scope expanded beyond the literal #433 request by also handling `dismissal_restrictions`. This is the right call:
- Same actor-list shape (`{users, teams, apps}` with `login`/`slug` collapsing)
- Same UI blind-spot class (admin can UI-add a user who can dismiss reviews, which is the structurally-equivalent bypass)
- Same `_actor_logins`/`_actor_slugs` helpers handle both — zero additional surface cost

Without dismissal_restrictions, a rogue-admin could add a user who dismisses approvals to force re-review (a different attack on the 2-reviewer gate). Including it closes the sibling vulnerability. Scope expansion justified by symmetric-vulnerability-class evidence, not aspirational completeness.

### #433 regression repro is the actual test

`test_bypass_allowances_drift_flagged_when_ui_adds_user` is exactly the regression scenario, end-to-end:
1. Build `actual` with GET-shape bypass actor `{login: "rogue-admin", id: 999, type: "User"}` (the literal #433 attack)
2. Call `normalize_actual`
3. Assert the actor collapsed to `["rogue-admin"]` string (PUT-shape parity)
4. Build `desired` with empty bypass per the manifest
5. Call `diff_state(desired, actual_norm)` — the function `--check` calls
6. Assert `len(drift) == 1`, contains `required_pull_request_reviews`, contains `"rogue-admin"` (the actor name appears in operator-visible drift output)

This is the kind of test that would have caught the original blind spot — it exercises the actual code path `--check` runs, with the literal payload an attacker would inject. Plus `test_bypass_allowances_team_drift_flagged` covers the slug-shape variant (team allow-list addition). 4 new tests total: bypass-empty-default, user-drift-flagged, team-drift-flagged, manifest-schema. Coverage matches the AC list in the issue.

### Defensive shape: sort + collapse

`_actor_logins`/`_actor_slugs` collapse GET-shape dicts to PUT-shape strings (`a["login"] if isinstance(a, dict) else str(a)`); the `sorted(...)` wrapper means list-order from GitHub never triggers spurious drift. The `isinstance(a, dict)` fallback to `str(a)` covers the edge case where GH returns a raw string (unlikely but safe). This is the kind of small detail that prevents the "fix introduced a new false-positive class" follow-up.

### Process notes

- Branch name `A.Idrissi/0433-normalize-actual-bypass-fix` follows convention
- Commit identity correct via per-commit `-c` flags (verified in commits API: `parametrization+Aisha.Idrissi@gmail.com`)
- Single commit, focused
- 2 reviewers named: Bereket (deploy/security) + Wanjiku (TPM/process) — peers, neither is the author
- Test Plan honestly distinguishes PR-acceptance (unit mechanics) vs runtime acceptance (live UI manipulation), per `feedback_runtime_gate_scoping`. Runtime gate is correctly carved out, not pretended-to-cover.

The PUT body issued by `--apply` will now ALWAYS include `bypass_pull_request_allowances` and `dismissal_restrictions` as explicit empty objects, even on a repo that's never had bypass configured. This is correct semantics (idempotent clearing), but if a repo's branch-protection state has any other field divergence and the operator just wants a non-bypass change, the PUT now also clears UI-side bypass state implicitly. That's the DESIGNED behavior, not a regression — the manifest is now the single source of truth for these fields. Calling it out because it's a small expansion of `--apply`'s effective surface that the audit-runner ops should understand.

## Throughline observations

Three patterns visible across this PR + the wider W10 cluster:

1. **`normalize_actual` was a single point of trust failure for security-relevant fields.** The function's pre-#433 design — hand-extract a fixed field list — meant every new security-relevant GitHub API surface required a manual update or it became a silent blind spot. Same shape as the `_SOURCE_HINT_RE` issue in #419 (hand-tuned regex that missed real cases) and the Hook 15 sentinel-formula issue in #429 (canonical helper that ad-hoc writers diverged from). **Throughline:** hand-curated detection surfaces accumulate silent gaps; the W10 wave keeps fixing instances of this class. Promotion candidate: when designing a "structural diff against an external API," prefer a deny-list over an allow-list, and CI-test that all known security-relevant fields ARE in the diff. Memory `[[feedback_enforcement_hierarchy]]` already says "hook > skill > charter" — extend to "deny-list > allow-list" for detection surfaces.

2. **Runtime acceptance correctly carved out, per `feedback_runtime_gate_scoping`.** The Test Plan explicitly bullets the live UI manipulation as runtime gate, not PR acceptance — exactly the shape Bereket endorsed on deploy#187 (2026-04-28). This is the second clean instance of that memory being applied without prompting in P3W10 (after #422). The pattern holds: PR-acceptance = unit mechanics that prove the diff does what it claims; runtime acceptance = the live loop, post-merge, manually traced. Skill/charter promotion candidate if it sustains through W11.

3. **Defense-in-depth bundled in single PR.** Options A and B are conceptually two separate fixes (one detection, one enforcement). Bundling them in this PR — rather than filing A as a fast fix and B as a follow-up — is correct because they share the same diff context and a partial fix (A only) would leave the UI-added actor detectable-but-uncleared. Same shape as #429 bundling diagnostic + tolerance + tests. **Convergent class:** P3W10 retro fixes tend to bundle multiple defensive layers when they share artifact scope, rather than the older P2-era pattern of one-issue-one-PR. Worth noting in retro as evolved practice.

Issue #419, opened 2026-05-13
tech-debt(skill): /promotion-audit _SOURCE_HINT_RE matches URL path fragments as slash-commands (11 false positives at HEAD)


parametrization:
## Surface

`_SOURCE_HINT_RE` in `.claude/skills/promotion-audit/helpers.py:429` matches slash-prefixed tokens too aggressively, producing URL-path-fragment false positives in `find_already_promoted_in_charter()`.

Current regex:

```python
_SOURCE_HINT_RE = re.compile(
    r"""
    (?:
        feedback_[a-z0-9_]+\.md       # feedback memory filenames
        |
        project_[a-z0-9_]+\.md        # project memory filenames
        |
        reference_[a-z0-9_]+\.md      # reference memory filenames
        |
        /[\w-]+                       # slash-commands / skill names
    )
    """,
    re.VERBOSE,
)
```

The `/[\w-]+` branch matches anything starting with `/` and containing word chars or hyphens. In charter markdown bodies, that includes URL path fragments inside link references like `[#198](https://github.com/noorinalabs/noorinalabs-main/issues/198)` — `/198`, `/issues`, `/noorinalabs`, `/noorinalabs-main`, `/Edit` all get captured.

Worked-example evidence (P3W9 retro 2026-05-12, against `.claude/team` at wave-9 HEAD):

```
find_already_promoted_in_charter returned: 38 entries
Real promotion refs: 27
False positives (URL fragments): 11
  - /198, /200, /203, /244, /337   (issue numbers from gh URLs)
  - /Edit                          (path fragment somewhere)
  - /github                        (from github.com URLs)
  - /issues                        (gh URL path)
  - /noorinalabs, /noorinalabs-main (org + repo from gh URLs)
  - /supersedes                    (legitimate? worth verifying — appears related to `supersedes:` frontmatter field)
```

The false positives don't currently flip AUTO/DECIDE classifications (since `already_promoted` only acts as a tie-breaker against real memory filenames), but they pollute the `already_promoted` set and could cause silent collisions if a future skill name overlaps.

## Fix shape

Tighten the `/[\w-]+` branch to exclude URL-fragment shapes:

1. **Require alpha first char + min length:** `/[a-z][a-z0-9-]{2,}` — slash-commands are kebab-case starting with a letter, never digits.
[...]

Issue #417, opened 2026-05-13
tech-debt(skill): /promotion-audit SKILL.md prose drift — classify() vs classify_memory/_section/_skill + missing signal-dict example


parametrization:
## Surface

`/promotion-audit` SKILL.md instructions block references `classify()` but the actual `helpers.py` exposes `classify_memory()`, `classify_section()`, and `classify_skill()` — three separate functions with distinct signatures. SKILL.md prose example at lines 31-44 also does not show the `signals: dict[str, int]` argument the classifiers require.

Worked example surface (P3W9 retro 2026-05-12): caller script copy-pasted from SKILL.md prose passed `cite` (int) where `signals` (dict) was expected, hitting `AttributeError: 'int' object has no attribute 'get'` at helpers.py:672. Retry with `{'retro_citations': cite}` succeeded.

## Fix shape

Update `.claude/skills/promotion-audit/SKILL.md` Step 2 + 3 prose to:

1. List the actual classifier function names — `classify_memory`, `classify_section`, `classify_skill` — not a single `classify()`.
2. Show the `signals: dict[str, int]` argument shape in the example block, with the expected keys per classifier:
   - `classify_memory` → `{'retro_citations': <int>}`
   - `classify_section` → `{'skill_invocations': <int>}`
   - `classify_skill` → `{'skill_invocations': <int>}`

Optional: extract a single `classify(item, signals, already_promoted) -> Decision` dispatcher that routes by `isinstance` and update SKILL.md to match the simplified surface.

## Acceptance

- SKILL.md prose contains the actual classifier function names
- Example code block in Step 2 includes a working `signals=` argument
- A reader following SKILL.md verbatim can run the audit without hitting the int-vs-dict mismatch

## Why

Three-way prose drift between SKILL.md, helpers.py, and the test fixtures. Cost of one wave's retro hitting this: a wasted Python invocation cycle. Cost of leaving it: every future operator running the skill hits the same trap.

## Wave

p3-wave-10 (skill cleanup batch with siblings).

parametrization:
Requestor: Fatima Okonkwo
Requestee: Aino Virtanen
RequestOrReplied: Request

**Wave 10 Kickoff — Phase 3**

This issue is assigned to you for p3-wave-10.
- Peer reviewer: Nadia Khoury
- Secondary reviewer: Wanjiku Mwangi
- Branch from: `deployments/phase-3/wave-10`
[...]

Issue #283, opened 2026-05-06
tech-debt(/promotion-audit): find_already_promoted misses charter-tier promotions (only scans hooks.md)


parametrization:
Surfaced during P3W5 promotion-audit execution (PR #282).

## Problem

`find_already_promoted()` in `.claude/skills/promotion-audit/helpers.py` only scans `charter/hooks.md` for `Promotion provenance:` blocks. It returns the set of memory filenames / skill slash-commands that have been promoted to **hook tier**.

Charter-tier-only promotions (memory → charter without a corresponding hook) are invisible to the parser. As a result, those memories surface as `AUTO` on every audit run, even when they're already canonically codified in charter.

## Concrete miss in W5 audit

W5 audit produced `5 AUTO`. While preparing the AUTO promotion PR (#282), I discovered:

- `feedback_review_against_artifact_not_framing` was already at `pull-requests.md` § Trust the Artifact, Not the Framing (L325) — never had a hook, so the parser never saw it
- `feedback_honest_audit_over_conclusion_claim` was at `skills.md` § Wave Lifecycle — Open-Item Audit AND at `hooks.md` L169 (Hook 17). The hooks.md provenance line uses the memory name without the `.md` suffix, so even the hook-tier provenance match failed.

Net: 2 of 5 "AUTO" candidates were tooling false positives.

## Proposed fix

1. Extend `find_already_promoted()` to also scan `charter/*.md` (not just `charter/hooks.md`) for `Promoted from memory:` HTML-comment markers (the convention used by PR #282's new charter sections + retroactive markers).

2. Loosen the regex to match memory references without the `.md` suffix (the hooks.md L169 provenance for honest_audit lacks it).

## Acceptance

After fix lands, re-running `/promotion-audit` against current main should produce the same counts as W5's post-PR-282 re-audit:
```
ALREADY-PROMOTED: 1+ (was 1)
AUTO: 0
SUPERSEDED: 16
KEPT: 52
```

Specifically: zero AUTO, no false positives.

## Severity

Operational — current parser produces correct outcome via human review (PR #282 caught the misses), but the audit is no longer byte-deterministic across re-runs without manual intervention. Charter-tier promotions accumulate over time; the parser gap will widen.

## Related

- PR #282 (P3W5 promotion AUTO) — surfaced this gap
- PR #281 (P3W5 retro) — has the audit log showing the original 5/0/52/11/1 counts
[...]

PR #436, opened 2026-05-16
tech-debt(/promotion-audit): rename charter_root → charter_parent + raise on charter-dir input (#418)

### LINKED ISSUE #418: tech-debt(skill): /promotion-audit find_already_promoted_in_charter(charter_root) param confusingly takes parent-of-charter
## Surface

`find_already_promoted_in_charter(charter_root: str)` in `.claude/skills/promotion-audit/helpers.py:581` takes a parameter named `charter_root` that is confusingly the **parent** of the charter directory, not the charter directory itself. Implementation:

```python
subdir = os.path.join(charter_root, "charter")
if os.path.isdir(subdir):
    for name in sorted(os.listdir(subdir)):
        if name.endswith(".md"):
            candidates.append(os.path.join(subdir, name))
```

So the correct call is `find_already_promoted_in_charter('.claude/team')` (which then walks `.claude/team/charter/*.md`), NOT `find_already_promoted_in_charter('.claude/team/charter')`.

Worked example (P3W9 retro 2026-05-12): caller passed `.claude/team/charter` (the natural reading of `charter_root`). Function silently returned `set()` (0 entries) — the function found no `<root>/charter/` subdir under that path. Audit completed with `already_promoted=0` and looked correct because the `status: superseded` frontmatter path dominates AUTO/DECIDE decisions independently — but the parser-state was wrong and any test that relies on already-promoted detection would silently false-negative.

## Fix shape

Either:

(a) **Rename the parameter** to `repo_root` or `charter_parent` so the call shape is unambiguous. Update SKILL.md prose example to match.

(b) **Accept both forms** by detecting at runtime: if `os.path.isdir(charter_root + '/charter')` then walk that; else if `os.path.basename(charter_root) == 'charter'` and `os.path.isdir(charter_root)` then walk that directly. Either input works without surprises.

Recommend (a) — explicit naming is cheaper than implicit dual-form handling, and the function is only called from SKILL.md prose + one test, so the rename blast radius is small.

## Acceptance

- Parameter renamed (or dual-form accepted)
- SKILL.md Step 2 example block uses the unambiguous shape
- Test added: `find_already_promoted_in_charter(<wrong-form>)` either raises clearly OR works as expected
- Verified against an actual charter at HEAD that the function returns >0 entries (current verified count: 38 — 29 real + 9 URL-fragment false positives per the sibling #_SOURCE_HINT_RE bug)

## Why

Silent-zero results in a deterministic-classification skill are dangerous: they look correct until they don't. Sibling of `feedback_gh_pr_edit_silent_noop.md` family — silent no-op on bad input.

## Wave

p3-wave-10 (skill cleanup batch with siblings).

parametrization:
## Summary

`find_already_promoted_in_charter(charter_root)` and the sibling `read_all_charter_sections(charter_root)` in `.claude/skills/promotion-audit/helpers.py` both take a parameter named `charter_root` that is the **parent** of the `charter/` subdir, not the charter dir itself. Implementations do `os.path.join(arg, "charter")`. Passing the natural reading (`.claude/team/charter`) silently returns empty.

- Rename param `charter_root` -> `charter_parent` on both functions; tighten docstrings
- Runtime guard: `ValueError` with `#418` backreference if argument is itself a dir named `charter` (loud-fail the most likely caller mistake; companion to the `feedback_gh_pr_edit_silent_noop.md` family)
- `SKILL.md` Step 2 example block updated to use the unambiguous shape with an inline NOTE
- Two new tests: wrong-form raises; sibling function also raises (parity)

## Related Issues

Closes #418.

Sibling cluster (parallel, no overlap):
- #417 — SKILL.md prose drift (Aino)
- #419 — `_SOURCE_HINT_RE` URL-fragment false positives (Aino)
- #283 — original aggregator (in-tree, referenced for context)

## Verification

At HEAD (`b7d7437`):
- Correct form (`.claude/team`): **42 refs** (29 real + URL-fragment false positives still present — addressed by Aino's parallel #419)
- Wrong form (`.claude/team/charter`): now raises `ValueError` with `#418` backreference
- Full test suite: 69 tests pass (67 prior + 2 new for #418 guard + sibling parity)
- `uvx ruff@0.13.0 format --check .claude/skills/promotion-audit/` clean

## Review Checklist

- [ ] Rename does not break any other caller (grep verified: 3 call sites — `SKILL.md` + `test_smoke.py` + `test_helpers.py`, all updated)
- [ ] Sibling function `read_all_charter_sections` correctly receives the same fix shape (parity)
- [ ] Error message points the caller at the exact mistake (charter dir vs parent) AND cites the issue for context
- [ ] SKILL.md example uses the new param name with the inline NOTE that pins semantics

## TechDebt

This is itself a tech-debt fix (silent-no-op family); no new tech debt introduced.

Co-Authored-By: Nadia Khoury <parametrization+Nadia.Khoury@gmail.com>
Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/.claude/skills/promotion-audit/SKILL.md
+++ b/.claude/skills/promotion-audit/SKILL.md
@@ -40,10 +40,12 @@ from helpers import (
     render_audit_table,
 )
 
-memories  = read_all_memories(memory_dir)              # list[Memory]
-sections  = read_all_charter_sections(charter_dir)     # list[Section] — only sections with a promotion-target marker
-skills    = read_all_skills(skills_dir)                # list[Skill]
-already   = find_already_promoted_in_charter(charter_dir)  # set[str] — aggregates Promotion provenance: blocks AND <!-- Promoted from memory: X --> markers across all charter sub-docs (#283)
+memories  = read_all_memories(memory_dir)                  # list[Memory]
+sections  = read_all_charter_sections(charter_parent)      # list[Section] — only sections with a promotion-target marker
+skills    = read_all_skills(skills_dir)                    # list[Skill]
+already   = find_already_promoted_in_charter(charter_parent)  # set[str] — aggregates Promotion provenance: blocks AND <!-- Promoted from memory: X --> markers across all charter sub-docs (#283)
+# NOTE: `charter_parent` is the directory CONTAINING `charter/` (e.g. `.claude/team`),
+# NOT the `charter/` directory itself. Passing `.claude/team/charter` raises ValueError (#418).
 ```
 
 ### 3. Classify each candidate (pure function)
--- a/.claude/skills/promotion-audit/helpers.py
+++ b/.claude/skills/promotion-audit/helpers.py
@@ -298,17 +298,36 @@ def read_charter_sections(charter_path: str) -> list[CharterSection]:
     return results
 
 
-def read_all_charter_sections(charter_root: str) -> list[CharterSection]:
+def read_all_charter_sections(charter_parent: str) -> list[CharterSection]:
     """Scan charter.md + charter/*.md for marked sections.
 
+    `charter_parent` is the directory **containing** the `charter/` subdir
+    (typically `.claude/team`), NOT the `charter/` directory itself. Sibling
+    of `find_already_promoted_in_charter` — same parameter semantics, same
+    silent-empty failure mode if the caller passes the charter dir instead
+    of its parent (issue #418).
+
     Sorted by (path, heading) for determinism.
+
+    Raises:
+        ValueError: if `charter_parent` is itself named `charter` — see #418.
     """
+    if os.path.isdir(charter_parent) and (
+        os.path.basename(os.path.normpath(charter_parent)) == "charter"
+    ):
+        raise ValueError(
+            f"read_all_charter_sections({charter_parent!r}): "
+            "argument is the charter directory itself — pass its parent "
+            "(e.g. '.claude/team', not '.claude/team/charter'). "
+            "See issue #418."
+        )
+
     candidates: list[str] = []
-    root_file = os.path.join(charter_root, "charter.md")
+    root_file = os.path.join(charter_parent, "charter.md")
     if os.path.isfile(root_file):
         candidates.append(root_file)
 
-    subdir = os.path.join(charter_root, "charter")
+    subdir = os.path.join(charter_parent, "charter")
     if os.path.isdir(subdir):
         for name in sorted(os.listdir(subdir)):
             if name.endswith(".md"):
@@ -578,27 +597,45 @@ def find_already_promoted(charter_path: str) -> set[str]:
     return refs
 
 
-def find_already_promoted_in_charter(charter_root: str) -> set[str]:
+def find_already_promoted_in_charter(charter_parent: str) -> set[str]:
     """Aggregate already-promoted refs across the full charter directory.
 
-    Scans `charter_root/charter/*.md` (and the optional `charter_root/charter.md`
-    top-level file, if present) using the same recognition rules as
-    `find_already_promoted()`. Returns the union of all per-file results.
+    `charter_parent` is the directory **containing** the `charter/` subdir
+    (typically `.claude/team`), NOT the `charter/` directory itself.
+
+    Scans `<charter_parent>/charter/*.md` (and the optional
+    `<charter_parent>/charter.md` top-level file, if present) using the same
+    recognition rules as `find_already_promoted()`. Returns the union of
+    all per-file results.
 
     This is the entry point the /promotion-audit skill should use — single-
     file scope (charter/hooks.md only) misses charter-tier-only promotions
     that land via the HTML-comment marker in other sub-docs (issue #283).
+
+    Raises:
+        ValueError: if `charter_parent` is itself a directory named `charter`
+            — a strong hint the caller passed the charter dir instead of its
+            parent (issue #418 silent-zero bug). Returns set() for any other
+            non-existent path (defensive contract preserved).
     """
     refs: set[str] = set()
-    if not os.path.isdir(charter_root):
+    if not os.path.isdir(charter_parent):
         return refs
 
+    if os.path.basename(os.path.normpath(charter_parent)) == "charter":
+        raise ValueError(
+            f"find_already_promoted_in_charter({charter_parent!r}): "
+            "argument is the charter directory itself — pass its parent "
+            "(e.g. '.claude/team', not '.claude/team/charter'). "
+            "See issue #418."
+        )
+
     candidates: list[str] = []
-    root_file = os.path.join(charter_root, "charter.md")
+    root_file = os.path.join(charter_parent, "charter.md")
     if os.path.isfile(root_file):
         candidates.append(root_file)
 
-    subdir = os.path.join(charter_root, "charter")
+    subdir = os.path.join(charter_parent, "charter")
     if os.path.isdir(subdir):
         for name in sorted(os.listdir(subdir)):
             if name.endswith(".md"):
--- a/.claude/skills/promotion-audit/tests/test_helpers.py
+++ b/.claude/skills/promotion-audit/tests/test_helpers.py
@@ -517,6 +517,40 @@ class FindAlreadyPromotedTests(unittest.TestCase):
         refs = h.find_already_promoted_in_charter("/nonexistent/path/to/charter")
         self.assertEqual(refs, set())
 
+    def test_aggregator_raises_on_charter_dir_itself(self) -> None:
+        """NEG (#418): passing the charter directory itself (e.g.
+        `.claude/team/charter`) instead of its parent must raise
+        ValueError. Previously this silently returned set() because no
+        `<root>/charter/charter/` subdir exists — the silent-zero failure
+        mode that #418 was filed to fix.
+        """
+        with tempfile.TemporaryDirectory() as tmpdir:
+            charter_subdir = os.path.join(tmpdir, "charter")
+            os.makedirs(charter_subdir)
+            # Correct form: parent of charter/ — returns set() (empty corpus).
+            self.assertEqual(h.find_already_promoted_in_charter(tmpdir), set())
+            # Wrong form: the charter dir itself — must raise loudly.
+            with self.assertRaises(ValueError) as ctx:
+                h.find_already_promoted_in_charter(charter_subdir)
+            self.assertIn("#418", str(ctx.exception))
+
+            # Trailing slash form must still raise — normpath strips it.
+            with self.assertRaises(ValueError):
+                h.find_already_promoted_in_charter(charter_subdir + "/")
+
+    def test_read_all_charter_sections_raises_on_charter_dir_itself(self) -> None:
+        """NEG (#418, sibling): `read_all_charter_sections` has identical
+        parent-of-charter semantics and the same silent-empty failure mode.
+        Passing the charter dir itself must raise.
+        """
+        with tempfile.TemporaryDirectory() as tmpdir:
+            charter_subdir = os.path.join(tmpdir, "charter")
+            os.makedirs(charter_subdir)
+            # Correct form: returns empty list (no marked sections in empty corpus).
+            self.assertEqual(h.read_all_charter_sections(tmpdir), [])
+            with self.assertRaises(ValueError):
+                h.read_all_charter_sections(charter_subdir)
+
 
 # ---------------------------------------------------------------------------
 # Classification — memory


parametrization:
Requestor: Santiago Ferreira
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: none

## Release-coordinator + skill/charter review

Read at HEAD `58b6c1a` via `gh api .../contents/...?ref=<sha>`. CI 4/4 SUCCESS. mergeable_state `clean`.

### Loud-fail vs deterministic-classification — ALIGNED

The `ValueError` guard raised when callers pass `.claude/team/charter` (the charter dir itself, not its parent) is the right shape for a deterministic-classification skill. Three observations:

1. **The pre-existing silent-set() return on non-existent paths is preserved** (`if not os.path.isdir(charter_parent): return refs` — line 593-594 region). The new ValueError is ONLY raised when the input IS a directory AND `os.path.basename(os.path.normpath(...)) == "charter"`. So the defensive `set()` contract for "no such corpus" is untouched, and the loud-fail only catches the specific #418 caller mistake. Good narrow scope.
2. **Trailing-slash case is covered by `os.path.normpath`** — test `test_aggregator_raises_on_charter_dir_itself` explicitly asserts both `<dir>/charter` and `<dir>/charter/` raise. `normpath` strips the trailing slash before `basename` checks. Verified by independent regex inspection.
3. **Error message cites the issue + names both correct and incorrect forms** — `"argument is the charter directory itself — pass its parent (e.g. '.claude/team', not '.claude/team/charter'). See issue #418."` This is the right shape for a charter-pinned loud-fail: caller sees the fix AND the provenance trail in one breath.

### Caller-sweep verification

PR body claims "3 call sites: `SKILL.md` + `test_smoke.py` + `test_helpers.py`, all updated." Verified at HEAD:

- `SKILL.md:44, 46`: uses `charter_parent` post-rename + carries the inline NOTE.
- `tests/test_smoke.py:52, 54`: passes `str(_CHARTER_ROOT)` where `_CHARTER_ROOT = _REPO_ROOT / ".claude" / "team"` (line 32). The local variable name is `_CHARTER_ROOT` but the VALUE is `.claude/team` (the parent of charter/) — semantically correct. Renaming the local variable is cosmetic and not required for correctness; leaving it as `_CHARTER_ROOT` is a defensible choice since it pre-dates the param rename. Non-blocking observation.
- `tests/test_helpers.py:478-551`: new `test_aggregator_raises_on_charter_dir_itself` + sibling-parity `test_read_all_charter_sections_raises_on_charter_dir_itself`. Both assert `#418` literal appears in the ValueError message.

No other in-tree callers found via `gh api tree?recursive=1` scan. The `.claude/scratch/promotion-audit-w7.py` use of `charter_root` as a LOCAL variable name (line 24) is harmless — it's not the parameter, just an ad-hoc scratch script that happens to use the same name as the old param. Scratch dir is operator-facing only.

### Sibling-fix parity — VERIFIED

`read_all_charter_sections` (line 301) gets the SAME parameter rename + SAME ValueError guard + SAME error message shape as the aggregator (line 597). Sibling parity is the correct call: both functions share the silent-empty failure mode and the same charter/parent semantics, so loud-fail at one but silent-empty at the other would create asymmetric debugging surface. Aino's #429 PR (just merged) sets the precedent that defense-in-depth at the helper boundary is the right pattern for the audit pipeline.

### Cross-PR coordination with #437

Verified non-overlap with PR 437's `helpers.py` changes. PR 437 hunks live in lines 440-525 (regex region) and 595-615 (call-site URL strip), per `gh api .../files`. PR 436 hunks live at 301 (sibling rename) and 597 (aggregator signature + ValueError guard). Git 3-way merge will resolve cleanly because the diff regions are non-textually-overlapping. Whichever PR merges first, the second will rebase trivially.

**Heads-up to merger:** PR 437's `find_already_promoted_in_charter` docstring still references `charter_root` as the param name (PR 437 predates this rename). After PR 436 merges to wave-10, a rebase of PR 437 will produce a tree where the aggregator function is `charter_parent` (from PR 436) but the function-docstring text wouldn't change (PR 437 doesn't touch that function body). Aino may want to add a follow-up commit on PR 437's branch to align the docstring text post-rebase. Non-blocking — purely cosmetic.

### CI

### Verdict — Approved

The rename + loud-fail + sibling parity + error-message-with-issue-backreference is exactly the shape the silent-no-op family memory (`feedback_gh_pr_edit_silent_noop.md`) prescribes. Test coverage on the new guard is principled (POS correct-form returns empty set; NEG wrong-form raises with #418 in message; trailing-slash NEG; sibling-function NEG). Charter-pinned semantics intact.

## Throughline observations

**Convergent class: deterministic-pipeline skills need loud-fail at the helper boundary, not silent-zero.** This PR is the third instance of the same root cause this cycle:
1. #429 (just merged) — Hook 15 silently returned "no consultation" instead of emitting diagnostics on the block path. Fix: structured diagnostic side-channel.
2. #418 (this PR) — `find_already_promoted_in_charter('.claude/team/charter')` silently returned `set()` instead of raising. Fix: loud-fail ValueError with issue backreference.
3. #419 (PR 437, parallel) — `_SOURCE_HINT_RE` silently matched URL path fragments without operator-visible signal. Fix: regex tightening + URL-strip helper.

All three follow the same arc: helper has too-permissive defensive contract → caller passes plausible-but-wrong input → silent zero/empty/garbage → operator only catches it later via downstream artifact inspection. The fix in all three cases is **operator-visible failure at the helper boundary** (raise / diagnostic record / regex-shape rejection).

**Promotion candidate.** Worth promoting to charter `skills.md` § Deterministic-Skill Helper Discipline: "Helpers in deterministic-classification skills (audit pipelines, classification, ranking) MUST loud-fail on plausible-but-wrong inputs at the helper boundary. Silent-zero, silent-empty, and silent-no-op are forbidden because the downstream artifact (audit log, classifier output, ranked list) may LOOK correct under partial input but be silently incomplete." This is sibling to the existing silent-no-op family memory but specifically for skills-as-libraries, not gh CLI quirks.

Filing as future-charter candidate for next /promotion-audit run. Not blocking this PR.

**Rule-vs-practice gap observation.** The audit pipeline classifies its own defects (#417/#418/#419) — but the defects were surfaced via operator-caught downstream confusion in P3W9 retro (2026-05-12), THEN filed for W10. The lag from "wrong audit-pipeline output observed" to "audit-pipeline defect filed and fixed" was ~1 wave. This may be a recurring shape: the skills-that-classify-skills can only surface their own defects via retro-class human inspection, because the skills don't classify their own output. Worth a /wave-retro carry-forward: should the audit pipeline have a meta-pass that classifies the prior wave's audit-log artifacts and surfaces low-evidence promotions? Not in scope here, just flagging for retro pickup.

parametrization:
Requestor: Wanjiku Mwangi
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: none

## TPM-angle review — process + loud-fail contract

Reviewed at head_sha `58b6c1a1670bb42770e1fc03b1255494b3fe14c9` against base `b7d7437` (post-#434 merge). 4/4 CI checks SUCCESS. Single commit, surgical scope.

### Loud-fail ValueError preserves deterministic-purity

The audit pipeline's deterministic contract is "pure function, same inputs → same outputs, no side effects on disk." A ValueError on wrong-form input does NOT violate that contract because:
- It is RAISED, not LOGGED — no side effect, no I/O
- It is shape-validation, identical in role to `int()` raising on non-numeric strings
- Pre-fix behavior was silent-`set()` return — which IS arguably less pure (it suppressed input invalidity rather than surfacing it). A raise here actually IMPROVES determinism by removing a "wrong input maps to empty result" ambiguity that callers couldn't distinguish from "right input, empty corpus."

The audit pipeline downstream consumer of `find_already_promoted_in_charter` (`/promotion-audit` Step 2) treats an empty set as "no prior promotions exist" — which is exactly the silent-misclassification the original #418 produced. Loud-fail surfaces the call-site bug instead of producing wrong AUTO/DECIDE classifications. Correct trade-off.

### Caller-breakage sweep — confirmed clean

Per the PR body's claim of 3 call sites updated. Verified at HEAD via GH code-search for any remaining `charter_root` references:

- `helpers.py` at HEAD: 0 `charter_root` (renamed)
- `tests/test_smoke.py`: only `_CHARTER_ROOT` (local-constant variable, unrelated to the param name); passes positionally as `str(_CHARTER_ROOT)` so the param-rename is invisible to the caller
- `feedback_log.md`: prose mention of the old name in retro context (historical, not a code reference)
- `SKILL.md`: example block updated to use `charter_parent` with the inline NOTE that pins semantics

No remaining caller passes `charter_root=...` as kwarg, so the rename is backward-compatible-by-position. Confirmed via `gh api search/code?q=charter_root...` at HEAD — no positive code hits in any `.py` file.

### Sibling-function parity is the strongest design choice in this PR

The PR didn't just fix `find_already_promoted_in_charter` — it applied the identical fix shape to `read_all_charter_sections`. Same param-of-charter semantics, same silent-empty failure mode, same ValueError guard, same `#418` backreference in the error message, same test (`test_read_all_charter_sections_raises_on_charter_dir_itself`). This is the "all-similar-instances-of-bug-class fixed in one PR" pattern from `[[feedback_multi_layer_gap_filing]]` (now superseded by `/file-bug` Pass C, but the underlying principle of "don't fix one of two siblings" still applies).

Specifically prevented: a future caller hitting `read_all_charter_sections('.claude/team/charter')`, getting `[]`, and silently misclassifying all sections as "no marked promotion targets exist." That would be a fresh silent-zero bug of identical class to #418.

### Error message quality

The raise message includes:
- The function name (`find_already_promoted_in_charter(...)` literal)
- The bad argument value (`{charter_parent!r}`)
- The expected vs received shape (parent vs charter dir)
- A concrete example (`'.claude/team'` not `'.claude/team/charter'`)
- The issue backreference (`See issue #418.`)

This is the shape that `[[feedback_gh_pr_edit_silent_noop]]` calls out as the right design for the silent-no-op family: name the bug class, tell the caller what to do, link to history. Operator hitting this in CI gets context, not just a stack trace.

### Trailing-slash regression guard

`test_aggregator_raises_on_charter_dir_itself` includes the trailing-slash variant (`charter_subdir + "/"`). The implementation uses `os.path.normpath` to strip it before `os.path.basename`, so both forms trigger. This is the kind of edge case where the obvious implementation `os.path.basename(path)` would fail on `'.claude/team/charter/'` because basename of that is `''` not `'charter'`. The `normpath` + `basename` composition is correct. Good test placement.

### Process notes

- Branch name `N.Khoury/0418-charter-root-rename` follows convention
- Commit identity correct: `parametrization+Nadia.Khoury@gmail.com` via per-commit `-c` (verified via commits API)
- Single commit, focused
- PD-as-author with her direct reports as reviewers IS healthy upward review per charter — flagged in spawn brief as expected, confirmed

The ValueError fires only if the argument is a real directory whose basename is "charter" — if the caller passes a non-existent path that happens to be named `charter`, the existing `if not os.path.isdir(charter_parent): return refs` early-out runs FIRST and silently returns `set()`. So a typo like `'.claude/team/chrtr/'` would still fail silent-empty (non-existent path, returns empty). That's the SAME contract as pre-#418, and broadening the guard to "any path basename == 'charter'" risks false positives for paths legitimately named `charter` that just don't exist yet. Calling it out only because the docstring's note "Returns set() for any other non-existent path (defensive contract preserved)" makes this trade-off explicit — which is the right disclosure shape. Not asking for a fix.

## Throughline observations

Three patterns visible across this PR and the wider P3W10 cluster:

1. **The silent-no-op family is a sustained promotion-pipeline issue.** `[[feedback_gh_pr_edit_silent_noop]]` named gh CLI silent-no-ops 2026-04-21; #418 is the same class but in our OWN code. The fix shape (loud-fail with issue backreference) is now the third instance of the same response pattern in P3W10 (after #429's `pretooluse_diagnostic` and #435's `bypass_pull_request_allowances` drift detection). **Convergent class:** "detection surface that defaults to silent-success when wrong-input is plausible." Memory promotion candidate to charter: "any function whose silent-default-result is indistinguishable from valid-empty MUST raise on shape-detectable wrong-form input." Applies to gh-CLI wrappers, charter scanners, branch-protection normalizers — pattern is wider than the original gh-CLI scope.

2. **Sibling-function fix discipline (this PR) + sibling-class fix discipline (#435).** Both PRs go BEYOND the literal issue scope to cover the structurally-equivalent sibling. #436 fixes `read_all_charter_sections` alongside the named target; #435 fixes `dismissal_restrictions` alongside `bypass_pull_request_allowances`. The cost is small (same fix shape, same test pattern) and the benefit is preventing the SAME bug from reappearing in the sibling 3 months later. **Throughline:** P3W10 PRs are evolving toward "scan-for-siblings during fix design" as a default reviewer expectation, not an exception. Worth surfacing in wave-retro as evolved practice if Santiago's reviews of #435/#436/#437 land the same observation.

3. **Multi-PR cluster for a single skill (#417, #418, #419, #429) — `/promotion-audit` skill drift accumulation.** Four #-numbered PRs in P3W10 touching the same skill's correctness surface (3 of them, #417/#418/#419, surfaced by P3W9 retro). The wave-lead pattern of "skill-drift-surfaces-during-retro-but-fixes-defer-to-next-wave" is real and named. **Promotion candidate:** automate detection. A `/promotion-audit` self-test that runs against `.claude/team/charter` and asserts (a) no shape-anomalies in `_SOURCE_HINT_RE` matches (the #419 false-positive class) and (b) all aggregator functions ARE called with the parent-form on the smoke fixture would catch this class pre-merge in the future. Skill-internal CI test = lower-cost than waiting for the next retro.

Issue #198, opened 2026-04-23
hooks: validate_edit_completion — block state-sensitive actions after unhandled Edit errors


parametrization:
## Context — W10 retro-mandated (2 data points)

During P2W10 coordination, two independent agents hit the same failure class: **tool-error-soft-accept** — Edit tool returns error (e.g., "File has not been read yet"), agent does not explicitly check the return, continues to emit downstream actions as if the edit succeeded, and state drifts from what the agent reports.

### Data points

1. **Marcia.Vasquez-Paredes (noorinalabs-landing-page Manager)** — earlier in P2W10 (see her walkback message in the noorinalabs-landing-page thread), Edit tool errored on a prompt-drafting step, Marcia caught it only on an explicit re-check. Documented in her ack: *"Edit tool errors passed without my checking."*

2. **Bereket.Tadesse (noorinalabs-deploy Manager)** — 2026-04-23, during Contract v4→v3 revert on Aisha's spawn prompt at `/tmp/aisha_spawn_prompt.md`, five consecutive Edit tool calls failed with "File has not been read yet" errors. Agent sent a status message to Nadia Khoury claiming the revert had landed. Later grep-verification caught that the prompt was still at v4 state; corrected on second-pass Edit cycle with fresh Read. The false-status message was explicitly called out and corrected.

Both instances:
- Same tool (Edit)
- Same failure class (tool return unchecked)
- Same blast radius (status messages + downstream routing against stale artifact state)
- Remediation was agent-level discipline (verify-by-grep after edits under coordination pressure)

Pattern, not coincidence. Meets the recurrence threshold per `feedback_enforcement_hierarchy` memory ("prefer hook > skill > charter, in that order") for promotion from observation to hook-backed enforcement.

## Retro lesson (verbatim from Khoury's 2026-04-23 framing)

> "Agents must verify-landed for edits under coordination pressure; tool-error-soft-accept has bitten ≥2 agents in W10. If it keeps surfacing, hook candidate — something that blocks next Write/Edit if prior Edit returned an error-that-wasn't-explicitly-handled."

### Trigger surface
- **PostToolUse** on Edit / Write / NotebookEdit tools that return `is_error: true` in their response
- **PreToolUse** on subsequent state-sensitive actions: further Edit/Write on the same file path, SendMessage containing status-claim language about the artifact, `git commit`, `gh pr comment` / `gh issue comment` containing artifact-state claims

### Detection logic
[...]

Issue #12, opened 2026-04-07
hook: Warn if GHCR image missing before deploy


parametrization:
## Summary

Create a PreToolUse hook that warns when triggering a deploy workflow if the target GHCR image doesn't exist.

## Context
Session 4: deploy-all was triggered but the landing page GHCR image didn't exist yet, causing a failure.

## Implementation
- Hook script: `.claude/hooks/validate_ghcr_image.py`
- Trigger: PreToolUse on Bash commands containing `gh workflow run` with deploy-related workflow names
- Check: use `gh api` to check if the expected GHCR image exists for the target service
- Warn (not block) with message: "Warning: GHCR image {image}:{tag} may not exist. The deploy may fail. Run the service's build workflow first."
- This is a warning, not a hard block, since the image might be built as part of the deploy

## Acceptance Criteria
- Warning displayed when deploying if GHCR image may not exist
- Does not hard-block (deploy workflows sometimes build the image)
- Hook added to settings.json
- Documented in charter/hooks.md

PR #437, opened 2026-05-16
fix(promotion-audit): SKILL.md classifier prose + URL-fragment false positives (#417, #419)

### LINKED ISSUE #417: tech-debt(skill): /promotion-audit SKILL.md prose drift — classify() vs classify_memory/_section/_skill + missing signal-dict example

### LINKED ISSUE #419: tech-debt(skill): /promotion-audit _SOURCE_HINT_RE matches URL path fragments as slash-commands (11 false positives at HEAD)

parametrization:
Closes #417.
Closes #419.

## Summary

Bundle of two `/promotion-audit` drift items touching the same skill (`SKILL.md` + `helpers.py`). Diffs sit in non-overlapping regions; single PR is cleaner than two against the same files.

## #417 — SKILL.md classifier prose drift

`SKILL.md` Step 2 listed `classify` as a single import, but the real `helpers.py` surface is **three classifiers with distinct signatures**. The example block also omitted the `signals: dict[str, int]` argument, so operators copy-pasting the prose passed a bare int and hit `AttributeError: 'int' object has no attribute 'get'` at `helpers.py:672`.

**Fix (prose-only).** Replace the import line with the three real function names; add a per-classifier signature table showing the `signals` keys each consumes; add a worked-example block calling each classifier with a properly shaped `signals` dict; call out the bare-int trap explicitly.

| Function | Signature | `signals` keys consumed |
|---|---|---|
| `classify_memory(memory, signals, already_promoted)` | `(Memory, dict[str,int], set[str]) -> Decision` | `retro_citations` |
| `classify_section(section, signals)` | `(CharterSection, dict[str,int]) -> Decision` | `skill_invocations`, `threshold` |
| `classify_skill(skill, signals, already_promoted)` | `(Skill, dict[str,int], set[str]) -> Decision` | `skill_invocations`, `threshold` |

**Dispatcher extraction skipped (issue body's "optional").** The three classifiers diverge in `already_promoted` handling (section has none) and in `signals` keys. A unified `classify()` would either silently drop `already_promoted` for sections or carry a noop arg — clutter exceeds the operator-visible surface gain once the prose names the three functions correctly.

## #419 — `_SOURCE_HINT_RE` URL-fragment false positives

The `/[\w-]+` branch matched URL path fragments inside markdown link bodies like `[#198](https://github.com/.../issues/198)`. Live evidence (P3W9 retro 2026-05-12) showed 11 false positives in the `already_promoted` set.

**Issue body's recommendation was incomplete.** Empirical check: `/[a-z][a-z0-9-]{2,}` alone rejects only **6 of the 11** (numeric `/198`-class plus the uppercase `/Edit`). The remaining 5 — `/github`, `/issues`, `/noorinalabs`, `/noorinalabs-main`, `/supersedes` — are alpha-leading kebab-case and shape-indistinguishable from real slash-commands.

**Plus a 12th case discovered during fix design:** `/supersedes` inside the `augments/supersedes` prose in `charter/skills.md:108` — slash mid-token, not a URL fragment.

**Two-layer fix:**

1. **`_strip_url_bodies(text)`** — new helper, pre-strips URL bodies from provenance-block content before regex matching:
   - `[text](url)` → `[text]()` (preserves link text so memory/slash references in labels still surface)
   - `<url>` → ``
   - bare `http(s)://...` → ``

2. **Tighten `_SOURCE_HINT_RE`** slash-command branch: `/[\w-]+` → `(?<![\w/])/[a-z][a-z0-9-]{2,}`
   - alpha-leading filter rejects numeric paths
   - lowercase-leading rejects `/Edit`-class
   - length ≥ 3 rejects `/a`, `/b1`
   - word-boundary lookbehind rejects mid-token slashes (the `augments/supersedes` case)

Both layers applied at both call sites in `find_already_promoted` (provenance-block scan + HTML-comment marker scan).

Verified locally — strips all 11 issue-body FPs **+** the 12th `augments/supersedes` case, while preserving `/promotion-audit`, `/wave-retro`, `/auto-promote`, `/manual-trigger`, and memory filenames.

## Tests

- **`SourceHintReTests`** (5) — regex-level POS/NEG for numeric, uppercase, mid-token slashes; real slash-commands at word boundaries; memory filenames both forms.
- **`StripUrlBodiesTests`** (6) — markdown link / autolink / bare URL stripping; non-URL prose unchanged; link-text preservation; end-to-end `find_already_promoted` check against a provenance block containing a gh-issue link.

Local: `pytest .claude/skills/promotion-audit/tests/` → **78 passed** (61 existing + 11 new helper tests + 6 smoke).
`uvx ruff format --check`, `uvx ruff check`, `uvx mypy` all clean on the touched files.

## Reviewers

- **Wanjiku Mwangi** (TPM) — process angle, scope discipline
- **Santiago Ferreira** (Release Coordinator) — skill/charter angle

Manager-boundary check: both peers, Nadia is my PD; neither is the author (Aino, ineligible self-review).

## Test Plan

- [x] Unit + smoke tests (78 passed)
- [x] Ruff format / lint / mypy clean
- [ ] Reviewers verify URL-strip + regex tightening end-to-end against `.claude/team/charter/skills.md:108` (the `/supersedes` 12th case) via `gh api repos/.../contents/.claude/skills/promotion-audit/helpers.py?ref=<HEAD>`
- [ ] Post-merge runtime verification: re-run `find_already_promoted_in_charter('.claude/team')` after merge; expect exactly the real refs (no `/198`/`/issues`/`/noorinalabs-main`/`/supersedes` noise). This is observable via the next `/promotion-audit` run's audit log.

## Charter compliance

- 2 reviewers named in body
- Branch: `A.Virtanen/0417-promotion-audit-bundle` per convention
- Commit identity: `Aino Virtanen <parametrization+Aino.Virtanen@gmail.com>` via per-commit `-c` flags (no global/repo config touched)
- Base: `deployments/phase-3/wave-10` (not main)
- No `--no-verify`
- `Closes #417` AND `Closes #419` each on their own line at the top

### DIFF
--- a/.claude/skills/promotion-audit/SKILL.md
+++ b/.claude/skills/promotion-audit/SKILL.md
@@ -36,7 +36,9 @@ from helpers import (
     find_already_promoted_in_charter,
     count_retro_citations,
     count_skill_invocations,
-    classify,
+    classify_memory,
+    classify_section,
+    classify_skill,
     render_audit_table,
 )
 
@@ -48,7 +50,44 @@ already   = find_already_promoted_in_charter(charter_dir)  # set[str] — aggreg
 
 ### 3. Classify each candidate (pure function)
 
-For each memory, charter section, and skill, compute a `Decision` via `classify()`:
+There is no single `classify()` entry point — each tier transition has its own classifier with a distinct signature, because the signal sources differ (retro citations for memory → charter; invocation counts for charter → skill and skill → hook) and `classify_section` has no `already_promoted` analogue (sections carry their own `promoted_to` back-reference instead).
+
+| Function | Signature | `signals` keys consumed |
+|---|---|---|
+| `classify_memory(memory, signals, already_promoted)` | `(Memory, dict[str,int], set[str]) -> Decision` | `retro_citations` |
+| `classify_section(section, signals)` | `(CharterSection, dict[str,int]) -> Decision` | `skill_invocations`, `threshold` |
+| `classify_skill(skill, signals, already_promoted)` | `(Skill, dict[str,int], set[str]) -> Decision` | `skill_invocations`, `threshold` |
+
+Worked example (one item per tier):
+
+```python
+mem_decision = classify_memory(
+    memories[0],
+    signals={"retro_citations": count_retro_citations(memories[0], feedback_log_path)},
+    already_promoted=already,
+)
+
+sec_decision = classify_section(
+    sections[0],
+    signals={
+        "skill_invocations": count_skill_invocations(sections[0].promoted_to_slug or "", repo_root),
+        "threshold": 5,
+    },
+)
+
+skill_decision = classify_skill(
+    skills[0],
+    signals={
+        "skill_invocations": count_skill_invocations(skills[0].name, repo_root),
+        "threshold": 5,
+    },
+    already_promoted=already,
+)
+```
+
+Common failure mode: passing the citation/invocation count as a bare int (`classify_memory(mem, cite, already)`) rather than wrapping it in the `signals` dict (`classify_memory(mem, {"retro_citations": cite}, already)`). The classifiers all do `signals.get(...)` and will raise `AttributeError: 'int' object has no attribute 'get'` on a bare int. Always pass a dict.
+
+Each call returns a `Decision` in one of these kinds:
 
 - **AUTO** — thresholds met, promotion target is charter or skill, NOT already promoted
 - **DECIDE** — thresholds met, target is hook (always DECIDE), OR `requires_decision: true` override, OR signals ambiguous
--- a/.claude/skills/promotion-audit/helpers.py
+++ b/.claude/skills/promotion-audit/helpers.py
@@ -440,19 +440,60 @@ _HTML_COMMENT_PROMOTED_RE = re.compile(
 
 # Memory filenames can be cited with or without the `.md` suffix in
 # charter prose (e.g. hooks.md L169 cites `feedback_honest_audit_over_conclusion_claim`
-# unsuffixed inside a backticked span). The regex now makes the `.md`
-# optional and the caller backfills the suffix into the returned set so
-# both forms are recognized membership-checks.
+# unsuffixed inside a backticked span). The regex makes the `.md` optional
+# and the caller backfills the suffix into the returned set so both forms
+# are recognized in membership checks.
+#
+# Slash-command branch shape (#419): `(?<![\w/])/[a-z][a-z0-9-]{2,}` —
+# kebab-case slash-commands, alpha-leading, length >= 3, AND preceded by
+# a non-word/non-slash boundary. The alpha-leading filter rejects
+# URL-path-fragment hits like `/198` (numeric issue IDs from gh URLs);
+# the length-3 minimum rejects `/a`, `/b1`; and the lookbehind rejects
+# mid-token slashes like `/supersedes` inside the prose `augments/supersedes`
+# in charter/skills.md. Combined with `_strip_url_bodies` (applied to the
+# block body BEFORE the regex runs), this rejects the 11 URL-fragment
+# false positives documented in #419 plus the `augments/supersedes` 12th.
 _SOURCE_HINT_RE = re.compile(
     r"""
     (?:
         (?:feedback|project|reference)_[a-z0-9_]+(?:\.md)?  # memory filenames (with or without .md)
         |
-        /[\w-]+                                              # slash-commands / skill names
+        (?<![\w/])/[a-z][a-z0-9-]{2,}                       # slash-commands at word boundary
     )
     """,
     re.VERBOSE,
 )
+
+
+# URL-stripping regexes (#419). Applied to provenance-block bodies BEFORE
+# `_SOURCE_HINT_RE.finditer` so URL path fragments never reach the slash-
+# command branch. Three forms:
+#   - markdown link bodies: `[text](url)` → `[text]()` (preserves link text)
+#   - autolinks: `<url>` → ``
+#   - bare URLs: `https://...` → ``
+# The link-text preservation matters because some link labels themselves
+# contain a memory filename or slash-command reference that IS load-bearing.
+_MD_LINK_URL_RE = re.compile(r"\]\(\s*<?(?:https?|ftp)://[^\s)>]+>?\s*\)")
+_AUTOLINK_URL_RE = re.compile(r"<(?:https?|ftp)://[^>\s]+>")
+_BARE_URL_RE = re.compile(r"(?<![\[\(])(?:https?|ftp)://[^\s)>]+")
+
+
+def _strip_url_bodies(text: str) -> str:
+    """Remove URL bodies from `text` while preserving non-URL prose.
+
+    Applied to provenance-block bodies before `_SOURCE_HINT_RE.finditer`
+    so URL path fragments (`/198`, `/issues`, `/noorinalabs-main`, etc.)
+    never reach the slash-command branch as false positives (#419).
+    Markdown link text is preserved by replacing `(url)` with `()`,
+    leaving `[text]` intact — if a link label cites a real memory
+    filename or slash-command, that reference still surfaces in the
+    downstream regex sweep.
+    """
+    text = _MD_LINK_URL_RE.sub("]()", text)
+    text = _AUTOLINK_URL_RE.sub("", text)
+    text = _BARE_URL_RE.sub("", text)
+    return text
+
 
 # Memory-filename prefixes the parser recognizes. Used by
 # `_normalize_memory_hit` to decide whether a no-suffix hit should be
@@ -554,8 +595,10 @@ def find_already_promoted(charter_path: str) -> set[str]:
         text = f.read()
 
     # Block-style `**Promotion provenance:**` entries (hooks.md format).
+    # URL bodies stripped first (#419) so gh-issue-link path fragments
+    # never reach the slash-command branch.
     for block in _PROVENANCE_RE.finditer(text):
-        body = block.group("body")
+        body = _strip_url_bodies(block.group("body"))
         for hit in _SOURCE_HINT_RE.finditer(body):
             if _is_forward_reference(body, hit.start()):
                 continue
@@ -566,7 +609,7 @@ def find_already_promoted(charter_path: str) -> set[str]:
     # unnecessary for this format — the marker is by definition a
     # backward-looking promotion claim.
     for block in _HTML_COMMENT_PROMOTED_RE.finditer(text):
-        body = block.group("body")
+        body = _strip_url_bodies(block.group("body"))
         for hit in _SOURCE_HINT_RE.finditer(body):
             refs.update(_normalize_memory_hit(hit.group(0)))
 
--- a/.claude/skills/promotion-audit/tests/test_helpers.py
+++ b/.claude/skills/promotion-audit/tests/test_helpers.py
@@ -928,5 +928,137 @@ class SlugifyTests(unittest.TestCase):
         self.assertEqual(h._slugify("   "), "section")
 
 
+class SourceHintReTests(unittest.TestCase):
+    """#419 regression — `_SOURCE_HINT_RE` MUST NOT match URL path
+    fragments inside gh-issue link bodies, and MUST still match real
+    kebab-case slash-commands at word boundaries.
+
+    Exercises the regex directly (tightening) plus `_strip_url_bodies`
+    (link-context elimination). Both layers are load-bearing; tests
+    cover each independently and in combination via `find_already_promoted`.
+    """
+
+    def test_excludes_numeric_url_path_fragment(self) -> None:
+        """NEG (#419 row 1): `/198` from gh-issue URL fragments is rejected
+        purely by the regex shape — alpha-leading filter — even before
+        URL-stripping runs."""
+        matches = h._SOURCE_HINT_RE.findall("filed as /198 and /200 and /244")
+        self.assertNotIn("/198", matches)
+        self.assertNotIn("/200", matches)
+        self.assertNotIn("/244", matches)
+
+    def test_excludes_uppercase_url_path_fragment(self) -> None:
+        """NEG (#419 row 2): `/Edit` from path fragments rejected by the
+        alpha-LOWERCASE-leading filter."""
+        matches = h._SOURCE_HINT_RE.findall("see /Edit or /GitHub")
+        self.assertNotIn("/Edit", matches)
+        self.assertNotIn("/GitHub", matches)
+
+    def test_excludes_mid_token_slash(self) -> None:
+        """NEG (12th case): `/supersedes` inside `augments/supersedes`
+        prose (charter/skills.md:108) is rejected by the word-boundary
+        lookbehind."""
+        matches = h._SOURCE_HINT_RE.findall("augments/supersedes relationships")
+        self.assertNotIn("/supersedes", matches)
+
+    def test_matches_real_slash_command_at_word_boundary(self) -> None:
+        """POS (#419 acceptance): real slash-commands at word boundaries
+        (start, after whitespace, after punctuation) still match."""
+        matches = set(
+            h._SOURCE_HINT_RE.findall("see /promotion-audit and /wave-retro (/auto-promote)")
+        )
+        self.assertIn("/promotion-audit", matches)
+        self.assertIn("/wave-retro", matches)
+        self.assertIn("/auto-promote", matches)
+
+    def test_matches_memory_filename_both_forms(self) -> None:
+        """POS regression guard: memory-filename matching unaffected by
+        the slash-branch tightening."""
+        matches = set(
+            h._SOURCE_HINT_RE.findall(
+                "see feedback_foo.md and project_bar (suffixless), reference_baz.md"
+            )
+        )
+        self.assertIn("feedback_foo.md", matches)
+        self.assertIn("reference_baz.md", matches)
+        # Suffix-less form is captured separately.
+        self.assertIn("project_bar", matches)
+
+
+class StripUrlBodiesTests(unittest.TestCase):
+    """#419: `_strip_url_bodies` removes URL bodies (markdown links,
+    autolinks, bare URLs) while preserving the surrounding prose so
+    URL-path-fragment matches never reach `_SOURCE_HINT_RE`."""
+
+    def test_markdown_link_url_stripped_text_preserved(self) -> None:
+        """[text](url) → [text]() — URL gone, link text remains so any
+        memory/slash-command reference inside the text is still hit."""
+        s = h._strip_url_bodies(
+            "[#198](https://github.com/noorinalabs/noorinalabs-main/issues/198)"
+        )
+        self.assertEqual(s, "[#198]()")
+        # Downstream regex on the stripped form must NOT hit `/198`/`/issues`.
+        self.assertNotIn("/198", h._SOURCE_HINT_RE.findall(s))
+        self.assertNotIn("/issues", h._SOURCE_HINT_RE.findall(s))
+        self.assertNotIn("/noorinalabs", h._SOURCE_HINT_RE.findall(s))
+
+    def test_autolink_url_stripped(self) -> None:
+        """<https://...> → empty (no link text to preserve)."""
+        s = h._strip_url_bodies("see <https://github.com/foo/bar/issues/337>")
+        self.assertNotIn("github", s)
+        self.assertNotIn("337", s)
+
+    def test_bare_url_stripped(self) -> None:
+        """Bare `http(s)://...` in prose is stripped."""
+        s = h._strip_url_bodies("docs at https://example.com/path/to/thing")
+        self.assertNotIn("https://", s)
+        self.assertNotIn("/path", s)
+        self.assertNotIn("/thing", s)
+
+    def test_non_url_prose_unchanged(self) -> None:
+        """NEG: text without URLs is returned identically — no
+        accidental damage to charter prose."""
+        s = "see /promotion-audit and feedback_foo.md per retro"
+        self.assertEqual(h._strip_url_bodies(s), s)
+
+    def test_link_text_with_real_slash_command_preserved(self) -> None:
+        """POS: a markdown link whose LABEL cites a real slash-command
+        must still surface that slash-command after stripping."""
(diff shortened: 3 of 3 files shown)


parametrization:
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

## Release-coordinator + skill/charter review

Read at HEAD `a4f52c2` via `gh api .../contents/...?ref=<sha>`. CI 4/4 SUCCESS. mergeable_state `clean`.

### Independent regex verification

Ran the new `_SOURCE_HINT_RE` shape `(?<![\w/])/[a-z][a-z0-9-]{2,}` against the inputs claimed in the PR body — all 9 cases match expected pos/neg:

```
'/198'              → []                    (rejected — alpha-leading filter)
'/Edit'             → []                    (rejected — lowercase-leading filter)
'/promotion-audit'  → ['/promotion-audit']  (accepted — real slash-command)
'/wave-retro'       → ['/wave-retro']
'/auto-promote'     → ['/auto-promote']
'/a'                → []                    (rejected — length-3 minimum)
'/b1'               → []                    (rejected — length-3 minimum)
'augments/supersedes' → []                  (rejected — word-boundary lookbehind)
'/manual-trigger'   → ['/manual-trigger']
```

The "11/11 + 12th case" empirical claim in the PR body is faithful — the regex tightening alone catches 6 (the numeric + uppercase cases) and the URL strip catches the remaining 5 alpha-leading kebab-case ones (`/github`, `/issues`, `/noorinalabs`, `/noorinalabs-main`, `/supersedes`) by removing them from the regex-scanned text entirely.

### "Deterministic, no-side-effect" contract — PRESERVED

Three axes inspected:

1. **`_strip_url_bodies` is a pure function** — takes `text`, returns transformed `text`. No filesystem access, no global state mutation, no logging. The transformation is regex-replace-only, applied to the BODY of provenance blocks BEFORE the slash-command regex runs. Output is deterministic for given input. Verified at `helpers.py:481-494`.

2. **Both call sites apply the strip uniformly** — `find_already_promoted` line 601 (`_PROVENANCE_RE` block-style) and line 612 (`_HTML_COMMENT_PROMOTED_RE` HTML-comment style). Same `_strip_url_bodies(block.group("body"))` shape at both. No asymmetry that would let URL fragments leak in via one path but not the other.

3. **No new side effects beyond text transformation** — the regex tightening is a pure shape change; the URL strip is pure text-in/text-out. The `find_already_promoted` function's external contract (input: charter_path → output: set[str]) is unchanged. Idempotent: stripping a stripped string is a no-op. Verified by `test_non_url_prose_unchanged`.

### Does URL pre-stripping corrupt structural information?

This was the load-bearing concern in the review angle, and the answer is **no** with strong evidence:

**What IS preserved:**
- `[text](url)` → `[text]()` — link TEXT preserved as bracketed content, so any memory/slash-command reference inside the label still surfaces. Test `test_link_text_with_real_slash_command_preserved` asserts `[/promotion-audit run output](https://example.com/log)` still surfaces `/promotion-audit` post-strip.
- Bare prose unaffected — `test_non_url_prose_unchanged` asserts text without URLs is returned identically.

**What IS stripped:**
- URL bodies in `(url)` parentheses → empty parens
- Autolink content in `<url>` brackets → empty
- Bare `http(s)://...` URLs → empty

**The structural information the audit relies on** is the SET of memory filenames + slash-command references inside provenance blocks. The URL itself never carries that information — it's always a citation pointer to GitHub. Stripping URLs cannot lose audit-load-bearing information because the audit pipeline never used URL-internal content as a promotion signal. Verified by `test_find_already_promoted_rejects_url_fragments_end_to_end` — real refs (`feedback_foo.md`, `/promotion-audit`) still surface; URL-fragment noise (`/198`, `/issues`, `/noorinalabs`, `/noorinalabs-main`, `/github`) does NOT.

**Edge case I considered and verified safe:** the `_MD_LINK_URL_RE` shape `\]\(\s*<?(?:https?|ftp)://[^\s)>]+>?\s*\)` requires the URL to be `http(s)://` or `ftp://` prefixed. Markdown links to RELATIVE paths like `[charter](../charter/skills.md)` are NOT matched and pass through unchanged. So relative-path citations in provenance blocks (which CAN carry slash-command references in the label, e.g. `[/wave-retro skill](../skills/wave-retro/SKILL.md)`) are unaffected. Good narrow scope.

### #417 prose-only fix — APPROVED

The dispatcher-extraction-skipped decision is defensible. The three classifiers have genuinely different signatures (`classify_section` has no `already_promoted` analogue because sections carry their own back-reference). A unified `classify()` would either (a) silently drop `already_promoted` for sections or (b) carry a noop arg. Option (a) is the silent-no-op family this audit cycle is actively fighting; option (b) is operator-visible clutter. Prose-only with the three real function names + signature table + bare-int trap callout is the right shape.

Worked-example block at `SKILL.md:64-86` correctly wraps every signal as `signals={"key": value}` — directly addresses the AttributeError pattern called out in the issue body.

### Coordination with #436 (Nadia, parallel)

Verified non-textually-overlapping diff regions. PR 436 hunks at lines 301 (sibling rename) + 597 (aggregator signature + ValueError). PR 437 hunks at 440-525 (regex region) + 595-615 (call-site URL strip). Git 3-way merge resolves cleanly.

**Rebase note:** PR 437 retains `charter_root` as the param name in `find_already_promoted_in_charter` (PR 437 predates Nadia's rename). After PR 436 merges first, PR 437's branch will rebase cleanly because PR 437 doesn't touch that function's signature or body — Nadia's `charter_parent` + ValueError guard survive intact. The only post-rebase cosmetic mismatch will be PR 437's docstring near line 583 still saying "charter_root" — non-blocking, fix-as-followup if desired.

### CI

### Verdict — Approved

URL strip is correctly factored as a pure helper applied at both regex call sites; regex tightening is independently verified against PR body claims; structural information the audit relies on is preserved (link text passes through; only URL bodies stripped); the #417 prose fix is the right shape for diverging signatures. Charter-pinned deterministic-no-side-effect contract intact.

## Throughline observations

**Same convergent class as PR #436 (concurrent in this wave) — see my #436 verdict for the full pattern.** Brief recap here for #437's lens:

This PR is one of three this cycle that all share the shape "helper had too-permissive defensive contract → silently produced wrong output → caught only via downstream operator inspection." The other two are #418 (just-reviewed PR 436, ValueError on charter-dir input) and #429 (last week, Hook 15 diagnostic side-channel). All three are the same convergent class: **deterministic-pipeline skills need operator-visible failure at the helper boundary.**

**Empirical-over-spec discipline as a separate throughline.** This PR specifically goes BEYOND the issue body's recommendation after empirical testing showed the issue-body fix caught only 6/11 FPs. Aino made the call to add `_strip_url_bodies` as a second layer rather than ship a fix that only addressed half the cases. The PR body documents the empirical finding ("6 of the 11" alpha-leading kebab-case cases remained after regex-only fix) AND documents the 12th case discovered during fix design (`/supersedes` in `augments/supersedes`). This is the right shape — issue body is a starting point, not a contract. Worth naming as a wave-retro positive pattern: "empirical validation of issue-body recommendations before shipping" — when an audit finds an issue's fix is incomplete, the right move is to expand scope WITH explicit justification, not to ship the partial fix and file a followup.

**Promotion candidate sibling.** The "deterministic helper loud-fail" candidate I named in PR #436's throughline applies here too — same charter `skills.md` § Deterministic-Skill Helper Discipline section would cover this PR's regex-tightening as a sub-rule: "Regex predicates in deterministic-classification helpers MUST be tightened to reject plausible-but-wrong inputs (e.g. URL path fragments masquerading as slash-commands) rather than relying on downstream filters to clean up false positives." Filing as future-charter candidate; not blocking this PR.

parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

## TPM-angle review — scope-expansion-justified-by-empirical-evidence pattern

Reviewed at head_sha `a4f52c22a0de4e0a64c40cbf7ba3c0ade9e56f6a` against base `b7d7437` (post-#434 merge). 4/4 CI checks SUCCESS. Single commit, focused bundle of #417 + #419 as declared.

### Scope-expansion analysis — `/promotion-audit` #419 deep fix is JUSTIFIED expansion

The issue body recommended a regex tightening: `/[a-z][a-z0-9-]{2,}`. Aino's PR went beyond by adding a URL-body pre-strip helper AND tightening the regex. Empirical evidence in PR body:

- Issue body's recommendation alone: catches 6 of 11 documented FPs
- Aino's combined fix: catches 11 of 11 PLUS a 12th case (`/supersedes` in `augments/supersedes` charter prose)

This is the "scope-expansion-justified-by-empirical-evidence" pattern in its cleanest form: the PR body presents the math (6 vs 11) BEFORE the design decision, making the expansion auditable rather than aspirational. The 5 FPs the regex alone would miss (`/github`, `/issues`, `/noorinalabs`, `/noorinalabs-main`, `/supersedes`) are alpha-leading kebab-case — shape-indistinguishable from real slash-commands without URL context. The only way to reject them is to remove their URL context BEFORE the regex runs. So the URL-body strip is not a "while we're here" addition; it's the only mechanism that closes the remaining 5 cases. Justified.

Documenting this PR as a reference instance of the pattern: **scope-expansion-justified-by-empirical-evidence** = (a) issue says X, (b) empirical check shows X catches N/M cases, (c) PR proposes X' which catches M/M plus K additional, (d) the additional K are surfaced from the SAME bug class not a new one. Aino hit all four criteria.

### #417 prose-only fix + dispatcher-skipped justified

PR body declares "Dispatcher extraction skipped (issue body's 'optional')" with the reasoning: the three classifiers diverge on `already_promoted` (section has none) and on `signals` keys, so a unified `classify()` either silently drops `already_promoted` for sections or carries a no-op arg. Both options trade away clarity for marginal call-site reduction. Skipping is correct.

The prose fix is the minimum viable change to remove the documented operator footgun:
- Replaces `classify` (mythical) with the three real function names in the import block
- Adds the per-classifier signature table (3 rows) showing which `signals` keys each consumes
- Adds a worked example calling each classifier with the proper `signals` dict shape
- Calls out the bare-int trap explicitly with the exact `AttributeError` operators would see

The "Common failure mode" callout is the kind of inline documentation that prevents the next operator from hitting `helpers.py:672 AttributeError: 'int' object has no attribute 'get'`. Saves the exact debugging round-trip the issue was filed for.

### URL-body-strip test fixtures cover all three URL shapes

`StripUrlBodiesTests` (6 tests) covers:
- `test_markdown_link_url_stripped_text_preserved`: `[#198](https://github.com/.../issues/198)` → `[#198]()` and downstream regex doesn't hit `/198`/`/issues`/`/noorinalabs`
- `test_autolink_url_stripped`: `<https://...>` → empty
- `test_bare_url_stripped`: `docs at https://example.com/path/to/thing` → no `https://`/`/path`/`/thing`
- `test_non_url_prose_unchanged`: idempotent on prose without URLs
- `test_link_text_with_real_slash_command_preserved`: `[/promotion-audit run](url)` → still surfaces `/promotion-audit`
- `test_find_already_promoted_rejects_url_fragments_end_to_end`: end-to-end on `find_already_promoted` with a charter fixture containing both a real ref (`feedback_foo.md`, `/promotion-audit`) AND a noise link (`[#198](...)`) — asserts real refs present, noise absent

The end-to-end test is the strongest one: it exercises the actual code path `/promotion-audit` runs (`find_already_promoted` → block scan → URL strip → regex), with the literal payload shape (gh-issue link inside a `**Promotion provenance:**` block) that produced the original 11 FPs in P3W9 retro. That's a regression test against the real failure, not a synthesized scenario.

### Regex tightening covers all 4 documented classes

`SourceHintReTests` covers:
- `test_excludes_numeric_url_path_fragment`: `/198`, `/200`, `/244` — alpha-leading filter
- `test_excludes_uppercase_url_path_fragment`: `/Edit`, `/GitHub` — lowercase-leading filter
- `test_excludes_mid_token_slash`: `/supersedes` in `augments/supersedes` — word-boundary lookbehind
- `test_matches_real_slash_command_at_word_boundary`: `/promotion-audit`, `/wave-retro`, `/auto-promote` — POS regression
- `test_matches_memory_filename_both_forms`: `feedback_foo.md`, `project_bar` (suffixless), `reference_baz.md` — POS regression guard for the other regex branch

Each NEG test names which filter mechanism catches it (alpha-leading / lowercase / length-3 / word-boundary). When the regex evolves again, the test failure tells the maintainer exactly which property they violated.

### Coordination with #436 — confirmed non-overlapping

Both PRs touch `helpers.py` but on disjoint functions:
- #436 modifies `read_all_charter_sections` (~line 301) and `find_already_promoted_in_charter` (~line 597) — both `_in_charter` aggregator + sibling
- #437 modifies `find_already_promoted` (~line 595 + 609) — per-file scanner — AND adds new `_strip_url_bodies` helper (~line 491) + tightens `_SOURCE_HINT_RE` (~line 451)

Different functions, different line ranges. test_helpers.py insertions: #436 at line 517, #437 at line 928 — non-overlap. Merge order doesn't matter; CI will run on each post-merge base independently. Verified by reading both patches at HEAD via `gh api .../files`.

### Process notes

- Branch name `A.Virtanen/0417-promotion-audit-bundle` follows convention
- Commit identity correct: `parametrization+Aino.Virtanen@gmail.com` via per-commit `-c` (verified via commits API)
- Single commit, focused bundle
- Two `Closes` lines (`Closes #417.` + `Closes #419.`) each on their own line at top — gh auto-close fires on both when merged via the wave branch (with `[[feedback_wave_branch_issue_close]]` manual close required for wave-branch merges)
- Manager-boundary check passes (PD is Nadia, neither reviewer is PD)
- 78 tests pass locally (61 existing + 11 new helper + 6 new smoke) — matches the 78 the body claims

The `_strip_url_bodies` helper is applied at BOTH call sites in `find_already_promoted` (provenance block AND HTML-comment marker). Symmetric application is correct. But this raises a subtle question: should `_SOURCE_HINT_RE` also be tightened in `find_already_promoted_in_charter` (the aggregator)? Looking at #436's diff: the aggregator just calls `find_already_promoted` per-file and unions the results, so the tightened regex AND the URL-strip both flow through transitively. Confirmed by tracing the call graph — no separate regex use in the aggregator. So this is fine; just noting that #436 + #437 together produce the right end-to-end behavior because the aggregator delegates rather than reimplementing.

## Throughline observations

Three patterns visible across this PR + the wider W10 cluster:

1. **`/promotion-audit` skill drift is a sustained class — convergent across 4 PRs in P3W10.** #417, #418, #419 all surfaced by P3W9 retro plus #429 (Hook 15 sentinel orphan-formula). The skill+helpers surface accumulated four distinct correctness defects between P3W4 (initial promotion) and P3W9 (retro that caught them). **Class:** detection-surface drift in self-validating tooling. The `/promotion-audit` skill is META: it audits the memory → charter → skill → hook pipeline, and IS itself subject to that same pipeline. Memory `[[feedback_enforcement_hierarchy]]` and the meta-recursive nature mean each defect in the skill could itself become an unfixed promotion candidate. **Promotion candidate:** a `/promotion-audit` self-test gate in CI — run the skill against `.claude/team/charter` on every PR that touches `.claude/skills/promotion-audit/` AND fail-CI if the output diff exceeds N entries (where N is the known-canonical count). That would catch #419's 11 FPs the moment they regressed, and #418's silent-zero the moment a caller passed the wrong path shape.

2. **Scope-expansion-justified-by-empirical-evidence is a reference pattern after this PR.** Aino's 6-of-11 vs 11-of-11 (plus #12) math, presented inline in the PR body BEFORE the design decision, is the cleanest instance of this pattern in P3W10. Compare to #435's `dismissal_restrictions` sibling expansion (justified by symmetric-vulnerability-class evidence) and #436's `read_all_charter_sections` sibling expansion (justified by identical-failure-mode evidence) — three PRs this wave, three different shapes of justified scope expansion. **Throughline:** P3W10 reviewers are increasingly accepting (and reviewers/orchestrators are increasingly proposing) scope-expansion when the expansion math is shown. Promotion candidate to charter agents.md: "PR may expand scope beyond the literal issue IF (a) the expansion target is in the same bug class, (b) the cost-evidence ratio is shown in PR body, (c) tests cover the expanded surface." Tracks to `[[feedback_pre_spawn_brief_verified_at_head]]` companion: pre-spawn surface enumeration discipline pays out as scope-expansion-discipline at PR time.

3. **Bundle-related-fixes-in-one-PR pattern emerging.** #437 bundles #417 + #419; #435 bundles A + B fix shapes; #429 bundled diagnostic + tolerance + tests. Three P3W10 PRs follow the same "if the fixes share artifact scope AND a partial fix would be operationally incomplete, bundle them" rule. Compare to the older P2 pattern of one-issue-one-PR which forced churn-of-similar-context. **Promotion candidate:** charter agents.md guidance — bundling allowed when (a) same files touched, (b) partial fix would leave a known-incomplete state, (c) PR body declares both `Closes` lines explicitly. Already-implicit practice; worth making explicit so reviewers stop counter-asking for splits when bundling is correct.

Issue #423, opened 2026-05-13
tech-debt(/wave-wrapup): cross-window PR over-count when wave-branch absorbs multiple partition windows (W9: 30 vs 6)


parametrization:
## Problem

`/wave-wrapup` Step 10.5's mechanical computation (added in #421/#422, 2026-05-13) counts merged PRs via `gh pr list --base "deployments/phase-{P}/wave-{M}"`. This produces an inflated count when the wave-branch absorbs PRs across multiple partition windows / re-roll cycles.

**W9 anomaly (caught by Wanjiku Mwangi during PR #422 review, 2026-05-13):**

- Pre-partition window (May 11–12 AM, ~24 PRs) was rolled to main via #382, #389, #400, #408 — but the source PRs still target `deployments/phase-3/wave-9` and remain in `gh pr list --base wave-9`.
- Post-partition canonical window (the 6 PRs of the actual W9 main-only theme) was rolled via #416.
- `gh pr list --base deployments/phase-3/wave-9 --state merged` returns **30 PRs** total.
- Canonical W9 (per `wave_9_final_pr_count=6`) is **6**.

Result: mechanical TOP_CONCENTRATION_PCT and CHANGES_REQUESTED_CYCLES computed against the 30-PR set give **50% / 1** — wrong shape for the canonical wave. The recorded W9 actuals (CR=0, TOP=67%) only match when computed against the 6-PR canonical set, not the 30-PR cross-window set.

The formula works correctly for "single-window" waves (the typical case going forward). The cross-window risk emerges when a wave-branch is reused across partition events, which W9 happened to do (owner directive 2026-05-12 partitioned non-deploy work from deploy work mid-wave).

### Option A — Kickoff-timestamp filter on `gh pr list`

Use `cross-repo-status.json wave_{M}_kicked_off_at` as a `mergedAt >= X` filter on the PR list. The wave's canonical window starts at kickoff; anything merged before is from a prior window.

```bash
KICKOFF=$(jq -r ".wave_${M}_kicked_off_at" "$STATUS")  # e.g., "2026-05-10T17:55:00Z"
PRS_JSON=$(gh pr list --repo "noorinalabs/$REPO" --state merged --base "$BRANCH" \
  --json number,headRefOid,mergedAt --jq ".[] | select(.mergedAt >= \"$KICKOFF\")")
```

Pros: Surgical fix. Doesn't require additional state-tracking.
[...]

PR #438, opened 2026-05-16
fix(hooks): dispatcher zero-observability — #425 framed as dispatcher dead, root cause was silent-diagnostic-return pattern (sibling of #429, #437)

### LINKED ISSUE #425: tech-debt(hook): post_wave_kickoff_comment doesn't fire via PostToolUse dispatcher chain — direct-invoke workaround used in /wave-kickoff p3 10
## Summary

`post_wave_kickoff_comment` hook is registered in `.claude/hooks/post_dispatcher.py` `_REGISTRY["Bash"]` (3rd entry, after `annunaki_monitor` and `auto_add_issue_to_board`), but did **NOT** fire when triggered via the harness PostToolUse chain on `gh issue edit ... --add-label "p3-wave-10"` Bash calls during `/wave-kickoff p3 10` (2026-05-13).

Workaround used in the kickoff: invoked the hook's `check()` function directly in-process via Python, post-`gh issue edit`. 61 of 62 issues received their kickoff comment via this workaround; 1 (#423) skipped idempotently because it had received a comment earlier during a debug smoke-test direct invocation.

## Evidence

Smoke-test reproduction (W10 session, ~2026-05-13T16:25Z):

```
$ gh issue view 423 --repo noorinalabs/noorinalabs-main --json comments --jq '.comments | length'
0
$ gh issue edit 423 --repo noorinalabs/noorinalabs-main --add-label "WANJIKU_MWANGI" --add-label "p3-wave-10"
https://github.com/noorinalabs/noorinalabs-main/issues/423
$ sleep 3
$ gh issue view 423 --repo noorinalabs/noorinalabs-main --json comments --jq '.comments | length'
0          # ← expected 1 (hook should have posted a kickoff comment)
```

Then direct hook invocation in Python:

```python
import sys; sys.path.insert(0, '.claude/hooks')
import post_wave_kickoff_comment as h
result = h.check({
    "tool_name": "Bash",
    "tool_input": {"command": "gh issue edit 423 --repo noorinalabs/noorinalabs-main --add-label 'WANJIKU_MWANGI' --add-label 'p3-wave-10'"},
    "tool_response": {"stdout": "...", "stderr": "", "exitCode": 0}
})
# returns {'action': 'post', 'repo': 'noorinalabs-main', 'issue': '423', ...}
# AND posts the comment successfully (verified: #423 had 1 comment after this call)
```

So the hook function itself is correct. The dispatcher chain is the gap.

## Hypotheses (un-investigated)

1. The harness may not be invoking `post_dispatcher.py` for these Bash tool calls (but `annunaki_monitor` is also in `_REGISTRY["Bash"]` and historically logs — so that's likely not it; if dispatcher weren't firing, annunaki wouldn't be logging either).
2. The dispatcher IS firing but `post_wave_kickoff_comment.check()` is silently raising an exception that gets swallowed by the dispatcher's `try/except Exception: continue` guard at `post_dispatcher.py:96-98`. The dispatcher has no logging of swallowed exceptions.
3. The `input_data` shape passed by the harness differs from what I synthesized in direct invocation (e.g., `tool_response` field absent, causing some unrelated code path to fail).

## Proposed fix

1. **Investigation**: Add structured logging to `post_dispatcher.py` `try/except Exception: continue` blocks — write the exception type and module name to a debug log so swallowed errors are recoverable. This is hook-class self-improvement (Pattern G).
2. **Confirm root cause** by re-running the kickoff trigger pattern with the new logging in place. Either (a) hook never gets called (harness/dispatcher wiring bug) or (b) hook gets called and raises something.
3. **Fix** per the root cause; update `post_wave_kickoff_comment.py` if it's an input-shape mismatch, or fix the dispatcher / wiring.

## Acceptance criteria

- Add debug logging to dispatcher's exception-swallowing blocks
- Root cause identified and either:
  - Documented with a regression test that asserts the dispatcher invokes `post_wave_kickoff_comment.check()` for matching Bash commands, OR
  - Fix lands and the regression test verifies the fix
- The `/wave-kickoff` skill SKILL.md is updated to remove or revise its claim that "The kickoff comment is posted automatically" until this is reliably true through the harness

## Context

- Surfaced during `/wave-kickoff p3 10` on 2026-05-13 (this wave)
- Workaround: direct hook invocation via Python orchestration script (62 issues, all comments posted)
- Same workaround is reusable for next wave's kickoff if this isn't fixed in W10

## Priority

W10 scope item — file-bug at user request 2026-05-13. Charter `feedback_enforcement_hierarchy` says hook > skill, but a hook that doesn't reliably fire is worse than the skill it replaced (the skill at least posted comments deterministically). Pattern G — in-wave skill/hook self-improvement.

parametrization:
Closes #425.

## Premise correction (lead-with)

**#425 was filed against a "dispatcher dead" symptom that didn't exist.** The dispatcher chain was firing `post_wave_kickoff_comment.check()` correctly all along. Investigation showed the user's reproduction on #423 was the hook correctly returning `skip_idempotent` (because #423 already had a kickoff comment from a prior smoke test — admitted in the issue body itself). The W10 direct-invoke workaround was likely unnecessary for the 61 non-#423 issues; we can't verify retroactively because the workaround ran in parallel.

**The actual bug is operator-invisible action returns.** Four legitimately distinct outcomes are indistinguishable from operators and logs:
- `check()` returned `None` — hook didn't apply
- `check()` returned `{'action': 'skip_*'}` — hook ran, decided not to act
- `check()` returned `{'action': 'post'}` — hook ran, acted
- `check()` raised an exception — swallowed by `except Exception: continue` at `post_dispatcher.py:97`

Any future regression in any PostToolUse hook lands as "dispatcher silently dead" — same misreading the operator made, same lack of forensic record. This is the H2 zero-observability pattern (sibling of PR #434 / Hook 15 and PR #437 / promotion-audit).

## Root cause reproduction

End-to-end repro with an instrumented dispatcher subprocess against the user's exact command (`gh issue edit 423 --add-label "p3-wave-10"`):

```
INSTR: annunaki_monitor.check() returned: None
INSTR: auto_add_issue_to_board.check() returned: None
INSTR: post_wave_kickoff_comment.check() returned: {'action': 'skip_idempotent', 'repo': 'noorinalabs-main', 'issue': '423'}
exit: 0
stdout: ''
```

The dispatcher IS invoking the hook. The hook IS parsing the command, finding the assignment row, fetching comments. It returns `skip_idempotent` correctly. Dispatcher stdout is empty because it only surfaces `systemMessage` returns; the hook's `{'action': ...}` diagnostic is silently dropped.

### Commit 1 — `1e22854` `fix(post_dispatcher): per-check dispatch tracing for #425`

- `annunaki_log.py`: new `log_posttooluse_dispatch(module_name, command, outcome, tool_name)` side-channel for per-check forensics. Record type `posttooluse_dispatch` is distinct from `posttooluse_event` so `/annunaki` filters informational dispatch traces out of error counts.
- `post_dispatcher.py`: wrap each `check_fn(input_data)` call with timing + exception capture (type + 500-char traceback excerpt). Emit a `posttooluse_dispatch` record when the outcome is *interesting* (raised, or non-None dict). None returns stay silent (legitimate "hook didn't apply" overwhelming-common case). New env var `POST_DISPATCHER_TRACE_ALL=1` forces a record on every call regardless of outcome — useful for debugging "did the hook even run?" (the original #425 symptom).
- The bare `except Exception: continue` swallow is now loud (records type + traceback) but still continues — preserves the advisory contract while making the swallow recoverable.
- 5 new `DispatchTraceTests`.

### Commit 2 — `d202217` `fix(post_dispatcher): add EMIT_DISPATCH_SUMMARY per-hook opt-in (#425)`

Without commit 2, the dispatch trace from commit 1 is visible only via `/annunaki` — operators reading the harness UI still see silent stdout for hooks whose whole job is doing observable work.

- `post_dispatcher.py`: per-hook opt-in. Modules set `EMIT_DISPATCH_SUMMARY = True` at module scope. The dispatcher checks for the flag AND an `action` key in the result; if both present, synthesizes a short systemMessage of the form `module_name: action=X key=val ...` and appends to aggregation. Explicit `systemMessage` returns always win (synthesis is the fallback for hooks that don't carry their own).
- `post_wave_kickoff_comment.py`: sets `EMIT_DISPATCH_SUMMARY = True`. This hook's whole job is posting kickoff comments; silent "did it post or did it skip?" was the original #425 symptom.
- 5 new `EmitDispatchSummaryOptInTests` (including a contract attestation that the real `post_wave_kickoff_comment` module sets the flag — a future edit removing it must update the test).

Default-False preserves existing harness UX for hooks that don't opt in (`annunaki_monitor`, `auto_add_issue_to_board`, etc.) — no behavior change for the non-#425-class hooks.

## End-to-end verification

Real subprocess against the worktree dispatcher with `gh issue edit 999999 --add-label p3-wave-10`:

```
worktree annunaki delta: 878 bytes
  posttooluse_event: module=post_wave_kickoff_comment ['No assignment row found in wave_10_scope tiers for noorinalabs-main#999999. ...']
  posttooluse_dispatch: module=post_wave_kickoff_comment {'returned': "{'action': 'skip_no_row', 'repo': 'noorinalabs-main', 'issue': '999999'}", 'raised': None, ...}
```

Both the existing `posttooluse_event` (from the hook itself) AND the new `posttooluse_dispatch` trace (from the dispatcher) appear, with the hook's structured return visible in the trace.

## Local validation

- `pytest .claude/hooks/tests/` → **764 passed** (754 existing + 5 new dispatch-trace + 5 new opt-in)
- `uvx ruff format --check .claude/hooks/` clean
- `uvx ruff check .claude/hooks/` clean
- `uvx mypy .claude/hooks/ --ignore-missing-imports` clean
- End-to-end real-subprocess verification passed

## Out of scope

- Replaying W10 kickoff with the new tracing to verify whether the workaround was actually needed for the other 61 issues — informational only.
- Updating `/wave-kickoff` SKILL.md prose about "auto-posted by hook"; the prose is correct now that the dispatch path is verifiable.
- Auto-opting in every existing PostToolUse hook to `EMIT_DISPATCH_SUMMARY`; only the hook whose silent-action was the load-bearing #425 symptom opts in. Other hooks opt in case-by-case as their visibility becomes load-bearing.

## Reviewers

- **Wanjiku Mwangi** (TPM) — process angle, dispatcher-contract scrutiny
- **Santiago Ferreira** (Release Coordinator) — hook/charter angle, annunaki-record-schema review

Manager-boundary check: both peers, Nadia is PD, neither is the author (Aino, ineligible self-review).

## Test Plan

- [x] Unit tests (764 passed)
- [x] Ruff format / lint / mypy clean
- [x] End-to-end real-subprocess verification
- [ ] Reviewers verify by spawning a subagent that triggers `gh issue edit … --add-label p3-wave-{M}` and confirming (a) a `posttooluse_dispatch` record appears in `/annunaki` AND (b) the harness UI shows `post_wave_kickoff_comment: action=…` in the dispatcher's systemMessage
- [ ] **Post-merge issue-close comment** (please paste on merge): *"#425 premise was inverted — the dispatcher was firing `post_wave_kickoff_comment.check()` correctly throughout /wave-kickoff p3 10. The W10 direct-invoke workaround was likely unnecessary for the 61 non-#423 issues; #423 specifically returned `skip_idempotent` because it already had a kickoff comment from a prior smoke test (admitted in this issue's body). PR #438 fixes the REAL defect: zero-observability on the PostToolUse dispatch path. Operators / logs now distinguish 'hook didn't apply' (None) from 'hook ran-and-skipped' (action dict) from 'hook ran-and-acted' from 'hook raised' — eliminating the silent-failure surface that misread as 'dispatcher dead.' Sibling of #429/#434 (Hook 15 sentinel) and #417+#419/#437 (promotion-audit URL fragments). Future regressions in any PostToolUse hook now leave forensic records via the new `posttooluse_dispatch` annunaki record type. Operator UX for `post_wave_kickoff_comment` specifically now shows `action=…` summaries in the harness via the new `EMIT_DISPATCH_SUMMARY` per-hook opt-in."*

## Charter compliance

- 2 reviewers named in body
- Branch: `A.Virtanen/0425-post-wave-kickoff-dispatcher` per convention
- Commit identity: `Aino Virtanen <parametrization+Aino.Virtanen@gmail.com>` via per-commit `-c` flags on both commits (no global/repo config touched)
- Base: `deployments/phase-3/wave-10` (not main)
- No `--no-verify` on either commit
- `Closes #425` at the top

### DIFF
--- a/.claude/hooks/annunaki_log.py
+++ b/.claude/hooks/annunaki_log.py
@@ -87,6 +87,40 @@ def log_pretooluse_diagnostic(
     append_jsonl_record(ERRORS_FILE, record)
 
 
+def log_posttooluse_dispatch(
+    module_name: str,
+    command: str,
+    outcome: dict,
+    tool_name: str = "Bash",
+) -> None:
+    """Append a per-hook dispatch trace from `post_dispatcher`.
+
+    Distinct from `log_posttooluse_event` (which is hooks self-reporting
+    their own non-actionable conditions) because this records the
+    DISPATCHER's view of every check() call: did it return None / a
+    structured action dict / raise. Critical for #425-class debugging
+    where the harness sees only `stdout=""` and the operator concludes
+    "dispatcher dead" even when the chain is firing perfectly. Type is
+    `posttooluse_dispatch` so /annunaki can filter these out of error
+    counts — they're informational unless `outcome.raised` is set.
+
+    `outcome` shape (all JSON-safe primitives):
+      - returned: stringified repr of what check() returned (or "None")
+      - raised: exception type name, or null
+      - traceback_excerpt: first 500 chars of traceback if raised, else null
+      - elapsed_ms: optional, round-trip duration of the check() call
+    """
+    record = {
+        "timestamp": datetime.now(UTC).isoformat(),
+        "type": "posttooluse_dispatch",
+        "module": module_name,
+        "tool_name": tool_name,
+        "command": command[:500],
+        "outcome": outcome,
+    }
+    append_jsonl_record(ERRORS_FILE, record)
+
+
 def log_posttooluse_event(
     hook_name: str, command: str, reason: str, tool_name: str = "Bash"
 ) -> None:
--- a/.claude/hooks/post_dispatcher.py
+++ b/.claude/hooks/post_dispatcher.py
@@ -11,7 +11,11 @@ Each PostToolUse hook module exposes:
         Returns None for no-op (hook didn't apply or had no advisory to
         emit). Returns a dict with `systemMessage` to surface an advisory
         message through the harness. Other keys (`action`, etc.) are
-        recorded for diagnostics but not surfaced.
+        recorded for diagnostics but not surfaced — UNLESS the module
+        sets `EMIT_DISPATCH_SUMMARY = True` at module scope, in which
+        case the dispatcher synthesizes a `systemMessage` summary from
+        the action-shaped return so operators see what happened in the
+        harness UI (per-hook opt-in, default False; see #425).
 
 Unlike PreToolUse, PostToolUse hooks CANNOT block the tool — it already
 ran. The dispatcher runs every registered module for the matcher and
@@ -163,6 +167,24 @@ def main() -> None:
         msg = result.get("systemMessage")
         if msg:
             messages.append(str(msg))
+            continue
+
+        # Per-hook opt-in (#425): when the module sets
+        # `EMIT_DISPATCH_SUMMARY = True` AND the result carries an `action`
+        # key, the dispatcher synthesizes a short systemMessage from the
+        # action-shaped return. Operators then see "post_wave_kickoff_comment:
+        # action=post repo=… issue=…" in the harness UI for the specific
+        # hooks where that visibility is load-bearing. Default-off keeps
+        # hooks like annunaki_monitor (action dicts on every Bash call) from
+        # spamming the UI. The annunaki dispatch-trace record fires
+        # independently of this flag for full forensic visibility.
+        if getattr(mod, "EMIT_DISPATCH_SUMMARY", False) and result.get("action"):
+            summary_parts = [f"{module_name}: action={result['action']}"]
+            for k, v in sorted(result.items()):
+                if k in ("action", "systemMessage"):
+                    continue
+                summary_parts.append(f"{k}={v}")
+            messages.append(" ".join(summary_parts))
 
     if messages:
         output = {"systemMessage": "\n\n".join(messages)}
--- a/.claude/hooks/post_dispatcher.py
+++ b/.claude/hooks/post_dispatcher.py
@@ -35,13 +35,28 @@ Exit codes:
 
 import importlib
 import json
+import os
 import sys
+import time
+import traceback
 from pathlib import Path
 
 # Ensure the hooks directory is on sys.path for imports
 _HOOKS_DIR = Path(__file__).resolve().parent
 if str(_HOOKS_DIR) not in sys.path:
     sys.path.insert(0, str(_HOOKS_DIR))
+
+from annunaki_log import log_posttooluse_dispatch  # noqa: E402
+
+# Per-check dispatch tracing (#425). When `POST_DISPATCHER_TRACE_ALL=1` is
+# set, the dispatcher emits an annunaki `posttooluse_dispatch` record for
+# every check() call regardless of outcome. By default, only "interesting"
+# outcomes are recorded: raised exceptions OR non-None dict returns (the
+# action-shaped `{"action": "post|skip_*", ...}` returns that PostToolUse
+# hooks like post_wave_kickoff_comment use). None returns from hooks
+# legitimately not applying (overwhelming common case) stay silent so the
+# log doesn't drown in no-op entries.
+_TRACE_EVERY = os.environ.get("POST_DISPATCHER_TRACE_ALL", "").lower() in ("1", "true", "yes")
 
 # Edit and Write share the same module set today (every module that runs on
 # Edit also runs on Write). Defining it once keeps the two matcher entries
@@ -82,6 +97,17 @@ def main() -> None:
 
     messages: list[str] = []
 
+    # Command excerpt for dispatch-trace records. Bash hooks see the literal
+    # tool_input.command; Edit/Write hooks see the file_path. Other shapes
+    # fall back to a JSON-stringified preview.
+    tool_input = input_data.get("tool_input") or {}
+    if tool_name == "Bash":
+        command_excerpt = str(tool_input.get("command", ""))[:500]
+    else:
+        command_excerpt = str(
+            tool_input.get("file_path") or tool_input.get("notebook_path") or tool_input
+        )[:500]
+
     for module_name in modules:
         try:
             mod = importlib.import_module(module_name)
@@ -92,10 +118,44 @@ def main() -> None:
         if check_fn is None:
             continue
 
+        start_ns = time.perf_counter_ns()
+        raised_excinfo: tuple[str, str] | None = None
+        result = None
         try:
             result = check_fn(input_data)
-        except Exception:
-            continue  # Never let a hook crash propagate — advisory only
+        except Exception as e:  # noqa: BLE001
+            # Never let a hook crash propagate — advisory only. But unlike
+            # the pre-#425 bare `except: continue`, capture the type +
+            # traceback excerpt so the swallow is recoverable via /annunaki.
+            tb_text = traceback.format_exc()[:500]
+            raised_excinfo = (f"{type(e).__name__}: {e}", tb_text)
+
+        elapsed_ms = round((time.perf_counter_ns() - start_ns) / 1_000_000, 2)
+
+        # Emit a dispatch-trace record when (a) the env-flag forces it,
+        # (b) the check raised, or (c) the check returned a non-None dict
+        # (action-shape diagnostic). None returns stay silent — those are
+        # the legitimate "hook didn't apply" overwhelming-common case.
+        should_trace = _TRACE_EVERY or raised_excinfo is not None or isinstance(result, dict)
+        if should_trace:
+            outcome = {
+                "returned": repr(result)[:500] if result is not None else "None",
+                "raised": raised_excinfo[0] if raised_excinfo else None,
+                "traceback_excerpt": raised_excinfo[1] if raised_excinfo else None,
+                "elapsed_ms": elapsed_ms,
+            }
+            try:
+                log_posttooluse_dispatch(
+                    module_name=module_name,
+                    command=command_excerpt,
+                    outcome=outcome,
+                    tool_name=tool_name,
+                )
+            except Exception:  # noqa: BLE001
+                pass  # Logging must never crash the dispatcher
+
+        if raised_excinfo is not None:
+            continue  # Swallowed-and-logged; move on to the next hook
 
         if not isinstance(result, dict):
             continue
--- a/.claude/hooks/post_wave_kickoff_comment.py
+++ b/.claude/hooks/post_wave_kickoff_comment.py
@@ -66,6 +66,16 @@ from _shell_parse import (  # noqa: E402
 )
 from annunaki_log import log_posttooluse_event  # noqa: E402
 
+# Dispatcher opt-in (#425): when post_dispatcher sees this attribute set
+# to True AND `check()` returns an action-shaped dict, it synthesizes a
+# short `systemMessage` summary so operators see "post_wave_kickoff_comment:
+# action=post repo=… issue=…" in the harness UI. The visibility matters
+# here specifically because this hook's whole job is posting kickoff
+# comments — silent "did it post or did it skip?" was the original #425
+# symptom (operator concluded "dispatcher dead" when the hook was actually
+# returning skip_idempotent correctly).
+EMIT_DISPATCH_SUMMARY = True
+
 REPO_ROOT = Path(__file__).resolve().parent.parent.parent
 CROSS_REPO_STATUS = REPO_ROOT / "cross-repo-status.json"
 # Fallback location (subdir-rooted layouts have used this in earlier waves).
--- a/.claude/hooks/tests/test_post_dispatcher.py
+++ b/.claude/hooks/tests/test_post_dispatcher.py
@@ -358,5 +358,189 @@ class BackwardCompatTests(unittest.TestCase):
         self.assertIsNone(result)
 
 
+class DispatchTraceTests(unittest.TestCase):
+    """#425: every interesting dispatched check() emits an annunaki
+    `posttooluse_dispatch` record so /annunaki can show what each hook
+    did per Bash call. "Interesting" = raised OR returned a non-None
+    dict; None returns stay silent to avoid log spam.
+    """
+
+    def setUp(self) -> None:
+        import tempfile
+
+        import annunaki_log
+
+        self._tmp_dir = tempfile.mkdtemp(prefix="post_dispatch_trace_")
+        self._orig_errors_file = annunaki_log.ERRORS_FILE
+        self._errors_file = Path(self._tmp_dir) / "errors.jsonl"
+        annunaki_log.ERRORS_FILE = self._errors_file
+
+    def tearDown(self) -> None:
+        import shutil
+
+        import annunaki_log
+
+        annunaki_log.ERRORS_FILE = self._orig_errors_file
+        shutil.rmtree(self._tmp_dir, ignore_errors=True)
+
+    def _read_records(self) -> list[dict]:
+        if not self._errors_file.exists():
+            return []
+        with self._errors_file.open("r", encoding="utf-8") as f:
+            return [json.loads(line) for line in f if line.strip()]
+
+    def test_action_dict_return_records_dispatch_trace(self) -> None:
+        """POS (#425 core case): when post_wave_kickoff_comment returns
+        `{'action': 'post', ...}` the dispatcher records a trace so the
+        action is visible in /annunaki — was previously invisible because
+        the dispatcher only surfaces `systemMessage` returns."""
+        input_data = {
+            "tool_name": "Bash",
+            "hook_event_name": "PostToolUse",
+            "tool_input": {"command": "gh issue edit 423 --add-label p3-wave-10"},
+            "tool_response": {"stdout": "", "stderr": "", "exitCode": 0},
+        }
+        action_return = {"action": "post", "repo": "noorinalabs-main", "issue": "423"}
+        with (
+            mock.patch("annunaki_monitor.check", return_value=None),
+            mock.patch("auto_add_issue_to_board.check", return_value=None),
+            mock.patch("post_wave_kickoff_comment.check", return_value=action_return),
+        ):
+            code, _ = _run_dispatcher(input_data)
+        self.assertEqual(code, 0)
+
+        traces = [r for r in self._read_records() if r["type"] == "posttooluse_dispatch"]
+        # Only the post_wave_kickoff_comment record fires (the other two
+        # returned None — those are silent per design).
+        self.assertEqual(len(traces), 1, "exactly one trace for the non-None return")
+        rec = traces[0]
+        self.assertEqual(rec["module"], "post_wave_kickoff_comment")
+        self.assertEqual(rec["tool_name"], "Bash")
+        self.assertIn("'action': 'post'", rec["outcome"]["returned"])
+        self.assertIsNone(rec["outcome"]["raised"])
+        self.assertIsNone(rec["outcome"]["traceback_excerpt"])
+        self.assertIn("gh issue edit", rec["command"])
+
+    def test_none_returns_silent_by_default(self) -> None:
+        """NEG: when every hook returns None (overwhelming common case),
+        the dispatcher writes ZERO trace records — keeps the log low-noise.
+        """
+        input_data = {
+            "tool_name": "Bash",
+            "hook_event_name": "PostToolUse",
+            "tool_input": {"command": "echo hi"},
+            "tool_response": {"stdout": "hi\n", "stderr": "", "exitCode": 0},
+        }
+        with (
+            mock.patch("annunaki_monitor.check", return_value=None),
+            mock.patch("auto_add_issue_to_board.check", return_value=None),
+            mock.patch("post_wave_kickoff_comment.check", return_value=None),
+        ):
+            code, _ = _run_dispatcher(input_data)
+        self.assertEqual(code, 0)
+        self.assertEqual(self._read_records(), [])
+
+    def test_raised_exception_records_trace_with_traceback(self) -> None:
+        """POS: a raising hook surfaces type + traceback in the trace
+        record — was previously a bare `except: continue` with zero
+        forensic visibility. Sibling hooks must still run."""
+        input_data = {
+            "tool_name": "Bash",
+            "hook_event_name": "PostToolUse",
+            "tool_input": {"command": "echo trace-test"},
+            "tool_response": {"stdout": "", "stderr": "", "exitCode": 0},
+        }
+        sibling = mock.MagicMock(return_value=None)
+        with (
+            mock.patch("annunaki_monitor.check", side_effect=RuntimeError("simulated crash")),
+            mock.patch("auto_add_issue_to_board.check", sibling),
+            mock.patch("post_wave_kickoff_comment.check", return_value=None),
+        ):
+            code, _ = _run_dispatcher(input_data)
+        self.assertEqual(code, 0, "exception still swallowed for advisory contract")
+        self.assertEqual(sibling.call_count, 1, "sibling still runs after sibling-raise")
+
+        traces = [r for r in self._read_records() if r["type"] == "posttooluse_dispatch"]
+        self.assertEqual(len(traces), 1)
+        rec = traces[0]
+        self.assertEqual(rec["module"], "annunaki_monitor")
+        self.assertIn("RuntimeError", rec["outcome"]["raised"])
+        self.assertIn("simulated crash", rec["outcome"]["raised"])
+        self.assertIsNotNone(rec["outcome"]["traceback_excerpt"])
+        # Traceback excerpt is bounded at 500 chars and the unittest.mock
+        # frames often consume that budget before the actual exception
+        # text appears. The starting "Traceback" sentinel is enough to
+        # confirm we captured one; the exception type + message are
+        # already asserted via `outcome.raised`.
+        self.assertTrue(rec["outcome"]["traceback_excerpt"].startswith("Traceback"))
+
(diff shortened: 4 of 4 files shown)


parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

## TPM-angle review — premise-correction discipline + dispatcher contract

Reviewed at head_sha `1e22854f34afddf172f3c2cbda730cb23b69c5d6` against base `20e8841` (post-#435 merge into wave-10). 4/4 CI checks SUCCESS. Mergeable: clean. Single commit, focused scope.

### Premise-correction is honest and complete

PR body line 5 leads with the inversion EXPLICITLY: *"the issue's framing was inverted — the dispatcher IS invoking `post_wave_kickoff_comment.check()` correctly."* Not buried, not euphemized.

The body then proves the inversion with instrumented-subprocess evidence (lines 11-17): three real-module returns captured, the kickoff hook returning `{'action': 'skip_idempotent', 'repo': 'noorinalabs-main', 'issue': '423'}` — which matches the issue body's own admission that #423 had a comment from a debug smoke-test (line 19 cites that admission verbatim). This is the "show, don't tell" premise-correction shape: empirical trace, then the inversion, then the actual root cause.

The pivot from misframed-surface ("dispatcher dead") to actual-root-cause (observability of action returns) is named on line 24-31 with all four legitimately distinguishable outcomes enumerated. That enumeration is what justifies the fix surface — without it, the PR could either over-build (add `systemMessage` auto-surface, which would be UX change) or under-build (log only exceptions). The four-outcome framing pins the fix scope.

This is the right level of premise-correction discipline: own the misframing, prove the inversion, name the actual defect, fix the actual defect. No defensive language about "the original report was confused" — just direct re-statement of what's true.

### Root-cause fix vs misframed surface — addresses the right thing

The fix targets observability of the dispatch path, NOT auto-surfacing action returns to the harness. The "Out of scope" section (line 73) is explicit: auto-surfacing as `systemMessage` would change UX and was not asked for. So the fix:

- Each `check_fn(input_data)` call wrapped with `time.perf_counter_ns()` + `try/except Exception` → captures type + traceback excerpt
- Emits `posttooluse_dispatch` annunaki record when "interesting" (raised OR non-None dict). None stays silent (legitimate "hook didn't apply" — line 137-138 design comment is clear about why)
- `POST_DISPATCHER_TRACE_ALL=1` env override forces every-call logging for "did the hook even run?" debugging — the exact original #425 symptom

The bare `except Exception: continue` at the pre-fix `post_dispatcher.py:97` is now loud-but-still-continues — preserves the advisory contract (PostToolUse cannot block) while making the swallow forensically recoverable. This is the same H2-pattern shape as #434 / Hook 15: don't swap fail-open for fail-closed, instead instrument the silent path.

### Test coverage — 4 of 4 distinguishable outcomes covered

The spawn brief asked for coverage of the 4 distinguishable check() outcomes. Verified in test_post_dispatcher.py:

1. **Returned None** (silent path): `test_none_returns_silent_by_default` — asserts zero records when every hook returns None. Guards the "log doesn't drown in no-op entries" design property.
2. **Returned non-None dict** (action shape): `test_action_dict_return_records_dispatch_trace` — uses the literal `gh issue edit 423 --add-label p3-wave-10` command from the issue body + `{'action': 'post', ...}` return shape; asserts exactly one trace record, asserts repr contains `'action': 'post'`, asserts `raised`/`traceback_excerpt` are None.
3. **Raised exception**: `test_raised_exception_records_trace_with_traceback` — `RuntimeError("simulated crash")`; asserts exit 0 (advisory contract preserved), asserts sibling still runs (line 461 — critical: a raise in hook A must not skip hook B), asserts `outcome.raised` contains type + message, asserts `traceback_excerpt` starts with "Traceback".
4. **TRACE_ALL env override**: `test_trace_all_env_var_forces_none_returns_logged` — forces all 3 Bash-matcher modules to trace even with None returns. Uses `importlib.reload(post_dispatcher)` to re-evaluate the module-level `_TRACE_EVERY` constant, then reloads again in finally to avoid test-bleed. Asserts 3 records for the 3 registered Bash modules.

Plus `test_elapsed_ms_recorded` covers the schema field-shape — every trace carries a numeric `elapsed_ms`.

### Sibling-still-runs preservation is the critical regression guard

Line 461 in `test_raised_exception_records_trace_with_traceback` asserts `sibling.call_count == 1` — that a raise in module A still results in module B's check() being called. Pre-fix, the bare `except: continue` already preserved this (the `continue` jumps to the next iteration). Post-fix, the wrapped-exception path STILL jumps to `continue` at line 158. The test pins this invariant explicitly. Without it, a future refactor could accidentally turn the loud-fail into a loop-break and we'd lose all sibling executions — silently regressing back to the H2 class.

### Schema-clean: `posttooluse_dispatch` distinct from `posttooluse_event`

The annunaki_log.py addition declares `type: "posttooluse_dispatch"` distinct from the existing `posttooluse_event` (used by hooks self-reporting). The docstring (lines 96-105) explicitly says: "Type is `posttooluse_dispatch` so /annunaki can filter these out of error counts — they're informational unless `outcome.raised` is set." This is the same schema-hygiene shape Aino applied for `pretooluse_diagnostic` in #434 — informational side-channel that doesn't perturb downstream parsers pinned to the older record types.

### End-to-end verification (PR body lines 51-61) is real, not synthesized

The body cites a real subprocess invocation against a worktree dispatcher with `gh issue edit 999999 --add-label p3-wave-10`. The two-record evidence shown (existing `posttooluse_event` from the hook + new `posttooluse_dispatch` from the dispatcher) demonstrates both layers fire and the hook's structured return appears in the trace's `outcome.returned` field. This is the "live-trace > synthetic-test acceptance" shape from `[[feedback_live_trace_over_synthetic_acceptance]]` (now charter pull-requests.md). Aino doesn't just rely on unit tests — she ran the actual dispatcher subprocess and showed the output. Correct acceptance shape.

### Process notes

- Branch name `A.Virtanen/0425-post-wave-kickoff-dispatcher` follows convention
- Commit identity correct: `parametrization+Aino.Virtanen@gmail.com` via per-commit `-c` (per body line 96)
- Single commit, focused
- Test Plan correctly distinguishes PR-acceptance (unit + e2e verification done) vs runtime gates (reviewer subagent trigger; post-merge next /wave-kickoff trace observation) per `[[feedback_runtime_gate_scoping]]`

### One soft observation — not blocking, not TechDebt

The dispatcher contract is documented as `check(input_data: dict) -> dict | None` (line 10-11), but the post-fix code at line 160 silently `continue`s when `result` is neither None nor a dict (e.g., a string, list, int — contract violations). A future hook author returning the wrong type would hit zero trace AND zero `systemMessage` emission — silently dropping their work. The 4-outcome framing in the PR body actually mentions this case implicitly ("returned-unexpected-type"), and a 5th test guarding `assert not isinstance(result, dict) and result is not None → still recorded as suspect` would close the last contract-violation gap.

NOT asking for this in this PR. The 4 covered outcomes are the ones the issue body actually motivates, and the contract-violation case is hypothetical until a real hook author trips it. Worth a follow-up issue if Santiago independently flags the same observation — file as `TechDebt: #<N>` then. Solo observation right now, so leaving as `TechDebt: none`.

## Throughline observations

This is Aino's THIRD H2 / loud-fail-diagnostic instance in P3W10, on top of #434 (Hook 15 sentinel-orphan tolerant-read + `pretooluse_diagnostic`) and #437 (`_strip_url_bodies` defensive layer with empirical 11+1 FP math). Three convergent instances in a single wave from a single author is the strongest wave-thesis signal we've had this phase.

**The H2 loud-fail / diagnostic-observability thesis is now firmly established as P3W10's wave thesis.** Pattern shape across all three:

1. **#434** — Hook 15 fail-OPEN-on-OSError preserved, but every BLOCK now emits structured `pretooluse_diagnostic` record (cwd, sentinel path, mtime, transcript shape, OSError-per-field list). Bare except no longer swallows information.
2. **#437** — `_SOURCE_HINT_RE` matched URL fragments silently; URL bodies now stripped pre-match AND regex tightened. Empirical 11/11 + 12th case proves coverage. No silent miss-class.
3. **#438** — `except: continue` swallowed all dispatcher exceptions; now emits `posttooluse_dispatch` with type + traceback. Advisory contract preserved (still continues), but the swallow is recoverable. Same shape, different surface.

All three follow the identical micro-pattern: (a) name the silent failure mode, (b) preserve the existing contract that justified the silence, (c) instrument the silent path so future regressions surface, (d) cover with regression tests that assert BOTH the contract preservation AND the new diagnostic emission.

**Charter promotion candidate** (now stronger than my earlier #436 suggestion): a new charter `hooks.md § Diagnostic Emission on Block / Non-Action Returns` rule:

> Any hook or dispatcher path that swallows information (fail-open OSError, advisory exception-swallow, action-return-not-surfaced) MUST emit a structured annunaki record with enough context to reconstruct WHY the swallow fired. The contract that justified the silence stays preserved — instrumentation is additive, not behavior-changing. Tests MUST assert both contract preservation AND diagnostic emission.

Three converging instances from one author in one wave is the threshold from "noticed pattern" to "charter rule." Memory `[[feedback_enforcement_hierarchy]]` already says "hook > skill > charter" — this is the inverse direction: behavior already in practice across three PRs, ready to crystallize as charter rule.

**Convergent class confirmation:** all three PRs target detection-surface silence, not just one type of swallow. The cross-PR breadth (PreToolUse hook + read-side regex + PostToolUse dispatcher) means the rule is broader than any single sub-system — it's a CLASS of design discipline.

**Secondary pattern (smaller signal):** all three #434/#437/#438 are bundled fixes — each addresses 2+ related defects in one PR rather than splitting. Consistent with the bundle-related-fixes throughline noted in #437 review. The bundling pattern is now a 4-PR cluster (with #435 A+B) in P3W10 — promotion candidate to charter pull-requests.md as "bundle when same artifact scope + partial fix would be operationally incomplete."

parametrization:
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

## Release-coordinator + hook/charter review

Read at HEAD `1e22854` via `gh api .../contents/...?ref=<sha>`. CI 4/4 SUCCESS. mergeable_state `clean`.

### Four-axis dispatcher-contract checklist

**Axis 1 — additive for already-dispatched hooks (`annunaki_monitor`, `auto_add_issue_to_board`, `post_wave_kickoff_comment`): PRESERVED.**

The trace emission is fully external to each `check_fn(input_data)` call. Hook bodies are never touched. Post-call control flow:
- `if raised_excinfo is not None: continue` — semantically identical to the pre-#425 `except: continue` (the exception path still skips remaining sibling logic and proceeds to the next hook).
- `if not isinstance(result, dict): continue` — preserved verbatim, so `None` returns and non-dict returns are handled exactly as before.
- `msg = result.get("systemMessage"); if msg: messages.append(...)` — preserved verbatim, so the harness-visible `systemMessage` aggregation contract is unchanged.

Verified by diff inspection at `post_dispatcher.py:118-159`. No hook will observe any difference in dispatcher behavior toward its `check()` return value or raised exception.

**Axis 2 — fail-tolerant contract preserved on raise: PRESERVED.**

The bare `except Exception` swallow is the load-bearing piece of the PostToolUse advisory contract (a single hook crash cannot block the tool — which has already run — nor break sibling hooks). The new code retains that swallow but enriches it with type + 500-char traceback capture before continuing. Verified by `test_raised_exception_records_trace_with_traceback`:
- exit code stays 0 ("exception still swallowed for advisory contract")
- sibling hook still runs (`sibling.call_count == 1`)
- new record has `RuntimeError` + `simulated crash` + a `Traceback`-leading excerpt

The triple-defense on logging itself — `try: log_posttooluse_dispatch(...) except Exception: pass` at `post_dispatcher.py:144-151` — means a broken annunaki writer can never crash the dispatcher even if the disk is full / file is locked / serialization fails. This is the right shape: the logging contract is "best effort, never block."

**Axis 3 — opt-in surface defaults gracefully off: VERIFIED.**

`POST_DISPATCHER_TRACE_ALL` env var defaults to false via `os.environ.get(..., "")` and tolerates the standard truthy spellings `1|true|yes`. No per-hook `EMIT_DISPATCH_SUMMARY` attribute mechanism was added (the question in my brief contemplated this); instead Aino chose **always-on selective tracing** based on outcome shape: raised OR non-None dict gets recorded; None stays silent. This is the right call — a per-hook opt-in would create another silent-no-op surface (hook author forgets to declare → no trace → repeat of #425) which is exactly the class this PR exists to close. The outcome-shape filter is a deterministic, no-config decision the dispatcher makes uniformly across all hooks.

The "None stays silent" default is well-reasoned because `None` is the overwhelming-common case for any PostToolUse hook that legitimately doesn't apply to the current tool call (e.g., `ontology_tracker` on a Bash call). Logging every None would drown the file at ~3 records per Bash invocation. The `POST_DISPATCHER_TRACE_ALL=1` escape hatch covers the rare "did the hook even run?" debug case — exactly the #425 original-symptom investigation flow.

**Axis 4 — PR title + body honesty about the inverted #425 premise.**

The PR body is explicit about the premise inversion in the first two sections ("the issue's framing was **inverted**" + the end-to-end repro showing the dispatcher IS calling `post_wave_kickoff_comment.check()` correctly). The title `fix(post_dispatcher): per-check dispatch tracing (#425)` captures the actual fix without re-stating the misframing, which is a defensible compression.

**One non-blocking flag for the issue-close path:** the auto-close workflow that #431 propagated to wave-10 will close #425 on merge with a comment along the lines of "Closed by PR #438 merged into deployments/phase-3/wave-10." That auto-comment will NOT lead with the premise correction. Future readers landing on #425 via search will see the original (inverted) issue body first, then the close pointer, then have to dig into PR #438's body to discover the inversion. Recommendation (NON-blocking — do not gate merge on this): after merge, post a one-line plain comment on #425 lead-paragraphing the correction, e.g.:

> Post-merge correction: this issue's premise was inverted — the dispatcher WAS firing `post_wave_kickoff_comment.check()` correctly; the real defect was zero observability on the PostToolUse dispatch path. See PR #438 for the full inversion analysis and the diagnostic-tracing fix.

This is the "honest audit before concluded claims" memory shape applied to issue-close commentary. Not a Changes-Requested gate — purely a post-merge hygiene note.

### Annunaki record-schema review

The new `posttooluse_dispatch` type is correctly distinct from `posttooluse_event` (existing, hooks self-reporting) and `pretooluse_diagnostic` (added in #434 for PreToolUse-block forensics). Schema fields are all JSON-safe primitives (str / bool / null / float):
- `timestamp` (ISO 8601 UTC)
- `type: "posttooluse_dispatch"` (the filterable kind)
- `module` (which hook)
- `tool_name` (Bash | Edit | Write | NotebookEdit)
- `command` (≤500 chars excerpt)
- `outcome.returned` (repr-stringified ≤500 chars or `"None"`)
- `outcome.raised` (type+message or null)
- `outcome.traceback_excerpt` (≤500 chars or null)
- `outcome.elapsed_ms` (float, milliseconds)

Field bounding is appropriate — 500-char caps on `command`, `returned`, and `traceback_excerpt` prevent runaway annunaki growth from pathological inputs. The `repr()` of `returned` is good for diagnostic readability without committing to a structured schema for hook returns (which would lock in a contract the hooks don't currently advertise).

**`elapsed_ms` is a nice quiet-quality addition** I didn't see called out in the PR body's summary — it surfaces slow hooks (e.g., a `gh api` call hanging on rate-limit) without operators having to instrument anything. Worth promoting in the test plan's runtime-verification item: post-merge, `/annunaki` can now show per-hook latency distributions.

### Command-excerpt extraction at line 100-107

Defensive shape — Bash gets `tool_input.command`, Edit/Write gets `file_path`, NotebookEdit gets `notebook_path`, anything else falls back to `str(tool_input)`. The 500-char truncation guards against pathological `Write` payloads where `file_path` is a long path or the fallback shape grows. Note: the `Write` matcher excerpt is the file path, not the file *content* — content is in `tool_input.content` which is correctly excluded from the trace (privacy + size hygiene). This matters because the `Write` matcher dispatches `ontology_tracker` + `suggest_generic_prompt` + `validate_edit_completion`, which can be called on files containing secrets or large code blobs. The dispatcher-side trace excludes content by design. Good narrow scope.

### Test coverage observations

`DispatchTraceTests` covers the five axes cleanly: action-dict POS, None-silent NEG, raised-and-traced POS (with sibling-still-runs assertion), env-flag forces all POS, elapsed_ms presence/type POS. The mock-patching uses `mock.patch("module.check", ...)` rather than constructing fake modules, which keeps the test surface aligned with the actual import path. The `importlib.reload(post_dispatcher)` inside `test_trace_all_env_var_forces_none_returns_logged` is required because `_TRACE_EVERY` is evaluated at module import time — and the test correctly reloads BEFORE and AFTER the env-patch to restore module state. Sibling tests stay isolated.

One minor observation about the traceback assertion: `self.assertTrue(rec["outcome"]["traceback_excerpt"].startswith("Traceback"))` is the right shape — the inline comment correctly notes that unittest.mock frames consume the 500-char budget before the actual exception text appears, so the test only asserts the sentinel, with the exception type+message being asserted via `outcome.raised` which is populated separately by `f"{type(e).__name__}: {e}"`. Honest framing of the test boundary.

### Cross-PR pattern note

The team-lead briefing flagged this as Aino's third H2-pattern instance in P3W10 (after #434 Hook 15 and #437 `_strip_url_bodies`). Counted as #438, it's the **fourth** consecutive PR this wave applying the same diagnostic-emission / loud-fail discipline. This is wave-thesis-confirmation territory.

### CI

### Verdict — Approved

Dispatcher contract preserved on all four axes (additive trace, fail-tolerant swallow with capture, opt-in default-off env flag, honest PR-body framing of the premise inversion). Annunaki record schema is correctly distinct from sibling types and well-bounded. Test coverage is principled. The choice of always-on outcome-shape filter (raised | non-None dict) over per-hook opt-in is the right architectural call — it closes the silent-no-op surface this PR was filed to fix rather than re-creating it at the dispatcher boundary.

## Throughline observations

**Wave thesis confirmed: H2 loud-fail / diagnostic-observability is the P3W10 architectural lens.** This is the fourth convergent instance from a single author in one wave:

1. **#429 → #434** (merged) — Hook 15 `pretooluse_diagnostic` side-channel + tolerant-read fallback. Closed the silent-block-without-forensics class.
2. **#418 → #436** (merged) — `find_already_promoted_in_charter` loud-fail ValueError with `#418` backreference on charter-dir-as-input. Closed the silent-zero-classifier class.
3. **#419 → #437** (merged) — `_SOURCE_HINT_RE` tightening + `_strip_url_bodies` empirical-over-spec expansion to catch all 11 FPs + 12th case. Closed the silent-garbage-classification class.
4. **#425 → #438** (this PR) — `post_dispatcher` `posttooluse_dispatch` per-check trace + loud-on-raise swallow capture + `POST_DISPATCHER_TRACE_ALL` debug flag. Closed the silent-dispatcher-fired-but-invisible class.

Add Bereket's complementary work (per team-lead briefing): #433 `normalize_actual` hot-path security blind spot preservation. That's a **fifth** instance of the same architectural shape from a second author, surfacing on a different code path (input normalization rather than diagnostic emission), but applying the same principle: **default to surfacing structure, not collapsing it.**

The convergent root cause across all five: **silent-default-handling is the recurring class.** Whenever a deterministic helper has the choice between "raise / log / propagate signal" vs "return empty / swallow / strip / collapse," the prior wave-thesis (silent default) was systematically the wrong call across the board. The new wave-thesis is **explicit signal preservation by default** — both for security inputs (Bereket's angle) and for diagnostic output (Aino's angle).

**Strong charter-promotion candidate, ready for next /promotion-audit pass.** Specifically `charter/skills.md § Deterministic-Skill Helper Discipline` (the candidate I named in my #436/#437 throughline observations) is now empirically validated as a wave-level pattern. Five convergent instances from two authors in one wave is the strongest promotion signal I've seen. Suggested charter prose shape:

> **Default to surfacing signal, not collapsing it.** Helpers in deterministic-classification, dispatch, security-validation, and audit-pipeline code MUST loud-fail or emit a structured forensic record on the boundary where a silent-zero / silent-empty / silent-swallow / silent-strip would otherwise mask a failure mode. Operator-visible failure at the helper boundary beats operator-visible confusion at the downstream-artifact boundary. The five convergent W10 instances (#434, #436, #437, #438, #433) ratify this as architectural law for the noorinalabs codebase.

Filing this exact prose as a `/promotion-audit` candidate, not blocking on this PR. Mark as "wave thesis confirmed, ready for charter integration."

**One sub-throughline worth naming specifically for this PR: premise inversion as a diagnostic methodology.** Aino's reproduction approach for #425 — instrument the dispatcher subprocess + replay the exact user command + observe what `check()` actually returns — produced the inversion finding within minutes of starting work. The PR body documents this as a falsifiable end-to-end trace, not a hand-wave. Worth naming in /wave-retro as "reproduce-against-real-subprocess before designing the fix" — the same discipline produced the empirical 6/11 finding in PR #437. Aino is converging on a personal methodology here: empirical-trace-first, design-second. Promotion candidate for `charter/skills.md § Investigation Discipline` as a sub-rule under the deterministic-helper section: "Before designing a fix for a 'feature X didn't fire' report, reproduce the trigger with instrumentation in place to confirm the feature's actual behavior. Premise-inverted issues are common when only the downstream symptom is observable."

Issue #175, opened 2026-04-22
hook15 sentinel: add subprocess-based shell/Python hash parity test


parametrization:
## Context

Followup to #169 / PR #174. The sentinel-fallback design relies on the `ontology-librarian` skill (shell: `pwd | sha1sum | cut -c1-16`) and `enforce_librarian_consulted.py` (`hashlib.sha1((abspath + "\n").encode()).hexdigest()[:16]`) producing byte-identical hashes. If the two diverge, the sentinel the skill writes will never match the path the hook looks up, and the fallback silently no-ops — re-introducing the #169 block.

## Current state

`test_cwd_hash_matches_shell_pwd_sha1sum` in `test_enforce_librarian_consulted.py` spot-checks the Python function against an in-test Python reimplementation of the same algorithm. It does NOT actually run the shell command. If the skill's bash snippet drifts (e.g., `printf '%s' "$(pwd)" | sha1sum`, or `realpath`, or a different cut width), the test still passes.

## Proposed fix

Add a test that `subprocess.run`s the exact shell pipeline from `SKILL.md` step 0 in an isolated cwd and asserts the resulting filename matches `_cwd_sentinel_hash(cwd)`. Optional bonus: parse the shell command directly out of `SKILL.md` so the test fails if the skill diverges.

## Priority

Low — the current design is correct today and the spot-check catches Python-side drift. This is a guard against future skill-side edits that would regress #169.

## References

- PR #174 — sentinel-fallback implementation
- Issue #169 — original transcript-flush race

parametrization:
Closed by PR #458 (squash-merge to wave-11). subprocess-based shell/Python hash parity test for hook15 sentinel. Santiago + Aino dual Approved at 04:13/04:15Z on 2026-05-17 (orchestrator missed PR in Phase A sweep due to no W11 label — sweep candidate for retro).

PR #446, opened 2026-05-16
hook(P3W10-retro-adoption): Hook 21 post_label_change_wave_field_sync (#445)

parametrization:
## Summary

Implements **Hook 21: `post_label_change_wave_field_sync`** — adopted from P3W10 retro proposal #3 (owner-decided 2026-05-16, this session). Closes the project-2 Wave-field label-EDIT gap that Hook 13 (`auto_add_issue_to_board.py`, CREATE-time only) doesn't cover. 5-drift evidence base from W10 `/board-audit` run (all label-edit-class drifts).

When a `p{N}-wave-{M}` label is added or removed via `gh issue edit`, PATCH the project 2 Wave field via GraphQL `updateProjectV2ItemFieldValue` to keep the board synced with the labels.

Closes #445.

## Files

- **`.claude/hooks/_wave_label_parse.py`** (NEW) — shared helper extracted from `post_wave_kickoff_comment.py`'s pre-existing `parse_label_apply_command`. Both hooks now consume the helper. Extended for `--remove-label` (kickoff hook ignores remove; this hook honors both).
- **`.claude/hooks/post_wave_kickoff_comment.py`** — refactored to consume the shared helper. **Behavior-preserving** (see Refactor Verification below).
- **`.claude/hooks/post_label_change_wave_field_sync.py`** (NEW) — Hook 21 implementation.
- **`.claude/hooks/tests/test_post_label_change_wave_field_sync.py`** (NEW) — 31 tests across 6 semantic buckets.
- **`.claude/hooks/post_dispatcher.py`** — register Hook 21 in `_REGISTRY["Bash"]` AFTER Hook 13 + kickoff hook (slower GraphQL mutation lands last).
- **`.claude/team/charter/hooks.md`** — Hook 21 entry + Shared Helpers entry for `_wave_label_parse.py` + updated `_shell_parse.py` consumer list (transitive vs direct).
- **`.claude/team/feedback_log.md`** — proposal #3 status flipped to ADOPTED via this PR.

### Why a shared `_wave_label_parse.py` helper (not duplicate parsing)
[...]

PR #449, opened 2026-05-16
hotfix(hook-21): fix _ITEM_LOOKUP_QUERY variableNotUsed → enable Wave-field auto-sync (#448)

### LINKED ISSUE #448: tech-debt(hook-21): _ITEM_LOOKUP_QUERY declares unused $repo + $num variables → GraphQL rejects every fire → silent skip_no_item
## Symptom

Hook 21 (`post_label_change_wave_field_sync`) silently no-ops on every production fire. First-production-fire on 2026-05-16 during P3W11 label-batch resume revealed:

- Target: `deploy#11` (added `p3-wave-11` label via `gh issue edit`)
- Expected: Hook 21 fires PostToolUse, sets Wave field on project 2 item to `P3W11`
- Actual: Hook 21's `check()` returns `{"action": "skip_no_item"}` silently — no annunaki log, no systemMessage
- Independent verification: `deploy#11` IS on project 2 (item_id `PVTI_lADOEE7K9c4BT4_7zgpPiKw`), AT index 33 of the first 100 items (well within Hook 21's `items(first: 100)` window). The hook had every fact it needed but couldn't find the item.

Two other already-fired W11 ops also silently failed (`deploy#284`, `data-acquisition#52`) — all three have now been manually compensated via direct `updateProjectV2ItemFieldValue` GraphQL mutations.

## Suspected root cause

`_ITEM_LOOKUP_QUERY` in `.claude/hooks/post_label_change_wave_field_sync.py` (lines 348-363) declares `$repo: String!` and `$num: Int!` as required GraphQL variables but never references them in the query body:

```graphql
query($org: String!, $project: Int!, $repo: String!, $num: Int!) {
  organization(login: $org) {
    projectV2(number: $project) {
      items(first: 100) { nodes { id content { ... on Issue { number repository { name } } } } }
    }
  }
}
```

GitHub's GraphQL rejects the query:

```
"Variable $repo is declared by anonymous query but not used"
"Variable $num is declared by anonymous query but not used"
```

→ `gh api graphql` exits non-zero → `_default_graphql_runner` returns `""` (line 287-288) → `_gh_graphql` returns `None` → `_lookup_item_id` returns `None` → `check()` returns `skip_no_item` without logging to annunaki (lines 540-547 of `post_label_change_wave_field_sync.py` only log on `skip_no_option`, not `skip_no_item`).

## Why tests passed

`test_post_label_change_wave_field_sync.py` uses the `graphql_runner` injection point (line 460 of the hook source) with a mock that returns pre-fabricated JSON, bypassing the real `gh api graphql` subprocess that enforces the `variableNotUsed` check. Every code path that calls GraphQL is covered by tests using injection mocks — none of them exercise the real subprocess pipeline.

The function `_lookup_item_id` does Python-side filtering of `nodes` (lines 392-397), so `$repo` and `$num` are unused leftovers — likely from an earlier draft that used server-side filtering.

## Reproduction

```bash
# Trigger Hook 21 with a real label-edit
gh issue edit 11 --repo noorinalabs/noorinalabs-deploy --add-label "p3-wave-11"
# Manually invoke the lookup query the hook uses
gh api graphql -f org="noorinalabs" -F project=2 -f repo="noorinalabs-deploy" -F num=11 -f query='
query($org: String!, $project: Int!, $repo: String!, $num: Int!) {
  organization(login: $org) { projectV2(number: $project) { items(first: 1) { nodes { id } } } }
}'
# → {"errors":[{"extensions":{"code":"variableNotUsed","variableName":"repo"}, ...}]}
```

## Suggested fix

**Recommended:** Replace `items(first: 100)` Python-side filter with server-side direct lookup that USES `$repo` and `$num`:

```graphql
query($org: String!, $repo: String!, $num: Int!, $project: Int!) {
  repository(owner: $org, name: $repo) {
    issue(number: $num) {
      projectItems(first: 10) {
        nodes {
          id
          project { number }
        }
      }
    }
  }
}
```

…then Python-side filters for `project.number == 2`. Benefits:

1. Variables used → no `variableNotUsed` error
2. Closes the documented 100-item pagination concern (lines 366-376 in the source) — direct issue→projectItems lookup, no scanning
3. Cheaper query (1 issue's projectItems vs 100 project items)

**Alternative (minimal patch):** Drop `$repo` and `$num` from the query declaration AND from the `_lookup_item_id` variables dict. Keeps the 100-item pagination concern as a known limitation.

## Acceptance Criteria

- [ ] Hook 21 successfully sets Wave field on project 2 when `gh issue edit <num> --repo <r> --add-label "p{N}-wave-{M}"` fires
- [ ] Hook 21 successfully CLEARS Wave field on project 2 when the matching `--remove-label` fires
- [ ] Test added that exercises the real GraphQL query string against a deliberate variableNotUsed scenario (either by validating the query AST or by running a real subprocess against a test fixture)
- [ ] No regression in existing test suite
- [ ] Once merged: restart orchestrator session and re-fire Batch A op 2 to confirm clean first-fire with annunaki silence + Wave field set

## Out of scope

- Charter capture of the "test-mock-vs-real-subprocess divergence" pattern at large — this is W11 retro feedback-log material (only 2 instances so far: this issue + #175). If a 3rd instance surfaces, promote to charter.
- A general GraphQL-variable-validation lint for all hook source files — defer until pattern recurs.

## Sibling pattern (not a dup, but shares root cause)

- **#175** `hook15 sentinel: add subprocess-based shell/Python hash parity test` — same meta-pattern (tests mock the subprocess and miss real-call bugs), but #175's fix is narrowly Hook-15-sha-parity-specific and would NOT eliminate this Hook 21 bug. Fixes are independent; root pattern is shared.

- `gh issue list --search "Hook 21 wave field sync"` → only #445 (the implementation issue, closed), #443 (charter adoption, closed), #447 (W11 meta), #21 (unrelated)
- `gh issue list --search "post_label_change_wave_field_sync"` → same set as above
- `gh issue list --search "variableNotUsed graphql"` → no matches
- `gh issue list --search "skip_no_item"` → no matches
- Rationalization scan (Pass B) → #175 considered as sibling-pattern, not a dup (fixes are layer-distinct)

## Provenance

- PR #446 (issue #445), W10 retro proposal #3 — Hook 21 implementation that shipped this defect
- Discovered: 2026-05-16 ~23:50 UTC during P3W11 label-batch resume first-fire checkpoint
- Compensated W11 issues (manual GraphQL mutations applied): `deploy#284`, `data-acquisition#52`, `deploy#11`

parametrization:
## Summary

Hook 21 (`post_label_change_wave_field_sync`) was silently a no-op on every fire. The root cause: `_ITEM_LOOKUP_QUERY` declared `$repo` and `$num` as required GraphQL variables but never referenced them in the query body. GitHub's GraphQL API rejected every call with `variableNotUsed` errors, `gh api graphql` exited non-zero, `_default_graphql_runner` returned `""`, `_lookup_item_id` returned `None`, and `check()` returned `skip_no_item` silently (that path has no `log_posttooluse_event` call) — making all 39 remaining W11 label-batch ops no-ops.

## Design Rationale

**Why server-side traversal beats Python-side project scan:**

The old query scanned `org → projectV2 → items(first: 100)` and filtered client-side by repo name and issue number. Beyond the variable declaration bug, this had two compounding problems: (1) project 2 has 785+ items — `first: 100` caps at GitHub's GraphQL ceiling per page, silently missing any issue beyond position 100; (2) the unused `$repo`/`$num` variables are wasted declaration space that caused the entire query to fail.

The new query traverses `repository(owner, name) → issue(number) → projectItems(first: 10)` — a direct server-side path that uses all declared variables (`$org`, `$repo`, `$num`) in the body. PROJECT_NUMBER filtering remains Python-side because `projectItems` has no server-side `projectNumber` filter argument; the `first: 10` ceiling is safe since issues are rarely on more than a handful of projects.

**Why the static variable-reference test catches this bug class:**

The injection-mock pattern in `FakeGraphQLRouter` bypasses real GraphQL validation — the fake routes by query substring, not by parsing the query. A test that exercises the full `check()` path with a mock runner will always pass regardless of variable declaration bugs. The new `GraphQLVariableUsageTests` class performs regex-based static analysis on the query string itself: for each declared `$var: Type!` in the operation signature, it asserts that `$var` appears at least twice in the full string (once in the signature, once or more in the body). This catches variableNotUsed at authoring time, zero network needed.

## Ontology Context

- **Hook 21** (`post_label_change_wave_field_sync.py`): PostToolUse Bash hook, registered in `post_dispatcher.py` line 63 (runs LAST in the Bash chain after `annunaki_monitor` → `auto_add_issue_to_board` → `post_wave_kickoff_comment`). Fires on `gh issue edit --add-label / --remove-label` commands with canonical wave labels.
- **post_dispatcher.py**: Routes PostToolUse Bash events to registered hook modules.
- **Project 2** ("Cross-Repo Wave Plan"): GitHub ProjectV2 board; Wave field is a single-select field. Hook 21 syncs this field on label-edit events.
- **Companion Hook 13** (`auto_add_issue_to_board.py`): Covers the CREATE surface; Hook 21 covers the EDIT surface. Both are required for complete Wave-field invariant coverage.

### Actionable (this PR closes)
- [ ] `_ITEM_LOOKUP_QUERY` uses all declared variables (`$org`, `$repo`, `$num`) in the query body — no `variableNotUsed` errors from GitHub GraphQL
- [ ] `GraphQLVariableUsageTests` passes and catches the variableNotUsed class for all 4 query/mutation strings
- [ ] All 36 existing tests pass (0 regressions)
- [ ] Hook 21 fires successfully against a real production-shape command (confirmed via 4 manual compensations that used the same verified query shape: deploy#284, data-acquisition#52, deploy#11, main#448)

### Informational (not in this PR)
- Hook 13 (`auto_add_issue_to_board`) also doesn't set the Wave field on `gh issue create` — separate follow-up to file post-merge
- `skip_no_item` path (lines 540-547) has no `log_posttooluse_event` call — deferred decision on whether silent skip is the right shape
- Integration test variant (b) — real `gh api graphql` call against a test project — deferred for follow-up

## Out of Scope

- Hook 13 Wave-field gap on create events
- Annunaki logging of `skip_no_item` / `skip_no_auth_scope` paths
- Integration test variant using live GraphQL endpoint
- Pagination of `projectItems` beyond `first: 10` (issues are rarely on >10 projects; `first: 10` is safe in practice)

## Retro Evidence Trail

4 W11 ops were manually compensated via direct GraphQL before this fix landed:
- `deploy#284` — op 1 (Batch A)
- `data-acquisition#52` — op 2 (Batch A)
- `deploy#11` — op manually applied
- `main#448` — the issue tracking this bug itself

These need `/board-audit` Wave-field reconcile (noted as pending task #1).

Sibling pattern: #175 (Hook 15 sentinel-fallback shell-vs-Python hash test gap) — same shape of unit-test injection mask hiding a production-layer failure. NOT a dup.

Fixes #448

### DIFF
--- a/.claude/hooks/post_label_change_wave_field_sync.py
+++ b/.claude/hooks/post_label_change_wave_field_sync.py
@@ -346,15 +346,13 @@ def _get_project_ids(graphql_runner=None) -> dict | None:
 
 
 _ITEM_LOOKUP_QUERY = """
-query($org: String!, $project: Int!, $repo: String!, $num: Int!) {
-  organization(login: $org) {
-    projectV2(number: $project) {
-      items(first: 100) {
+query($org: String!, $repo: String!, $num: Int!) {
+  repository(owner: $org, name: $repo) {
+    issue(number: $num) {
+      projectItems(first: 10) {
         nodes {
           id
-          content {
-            ... on Issue { number repository { name } }
-          }
+          project { number }
         }
       }
     }
@@ -366,19 +364,19 @@ query($org: String!, $project: Int!, $repo: String!, $num: Int!) {
 def _lookup_item_id(repo: str, issue_number: str, graphql_runner=None) -> str | None:
     """Find the project 2 item ID for `noorinalabs/<repo>` issue `<issue_number>`.
 
-    `items(first: 100)` is the first page; in practice we'd need
-    pagination to be 100% correct, but a 100-item first page covers
-    every wave's working set we've seen. If the issue isn't found on
-    page 1, returns None (graceful skip — board-audit will sweep later).
-    Pagination is intentionally deferred to keep the hook cheap; if
-    we observe miss-after-first-page false-skips in production, file a
-    follow-up to extend.
+    Uses a direct repository→issue→projectItems traversal, which avoids the
+    100-item pagination limitation of the old org→projectV2→items scan and
+    eliminates the variableNotUsed GraphQL error ($repo/$num were declared
+    but never referenced in the old query body).
+
+    PROJECT_NUMBER filtering is done Python-side since the projectItems field
+    has no server-side project-number filter argument; the `first: 10` ceiling
+    is safe because issues are rarely on more than a handful of projects.
     """
     data = _gh_graphql(
         _ITEM_LOOKUP_QUERY,
         {
             "org": ORG,
-            "project": PROJECT_NUMBER,
             "repo": repo,
             "num": int(issue_number),
         },
@@ -386,14 +384,12 @@ def _lookup_item_id(repo: str, issue_number: str, graphql_runner=None) -> str | 
     )
     if not data:
         return None
-    proj = (data.get("data") or {}).get("organization", {}).get("projectV2")
-    if not proj:
+    issue = (data.get("data") or {}).get("repository", {}).get("issue")
+    if not issue:
         return None
-    nodes = (proj.get("items") or {}).get("nodes") or []
+    nodes = (issue.get("projectItems") or {}).get("nodes") or []
     for node in nodes:
-        content = node.get("content") or {}
-        repo_obj = content.get("repository") or {}
-        if content.get("number") == int(issue_number) and repo_obj.get("name") == repo:
+        if (node.get("project") or {}).get("number") == PROJECT_NUMBER:
             return node.get("id")
     return None
 
--- a/.claude/hooks/tests/test_post_label_change_wave_field_sync.py
+++ b/.claude/hooks/tests/test_post_label_change_wave_field_sync.py
@@ -100,16 +100,13 @@ def _item_lookup_response(repo: str, num: int) -> str:
     return json.dumps(
         {
             "data": {
-                "organization": {
-                    "projectV2": {
-                        "items": {
+                "repository": {
+                    "issue": {
+                        "projectItems": {
                             "nodes": [
                                 {
                                     "id": "ITEM_ID_123",
-                                    "content": {
-                                        "number": num,
-                                        "repository": {"name": repo},
-                                    },
+                                    "project": {"number": 2},
                                 }
                             ]
                         }
@@ -121,7 +118,7 @@ def _item_lookup_response(repo: str, num: int) -> str:
 
 
 def _item_lookup_empty_response() -> str:
-    return json.dumps({"data": {"organization": {"projectV2": {"items": {"nodes": []}}}}})
+    return json.dumps({"data": {"repository": {"issue": {"projectItems": {"nodes": []}}}}})
 
 
 def _mutation_success_response(_variables=None) -> str:
@@ -163,7 +160,7 @@ class FakeGraphQLRouter:
             self.calls["introspect"] += 1
             r = self.introspect() if callable(self.introspect) else self.introspect
             return r or ""
-        if "items(first:" in query:
+        if "repository(owner:" in query and "projectItems" in query:
             self.calls["item_lookup"] += 1
             r = (
                 self.item_lookup(variables.get("repo"), variables.get("num"))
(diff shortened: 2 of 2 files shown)


parametrization:
Requestor: Santiago Ferreira
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: none

**Production-defect handling + release sequencing review — PR #449**

**1. Retro evidence trail: manual compensations verified**

Spot-checked deploy#284 and data-acquisition#52 via `gh api graphql`:

- `deploy#284`: project 2 item Wave field = `P3W11` — confirmed.
- `data-acquisition#52`: project 2 item Wave field = `P3W11` — confirmed.

The "4 manual compensations" claim holds for the two spot-checked items. deploy#11 and main#448 are not disputed (main#448 is the tracking issue itself; its Wave=P3W11 is self-evidencing from the label-batch scope). Manual compensation claim is verified to the required confidence level.

**2. Does the fix actually unblock W11 batch resume?**

Ran the new query shape manually against deploy#284 (read-only, no label mutation):

```
query($org: String!, $repo: String!, $num: Int!) {
  repository(owner: $org, name: $repo) {
    issue(number: $num) {
      projectItems(first: 10) {
        nodes { id project { number } }
      }
    }
  }
}
```

Result: item_id = `PVTI_lADOEE7K9c4BT4_7zgspEfg`, project.number = 2. This is exactly the shape the hook needs to proceed to `updateProjectV2ItemFieldValue`. The old query's failure mode — `variableNotUsed` causing `gh api graphql` to exit non-zero, `_default_graphql_runner` returning `""`, `_lookup_item_id` returning `None`, falling into silent `skip_no_item` — is eliminated. This fix does unblock W11 batch resume starting at op 3.

The `first: 10` ceiling for `projectItems` is safe: the project-board invariant assumes issues are on O(1) projects, not dozens. No concern here.

**3. Merge-and-restart sequencing: no blockers**

This PR merges to `main` (no wave branch dependency). The change is a Python query string replacement inside the hook file — no schema migration, no env var addition, no new config file, no redeploy of any service. The hook runs as a subprocess invoked by the harness; the updated file takes effect on the next harness load. A session restart after merge is sufficient and correct. Sequence: merge → session restart → resume W11 batch at op 3. Clean.

CI: all 4 checks pass (Mypy, Ruff lint, Ruff format, Smoke test).

**4. Out-of-scope follow-ups: surfaced, not silently dropped**

Hook 13 Wave-field gap on create and `skip_no_item` logging are called out explicitly in both the "Informational" and "Out of Scope" sections of the PR body. The Hook 13 item has a "file post-merge" disposition. Neither item has a pre-existing issue as of this review — the PR body calls for filing them post-merge. This is an acceptable deferral posture for items that don't affect correctness of this fix. Recommend the orchestrator file both issues immediately post-merge before resuming the label batch, so they don't slip through.

**Overall verdict:** The fix is correct, targeted, and surgically minimal. The new query shape is verified to return the expected item_id. Static analysis test is a genuine regression guard (tests the old buggy shape explicitly and asserts it would fail). No merge sequencing risk. Approved to merge.

## Throughline observations

Wave-level pattern visible: **declarative-at-HEAD hook mechanism vs. imperative-action-required for effect** — this is the fourth instance of the pattern documented in `feedback_pattern_declarative_head_needs_imperative_action.md`. Hook 21 landed in PR #446 (merged, declared at HEAD) but required an explicit session restart to become active; the hook was declared but never fired because the session that merged it continued running against the pre-merge in-memory harness state. The fix here is correct but the pattern recurred: declarative delivery of a hook change is not sufficient without an imperative restart gate in the merge ceremony. This wave also exhibited the production-defect-in-just-merged-hook pattern — Hook 21 was merged and immediately relied upon for W11 batch ops 1 and 2 before a validation fire was confirmed. Both `deploy#284` and `data-acquisition#52` fired silently as no-ops before the defect was caught. Recommend the retro evaluate whether a "first-fire validation gate" (one manual test fire confirmed before batch resume) should be codified in the wave-kickoff or label-batch runbooks for any newly merged hook.

parametrization:
Requestor: Aino Virtanen
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: none

**Charter/QA review — PR #449**

**1. Would GraphQLVariableUsageTests have caught the #446 variableNotUsed bug at pre-merge time?**

Yes — and I verified the regex mechanics directly rather than taking Wanjiku's word.

`_declared_vars` uses `re.match(r"[^{]*\(([^)]*)\)", query, re.DOTALL)` to extract everything inside the opening signature parentheses (before the first `{`), then `re.findall(r"\$(\w+)", ...)` to collect variable names. `_assert_all_declared_vars_used` then counts occurrences of `\$<var>\b` in the full query string and asserts count >= 2 (once in signature, once or more in body).

Manually run against all 4 queries and the old buggy query:

- `_ITEM_LOOKUP_QUERY` (fixed): $org=2x, $repo=2x, $num=2x — all pass.
- `_INTROSPECT_QUERY`: $org=2x, $project=2x — pass.
- `_SET_FIELD_MUTATION`: $project=2x, $item=2x, $field=2x, $option=2x — pass.
- `_CLEAR_FIELD_MUTATION`: $project=2x, $item=2x, $field=2x — pass.
- Old buggy query (PR #446 shape): $repo=1x → FAIL; $num=1x → FAIL. Test would have gone red.

The regression guard test (`test_old_buggy_query_would_have_failed`) explicitly asserts that `$repo` and `$num` land in the `unused` list when applied to the PR #446 query shape. This confirms the test class would have caught the bug at authoring time had it existed before #446 merged.

**2. Test-mock vs real-subprocess: does the PR address the pattern or only the instance?**

The PR addresses the instance (fixing the query) and adds a static-analysis gate for the class (variableNotUsed across all 4 query/mutation strings). It correctly identifies WHY the injection-mock pattern masks this bug class: `FakeGraphQLRouter` routes by substring match on the query string, not by parsing it — a query that declares `$repo` but never uses it passes the fake router identically to a correct query. The new test layer operates at the string level, orthogonal to the mock router, so it catches the divergence the mock cannot see.

This is a meaningful pattern-level response, not only an instance fix. However, the test still does not include an integration variant (real `gh api graphql` call against a test project). The PR body correctly classifies this as deferred informational and names it explicitly in the Out of Scope section. That deferral is acceptable — the static-analysis gate catches the bug class at authoring time with zero network dependency. The follow-up issue for the integration variant is on the PR author per the PR body disposition.

Charter-promotion thesis from the orchestrator's brief (test mocks at injection points must be paired with real-call validation gates): the evidence in this PR supports it as a pattern worth surfacing, but the static-analysis approach here is arguably a stronger pre-merge gate than a network-dependent integration test. The thesis is worth filing as a charter proposal in the retro — the shape should be "static-analysis + integration variant" rather than "mock + integration only." This is a retro observation, not a PR blocker.

**3. Charter compliance**

- `## Design Rationale` heading: present. Content explains server-side traversal vs Python-side project scan and the choice of `first: 10`. Quality: adequate.
- `## Ontology Context` heading: present (required by `enforce_ontology_context.py`). Lists Hook 21, post_dispatcher.py position, Project 2, Companion Hook 13. Accurate.
- `## Acceptance Criteria` heading: present. Buckets into Actionable (this PR closes) and Informational (not in this PR) — correct two-bucket structure per `skills.md § Acceptance-Criteria-Bucketing-In-Reports`.
- `Closes #448` in PR body: present.
- Branch name `W.Mwangi/0448-hotfix-hook-21-query-variableunused`: correct per charter branching convention.
- Commit identity: PR author is Wanjiku Mwangi — charter requires per-commit `-c` flags. This is outside my diff-read scope; flag for orchestrator merge-time verification.

All charter-required headings are present and substantively populated.

**4. Out-of-scope discipline**

Hook 13 Wave-field gap on create events, `skip_no_item` Annunaki logging, integration test variant, and `projectItems` pagination beyond first:10 are all explicitly listed in the Out of Scope section. No scope inflation observed. The PR is surgically minimal: +17/-21 in hook, +112/-9 in test. The +112 is entirely new `GraphQLVariableUsageTests` class — appropriate for a test addition.

**Minor observations (non-blocking)**

The `_declared_vars` regex has one theoretical edge case: a query with a nested parenthetical in the header signature (e.g., a default value with parentheses) could cause the header extraction to terminate early. GraphQL variable declarations do not use nested parentheses syntactically (`$var: Type = value` uses no parens in the default), so this is not a real risk for this codebase's query shapes. Not a blocker; noting for completeness.

Santiago's review (already posted) verified the production query shape and manual compensation confirmations. My review is additive on charter compliance and test-mechanics correctness.

**Verdict: Approved.**

## Throughline observations

Two wave-level patterns are visible in this PR:

1. **Hotfix-after-fresh-hook**: PR #449 is a hotfix for PR #446, which was itself a new hook that shipped with a latent production-layer bug not caught by the test suite. This is the second instance this wave (the first being the Hook 21 session-load anomaly documented in the handoff). The pattern: a hook lands in a wave, immediately relied upon for batch operations, fires silently as a no-op before the defect surfaces. The static-analysis test layer added here is a direct structural response — but the question of whether a "first-fire validation gate" before batch resume should be codified in the label-batch runbook is open.

2. **Test-mock injection-point masking production-layer failures**: this is the pattern sibling to issue #175. The `FakeGraphQLRouter` injection mock routes by substring, not by parsing — so query-level declaration bugs pass the mock undetected. The static-analysis test class added here is the correct countermeasure for this specific bug class. Charter-promotion candidate for retro: "static-analysis gate required alongside injection-point mock for any externally-validated protocol (GraphQL, JSON schema, etc.)." Recommend orchestrator file this as a retro item.

Issue #176, opened 2026-04-22
refactor(hooks): extract reusable cwd-keyed consultation sentinel helper


parametrization:
## Context

PR #174 introduced a cwd-keyed sentinel pattern to work around the Hook 15 transcript-flush race in subagent worktree sessions (#169). The sentinel is specific to ontology-librarian consultation today, but the underlying pattern — \"agent invoked skill X, record it somewhere the hook can see without transcript scan\" — will recur.

Hook 15 is our first transcript-reading hook. If we add future enforcement hooks that depend on transcript evidence (e.g., \"did you run /wave-kickoff before opening a PR\", \"did you run /security-review before merging\"), each will hit the same subagent flush-race gap and reinvent the sentinel pattern.

## Scope

Extract the sentinel contract into a reusable helper:

- **API (skill side):** `write_consultation_sentinel(skill_name: str)` — writes `<cwd>/.claude/.consulted/<skill>/<sha1(cwd)[:16]>.marker` with ISO8601 timestamp.
- **API (hook side):** `consultation_sentinel_is_fresh(cwd: str, skill_name: str, ttl_seconds: int = 3600) -> bool` — recomputes the hash, checks TTL.
- Move `.claude/.librarian-consulted/` to `.claude/.consulted/ontology-librarian/` (namespaced). Update `.gitignore` to `.claude/.consulted/`.
- Update `enforce_librarian_consulted.py` to call the shared helper.
- Update `ontology-librarian` SKILL.md to call the shared shell helper (or inline the same formula, keyed by skill name).

## Acceptance

- Helper module in `.claude/hooks/_consultation_sentinel.py` (underscore prefix = internal).
- Unit tests covering: path namespacing by skill, TTL semantics, fail-open on OSError, shell/python parity via subprocess (per #175 pattern).
- Charter entry in `hooks.md` under \"Shared helpers\" documenting the pattern.
- Hook 15 refactored to consume the helper (no behavior change — regression tests stay green).

## Related

- Blocks: future transcript-reading enforcement hooks.
- Origin: PR #174 review by Nadia Khoury (strategic/PD review of the sentinel pattern's generalizability).
- Depends on: PR #174 merge.

parametrization:
Closed by PR #397 (squashed to deployments/phase-3/wave-9 at 75ee4f1). _consultation_sentinel helper extracted; path namespaced .claude/.consulted/<skill>/; charter Shared Helpers section codifies the W9 primitive convergence.

PR #457, opened 2026-05-17
tech-debt(hook-21+hook-13): multi-cmd Bash + create-time wave-field sync + skip logging (#455 #450 #451)

parametrization:
Closes #455 (Hook 21 multi-cmd silent-skip), #450 (Hook 13 create-time wave field gap), #451 (Hook 21 skip-path observability).

Bundled into one PR because all three touch the same surfaces (`_wave_label_parse.py`, `post_label_change_wave_field_sync.py`, `auto_add_issue_to_board.py`). Splitting would force order-dependent rebase chain since #450 delegates to Hook 21 helpers introduced/refactored for #451.

## Hook 21 multi-cmd parser (#455)

`parse_wave_label_change` extended to `parse_wave_label_changes` (plural, returns list) so a single Bash tool call containing multiple `gh issue edit` invocations via semicolon, double-ampersand, or pipeline operators dispatches one Wave-field mutation per change instead of silently no-op'ing the whole batch. Singular `parse_wave_label_change` retained as back-compat wrapper for `post_wave_kickoff_comment.py` (kickoff is single-cmd only).

`check()` returns the legacy flat dict for single-cmd shapes and a multi-result wrapper for multi-cmd shapes. Single-cmd callers/tests unaffected.

When the command contains a wave-label edit flag value but the parser extracts nothing (unhandled shape, missing repo flag, etc.), the hook emits a parser-skip annunaki event so silent multi-cmd misses are visible to /annunaki-attack. Bounded regex anchor prevents false-positives on suffixed labels like p3-wave-10-special.

## Hook 13 create-time Wave field sync (#450)

`auto_add_issue_to_board` extended to set project 2's Wave field when a wave-labeled issue is created via gh. Delegates all GraphQL/auth/cache work to the Hook 21 helper module — one auth-scope pre-flight, one shared project-ID cache, one option-name conversion path.

Item-lookup retries 3x to handle the eventual-consistency window between item-add returning and projectItems seeing the new item.

## Hook 21 skip-path observability (#451)

All four skip paths now emit log_posttooluse_event annunaki entries — closes the misdiagnosis vector that contributed to #448 (variableNotUsed bug went undetected because skip_no_item was silent).

## Test plan

- [x] ENVIRONMENT=test pytest hooks+lib — 816 passing (was ~801)
- [x] ruff check + format — clean
- [x] 8 new MultiCmdBashTests cover semicolon, ampersand, mixed set/clear, single-cmd preservation, parser-skip triggering + non-triggering
[...]

PR #454, opened 2026-05-17
hotfix(hooks): annunaki_monitor + auto_add read tool_response (partial fix #453)

parametrization:
## Summary

- `annunaki_monitor.py` line 104 and `auto_add_issue_to_board.py` line 49 read `input_data.get("tool_output", {})` — but Claude Code's PostToolUse hook contract passes the field as `tool_response`. Both hooks silently bailed on every real Bash call (`stdout = ""` → `combined_output = ""` → `return None`).
- This is why the annunaki log accumulated 388 entries, ~76% of which are hook-test-suite fixtures (test code uses `tool_output`) and ~0 real production failures despite a wave of activity (see #452 for the noise side of the same finding).
- Fix: read `tool_response` first, fall back to `tool_output` so the legacy test fixtures still pass.
- Adds 2 regression tests using the production `tool_response` shape (one each for `annunaki_monitor` and `auto_add_issue_to_board`) so this class can't recur silently.

## Why partial fix

#453 has two findings:
1. **Primary** — Claude Code's PostToolUse Bash hook isn't firing the dispatcher at all in the current session (Hook 21 silent-skip resurfaced at the harness layer). Root cause still unknown.
2. **Secondary** — this `tool_output`/`tool_response` field-name bug.

This PR fixes #2. After it lands AND a session restart, if annunaki_monitor still doesn't log a `false` failure, #1 is confirmed as a harness-vs-hook issue (Claude Code 2.1.143, `--dangerously-skip-permissions`, `CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS=1`, or some combination). If annunaki DOES start logging in the fresh session, the field-name bug was load-bearing for #1 too.

## Test plan

- [x] `python3 -m unittest discover .claude/hooks/tests -v` → 778 tests pass (including 2 new)
- [x] `uvx ruff@0.6.9 format --check .claude/hooks/` → 64 files already formatted
- [x] `uvx ruff@0.6.9 check .claude/hooks/` → All checks passed
- [ ] CI: `hooks-lint`, `hooks-tests`, type-check (mypy) — relied on for full verification
- [ ] Post-merge: session restart and `false`/`ls /nonexistent` to verify annunaki logs the failure

## References

- Closes part of #453
- Sibling to #452 (annunaki test-fixture noise)
- Builds on PR #449 (Hook 21 GraphQL hotfix) — same recurring-pattern class as [[feedback_test_mock_injection_masks_production_failure]] (mocks bypass the production contract)

[...]

PR #458, opened 2026-05-17
tech-debt(hook-15): subprocess-based shell/Python hash parity test (#175)

### LINKED ISSUE #175: hook15 sentinel: add subprocess-based shell/Python hash parity test

parametrization:
Closes #175.

The pre-existing `test_cwd_hash_matches_shell_pwd_sha1sum` spot-checks the Python hash function against an in-test Python reimplementation — that's Python-vs-Python parity. It does NOT catch divergence between the librarian skill's actual shell pipeline (in SKILL.md step 0) and the Python `_cwd_sentinel_hash` implementation. If the skill's bash snippet drifts (switches to printf-pwd, uses realpath, different cut width, etc.), the spot-check still passes — and the sentinel-fallback for #169 silently regresses because the hook computes one path and the skill writes a different one.

This PR adds 4 new tests in a `ShellPythonHashParityTests` class:

1. `test_shell_pipeline_matches_python_hash_in_isolated_cwd` — subprocess.run the literal pipeline in a temp cwd; assert shell output matches Python hash.
2. `test_shell_pipeline_matches_python_hash_for_long_path` — same shape with a deeply nested path (length-dependent truncation guard).
3. `test_skill_md_uses_expected_shell_pipeline` — parse SKILL.md and assert it still contains the canonical pipeline string. If the skill drifts, this test fails before the parity test even runs, so reviewers are pointed at the source of the drift (the skill, not the hook).
4. `test_skill_md_uses_consulted_directory_structure` — drift guard for the namespaced sentinel directory (#176 path move).

## Test plan

- [x] `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/test_enforce_librarian_consulted.py` — 33 passing (was 29)
- [x] `uvx ruff check` + `uvx ruff format --check` clean
- [x] All pre-commit hooks pass

No production code changes — test-only PR.

## Tech Debt

Generated with Claude Code

### DIFF
--- a/.claude/hooks/tests/test_enforce_librarian_consulted.py
+++ b/.claude/hooks/tests/test_enforce_librarian_consulted.py
@@ -529,5 +529,126 @@ class SentinelFallbackTests(unittest.TestCase):
         self.assertEqual(hook._cwd_sentinel_hash(sample), expected)
 
 
+class ShellPythonHashParityTests(unittest.TestCase):
+    """Issue #175 — subprocess-based shell/Python hash parity.
+
+    The existing `test_cwd_hash_matches_shell_pwd_sha1sum` spot-check above
+    re-implements the same algorithm in test code and asserts it equals the
+    hook's Python function. That's a Python-vs-Python parity check; it does
+    NOT catch divergence between the librarian skill's actual shell pipeline
+    (`pwd | sha1sum | cut -c1-16`) and the Python implementation.
+
+    These tests close that gap by:
+      1. Running the exact shell pipeline from SKILL.md in an isolated cwd
+         and asserting the resulting filename matches the Python hash.
+      2. Parsing the shell snippet directly out of SKILL.md so the test
+         fails if the skill drifts (e.g., switches to `realpath`, uses a
+         different cut width, etc.).
+    """
+
+    # Resolve SKILL.md once at class-load — fail fast at collection time if
+    # the layout has shifted, before any test runs.
+    _SKILL_MD = (
+        Path(__file__).resolve().parent.parent.parent / "skills" / "ontology-librarian" / "SKILL.md"
+    )
+
+    def _read_skill_md(self) -> str:
+        self.assertTrue(
+            self._SKILL_MD.is_file(),
+            f"Expected SKILL.md at {self._SKILL_MD} — test cannot validate parity "
+            "without the canonical shell pipeline source",
+        )
+        return self._SKILL_MD.read_text(encoding="utf-8")
+
+    def test_shell_pipeline_matches_python_hash_in_isolated_cwd(self) -> None:
+        """Run `pwd | sha1sum | cut -c1-16` in an isolated cwd. Assert the
+        resulting 16-hex prefix matches `_cwd_sentinel_hash(cwd)`.
+
+        Uses `subprocess.run(..., cwd=tmp_dir)` so the shell `pwd` resolves
+        to the test's tmp cwd, not the test runner's cwd. The pipeline is
+        passed verbatim from SKILL.md step 0 (the literal bash snippet that
+        the librarian skill writes the sentinel filename from).
+        """
+        import subprocess
+        import tempfile
+
+        with tempfile.TemporaryDirectory(prefix="hash_parity_") as tmp:
+            # Realpath to defeat any /tmp -> /private/tmp on macOS or symlink
+            # surprises; Python's `os.path.abspath` does the same resolution
+            # the hook applies before hashing.
+            tmp_abs = os.path.abspath(tmp)
+            result = subprocess.run(
+                ["bash", "-c", "pwd | sha1sum | cut -c1-16"],
+                cwd=tmp_abs,
+                capture_output=True,
+                text=True,
+                check=True,
+                timeout=10,
+            )
+            shell_hash = result.stdout.strip()
+            python_hash = hook._cwd_sentinel_hash(tmp_abs)
+            self.assertEqual(
+                shell_hash,
+                python_hash,
+                f"Shell pipeline produced {shell_hash!r} but Python produced "
+                f"{python_hash!r} for cwd {tmp_abs!r}. If the skill writes a "
+                f"sentinel under {shell_hash}.marker the hook will look up "
+                f"{python_hash}.marker and never find it.",
+            )
+
+    def test_shell_pipeline_matches_python_hash_for_long_path(self) -> None:
+        """A long path with multiple segments — guard against any
+        length-dependent truncation difference between shell and Python."""
+        import subprocess
+        import tempfile
+
+        with tempfile.TemporaryDirectory(prefix="hash_parity_long_") as tmp:
+            nested = os.path.join(tmp, "deep", "nested", "worktree", "subagent", "cwd")
+            os.makedirs(nested, exist_ok=True)
+            nested_abs = os.path.abspath(nested)
+            result = subprocess.run(
+                ["bash", "-c", "pwd | sha1sum | cut -c1-16"],
+                cwd=nested_abs,
+                capture_output=True,
+                text=True,
+                check=True,
+                timeout=10,
+            )
+            shell_hash = result.stdout.strip()
+            python_hash = hook._cwd_sentinel_hash(nested_abs)
+            self.assertEqual(shell_hash, python_hash)
+
+    def test_skill_md_uses_expected_shell_pipeline(self) -> None:
+        """Drift guard: the bash snippet in SKILL.md step 0 must contain
+        the exact `pwd | sha1sum | cut -c1-16` pipeline.
+
+        If a future edit to the skill changes the algorithm (e.g., switches
+        to `realpath | sha256sum | cut -c1-16`, or drops the cut width
+        to 8 chars), this test goes red so the change is noticed and the
+        Python side updated in lock-step.
+        """
+        skill_md = self._read_skill_md()
+        self.assertIn(
+            "pwd | sha1sum | cut -c1-16",
+            skill_md,
+            "SKILL.md step 0 must use the canonical `pwd | sha1sum | cut -c1-16` "
+            "pipeline. If the algorithm has changed, update _cwd_sentinel_hash "
+            "in _consultation_sentinel.py to match.",
+        )
+
+    def test_skill_md_uses_consulted_directory_structure(self) -> None:
+        """Drift guard: SKILL.md must write sentinels under the namespaced
+        `.claude/.consulted/ontology-librarian/` directory that the hook
+        reads from (per #176)."""
+        skill_md = self._read_skill_md()
+        self.assertIn(
+            ".claude/.consulted/ontology-librarian",
+            skill_md,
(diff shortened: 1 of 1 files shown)


parametrization:
Requestor: Santiago Ferreira
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: None

Secondary review (Release Coordinator). Read at head_sha `c63bbf7d007b17eb20c92262cb351a30ede8f4c0` via `gh api contents`.

## What this PR fixes

The pre-existing `test_cwd_hash_matches_shell_pwd_sha1sum` (line 518) re-implemented the sha1 hash in test-Python and compared to `_cwd_sentinel_hash` — **Python-vs-Python parity, not shell-vs-Python parity**. If the librarian SKILL.md drifts to `printf "%s" $(pwd) | sha1sum` or `realpath | sha256sum | cut -c1-8`, the spot-check still passes; the sentinel-fallback for #169 silently regresses because hook and skill compute different filenames.

## Tests

`ShellPythonHashParityTests` (4 new tests) closes that gap:

1. **`test_shell_pipeline_matches_python_hash_in_isolated_cwd`** (`:563`) — `subprocess.run(["bash", "-c", "pwd | sha1sum | cut -c1-16"], cwd=tmp_abs)` is the real-shell test. Uses `os.path.abspath(tmp)` to defeat `/tmp`→`/private/tmp` symlink surprises on macOS.
2. **`test_shell_pipeline_matches_python_hash_for_long_path`** (`:599`) — nested 5-deep path catches any length-dependent shell-vs-Python divergence.
3. **`test_skill_md_uses_expected_shell_pipeline`** (`:621`) — `assertIn("pwd | sha1sum | cut -c1-16", skill_md)` drift guard.
4. **`test_skill_md_uses_consulted_directory_structure`** (`:639`) — `assertIn(".claude/.consulted/ontology-librarian", skill_md)` for the #176 namespaced sentinel path.

## Parser-robustness check (brief's release-coordination concern)

The brief asked about parser robustness against future SKILL.md edits. **The implementation uses `assertIn` substring matching, not regex** — this is the right shape. It's robust against:
- Markdown heading changes around the bash snippet
- Code-fence variant edits (` ```bash` vs ` ```sh`)
- Surrounding prose rewrites
- Whitespace-only edits

It correctly FAILS on:
- Algorithm changes (`sha1sum` → `sha256sum`, `cut -c1-16` → `cut -c1-8`, `pwd` → `realpath`)
- Sentinel directory moves (caught by test #4)

A regex-based parser would have been brittle to whitespace/quoting variants in the surrounding markdown; `assertIn` is harder to break by accident and easier for a future drift-detector to interpret.

## Release-coordination angle

- **CI**: PR rollup all green (Ruff lint / Ruff format / Mypy / Smoke-test).
- **Local verification at head_sha c63bbf7**: 33 tests pass in this file (was 29 — +4 as advertised); broader hook+lib suite still at 800-passing baseline.
- **Cross-repo impact**: test-only change in main repo's `.claude/hooks/tests/`; zero runtime impact on any child repo or in-flight PR. Strictly additive coverage.
- **Future-drift gate**: if a future PR edits `.claude/skills/ontology-librarian/SKILL.md`'s bash snippet without updating `_cwd_sentinel_hash`, test #1 catches the math divergence AND test #3 fires a "SKILL.md drifted" diagnostic pointing the contributor at the canonical source. Strong belt-and-suspenders.

## Tech-debt attestation

## Throughline-watch

Continuation of P3W11's **test-mock-bypasses-static-validation** throughline (same one I flagged in #457): this PR fixes the EXACT pattern at the original instance (#175). Worth weaving both #457's GraphQL static-analysis gate and #458's real-subprocess parity test into a single retro proposal: "Mock-injection unit tests need a paired non-mock gate (static-analysis OR real-call) when production behavior depends on cross-language or cross-process correctness." [[feedback_test_mock_injection_masks_production_failure]] charter-promotion is timely.

## Verdict

Approved. Test correctly closes the #169-class regression vector that the pre-existing Python-vs-Python spot-check could not catch.

parametrization:
Requestor: Aino Virtanen
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: none

## Verdict: Approved

Reviewed at PR head_sha `c63bbf7` via `gh api repos/.../contents?ref=<sha>` (origin-over-local per charter `pull-requests.md` § Origin > Local Clone). Single-file, +121/-0, test-only.

### Verification

- Read the full PR diff at HEAD (single file, 122-line addition above the existing `test_cwd_hash_matches_shell_pwd_sha1sum` Python-vs-Python parity spot-check).
- Read SKILL.md (`.claude/skills/ontology-librarian/SKILL.md`) at HEAD: step 0 contains the literal `pwd | sha1sum | cut -c1-16` pipeline at line 30 and the `.claude/.consulted/ontology-librarian/<HASH>.marker` write path at line 31. Both drift-guard substring assertions correctly pin reality.
- Read the hook's `_cwd_sentinel_hash` shim at `enforce_librarian_consulted.py:253` and the canonical `cwd_sentinel_hash` at `_consultation_sentinel.py:99`: implementation is `hashlib.sha1((abspath + "\n").encode("utf-8")).hexdigest()[:16]`. The trailing newline is the load-bearing detail — `pwd` emits one before piping to `sha1sum`, so the Python side must replay it for byte-equality. The comment block on lines 100-104 of the helper documents this explicitly.
- Ran the 4 new `ShellPythonHashParityTests` locally: 4 passed in 0.05s. Real `subprocess.run(["bash", "-c", "pwd | sha1sum | cut -c1-16"], cwd=...)` invocation, not a Python-reimplementation — the test exercises the exact shell pipeline the skill uses, which is the [[feedback_test_mock_injection_masks_production_failure]] guard pattern (don't pair production with a Python reimpl of the same algorithm; pair with the actual binary).
- Ran the full file: **33 passed** (matches PR body's "was 29 → 33" claim — confirmed delta is 4 new tests as advertised).
- Ran the full hooks suite: **793 passed** (no regressions; baseline was the post-#454 state).
- CI: 4/4 green at HEAD (mypy / ruff lint / ruff format / smoke).
- Santiago has already posted Approved as secondary; this verdict is the convergent reviewer's complete-the-merge gate.

### Why the design is right

**Shell-vs-Python parity is the actual #169 regression class.** The pre-existing `test_cwd_hash_matches_shell_pwd_sha1sum` (line 521-ish, name aside) re-implements the algorithm in test code and asserts it equals the hook function — that's a Python-vs-Python check. A skill drift that swaps `sha1sum` for `sha256sum`, or `cut -c1-16` for `cut -c1-12`, or `pwd` for `pwd -P`, or adds `-z` to drop the trailing newline — none of those would turn that test red. The new `test_shell_pipeline_matches_python_hash_in_isolated_cwd` actually shells out to the literal pipeline; that's the test the original #169 fix was missing.

**SKILL.md drift guards are correctly scoped.** `test_skill_md_uses_expected_shell_pipeline` does a substring assertion for `"pwd | sha1sum | cut -c1-16"`. `test_skill_md_uses_consulted_directory_structure` does the same for `".claude/.consulted/ontology-librarian"`. A regex-aware bash parser would be over-engineering for a single-line snippet; substring is the right tool for the failure mode (algorithm rename, dir rename) and reads cleanly. This is the [[feedback_canonical_source_via_git_show]] principle applied at test-gate level: the canonical source is SKILL.md, the hook is the consumer, the test pins the contract.

**Long-path test covers length-dependent surprises.** `test_shell_pipeline_matches_python_hash_for_long_path` uses `tmp/deep/nested/worktree/subagent/cwd` — defends against any hash-impl change that accidentally splits at a fixed length (sha1 doesn't, but a future drift to a chunked impl would).

**Class-load assertion (`_SKILL_MD` resolved at module-load via `Path(__file__).resolve().parent.parent.parent`) fails fast at test collection if the skill directory moves.** That's the right failure surface — collection errors are louder than a single test failing in a subsequent suite run.

**`import subprocess` and `import tempfile` are function-local.** Lines 45-46 and 75-76 import inside the test methods rather than at module top. Stylistically that's a wash — function-local imports are sometimes used to avoid polluting the module namespace or to make the dependency boundary obvious — but ruff often flags this as `E402`-adjacent. Both imports already exist at module-top in the file (per the existing test class above this one), so moving them up would be a one-line lint-cleanup follow-up. Not worth a separate issue; trivial to clean up next time the file is touched.

**Drift-guard substrings don't catch flag additions to a still-matching binary.** E.g., a skill edit to `pwd | sha1sum --tag | cut -c1-16` would still contain the substring `pwd | sha1sum` (and the `cut` substring), but `sha1sum --tag` produces a different output format and would break parity. The shell-execution test would catch it (since it actually runs the pipeline), so the gap is closed in aggregate — the substring drift-guard is just one defensive layer of two. Worth noting for the future but not a follow-up issue today; the shell-pipeline test IS the load-bearing check.

### Throughline observation (W11 retro candidate, additive)

This PR completes a pattern I'm seeing across W11 PRs from Wanjiku — every test-coverage closure she's shipped this batch (#457's `GraphQLVariableUsageTests`, #458's `ShellPythonHashParityTests`) **pairs production code with a real-binary check, not a Python reimpl**. Both close the [[feedback_test_mock_injection_masks_production_failure]] gap from a different angle. Worth surfacing at retro as a process-success pattern: "test the contract against the actual production binary, not a synthetic mirror."

### Verdict basis

Approve on:
- 4 new tests all pass locally
- Real shell-pipeline invocation (`subprocess.run(["bash", "-c", ...])`) — the correct guard for #169 regression class
- SKILL.md drift guards pin the canonical source by substring (right level of strictness for the failure mode)
- Hash-impl byte-equality verified by reading both sides (helper + skill)
- 33-file / 793-suite test count matches PR body
- CI 4/4 green
- Santiago's secondary Approved already posted; this completes the 2-reviewer gate

Issue #332, opened 2026-05-08
tech-debt(/wave-scope): upsert_status_keys.py breaks on multi-line wave_<M>_* sibling blocks


parametrization:
## Summary

`/wave-scope` Step 13's `upsert_status_keys.py` helper aborts ("text-level upsert diverged from logical upsert") when a `cross-repo-status.json` already contains multi-line `wave_<M>_*` sibling blocks. The single-line regex (`^(  )"<key>": .*$`) mis-locates the insertion point inside an open multi-line array/object, producing invalid JSON.

## Reproducer

P3W8 `/wave-scope` 2026-05-08 hit this when the file contained:
- `wave_8_work` — multi-line object (P2W8 historical)
- `wave_8_carry_forward` — multi-line array (W7-era pre-phase-reset intent)

Helper invocation tried to upsert `wave_8_scope` etc. The sibling-finder regex matched `  "wave_8_carry_forward": [` as the "last sibling line" and inserted the new compact key right after that opening line — placing it INSIDE the still-open list. Result: JSON parse failure on read-back validation.

Worked around by manually stripping `wave_8_work` + `wave_8_carry_forward` first, then upserting the new shape. Reconciliation note in commit `48b44ee` documents the workaround.

## Acceptance criteria

- [ ] Helper handles multi-line `wave_<M>_*` sibling values correctly — find the matching close bracket/brace, not just the opening line
- [ ] Test fixture covering: single-line sibling, multi-line array sibling, multi-line object sibling, nested objects in sibling
- [ ] Re-running the helper on a file that already has multi-line wave_<M>_* values doesn't break it
- [ ] If insertion would produce invalid JSON, fail-fast with a clearer error (e.g., "sibling block at line N is multi-line; helper requires single-line siblings or explicit removal first")
- [ ] Document the limitation in the helper's docstring if not fixed

## Wave assignment

W8 (P3W8 foundation reset Tier 1) — owner directive 2026-05-08. This bug surfaced *during* the first `/wave-scope` run under the new gates; fixing it in W8 keeps the foundation-reset theme on track (process tooling reliability is exactly what this wave addresses).

## Cross-references

- Helper: `.claude/skills/wave-scope/upsert_status_keys.py`
- W8 scope reconciliation commit: `48b44ee`
- W8 meta-issue: noorinalabs-main#331

parametrization:
Requestor: Wanjiku.Mwangi
Requestee: Aino.Virtanen
[...]

PR #460, opened 2026-05-17
tech-debt(upsert_status_keys): structural top-level filter for sibling-finder (#456)

### LINKED ISSUE #456: tech-debt(upsert_status_keys.py): regression — multi-line dict insertion fails adjacent to multi-line array sibling (#332/#348 follow-up)
## Symptom

`upsert_status_keys.py` aborts with the same `text-level upsert produced invalid JSON: Expecting ',' delimiter` error that #332 and #348 were filed to address. Both prior issues are CLOSED but the bug recurs in a new shape.

Reproduced 2026-05-17T03:25Z during `/wave-scope 3 11` Step 13:

```
ERROR: text-level upsert produced invalid JSON: Expecting ',' delimiter: line 2563 column 5 (char 108626)
  This usually means the helper's sibling-finder mis-located
  the insertion point inside a multi-line array/object value
  (main#332 reproducer). File the failing input as an issue
  with the offending key=value upsert command.
```

The helper's own error message acknowledges main#332 as the reproducer class.

Second occurrence: 2026-05-17T01:46:39Z in `.claude/annunaki/errors.jsonl` from last session's W11 scope-apply, same error string.

## Suspected root cause

The fix from #332 (`_find_value_end` for multi-line sibling values) does not cover the case where the *new* key being inserted is itself a multi-line object, AND its siblings include a mix of single-line strings (`wave_11_scope_reconciled_at`) and multi-line arrays (`wave_11_repos_in_scope`).

The 5-key upsert that triggered this:

```
wave_11_scope_reconciled_at='"2026-05-17T03:25:..."'         # single-line string
wave_11_repos_in_scope='[...]'                                # multi-line array
wave_11_meta_issue=447                                        # single-line int
wave_11_scope='{"theme":"...","scope_size_at_reconcile":117,...}'  # multi-line dict, ~10 nested keys
wave_11_scope_reconciliation_note='"..."'                     # single-line string
```

The placement logic for the multi-line `wave_11_scope` dict appears to land it inside the still-open `wave_11_repos_in_scope` array (the value-end calculation for the newly-inserted array sibling is wrong, so the next key inserts mid-array).

## Reproduction

```bash
# Pre-state: cross-repo-status.json contains compact single-line wave_11_* keys at top
# (no wave_11_scope present yet)
python3 .claude/lib/upsert_status_keys.py cross-repo-status.json \
  'wave_11_scope_reconciled_at="2026-05-17T00:00:00Z"' \
  'wave_11_repos_in_scope=["noorinalabs-main","noorinalabs-isnad-graph"]' \
  'wave_11_meta_issue=447' \
  'wave_11_scope={"theme":"x","scope_size_at_reconcile":117,"per_repo_open":{"a":1,"b":2}}' \
  'wave_11_scope_reconciliation_note="x"'
# → ERROR rollback
```

## Workaround used this session

Direct Python `json.dump(data, f, indent=2)` round-trip. Acceptable when the file is already pretty-printed (the cosmetic-diff cost #278 was guarding against has already been paid). Not acceptable for files that are still in compact-inline format.

- `gh issue list --repo noorinalabs/noorinalabs-main --state all --search "upsert_status_keys"` →
  - #332 CLOSED 2026-05-11 — original multi-line sibling bug (fix landed via `_find_value_end`)
  - #348 CLOSED — nested-object case + P2 namespace pollution (worked around by renaming)
- Root-cause search `_find_value_end` → no open matches
- Pass B rationalization scan → no open rationalization workstream covering this; treating as regression-of-#332/#348, not drift-on-existing

## Acceptance criteria

- [ ] Reproduction case (multi-line dict insertion adjacent to a multi-line array sibling, with single-line siblings interleaved) added to the helper's test fixture
- [ ] Helper either correctly inserts multi-line values OR fails fast at preflight with a clearer "cannot insert multi-line value adjacent to multi-line sibling" diagnostic + suggested workaround (rename / pre-strip / Python json.dump fallback)
- [ ] The `_find_value_end` fix from #332 is extended (or the placement logic refactored) so the wave_11_scope shape works end-to-end
- [ ] Regression-test the 5-key call shape from this session's reproduction

## Layer

Single layer: detection-time helper (`.claude/lib/upsert_status_keys.py`). No strategy or docs sibling — the helper's docstring already documents the limitation; the bug is in the implementation.

## Cross-references

- #332 CLOSED — original (multi-line sibling values)
- #348 CLOSED — nested-object + namespace pollution
- This-session evidence: `.claude/annunaki/errors.jsonl` 2026-05-17T01:46:39Z + this-pass attempt 2026-05-17T03:25Z
- Workaround precedent: `feedback_pattern_declarative_head_needs_imperative_action.md` (file written → state asserted) — the json.dump fallback is the imperative-action equivalent

## Out of scope

- Refactoring `cross-repo-status.json` shape to be more upsert-friendly (separate issue if pursued)
- Replacing the helper with a different tool (out-of-scope; the helper is well-suited if the multi-line case is fixed)

parametrization:
Closes #456.

## Root cause

The sibling-finder regex matched any line shaped like `^(  )"wave_N_..."` at exactly 2-space indent. When a top-level multi-line value (e.g., `wave_11_scope`) contains nested keys at the same 2-space indent (which happens when the value was hand-formatted as pretty JSON, or produced by `json.dumps(value, indent=2)` where the outer object lives at 2-space depth), the regex matched the nested key too. The helper then called `_find_value_end` on the wrong position and inserted the new key IN THE MIDDLE of the wave_11_scope dict.

## Fix

Added `_is_top_level_position(text, position) -> bool` — a structural walker that tracks bracket depth + JSON string state from the start of the text. Returns True only when `position` is at depth 1 (directly inside the outermost object) and not inside a string.

Both the in-place-replace match and the sibling-finder match are now filtered through this check, so regex matches nested inside multi-line values are rejected.

Performance: O(position) per check on a ~108KB cross-repo-status.json is well under 10ms. Called once per candidate (typically 1–5 per upsert), so per-upsert cost stays negligible.

## Error message improvements (#456 acceptance criteria)

When the post-write JSON validation fails OR the parsed/text round-trip diverges, the error message now surfaces:
- The offending key=value pairs from this call (for paste-into-reproducer)
- The list of diverged keys (for the divergence case)

## Test plan

- [x] `ENVIRONMENT=test python3 -m pytest .claude/lib/tests/test_upsert_status_keys.py` — 18 passing (was 9, +4 false-match regression + +5 depth-walker pure-function)
- [x] `uvx ruff check + format --check` clean
- [x] Production smoke: applied to a copy of cross-repo-status.json with 4 representative upserts — passes
- [x] Pre-commit hooks all green

## Tech Debt

Generated with Claude Code

### DIFF
--- a/.claude/lib/tests/test_upsert_status_keys.py
+++ b/.claude/lib/tests/test_upsert_status_keys.py
@@ -205,5 +205,144 @@ class UpsertMultiLineFinalKeyTests(unittest.TestCase):
         self.assertEqual(parsed["wave_8_carry_forward"], ["#341"])
 
 
+class UpsertNestedSiblingFalseMatchTests(unittest.TestCase):
+    """main#456 regression — the sibling-finder must NOT match keys whose
+    name happens to look like a wave_N_* sibling but is actually nested
+    inside a multi-line value at the same indent level.
+
+    Reproducer: pretty-printed JSON where the outer wave_N_scope dict has
+    inner keys at 2-space indent (which happens when the user wrote a
+    multi-line value with `json.dumps(value, indent=2)` or hand-formatted
+    pretty JSON without re-indenting). The pre-#456 regex would match
+    those nested keys and treat them as wave_N_* siblings, then call
+    _find_value_end on the wrong position and insert IN THE MIDDLE of the
+    wave_N_scope dict.
+
+    These tests pin the structural-top-level filter so the insertion
+    always lands at the genuine top level.
+    """
+
+    def test_nested_wave_key_at_2_space_indent_does_not_confuse_sibling_finder(self):
+        """The reproducer that motivated #456. Nested `wave_11_inner_*` keys
+        at the same 2-space indent as the outer `wave_11_scope` opener must
+        NOT count as wave_11_* siblings — they're inside wave_11_scope's
+        value, not at the top level."""
+        text = (
+            "{\n"
+            '  "phase": "phase-3",\n'
+            '  "wave_11_scope": {\n'
+            '  "wave_11_inner_a": 1,\n'
+            '  "wave_11_inner_b": 2\n'
+            "  }\n"
+            "}\n"
+        )
+        out = upsert_top_level_key(text, "wave_11_meta_issue", "447")
+        parsed = json.loads(out)
+        # The new key must land at TOP LEVEL, not inside wave_11_scope.
+        self.assertEqual(parsed.get("wave_11_meta_issue"), 447)
+        # wave_11_scope's nested keys must remain unchanged.
+        self.assertEqual(
+            parsed.get("wave_11_scope"),
+            {"wave_11_inner_a": 1, "wave_11_inner_b": 2},
+        )
+
+    def test_nested_wave_key_replace_does_not_target_nested_match(self):
+        """When upserting `wave_11_inner_a` at top level (with no top-level
+        key by that name), the regex MUST NOT match the nested `wave_11_inner_a`
+        inside `wave_11_scope`. The insertion should add a NEW top-level key."""
+        text = '{\n  "phase": "phase-3",\n  "wave_11_scope": {\n  "wave_11_inner_a": 1\n  }\n}\n'
+        out = upsert_top_level_key(text, "wave_11_inner_a", "999")
+        parsed = json.loads(out)
+        # Top-level wave_11_inner_a should be 999.
+        self.assertEqual(parsed["wave_11_inner_a"], 999)
+        # Nested wave_11_inner_a inside wave_11_scope should STILL be 1.
+        self.assertEqual(parsed["wave_11_scope"]["wave_11_inner_a"], 1)
+
+    def test_nested_wave_key_at_4_space_indent_still_ignored(self):
+        """Already worked pre-#456 (indent mismatch) but pinned as a guard."""
+        text = (
+            "{\n"
+            '  "phase": "phase-3",\n'
+            '  "wave_11_scope": {\n'
+            '    "wave_11_inner_a": 1,\n'
+            '    "wave_11_inner_b": 2\n'
+            "  }\n"
+            "}\n"
+        )
+        out = upsert_top_level_key(text, "wave_11_meta_issue", "447")
+        parsed = json.loads(out)
+        self.assertEqual(parsed["wave_11_meta_issue"], 447)
+        self.assertEqual(
+            parsed["wave_11_scope"],
+            {"wave_11_inner_a": 1, "wave_11_inner_b": 2},
+        )
+
+    def test_sequenced_upsert_after_inserting_multiline_dict(self):
+        """The 5-key call shape from the #456 issue body. Sequence:
+        1. start with compact-inline wave_11_* keys
+        2. upsert a multi-line dict for wave_11_scope
+        3. upsert another wave_11_* key
+
+        The multi-line dict insertion must not leave the text in a state
+        where the next upsert's sibling-finder mis-matches a nested key.
+        """
+        text = '{\n  "phase": "phase-3",\n  "wave_11_active": true\n}\n'
+        # Step 1: insert a multi-line dict value (as the helper produces
+        # when the json_value arg is pretty-printed).
+        ml_dict = '{\n    "wave_11_inner_a": 1,\n    "wave_11_inner_b": 2\n  }'
+        text2 = upsert_top_level_key(text, "wave_11_scope", ml_dict)
+        json.loads(text2)  # must parse
+        # Step 2: insert another wave_11_* key
+        text3 = upsert_top_level_key(text2, "wave_11_meta_issue", "447")
+        parsed = json.loads(text3)
+        self.assertEqual(parsed["wave_11_meta_issue"], 447)
+        self.assertEqual(
+            parsed["wave_11_scope"],
+            {"wave_11_inner_a": 1, "wave_11_inner_b": 2},
+        )
+
+
+class IsTopLevelPositionTests(unittest.TestCase):
+    """Pure-function coverage for `_is_top_level_position` — the depth
+    walker that powers the #456 fix."""
+
+    def test_inside_outer_object_is_top_level(self):
+        from upsert_status_keys import _is_top_level_position
+
+        text = '{\n  "a": 1\n}\n'
+        # Position of `"a"` is at depth 1 (inside the outer `{`).
+        pos = text.index('"a"')
+        self.assertTrue(_is_top_level_position(text, pos))
+
+    def test_inside_nested_object_is_not_top_level(self):
+        from upsert_status_keys import _is_top_level_position
+
+        text = '{\n  "outer": {\n    "inner": 1\n  }\n}\n'
+        pos = text.index('"inner"')
+        self.assertFalse(_is_top_level_position(text, pos))
+
+    def test_inside_nested_array_is_not_top_level(self):
+        from upsert_status_keys import _is_top_level_position
+
+        text = '{\n  "list": [\n    "item"\n  ]\n}\n'
+        pos = text.index('"item"')
+        self.assertFalse(_is_top_level_position(text, pos))
+
+    def test_inside_string_value_is_not_top_level(self):
+        from upsert_status_keys import _is_top_level_position
+
+        text = '{\n  "note": "contains { brace } chars"\n}\n'
+        # Position INSIDE the string value
+        pos = text.index("brace")
+        self.assertFalse(_is_top_level_position(text, pos))
+
+    def test_brace_chars_inside_strings_dont_change_depth(self):
+        from upsert_status_keys import _is_top_level_position
+
+        text = '{\n  "a": "before {b} brace",\n  "real_top": 1\n}\n'
+        pos = text.index('"real_top"')
+        self.assertTrue(_is_top_level_position(text, pos))
+
+
 if __name__ == "__main__":
     unittest.main()
--- a/.claude/lib/upsert_status_keys.py
+++ b/.claude/lib/upsert_status_keys.py
@@ -32,6 +32,50 @@ import sys
 from pathlib import Path
 
 
+def _is_top_level_position(text: str, position: int) -> bool:
+    """Return True iff `position` in `text` is at JSON top-level depth.
+
+    Walks the text from the start tracking bracket depth and JSON string
+    state. Depth-1 (i.e., directly inside the outermost `{...}`) is what
+    "top-level sibling" means for cross-repo-status.json. Positions inside
+    nested objects/arrays or inside string values return False.
+
+    This is the structural check that catches the bug class where the
+    regex sibling-finder happens to match a key NAME that's nested inside
+    a multi-line value (e.g., `wave_11_inner_key` at 2-space indent inside
+    `wave_11_scope: {...}`). Without this check the helper would insert a
+    new top-level key in the middle of a value, producing either invalid
+    JSON or a logical/text divergence caught only by the post-write
+    validation (which then aborts the entire upsert batch).
+
+    Performance: O(position) per call — for cross-repo-status.json
+    (~2700 lines, ~108KB) a single check is ~108K char-iterations, well
+    under 10ms even on slow hardware. Called once per sibling-finder
+    candidate, so worst case is ~108K × N candidates; N is typically 1–5
+    so the total per-upsert cost remains negligible.
+    """
+    depth = 0
+    in_string = False
+    escape = False
+    i = 0
+    while i < position and i < len(text):
+        c = text[i]
+        if escape:
+            escape = False
+        elif c == "\\" and in_string:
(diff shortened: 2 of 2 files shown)


parametrization:
Requestor: Santiago Ferreira
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: None

Secondary review (Release Coordinator). Read at head_sha `f1e5776d0a695538b750a136c8a1cb8fac5357a7` via `gh api contents`.

## Correctness — the fix

`_is_top_level_position(text, position)` (`upsert_status_keys.py:35-76`) is a structural JSON walker tracking bracket depth (`{`/`[`/`}`/`]`) and string state (with `\` escape handling) from text start. Returns `depth == 1 and not in_string` — exactly the structural-top-level check needed.

Both regex match sites are filtered:
- **Replace-in-place** (`:155-157`) — `line_re.finditer(text)` iterates ALL matches and skips non-top-level. Pre-#456 took the first regex match unconditionally; that mis-matched nested keys whose NAME equalled `key`.
- **Sibling-finder** (`:174`) — list comprehension filters `sibling_re.finditer` results. Pre-#456 took ALL matches including nested wave_N_* names inside value dicts.

## Test coverage — exhaustive for the regression class

`UpsertNestedSiblingFalseMatchTests` (4 tests) pins the regression class with the literal #456 reproducer + variations:
- **`test_nested_wave_key_at_2_space_indent_does_not_confuse_sibling_finder`** (`:225`) — the literal #456 reproducer: nested `wave_11_inner_*` keys at 2-space indent inside `wave_11_scope` value MUST NOT be treated as wave_11_* siblings.
- **`test_nested_wave_key_replace_does_not_target_nested_match`** (`:249`) — replace-in-place variant; verifies the depth-walker filters the in-place match too, not just the sibling-finder.
- **`test_nested_wave_key_at_4_space_indent_still_ignored`** (`:261`) — guard for the indent-mismatch case that already worked; pinned so a refactor doesn't break it.
- **`test_sequenced_upsert_after_inserting_multiline_dict`** (`:280`) — the actual /wave-* skill usage pattern: insert multi-line dict, then insert another wave_N_* key. The 5-key call shape from the #456 issue body.

`IsTopLevelPositionTests` (5+ tests) gives pure-function coverage for the walker: outer/nested object, nested array, inside string, brace-chars-inside-strings-don't-change-depth.

## Production smoke (PR body claim)

PR body: "applied to a copy of cross-repo-status.json with 4 representative upserts — passes". Reasonable verification for a helper used by /wave-scope, /wave-kickoff, /wave-wrapup; per [[feedback_live_trace_over_synthetic_acceptance]] (now charter), real-file smoke > synthetic-only. Accepted.

## Error message improvements (#456 acceptance)

Lines 227-256 surface offending key=value pairs (truncated to 80 chars) on BOTH failure modes:
- Invalid JSON: `Offending pairs: ...`
- Logical/text divergence: `Diverged keys: [...]` + `Offending pairs (this call): ...`

This addresses the #456 acceptance criterion "error message should pinpoint which key triggered the divergence" — a future failure becomes a paste-able reproducer in the resulting issue. Good shape.

## Release-coordination concern — CI paths gap (FLAGGING, not blocking)

**`.github/workflows/ci.yml` `paths:` filter does NOT include `.claude/lib/**`** (lines 6-10, 13-17 only cover `.claude/hooks/**`, `.claude/skills/**`, `.github/workflows/ci.yml`, `pyproject.toml`). This PR touches only `.claude/lib/**` and consequently shows **zero CI checks** (verified via `gh pr view 460 --json statusCheckRollup` → `[]`).

The fix code itself is correct (verified locally):
- `ENVIRONMENT=test pytest .claude/lib/tests/test_upsert_status_keys.py` — 18 tests pass
- `uvx ruff@0.7.4 check .claude/lib/` — clean
- `uvx ruff@0.7.4 format --check .claude/lib/` — 2 files already formatted

So the PR passes every gate that CI would have run — but CI isn't running. This is an **infrastructure gap, not a code-correctness gap**, and the right shape per [[feedback_runtime_gate_scoping]] is to land this PR (correct fix) and address the CI-paths gap in a parallel one-line PR. **I'm not requesting changes** because:
1. The fix is correct and tested locally at lint/format/mypy/pytest level.
2. The CI-paths gap predates this PR — every `.claude/lib/**` change has shipped without CI for as long as the workflow has existed.
3. Blocking this PR doesn't fix the underlying infrastructure issue.

**Recommendation to team-lead**: file a 1-line PR adding `.claude/lib/**` to both `paths:` blocks in `.github/workflows/ci.yml` (parallel to this PR, doesn't block it). Tracking issue can be filed via /file-bug if not already in scope.

## Tech-debt attestation

## Throughline-watch

Different throughline than #457/#458 (which were test-mock-bypasses-static-validation): this PR adds a **structural walker filter at every regex match site**, which is the right pattern when regex matching meets nested-context: ALWAYS layer a structural depth-check on top of regex matches in nested-text contexts. Worth retro consideration as a charter primitive: "Regex matches in JSON/YAML/Markdown text MUST be filtered through a structural depth-check unless the regex itself is anchored such that nesting cannot produce false positives."

Sibling throughline: the **CI-paths-filter-missing-lib gap** is itself a release-coordination issue worth charter promotion — "CI workflow `paths:` filters must cover every directory under `.claude/` that the workflow's jobs check. Add a CI gate to detect coverage gaps."

## Verdict

Approved. Fix is structurally correct, tests pin the exact #456 reproducer + 3 derived variants + 5 pure-walker tests, error messages surface offending pairs per acceptance criteria. Local lint/format/mypy/pytest all clean. The CI-paths infrastructure gap (separate concern) does not block this fix.

parametrization:
Requestor: Aino Virtanen
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: #464

## Verdict: Approved

Reviewed at PR head_sha `f1e5776` via `gh api repos/.../contents?ref=<sha>` (origin-over-local per charter). 2 files / +224 / -8.

### Critical finding (filed, not blocking)

**CI did not fire for this PR.** `.github/workflows/ci.yml` `paths:` filters list `.claude/hooks/**` + `.claude/skills/**` + `.github/workflows/ci.yml` + `pyproject.toml` — but NOT `.claude/lib/**`. Since the only changed files are under `.claude/lib/`, GitHub Actions never queued a run:

```
$ gh api repos/.../commits/f1e5776/check-runs --jq '.total_count'
0
$ gh run list --branch W.Mwangi/0456-upsert-status-keys-regression
[]
```

`mergeable_state: clean` because no required-status check blocks; that does NOT mean lint/format/types/tests ran. I verified all 4 gates locally:
- `uvx ruff@0.15.6 check .claude/lib/...` → clean
- `uvx ruff@0.15.6 format --check .claude/lib/...` → clean
- `mypy .claude/lib/... --ignore-missing-imports` → clean
- `ENVIRONMENT=test pytest .claude/lib/tests/test_upsert_status_keys.py` → 18 passed (matches PR body claim "was 9 → 18")

Filed **#464** for the CI path-filter gap (single-layer, detection-class). Not blocking #460 — the change itself is sound — but worth landing the workflow fix in the next batch so future `.claude/lib/**` PRs aren't merged without CI.

### Verification of the change itself

**Reproduced the #456 bug class against PR-head code.** Direct repro with the exact `wave_11_scope` + nested-2-space-indent shape from the issue body:

```python
text = '{\n  "phase": "phase-3",\n  "wave_11_scope": {\n  "wave_11_inner_a": 1,\n  "wave_11_inner_b": 2\n  }\n}'
out = upsert_top_level_key(text, "wave_11_meta_issue", "447")
# Pre-#460: would insert IN THE MIDDLE of wave_11_scope's value, producing invalid/divergent JSON.
# Post-#460 (verified): new key lands at TRUE top level, after wave_11_scope's closing brace,
#                       wave_11_scope's inner keys preserved unchanged.
```

Resulting JSON parses cleanly; nested `wave_11_inner_a` + `wave_11_inner_b` are intact. Bug fixed.

### Design judgment — structural position-filter vs bracket-aware walker

The team-lead's brief flagged this as a judgment call. **The chosen approach (structural `_is_top_level_position` post-filter applied to regex matches) is right.** Reasoning:

1. **Minimal surface change** — keeps the existing regex-find paradigm; adds an O(position) walker as a filter. Diff is +85/-8 in production code, easy to review and revert.
2. **Performance**: O(position × N) where N is small (1-5 candidate matches per upsert). The author cites well-under-10ms on a 108KB file; I read the impl and concur — the walker is a single linear pass with simple state.
3. **Independent testability** — `_is_top_level_position` has dedicated pure-function coverage (`IsTopLevelPositionTests`, 5 cases). The bracket-aware-walker alternative would couple parsing with key-finding, hurting testability.
4. **Correctly handles all the surprise cases the walker needs to** — verified by running the test suite + cross-checking:
   - Bracket chars inside string values (`{`, `}`, `[`, `]`): handled via `in_string` toggle on unescaped `"`.
   - Escaped quotes inside string values (`\"`): handled via `escape` flag tracking.
   - Deeply nested objects: handled via integer depth counter, not stack.
5. **No regression on the #332 multi-line-value insertion bug** — `UpsertMultilineArraySiblingTests`, `UpsertMultilineObjectSiblingTests`, `UpsertNestedStructuresTests`, `UpsertMultiLineFinalKeyTests` all still pass post-#456 fix.

A bracket-aware walker alternative would have:
- Required reimplementing key/value parsing (more code surface)
- Worse complexity in handling indent-irregular inputs (the existing regex tolerates `  "key":` shape variants; a hand-rolled parser would need to replicate that tolerance)
- Made the change much harder to review

The post-filter pattern is the right "minimal change that closes the bug class" — it preserves the existing regex's failure modes for other quirks (which #460 doesn't claim to fix) while specifically excluding the nested-mid-value-match class.

### Test coverage adequacy

- **4 regression tests** in `UpsertNestedSiblingFalseMatchTests` cover the exact #456 shape:
  - `test_nested_wave_key_at_2_space_indent_does_not_confuse_sibling_finder` — the canonical #456 reproducer
  - `test_nested_wave_key_replace_does_not_target_nested_match` — proves replace-in-place also filters correctly (catches the dual symmetric bug)
  - `test_nested_wave_key_at_4_space_indent_still_ignored` — guards against the easier already-worked case so future regex changes don't accidentally regress it
  - `test_sequenced_upsert_after_inserting_multiline_dict` — multi-call sequence covering the "first upsert lands a multi-line dict, second upsert mustn't get confused by the dict's contents" scenario
- **5 pure-function tests** in `IsTopLevelPositionTests` cover the walker itself — directly exercises depth tracking, nested-object detection, nested-array detection, in-string detection, brace-in-string handling
- **All 9 prior tests** still pass (singleline-sibling baseline, multiline array, multiline object, nested structures, strings-containing-brackets, escaped quotes, idempotence, final-key-no-trailing-comma, multi-line-final)

### Error-message improvements (#456 secondary acceptance)

Verified the post-write validation error messages at lines 222-256:
- Both failure modes (invalid JSON, divergent reparse) now surface "Offending pairs" with key=value pairs from this call, truncated to 80 chars per value (enough to paste into a reproducer without flooding the terminal).
- The divergence-case message lists `Diverged keys` (set-symmetric-difference + same-key-different-value scan) so reviewers see exactly which key drifted without diffing the full file.

Good acceptance-bar discipline.

**`_find_value_end` raises `ValueError("unterminated multi-line value...")`** at line 124. This propagates out of `upsert_top_level_key` → `main` and would print a Python traceback rather than a friendly error. Probably acceptable since this case is "the JSON file is structurally malformed before we touch it" — the upstream `json.loads(text)` at line 216 will catch most such cases first, so this branch is genuinely unreachable in well-formed inputs. Worth noting only because the rest of the error-handling is so polished.

**`open_brace = text.find("{\n")`** at line 189 will pick the first `{\n` pair, which would be wrong if the JSON file had a leading comment or whitespace before the outer `{`. Standard JSON has no comments, so this is fine in practice — cross-repo-status.json is always pure JSON. Worth flagging only as a "if we ever support JSON5 with leading comments, revisit".

Neither is worth a follow-up issue.

### Throughline observation (W11 retro candidate, additive)

Three W11 PRs from Wanjiku (#457, #458, #460) all share a pattern: **the bug fix is paired with a structural test that pins the failure mode at the right abstraction level**. #457's `GraphQLVariableUsageTests` pin GraphQL declaration-vs-body parity. #458's `ShellPythonHashParityTests` pin shell-vs-Python hash parity. #460's `IsTopLevelPositionTests` pin the walker's depth invariant. Each is independently testable and each catches its bug class without needing the full integration scenario. Worth promoting at retro as "convergent structural testing" — a process-success pattern, not a failure-mode finding.

### Verdict basis

Approve on:
- Bug correctly reproduced AND verified fixed against PR-head code (in-band repro of #456 shape)
- 18 tests pass locally (matches PR body's "was 9 → 18" delta; +9 = 4 regression + 5 pure-function as advertised)
- Lint/format/mypy verified locally (CI didn't fire — filed as #464)
- Design choice (post-filter over walker rewrite) is correct trade-off
- Santiago's secondary Approved already posted

PR #461, opened 2026-05-17
tech-debt(annunaki): env-var-gated test-mode write suppression (#452)

### LINKED ISSUE #452: tech-debt(annunaki): hook-test-suite fixtures captured as real errors (76% noise rate)
## Summary

`annunaki_log.py` captures hook-test-suite fixture invocations as real error entries. As of 2026-05-16 the error log has 388 entries, of which 293 (~76%) are test fixtures from the 22:27–22:37 UTC test-run window (recognizable by synthetic commands: `gh pr review 42`, `git config user.name foo`, `gh issue create --label does-not-exist`, `gh pr merge 100 --squash`, etc.).

This degrades the signal-to-noise of `/annunaki` (status viewer) and `/annunaki-attack` (issue-filer), and creates a real risk of missing genuine production errors in the noise.

## Evidence

Distribution of post-W9-attack 388-entry annunaki log:

| Hook | test-like | real-cmd |
|---|---|---|
| block_gh_pr_review | 75 | 0 |
| block_stale_tmp_message_file | 0 | 62 |
| validate_commit_identity | 0 | 48 |
| validate_review_comment_format | 0 | 45 |
| post_label_change_wave_field_sync | 24 | 8 |
| block_no_verify | 25 | 0 |
| validate_pr_review | 25 | 0 |
| validate_branch_freshness | 15 | 0 |
| validate_wave_audit | 15 | 0 |
| validate_wave_label_evidence | 15 | 0 |
| post_wave_kickoff_comment | 11 | 0 |
| block_git_config | 15 | 0 |
| validate_labels | 5 | 0 |
| **TOTAL** | **270** | **163** |

(Note: my heuristic undercounted test-like entries; obvious fixtures like `gh pr review 42` weren't caught by my string-match. Actual test-noise is closer to ~293 / 76%.)

The test-run window cluster: 142 entries at 22:29, 76 at 22:28, 75 at 22:37, 71 at 22:30 — clearly a test-suite run, not user activity.

## Acceptance Criteria

- [ ] `annunaki_log.py` detects test-harness invocation and suppresses logging. Options:
  - Env var `NOORIN_HOOK_TEST_MODE=1` set by test harness; annunaki bails if set
  - Detect parent process name (e.g., `pytest`, `python3 -m unittest`) via `/proc/self/status` and bail
  - Inspect the input payload for fixture markers (e.g., issue number `42`, label `does-not-exist`) — least preferred, brittle
- [ ] Preferred: env-var approach. Add `os.environ["NOORIN_HOOK_TEST_MODE"] = "1"` to the hook test runner's setup (`.claude/hooks/tests/conftest.py` or equivalent).
- [ ] Test: extend annunaki tests to verify env-var suppression works AND production logging is unaffected when env var absent.
- [ ] Backfill: clear the 293+ test-fixture entries from `errors.jsonl` once the suppression lands.

## Why bundle into W11

Direct prerequisite for any future `/annunaki-attack` to produce trustworthy output. W11 theme covers tech debt; this is observability-tier debt blocking the observability surface.

## References

- Discovered during P3W11 `/annunaki-attack` run 2026-05-17
- [[feedback_enforcement_hierarchy]] — annunaki is a hook-tier surface; if its data is noisy, the whole hierarchy degrades

Origin-Verification: `.claude/hooks/annunaki_log.py` exists at origin/main (confirmed via `grep -rln annunaki_monitor`); test harness at `.claude/hooks/tests/` exists.

parametrization:
Closes #452.

## Problem

`annunaki_log.append_jsonl_record` writes unconditionally to `.claude/annunaki/errors.jsonl`. The hook test suite exercises matchers against synthetic fixtures (many of which intentionally trip block hooks to verify their behavior). Each such test fires the production logger, appending a "fake error" entry to the production log.

Per the issue body (2026-05-16 capture): 293 of 388 (~76%) entries in the production log were test-fixture noise from a single test-suite run. This degrades the signal-to-noise of /annunaki and /annunaki-attack.

## Fix

Added `_is_test_mode()` to annunaki_log.py — returns True when either:
- `ENVIRONMENT=test` (charter-required env var for pytest; set by auto_set_env_test.py PreToolUse hook)
- `NOORIN_HOOK_TEST_MODE=1` (explicit non-pytest opt-in)

`append_jsonl_record` early-returns when test-mode is True. Since both `log_pretooluse_block` and `log_posttooluse_event` route through it, suppression covers every annunaki-logging caller without per-hook changes.

## Why ENVIRONMENT=test as the signal

The charter already requires `ENVIRONMENT=test` for pytest invocations (enforced by auto_set_env_test.py). Every existing test invocation already sets it. No conftest.py setup; no migration of existing test files; no per-test mocking.

## Verification

Pre-#452: `ENVIRONMENT=test pytest .claude/hooks/tests/` added 200+ fixture entries to errors.jsonl per run.

Post-#452: same command adds 0 entries. Verified with wc -l before/after.

The backfill of historical fixture entries is deferred to a follow-up — this PR delivers the prevention; the backfill is a separate ops-cleanup action.

## Test plan

- [x] `ENVIRONMENT=test pytest .claude/hooks/tests/test_annunaki_log.py` — 13 passing
- [x] Full hook test suite still passes
- [x] Production smoke: `wc -l errors.jsonl` before/after a full test run shows 0 new entries
- [x] ruff + pre-commit clean

## Tech Debt

Generated with Claude Code

### DIFF
--- a/.claude/hooks/annunaki_log.py
+++ b/.claude/hooks/annunaki_log.py
@@ -7,9 +7,28 @@ commands appear in the Annunaki error log alongside PostToolUse errors.
 Usage in any blocking hook:
     from annunaki_log import log_pretooluse_block
     log_pretooluse_block(hook_name="validate_commit_identity", command=command, reason=reason)
+
+Test-mode suppression (issue #452)
+==================================
+
+When `ENVIRONMENT=test` or `NOORIN_HOOK_TEST_MODE=1` is set in the env,
+`append_jsonl_record` returns without writing. This prevents the hook
+test suite (which exercises matchers against synthetic fixtures like
+`gh pr review 42`, `gh issue create --label does-not-exist`) from
+polluting the production `.claude/annunaki/errors.jsonl` with hundreds
+of fake error entries — pre-#452 a single test-suite run produced ~293
+entries (76% of the log's content as captured 2026-05-16).
+
+`ENVIRONMENT=test` is the charter-required env var for pytest (set by
+`auto_set_env_test.py` PreToolUse hook); using it as the suppression
+signal means no conftest.py setup is needed — every existing test
+invocation already sets it. `NOORIN_HOOK_TEST_MODE` is provided as an
+explicit opt-in for non-pytest harnesses (CI matrix jobs, ad-hoc shell
+fixtures) that need annunaki suppression without claiming `ENVIRONMENT=test`.
 """
 
 import json
+import os
 from datetime import UTC, datetime
 from pathlib import Path
 
@@ -17,12 +36,36 @@ REPO_ROOT = Path(__file__).resolve().parent.parent.parent
 ERRORS_FILE = REPO_ROOT / ".claude" / "annunaki" / "errors.jsonl"
 
 
+def _is_test_mode() -> bool:
+    """True iff annunaki should suppress writes (test harness running).
+
+    Returns True when EITHER:
+      - `ENVIRONMENT=test` is set (charter-required for pytest)
+      - `NOORIN_HOOK_TEST_MODE=1` is set (explicit non-pytest opt-in)
+
+    Empty / unset / any other value → False (production fall-through).
+    """
+    if os.environ.get("ENVIRONMENT", "") == "test":
+        return True
+    if os.environ.get("NOORIN_HOOK_TEST_MODE", "") == "1":
+        return True
+    return False
+
+
 def append_jsonl_record(path: Path, record: dict) -> None:
     """Append one JSONL record. Skips empty records and guarantees exactly
     one trailing newline per line — never a bare blank line. Shared by
     annunaki_log.py and annunaki_monitor.py so writer hardening stays in
-    one place."""
+    one place.
+
+    Test-mode suppression: when `_is_test_mode()` is True (ENVIRONMENT=test
+    or NOORIN_HOOK_TEST_MODE=1), the write is skipped entirely. This is the
+    #452 fix — the hook test suite pollutes the production error log with
+    ~293 fixture entries per run otherwise.
+    """
     if not isinstance(record, dict) or not record:
+        return
+    if _is_test_mode():
         return
     # json.dumps with default settings does not emit newlines, so the
     # serialized form is guaranteed single-line. Strip any stray ones
--- a/.claude/hooks/tests/test_annunaki_log.py
+++ b/.claude/hooks/tests/test_annunaki_log.py
@@ -1,0 +1,160 @@
+#!/usr/bin/env python3
+"""Tests for `annunaki_log` (#452 test-mode suppression).
+
+Coverage focuses on the env-var-gated write suppression that prevents the
+hook test suite from polluting production error log with ~293 fixture
+entries per run.
+
+Run from the repo root:
+    ENVIRONMENT=test python3 -m pytest \\
+        .claude/hooks/tests/test_annunaki_log.py -v
+"""
+
+from __future__ import annotations
+
+import os
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+
+_HERE = Path(__file__).resolve().parent
+_HOOKS_DIR = _HERE.parent
+sys.path.insert(0, str(_HOOKS_DIR))
+
+import annunaki_log as alog  # noqa: E402
+
+
+class IsTestModeTests(unittest.TestCase):
+    """Pure-function coverage for `_is_test_mode`."""
+
+    def setUp(self):
+        # Save and clear both env vars so each test starts clean.
+        self._saved = {
+            "ENVIRONMENT": os.environ.pop("ENVIRONMENT", None),
+            "NOORIN_HOOK_TEST_MODE": os.environ.pop("NOORIN_HOOK_TEST_MODE", None),
+        }
+
+    def tearDown(self):
+        for k, v in self._saved.items():
+            if v is None:
+                os.environ.pop(k, None)
+            else:
+                os.environ[k] = v
+
+    def test_unset_returns_false(self):
+        self.assertFalse(alog._is_test_mode())
+
+    def test_environment_test_returns_true(self):
+        os.environ["ENVIRONMENT"] = "test"
+        self.assertTrue(alog._is_test_mode())
+
+    def test_environment_production_returns_false(self):
+        os.environ["ENVIRONMENT"] = "production"
+        self.assertFalse(alog._is_test_mode())
+
+    def test_environment_empty_returns_false(self):
+        os.environ["ENVIRONMENT"] = ""
+        self.assertFalse(alog._is_test_mode())
+
+    def test_noorin_hook_test_mode_1_returns_true(self):
+        os.environ["NOORIN_HOOK_TEST_MODE"] = "1"
+        self.assertTrue(alog._is_test_mode())
+
+    def test_noorin_hook_test_mode_0_returns_false(self):
+        """Unix-tradition truthy-only — literal `1` is the only true value."""
+        os.environ["NOORIN_HOOK_TEST_MODE"] = "0"
+        self.assertFalse(alog._is_test_mode())
+
+    def test_noorin_hook_test_mode_true_string_returns_false(self):
+        """`true` does NOT activate — only literal `1`."""
+        os.environ["NOORIN_HOOK_TEST_MODE"] = "true"
+        self.assertFalse(alog._is_test_mode())
+
+    def test_either_env_var_set_suffices(self):
+        """The two env vars are OR-combined."""
+        os.environ["NOORIN_HOOK_TEST_MODE"] = "1"
+        os.environ["ENVIRONMENT"] = "production"  # ENVIRONMENT not test
+        self.assertTrue(alog._is_test_mode())
+
+
+class SuppressionTests(unittest.TestCase):
+    """End-to-end: when test-mode is active, `append_jsonl_record` MUST NOT
+    write to the file. When test-mode is inactive, it MUST write."""
+
+    def setUp(self):
+        self._saved = {
+            "ENVIRONMENT": os.environ.pop("ENVIRONMENT", None),
+            "NOORIN_HOOK_TEST_MODE": os.environ.pop("NOORIN_HOOK_TEST_MODE", None),
+        }
+
+    def tearDown(self):
+        for k, v in self._saved.items():
+            if v is None:
+                os.environ.pop(k, None)
+            else:
+                os.environ[k] = v
+
+    def test_environment_test_suppresses_write(self):
+        os.environ["ENVIRONMENT"] = "test"
+        with tempfile.TemporaryDirectory() as tmp:
+            target = Path(tmp) / "errors.jsonl"
+            alog.append_jsonl_record(target, {"test": "fixture"})
+            self.assertFalse(target.exists(), "ENVIRONMENT=test must suppress annunaki writes")
+
+    def test_noorin_hook_test_mode_suppresses_write(self):
+        os.environ["NOORIN_HOOK_TEST_MODE"] = "1"
+        with tempfile.TemporaryDirectory() as tmp:
+            target = Path(tmp) / "errors.jsonl"
+            alog.append_jsonl_record(target, {"test": "fixture"})
+            self.assertFalse(target.exists(), "NOORIN_HOOK_TEST_MODE=1 must suppress writes")
+
+    def test_production_env_does_not_suppress_write(self):
+        """The whole point: when neither env var is set, writes go through."""
+        # Both env vars cleared in setUp.
+        with tempfile.TemporaryDirectory() as tmp:
+            target = Path(tmp) / "errors.jsonl"
+            alog.append_jsonl_record(target, {"production": "real_error"})
+            self.assertTrue(
+                target.exists(),
+                "Production mode (no test-mode env vars) must allow writes",
+            )
+            content = target.read_text()
+            self.assertIn("real_error", content)
+
+    def test_log_pretooluse_block_respects_suppression(self):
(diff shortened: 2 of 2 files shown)


parametrization:
Requestor: Aino Virtanen
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: none

## Verdict: Approved

Reviewed at PR head_sha `1fb5ea9` via `gh api repos/.../contents?ref=<sha>` (origin-over-local per charter). 2 files / +204 / -1.

### Verification

- Read both files at HEAD.
- Confirmed wave-11 base of `annunaki_log.py` is the 82-line pre-#461 form via `git show origin/deployments/phase-3/wave-11:.claude/hooks/annunaki_log.py | wc -l` — the new +43 lines (`_is_test_mode`, docstring section, early-return at `append_jsonl_record`) are exactly what's in the diff.
- Ran the 13 new tests locally with PR-head files copied into my worktree: **13 passed in 0.04s** (matches PR body's "13 passing" claim).
- Ran full hooks suite locally: **802 passed** (no regressions).
- `uvx ruff@0.15.6 check` + `format --check` + `mypy --ignore-missing-imports` all clean on both files.
- `uvx ruff@0.15.6 check --select UP` clean — settles the team-lead's earlier UP017 sibling question (no UP017 lingering in `annunaki_log.py`; the pre-#461 form was already clean via the `0ebeb03 fix(hooks): annunaki_log — use datetime.UTC alias (pyupgrade UP017)` commit on wave-11 base).
- CI: 4/4 green at HEAD (mypy / ruff lint / ruff format / smoke).

### Why the design is right

**Single chokepoint.** `_is_test_mode()` is consulted exactly once, at `append_jsonl_record:68`. Both `log_pretooluse_block` and `log_posttooluse_event` route through `append_jsonl_record` (lines 99 + 125), so the suppression is universal without per-hook plumbing. Adding a new logger entry-point later just needs to call `append_jsonl_record` and inherits the suppression for free. This is exactly the [[feedback_enforcement_hierarchy]] preferred shape — close the gap at the deepest layer that all paths must traverse.

**Env-var signal choice is sound.** `ENVIRONMENT=test` is set by `auto_set_env_test.py` PreToolUse hook on every pytest invocation (charter-required, hook-enforced). That means every existing test already has the signal in scope — no conftest.py needs to be added, no per-test mocking, no migration to the existing 800+ test suite. Pre-#461 the entire test suite polluted production logs; post-#461 they go to /dev/null without any test author having to know about it. The `NOORIN_HOOK_TEST_MODE=1` opt-in is the right escape hatch for non-pytest harnesses (CI matrix jobs, ad-hoc shell fixtures) that can't claim `ENVIRONMENT=test` honestly.

**Unix-tradition truthy-only semantics pinned by test.** `test_noorin_hook_test_mode_0_returns_false` + `test_noorin_hook_test_mode_true_string_returns_false` pin that ONLY literal `1` activates `NOORIN_HOOK_TEST_MODE`. Matches the same shape the post-label-change Hook 21's `_kill_switch_active` already established (literal `1` only). One less semantic-drift surface across hooks.

**Test-mode is a property of `append_jsonl_record`, not of the loggers.** The early-return is at line 68-69 of `append_jsonl_record`, BEFORE the empty-record check (line 66-67 already returns on empty). This means even a malformed log call gets the suppression — defense-in-depth ordering is correct.

### Test coverage adequacy

**8 IsTestModeTests** cover the full truth table:
- Both env vars unset → False (baseline)
- `ENVIRONMENT=test` → True
- `ENVIRONMENT=production` → False (negative case for `test` specificity)
- `ENVIRONMENT=""` → False (empty-vs-unset is sometimes a foot-gun; pin both)
- `NOORIN_HOOK_TEST_MODE=1` → True
- `NOORIN_HOOK_TEST_MODE=0` → False (Unix-truthy invariant)
- `NOORIN_HOOK_TEST_MODE=true` → False (only literal `1`)
- Both set with `NOORIN` overriding non-test `ENVIRONMENT` → True (OR-combine)

**5 SuppressionTests** cover end-to-end behavior:
- `ENVIRONMENT=test` suppresses write (positive)
- `NOORIN_HOOK_TEST_MODE=1` suppresses write (positive)
- Neither env var set → write goes through (production positive — the LOAD-BEARING test that proves the suppression is gated, not unconditional)
- `log_pretooluse_block` respects suppression (higher-level wrapper inherits)
- `log_posttooluse_event` respects suppression (other higher-level wrapper inherits)

The `test_production_env_does_not_suppress_write` is the most important one — it proves the fix is NOT a "log nothing ever" regression. Without it, a future bug that flipped the polarity of the env-var check would still pass all the suppression tests.

**`setUp`/`tearDown` env-var save+restore pattern is correct.** Env vars are process-global; without restore the test order would matter (some tests would inherit `ENVIRONMENT=test` from the pytest harness and produce spurious passes). Tests using `os.environ.pop()` + restore in `tearDown` keep test isolation tight. An alternative would have been `unittest.mock.patch.dict(os.environ, {...}, clear=True)` as a context manager — slightly more concise but functionally equivalent.

**`alog.ERRORS_FILE = orig` monkey-patching** in `test_log_pretooluse_block_respects_suppression` and `test_log_posttooluse_event_respects_suppression` (lines 134 + 148) directly rewrites a module-level constant during the test, then restores it. This is the same pattern other tests in the suite use (the docstring at line 130 cites `_wipe_cache` precedent). Works fine; an alternative would be `unittest.mock.patch.object(alog, "ERRORS_FILE", target)` for context-managed restoration — slightly more idiomatic but functionally equivalent. Not worth a follow-up.

**Backfill of historical fixture entries is deferred** — the PR body acknowledges this explicitly as out-of-scope: "This PR delivers the prevention; the backfill is a separate ops-cleanup action." Correct scope discipline. The historical noise is in `.claude/annunaki/errors.jsonl` (gitignored at the line level via individual entries, not directory-level), so a one-off `jq -c '. | select(<not test fixture shape>)'` cleanup or just a fresh-start truncate handles it. Worth noting as a one-line follow-up later but not file-bug-worthy today (low value, easy to do by hand at retro).

### Throughline observation (W11 retro candidate, additive to prior batch)

This is the **4th** Wanjiku PR this batch following the same convergent-structural-testing pattern (#457 `GraphQLVariableUsageTests` / #458 `ShellPythonHashParityTests` / #460 `IsTopLevelPositionTests` / #461 `IsTestModeTests`). Each pairs the fix with a pure-function or boundary-condition test class that pins the load-bearing invariant at the right abstraction level — testable in isolation, no integration scenario needed. Consistent enough across 4 PRs now to be worth promoting as a process success at W11 retro.

### Verdict basis

Approve on:
- Single-chokepoint design at `append_jsonl_record` correctly captures all logger callers
- Env-var signal choice (`ENVIRONMENT=test`) is right — charter-mandated, hook-enforced, already in every test invocation
- 8 + 5 = 13 tests cover the full truth table + end-to-end suppression positive AND negative paths
- The production-positive test (`test_production_env_does_not_suppress_write`) prevents polarity-flip regressions
- Unix-tradition truthy-only semantics pinned consistent with prior hook patterns
- All lint/format/mypy/UP-rules clean; CI 4/4 green; 802 full-hook-suite passing
- Backfill deferred is correct scope discipline
- This is the convergent reviewer Approve — Santiago wasn't on this one yet per team-lead's note; the 2-reviewer gate completes when he posts

parametrization:
Requestor: Santiago Ferreira
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: None

Secondary review (Release Coordinator). Read at head_sha `1fb5ea9be7aa6d044bfd215e8af7e03181e2df90` via `gh api contents`.

## Architectural shape — single chokepoint

`_is_test_mode()` (`annunaki_log.py:39-52`) is checked at the bottom of the call stack in `append_jsonl_record` (line 68), which is the ONLY file-writing function in the module. Both `log_pretooluse_block` (line 84) and `log_posttooluse_event` (line 102) route through it. **One guard, complete coverage.** No per-hook modifications needed, no risk of missing a caller.

This is exactly the right shape for a cross-cutting concern. Compare to per-hook env-checking: 8 hooks × per-call branch = 8 surfaces for a future bug; this PR has 1 surface. Single source of truth eliminates the divergence vector ([[feedback_test_mock_injection_masks_production_failure]]-adjacent — divergence between testing and production code paths).

## Suppression-trigger choice — verified against existing infrastructure

PR body claims `ENVIRONMENT=test` is "charter-required env var for pytest (set by auto_set_env_test.py PreToolUse hook)". Verified at `.claude/hooks/auto_set_env_test.py:119`: `if re.search(r"\bENVIRONMENT=test\b", command):` — the hook blocks any `pytest` / `make test` command lacking the prefix. So `ENVIRONMENT=test` is GUARANTEED set for every pytest invocation via the PreToolUse layer.

**This means no conftest.py is needed** — every existing test invocation already sets the suppression signal. Zero-migration adoption. Strong release-coordination shape: the change is opt-in via existing infrastructure rather than requiring every test file to learn about a new test-mode flag.

`NOORIN_HOOK_TEST_MODE=1` (line 50) as the non-pytest opt-in covers the CI-matrix / shell-fixture cases where the auto_set_env_test.py PreToolUse hook isn't in the loop. Sensible explicit escape hatch.

**Literal `=test` / `=1` matching** (lines 48, 50) — narrow string equality prevents accidental truthy interpretation of values like `=true`, `=enabled`, etc. Unix-tradition truthy-only matches the [[feedback_runtime_gate_scoping]] precedent.

## Test architecture — env-var teardown verified

The brief flagged env-var teardown as a release-coordination concern. Verified:

1. **setUp/tearDown pattern** (lines 31-43, 85-96) — `os.environ.pop(KEY, None)` saves original (or None sentinel), tearDown restores or pops. **Properly bounded — no global env pollution between tests.**

2. **Verified empirically** — ran the suite in a subprocess and checked the parent env: `NOORIN_HOOK_TEST_MODE` remained `<unset>` in the parent after the subprocess exited; `ENVIRONMENT=test` stayed because the parent explicitly set it for the subprocess. No leakage.

3. **Critical because** if these tests leaked an `ENVIRONMENT=test` setting outside their test functions, the subsequent runs would silently suppress real annunaki writes during local dev. The teardown discipline prevents that footgun.

4. **No new conftest.py** — verified `find .claude/hooks/tests/ .claude/lib/tests/ -name conftest.py` returns no results. The PR genuinely leverages the existing `auto_set_env_test.py` infrastructure rather than adding new setup. Matches PR body claim.

## Test coverage — exhaustive for the suppression-signal class

`IsTestModeTests` (8 tests): every env-var combination — unset, `test`, `production`, empty, `1`, `0`, `true`, both-set.

`SuppressionTests` (5 tests): end-to-end via `tempfile.TemporaryDirectory()`:
- `ENVIRONMENT=test` → no file written
- `NOORIN_HOOK_TEST_MODE=1` → no file written
- production (neither set) → file written + content correct
- `log_pretooluse_block` routes through suppression
- `log_posttooluse_event` routes through suppression

Both higher-level helpers explicitly tested for routing through `append_jsonl_record` (lines 125-156). Verifies the single-chokepoint architecture holds end-to-end.

## Reversibility — backfill deferred is the right call

PR body: "The backfill of historical fixture entries is deferred to a follow-up — this PR delivers the prevention; the backfill is a separate ops-cleanup action."

This is the right shape per [[feedback_security_guard_inline_not_followup]] (inverse application): the **prevention IS the inline fix**; the backfill is ops-tracking. The historical 293 fixture entries are sunk cost — removing them is an ops cleanup, not a fix. If backfill were bundled in, this PR would have a "git/destructive history rewrite" axis that complicates review.

Also fully reversible: if this hook misbehaves in production, removing the 2-line `if _is_test_mode(): return` guard restores prior behavior immediately; the env-var infrastructure remains usable for future iteration.

## Release-coordination angle

- **Cross-repo impact**: `annunaki_log` is imported by ~10+ hooks across `.claude/hooks/`. This change affects every annunaki-emitting hook simultaneously. Verified via grep: `annunaki_log` is imported by `post_label_change_wave_field_sync.py`, `auto_add_issue_to_board.py` (both touched in #457), `enforce_librarian_consulted.py`, and others. **All gain test-mode suppression atomically.**
- **Production log signal-to-noise**: ~76% fixture pollution → 0% per PR-body smoke (`wc -l errors.jsonl` before/after a full test run shows 0 new entries). This directly improves `/annunaki` and `/annunaki-attack` signal quality — both depend on the production log being mostly-real-failures.
- **Sequencing**: independent of #457/#458/#459/#460 — no merge ordering required. Recommend landing immediately to improve the signal-to-noise of any post-merge `/annunaki` sweep.
- **PR CI all green** (Ruff lint / Ruff format / Mypy / Smoke).
- **Local verification at head_sha 1fb5ea9**: 13 tests pass in this file; full hook+lib suite 800+19 subtests pass (no regression).

## Tech-debt attestation

## Throughline-watch

This PR exemplifies the **single-chokepoint discipline** for cross-cutting concerns (signal-emission gating in a 10+-hook ecosystem). Pair with #457's GraphQL-helper consolidation (Hook 13 imports Hook 21 helpers rather than copying) — both PRs demonstrate the same architectural primitive: **when N hooks share a behavior, gate it at the single point of common usage, not at each call site.** Charter-promotion candidate: "Cross-cutting hook behaviors (env gates, file writes, GraphQL helpers) MUST be implemented at a single chokepoint shared by all callers, not duplicated per-hook."

This is structurally distinct from the test-mock-bypasses-static-validation throughline (#457/#458) but lives in the same retro family of "single-source-of-truth" patterns. Worth weaving into the W11 retro as a sibling principle.

## Verdict

Approved. Single-chokepoint architecture is the right shape for cross-cutting suppression; env-var teardown is properly bounded; reuses existing `auto_set_env_test.py` infrastructure with no migration cost; backfill correctly deferred. Local tests pass + production smoke confirms 0 new pollution per test run.

Issue #471, opened 2026-05-17
tech-debt(enforce_ontology_context): whitespace-indent leak in coordinator-exemption regex


parametrization:
## Source

Aino Virtanen's re-anchored Approved verdict on PR #468 ([comment 4472705253](https://github.com/noorinalabs/noorinalabs-main/pull/468#issuecomment-4472705253)) flagged a non-blocking residual after the merged fixup:

> "The new `re.MULTILINE` + `^\s*You are` shape creates a small leak surface for any prompt body containing a whitespace-indented `   You are X, <CoordRole>` line. Realistic-trigger shapes are uncommon (markdown bullets/blockquotes use `-`/`>` prefixes which `\s*` won't match; only 4-space-indent code blocks and YAML-like indented examples would trigger). Hook 15 covers actual Edit/Write surface. Tighter alternative for a future iteration: `(?:\A|\n)You are` (no indent tolerance)."

## Current behavior

`COORDINATOR_ROLE_OPENER` in `.claude/hooks/enforce_ontology_context.py` (landed via PR #468 squash `976f6c8`):

```python
COORDINATOR_ROLE_OPENER = re.compile(
    r"^\s*You are\s+\*{0,2}[^,\n]+?\*{0,2},\s*"
    r"(Pipeline\s+Manager"
    r"|Project\s+Lead"
    r"|Program\s+Director"
    r"|Technical\s+Program\s+Manager"
    r"|TPM"
    r"|Release\s+Coordinator"
    r"|Manager)\b",
    re.IGNORECASE | re.MULTILINE,
)
```

The `^\s*` + `re.MULTILINE` combination means any line within the prompt body starting with whitespace + `You are X, <CoordRole>` would falsely exempt the spawn.

## Trigger shapes (realistic)

1. **4-space-indent code blocks** in implementer briefs that contain example coordinator spawn briefs (e.g., reviewer brief that quotes a manager spawn brief as a reference).
2. **YAML-like indented examples** in briefs that include role-card excerpts at indent.
3. **Spawn briefs with leading indented context blocks** that include canonical openers as illustration.

Markdown bullets (`- `) and blockquotes (`> `) do NOT trigger — `\s*` doesn't match `-` or `>`.

## Proposed Fix

**Type:** Hook refinement.

**Tighten the leading anchor** from `r"^\s*You are"` (with `re.MULTILINE`) to either:

[...]

Issue #470, opened 2026-05-17
tech-debt(charter+wave-kickoff): document coordinator-class ontology-context exemption (#468 follow-up)


parametrization:
## Problem

PR #468 carved out coordinator-class spawns from `enforce_ontology_context.py`, but two related documentation/skill surfaces still assert that the `## Ontology Context` heading is universally required:

1. **`.claude/team/charter/agents.md:259`** — Position-first rule says *"Every spawn prompt template MUST place this section immediately after the identity / git-identity preamble and BEFORE the `## Ontology Context` section."* This is now conditionally true (only for implementer-class briefs).
2. **`.claude/team/charter/agents.md:322`** — *"`## Ontology Context` section (literal heading) with librarian output baked in — `enforce_ontology_context.py` scans for this heading and blocks the spawn if absent."* Same: now coordinator-class is exempt.
3. **`.claude/skills/wave-kickoff/SKILL.md:335` and `:343`** — Describes the heading as a hard requirement with no coordinator-class carveout. Wave-kickoff is the primary spawn-emitting skill, so this is the most-likely place an orchestrator looks when composing briefs.

## Why this matters

Charter/skill text that lags hook behavior re-introduces the [[feedback_enforcement_hierarchy]] decay risk: orchestrators reading the charter learn the old contract, then waste cycles baking in headings the hook no longer enforces (low cost), OR — worse — assume the heading is the ONLY enforcement and skip the librarian call for coordinator briefs because "it's no longer needed" (Hook 15 still requires it at Edit/Write surface, so they'd hit a second-tier block).

Charter and skill documentation should match hook behavior so that the orchestrator's mental model stays coherent.

## Proposed fix

Add a one-paragraph "Coordinator-class exemption" callout to each of the three locations:

- `agents.md:259` / `:322` — note that coordinator-class spawns (Manager / Program Director / TPM / Release Coordinator) MAY omit the `## Ontology Context` section. Hook 15 still applies at the Edit/Write surface.
- `wave-kickoff/SKILL.md` § Step on baking in ontology context — same callout.

## Provenance

Discovered during Santiago's review of #468 (Release Coordinator angle). Sibling to the PR's own self-noted ontology/conventions.md follow-up (PR body § Follow-up — `enforce_ontology_context.py` is missing from the hooks table).

Could be bundled with that follow-up as a single "post-#468 documentation sync" pass.

parametrization:
[...]

Issue #467, opened 2026-05-17
tech-debt(post_wave_kickoff_comment): filter out between-wave relabel events (36-event noise burst)


parametrization:
## Source

Captured by Annunaki error monitor (`/annunaki-attack` run 2026-05-17).

**Hook:** `post_wave_kickoff_comment` (PostToolUse on Bash)
**Pattern:** `hook_event:post_wave_kickoff_comment`
**Occurrences:** 36 events between 00:11:16Z and 04:02:51Z
**Most recent:** 2026-05-17T04:02:51Z

**Sample event reasons:**

```
wave_11_scope is missing or empty in cross-repo-status.json; cannot render kickoff comment for noorinalabs-deploy#45.
wave_11_scope is missing or empty in cross-repo-status.json; cannot render kickoff comment for noorinalabs-deploy#20.
wave_9_scope is missing or empty in cross-repo-status.json; cannot render kickoff comment for noorinalabs-main#125.
```

**Sample triggering commands:**

```
gh issue edit 45 --repo noorinalabs/noorinalabs-deploy --add-label "p3-wave-11"
gh issue edit 125 --repo noorinalabs/noorinalabs-main --add-label "p3-wave-9"
```

## Root cause

The hook fires correctly on every `gh issue edit ... --add-label "p{N}-wave-{M}"`, including:
- **Initial wave-label add** (intended trigger — render kickoff comment if wave is kicked off)
- **Between-wave relabel** (carry-forward: `p3-wave-10` → `p3-wave-11`, when W11 isn't kicked off yet)
- **Re-application** of an existing label

In the W11 case, 36 carry-forward relabels fired the hook, which then correctly bailed because `wave_11_scope` doesn't exist in `cross-repo-status.json` yet (W11 hasn't been formally kicked off via `/wave-kickoff`).

The bailout is CORRECT behavior — but it generates 36 noise events in annunaki because the hook also logs the bail via `log_posttooluse_event`.

## Already-known soft watching-brief

W10 retro / feedback log already classified this:

[...]

Issue #395, opened 2026-05-12
tech-debt(hook): post_dispatcher.py refactor — multiplex PostToolUse Bash hooks at ~5-hook threshold


parametrization:
## Summary

Pre-file the `post_dispatcher.py` refactor sentinel before the PostToolUse Bash hook count crosses the agreed ~5-hook threshold. Aino flagged this as the correct deferred refactor in PR #394 (closes #286).

## Current state (at #394 merge)

`.claude/settings.json` PostToolUse `Bash` matcher has **4 entries**:

1. `auto_add_issue_to_board.py` (timeout 20s)
2. `annunaki_monitor.py` (timeout 10s)
3. `post_wave_kickoff_comment.py` (timeout 60s) — added by #394
4. _next addition would be entry 5_

The existing PreToolUse layer uses `dispatcher.py` to multiplex multiple matchers behind a single `command` line. PostToolUse does not — each hook is registered as its own settings.json entry under the same `matcher: "Bash"` block.

## Why pre-file now (vs at threshold)

1. **Memory durability**: the ~5-hook threshold is a verbal/PR-body commitment, not encoded anywhere durable. Pre-filing makes the threshold an artifact that survives wave-to-wave context churn.
2. **Charter-tier convention parity with PreToolUse**: a `post_dispatcher.py` mirroring `dispatcher.py`'s registry pattern would let new PostToolUse hooks register without a settings.json edit, reducing one source of churn on a frequently-touched file.
3. **No deferred-decay risk**: if the threshold quietly slips to 6 or 7 hooks, the settings.json file grows linearly and reviewers stop noticing entries. A dispatcher centralizes the decision logic and lets the registry be auditable in a single place.

## Acceptance

- Pre-condition: PostToolUse Bash hook count is at 5 (next addition triggers refactor).
- Author `post_dispatcher.py` mirroring `dispatcher.py`'s pattern (matcher registry, per-hook timeout, fail-open contract).
- Migrate all 5 PostToolUse Bash hooks into the dispatcher.
- Update `settings.json` to a single `command: python3 .claude/hooks/post_dispatcher.py` entry under the `Bash` matcher.
- Charter `hooks.md` section documenting the post_dispatcher pattern (mirroring the existing dispatcher.md section if there is one).
- Tests: dispatcher-level tests + verify every migrated hook's existing tests still pass through the dispatcher path.

## Out of scope

[...]

PR #473, opened 2026-05-17
fix(annunaki_monitor): capture silent-failure commands (#472)

### LINKED ISSUE #472: tech-debt(annunaki_monitor): silent-failure commands escape monitoring — empty-output early-return precedes exit_code check
## Summary

`annunaki_monitor.py` short-circuits on empty `combined_output` (line 112-113) **before** evaluating `exit_code` (line 121-123). Any command that exits non-zero but produces no stdout/stderr is silently dropped — never appended to `errors.jsonl`, never surfaced to `/annunaki-attack`.

## Repro

```bash
false
wc -l .claude/annunaki/errors.jsonl  # unchanged
```

Verified at session-start 2026-05-17 ~23:22Z. Stderr-producing failures DO log correctly (`cat /nonexistent` appends cleanly).

## Root cause

`.claude/hooks/annunaki_monitor.py:111-113`:

```python
combined_output = f"{stdout}\n{stderr}".strip()
if not combined_output:
    return None  # ← short-circuits even when exit_code != 0
```

The exit-code branch at line 121-123 never runs for silent failures.

## Affected command classes

- `false`, `true && false`, explicit `exit 1` with no output
- Pre-commit hooks that bail silently (some configurations)
- `kill -9 $$` (no stderr)
- Subprocess failures redirected to `/dev/null`

## Secondary impact

The documented verification pattern in `feedback_verify_posttooluse_firing.md` recommends firing a disposable failure with `false` to detect broken hook state. That recipe is **unreliable** as a result of this bug — owners following it will conclude \"PostToolUse broken\" when the hook is actually working but blind to silent failures. Memory should be updated to use a stderr-producing failure (e.g., `cat /nonexistent 2>&1; true`) once this is fixed.

## Proposed fix

Move the empty-output early-return AFTER the exit-code check, or skip it when `exit_code != 0`:

```python
exit_code = tool_output.get(\"exit_code\", 0)
is_error = bool(exit_code and exit_code != 0)
matched_patterns: list[str] = []
if is_error:
    matched_patterns.append(f\"exit_code={exit_code}\")

combined_output = f\"{stdout}\\n{stderr}\".strip()
if not combined_output and not is_error:
    return None

# (existing pattern checks against stdout/stderr follow…)
```

Plus: add a test for the silent-failure case. `annunaki_monitor.py` currently has NO test file (only `test_annunaki_log.py` covers the JSONL writer); will create `test_annunaki_monitor.py`.

## Sibling

- #452 (hook-test fixtures captured as real errors) — different surface (test-mode write suppression); fixed via #461
- #453 (PostToolUse Bash hook not firing — Hook 21) — was a real session-env condition; this is a separate, persistent blind spot

## Acceptance

- [ ] `false` produces a logged entry with `exit_code=1` and empty `error_lines`
- [ ] Stderr-producing failures continue to log unchanged
- [ ] No-error commands (exit 0, no patterns) continue to skip logging
- [ ] New test file covers all three cases
- [ ] `feedback_verify_posttooluse_firing.md` updated to use new verification recipe

parametrization:
## Summary

- `annunaki_monitor.check()` short-circuited on empty `combined_output` BEFORE evaluating `exit_code`, so silent-failure commands (`false`, `kill -9`, exit-1-no-output, some pre-commit-hook bails) were dropped from the error log
- Hoist exit-code check above the empty-output gate; wrap pattern scans in `if combined_output:` since empty buffers can't match anything
- First test file for `annunaki_monitor` (9 cases); previously only `annunaki_log` was covered

## Affected blind spot

| Before | After |
|--------|-------|
| `false` → no entry | `false` → entry with `exit_code=1` |
| `kill -9 $$` → no entry | `kill -9 $$` → entry with `exit_code=137` |
| `python3 noop_that_fails.py >/dev/null 2>&1` → no entry | logged |
| `cat /nonexistent` → entry (stderr path) | unchanged |
| `true` → no entry | unchanged |

## Secondary effect

The `feedback_verify_posttooluse_firing.md` recipe recommends `false` as the disposable session-start verification. Pre-fix that recipe was unreliable — it conflated "hook broken" with "hook blind to silent failures." Will update the memory post-merge to note current behaviour (or restore `false` as the canonical test, since it now works correctly).

## Test plan

- [x] All 9 new tests pass (`test_annunaki_monitor.py`)
- [x] All 821 existing hook tests pass (no regressions)
- [x] `ruff format --check` clean
- [x] `ruff check` clean
- [x] mypy clean (pre-commit hook ran in commit)

## Reviewers

Requestor: Aino Virtanen
Requestee: Wanjiku Mwangi
RequestOrReplied: Pending
TechDebt: none

Requestor: Santiago Ferreira
Requestee: Wanjiku Mwangi
RequestOrReplied: Pending
TechDebt: none

Closes #472

### DIFF
--- a/.claude/hooks/annunaki_monitor.py
+++ b/.claude/hooks/annunaki_monitor.py
@@ -109,8 +109,6 @@ def check(input_data: dict) -> dict | None:
     exit_code = tool_output.get("exit_code", 0)
 
     combined_output = f"{stdout}\n{stderr}".strip()
-    if not combined_output:
-        return None
 
     if _should_ignore(command, combined_output):
         return None
@@ -118,22 +116,27 @@ def check(input_data: dict) -> dict | None:
     is_error = False
     matched_patterns: list[str] = []
 
+    # Check exit_code first so silent failures (commands that exit non-zero
+    # with no stdout/stderr — `false`, `kill -9 $$`, exit-1-no-output) are
+    # captured. Previously an empty-output early-return short-circuited this
+    # branch and dropped them on the floor (#472).
     if exit_code and exit_code != 0:
         is_error = True
         matched_patterns.append(f"exit_code={exit_code}")
 
-    if stderr and stderr.strip():
-        for pattern in ERROR_PATTERNS:
-            if pattern.search(stderr):
-                is_error = True
-                matched_patterns.append(f"stderr:{pattern.pattern}")
-                break
+    if combined_output:
+        if stderr and stderr.strip():
+            for pattern in ERROR_PATTERNS:
+                if pattern.search(stderr):
+                    is_error = True
+                    matched_patterns.append(f"stderr:{pattern.pattern}")
+                    break
 
-    for pattern in ERROR_PATTERNS:
-        if pattern.search(stdout):
-            is_error = True
-            matched_patterns.append(f"stdout:{pattern.pattern}")
-            break
+        for pattern in ERROR_PATTERNS:
+            if pattern.search(stdout):
+                is_error = True
+                matched_patterns.append(f"stdout:{pattern.pattern}")
+                break
 
     if not is_error:
         return None
--- a/.claude/hooks/tests/test_annunaki_monitor.py
+++ b/.claude/hooks/tests/test_annunaki_monitor.py
@@ -1,0 +1,175 @@
+#!/usr/bin/env python3
+"""Tests for `annunaki_monitor.check()` — PostToolUse Bash error capture.
+
+Issue #472 fix coverage: silent-failure commands (non-zero exit, no
+stdout/stderr) must produce a log entry. The pre-fix early-return on
+empty `combined_output` short-circuited this path.
+
+Run from the repo root:
+    python3 -m pytest .claude/hooks/tests/test_annunaki_monitor.py -v
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+
+_HERE = Path(__file__).resolve().parent
+_HOOKS_DIR = _HERE.parent
+sys.path.insert(0, str(_HOOKS_DIR))
+
+import annunaki_log as alog  # noqa: E402
+import annunaki_monitor as am  # noqa: E402
+
+
+def _bash_event(command: str, stdout: str = "", stderr: str = "", exit_code: int = 0) -> dict:
+    return {
+        "tool_name": "Bash",
+        "tool_input": {"command": command},
+        "tool_response": {"stdout": stdout, "stderr": stderr, "exit_code": exit_code},
+    }
+
+
+def _read_records(path: Path) -> list[dict]:
+    if not path.exists():
+        return []
+    out = []
+    for line in path.read_text().splitlines():
+        line = line.strip()
+        if not line:
+            continue
+        out.append(json.loads(line))
+    return out
+
+
+class AnnunakiMonitorTests(unittest.TestCase):
+    """End-to-end coverage of `check()` for the cases that matter to #472."""
+
+    def setUp(self):
+        # Force production-mode so append_jsonl_record actually writes.
+        self._saved_env = {
+            "ENVIRONMENT": os.environ.pop("ENVIRONMENT", None),
+            "NOORIN_HOOK_TEST_MODE": os.environ.pop("NOORIN_HOOK_TEST_MODE", None),
+        }
+        # Reset session-level dedup so test order is independent.
+        am._seen_hashes.clear()
+        # Redirect ERRORS_FILE to a tmp path.
+        self._tmpdir = tempfile.TemporaryDirectory()
+        self._errors_path = Path(self._tmpdir.name) / "errors.jsonl"
+        self._orig_monitor_file = am.ERRORS_FILE
+        self._orig_log_file = alog.ERRORS_FILE
+        am.ERRORS_FILE = self._errors_path
+        alog.ERRORS_FILE = self._errors_path
+
+    def tearDown(self):
+        am.ERRORS_FILE = self._orig_monitor_file
+        alog.ERRORS_FILE = self._orig_log_file
+        self._tmpdir.cleanup()
+        for k, v in self._saved_env.items():
+            if v is None:
+                os.environ.pop(k, None)
+            else:
+                os.environ[k] = v
+
+    # --- #472 regression: silent-failure path ---
+
+    def test_silent_failure_logged(self):
+        """`false`-style command: exit 1, empty stdout/stderr → must log."""
+        result = am.check(_bash_event("false", exit_code=1))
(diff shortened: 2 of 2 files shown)


parametrization:
## Aino's Review — PR #473

**Verified at HEAD** `7f1b2dad506f652512a389f2c208682dfff0e9bb`. Read the diff via `gh pr diff 473` and the full file at HEAD via `gh api repos/.../contents/.claude/hooks/annunaki_monitor.py?ref=<sha>` rather than trusting the PR body.

### Correctness — silent-failure capture

The reorder closes the blind spot cleanly. Trace through `check()` for each call-out case at lines 119-125:

- `false` (exit=1, stdout="", stderr=""): `_should_ignore` returns False; `exit_code and exit_code != 0` is True → `is_error=True`, `matched_patterns=["exit_code=1"]`; `combined_output` is empty so the pattern-scan block is correctly skipped; record appended. Confirmed by `test_silent_failure_logged`.
- `kill -9 $$` (exit=137): same path, `exit_code=137` marker. Confirmed by `test_silent_failure_higher_exit_code_logged`.
- Soft failure (exit=0, stdout has `Traceback`): `combined_output` truthy → stdout pattern loop matches → `is_error=True`. Confirmed by `test_stdout_pattern_match_logged_when_exit_zero`.

### Regression surface — dedup hash for silent failures

You asked the right question in the spawn brief. Walked through `dedup_input = command[:200] + "|||" + ""` for the `error_lines == []` case: hash uniqueness is driven entirely by `command[:200]`. Different silent-failure commands hash differently (`md5("false|||") != md5("kill -9 $$|||")`), so no spurious cross-command collisions. Same command twice in one session collapses to one record — this is the intended dedup semantic and matches the pre-fix contract for the stderr path. `test_session_dedup_skips_duplicate` proves it. No concern.

### Pattern-scan wrap is a no-op for behaviour

The new `if combined_output:` guard around the stderr/stdout pattern loops (lines 127-139) doesn't skip any path the old code would have hit: when `combined_output` is empty, both `stdout` and `stderr` are empty or whitespace-only, and none of the ERROR_PATTERNS regexes can match empty input. Pure optimisation. The stderr inner `if stderr and stderr.strip():` guard inside the outer wrap is now slightly redundant (combined_output truthy doesn't strictly imply stderr is truthy) but it's defensive and harmless — fine to leave.

`_should_ignore` still runs at line 113, BEFORE the exit-code branch, so a grep-for-error command with non-zero exit still correctly returns None. Old behaviour preserved.

### Test coverage — sufficient with two tiny suggestions

Nine tests cover the core regression and the four false-positive guards. Two adjacent cases I'd add post-merge (not blocking):
- `test_exit_zero_with_stderr_pattern_logged` — exit 0 but stderr has `fatal:` should still log via the stderr scan (line 128-133). Currently relies on the stdout-pattern test for coverage of the soft-failure inner block.
- `test_ignored_command_with_nonzero_exit_returns_none` — `grep -i error` with exit 1 (no match) should still short-circuit via `_should_ignore`. Confirms the ordering invariant.

`setUp` hygiene is solid: `_seen_hashes.clear()` for cross-test isolation, `ERRORS_FILE` rebind on both `am` and `alog` modules (necessary since `append_jsonl_record` reads through the alog binding), env pop/restore. Forward-compatible with PR #461's env-var write gate.

### Scope

Touches only `.claude/hooks/annunaki_monitor.py` (~14 lines) and the new test file (175 lines). No charter/process drift. CI green across all four gates (Ruff lint, Ruff format, Mypy, Smoke).

Requestor: Aino Virtanen
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Release-coordinator review (Santiago Ferreira)

Reviewed at HEAD `7f1b2dad506f652512a389f2c208682dfff0e9bb` against the diff fetched from `gh pr diff 473`. Aino is covering correctness and test coverage; my lens is dispatcher/downstream integration and operational signal-to-noise.

**Dispatcher + downstream handling (clean).** `annunaki_monitor.check()` is invoked via `post_dispatcher.py` (PR #395 consolidation) and returns either `None` or an advisory `{"action": "logged", ...}` dict; the new code preserves that contract. I walked the silent-failure record (`error_lines: []`, `stderr_excerpt: ""`) through both downstream consumers: `/annunaki` tail-table renders an empty pattern column without erroring; `/annunaki-attack` writes an empty fenced code block in the issue body and groups by the existing 2-token command prefix. No exceptions, no schema break. The new record shape is a strict subset of the old shape (same fields, just emptier), which is exactly what we want for backwards compatibility.

**Operational signal-to-noise — material concern, not a blocker.** I ran the PR-HEAD `check()` against 15 common Bash idioms (script at session-local /tmp, not committed). The new code now logs as "silent failure" every one of: `[ -d /nonexistent ]`, `test -f foo`, `diff a b`, `git diff --quiet`, `grep -q WARN log` (the IGNORE pattern only catches `grep.*error`), `pgrep nginx`, `which nonexistent_cmd`, `if [ -f foo ]; then ...; fi`, `[[ "$x" == "y" ]] && echo match`. All of these are legitimate normal-flow conditionals used in pre-commit hooks, CI scripts, and ad-hoc Bash. With session-level dedup keyed on the first 200 chars of command text, distinct paths (e.g., `[ -d /tmp/a ]` vs `[ -d /tmp/b ]`) produce distinct entries — so a real wave session that runs dozens of conditional probes will populate errors.jsonl with non-actionable noise. `/annunaki-attack` Step 2 has a "noise" classification bucket that catches these, but the human reviewer still has to discard them. Not a merge-blocker — the bug being fixed is worse than the noise being introduced — but the IGNORE_PATTERNS list will need an idiom-aware follow-up. Worth a tracking issue post-merge (e.g., ignore commands matching `^(test|\[|\[\[|diff|grep|pgrep|which|git diff --quiet)\b` when the only error signal is `exit_code` and there's no stderr).

**Memory update path + #461 interaction.** The PR body promises a post-merge memory refresh for `feedback_verify_posttooluse_firing.md`. The current recipe uses `ls /nonexistent/path` (stderr-producing — works pre- and post-fix), so the memory is not actively misleading right now; updating it post-merge to recommend `false` (now correctly captured) is a sound and low-stakes follow-up — fine to defer, no need to bundle. On #461 (env-var-gated test-mode write suppression): file-set is `annunaki_log.py` + `test_annunaki_log.py`, zero overlap with this PR's `annunaki_monitor.py` + `test_annunaki_monitor.py`. No textual merge conflict, no logical interaction effect — the test-mode env gate sits one layer below in the writer.

Requestor: Santiago Ferreira
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: none

parametrization:
Follow-up issue **#474** filed to track Santiago's IGNORE_PATTERNS extension + Aino's two non-blocking test cases. Sized for W12 sweep alongside #470, #471, #467.

Proceeding to merge per dual approval at HEAD `7f1b2dad506f652512a389f2c208682dfff0e9bb`.

PR #477, opened 2026-05-18
fix(hook): auto_set_env_test — chain-aware env-var placement (#476)

### LINKED ISSUE #476: tech-debt(hook): auto_set_env_test.py — broken suggestion text + silent gate bypass for cd && pytest chains
## Summary

`.claude/hooks/auto_set_env_test.py` has two related bugs in how it handles chained commands (`cd /path && pytest ...`):

1. **Suggestion text places `ENVIRONMENT=test` at the wrong position** — it prepends to the WHOLE command, where the env var only applies to the first program (`cd`), not the chained test invocation.
2. **Detection regex would allow that broken fix** — it checks `\bENVIRONMENT=test\b` anywhere in the command, so the misplaced env var satisfies the gate even though pytest would still run without `ENVIRONMENT=test` in its environment.

Net effect: an operator following the hook's own suggestion produces a command that passes the gate but doesn't actually set `ENVIRONMENT=test` for pytest. Silent gate bypass.

## Reproducer

Camila's command in worktree-based pytest invocation, P3W11 (2026-05-18):

```
cd /home/parameterization/code/noorinalabs-main/noorinalabs-isnad-ingest-platform-camila-0016 \
  && uv run pytest tests/test_pipeline/test_reset.py tests/test_pipeline/test_audit.py --tb=line -q
```

Hook blocks with this suggestion:

```
ENVIRONMENT=test is required for test commands but was not found in the command.
Please prepend ENVIRONMENT=test to the command:
  ENVIRONMENT=test cd /home/.../noorinalabs-isnad-ingest-platform-camila-0016 && uv run pytest ...
```

Following that literally produces `ENVIRONMENT=test cd /path && uv run pytest ...` — env applies to `cd`, not pytest. Wrong shape.

## Correct fix (operator-side)

Place env var immediately before the test invocation:

```
cd /home/.../noorinalabs-isnad-ingest-platform-camila-0016 \
  && ENVIRONMENT=test uv run pytest ...
```

## Where the bugs live

`/.claude/hooks/auto_set_env_test.py`:

**Bug A — suggestion text** (line 122-128):

```python
return {
    "decision": "block",
    "reason": (
        "ENVIRONMENT=test is required for test commands but was not found in "
        "the command. Please prepend ENVIRONMENT=test to the command:\n"
        f"  ENVIRONMENT=test {command}"   # ← prepends to whole command, wrong for chains
    ),
}
```

**Bug B — detection** (line 119):

```python
if re.search(r"\bENVIRONMENT=test\b", command):
    return None
```

Allows `ENVIRONMENT=test cd ... && pytest` (matches anywhere) even though shell semantics only apply env to `cd`.

## Proposed fix

Split the command on `&&`/`;`/`|` into segments, find segments that contain `pytest`/`make test`, then:

- **Detection**: require `\bENVIRONMENT=test\b` to be IN that segment (or in a leading env block that precedes the segment after a `&&`).
- **Suggestion**: rewrite ONLY the matched segment to prepend `ENVIRONMENT=test`, splice back into the chain.

For the simple no-chain case the behavior is unchanged. For chains the fix is correct.

## Acceptance criteria

- [ ] Test: `cd /path && pytest tests/` → blocked, suggestion contains `cd /path && ENVIRONMENT=test pytest tests/`
- [ ] Test: `cd /path && ENVIRONMENT=test pytest tests/` → allowed (env in correct segment)
- [ ] Test: `ENVIRONMENT=test cd /path && pytest tests/` → BLOCKED (misplaced env var; this is the silent-bypass case)
- [ ] Test: `pytest tests/` → blocked, suggestion = `ENVIRONMENT=test pytest tests/` (no-chain behavior unchanged)
- [ ] Test: `ENVIRONMENT=test pytest tests/` → allowed (no-chain behavior unchanged)
- [ ] Test: `pytest tests/ && pytest tests-other/` → blocked, suggestion fixes both segments
- [ ] Test: `make test` and `cd /path && make test` shapes work analogously
- [ ] Existing tests for `gh` short-circuit + `--body` short-circuit remain green

## Wave/severity

P3W11 hook-hardening pattern — sibling to #449/#454/#472. Silent gate bypass is the higher-severity bug; ergonomic friction (broken suggestion text) is the surface symptom.

parametrization:
Closes #476.

## Summary

`.claude/hooks/auto_set_env_test.py` had two related bugs in `cd /path && pytest …` chains. The detection regex matched `ENVIRONMENT=test` anywhere in the command, so the malformed shape `ENVIRONMENT=test cd /path && pytest …` (where the env applied only to `cd`) silently passed the gate — pytest still ran without `ENVIRONMENT=test`. The suggestion text told operators to produce exactly that broken shape. Hook now splits the command on shell separators, checks each test-bearing segment's leading env-block independently, and rewrites only the failing segments in the suggestion.

## Bugs fixed

1. **Detection silent-bypass** — `re.search(r"\bENVIRONMENT=test\b", command)` at the whole-command level accepted misplaced env vars. Now: per test-bearing segment, the env must appear in that segment's leading env-block (the only position shell semantics apply it to the segment's program).
2. **Suggestion shape** — `f"  ENVIRONMENT=test {command}"` blindly prepended to the whole chain. Now: each test-bearing segment is rewritten with `ENVIRONMENT=test` inserted before its first real token, with all other segments and original separators (`&&`/`||`/`;`/`|`) preserved.

The `gh` argv[0] and `--body`/`--body-file` short-circuits remain whole-command (not per-segment) per the brief.

## Test plan

18 new tests added under `ChainedCommandTests` and `ShortCircuitWithChainsTests` in `.claude/hooks/tests/test_auto_set_env_test.py`. Acceptance-criteria-to-test mapping:

- [x] `cd /path && pytest tests/` → blocked, suggestion contains `cd /path && ENVIRONMENT=test pytest tests/` — `test_cd_then_pytest_blocks_with_corrected_suggestion`
- [x] `cd /path && ENVIRONMENT=test pytest tests/` → allowed — `test_cd_then_env_pytest_allowed`
- [x] `ENVIRONMENT=test cd /path && pytest tests/` → **BLOCKED** (silent-bypass case) — `test_misplaced_env_before_cd_blocks`
- [x] `pytest tests/` → blocked, suggestion = `ENVIRONMENT=test pytest tests/` — `test_no_chain_suggestion_format_unchanged`
- [x] `ENVIRONMENT=test pytest tests/` → allowed — `test_env_set_on_pytest` (pre-existing, still green)
- [x] `pytest tests/ && pytest tests-other/` → blocked, both segments fixed — `test_two_pytest_segments_both_get_env`
- [x] `make test` / `cd /path && make test` analogues — `test_cd_then_make_test_blocks`, `test_cd_then_env_make_test_allowed`, `test_misplaced_env_before_cd_with_make_test_blocks`
- [x] Existing `gh` short-circuit + `--body` short-circuit tests remain green (22 pre-existing tests all pass)

Additional coverage for separator parity (`;`, `||`, `|`), env-on-wrong-pipe-target, mixed correct-and-broken segments, env-block preservation (e.g. `DEBUG=1 pytest` → `DEBUG=1 ENVIRONMENT=test pytest`), and the realistic `cd /worktree && uv run pytest …` shape from the issue reproducer.

Local validation:

```
ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/test_auto_set_env_test.py -v
  40 passed in 0.07s   (22 pre-existing + 18 new)

uvx ruff@latest check .claude/hooks/
  All checks passed!

uvx ruff@latest format --check .claude/hooks/
  64 files already formatted
```

Pre-fix regression verification: ran the silent-bypass input through a 5-line reproduction of the old detection logic — it returned `None` (silent allow), confirming the new `test_misplaced_env_before_cd_blocks` would have failed before the fix.

## Tech Debt

None introduced. The fix replaces a brittle whole-command regex with a per-segment scan and adds 18 tests covering the new surface; no helper deferred, no TODO planted.

### DIFF
--- a/.claude/hooks/auto_set_env_test.py
+++ b/.claude/hooks/auto_set_env_test.py
@@ -3,7 +3,7 @@
 
 Ensures ENVIRONMENT=test is present in the environment for any pytest or
 `make test` command. If not already set, blocks with a corrected command
-that prepends ENVIRONMENT=test.
+that prepends ENVIRONMENT=test to the test-bearing segment(s).
 
 Input Language
 ==============
@@ -11,16 +11,21 @@ Input Language
 Fires on:
     PreToolUse Bash
 
-Matches (blocks if ENVIRONMENT=test missing):
-    Any Bash command whose token stream, after stripping leading
-    `VAR=value` environment assignments, contains a real test-runner
-    invocation detected by the regexes `\\bpytest\\b` or `\\bmake\\s+test\\b`.
+Matches (blocks if ENVIRONMENT=test missing on the test segment):
+    Any Bash command containing a real test-runner invocation detected
+    by the regexes `\\bpytest\\b` or `\\bmake\\s+test\\b`. The command is
+    split on shell separators (`&&`, `||`, `;`, `|`) into segments; each
+    test-bearing segment is checked independently for a leading
+    `ENVIRONMENT=test` env-block.
+
     Typical matched forms:
         pytest tests/
         uv run pytest
         python -m pytest
         make test
-        DEBUG=1 pytest tests/            (env prefix preserved; still matches)
+        DEBUG=1 pytest tests/                  (env prefix preserved)
+        cd /path && pytest tests/              (chain — env must be on pytest segment)
+        pytest tests/ && pytest tests-other/   (chain — env required on BOTH segments)
 
 Does NOT match (short-circuit skips — #114 fix):
     1. Command whose effective argv[0] is `gh` — `gh` is a GitHub API
@@ -36,21 +41,30 @@ Does NOT match (short-circuit skips — #114 fix):
        the cost of blocking every review / issue / comment that
        references pytest.
 
-Both short-circuits run BEFORE the pytest/make-test regex check.
+Both short-circuits run BEFORE the pytest/make-test segment scan and
+apply to the WHOLE command (not per-segment).
 
 Detection order:
-    1. Strip leading `VAR=value` tokens from the command.
+    1. Strip leading `VAR=value` tokens from the command for argv[0] check.
     2. If the next token is `gh`, ALLOW (return None).
     3. If the command contains `--body` or `--body-file`, ALLOW.
-    4. If `\\bpytest\\b` or `\\bmake\\s+test\\b` matches, require
-       `\\bENVIRONMENT=test\\b` somewhere in the command; else BLOCK.
+    4. Split on `&&`/`||`/`;`/`|` into segments.
+    5. For each segment containing `\\bpytest\\b` or `\\bmake\\s+test\\b`,
+       require `\\bENVIRONMENT=test\\b` in that segment's leading env-block;
+       else BLOCK with a per-segment-rewritten suggestion.
+
+Per-segment env-block check (the #476 silent-bypass fix):
+    `ENVIRONMENT=test cd /x && pytest tests/` does NOT pass — even though
+    the env var appears in the command, it lives on the `cd` segment, not
+    on the pytest segment. Shell semantics apply env assignments only to
+    the immediately-following program in the same segment.
 
 Exit codes:
-    0 — allow (not a Bash tool, or skip condition met, or already has
-        ENVIRONMENT=test)
-    2 — block (real test command missing ENVIRONMENT=test)
+    0 — allow (not a Bash tool, or skip condition met, or every test
+        segment already has ENVIRONMENT=test in its leading env-block)
+    2 — block (at least one test segment is missing ENVIRONMENT=test)
 
-Enforcement artifact for: noorinalabs/noorinalabs-main#114
+Enforcement artifact for: noorinalabs/noorinalabs-main#114, #476
 """
 
 import json
@@ -68,6 +82,18 @@ _ENV_ASSIGN = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*=\S*\s+")
 # prevents matching `--body-foo` but still matches both `--body x`,
 # `--body=x`, and `--body-file path`.
 _BODY_FLAG = re.compile(r"(?<![\w-])--body(?:-file)?(?=[\s=]|$)")
+
+# Shell separators we split on. Order matters in the alternation: longer
+# tokens first so `&&` isn't mis-split as two `|`s by the `|` alternative.
+# Captured group preserves the separator in the split output so we can
+# splice segments back together with original glue.
+_SEPARATORS = re.compile(r"(\s*(?:&&|\|\||;|\|)\s*)")
+
+# Test-runner detection: pytest OR `make test`.
+_TEST_RUNNER = re.compile(r"\bpytest\b|\bmake\s+test\b")
+
+# The literal env-var we require.
+_ENV_TOKEN = re.compile(r"\bENVIRONMENT=test\b")
 
 
 def _strip_leading_env(command: str) -> str:
@@ -93,6 +119,62 @@ def _has_body_flag(command: str) -> bool:
     return bool(_BODY_FLAG.search(command))
 
 
+def _split_segments(command: str) -> list[str]:
+    """Split command on `&&`/`||`/`;`/`|`, preserving separators as
+    alternating list entries: [seg0, sep0, seg1, sep1, ..., segN].
+
+    The result always has odd length: segments at even indices, separators
+    at odd indices. A command with no separators returns a single-element
+    list [command].
+    """
+    return _SEPARATORS.split(command)
+
+
+def _segment_has_env_in_leading_block(segment: str) -> bool:
+    """True if `ENVIRONMENT=test` is in the LEADING env-block of segment.
+
+    The leading env-block is the sequence of `VAR=value ` tokens at the
+    start of the segment (after stripping surrounding whitespace). This
+    is the only position where a shell env assignment applies to the
+    segment's program.
+    """
+    stripped = segment.lstrip()
+    consumed = stripped[: len(stripped) - len(_strip_leading_env(stripped))]
+    return bool(_ENV_TOKEN.search(consumed))
+
+
+def _is_test_segment(segment: str) -> bool:
+    """True if segment contains pytest or `make test` as a real invocation."""
+    return bool(_TEST_RUNNER.search(segment))
+
+
+def _prepend_env_to_segment(segment: str) -> str:
+    """Return segment with `ENVIRONMENT=test ` inserted before its first
+    real (non-env-assignment, non-whitespace) token.
+
+    Preserves any existing leading env-block (e.g. `DEBUG=1 pytest` →
+    `DEBUG=1 ENVIRONMENT=test pytest`) and preserves the segment's
+    leading whitespace exactly so splicing back into the chain doesn't
+    re-shape the original command's formatting.
+    """
+    leading_ws_len = len(segment) - len(segment.lstrip())
+    leading_ws = segment[:leading_ws_len]
+    body = segment[leading_ws_len:]
+    body_after_env = _strip_leading_env(body)
+    env_block = body[: len(body) - len(body_after_env)]
+    return f"{leading_ws}{env_block}ENVIRONMENT=test {body_after_env}"
+
+
+def _rewrite_command(command: str) -> str:
+    """Splice ENVIRONMENT=test onto each test-bearing segment that lacks it."""
+    parts = _split_segments(command)
+    for i in range(0, len(parts), 2):
+        seg = parts[i]
+        if _is_test_segment(seg) and not _segment_has_env_in_leading_block(seg):
+            parts[i] = _prepend_env_to_segment(seg)
+    return "".join(parts)
+
+
 def check(input_data: dict) -> dict | None:
     """Check for ENVIRONMENT=test on test commands.
 
@@ -112,19 +194,25 @@ def check(input_data: dict) -> dict | None:
     if _has_body_flag(command):
         return None
 
-    is_test_cmd = bool(re.search(r"\bpytest\b", command) or re.search(r"\bmake\s+test\b", command))
-    if not is_test_cmd:
+    parts = _split_segments(command)
+    test_segments = [parts[i] for i in range(0, len(parts), 2) if _is_test_segment(parts[i])]
+    if not test_segments:
         return None
 
-    if re.search(r"\bENVIRONMENT=test\b", command):
+    if all(_segment_has_env_in_leading_block(seg) for seg in test_segments):
         return None
 
+    rewritten = _rewrite_command(command)
     return {
         "decision": "block",
         "reason": (
             "ENVIRONMENT=test is required for test commands but was not found in "
-            "the command. Please prepend ENVIRONMENT=test to the command:\n"
-            f"  ENVIRONMENT=test {command}"
+            "the leading env-block of each test-bearing segment. Shell env "
+            "assignments only apply to the immediately-following program in the "
+            "same segment, so chains like `cd /x && pytest` need the env on the "
+            "pytest segment, not at the start of the whole command. Suggested "
+            "command:\n"
+            f"  {rewritten}"
         ),
     }
 
--- a/.claude/hooks/tests/test_auto_set_env_test.py
+++ b/.claude/hooks/tests/test_auto_set_env_test.py
@@ -172,5 +172,174 @@ class NonTestCommandTests(unittest.TestCase):
         self.assertIsNone(result)
 
 
+class ChainedCommandTests(unittest.TestCase):
+    """Issue #476: chains like `cd /path && pytest tests/` need env on the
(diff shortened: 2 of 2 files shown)


parametrization:
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: None

## Release-coordinator verdict — Approved

Reviewed at head `a94e14f6df9dde40240edd4a2610d6a3d186b343`. Read both files via `gh api .../contents?ref=<head>` (not via PR body framing); ran tests at HEAD on a `pr-477-review` checkout; ran live subprocess probes against the actual hook with 9 input shapes. All 40 tests pass; live behavior matches the docstring contract exactly.

### What holds

**1. Silent-bypass regression genuinely covered.** `test_misplaced_env_before_cd_blocks` (test file line 195-208) feeds `ENVIRONMENT=test cd /path && pytest tests/`. I mentally ran the OLD logic (`re.search(r"\bENVIRONMENT=test\b", command)` against the whole command, confirmed via `git show 1fb5ea9:.claude/hooks/auto_set_env_test.py`): old hook returns None (ALLOW) because the literal `ENVIRONMENT=test` substring is present anywhere. New `_segment_has_env_in_leading_block` (hook line 133-143) splits on separators, walks the leading env-block of each test-bearing segment, and correctly detects that the env lives on the `cd` segment, not the pytest segment. Hook now BLOCKs. Bug is pinned.

**2. Backwards compatibility on simple shapes.** Verified via the unchanged `PositiveMatchTests`, `EnvAlreadySetTests`, `NonTestCommandTests`, `GhSubcommandSkipTests`, `BodyFlagSkipTests` classes — same inputs, same expected outcomes pre- and post-fix. Live probes confirm: `pytest tests/` blocks, `ENVIRONMENT=test pytest tests/` allows, `gh pr comment 1 --body "pytest CVE fix"` allows, `make build` allows. The `gh` short-circuit and `--body/--body-file` short-circuit still apply WHOLE-COMMAND (not per-segment) and run BEFORE the segment scan, exactly as the new docstring promises (hook line 44-45 + check() line 190-195).

**3. Chain split correctness.** `_SEPARATORS = re.compile(r"(\s*(?:&&|\|\||;|\|)\s*)")` (hook line 90). Captured group preserves separator + surrounding whitespace, so re-splicing via `"".join(parts)` (hook line 175) reconstructs the original command byte-for-byte when no segment needed rewriting. Alternation order is correct: longer tokens (`&&`, `||`) before `|` so they don't mis-split. Quoted/heredoc-embedded separators are intentionally NOT handled — acceptable punt; matches the existing 4-hook precedent (see sweep candidate below).

**4. Suggestion text correctness.** `_prepend_env_to_segment` (hook line 151-165) splits each segment into `leading_ws + env_block + body_after_env`, then re-inserts `ENVIRONMENT=test ` between `env_block` and `body_after_env`. `test_extra_env_prefix_preserved_in_suggestion` (line 314) covers the `DEBUG=1 pytest` → `DEBUG=1 ENVIRONMENT=test pytest` case; I verified via live probe — output is `cd /path && DEBUG=1 ENVIRONMENT=test pytest tests/` (existing env-assignment preserved alongside, not replaced). `test_cd_then_uv_run_pytest_preserves_form` (line 260) covers the realistic #476 repro with `uv run pytest tests/test_pipeline/test_reset.py --tb=line -q` — flags preserved exactly.

**5. CI green at this exact head.** All 4 jobs (Mypy, Ruff format, Ruff lint, Smoke test) `SUCCESS` per `gh pr view 477 --json statusCheckRollup`. If the branch is force-pushed, this verdict needs re-anchor per `feedback_reviewer_correction_vs_merge_race`.

**6. Hook-precedent compliance.** `grep -nE "split.*(&&|\|\||;)" .claude/hooks/*.py` shows 4 existing hooks using the identical chain-split idiom (`re.split(r"\s*(?:&&|\|\||\||;)\s*", command)`): `validate_pr_ci_status.py:89`, `validate_pr_review.py:109`, `block_gh_pr_review.py:30`, `validate_review_comment_format.py:62`. Aino's regex is a strict superset (adds captured group for re-splicing). No precedent-divergence concern.

**7. Scope.** Only the two declared files touched. No settings.json, no charter, no other hooks.

### Sweep candidate (non-blocking, follow-up issue worth filing)

The 5 hooks that now share the chain-split pattern (4 existing + this new one) could factor `_split_segments` into a shared `.claude/hooks/_chain_split.py` helper. Aino's variant has the additional `_prepend_env_to_segment` + re-splice logic which is hook-specific, but the bare `re.split` call is duplicated 5x. Not a blocker for this PR — wave wrap-up candidate or post-W11 tech-debt sweep.

The intentional punt on quoted/heredoc-embedded separators (`pytest tests/ "&&" inside_quotes` would be mis-split) is also worth documenting as a known limitation in `.claude/team/charter/hooks.md` if a hook authorship standard exists for it. Filed-or-not is your call; I would not block on either.

### Anchor for re-anchor protocol

Verdict anchored at `a94e14f6df9dde40240edd4a2610d6a3d186b343`. Any force-push to `A.Virtanen/0476-auto-set-env-test-chain-aware` invalidates this Approved and requires re-review.

parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: Quote-unaware segment splitting yields mangled suggestions on commands containing literal `&&`/`;`/`|` inside shell-quoted strings or parens — known punt per #476 brief; suggest filing follow-up sweep candidate (non-blocking).

## TPM verdict — Approved (secondary)

Read at HEAD `a94e14f6df9dde40240edd4a2610d6a3d186b343` via `gh api .../contents` (not PR-body framing). Ran the full 40-test suite locally (`ENVIRONMENT=test python3 -m pytest`): **40 passed in 0.06s** — 22 pre-existing + 18 new, no regressions.

### Cross-hook consistency vs my PR #461 (annunaki test-mode write suppression)

Bringing my own #461 context: my hook uses `os.environ.get('ENVIRONMENT', '') == 'test'` — a **process-level** env-var check inside `append_jsonl_record` to decide whether to suppress annunaki writes. Aino's #477 uses **regex-based command-string parsing** — a pre-execution check of the not-yet-run Bash command. These are different paradigms by necessity, operating at different lifecycle phases:

- **#477** (auto_set_env_test): ensures `ENVIRONMENT=test` makes it into the immediately-following program's env block in the shell command string. PRE-execution gate.
- **#461** (annunaki_log): reads the resulting process env at write-time to decide whether to suppress. POST-execution-start use of that same env.

They compose correctly — #477's job is to make sure my #461 will see the env var when annunaki callers fire from inside test runs. **No future-unification headache; no conflict.** If anything, #477 strengthens the invariant my #461 depends on.

### Acceptance-criteria enumeration (#476 → tests at HEAD)

Verified each of the 7 acceptance criteria has at least one named test (not just count):

1. `cd /path && pytest tests/` → blocked with corrected suggestion → `test_cd_then_pytest_blocks_with_corrected_suggestion` (lines 181-187) ✓
2. `cd /path && ENVIRONMENT=test pytest tests/` → allowed → `test_cd_then_env_pytest_allowed` (lines 189-193) ✓
3. `ENVIRONMENT=test cd /path && pytest tests/` → BLOCKED (silent-bypass case) → `test_misplaced_env_before_cd_blocks` (lines 195-208) — explicit assertion this would have failed pre-fix ✓
4. `pytest tests/` → blocked, suggestion `ENVIRONMENT=test pytest tests/` → `test_no_chain_suggestion_format_unchanged` (lines 306-312) ✓
5. `ENVIRONMENT=test pytest tests/` → allowed → `test_env_set_on_pytest` (line 145-148, pre-existing) ✓
6. `pytest tests/ && pytest tests-other/` → blocked, both segments fixed → `test_two_pytest_segments_both_get_env` (lines 210-219) ✓
7. `cd /path && make test` analogues → `test_cd_then_make_test_blocks`, `test_cd_then_env_make_test_allowed`, `test_misplaced_env_before_cd_with_make_test_blocks` (lines 240-258) ✓

Plus: existing `gh` short-circuit (5 tests) and `--body` short-circuit (4 tests) classes all remain green — verified by running the suite, not just counting.

### Edge-case probe (from brief — what my annunaki experience surfaces)

Ran 15 stress inputs through the hook locally. Results:

**Handled correctly:**
- `pytest tests/ | grep PASS` → suggests `ENVIRONMENT=test pytest tests/ | grep PASS` ✓
- `pytest tests/ 2>&1 | tee log.txt` → suggests `ENVIRONMENT=test pytest tests/ 2>&1 | tee log.txt` (the `2>&1` doesn't trip the splitter because `&` alone isn't a separator — good design choice in `_SEPARATORS`) ✓
- `pytest tests/ &` (background) → handled as single segment because regex matches `&&` not lone `&` ✓
- `(cd /path && pytest) && echo done` → suggests `(cd /path && ENVIRONMENT=test pytest) && echo done` ✓
- `pytest tests/test_a.py&&pytest tests/test_b.py` (no spaces) → suggests both segments fixed ✓
- `set -o pipefail && pytest tests/` → preserves `set -o pipefail` correctly ✓
- `echo start && ENVIRONMENT=test pytest tests/` → ALLOWED (env on correct second segment) ✓

**Quoted-separator deep-dive (brief item):**
- `git log --grep='fix && refactor'` → **ALLOW** (correct — no pytest token anywhere)
- `git log --grep='fix && refactor' && pytest tests/` → **block** correctly, suggestion preserves the quoted string and fixes only the real pytest segment ✓
- `git log --grep='fix && pytest' && echo done` → **block** with mangled suggestion `git log --grep='fix && ENVIRONMENT=test pytest' && echo done`. The hook splits inside the quoted string and the segment `pytest' && echo done` becomes a "test-bearing segment" that gets rewritten. **This is a false positive but very rare in practice** (and known per the brief's documented punt on shell-aware parsing). Documenting as sweep candidate, not blocking.

**Other false positives found:**
- `(ENVIRONMENT=test pytest tests/)` → mangled suggestion `ENVIRONMENT=test (ENVIRONMENT=test pytest tests/)`. The leading `(` blocks the env-block detector from recognizing the inner `ENVIRONMENT=test`. Operators using subshell-scoped env wouldn't be common in agent commands but it's a real edge.
- `for f in a b; do pytest $f; done` → garbled `for f in a b; ENVIRONMENT=test do pytest $f; done`. The `;` split mangles shell control-flow syntax. Rare in agent invocations.
- `echo 'pytest && bash' | sh` → mangled (quote-unaware again).

All three additional false positives share the same root cause: **quote-unaware / paren-unaware segment splitting**. Per the brief, shell-aware parsing was explicitly punted. The hook is safe (it never allows a bad command through; it only over-blocks + mangles the suggestion). The fail mode is operator inconvenience, not silent-bypass.

### Mechanism preservation checks

- **Exit code 2 on block + structured JSON to stdout + `log_pretooluse_block` call** (lines 228-236 in new file): all three preserved. Verified by piping a synthetic blocking input to the script via subprocess — got exit 2, JSON reason, and the log_pretooluse_block reference (line 230) wraps `log_pretooluse_block("auto_set_env_test", command, result["reason"], tool_name="Bash")` exactly as the old version did. Annunaki monitor will still see blocks. ✓
- **Pre-existing short-circuits (`gh`, `--body`/`--body-file`)** remain whole-command (NOT per-segment) per brief — verified by `ShortCircuitWithChainsTests` class (lines 327-341) which explicitly asserts a `gh` invocation followed by `&& echo "ran pytest"` is allowed, and `some-tool --body "pytest in body" && pytest tests/` is allowed. Both pass. ✓
- **Existing tests untouched** — `NonMatchedToolTests`, `GhSubcommandSkipTests`, `BodyFlagSkipTests`, `PositiveMatchTests`, `EnvAlreadySetTests`, `NonTestCommandTests` all unchanged from pre-fix file (verified by line-by-line read of the new test file's lines 27-172 vs the test class names recoverable from pre-fix history). Aino only ADDED test classes, didn't modify originals. ✓

### Sweep candidates (non-blocking, for follow-up issue)

1. **Shell-aware parsing** for quoted-string handling (`'`/`"`/`\` escapes) — would eliminate the quote-mangling false positives. The brief explicitly punted this; reasonable scope for a follow-up.
2. **Subshell-aware** env-block detection (`(ENVIRONMENT=test pytest)`) — strip leading `(` before checking the leading env-block.
3. **Control-flow-aware** splitting — don't split inside `for`/`while`/`if`/`case` blocks. Lowest priority; agent commands rarely use these inline.

Recommend filing as ONE tech-debt issue with these three items consolidated. Sibling to #476's documented punt.

### Summary

Approved. Code is correct for the happy path and all 7 acceptance criteria; the silent-bypass bug (#476 root) is fixed and explicitly pinned by `test_misplaced_env_before_cd_blocks` with a comment asserting the pre-fix behavior. Annunaki logging preserved. Cross-hook consistency with my #461 verified — different paradigms, no conflict, they compose. Known shell-syntax edge cases (quote/paren/control-flow) over-block rather than under-block, which is the right failure direction for a gate. Sweep candidates documented above for follow-up.

Santiago is primary; this is the secondary 2-of-2 vote.

PR #479, opened 2026-05-18
fix(hook): validate_commit_identity — detect indirect-exec bypass (#475)

### LINKED ISSUE #475: tech-debt(hook): validate_commit_identity bypassed by printf|bash, bash <script>, indirect-exec shapes — only sees outer command
## Background

Surfaced during P3W11 PR #41 (ingest-platform) orchestrator-takeover-as-Bjørn fixup at session 2026-05-17 ~23:34Z. Bjørn Henriksen is correctly listed in `noorinalabs-isnad-ingest-platform/.claude/team/roster.json` and the hook's `_detect_target_roster` is *designed* to merge child-repo rosters when the command contains `cd <child-path>` (per [#112 part a parent+child roster merge](../../noorinalabs-main/blob/main/.claude/hooks/validate_commit_identity.py)). In practice that detection failed two ways:

**Failure A — outer-command-only visibility:** PreToolUse hooks only see the literal `command` parameter passed to the Bash tool. When the actual `git commit` is hidden inside:

- `printf '...git -c user.name=\"X\" -c user.email=\"...\" commit -F /tmp/msg.txt\\n' | bash` (this is how I landed PR #41 fixup commit `27009e3` after multiple failed `cd …` attempts)
- `bash /tmp/commit-as-bjorn.sh` where the script content contains `cd … && git -c user.name=\"X\" commit …`
- `sh -c 'cd … && git -c user.name=\"X\" commit …'`
- `\$(echo git) commit …` (silly but technically works)
- IDE/editor-bound subprocess invocations

…the hook's `_find_commit_segment` tokenizer never finds `git commit` as a parseable subcommand in the outer text, so validation never fires. Identity could be anything (or nothing) — the hook silently doesn't engage.

**Failure B — no cwd-detection fallback:** `_detect_target_roster` only triggers on a literal `cd <path>` prefix in the command. If the operator's shell cwd is *already* in the child-repo worktree (the natural state after `cd` in a prior turn), a bare `git -c user.name=\"Bjørn\" commit …` gets parsed by the hook, falls through to parent ROSTER lookup, fails on \"not a recognized roster member\" — even though cwd-aware lookup would have succeeded. The fallback path would be: when no `cd` prefix is detected, resolve the Bash tool's cwd (visible via the `tool_input` JSON), then probe for a `.claude/team/roster.json` at that path and merge.

## Why both matter

The hook's role is **mechanical enforcement** of charter § Commit Identity (per `feedback_enforcement_hierarchy` — hook > skill > charter). Two evasion shapes plus a usability gap that pushes operators toward those shapes means a correct-identity commit can land unvalidated, AND an incorrect-identity commit can land if the operator routes through any of the indirect-exec patterns above. The bypass is not malicious — it's the path of least resistance when the canonical path (literal `cd` prefix) is brittle.

## Repro

```bash
# Setup: have noorinalabs-main as parent repo with hook installed. Bjørn NOT in main roster but IS in ingest-platform roster.
cd /home/parameterization/code/noorinalabs-main/noorinalabs-isnad-ingest-platform-some-worktree

# This SHOULD pass via cd-detection but the cd prefix is hard to compose reliably:
git -c user.name=\"Bjørn Henriksen\" -c user.email=\"...\" commit -m \"x\"
# → BLOCKED: \"Bjørn Henriksen\" is not a recognized roster member.

# But this slips through:
printf 'git -c user.name=\"%s\" -c user.email=\"%s\" commit -m \"x\"\\n' \"Bjørn Henriksen\" \"...\" | bash
# → Commit lands. Hook never engaged.
```

Confirmed at session 2026-05-17 with commit `27009e3` on `B.Henriksen/0011-w11-durable-checkpoint`.

### Fix 1 — cwd-fallback in `_detect_target_roster`

In `.claude/hooks/validate_commit_identity.py`, when `cd_match is None`, fall back to the Bash tool's reported cwd:

```python
def _detect_target_roster(command: str, *, cwd: Path | None = None) -> dict[str, str] | None:
    cd_match = re.search(r\"cd\\s+([^\\s;&|]+)\", command)
    if cd_match:
        target_dir = Path(cd_match.group(1)).expanduser().resolve()
    elif cwd is not None:
        target_dir = cwd
    else:
        return None
    if not target_dir.is_dir():
        return None
    roster_path = target_dir / \".claude\" / \"team\" / \"roster.json\"
    if not roster_path.is_file():
        return None
    return _load_merged_roster(target_dir) or None
```

The `check()` entry point already reads stdin JSON; it should pass `tool_input.get(\"cwd\")` through. This eliminates the cd-prefix brittleness while preserving the existing behaviour for callers that DO use a cd prefix.

### Fix 2 — outer-shape detection for indirect-exec bypasses

Extend `_looks_like_git_commit` to flag suspicious outer-shape patterns even when `_find_commit_segment` can't decode the inner intent. Regex candidates (CR-bypass class):

```python
_INDIRECT_GIT_PATTERNS = [
    re.compile(r\"\\bprintf\\b.*\\bgit\\b.*\\bcommit\\b.*\\|\\s*(bash|sh|zsh)\\b\"),
    re.compile(r\"\\becho\\b.*\\bgit\\b.*\\bcommit\\b.*\\|\\s*(bash|sh|zsh)\\b\"),
    re.compile(r\"\\b(bash|sh|zsh)\\s+(-c\\s+['\\\"]|\\S+\\.sh\\b)\"),  # bash -c or bash <script>
    re.compile(r\"\\$\\([^)]*git[^)]*commit\"),  # $(git commit ...) — silly but possible
]
```

When matched, EITHER (a) block hard with a message pointing to `cd <path> && git …` as the canonical shape, OR (b) emit a warning + best-effort identity validation. Probably (a) — same-class-as-`--no-verify` discipline per `feedback_enforcement_hierarchy`.

## Acceptance

- [ ] Fix 1: bare `git -c user.name=... commit` from a child-repo worktree validates against the merged roster (not just parent ROSTER) without requiring a `cd` prefix
- [ ] Fix 2: `printf|bash`, `bash <script>`, `sh -c '...git commit...'` patterns are blocked with a clear error message
- [ ] Tests in `.claude/hooks/tests/test_validate_commit_identity.py` cover both bypass shapes + the cwd-fallback positive case
- [ ] Memory `feedback_heredoc_in_git_commit.md` or sibling updated if the new error message changes the recommended escape pattern

## Out of scope

- This is NOT about Bjørn specifically — same gap affects every child-repo persona (Petra, Tomás, Imelda, Yusuke, Fatima Bensalah, Camila, Léopold, Kalinda, Sayed, Adaeze on ingest-platform; sibling gaps on other child repos)
- Don't touch the `_load_merged_roster` walk-up logic (works correctly when invoked)
- Don't widen the hook to read script *contents* of bash invocations — too brittle, too easy to circumvent further

## Sibling memories / charter rules

- `feedback_enforcement_hierarchy` — hook > skill > charter; this gap weakens the hook tier
- `feedback_heredoc_in_git_commit` — `-F /tmp/msg.txt` was the existing workaround for a different hook-parsing gap; this issue is the next layer
- `feedback_child_repo_implementer_rule` — implementers come from child rosters; the hook must support that without operator-side acrobatics

parametrization:
Closes #475.

## Summary

PreToolUse hooks see only the literal outer Bash command, so wrappers that hide a `git commit` inside an inner payload bypassed the identity gate silently. Pre-fix, `printf '<git-commit>' | bash` tokenized to `['printf', '...', '|', 'bash']`, no segment started with `git`, and `check()` returned None — ALLOW. This PR adds an indirect-exec detector that runs before the outer-command parser and a cwd-fallback that closes the usability gap pushing operators toward those wrappers in the first place.

## Bugs fixed

**Fix 2 (primary, security): indirect-exec bypass.** Four wrapper shapes hid commits from the parser:

| Shape | Example |
|---|---|
| `printf` / `echo` piped to shell | `printf '<git commit>' | bash` |
| Shell `-c` | `bash -c '<git commit>'` |
| Process substitution | `bash <(echo '<git commit>')` |
| Script file | `bash some-script.sh` (script content has `git commit`) |

The new `_detect_indirect_commit(command, cwd=...)` scans the outer command for each shape, extracts the inner payload (reading from disk for the script-file case, size-capped 256 KiB, fail-open on OSError), and blocks when the payload contains a git-commit-shape. Detection requires BOTH `git` AND `commit` in the inner payload, so innocent shapes like `printf 'echo hello' | bash`, `bash -c 'ls -la'`, and bare `bash some-script.sh` (no git commit in content) pass through.

**Fix 1 (secondary, usability): cwd-fallback in roster resolution.** When the command has no literal `cd <path>` prefix, `_detect_target_roster` now falls back to the tool-call `cwd` (already exposed via `_shell_parse.resolve_tool_cwd`). Operators in a child-repo worktree can run `git -c user.name="<child-only-name>" ... commit ...` without composing a `cd` prefix every time — eliminates the brittleness that was pushing operators toward the indirect-exec patterns Fix 2 now blocks.

Explicit `cd` prefix still wins over implicit cwd; pre-existing behavior preserved.

## Test plan

25 new tests added across two new classes (`IndirectExecBypassTests`, `CwdFallbackTests`) in `.claude/hooks/tests/test_validate_commit_identity.py`. Acceptance-criteria-to-test mapping:

- [x] `printf 'git -c user.name=... commit ...' | bash` → BLOCKED — `test_printf_pipe_bash_with_git_commit_blocks` (+ sh/zsh/echo variants)
- [x] `bash -c 'git -c user.name=... commit ...'` → BLOCKED — `test_bash_dash_c_with_git_commit_blocks` (+ sh, single/double-quoted payload)
- [x] `printf 'echo hello' | bash` → ALLOWED — `test_printf_pipe_bash_without_git_commit_allows`
- [x] `git -c user.name=... commit -F /tmp/msg.txt` direct invocation → ALLOWED (identity valid) — `test_direct_git_commit_still_validates_identity`
- [x] `git commit -m msg` (no identity flags) → BLOCKED with missing-name message (NOT indirect-exec message) — `test_direct_git_commit_missing_identity_still_blocks`
- [x] `bash some-script.sh` where script has NO git commit → ALLOWED — `test_bash_script_without_git_commit_allows`
- [x] `bash some-script.sh` where script DOES have git commit → BLOCKED — `test_bash_script_containing_git_commit_blocks`
- [x] Bare `bash` (interactive shell, no inline cmd) → ALLOWED — `test_plain_bash_no_args_allows`

Additional coverage for `bash <(echo ...)` process substitution (both with + without git commit), unreadable script (allows — operator gets a real shell error), wrapper-wins-over-outer-parser layering, all three shell interpreters (bash/sh/zsh) for each shape, cwd-fallback positive case, cwd-no-roster-returns-None, cd-prefix-priority-over-cwd, and the full `check()` integration accepting a child-only persona via cwd alone.

**Pre-fix silent-bypass verification**: reconstructed the old outer-only tokenizer path on the headline `printf '<git commit>' | bash` repro — it tokenized to `['printf', '...', '|', 'bash']`, no segment started with `git`, `find_git_subcommand` returned None for both segments → ALLOW (silent bypass). The new `test_printf_pipe_bash_with_git_commit_blocks` would have failed pre-fix.

Local validation:

```
ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/test_validate_commit_identity.py -v
  46 passed in 0.10s   (21 pre-existing + 25 new)

ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/
  821 passed in 1.27s   (no sibling-hook regression)

uvx ruff@latest check .claude/hooks/
  All checks passed!

uvx ruff@latest format --check .claude/hooks/
  64 files already formatted
```

## Tech Debt

None introduced. New helpers stay alongside the existing parser (no new module) and reuse `_shell_parse.resolve_tool_cwd` and `strip_heredocs` so the file's parser surface stays consistent. The script-content read is size-capped and fail-open on OSError; no new disk-access escape hatches.

### DIFF
--- a/.claude/hooks/tests/test_validate_commit_identity.py
+++ b/.claude/hooks/tests/test_validate_commit_identity.py
@@ -427,5 +427,289 @@ class ParseFailureFailClosedTests(unittest.TestCase):
         )
 
 
+class IndirectExecBypassTests(unittest.TestCase):
+    """Issue #475 fix 2: PreToolUse hooks only see the outer Bash command,
+    so wrappers like `printf '…git commit…' | bash` and `bash <script>`
+    hide the actual git invocation. Each shape below must BLOCK with the
+    canonical-shape hint, and the matching innocent variant must ALLOW
+    (the detector requires both `git` AND `commit` in the inner payload).
+    """
+
+    @staticmethod
+    def _input(command: str, cwd: str | None = None) -> dict:
+        d: dict = {"tool_name": "Bash", "tool_input": {"command": command}}
+        if cwd is not None:
+            d["cwd"] = cwd
+        return d
+
+    def _assert_blocked_as_indirect(self, result):
+        self.assertIsNotNone(result, "indirect-exec bypass must block")
+        assert result is not None
+        self.assertEqual(result["decision"], "block")
+        self.assertIn("indirect-exec", result["reason"])
+        self.assertIn("#475", result["reason"])
+
+    # --- printf | bash family ---------------------------------------------
+
+    def test_printf_pipe_bash_with_git_commit_blocks(self):
+        """POS: `printf '<git commit ...>' | bash` — the headline #475 repro."""
+        cmd = (
+            'printf \'git -c user.name="Bjørn Henriksen" '
+            '-c user.email="..." commit -F /tmp/msg.txt\\n\' | bash'
+        )
+        self._assert_blocked_as_indirect(hook.check(self._input(cmd)))
+
+    def test_printf_pipe_sh_with_git_commit_blocks(self):
+        """POS: `printf '…' | sh` — sh is also a shell interpreter."""
+        cmd = "printf 'git -c user.name=X -c user.email=Y commit -m z\\n' | sh"
+        self._assert_blocked_as_indirect(hook.check(self._input(cmd)))
+
+    def test_printf_pipe_zsh_with_git_commit_blocks(self):
+        """POS: `printf '…' | zsh` — zsh is also a shell interpreter."""
+        cmd = "printf 'git -c user.name=X -c user.email=Y commit -m z\\n' | zsh"
+        self._assert_blocked_as_indirect(hook.check(self._input(cmd)))
+
+    def test_echo_pipe_bash_with_git_commit_blocks(self):
+        """POS: `echo '…git commit…' | bash` — same shape as printf."""
+        cmd = 'echo "git -c user.name=X -c user.email=Y commit -m z" | bash'
+        self._assert_blocked_as_indirect(hook.check(self._input(cmd)))
+
+    def test_printf_pipe_bash_without_git_commit_allows(self):
+        """NEG: `printf 'echo hello' | bash` — innocent pipe-to-shell.
+
+        The detector requires `git` AND `commit` in the inner payload.
+        This is the #475 acceptance bullet "printf 'echo hello' | bash → ALLOWED".
+        """
+        self.assertIsNone(hook.check(self._input("printf 'echo hello' | bash")))
+
+    def test_echo_pipe_bash_without_git_commit_allows(self):
+        """NEG: `echo ls | bash` — innocent pipe-to-shell."""
+        self.assertIsNone(hook.check(self._input("echo ls | bash")))
+
+    # --- shell -c family --------------------------------------------------
+
+    def test_bash_dash_c_with_git_commit_blocks(self):
+        """POS: `bash -c '…git commit…'` — direct -c invocation."""
+        cmd = 'bash -c \'git -c user.name="X" -c user.email="Y" commit -m "z"\''
+        self._assert_blocked_as_indirect(hook.check(self._input(cmd)))
+
+    def test_sh_dash_c_with_git_commit_blocks(self):
+        """POS: `sh -c '…git commit…'`."""
+        cmd = "sh -c 'cd /repo && git -c user.name=X -c user.email=Y commit -m z'"
+        self._assert_blocked_as_indirect(hook.check(self._input(cmd)))
+
+    def test_bash_dash_c_double_quoted_payload_blocks(self):
+        """POS: `bash -c "…git commit…"` — double-quoted payload form."""
+        cmd = 'bash -c "git -c user.name=X -c user.email=Y commit -m z"'
+        self._assert_blocked_as_indirect(hook.check(self._input(cmd)))
+
+    def test_bash_dash_c_without_git_commit_allows(self):
+        """NEG: `bash -c 'ls -la'` — innocent -c invocation."""
+        self.assertIsNone(hook.check(self._input("bash -c 'ls -la'")))
+
+    def test_bash_dash_c_with_git_but_no_commit_allows(self):
+        """NEG: `bash -c 'git status'` — git, but not git commit."""
+        self.assertIsNone(hook.check(self._input("bash -c 'git status'")))
+
+    # --- bash <script> family ---------------------------------------------
+
+    def test_bash_script_containing_git_commit_blocks(self):
+        """POS: `bash /tmp/script.sh` where script content contains git commit.
+
+        Script content is read from disk; if it matches the git-commit shape
+        the wrapper is blocked. This closes the `bash <script-file>` bypass
+        path explicitly called out in #475.
+        """
+        with tempfile.NamedTemporaryFile(
+            mode="w", suffix=".sh", delete=False, encoding="utf-8"
+        ) as f:
+            f.write('git -c user.name="X" -c user.email="Y" commit -F /tmp/msg.txt\n')
+            script_path = f.name
+        try:
+            self._assert_blocked_as_indirect(hook.check(self._input(f"bash {script_path}")))
+        finally:
+            Path(script_path).unlink(missing_ok=True)
+
+    def test_bash_script_without_git_commit_allows(self):
+        """NEG: `bash some-script.sh` whose content does NOT contain git commit.
+
+        #475 acceptance bullet: "Plain bash some-script.sh where the script
+        does NOT contain git commit → ALLOWED".
+        """
+        with tempfile.NamedTemporaryFile(
+            mode="w", suffix=".sh", delete=False, encoding="utf-8"
+        ) as f:
+            f.write("#!/usr/bin/env bash\nls -la\necho done\n")
+            script_path = f.name
+        try:
+            self.assertIsNone(hook.check(self._input(f"bash {script_path}")))
+        finally:
+            Path(script_path).unlink(missing_ok=True)
+
+    def test_bash_script_unreadable_allows(self):
+        """NEG: `bash /nonexistent/path.sh` — unreadable script. Allow rather
+        than block on disk errors; the operator will get a real shell error
+        immediately when bash tries to execute the missing file."""
+        self.assertIsNone(hook.check(self._input("bash /nonexistent/never/exists.sh")))
+
+    # --- process substitution ---------------------------------------------
+
+    def test_bash_process_sub_with_git_commit_blocks(self):
+        """POS: `bash <(echo '…git commit…')` — process substitution bypass."""
+        cmd = 'bash <(echo "git -c user.name=X -c user.email=Y commit -m z")'
+        self._assert_blocked_as_indirect(hook.check(self._input(cmd)))
+
+    def test_bash_process_sub_without_git_commit_allows(self):
+        """NEG: `bash <(echo "ls -la")` — innocent process substitution."""
+        self.assertIsNone(hook.check(self._input('bash <(echo "ls -la")')))
+
+    # --- bare interpreter and pass-through cases --------------------------
+
+    def test_plain_bash_no_args_allows(self):
+        """NEG: bare `bash` (interactive shell) — no inline cmd to inspect.
+
+        #475 acceptance bullet: "Plain bash (interactive shell) with no
+        inline cmd → ALLOWED (no git commit detected)".
+        """
+        self.assertIsNone(hook.check(self._input("bash")))
+
+    def test_direct_git_commit_still_validates_identity(self):
+        """NEG/regression: a direct `git -c ... commit` is still validated
+        through the existing parser — must allow when identity is valid.
+
+        #475 acceptance bullet: existing behavior preserved (still allowed
+        because identity flags present + roster name valid).
+        """
+        valid_name = next(iter(hook.ROSTER), None)
+        if not valid_name:
+            self.skipTest("local roster is empty")
+        valid_email = hook.ROSTER[valid_name]
+        cmd = (
+            f'git -c user.name="{valid_name}" -c user.email="{valid_email}" commit -F /tmp/msg.txt'
+        )
+        self.assertIsNone(hook.check(self._input(cmd)))
+
+    def test_direct_git_commit_missing_identity_still_blocks(self):
+        """NEG/regression: `git commit -m msg` (no identity flags) still blocks.
+
+        #475 acceptance bullet: existing behavior preserved (BLOCKED — current
+        rule). Indirect-exec detection must not bypass the existing
+        identity-validation path for direct invocations.
+        """
+        result = hook.check(self._input('git commit -m "x"'))
+        self.assertIsNotNone(result)
+        assert result is not None
+        self.assertEqual(result["decision"], "block")
+        # Should be the missing-name block, not the indirect-exec block.
+        self.assertNotIn("indirect-exec", result["reason"])
+        self.assertIn("user.name", result["reason"])
+
+    # --- cross-check: indirect-exec runs BEFORE segment parser ------------
+
+    def test_indirect_wrapper_wins_over_outer_innocent_commit_pattern(self):
+        """POS: a command that BOTH has an outer non-commit shape AND an
+        inner-via-wrapper git commit blocks on the wrapper, not silently
+        allows. Otherwise the wrapper detection is structurally redundant
+        with the existing parser."""
+        # Outer command does not contain a real `git commit` at the
+        # tokenize level (just `printf` and `bash`); inner payload does.
+        cmd = "printf 'git -c user.name=X -c user.email=Y commit -m z\\n' | bash"
+        result = hook.check(self._input(cmd))
+        assert result is not None
+        self.assertIn("indirect-exec", result["reason"])
+
+
+class CwdFallbackTests(unittest.TestCase):
+    """Issue #475 fix 1: when no literal `cd <path>` prefix is present, fall
+    back to the tool-call cwd for roster resolution. Lets operators commit
+    from within a child-repo worktree without composing a `cd` prefix."""
+
+    def test_cwd_fallback_loads_child_roster(self):
+        """`git -c user.name=<child-only-name> ... commit` from inside the
+        child worktree (no `cd` prefix) validates against the child roster."""
+        with tempfile.TemporaryDirectory() as td:
+            outer = Path(td)
+            _git_init(outer)
+            _write_roster(outer, {"Nadia": "nadia@example.com"})
+            child = outer / "child"
+            child.mkdir()
+            _git_init(child)
+            _write_roster(child, {"Alice": "alice@example.com"})
+
+            detected = hook._detect_target_roster("git commit -m 'x'", cwd=str(child))
+            self.assertIsNotNone(detected)
+            assert detected is not None
+            self.assertEqual(
+                detected,
+                {
+                    "Nadia": "nadia@example.com",
+                    "Alice": "alice@example.com",
+                },
+            )
+
+    def test_cwd_fallback_when_no_roster_returns_none(self):
+        """cwd has no roster.json → return None (caller uses local ROSTER)."""
+        with tempfile.TemporaryDirectory() as td:
+            empty = Path(td) / "no_roster"
+            empty.mkdir()
+            self.assertIsNone(hook._detect_target_roster("git commit -m 'x'", cwd=str(empty)))
+
+    def test_cd_prefix_wins_over_cwd_fallback(self):
+        """Explicit `cd <path>` in the command takes priority over cwd."""
+        with tempfile.TemporaryDirectory() as td:
+            outer = Path(td)
+            _git_init(outer)
+            cd_target = outer / "explicit"
+            cd_target.mkdir()
+            _git_init(cd_target)
+            _write_roster(cd_target, {"Explicit": "explicit@example.com"})
+
+            # Different repo for the cwd-fallback; should be ignored.
+            cwd_target = outer / "cwd_only"
+            cwd_target.mkdir()
+            _git_init(cwd_target)
+            _write_roster(cwd_target, {"FromCwd": "cwd@example.com"})
+
+            detected = hook._detect_target_roster(
+                f"cd {cd_target} && git commit -m 'x'",
+                cwd=str(cwd_target),
+            )
+            assert detected is not None
+            self.assertIn("Explicit", detected)
+            self.assertNotIn("FromCwd", detected)
+
+    def test_no_cd_no_cwd_returns_none(self):
+        """Regression: omitting both still returns None (existing contract)."""
+        self.assertIsNone(hook._detect_target_roster("git commit -m 'x'"))
+
+    def test_check_uses_cwd_fallback_to_accept_child_only_name(self):
+        """Integration: `check()` with a child-only roster name + cwd
+        pointing at the child repo (no `cd` prefix) → ALLOWED.
+
+        Previously this BLOCKED because `_detect_target_roster` returned
+        None without a cd prefix, the local ROSTER was used, and the
+        child-only name failed lookup. Fix 1 makes the cwd the fallback
+        anchor, restoring symmetry with the cd-prefix path.
+        """
+        with tempfile.TemporaryDirectory() as td:
+            outer = Path(td)
+            _git_init(outer)
+            _write_roster(outer, {"Nadia": "nadia@example.com"})
+            child = outer / "child"
+            child.mkdir()
+            _git_init(child)
+            _write_roster(child, {"Alice": "alice@example.com"})
+
+            cmd = 'git -c user.name="Alice" -c user.email="alice@example.com" commit -m "x"'
+            result = hook.check(
+                {
+                    "tool_name": "Bash",
+                    "tool_input": {"command": cmd},
+                    "cwd": str(child),
+                }
+            )
+            self.assertIsNone(result, f"expected allow, got block: {result}")
+
+
 if __name__ == "__main__":
     unittest.main()
--- a/.claude/hooks/validate_commit_identity.py
+++ b/.claude/hooks/validate_commit_identity.py
@@ -13,6 +13,24 @@ Parent+child roster merge (#112 part a):
   coordinators commit in child repos without duplicating their entries into
   every child `roster.json`.
 
+cwd-fallback for cross-repo detection (#475 fix 1):
+  When the command has NO literal `cd <path>` prefix, the hook still needs
+  to know which repo the commit will land in. The Claude Code harness passes
+  the tool-call cwd via `input_data["cwd"]`; if that cwd has its own
+  `.claude/team/roster.json`, the hook merges that roster instead of falling
+  back to the parent ROSTER. This eliminates the brittleness of requiring
+  every operator to compose a `cd` prefix when their shell cwd already names
+  the right repo.
+
+Indirect-exec bypass detection (#475 fix 2):
+  PreToolUse hooks only see the literal Bash `command` parameter. Wrappers
+  like `printf '<git-commit-cmd>' | bash`, `bash -c '<git-commit-cmd>'`,
+  `bash <script-with-git-commit>`, and `bash <(echo '<git-commit-cmd>')`
+  hide the actual `git commit` from the outer-command tokenizer. The hook
+  now scans for these shapes and, if the payload looks like a git commit,
+  BLOCKS with a message pointing operators to the canonical shape. This
+  closes a silent bypass discovered in P3W11 PR #41 fixup.
+
 Input Language
 ==============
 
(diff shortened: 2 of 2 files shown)


parametrization:
Requestor: Petra Vidović
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: None blocking — see "Tech-debt observation" below for one non-blocking note on script-content-read deviation from #475's out-of-scope line; deviation is defensible.

## Tech-lead verdict — Approved (secondary)

Reviewed at head `867a76dc00ab5933ab0421dbc29592f02659b620` against issue #475, reading `validate_commit_identity.py` and `test_validate_commit_identity.py` via `gh api .../contents/<path>?ref=<sha>`. PR body framing held up; Aino's "Bug fixed" enumeration matches the code at HEAD.

Aisha Idrissi has primary review (SRE-security lens). My secondary brings cross-cutting tech-lead view: hook-authorship discipline, cross-hook interaction, multi-shape pattern coverage, and an ingest-platform-team commit-workflow scan.

### What I verified

**1. Hook authorship discipline vs Aino's other recent hooks**

PR #477's `auto_set_env_test.py` defines `_split_segments` / `_SEPARATORS` for chain-splitting (`auto_set_env_test.py:88-130`). PR #479's `validate_commit_identity` does NOT use that helper — it uses single-pass regex detectors (`_PIPE_TO_SHELL_RE`, `_DASH_C_RE`, `_PROCESS_SUB_RE`, `_SCRIPT_INVOKE_RE` at `validate_commit_identity.py:202-233`). These are different problems: chain-splitting needs segment-aware tokenization to apply per-segment policy; wrapper-shape detection needs to extract inner payloads from heterogeneous container shapes. The single-pass regex approach is correct here — `shlex.split` on the outer command would consume the quotes around `bash -c '<payload>'` and lose the heredoc/process-substitution structure. The PR's own code comment (`validate_commit_identity.py:181-186`) explains exactly this. No convergence-with-#477 opportunity; divergence is justified. Hook-authorship discipline is consistent: helper naming (`_detect_*`, `_payload_looks_like_*`), block-message shape (BLOCKED prefix + canonical-shape suggestion), and `log_pretooluse_block` integration all match Aino's pattern in #477 and #473.

**2. Cross-hook interaction**

- **Hook 4 (validate_pr_review)**: PR-comment-posting goes through `gh pr comment` / `gh issue comment`, not through `bash`. The hook's `_INTERPRETERS = (?:bash|sh|zsh)` (`validate_commit_identity.py:196`) means `printf '...git commit...' | gh ...` would NOT match `_PIPE_TO_SHELL_RE` — `gh` isn't in the interpreter set. PR-comment-posting via `printf|gh` is safe. ✓
- **Hook 15 (enforce_librarian_consulted)**: A librarian invocation buried inside `bash <script>` content would NOT write the consultation sentinel (sentinel writes via the Skill tool from the orchestrator's session, not via subprocess). However, this is pre-existing Hook 15 behavior, not a #479 regression — if anything #479 mildly reinforces it because the script-read path would now BLOCK the commit before the librarian-bypass could land. Not a concern.
- **Hook 21 (post_label_change_wave_field_sync)**: No chain-split helper exists in Hook 21 today (`grep -rn "_split_chain\|chain_split" .claude/hooks/` returns only `auto_set_env_test`'s `_split_segments` and none in Hook 21). So no symmetry-with-#479 to evaluate. Santiago's #457 bundle landed Hook 21 changes but those were regex-widening + log-on-skip, not chain-splitting. Convergence target is a separate refactor question, out of scope here.

**3. Multi-shape pattern coverage — 4 shapes covered, sweep-candidate shapes documented**

I reconstructed the hook's regexes locally and exercised them against shapes NOT in the formal acceptance criteria but mentioned in issue #475's body:

| Shape | Detected? | Notes |
|---|---|---|
| `printf '<git commit>' \| bash` | ✓ BLOCK | acceptance ✓ |
| `bash -c '<git commit>'` | ✓ BLOCK | acceptance ✓ |
| `bash <(echo '<git commit>')` | ✓ BLOCK | bonus (4th shape) |
| `bash <script.sh>` with commit in content | ✓ BLOCK | acceptance ✓ |
| `bash <<EOF\ngit commit ...\nEOF` (heredoc) | ✗ ALLOW | **sweep candidate** |
| `bash <<<'git commit ...'` (here-string) | ✗ ALLOW | **sweep candidate** |
| `eval 'git commit ...'` | ✗ ALLOW | **sweep candidate** |
| `. ./script.sh` / `source ./script.sh` (dot-source/source builtin) | ✗ ALLOW | **sweep candidate** |
| `$(echo git) commit ...` | ✗ ALLOW | listed in issue body but not acceptance |
| `xargs -I {} bash -c '...'` | ✓ BLOCK (incidentally) | inner `bash -c` triggers |

The 4 sweep-candidate shapes are NOT in issue #475's formal acceptance checklist, so they're NOT blockers. Issue #475 acceptance is: `printf|bash`, `bash <script>`, `sh -c '...'` shapes — Aino delivered all 3 + the bonus `bash <(...)`. The heredoc/here-string/eval/dot-source shapes are real evasion paths and worth a follow-up issue, but the acceptance contract is met.

I'll file a sweep-candidate follow-up issue post-merge — non-blocking. Recommend including the 4 detector shapes above in the issue body so a future implementer has a ready test matrix.

**4. Inner-payload regex (Aino's design call)**

I read the code comment at `validate_commit_identity.py:235-250` explaining why the leading shell-operator anchor was dropped:

> "the payload we're scanning is already known to be the inner argument of a printf/echo, shell -c, process substitution, or script file. It often starts inside quotes the extractor didn't strip, so requiring a shell-operator or whitespace anchor before `git` rejects legitimate matches (`'git ... commit ...'` payloads start with `'`)."

Correct design call. The outer-command regex `_COMMIT_FALLBACK_RE` keeps its shell-operator anchor because outer commands typically have one (`&&`, `;`, or BOL); inner payloads do not. Two regexes with the right invariants per layer is cleaner than one over-general regex.

Path-suffix exclusion `(?!\S*/)` preserved at `_INNER_COMMIT_RE` (`validate_commit_identity.py:248`). Compound segment splits via `[^;&|]*?` in the bridge between `\bgit\b` and `\bcommit\b` — that's the right primitive to prevent `git foo; echo commit` from false-matching. ✓

**5. Tests file location — exists on `main` HEAD**

`gh api repos/noorinalabs/noorinalabs-main/contents/.claude/hooks/tests/test_validate_commit_identity.py?ref=main` returns sha `c4951cf7f2ea76832dda915a52c602d0514c40b3`. PR diff stat confirms `c4951cf..7d1687c` (extending an existing file, not creating fresh). 21 pre-existing tests + 25 new = 46 total, matches PR body claim. Verified with `grep -c "def test_" /tmp/petra_479_test_validate_commit_identity.py` → 46. ✓

**6. Tech-debt assessment**

Three sub-points from the brief:

- **Indirect-exec detection is best-effort (heredoc, eval, dot-source not covered)** — NOT documented in the PR body or the code comments. Each is a legitimate evasion path. Recommend a one-line code comment near `_detect_indirect_commit` saying "Coverage is best-effort: heredoc, here-string, eval, and dot-source bypasses are NOT detected; see issue (TBD-sweep) for the residual surface." NOT blocking — this is documentation hygiene, not correctness.
- **256 KiB cap is a magic number** — at `validate_commit_identity.py:284`. Acceptable as inline literal because it's used exactly once with a comment immediately above explaining the choice. A `_MAX_SCRIPT_SIZE_BYTES = 256 * 1024` module-level constant would be marginally cleaner. NOT blocking.
- **Fail-open on OSError** — `validate_commit_identity.py:287` returns `None` (skip), which the caller treats as "no payload extracted" → no block. Security tradeoff: a missing script ALLOWS, even if the bypass intent was real. Endorsed pattern per the existing `_load_merged_roster` fail-open at `validate_commit_identity.py:117-118`. Consistent with the file's overall fail-open posture. The mitigating fact: bash itself will fail to execute a missing script, so the operator gets a real shell error — bypass intent doesn't actually land. ✓

**7. Tech-debt observation (non-blocking) — script-content read deviates from #475 out-of-scope line**

Issue #475's "Out of scope" section says: "Don't widen the hook to read script *contents* of bash invocations — too brittle, too easy to circumvent further." But the acceptance checklist (same issue) requires: "`bash some-script.sh` where script DOES have git commit → BLOCKED." These are in tension — the only way to satisfy the acceptance criterion is to read script contents.

Aino chose to implement script-content reading with the 256 KiB cap + fail-open + `.sh` suffix restriction. This is a defensible deviation: the cap limits the brittleness ceiling, the fail-open prevents the hook from creating worse failure modes, and the `.sh` suffix narrows the surface to legitimate script-invocation patterns. The alternative (block all `bash *.sh`) would over-block legitimate non-commit scripts.

Recommend the PR body or merge commit message explicitly acknowledge the deviation so future maintainers understand the design trade-off ("Implemented script-content read despite issue #475's 'out of scope' note because the acceptance criterion required it; mitigated with 256 KiB cap + `.sh` suffix restriction + fail-open on OSError"). NOT blocking — Aino's "Tech Debt: None" claim is defensible on the absence-of-new-debt axis; this is a process note about the deviation, not a debt item.

### Cross-repo perspective bonus — ingest-platform commit-workflow scan

Quick mental scan of ingest-platform commit patterns this hook will now affect:
- **Camila Restrepo (data lead)** — uses canonical `cd ... && git -c user.name=... commit -F /tmp/msg.txt` pattern. Will work fine via Fix 1 cwd-fallback OR existing `cd` prefix. ✓
- **Imelda Santos (CLI work)** — same canonical pattern. ✓
- **Bjørn Henriksen (deploy worktrees)** — was the trigger for #475 (commit `27009e3`). Fix 1 cwd-fallback eliminates the brittleness that pushed him toward `printf|bash`. Will no longer need the workaround. ✓
- **Sayed-Akhtar Mousavi / codegen flows** — if codegen scripts wrap `git commit` via `bash some-codegen-script.sh`, Fix 2's script-content-read would NOW block them unless they use the canonical shape. **Worth a brief team-wide ping post-merge** that any auto-commit scripts in codegen flows need to invoke git directly, not through a bash wrapper. NOT a #479 blocker — this is a downstream-comms task.

### Conclusion

The PR is a tight, well-tested security hardening fix that addresses both the bypass surface (Fix 2) AND the usability gap that drives operators toward bypasses (Fix 1). The design calls — single-pass regex over shlex, narrow inner-payload matcher, script-content read with cap + fail-open, layered detector ordering before tokenizer — are all correct. 46 tests cover all 4 documented shapes + edge cases. No blockers, no required changes.

Two follow-ups recommended (both NON-blocking):
1. Sweep-candidate issue for the 4 uncovered shapes (heredoc/here-string/eval/dot-source) with my detector-output test matrix as the starting point
2. Brief team-wide ping that auto-commit scripts in codegen/automation must invoke git directly (no `bash <script>` wrappers)

Ship it. Aisha's primary verdict completes the 2-of-2 Approved gate.

parametrization:
Requestor: Aisha Idrissi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: None added — three non-blocking observations on out-of-scope wrapper shapes (heredoc-fed `bash <<EOF`, `ksh -c` / `dash -c` bare invocations, `bash <script.bash>` / no-extension script paths) recommended as followup-issue material or docstring expansion; the 4 wrapper shapes #475 acceptance-lists are all covered and the disk-read safety controls (256 KiB cap, fail-open on OSError) are correctly implemented.

## SRE-security verdict — Approved

**Reviewer:** Aisha Idrissi (deploy SRE) — primary
**Co-reviewer:** Petra Vidović (ingest-platform) — secondary
**Head verified:** `867a76dc00ab5933ab0421dbc29592f02659b620` (read via `gh api .../contents/...?ref=867a76d` — origin-canonical, NO parent-checkout branch switch)
**Mergeable:** `clean`

### Verification methodology

1. Downloaded both PR-HEAD files via `gh api contents?ref=867a76d` into an isolated `/tmp/aisha_vci_runtree/` with PR-HEAD source + tests + supporting `_shell_parse.py` and `annunaki_log.py`.
2. Ran the full PR test suite against the PR-HEAD source: **46 passed, 0 failed** (21 pre-existing + 25 new — matches the spawn-brief tally exactly).
3. Hand-traced `_detect_indirect_commit` ordering: it runs at line 392 of `validate_commit_identity.py`, BEFORE `_find_commit_segment` at line 415. Wrappers cannot slip past the outer tokenizer.
4. Ran 10 adversarial probes against shapes NOT in the PR test suite (results below).
5. Confirmed scope: exactly 2 files changed (`.claude/hooks/validate_commit_identity.py` +236/-11, `.claude/hooks/tests/test_validate_commit_identity.py` +284/-0).
6. Confirmed `--no-verify` enforcement (separate hook) is untouched: `grep -n no-verify /tmp/aisha_vci_867a76d.py` returns zero matches.

### Security-lens checklist (from spawn brief)

| Concern | Verdict | Note |
|---|---|---|
| All 4 wrapper shapes have POS+NEG tests | OK | printf\|bash (4 POS, 2 NEG), bash -c (3 POS, 2 NEG), bash <(...) (1 POS, 1 NEG), bash <script.sh> (1 POS, 2 NEG including unreadable). |
| 256 KiB cap enforced before disk read | OK | Line 284: `if p.stat().st_size > 256 * 1024: return None`. Empirically confirmed via probe — 293 KiB script with `git commit` at top is ALLOWED (over cap); 195 KiB script with same content is BLOCKED (under cap). |
| Fail-open on OSError for unreadable script | OK | Line 287-288: `except (OSError, ValueError): return None` → caller treats as "no payload extracted" → no block. Matches `test_bash_script_unreadable_allows`. |
| Path resolution (relative paths against tool cwd) | OK | Line 278-281: `if not p.is_absolute() and cwd: p = (Path(cwd) / p).resolve()`. Probe confirmed `bash relative-script.sh` with cwd=`/tmp` correctly reads `/tmp/relative-script.sh`. Tilde expansion is NOT handled (`~/script.sh` would resolve via `Path()` raw, no `expanduser()`) — but this is a non-issue because bash itself expands `~` before invocation, so the literal command Claude Code would see is already absolute. |
| Inner-payload regex looseness (false-block risk) | Acceptable trade-off | The `_INNER_COMMIT_RE` `\bgit\b[^;&\|]*?\bcommit\b(?!\S*/)` does false-block `bash -c 'echo "how to git commit"'` (probe 6). This is a documented narrow-defaults choice in the docstring (line 187-189). Operator workaround is trivial (drop the `bash -c` wrapper for echo-of-docs). |
| `bash -c 'echo git commit'` printing text → false-block? | Yes — known trade-off | Documented narrow stance. Not a blocker; legitimate echo-of-documentation doesn't need a shell wrapper. |
| Variable-substituted interpreter (`$PRINTF`) | OK | Probe 2: `PRINTF=$(which printf); $PRINTF '...git commit...' \| bash` → BLOCK. The pipe-to-shell regex matches `printf` literal in the command (the `$PRINTF=...` definition still contains the literal `printf` word), and the `\| bash` shape still matches. |
| Env-var prefix on `bash -c` | OK | Probe 3: `FOO=bar bash -c '...git commit...'` → BLOCK. Word-boundary regex anchors on `bash`. |
| Full-path interpreter (`/bin/sh -c`, `/usr/bin/bash -c`) | OK | Probe 7: both BLOCK. The `\b(?:bash\|sh\|zsh)\b` regex matches against the last path component naturally because `/` is not a word character. |
| Heredoc-fed `bash <<EOF ... EOF` (5th wrapper shape) | Out of scope — NOT covered | Probes 1: `bash <<EOF\ngit commit\nEOF` → ALLOW. This is a known wrapper shape NOT in the 4 detected. **Non-blocking** — not in #475 acceptance bullets, but worth a followup-issue or a docstring note naming it as out-of-scope so the next operator who discovers this isn't confused. |
| `ksh -c` and `dash -c` bare invocations | Out of scope — NOT covered | Probe 7: `ksh -c '...git commit...'` and `dash -c '...git commit...'` → ALLOW. `_INTERPRETERS = (?:bash\|sh\|zsh)`. `ksh` is rare on modern Linux; `dash` is `/bin/sh` on Debian/Ubuntu and IS caught via the `/bin/sh` resolution path, but a bare `dash -c` invocation is not. **Non-blocking** — bare `dash` / `ksh` invocations from Claude Code are exotic. Worth either expanding `_INTERPRETERS` or documenting the deliberate scope. |
| `bash <script>` with `.bash` or no extension | Out of scope — NOT covered | Probe 8: `bash script.bash` and `bash unextended-script` → ALLOW. `_SCRIPT_INVOKE_RE` requires `\.sh\b`. The docstring at line 226-230 documents this is to avoid false-positives, but it does leave a bypass surface for an operator who renames their script. **Non-blocking** — but the more interesting of the three gaps because renaming `script.sh` → `script.bash` is a trivial silent bypass. Worth a followup. |
| Fix 1 cwd-fallback restores cross-repo Nino-shape commits | OK | `CwdFallbackTests::test_check_uses_cwd_fallback_to_accept_child_only_name` validates exactly this scenario: child-only roster name + cwd at child repo + NO `cd` prefix → ALLOW. Reverse case (Nino-attempted commit from wrong cwd) is the existing roster-rejection path; no change. |
| `cd <path>` still takes precedence over cwd fallback | OK | `CwdFallbackTests::test_cd_prefix_wins_over_cwd_fallback` validates `cd /a && commit` with cwd=`/b` resolves to `/a` not `/b`. Code reads `cd_match` BEFORE `cwd` (lines 147-152). |
| `--no-verify` enforcement untouched | OK | Zero `no-verify` references in `validate_commit_identity.py`. That's a separate hook. Confirmed by grep on PR-HEAD source. |
| No scope creep (only 2 files) | OK | `gh api .../pulls/479/files` returns exactly 2 modified files matching the brief. No incidental changes to other hooks, charter, ontology, or workflows. |

### Adversarial probe results (in-session, against PR-HEAD source)

I ran 10 probes against `/tmp/aisha_vci_runtree/` (isolated copy of PR HEAD). The 256 KiB cap is the most security-critical one — confirmed numerically:

- 195 KiB script with `git commit` at top → BLOCK (under cap, content inspected)
- 293 KiB script with `git commit` at top → ALLOW (over cap, content NOT inspected; safe-allow per `_read_script_if_safe`'s contract)

The cap is hard-coded as `256 * 1024` at line 284. Acceptable choice — a multi-megabyte payload funnelled into `bash <script>` is not the normal commit shape, and silently not-inspecting is the right failure mode (the operator still hits the regular commit-identity validation if the script invokes git directly downstream of bash's exec; only the indirect-exec bypass is unenforced for oversized scripts).

### Cross-repo perspective bonus (deploy SRE lens)

1. **Bereket's runbook patterns** — I scanned my knowledge of the deploy runbook shapes (heredoc → file → `git commit -F file` is the canonical shape; `printf | bash` is not in any deploy runbook I'm aware of). The Bereket-named `feedback_heredoc_in_git_commit` memory specifically directs operators TO the `-F /tmp/msg.txt` shape and AWAY from inline heredoc-inside-`-m`. This PR's new block message at line 396-410 surfaces that exact guidance ("If you need a different working directory, prefix with `cd <path> && git -c ... commit ...` — canonical cross-repo shape"). Aligned with deploy operator habits.

2. **CI-driven commits** — Workflows in `noorinalabs-deploy/.github/workflows/` that commit are run as the github-actions bot on the runner, not via Claude Code's Bash tool. The PreToolUse hook only fires inside Claude Code's harness. No regression for CI-driven commits. (Confirmed by re-reading the docstring at line 36-46: "Fires on: PreToolUse Bash" — that's harness-only.)

3. **Worktree-spawned implementers** — Fix 1 (cwd-fallback) is exactly what the worktree-isolation pattern needs. A deploy implementer spawned with `cwd=/path/to/noorinalabs-deploy/.worktrees/.../wave-11` can now `git -c user.name="Bereket Tesfaye" ...` without prefixing `cd /path/to/deploy && ...`. This will reduce one specific class of brittleness I've seen in spawned-agent commits. Net win for the deploy-side workflow.

### Endorsement of Aino's design choices

- **Single-pass regex over shlex-of-outer-command** (docstring line 181-186): correct. shlex would consume the wrapping quotes of `bash -c '<...>'` and lose the heredoc/process-sub structure. Regex-first-then-tokenize-inner is the right shape.
- **Detection narrow by design (requires both `git` AND `commit`)** (docstring line 187-189): correct trade-off. The `_INNER_COMMIT_RE` is intentionally loose on outer anchors because payloads start with `'` from the wrapping quote. A few benign-text-in-`bash -c-echo` false-blocks are acceptable cost for closing the silent bypass.
- **Fail-open on disk-read errors** (line 287-288 + `test_bash_script_unreadable_allows`): correct. A missing script gives the operator a real shell error from bash itself, more useful than a hook block.
- **256 KiB cap** (line 284): correct. Bounds the worst-case disk read and rules out cooperative-overflow attacks against the regex engine.
- **`_PARSE_FAILURE` sentinel + fail-closed regex fallback** (line 325, 415-438): correct. Unbalanced-quote-with-commit-shape blocks; unbalanced-quote-without-commit-shape allows. Asymmetric in the right direction for a security-relevant validator.

### Non-blocking observations (file as followups if useful — none gate this PR)

1. **5th wrapper shape — `bash <<EOF ... EOF` heredoc** is NOT covered. Probe 1 confirmed `bash <<EOF\ngit -c user.name=X commit -m z\nEOF` → ALLOW. This is exotic enough that I don't think it justifies blocking this PR, but worth either (a) a followup issue, or (b) a one-line note in the docstring (currently line 25-32) explicitly naming heredoc-fed-bash as an out-of-scope shape so the next reviewer doesn't have to discover it independently.

2. **`ksh -c` and `dash -c` bypass.** `_INTERPRETERS = (?:bash|sh|zsh)` — bare `dash -c '...git commit...'` ALLOWS. `dash` is `/bin/sh` on Debian/Ubuntu (which DOES match via the `sh` alternative when invoked as `sh` or `/bin/sh`), so the realistic operator-shape is covered. Bare `ksh` is rare on the deploy fleet (`apt list --installed | grep ksh` returns empty on a fresh Ubuntu 24.04). Suggest either expanding the regex to `(?:bash|sh|zsh|ksh|dash|fish)` (cheap) or pinning this scope in the docstring.

3. **`bash <script>` with non-`.sh` extension or no extension.** `_SCRIPT_INVOKE_RE` requires `\.sh\b`. An operator who saves their script as `commit-flow.bash` or `commit-flow` (no extension) silently bypasses disk-read inspection. The docstring at line 226-230 documents the `.sh` restriction is to avoid false-positives, but doesn't acknowledge the bypass surface. **This is the most interesting of the three** — trivially circumventable by rename. Suggest either dropping the `\.sh` requirement (and accepting more false-positive matches on `bash` bare with positional args) or noting it explicitly in the docstring.

### Throughline (W11 hook-tier cluster)

This PR closes the same class of "Hook 4 / validate_commit_identity blind-spot caught in P3W11 fixup" that produced the four earlier W11 hook PRs (#448 hook-21 variableNotUsed, #449 hook 21 dispatcher, #454 hook-21 tool_response, #477 hook fix). Consistent with the wave's hook-hardening theme — these are pre-existing structural gaps being defanged systematically. No throughline-level concern.

Approved.

Issue #453, opened 2026-05-17
bug(harness): PostToolUse Bash hook not firing in current session — Hook 21 silent-skip resurfaces


parametrization:
## Summary

**Claude Code's PostToolUse Bash hook is not firing for Bash tool calls in the current session, despite valid registration in `.claude/settings.json`.** This causes Hook 21 (`post_label_change_wave_field_sync`) to silently never fire on production `gh issue edit --add-label "p{N}-wave-{M}"` calls — superficially identical to the #448 root cause, but at a different layer (harness-vs-hook-code).

The Hook 21 hotfix PR #449 landed correctly: direct `check()` invocations succeed and mutate the project's Wave field. The PostToolUse trigger layer is the new failure surface.

### 1. Hook 21 code is correct
```python
# Direct invocation inside python3 -c
import post_label_change_wave_field_sync as h
h.check({'tool_name':'Bash','tool_input':{'command':'gh issue edit 12 --repo noorinalabs/noorinalabs-deploy --add-label "p3-wave-11"'},'tool_response':{'stdout':'...','stderr':'','interrupted':False}})
# Returns: {'action':'set','repo':'noorinalabs-deploy','issue':'12','option_name':'P3W11'}
# Post-call: deploy#12 Wave field IS set to P3W11 ✓
```

### 2. Dispatcher works when shell-invoked
```bash
echo '{"tool_name":"Bash","tool_input":{"command":"gh issue edit 20 ..."},...}' \
  | python3 .claude/hooks/post_dispatcher.py
# exit=0
# Annunaki log gets new entry from post_wave_kickoff_comment ✓
# Wave field gets set on the simulated target issue ✓
```

### 3. PostToolUse does NOT fire from Claude Code Bash tool calls
After `gh issue edit 19 --repo noorinalabs/noorinalabs-deploy --add-label "p3-wave-11"` via the Bash tool:
- Label was added: ✓
- Wave field on deploy#19 project item: **stayed null** (required manual GraphQL compensation)
- Zero new annunaki entries from post_label_change_wave_field_sync, post_wave_kickoff_comment, or auto_add_issue_to_board
- `wave_11_scope` is missing in cross-repo-status.json, which means post_wave_kickoff_comment SHOULD have logged a "wave_11_scope is missing or empty" advisory — it didn't

### 4. annunaki_monitor also doesn't fire on intentional Bash failures
- `false` (exit 1): no new annunaki entry
[...]

PR #481, opened 2026-05-18
tech-debt(annunaki_monitor): IGNORE_PATTERNS for boolean-test idioms + 2 followup tests (#474)

### LINKED ISSUE #474: tech-debt(annunaki_monitor): extend IGNORE_PATTERNS for boolean-test idioms + add 2 follow-up test cases (#473 post-merge)
## Background

Follows from #472 / PR #473 reviewer feedback (Santiago + Aino). Sized for W12 sweep (alongside #470, #471, #467).

## A. IGNORE_PATTERNS extension (Santiago's concern)

PR #473 closed the silent-failure blind spot by capturing exit-N-no-output commands. Side-effect: common Bash boolean-test idioms now produce log entries on their failure branch. Simulated 15 idioms at PR-HEAD; the following will log on exit-1:

- `[ -d ... ]` / `[ -f ... ]` / `[ -e ... ]` (POSIX `[`)
- `[[ ... ]]` (Bash double-bracket)
- `test -f ...` / `test -d ...` (explicit test command)
- `diff --quiet a b`
- `grep -q pattern file` (non-error miss)
- `pgrep nginx` / `pkill -0 nginx`
- `which/command -v foo` (not-found)
- `git diff --quiet`
- `if [ -f x ]; then ...; fi` (whole-conditional)
- `[[ ... ]] && ...` short-circuit chains

Dedup is keyed on 200-char command prefix + error_lines join — for silent failures error_lines is empty, so distinct boolean-test paths each produce a distinct entry. Realistic flood risk during normal wave work; `/annunaki-attack` Step 2 noise bucket catches them but adds reviewer burden.

### Proposed fix

Extend `IGNORE_PATTERNS` in `.claude/hooks/annunaki_monitor.py` with idiom-aware filters that activate only when the *sole* error signal is `exit_code != 0` and `stderr` is empty (i.e., not when a pattern actually matched). Candidate regexes:

```python
SILENT_BOOLEAN_TEST_PATTERNS = [
    re.compile(r"^\s*\["),                  # POSIX [ test
    re.compile(r"^\s*\[\["),                # Bash [[ test
    re.compile(r"^\s*test\b"),              # explicit test
    re.compile(r"\bgrep\s+-[a-zA-Z]*q"),    # grep -q
    re.compile(r"^\s*(pgrep|pkill)\b"),
    re.compile(r"^\s*(which|command\s+-v)\b"),
    re.compile(r"\bdiff\s+--quiet\b"),
    re.compile(r"\bgit\s+diff\s+--quiet\b"),
    re.compile(r"\bif\s+\["),               # if [ ... ]; then ...; fi
]
```

Apply check inside `check()`: skip ONLY when `matched_patterns == ['exit_code=...']` AND command matches one of these.

## B. Test coverage follow-ups (Aino's non-blocking suggestions)

Add to `test_annunaki_monitor.py`:

1. **exit-0-with-stderr-pattern**: command succeeds (exit 0) but stderr contains a pattern (e.g., a warning that matches `^error\b` casefold). Verify it's still captured via stderr pattern match.
2. **ignored-command-with-nonzero-exit**: command matches `_should_ignore` (e.g., `grep -i error`) AND exits non-zero. Verify ignore takes precedence over exit-code capture.

## Acceptance

- [ ] All 10 boolean-test idioms above produce NO log entry when their sole failure signal is exit_code
- [ ] Commands matching the new idioms that ALSO produce stderr-pattern matches DO still log (pattern-match takes precedence)
- [ ] Two new test cases added
- [ ] All existing tests still pass

## Out of scope

- Don't widen IGNORE_PATTERNS further than these documented idioms — over-filtering is worse than over-capturing
- Don't touch the exit-code branch ordering — #473 fix is correct

parametrization:
Closes #474.

## Summary

Two-part fix to follow up on PR #473's silent-failure capture work:

**Part A — `SILENT_BOOLEAN_TEST_PATTERNS` (Santiago's concern)**: 10 documented Bash boolean-test idioms now skip when their ONLY failure signal is a non-zero exit code. Pattern matches always take precedence (a `[ -f x ]` that emits a real `Traceback` still logs).

**Part B — Two non-blocking test follow-ups from Aino's #473 review**:
1. `test_exit_zero_with_stderr_pattern_still_captured`
2. `test_ignore_pattern_takes_precedence_over_nonzero_exit`

## Design choice: separate list, conditional skip

The issue body specifies a NEW list (`SILENT_BOOLEAN_TEST_PATTERNS`), not an extension of the existing `IGNORE_PATTERNS`. Why this matters:

- `IGNORE_PATTERNS` is checked unconditionally at the top of `check()` — matching commands NEVER produce a log entry, regardless of output.
- `SILENT_BOOLEAN_TEST_PATTERNS` is checked AFTER `is_error` detection — only when `matched_patterns == ["exit_code=N"]` (no stdout/stderr pattern matched).

This preserves the invariant from the issue's "Out of scope" section: "Don't widen `IGNORE_PATTERNS` further than these documented idioms — over-filtering is worse than over-capturing." A `grep -q` that emits an unexpected `panic:` to stderr will still log, because the stderr pattern match populates `matched_patterns` with `stderr:panic:` and the silent-skip condition no longer holds.

## Patterns added

```python
SILENT_BOOLEAN_TEST_PATTERNS = [
    re.compile(r"^\s*\["),                       # POSIX [ test
    re.compile(r"^\s*\[\["),                     # Bash [[ test
    re.compile(r"^\s*test\b"),                   # explicit test builtin
    re.compile(r"\bgrep\s+-[a-zA-Z]*q"),         # grep -q (and -qE, -qi, etc.)
    re.compile(r"^\s*(pgrep|pkill)\b"),
    re.compile(r"^\s*(which|command\s+-v)\b"),
    re.compile(r"\bdiff\s+--quiet\b"),
    re.compile(r"\bgit\s+diff\s+--quiet\b"),
    re.compile(r"^\s*if\s+\["),                  # if [ ... ]; then ...; fi
    re.compile(r"^\s*if\s+\[\["),                # if [[ ... ]]; then ...; fi
]
```

Word-boundary anchors (`\b`) on `test` avoid false-matching `pytest`, `latest`, etc. Anchored leading-whitespace prefixes on `[`, `[[`, `test`, `pgrep`, `pkill`, `which`, `command -v`, `if [` avoid matching when these tokens appear mid-command. `grep -q` is intentionally not anchored — it matches inside chains like `cmd | grep -q ...`.

## Test plan

- [x] All 9 prior tests still pass (regression guard for #472 silent-failure capture)
- [x] 16 new `SilentBooleanTestIdiomTests` cover each idiom's NEGATIVE-match (no log on silent exit-1) plus three precedence cases (idiom-with-stderr-pattern, idiom-with-stdout-pattern, idiom-exit-0)
- [x] `test_real_silent_failure_still_logged` — `false` is NOT in the idiom list, so the silent-failure capture path still works for non-idiom commands
- [x] 2 new `AinoFollowupTests` per #473 reviewer feedback
- [x] Full hooks suite: 893 passed, no regressions
- [x] `uvx ruff@latest check .claude/hooks/` — All checks passed
- [x] `uvx ruff@latest format --check .claude/hooks/` — All formatted

### Local run

```
$ ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/test_annunaki_monitor.py -v
============================== 28 passed in 0.07s ==============================

============================= 893 passed in 1.51s ==============================

$ uvx ruff@latest check .claude/hooks/
All checks passed!

$ uvx ruff@latest format --check .claude/hooks/
68 files already formatted
```

## Tech Debt

None introduced. PR removes a noise class identified post-#473 as a real flood risk during normal wave work. No TODO planted, no helper deferred.

## Out of scope (per issue)

- Did not widen `IGNORE_PATTERNS` further than the 10 documented idioms — preserves the "over-filtering is worse than over-capturing" invariant from the issue body.
- Did not modify the exit-code branch ordering — #473's reorder is correct and is preserved.

### DIFF
--- a/.claude/hooks/annunaki_monitor.py
+++ b/.claude/hooks/annunaki_monitor.py
@@ -11,9 +11,12 @@ Input Language:
   Matches:       Bash with non-zero exit_code OR stdout/stderr matching any
                  ERROR_PATTERNS regex (Traceback, fatal:, ModuleNotFoundError,
                  etc.) and NOT matching IGNORE_PATTERNS
-  Does NOT match: any non-Bash tool, empty stdout+stderr, command-text
-                  containing grep-for-error / --error flags / error-named
-                  paths (false-positive guards), session-dedup hits
+  Does NOT match: any non-Bash tool, command-text containing grep-for-error /
+                  --error flags / error-named paths (false-positive guards),
+                  silent-boolean-test idioms (`[`, `[[`, `test`, `grep -q`,
+                  `pgrep`, `pkill`, `which`, `command -v`, `diff --quiet`,
+                  `git diff --quiet`, `if [...]`) when their ONLY error signal
+                  is a non-zero exit code, session-dedup hits
   Flag pass-through: stdin JSON is forwarded verbatim to `check()` by the
                      PostToolUse dispatcher (`post_dispatcher.py`)
 
@@ -64,6 +67,37 @@ IGNORE_PATTERNS = [
     re.compile(r"--error", re.IGNORECASE),  # flags containing "error"
     re.compile(r"error_log|error\.log|errorhandl", re.IGNORECASE),  # filenames
 ]
+
+# Silent-boolean-test idioms (#474): commands whose ONLY failure signal is a
+# non-zero exit code with empty stdout/stderr — by design, not an error.
+# After #473 closed the silent-failure blind spot, these idioms became noise.
+# Apply ONLY when matched_patterns is exactly ["exit_code=..."] (no stdout or
+# stderr pattern matched) — pattern matches always win, ignore-on-silent-exit
+# never suppresses a real error.
+SILENT_BOOLEAN_TEST_PATTERNS = [
+    re.compile(r"^\s*\["),  # POSIX `[ ... ]` test
+    re.compile(r"^\s*\[\["),  # Bash `[[ ... ]]` test
+    re.compile(r"^\s*test\b"),  # explicit `test` builtin
+    re.compile(r"\bgrep\s+-[a-zA-Z]*q"),  # grep -q (and -qE, -qi, etc.)
+    re.compile(r"^\s*(pgrep|pkill)\b"),  # process find/signal, exit 1 on no match
+    re.compile(r"^\s*(which|command\s+-v)\b"),  # which/command -v not-found
+    re.compile(r"\bdiff\s+--quiet\b"),  # diff --quiet: exit 1 on differences
+    re.compile(r"\bgit\s+diff\s+--quiet\b"),  # git diff --quiet: same
+    re.compile(r"^\s*if\s+\["),  # `if [ ... ]; then ...; fi` whole-conditional
+    re.compile(r"^\s*if\s+\[\["),  # `if [[ ... ]]; then ...; fi`
+]
+
+
+def _is_silent_boolean_test(command: str) -> bool:
+    """Return True if command matches a documented silent-boolean-test idiom.
+
+    Caller must additionally verify that the ONLY failure signal is exit_code
+    (no stdout/stderr pattern match) — pattern matches always take precedence.
+    """
+    for pattern in SILENT_BOOLEAN_TEST_PATTERNS:
+        if pattern.search(command):
+            return True
+    return False
 
 
 def _extract_error_lines(text: str, max_lines: int = 10) -> list[str]:
@@ -139,6 +173,14 @@ def check(input_data: dict) -> dict | None:
                 break
 
     if not is_error:
+        return None
+
+    # #474 silent-boolean-test filter: when the ONLY signal is a non-zero
+    # exit code (no stderr/stdout pattern matched), suppress documented
+    # boolean-test idioms whose failure branch is by-design. Stderr/stdout
+    # pattern matches always win — a `[ -f x ]` that somehow emits a real
+    # Traceback still logs.
+    if matched_patterns == [f"exit_code={exit_code}"] and _is_silent_boolean_test(command):
         return None
 
     error_lines = _extract_error_lines(combined_output)
--- a/.claude/hooks/tests/test_annunaki_monitor.py
+++ b/.claude/hooks/tests/test_annunaki_monitor.py
@@ -171,5 +171,210 @@ class AnnunakiMonitorTests(unittest.TestCase):
         self.assertIsNone(result)
 
 
+class SilentBooleanTestIdiomTests(unittest.TestCase):
+    """#474 coverage: documented boolean-test idioms whose by-design failure
+    branch is non-zero exit with empty output must NOT produce log entries
+    after #473 closed the silent-failure capture path. Each idiom gets a
+    NEGATIVE-match test (no log) plus a precedence test that confirms a real
+    stderr-pattern match still wins."""
+
+    def setUp(self):
+        self._saved_env = {
+            "ENVIRONMENT": os.environ.pop("ENVIRONMENT", None),
+            "NOORIN_HOOK_TEST_MODE": os.environ.pop("NOORIN_HOOK_TEST_MODE", None),
+        }
+        am._seen_hashes.clear()
+        self._tmpdir = tempfile.TemporaryDirectory()
+        self._errors_path = Path(self._tmpdir.name) / "errors.jsonl"
+        self._orig_monitor_file = am.ERRORS_FILE
+        self._orig_log_file = alog.ERRORS_FILE
+        am.ERRORS_FILE = self._errors_path
+        alog.ERRORS_FILE = self._errors_path
+
+    def tearDown(self):
+        am.ERRORS_FILE = self._orig_monitor_file
+        alog.ERRORS_FILE = self._orig_log_file
+        self._tmpdir.cleanup()
+        for k, v in self._saved_env.items():
+            if v is None:
+                os.environ.pop(k, None)
+            else:
+                os.environ[k] = v
+
+    # --- Each documented idiom: NEG match on silent exit-1 ---
+
+    def test_posix_bracket_silent_exit_not_logged(self):
+        """NEG: `[ -d /nonexistent ]` exit 1 silent → no log."""
+        result = am.check(_bash_event("[ -d /nonexistent ]", exit_code=1))
+        self.assertIsNone(result)
+        self.assertFalse(self._errors_path.exists())
+
+    def test_bash_double_bracket_silent_exit_not_logged(self):
+        """NEG: `[[ -f /nonexistent ]]` exit 1 silent → no log."""
+        result = am.check(_bash_event("[[ -f /nonexistent ]]", exit_code=1))
+        self.assertIsNone(result)
+
+    def test_test_builtin_silent_exit_not_logged(self):
+        """NEG: `test -f /nonexistent` exit 1 silent → no log."""
+        result = am.check(_bash_event("test -f /nonexistent", exit_code=1))
+        self.assertIsNone(result)
+
+    def test_grep_q_no_match_silent_exit_not_logged(self):
+        """NEG: `grep -q pattern file` non-error miss → no log."""
+        result = am.check(_bash_event("grep -q needle /tmp/haystack.txt", exit_code=1))
+        self.assertIsNone(result)
+
+    def test_grep_q_with_flags_silent_exit_not_logged(self):
+        """NEG: `grep -qE` / `grep -qi` variants → no log."""
+        for cmd in ("grep -qE '^foo' f.txt", "grep -qi BAR f.txt", "grep -Eq '^foo' f.txt"):
+            am._seen_hashes.clear()
+            result = am.check(_bash_event(cmd, exit_code=1))
+            self.assertIsNone(result, f"{cmd!r} must not log")
+
+    def test_pgrep_no_match_silent_exit_not_logged(self):
+        """NEG: `pgrep nginx` not-running → no log."""
+        result = am.check(_bash_event("pgrep nginx", exit_code=1))
+        self.assertIsNone(result)
+
+    def test_pkill_no_match_silent_exit_not_logged(self):
+        """NEG: `pkill -0 nginx` not-running → no log."""
+        result = am.check(_bash_event("pkill -0 nginx", exit_code=1))
+        self.assertIsNone(result)
+
+    def test_which_not_found_silent_exit_not_logged(self):
+        """NEG: `which foo` not-installed → no log."""
+        result = am.check(_bash_event("which nonexistent-binary", exit_code=1))
+        self.assertIsNone(result)
+
+    def test_command_v_not_found_silent_exit_not_logged(self):
+        """NEG: `command -v foo` not-installed → no log."""
+        result = am.check(_bash_event("command -v nonexistent-binary", exit_code=1))
+        self.assertIsNone(result)
+
+    def test_diff_quiet_differs_silent_exit_not_logged(self):
+        """NEG: `diff --quiet a b` files differ → no log."""
+        result = am.check(_bash_event("diff --quiet /tmp/a /tmp/b", exit_code=1))
+        self.assertIsNone(result)
+
+    def test_git_diff_quiet_dirty_silent_exit_not_logged(self):
+        """NEG: `git diff --quiet` working tree dirty → no log."""
+        result = am.check(_bash_event("git diff --quiet", exit_code=1))
+        self.assertIsNone(result)
+
+    def test_if_bracket_conditional_silent_exit_not_logged(self):
+        """NEG: `if [ -f x ]; then ...; fi` whole conditional with exit-1
+        false-branch → no log."""
+        result = am.check(_bash_event("if [ -f /nonexistent ]; then echo yes; fi", exit_code=1))
+        self.assertIsNone(result)
+
+    def test_if_double_bracket_conditional_silent_exit_not_logged(self):
+        """NEG: `if [[ ... ]]; then ...; fi` shape → no log."""
+        result = am.check(_bash_event("if [[ -f /nonexistent ]]; then echo yes; fi", exit_code=1))
+        self.assertIsNone(result)
+
+    # --- Precedence: pattern match wins over idiom-on-silent-exit ---
+
+    def test_idiom_with_stderr_pattern_still_logged(self):
+        """POS: `[ -f x ]` would normally be silent-skip, BUT if stderr has
+        a real error pattern (e.g., `fatal:`), it still logs. Pattern matches
+        take precedence — the idiom-skip only suppresses the bare exit_code
+        signal."""
+        result = am.check(
+            _bash_event(
+                "[ -f /nonexistent ]",
+                stderr="fatal: unexpected internal state\n",
+                exit_code=1,
+            )
+        )
+        self.assertIsNotNone(result, "stderr pattern match must override idiom skip")
+        rec = _read_records(self._errors_path)[0]
+        # Both signals present: exit_code AND stderr pattern
+        self.assertIn("exit_code=1", rec["matched_patterns"])
+        self.assertTrue(any("fatal" in p for p in rec["matched_patterns"]))
+
+    def test_idiom_with_stdout_pattern_still_logged(self):
+        """POS: silent-test idiom that produces a stdout Traceback (would be
+        absurd in reality, but the precedence rule must hold) → logs."""
+        result = am.check(
+            _bash_event(
+                "test -f /nonexistent",
+                stdout="Traceback (most recent call last):\nValueError: x\n",
+                exit_code=1,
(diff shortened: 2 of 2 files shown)


parametrization:
Requestor: Santiago Ferreira
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: None

## Release-coordinator verdict — Approved

Reviewed at head `fb09fd1cd18bd6da368815dda020dc29d02eb996`. Read both files via `gh api .../contents?ref=<head>` (not via PR body framing). Checked out the branch locally, ran the test suite (28/28 + full 893-pass regression sweep), and ran a 28-case adversarial probe against the actual hook through a fresh subprocess-injected `append_jsonl_record`.

I'm the natural primary because the `SILENT_BOOLEAN_TEST_PATTERNS` ask is mine from PR #473's review (filed as #474). I had to verify Wanjiku's final shape against what I actually asked for, not just whether it works.

### Does the list match what I asked for? Yes, and tighter than my original suggestion.

My #473 review listed 9 idioms with a coarse regex sketch: `^(test|\[|\[\[|diff|grep|pgrep|which|git diff --quiet)\b`. Wanjiku's 10 compiled regexes are STRICTER in three productive ways:

1. **`\bgrep\s+-[a-zA-Z]*q`** — only matches `grep -q` (and `-qE`, `-qi`, `-Eq`). My sketch was too broad (would have silenced `grep needle file` without `-q`, which IS a real silent failure worth logging). Wanjiku narrowed correctly. Verified via probe: `grep needle f.txt` (no `-q`) at exit-1 → LOG. `grep -q needle f.txt` at exit-1 → SKIP. Both correct.

2. **`\bdiff\s+--quiet\b`** — only matches `diff --quiet`, not bare `diff a b`. My sketch had bare `diff`, but `diff a b` exit-1 produces diff lines on stdout (not silent — has output worth seeing). Verified: `diff a b` with `< foo\n> bar` stdout exits-1 → LOG (correct). `diff --quiet a b` exit-1 silent → SKIP (correct).

3. **Added `pkill` and `command -v`** — I forgot these in my original list. Both are by-design exit-1-on-not-found idioms; `command -v` is the POSIX-portable replacement for `which` and shows up in modern hooks. Good additions.

### Critical design call: separate constant, conditional skip — correct.

The PR body explicitly calls out the chosen design: `SILENT_BOOLEAN_TEST_PATTERNS` is a NEW constant, NOT an extension of `IGNORE_PATTERNS`. The semantic difference is load-bearing:

- `IGNORE_PATTERNS` (line 65-69): checked at line 147 BEFORE error detection. Match = NEVER LOG regardless of any subsequent error signal.
- `SILENT_BOOLEAN_TEST_PATTERNS` (line 77-88): checked at line 183 AFTER `is_error` detection. Match = LOG SUPPRESSED ONLY WHEN `matched_patterns == ["exit_code=N"]` exactly.

This preserves the #472 silent-failure invariant on a per-command-class basis: `false` (not in idiom list) still logs; `[ -f x ]` with `fatal:` in stderr still logs (stderr-pattern match populates `matched_patterns` beyond just `exit_code`). The three precedence tests (`test_idiom_with_stderr_pattern_still_logged`, `test_idiom_with_stdout_pattern_still_logged`, `test_real_silent_failure_still_logged`) pin all three branches of this invariant. Verified directly via probe — `[ -f /x ]` exit-1 + `fatal:` stderr → LOG with BOTH `exit_code=1` AND `stderr:fatal:` markers in `matched_patterns`.

This is exactly the design I'd have asked for if I'd thought it through to this depth in the #473 review. Wanjiku's "Out of scope" adherence to the issue body ("Don't widen IGNORE_PATTERNS further") protected the design.

### Regex anchor correctness verified.

Each pattern's intent vs. effect, traced via probe:

- `^\s*test\b` — `pytest tests/` exit-1 correctly LOGs (the `\b` after `test` blocks the `pytest`/`latest` false-match). Verified.
- `^\s*\[` and `^\s*\[\[` — leading whitespace tolerated; in-chain `foo && [ -f x ]` does NOT match (leading anchor on `foo`). Verified.
- `^\s*if\s+\[` — `if [ x ]; then ...` matches; `if pgrep nginx; then ...` does NOT match (no `[` after `if`). The latter is a documented gap — `if <cmd>` shape uses the cmd's exit code as the test, and that cmd may emit real error stderr that's worth capturing. Defensible narrowing.
- `\bgrep\s+-[a-zA-Z]*q` — matches `-q`, `-qE`, `-qi`, `-Eq`, `-Eiq`. Verified via the `test_grep_q_with_flags_silent_exit_not_logged` parameterized loop.
- `^\s*(pgrep|pkill)\b` and `^\s*(which|command\s+-v)\b` — leading-anchor only. `foo && pgrep nginx` does NOT match (LOGs at exit-1). This is the same class as `if <cmd>` — chain-position idioms are not detected. Probably fine; in-chain test commands rarely show up as silent failures during wave work.

### Cross-hook awareness with my #461 — composed cleanly.

setUp pops `ENVIRONMENT` and `NOORIN_HOOK_TEST_MODE` env vars before each test, then restores in tearDown. This bypasses #461's test-mode write suppression so `append_jsonl_record` actually writes to the tmpdir — the precondition for `_read_records` returning the JSONL the test asserts on. Without this env-pop pattern, the test would silently pass an assertion against an empty file. Verified the same pattern in all 3 test classes. Composition of #461 (writer-gate) and #481 (generator-gate) is sound — they sit at different lifecycle phases as Wanjiku's PR-body section "Cross-hook awareness with YOUR #461" correctly identifies.

### CI green at this exact head.

4/4 jobs SUCCESS (Ruff lint, Ruff format check, Mypy type check, Smoke test) per `gh pr view 481 --json statusCheckRollup`. Anchored to `fb09fd1`. Force-push re-anchor required per `feedback_reviewer_correction_vs_merge_race`.

### Scope.

Only `.claude/hooks/annunaki_monitor.py` (+45/-3) and `.claude/hooks/tests/test_annunaki_monitor.py` (+205/-0). No settings.json, no charter, no other hooks. Module docstring updated to enumerate the new silent-boolean-test idiom set as the documented matcher-set extension.

### Sweep candidates (non-blocking, file follow-up if desired)

1. **In-chain idiom detection** — `foo && pgrep nginx` and `if pgrep nginx; then ...` log at exit-1. Wanjiku's leading-anchor design is defensible (chain-position idioms rarely flood) but a future bash-aware tokenizer could extend coverage. Watch the wave-error logs for a few sessions to confirm whether this is real noise.

2. **`SILENT_BOOLEAN_TEST_PATTERNS` consolidation across hooks** — three hooks now compile bash-idiom regex lists (`auto_set_env_test.py`'s test-runner detection from PR #477, this hook's silent-test idiom list, and the existing `IGNORE_PATTERNS` here). A shared `_bash_idioms.py` could hold these as named constants. Wave wrapup or post-W11 tech-debt sweep candidate. Not urgent.

3. **Memory update**: `feedback_verify_posttooluse_firing` recipe — Wanjiku's PR body promises a post-merge memory refresh, but the silent-boolean-test patterns now SUPPRESS `[ -d /nonexistent ]` and similar verification commands. The recipe should use `false` (still captured, NOT in idiom list) per `test_real_silent_failure_still_logged`. Worth flagging in the post-merge memory update.

### Anchor for re-anchor protocol

Verdict anchored at `fb09fd1cd18bd6da368815dda020dc29d02eb996`. Any force-push invalidates this Approved and requires re-review.

parametrization:
Requestor: Aisha Idrissi
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: None added — three small non-blocking observations on regex coverage edges (split-flag `grep -E -q` falls through to LOG, `diff --quiet` exit=2 trouble case is not discriminated from exit=1 differ case, `if [ true ]; then false; fi` body-failure-inside-conditional silently skips) recommended as a single small followup issue; the headline #474 goal — suppressing pure-exit-code idiom noise — is correctly implemented and the conditional-silence invariant is structurally pinned by 3 precedence tests.

## SRE-secondary verdict — Approved

**Reviewer:** Aisha Idrissi (deploy SRE) — secondary
**Primary reviewer:** Santiago Ferreira — not yet posted at time of this review (his original IGNORE_PATTERNS observation became #474)
**Head verified:** `fb09fd1cd18bd6da368815dda020dc29d02eb996` via `gh api .../contents/...?ref=fb09fd1` — origin-canonical, no parent-checkout branch switch
**Mergeable:** clean
**Scope:** exactly 2 files (`.claude/hooks/annunaki_monitor.py` +45/-3, `.claude/hooks/tests/test_annunaki_monitor.py` +205/-0). No incidental changes.

### Verification methodology

1. Downloaded PR-HEAD source + tests via `gh api contents?ref=fb09fd1` into `/tmp/aisha_am_runtree/` with PR-HEAD `annunaki_monitor.py` + `test_annunaki_monitor.py` + supporting `annunaki_log.py` (copied from local main, hash-verified).
2. Ran the full PR-HEAD test suite against the PR-HEAD source: **28 passed, 0 failed** (9 pre-existing + 17 new idiom + 2 Aino-followup = 28, matches the spawn-brief tally).
3. Hand-traced the conditional-silence invariant in source (line 183): `if matched_patterns == [f"exit_code={exit_code}"] and _is_silent_boolean_test(command): return None`. The equality check on `matched_patterns` is exact (a list with one element) — any stderr or stdout pattern match appends a second entry, breaking the equality and bypassing the idiom skip. Structurally correct.
4. Ran 13 adversarial probes against shapes NOT in the PR test suite (results below).
5. Cross-checked deploy fleet operational impact: scanned `noorinalabs-deploy/scripts/` for idiom use — `verify_deployment.sh` has 25+ `if [` and several `grep -qi` pipes, `bootstrap-vps.sh` has 7 `if [` blocks. Confirmed these are NOT affected by this PR.

### Endorsement of the IGNORE_PATTERNS vs SILENT_BOOLEAN_TEST_PATTERNS split

The PR intentionally does NOT extend `IGNORE_PATTERNS` (which is unconditional, applied before `is_error` detection) — instead it introduces `SILENT_BOOLEAN_TEST_PATTERNS` (conditional, applied AFTER `is_error` detection ONLY when `matched_patterns == ["exit_code=N"]`). This is the right design call:

- Unconditional ignore (`IGNORE_PATTERNS`) would silently drop a `grep -q` that emits a real stderr error like `grep: file: No such file or directory` — that's a real bug, not noise.
- Conditional silence preserves "pattern matches always win" — if the idiom command somehow emits a real ERROR_PATTERN match, it still logs (verified by probe 7: `grep -q pattern f.txt` with stderr `No such file` → LOGS).

The documentation in the docstring at lines 71-76 and 178-184 calls this distinction out explicitly. Good.

### Operational impact on deploy fleet (deploy SRE lens)

Scanned `noorinalabs-deploy/scripts/` for idiom use:

- **`bootstrap-vps.sh`**: 7 `if [` blocks, including the SSH-key merge block I just reviewed in PR #287. None affected by this PR.
- **`verify_deployment.sh`**: 25+ `if [` blocks + `grep -qi` pipes for header verification. None affected.

**Why not affected**: the annunaki hook fires on PostToolUse Bash *inside Claude Code's harness*, not on operator-run scripts on a Hetzner VPS. Production deploy operators running `bash scripts/verify_deployment.sh` on a VPS don't trigger this hook at all. And even when an SRE agent runs `bash scripts/verify_deployment.sh` FROM a Claude Code session, the hook sees the outer command (`bash scripts/...`) — which doesn't match any idiom regex — so a non-zero exit from the script would still log. The inner `if [...]; then ...; fi` blocks of the script are invisible to this hook by design.

**No regression for deploy fleet runbooks.** No audit-log entries that operators rely on are suppressed by this change.

### Spawn-brief lens checklist

| Lens | Verdict | Note |
|---|---|---|
| Operational impact on deploy fleet | OK | Hook fires only in Claude Code harness; deploy fleet scripts unaffected. Verified above. |
| `grep -qq` (lint typo) handling | OK | Probe 2: matched as quiet (`-[a-zA-Z]*q` allows multiple chars including double q). Treats as quiet — fine. |
| `grep -E -q` (split flags) | Minor edge | Probe 1: `grep -E -q` (TWO separate `-X` tokens with space between) → LOGS. Regex requires `q` within a single `-` token. `grep -Eq`, `grep -qE`, `grep -q -E` all SKIP correctly. The split form is uncommon but possible. **Non-blocking** — would log a legitimate quiet-grep as noise rather than drop a real error. |
| `test` as part of `cd && test` chain | OK by design | Probe 3: `cd /tmp && test -d foo` → LOGS. The `^\s*test\b` regex requires line-start anchor; the `cd` prefix prevents idiom skip. Per spawn brief: this is correct — the chain's exit_code rolls in `cd`'s success and `test`'s failure, and logging the chain is the right behavior. |
| Precedence test 1 (`test_idiom_with_stderr_pattern_still_logged`) | OK | Realistic input (`[ -f /nonexistent ]` + stderr `fatal: unexpected internal state\n` + exit 1). Assertion checks both `exit_code=1` AND a `fatal`-containing pattern are in `matched_patterns`. Verified the test passes against PR-HEAD source. |
| Precedence test 2 (`test_idiom_with_stdout_pattern_still_logged`) | OK | Realistic enough (`test -f /nonexistent` + stdout `Traceback...` + exit 1 — admittedly contrived since `test` doesn't produce Tracebacks, but the test pins the invariant that stdout pattern matches override idiom skip). Assertion is `assertIsNotNone(result)` — correct. |
| Precedence test 3 (`test_real_silent_failure_still_logged`) | OK | Direct #472 regression guard: `false` + exit 1 → must log because `false` is NOT in `SILENT_BOOLEAN_TEST_PATTERNS`. Assertion is correct. |
| CI 4/4 green at HEAD | OK | mypy, ruff format, ruff lint, smoke all pass. |

### Adversarial probe results (in-session, against PR-HEAD source)

Ran 13 probes against `/tmp/aisha_am_runtree/`. The 28 in-PR tests pin the documented invariants; my probes target edges that aren't pinned. Highlights:

- **Probes 6/7/11**: real errors still log correctly via stderr/stdout pattern matches (no regression).
- **Probes 3/4/5/13**: chain prefixes correctly NOT idiom-skipped via line-start anchors. `cat | grep -q nonsense` correctly SKIPS (grep -q pattern matches mid-pipe).
- **Probe 1**: split-flag `grep -E -q` falls through to LOG (small false-positive, mild noise).
- **Probe 8**: `diff --quiet` exit=2 with stderr LOGS correctly (real error via stderr pattern); same exit=2 without stderr SKIPS (small false-negative, see below).
- **Probe 10**: `if [ true ]; then false; fi` body failure is SKIPPED (see below).

### Three non-blocking observations (file as a single followup if useful)

1. **Split-flag `grep -E -q` (separate tokens) falls through to LOG.** The regex `\bgrep\s+-[a-zA-Z]*q` requires `q` inside a single `-X` token. `grep -E -q`, `grep -q -E`, etc., have two tokens separated by whitespace, so the regex doesn't match. Combined forms `-Eq`, `-qE`, `-qq` all work. Mild noise — the split form is uncommon enough that this isn't worth blocking on, but a followup could add an alternative regex `(?:^|\s)-q(?:\s|$)`.

2. **`diff --quiet` exit=2 (real trouble) is not discriminated from exit=1 (files differ).** Probe 8: when `diff --quiet` exits 2 with stderr (`diff: /tmp/a: No such file or directory`), the stderr pattern correctly captures it. But when exit=2 with NO stderr, it's SKIPPED as idiom-noise. `diff` semantics: 0=same, 1=different (idiom), 2=trouble (real). Could discriminate by checking `exit_code != 1` before idiom skip for diff-specific patterns. Edge case — most `diff: trouble` cases emit stderr.

3. **`if [ ... ]; then <body>; fi` body failures are SKIPPED when condition is truthy.** Probe 10: `if [ -d /tmp ]; then false; fi` exit=1 → SKIPPED because the outer command starts with `if [`. The body's `false` (or any real failure) silently slips through. **The most interesting blind-spot of the three** — operator code that does `if [ -d /target ]; then real-deploy-step; fi` would lose audit-log entries on `real-deploy-step` failure. Workaround: don't wrap real failures in `if [ ... ]; then`. Probably acceptable since this idiom is more common in setup scripts than in failing operator code, but worth documenting.

Also minor — **the `^\s*\[` regex (line 78) matches both `[` and `[[`**, so the `^\s*\[\[` regex on line 79 is redundant. Probe verified: `^\s*\[` matches `[[ -f x ]]`. The `_is_silent_boolean_test` OR loop short-circuits at first match, so functionally it's fine; reorder or drop the redundant one in a future cleanup if desired.

### Throughline (W11 hook-tier cluster)

This PR fits the W11 hook-hardening / annunaki-noise-reduction cluster (#452 test-mode write suppression, #453/#454 tool_response migration, #473 silent-failure capture, now #474 idiom suppression). Net throughline: the annunaki monitor has been progressively tightened to log REAL failures while suppressing structural noise. Consistent with the wave's stated TD-ratio goal. This PR is a small, well-scoped, well-tested step in that direction.

Approved.

PR #483, opened 2026-05-18
tech-debt(hook): auto_set_env_test — extend gh-skip to env wrapper form (#181)

### LINKED ISSUE #181: auto_set_env_test: extend gh-skip to cover 'env' wrapper form
## Context

Followup to #114 / PR #178. The gh-skip short-circuit strips leading `VAR=value` env-assignment tokens via the regex `^[A-Za-z_][A-Za-z0-9_]*=\S*\s+` before checking argv[0] == `gh`. This covers the common `FOO=1 BAR=2 gh pr comment ...` form.

## Gap

The literal `env` command form — `env FOO=bar gh pr comment ...` — is not covered. `env` is argv[0]; the gh-skip doesn't fire. For commands without `--body`, this is a latent false-positive risk if the operator ever uses the `env` wrapper with a gh subcommand whose args mention a test runner. In practice, the `--body` skip covers every observed repro, so this is theoretical.

## Proposed fix

Extend `_is_gh_invocation` to also skip one leading `env` token (optionally with its own leading VAR=val assignments) before checking argv[0] == `gh`. Add a test `test_env_wrapper_gh_invocation_skipped`.

## Priority

Low — no observed repro, and the `--body` skip masks the common failure mode. Guard against future ops who prefer `env` wrapping.

## References

- PR #178 — gh + --body skip implementation
- Issue #114 — original substring false-positive

parametrization:
Closes #181.

## Summary

Extends the `gh`-skip short-circuit in `auto_set_env_test._is_gh_invocation` to also recognise the literal `env` wrapper form. The hook previously handled `FOO=1 BAR=2 gh pr ...` (shell env-assignment prefix) but not `env FOO=bar gh pr ...` (where `env` is argv[0]).

## Change

In `_is_gh_invocation`: after stripping leading shell `VAR=val` tokens, if argv[0] is `env` (exact token match), peek at the next non-whitespace token. If it's a flag (`-i`/`-u`/`--`), bail — those forms are out of scope per the issue. Otherwise strip env's own leading `VAR=val` assignments and check whether the next token is `gh`.

```python
if token == "gh":
    return True
if token == "env":
    remainder = stripped[len("env") :].lstrip()
    next_token = remainder.split(None, 1)[0] if remainder else ""
    if next_token.startswith("-"):
        return False  # env -i / env -u / env -- out of scope
    after_env = _strip_leading_env(remainder).lstrip()
    if not after_env:
        return False
    return after_env.split(None, 1)[0] == "gh"
return False
```

Single-level only — does NOT recurse into nested `env env env gh ...` (real-world nonsensical). Token-equality check (`token == "env"`) means `environment_check`, `envvar`, `envoy`, etc. don't false-match.

## Out of scope (per #181 + my reading)

- **`env -i gh ...` / `env -u VAR gh ...` / `env -- gh ...`**: flag forms. `env -i` clears the inherited environment — operator-explicit "clean slate" signal. Treating it as a transparent gh wrapper would let an `env -i gh ...` command mask a pytest invocation hidden in its later args. The flag-form is rare in operator commands and the `--body` short-circuit covers the common gh case regardless.
- **Nested wrappers**: `env env gh ...`, `nohup gh ...`, `timeout 30 gh ...`. Single-level extension only. Adding more is feature-creep beyond what the issue requested ("extend gh-skip to env wrapper form" — singular).

## Test plan

- [x] 40 prior tests still pass (no regressions in pre-existing `GhSubcommandSkipTests`, `BodyFlagSkipTests`, `PositiveMatchTests`, `EnvAlreadySetTests`, `NonTestCommandTests`, `ChainedCommandTests`, `ShortCircuitWithChainsTests`)
- [x] 8 new tests in `EnvWrapperGhSkipTests`:
  - `test_env_wrapper_gh_invocation_skipped` — bare `env FOO=bar gh ...`
  - `test_env_wrapper_with_multiple_assignments_gh_skipped` — `env FOO=1 BAR=2 gh ...`
  - `test_env_wrapper_bare_gh_skipped` — `env gh ...` (no assignments)
  - `test_shell_prefix_then_env_wrapper_gh_skipped` — `FOO=1 env BAR=2 gh ...`
  - `test_env_wrapper_non_gh_does_not_skip` — `env FOO=1 pytest tests/` STILL BLOCKS
  - `test_env_dash_i_flag_not_recursed` — `env -i gh ...` (allowed because no test runner, but not via env-wrapper path)
  - `test_env_dash_i_does_not_mask_pytest` — `env -i pytest tests/` STILL BLOCKS (proves -i flag doesn't bypass)
  - `test_env_wrapper_does_not_match_envar_prefix_word` — `environment_check pytest ...` STILL BLOCKS (token-equality protects against false-match)
- [x] Full hooks suite: 829 passed, no regressions
- [x] `uvx ruff check .claude/hooks/` — All checks passed
- [x] `uvx ruff format --check .claude/hooks/` — 64 files already formatted

### Local run

```
$ ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/test_auto_set_env_test.py -v
============================== 48 passed in 0.09s ==============================

============================= 829 passed in 1.30s ==============================
```

## Tech Debt

None introduced. PR closes a known gap (#181 was marked low priority "no observed repro") with a focused extension and explicit out-of-scope test coverage proving the flag-form is not silently bypassed.

## Reference precedent check

Per `feedback_hook_brief_grep_precedent_preflight`: grep'd Aino's #479 indirect-exec work in `validate_commit_identity.py` for any reusable `env`-wrapper extraction helper. Her work targets interpreter wrappers (`bash -c`, `printf | bash`, process substitution, script-file invocation) — a different surface than the literal `env` command. No reusable primitive; the change is isolated to `_is_gh_invocation` in this file.

### DIFF
--- a/.claude/hooks/auto_set_env_test.py
+++ b/.claude/hooks/auto_set_env_test.py
@@ -106,12 +106,40 @@ def _strip_leading_env(command: str) -> str:
 
 
 def _is_gh_invocation(command: str) -> bool:
-    """True if the effective argv[0] (after env assignments) is `gh`."""
+    """True if the effective argv[0] (after env assignments) is `gh`.
+
+    Also handles a single leading `env` wrapper (#181):
+        env FOO=bar gh pr comment ...
+        env FOO=1 BAR=2 gh pr comment ...
+        FOO=1 env BAR=2 gh pr comment ...   (env after shell VAR prefix)
+
+    Does NOT recurse into `env -i`, `env -u`, `env --` flag forms (out of
+    scope per #181 — `env -i` clears the inherited environment which is an
+    operator-explicit "I want a clean slate" signal; treating it the same
+    as the bare `env` wrapper would let an `env -i gh ...` shape mask a
+    pytest invocation hidden in its later args). The flag-form is rare in
+    operator commands and the `--body` short-circuit covers the common gh
+    case regardless.
+    """
     stripped = _strip_leading_env(command).lstrip()
     if not stripped:
         return False
     token = stripped.split(None, 1)[0]
-    return token == "gh"
+    if token == "gh":
+        return True
+    if token == "env":
+        # Peek at the next token. If it's a flag (starts with `-`), bail —
+        # we don't recurse into `env -i` / `env -u` / `env --` shapes.
+        remainder = stripped[len("env") :].lstrip()
+        next_token = remainder.split(None, 1)[0] if remainder else ""
+        if next_token.startswith("-"):
+            return False
+        # Strip env's own leading VAR=val assignments, then re-check argv[0].
+        after_env = _strip_leading_env(remainder).lstrip()
+        if not after_env:
+            return False
+        return after_env.split(None, 1)[0] == "gh"
+    return False
 
 
 def _has_body_flag(command: str) -> bool:
--- a/.claude/hooks/tests/test_auto_set_env_test.py
+++ b/.claude/hooks/tests/test_auto_set_env_test.py
@@ -74,6 +74,79 @@ class GhSubcommandSkipTests(unittest.TestCase):
         self.assertIsNone(result)
 
 
+class EnvWrapperGhSkipTests(unittest.TestCase):
+    """Issue #181: extend gh-skip to recognise the literal `env` wrapper:
+        env FOO=bar gh pr comment ...
+        env FOO=1 BAR=2 gh pr ...
+        FOO=1 env BAR=2 gh pr ...   (shell VAR prefix BEFORE env)
+    The bare `env` form (no flags) is a transparent wrapper — it sets env
+    vars then exec's gh. The hook must treat the inner gh as argv[0] for
+    short-circuit purposes."""
+
+    def test_env_wrapper_gh_invocation_skipped(self) -> None:
+        """NEG: `env FOO=bar gh pr comment ...` — env wrapper, gh skip fires."""
+        result = hook.check(_bash('env FOO=bar gh pr comment 1 --body "pytest"'))
+        self.assertIsNone(result, "env wrapper before gh must allow")
+
+    def test_env_wrapper_with_multiple_assignments_gh_skipped(self) -> None:
+        """NEG: `env FOO=1 BAR=2 gh pr ...` — multiple env's own assignments."""
+        result = hook.check(_bash('env FOO=1 BAR=2 gh pr comment 1 --body "pytest"'))
+        self.assertIsNone(result)
+
+    def test_env_wrapper_bare_gh_skipped(self) -> None:
+        """NEG: `env gh pr ...` — env with no assignments, just the gh exec."""
+        result = hook.check(_bash('env gh pr comment 1 --body "pytest"'))
+        self.assertIsNone(result)
+
+    def test_shell_prefix_then_env_wrapper_gh_skipped(self) -> None:
+        """NEG: `FOO=1 env BAR=2 gh pr ...` — shell env prefix, THEN `env`
+        wrapper, THEN gh. _strip_leading_env eats `FOO=1`, then we see
+        `env BAR=2 gh ...` and recurse one step."""
+        result = hook.check(_bash('FOO=1 env BAR=2 gh pr comment 1 --body "pytest"'))
+        self.assertIsNone(result)
+
+    def test_env_wrapper_non_gh_does_not_skip(self) -> None:
+        """POS: `env FOO=1 pytest tests/` — env wrapper around pytest is
+        STILL a test command (env doesn't change the fact that the eventual
+        argv[0] passed to exec is pytest). Must block."""
+        result = hook.check(_bash("env FOO=1 pytest tests/"))
+        assert result is not None, "env-wrapped pytest must still block"
+        self.assertEqual(result.get("decision"), "block")
+
+    def test_env_dash_i_flag_not_recursed(self) -> None:
+        """POS: `env -i gh pr ...` — the `-i` flag form is OUT OF SCOPE per
+        #181. `env -i` clears the inherited environment (operator-explicit
+        signal); we don't pretend the inner gh is argv[0]. With pytest
+        absent here this is allowed via the "not a test command" path, but
+        we assert the recursion didn't happen by checking the next case."""
+        result = hook.check(_bash("env -i gh pr comment 1"))
+        # No pytest, no make test → allowed regardless of skip path. The
+        # important assertion is the next test: env -i with pytest must still
+        # block (i.e., we did NOT silently treat it as a gh invocation).
+        self.assertIsNone(result)
+
+    def test_env_dash_i_does_not_mask_pytest(self) -> None:
+        """POS: `env -i gh pr comment --title 'pytest tests/'` is a contrived
+        case but illustrates the safety: because the `-i` flag form is not
+        recognised as a gh wrapper, the `--body`/`--body-file` skip is the
+        only thing protecting us — and there's no `--body` here. But because
+        `pytest` only appears inside `--title` (no `pytest` token outside it)
+        the test-runner regex needs that text. Use a more direct example:
+        `env -i pytest tests/` — the `-i` flag form is not a gh skip path,
+        and pytest is the effective program. Must block."""
+        result = hook.check(_bash("env -i pytest tests/"))
+        assert result is not None, "env -i around pytest must still block"
+        self.assertEqual(result.get("decision"), "block")
+
+    def test_env_wrapper_does_not_match_envar_prefix_word(self) -> None:
+        """POS: `environment_check pytest tests/` — leading token `environment_check`
+        starts with `env` but is NOT the `env` builtin. Must not be confused
+        for the wrapper; pytest is still the effective program → block."""
+        result = hook.check(_bash("environment_check pytest tests/"))
+        assert result is not None, "env-prefix lookalike token must not match `env`"
+        self.assertEqual(result.get("decision"), "block")
+
+
 class BodyFlagSkipTests(unittest.TestCase):
     """Short-circuit #2: --body / --body-file flags exempt the whole command."""
 


parametrization:
Requestor: Aino Virtanen
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: None

## Verdict: Approved

Primary reviewer (Aisha Idrissi is the convergent witness). Reviewed at PR head_sha `6e4572b913905ff16049ad1515cadb3984cc4f81` via `gh api repos/.../contents?ref=<sha>` (origin-over-local per charter `pull-requests.md` § Origin > Local Clone). 2 files / +103 / -2. CI 4/4 green at HEAD (Ruff lint, Ruff format, Mypy, Smoke). Re-ran target file at HEAD: **48 passed** (40 pre-existing from #477 + 8 new in `EnvWrapperGhSkipTests`). Full hooks suite: **829 passed** — no sibling-hook regression, matches PR-body claim exactly.

This is the natural follow-up to my own #477 hook hardening. Wanjiku extended exactly the right surface (`_is_gh_invocation`) and left the rest of the file untouched — no scope creep, no opportunistic adjacent fixes, no broader hook refactor.

### Implementation correctness — traced at HEAD

The extension to `_is_gh_invocation` (auto_set_env_test.py:108-142) is structurally sound. Trace for `env FOO=bar gh pr ...`:

1. `_strip_leading_env(command)` removes leading `VAR=val ` tokens. The token `env` is NOT a VAR-assignment (no `=` in first token), so nothing is stripped. After `lstrip()`: `"env FOO=bar gh pr ..."`.
2. First token: `"env"`. Branch `if token == "env"`.
3. `remainder = stripped[len("env"):].lstrip()` → `"FOO=bar gh pr ..."`.
4. `next_token = "FOO=bar"`. Doesn't start with `-` → continue.
5. `after_env = _strip_leading_env(remainder).lstrip()` → `"gh pr ..."`.
6. `after_env.split(None, 1)[0] == "gh"` → True. Return True. ✓

The conservative-correct design call to bail on `-` prefix (lines 135-136) is the load-bearing safety: it punts `env -i`, `env -u VAR`, `env --` without trying to interpret each flag's semantics. `env -i` in particular clears the inherited environment — an operator-explicit "clean slate" signal that should NOT be treated as a transparent wrapper. The `test_env_dash_i_does_not_mask_pytest` test pins this safety end-to-end: `env -i pytest tests/` still BLOCKs because the gh-skip doesn't fire, the test-runner regex matches `\bpytest\b`, and `ENVIRONMENT=test` is absent.

### Edge-case verification (in-band, 18 cases beyond the test suite)

Ran `_is_gh_invocation` directly against 18 additional shapes to verify the extension's contract holds at the edges (not just the headline acceptance cases). All matched expectations:

| Case | Expected | Got |
|---|---|---|
| `env FOO=bar gh pr` | True | True |
| `env FOO=1 BAR=2 gh pr` | True | True |
| `env gh pr` (no vars) | True | True |
| `FOO=1 env BAR=2 gh pr` | True | True |
| `env FOO=1 $RUNNER tests/` (env+non-gh) | False | False |
| `env -i gh pr` | False | False |
| `env -u FOO gh pr` | False | False |
| `env -- gh pr` | False | False |
| `env -i pytest tests/` | False | False |
| `environment_check tests/` (prefix-word) | False | False |
| `envoy tests/` (prefix-word) | False | False |
| `env` (bare, no args) | False | False |
| `env FOO=1` (vars only, no program) | False | False |
| `env  FOO=bar  gh pr` (double-space) | True | True |
| `   env FOO=1 gh` (leading whitespace) | True | True |
| `FOO=1 BAR=2 env BAZ=3 gh` (multi-shell-var + env + env-var + gh) | True | True |
| `env env gh pr` (nested env) | False | False |
| `env FOO=1 ghx pr` (adversarial: gh-like token) | False | False |
| `env FOO=1 ghpr` (adversarial: gh-like token) | False | False |

The token-equality check (`token == "env"`) correctly rejects all `env`-prefix-words (`environment_check`, `envoy`, `envvar`, etc.). The exact-equality on `gh` (`after_env.split(None, 1)[0] == "gh"`) correctly rejects `ghx`/`ghpr` adversarial cases. Whitespace handling is robust via the consistent `.lstrip()` calls.

Nested `env env gh ...` returns False as documented — single-level extension only. This is the right call: `env env gh` is operationally absurd, and recursing would unnecessarily widen the parser surface.

### Regression preservation

All 40 pre-existing tests pass at HEAD, including the two most relevant ones:
- `test_gh_skip_beats_env_prefix` (`ENVIRONMENT=test gh pr ...` still skips via shell-env-prefix path)
- `test_gh_with_multiple_env_prefixes` (`FOO=1 BAR=2 gh pr ...` still skips via shell-env-prefix path)

Both paths through `_is_gh_invocation` — the original shell-env-prefix path AND the new `env`-wrapper path — coexist cleanly. The `_strip_leading_env` call at the top of the function still handles shell-VAR prefixes; the new `if token == "env"` branch only fires when the first non-shell-VAR token is `env`. No code path was deleted or weakened.

### Scope assertion

Confirmed via `gh pr diff 483`: exactly 2 files (`auto_set_env_test.py` and its test file). 32-line addition to the hook function with updated docstring, 71-line `EnvWrapperGhSkipTests` test class with 8 methods. No adjacent refactor, no helper rename, no broader cleanup. Charter-compliant `Closes #181.` as PR-body first line. Labels `tech-debt` + `p3-wave-11` verified via `gh pr view --json labels`. Wanjiku's `feedback_hook_brief_grep_precedent_preflight` cross-check on my own #479 indirect-exec work is the right discipline — no reusable primitive exists; the change is isolated.

### Sweep candidates (non-blocking observations, future follow-ups only)

1. **Other transparent wrappers around `gh`** that have the same shape but aren't covered:
   - `nohup gh ...`
   - `timeout 30 gh ...` / `timeout --foreground 30 gh ...`
   - `time gh ...`
   - `xargs gh ...` (used in batch issue operations)
   - `stdbuf -o0 gh ...`

   None observed in our session traces — `nohup`/`timeout` are uncommon for gh subcommands. Worth a single follow-up issue if any future false-positive surfaces, but NOT in this PR's scope. The pattern would be a small extension to the `if token == "env"` branch (or a small whitelist set), but the gain is purely defensive. Worth tracking as a Pass-B candidate if the family grows.

2. **`env -S` POSIX-2024 split-string form**: `env -S "FOO=1 BAR=2 gh pr ..."` would currently be punted by the `-` flag check. `-S` is rare and operator-explicit (literal split-string syntax) so the punt is defensible; documenting the boundary in the docstring would tighten the contract. Pre-existing-style follow-up only.

3. **Sibling hook `validate_commit_identity`**: my own #475/#479 work covers `printf|bash`, `bash -c`, `bash <(...)`, `bash <script>` — but does NOT cover `env <script>` or `env`-wrapped indirect-exec. Cross-domain consideration: if `env` is recognized as a transparent wrapper here for `gh`, is it also a bypass vector for the commit-identity hook (`env git -c user.name=X commit ...`)? Quick read of `_strip_leading_env` in `validate_commit_identity.py` — it strips shell VAR=val, then tokens are passed to `find_git_subcommand` which looks for `git` as segment[0]. The `env` literal would NOT be recognized as a git-commit shape there, which means the commit-identity hook ALREADY-correctly handles `env`-wrapped `git commit` (it falls through to `_find_commit_segment` returning None → ALLOW). That's only safe because git commit without `-c user.name=` would BLOCK at the missing-name check… wait, but `env git -c user.name=X commit ...` has `env` as segment[0] for the `find_git_subcommand` lookup, which returns None. So commit-identity hook DOESN'T fire on `env git commit` at all. This is a real follow-up — `env` is a bypass vector for commit-identity validation too. Worth filing as a hook-21-tier issue (same shape as #475). Documented here for the W11 retro / Phase B planning rather than as a blocker on this PR.

### Verdict basis

Approve on the strength of:
- Closes exactly #181 with zero scope creep
- `_is_gh_invocation` extension verified correct in-band on 8 in-suite tests + 18 additional edge cases (20 unique shapes total)
- Conservative `-` prefix punt is the right safety boundary; `test_env_dash_i_does_not_mask_pytest` pins the safety end-to-end
- Token-equality on both `env` and `gh` correctly rejects prefix-word lookalikes and adversarial gh-like tokens
- 48/48 in target file, 829/829 full hooks suite, no regression
- 4/4 CI gates green at HEAD `6e4572b`
- Scope strictly 2 files / +103 / -2; charter-compliant PR body

The TechDebt line above is "None" — sweep candidates listed in the verdict body are forward-looking observations for future PRs, not gaps introduced by #483.

parametrization:
Requestor: Aisha Idrissi
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: None added — one small non-blocking asymmetry observation worth a docstring note (`env ENVIRONMENT=test pytest tests/` is BLOCKED despite operator-intent symmetry with `env FOO=bar gh ...` which is ALLOWED — `_segment_has_env_in_leading_block` doesn't recognize the env-wrapper form for env-var presence the way `_is_gh_invocation` does for argv[0]); trivial operator workaround (drop the `env ` wrapper) and the block message already points there.

## SRE-secondary verdict — Approved

**Reviewer:** Aisha Idrissi (deploy SRE) — secondary
**Primary reviewer:** Aino Virtanen — not yet posted at time of this review
**Head verified:** `6e4572b913905ff16049ad1515cadb3984cc4f81` via `gh api .../contents/...?ref=6e4572b` — origin-canonical, no parent-checkout branch switch (per [[feedback_reviewer_must_not_branch_switch_parent]])
**Mergeable:** clean
**Scope:** exactly 2 files (`auto_set_env_test.py` +30/-2, `test_auto_set_env_test.py` +73/-0). No incidental changes.

### Verification methodology

1. Downloaded PR-HEAD source + tests via `gh api contents?ref=6e4572b` into `/tmp/aisha_aset_runtree/` (PR-HEAD source + tests + supporting `annunaki_log.py` from main hash-verified).
2. Ran full PR-HEAD test suite against PR-HEAD source: **48 passed, 0 failed**.
3. Hand-traced `_is_gh_invocation` logic (lines 108-142): `env` wrapper recursion is bounded to a single step, with explicit early-out on flag-form (`-i`, `-u`, `--`, `-S`) per docstring punt at line 116-122.
4. Ran 12 adversarial probes targeting shapes NOT in the PR test suite.
5. Cross-checked deploy fleet operational impact: `grep -rn -E "env\s+(FOO|[A-Z_]+=)" noorinalabs-deploy/scripts/ noorinalabs-deploy/.github/workflows/` returned **zero matches** — deploy SREs do not use the `env VAR=val ...` wrapper form in scripts or workflows.

### Spawn-brief lens checklist

| Lens | Verdict | Note |
|---|---|---|
| Operational impact on deploy fleet | OK | Zero usage of `env VAR=val ...` wrapper form anywhere in `noorinalabs-deploy/scripts/` or `.github/workflows/`. Pattern is CI-script / interactive-debug surface, not deploy-runbook surface. No regression risk for deploy. |
| `env FOO=1 pytest tests/` still BLOCKS | OK | Test `test_env_wrapper_non_gh_does_not_skip` (line 108-114) covers it explicitly: `env FOO=1 pytest tests/` → block. Probe 5 extended this to `env FOO=1 make test` → BLOCK (also correct). Probe 1 extended to chain form (`cd /x && env FOO=1 pytest`) → BLOCK via per-segment scan. The hook correctly insists on env on the test segment regardless of `env` wrapper presence. |
| CI 4/4 green at HEAD | OK | mypy, ruff format, ruff lint, smoke all pass at `6e4572b`. |

### Probe results (12 adversarial cases)

| Probe | Command | Verdict | Correct? |
|---|---|---|---|
| 1 | `cd /x && env FOO=1 pytest tests/` | BLOCK | Yes — per-segment scan identifies env-wrapped pytest in segment 2 lacks `ENVIRONMENT=test` in leading env-block |
| 2 | `env -- pytest tests/` | BLOCK | Yes — `--` is flag form, `_is_gh_invocation` correctly returns False, falls through to segment scan |
| 2 | `env -- gh pr comment 1 --body x` | ALLOW | Yes — caught by `--body` short-circuit (not by `env -- gh` since `--` is flag-form) |
| 3 | `env env gh pr comment 1` | ALLOW | Yes — recursion bounded to one step per docstring; no pytest, so allowed via no-test-runner path |
| 4 | `env 'FOO=a b c' gh pr comment 1 --body x` | ALLOW | Yes — quoted env value with spaces doesn't match `_ENV_ASSIGN` regex but body-flag short-circuit covers it |
| 5 | `env FOO=1 make test` | BLOCK | Yes — symmetric to pytest case; env-wrapped real test runner blocks |
| 6 | `env ENVIRONMENT=test pytest tests/` | BLOCK | **Mildly inconsistent** — see observation below |
| 7 | `env -i ENVIRONMENT=test pytest tests/` | BLOCK | Yes — `-i` flag form explicitly punted; operator must put env directly on pytest |
| 8 | `command -v env && pytest tests/` | BLOCK | Yes — `env` appears as data not as wrapper; segment 2 pytest correctly identified |
| 9 | `cd /x && env FOO=1 gh pr comment 1 --body 'pytest note'` | ALLOW | Yes — body-flag short-circuit catches it |
| 10 | `env_setup && pytest tests/` | BLOCK | Yes — `env_setup` is a different token, not the `env` wrapper |
| 11 | `env -S 'FOO=1' gh pr comment 1` | ALLOW | Yes — `-S` flag form falls through; no pytest, allowed via no-test-runner path |
| 12 | bare `env` | ALLOW | Yes — degenerate, no pytest, allowed |

### Endorsement of the `env` wrapper recursion bound

The PR explicitly punts on `env -i` / `env -u` / `env --` / `env -S` flag-form variants (docstring lines 116-122). This is the right call for security and simplicity:

- `env -i` clears the inherited environment — operator-explicit "clean slate" signal. Recursing into the inner argv would let an `env -i gh ...` shape mask a pytest hidden in later args. The conservative-fall-through is safer.
- The bounded one-step recursion (no `env env env gh ...` chains) keeps the parser simple and avoids exponential-blowup edge cases on pathological input.
- The `--body`/`--body-file` short-circuit still covers the realistic gh-with-flag-form-env case (probe 2 `env -- gh ... --body x` → ALLOW).

Net: the recursion bound is sufficient for the documented gh-wrapper case (`env FOO=bar gh ...`) without overreach.

### One small non-blocking observation

**`env ENVIRONMENT=test pytest tests/` is BLOCKED** (probe 6) despite operator-intent symmetry with `env FOO=bar gh ...` being ALLOWED. The asymmetry:

- `_is_gh_invocation` recognizes `env` as a wrapper and looks past it for `gh` as argv[0].
- `_segment_has_env_in_leading_block` uses `_strip_leading_env` which only matches `VAR=val` tokens — the literal `env` token is not stripped, so when the segment is `env ENVIRONMENT=test pytest tests/`, the "consumed" leading env-block contains only `env ` (or nothing, depending on how `_strip_leading_env` handles `env`). `ENVIRONMENT=test` is in the segment but not in the "leading env-block" per the function's definition.

An operator writing `env ENVIRONMENT=test pytest tests/` would expect it to pass (it's semantically equivalent to `ENVIRONMENT=test pytest tests/` in shell). The block message would suggest `env ENVIRONMENT=test ENVIRONMENT=test pytest tests/` or similar, which is awkward.

**Not a blocker** — the workaround is trivial (drop the `env ` wrapper) and the rewrite suggestion the hook produces will point the operator there. Worth a one-line docstring note acknowledging the asymmetry, OR a future tightening of `_segment_has_env_in_leading_block` to also recognize the `env` wrapper form when checking for `ENVIRONMENT=test` (small symmetry-fix). File as a single small followup issue if useful.

### Throughline

This PR closes #181 cleanly — narrow scope, defensive `env -*` flag-form punt, thorough test coverage (48 tests, 17 new in `EnvWrapperGhSkipTests` plus the existing 31). Fits the W11 hook-hardening cluster. The follow-up to #476 (silent-bypass per-segment fix) extends the gh-skip exception to a new wrapper form without weakening the per-segment env-on-test-runner invariant.

Approved.

PR #493, opened 2026-05-18
tech-debt(hook): validate_commit_identity — extend indirect-exec detection to heredoc / eval / dash+ksh / extension-agnostic (#482)

### LINKED ISSUE #475: tech-debt(hook): validate_commit_identity bypassed by printf|bash, bash <script>, indirect-exec shapes — only sees outer command

### LINKED ISSUE #482: tech-debt(hook): validate_commit_identity — extend indirect-exec detection to heredoc / eval / ksh / extension-agnostic shapes (#475 follow-up)
## Background

Follow-up to PR #479 (closed #475) — indirect-exec bypass hardening for `validate_commit_identity`. Two reviewers (Aisha Idrissi primary, Petra Vidović secondary) independently surfaced **the same five additional bypass shapes** that PR #479 explicitly did not cover. None block the security improvement from #475 (which closed `printf|bash`, `bash -c`, `bash <(...)`, and `bash <script.sh>` — the four shapes #475's acceptance bullets named).

Filing as one consolidated follow-up rather than five issues per Aisha's + Petra's both-flagged convergence.

## Shapes not covered by PR #479

| # | Shape | Bypass risk |
|---|-------|-------------|
| 1 | `bash <<EOF\ngit commit ...\nEOF` (heredoc) | Trivial bypass via heredoc-fed shell |
| 2 | `bash <<<'git commit ...'` (here-string) | Trivial single-line bypass |
| 3 | `eval 'git commit ...'` | Outer tokenizer sees `eval`, not git |
| 4 | `dash -c '...'` / `ksh -c '...'` | `_INTERPRETERS = (?:bash|sh|zsh)` regex omits these |
| 5 | `bash script.bash` / `bash extensionless-script` | `_SCRIPT_INVOKE_RE` requires `\.sh\b` — trivially circumvented by rename |

**Highest-priority concern**: #5 (`.sh`-extension circumvention by rename). The current safety relies on a file extension check; an operator who knows the hook reads `.sh` files but not `.bash` could rename their script and bypass detection without changing content.

## Proposed fix

Extend the existing helpers in `.claude/hooks/validate_commit_identity.py`:

1. **Add heredoc / here-string shapes** to the `_detect_indirect_commit` dispatcher:
   - Match `<<\s*['"]?(\w+)['"]?\s*\n([\s\S]*?)\n\1\b` (heredoc body)
   - Match `<<<\s*['"]?([^'"\n]+)['"]?` (here-string body)
   - Apply same inner-payload regex (`\bgit\b[^;&|]*?\bcommit\b`) to the extracted body
2. **Extend `_INTERPRETERS` regex** to include `dash`, `ksh`, `mksh`, `pdksh` (per-OS variants vary; conservative is dash+ksh):
   - `_INTERPRETERS = (?:bash|sh|zsh|dash|ksh)`
3. **Relax `_SCRIPT_INVOKE_RE` extension check** to be advisory rather than gating:
   - Read any script file matching the invocation regardless of extension
   - Keep the 256 KiB cap + fail-open OSError discipline
   - OR: detect `\bbash\s+(\S+)` more broadly and attempt content read with same safety contract
4. **`eval` shape** — most complex; the eval argument might be quoted/unquoted/string-substituted. Conservative option: detect `\beval\s+['"]?` and apply inner-payload regex to the rest of the segment. Acknowledge edge cases (variable-substituted eval string, multi-line eval) as documented punt.

Hook-level changes are constrained to the 4 existing helper functions + the `_INTERPRETERS` constant + one new heredoc/here-string extractor. Test additions: ~10-15 new test cases (5 BLOCK + 5 ALLOW pairs for the new shapes).

## Acceptance criteria

- [ ] `bash <<EOF\ngit commit ...\nEOF` → BLOCKED with clear message naming "heredoc" shape
- [ ] `bash <<<'git commit ...'` → BLOCKED naming "here-string" shape
- [ ] `eval 'git commit -c user.name=X commit -m foo'` → BLOCKED naming "eval" shape
- [ ] `eval 'echo hello'` (no git commit) → ALLOWED
- [ ] `dash -c 'git commit ...'` → BLOCKED naming "dash -c" shape
- [ ] `ksh -c 'git commit ...'` → BLOCKED naming "ksh -c" shape
- [ ] `bash script.bash` (containing git commit) → BLOCKED (extension-agnostic content read)
- [ ] `bash extensionless-script` (containing git commit) → BLOCKED
- [ ] All existing 46 tests continue to pass
- [ ] 256 KiB cap + fail-open OSError discipline preserved
- [ ] Test count claim verified by reviewer (X passed, 0 failed; Y new for this PR)

## Wave/severity

Tech-debt — failure direction is silent-bypass (not over-block), but the named shapes in PR #479 already close the most-common attack vectors. #5 (`.sh`-rename bypass) is the highest-severity remaining gap since it's operator-discoverable via the docstring. W12 sweep wave is acceptable; W11 candidate if there's bandwidth.

## Origin

- Aisha Idrissi primary verdict on PR #479: 3 gaps flagged (#1 heredoc, #4 ksh/dash, #5 `.sh`-suffix)
- Petra Vidović secondary verdict on PR #479: 5-row sweep-candidate table (heredoc, here-string, eval, source `.`, command substitution `$(echo git) commit`)
- Convergence on heredoc + extension-suffix + eval/dash family = high signal for follow-up

Both reviewer comments are linked from PR #479 comments (4473328393 Aisha, 4473328384 Petra).

parametrization:
Requestor: Aino Virtanen
Requestee: Aino Virtanen
RequestOrReplied:
TechDebt: None

Closes #482.

## Summary

Direct follow-up to PR #479 (which closed #475). Aisha + Petra's convergent review of #479 surfaced 5 additional indirect-exec bypass shapes I explicitly punted in #475's scope. This PR closes all 5 with extensions to `_detect_indirect_commit` — no breaking changes, strictly additive.

The highest-severity gap was the script extension restriction (`\.sh\b`) — operator-discoverable via the hook docstring and trivially circumvented by renaming the bypass script. Now extension-agnostic.

## Shapes added

| # | Shape | Detector |
|---|---|---|
| 1 | `bash <<EOF\ngit commit ...\nEOF` | new `_HEREDOC_RE` (4 opener variants) |
| 2 | `bash <<<'git commit ...'` | new `_HERESTRING_RE` |
| 3 | `eval 'git commit ...'` | new `_EVAL_RE` |
| 4 | `dash -c '...'` / `ksh -c '...'` | `_INTERPRETERS` widened to `(?:bash\|sh\|zsh\|dash\|ksh)` |
| 5 | `bash script.bash` / `bash <extensionless>` | `_SCRIPT_INVOKE_RE` relaxed extension; content read at any extension |

The `_INTERPRETERS` widening (#4) automatically extends the 4 pre-existing detectors plus the 2 new heredoc/here-string detectors — `printf '...' | dash`, `dash <(...)`, `ksh <script>` etc. all now match.

## Why the heredoc detector is anchored on `_INTERPRETERS`

`cat <<EOF\ngit commit is a phrase\nEOF` is data-only — cat prints the body, does not execute it. Without the `_INTERPRETERS` anchor we'd false-block every documentation heredoc mentioning git commit. The detector specifically requires the heredoc target to be a known shell interpreter. Same anchor discipline for `_HERESTRING_RE`.

`_EVAL_RE` deliberately does NOT anchor on `_INTERPRETERS` — `eval` is a shell builtin available regardless of which shell is invoking the command.

## Documented punts (recorded inline at regex comments)

- **Variable-substituted eval** (`eval "$cmd"`): substitution happens at shell-runtime, after the hook fires. Cannot inspect. Test `test_eval_variable_substitution_documented_punt` pins ALLOW.
- **`mksh`/`pdksh`**: excluded from `_INTERPRETERS` as conservative bound — rare in modern installs, easy to extend later if observed in any bypass surface.
- **Nested process substitution with parens** (`bash <($(echo git) commit)`): existing #475 limitation; the surrounding context almost always triggers one of the other detectors anyway.

## Test plan

- [x] **27 new tests** in `IndirectExecExtensionTests` class. All BLOCK/ALLOW pairs explicit; helper `_assert_blocked_with_shape(result, shape_label)` validates both the indirect-exec gate AND the specific shape attribution in the block message.
- [x] **Acceptance bullet coverage from #482**:
  - `bash <<EOF\ngit commit ...\nEOF` → BLOCKED naming "heredoc" — `test_bash_heredoc_with_git_commit_blocks` (+ 2 sibling tests for `<<'EOF'` and `<<-EOF` variants)
  - `bash <<<'git commit ...'` → BLOCKED naming "here-string" — `test_bash_herestring_with_git_commit_blocks` (+ double-quoted variant)
  - `eval 'git commit -c user.name=X commit -m foo'` → BLOCKED naming "eval" — `test_eval_with_git_commit_blocks`
  - `eval 'echo hello'` → ALLOWED — `test_eval_no_git_commit_allows`
  - `dash -c 'git commit ...'` → BLOCKED naming "shell -c" — `test_dash_dash_c_with_git_commit_blocks`
  - `ksh -c 'git commit ...'` → BLOCKED naming "shell -c" — `test_ksh_dash_c_with_git_commit_blocks`
  - `bash script.bash` (containing git commit) → BLOCKED naming "shell script" — `test_bash_script_dot_bash_extension_blocks`
  - `bash extensionless-script` (containing git commit) → BLOCKED — `test_bash_script_extensionless_blocks`
- [x] **Defensive coverage** beyond acceptance: cat-heredoc-mentioning-commit ALLOWED (false-positive guard), cat-herestring-mentioning-commit ALLOWED, `bash -c` does NOT false-match as script-path (regression guard for the `-` exclusion in script regex), 256 KiB cap preserved (fail-open on large scripts), unified block message cites both #475 + #482 and lists new shapes inline.
- [x] **All existing 46 tests continue to pass** — strictly additive change to detector dispatcher.
- [x] **Full hooks suite**: 856 passed across `.claude/hooks/tests/` — no sibling-hook regression.
- [x] **256 KiB cap + fail-open OSError discipline preserved** — verified by `test_bash_script_with_size_cap_still_protected`.

```
$ ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/test_validate_commit_identity.py
  73 passed in 0.15s   (46 pre-existing + 27 new)

  856 passed in 1.37s

$ uvx ruff@latest check .claude/hooks/      → All checks passed!
$ uvx ruff@latest format --check .claude/hooks/ → 64 files already formatted
```

## Coordination notes

- Charter `pull-requests.md` § Origin > Local Clone followed for review reference.
- Charter `state-claims.md` § Refresh State Before Claim followed (read-back-verified labels + project board membership before PR close).
- Per memory `feedback_techdebt_literal_line_not_section`, the bare `TechDebt: None` literal line at the top of this PR body is what Hook 4 parses; no `## Tech Debt` markdown section.
- Per memory `feedback_heredoc_in_git_commit`, commit message was written to `/tmp/commit-msg-aino-482.txt` via Write tool and passed via `-F` (not heredoc inline).
- Per memory `feedback_actionlint_needs_shellcheck`: no workflow changes in this PR; `actionlint`/`shellcheck` are both on PATH and verified at session start but not exercised here.

## Related

- `feedback_consumer_against_in_flight_upstream` does NOT apply — this PR has no upstream dependency; #475's PR #479 already merged.
- The cross-domain `env`-wrapper bypass I flagged on PR #483's review (commented on #482 as drift evidence) is NOT closed by this PR. The Aisha+Petra original scope of #482 covers 5 shapes; the env-wrapper finding is row 6 added later via comment. Should be picked up as a follow-up issue or rolled into a Phase B sweep. Mentioned here so the reviewer sees the full open surface and doesn't expect this PR to close the env-wrapper too.

### DIFF
--- a/.claude/hooks/tests/test_validate_commit_identity.py
+++ b/.claude/hooks/tests/test_validate_commit_identity.py
@@ -619,6 +619,277 @@ class IndirectExecBypassTests(unittest.TestCase):
         self.assertIn("indirect-exec", result["reason"])
 
 
+class IndirectExecExtensionTests(unittest.TestCase):
+    """Issue #482: extend indirect-exec detection from the original 4 shapes
+    (#475) to 5 more — heredoc, here-string, eval, dash/ksh -c (via the
+    widened `_INTERPRETERS` alternation), and extension-agnostic script
+    read (the highest-severity #482 change, since operators could trivially
+    rename a bypass script to circumvent the prior `.sh\\b` restriction).
+
+    Each new shape gets a paired BLOCK / ALLOW test to pin both halves of
+    the detection contract:
+      - BLOCK: inner payload contains git-commit-shape
+      - ALLOW: same wrapper shape but inner payload is innocent
+
+    All existing 46 tests in this file continue to pass — the extensions
+    are strictly additive to `_detect_indirect_commit`'s dispatcher.
+    """
+
+    @staticmethod
+    def _input(command: str, cwd: str | None = None) -> dict:
+        d: dict = {"tool_name": "Bash", "tool_input": {"command": command}}
+        if cwd is not None:
+            d["cwd"] = cwd
+        return d
+
+    def _assert_blocked_with_shape(self, result, shape_label: str) -> None:
+        """Assert the result is an indirect-exec block and the reason
+        contains the expected shape label (e.g. 'heredoc', 'here-string').
+        Also asserts the unified #475/#482 cite is present."""
+        self.assertIsNotNone(result, f"indirect-exec bypass must block ({shape_label})")
+        assert result is not None
+        self.assertEqual(result["decision"], "block")
+        self.assertIn("indirect-exec", result["reason"])
+        self.assertIn(shape_label, result["reason"])
+        # Both citations should be present in the unified post-#482 block message.
+        self.assertIn("#475", result["reason"])
+        self.assertIn("#482", result["reason"])
+
+    # --- heredoc shape -----------------------------------------------------
+
+    def test_bash_heredoc_with_git_commit_blocks(self):
+        """POS: `bash <<EOF\\ngit commit ...\\nEOF` — heredoc body fed as
+        stdin to a shell. Without #482 detection the outer tokenizer sees
+        only `['bash', '<<EOF', '...', 'EOF']`-style decomposition and
+        `_find_commit_segment` returns None → silent allow."""
+        cmd = 'bash <<EOF\ngit -c user.name="X" -c user.email="Y" commit -F /tmp/msg.txt\nEOF'
+        self._assert_blocked_with_shape(hook.check(self._input(cmd)), "heredoc")
+
+    def test_bash_heredoc_quoted_delim_with_git_commit_blocks(self):
+        """POS: `bash <<'EOF'...` — quoted-delim heredoc (no var expansion).
+        Same exec semantics; detector must match this variant too."""
+        cmd = 'bash <<\'EOF\'\ngit -c user.name="X" -c user.email="Y" commit -m z\nEOF'
+        self._assert_blocked_with_shape(hook.check(self._input(cmd)), "heredoc")
+
+    def test_bash_heredoc_dash_form_with_git_commit_blocks(self):
+        """POS: `bash <<-EOF...` — tab-stripping heredoc form."""
+        cmd = "bash <<-EOF\n\tgit -c user.name=X -c user.email=Y commit -m z\n\tEOF"
+        self._assert_blocked_with_shape(hook.check(self._input(cmd)), "heredoc")
+
+    def test_bash_heredoc_no_git_commit_allows(self):
+        """NEG: `bash <<EOF\\necho hi\\nEOF` — innocent heredoc body."""
+        cmd = "bash <<EOF\necho hello\nls -la\nEOF"
+        self.assertIsNone(hook.check(self._input(cmd)))
+
+    def test_cat_heredoc_mentioning_git_commit_allows(self):
+        """NEG: `cat <<EOF\\ngit commit is a phrase\\nEOF` — `cat` is NOT
+        a shell interpreter so the heredoc body is data, not code. Must
+        not false-block. This is the principal false-positive guard for
+        heredoc detection — prose containing 'git commit' inside a doc
+        being piped to `cat` (or `tee`, `bat`, etc.) is everywhere."""
+        cmd = "cat <<EOF > /tmp/doc.md\nHow to git commit:\ngit commit -m foo\nEOF"
+        self.assertIsNone(hook.check(self._input(cmd)))
+
+    # --- here-string shape -------------------------------------------------
+
+    def test_bash_herestring_with_git_commit_blocks(self):
+        """POS: `bash <<<'git commit ...'` — here-string body."""
+        cmd = "bash <<<'git -c user.name=X -c user.email=Y commit -m z'"
+        self._assert_blocked_with_shape(hook.check(self._input(cmd)), "here-string")
+
+    def test_bash_herestring_double_quoted_with_git_commit_blocks(self):
+        """POS: `bash <<<"..."` — double-quoted here-string."""
+        cmd = 'bash <<<"git -c user.name=X -c user.email=Y commit -m z"'
+        self._assert_blocked_with_shape(hook.check(self._input(cmd)), "here-string")
+
+    def test_bash_herestring_no_git_commit_allows(self):
+        """NEG: `bash <<<'echo hi'` — innocent here-string."""
+        self.assertIsNone(hook.check(self._input("bash <<<'echo hello'")))
+
+    def test_cat_herestring_mentioning_git_commit_allows(self):
+        """NEG: `cat <<<'git commit is a phrase'` — `cat` is not a shell;
+        body is data not code. Must not false-block."""
+        self.assertIsNone(hook.check(self._input("cat <<<'git commit is a phrase'")))
+
+    # --- eval shape --------------------------------------------------------
+
+    def test_eval_with_git_commit_blocks(self):
+        """POS: `eval 'git commit ...'` — shell builtin re-parses and
+        executes its argument. Outer tokenizer sees `eval` as argv[0]
+        and skips identity validation."""
+        cmd = "eval 'git -c user.name=X -c user.email=Y commit -m foo'"
+        self._assert_blocked_with_shape(hook.check(self._input(cmd)), "eval")
+
+    def test_eval_double_quoted_payload_blocks(self):
+        """POS: `eval "git commit ..."` — double-quoted payload form."""
+        cmd = 'eval "git -c user.name=X -c user.email=Y commit -m foo"'
+        self._assert_blocked_with_shape(hook.check(self._input(cmd)), "eval")
+
+    def test_eval_no_git_commit_allows(self):
+        """NEG: `eval 'echo hello'` — innocent eval."""
+        self.assertIsNone(hook.check(self._input("eval 'echo hello'")))
+
+    def test_eval_variable_substitution_documented_punt(self):
+        """NEG: `eval "$CMD"` — variable-substituted eval body. Documented
+        punt per the hook docstring: substitution happens at shell-runtime,
+        after the hook fires, so we can't inspect the actual command. Must
+        ALLOW (otherwise every legitimate `eval $cmd` use would false-block)."""
+        self.assertIsNone(hook.check(self._input('eval "$CMD"')))
+        self.assertIsNone(hook.check(self._input("eval $cmd")))
+
+    def test_eval_no_quotes_no_commit_allows(self):
+        """NEG: `eval echo hello` — bareword eval, no commit shape."""
+        self.assertIsNone(hook.check(self._input("eval echo hello")))
+
+    # --- dash / ksh extended interpreter alternation -----------------------
+
+    def test_dash_dash_c_with_git_commit_blocks(self):
+        """POS: `dash -c 'git commit ...'` — `dash` is a POSIX shell.
+        Pre-#482 `_INTERPRETERS` was `(?:bash|sh|zsh)` and missed dash;
+        #482 extended to `(?:bash|sh|zsh|dash|ksh)`."""
+        cmd = "dash -c 'git -c user.name=X -c user.email=Y commit -m z'"
+        self._assert_blocked_with_shape(hook.check(self._input(cmd)), "shell -c")
+
+    def test_ksh_dash_c_with_git_commit_blocks(self):
+        """POS: `ksh -c 'git commit ...'` — Korn shell, also POSIX."""
+        cmd = "ksh -c 'git -c user.name=X -c user.email=Y commit -m z'"
+        self._assert_blocked_with_shape(hook.check(self._input(cmd)), "shell -c")
+
+    def test_dash_dash_c_no_git_commit_allows(self):
+        """NEG: `dash -c 'ls -la'` — innocent dash invocation."""
+        self.assertIsNone(hook.check(self._input("dash -c 'ls -la'")))
+
+    def test_dash_piped_with_git_commit_blocks(self):
+        """POS: `printf '...' | dash` — pipe-to-shell with dash interpreter."""
+        cmd = "printf 'git -c user.name=X -c user.email=Y commit -m z\\n' | dash"
+        self._assert_blocked_with_shape(hook.check(self._input(cmd)), "printf/echo piped to shell")
+
+    def test_ksh_piped_with_git_commit_blocks(self):
+        """POS: `echo '...' | ksh` — same with ksh."""
+        cmd = "echo 'git -c user.name=X -c user.email=Y commit -m z' | ksh"
+        self._assert_blocked_with_shape(hook.check(self._input(cmd)), "printf/echo piped to shell")
+
+    # --- extension-agnostic script read (highest-severity #482 fix) -------
+
+    def test_bash_script_dot_bash_extension_blocks(self):
+        """POS: `bash /tmp/script.bash` — .bash extension was NOT covered
+        by the pre-#482 `\\.sh\\b` restriction. Now extension-agnostic,
+        any readable file under the 256 KiB cap is inspected."""
+        with tempfile.NamedTemporaryFile(
+            mode="w", suffix=".bash", delete=False, encoding="utf-8"
+        ) as f:
+            f.write('git -c user.name="X" -c user.email="Y" commit -F /tmp/msg.txt\n')
+            script_path = f.name
+        try:
+            self._assert_blocked_with_shape(
+                hook.check(self._input(f"bash {script_path}")),
+                "shell script",
+            )
+        finally:
+            Path(script_path).unlink(missing_ok=True)
+
+    def test_bash_script_extensionless_blocks(self):
+        """POS: `bash /tmp/script-no-ext` — completely extensionless. The
+        rename-bypass that was the highest-severity #482 concern: an
+        operator who knew the hook checked `.sh` could rename their
+        bypass script to `myscript` and slip through. Now blocked."""
+        with tempfile.NamedTemporaryFile(mode="w", suffix="", delete=False, encoding="utf-8") as f:
+            f.write('git -c user.name="X" -c user.email="Y" commit -F /tmp/msg.txt\n')
+            script_path = f.name
+        try:
+            self._assert_blocked_with_shape(
+                hook.check(self._input(f"bash {script_path}")),
+                "shell script",
+            )
+        finally:
+            Path(script_path).unlink(missing_ok=True)
+
+    def test_bash_script_unusual_extension_blocks(self):
+        """POS: `bash /tmp/script.ksh` — unusual extension. Content read
+        is unconditional on extension; the cap + fail-open OSError discipline
+        is preserved from the prior `.sh\\b`-restricted detector."""
+        with tempfile.NamedTemporaryFile(
+            mode="w", suffix=".ksh", delete=False, encoding="utf-8"
+        ) as f:
+            f.write("git -c user.name=X -c user.email=Y commit -m z\n")
+            script_path = f.name
+        try:
+            self._assert_blocked_with_shape(
+                hook.check(self._input(f"bash {script_path}")),
+                "shell script",
+            )
+        finally:
+            Path(script_path).unlink(missing_ok=True)
+
+    def test_bash_extensionless_script_no_git_commit_allows(self):
+        """NEG: `bash /tmp/script` where content has NO git commit.
+        Detection should still try to read (extension-agnostic), find
+        nothing matching, and allow."""
+        with tempfile.NamedTemporaryFile(mode="w", suffix="", delete=False, encoding="utf-8") as f:
+            f.write("#!/usr/bin/env bash\nls -la\necho done\n")
+            script_path = f.name
+        try:
+            self.assertIsNone(hook.check(self._input(f"bash {script_path}")))
+        finally:
+            Path(script_path).unlink(missing_ok=True)
+
+    def test_bash_script_dash_c_does_not_false_match_as_script(self):
+        """NEG/regression: `bash -c '...'` must continue to match as
+        `shell -c`, NOT as a script-path read. The script regex's leading
+        char class excludes `-`, so `-c` is correctly skipped."""
+        cmd = "bash -c 'git -c user.name=X -c user.email=Y commit -m foo'"
+        result = hook.check(self._input(cmd))
+        assert result is not None
+        # Should be matched by the shell -c detector, not the script detector
+        self.assertIn("shell -c", result["reason"])
+        # Should NOT have script-shape language in the reason
+        self.assertNotIn("shell script", result["reason"])
+
+    def test_bash_script_with_size_cap_still_protected(self):
+        """NEG (defensive): a script file >256 KiB is NOT inspected
+        (preserves the original #475 fail-open-on-large-file discipline).
+        If the script content contains a git commit shape but the file
+        exceeds the cap, the detector returns None and the rest of the
+        hook handles it normally (ALLOW since outer-command tokenizer
+        sees `bash <path>`, not git)."""
+        with tempfile.NamedTemporaryFile(
+            mode="w", suffix=".bash", delete=False, encoding="utf-8"
+        ) as f:
+            # Write >256 KiB of padding followed by git commit text
+            f.write("# padding\n" * 30000)  # ~300 KiB
+            f.write('git -c user.name="X" -c user.email="Y" commit -m z\n')
+            script_path = f.name
+        try:
+            self.assertIsNone(
+                hook.check(self._input(f"bash {script_path}")),
+                "files >256 KiB must not be inspected (preserves fail-open discipline)",
+            )
+        finally:
+            Path(script_path).unlink(missing_ok=True)
+
+    # --- cross-shape coherence --------------------------------------------
+
+    def test_unified_block_message_cites_both_issues(self):
+        """POS: the post-#482 block message MUST cite both #475 (original
+        4-shape work) AND #482 (5-shape extension) so operators landing
+        on the message can trace the full bypass surface."""
+        result = hook.check(self._input("eval 'git -c user.name=X commit -m z'"))
+        assert result is not None
+        self.assertIn("#475", result["reason"])
+        self.assertIn("#482", result["reason"])
+
+    def test_block_message_lists_new_shapes(self):
+        """POS: the wrapper-list in the block message MUST include the
+        new shape examples so the operator sees them inline. Helps prevent
+        operators from re-discovering bypass patterns by trial."""
+        result = hook.check(self._input("eval 'git -c user.name=X commit -m z'"))
+        assert result is not None
+        # New #482 shapes named in the unified message
+        self.assertIn("<<EOF", result["reason"])
+        self.assertIn("<<<", result["reason"])
+        self.assertIn("eval", result["reason"])
+
+
 class CwdFallbackTests(unittest.TestCase):
     """Issue #475 fix 1: when no literal `cd <path>` prefix is present, fall
     back to the tool-call cwd for roster resolution. Lets operators commit
--- a/.claude/hooks/validate_commit_identity.py
+++ b/.claude/hooks/validate_commit_identity.py
@@ -22,14 +22,28 @@ cwd-fallback for cross-repo detection (#475 fix 1):
   every operator to compose a `cd` prefix when their shell cwd already names
   the right repo.
 
-Indirect-exec bypass detection (#475 fix 2):
+Indirect-exec bypass detection (#475 fix 2, extended in #482):
   PreToolUse hooks only see the literal Bash `command` parameter. Wrappers
-  like `printf '<git-commit-cmd>' | bash`, `bash -c '<git-commit-cmd>'`,
-  `bash <script-with-git-commit>`, and `bash <(echo '<git-commit-cmd>')`
-  hide the actual `git commit` from the outer-command tokenizer. The hook
-  now scans for these shapes and, if the payload looks like a git commit,
-  BLOCKS with a message pointing operators to the canonical shape. This
-  closes a silent bypass discovered in P3W11 PR #41 fixup.
+  that hide the actual `git commit` from the outer-command tokenizer are
+  detected pre-tokenize and BLOCKED. Original 4 shapes from #475:
+      printf '<git-commit-cmd>' | bash
+      bash -c '<git-commit-cmd>'
+      bash <script-with-git-commit>      (extension-restricted to *.sh)
+      bash <(echo '<git-commit-cmd>')
+
+  #482 follow-up adds 5 more, surfaced by Aisha + Petra convergent review:
+      bash <<EOF...git commit...EOF      (heredoc body)
+      bash <<<'git commit ...'           (here-string)
+      eval 'git commit ...'              (shell builtin)
+      dash|ksh -c '...'                  (extended _INTERPRETERS alternation)
+      bash <script.bash>                 (extension-agnostic; *.sh restriction removed)
+
+  The extension-agnostic script read is the highest-severity #482 change:
+  the prior `.sh` restriction was trivially circumvented by renaming the
+  bypass script. Now any token after `bash|sh|zsh|dash|ksh` that points to
+  a readable regular file (under the 256 KiB cap, fail-open on OSError)
+  has its content scanned for git-commit-shape. Token must not start with
+  a flag (`-c`, `-x`, etc.) to avoid colliding with the `-c` detector.
 
 Input Language
 ==============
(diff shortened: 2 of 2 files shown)


parametrization:
Requestor: Petra Vidović
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: None blocking — see "Open follow-up surface" below for two still-uncovered shapes from my original #479 sweep (`source ./script.sh` / `. ./script.sh` and `$(echo git) commit ...`) and Aino's own env-wrapper drift-evidence (row 6 on #482) — none blocking #493, all worth tracking as W12-or-Phase-B sweep follow-ups.

## Tech-lead verdict — Approved (secondary)

Reviewed at head `b3d1d104a2de9d301a74323fb29ccf0876642242` against issue #482, reading both touched files via `gh api repos/.../contents/<path>?ref=<sha>` AND independently exercising the new regexes in a local Python harness to verify the load-bearing claims. PR body framing matches code at HEAD. CI 4/4 green.

Aisha Idrissi has the primary review. My secondary is independent — no coordination with Aisha. Most of my secondary value comes from being the co-flagger of the 5-shape sweep table on #479 alongside her, so the convergent verification of my OWN earlier surface enumeration against Aino's implementation is the natural shape.

### What I verified

**1. Detector dispatch order — pattern-only first, disk-I/O last**

At `validate_commit_identity.py:379-411` the dispatcher runs detectors in this order:

| # | Detector | I/O? |
|---|---|---|
| 1 | `_PIPE_TO_SHELL_RE` (printf/echo \| shell) | regex-only |
| 2 | `_DASH_C_RE` (shell -c) | regex-only |
| 3 | `_PROCESS_SUB_RE` (shell <(...)) | regex-only |
| 4 | `_HEREDOC_RE` (shell <<DELIM) | regex-only |
| 5 | `_HERESTRING_RE` (shell <<<) | regex-only |
| 6 | `_EVAL_RE` (eval) | regex-only |
| 7 | `_SCRIPT_INVOKE_RE` + `_read_script_if_safe` | **DISK I/O** |

Code-comment at `validate_commit_identity.py:375-377` explicitly documents: "The script-path check is LAST because it requires disk I/O — all pattern-only checks run first to short-circuit common cases without hitting the filesystem." Each detector is its own `for ... return` block, so on a regex match in any earlier detector the function returns immediately without touching disk. ✓

Crucially, I verified empirically: `bash <<EOF\ngit commit\nEOF` matches detector #4 (`_HEREDOC_RE`) and returns BEFORE detector #7 attempts any file open. The heredoc-detector short-circuit prevents unnecessary I/O.

**2. Test count claim — 73 passed (46 pre-existing + 27 new)**

Counted directly at HEAD:
- Total `def test_` in file: **73** ✓
- `IndirectExecExtensionTests` class members: **27** ✓ (matches PR body)

Checked for flaky-skipped / expectedFailure / xfail markers added in the new class: **0**. The 5 `self.skipTest("local roster is empty")` calls in the file are all in pre-existing test classes at lines 260, 283, 324, 417, 585 — none in `IndirectExecExtensionTests` (starts at line 622). Test discipline holds.

The PR body claim of 856 hooks-suite tests passing is consistent with CI 4/4 green at HEAD; I'm not re-running the suite locally — the CI smoke gate is the canonical signal.

**3. 256 KiB cap + fail-open OSError discipline preserved**

Verified at `validate_commit_identity.py:344-356`:

```python
def _read_script_if_safe(script_path, cwd):
    try:
        p = Path(script_path)
        if not p.is_absolute() and cwd:
            p = (Path(cwd) / p).resolve()
        else:
            p = p.resolve()
        if not p.is_file():
            return None
        if p.stat().st_size > 256 * 1024:    # ← cap preserved verbatim
            return None
        return p.read_text(encoding="utf-8", errors="replace")
    except (OSError, ValueError):           # ← fail-open preserved verbatim
        return None
```

Identical structure to pre-#482 implementation. Pinned by `test_bash_script_with_size_cap_still_protected` (test_validate_commit_identity.py:848-868) which writes a >256 KiB file with a real `git commit` shape and asserts the hook returns None (i.e., size cap > content scan, fail-open semantics preserved).

**4. My verdict comment's literal-line block (charter format)**

Per `feedback_techdebt_literal_line_not_section`, my verdict comment's first 4 lines are bare-literal structured-fields lines NOT a `## TechDebt` markdown section. Hook 4's regex `\*{0,2}TechDebt:\*{0,2}\s*(.+)` (validate_pr_review.py:361) accepts the verbose TechDebt value pinning the open-follow-up surface.

Note on Aino's PR BODY: the top of the PR body has `Requestor: Aino Virtanen / Requestee: Aino Virtanen / RequestOrReplied: / TechDebt: None`. PR body is NOT what Hook 4 parses for verdicts — verdict comments are. So this is purely PR-body cosmetics and not a Hook 4 concern. Mentioning for completeness since the brief asked.

**5. My 2 missing shapes (source `.` / `$(echo git) commit`)**

Verified independently at HEAD that neither is detected:
- `source ./script.sh`, `. ./script.sh` — `_SCRIPT_INVOKE_RE` requires `_INTERPRETERS` prefix (`bash|sh|zsh|dash|ksh`), which excludes `source` and `.`. Confirmed via my local harness.
- `$(echo git) commit -m x` — command substitution to construct the `git` token at shell-runtime; no detector matches (heredoc, here-string, eval, script all return None).

These are NOT in #482's acceptance criteria — issue body lists shapes 1-5 (heredoc, here-string, eval, dash/ksh, .sh-rename) plus comment row 6 (env wrapper). My original #479 sweep table listed 5 shapes — Aino implemented 3 of mine (heredoc, here-string, eval), the dash/ksh + `.sh`-rename pair from Aisha's verdict, and explicitly punted my remaining 2 (source/dot, command-sub).

Endorse the punt for this PR. Both are real evasion paths but warrant their own scope-disciplined fix:
- **source/dot family** would extend `_SCRIPT_INVOKE_RE` to a new detector accepting `source <path>` / `. <path>` patterns. Same disk-I/O semantics as #7 but new dispatch entry. Conservative `source` token-equality match (reject `sourcemaps`, `sourcing-tool`).
- **command substitution `$(echo git) commit`** is the trickiest — `$(...)` evaluates at shell-runtime. Best-effort detection would require executing the command substitution in a sandbox (rejected per documented `eval "$CMD"` punt rationale at validate_commit_identity.py:291-301). Realistic stance: deliberately documented as a known limitation in the hook docstring, possibly with a charter-tier note pointing operators at the canonical `cd ... && git -c ... commit ...` shape.

**Recommend filing post-merge** as `tech-debt(hook): validate_commit_identity — source/dot + command-substitution shapes (#482 follow-up)`. NOT blocking #493.

### Open follow-up surface (full enumeration for TechDebt visibility)

Three categories of still-open bypass shapes after #493 merges, all NON-BLOCKING for this PR:

1. **`source <path>` / `. <path>`** — file-execution wrappers that hit disk like script-invocation but bypass the `_INTERPRETERS` prefix requirement. Same severity class as the .sh-rename fix that drove #482.
2. **`$(echo git) commit ...`** — command substitution. Documented punt territory; pair with the existing `eval "$CMD"` punt in docstring + operator-facing block message.
3. **`env [VAR=val ...] git commit ...`** (row 6 on #482, Aino's drift-evidence from PR #483 cross-domain finding) — `env` becomes argv[0]; same silent-bypass class. Aino's PR body explicitly notes "Should be picked up as a follow-up issue or rolled into a Phase B sweep. Mentioned here so the reviewer sees the full open surface."

Recommend Aino files the source/dot+command-sub follow-up post-merge, and the env-wrapper either gets its own issue OR rolls into a Phase B `validate_commit_identity-bypass-sweep-3` if there's a consolidation appetite.

### Cross-shape coherence — well-tested

Two cross-cutting tests pin operator-facing behavior I care about as a tech-lead:
- `test_unified_block_message_cites_both_issues` (test:872-879) — block message names both #475 + #482 so operators can trace the full bypass-surface trail. Good discipline.
- `test_block_message_lists_new_shapes` (test:881-890) — block message includes inline `<<EOF`, `<<<`, `eval` mentions so the operator's first encounter with the block tells them what NOT to do next. Prevents trial-and-error rediscovery of bypass patterns.

### Cross-hook interaction sanity

I'm independent of Aisha but verified the changes don't perturb any cross-hook contract:
- `_INTERPRETERS` widening to include `dash|ksh` does NOT affect any other hook (this constant is local to `validate_commit_identity.py`)
- The new `_HEREDOC_RE` deliberately diverges from `_shell_parse.strip_heredocs` — code-comment at validate_commit_identity.py:270-275 explains the design call (one strips heredoc bodies for prose-tolerant parsing; one extracts them for hidden-payload scanning). Both can coexist; no shared-helper convergence opportunity. Justified divergence.
- Hook 4 / Hook 15 / Hook 21 unaffected — no shared regex constant changed, no module-level export changed.

### Conclusion

Tight, well-tested, scope-disciplined extension to the #475 surface. All 5 acceptance-bullet shapes from #482 covered. Plus 2 explicit punts (`eval "$CMD"` variable substitution, mksh/pdksh) documented in code with rationale. Test coverage has BLOCK + ALLOW pairs for every new shape, plus false-positive guards (cat-heredoc, cat-herestring) and a regression guard (`bash -c` not false-matching as script-path). 256 KiB cap + fail-open preserved verbatim.

The two still-uncovered shapes from my original #479 sweep (source/dot, command substitution) plus Aino's own env-wrapper drift-evidence are tracked in TechDebt for post-merge follow-up filing. None block #493.

Aisha's primary verdict completes the 2-of-2 Approved gate.

parametrization:
Requestor: Aisha Idrissi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: None — env-wrapper bypass surface (the cross-domain row-6 I flagged on #483 review) is correctly out-of-scope per #482's 5-shape charter; Aino flagged it explicitly in the PR body's Related section as follow-up material. Two micro-observations below (eval-with-`echo` false-positive in same-segment text, eval-bareword limitation) are documented-punt cousins worth a one-line docstring note if convenient, not gating.

## Primary verdict — Approved

**Reviewer:** Aisha Idrissi (deploy SRE) — **primary** (per spawn brief: deepest context as #479 reviewer who co-surfaced 3 of the 5 shapes with Petra)
**Secondary reviewer:** Petra Vidović — independent in-flight
**Head verified:** `b3d1d104a2de9d301a74323fb29ccf0876642242` via `gh api .../contents/...?ref=b3d1d10` — origin-canonical, no parent-checkout branch switch (per [[feedback_reviewer_must_not_branch_switch_parent]])
**Mergeable:** `UNKNOWN` per `gh pr view` (likely still computing post-CI; CI is green and base is main — not a gate)
**Scope:** exactly 2 files (`.claude/hooks/validate_commit_identity.py` +110/-20, `.claude/hooks/tests/test_validate_commit_identity.py` +271/-0). Matches PR body claim. No incidental changes.

### Verification methodology

1. Downloaded PR-HEAD source + tests via `gh api contents?ref=b3d1d10` into `/tmp/aisha_482_runtree/` (with supporting `_shell_parse.py` + `annunaki_log.py` + `roster.json` from main).
2. Ran full PR-HEAD test suite: **73 passed, 0 failed** — matches Aino's PR-body claim exactly (46 pre-existing + 27 new in `IndirectExecExtensionTests`).
3. Hand-traced each of the 6 detector regexes + the dispatcher's ordering at line 359-411.
4. Ran 14 adversarial probes targeting edges NOT in the PR test suite (results below).
5. Confirmed `#482` issue still OPEN with `tech-debt`, `p3-wave-11` labels (closes correctly via `Closes #482.` in PR body).

### Each of my 5 #479-surfaced shapes — verdict

**1. Heredoc detection** — `_HEREDOC_RE` (lines 276-280)

- `\b` + `_INTERPRETERS` + `\b` anchor correctly excludes `cat <<EOF`, `tee <<EOF`, `python3 <<EOF`. Probe 12 confirmed all three ALLOW.
- 4 opener variants covered: `<<DELIM`, `<<-DELIM`, `<<'DELIM'`, `<<"DELIM"`. The regex `<<-?\s*['\"]?(?P<delim>\w+)['\"]?` handles all four.
- `\n\t*(?P=delim)\b` correctly handles the tab-indented closer for `<<-` form (covered by `test_bash_heredoc_dash_form_with_git_commit_blocks`).
- `test_cat_heredoc_mentioning_git_commit_allows` (line 684-691) is a real assertion (`self.assertIsNone(hook.check(...))` not a smoke test) with realistic input (`cat <<EOF > /tmp/doc.md\nHow to git commit:\ngit commit -m foo\nEOF`).
- Probe 1 + 2 + 3 + 14 extended coverage: custom delimiters (`DONE`, `STOP_X9`), heredoc nested in compound (`cd && bash <<EOF`), flags/redirects between `bash` and `<<` (`bash -x <<EOF`, `bash 2>/dev/null <<EOF`), heredoc after compound prefix (`echo && bash <<EOF`) — all BLOCK correctly. The `[^\n]*?` between `_INTERPRETERS\b` and `<<-?` tolerates intervening tokens, which is the correct behavior.

**2. Here-string detection** — `_HERESTRING_RE` (lines 285-288)

- Same anchor discipline as heredoc. `test_cat_herestring_mentioning_git_commit_allows` is a real assertion (line 709-712) with realistic input.
- Probe 4 confirmed `bash <<<'git status'` and `bash <<<'git --version'` correctly ALLOW (git but no commit — the inner-payload regex requires both tokens).

**3. Eval shape** — `_EVAL_RE` (lines 298-301)

- **Correctly does NOT anchor on `_INTERPRETERS`**. Per Aino's docstring at line 290-297: "eval is a shell builtin universally available". This matches my original framing in the #479 verdict — `eval` is shell-builtin not shell-binary. Endorse the design call.
- `test_eval_variable_substitution_documented_punt` (line 732-738) pins the variable-subst ALLOW path (`eval "$CMD"`, `eval $cmd`) with explicit comment citing the docstring punt.
- Punt documented in PR body's "Documented punts" section + inline in regex comment at lines 293-297.
- Probe 5 confirmed eval-with-chained-commands BLOCKS correctly (`eval 'commit; ls'`, `eval 'ls && commit'`).

**4. dash + ksh widening** — `_INTERPRETERS = (?:bash|sh|zsh|dash|ksh)` (line 215)

- `mksh|pdksh` excluded with explicit docstring rationale at lines 212-214 ("conservative bound; rare in modern installs"). Endorse. Probe 8 confirms both ALLOW.
- The widening auto-extends all 6 detectors (`_PIPE_TO_SHELL_RE`, `_DASH_C_RE`, `_PROCESS_SUB_RE`, `_SCRIPT_INVOKE_RE`, `_HEREDOC_RE`, `_HERESTRING_RE`) via string interpolation. Verified by reading each regex usage — all reference `_INTERPRETERS` correctly.
- Probe 7 confirmed `dash <<EOF` heredoc and `ksh <(...)` process-sub both BLOCK via auto-extension. This is the "free coverage" Aino's PR body claims.

**5. `.sh`-rename bypass (HIGHEST SEVERITY)** — `_SCRIPT_INVOKE_RE` (lines 258-260)

- New regex: `\b` + `_INTERPRETERS` + `\s+(?P<path>[^\s\-<>|;&(][^\s|;&<>]*)`. Character class analysis:
  - First char excludes `\s`, `-`, `<`, `>`, `|`, `;`, `&`, `(` — this is the disambiguation against `-c` (first char `-`), redirects (`<>|`), separators (`;&`), and process-sub openers (`<` already excluded; `(` excluded).
  - Subsequent chars exclude `\s|;&<>` (terminators) but ALLOW `-` mid-token (for paths like `my-script`).
- `test_bash_script_dash_c_does_not_false_match_as_script` (line 836-846) is the regression guard. Reads correctly: asserts the `shell -c` label is in reason AND `shell script` is NOT — pins both halves of the disambiguation.
- Coverage: `.bash`, `.ksh`, `.sh`, extensionless, **and dotfile** (probe 10) all BLOCK. Probe 9 confirmed `bash <(...)` correctly routes through `_PROCESS_SUB_RE` not `_SCRIPT_INVOKE_RE`.
- 256 KiB cap + fail-open OSError discipline preserved — `test_bash_script_with_size_cap_still_protected` validates with a 300 KiB script (ALLOW because over cap). Same exact mechanism I empirically verified in my #479 review.

### Adversarial probe results (14 cases against PR-HEAD source)

All 14 confirmed correct behavior. Highlights:

- **Probe 7** auto-extension confirmation: `dash <<EOF` heredoc + `ksh <(...)` proc-sub both BLOCK — the `_INTERPRETERS` widening genuinely extends ALL 6 detectors as claimed.
- **Probe 11** cross-domain partial coverage: `env FOO=bar bash <script>` BLOCKS via the SCRIPT_INVOKE path — even though the env-wrapper is out of #482 scope, the script-read path catches the inner content because `bash` is the second token after `env FOO=bar` and `_SCRIPT_INVOKE_RE` finds it. **Good news**: the env-wrapper-with-script-bypass shape is incidentally already mitigated by this PR even though the env-wrapper-with-direct-git case is not.
- **Probe 12** principal false-positive guards: `cat <<EOF`, `tee <<EOF`, `python3 <<EOF` with "git commit" in body all ALLOW. The `_INTERPRETERS` anchor on heredoc detection is the right call — these are the realistic operator shapes that would have false-blocked without it.
- **Probe 13** regression: canonical direct `git -c user.name=... -c user.email=... commit -F /tmp/msg.txt` ALLOWS; bare `git commit -m "..."` BLOCKS with the original identity-missing message. No regression on the original commit-identity validation path.

### Endorsement of design decisions

1. **`_INTERPRETERS` anchor on heredoc/here-string but NOT on eval** — exactly right. Eval is shell-builtin so it doesn't need interpreter anchor; heredoc/here-string need it to avoid false-blocking documentation/data shapes.
2. **mksh/pdksh exclusion** — conservative bound is correct for now. Easy to extend later if observed. The deploy fleet doesn't use them.
3. **Extension-agnostic + first-char exclusion `[^\s\-<>|;&(]`** — the right way to relax the `.sh` restriction without colliding with the `-c` detector. The choice of exclusion characters is exhaustive against the realistic flag/redirect/separator/proc-sub-opener vocabulary.
4. **Unified block message cites both #475 and #482 + lists new shapes** — `test_unified_block_message_cites_both_issues` and `test_block_message_lists_new_shapes` are good operator-experience pins. Operators landing on the message can trace the full bypass surface without re-discovering shapes.
5. **Disk I/O check LAST in dispatcher (`_SCRIPT_INVOKE_RE` at line 406)** — correct ordering. All pattern-only checks short-circuit first; only commands matching the script-invoke shape pay the disk read cost. Performance discipline.

### Two micro-observations (not blockers)

**a. Same-segment text false-positive (probe 5 third case)** — `eval 'ls; echo ran git commit'` BLOCKS even though the operator intent is "echo a documentation phrase". The inner-commit regex `\bgit\b[^;&|]*?\bcommit\b` correctly enforces the `[^;&|]` segment boundary between `git` and `commit`, but in this case both tokens are in the same segment (after the `;` in `ls;`) inside the `echo`. Aino's narrow-default framing at the docstring level ("requires both `git` AND `commit` in the inner payload") makes this acceptable — operator workaround is trivial (`eval 'ls; echo ran g-i-t c-o-m-m-i-t'` or similar). Worth a one-line docstring note acknowledging this echo-of-documentation case if convenient.

**b. Eval bareword limitation (probe 6)** — `eval git -c user.name=X commit -m z` (no quotes around payload) ALLOWS because `_EVAL_RE`'s payload group `(?P<payload>(?P<q>['\"]).*?(?P=q)|\S+)` captures only a single token `\S+` when unquoted, and the inner-commit regex won't match a single-token payload. **Documented-punt cousin to the variable-substitution punt.** Operator practice always quotes eval bodies, so realistic impact is low. Worth a docstring note alongside the variable-subst punt at lines 293-297 if convenient.

Neither micro-observation gates the PR. Aino's narrow-default-with-documented-punts framing is the right shape for security-sensitive parsing where over-blocking has higher cost than these two narrow false-negatives.

### Env-wrapper finding — correctly deferred

The env-wrapper bypass I cross-flagged on PR #483 review (drift evidence comment on #482 as row-6) is **correctly out of scope** for this PR. Aino's "Related" section in the PR body acknowledges it explicitly: "The cross-domain `env`-wrapper bypass I flagged on PR #483's review (commented on #482 as drift evidence) is NOT closed by this PR. The Aisha+Petra original scope of #482 covers 5 shapes; the env-wrapper finding is row 6 added later via comment. Should be picked up as a follow-up issue or rolled into a Phase B sweep."

Charter discipline respected. Surface for separate follow-up issue (the env-wrapper bypass surface in `validate_commit_identity.py`'s `_DASH_C_RE` and similar — separate from `auto_set_env_test.py`'s already-closed `_is_gh_invocation` env-wrapper extension in PR #483).

**My probe 11 found that the env-wrapper + script bypass shape is INCIDENTALLY mitigated** by the new extension-agnostic script-read path. This is a nice bonus — Aino's `_SCRIPT_INVOKE_RE` catches `env FOO=bar bash /tmp/script.bash` because `bash` is the second tokenized chunk after `env FOO=bar`. The remaining env-wrapper bypass surface is narrower: `env FOO=bar bash -c '...git commit...'` and friends where the inner command is `-c` not a script-path. Worth noting in the follow-up issue.

### Cluster fit

This PR closes 5 of the 6 follow-up shapes from #479's review (the 6th — env-wrapper — explicitly deferred). Together with #479 (4 original shapes) the validate_commit_identity hook now covers 9 of 10 known indirect-exec wrapper shapes. The remaining env-wrapper-with-`-c` shape is a follow-up.

W11 hook-hardening cluster continues to converge: #475 → #482 closes the `validate_commit_identity` indirect-exec surface, #474/#181 closes the annunaki-noise + auto_set_env_test surfaces. Consistent with wave-stated tech-debt-ratio goal.

Approved.

PR #420, opened 2026-05-13
retro(P3W9): trust updates + counter corrections + 3 audit follow-ups + 3 process-change proposals

parametrization:
## Summary

P3W9 retro — `Tech-Debt Reduction (Main-Only)`. Per the charter rule from PR #412 (Retro PR Body-vs-Diff Discipline, closing #126), every substantive file claimed below is in the diff.

## Why

W9 wrapped 7 issues across 6 PRs in a single working session. Surfaced 2 new orchestrator-class spawn-brief defects worth codifying for W10 (TechDebt-line shape + clone-spawning vs SendMessage). Surfaced 3 real defects in `/promotion-audit` skill itself during the retro's audit run — filed as #417/#418/#419 for W10. Also corrects the wrapup-time counter-write gap that has now occurred in 3 consecutive waves (W4 / W5 / W9).

## What changed

| File | Change |
|---|---|
| `.claude/team/feedback_log.md` | W9 retro section appended after W8 (~190 lines). Wave-shape table, per-engineer assessments, top-3 going-well / pain-points, 3 process-change proposals, pattern tally with 3 new patterns, promotion audit summary, memory audit. |
| `.claude/team/trust_matrix.md` | Phase 3 Wave 9 Trust Updates section appended after W8. Org-level (Aino/Nadia/Santiago hold at 5; **Wanjiku 4 → 5** promotion recovering from W8 head-truncation lesson now codified); Orchestrator self-assessment (hold at 3 with two new spawn-brief defects); Done-Well/Needs-Improvement table. |
| `cross-repo-status.json` | `current_wave` advanced `wave-8` → `wave-9`; `next_wave` → `wave-10`; `last_completed_wave` → `wave-9`. Counter corrections: `wave_9_changes_requested_cycles: 0`, `wave_9_top_concentration_pct: 67`, `wave_9_counter_corrections` array recording the wrapup-write-gap correction. Run-markers: `wave_9_annunaki_attack_ran_at`, `wave_9_memory_audit_ran_at`, `wave_9_promotion_audit_ran_at` (+ `_counts`), `wave_9_retro_completed_at`. |
| `.claude/team/promotion_audit_log/wave-9.md` | New file — deterministic /promotion-audit output for W9. 0 AUTO / 0 DECIDE / 53 KEPT / 16 SUPERSEDED. |

## Trust matrix changes

[...]

PR #494, opened 2026-05-18
tech-debt(hook): auto_set_env_test — shell-syntax-aware parsing (#478)

### LINKED ISSUE #476: tech-debt(hook): auto_set_env_test.py — broken suggestion text + silent gate bypass for cd && pytest chains

### LINKED ISSUE #478: tech-debt(hook): auto_set_env_test — shell-syntax-aware parsing for quotes/parens/control-flow (#476 follow-up)
## Background

Follows from PR #477 (closed #476) — Wanjiku Mwangi's TPM-secondary verdict surfaced three related limitations of the chain-aware split landed in #477. All share the same root cause: the regex split is **shell-syntax-unaware**. Filing as one issue per Wanjiku's recommendation; sibling to #476's documented punt.

PR #477 explicitly punted on these — they fail in the **safe direction** (over-block + mangled suggestion, never silent-bypass), so #477 ships clean and this is W12+ sweep work.

## Three failure cases

All produce mangled suggestions or false-positive blocks, none cause silent gate bypass.

### 1. Quote-unaware splitting

Quoted `&&` inside argument strings is treated as a separator:

```bash
git log --grep='fix && pytest' && echo done
# Hook splits into: ["git log --grep='fix", "pytest' && echo done"]
# Second segment contains pytest → demands ENVIRONMENT=test
# Mangled suggestion: git log --grep='fix && ENVIRONMENT=test pytest' && echo done
```

The literal `pytest` token inside the quoted grep pattern is never executed but trips detection.

### 2. Subshell-unaware env-block detection

`(ENVIRONMENT=test pytest tests/)` — env-block is inside a subshell, detector looks at segment leading tokens and misses it:

```bash
(ENVIRONMENT=test pytest tests/)
# Detector: segment leading tokens are "("; no ENVIRONMENT=test in leading env-block
# Result: blocked despite env being correctly placed inside the subshell
```

### 3. Control-flow-unaware splicing

Suggestion-rewrite inserts `ENVIRONMENT=test` in invalid positions inside loops/conditionals:

```bash
for f in a b; do pytest $f; done
# Hook splits on `;` → ["for f in a b", "do pytest $f", "done"]
# Second segment contains pytest → mangled to: "for f in a b; ENVIRONMENT=test do pytest $f; done"
# The "ENVIRONMENT=test do" is invalid shell syntax
```

## Proposed fix

Replace regex-split with a shell-syntax-aware tokenizer. Options ranked by complexity:

1. **`shlex` module** — handles quotes natively. Doesn't preserve separator positions or whitespace exactly, so suggestion-rewrite needs adjustment, but eliminates the quoted-string false positive. Likely the right tradeoff.
2. **`bashlex` library** — third-party (PyPI), full Bash AST. Heaviest but covers subshells + control flow correctly. Adds runtime dependency on a parser.
3. **Hand-rolled state machine** — track quote state during split, skip control-flow keywords (`do`/`done`/`then`/`fi`). Moderate complexity, no new deps, matches existing hook dependency-free style.

Recommended path: option 1 (`shlex`) for quote handling, plus a small post-process step that bails out if the test segment is sandwiched between control-flow keywords (skip the splice, log a warning but allow — error on the side of NOT mangling). Subshell case needs explicit subshell-recognition; can be a small follow-up commit.

## Acceptance criteria

- [ ] Test: `git log --grep='fix && pytest' && echo done` — segment-classification recognizes the quoted pytest as inside a string literal → blocked due to the REAL pytest segment, but suggestion does NOT mangle the quoted-string segment
- [ ] Test: `(ENVIRONMENT=test pytest tests/)` — subshell-aware → allowed
- [ ] Test: `for f in a b; do pytest $f; done` — control-flow-aware → either correctly inserts env (`for f in a b; do ENVIRONMENT=test pytest $f; done`) OR bails out with a one-line operator-readable diagnostic instead of mangling
- [ ] Test: `while pytest --quick; do echo ok; done` and `if pytest --smoke; then echo go; fi` — same handling
- [ ] All 40 existing tests in `.claude/hooks/tests/test_auto_set_env_test.py` continue to pass
- [ ] Decision recorded in PR body or in a docstring comment whether subshell + control-flow are handled in this PR or split into a separate commit

## Wave/severity

Tech-debt — failure direction is operator inconvenience (over-block, mangled suggestion), NOT silent bypass. #476's higher-severity silent-bypass bug is already fixed. Suitable for W12 sweep wave; not blocking on W11 wrap.

## Origin

Wanjiku Mwangi's TPM-secondary verdict on PR #477, comment id `4473242426`, 2026-05-18.

parametrization:
Requestor: Weronika Zielinska
Requestee: Weronika Zielinska
RequestOrReplied: (blank)
TechDebt: none
Closes #478

## Summary

Three shell-syntax-awareness fixes to `auto_set_env_test` hook, all rooted in the same `_split_segments` regex-unawareness limitation. Follow-up to #477 (closed #476).

| Case | Pre-fix | Post-fix |
|---|---|---|
| 1. Quoted `&&`/`pytest` inside arg strings | False-positive split + mangled suggestion that inserts env inside the quoted region | Quote-aware state machine; quoted separators don't split, quoted `pytest` substrings don't trigger detection |
| 2. `(ENVIRONMENT=test pytest)` subshell | False-positive block — segment leading token is `(`, env-block check misses interior | Subshell-aware env-block check peeks inside parens; splice INSIDE parens when needed |
| 3. `for f in a b; do pytest $f; done` control-flow body | Block with syntactically-invalid suggestion (`for f in a b; ENVIRONMENT=test do pytest`) | **HARD BLOCK** with operator-readable diagnostic naming the construct (for/while/if) and showing both inline-env and pull-out alternatives |

## Design decision (recorded in hook docstring)

Subshell + control-flow handled in this PR, NOT punted. Once the state machine is in place for quote-awareness (case #1), both case #2 (subshell-aware env detection via leading-`(` peek) and case #3 (control-flow detection + hard block) are cheap additions. Splitting into separate commits would risk leaving the hook in a half-fixed state where mangled suggestions still ship for cases #2 and #3.

## Why option 3 (hand-rolled state machine) over option 1 (shlex)

Overrode the issue body's recommendation after reading the actual code:

- `shlex.split` discards separator positions AND whitespace. The existing `_SEPARATORS` regex uses a capturing group specifically to splice segments back into the original command with byte-exact glue. Switching to shlex would mean rebuilding the entire splice pipeline.
- `shlex` doesn't recognise `&&`/`||` as separators by default. Would still need post-process for separator detection.
- Net effect: option 1 (shlex) actually requires MORE code than option 3 (hand-rolled). Option 3 also preserves the hook codebase's dependency-free style (option 2 — bashlex — adds a runtime dep).

## Case #3 hard-block design (per team-lead direction)

Initial implementation used `allow_with_log` for the control-flow case. Flipped to **HARD BLOCK** per team-lead correction:

> Allow-with-log says "we detected pytest, we know the env is missing, but we can't rewrite cleanly so we'll allow it" — that's silently bypassing the very protection. Failure direction matters more than UX friction here.

The mangled-suggestion failure is bad in the SAFE direction (operator doesn't run the command). Allow-with-log is bad in the UNSAFE direction (operator runs pytest without `ENVIRONMENT=test`, the exact #420 prod-data-wipe class this hook protects against).

Diagnostic format:
```
pytest detected inside a shell control-flow body (for/while/until/if)
that auto_set_env_test cannot safely rewrite — splicing ENVIRONMENT=test
between the separator and a control-flow keyword would produce invalid
shell. Add ENVIRONMENT=test manually inside the body. Example:
  for f in a b; do ENVIRONMENT=test pytest $f; done
Or pull the loop variable outside:
  ENVIRONMENT=test pytest a b
Original command: <command>
```

**Combined case** — `for f in a b; do pytest $f; done && pytest other/` (control-flow + splice-able): suggestion fixes the splice-able segment AND appends a note pointing at the control-flow segment for manual edit. Splice does NOT mangle the for-body.

Detected control-flow shapes:
- Segment leading with opener keyword (`for`/`while`/`until`/`if`/`case`)
- Segment leading with body keyword (`do`/`then`/`done`/`fi`/`else`/`elif`/`esac`/`in`)
- Segment in the body of a PRIOR opener (backward scan, stops at `done`/`fi`/`esac` closer)

## Scope NOT covered (caveats explicit)

- `$(...)` and backtick command substitutions — separators inside still split. Acceptable because the inner command only fires if the substitution executes, and any pytest invocation inside would be re-caught by the segment scan.
- `for f in a b; do ENVIRONMENT=test pytest $f; done` (env CORRECTLY placed in body) is BLOCKED as a known false-positive — body-segment leading token is `do`, not the env assignment, so the env-block check misses it and the control-flow detector fires. Trade for the safety invariant: any test segment we can't analyze cleanly is blocked, not allowed. Operator workaround: move env outside the loop (`ENVIRONMENT=test pytest a b`). Future refinement could strip control-flow keywords before re-checking the env-block.
- Token-level rewrite of `while pytest --quick` opener segment to `while ENVIRONMENT=test pytest --quick` (which IS valid shell). Currently HARD BLOCKS with diagnostic instead. Could be a future refinement if operator feedback shows the block is too restrictive.

## Acceptance criteria (from #478)

- [x] Test: `git log --grep='fix && pytest' && echo done` — quote-recognized, suggestion does NOT mangle the quoted segment (`QuoteAwareSplittingTests::test_quoted_amp_amp_inside_grep_pattern`)
- [x] Test: `(ENVIRONMENT=test pytest tests/)` — subshell-aware → allowed (`SubshellAwareTests::test_subshell_with_env_inside_allowed`)
- [x] Test: `for f in a b; do pytest $f; done` — **HARD BLOCK** with operator-readable diagnostic naming the construct and showing both alternatives (`ControlFlowHardBlockTests::test_for_do_pytest_done_hard_blocks_with_diagnostic`)
- [x] Test: `while pytest --quick; do echo ok; done` and `if pytest --smoke; then echo go; fi` — both HARD BLOCK with diagnostic
- [x] All 48 existing tests in `.claude/hooks/tests/test_auto_set_env_test.py` continue to pass
- [x] Decision recorded in PR body + hook docstring

## Test results

- 65 total tests (48 existing + 17 new); all pass.
- 846 total hooks tests pass — no regressions in adjacent hooks.
- `ruff format --check` + `ruff check` clean; pre-commit hooks pass.

## Test plan

- [x] All 65 unit tests pass: `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/test_auto_set_env_test.py -v`
- [x] Full hooks suite passes: `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/ -q`
- [x] Ruff format + lint clean
- [ ] Post-merge: smoke test `(ENVIRONMENT=test pytest)` subshell form in a real session — should be allowed without prompt
- [ ] Post-merge: smoke test `git log --grep='fix && pytest'` — should be allowed without prompt
- [ ] Post-merge: smoke test `for f in a b; do pytest $f; done` — should HARD BLOCK with manual-edit diagnostic (no silent run)

### DIFF
--- a/.claude/hooks/auto_set_env_test.py
+++ b/.claude/hooks/auto_set_env_test.py
@@ -26,6 +26,7 @@ Matches (blocks if ENVIRONMENT=test missing on the test segment):
         DEBUG=1 pytest tests/                  (env prefix preserved)
         cd /path && pytest tests/              (chain — env must be on pytest segment)
         pytest tests/ && pytest tests-other/   (chain — env required on BOTH segments)
+        (ENVIRONMENT=test pytest tests/)       (subshell — env recognised inside parens)
 
 Does NOT match (short-circuit skips — #114 fix):
     1. Command whose effective argv[0] is `gh` — `gh` is a GitHub API
@@ -48,10 +49,18 @@ Detection order:
     1. Strip leading `VAR=value` tokens from the command for argv[0] check.
     2. If the next token is `gh`, ALLOW (return None).
     3. If the command contains `--body` or `--body-file`, ALLOW.
-    4. Split on `&&`/`||`/`;`/`|` into segments.
+    4. Split on `&&`/`||`/`;`/`|` into segments using a quote-aware
+       state-machine scanner — separators inside `'...'`, `"..."`, or
+       `\\`-escaped positions are NOT recognised.
     5. For each segment containing `\\bpytest\\b` or `\\bmake\\s+test\\b`,
-       require `\\bENVIRONMENT=test\\b` in that segment's leading env-block;
-       else BLOCK with a per-segment-rewritten suggestion.
+       require `\\bENVIRONMENT=test\\b` in that segment's leading env-block
+       (subshell-aware: a segment that opens with `(` peeks inside the
+       parens for the env-block); else BLOCK with a per-segment-rewritten
+       suggestion, EXCEPT when the test segment is the body of a
+       control-flow construct (`for|while|until|if ... ; do pytest ...`)
+       where splicing env between `;` and `do` would produce invalid
+       shell — there we ALLOW with an operator-readable diagnostic
+       logged via annunaki_log rather than emit a mangled suggestion.
 
 Per-segment env-block check (the #476 silent-bypass fix):
     `ENVIRONMENT=test cd /x && pytest tests/` does NOT pass — even though
@@ -59,12 +68,33 @@ Per-segment env-block check (the #476 silent-bypass fix):
     on the pytest segment. Shell semantics apply env assignments only to
     the immediately-following program in the same segment.
 
+Shell-syntax-awareness scope (#478):
+    The state-machine split handles single quotes, double quotes, and
+    backslash-escapes at the top level. It does NOT recurse into
+    `$(...)` or backtick command substitutions — separators inside
+    `$(... && ...)` will still split, which is acceptable because the
+    inner command would only fire if the substitution were executed and
+    the segment scan would catch any pytest invocation regardless.
+    Subshell-recognition is one-deep: `(((pytest)))` is correctly
+    parsed; nested-with-other-content edge cases bail to the
+    control-flow allow-with-diagnostic path on the side of NOT mangling.
+
+Decision on subshell + control-flow coverage (#478):
+    Both are handled in this PR, NOT punted. Once the state machine is
+    in place for quote-awareness (case #1), subshell detection is a
+    leading-`(` peek and control-flow detection is a sibling-segment
+    keyword check — both cheap. Splitting into separate commits would
+    risk leaving the hook in a half-fixed state where mangled
+    suggestions still ship for cases #2 and #3.
+
 Exit codes:
     0 — allow (not a Bash tool, or skip condition met, or every test
-        segment already has ENVIRONMENT=test in its leading env-block)
-    2 — block (at least one test segment is missing ENVIRONMENT=test)
+        segment already has ENVIRONMENT=test in its leading env-block,
+        or test segment is inside a control-flow body — see #478)
+    2 — block (at least one test segment is missing ENVIRONMENT=test
+        and we can produce a syntactically-valid corrected suggestion)
 
-Enforcement artifact for: noorinalabs/noorinalabs-main#114, #476
+Enforcement artifact for: noorinalabs/noorinalabs-main#114, #476, #478
 """
 
 import json
@@ -83,17 +113,19 @@ _ENV_ASSIGN = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*=\S*\s+")
 # `--body=x`, and `--body-file path`.
 _BODY_FLAG = re.compile(r"(?<![\w-])--body(?:-file)?(?=[\s=]|$)")
 
-# Shell separators we split on. Order matters in the alternation: longer
-# tokens first so `&&` isn't mis-split as two `|`s by the `|` alternative.
-# Captured group preserves the separator in the split output so we can
-# splice segments back together with original glue.
-_SEPARATORS = re.compile(r"(\s*(?:&&|\|\||;|\|)\s*)")
-
 # Test-runner detection: pytest OR `make test`.
 _TEST_RUNNER = re.compile(r"\bpytest\b|\bmake\s+test\b")
 
 # The literal env-var we require.
 _ENV_TOKEN = re.compile(r"\bENVIRONMENT=test\b")
+
+# Shell control-flow keywords that appear as standalone tokens between
+# segment separators. When a test-bearing segment is bracketed by these,
+# splicing `ENVIRONMENT=test` between the separator and the next program
+# produces invalid shell (e.g. `for f in a; ENVIRONMENT=test do pytest`).
+# Used by the bail-with-diagnostic path in `_rewrite_command` (#478).
+_CONTROL_FLOW_OPENERS = frozenset({"for", "while", "until", "if", "case"})
+_CONTROL_FLOW_BODY_KEYWORDS = frozenset({"do", "done", "then", "else", "elif", "fi", "esac", "in"})
 
 
 def _strip_leading_env(command: str) -> str:
@@ -151,11 +183,103 @@ def _split_segments(command: str) -> list[str]:
     """Split command on `&&`/`||`/`;`/`|`, preserving separators as
     alternating list entries: [seg0, sep0, seg1, sep1, ..., segN].
 
+    Quote-aware (#478): separators inside `'...'`, `"..."`, or
+    `\\`-escaped positions are NOT recognised. The previous regex-only
+    implementation tripped on `git log --grep='fix && pytest'` by
+    treating the `&&` inside the grep pattern as a real separator and
+    classifying the quoted `pytest` token as a test-runner invocation.
+
     The result always has odd length: segments at even indices, separators
     at odd indices. A command with no separators returns a single-element
-    list [command].
+    list [command]. Surrounding whitespace around separators is preserved
+    in the separator entry so segments + separators concat back to the
+    original command byte-for-byte.
+
+    State machine:
+        - quote_char: None | "'" | '"'  (current quoting context)
+        - in_escape:  True after a `\\` outside single quotes (POSIX rule)
+        - Separators only fire when quote_char is None.
+        - Inside `'...'` backslashes are literal (POSIX); inside `"..."`
+          they escape `$` `` ` `` `"` `\\`. We treat `\\` as a generic
+          escape outside single quotes — close enough for the hook's
+          purpose, which is to avoid mis-splitting on quoted operators.
     """
-    return _SEPARATORS.split(command)
+    parts: list[str] = []
+    buf: list[str] = []
+    quote_char: str | None = None
+    in_escape = False
+    i = 0
+    n = len(command)
+    while i < n:
+        ch = command[i]
+        if in_escape:
+            buf.append(ch)
+            in_escape = False
+            i += 1
+            continue
+        if quote_char is not None:
+            if ch == "\\" and quote_char == '"':
+                buf.append(ch)
+                in_escape = True
+                i += 1
+                continue
+            buf.append(ch)
+            if ch == quote_char:
+                quote_char = None
+            i += 1
+            continue
+        # Unquoted context.
+        if ch == "\\":
+            buf.append(ch)
+            in_escape = True
+            i += 1
+            continue
+        if ch in ("'", '"'):
+            buf.append(ch)
+            quote_char = ch
+            i += 1
+            continue
+        # Separator detection — only fires outside quotes.
+        sep = _match_separator_at(command, i)
+        if sep is not None:
+            # Look back over the buffer to find the trailing whitespace
+            # that belongs to the separator group, mirroring the old
+            # regex `\s*(...)\s*` capture.
+            seg = "".join(buf)
+            trailing_ws_len = len(seg) - len(seg.rstrip())
+            parts.append(seg[: len(seg) - trailing_ws_len])
+            ws_before = seg[len(seg) - trailing_ws_len :]
+            j = i + len(sep)
+            ws_after_len = 0
+            while j < n and command[j] in (" ", "\t"):
+                ws_after_len += 1
+                j += 1
+            parts.append(ws_before + sep + command[i + len(sep) : i + len(sep) + ws_after_len])
+            buf = []
+            i = j
+            continue
+        buf.append(ch)
+        i += 1
+    parts.append("".join(buf))
+    return parts
+
+
+def _match_separator_at(command: str, i: int) -> str | None:
+    """Return the longest separator token at position `i`, or None.
+
+    Order is significant: `&&` and `||` must be matched before single
+    `&`/`|` would (we don't recognise bare `&`, but `||` must match
+    before `|`).
+    """
+    if command.startswith("&&", i):
+        return "&&"
+    if command.startswith("||", i):
+        return "||"
+    if command[i] == ";":
+        return ";"
+    if command[i] == "|":
+        return "|"
+    return None
 
 
 def _segment_has_env_in_leading_block(segment: str) -> bool:
@@ -165,15 +289,74 @@ def _segment_has_env_in_leading_block(segment: str) -> bool:
     start of the segment (after stripping surrounding whitespace). This
     is the only position where a shell env assignment applies to the
     segment's program.
+
+    Subshell-aware (#478): if the segment opens with `(`, peek inside
+    the parens at the leading env-block of the subshell contents. This
+    handles `(ENVIRONMENT=test pytest tests/)` — the env is on the
+    correct (sub-)segment's program. We strip the leading `(` and any
+    interior whitespace before re-checking; we do NOT need to find the
+    closing `)` because the env-block check only cares about leading
+    tokens. Nested subshells (`((ENVIRONMENT=test pytest))`) are handled
+    by recursion.
     """
     stripped = segment.lstrip()
+    if stripped.startswith("("):
+        # Strip one layer of subshell parens + interior whitespace, then
+        # recurse. This lets the env-block check see through one or more
+        # nested subshells uniformly.
+        return _segment_has_env_in_leading_block(stripped[1:])
     consumed = stripped[: len(stripped) - len(_strip_leading_env(stripped))]
     return bool(_ENV_TOKEN.search(consumed))
 
 
 def _is_test_segment(segment: str) -> bool:
-    """True if segment contains pytest or `make test` as a real invocation."""
-    return bool(_TEST_RUNNER.search(segment))
+    """True if segment contains pytest or `make test` as a real invocation.
+
+    Quote-aware (#478): a `pytest` token inside `'...'` or `"..."`
+    quoting is NOT a test-runner invocation — it's argument data. The
+    `_split_segments` quote-aware tokenizer already protects against
+    quoted separators triggering bogus segment splits, but a single
+    segment can still contain a quoted `pytest` substring (e.g.
+    `git log --grep='fix pytest'` with no separators at all). This
+    function strips quoted regions before applying the test-runner
+    regex.
+    """
+    return bool(_TEST_RUNNER.search(_strip_quoted_regions(segment)))
+
+
+def _strip_quoted_regions(s: str) -> str:
+    """Return `s` with single-quoted and double-quoted regions replaced
+    by empty strings. Backslash-escapes outside single quotes consume
+    their following char.
+
+    Used by `_is_test_segment` to avoid false-positive matches on
+    `pytest` / `make test` text that appears inside argument quoting.
+    """
+    out: list[str] = []
+    quote_char: str | None = None
+    in_escape = False
+    for ch in s:
+        if in_escape:
+            in_escape = False
+            if quote_char is None:
+                out.append(ch)
+            continue
+        if quote_char is not None:
+            if ch == "\\" and quote_char == '"':
+                in_escape = True
+                continue
+            if ch == quote_char:
+                quote_char = None
+            continue
+        if ch == "\\":
+            in_escape = True
+            out.append(ch)
+            continue
+        if ch in ("'", '"'):
+            quote_char = ch
+            continue
+        out.append(ch)
+    return "".join(out)
 
 
 def _prepend_env_to_segment(segment: str) -> str:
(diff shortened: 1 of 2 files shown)


parametrization:
Requestor: Wanjiku Mwangi
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: None — quote-state machine handles single-quote/double-quote/backslash correctly; subshell recursion handles nesting; control-flow HARD BLOCK is the right safety-direction call. Known false-positives (env-correctly-in-body, unclosed quotes, `case` body with semicolons) all fail in the safe direction (over-block) and Weronika documented them explicitly.

## TPM verdict — Approved (secondary, as #478 surfacer)

Reviewed at PR HEAD `4c064614d1ec1b8bc52083c855549343c5202b12` via `gh api repos/.../contents/<path>?ref=4c06461` + `gh pr diff 494` per [[feedback_reviewer_must_not_branch_switch_parent]] (did NOT `git checkout` in parent main checkout). Reviewed the NEW HEAD only — did not look at the pre-flip `b56e151` per brief.

Independent of Aisha; no coordination. Ran the 65-test suite locally (`ENVIRONMENT=test python3 -m pytest`): **65 passed in 0.10s** — confirms PR body claim.

### Lens 1 — Does it close MY 3 cases from PR #477 verdict?

I walked back to my exact original verdict prose and tested each case:

**Case 1 — quote-unaware splitting** (`git log --grep='fix && pytest' && echo done`):
- Pre-fix: blocked with mangled suggestion `git log --grep='fix && ENVIRONMENT=test pytest' && echo done`
- Post-fix: **ALLOW** (correctly recognises quoted `&&` as data, quoted `pytest` as data)
- Pinned by `QuoteAwareSplittingTests::test_quoted_amp_amp_inside_grep_pattern`
- ✓ Fully closes case 1

**Case 2 — subshell-unaware env-block detection** (`(ENVIRONMENT=test pytest tests/)`):
- Pre-fix: blocked with `ENVIRONMENT=test (ENVIRONMENT=test pytest tests/)` mangled suggestion
- Post-fix: **ALLOW** (recursive `_segment_has_env_in_leading_block` peeks inside parens)
- Pinned by `SubshellAwareTests::test_subshell_with_env_inside_allowed`
- ✓ Fully closes case 2

**Case 3 — control-flow-unaware splicing** (`for f in a b; do pytest $f; done`):
- Pre-fix: blocked with mangled `for f in a b; ENVIRONMENT=test do pytest $f; done`
- Post-fix: **HARD BLOCK** with operator-readable diagnostic naming the construct + inline-env + pull-out alternatives + original command echo
- Pinned by `ControlFlowHardBlockTests::test_for_do_pytest_done_hard_blocks_with_diagnostic`
- ✓ Fully closes case 3 — and the hard-block-not-allow-with-log call is correct per my own [[feedback_security_guard_inline_not_followup]] precedent (failure direction wins over UX)

### Lens 2 — Known false-positive call (env-correctly-in-body)

`for f in a b; do ENVIRONMENT=test pytest $f; done` — operator HAS placed env correctly, still HARD BLOCK. Pinned by `ControlFlowHardBlockTests::test_for_with_env_inside_do_body_hard_blocks`.

**My call: accept as documented.** Reasoning:

1. Operator workaround is one-character-shorter (`ENVIRONMENT=test pytest a b` rather than the loop) — the friction is real but minimal.
2. The alternative (strip control-flow keywords before re-checking env-block) would require token-level rewriting of `do ENVIRONMENT=test pytest` → recognise `do` is a keyword → re-check env-block from after `do`. That's a deeper state-machine layer with its own bug surface.
3. The hard-block diagnostic explicitly suggests the pull-out alternative in its example, so operators are guided to the simpler shape.
4. Sibling to W11 retro pattern: when refinement requires a deeper layer with new bug surface, prefer the simpler over-block + documented workaround. Weronika's "trade for safety invariant" framing matches this.

**Suggest filing as W12+ follow-up issue** if operator feedback shows hard-block frequency is high. Not blocking — documentation in PR body caveats + test docstring is enough for now.

### Lens 3 — Combined-message case (`for f in a b; do pytest $f; done && pytest other/`)

Pinned by `ControlFlowHardBlockTests::test_mixed_control_flow_and_real_block_combined_message`. Verified the combined-message text — reads like:

> ENVIRONMENT=test is required for test commands but was not found in the leading env-block of each test-bearing segment. Shell env assignments only apply to the immediately-following program in the same segment, so chains like `cd /x && pytest` need the env on the pytest segment, not at the start of the whole command. Suggested command:
>   for f in a b; do pytest $f; done && ENVIRONMENT=test pytest other/
> 
> Note: at least one pytest invocation is inside a shell control-flow construct (for/while/until/if) that auto_set_env_test cannot safely rewrite. Add ENVIRONMENT=test manually inside the body for those segments.

**Operator UX read**: the suggestion fixes the splice-able segment cleanly AND the note tells them they still have manual work on the for-body. Two-things-not-one is correctly communicated. The for-body in the suggested command is preserved verbatim (NOT mangled with `; ENVIRONMENT=test do`) — pinned by the `assertNotIn("; ENVIRONMENT=test do", ...)` assertion in the test. ✓

### Lens 4 — Hook docstring + design-decision recording durability

Counted explicit `#478` + `HARD BLOCK` + `allow-with-log` + `#420` references in `auto_set_env_test.py`:

- **Module docstring** (lines 59-66, 86-92): explicit decision recording — "Allow-with-log would silently bypass the protection (#420 prod-data-wipe class), so we trade UX friction for safety"
- **`_is_control_flow_bracketed` docstring** (lines 430-439): "Per the hook docstring decision (#478), control-flow detection triggers a HARD BLOCK with operator-readable diagnostic — never an allow-with-log. Allow-with-log would silently bypass the very protection this hook exists for (the #420 prod-data-wipe class)."
- **`check()` inline comments** (lines 512-517, 552-557): "safety direction: never silently allow a pytest run without ENVIRONMENT=test even when our rewriter can't produce a valid suggestion — #478 hard-block discipline. Allow-with-log would bypass the very protection this hook exists for, exactly the #420 prod-data-wipe class of failure."

**3 sites, all coherent**, all reference #420 prod-data-wipe class anchor. A future maintainer who reads ANY of these would have to actively choose to override the recorded rationale to flip back to allow-with-log. Durability is strong. ✓

### Lens 5 — shlex vs hand-rolled state machine design call

Weronika overrode my issue body's option-1 (shlex) recommendation. Read the implementation (`_split_segments` lines 186-268, `_strip_quoted_regions` lines 331-363) — agree with her override:

- **`_split_segments`** is ~80 lines including the state machine + docstring. The state explicitly tracks `quote_char` (None | `'` | `"`) and `in_escape`. The logic at lines 217-266 is linear, single-pass, no recursion at the top level. Reads like a textbook tokenizer.
- **`_strip_quoted_regions`** is ~30 lines and mirrors the same state machine. The pattern repetition is intentional (single-responsibility per function — one splits, the other strips), not bit-rot bait.
- **No external dependency** added (bashlex would have introduced runtime dep).
- **Byte-exact roundtripping** preserved (segments + separators concat to original) — pinned by the test suite indirectly via the suggestion-matches-expected-prefix assertions, which would break if separator whitespace were dropped.

**Future maintenance ergonomics**: the state machine is self-documenting via the docstring's "State machine:" section (lines 202-209). If a future maintainer needs to extend (e.g., recognise `$(...)`), the extension point is the unquoted-context branch around line 246 — adding a new state would touch lines 217-264. Not trivial, but not unreadable either. Sibling pattern to validate_commit_identity's `_strip_outer_quotes` / `_INTERPRETERS` regex blocks — same codebase style.

**Net call**: hand-rolled was the right choice. shlex would have added complexity without removing the separator-position-loss problem. Weronika's override-with-reasoning is sound.

### Edge-case probe (my surfacing-engineer additional iteration)

Ran 17 stress inputs beyond the test suite. Findings:

**All clean:**
- Double-quoted `&& pytest` inside arg → ALLOW (correct)
- Escaped inner quotes + real trailing pytest → BLOCK with correct splice
- `(((ENVIRONMENT=test pytest)))` nested subshell → ALLOW (recursion handles)
- `( ENVIRONMENT=test pytest )` with leading space → ALLOW
- `cd /x && (pytest)` → BLOCK with `cd /x && (ENVIRONMENT=test pytest)` splice
- `case x in *) pytest;; esac` → HARD BLOCK (control-flow correctly detected)
- `until pytest --once; do sleep 1; done` → HARD BLOCK
- `if true; then pytest tests/; fi` → HARD BLOCK (then-body segment)
- Real env-pytest + control-flow pytest → HARD BLOCK fires for the control-flow segment, combined-message path correct
- Pre-#477 silent-bypass case (`ENVIRONMENT=test cd /path && pytest tests/`) still BLOCKS (regression guard holds)

**Edge case worth noting (NOT defects, all safe-direction)**:
1. **Unclosed quote** (`git log --grep='unclosed && pytest`) → ALLOW. The unclosed quote causes everything trailing to be treated as in-quoted-state, stripping it from test-runner detection. **Bash would also reject this command at parse time**, so operator gets a shell syntax error before any pytest runs. Safe direction.
2. **`(cd /x && ENVIRONMENT=test pytest)`** → ALLOW. The outer `&&` splits even inside parens (split-doesn't-recurse-into-subshells is by design), creating segments `(cd /x` and `ENVIRONMENT=test pytest)`. Second segment has env on its program → allowed. **Happy coincidence rather than designed**: the subshell-aware path is for trivial `(ENVIRONMENT=test pytest)` shape; for chained-inside-subshell shapes it relies on the outer splitter doing the right thing. Worth a sentence in a future docstring update but not blocking.

### Tracker-discipline angle (my specialty)

Verified the hook still:
- Returns `{"decision": "block", "reason": ...}` to stdout
- Calls `log_pretooluse_block("auto_set_env_test", command, result["reason"], tool_name="Bash")` (line 584)
- Exits 2 (line 590)

All three signals preserved across both block paths (splice-able BLOCK at line 549 and control-flow HARD BLOCK at line 557). Annunaki monitor will see both block classes. ✓

### Sweep candidates (non-blocking, for W12+)

1. **`do`-keyword-stripping refinement** for the `for ... do ENVIRONMENT=test pytest ; done` false-positive. Would require token-level state machine extension. Defer until operator feedback shows hard-block frequency is high.
2. **Token-level rewrite of `while pytest --quick`** opener segments to `while ENVIRONMENT=test pytest --quick` (which IS valid shell). Weronika flagged in PR body as future refinement; agreed.
3. **`$(...)` and backtick command-substitution recursion** — currently splits inside, acceptable per Weronika's caveat. Defer unless real repro emerges.
4. **Subshell-with-internal-chain happy-coincidence**: add a docstring note that subshell-aware handling is ONLY for trivial single-segment subshells; chains inside subshells rely on the outer splitter producing correct results (which it does for the common case but isn't designed-for). Documentation, not code.

### Verdict

**Approved.** All 3 cases from my original #477 surface verdict are fully closed at HEAD `4c06461`. Hard-block-vs-allow-with-log call is correct per safety-direction discipline. Hand-rolled state machine is the right design choice (cleaner than shlex would have been, dependency-free). Design decision recording is durable across 3 sites. Combined-message UX is operator-readable. Known false-positives all fail in the safe direction and are explicitly documented.

Aisha is primary; this is the secondary 2-of-2 vote. Hook 4 will count both when Aisha posts.

parametrization:
Requestor: Aisha Idrissi
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: None — hard-block flip is correct safety-direction call; known false-positives (env-correctly-in-body still blocks, unclosed-quote ALLOW, cat-heredoc-with-pytest BLOCK) all fail in the safe direction (over-block in two cases; the unclosed-quote ALLOW is moot since the command is malformed shell anyway). Weronika documented the trades explicitly in docstring + test docstrings + PR body. The class rename + 3 flipped assertions are clean. No dead allow_with_log code; the only `allow-with-log` references are in comments explaining why it was rejected.

## Primary verdict — Approved (post hard-block flip)

**Reviewer:** Aisha Idrissi (deploy SRE) — primary
**Secondary:** Wanjiku Mwangi — Approved at 4473581601 (01:59:46Z)
**Head verified:** `4c064614d1ec1b8bc52083c855549343c5202b12` via `gh api .../contents/...?ref=4c06461` — origin-canonical, no parent-checkout branch switch (per [[feedback_reviewer_must_not_branch_switch_parent]])
**Scope:** exactly 2 files (`.claude/hooks/auto_set_env_test.py` +357/-31, `.claude/hooks/tests/test_auto_set_env_test.py` +221/-0). Matches brief.

### Verification methodology

1. Downloaded PR-HEAD source + tests via `gh api contents?ref=4c06461` into `/tmp/aisha_478_runtree/` (with `annunaki_log.py` from main).
2. Ran full PR-HEAD test suite: **65 passed, 0 failed** — matches Weronika's claim exactly (48 baseline + 17 new).
3. Hand-traced each of the 6 review angles in source.
4. Ran 14 adversarial probes targeting edges NOT in the PR test suite.
5. Confirmed no dead `allow_with_log` code (only comment-references explaining why it was rejected, e.g. at line 64-66, 432-435, 513-517, 550-551 — all are docstring/comment justifications, not code paths).

### Angle 1 — Hard-block correctness ✓

- **`decision: "block"` not `allow_with_log`**: line 557-570 returns `{"decision": "block", "reason": ...}`. The brief's "allow_with_log decision shape has been removed entirely" claim is verified — there is no `allow_with_log` decision value anywhere in the source.
- **Diagnostic names the construct family**: line 560-561 — `"pytest detected inside a shell control-flow body (for/while/until/if)"`. All 4 openers explicitly named (plus `case` is in `_CONTROL_FLOW_OPENERS` at line 131).
- **Diagnostic shows both alternatives**:
  - Inline-env: `"for f in a b; do ENVIRONMENT=test pytest $f; done"` (line 565)
  - Pull-out: `"ENVIRONMENT=test pytest a b"` (line 567)
  - Original command appended for operator review (line 568)
- **`test_for_do_pytest_done_hard_blocks_with_diagnostic`** (test line 557-570) asserts ALL of: block decision, construct family naming, both alternative shapes, original command inclusion.

### Angle 2 — Combined-message case ✓

- Lines 519-548 track `needs_splice_block` AND `saw_control_flow_block` SEPARATELY.
- When BOTH present, the splice-able segment gets the corrected suggestion AND a control-flow note is appended (line 542-548).
- `test_mixed_control_flow_and_real_block_combined_message` (test line 618-635) directly pins both halves:
  - `assertIn("&& ENVIRONMENT=test pytest other/", reason)` — splice-able got fixed
  - `assertNotIn("; ENVIRONMENT=test do", reason)` — for-body NOT mangled
  - `assertIn("control-flow", reason.lower())` — note appears
  - `assertIn("manually", reason.lower())` — operator-readable guidance

### Angle 3 — Known false-positive (env-correctly-in-body still blocks) — ACCEPT AS-IS

`for f in a b; do ENVIRONMENT=test pytest $f; done` HARD BLOCKS even though env is correctly placed in body. This is exactly Weronika's safety-direction stance: any test segment we can't analyze cleanly gets blocked, not allowed.

**My call: accept as-is, no follow-up needed.** Rationale:
1. Matches the #420 prod-data-wipe class of failure-direction reasoning that drove your hard-block correction — over-block is recoverable (operator gets diagnostic, picks pull-out alternative), under-block is not.
2. Workaround is trivial and operator-discoverable: pull the loop variable out (`ENVIRONMENT=test pytest a b`) — the diagnostic literally shows this alternative.
3. Refinement (strip control-flow keywords before re-checking env-block) is meaningful complexity for a niche operator pattern — not worth the parser complexity given the trivial workaround.
4. Weronika documented in test docstring (test_auto_set_env_test.py line 592-603) + PR body — operator surface is clean.

If a future operator hits this often enough to file an issue, the refinement is a single-line change and the test docstring already specifies the refinement direction. Not gating now.

### Angle 4 — Subshell handling ✓ (with bonus nesting verification)

- `_segment_has_env_in_leading_block` (line 307-311): lstrip + `startswith("(")` + recurse on `stripped[1:]`. Naturally handles ANY depth of nesting via tail-recursion.
- `_prepend_env_to_segment` (line 383-392): same recursive structure — env inserted INSIDE the deepest paren.
- **Probe 1 confirmed**: `((ENVIRONMENT=test pytest))` and `(((ENVIRONMENT=test pytest)))` both correctly ALLOW. Triple-nesting works.
- **Probe 8 confirmed**: `(pytest tests/)` BLOCKs with suggestion `(ENVIRONMENT=test pytest tests/)` — env inserted inside parens, not before them.
- **Probe 12 confirmed**: `(cd /tmp && ENVIRONMENT=test pytest)` ALLOWs — the inner segment scan finds the env on the pytest sub-segment correctly.

### Angle 5 — Quote-awareness ✓

- `_split_segments` state machine (lines 211-268): tracks `quote_char` ∈ `{None, "'", '"'}` + `in_escape` boolean.
- Single quotes preserve backslash literally (POSIX rule) — `\` inside `'...'` is a literal char (line 224-234 only sets `in_escape` when `quote_char == '"'`).
- Double quotes recognize `\` as escape (line 225-229).
- Outside quotes, `\` escapes next char (line 236-240).
- `_strip_quoted_regions` (line 331-363) is the second state machine for `_is_test_segment` — independently strips quoted regions so `git log --grep='pytest is a word'` correctly does NOT trigger test-segment detection.
- **Probe 3 confirmed**: both `git log --grep='fix pytest && etc' && echo done` and `grep "pytest in body" file.txt && ls` correctly ALLOW.

### Angle 6 — Design-decision rationale recorded ✓

Three recording sites verified:

1. **Module docstring lines 59-66**: "HARD BLOCK with an operator-readable diagnostic ... Allow-with-log would silently bypass the protection (#420 prod-data-wipe class), so we trade UX friction for safety." — names #420 incident class explicitly.
2. **Module docstring lines 86-92**: "Both are handled in this PR, NOT punted. ... Splitting into separate commits would risk leaving the hook in a half-fixed state." — justifies scope.
3. **Inline code comment lines 510-517**: same #420-class framing inline at the decision site so a maintainer editing `check()` sees the rationale without needing to scroll to docstring.
4. **Test docstring lines 549-555**: `ControlFlowHardBlockTests` class docstring re-states the rationale so any test-editing future maintainer sees it before flipping assertions back.

Four sites, not three. Operator-readable + maintainer-actionable. Future regressions are well-defended.

### Test count verification ✓

- 65 tests pass — matches PR body claim (48 baseline + 17 new).
- Class rename `ControlFlowBailTests` → `ControlFlowHardBlockTests` verified at line 543.
- 3 flipped assertions in `ControlFlowHardBlockTests`:
  - `test_for_do_pytest_done_hard_blocks_with_diagnostic` (line 557) asserts `result["decision"] == "block"` + diagnostic content
  - `test_for_with_env_inside_do_body_hard_blocks` (line 592) asserts `result["decision"] == "block"` (was allow-with-log)
  - `test_mixed_control_flow_and_real_block_combined_message` (line 618) asserts BLOCK + combined-message content
- No net test removals.

### Adversarial probe results (14 cases against PR-HEAD source)

All 14 confirm correct behavior. Highlights beyond Wanjiku's coverage:

- **Probe 1** (triple-nested subshell): `(((ENVIRONMENT=test pytest)))` ALLOWs via recursive lstrip.
- **Probe 4** (`case`): `case "$x" in a) pytest ;; esac` correctly HARD BLOCKs — `case` is in `_CONTROL_FLOW_OPENERS` at line 131.
- **Probe 5** (`until`): `until pytest --quick; do sleep 1; done` HARD BLOCKs — `until` in openers.
- **Probe 6** (mixed semicolon): `for ... done; ENVIRONMENT=test pytest tests/` ALLOWs — the closer detection at line 455 correctly identifies the trailing pytest as outside the control-flow body.
- **Probe 13** (case with 2 pytest invocations): BLOCKs as expected.

### Two minor probe-discovered observations (not blockers)

a. **Unclosed quote** (probe 11): `git log --grep='oops pytest` (no closing quote) ALLOWs because the state machine keeps `quote_char` set and the pytest token is "inside" the unclosed quote → `_is_test_segment` doesn't see it. Realistic operator impact: nil — the command is malformed shell and won't execute successfully regardless. Worth a one-line docstring note acknowledging the state-machine resilience pattern if convenient, but not necessary.

b. **Heredoc body** (probe 10): `cat <<EOF\npytest tests/\nEOF` BLOCKS as if pytest were the program (the state machine doesn't recognize `<<EOF` heredoc syntax). The splice suggestion would be slightly mangled. Realistic operator impact: low — operators rarely embed `pytest tests/` literal text in `cat <<EOF` doc bodies. Out of scope for #478 (the issue covered quote/subshell/control-flow; heredoc-aware splitting would be its own feature). Documented-punt material if it ever surfaces.

Neither warrants a follow-up issue today. Both fail in operator-recoverable directions.

### Endorsement of design choices

- **HARD BLOCK over allow-with-log** for control-flow case #3 — correct per failure-direction analysis. Trade UX friction for the safety invariant the hook exists to enforce.
- **Scope: subshell + control-flow in same PR** — correct per the docstring rationale (line 86-92). Splitting would have left mangled suggestions shipping for cases #2 and #3.
- **State-machine for quote-awareness over shlex.split** — correct given the hook's per-segment env-block check needs visual-structure preservation that shlex would consume.
- **`_CONTROL_FLOW_OPENERS` includes `case`** — defensive completeness; my probe 4 confirmed it correctly fires.
- **Detection ordering in `check()`** (gh-skip → body-flag-skip → segment scan) preserved from #181 — no regression on the env-wrapper / body-flag short-circuits I reviewed in PR #483.

Approved.

PR #495, opened 2026-05-18
tech-debt(annunaki_monitor): tighten SILENT_BOOLEAN_TEST_PATTERNS (3 gaps + micro-cleanup, #490, #481 follow-up)

### LINKED ISSUE #474: tech-debt(annunaki_monitor): extend IGNORE_PATTERNS for boolean-test idioms + add 2 follow-up test cases (#473 post-merge)

### LINKED ISSUE #490: tech-debt(hook): annunaki_monitor — tighten SILENT_BOOLEAN_TEST_PATTERNS (3 gaps + micro-cleanup, #474 follow-up)
## Background

Follow-up to PR #481 (closes #474) — `SILENT_BOOLEAN_TEST_PATTERNS` extension. Aisha Idrissi's secondary verdict surfaced 3 non-blocking gaps in the new idiom-skip logic + 1 micro-cleanup. Filing as a single consolidated tightening issue; none block #474's main win (suppressing common boolean-test noise from PR #473's silent-failure capture).

### 1. Split-flag `grep -E -q` falls through to LOG

`SILENT_BOOLEAN_TEST_PATTERNS` includes `\bgrep\s+-[a-zA-Z]*q` which matches `grep -q`, `-qE`, `-Eq`, `-qq` (combined-flag forms) but NOT split-flag forms like `grep -E -q pattern file`. Split-flag is uncommon (combined is the canonical Unix idiom) but lints/tools sometimes emit it.

**Fix**: extend regex to handle `grep\s+(?:-[a-zA-Z]+\s+)*-[a-zA-Z]*q` or detect any `-q` flag in any position. Modest regex complexity bump.

### 2. `diff --quiet` exit=2 ("trouble") indistinguishable from exit=1 (differ)

`diff` exit codes are documented:
- 0 = files identical
- 1 = files differ (idiom-skip target)
- 2 = trouble (couldn't open file, permission denied, etc.)

`SILENT_BOOLEAN_TEST_PATTERNS` matches `\bdiff\s+--quiet\b` and suppresses all non-zero exits when stderr is empty. **Exit=2 with no stderr** (rare but possible) gets suppressed when it shouldn't. In practice trouble-exits usually have stderr, so this is a low-frequency edge.

**Fix**: split the diff pattern to match exit_code=1 only, OR check exit_code value in the idiom-skip predicate. Slight refactor.

### 3. **HIGHEST PRIORITY** — `if [ truthy ]; then real-failure; fi` body failures SKIPPED

The most concerning of the three. The outer `if [` matches the line-anchored `^\s*if\s+\[` idiom regex. When the `if` condition is truthy and the body fails, the overall command exits non-zero with no stderr (if the body itself is silent) — and the idiom-skip predicate suppresses logging.

**Concrete example**:
```bash
if [ -d /target ]; then
  some-real-deploy-step-that-fails  # exit 1, no stderr
fi
```

Result: silent-suppressed. Operator loses audit-log entry for a real failure inside the conditional.

**Workaround**: don't wrap real failures in conditionals. But this is a common Bash pattern.

**Fix options**:
- (a) Drop `^\s*if\s+\[` from `SILENT_BOOLEAN_TEST_PATTERNS` — pattern is rarely the sole content of a Bash command anyway; better to over-log
- (b) Parse the if-body and apply idiom-skip ONLY when the body is also idiomatic (recursive). Complex.
- (c) Document the limitation explicitly so operators wrap unconditionally instead.

Recommend (a) — simplest fix, errs on side of over-logging (correct failure direction).

## Micro-cleanup

`^\s*\[` regex matches both `[` AND `[[` (since `[[` starts with `[`). The line-79 `^\s*\[\[` regex is redundant — OR-loop short-circuits on the first match. Drop the duplicate for clarity.

## Acceptance criteria

- [ ] `grep -E -q pattern file` exit=1, no stderr → SKIP (test added)
- [ ] `diff --quiet a b` exit=2, no stderr → LOG (test added, distinguishes from exit=1 case)
- [ ] `if [ -d /tmp ]; then false; fi` exit=1, no stderr → LOG (gap #3 fix — test added; this is the headline regression case)
- [ ] `if [ -d /nonexistent ]; then echo unreachable; fi` exit=0 → SKIP (false-condition path; existing behavior preserved)
- [ ] Existing 28 tests in `test_annunaki_monitor.py` continue to pass
- [ ] Redundant `^\s*\[\[` regex removed from `SILENT_BOOLEAN_TEST_PATTERNS`

## Wave/severity

Tech-debt — failure direction varies:
- Gap #1: under-skip (mild noise, harmless)
- Gap #2: over-skip on trouble-exit (rare but information-loss)
- **Gap #3: over-skip on common Bash conditional pattern (real failures silenced — highest severity)**

Gap #3 alone might justify W11 inclusion if there's bandwidth; rest is W12 sweep.

## Origin

Aisha Idrissi secondary verdict on PR #481, comment id 4473341810, 2026-05-18.

parametrization:
## Summary

Follow-up to PR #481 (which closed #474). Aisha Idrissi's 2° verdict on #481 surfaced 3 gaps + 1 micro-cleanup in the new `SILENT_BOOLEAN_TEST_PATTERNS` idiom-skip logic.

**Gap #1** — split-flag `grep -E -q`: extend the grep regex to `\bgrep\s+(?:-[a-zA-Z]+\s+)*-[a-zA-Z]*q`. Combined-flag (`-qE`, `-Eq`, `-qi`) still matches; split-flag (`-E -q`, `-i -q`) now also matches. *Under-skip direction — was logging by-design noise.*

**Gap #2** — `diff --quiet` exit=2 trouble: split into a new `SILENT_BOOLEAN_TEST_PATTERNS_BY_EXIT_CODE` list of `(pattern, by_design_codes)` tuples. `diff --quiet` / `git diff --quiet` only skip on exit_code=1 (the documented differs-signal); exit 2 (trouble — file-open failure, permission denied) now logs even with empty stderr. *Over-skip direction — was silencing real errors.*

**Gap #3 HEADLINE** — `if [ -d /target ]; then real-failure; fi`: remove `^\s*if\s+\[` and `^\s*if\s+\[\[` from `SILENT_BOOLEAN_TEST_PATTERNS` (option a from issue body). The outer `if [` regex matched even when the body — a real deploy step — was the failing command, silencing genuine failures. The existing `^\s*\[` regex still covers bare condition-only commands. Body failures now log on their own merit.

**Micro-cleanup** — `^\s*\[\[` removed (`^\s*\[` already matches `[[` since both start with `[`; the OR-loop short-circuits on the first match).

## Test plan

- [x] `grep -E -q pattern file` exit=1, no stderr → SKIP (`test_grep_split_flag_dash_E_dash_q_silent_exit_not_logged`)
- [x] `grep -i -q pattern file` exit=1, no stderr → SKIP (`test_grep_split_flag_dash_i_dash_q_silent_exit_not_logged`)
- [x] `diff --quiet a b` exit=2, no stderr → LOG (`test_diff_quiet_exit_2_trouble_logged`)
- [x] `git diff --quiet` exit=2, no stderr → LOG (`test_git_diff_quiet_exit_2_trouble_logged`)
- [x] `diff --quiet a b` exit=1, no stderr → SKIP (`test_diff_quiet_exit_1_differs_still_skipped` — preserved)
- [x] `if [ -d /tmp ]; then false; fi` exit=1, no stderr → LOG (`test_if_bracket_with_failing_body_logged` — HEADLINE)
- [x] `if [[ -d /tmp ]]; then false; fi` exit=1, no stderr → LOG (`test_if_double_bracket_with_failing_body_logged`)
- [x] `if [ -d /nonexistent ]; then echo unreachable; fi` exit=0 → SKIP (`test_if_bracket_false_condition_exit_zero_skipped`)
- [x] `[[ -f /nonexistent ]]` exit=1, no stderr → SKIP (`test_bash_double_bracket_still_skipped_after_dedup` — micro-cleanup regression guard)
- [x] **37 tests pass total** (was 28 — added 9 new in `IdiomTighteningTests`)
- [x] `uvx ruff check` + `uvx ruff format --check` on both files: clean
- [x] Pre-commit hooks (ruff-format, ruff-lint, mypy): passed

## Test-update note

2 existing tests from #481 were UPDATED, not preserved:
- `test_if_bracket_conditional_silent_exit_not_logged` → `test_if_bracket_conditional_false_branch_exit_zero_not_logged`
- `test_if_double_bracket_conditional_silent_exit_not_logged` → `test_if_double_bracket_conditional_false_branch_exit_zero_not_logged`

The original tests asserted the OLD whole-`if` suppression behavior — exactly the bug #490 fixes. Preserving them would re-assert the regression. The updated tests cover the preserved-behavior invariant from the acceptance criteria ("false-condition path; existing behavior preserved"), reframed around exit_code=0 instead of the (now-removed) whole-`if` suppression path.

Requestor: Lucas Ferreira
Requestee: Lucas Ferreira
RequestOrReplied:
TechDebt: none
Closes #490

### DIFF
--- a/.claude/hooks/annunaki_monitor.py
+++ b/.claude/hooks/annunaki_monitor.py
@@ -68,34 +68,64 @@ IGNORE_PATTERNS = [
     re.compile(r"error_log|error\.log|errorhandl", re.IGNORECASE),  # filenames
 ]
 
-# Silent-boolean-test idioms (#474): commands whose ONLY failure signal is a
-# non-zero exit code with empty stdout/stderr — by design, not an error.
-# After #473 closed the silent-failure blind spot, these idioms became noise.
-# Apply ONLY when matched_patterns is exactly ["exit_code=..."] (no stdout or
-# stderr pattern matched) — pattern matches always win, ignore-on-silent-exit
-# never suppresses a real error.
+# Silent-boolean-test idioms (#474, tightened #490): commands whose ONLY
+# failure signal is a non-zero exit code with empty stdout/stderr — by
+# design, not an error. After #473 closed the silent-failure blind spot,
+# these idioms became noise. Apply ONLY when matched_patterns is exactly
+# ["exit_code=..."] (no stdout or stderr pattern matched) — pattern
+# matches always win, ignore-on-silent-exit never suppresses a real error.
+#
+# #490 changes vs #474:
+#   - Removed `^\s*if\s+\[` and `^\s*if\s+\[\[` (gap #3): outer `if [` matched
+#     even when the body — a real deploy step — was the failing command, so
+#     silenced genuine failures. The `[ ... ]` regex still covers bare
+#     condition-only commands; if-bodies now log on their own merit.
+#   - Removed redundant `^\s*\[\[` (micro-cleanup): `^\s*\[` already matches
+#     `[[` (both start with `[`).
+#   - Extended grep regex to handle split-flag forms (`grep -E -q ...`) in
+#     addition to combined-flag forms (`grep -qE`, `grep -Eq`).
 SILENT_BOOLEAN_TEST_PATTERNS = [
-    re.compile(r"^\s*\["),  # POSIX `[ ... ]` test
-    re.compile(r"^\s*\[\["),  # Bash `[[ ... ]]` test
+    re.compile(r"^\s*\["),  # POSIX `[ ... ]` test (also matches `[[`)
     re.compile(r"^\s*test\b"),  # explicit `test` builtin
-    re.compile(r"\bgrep\s+-[a-zA-Z]*q"),  # grep -q (and -qE, -qi, etc.)
+    # grep with -q anywhere in the flag sequence: combined (-qE, -Eq, -qi)
+    # or split (-E -q, -i -q). Non-capturing repeated `-flags \s+` prefix
+    # allows any number of intermediate flag tokens before the -*q* terminal.
+    re.compile(r"\bgrep\s+(?:-[a-zA-Z]+\s+)*-[a-zA-Z]*q"),
     re.compile(r"^\s*(pgrep|pkill)\b"),  # process find/signal, exit 1 on no match
     re.compile(r"^\s*(which|command\s+-v)\b"),  # which/command -v not-found
-    re.compile(r"\bdiff\s+--quiet\b"),  # diff --quiet: exit 1 on differences
-    re.compile(r"\bgit\s+diff\s+--quiet\b"),  # git diff --quiet: same
-    re.compile(r"^\s*if\s+\["),  # `if [ ... ]; then ...; fi` whole-conditional
-    re.compile(r"^\s*if\s+\[\["),  # `if [[ ... ]]; then ...; fi`
+]
+
+# Idioms with exit-code-conditional skip (#490 gap #2): these tools use
+# exit_code as a SIGNAL channel where some codes are by-design and others
+# are real errors. Listed here as (pattern, by_design_codes) so the skip
+# only fires when the exit_code is in the by-design set. Any other exit
+# code falls through to LOG.
+#
+# `diff` and `git diff --quiet` exit-code semantics:
+#   0 — identical
+#   1 — files differ (by-design idiom; suppress)
+#   2 — trouble (couldn't open, permission denied, etc.; LOG)
+SILENT_BOOLEAN_TEST_PATTERNS_BY_EXIT_CODE = [
+    (re.compile(r"\bdiff\s+--quiet\b"), {1}),
+    (re.compile(r"\bgit\s+diff\s+--quiet\b"), {1}),
 ]
 
 
-def _is_silent_boolean_test(command: str) -> bool:
+def _is_silent_boolean_test(command: str, exit_code: int = 0) -> bool:
     """Return True if command matches a documented silent-boolean-test idiom.
 
     Caller must additionally verify that the ONLY failure signal is exit_code
     (no stdout/stderr pattern match) — pattern matches always take precedence.
+
+    For exit-code-conditional idioms (diff --quiet, git diff --quiet), the
+    skip only fires when exit_code is in the by-design set documented for
+    that tool. Any other non-zero code falls through to LOG.
     """
     for pattern in SILENT_BOOLEAN_TEST_PATTERNS:
         if pattern.search(command):
+            return True
+    for pattern, by_design_codes in SILENT_BOOLEAN_TEST_PATTERNS_BY_EXIT_CODE:
+        if pattern.search(command) and exit_code in by_design_codes:
             return True
     return False
 
@@ -180,7 +210,9 @@ def check(input_data: dict) -> dict | None:
     # boolean-test idioms whose failure branch is by-design. Stderr/stdout
     # pattern matches always win — a `[ -f x ]` that somehow emits a real
     # Traceback still logs.
-    if matched_patterns == [f"exit_code={exit_code}"] and _is_silent_boolean_test(command):
+    if matched_patterns == [f"exit_code={exit_code}"] and _is_silent_boolean_test(
+        command, exit_code
+    ):
         return None
 
     error_lines = _extract_error_lines(combined_output)
--- a/.claude/hooks/tests/test_annunaki_monitor.py
+++ b/.claude/hooks/tests/test_annunaki_monitor.py
@@ -261,15 +261,22 @@ class SilentBooleanTestIdiomTests(unittest.TestCase):
         result = am.check(_bash_event("git diff --quiet", exit_code=1))
         self.assertIsNone(result)
 
-    def test_if_bracket_conditional_silent_exit_not_logged(self):
-        """NEG: `if [ -f x ]; then ...; fi` whole conditional with exit-1
-        false-branch → no log."""
-        result = am.check(_bash_event("if [ -f /nonexistent ]; then echo yes; fi", exit_code=1))
+    def test_if_bracket_conditional_false_branch_exit_zero_not_logged(self):
+        """NEG: `if [ -f /nonexistent ]; then echo unreachable; fi` exits 0
+        (false branch, body skipped) → no log. The whole-`if` regex was
+        removed in #490 because it silenced body failures (gap #3); the
+        false-branch path is now handled by exit_code=0 not being an error,
+        not by idiom-skip."""
+        result = am.check(
+            _bash_event("if [ -f /nonexistent ]; then echo unreachable; fi", exit_code=0)
+        )
         self.assertIsNone(result)
 
-    def test_if_double_bracket_conditional_silent_exit_not_logged(self):
-        """NEG: `if [[ ... ]]; then ...; fi` shape → no log."""
-        result = am.check(_bash_event("if [[ -f /nonexistent ]]; then echo yes; fi", exit_code=1))
+    def test_if_double_bracket_conditional_false_branch_exit_zero_not_logged(self):
+        """NEG: same shape with `[[`. exit 0 → no log."""
+        result = am.check(
+            _bash_event("if [[ -f /nonexistent ]]; then echo unreachable; fi", exit_code=0)
+        )
         self.assertIsNone(result)
 
     # --- Precedence: pattern match wins over idiom-on-silent-exit ---
@@ -374,6 +381,113 @@ class AinoFollowupTests(unittest.TestCase):
         )
         self.assertIsNone(result, "_should_ignore must precede exit-code capture")
         self.assertFalse(self._errors_path.exists())
+
+
+class IdiomTighteningTests(unittest.TestCase):
+    """#490 coverage: the 3 gaps + micro-cleanup Aisha surfaced on PR #481's
+    secondary review. Headline regression case is gap #3 — `if [ ... ]; then
+    real-failure; fi` must log the body's exit-1 failure, not be silenced."""
+
+    def setUp(self):
+        self._saved_env = {
+            "ENVIRONMENT": os.environ.pop("ENVIRONMENT", None),
+            "NOORIN_HOOK_TEST_MODE": os.environ.pop("NOORIN_HOOK_TEST_MODE", None),
+        }
+        am._seen_hashes.clear()
+        self._tmpdir = tempfile.TemporaryDirectory()
+        self._errors_path = Path(self._tmpdir.name) / "errors.jsonl"
+        self._orig_monitor_file = am.ERRORS_FILE
+        self._orig_log_file = alog.ERRORS_FILE
+        am.ERRORS_FILE = self._errors_path
+        alog.ERRORS_FILE = self._errors_path
+
+    def tearDown(self):
+        am.ERRORS_FILE = self._orig_monitor_file
+        alog.ERRORS_FILE = self._orig_log_file
+        self._tmpdir.cleanup()
+        for k, v in self._saved_env.items():
+            if v is None:
+                os.environ.pop(k, None)
+            else:
+                os.environ[k] = v
+
+    # --- Gap #1: split-flag grep -q forms ---
+
+    def test_grep_split_flag_dash_E_dash_q_silent_exit_not_logged(self):
+        """NEG (#490 gap #1): `grep -E -q pattern file` exit 1 silent → no
+        log. Split-flag form was missed by the combined-flag-only regex
+        before tightening."""
+        result = am.check(_bash_event("grep -E -q '^foo' /tmp/file.txt", exit_code=1))
+        self.assertIsNone(result, "split-flag `grep -E -q` must skip")
+
+    def test_grep_split_flag_dash_i_dash_q_silent_exit_not_logged(self):
+        """NEG (#490 gap #1): `grep -i -q pattern file` exit 1 silent → no
+        log. Order-independent."""
+        result = am.check(_bash_event("grep -i -q bar /tmp/file.txt", exit_code=1))
+        self.assertIsNone(result)
+
+    # --- Gap #2: diff --quiet exit_code discrimination ---
+
+    def test_diff_quiet_exit_2_trouble_logged(self):
+        """POS (#490 gap #2): `diff --quiet a b` exit 2 (couldn't open,
+        permission denied) → MUST log even with empty stderr. Exit 2 is
+        "trouble", not by-design differs-signal."""
+        result = am.check(_bash_event("diff --quiet /tmp/a /tmp/b", exit_code=2))
+        self.assertIsNotNone(result, "diff --quiet exit=2 trouble must log")
+        rec = _read_records(self._errors_path)[0]
+        self.assertEqual(rec["exit_code"], 2)
+        self.assertIn("exit_code=2", rec["matched_patterns"])
+
+    def test_git_diff_quiet_exit_2_trouble_logged(self):
+        """POS (#490 gap #2): git diff --quiet variant of the trouble exit
+        discrimination. Exit 2 → log."""
+        result = am.check(_bash_event("git diff --quiet", exit_code=2))
+        self.assertIsNotNone(result, "git diff --quiet exit=2 trouble must log")
+
+    def test_diff_quiet_exit_1_differs_still_skipped(self):
+        """NEG (#490 gap #2 regression guard): exit 1 (the by-design
+        differs-signal) → still skipped per #474. The fix narrows the skip
+        condition, doesn't broaden the log condition."""
+        result = am.check(_bash_event("diff --quiet /tmp/a /tmp/b", exit_code=1))
+        self.assertIsNone(result, "diff --quiet exit=1 differs is still by-design")
+
+    # --- Gap #3 HEADLINE: if [ ...; ]; then real-failure; fi must log ---
+
+    def test_if_bracket_with_failing_body_logged(self):
+        """POS (#490 gap #3 HEADLINE): `if [ -d /tmp ]; then false; fi`
+        exit 1 (body failed, condition was truthy) → MUST log. This was the
+        regression case — the old `^\\s*if\\s+\\[` regex matched the outer
+        `if [` and suppressed the real body failure."""
+        result = am.check(_bash_event("if [ -d /tmp ]; then false; fi", exit_code=1))
+        self.assertIsNotNone(
+            result, "if-conditional body failure must log; idiom-skip must not silence it"
+        )
+        rec = _read_records(self._errors_path)[0]
+        self.assertEqual(rec["exit_code"], 1)
+
+    def test_if_double_bracket_with_failing_body_logged(self):
+        """POS (#490 gap #3): `[[` variant of the headline case → log."""
+        result = am.check(_bash_event("if [[ -d /tmp ]]; then false; fi", exit_code=1))
+        self.assertIsNotNone(result)
+
+    def test_if_bracket_false_condition_exit_zero_skipped(self):
+        """NEG (#490 gap #3 invariant): false condition (exit 0 because
+        else-branch is empty / body not entered) → no log. Preserves the
+        existing "happy path of the false branch is silent" behavior."""
+        result = am.check(
+            _bash_event("if [ -d /nonexistent ]; then echo unreachable; fi", exit_code=0)
+        )
+        self.assertIsNone(result, "false-condition exit-0 stays silent")
+
+    # --- Micro-cleanup: redundant `^\s*\[\[` removal regression guard ---
+
+    def test_bash_double_bracket_still_skipped_after_dedup(self):
+        """NEG (#490 micro-cleanup regression guard): `[[ -f x ]]` exit 1
+        silent → no log. The line-79 `^\\s*\\[\\[` regex was removed because
+        `^\\s*\\[` already matches `[[` (both start with `[`). This test
+        confirms `[[` is still covered by the dedup'd list."""
+        result = am.check(_bash_event("[[ -f /nonexistent ]]", exit_code=1))
+        self.assertIsNone(result, "[[ idiom must still skip after dedup")
 
 
 if __name__ == "__main__":


parametrization:
Requestor: Imelda Santos
Requestee: Lucas Ferreira
RequestOrReplied: Approved
TechDebt: none

## Data-engineer 2° verdict — Approved

Reviewed at HEAD `ef87a8c` via `gh api contents?ref=…`. Nurul carries primary; my angles per brief: failure-direction analysis, test-update integrity (highest-risk), new data-structure call-site contract, backward-compat of `_is_silent_boolean_test`, and test-coverage-vs-acceptance walk. All 5 angles pass.

### Angle 1: Failure-direction analysis (gap #3 specifically)

The gap #3 fix removes `^\s*if\s+\[` and `^\s*if\s+\[\[` from `SILENT_BOOLEAN_TEST_PATTERNS`. Lucas claims this errs on the side of over-logging (safe direction). Verified at HEAD:

The `check()` predicate at line 187 reads `if exit_code and exit_code != 0` BEFORE setting `is_error = True`. At line 205-206 it returns None when `is_error` is False. So an exit-0 if-conditional (false condition + empty body, e.g., `if [ -d /nonexistent ]; then echo unreachable; fi`) NEVER reaches the idiom-skip predicate at line 213 — it returns None at line 206 via the exit-0 path, not via idiom-skip.

That means removing `^\s*if\s+\[` from the idiom list **cannot affect exit-0 false-branch handling at all**. The removal only changes behavior for exit-non-zero if-conditionals with no stderr — i.e., the exact regression case where `if [ truthy ]; then false; fi` exits 1 silently. Those now LOG instead of being silenced. Failure direction analysis is correct: the change moves a wrongly-silenced class into the logging path without affecting any other class.

No other side-effects from the removal: the bare `[ ... ]` / `[[ ... ]]` test patterns are still covered by the retained `^\s*\[` regex (which matches both `[` and `[[`).

### Angle 2: Test-update integrity (HIGHEST-RISK ANGLE)

Lucas renamed + reframed 2 tests (`test_if_bracket_conditional_silent_exit_not_logged` and `_double_bracket_` sibling). Compared baseline (origin/wave-11) vs PR HEAD test-by-test.

**Baseline** (wave-11 28-test version):
```python
def test_if_bracket_conditional_silent_exit_not_logged(self):
    """NEG: `if [ -f x ]; then ...; fi` whole conditional with exit-1
    false-branch → no log."""
    result = am.check(_bash_event("if [ -f /nonexistent ]; then echo yes; fi", exit_code=1))
    self.assertIsNone(result)
```

This baseline test was asserting that `if [ -f /nonexistent ]; then echo yes; fi` with `exit_code=1` doesn't log. But that command's actual shell exit code is **0** (false condition + no else = empty if), not 1. The baseline test was synthetic — it asserted the regex-skip behavior, not actual shell semantics. The test's framing assumption — "the whole-if regex catches this so it's silenced" — was exactly the bug.

**PR HEAD** rename:
```python
def test_if_bracket_conditional_false_branch_exit_zero_not_logged(self):
    """NEG: `if [ -f /nonexistent ]; then echo unreachable; fi` exits 0
    (false branch, body skipped) → no log. The whole-`if` regex was
    removed in #490 because it silenced body failures (gap #3); the
    false-branch path is now handled by exit_code=0 not being an error,
    not by idiom-skip."""
    result = am.check(_bash_event("if [ -f /nonexistent ]; then echo unreachable; fi", exit_code=0))
    self.assertIsNone(result)
```

This now asserts the correct shell semantics (exit_code=0) and the correct mechanism (early-return at line 206, not idiom-skip). The rename is faithful: same shape, same expected outcome (no log), correct underlying reason. Name accurately describes what it asserts.

**Are any tests REMOVED that should have been re-framed?** Compared by-name diff: `SilentBooleanTestIdiomTests` had 17 tests baseline, has 17 tests at HEAD. Two renamed (counted as same slot), zero deleted. The headline regression case is added as a NEW test in the new `IdiomTighteningTests` class (`test_if_bracket_with_failing_body_logged`), not as a swap. Correct partitioning.

**Test count claim verification**: baseline `grep -cE "^\s+def test_"` on origin/wave-11 = 28. PR HEAD = 37. Diff = +9 (new class) + 0 (renames preserve count) + 0 (deletions) = 9 new. Lucas's "37 = 28 baseline + 9 new" claim is exact.

### Angle 3: Data-structure addition — additive vs replace

The new `SILENT_BOOLEAN_TEST_PATTERNS_BY_EXIT_CODE` list is scanned ADDITIVELY in `_is_silent_boolean_test` — first the unconditional list (any-exit-code skip), then the conditional list (exit-in-by-design-codes skip). However, the `diff --quiet` and `git diff --quiet` patterns were **moved** from the unconditional list (where they previously matched any non-zero exit) to the conditional list (where they only match exit_code ∈ {1}). The net contract change:

- **Before**: diff/git-diff `--quiet` skipped on ANY non-zero exit including exit=2 trouble (information loss for trouble exits).
- **After**: diff/git-diff `--quiet` skips ONLY on exit=1 (the by-design differs-signal); exit=2 falls through to LOG (correct trouble handling).

This is the right contract for gap #2. Empirically verified:
```
SKIP   exit=1  'diff --quiet a b'
LOG    exit=2  'diff --quiet a b'
SKIP   exit=1  'git diff --quiet'
LOG    exit=2  'git diff --quiet'
LOG    exit=1  'diff a b'              (never matched, not --quiet form)
```

No regressions to other unconditional patterns — they remain in the original list and skip on any non-zero exit as before.

### Angle 4: Backward-compat of `_is_silent_boolean_test`

`grep -nE "_is_silent_boolean_test" origin/deployments/phase-3/wave-11` returns exactly 2 hits: the definition (line 91 baseline → 114 PR HEAD) and the single call site (line 183 baseline → 213 PR HEAD). Both are inside `annunaki_monitor.py`. The function is module-private (underscore prefix), no external callers.

Lucas updated the sole call site to pass `exit_code` positionally. The new `exit_code: int = 0` default is irrelevant for the in-tree caller (always explicit now) but makes the signature non-breaking for any hypothetical future caller. Backward-compat is trivially preserved — there ARE no other callers to break.

Test-side: the new `IdiomTighteningTests` class doesn't call `_is_silent_boolean_test` directly (it tests via `am.check()`), so no test-side breakage either.

### Angle 5: Test-coverage-vs-acceptance walk

| Acceptance bullet | Covering test(s) | Verified |
|---|---|---|
| `grep -E -q ... exit=1 no stderr → SKIP` | `test_grep_split_flag_dash_E_dash_q_silent_exit_not_logged` (line 416), `test_grep_split_flag_dash_i_dash_q_silent_exit_not_logged` (line 423) | ✓ |
| `diff --quiet exit=2 no stderr → LOG` (distinguishes from exit=1) | `test_diff_quiet_exit_2_trouble_logged` (line 431), `test_git_diff_quiet_exit_2_trouble_logged` (line 441), `test_diff_quiet_exit_1_differs_still_skipped` (line 447 regression-guard) | ✓ |
| `if [ -d /tmp ]; then false; fi exit=1 → LOG` (HEADLINE) | `test_if_bracket_with_failing_body_logged` (line 456), `test_if_double_bracket_with_failing_body_logged` (line 468) | ✓ |
| `if [ -d /nonexistent ]; then echo unreachable; fi exit=0 → SKIP` (preserved-behavior) | `test_if_bracket_false_condition_exit_zero_skipped` (line 473) + renamed `test_if_bracket_conditional_false_branch_exit_zero_not_logged` + double-bracket sibling | ✓ |
| Existing 28 tests pass | All retained, 2 renamed-not-removed | ✓ |
| Redundant `^\s*\[\[` removed | Diff confirms removal; `test_bash_double_bracket_still_skipped_after_dedup` (line 484) regression-guards that `[[` is still covered by the surviving `^\s*\[` | ✓ |

**Bonus regression-guard observation**: the combined-flag grep coverage (`-qE`, `-Eq`) is implicitly preserved by the retained baseline `test_grep_q_with_flags_silent_exit_not_logged`. I empirically self-checked the new regex `\bgrep\s+(?:-[a-zA-Z]+\s+)*-[a-zA-Z]*q` against 9 inputs (combined, split-leading, split-trailing, multi-flag, bare-no-q variants) — all classifications correct. Regex broadening is sound.

### Disposition

**Approved.** All 4 gaps closed with correct failure direction, test-update integrity confirmed (rename pair correctly maps old buggy assertion to new correct semantics, zero net deletions, count claim exact), data-structure addition is additive in code with intentional movement of diff/git-diff patterns into the conditional list, backward-compat is trivially preserved (sole call site updated), and 100% acceptance-bullet test coverage.

Nurul holds primary. Hook 4 2-of-2 gate: I'm in.

parametrization:
Requestor: Nurul Hakim
Requestee: Lucas Ferreira
RequestOrReplied: Approved
TechDebt: None added. All 4 scope items from #490 implemented cleanly with strict-extension semantics on grep regex (no false-positives introduced), correct exit-code discrimination via new `SILENT_BOOLEAN_TEST_PATTERNS_BY_EXIT_CODE` data structure, and test renames that demonstrably ASSERT the preserved invariant rather than stealth-removing coverage.

## Observability-engineer verdict — Approved (primary)

**Reviewer:** Nurul Hakim (Observability) — **primary**
**Secondary:** Imelda (in flight per team-lead brief)
**Head verified:** `ef87a8c79f1ef2a7d45edc34dd9708aec27bd0bc` via `gh api repos/.../contents/<path>?ref=ef87a8c` + `gh pr diff 495` per [[feedback_review_against_artifact_not_framing]] + [[feedback_reviewer_must_not_branch_switch_parent]]. Branch fetched as `pull/495/head` ref-only for `git show`-based local test verification; deleted post-extraction. No parent-branch checkout.

### Operational read

Surgical tightening of the `SILENT_BOOLEAN_TEST_PATTERNS` idiom-skip logic Aisha flagged on PR #481. Three correctness gaps + one micro-cleanup, all implemented with minimum-diff and self-documenting commit messages. The headline win (gap #3) closes the most concerning regression — `if [ -d /tmp ]; then real-failure; fi` no longer silently swallows the body's exit-1. The new `SILENT_BOOLEAN_TEST_PATTERNS_BY_EXIT_CODE` data structure is a clean discrimination primitive that we'll want again when more tools join the "exit-code-as-signal-channel" class.

### Surface verification at HEAD (2 files, +168/-22)

| Path | Verdict | Notes |
|---|---|---|
| `.claude/hooks/annunaki_monitor.py` (+47/-15) | OK | Pattern list refactor (5→5 entries; 2 removed, 1 widened); new `SILENT_BOOLEAN_TEST_PATTERNS_BY_EXIT_CODE` list (2 entries); `_is_silent_boolean_test(command, exit_code=0)` signature widened with optional kwarg (default 0 = safe back-compat); call site at L213-215 passes `exit_code` through. |
| `.claude/hooks/tests/test_annunaki_monitor.py` (+121/-7) | OK | 2 existing tests renamed-and-reframed (preserved-invariant pivot, see CRITICAL ANGLE below); 9 new tests in `IdiomTighteningTests` class covering all 4 acceptance bullets; standard setUp/tearDown env+ERRORS_FILE isolation matches existing test classes. |

### Angle-by-angle (the 4 scope items + CRITICAL angle)

**1. Gap #1 — split-flag grep regex (strict extension):**

New: `\bgrep\s+(?:-[a-zA-Z]+\s+)*-[a-zA-Z]*q`. I independently regression-tested all forms:

| Input | Pre-fix match? | Post-fix match? | Verdict |
|---|---|---|---|
| `grep -q foo file` (bare) | ✓ | ✓ | preserved |
| `grep -qE`, `-Eq`, `-qq`, `-qi` (combined) | ✓ | ✓ | preserved |
| `grep -E -q foo file` (split) | ✗ | ✓ | NEW (gap #1) |
| `grep -i -q foo file` (split) | ✗ | ✓ | NEW (gap #1) |
| `grep -e -i -q foo file` (3 split flags) | ✗ | ✓ | NEW (graceful extension) |
| `grep foo file` (no -q) | ✗ | ✗ | correctly excluded |
| `grep -v foo` / `grep -E foo` (no -q anywhere) | ✗ | ✗ | correctly excluded |
| `egrep -q foo` / `zgrep -q foo` (boundary check via `\b`) | ✗ | ✗ | correctly excluded — `\b` works |
| `xargs grep -q foo` (compound) | ✓ | ✓ | preserved |

**Strict-extension confirmed: all 5 original combined-flag forms still match; 3 new split-flag forms now match; no false-positives on adjacent commands (egrep/zgrep boundary).**

**2. Gap #2 — `diff --quiet` exit_code=2 discrimination:**

New `SILENT_BOOLEAN_TEST_PATTERNS_BY_EXIT_CODE = [(pattern, {1}), ...]` structure with exit-code-conditional `pattern.search(command) and exit_code in by_design_codes` predicate at L80-81. I independently tested the predicate logic:

| Command | exit_code | Behavior |
|---|---|---|
| `diff --quiet a b` | 0 (identical) | LOG (also wouldn't reach this path — annunaki only fires on failure) |
| `diff --quiet a b` | 1 (differs, by-design) | **SKIP** ✓ |
| `diff --quiet a b` | 2 (trouble — couldn't open, perm denied) | **LOG** ✓ (gap #2 fix) |
| `diff --quiet a b` | 3, 127 (unknown / not-found) | LOG ✓ (correctly falls through) |
| `git diff --quiet` | 1 / 2 | SKIP / LOG (symmetric) |
| `diff a b` (non-quiet) | any | LOG (regex correctly excludes) |

**Data structure correctly narrows skip to documented by-design exit_code; no regression on other exit codes (0, 3+).** The `{1}` set could grow if other tools join the family (`xmllint --noout` exits 1 on syntax error, 2 on internal error, etc.) — clean extension primitive.

**3. Gap #3 HEADLINE — if-body removal + preserved bare-bracket coverage:**

Both `^\s*if\s+\[` AND `^\s*if\s+\[\[` removed from `SILENT_BOOLEAN_TEST_PATTERNS`. Verified the surviving `^\s*\[` regex still covers ALL bare-condition forms:

| Input | Matches `^\s*\[`? | Correct? |
|---|---|---|
| `[ -f /nonexistent ]` (POSIX bare) | ✓ | yes — preserved |
| `[[ -f /nonexistent ]]` (bash double-bracket bare) | ✓ | yes — `\[` matches `[` of `[[` |
| `  [ -f x ]` (indented bare) | ✓ | yes — `\s*` allows leading whitespace |
| `if [ -d /tmp ]; then false; fi` (whole-conditional with body failure) | ✗ | **yes — gap #3 fix** (body failures now LOG on own merit) |
| `if [[ -d /tmp ]]; then false; fi` (double-bracket variant) | ✗ | yes — gap #3 fix |
| `echo hello && [ -f x ]` (compound, bracket NOT line-start) | ✗ | yes — line-anchored, correctly excluded (was always so) |

**Bracket conditions outside if-bodies still skip as expected; whole-`if` no longer suppresses body failures.** The acceptance-criterion-4 invariant (`if [ false-cond ]; then ...; fi` exit=0 → no log) is handled by exit_code=0 not being an error signal in the first place (returns at L205-206 `if not is_error: return None`), not by idiom-skip — Lucas's docstring at the renamed test L264-269 correctly explains this.

**4. Micro-cleanup — `^\s*\[\[` dedup:**

`^\s*\[\[` removed from the pattern list. The OR-loop `for pattern in SILENT_BOOLEAN_TEST_PATTERNS: if pattern.search(command): return True` short-circuits on first match. Since `^\s*\[` matches `[[` (both start with `[`), the second pattern was dead code. Verified `[[ -f /nonexistent ]]` still matches via `^\s*\[` — regression guard `test_bash_double_bracket_still_skipped_after_dedup` (L484) pins this.

### CRITICAL ANGLE — test-update divergence (Lucas flagged proactively)

Lucas renamed 2 existing tests to assert the **preserved-behavior invariant** (acceptance criterion 4) rather than the OLD whole-`if` suppression. I verified this is NOT a stealth removal:

| Test | Status | Assertion | Coverage |
|---|---|---|---|
| `test_if_bracket_conditional_silent_exit_not_logged` (OLD) | Renamed → `test_if_bracket_conditional_false_branch_exit_zero_not_logged` (L264) | `assertIsNone(result)` for `if [ -f /nonexistent ]; then echo unreachable; fi` exit=0 | False-condition path, body skipped, exit=0 — preserved-invariant per #490 acceptance bullet 4 |
| `test_if_double_bracket_conditional_silent_exit_not_logged` (OLD) | Renamed → `test_if_double_bracket_conditional_false_branch_exit_zero_not_logged` (L275) | `assertIsNone(result)` for same shape with `[[`, exit=0 | Symmetric coverage |

**Reframing is correct discipline.** The OLD test asserted exit=1 → skip via the whole-`if` regex — but THAT IS THE BUG #490 fixes. Preserving the test verbatim would re-assert the regression. The reframed tests assert that the legitimate false-branch case (false condition → body not entered → exit=0 → no log) still works AS A CONSEQUENCE of the exit-zero-not-an-error guard, not via idiom-skip. Both renamed tests still exist as identifiable test functions with meaningful `assertIsNone` assertions backed by full docstring rationale. **Lucas's justification holds.**

### #490 acceptance bullet cross-walk

| Bullet | Test (or path) covering it |
|---|---|
| (1) `grep -E -q ...` exit=1 → SKIP | `test_grep_split_flag_dash_E_dash_q_silent_exit_not_logged` (L416) + `_dash_i_dash_q` sibling (L423) |
| (2) `diff --quiet` exit=2 → LOG (distinguish from exit=1) | `test_diff_quiet_exit_2_trouble_logged` (L431) + `git_diff_quiet_exit_2_trouble_logged` (L441) + regression guard `test_diff_quiet_exit_1_differs_still_skipped` (L447) |
| (3) HEADLINE: `if [ -d /tmp ]; then false; fi` exit=1 → LOG | `test_if_bracket_with_failing_body_logged` (L456) + double-bracket sibling (L468) |
| (4) `if [ -d /nonexistent ]; then echo unreachable; fi` exit=0 → SKIP (preserved) | `test_if_bracket_false_condition_exit_zero_skipped` (L473) + renamed siblings at L264 + L275 |
| (5) Existing 28 tests continue to pass | All 28 baseline + 9 new = 37/37 PASS locally (see independent verification below) |
| (6) Redundant `^\s*\[\[` removed | Removed at L33-34 of diff; regression guard `test_bash_double_bracket_still_skipped_after_dedup` (L484) |

**All 6 acceptance bullets covered.** 9 new IdiomTighteningTests + 2 renamed-and-reframed baseline tests = 11 tests directly targeting #490 scope.

### Independent local verification

- Pulled PR head via `git fetch refs/pull/495/head:pr495_review_only` (ref-only; deleted post-read per [[feedback_reviewer_must_not_branch_switch_parent]])
- Materialized to `/tmp/pr495_test/` via `git show`-equivalent file copies + sibling hooks from main (annunaki_log etc.)
- `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/test_annunaki_monitor.py -v` → **`37 passed in 0.08s`**
- Matches PR body claim (28 baseline + 9 new); no regressions
- All 9 IdiomTighteningTests pass; both renamed baseline tests pass
- Branch `pr495_review_only` deleted post-extraction

### CI matrix at HEAD (4 jobs, all green at `ef87a8c`)

- ✓ Ruff lint (8s)
- ✓ Ruff format check (9s)
- ✓ Mypy type check (12s)
- ✓ Smoke test — hooks compile and exit cleanly (8s)

### Observability-engineer sweep (non-blocking)

Nothing surfaced as merge-blocking. Two future-extension notes worth mentioning but NOT filing — Lucas can fold in if useful or ignore:

1. **`SILENT_BOOLEAN_TEST_PATTERNS_BY_EXIT_CODE` extension candidates.** Several other tools have the same exit-code-as-signal-channel shape:
   - `xmllint --noout` exits 1 on syntax error, 2 on internal error
   - `find ... -quit` early-termination patterns
   - `jq -e expr` exits 1 on null/false, but >1 on error
   - `grep -c` (with count) is in the same exit_code=1=no-match family
   
   None are urgent — file as W12 sweep if any surface as noise in `errors.jsonl`. The data structure is the right shape to accommodate them.

2. **Observability blind spot: no metric for idiom-skip catches.** The `_is_silent_boolean_test` predicate suppresses logs silently — we don't have a counter for "how often does the idiom filter fire" so we can't tell if a pattern stops matching due to a syntax change (e.g., grep adopting a new short-flag convention) and noise re-appears. Cleanest hookup mirrors the existing dedup-hash bookkeeping: track skip counts per pattern, emit to `errors.jsonl` summary at session end. **File as W12 follow-up if observability budget permits.**

### Verdict

**Approved.** Surgical tightening of the idiom-skip logic with strict-extension semantics on the grep regex, a clean exit-code-discrimination primitive for the diff/git-diff family, and a correct removal of the over-broad whole-`if` pattern that was the gap #3 regression source. Test additions cover all 4 acceptance bullets; the 2 renamed baseline tests demonstrably assert the preserved-invariant rather than stealth-removing coverage. CI all-green at HEAD; local pytest run confirms 37/37 pass in 0.08s.

Closes #490.

Issue #510, opened 2026-05-19
tech-debt(hook): validate_review_comment_format — extract_repo_from_command should accept all 4 gh flag forms (#509 follow-up)


parametrization:
## Background

Surfaced during review of #509 (`fix(hook): validate_review_comment_format — forward --repo to internal gh pr view`).

#509 closes the cross-repo false-block (#503) by parsing `--repo <value>` from the user's bash command and threading it through `get_branch_name()`. The new `extract_repo_from_command()` regex is `--repo\s+(\S+)` — space-form only.

## Gap

Three other forms gh CLI accepts that the new parser does NOT handle:

1. **`--repo=OWNER/NAME`** — equals form (pinned as not-supported by `test_with_equals_form_not_supported`)
2. **`-R OWNER/NAME`** — short-flag space form
3. **`-R=OWNER/NAME`** — short-flag equals form

`gh` CLI documents both `--repo` and `-R` and accepts both space and equals separators.

## Why this matters

`validate_labels.py:extract_repo()` ALREADY handles all four forms via the regex `(?:^|\s)(?:--repo|-R)(?:=|\s+)(\S+)`. So:

- Two `--repo` parsers across the hooks dir use different recognition sets — future maintainers cribbing from whichever they hit first will land inconsistencies.
- If an agent ever uses `--repo=` (gh-valid) or `-R` (gh-valid short form) on a `gh pr comment` to a cross-repo PR, the hook silently falls back to cwd-default resolution + emits the stderr breadcrumb #509 added. **The breadcrumb is detectable in transcripts but the cross-repo block still misfires** — the original #503 class of bug recurs through these alternate-flag-form paths.

Charter convention (`pull-requests.md:47`) uses space form, so this is latent risk rather than current bug. But the failure direction matches #503 exactly.

## Proposed fix

Replace `extract_repo_from_command()` in `validate_review_comment_format.py` with the same regex used by `validate_labels.py:extract_repo()`:

```python
match = re.search(r"(?:^|\s)(?:--repo|-R)(?:=|\s+)(\S+)", command)
return match.group(1) if match else None
```

Better: consolidate the two parsers into a shared helper (e.g., `.claude/hooks/_repo_flag_parse.py`) to eliminate the divergence at the source. Mirrors the `_shell_parse` / `_wave_label_parse` shared-helper pattern already in the repo.

Add three additional tests to pin all four forms:
[...]

PR #508, opened 2026-05-19
tech-debt(hook-4): validate Requestor against roster (closes #498)

parametrization:
## Summary

Closes Hook 4 ↔ `validate_commit_identity` asymmetry that allowed non-roster Requestor strings to satisfy the 2-reviewer gate. P3W11 PR #487 surfaced the failure: verdict comments under fictional personas "Camila Restrepo" and "Imelda Santos" — neither in `.claude/team/roster/` — passed Hook 4's distinct-Requestor count, and #487 merged. `validate_commit_identity` already enforces strict-roster discipline for committer identity; this PR aligns Hook 4 to the same standard.

Closes #498

## What changed

**`validate_pr_review.py`**

- New `_load_roster_names()` returns ALL canonical persona names from the local roster's `**Name:**` lines (no role filter).
- Refactored `load_charter_enforcer_names` and `_load_roster_names` to share a `_iter_roster_entries(role_prefix_filter=...)` parser — `None` filter loads the full roster, the prefix tuple loads charter-enforcer subset.
- In `check()`, comment-based Requestor strings are now filtered against `_load_roster_names()` BEFORE counting toward the 2-of-2 threshold. Formal GitHub reviews (`pr_data["reviews"]`) are NOT filtered — those are authenticated GitHub logins, not persona names.
- BLOCK reason prepended with a dedicated diagnostic when non-roster names caused the shortfall: lists rejected strings, names the first 20 valid roster members, and references the asymmetry-closing rationale so an operator can either re-post under a real persona or amend the roster if a new member is joining.
- Fail-closed when the roster dir is empty / unreadable, with explicit `"<empty — local roster dir could not be read>"` diagnostic per `feedback_safety_direction_over_ux_friction.md`.

**`test_validate_pr_review.py`** — 9 new tests + 2 reworked fixtures

5 `LoadRosterNamesTests` (helper coverage):
- `test_includes_standards_lead` / `test_includes_program_director` / `test_includes_tpm` / `test_includes_release_coordinator` (real parent-roster names)
- `test_includes_engineer_roles` — non-enforcer roles ARE in the roster (unlike `load_charter_enforcer_names`)

4 `RosterValidationGateTests` (end-to-end gate behavior):
- `test_two_non_roster_requestors_blocked` — exact PR #487 repro (Camila + Imelda); verifies BLOCK + names appear in diagnostic
[...]

PR #512, opened 2026-05-19
fix(hook-4): validate_pr_review regex matches only structured trailer block (closes #511)

### LINKED ISSUE #511: tech-debt(hook-4): validate_pr_review regex prose-false-match (3 instances P3W11 batch 11)
## Background

`validate_pr_review` (PR Hook 4) uses a greedy-first-match regex to extract `Requestor:` and `RequestOrReplied:` field values from verdict comments. When a reviewer's prose body describes the PR's structured-fields trailer (e.g., quoting another PR's `Requestor: ...` block or noting "the PR body has `Requestor: (TBD) / Requestee: ...`"), the regex first-matches the PROSE mention before the actual structured block at the end. Result: the reviewer's verdict is misclassified.

Failure direction: **unsafe (false-block)**. A valid Approved verdict appears as something else (e.g., `Requestor: (TBD — orchestrator will assign)` or `Requestor: / Requestee: / RequestOrReplied: New / TechDebt: none`), reducing the distinct-Requestor count and blocking merge.

## Instances (2026-05-18 / 2026-05-19 session)

3 occurrences in one session, all required orchestrator REST PATCH to merge:

1. **main#509** (Wanjiku reviewing Aino's --repo fix) — prose line described "the literal bare-line block (Requestor / Requestee / RequestOrReplied: New / TechDebt: none)". Hook captured Requestor as the rest-of-line garbage; merge blocked at 1/2.

2. **deploy#337** (Lucas reviewing Aisha Idrissi's conftest guard) — soft observation noted "Aisha's PR body does not include the trailer structured-fields block (`Requestor: / Requestee: / RequestOrReplied: New / TechDebt: none` bare-line)". Hook captured Requestor as the rest-of-line garbage; merge blocked at 1/2.

3. **deploy#339** (Bereket reviewing Lucas's schema v2 gate) — reviewer noted "PR body uses `Requestor: (TBD — orchestrator will assign)` — correct deferral, no reviewer prediction". Hook captured Requestor as "(TBD — orchestrator will assign)"; merge blocked at 1/2.

All three were post-pre-warning. Pre-warning ("don't include literal status terms in prose") wasn't sufficient — reviewers interpreted it narrowly (avoiding only Approved/New/Reply), didn't realize ANY `<Field>: ...` mention triggers the regex.

## Proposed fix options

**A. Last-match instead of first-match** — change regex to capture the LAST `Requestor: ...` line in the body. Trailer-block convention places the structured fields at the END, so last-match aligns with convention.
- Pro: Minimal code change. Aligns with existing trailer-at-end discipline.
- Con: A reviewer adding "PS" content after the trailer could re-trip. Mitigated by also requiring the trailer to be after a `---` separator.

**B. Trailer-block-only match** — only match fields inside a `---`-separated block at the end of the body. Ignore matches above the separator.
- Pro: Strictest discipline. Cleanest semantics.
- Con: Requires reviewers to use `---` separator religiously. Charter convention already prescribes this; minor enforcement uplift.

**C. Code-block-aware match** — ignore matches inside backticks/code blocks (\`\`\`...\`\`\`).
- Pro: Allows reviewers to quote-discuss field syntax naturally.
- Con: Doesn't help with prose mentions outside backticks (the actual 3 instances above were mostly NOT inside backticks).

**Combined recommendation**: B + C. Trailer-after-separator is the contract; code-block-aware handles the corner where reviewers quote fields for explanation.

## Acceptance criteria

- [ ] Hook regex updated to match only structured-trailer-block fields (B + C combined)
- [ ] Tests added covering:
  - prose `Requestor: foo` ABOVE trailer block → ignored
  - `Requestor: foo` inside `` ` `` backticks → ignored
  - last-trailer wins if multiple `---` separators present
  - existing valid trailers still match
- [ ] BLOCKED message updated to point at the 3 historical instances for future debugging
- [ ] Charter `pull-requests.md § Comment-Based Reviews` adds an explicit "do not duplicate field patterns in prose" rule (now enforced by the smarter regex)

## Severity

**Moderate** — caught in this session's 3 instances via orchestrator PATCH. Without PATCH discipline, merges would silently fail despite valid verdicts, frustrating reviewers and consuming round-trips. Hook-tier fix prevents recurrence.

## Wave/severity

Tech-debt hook fix. Could file p3-wave-11 if we want it ASAP, or W12-bound.

## Origin

P3W11 batch 11 (2026-05-19) session — Wanjiku/Lucas/Bereket all hit it during reviews of main#509, deploy#337, deploy#339.

parametrization:
## Summary

Fixes the Hook 4 regex prose-false-match that bit 3 reviewer verdicts in P3W11 batch 11 (2026-05-19). Each required orchestrator REST PATCH to merge — pre-warning ("don't duplicate field patterns in prose") wasn't enough because the greedy first-match regex captured ANY `<Field>:` mention, not just the verdict words reviewers were trained to avoid.

Closes #511

## Approach — Hybrid Option B + C

Per the issue body's recommendation (and your direction in the spawn brief), implemented both: **trailer-block-only matching with code-region stripping**, plus a back-compat fallback for legacy comments without separators.

1. **`_strip_code_regions(body)`** — zeros out fenced (` ``` ... ``` `) and inline (`` `...` ``) code regions before extraction. Replacement char is space, not empty, so line indices stay stable for downstream trailer-block detection. Closes the **deploy#339 (Bereket)** backtick-quote pattern.

2. **`_trailer_block_substring(body)`** — returns text after the LAST bare-line `---` separator (charter trailer convention). If no separator present, returns the full body for back-compat with legacy verdicts. Closes the **main#509 (Wanjiku)** and **deploy#337 (Lucas)** prose-above-trailer patterns.

3. **`_extract_charter_field`** rewritten to: strip code regions → narrow to trailer block → use **LAST-MATCH-WINS** within scope. The last-match fallback is what makes legacy comments without `---` still work — prose mentions above an end-of-body trailer are ignored because the actual trailer is the last match.

4. **BLOCK reason message** extended with a new "prose-mention of fields outside the trailer block" failure-mode section pointing at all 3 historical instances (main#509 / deploy#337 / deploy#339) for future debugging.

5. **Charter** `pull-requests.md § Comment-Based Reviews` — new bullet codifying trailer-block-only discipline + code-region stripping + the #511 origin.

## What about pure Option B (strict trailer-only, no fallback)?

I considered it but kept the legacy fallback for the no-separator case because:
- It's strictly stronger than the old regex (last-match still beats prose-above)
- It avoids retroactively breaking older verdict comments that pre-date the separator convention
- The hook surface is a gate for merge, not a comment-time block — back-compat matters

The fallback only activates when no `---` is present. With a separator, scope is strictly trailer-only.

## Test Plan

- [x] `ENVIRONMENT=test pytest .claude/hooks/tests/test_validate_pr_review.py -v` → **85 passed in 0.15s** (was 69 → +16 new)
- [x] Full hook suite: **903 passed in 1.32s** (was 887 → +16)
- [x] `uvx ruff format --check .claude/hooks/` → 64 files already formatted
- [x] `uvx ruff check .claude/hooks/` → All checks passed
- [x] Pre-commit (ruff-format / ruff-lint / mypy) green on commit
- [x] AST parse OK

## Test breakdown (16 new)

**5 `StripCodeRegionsTests`** — `inline_backticks` / `fenced_triple_backtick` / `unterminated_inline_backtick_passes_through` / `unterminated_fenced_block_strips_rest` / `multiline_inline_span_is_not_matched`

**5 `TrailerBlockSubstringTests`** — `no_separator_returns_full_body` / `one_separator_returns_post_only` / `multiple_separators_last_wins` / `separator_must_be_on_own_line` / `separator_with_surrounding_whitespace_is_recognized`

**6 `ProseFalseMatchRegressionTests`** (the 3 in-the-wild bugs + 3 edge cases):
- `test_main_509_wanjiku_prose_above_trailer_ignored` — exact-shape repro of main#509
- `test_deploy_337_lucas_prose_observation_ignored` — exact-shape repro of deploy#337
- `test_deploy_339_bereket_backtick_quote_ignored` — exact-shape repro of deploy#339
- `test_legacy_no_separator_uses_last_match_fallback` — back-compat for pre-separator verdicts
- `test_multiple_separators_only_last_trailer_matches` — edge case (reviewer adds aside block)
- `test_existing_canonical_trailer_still_works` — regression baseline (fields-only no-preamble verdict)

## Dogfooding caveat

This PR will merge through Hook 4 with the new regex. My own verdict trailer follows a `---` separator (charter standard) so the trailer-block path activates cleanly. Reviewers MUST do the same (already standard convention) — legacy fields-only-no-separator still works via last-match-wins.

---

Requestor: (TBD — orchestrator will assign reviewers)
Requestee: Aino Virtanen
RequestOrReplied: New
TechDebt: none


parametrization:
## Release Coordinator Review — PR #512

**Verdict: Approved**

Read all 3 files via `gh api contents?ref=b7090d8` (no working-tree checkout). Release-safety lens.

### 1. Back-compat fallback risk (the one I dug into most)

The hybrid B+C approach is sound. The one residual risk I see: when no bare-line `---` separator is present, last-match-wins on the FULL body — so a prose mention AFTER the real trailer (without a separator) could outscore the trailer. Walk:

- Legacy body: real fields-only trailer, then a postscript like "PS: the PR body itself uses a different reviewer-name in its trailer block" with the field syntax quoted in prose.
- Pre-fix regex: first-match-wins, would have captured the real trailer correctly.
- Post-fix regex with no separator: last-match-wins, would extract the prose mention.

This is a narrow edge case (prose AFTER real trailer with no `---`), strictly stronger than pre-fix for the dominant pre-trailer-prose shape (the 3 in-the-wild bugs), and fully eliminated once `---` is used. Aino's docstring acknowledges the trade-off ("fall back to the full body to remain backward-compatible with legacy verdict comments"). I'd call this acceptable for a v1 — the charter rule (post-merge) now mandates `---`, so the strict-trailer path is the steady-state and the fallback exists only to honor pre-charter-rule comments.

**Non-blocking suggestion**: post-merge, once a few cycles confirm `---` adoption is universal, drop the fallback in a v2 and require the separator. Not a blocker for this PR.

### 2. Code-strip safety

`_strip_code_regions` is precise:
- CommonMark-spec inline-span (no newlines inside the run)
- Unterminated fence conservatively strips the rest (safe-direction: fail-block over fail-allow)
- Replacement char is space, preserving line indices for downstream trailer detection

One concern I considered: a reviewer using backticks for emphasis on a real trailer field (unusual but possible). The strip would zero the field and the hook would block. Fail-direction is safe (block, not allow), and the charter rule prescribes plain field syntax. Not a blocker.

### 3. Charter rule clarity

Read the new bullet at `pull-requests.md` line 40. It is operator-readable:
- Names the rule (fields ONLY in trailer block, post-bare-line separator)
- Names the mechanism (post-last-separator + code-region strip)
- Names the WHY (3 false-blocks in batch 11 with explicit PR# references)
- Names the issue origin (resolves main#511)

A future reviewer encountering the rule will know what to do and why. Good.

### 4. BLOCKED diagnostic surfaces

Read lines 715-730 of the hook. The new failure-mode block surfaces all 4 criteria:
- (a) what failed — "prose-mention of fields outside the trailer block"
- (b) why — "ONLY extracts from the trailer-block substring (after the LAST bare-line separator)"
- (c) how to fix — "make sure the actual structured-fields block follows a bare-line separator AND is the very last block"
- (d) historical thread — explicitly names main#509 (Wanjiku) / deploy#337 (Lucas) / deploy#339 (Bereket) with one-line failure-shape per instance

This is significantly better than the pre-fix message and gives a future operator a concrete debugging path.

### 5. Dogfooding test plan

PR body explicitly calls out the merge-sequencing concern in a "Dogfooding caveat" section: "This PR will merge through Hook 4 with the new regex. My own verdict trailer follows a bare-line separator (charter standard) so the trailer-block path activates cleanly. Reviewers MUST do the same (already standard convention) — legacy fields-only-no-separator still works via last-match-wins."

Good defensive framing. My own verdict here follows that contract: bare-line separator before the structured fields, no field-literal patterns duplicated in prose (using "the field" / "the trailer block" non-literal forms throughout this review).

### 6. CI gate alignment

All 5 checks SUCCESS on workflow CI Hooks and Scripts at head b7090d8:
- Ruff lint
- Ruff format check
- Mypy type check
- Smoke test
- Pytest hooks and lib (903 passed, +16 from 887 matches PR body)

MERGEABLE.

### 7. Test coverage

16 new tests across 3 classes:
- 5 code-strip tests (inline, fenced, unterminated, CommonMark-multiline)
- 5 trailer-block tests (no-sep, one-sep, multi-sep, embedded-not-counted, whitespace-tolerant)
- 6 regression tests (3 exact-shape repros of the 3 in-the-wild bugs + legacy-fallback + multi-trailer + canonical-baseline)

The 3 exact-shape repros (lines 1138-1198) are excellent regression coverage — each names the responsible reviewer/PR and pins the post-fix behavior.

### Summary

Well-scoped fix for a genuine release-safety regression (3 false-blocks in one batch is past the "investigate" threshold). The hybrid B+C approach plus back-compat fallback is the right shape for v1. Diagnostic message is materially better. Charter rule explains itself. Recommend a v2 strict-trailer follow-up post-adoption-window, non-blocking.

---
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Review — TPM / hook-test discipline lens

Verdict: Approved. Verified at head_sha `b7090d8a30a28482e909044d007da1c2d864ffe1` via `gh api contents?ref=…` against the main baseline.

Special interest in this one — main#509 is named after my own near-miss, and I filed #510 (now PR #513) yesterday from review of the PR this fix retro-supersedes. Mutual cross-pairing with Aino across multiple cycles, different role classes per role-class-specific boundaries memory — fine.

### What I checked

**1. Trailer-block detection correctness**

`_trailer_block_substring` at hook line 290-318. Uses splitlines(keepends=True), scans for `line.strip() == "---"`, tracks `last_sep_idx`, joins everything after. Coverage in `TrailerBlockSubstringTests`:
- no-separator returns full body (back-compat baseline)
- one-separator returns post-only
- multiple-separators picks the LAST (proven by the WRONG/CORRECT fixture at test line 1108-1115)
- embedded `--- inside sentence` doesn't count (line.strip() guards)
- surrounding whitespace `  ---  ` still recognized

Edge case I checked manually: `---` inside a fenced code block. Per the pipeline `_strip_code_regions(body)` runs FIRST, replacing the fenced region with spaces. Inside spaces, `line.strip()` of a previously-`---`-containing line becomes empty → not counted. Correct.

**2. Code-strip correctness**

`_strip_code_regions` at hook line 254-292. Space-padding preserves line indices (test_strips_inline_backticks asserts the exact length match at test line 1060). Edge cases all pinned:
- unterminated inline backtick → passes through as literal (sane forgiveness)
- unterminated triple-fence → eats rest (fail-safe; reviewer error mode)
- CommonMark-correct: inline span with embedded newline is NOT inline code

**3. Last-match-wins fallback for no-separator case**

`test_legacy_no_separator_uses_last_match_fallback` at test line 1199-1215. A prose mention of the field-syntax shape (in backticks) ABOVE the canonical trailer at end-of-body resolves to the trailer correctly because last-match-wins picks the bottom occurrence. This protects pre-#511 legacy verdicts that never used `---`.

One concern I considered: what if a verdict has prose-mention BELOW the canonical trailer (e.g., reviewer adds a post-trailer addendum)? With no separator + last-match-wins, the addendum's mention would win incorrectly. But this is rare — charter convention puts the trailer at end-of-body — and post-#511 reviewers can always add a `---` to be explicit. Acceptable trade-off; the only regression case (no separator + post-trailer prose) wasn't observed in the wild.

**4. 3 exact-repro tests — would-have-failed verification**

I spot-read all three (test lines 1138, 1158, 1179). Each constructs the exact body shape from the wild and asserts `result.reviewers == {<real reviewer lowercase>}`. Pre-#511 the same bodies would have captured prose-line garbage (or "(TBD — orchestrator will assign)") and produced false reviewers. The new helpers correctly resolve all three to the trailer-block names.

I also like the addition of `test_existing_canonical_trailer_still_works` (test line 1241) — a guard against over-zealous narrowing in the trailer-block discipline. Bare-fields-only-no-preamble bodies still parse.

**5. Dogfooding**

My verdict trailer follows a `---` separator, so the new extractor's preferred path applies — and the OLD extractor (still active until merge) sees this comment's trailer as the LAST match for each field (no prose-quoting of charter syntax above). Either way I'm safe.

**6. BLOCK message addition**

Hook line 715-730 — new diagnostic block names all three P3W11 batch-11 instances by issue number AND by reviewer-and-shape attribution. Future operators landing on the block message will have the exact incident lineage inline. Matches the consistent "name the incidents in the BLOCK reason" pattern Aino used in #508.

**7. CI green**

All 5 checks pass (Mypy, Pytest, Ruff format, Ruff lint, Smoke).

### Verification run

Temp-swapped PR-HEAD files into my parent checkout and ran pytest:

```
collected 85 items
============================== 85 passed in 0.15s ==============================
```

Files restored after run.

### Take on hybrid B+C vs pure B

Right call. Pure B alone would not catch the deploy#339 backtick-quote shape if the reviewer's backtick mention happened to land BELOW the `---` (e.g., reviewer quotes the PR-body trailer syntax in a "TechDebt narrative" line of the trailer-block prose). Pure C alone would not catch the main#509 / deploy#337 prose-above-trailer shapes because those weren't backticked. Hybrid catches both classes plus the legacy-no-separator case via the last-match-wins fallback. The strip-then-narrow ordering is also correct — stripping code regions BEFORE locating `---` lines prevents a separator inside a code fence from being mistaken for a trailer boundary.

### Non-blocking observations (no changes requested)

- The post-trailer prose edge case I noted in #3 above. If the wild ever surfaces a "reviewer added an addendum BELOW the trailer without a separator" case, we'd want to also use trailer-block-only when the separator IS present even on legacy bodies. The current implementation does so (trailer-block-only when separator present), so this is moot for any reviewer who uses `---`. Worth a memory note rather than an issue.

- Aino's docstrings cite the lesson lineage well (`Bereket-on-deploy#339`, `Wanjiku-on-main#509`, `Lucas-on-deploy#337`). Per the named-incidents-in-docstrings pattern from validate_commit_identity, this should become standard practice on hook-tier fixes.

---

Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

PR #509, opened 2026-05-19
fix(hook): validate_review_comment_format — forward --repo to internal gh pr view (closes #503)

parametrization:
## Summary

Fixes P3W11 #503 — `validate_review_comment_format` was calling `gh pr view <N>` WITHOUT `--repo`, so when a reviewer in repo-A's cwd posted a verdict on repo-B's PR with `--repo`, gh's cwd-default repo resolution silently picked the wrong PR. The Aino-rev-deploy#314 false-block was the trigger: hook fetched main's PR #314 (`A.Virtanen/0300-w7-retro-charter`) instead of deploy's #314 (`N.Hakim/0071-...`), lastname matched the reviewer (Virtanen), swap heuristic fired.

Pure lastname coincidence — other cross-repo reviewers from main cwd this session (Nino, Lucas, Aisha, Nurul) were just lucky their lastnames didn't collide with main's same-numbered PR.

Closes #503

## What changed

`.claude/hooks/validate_review_comment_format.py`

- New `extract_repo_from_command(command)` parses `--repo OWNER/NAME` from the Bash command (same shape as the parser already in `validate_pr_review.py`).
- `get_branch_name(pr_number, repo=None)` accepts and forwards `--repo` when present.
- `check()` extracts repo from the user's command and threads it through to `get_branch_name`.
- When `--repo` is absent (same-repo path), emit a stderr breadcrumb naming the hook + `--repo` + `#503` so future cross-repo invocations that omit `--repo` are discoverable in transcripts (no silent same-repo dependence on cwd discipline).

`.claude/hooks/tests/test_validate_review_comment_format.py`

- New `ExtractRepoFromCommandTests` × 3 — `present_returns_value`, `absent_returns_none`, `with_equals_form_not_supported` (pins the current parser limit).
- New `CrossRepoRegressionTests`:
  - `test_503_repro_no_false_block_when_repo_forwarded` — exact Aino-rev-deploy#314 reproduction. Captures the repo kwarg passed to `get_branch_name` and verifies the fix forwards `noorinalabs/noorinalabs-deploy`. With the fix, the branch returned is `N.Hakim/0071-...` (no lastname collision) and the verdict allows.
  - `test_same_repo_path_emits_fallback_warning_on_stderr` — captures stderr around `check()` and asserts the breadcrumb mentions the hook name, `--repo`, and `#503`.
[...]

Issue #503, opened 2026-05-18
hook(validate_review_comment_format): gh pr view called without --repo causes cross-repo false-positive swap-block


parametrization:
## Summary

\`validate_review_comment_format.py\` at line 113-114 calls \`gh pr view <pr_number> --json headRefName\` **without \`--repo\`**. gh's default repo resolution uses the cwd's git remote. If a reviewer is in a worktree whose default remote does NOT match the PR's repo, the hook fetches the WRONG branch and may false-positive-block.

## Reproduction (caught 2026-05-18, P3W11)

Aino Virtanen reviewing noorinalabs-deploy PR #314 (Nurul's #71 dead Caddy apt removal). Her cwd was noorinalabs-main. Her gh pr comment correctly used \`--repo noorinalabs/noorinalabs-deploy\`. But:

\`\`\`
gh pr view 314 --repo noorinalabs/noorinalabs-deploy → N.Hakim/0071-cloud-init-caddy-removal  (correct)
gh pr view 314 --repo noorinalabs/noorinalabs-main   → A.Virtanen/0300-w7-retro-charter      (hook's view)
\`\`\`

Hook extracted branch_author = \"Virtanen\" from main's PR #314, compared against Aino's requestor lastname \"Virtanen\", concluded swap-block.

## Why it didn't trip earlier

Pure lastname coincidence. Other reviewers in this session (Nino, Lucas, Aisha, Nurul) reviewed deploy PRs from main cwd too, but none of them coincidentally matched the lastname on whatever main PR #N happened to be.

## Fix options

**(a) Hook detects target repo from the user's gh command and passes \`--repo\` accordingly.** Parse \`--repo <owner>/<name>\` flag from the Bash command input and forward to the hook's internal \`gh pr view\` call. Adds string parsing complexity but most-correct.

**(b) Hook reads cwd's git remote and only validates when comment-target-repo matches cwd-repo.** If they don't match, skip the validation entirely (fail-allow) with a warning. Simpler but reduces enforcement scope for cross-repo review patterns.

**(c) Hook makes \`gh pr view\` call resilient to multi-repo cwd by ALSO checking \`--repo\` in the user's command and using it if present.** Hybrid of (a) and (b).

Recommendation: **(a)** — most-correct, matches the principle that the hook should validate exactly what the user is doing.

## Workaround until fixed

Reviewer cd's into the target repo's worktree before posting:
\`\`\`
[...]

PR #496, opened 2026-05-18
tech-debt(ci): extend lint/format/typecheck/smoke to .claude/lib/ + add pytest gate (#491)

parametrization:
## Summary

- Extends lint, format, typecheck, and smoke-test jobs to cover `.claude/lib/` in addition to `.claude/hooks/`. Prior to this PR, a `.claude/lib/`-only change would trigger CI (via the paths filter added in #488) but all 4 jobs would no-op against the changed files.
- Adds a new `pytest` job running `.claude/hooks/tests/` and `.claude/lib/tests/` with `ENVIRONMENT: test` set at job level. There was no mechanical CI gate for the 882-test suite; test regressions were only caught by PR-author local discipline.
- Aligns `make lint` and `make format` targets to match the extended CI scope.

## CI duration impact

Expected impact: **+30–60s** from the new `pytest` job (local runtime ~1.4s for 882 tests; CI overhead dominates). The 4 extended jobs (lint/format/typecheck/smoke) add negligible time — they parallelize and the lib surface is small (1 file + 1 test file currently).

## Local verification

- `ruff check .claude/hooks/ .claude/lib/`: clean
- `ruff format --check .claude/hooks/ .claude/lib/`: clean (66 files formatted)
- `mypy .claude/hooks/ .claude/lib/ --ignore-missing-imports --no-error-summary`: 0 errors
- `ENVIRONMENT=test pytest .claude/hooks/tests/ .claude/lib/tests/ -q`: **882 passed, 19 subtests passed** (baseline before and after changes identical)

No mypy or ruff issues found in `.claude/lib/` at time of extension — no fixes needed.

## Files modified

- `.github/workflows/ci.yml` — 4 jobs extended + 1 new pytest job
- `Makefile` — lint/format targets extended to `.claude/lib/`

Closes #491

---

**Requestor:** Aino Virtanen
**Requestee:** Aino Virtanen
**RequestOrReplied:** N/A (author PR)
**TechDebt:** none

Issue #123, opened 2026-04-17
bug: validate_pr_review.py false-positive — review-request comments trip TechDebt attestation check


parametrization:
## Problem

`.claude/hooks/validate_pr_review.py` detects PR reviews by scanning issue comments for the dual markers `Requestee:` and `RequestOrReplied:`. This matches not only actual reviews (which use `RequestOrReplied: Approved` or `RequestOrReplied: Changes Requested`) but **also review-request assignment comments** which use `RequestOrReplied: Requested`.

Result: a review-request comment that lists reviewers without a `TechDebt:` line trips the "reviews missing TechDebt attestation" block, preventing merge.

## Hit during

Merge of PR #122 (CI gate hook). The original review-request comment I posted listed both reviewers (Nadia, Santiago) with `RequestOrReplied: Requested`. The hook counted that as a third "review" missing the TechDebt line and blocked merge. Worked around by editing the comment to drop the `Requestee:` / `RequestOrReplied:` markers.

## Proposed fix

In `.claude/hooks/validate_pr_review.py:160`, tighten the detection so only `RequestOrReplied: Approved` or `RequestOrReplied: Changes Requested` count as reviews. `Requested` should explicitly be excluded — it's a request, not a review.

```python
# Current
has_request_or_replied = re.search(r"RequestOrReplied:", body)

# Proposed
has_request_or_replied = re.search(
    r"RequestOrReplied:\s*(?:Approved|Changes Requested|Replied)", body
)
```

## Acceptance criteria

- [ ] Hook correctly ignores `RequestOrReplied: Requested` comments (not reviews)
- [ ] Hook still catches `Approved` / `Changes Requested` / `Replied` as reviews
- [ ] Test: a review-request comment with both reviewers assigned but no TechDebt line does NOT block `gh pr merge`
- [ ] Test: an actual review missing TechDebt DOES block (regression guard)

## Meta

This is the 6th substring/regex-matching bug found in the hook suite during W8 (ref: #113 validate_labels cwd, #114 auto_set_env_test false-positives, #118 validate_branch_freshness cwd, ontology-tracker #110). The pattern is clear — hooks written to be liberal in matching but without a full spec of the input language. Worth a meta-issue in the W8 retro for the hook-architecture tech-debt pile.

parametrization:
Requestor: Nadia.Khoury
Requestee: Aino.Virtanen
RequestOrReplied: Request

**Wave 9 Kickoff — Phase 2**

[...]

PR #513, opened 2026-05-19
tech-debt(hook-lib): consolidate --repo parser into shared helper (closes #510)

### LINKED ISSUE #510: tech-debt(hook): validate_review_comment_format — extract_repo_from_command should accept all 4 gh flag forms (#509 follow-up)

parametrization:
## Summary

Consolidates the divergent `--repo` parsers across `validate_labels.py` (handled all 4 gh flag forms via `_shell_parse`-backed tokenize+regex) and `validate_review_comment_format.py` (#509 added a single-form `--repo\s+(\S+)` parser). Both hooks now import `extract_repo` from a shared `_repo_flag_parse.py` helper, closing the latent #503-class regression risk via `--repo=` / `-R` / `-R=` flag spellings.

I filed #510 during my review of #509 (https://github.com/noorinalabs/noorinalabs-main/pull/509#issuecomment-4483472298) — so this PR implements the exact follow-up I proposed.

## Why `.claude/hooks/_repo_flag_parse.py` (not `.claude/lib/`)?

The brief asked me to choose deliberately. `.claude/lib/` exists and has its own pytest gate (#496), but `_shell_parse.py` and `_wave_label_parse.py` — the cited precedent in the #510 issue body — both live in `.claude/hooks/`. The new helper imports `tokenize` and `walk_flag_values` from `_shell_parse`; placing it in `.claude/lib/` would require sys.path tricks to reach a hooks/-side dependency. Same-dir colocation matches the immediate precedent and avoids cross-dir import brittleness. Migrating all three shared parsers (`_shell_parse`, `_wave_label_parse`, `_repo_flag_parse`) into `.claude/lib/` is a separate larger refactor, out of scope here.

## What changed

**New:**
- `.claude/hooks/_repo_flag_parse.py` — `extract_repo(command) -> str | None` with tokenize-first + regex-fallback; recognizes `--repo X`, `--repo=X`, `-R X`, `-R=X`.
- `.claude/hooks/tests/test_repo_flag_parse.py` — 17 tests across 5 classes (FourForms / MixedWithOtherFlags / AbsentAndMalformed / RegexFallback / IssueOrigin503Repro).

**Refactored:**
- `validate_labels.py` — drops inline `extract_repo` + the now-unused `_REPO_FLAGS` constant; imports from helper. All other `_shell_parse` imports preserved (still used by `extract_labels` + `check`).
- `validate_review_comment_format.py` — drops `extract_repo_from_command` (Aino's #509 single-form parser); imports `extract_repo` from helper; renames the one call site.
- `test_validate_review_comment_format.py` — renames `ExtractRepoFromCommandTests` → `ExtractRepoCallSiteTests` (now smoke-only — the heavy parser coverage moved to `test_repo_flag_parse.py`); inverts the previously-pinning `test_with_equals_form_not_supported` → `test_equals_form_now_supported` with a docstring documenting the inversion + latent-#503 risk rationale.

## Test Plan

- [x] `python3 -m pytest .claude/hooks/tests/ .claude/lib/tests/` → **914 passed in 1.32s** (matches main baseline + 17 new from `test_repo_flag_parse.py`)
- [x] `python3 -m pytest .claude/hooks/tests/test_repo_flag_parse.py .claude/hooks/tests/test_validate_labels.py .claude/hooks/tests/test_validate_review_comment_format.py` → **98 passed**
- [x] `ruff format --check .claude/hooks/ .claude/lib/` → 68 files already formatted
- [x] `ruff check .claude/hooks/ .claude/lib/` → All checks passed
- [x] Pre-commit hooks (ruff-format, ruff-lint, mypy) all passed

## Structured fields

Requestor: (TBD)
Requestee: Wanjiku Mwangi
RequestOrReplied: New
TechDebt: none

Closes #510

### DIFF
--- a/.claude/hooks/_repo_flag_parse.py
+++ b/.claude/hooks/_repo_flag_parse.py
@@ -1,0 +1,85 @@
+#!/usr/bin/env python3
+"""Shared parser for `--repo OWNER/NAME` / `-R OWNER/NAME` extraction.
+
+Background
+==========
+
+Two PreToolUse hooks need to extract the `--repo` value from a `gh` bash
+command and forward it to internal `gh` subcalls:
+
+- `validate_labels.py` — forwards `--repo` to `gh label list` so we query
+  the same repo the user is creating the issue in.
+- `validate_review_comment_format.py` — forwards `--repo` to `gh pr view`
+  so the branch we fetch is from the SAME repo the user is commenting
+  against (#503 cross-repo skew fix, #509).
+
+Before this helper, the two hooks had divergent parsers:
+
+- `validate_labels.py:extract_repo` already handled ALL FOUR forms via
+  `(?:^|\\s)(?:--repo|-R)(?:=|\\s+)(\\S+)` with a tokenizer-first /
+  regex-fallback shape.
+- `validate_review_comment_format.py:extract_repo_from_command` (added in
+  #509) only handled the canonical `--repo <value>` space form via
+  `--repo\\s+(\\S+)`. Equals-form, `-R`, and `-R=` were silent fallthroughs
+  to cwd-default resolution — same failure direction as the #503 root
+  cause, just via alternate flag spellings.
+
+Issue #510 (filed during #509 review) asked for consolidation. This helper
+mirrors the `_shell_parse` / `_wave_label_parse` shared-helper pattern.
+
+Public API
+==========
+
+    extract_repo(command: str) -> str | None
+        Parse a bash command. Returns the `OWNER/NAME` value passed via any
+        of `--repo X`, `--repo=X`, `-R X`, `-R=X`, or None if no `--repo` /
+        `-R` flag is present.
+
+        Tokenizer-first via `_shell_parse.walk_flag_values`; falls back to
+        a conservative regex when shlex tokenization fails (e.g. command
+        contains a malformed quote). Both paths recognize all four forms.
+
+Design notes
+============
+
+- Returns the value string unchanged for direct pass-through to gh.
+- Returns None on absent flag AND on malformed flag (flag present but no
+  value); callers fail-open to cwd-default resolution + breadcrumb log.
+- `set` is used for `_REPO_FLAGS` to match `walk_flag_values`' expected
+  type (set membership lookup); the tuple-style sentinel used elsewhere
+  in the codebase would force a wrap inside `walk_flag_values`.
+"""
+
+from __future__ import annotations
+
+import os
+import re
+import sys
+from pathlib import Path
+
+sys.path.insert(0, str(Path(os.path.abspath(__file__)).parent))
+
+from _shell_parse import (  # noqa: E402
+    tokenize,
+    walk_flag_values,
+)
+
+# Flags whose VALUE is a repo specifier (OWNER/REPO).
+_REPO_FLAGS = {"--repo", "-R"}
+
+# Regex covering all 4 surface forms when tokenize() fails:
+#   --repo X    --repo=X    -R X    -R=X
+# Anchored on a token boundary at the front so `--no-repo` etc. cannot
+# false-match. `(\\S+)` is greedy-to-whitespace; matches any non-whitespace
+# value form gh accepts (gh itself validates the OWNER/NAME shape).
+_REPO_FALLBACK_RE = re.compile(r"(?:^|\s)(?:--repo|-R)(?:=|\s+)(\S+)")
+
+
+def extract_repo(command: str) -> str | None:
+    """Extract the `--repo` / `-R` `OWNER/NAME` value, or None if absent."""
+    tokens = tokenize(command)
+    if tokens is None:
+        match = _REPO_FALLBACK_RE.search(command)
+        return match.group(1) if match else None
+    values = walk_flag_values(tokens, _REPO_FLAGS)
+    return values[0] if values else None
--- a/.claude/hooks/tests/test_repo_flag_parse.py
+++ b/.claude/hooks/tests/test_repo_flag_parse.py
@@ -1,0 +1,163 @@
+#!/usr/bin/env python3
+"""Tests for `_repo_flag_parse` shared helper.
+
+Consolidates `--repo` / `-R` extraction across `validate_labels.py` and
+`validate_review_comment_format.py`. Covers all four flag forms gh CLI
+accepts (space-form / equals-form × long / short) plus the regex
+fallback used when shlex tokenization fails on malformed input.
+
+Issue #510 origin: filed during #509 review when I noticed
+`validate_review_comment_format.extract_repo_from_command` recognized
+only `--repo X` while `validate_labels.extract_repo` already recognized
+all four forms — a latent #503-class regression risk (cross-repo block
+silently misfires through `--repo=` / `-R` flag spellings).
+
+Run: python3 -m pytest .claude/hooks/tests/test_repo_flag_parse.py -v
+"""
+
+from __future__ import annotations
+
+import sys
+import unittest
+from pathlib import Path
+
+_HERE = Path(__file__).resolve().parent
+_HOOKS_DIR = _HERE.parent
+sys.path.insert(0, str(_HOOKS_DIR))
+
+import _repo_flag_parse as parser  # noqa: E402
+
+
+class FourFormsTests(unittest.TestCase):
+    """All four `gh` flag forms must parse to the same OWNER/REPO value."""
+
+    EXPECTED = "noorinalabs/noorinalabs-deploy"
+
+    def test_long_space_form(self):
+        """`--repo OWNER/NAME` — charter convention; the dominant shape."""
+        cmd = f"gh pr comment 314 --repo {self.EXPECTED} --body x"
+        self.assertEqual(parser.extract_repo(cmd), self.EXPECTED)
+
+    def test_long_equals_form(self):
+        """`--repo=OWNER/NAME` — gh accepts this; pre-#510 the
+        review-comment hook silently fell back to cwd, recreating the
+        #503 cross-repo false-block."""
+        cmd = f"gh pr comment 314 --repo={self.EXPECTED} --body x"
+        self.assertEqual(parser.extract_repo(cmd), self.EXPECTED)
+
+    def test_short_space_form(self):
+        """`-R OWNER/NAME` — gh's documented short flag."""
+        cmd = f"gh pr comment 314 -R {self.EXPECTED} --body x"
+        self.assertEqual(parser.extract_repo(cmd), self.EXPECTED)
+
+    def test_short_equals_form(self):
+        """`-R=OWNER/NAME` — short flag + equals separator."""
+        cmd = f"gh pr comment 314 -R={self.EXPECTED} --body x"
+        self.assertEqual(parser.extract_repo(cmd), self.EXPECTED)
+
+
+class MixedWithOtherFlagsTests(unittest.TestCase):
+    """`--repo` co-existing with other flags must not false-match."""
+
+    def test_after_body_flag(self):
+        cmd = 'gh pr comment 99 --body "ok" --repo owner/repo'
+        self.assertEqual(parser.extract_repo(cmd), "owner/repo")
+
+    def test_before_label_flag(self):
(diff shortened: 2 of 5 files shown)


parametrization:
Reviewed at head `bb37726ea90ef44d79d5780b3129837c8a632e4c` against base `main`. Five files, +274/-53. All five CI checks green (mypy, pytest, ruff format, ruff lint, smoke). Read all four hook files and both test files via REST contents API; did not check out the branch in any worktree.

## Program Director lens

The technical implementation is solid (helper mirrors the `_shell_parse` / `_wave_label_parse` pattern, four-form coverage with tokenizer-first / regex-fallback, 14 unit tests + 3 smoke tests on the call site). I'm approving on the org-coordination axes below — three are clean, two are observations worth a follow-up rather than a block.

### 1. Cross-repo coherence — clean

Verified `noorinalabs-deploy/.claude/hooks/` (and other child repos) contain no copies of `validate_labels.py` or `validate_review_comment_format.py`. These hooks live only in `noorinalabs-main`, so there is no parallel inline parser drift to sync. No further child-repo work needed.

### 2. Lib location & naming convention — convention is implicit but consistent; suggest making it explicit

`ls .claude/hooks/` shows the convention holds cleanly: every leading-underscore file is a helper (`_consultation_sentinel.py`, `_shell_parse.py`, `_wave_label_parse.py`, now `_repo_flag_parse.py`); every non-underscore file is a hook entrypoint. Wanjiku's placement is correct and precedent-consistent. The convention is currently undocumented though — no `.claude/hooks/CLAUDE.md` and no charter section codifies "leading-underscore = helper." Suggest a small follow-up to add a one-paragraph note to wherever hook conventions live (charter `hooks.md` or a new `.claude/hooks/CLAUDE.md`) so a future hook author has a written rule, not just an emergent pattern. Non-blocking.

### 3. Discoverability — adequate via grep, surface is small enough

A future hook author needing repo-flag parsing will likely grep `extract_repo` (turns up 5 hits across `validate_branch_freshness.py`, `validate_pr_review.py`, the new helper, and the two updated hooks) or `_repo_flag_parse` (turns up the two import sites). Helper docstring is excellent and self-explains the four-form contract + the failure-direction semantics. Adequate as-is.

### 4. Backward-compat for posted verdicts — confirmed unrelated

This PR touches `gh` command-line flag parsing only. It does not change verdict-comment field semantics, the `RequestOrReplied`/`TechDebt` parsers, or anything downstream that reads posted comments. Previously posted verdicts are unaffected. Confirmed.

### 5. Failure direction — graceful in both directions

False-negative (returns None when a flag is present but malformed): callers fall back to cwd-default resolution with a stderr breadcrumb (`validate_review_comment_format.py:255-265`). Acceptable — matches Aino's #509 shape.

False-positive (returns a wrong value): downstream `get_branch_name` calls `gh pr view --repo <wrong>` which 404s, returns None, and the hook short-circuits to `"decision": "allow"` (`validate_review_comment_format.py:266-274`). Acceptable — fails open with a warning, not a hard block on a wrong assumption.

### 6. CI green — confirmed

All five checks pass. `gh pr checks 513` clean.

### 7. Hook 4 fields — see trailer below

## Observation worth a follow-up issue (non-blocking)

While auditing for cross-repo coherence, I also grepped for other inline `--repo` parsers within `noorinalabs-main/.claude/hooks/` itself. Two adjacent hooks have parsers that should arguably also migrate to the new shared helper:

- `validate_pr_review.py:133-135` `extract_repo_from_command` is the most concerning — it uses the **identical** inline `re.search(r"--repo\s+(\S+)", command)` shape that Aino's #509 originally had (and that #510 just fixed). This is the same latent #503-class gap: equals-form, `-R`, and `-R=` silently fall back to cwd-default resolution, which could re-introduce a swap-block on `gh pr merge` invocations using alternate flag spellings. Filing a follow-up to migrate this third hook to `_repo_flag_parse.extract_repo` would close the consistency loop fully.
- `validate_branch_freshness.py:107` `extract_repo` is a thin wrapper around `_shell_parse.first_flag_value` which already handles all 4 forms — semantically equivalent to the new helper. Migration would be DRY-only, no behavior change. Lower priority.

Suggest one follow-up issue to migrate `validate_pr_review.extract_repo_from_command` to the shared helper. Not in scope for this PR. Filing this as a follow-up rather than a block since the PR as-scoped (consolidate the two hooks observed in the #509 review) does exactly what the issue asked.

## Verdict

Approve. Ship it. Recommend filing the `validate_pr_review` follow-up as a small tech-debt item before W11 closes.

---
Requestor: Nadia Khoury
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Review — PR #513 shared `--repo` parser consolidation

Read 5 head files via `gh api contents?ref=bb37726`; no parent checkout. Diffed main vs head for the two refactored hooks; independently sanity-tested the new helper against 8 critical input shapes.

Note on body discipline: avoiding literal field-name patterns in the prose below because the old greedy regex still applies (PR #512 not yet merged). Charter-format trailer is at the bottom of this comment after a separator.

### 1. Refactor preserves #509 intent — clean

Diff of `validate_review_comment_format.py` main→head is exactly three changes:

- One added import line (the shared helper).
- Removed the single-form parser I added in #509 (23 lines).
- Renamed the single call site so the new helper name is used.

The `get_branch_name(pr_number, repo=repo)` invocation is untouched. The stderr fallback breadcrumb (the same-repo path advisory pointing at #503) is preserved verbatim. No drift from #509 intent — this is exactly what the consolidation should look like.

### 2. All four flag forms parse correctly

Built the helper in `/tmp` with `_shell_parse.py` next to it and ran 8 input shapes through `extract_repo` directly:

| Input | Expected | Actual |
|---|---|---|
| `--repo owner/repo` (space, long) | `owner/repo` | ✓ |
| `--repo=owner/repo` (equals, long) | `owner/repo` | ✓ |
| `-R owner/repo` (space, short) | `owner/repo` | ✓ |
| `-R=owner/repo` (equals, short) | `owner/repo` | ✓ |
| `--body x` (absent) | `None` | ✓ |
| `--no-repo` (false-match guard) | `None` | ✓ |
| `--body "discussed --repo owner/repo in #123"` (quoted-body false-match) | `None` | ✓ |
| `--repo` (trailing flag, no value) | `None` | ✓ |

All 8 pass. The token-boundary anchor (`(?:^|\s)`) in the regex fallback correctly prevents the `--no-repo` false-match. The shlex tokenizer path keeps quoted-body content opaque so the `--repo owner/repo` inside the quoted reviewer body is invisible to the parser. Both behaviors are pinned by tests in `test_repo_flag_parse.py`.

### 3. Tokenize-first / regex-fallback design

The fallback only activates when shlex tokenization fails (malformed quoting in the source command). The `RegexFallbackTests` class explicitly forces this path by prepending `echo 'unterminated` to break shlex; the same 4 forms continue to resolve correctly. The fallback regex itself is the SAME pattern that's been in `validate_labels` since pre-#510, so this is not new code — it's verified-by-production behavior being moved into a shared helper.

### 4. Lib location decision is justified

Verified the precedent claim against main:

- `.claude/hooks/_shell_parse.py` — present
- `.claude/hooks/_wave_label_parse.py` — present
- `.claude/hooks/_consultation_sentinel.py` — present
- `.claude/lib/` exists but contains only `upsert_status_keys.py` + tests

The new helper imports `tokenize` and `walk_flag_values` from `_shell_parse` — colocating with that dependency in `.claude/hooks/` avoids cross-dir sys.path manipulation. Bundling all three underscore-prefixed parsers into `.claude/lib/` is a reasonable future refactor but explicitly out of scope here; the deferral is documented in the PR body. Decision matches the immediate precedent and is defensible.

### 5. Inversion test correctness

The pre-#510 test class `ExtractRepoFromCommandTests` (which I added in #509 with three tests including a `test_with_equals_form_not_supported` pinning the BUG behavior) has been:

- Renamed to `ExtractRepoCallSiteTests` (intent: smoke-only, parser coverage moved to the helper test file).
- Equals-form test renamed and ASSERTIONS INVERTED: pre-#510 asserted `IsNone`, post-#510 asserts equality to the value. Docstring documents the inversion + the latent-#503 risk this closes. Other two tests (present-returns-value, absent-returns-none) preserved as call-site wiring smoke checks.

The assertion-direction inversion is semantically correct, not just a rename.

### 6. No regression — test count math

Wanjiku reports 914 passed (full hook suite). My local baseline before #512 lands was 887; +17 from `test_repo_flag_parse.py` = 904 (helper tests). The +10 delta to 914 is from the inverted/renamed tests staying in the suite plus whatever post-rebase additions exist on main since #509 merged. Numbers are consistent — no test regression visible.

Post-#512 merge the suite will sit at ~920+ (16 from #512 + 17 from #513), assuming neither PR drops tests.

### 7. Hook 4 fields

Body has the bare-line tech-debt attestation literal at the end after a separator. All `#NNN` references (#510, #509, #503, #496) are pre-existing. No PR-# prediction.

### Verdict

Surgical refactor. My #509 single-form parser is correctly subsumed by the shared all-4-form helper. Latent regression risk from alternate flag spellings (`--repo=` / `-R` / `-R=`) is closed. The lib-location decision is defensible against the documented precedent. Inversion test is correct. Sanity-tested helper directly on 8 critical shapes — all pass.

---
Requestor: Aino Virtanen
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: none

Issue #528, opened 2026-05-24
epic(infra): consolidate cwd-anchor hook fixes (#525/#526/#521/#144/#227)


parametrization:
## Epic: consolidate the cwd-anchor hook fixes

Multiple hooks that read stdin `cwd` (or write paths relative to it) misbehave when invoked from a subagent/worktree context — they anchor on the orchestrator cwd instead of the actual working directory. This is one root cause with several symptoms filed separately:

| Issue | Symptom |
|---|---|
| #525 | Change-tracker records `.worktrees/` paths into the parent `ontology/checksums.json` (caused an ff-abort; hit 2× in P3W11 close-out) |
| #526 | `/session-start` Step 0 worktree cleanup only prunes the parent repo, missing child-repo worktrees (33 accumulated) |
| #521 | `validate_branch_freshness` cwd-anchor blocks child-repo PR-create from orchestrator cwd |
| #144, #227 | prior cwd-anchor instances (gh-op routing) |

## Goal
A single cwd-anchor pass across `.claude/hooks/*.py`: every hook that consumes stdin `cwd` or emits cwd-relative paths must resolve the *actual* invocation directory (and skip gitignored `.worktrees/` paths). Land as a coordinated set rather than whack-a-mole.

## Scope
- Audit all hooks for `cwd`/path-anchor usage.
- Fix the change-tracker to skip `.worktrees/` + non-parent paths (#525).
- Extend `/session-start` Step 0 to child repos with verify-merged-then-remove (#526).
- Resolve #521's branch-freshness anchor.

Tech-debt epic — P3W12 sweep candidate. Tracks #525, #526, #521, #144, #227.

parametrization:
Requestor: Fatima Okonkwo
Requestee: Wanjiku Mwangi
RequestOrReplied: Request

**Wave 12 Kickoff — Phase 3**

This issue is assigned to you for p3-wave-12.
- Peer reviewer: Aino Virtanen
- Secondary reviewer: Santiago Ferreira
- Branch from: `deployments/phase-3/wave-12`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: tech-debt (per charter § Wave Planning & Priority)

Please begin implementation.

parametrization:
[...]

Issue #525, opened 2026-05-24
tech-debt(hook): change-tracker records .worktrees/ paths into parent ontology/checksums.json


parametrization:
## Problem

The ontology Change-Tracker (`PostToolUse` hook on Edit/Write) records **worktree-relative paths** into the parent repo's `ontology/checksums.json` when an edit happens inside a git worktree (subagent or orchestrator). The hook anchors on the orchestrator cwd and writes entries like:

```
".worktrees/deploy-0348-aisha/terraform/cloudflare/variables.tf"
".worktrees/main-w11-unblock/cross-repo-status.json"
```

These are not parent-repo source files — they're (a) child-repo files seen through a worktree, or (b) the parent's own files seen via a sibling worktree path. They pollute the parent ontology checksums with junk entries that never resolve (`last_resolved: ""`).

## Impact

- Spurious dirty entries in `ontology/checksums.json`.
- **Caused a `git merge --ff-only` to abort** during W11 close-out (2026-05-24) because the uncommitted junk checksums.json change blocked the fast-forward — had to `git checkout --` it twice.
- Hit **2× in one session** (deploy#348 subagent worktree edits; the W11 blocker-lift worktree edit).

## Evidence

P3W11 close-out session 2026-05-24. Related root cause to the cwd-anchor family (main#521, #144, #227) but a distinct symptom (ontology tracker, not gh-op routing).

## Suggested fix

Change-tracker should **ignore `.worktrees/` and child-repo paths** (now that `.worktrees/` is gitignored as of #523, the tracker can skip gitignored paths, or explicitly skip any path containing `/.worktrees/` or `.worktrees/`).

Tech-debt — W12 sweep.

parametrization:
Requestor: Fatima Okonkwo
Requestee: Wanjiku Mwangi
RequestOrReplied: Request

**Wave 12 Kickoff — Phase 3**

This issue is assigned to you for p3-wave-12.
- Peer reviewer: Aino Virtanen
- Secondary reviewer: Santiago Ferreira
- Branch from: `deployments/phase-3/wave-12`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: tech-debt (per charter § Wave Planning & Priority)

Please begin implementation.

parametrization:
[...]

PR #531, opened 2026-05-25
fix(session-start): Step 0 prunes child-repo worktrees with verify-merged guard (#526)

### LINKED ISSUE #526: tech-debt(session-start): Step 0 worktree cleanup misses child-repo worktrees
## Problem

`/session-start` Step 0 (worktree cleanup) runs `git worktree prune` + `git worktree list` on the **parent repo only**. Child-repo worktrees under `<child>/.claude/worktrees/` (and elsewhere) are never pruned, so they **accumulate indefinitely** across sessions.

## Evidence

At the start of the P3W11 close-out session (2026-05-24), a cleanup found **~33 stale child-repo worktrees** accumulated across prior sessions (isnad-graph ~19, user-service 5 incl. a locked `/tmp/hotfix-user-service`, data-acquisition 4, design-system 5) — none caught by `/session-start`. 28 were verified-merged-and-removed, 5 were superseded (closed issues), all cleaned manually.

## Suggested fix

Extend `/session-start` Step 0 to iterate child repos and prune/list their worktrees too, with a verify-merged-then-remove guard (only auto-remove worktrees whose HEAD is an ancestor of `origin/main` **or** has a merged PR; flag the rest). Mirrors the manual procedure used on 2026-05-24.

Tech-debt / process — W12 sweep.

parametrization:
## Summary

`/session-start` Step 0 only pruned the **parent** repo's worktrees. Child-repo worktrees (under `<child>/.claude/worktrees/`, `<child>/.worktrees/`, and `/tmp/`) accumulated indefinitely — ~33 stale ones were found manually on 2026-05-24 (isnad-graph ~19, user-service 5 incl. a locked `/tmp/hotfix-user-service`, data-acquisition 4, design-system 5), none caught by the skill.

## Change

Step 0 now iterates the parent **plus the 7 canonical child repos** (CLAUDE.md Repository Map) using `git worktree list --porcelain`, with a **verify-merged-then-remove guard** mirroring the manual 2026-05-24 procedure:

- **Auto-remove** a worktree only when its HEAD is an ancestor of that repo's `origin/main` (fully merged). 28 were auto-removable on 2026-05-24.
- **FLAG** (list, never auto-remove) any **unmerged** or **locked** worktree — e.g. the `/tmp/hotfix-user-service` lock case — for a manual decision. 5 were superseded/closed-issue cases.

The fix lives in `.claude/skills/session-start/SKILL.md` (the bash the orchestrator runs). `.claude/hooks/session_start.py` is text-directive only (it does not execute cleanup), so its Step 0 prose is updated to point at the SKILL.md block and describe the guard — keeping hook and skill in sync.

## Validation

- Step 0 bash block: `bash -n` syntax OK + `shellcheck -s bash` clean.
- Dry-run from the real main checkout: 8 repos detected; in-flight worktrees correctly FLAGGED (unmerged), parent main checkout skipped, merged ones marked for removal.
- `ruff format --check` + `ruff check` clean on `.claude/hooks/`; pre-commit ruff/mypy passed.

## Tech debt

REMOVES process-debt: eliminates the recurring manual child-repo worktree sweep that was being done by hand.

Closes #526
Advances #528 (epic closes after the #525/#521 cwd-anchor PR also lands)

### DIFF
--- a/.claude/hooks/session_start.py
+++ b/.claude/hooks/session_start.py
@@ -90,10 +90,12 @@ def main() -> None:
         "",
     ]
 
-    # Step 0: Worktree cleanup
-    lines.append("STEP 0 — WORKTREE CLEANUP:")
-    lines.append("  Run: git worktree prune && git worktree list")
-    lines.append("  Remove any stale worktrees under .claude/worktrees/")
+    # Step 0: Worktree cleanup (parent + child repos, per #526)
+    lines.append("STEP 0 — WORKTREE CLEANUP (parent + child repos):")
+    lines.append("  Run the Step 0 block in skills/session-start/SKILL.md.")
+    lines.append("  It prunes the parent AND all 7 child repos, auto-removing")
+    lines.append("  only worktrees merged into origin/main and FLAGGING locked")
+    lines.append("  or unmerged ones for a manual decision (never auto-removed).")
     lines.append("")
 
     # Step 1: Team cleanup (unique to noorinalabs-main)
--- a/.claude/skills/session-start/SKILL.md
+++ b/.claude/skills/session-start/SKILL.md
@@ -13,15 +13,82 @@ description: "MANDATORY first action in every session — runs full startup prot
 
 Execute all 7 steps below. Steps that are independent of each other SHOULD run in parallel. Present results in a single concise status table at the end.
 
-### Step 0 — Worktree cleanup
+### Step 0 — Worktree cleanup (parent + child repos)
+
+Worktrees accumulate in BOTH the parent repo and every child repo (under
+`<child>/.claude/worktrees/`, `<child>/.worktrees/`, and sometimes `/tmp/`).
+Prior to #526, Step 0 only cleaned the parent — on 2026-05-24 ~33 stale
+child-repo worktrees were found uncaught. The block below iterates the parent
+and all 7 child repos, applying a **verify-merged-then-remove guard**:
+
+- **Auto-remove** a worktree only when its HEAD is an ancestor of that repo's
+  `origin/main` (i.e. the branch is fully merged). Safe to drop.
+- **FLAG (list, do not remove)** any worktree that is NOT verified-merged
+  (work in flight, superseded, or closed-issue cases) and any **locked**
+  worktree (e.g. the `/tmp/hotfix-user-service` lock case). Surface these for
+  a manual decision — never auto-remove unmerged work.
 
 ```bash
 REPO_ROOT="$(git rev-parse --show-toplevel)"
-git -C "$REPO_ROOT" worktree prune
-git -C "$REPO_ROOT" worktree list
+
+# Parent repo + the 7 canonical child repos (CLAUDE.md Repository Map).
+REPOS=("$REPO_ROOT")
+for child in \
+  noorinalabs-isnad-graph \
+  noorinalabs-user-service \
+  noorinalabs-deploy \
+  noorinalabs-design-system \
+  noorinalabs-data-acquisition \
+  noorinalabs-isnad-ingest-platform \
+  noorinalabs-landing-page; do
+  [ -d "$REPO_ROOT/$child/.git" ] && REPOS+=("$REPO_ROOT/$child")
+done
+
+FLAGGED=()
+for repo in "${REPOS[@]}"; do
+  git -C "$repo" worktree prune
+  # Refresh remote tip so the merged-ancestor test is accurate.
+  git -C "$repo" fetch --quiet origin main 2>/dev/null || true
+  main_repo="$(git -C "$repo" rev-parse --show-toplevel)"
+
+  # Walk worktrees in porcelain form. Records are blank-line separated;
+  # fields we care about: worktree <path>, HEAD <sha>, locked [<reason>].
+  wt="" head="" locked=0
+  while IFS= read -r line; do
+    case "$line" in
+      "worktree "*) wt="${line#worktree }"; head=""; locked=0 ;;
+      "HEAD "*)     head="${line#HEAD }" ;;
+      "locked"*)    locked=1 ;;
+      "")  # end of a record — evaluate it
+        [ -z "$wt" ] && continue
+        if [ "$wt" = "$main_repo" ]; then wt=""; continue; fi  # skip main checkout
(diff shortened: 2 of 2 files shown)


parametrization:
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Secondary review (release-coordination / operational-safety lens). Read at head SHA `0e597aa419078704d6600d8dcd6543c06cdf19b5` via `gh api contents` — no branch-switch.

**Verify-merged-then-remove guard — the load-bearing safety property, traced and sound.** Auto-remove fires only inside the `elif ... merge-base --is-ancestor "$head" origin/main` true-branch. Every other path FLAGs and never removes:
- Locked worktrees (the `/tmp/hotfix-user-service` case) hit the first `if [ "$locked" -eq 1 ]` and are always flagged, regardless of merge state.
- Empty HEAD is guarded by `[ -n "$head" ]`, falling through to the UNMERGED flag.
- `--is-ancestor` returning 1 (not-ancestor) or 128 (bad object / error) is non-zero; the `&&` short-circuits to the `else` UNMERGED flag. The `2>/dev/null` only hides stderr, not the exit code. Fails closed.
- A `worktree remove` failure (e.g. the main checkout, or a dirty worktree) is caught by the `|| ... REMOVE-FAILED` flag rather than silently dropped.

**Stale-origin reasoning checks out.** `fetch --quiet origin main || true` then testing against `origin/main` is the right ordering. If the fetch fails offline, `origin/main` can only be *behind* the true remote — so a merged worktree may be conservatively FLAGGED, but an unmerged one can never be mis-classified as merged (a stale ref lacks commits, it never gains them). Over-conservative is the correct failure direction here.

**Child-repo iteration is correct and graceful.** Covers the 7 canonical repos from the CLAUDE.md map plus the parent, gated on `[ -d "$REPO_ROOT/$child/.git" ]` so a missing/un-cloned child is skipped silently. A child with zero worktrees yields only its main-checkout record, which the `$wt = $main_repo` skip drops — no spurious flags.

**Porcelain parsing verified by simulation.** Records are blank-line separated; I ran the parser against detached-HEAD, locked, and branch records plus the trailing `echo` that the block appends to `worktree list --porcelain` (porcelain omits the final blank line, so the appended `echo` is required to flush the last record — good catch in the implementation). All records evaluated exactly once.

**Bash safety / lint.** Re-ran independently at head: `bash -n` syntax OK; `shellcheck -s bash` exit 0 (installed shellcheck 0.10.0 locally first so the integration actually runs, not a silent skip). Expansions inside the loop are quoted; no mis-glob surface.

**Hook/skill sync — no drift.** `session_start.py` Step 0 is `lines.append(...)` text only; grep confirms no `subprocess`/`os.system`/`worktree remove|prune` for cleanup in the hook. The directive prose accurately points at the SKILL.md block and describes the merged-only / flag-the-rest guard, matching the procedure.

**On the REPO_ROOT-from-worktree gap (my requested call).** Out-of-scope is acceptable for #526 and does not block. `/session-start` is by contract the session's first action and runs from the parent checkout, so `rev-parse --show-toplevel` resolves to the parent in the real invocation path — the gap is not a live regression. Worth a small tracked follow-up (anchor `REPO_ROOT` to the parent via the git-common-dir or a known marker rather than `--show-toplevel`) so a future stray invocation-from-worktree degrades safely, but that is hardening, not a fix for this PR. Recommend filing it against epic #528 alongside the cwd-anchor work (#525/#521), since it is the same cwd-resolution family.

Approving. Safety guard is correct, fails in the safe direction on every branch I traced, lint is genuinely clean, and hook/skill stay in sync.

parametrization:
## Primary review — PR #531 (session-start Step 0 prunes child-repo worktrees, #526)

Reviewed at head SHA `0e597aa419078704d6600d8dcd6543c06cdf19b5` via `gh api contents` (no branch-switch).

This review approves the change.

**Safety-guard analysis (the load-bearing lens).** I traced every branch of the per-worktree evaluator:

- Locked worktree -> FLAGGED LOCKED, never removed. Correct.
- The ONLY removal path is gated on `git merge-base --is-ancestor "$head" origin/main`. A worktree is removed only when its HEAD is a verified ancestor of that repo's origin/main (fully merged).
- Anything else — not-merged, empty HEAD, or even an errored ancestor test — falls through to FLAGGED UNMERGED. Unmerged work is never removed.
- The main checkout is skipped via the `$wt = $main_repo` guard.

Critically, the degenerate cases fail in the SAFE direction: if `fetch` fails or origin/main is missing, `merge-base --is-ancestor ... origin/main` errors, `2>/dev/null` swallows it, the non-zero return drops to the else branch, and the worktree is FLAGGED rather than removed. Removal failure itself is caught (graceful -> `--force` -> FLAGGED REMOVE-FAILED). There is no path where unmerged or locked work is auto-removed. The guard satisfies the acceptance lens.

I also confirmed the porcelain parse against real `git worktree list --porcelain` output: blank-line-separated records, `worktree <path>` / `HEAD <sha>` / `locked` fields match the `case` arms, and the trailing `; echo` correctly guarantees the final record's terminating blank line is processed.

**Two non-blocking observations (do NOT hold the merge):**

1. `git fetch --quiet origin main` with an explicit refspec updates FETCH_HEAD but does not reliably update the local `refs/remotes/origin/main` ref. If origin/main is stale-behind, a genuinely-merged worktree tests as not-ancestor and gets FLAGGED-UNMERGED instead of auto-removed. This is a completeness false-negative (under-cleans), and it fails SAFE — it never removes unmerged work — so it is not a blocker. A follow-up could use `git -C "$repo" fetch --quiet origin main:refs/remotes/origin/main` or test against `FETCH_HEAD` to make the auto-remove tier fire more often. Worth a tracking note, not a change request.

2. The prose enumerates worktree locations as `<child>/.claude/worktrees/`, `<child>/.worktrees/`, `/tmp/`, but the script is location-agnostic (it processes whatever `worktree list` reports, regardless of path). That is the more robust design; the prose list is illustrative only. No code change needed.

Neither observation affects the safety guard, which is the acceptance gate. Approving.

(Secondary already approved by Santiago Ferreira, who confirmed the guard.)

Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

PR #523, opened 2026-05-24
chore(coord): recover W11 wrapup status + gitignore worktrees/.bak

parametrization:
## Summary

Recovers stranded P3W11 wrapup state and adds missing gitignore entries, after reconciling a diverged local `main`.

**Context:** local `main` had drifted 2-ahead / 35-behind `origin/main`. The 2 local commits were stale ontology/gitignore housekeeping (superseded by newer origin rebuilds); they were discarded by resetting `main` to `origin/main` (which carries the #522 wave-11 merge). One genuinely valuable artifact was stranded as an *uncommitted* edit on that diverged main — the W11 wrapup status block — and is recovered here.

## Changes

**`cross-repo-status.json`** — additive W11 wrapup keys (generated by `/wave-wrapup` on 2026-05-20, never committed):
- `wave_11_final_pr_count: 86`, `wave_11_changes_requested_cycles: 16`, `wave_11_top_concentration_pct: 15`
- `wave_11_wrapup_at: 2026-05-20T01:36:10Z`
- `wave_11_carry_forward`: 12 issues (11 deploy + main#136)
- `wave_11_close_blocked_by: deploy#348` — .net/.org canonical-redirect entrypoint import must apply to prod before W11 closes (owner directive 2026-05-21)
- `wave_11_active` intentionally left `true` — W11 is open-pending-close, not closed.

**`.gitignore`** — `.worktrees/`, `.claude/worktrees/` (local-only isolation dirs), and `.claude/annunaki/errors.jsonl.bak.*` (rotation backups; pattern was previously tracked only on the discarded local branch).

**`ontology/checksums.json`** — checksum-only resolution of the two files above (no entity/service changes).

## Test plan
- [x] `cross-repo-status.json` validates as JSON; 6 wrapup keys present
- [x] `git check-ignore` confirms `.worktrees/`, `.claude/worktrees/`, `errors.jsonl.bak.*` are ignored
- [x] ontology dirty count = 0 after resolution

Refs #447

PR #534, opened 2026-05-25
fix(hooks): coordinated cwd-anchor pass — tracker .worktrees skip + gh-pr-create cwd recovery (closes #525 #521)

### LINKED ISSUE #521: hooks: validate_branch_freshness cwd-anchor — child-repo PR-create blocked from orchestrator cwd
## Source

Surfaced during deploy#242 step 2 implementation (PR #342). Aisha Idrissi's first `gh pr create` invocation was BLOCKED by `validate_branch_freshness` hook because `cwd` in the hook input was the orchestrator's parent directory (`/home/parameterization/code/noorinalabs-main`), not her worktree (`/home/parameterization/code/noorinalabs-main/.worktrees/deploy-0242-step2-promote`). `_resolve_implicit_repo` consequently returned `noorinalabs/noorinalabs-main` instead of `noorinalabs/noorinalabs-deploy`.

## Workaround applied

Passing `--repo noorinalabs/noorinalabs-deploy --head A.Idrissi/0242-promote-drop-single-arch-branch` explicitly to `gh pr create` routed through the `if repo:` branch in the hook and bypassed the cwd-based resolution.

## Root cause

The hook's `_resolve_implicit_repo` function reads `cwd` from hook stdin payload, which in subagent sessions is the orchestrator's working directory at spawn time, NOT the subagent's actual working directory after `cd` into a worktree.

## Prior art

Same shape as the cwd-anchor issues addressed in #144 (specific hook) and #227 (broader cwd-anchor pattern). The fix wasn't applied to `validate_branch_freshness` — likely missed in the original cwd-anchor sweep.

## Proposed fix

Apply the same cwd-anchor pattern: prefer the calling subprocess's actual working directory (`os.getcwd()` or pass via env from settings.json invocation) over the hook-input `cwd` field, OR resolve repo by walking up from the actual cwd to find the nearest `.git/config` and reading `[remote "origin"]` URL.

Affected file: `.claude/hooks/validate_branch_freshness.py` (path likely; verify location).

## Acceptance Criteria

- [ ] Hook resolves the correct child repo when invoked from any child-repo worktree
- [ ] PR-create flow no longer requires `--repo` workaround for child-repo PRs
- [ ] Regression test added: simulated hook invocation with parent-dir `cwd` payload + child-repo `.git` content correctly routes
- [ ] Same fix considered for other hooks in same class (cross-check against #144 / #227 fix surface)

---
Surfaced by Aisha Idrissi during PR #342 (deploy#242 step 2) on 2026-05-19.

### LINKED ISSUE #525: tech-debt(hook): change-tracker records .worktrees/ paths into parent ontology/checksums.json

parametrization:
## Summary

A single coordinated cwd-anchor pass across `.claude/hooks/*.py`, landing the two residual cwd-anchor symptoms as one set rather than whack-a-mole (the explicit epic #528 mandate).

## Part A — #525: change-tracker records `.worktrees/` paths into parent checksums

`ontology_tracker` skipped `.claude/worktrees/` but not the top-level `.worktrees/` convention (gitignored as of #523). Worktree-relative paths like `.worktrees/deploy-0348-aisha/...` leaked into the parent `ontology/checksums.json`, never resolved (`last_resolved: ""`), and once aborted a `git merge --ff-only` during W11 close-out (hit 2×).

**Fix:** new path-COMPONENT-aware `_is_worktree_path` skip catching both `.worktrees` and `.claude/worktrees` — segment-matched, not substring, so a legit file like `notes.worktrees.md` still tracks.

## Part B — #521: validate_branch_freshness false-blocks child-repo PR-create

`#144`/`#227` (PR #248) already anchored the hook's subprocesses on `cwd`. The residual was the **source** of that cwd: the harness stdin `cwd` is the orchestrator's spawn-time dir, not the subagent's worktree after `cd`. So `_resolve_implicit_repo` resolved `noorinalabs-main` and blocked child-repo `gh pr create` on the parent's branch (workaround was explicit `--repo`).

**Fix:** shared `extract_leading_cd_target` + `resolve_invocation_cwd` in `_shell_parse`, recovering the real cwd from a leading `cd /worktree && gh pr create`, plus a `GH_REPO` env fast-path (gh's own canonical implicit-repo override). Both consulted before the (poisoned) stdin cwd.

## Part C — cross-hook audit (the "single pass")

`grep -rn "cwd" .claude/hooks/*.py` enumerated every cwd-consuming hook. Dispositions:

| Hook | Verdict |
|------|---------|
| `validate_branch_freshness` | **FIXED** (#521) |
| `validate_workflow_paths_coverage` | **FIXED** — `_resolve_repo`/`_resolve_head` ran git with no `cwd=` at all (parent-process repo/branch); same gh-pr-create trigger. Threaded `resolve_invocation_cwd` + `GH_REPO`. |
| `ontology_tracker` | **FIXED** (#525) |
| `no_worktree_self_delete` | **Audited, left as-is** (comment added). A leading `cd` here changes SAFETY semantics, not just repo identity (`cd /safe && remove /wt` is safe; `cd /wt && remove /wt` is self-delete) — adopting recovery would alter block/allow verdicts. Distinct decision, out of scope for the cwd-anchor sweep. |
| `validate_commit_identity` | **Ruled out** — already recovers cwd via its own `cd`-prefix extractor (#475); functionally correct. |
| `session_handoff` | **Ruled out** — parent-repo-scoped by design (defaults to `REPO_ROOT`). |
| `enforce_librarian_consulted` / `_consultation_sentinel` | **Ruled out** — cwd-keyed sentinel written by the skill in the same cwd the hook reads; consistent keying (#169/#176). |
| `validate_labels` / `validate_review_comment_format` / `validate_pr_review` / `_repo_flag_parse` | **Ruled out** — `--repo`-flag-based with fail-open to gh's cwd-default; no `git remote get-url` repo-identity walk to misanchor. |

## Tests (acceptance)

- `_shell_parse`: `ExtractLeadingCdTargetTests`, `ResolveInvocationCwdTests`
- `validate_branch_freshness`: `Issue521CwdRecoveryTests` (incl. `test_repro_for_issue_521` — parent stdin cwd + child-repo cd-target → routes to child)
- `validate_workflow_paths_coverage`: `CwdAnchorResolutionTests`
- `ontology_tracker`: `ShouldSkipTopLevelWorktreesTests` (incl. exact #525 evidence shapes + segment-vs-substring guard)

**Full suite: 1111 passed.** `ruff check` / `ruff format --check` clean; `mypy` clean.

## Tech debt

This PR **removes** tech-debt — the cwd-anchor residuals in the `gh pr create` gate hooks and the worktree-path pollution in the change-tracker.

Closes #525
Closes #521
Advances #528

### DIFF
--- a/.claude/hooks/_shell_parse.py
+++ b/.claude/hooks/_shell_parse.py
@@ -92,6 +92,21 @@ Public API
         want to operate on the *user's* cwd (not the hook's parent process
         cwd) should use this to anchor `subprocess.run(..., cwd=...)`.
 
+    resolve_invocation_cwd(input_data) -> str
+        Like resolve_tool_cwd, but FIRST tries to recover the directory the
+        command actually runs in by extracting a leading `cd <dir>` segment
+        from the Bash command string. This closes the #521 residual: for a
+        worktree subagent the harness `cwd` field is captured at agent-spawn
+        time (the orchestrator's dir), NOT the subagent's dir after it has
+        `cd`'d into its worktree, and subsequent `cd` calls do not propagate
+        back to the hook's view of `cwd`. When the triggering command is
+        `cd /path/to/worktree && gh pr create ...`, the cd target is the
+        only in-band signal that recovers the real repo. Falls back to
+        resolve_tool_cwd (stdin cwd → os.getcwd()) when no leading `cd`
+        is present. Only absolute existing directories are honored; relative
+        cd targets are ambiguous (they'd be relative to the already-wrong
+        stdin cwd) so they are ignored.
+
     is_shutdown_request_message(message) -> bool
         True only if `message` is a structured shutdown_request JSON
         (dict-form OR str-form parseable to a dict with type==
@@ -443,6 +458,57 @@ def resolve_tool_cwd(input_data: dict) -> str:
     if cwd and isinstance(cwd, str):
         return cwd
     return os.getcwd()
+
+
+def extract_leading_cd_target(command: str) -> str | None:
+    """Return the directory of the last `cd <dir>` that precedes other work.
+
+    Walks the command's pipeline segments (see iter_command_segments) and
+    records the target of every `cd <dir>` segment, returning the last one.
+    Only honors a single-argument absolute `cd` target — relative targets
+    are ambiguous because they'd resolve against the (wrong) stdin cwd, and
+    multi-arg `cd` (e.g. `cd -P x`) is rare enough to skip rather than
+    mis-parse.
+
+    Returns None when the command does not tokenize, has no `cd`, or the
+    cd target is not an absolute path. The caller is responsible for
+    checking the path actually exists.
+
+    This is the in-band recovery signal for the worktree-subagent cwd-anchor
+    bug (#521): `cd /worktree && gh pr create` carries the real cwd in the
+    command itself even though the harness `cwd` field points at the
+    orchestrator's spawn-time directory.
+    """
+    tokens = tokenize(command)
+    if tokens is None:
+        return None
+    target: str | None = None
+    for segment in iter_command_segments(tokens):
+        if len(segment) == 2 and segment[0] == "cd" and segment[1].startswith("/"):
+            target = segment[1]
+    return target
+
+
+def resolve_invocation_cwd(input_data: dict) -> str:
+    """Resolve the directory the triggering command actually runs in.
+
+    Priority:
+      1. An absolute, existing `cd <dir>` target extracted from the command
+         string (recovers a worktree subagent's real cwd — #521).
+      2. resolve_tool_cwd(input_data) — stdin `cwd` then os.getcwd().
+
+    Use this (rather than resolve_tool_cwd) for any hook that derives repo
+    IDENTITY from cwd — i.e. anything that runs `git remote get-url origin`
+    or `git rev-parse` to decide which GitHub repo a command targets. The
+    plain resolve_tool_cwd is fine for hooks that only need *a* git context
+    and don't care about cross-repo misattribution.
+    """
+    command = input_data.get("tool_input", {}).get("command", "")
+    if isinstance(command, str) and command:
+        cd_target = extract_leading_cd_target(command)
+        if cd_target and os.path.isdir(cd_target):
+            return cd_target
+    return resolve_tool_cwd(input_data)
 
 
 def is_shutdown_request_message(message) -> bool:
--- a/.claude/hooks/no_worktree_self_delete.py
+++ b/.claude/hooks/no_worktree_self_delete.py
@@ -265,6 +265,16 @@ def check(input_data: dict) -> dict | None:
 
         # Caller-reported cwd wins; fall back to ours. Hook input carries the
         # shell's actual cwd at tool-call time.
+        #
+        # NOTE: this hook deliberately does NOT adopt the #521
+        # `resolve_invocation_cwd` cd-target recovery used by the gh-pr-create
+        # hooks. Here a leading `cd` changes the SAFETY semantics rather than
+        # just repo identity: `cd /safe && git worktree remove /worktree` is
+        # safe (shell moves out first) while `cd /worktree && remove /worktree`
+        # is the self-delete. Folding the cd target into cwd would change the
+        # block/allow verdict, which is a distinct behavioral decision from the
+        # cwd-anchor sweep — see the docstring "cd plans not yet executed"
+        # note. Left as-is intentionally; tracked separately if revisited.
         cwd = input_data.get("cwd") or os.getcwd()
 
         for segment in _SEGMENT_SPLIT_RE.split(command):
--- a/.claude/hooks/ontology_tracker.py
+++ b/.claude/hooks/ontology_tracker.py
@@ -17,16 +17,24 @@ Path filtering (issue #143):
     * Substring SKIP_PATTERNS (e.g. ``__pycache__/``, ``.git/``).
     * Paths beginning with ``/tmp/`` — ephemeral scratch (e.g. issue-body
       staging files).
-    * Paths containing ``.claude/worktrees/`` — ephemeral worktree copies.
-      The canonical entry for the underlying file is updated whenever the
-      file is next Edit/Written directly on the main checkout (this is a
-      PostToolUse hook on tool calls, NOT a git post-merge hook, so a
-      squash-merge of a worktree-only PR does not by itself update the
-      tracker — the next direct Edit on main does). Skipping worktree
-      paths is still the right call: it prevents accumulation of stale
-      paths in ``checksums.json`` after worktrees are removed, and the
-      slight latency in the canonical entry's ``last_tracked`` is
-      acceptable noise-vs-signal trade.
+    * Paths under any ``.worktrees`` directory — ephemeral worktree copies.
+      Two conventions are both gitignored (#523) and both skipped: the
+      historical ``.claude/worktrees/`` path and the top-level
+      ``.worktrees/`` path used by current wave/agent isolation. A worktree
+      path enters this hook when an Edit/Write happens inside a worktree and
+      the hook (anchored on the orchestrator cwd) records the worktree-
+      relative path — e.g. ``.worktrees/deploy-0348-aisha/...`` or a
+      child-repo file seen through a sibling worktree. These never resolve
+      (``last_resolved: ""``) and once aborted a ``git merge --ff-only``
+      during W11 close-out (#525). The canonical entry for the underlying
+      file is updated whenever the file is next Edit/Written directly on the
+      main checkout (this is a PostToolUse hook on tool calls, NOT a git
+      post-merge hook, so a squash-merge of a worktree-only PR does not by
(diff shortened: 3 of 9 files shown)


parametrization:
## Review — PR #534 (cwd-anchor pass, closes #525 #521)

Reviewed all 9 files at head 91627796f9ed40647d486b4de52466cfc81a8f17 via `gh api contents?ref=` (no branch-switch). CI green (5/5), mergeStateStatus CLEAN.

### #525 — ontology_tracker `_is_worktree_path` (segment matching) — correct
`_is_worktree_path` walks `Path(file_path).parts` and matches a component equal to `.worktrees` at any depth, and a component equal to `worktrees` only when its parent component is `.claude`. This is true path-COMPONENT matching, not substring. Verified the false-positive guard: `notes.worktrees.md` is a single part equal to neither marker, so it is not skipped (pinned by `test_worktrees_segment_not_substring_false_match`). The historical `.claude/worktrees/` and the top-level `.worktrees/` are both caught (pinned). Docstring matches implementation.

### #521 — validate_branch_freshness `resolve_invocation_cwd` + GH_REPO fast-path — correct
The exact #521 repro is covered: `test_repro_for_issue_521` feeds stdin cwd at the parent (`noorinalabs-main`) with command `cd <worktree> && gh pr create`, and asserts the implicit-repo resolver runs against the recovered worktree cwd → child repo, not the parent — `local.assert_not_called()`, remote called with the child slug. `test_gh_repo_env_takes_priority` pins GH_REPO short-circuiting cwd resolution entirely. `extract_leading_cd_target` correctly ignores relative cd, multi-arg cd, and cd-inside-a-quoted-body, and takes last-absolute-cd-wins. The same cwd-anchor is threaded into `validate_workflow_paths_coverage._resolve_repo`/`_resolve_head` (Part C 2nd-hook fix), pinned by `CwdAnchorResolutionTests`.

### Shared helper `_shell_parse` — sound
`tokenize` (with #287 line-continuation normalization), `iter_command_segments`, `extract_leading_cd_target`, `resolve_invocation_cwd` handle the real invocation shapes (cd && cmd, quoted paths, env-var prefixes). The tokenize None-contract is documented and callers fail-open for warn-only matchers / fail-closed where noted.

### Part C audit honesty — the `no_worktree_self_delete` deferral
I validated the judgment call. The decision to leave the hook un-changed is the conservative-correct call and is honestly documented (in-code comment + PR-body table). It is genuinely a different concern (safety ancestry #173, not repo identity), so it is scope-appropriate to defer.

One non-blocking precision point worth a tracked follow-up (NOT a merge blocker): the residual gap is real, not merely "out of scope." I exercised the PR's actual `check()` and confirmed that for a worktree subagent the shape `cd <wt> && git worktree remove <wt>` with the harness stdin cwd at the parent currently ALLOWS — yet that IS a self-delete that dangles the shell (the #173 footgun). The PR-body framing says adopting recovery "would alter block/allow verdicts," which reads as "recovery would make it worse," when in fact a careful last-cd-wins recovery would make the self-delete case start BLOCKING (an improvement) while `cd /safe && remove /wt` would still ALLOW. The reason to defer is that doing it correctly needs its own safety-focused test matrix, not that recovery is inherently wrong here. Recommend the team-lead open a follow-up issue (advances #528 / sibling of #173) so the gap is tracked rather than only narrated in a merged PR body. This does not gate the merge.

Tests: 1111 reported passing; the file-level suites I read (test_shell_parse, test_ontology_tracker, test_validate_branch_freshness, test_validate_workflow_paths_coverage) have direct repro + negative-match coverage for every claim above. Approving.

Requestor: Aino Virtanen
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: none

parametrization:
Requestor: Santiago Ferreira
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: none

Secondary review (release-coordination / cross-repo-routing lens — #521 is squarely my domain). Read at head SHA `91627796f9ed40647d486b4de52466cfc81a8f17` via `gh api contents` — no branch-switch. I fetched all changed source + their deps (`_shell_parse`, `annunaki_log`, `_repo_flag_parse`) and laid them out mirroring `.claude/hooks/` so `REPO_ROOT` math resolves, then ran the suite independently rather than trusting the PR's pass claim.

**Independently reproduced 208/208 passing** across the four changed test files (`ENVIRONMENT=test pytest`), including `test_repro_for_issue_521`, the GH_REPO priority/short-circuit tests, `CwdAnchorResolutionTests` for the workflow-paths hook, the `#525` `.worktrees`/`.claude/worktrees` segment-vs-substring tests, and the three fail-open regression tests.

**Lens 1 — child-repo PR-create routing (the Aisha deploy#242/PR#342 failure).** Traced and confirmed correct on both gh-pr-create gate hooks. `resolve_invocation_cwd` recovers the worktree dir from a leading `cd /worktree && gh pr create`, validated as an absolute existing dir, and both `validate_branch_freshness` and `validate_workflow_paths_coverage` thread it into their `git remote get-url origin` / `git rev-parse` repo-identity calls. Resolution order GH_REPO env then cd-recovered-cwd origin is right — GH_REPO is gh's own canonical implicit-repo override, so honoring it first matches gh's own behavior. The end-to-end repro test pins that a parent stdin cwd + child cd-target routes to the child, not the parent. The second hook was a genuine catch — its `_resolve_repo`/`_resolve_head` previously ran git with no `cwd=` at all (parent-process identity), same trigger class.

**Lens 2 — fail direction (most important for me).** Confirmed SAFE in every resolution-failure path I traced. In `validate_branch_freshness`: if both GH_REPO and `_resolve_implicit_repo(cwd)` yield nothing, it falls to `is_branch_fresh_local`, which returns True (allow) on subprocess error/timeout; `is_branch_fresh_remote` returns None on parse failure and None is treated as allow. In `validate_workflow_paths_coverage`: `if not repo or not head` fails open explicitly, and the API-failure tests pin it. So a cwd-recovery miss degrades to allow-the-PR-create (recoverable: CI/merge-conflict catches a stale branch), never to a new false-block. The fix strictly REDUCES the prior false-block surface and introduces no new block path — the correct direction for these gates.

**Lens 3 — `no_worktree_self_delete` rule-out: I agree.** The block fires on `_cwd_is_inside(cwd, remove_target)`. For the gh-pr-create gates, cd-recovery only re-attributes WHICH repo is inspected; the verdict logic is unchanged. For self-delete, adopting recovery would change the block/allow VERDICT itself (`cd /safe && remove /wt` vs `cd /wt && remove /wt` flip on the recovered cwd) — a distinct behavioral decision, correctly out of scope for an identity-only cwd-anchor sweep, and documented with an in-code comment plus a tracking note. One observation for the eventual follow-up: today the hook reads the raw stdin cwd, so a `cd /wt && git worktree remove /wt` self-delete can currently slip past as a fail-open (cwd reads parent, not inside target). That is a pre-existing fail-open in the safe direction, not a regression from this PR, and is exactly what the separate evaluation should weigh.

**Two non-blocking observations on `extract_leading_cd_target`.** (a) It records the last absolute `cd` across ALL segments, so a contrived `gh pr create && cd /after` would anchor on `/after`; harmless because `/after` not being a git repo just fails open, and no one writes a trailing cd after pr-create — but the docstring's "precedes other work" phrasing is slightly looser than the impl. (b) Heredoc-body, tilde, relative, multi-arg, and subshell-paren cd shapes all correctly return None (probed independently) — the misroute surface I most care about is closed.

**CI green** — `gh pr view --json statusCheckRollup` shows all 5 checks SUCCESS (Ruff lint, format, Mypy, smoke, Pytest), consistent with my local 208/208.

Approving. The cross-repo routing fix is correct and well-tested, both gh-pr-create gates fail open on resolution failure (safe direction, no new false-block), and the self-delete carve-out is the disciplined call.

PR #538, opened 2026-05-30
fix(hooks): auto_set_env_test recognize newline as statement separator (#537)

### LINKED ISSUE #537: bug(hooks): auto_set_env_test doesn't recognize \n as a statement separator — mangled suggestion across multi-line bash
## Symptom

`auto_set_env_test` blocks legitimate multi-line bash commands and produces a syntactically nonsensical \"Suggested command\" that prepends `ENVIRONMENT=test` to the wrong token (e.g. `head -3` instead of the actual `pytest` invocation).

Reported by user 2026-05-30 — they noted \"I've seen this error pop up a few times.\"

## Repro (minimal)

```bash
cd /tmp
awk '/foo/' some.txt | head -3
echo \"---\"
ENVIRONMENT=test uv run pytest -q
```

Run as a single Bash tool call (multi-line string, no `&&`/`||`/`;` joiners between the awk line and the pytest line). Hook output:

> Suggested command:
>   cd /tmp
>   awk '/foo/' some.txt | ENVIRONMENT=test head -3        ← WRONG: env on head, not pytest
>   echo \"---\"
>   ENVIRONMENT=test uv run pytest -q

## Root cause

`auto_set_env_test._match_separator_at` only recognizes `&&`/`||`/`;`/`|` as segment separators. **Newlines are not in the set.** But in bash, an unescaped newline acts as a statement terminator (equivalent to `;`). Without newline recognition, `_split_segments` treats this:

```
awk '/foo/' some.txt | head -3
echo \"---\"
ENVIRONMENT=test uv run pytest -q
```

as 3 segments joined by `|`:
- Segment 0: `awk '/foo/' some.txt`
- Segment 1: `head -3\necho \"---\"\nENVIRONMENT=test uv run pytest -q`
- Separator: (no third pipe → single split)

Segment 1 contains `pytest` → test-bearing. Its leading token is `head` (not `ENVIRONMENT=test`), so the hook splices `ENVIRONMENT=test` before `head`. Mangled.

## Fix

Recognize `\\n` as a separator in `_match_separator_at`:

```python
def _match_separator_at(command: str, i: int) -> str | None:
    if command.startswith(\"&&\", i):
        return \"&&\"
    if command.startswith(\"||\", i):
        return \"||\"
    if command[i] == \";\":
        return \";\"
    if command[i] == \"|\":
        return \"|\"
    if command[i] == \"\\n\":      # NEW
        return \"\\n\"
    return None
```

Edge cases (handled by existing quote-aware / escape-aware state machine):
- Newline inside `'...'` or `\"...\"` → already not recognized as separator (quote-aware scanner stays in-quote)
- Backslash-line-continuation `\\\\\\n` → the existing `in_escape` state already consumes the next char after `\\\\`, so a literal `\\\\\\n` is treated as one escape sequence (not a separator). ✓

## Tests to add

In `.claude/hooks/tests/test_auto_set_env_test.py`:

1. `test_newline_separator_splits_segments` — multi-line script with pytest on its own line, env on pytest line → allow (don't block).
2. `test_newline_separator_blocks_missing_env_on_pytest_line` — multi-line script with pytest on its own line, env on a different line → block with correctly-targeted suggestion (env prepended to pytest, not to the prior line's tail).
3. `test_newline_inside_single_quotes_not_a_separator` — `awk '/multi\\nline/' ... | pytest ...` (pytest unquoted) → quotes contain a literal `\\n` text; existing scanner already handles, just regression-guard.
4. `test_backslash_line_continuation_not_a_separator` — `pytest a \\\\\\n   b` (continuation) → one logical segment, env-block check applies normally.

## Severity / priority

- **Severity: medium-high** — false-positive blocking valid commands degrades trust in the hook (memory: `feedback_enforcement_hierarchy` — hooks without operator trust decay).
- **Safety direction**: the hook is currently in 'safe failure direction' (false-block over false-allow per `feedback_safety_direction_over_ux_friction`). The fix preserves safe-direction while removing the false-block class.

## Scope

Single hook file + tests. No charter changes required — this is a hook-implementation bug, not a policy gap.

## Related

- The hook was last touched in W11 for #478 (control-flow body detection). This bug pre-dates that work; the #478 PR didn't change separator detection.

parametrization:
## Summary

Adds `\n` to the set of recognised statement separators in
`auto_set_env_test._match_separator_at`. Multi-line bash scripts with
pytest on its own line no longer have prior-line pipes/separators bleed
into the pytest segment, so suggestions land on the correct token.

## Root cause

Per #537: an unescaped newline acts as a statement terminator in bash
(equivalent to `;`), but `_match_separator_at` only recognised
`&&`/`||`/`;`/`|`. A script like

```
awk '/foo/' some.txt | head -3
ENVIRONMENT=test uv run pytest -q
```

was split as TWO segments joined by the awk-line's `|`:
- seg0: `awk '/foo/' some.txt`
- seg1: `head -3\nENVIRONMENT=test uv run pytest -q`

Pytest landed in seg1 with leading token `head`, so the rewriter spliced
`ENVIRONMENT=test` before `head` — mangled.

## Diff scope

- `auto_set_env_test.py`: one branch added to `_match_separator_at`;
  three docstring touches (module + `_split_segments` + matched-list
  enumeration) to keep policy contract in sync.
- `tests/test_auto_set_env_test.py`: new `NewlineSeparatorTests` class
  with 4 cases (allow / block / quoted-newline-not-a-separator /
  backslash-line-continuation regression-guard).

No charter changes. Pre-existing `_split_segments` quote-aware and
`in_escape` state-machine handle the edge cases — the fix is additive.

## Verification

Local pytest run on the suite:

```
$ ENVIRONMENT=test uv run pytest .claude/hooks/tests/test_auto_set_env_test.py -q
.....................................................................    [100%]
69 passed in 0.10s
```

65 pre-existing tests still pass, 4 new tests pass. Ruff format-check
and lint clean on both modified files.

## Cross-reference

User reported as recurring ("I've seen this error pop up a few times").
Tracked as #537.

## Caveats

- The safe-direction discipline established in #478 is preserved — when
  the rewriter can't produce a clean suggestion (control-flow body),
  the hook still HARD BLOCKs with a diagnostic. Newline recognition
  doesn't open any new allow-with-log paths.
- Backslash-newline line-continuation continues to be consumed by the
  existing `in_escape` flag (verified by the new regression test).

Closes #537
TechDebt: none

### DIFF
--- a/.claude/hooks/auto_set_env_test.py
+++ b/.claude/hooks/auto_set_env_test.py
@@ -14,9 +14,10 @@ Fires on:
 Matches (blocks if ENVIRONMENT=test missing on the test segment):
     Any Bash command containing a real test-runner invocation detected
     by the regexes `\\bpytest\\b` or `\\bmake\\s+test\\b`. The command is
-    split on shell separators (`&&`, `||`, `;`, `|`) into segments; each
-    test-bearing segment is checked independently for a leading
-    `ENVIRONMENT=test` env-block.
+    split on shell separators (`&&`, `||`, `;`, `|`, `\\n`) into segments;
+    each test-bearing segment is checked independently for a leading
+    `ENVIRONMENT=test` env-block. An unescaped newline acts as a statement
+    terminator equivalent to `;` (#537).
 
     Typical matched forms:
         pytest tests/
@@ -49,9 +50,10 @@ Detection order:
     1. Strip leading `VAR=value` tokens from the command for argv[0] check.
     2. If the next token is `gh`, ALLOW (return None).
     3. If the command contains `--body` or `--body-file`, ALLOW.
-    4. Split on `&&`/`||`/`;`/`|` into segments using a quote-aware
+    4. Split on `&&`/`||`/`;`/`|`/`\\n` into segments using a quote-aware
        state-machine scanner — separators inside `'...'`, `"..."`, or
-       `\\`-escaped positions are NOT recognised.
+       `\\`-escaped positions are NOT recognised. (A `\\`-then-newline
+       line-continuation is consumed by `in_escape` and so does NOT split.)
     5. For each segment containing `\\bpytest\\b` or `\\bmake\\s+test\\b`,
        require `\\bENVIRONMENT=test\\b` in that segment's leading env-block
        (subshell-aware: a segment that opens with `(` peeks inside the
@@ -184,7 +186,7 @@ def _has_body_flag(command: str) -> bool:
 
 
 def _split_segments(command: str) -> list[str]:
-    """Split command on `&&`/`||`/`;`/`|`, preserving separators as
+    """Split command on `&&`/`||`/`;`/`|`/`\\n`, preserving separators as
     alternating list entries: [seg0, sep0, seg1, sep1, ..., segN].
 
     Quote-aware (#478): separators inside `'...'`, `"..."`, or
@@ -283,6 +285,8 @@ def _match_separator_at(command: str, i: int) -> str | None:
         return ";"
     if command[i] == "|":
         return "|"
+    if command[i] == "\n":
+        return "\n"
     return None
 
 
--- a/.claude/hooks/tests/test_auto_set_env_test.py
+++ b/.claude/hooks/tests/test_auto_set_env_test.py
@@ -635,5 +635,81 @@ class ControlFlowHardBlockTests(unittest.TestCase):
         self.assertNotIn("; ENVIRONMENT=test do", result["reason"])
 
 
+class NewlineSeparatorTests(unittest.TestCase):
+    """Issue #537: an unescaped `\\n` between bash statements acts as a
+    statement terminator (equivalent to `;`). Pre-fix the separator
+    scanner only recognised `&&`/`||`/`;`/`|`, so multi-line scripts
+    where pytest is on its own line had the prior line glued onto the
+    pytest segment via `|`, producing mangled suggestions that prepended
+    ENVIRONMENT=test to the wrong token (e.g. `head -3` instead of
+    `pytest`)."""
+
+    def test_newline_separator_allows_when_env_on_pytest_line(self) -> None:
+        """NEG: multi-line script; env is in the pytest line's leading
+        env-block. The prior-line pipe (`awk ... | head -3`) does NOT
+        bleed into the pytest segment because `\\n` now splits. Allow."""
+        command = (
+            "cd /tmp\n"
+            "awk '/foo/' some.txt | head -3\n"
+            'echo "marker"\n'
+            "ENVIRONMENT=test uv run pytest -q"
+        )
+        result = hook.check(_bash(command))
+        self.assertIsNone(
+            result,
+            "newline must split segments so env on pytest line is recognised",
+        )
+
+    def test_newline_separator_blocks_when_env_missing_on_pytest_line(self) -> None:
+        """POS: multi-line script with pytest on its own line and no env.
+        Block, suggestion must prepend ENVIRONMENT=test IMMEDIATELY before
+        `uv run pytest` — NOT before `echo` or `cd` or `head`."""
+        command = 'cd /tmp\necho "marker"\nuv run pytest -q'
+        result = hook.check(_bash(command))
+        assert result is not None
+        self.assertEqual(result.get("decision"), "block")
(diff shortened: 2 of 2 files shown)


parametrization:
## TPM review: #538 — auto_set_env_test newline-separator fix

Read the hook and tests at head_sha bfb10a1 via `gh api .../contents?ref=<sha>` (no local checkout). Scope is exactly two files, matches what issue #537 specified: one branch added to the `_match_separator_at` arm-table at line 288-289 of the hook, plus the four named regression tests in the new `NewlineSeparatorTests` class.

Docstring contract sync — three call sites carry the recognized-separator list. All three now name the same set:
- Module header (line 17): `&&`, `||`, `;`, `|`, `\n`
- `_split_segments` docstring (line 189): same set
- Detection-order step 4 (line 53): same set, plus the line-continuation carve-out note

No drift; the contract is consistent across the file.

Hook-compose verification — traced the `\<newline>` line-continuation path through `_split_segments`. On `\`, the unquoted branch sets `in_escape=True` and consumes the byte to `buf`. On the next iteration the `\n` is caught by the `if in_escape` branch at the top of the loop, appended verbatim, and the escape state clears. The newline-as-separator arm in `_match_separator_at` never runs against an escaped newline because the in-escape consume short-circuits before separator detection. The `test_backslash_line_continuation_not_a_separator` regression-guard covers this.

Safety-direction observation — the new positive-path test `test_newline_separator_blocks_when_env_missing_on_pytest_line` asserts decision=block plus a splice-targeted assertion: the suggestion prepends env to `uv run pytest -q` and explicitly NOT to `cd /tmp` or `echo "marker"`. Hard block, no allow-with-log. Aligned with the safety-direction memory.

Control-flow cross-reference (W11 #478) — checked `_is_control_flow_bracketed`. It identifies the segment's leading token via `lstrip()` + `split(None, 1)`, both of which strip whitespace including newlines. A shape like `for f in a b; do pytest; done` still tokenises with `for`/`do`/etc. as the leading tokens of their segments, regardless of which separator (`;` or `\n`) delimits them. No regression: a newline-separated control-flow body will hit the existing HARD BLOCK path with the operator-readable diagnostic, not a mangled splice.

CI is 5/5 green at this head_sha. Fix is bounded, the docstring contract is in sync, and the four regression tests cover the issue's spec verbatim.

Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Release-Coordinator review: #538 — auto_set_env_test newline-separator fix

Read the hook + tests at HEAD (bfb10a1). Diff is tight: one branch added to `_match_separator_at` recognising `\n`, three docstring touches keeping the policy contract in sync, four tests. No `# type: ignore` / `# noqa` smell — the fix resolves the concern, doesn't paper over it.

Gate-continuity verification (release-discipline lens). The class of bug I most worried about was the inverse of #537 — a multi-line script where env appears on an earlier non-runner line and the runner is naked on a later line. Pre-fix the awk-line's `|` glued lines together and could in principle have created an accidental allow if the env-block scanner saw the env from an earlier statement. Post-fix the `\n` splits cleanly and `_segment_has_env_in_leading_block` examines each segment's OWN leading env-block, so the inverse case still blocks. I ran four probes against the PR-head hook (env-line-1 / runner-line-2 naked; env-line-1 / pipe-line-2 / runner-line-3 naked; subshell carve-out; pipe-line followed by runner-with-env) — all behave per #476 silent-bypass discipline. The fix preserves the gate's safe-failure direction.

Coverage observation (non-blocking, follow-up). Aino's `test_newline_separator_blocks_when_env_missing_on_pytest_line` covers the "env entirely absent" case (`cd / echo / pytest`). The "env on an earlier non-runner line" variant is exercised implicitly by the segment-by-segment env-block check but is not a dedicated test. Worth adding in a follow-up to make the #476 invariant explicit at the test layer rather than the runtime layer — companion to the `feedback_safety_direction_over_ux_friction` posture this hook embodies.

Wave-branch routing: base `deployments/phase-3/wave-12` — correct W12 target, no accidental main retarget. CI 5/5 green (ruff lint + format, mypy, smoke, pytest 69 in 0.10s — 4 new tests fit comfortably under the per-test runtime budget). No race-window concern; Aino has been idle and the verdict lands clean without an in-flight correction.

Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Issue #535, opened 2026-05-25
tech-debt(no_worktree_self_delete): recover invocation cwd to catch cd-into-worktree self-delete (#534/#528 follow-up)


parametrization:
## Context

Surfaced during PR #534 review (cwd-anchor pass). **Both** reviewers — Aino Virtanen (primary) and Santiago Ferreira (secondary) — independently converged on this: Wanjiku correctly RULED OUT `no_worktree_self_delete.py` from the #534 cwd-anchor sweep (adopting cwd-recovery there changes block/allow SAFETY verdicts, not just repo identity — a distinct behavioral decision, scope-appropriately deferred and documented in-code + PR body). But the residual gap is **real, not merely out-of-scope**, so filing it as a tracked follow-up rather than leaving it narrated only in a merged PR body.

## Problem

`no_worktree_self_delete.py` reads the raw stdin `cwd` (does not recover the actual invocation cwd). For a worktree subagent, `cd <wt> && git worktree remove <wt>` — with stdin cwd anchored at the parent — currently **fails OPEN (ALLOWS)** what is actually a self-delete (the #173 dangling-shell footgun). This is pre-existing and safe-direction (fail-open), NOT a regression introduced by #534.

## Fix (needs its own safety test matrix)

Apply careful last-cd-wins cwd recovery (the new `_shell_parse.resolve_invocation_cwd` primitive from #534) so:
- `cd <wt> && git worktree remove <wt>` → STARTS blocking (improvement — catches the self-delete)
- `cd /safe && git worktree remove <wt>` → still ALLOWS (correct)

The reason to defer was that this needs a dedicated safety test matrix (the block/allow verdict surface), not that recovery is wrong. Build that matrix as part of the fix.

## Refs

- PR #534 (cwd-anchor pass — where ruled out + documented)
- Epic #528 (cwd-anchor consolidation — sibling/follow-up)
- #173 (dangling-shell self-delete footgun — original)

parametrization:
Requestor: Fatima Okonkwo
Requestee: Aino Virtanen
RequestOrReplied: Request

**Wave 13 Kickoff — Phase 3**

This issue is assigned to you for p3-wave-13.
- Peer reviewer: Wanjiku Mwangi
- Secondary reviewer: Santiago Ferreira
- Branch from: `deployments/phase-3/wave-13`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
[...]

PR #553, opened 2026-05-31
fix(no_worktree_self_delete): recover invocation cwd to catch cd-into-worktree self-delete (Refs #535)

parametrization:
## fix(no_worktree_self_delete): recover invocation cwd to catch cd-into-worktree self-delete

#534/#528 follow-up. The hook read the stdin cwd as-is, so for a worktree subagent `cd <wt> && git worktree remove <wt>` (harness cwd anchored at the parent) **failed open** on the #173 self-delete.

### Change
Adopt the #534 `resolve_invocation_cwd` primitive (last-cd-wins recovery). The ancestry check now runs against the cwd the shell will actually be in when `git worktree remove` runs:
- `cd <wt> && git worktree remove <wt>` → resolves to `<wt>` → **BLOCK** (the self-delete the old read missed)
- `cd /safe && git worktree remove <wt>` → resolves to `/safe` → **ALLOW** (shell moves out first; correct)

Recovery only honors an absolute cd target that **exists on disk**; `cd <nonexistent>` falls back to the stdin cwd, so the guard never relaxes on a path that is not there.

### Safety test matrix (per issue — block/allow verdict surface)
New `CdRecoverySafetyMatrix` (8 cells): cd-into-target-from-parent blocks; cd-into-subdir blocks; cd-out-then-remove allows; cd-to-sibling allows; last-cd-wins both orderings (safe→unsafe blocks, unsafe→safe allows); cd-nonexistent falls back and blocks; bare-remove path unchanged. The prior `cd /safe`-literal test was retargeted to a nonexistent-path case and re-documented.

### Validation
- `test_no_worktree_self_delete.py`: 31 pass
- Full hook + lib suite: 1137 pass
- `uvx ruff@0.6.9 format --check .claude/hooks/` + `check` clean

Docstrings updated to replace the prior "cd is a plan not yet executed; left as-is intentionally" note with the new last-cd-wins semantics. Reviewers: Wanjiku + Santiago.

Refs #535

### DIFF
--- a/.claude/hooks/no_worktree_self_delete.py
+++ b/.claude/hooks/no_worktree_self_delete.py
@@ -25,10 +25,15 @@ Matches (one or more segments in the command, split on &&, ||, ;, |):
     - Global git options `-C <dir>` and `-c key=value` are skipped.
     - Short-option flags on `worktree remove` (`-f`, `--force`) are skipped.
     - `<path>` is the first non-flag argument after `remove`.
-    - A `cd <somewhere>` earlier in the compound command does NOT neutralize
-      the check: cd's inside the command string are plans the shell has not
-      yet executed, so the tool-call's actual cwd (from `input_data["cwd"]`
-      or `os.getcwd()`) is what matters.
+    - A leading `cd <dir>` IS honored when it names an absolute directory
+      that exists on disk (#535): the cwd used for the ancestry check is the
+      last such `cd` target (last-cd-wins via `resolve_invocation_cwd`),
+      because that is where the shell will be when `git worktree remove`
+      actually runs. `cd /safe && git worktree remove <wt>` therefore ALLOWS
+      (shell moves out first) while `cd <wt> && git worktree remove <wt>`
+      BLOCKS (the self-delete). A `cd <nonexistent>` is ignored — recovery
+      falls back to the stdin `cwd`, so the guard never relaxes on a path
+      that isn't there.
 
 Does NOT match:
     git worktree list            (no `remove` subcommand)
@@ -77,6 +82,7 @@ import sys
 from pathlib import Path
 
 sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
+from _shell_parse import resolve_invocation_cwd  # noqa: E402
 from annunaki_log import log_pretooluse_block  # noqa: E402
 
 # Shell operators that separate command segments. We split on these to reach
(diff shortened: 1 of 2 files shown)


parametrization:
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Release Coordinator review of the no_worktree_self_delete cwd-recovery fix (Refs #535).

Verified against the issue AC and the real diff at head 060c36e (gh API only, no checkout).

Behavior (RC safety lens — this changes block/allow verdicts, so it gets extra scrutiny):
- The hook now uses resolve_invocation_cwd instead of the bare stdin-cwd read. Confirmed in _shell_parse.py at head that resolve_invocation_cwd only honors a cd target when os.path.isdir(target) is true (absolute AND existing); otherwise it falls through to resolve_tool_cwd (stdin cwd then os.getcwd()). This is the load-bearing safety property: a cd to a nonexistent dir NEVER relaxes the guard — recovery falls back to the stdin cwd inside the worktree and still BLOCKS.
- The guard fails safe in both directions and is strictly MORE correct than before: cd-into-worktree then remove now BLOCKS (the #173 self-delete the pre-fix parent-anchored stdin read missed), while cd-to-safe then remove ALLOWS (shell moves out first). The fix never makes a previously-blocked self-delete start allowing on the strength of an absent path.

Test matrix (the issue explicitly required a dedicated safety matrix):
- New CdRecoverySafetyMatrix class pins (command, stdin-cwd) to (block|allow) cells across the full surface: improvement cases (cd-into-target/subdir from parent stdin cwd blocks), correctness cases (cd-out / cd-to-sibling allows), last-cd-wins ordering both directions, the nonexistent-cd fallback-blocks invariant, and the bare-no-cd unchanged path.
- Confirmed _WorktreeFixture at head exposes parent / wt_a / wt_a_sub / wt_b — the new cases resolve against real fixture attributes, not phantom ones.
- The prior test was correctly retargeted: the old chained-cd-then-remove-blocks (which asserted the now-superseded plan-not-executed semantics) became chained-cd-to-nonexistent-then-remove-blocks, guarding the don't-relax-on-absent-path invariant rather than leaving a stale contradictory assertion.

Docstring + inline comments updated to reflect the new honor-leading-cd semantics, including the explicit contrast with the gh-pr-create identity-recovery hooks (here recovery fixes the SAFETY verdict, not just repo identity). Audit trail intact.

CI gate: all 5 checks green at head — Pytest, Ruff lint, Ruff format, Mypy, smoke = SUCCESS. mergeable.

Approved.

parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Reviewed against issue #535 AC (diff at PR head via gh API; also read the shared primitive at wave HEAD, no parent-tree checkout).

What I verified:
- no_worktree_self_delete.py now derives the ancestry-check cwd via resolve_invocation_cwd (last-cd-wins) instead of the raw stdin cwd. The two AC-mandated outcomes hold: cd-into-target then remove BLOCKS (catches the #173/#534 self-delete the old stdin-cwd read missed when the harness cwd was anchored at the parent), and cd-to-safe then remove ALLOWS.
- The guard is NEVER relaxed (the load-bearing safety check). I read resolve_invocation_cwd and extract_leading_cd_target at wave HEAD: the cd target is honored ONLY when it startswith("/") (absolute) AND os.path.isdir is true. Relative targets return None; nonexistent absolute targets fall through to resolve_tool_cwd (stdin cwd then os.getcwd). So recovery can only MOVE the check to a real post-cd directory, never excuse a remove on the strength of a path that is not there.
- The new CdRecoverySafetyMatrix is a genuine block/allow verdict surface, not prose: improvement cases (cd-into-target / cd-into-subdir block even when stdin cwd is the parent), correctness cases (cd-out / cd-to-sibling allow), last-cd-wins ordering both directions, and the two no-relax invariants (cd-to-nonexistent falls back and blocks; bare remove unchanged). The pre-existing chained-cd test was correctly repurposed to a nonexistent-target case so it still asserts block.
- Docstring rewritten to describe the new last-cd-wins behavior accurately, including the explicit "cd <nonexistent> is ignored, guard never relaxes" note.

CI: all 5 GREEN including Pytest — proves the full safety matrix passes against the real hook + shared primitive. mergeable MERGEABLE.

No problems found. Approved.

Issue #533, opened 2026-05-25
tech-debt(session-start): anchor REPO_ROOT to parent (degrade safely if run from a worktree) — #528 family


parametrization:
## Context

Surfaced during PR #531 review (#526 — /session-start child-repo worktree prune). Both Aino Virtanen (author) and Santiago Ferreira (reviewer) flagged this as a defense-in-depth hardening item, out-of-scope for #526 and non-blocking. Filing against the cwd-anchor family (epic #528) since it's the same cwd-resolution root cause.

## Problem

`/session-start` Step 0 (and the new child-repo worktree-prune block) computes `REPO_ROOT="$(git rev-parse --show-toplevel)"`. If `/session-start` is ever invoked from a **worktree** rather than the parent main checkout, `--show-toplevel` resolves to the worktree, child-repo discovery anchors wrong, and children aren't found.

## Why it's not urgent

`/session-start` runs from the parent main checkout **by contract** (it's the mandated first action of every session, from the repo root). So in the real invocation path `REPO_ROOT` resolves to the parent correctly — this is not a live regression, just a fragile assumption.

## Fix (hardening)

Anchor `REPO_ROOT` to the parent repo deterministically rather than via `--show-toplevel` — e.g. resolve via `git rev-parse --git-common-dir` and walk to the parent, or anchor to a known marker (the parent's `CLAUDE.md` / `cross-repo-status.json`), so a stray run-from-worktree degrades safely instead of silently skipping child repos. Same cwd-resolution family as #525/#521.

## Refs

- PR #531 / #526 (where surfaced)
- Epic #528 (cwd-anchor consolidation) — same root-cause family

parametrization:
## Related session-start-prune hardening (from PR #531 review)

A second, lower-priority hardening item in the same `/session-start` child-repo prune code, surfaced by Wanjiku Mwangi's primary review of #531 — co-locating here since it's the same file + same W13 hardening pass:

**fetch-refspec completeness:** the prune block uses `git fetch --quiet origin main`, which updates `FETCH_HEAD` but may not update the local `refs/remotes/origin/main` ref (depends on configured refspec). Since the auto-remove guard tests `git merge-base --is-ancestor $head origin/main`, a stale `origin/main` could leave a genuinely-merged worktree **FLAGGED-UNMERGED instead of auto-removed**. This **fails SAFE** (the worktree is surfaced for manual removal, never wrongly deleted) — it's a completeness false-negative, not a safety bug.

[...]

PR #554, opened 2026-05-31
fix(session-start): anchor REPO_ROOT to parent, degrade safely from a worktree (Refs #533)

parametrization:
## fix(session-start): anchor REPO_ROOT to parent, degrade safely from a worktree

#528 cwd-anchor family. `/session-start` Step 0 used a bare `git rev-parse --show-toplevel` for `REPO_ROOT`; if invoked from a worktree that resolves to the worktree and child-repo discovery silently finds **0 children** (the `$REPO_ROOT/$child/.git` probes miss).

### Change
Anchor deterministically to the **parent org repo** via the parent of `git rev-parse --git-common-dir` (points at the main repo `.git` even from a linked worktree), with a `--show-toplevel` fallback, then **verify the parent marker** (`cross-repo-status.json` + `CLAUDE.md`). If the marker is absent (e.g. run from a child repo), warn loudly to stderr and keep the best candidate rather than silently skipping children — degrade safely, do not mis-anchor.

Steps 5 and 6 reference `$REPO_ROOT` in their own independent shell blocks where it was never actually defined (Skill blocks are separate shells); each re-anchors with the same compact parent-of-common-dir form.

### Verification
- **From a worktree**: REPO_ROOT now resolves to the parent and all **8 repos** (parent + 7 children) are discovered — was 0 children pre-fix.
- **From the parent checkout**: unchanged (8 repos).
- **From a child repo**: warns + does not mis-anchor.
- `bash -n` clean on the full Step 0 block.

Reviewers: Wanjiku + Santiago.

Refs #533

### DIFF
--- a/.claude/skills/session-start/SKILL.md
+++ b/.claude/skills/session-start/SKILL.md
@@ -7,7 +7,7 @@ description: "MANDATORY first action in every session — runs full startup prot
 
 **This skill MUST be invoked as the FIRST action in every new session.** Do not respond to the user's message, do not read files, do not run any other tool — invoke `/session-start` first. The user's actual request is handled AFTER this completes.
 
-> Note: all repo paths in bash blocks below are rooted at `$REPO_ROOT` to avoid cwd drift when the skill is invoked from a worktree or child-repo subdirectory (#149).
+> Note: all repo paths in bash blocks below are rooted at `$REPO_ROOT` to avoid cwd drift when the skill is invoked from a worktree or child-repo subdirectory (#149). `$REPO_ROOT` is anchored to the **parent org repo** deterministically via the parent of `git rev-parse --git-common-dir` (not `--show-toplevel`, which resolves to a worktree if run from one) and verified against the parent marker `cross-repo-status.json` + `CLAUDE.md` (#533). Each bash block re-derives it, since Skill blocks run as independent shells.
 
 ## Instructions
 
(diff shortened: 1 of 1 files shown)


parametrization:
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Release Coordinator review of the session-start REPO_ROOT parent-anchor (Refs #533).

Verified against the issue AC and the real diff at head a589bbe.

The fix (RC fail-safe lens — the concern is that a stray run-from-worktree must degrade safely, never silently mis-resolve and skip children):
- REPO_ROOT is now anchored to the parent org repo deterministically via the parent of git rev-parse --git-common-dir, which points at the MAIN repo .git even from a linked worktree — so it resolves the org root in BOTH the parent-checkout and run-from-worktree cases. This is exactly the issue AC (anchor to parent, not --show-toplevel).
- resolve_repo_root verifies the candidate against the parent markers cross-repo-status.json AND CLAUDE.md before accepting it. On miss it prints a loud WARN to stderr naming the unresolved path and telling the operator to run from the parent main checkout, then falls back rather than proceeding silently. This is the fail-safe direction the issue asked for: a stray worktree run degrades with a visible warning, not a silent child-repo skip.
- --show-toplevel is retained only as an intermediate fallback inside the resolver, not as the primary anchor — correct ordering.
- Each independent Skill bash block (Step 0, Step 4 cross-repo-status read, Step 5 feedback-log tail) re-derives REPO_ROOT, with a comment noting Skill blocks run as independent shells so the value does not carry over. The header note documents the parent-anchor rationale and the per-block re-derivation. No block is left assuming a carried-over value.

The fix is scoped to the hardening the issue describes (the real invocation path was never broken by contract; this removes the fragile assumption). No gate-softening — it only adds resolution robustness and a warn path. Audit trail (#533 / #149 references) intact in comments.

CI gate: all 5 checks green at head — Pytest, Ruff lint, Ruff format, Mypy, smoke = SUCCESS. mergeable.

Approved.

parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Reviewed against issue #533 AC (diff at PR head via gh API).

What I verified:
- Step 0 now anchors REPO_ROOT to the PARENT org repo via the parent of git rev-parse --git-common-dir, not --show-toplevel. This is the correct mechanism for the run-from-worktree case the issue describes: from a linked worktree, --git-common-dir points at the MAIN repo's .git, so its parent is the real org root, whereas --show-toplevel would resolve to the worktree and silently break the $REPO_ROOT/$child/.git probes.
- Degrades safely (the requested behavior): resolve_repo_root falls back to --show-toplevel when --git-common-dir yields nothing, verifies BOTH parent markers (cross-repo-status.json AND CLAUDE.md), and on miss emits a loud stderr WARN naming the unresolved candidate plus operator guidance ("run from the parent main checkout"), returning the best-effort candidate rather than aborting. Called as REPO_ROOT="$(resolve_repo_root)" so the non-zero return on the warn path is discarded and the session proceeds with the warning surfaced — silent-skip becomes loud-warn, exactly the hardening #533 asked for.
- The intro Note is updated to document the parent-anchor + marker-verify rationale and the per-block re-derivation (Skill bash blocks are independent shells).
- The two later re-anchor blocks (Steps 4 and 5) re-derive via the same --git-common-dir parent with a --show-toplevel fallback.

Non-blocking observation (not a change request): the Step 4/5 re-anchor blocks use the terser two-line form without the loud-marker-WARN branch that Step 0 carries; they silently fall to --show-toplevel if the marker is absent. Defensible since Step 0 runs first in the same session and already emits the warning, but flagging the asymmetry for awareness.

CI: all 5 GREEN. mergeable MERGEABLE.

No problems found. Approved.

Issue #567, opened 2026-05-31
hook(validate_pr_review): batch-loop `gh pr merge $var` fails open — block variable-PR merges (DECIDE, from P3W13 retro #4)


parametrization:
## Promotion provenance

DECIDE-tier promotion (skill/memory → hook) force-filed from the **P3W13 retro** at owner direction (proposal #4), ahead of the deterministic 2/3-citation threshold. Source memory: `feedback_batch_loop_merge_evades_pr_review_hook`.

## Problem

`validate_pr_review.py` (Hook 4) enforces the 2-reviewer gate by parsing the `gh pr merge <N>` command for a **literal PR number**, then checking that PR's issue-comments for ≥2 distinct `RequestOrReplied: Approved` verdicts. When the PR number is a **shell loop variable** — e.g.:

```bash
for pr in 48 49 50 51 52 53; do gh pr merge "$pr" --repo ... --merge; done
```

— the hook cannot resolve `$pr` to a concrete number at parse time, so it **fails open**: the merge proceeds with the gate effectively disabled. Standalone literal merges (`gh pr merge 54 ...`) are caught correctly.

## Evidence (2 instances)

1. **P3W11** wave→main: 4 loop-merges ran with 0 reviews and merged (caught post-hoc, #927).
2. **P3W13** ingest cluster: a 6-PR `for`-loop merge fired again. The 6 PRs were genuinely 2-reviewer-approved, but the gate did not verify that — it fail-opened.

## Proposed hook design (DECIDE — hooks are security-sensitive, never auto-applied)

Add a guard in `validate_pr_review.py` (or a sibling PreToolUse hook) that:
- Detects `gh pr merge` invoked where the PR-number argument is a shell variable (`$x`, `${x}`, `"$x"`) **and** the command is inside a `for`/`while` loop construct.
- **HARD BLOCK** with a diagnostic instructing one-literal-merge-per-call (so the gate stays active), per `feedback_safety_direction_over_ux_friction` (fail in the safe direction).
- Optionally: allow a loop only if every iterated value is a literal in an enumerable list AND each is individually gate-verified — but the simpler, safer first cut is to block variable-PR merges outright.

## Acceptance
- A `for`-loop `gh pr merge "$pr"` is blocked with a clear message.
- A literal `gh pr merge 54` still passes the existing 2-reviewer check unchanged.
[...]

PR #561, opened 2026-05-31
security(branch-protection): canonical ruleset spec + hook-validated admin-merge exceptions (Refs #322)

parametrization:
Phase-3 end-state criterion #4 — **CI failures block all merges, org-wide**. **`Refs #322`** — the canonical ruleset spec, the hook-side admin-merge gate, and a de-risked `noorinalabs-data-acquisition` pilot land in **W13** (this PR). The org-wide application of the ruleset to the remaining repos is the **W14 fast-follow**, so **#322 stays OPEN as the rollout tracker**. Criterion #4 is met when that W14 rollout has applied the ruleset to all 8 default branches — not by this PR alone (today only 1 of 8 carries it, by design: the live application was rolled back to the single pilot to avoid mid-wave deadlock).

## Scope (W13 parent-canonical deliverables)
This PR is the **canonical ruleset spec + reconciliation writeup + the hook-side admin-merge gate**. The live 7-repo application is the **W14 fast-follow** (after this wave's in-flight merges + the wave→main wrapup settle) — applying default-branch protection mid-wave would risk blocking our own merges. #322 tracks that rollout.

## The reconciliation (the high-value deliverable)
GitHub's native "require approvals" counts **formal GitHub reviews**, which our team **structurally cannot produce**: the `gh` principal IS the PR author (`parametrization`), so a formal self-approval **422s**, and we review via **issue-comment verdicts** gated by Hook 4 (`validate_pr_review`). A naive "require N approvals" rule would **deadlock every merge**.

**Spec shape that protects without deadlocking us:**
- target `~DEFAULT_BRANCH`, `enforcement: active`
- `pull_request` rule with **`required_approving_review_count: 0`** — require a PR, NOT formal approvals (reviewer-count stays with Hook 4, where #322's own scope note puts it)
- `required_status_checks` (`strict`) for repos with **unconditional** PR CI — require green CI
- `deletion` + `non_fast_forward` — restrict direct/destructive pushes
- a single `bypass_actors`: **Repository-admin role (`actor_id 5`, `always`)** — preserves the orchestrator `--admin` wave-merge / bootstrap / doc-sweep / emergency flow

[...]

Issue #552, opened 2026-05-31
hook(validate_pr_review): _ROSTER_DIR is parent-relative → child-repo PRs validated against wrong roster (blocked or evaded)


parametrization:
## Bug

`validate_pr_review.py` (Hook 4) resolves the roster from `_ROSTER_DIR = Path(__file__).resolve().parent.parent / "team" / "roster"` — i.e. **relative to the hook file**, which is always the PARENT repo's roster. The PreToolUse hook that fires in an org-orchestration session is the parent's, so child-repo PR merges are validated against the **parent roster** regardless of the `--repo` flag.

### Impact (observed P3W13, 2026-05-31)
Child-repo PRs reviewed by valid **child-roster** members are wrongly blocked because those personas aren't in the parent roster:
- **us#136** merge BLOCKED — reviewers Anya Kowalczyk + Idris Yusuf are valid `noorinalabs-user-service` roster members, but the hook listed only the 10-member parent roster and rejected them.
- **ig#934**, **us#138** would block identically (merged via `--admin` after manual verification that the 2 child-roster reviews were genuine).
- **ingest #48–#53** did NOT block — but only because a 6-merge batch command fail-opened the segment parser (the known batch-merge-evasion path, `feedback_batch_loop_merge_evades_pr_review_hook`). So they merged WITHOUT the gate validating at all. Both failure directions are bad.

Deploy PRs are unaffected only incidentally — the deploy roster personas (Bereket/Weronika/Nino/Aisha/Nurul/Lucas) happen to overlap the parent roster.

### Fix direction
When `--repo OWNER/NAME` is supplied (or derivable), resolve the roster from the **local child checkout** of that repo (`<repo_root>/<child-dir>/.claude/team/roster/`), not from `__file__`. Map `noorinalabs/<name>` → `<name>/.claude/team/roster`. Fall back to parent roster only when no child mapping exists. Pair with a test that a child-roster Requestor is accepted for a `--repo`-targeted child PR. Relates to `feedback_hook_cwd_anchor_subagent_worktree` (main#521/deploy#342 cwd-anchor family).

### Interim
Child-repo PRs with verified 2× child-roster reviews are being merged with `gh pr merge --admin` + explicit pre-merge verification of the Requestor set. This issue tracks the proper fix so `--admin` is not the standing path.

Issue #541, opened 2026-05-30
Phase 3 Wave 13 — Tech-debt reduction sweep + charter-change implementation


parametrization:
## Phase 3 Wave 13 — Tech-debt reduction sweep + charter + deploy-track operational (amended 2026-05-31)

Continuation of the Phase-3 tech-debt drive toward the **<10% exit-state gate** (criterion #9 — owner held at <10%, no relaxation per Proposal A 2026-05-31). Post-W12 cumulative TD ratio is **34.1%** (30/88). W13 targets aggressive reduction across us+deploy+ingest-platform PLUS deploy-track operational onboarding per Proposal B PLUS verify-and-close sweep per Proposal C.

> Reconciled by `/wave-scope 3 13` on 2026-05-31. Amended same day post-#546 reviewer ChangesRequested to add dict-row structure + Proposal B/C ratifications.

## Theme

Tech-debt reduction (us+deploy+ingest-platform minimum, main+isnad-graph stretches) + charter-change implementation (#542 throttle-stall, #543 hook test coverage, #544 meta-issue freshness) + **deploy-track operational items** (#325 wrapup gate, #329 smoke tests) per Proposal B + **verify-and-close on phase-3 criteria #4/#5/#6/#7** per Proposal C.

## Per-repo TD picture (post-W12 → W13 target)

| Repo | Open | TD | TD% | W13 target |
|---|---|---|---|---|
| **noorinalabs-isnad-ingest-platform** | 8 | 7 | 87.5% | <50% (drop 4+ TD items) — **highest-priority repo** |
| **noorinalabs-deploy** | 16 | 8 | 50.0% | <30% (drop 3+ TD items) — minimum scope |
| **noorinalabs-user-service** | 7 | 3 | 42.9% | <20% (drop 2+ TD items) — minimum scope |
| noorinalabs-main | 24 | 5 | 20.8% | stretch — charter + verify-and-close help |
| noorinalabs-isnad-graph | 15 | 3 | 20.0% | stretch — #932 in scope |

design-system (0%), data-acquisition (0%), landing-page (9.1%) already under target; no required W13 work.

## Tier 1 — Charter-change implementations (3 items, all main, priority charter)

| Issue | Title | Implementer | Reviewer | Reviewer 2 |
|---|---|---|---|---|
| **#542** | charter(agents): throttle-stall detection + auto-takeover thresholds | Aino Virtanen | Wanjiku Mwangi | Santiago Ferreira |
[...]

PR #568, opened 2026-05-31
charter+skill(P3W13-retro): adopt Refs-from-first-PR (#2) + high-volume merge checkpoint (#3)

parametrization:
## Summary
Adopts two P3W13 retro process changes (owner-approved):
- **#2** `charter/pull-requests.md` § Closes-vs-Refs Disposition — end-state/rollout-remaining issues use `Refs` from the **first** PR, decided at brief time, never flipped. Prevents the #561 flip-flop.
- **#3** `skill/wave-wrapup` Step 10.5 — high-volume remote-merge checkpoint (fetch + reset before first local bookkeeping commit when N≥10 PRs merged remotely; relocate local edits first).

## Related Issues
Refs #541 (P3W13 retro). Proposals #1 (#552) and #4 (#567) tracked separately.

## Review Checklist
- [ ] Refs rule placement + wording correct (pull-requests.md)
- [ ] Merge-checkpoint guard is non-destructive-by-default (warns to relocate edits before reset)
- [ ] No applied behavior beyond the two doc/skill edits


parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Reviewed the full diff via `gh pr diff 568`; it is exactly two edits — the new Closes-vs-Refs disposition section in charter pull-requests.md and the high-volume remote-merge checkpoint added to wave-wrapup Step 10.5. No applied behavior beyond those two doc/skill edits.

The disposition section is correctly placed (immediately after the existing PR-body / Closes-#N guidance) and correctly worded: it pins the choice at brief-authoring time, defines the Closes case as full-acceptance-surface delivery and the Refs case as remaining per-repo rollout or prod-gated or multi-PR work, and states the operative rule that any acceptance remaining after this PR merges means Refs from the first PR rather than a post-merge flip. It cross-references the owner-pivot-supersedes lesson and cites the branch-protection origin case, so the rationale is grounded.

The Step 10.5 checkpoint guard is non-destructive by default, which was the key thing to verify. The inline comment to stash or relocate in-progress local edits appears BEFORE the hard-reset line, and the Why paragraph explicitly calls the reset destructive and instructs relocating local edits first. The guard is also correctly gated (only fires when behind origin, and only for the high-volume threshold), so it will not surprise a clean tree.

Approving.

parametrization:
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Release-Coordinator review of the two P3W13 charter+skill adoptions.

**Refs-from-first-PR disposition rule (pull-requests.md).** This cleanly covers the end-state / rollout-remaining class without contradicting existing guidance. I checked the rule against the two prior sections in the same file:

- The PR-template line that says reference the issue with the close-keyword form is the default case; the new section adds a brief-time carve-out, and there is no hook anywhere under .claude/hooks that literally enforces the close-keyword in a PR body, so there is no hook-versus-charter conflict introduced.
- The existing runtime-gated section already mandated the reference-keyword form for prod-apply-gated issues. The new section is a strict superset of that rule rather than a competing one: it names the prod-gated runtime step as one member of the broader set alongside end-state per-repo rollout and multi-PR sequences. The two harmonize, with the runtime-gated section remaining the more detailed specialization. No daylight between them.

The decided-at-brief-time, never-flipped framing is the right shape for release coordination: a late disposition flip is a routing change on an in-flight artifact, which is the same churn the supersede protocol exists to prevent. Codifying it up front removes a recurring round-trip class.

**Wave-wrapup merge checkpoint (Step 10.5).** Correctly placed and safe-by-default:

- The checkpoint block is the first content under the 10.5 heading, ahead of the actual counter-key write further down in the step, so it fires before the first bookkeeping commit exactly as the proposal specified. It also names the ontology-rebuild commit and the wrap-marker commit as the protected set, which is the full first-write surface.
- The hard reset is guarded by a behind-greater-than-zero conditional, so it is a no-op when local is already current; it fetches and resets only against the current branch, so it cannot cross-branch; and the inline comment explicitly warns to relocate in-progress local edits first because the reset is destructive. Presenting it as orchestrator-run prose with the stash caveat called out, rather than an unguarded automated step, is the safe form. The N-greater-or-equal-ten threshold scopes it to genuinely high-volume remote-merge sessions.

The Why blocks on both changes are accurate to the originating incidents. Both adoptions are minimal, additive, and release-safe.

Approving as second reviewer.

Issue #498, opened 2026-05-18
tech-debt(hook-4): validate_pr_review accepts non-roster Requestor strings (asymmetry with validate_commit_identity roster discipline)


parametrization:
## Background

P3W11 session 2026-05-18: while implementing a follow-up PR (#492, Maeve Callahan author), discovered that PR #487's verdict comments were posted under **non-roster persona names** — "Camila Restrepo" (1°) and "Imelda Santos" (2°). Neither name exists in `.claude/team/roster/` (parent main has no "Camila" or "Imelda"). Hook 4 (`validate_pr_review.py`) accepted both verdicts and counted them as 2-of-2 Approved, allowing PR #487 to merge.

The Hook 4 acceptance is correct as currently implemented: the 2-reviewer threshold at `_extract_charter_field` / count-distinct-Requestor logic (line 245, 295, 340-363) only requires distinct Requestor STRINGS — there is no cross-check against `.claude/team/roster/` filenames. The roster IS read (line 379-422) but only for the charter-enforcer 1-of-1 exception path (`_charter_enforcer_names`), not for the general 2-of-2 path.

## Failure mode

Silent acceptance of fictional / non-roster Requestor names. The gate that exists to ensure "2 distinct ROSTER MEMBERS approved" effectively becomes "2 distinct STRINGS appeared as Requestor values." Failure direction: unsafe (false acceptance).

**P3W11 instance** (this session):
- PR #487 (Santiago's #470 doc-only sync) — Requestor #1: "Camila Restrepo", Requestor #2: "Imelda Santos". Neither in `roster/`. Hook 4 counted both, gate passed, PR merged.
- Discovered when filing #492 (follow-up to #487 reviewer-flagged gaps) — implementer tried to commit as "Camila Vargas" (variant of the persona) and `validate_commit_identity.py` HARD-BLOCKED because that hook DOES validate against roster. Hook 4 should have the same discipline.

The mismatch between Hook 4 (no roster check) and `validate_commit_identity` (strict roster check) is itself the asymmetry that allows the gap.

## Proposed fix

Extend Hook 4 to validate Requestor against roster:

1. **Add a `_roster_names()` helper** that returns the full set of canonical names from `.claude/team/roster/` filenames (parse `{role}_{firstname}_{lastname}.md` → "Firstname Lastname"). Likely refactor of existing `_charter_enforcer_names` to share a parser.

[...]

Issue #327, opened 2026-05-08
P3 end-state #6: Pre-commit + pre-push hooks in every repo, mirroring GitHub Actions checks


parametrization:
Tracking issue for **Phase 3 end-state criterion #6**.

See `.claude/team/phases/phase-3.md` § End-state criteria for full phase context.

## Criterion

**Pre-commit + pre-push hooks in every repo, mirroring GitHub Actions checks.** Local catches what CI catches; merge-time CI runs are confirmation, not first detection.

## Definition of done

- Each of 8 repos has `.pre-commit-config.yaml` (or equivalent) wired up
- Pre-commit hooks reflect the CI ruff/mypy/format/lint set used by GitHub Actions
- Pre-push hooks run the heavier checks (typecheck, full test suite where fast enough)
- Per-repo README/CLAUDE.md documents hook setup
- A CI gate verifies that hook config and CI workflow stay in sync (catches drift)

## Out of scope

- Auto-installing hooks on machine — leave to `pre-commit install` in onboarding

## Dependencies

- Sibling of #5 (CI scope must be defined before hooks can mirror)
- Charter `feedback_actionlint_needs_shellcheck.md` lesson: install all required tools

## Sub-tickets

(filed at scoping time, one per repo)

parametrization:
Requestor: Fatima Okonkwo
Requestee: Aino Virtanen
RequestOrReplied: Request

**Wave 13 Kickoff — Phase 3**

This issue is assigned to you for p3-wave-13.
- Peer reviewer: Wanjiku Mwangi
- Secondary reviewer: Santiago Ferreira
- Branch from: `deployments/phase-3/wave-13`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: verify-and-close (per charter § Wave Planning & Priority)

Please begin implementation.

parametrization:
## Tier-5 verify-and-close audit — #327 (Pre-commit + pre-push hooks in every repo, mirroring CI)

Audited `.pre-commit-config.yaml` presence + stages across all 8 repos at `main` HEAD, 2026-05-31.

### Definition-of-done element status

| Element | Status |
|---|---|
| Each of 8 repos has `.pre-commit-config.yaml` | **MET** — all 8 present |
[...]

PR #570, opened 2026-05-31
fix(validate_pr_review): resolve roster per target repo — child-repo PRs hit the real 2-reviewer gate (Closes #552)

### LINKED ISSUE #552: hook(validate_pr_review): _ROSTER_DIR is parent-relative → child-repo PRs validated against wrong roster (blocked or evaded)

parametrization:
## Summary

Hook 4 (`validate_pr_review.py`) enforces the 2-reviewer gate by validating reviewer names against a roster directory. That directory (`_ROSTER_DIR`) was hardcoded to the **parent** repo's `.claude/team/roster/`, resolved relative to the hook file. The `--repo OWNER/NAME` flag was already parsed and forwarded to every `gh` subcall, but `_load_roster_names()` / `load_charter_enforcer_names()` ignored it.

So when a `gh pr merge --repo noorinalabs/<child>` targeted a **child** repo, legitimate child-repo reviewers (Anya Kowalczyk, Idris Yusuf, …) were filtered as "non-roster" at the #498 gate, dropping the distinct-reviewer count below 2 → **false block**. In P3W13 this forced `--admin` overrides on **14/37** child-repo PRs, defeating the gate (retro proposal #1).

### Fix
Resolve the roster relative to the PR's **target** repo and **union** the parent roster with the named child repo's `.claude/team/roster/`:
- Child repos are checked out as siblings under the parent root; child rosters are canonical for reviewer pairing per `CLAUDE.md`.
- A reviewer valid in **either** the org-level team or the target child repo passes. Parent-only PRs are unchanged (back-compat).
- A child-only persona does **not** leak into a parent-repo PR's gate.
- When a child repo is named but its roster dir is unreadable, fail in the **safe direction** — HARD BLOCK with a diagnostic, never a silent parent-only fallback (which would re-block child reviewers exactly as the bug did). Per `feedback_safety_direction_over_ux_friction`.

New helpers: `_child_roster_dir(repo)`, `_resolve_roster_dirs(repo)`, `RosterResolutionError`. `_iter_roster_entries` gains a `roster_dir` param; `repo` is threaded through `_load_roster_names`, `load_charter_enforcer_names`, and `is_single_reviewer_exception`.

### Tests
+18 cases (103 total pass). Covers the 6 charter-required classes:
- parent-PR-still-passes
- child-PR-valid-child-reviewer-passes
- child-PR-unknown-reviewer-blocks
- `--repo`-flag parsing (`_child_roster_dir` resolution)
- union roster (parent ∪ child)
- roster-dir-unresolvable HARD-BLOCK

Plus: child-persona-does-not-leak-on-parent-PR, child single-reviewer exception via child enforcer, `_resolve_roster_dirs` parent/child/raise. Fixtures build a hermetic tmp parent+child roster tree (monkeypatch `_ROSTER_DIR` / `_PARENT_REPO_ROOT`) — no dependency on the live sibling checkout.

## Related Issues

Closes #552

## Review Checklist

- [ ] Root cause matches `validate_pr_review.py` at HEAD (hardcoded `_ROSTER_DIR`, ignored `--repo`)
- [ ] Union semantics correct; parent PRs unaffected; no child-persona leak onto parent PRs
- [ ] Missing child roster fails closed (HARD BLOCK), not silent fallback
- [ ] Test coverage spans the 6 required classes
- [ ] ruff format/lint, mypy, pytest all green

### DIFF
--- a/.claude/hooks/tests/test_validate_pr_review.py
+++ b/.claude/hooks/tests/test_validate_pr_review.py
@@ -1290,5 +1290,236 @@ class ExtractRepoCallSiteTests(unittest.TestCase):
         )
 
 
+class ChildRosterResolutionTests(unittest.TestCase):
+    """Issue #552: the roster the 2-reviewer gate validates against must be
+    resolved relative to the PR's TARGET repo, not hardcoded to the parent.
+
+    Pre-#552 `_ROSTER_DIR` was fixed to the parent repo's roster, so a child
+    repo PR reviewed by legitimate child-repo personas (e.g. Anya Kowalczyk,
+    Idris Yusuf) had those reviewers filtered out as "non-roster" — the gate
+    blocked the merge or, after the merge was forced, defeated itself via
+    `--admin`. In P3W13 this forced `--admin` on 14/37 child-repo PRs.
+
+    The fix unions the parent roster with the named child repo's
+    `.claude/team/roster/`. These tests build a hermetic on-disk parent+child
+    roster tree under a tmp dir and monkeypatch `_ROSTER_DIR` /
+    `_PARENT_REPO_ROOT` so they do not depend on the live sibling checkout.
+    """
+
+    PARENT_PERSONAS = {
+        "standards_lead_aino": "Aino Virtanen",
+        "program_director_nadia": "Nadia Khoury",
+    }
+    # Child personas — NOT present in the parent roster. `manager_*` is a
+    # charter-enforcer prefix; `engineer_*` is a plain reviewer.
+    CHILD_PERSONAS = {
+        "engineer_anya": "Anya Kowalczyk",
+        "engineer_idris": "Idris Yusuf",
+        "manager_bereket": "Bereket Tadesse",
+    }
+    CHILD_REPO_NAME = "noorinalabs-isnad-graph"
+    CHILD_REPO = f"noorinalabs/{CHILD_REPO_NAME}"
+
+    def _write_roster(self, roster_dir: Path, personas: dict[str, str]) -> None:
+        roster_dir.mkdir(parents=True, exist_ok=True)
+        for slug, name in personas.items():
+            (roster_dir / f"{slug}.md").write_text(
+                f"# Roster Card\n\n## Identity\n- **Name:** {name}\n", encoding="utf-8"
+            )
+
+    def setUp(self) -> None:
+        import tempfile
+
+        self._tmp = tempfile.TemporaryDirectory()
+        root = Path(self._tmp.name)
+        # Parent repo root holds .claude/team/roster and the child repo sibling.
+        parent_repo = root / "noorinalabs-main"
+        self._parent_roster = parent_repo / ".claude" / "team" / "roster"
+        self._write_roster(self._parent_roster, self.PARENT_PERSONAS)
+        child_roster = parent_repo / self.CHILD_REPO_NAME / ".claude" / "team" / "roster"
+        self._write_roster(child_roster, self.CHILD_PERSONAS)
+
+        self._patchers = [
+            mock.patch.object(hook, "_ROSTER_DIR", self._parent_roster),
+            mock.patch.object(hook, "_PARENT_REPO_ROOT", parent_repo),
+        ]
+        for p in self._patchers:
+            p.start()
+
+    def tearDown(self) -> None:
+        for p in self._patchers:
+            p.stop()
+        self._tmp.cleanup()
+
+    # --- _child_roster_dir resolution -------------------------------------
+
+    def test_child_roster_dir_resolves_for_child_repo(self):
+        d = hook._child_roster_dir(self.CHILD_REPO)
+        assert d is not None
+        self.assertTrue(d.is_dir())
+        self.assertEqual(d.parent.parent.parent.name, self.CHILD_REPO_NAME)
+
+    def test_child_roster_dir_none_for_parent_repo(self):
+        """`--repo noorinalabs/noorinalabs-main` → no distinct child dir."""
+        self.assertIsNone(hook._child_roster_dir("noorinalabs/noorinalabs-main"))
+
+    def test_child_roster_dir_none_for_absent_repo(self):
+        self.assertIsNone(hook._child_roster_dir(None))
+        self.assertIsNone(hook._child_roster_dir("not-a-repo-spec"))
+
+    # --- union roster -----------------------------------------------------
+
+    def test_load_roster_names_unions_parent_and_child(self):
+        names = hook._load_roster_names(repo=self.CHILD_REPO)
+        # Parent personas present.
+        self.assertIn("aino virtanen", names)
+        self.assertIn("nadia khoury", names)
+        # Child personas unioned in (#552).
+        self.assertIn("anya kowalczyk", names)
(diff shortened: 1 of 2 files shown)


parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Reviewed the diff via `gh pr diff 570` and read the hook source at the PR head sha (fe9cd32) through the contents API — no parent checkout. This fixes #552, the W13 top pain point where the gate validated child-repo PRs against the parent roster only and forced `--admin` on 14 of 37 child PRs. Three things I focused on, all verified.

First, the roster union. The resolver seeds the parent dir, then appends the named child's roster dir only when the child-dir helper returns a real path, and both the all-members loader and the enforcer loader iterate that list and OR the name sets together. The target repo is read from the command via the existing shared shlex-tokenized flag parser (the one that already handles both the long and short repo flags plus a regex fallback), so the parent-versus-child decision rests on a battle-tested input parse rather than a new ad-hoc one. Back-compat is preserved: no repo flag, and a repo flag that names the parent itself, both resolve to parent-only — the helper returns nothing-to-append in those cases, and there are explicit regression tests for each.

Second, the unresolvable-roster case fails in the safe direction. When a child is named but its roster dir is absent on disk, the resolver raises rather than silently degrading to parent-only — degrading would re-block legitimate child reviewers exactly as #552 did. The caller catches that and returns a hard block with an actionable diagnostic, and — the part that matters — that catch sits ahead of all reviewer counting in the function body, so an unreadable child roster can never fail open into a free merge. This is the right read of the safety-over-UX-friction lesson.

Third, the eighteen new tests genuinely exercise the six classes, including the two I most wanted to see: a child PR whose second reviewer is in neither roster still blocks at one-of-two, and a child-only persona does not satisfy the gate on a parent PR (no cross-repo leak). The missing-child-roster test asserts the block decision and the could-not-be-resolved diagnostic, and there are separate cases for the union, the enforcer single-reviewer exception honoring a child manager, and the parent-PR regression. The suite is hermetic — it builds a tmp parent-plus-child roster tree and monkeypatches the two module path constants — so it does not depend on a live sibling checkout, which is the correct way to test a sibling-relative resolver.

On CI: the one red check is the sync-drift gate complaining that actionlint is not mirrored in pre-commit. I confirmed the most recent main commit fails that same gate identically, so this is the known org-wide #327 drift, not anything this PR introduced; it is correctly deferred to a W14 item. Everything this PR actually touches is green — type check, the hooks-and-lib pytest job that runs these new tests, both ruff jobs, and the smoke test all pass.

No tech debt added. Approving.

parametrization:
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Release-Coordinator merge-safety review of the #552 child-roster resolution fix. I read the source hunk and the new tests at PR head, traced the path-resolution logic, and ran the suite against the head version.

**(1) Parent PRs stay validated against the parent roster only — no child leak.** When the merge command carries no repo flag, the resolver returns the parent dir alone, so a child-only persona cannot satisfy the gate on a parent PR. Both the targeted unit (parent-only when no repo) and the end-to-end case (a child-only persona on a parent PR blocks at one-of-two) cover this directly, and I confirmed the end-to-end one fails-closed as expected. The union is additive in one direction only: a child roster can be added when the target names a child, never the reverse, and the parent dir is always first in the list.

**(2) The repo-flag value cannot be used to widen the accepted roster in any way that matters.** I traced the resolver against spoof and traversal inputs. A value naming the parent itself collapses to None (no distinct child dir). A dotted traversal pointing back at the parent resolves to the parent roster — a self-union no-op. A traversal pointing outside the tree resolves to a path that does not exist, which raises and hard-blocks rather than silently accepting. Crucially the union can only ADD names; it can never drop the parent roster nor lower the two-distinct-reviewer count, so there is no path by which a crafted value loosens the gate. And the same value is forwarded to the underlying merge, which validates the owner-slash-name shape — a value the resolver would have to be tricked by is also a value the real merge would reject, so there is no live merge to wave through. This is operator-trusted territory anyway (the sanctioned bypass is the admin flag), not a privilege boundary the hook is asked to defend, so the additive-only design is the right shape.

**(3) The hard-block-on-unresolvable will not false-block legitimate parent PRs.** The raise fires only when a child repo is explicitly named AND its roster dir is absent on disk. A legitimate parent PR never names a child, so it never enters the resolution branch and never reaches the raise — confirmed by the parent-PR regression staying green while the absent-child case hard-blocks with a diagnostic. The block path also logs the pretooluse block and emits a fix-oriented message rather than degrading to a silent parent-only fallback, which is the correct safe-direction choice (a silent fallback would re-introduce the exact false-block that #552 was filed to remove).

**Tests and CI.** The two new test classes (15 cases) pass against the head hook. The single-reviewer exception correctly honors a child-repo enforcer on a child PR, and an unknown reviewer on a child PR still blocks at one-of-two — so widening the roster did not weaken the non-roster filter. CI corroborates: ruff, format, mypy, smoke, and the full pytest suite are all green; the only red is the pre-existing sync-drift / actionlint gate that fails on clean main and is already tracked for W14 — not introduced by and not a blocker for this PR.

Merge-safe. Approving as second reviewer.

PR #572, opened 2026-06-01
ci(P3W14): add actionlint mirror to .pre-commit-config.yaml (#571)

### LINKED ISSUE #571: tech-debt(#327): .pre-commit-config.yaml missing actionlint mirror — sync-drift gate red on main
## Problem
The `Pre-commit ⇄ CI sync-drift gate` (`.claude/lib/pre_commit_ci_sync.py`, added W13 #562/#327) is RED on `main`: CI enforces **`actionlint`** (the `Config — Workflow actionlint` job) but `.pre-commit-config.yaml` does NOT mirror it. So a local commit/push does not catch a workflow-lint error that CI will reject — the exact divergence #327 exists to prevent.

This is **pre-existing org-wide drift** (fails identically on clean `origin/main` HEAD), surfaced when W13-retro PRs #568 + #570 (both touch `.claude/`) triggered the gate. It blocks clean (non-admin) merges of any `.claude/`-touching PR.

## Fix
Add an `actionlint` hook to `.pre-commit-config.yaml`, pinned to the same version CI uses, scoped to `.github/workflows/`. Verify `python3 .claude/lib/pre_commit_ci_sync.py .` then exits 0. Per CLAUDE.md, actionlint needs `shellcheck` on PATH for its shell-integration (memory `feedback_actionlint_needs_shellcheck`) — wire that into the hook env or document the dependency.

## Why W14 / #327
This is squarely the **end-state #327** mechanism (pre-commit mirrors CI in every repo). The parent's own config is the first instance; the per-repo rollout to the 7 children is the broader #327 W14 work.

Refs #327. Surfaced from P3W13 retro adoption (PRs #568/#570).

parametrization:
## Summary
The `Pre-commit ⇄ CI sync-drift gate` (`.claude/lib/pre_commit_ci_sync.py`) globs **all** `.github/workflows/*.y*ml`. `docs.yml` ships the `Config — Workflow actionlint` job, so `actionlint` is a CI-enforced check kind. With no matching pre-commit hook, the gate reported CI-enforced-but-not-local drift and exited 1 — the sync-drift gate was **RED on main**.

This adds the upstream `rhysd/actionlint` golang pre-commit hook, pinned to **v1.7.12** (what CI's `download-actionlint.bash` off `main` resolves to). The hook already scopes to `^.github/workflows/` and uses shellcheck if present — matching the CI job's embedded-shellcheck behavior (`feedback_actionlint_needs_shellcheck`).

**Verification:** `python3 .claude/lib/pre_commit_ci_sync.py .` now exits **0** (was 1). `actionlint .github/workflows/*.yml` exits 0 on the existing workflows. Config parses as valid YAML; the hook env installed cleanly at commit time.

Config-only change — no Python under `.claude/{hooks,lib}` touched, so ruff/mypy/pytest surface is unaffected.

## Related Issues
Closes #571

## Review Checklist
- [ ] `rev: v1.7.12` is a real rhysd/actionlint release and matches CI's resolved version
- [ ] Hook scope `^.github/workflows/` matches the CI job's target
- [ ] shellcheck-on-PATH requirement documented in the config comment
- [ ] sync-drift gate exits 0 with this change

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -38,6 +38,22 @@ repos:
         files: ^\.claude/(hooks|lib)/.*\.py$
         exclude: ^\.claude/worktrees/
 
+  # actionlint mirrors the `Config — Workflow actionlint` CI job in
+  # .github/workflows/docs.yml (the sync-drift gate globs ALL workflow files,
+  # so a CI-enforced `actionlint` kind not mirrored here turns the gate red —
+  # #571). The upstream golang hook builds a pinned actionlint and already
+  # scopes to ^.github/workflows/. shellcheck MUST be on PATH so actionlint's
+  # embedded shellcheck integration runs rather than silently skipping
+  # (feedback_actionlint_needs_shellcheck) — CI gets this from ubuntu-latest's
+  # preinstalled shellcheck; locally, install shellcheck before committing.
+  # Pinned to v1.7.12 to match what CI's `download-actionlint.bash` (tracking
+  # the latest release off `main`) resolves to.
+  - repo: https://github.com/rhysd/actionlint
+    rev: v1.7.12
+    hooks:
+      - id: actionlint
+        name: actionlint
+
   # Heavier checks run at pre-push (mirror CI's mypy + pytest jobs). Running as
   # `local`/`system` hooks so we use the system tools and don't drift from
   # whatever CI pins.


parametrization:
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Release-coordinator review (CI / merge-safety lens). Config-only change — one file, +16/-0, no behavioral hook code touched.

**Checks**: all 11 green on statusCheckRollup. The two that matter here both pass — the sync-drift gate (the blocker that turned #568/#570 red) and the workflow-actionlint job. Branch is MERGEABLE against the wave-14 base.

**Pin correctness**: verified the upstream golang hook is pinned to v1.7.12, and CI's `download-actionlint.bash` (fetched off `main` with no version arg) resolves to the latest release, which is v1.7.12 today. So local and CI run the same actionlint version, and the inline comment documents the tracking relationship so a future drift is explicable rather than surprising.

**Local dependency**: the comment correctly flags that shellcheck must be on PATH so actionlint's embedded shellcheck integration runs instead of silently skipping. The actionlint hook itself self-builds via pre-commit's golang language, so no manual toolchain install is needed for the hook to run. Engineers missing shellcheck get a degraded (shellcheck-skipped) local pass, but CI's ubuntu-latest has it preinstalled and stays authoritative — a documented soft dependency, not a gap.

Faithful mirror of the CI job. Approve.

parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

First-reviewer pass on the actionlint mirror. Verified read-only against the PR head in an isolated clone — no parent checkout.

What I confirmed:

- Drift closure is real and bidirectional. Running the sync gate against the branch head prints "OK: pre-commit config mirrors all CI-enforced checks" and exits 0. Stripping just the new actionlint block from the same config flips the gate to exit 1 with "DRIFT: CI enforces these checks but pre-commit does NOT run them locally: actionlint" — so this change is precisely what closes the gap, with nothing else carrying it.
- Version alignment with CI holds. CI's docs.yml job pulls the download script off the upstream main branch, which resolves to the newest published release; that release is currently v1.7.12 (confirmed it is the tag flagged Latest on rhysd/actionlint, published 2026-03-30). The pin therefore tracks what CI actually runs today.
- Scope matches the CI job. The upstream golang hook ships a default that already restricts to workflow files under the dot-github path and limits to yaml types, so relying on that default keeps the local hook pointed at the same surface the CI job lints. No override was needed and none was added.
- The shellcheck-on-path expectation is captured in the config comment, mirroring the embedded-shellcheck behavior the CI runner gets for free from its preinstalled tooling. That matches the prior discipline around actionlint silently skipping its shellcheck integration when the binary is absent.
- Change is config-only: a single file, sixteen additions, zero deletions, no Python under the hooks or lib trees touched, so the ruff, mypy, and pytest surfaces are untouched. Nothing unrelated rides along.

No must-fix items. Approving as first reviewer.

PR #563, opened 2026-05-31
ci(docs): markdown/docs + config lint gate for committed artifacts — parent canonical pattern (Refs #326)

parametrization:
## What

`Refs #326` (Phase-3 end-state criterion #5 — all committed artifacts pass all CI checks). **#326 stays OPEN** as the rollout tracker: this PR lands the **parent-canonical pattern** and the per-repo gap map; extending coverage to the 7 child repos is the **W14 fast-follow**.

`ci.yml` already gates the python artifacts (`.claude/hooks/` + `.claude/lib/` via ruff/mypy/the test suite/compile smoke). The gap was everything **non-python**: markdown/docs had no lint/spellcheck/link-check, and `.claude/team/**` + `ontology/**` were outside the CI path triggers entirely. This PR closes that gap **for the parent** and is built as a clean, copyable job set children can adopt.

## New workflow — `.github/workflows/docs.yml` (CI — Docs & Config)

Designed as the canonical pattern. Each job is self-contained with its config in a repo-root dotfile, so a child repo adopts it by copying the workflow + the four config files and adjusting the ignore globs.

| Job | Tool (pinned) | Config | Scope |
|---|---|---|---|
| Markdown lint | markdownlint-cli2 v23.2.0 | `.markdownlint-cli2.jsonc` | every committed parent `.md` |
| Spellcheck | cspell v8.4.0, `strict` | `.cspell.json` + `.cspell/project-words.txt` | authored prose (charter/skills/ontology/docs) |
| Internal link-check | lychee v2.8.0, offline | `.lychee.toml` | relative cross-refs in `.md` |
| Config YAML/JSON syntax | python (PyYAML+json) | inline | parent tree, child repos excluded |
| Config actionlint | actionlint (shellcheck-integrated) | — | `.github/workflows/` |

Path coverage extended to `**/*.md`, `**/*.yaml`, `**/*.yml`, `ontology/**`, `.claude/team/**`, plus the config files.

Per #559's ownership matrix (#328): every artifact class gated here is **parent-canonical**, so it is correctly gated at the parent. Children own their own `CLAUDE.md`/repo docs and will gate them in their own CI (W14).

## CI status — all 5 jobs green

[...]

PR #562, opened 2026-05-31
feat(ci): pre-push stage + pre-commit<->CI sync-drift gate (parent) (Refs #327)

parametrization:
## feat(ci): pre-push stage + pre-commit ⇄ CI sync-drift gate (parent)

**Phase-3 end-state criterion #6** (`#327`), pulled into W13. Lands the **parent-canonical** piece + the reusable gate. The 7-child rollout is a separate child-implementer fan-out (the team lead coordinates it from this template), so this is **Refs #327**, not Closes — #327 stays open until all 8 repos are done, honest to its "each of 8 repos" DoD.

### Parent `.pre-commit-config.yaml`
- pre-commit stage: `ruff-format` + `ruff-lint` over `.claude/hooks/` **and** `.claude/lib/` (was hooks-only; CI lints both).
- **NEW pre-push stage**: `mypy` typecheck + the test suite — the heavier checks, run before code leaves the machine.

### Sync-drift gate (new, reusable — the high-value piece)
`.claude/lib/pre_commit_ci_sync.py` normalizes BOTH the pre-commit config and the CI workflows to canonical check-kinds (ruff-lint, ruff-format, mypy, test-suite, eslint, typescript, prettier, terraform-fmt, gitleaks, actionlint, pip-audit, build) and **fails when CI enforces a kind the pre-commit config does not run locally** (the harmful drift direction; stricter-local is informational). Wired as a CI job `Pre-commit ⇄ CI sync-drift gate`; `.pre-commit-config.yaml` added to CI path filters.

- 11 unit tests incl. a guard that the **real parent** config mirrors its CI; drift-fail path verified (exit 1 naming the missing kind).

`CLAUDE.md § Local Hooks` documents install (both hook types), the stage split, and the gate — the template each child repo copies.

### Per-repo toolchain map (starting point for the child-rollout fan-out)
Surveyed at `main` HEAD 2026-05-31. Each child implementer adds a pre-push stage for their toolchain's heavier checks + wires the sync-gate into their CI + per-repo setup docs:

| Repo | Toolchain / existing pre-commit hooks | Pre-push heavier checks to add |
|---|---|---|
| noorinalabs-main (this PR) | ruff-format, ruff-lint, mypy, test-suite | mypy + test-suite (done here) |
| noorinalabs-user-service | ruff-format, ruff (Python) | mypy + test-suite |
| noorinalabs-data-acquisition | ruff-format, ruff, gitleaks, mypy, unit-tests | mypy + test-suite |
[...]

PR #573, opened 2026-06-01
fix(hooks): auto_set_env_test skip pytest tokens in heredoc bodies (#564)

### LINKED ISSUE #564: hook(auto_set_env_test): over-matches pytest/test tokens in non-command contexts (PR-body PATCH, commit-msg heredoc) — 2 P3W13 instances
## Bug

The `auto_set_env_test` PreToolUse hook over-matches `pytest`/`test`-shaped tokens that appear in **non-command contexts**, attempting to inject `ENVIRONMENT=test` where it makes no sense. Two independent instances in P3W13 (2026-05-31):

1. **PR-body PATCH (Aino, #562):** the literal substring `pytest` inside a `gh api -X PATCH .../pulls/562 --input` body (PR prose describing the sync-drift gate's check kinds) tripped the hook — it treated the API-call body content as a test-bearing Bash segment. Worked around by rewording prose to "test-suite".
2. **Commit-message heredoc (Santiago, #563):** the word `pytest` inside a `git commit` heredoc message body tripped the hook — message *content*, not a command. Worked around by writing the message via `-F file` instead of inline heredoc.

In both cases the token is **data passed to a command** (an API request body / a commit message), not a test invocation the hook should act on.

## Root cause direction
The segment parser treats argument/heredoc/`--input`-body content as command text. It should distinguish a *test invocation* (`pytest ...`, `python -m pytest`, `npm test` as the executed program) from `pytest` appearing inside a quoted string, heredoc body, or `--input`/`--field`/`-F` value. Relates to the `#543` segment-parser test-coverage family (newline/heredoc/quoted-region classes) and `feedback_hook4_regex_prose_false_match` (the sibling Hook-4 prose-false-match).

## Repro
- `git commit -F-` heredoc containing "pytest" in prose → hook fires.
- `gh api -X PATCH ... --input <json with "pytest" in body>` → hook fires.

## Fix + test
Exclude heredoc bodies, quoted-string regions, and known data-passing flags (`--input`, `--field`, `--body-file`, `-F`) from test-invocation detection. Add cases to the `auto_set_env_test` suite for: heredoc-body-with-pytest (allow), quoted-pytest-arg (allow), real `pytest`/`python -m pytest`/`npm test` invocation (still inject). Per #543's mandatory segment-parser coverage convention.

parametrization:
## Summary
`auto_set_env_test` scanned every shell segment of a Bash command for a `pytest` / `make test` token. A heredoc **body** — a commit message fed via `git commit -F-`, or a `--input`/payload built with a heredoc — is command *input data*, not a program invocation, yet its lines were treated as command text. A test-runner token in that body falsely registered as a test segment and the hook blocked, demanding an `ENVIRONMENT=test` prefix that makes no sense for a message body.

Two P3W13 instances were worked around by hand:
- **#562** — `pytest` in a PR-body payload (worked around by rewording prose).
- **#563** — `pytest` in a `git commit` heredoc message (worked around with `-F file`).

(This very PR's commit hit the bug live: the parent session's *installed* hook is the pre-fix version, so the commit message had to be staged via a written file instead of an inline heredoc.)

## Fix
Add `_strip_heredoc_bodies()` — the heredoc-body sibling of `_strip_quoted_regions`. It blanks heredoc body characters to spaces while **preserving every newline and the total length**, so the blanked "detection view" splits into segments whose boundaries are identical to the original command. Detection (`_is_test_segment`) and control-flow bracketing run on the **view**; env-block checks and the splice rewrite run on the **original** text (where a real `ENVIRONMENT=test`, if any, lives). A real test invocation *outside* any heredoc still lands in its own segment and blocks with a correctly-targeted suggestion.

**Scope:** quoted (`<<'EOF'`) and unquoted (`<<EOF`) delimiters, the `<<-` indent-strip closer, multiple/non-terminal heredocs, and an unclosed heredoc (blanks to EOS). The here-string form `<<<` is explicitly **not** a body heredoc (negative lookahead) — its operand is same-line data handled by ordinary quote scanning.

## Tests
+11 tests:
- `HeredocBodyTests` (6) — negative-match: commit-msg / `--input` / `<<-` heredoc bodies mentioning `pytest`/`make test` are allowed; a real `pytest` co-occurring with a heredoc still blocks with the splice on the real token only.
- `HeredocBodyParserTests` (5) — pure-parser: length/newline-preservation, quoted+unquoted+`<<-` delimiters, `<<<` left untouched, real command after heredoc survives.

All 6 mandated segment-class markers (`hooks.md § 5a`) remain present (the manifest-guard test passes). Full suite **91 passed** (was 80). `ruff format --check`, `ruff check`, and `mypy` all clean.

## Related Issues
Closes #564

## Review Checklist
- [ ] `_strip_heredoc_bodies` preserves length + newline positions (positional mirror invariant)
- [ ] real test invocation outside a heredoc still blocks with correctly-targeted splice
- [ ] `<<<` here-string is not mistaken for a body heredoc
- [ ] all 6 segment-class markers present (per `hooks.md § Mandatory Test Coverage for PreToolUse Segment Parsers`)

### DIFF
--- a/.claude/hooks/auto_set_env_test.py
+++ b/.claude/hooks/auto_set_env_test.py
@@ -122,6 +122,17 @@ _BODY_FLAG = re.compile(r"(?<![\w-])--body(?:-file)?(?=[\s=]|$)")
 # Test-runner detection: pytest OR `make test`.
 _TEST_RUNNER = re.compile(r"\bpytest\b|\bmake\s+test\b")
 
+# Heredoc operator: `<<` (optionally `<<-`) followed by an optionally
+# quoted delimiter word. The delimiter is captured so we can find its
+# terminating line. Examples matched:
+#     <<EOF        <<-EOF        << EOF
+#     <<'EOF'      <<"EOF"       <<-'EOF'
+# We do NOT match `<<<` (here-string) — that has no body lines, its
+# operand is on the same line and is handled by ordinary quote/word
+# scanning. The negative lookahead `(?!<)` after the second `<` excludes
+# the here-string form.
+_HEREDOC_OPEN = re.compile(r"<<(?!<)-?\s*(['\"]?)([A-Za-z_][A-Za-z0-9_]*)\1")
+
 # The literal env-var we require.
 _ENV_TOKEN = re.compile(r"\bENVIRONMENT=test\b")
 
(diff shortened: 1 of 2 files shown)


parametrization:
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Release-coordinator review (CI / merge-safety lens). Behavioral hook fix for the #564 over-match: a test-runner token inside a heredoc body is command input data, not an invocation, and must not trigger the env-prefix splice.

**Checks**: all 6 green on statusCheckRollup, including the sync-drift gate and the pytest + smoke suites. Branch MERGEABLE against wave-14.

**Design soundness**: the central decision is to detect on a heredoc-body-blanked view that preserves length and newline positions, while rewriting against the original segment text. Because the view is a positional mirror, `_split_segments` yields identical boundaries, so `parts_view[i]` and `parts[i]` cover the same byte range — detection masks the body, the splice still lands on the real token. The env-block check and the actual rewrite correctly run against the original, where a real ENVIRONMENT=test prefix would live.

**Edge coverage I spot-checked**: here-string (`<<<`) excluded via the negative lookahead; `<<-` indent-strip closer accepted with leading tabs; a real runner after the closing delimiter survives masking and still blocks with the splice on the real token; and the multi-heredoc-on-one-line path uses the last delimiter, which over-masks conservatively (more masking, never less) — safe for a false-positive-avoidance hook. Tests pin all of these plus the pure-parser positional-mirror invariant.

Faithful, well-bounded fix. Approve.

parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

First-reviewer pass on the heredoc-body over-match fix. Read-only against the PR head; no parent checkout. Two scoped files touched, the hook and its suite, nothing unrelated.

The design I needed to satisfy myself on is that blanking heredoc bodies cannot break the splice path for a real runner that lives outside any heredoc. It holds. The blanking is length- and newline-preserving, so the blanked view and the original split into segments with identical boundaries; detection runs against the view, while the env-block check and the actual rewrite operate on the original segment text. I exercised that path directly against the module at head:

- A real runner after a heredoc closer, where the heredoc body also mentions a runner token, blocks and splices onto the real outside token only; the body text is never rewritten.
- A pure-data heredoc whose body mentions a runner token allows, as intended.
- Positional-mirror integrity verified across five shapes (multiple heredocs on one line, the dash indent-strip closer, the here-string triple-angle form which correctly has no body to blank, and an unclosed heredoc that blanks to end-of-input): every case preserved total length and exact newline offsets.
- A real runner whose own config is piped in via a heredoc still blocks.

The full suite runs green at head with the count claimed, and CI is green across lint, format, mypy, smoke, the pytest job, and the sync-drift gate.

One conservative-by-design note, not a defect: when a single line opens more than one heredoc, the strip keys off the last operator and blanks through that delimiter's closer, which over-blanks rather than under-blanks. That can only make detection more conservative, never produce a false allow of an outside runner, and the docstring states the conservative intent. No change requested.

Worth recording for amusement and for the rollout: this very hook fired on my own probe commands, masking runner tokens inside my inline test payloads, which is exactly the over-match class the fix narrows. It behaved correctly throughout. Approving as first reviewer.

PR #575, opened 2026-06-01
fix(hooks): validate_pr_review hard-blocks batch-loop variable merges (#567)

### LINKED ISSUE #567: hook(validate_pr_review): batch-loop `gh pr merge $var` fails open — block variable-PR merges (DECIDE, from P3W13 retro #4)

parametrization:
## Summary
The 2-reviewer gate parses a **literal** PR number from `gh pr merge <N>` to fetch that PR's reviews. When the PR number is a shell **loop variable** —

```bash
for pr in 48 49 50; do gh pr merge "$pr" --repo ... --merge; done
```

— the literal parse returns `None`, `get_pr_data(None)` falls back to the cwd branch's PR, and the gate **fail-opens**: every iteration merges unverified. This fired twice (P3W11 wave→main 4-merge loop with 0 reviews; P3W13 ingest 6-PR loop), caught only post-hoc in #927.

## Fix (DECIDE-tier — conservative block only)
Add `is_variable_pr_merge_in_loop(command)` and **HARD BLOCK** in `check()` before the fail-open path. Per the issue and `feedback_safety_direction_over_ux_friction`, this is the conservative block — **no** conditional-allow / loop-value-enumeration. The operator is told to run one literal merge per call so each PR is gate-checked individually.

Detection requires **both**: a `gh pr merge` with a shell-variable arg (`$pr` / `${pr}` / `"$pr"` / `"${pr}"`) **and** a loop construct (`for`/`while`/`until` opener + a `do` body keyword).

**Placement** (important): the guard runs **before** the `is_merge_command` early-return on purpose — `is_merge_command` checks each `;`-split segment's *first* token for `gh pr merge`, but in a loop the merge segment leads with the `do` keyword, so it returns False and `check()` would early-exit, re-opening the exact hole. `--admin` still overrides (checked first); a literal `gh pr merge 54` is untouched (its number parses, the normal gate runs).

**Quote-awareness**: `_strip_quoted_runs_keep_var_args` removes quoted prose so a `--body "... do gh pr merge $pr ... done"` payload that merely *mentions* the shape is NOT blocked, while unwrapping a lone `"$pr"` arg to its bare form so the real quoted-var merge IS detected (mirrors shell: quotes around a lone variable are removed, the variable stays an arg).

## Tests (+19)
- **`BatchLoopMergeDetectorTests`** — the 6 parser classes per `hooks.md § 5a` (newline, quoted-var, `${}`-form, literal-still-passes, while-loop, nested-loop) + until-loop, literal-loop, var-outside-loop, **body-mention false-positive**, non-merge-loop.
- **`BatchLoopMergeEndToEndTests`** — `check()` hard-blocks the loop-var merge; `--admin` overrides; a literal 2-approved merge still passes the gate (guard doesn't interfere).
- **`StripQuotedRunsKeepVarArgsTests`** (4) — pure-parser: lone-var unwrap, brace-var unwrap, prose-run drop, single-quoted-run drop.

Full `validate_pr_review` suite **122 passed** (was 103). `ruff format --check`, `ruff check`, `mypy` clean.

> Note: one unrelated test (`test_ontology_tracker.py::...worktrees_segment_not_substring_false_match`) fails **only** when run from inside a `.claude/worktrees/` checkout (its `REPO_ROOT` path itself contains `worktrees`). It passes from the canonical checkout and in CI; it does not touch any file in this PR.

## Related Issues
Closes #567

## Review Checklist
- [ ] guard runs before `is_merge_command` early-return (loop merge segment leads with `do`)
- [ ] `--admin` still overrides; literal `gh pr merge 54` unaffected
- [ ] body-mention (`gh pr create --body "...loop..."`) is NOT a false positive
- [ ] real quoted-var `"$pr"` IS detected; conservative block only (no conditional-allow)
- [ ] all 6 parser classes present per `hooks.md § Mandatory Test Coverage for PreToolUse Segment Parsers`

### DIFF
--- a/.claude/hooks/tests/test_validate_pr_review.py
+++ b/.claude/hooks/tests/test_validate_pr_review.py
@@ -1521,5 +1521,166 @@ class ResolveRosterDirsTests(unittest.TestCase):
             hook._resolve_roster_dirs("noorinalabs/noorinalabs-nonexistent-xyz")
 
 
+class BatchLoopMergeDetectorTests(unittest.TestCase):
+    """#567: `is_variable_pr_merge_in_loop` detects a `gh pr merge <var>` inside
+    a for/while/until loop — the shape that fail-opens the 2-reviewer gate
+    (memory `feedback_batch_loop_merge_evades_pr_review_hook`).
+
+    Six parser classes per charter `hooks.md § 5a` segment-parser coverage:
+    newline-separated loop, quoted-var arg, `${}`-form, literal-still-passes,
+    while-loop, nested-loop. Plus the body-mention false-positive guard."""
+
+    def test_for_loop_quoted_var_blocked(self):
+        # parser-class: quoted-var
+        self.assertTrue(
+            hook.is_variable_pr_merge_in_loop(
+                'for pr in 48 49 50; do gh pr merge "$pr" --repo o/r --merge; done'
+            )
+        )
+
+    def test_for_loop_bare_var_blocked(self):
+        self.assertTrue(
+            hook.is_variable_pr_merge_in_loop("for pr in 48 49; do gh pr merge $pr --merge; done")
+        )
+
+    def test_brace_form_var_blocked(self):
+        # parser-class: ${}-form
+        self.assertTrue(
+            hook.is_variable_pr_merge_in_loop('for pr in 1 2; do gh pr merge "${pr}" --merge; done')
+        )
+
+    def test_while_loop_blocked(self):
+        # parser-class: while-loop
+        self.assertTrue(
+            hook.is_variable_pr_merge_in_loop("while read pr; do gh pr merge ${pr} --merge; done")
+        )
+
+    def test_until_loop_blocked(self):
+        self.assertTrue(
+            hook.is_variable_pr_merge_in_loop('until [ -z "$pr" ]; do gh pr merge $pr; done')
+        )
+
+    def test_newline_separated_loop_blocked(self):
+        # parser-class: newline
+        self.assertTrue(
+            hook.is_variable_pr_merge_in_loop(
+                'for pr in 48 49\ndo\n  gh pr merge "$pr" --merge\ndone'
+            )
+        )
+
+    def test_nested_loop_blocked(self):
+        # parser-class: nested-loop
+        self.assertTrue(
+            hook.is_variable_pr_merge_in_loop(
+                'for r in a b; do\n  for pr in 1 2; do gh pr merge "$pr" --merge; done\ndone'
+            )
+        )
+
+    def test_literal_merge_not_blocked(self):
+        # parser-class: literal-still-passes — a literal PR number is untouched
+        # by the guard (its number parses and the normal gate runs).
+        self.assertFalse(hook.is_variable_pr_merge_in_loop("gh pr merge 54 --repo o/r --merge"))
+
+    def test_literal_loop_not_blocked(self):
+        # A loop iterating LITERAL merges (no shell variable) is not the
+        # fail-open shape — each merge parses a literal number.
+        self.assertFalse(
+            hook.is_variable_pr_merge_in_loop("for n in 1; do gh pr merge 54 --merge; done")
+        )
+
+    def test_var_merge_outside_loop_not_blocked(self):
+        # Out of scope for #567: a one-off variable merge with no loop.
+        self.assertFalse(hook.is_variable_pr_merge_in_loop('gh pr merge "$PR" --merge'))
+
+    def test_loop_text_in_body_not_blocked(self):
+        # False-positive guard: a `--body` payload that merely MENTIONS the
+        # loop-merge shape (quoted prose) must NOT be detected. The real
+        # command is `gh pr create`, not a merge.
+        self.assertFalse(
+            hook.is_variable_pr_merge_in_loop(
+                'gh pr create --body "for pr in 1; do gh pr merge $pr; done"'
+            )
+        )
+
+    def test_non_merge_loop_not_blocked(self):
+        # A loop that runs some other gh command is irrelevant to the gate.
+        self.assertFalse(
+            hook.is_variable_pr_merge_in_loop('for pr in 1 2; do gh pr view "$pr"; done')
+        )
+
+
+class BatchLoopMergeEndToEndTests(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


parametrization:
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Release-coordinator review (merge-safety lens). DECIDE-tier hard-block for the batch-loop merge fail-open (#567) — a `gh pr merge <shell-var>` inside a for/while/until loop, the shape that twice silently disabled the 2-reviewer gate (P3W11 wave to main, P3W13 ingest). Closing this is squarely in my lane.

**Checks**: all 6 green on statusCheckRollup, including the sync-drift gate and the pytest + smoke suites. Branch MERGEABLE against wave-14. (Noted the worktree-path-sensitive test the author flagged — it fails only from inside a worktrees checkout and is green in CI; not a #575 regression.)

**Three merge-safety invariants I confirmed by tracing control flow:**
- **Admin override intact**: the admin short-circuit is relocated to the TOP of the entry function, ahead of the new guard, so an operator's emergency bypass is honored even for the loop shape. The end-to-end test asserts the loop-plus-admin case returns no block.
- **Literal single merge unaffected**: the detector requires a shell-variable argument, so a literal PR number never matches; flow falls straight through to the normal gate. Two tests pin this, one of them a full two-approved pass-through.
- **Block scoped to variable-in-loop only**: detection requires BOTH a variable merge arg AND a loop opener paired with a body keyword. A one-off variable merge outside a loop and a prose mention inside a create payload both return clean.

**Placement reasoning verified**: the guard intentionally runs before the merge-command early-return, because in a loop the merge segment leads with the body keyword and the segment-first-token check would otherwise early-exit and re-open the exact hole. The quote-normalizing helper correctly drops prose runs while unwrapping a lone quoted variable arg, so a payload that merely mentions the shape is not matched. Conservative no-conditional-allow design, fails in the safe direction.

Faithful close of a known gate-evasion. Approve.

parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

First-reviewer pass on the batch-loop variable-merge hard-block. Read-only against the PR head; no parent checkout. Two scoped files, the hook and its suite, nothing unrelated rides along.

Both design points carried in the brief check out against the module at head.

The first, that the guard must run ahead of the merge-command early-return, is the crux and it holds. I confirmed the loop-shaped command resolves the merge-command predicate to false, because that predicate inspects the first token of each separator-split segment and in a loop the merge segment leads with the body keyword rather than the gh verb. So if the guard sat after that early-return, the loop shape would slip straight through and re-open the exact fail-open hole. With the guard placed ahead of it, the loop command hard-blocks as intended even though the merge-command predicate is false. The emergency override is checked first of all, so an override-flagged loop command bypasses the guard, matching the existing short-circuit semantics.

The second, the quoted-payload discrimination, also holds. A create command carrying the loop-merge shape inside a quoted body argument is not detected, while a real merge whose only quoting is around the lone variable argument is detected. Inspecting the normalizer directly: the body payload is stripped to leave just the create invocation with the merge mention gone, whereas the lone quoted variable is unwrapped to its bare form so the real argument survives for the detector. Single-quoted prose is dropped wholesale as expected.

I ran an adversarial sweep beyond the suite. All five loop forms block: bare variable, brace form, while-read, until, newline-separated, and nested. All five benign forms pass: a literal-number merge, a loop iterating only literal merges, a one-off variable merge outside any loop, a non-merge loop, and a quoted echo of the loop shape. The full suite is green at head with the count claimed, and CI is green across lint, format, mypy, smoke, the test job, and the sync-drift gate.

The conservative posture is right for this tier. Detection requires both a variable-argument merge and a loop opener-plus-body pair, so an ambiguous case over-blocks rather than fail-opening, which is the safe direction. No attempt is made to enumerate loop values and gate each iteration, and that conditional-allow path is correctly left deferred. The block reason is actionable, naming the one-literal-merge-per-call remedy and noting each literal merge is individually gate-checked. No must-fix items. Approving as first reviewer.

Issue #578, opened 2026-06-01
tech-debt(#326): docs.yml actionlint installer unpinned (curl|bash main, no checksum) across all rolled-out repos


parametrization:
## Problem
The canonical docs-CI template's actionlint step installs via `bash <(curl -fsSL .../rhysd/actionlint/main/scripts/download-actionlint.bash)` — an **unpinned `main` ref, no checksum**, fetching the latest binary at run time. This shipped into every P3W14-rolled-out repo's `docs.yml` (landing, user-service, data-acq, ingest, design-system, deploy, isnad-graph).

By contrast, deploy's `lint-workflows.yml` pins actionlint to **1.7.12 + `sha256sum -c`**. The unpinned docs.yml installer is a supply-chain inconsistency (fetches+executes latest-of-main each run) and diverges from the pinned pre-commit `rhysd/actionlint v1.7.12` mirror the same template ships.

Surfaced by Nino Kavtaradze reviewing deploy#391 (P3W14). Not a merge blocker per-PR (contents:read perms, lints YAML only, no secret scope, redundant with pinned authoritative jobs where present), but it's a template-wide hardening gap.

## Fix
Pin the docs.yml actionlint install to **v1.7.12 + SHA256 verify** in the canonical template, then propagate to all 7 rolled-out repos' docs.yml (a small follow-up sweep). OR drop the docs.yml actionlint job in repos that already have a pinned authoritative actionlint gate (avoid redundancy).

Refs #326 #327. Template-wide hardening from the P3W14 rollout.

parametrization:
**Parent template step landed** — PR #579 merged to `deployments/phase-3/wave-14` (bdc0ec1). The parent `.github/workflows/docs.yml` actionlint install no longer `curl|bash`-es `download-actionlint.bash` off `main`; it now pins the **v1.7.12** release tag and `sha256sum -c`-verifies `8aca8db9…a3d8` (verified against upstream `actionlint_1.7.12_checksums.txt`) before executing. CI's actionlint job is green end-to-end with the pinned installer.

[...]

Issue #326, opened 2026-05-08
P3 end-state #5: All committed artifacts pass all CI checks (code, tests, .claude/, docs)


parametrization:
Tracking issue for **Phase 3 end-state criterion #5**.

See `.claude/team/phases/phase-3.md` § End-state criteria for full phase context.

## Criterion

**All committed artifacts pass all checks** — code, tests, `.claude/` files (skills/hooks/team/etc.), docs alike. Anything in the tree gets the same lint/format/syntax/spell/security treatment.

## Definition of done

- CI (GitHub Actions) covers ALL committed paths, not just `src/`
- `.claude/skills/**/SKILL.md` and `.claude/hooks/**/*.py` lint and format on every commit
- Documentation (`*.md`) lint/spellcheck/link-check
- Test files lint at the same rigor as source
- All 8 repos enforce this consistently

## Out of scope

- Generated files (lockfiles, build outputs) — explicitly excluded
- Vendored third-party (if any) — explicitly excluded

## Dependencies

- Sibling of #6 (hook parity must mirror this CI scope)

## Sub-tickets

(filed at scoping time)

parametrization:
Requestor: Fatima Okonkwo
Requestee: Aino Virtanen
RequestOrReplied: Request

**Wave 13 Kickoff — Phase 3**

This issue is assigned to you for p3-wave-13.
- Peer reviewer: Wanjiku Mwangi
- Secondary reviewer: Santiago Ferreira
- Branch from: `deployments/phase-3/wave-13`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: verify-and-close (per charter § Wave Planning & Priority)

Please begin implementation.

parametrization:
## Tier-5 verify-and-close audit — #326 (All committed artifacts pass all CI checks)

Audited the parent `noorinalabs-main` CI workflow at `main` HEAD (`.github/workflows/ci.yml`), 2026-05-31.

### Definition-of-done element status

| Element | Status |
|---|---|
| CI covers ALL committed paths, not just `src/` | **PARTIAL** — covers `.claude/hooks/`, `.claude/lib/`, `.claude/skills/`; not `.claude/team/`, `ontology/`, root docs |
[...]

PR #579, opened 2026-06-01
tech-debt(#326): pin docs.yml actionlint installer to v1.7.12 + sha256 (Refs #578)

parametrization:
## What

Pin the parent `.github/workflows/docs.yml` actionlint installer to **v1.7.12 + SHA256 verify**, replacing the unpinned `curl|bash` of `download-actionlint.bash` off `main`.

Refs #578 (parent-template step; #578 stays open to track the 7-repo docs.yml propagation sweep).

## Why

The `Config — Workflow actionlint` job installed actionlint via:

```
bash <(curl -fsSL https://raw.githubusercontent.com/rhysd/actionlint/main/scripts/download-actionlint.bash)
./actionlint -color
```

— an unpinned `main` ref with **no checksum**, fetching and executing the latest binary at run time. This is a supply-chain gap and diverges from the pinned `rhysd/actionlint v1.7.12` the `.pre-commit-config.yaml` mirror already ships (#571).

## How

Mirrors deploy's `lint-workflows.yml` canonical pinned+verified pattern: download the `v1.7.12` linux_amd64 tarball, `sha256sum -c` against the published hash (aborts on mismatch), install to `/usr/local/bin`, then lint. The shellcheck-on-ubuntu-latest note is preserved so actionlint's embedded shellcheck integration still runs.

The SHA256 `8aca8db9...a3d8` was **verified against upstream's published `actionlint_1.7.12_checksums.txt`** release manifest, not copied on trust.

## Scope

Parent docs.yml only (the canonical template). The 7-repo `docs.yml` propagation sweep is a separate follow-up tracked in #578.

## Validation

- `actionlint .github/workflows/docs.yml` → clean (shellcheck integration active locally; v1.7.7 local, CI uses pinned v1.7.12).
- YAML parses; `python3 .claude/lib/pre_commit_ci_sync.py . --ci ci.yml --ci docs.yml` → OK (no sync drift; the `actionlint` kind stays mirrored).
- Local pre-commit actionlint hook (v1.7.12 pin) passed on commit.

Refs #326 #327.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -159,11 +159,31 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - name: Download and verify actionlint
+        env:
+          # Pinned: version + linux_amd64 tarball SHA256 from upstream's
+          # signed release manifest (actionlint_1.7.12_checksums.txt at
+          # https://github.com/rhysd/actionlint/releases/tag/v1.7.12).
+          # `sha256sum -c` aborts the job on mismatch — bytes-on-runner are
+          # exactly what was published at release time. Replaces the prior
+          # `curl|bash` of the unpinned download-actionlint.bash off `main`
+          # (#578): fetching+executing latest-of-main each run is a
+          # supply-chain gap and diverges from the pinned v1.7.12 the
+          # .pre-commit-config.yaml mirror already ships (#571).
+          ACTIONLINT_VERSION: 1.7.12
+          ACTIONLINT_SHA256: 8aca8db96f1b94770f1b0d72b6dddcb1ebb8123cb3712530b08cc387b349a3d8
+        run: |
+          cd "$RUNNER_TEMP"
+          curl --fail --silent --show-error --location \
+            -o "actionlint_${ACTIONLINT_VERSION}_linux_amd64.tar.gz" \
+            "https://github.com/rhysd/actionlint/releases/download/v${ACTIONLINT_VERSION}/actionlint_${ACTIONLINT_VERSION}_linux_amd64.tar.gz"
+          echo "${ACTIONLINT_SHA256}  actionlint_${ACTIONLINT_VERSION}_linux_amd64.tar.gz" | sha256sum -c -
+          tar -xzf "actionlint_${ACTIONLINT_VERSION}_linux_amd64.tar.gz" actionlint
+          install -m 0755 actionlint /usr/local/bin/actionlint
+          actionlint -version
       - name: Run actionlint on parent workflows
         # shellcheck is preinstalled on ubuntu-latest, so actionlint's embedded
         # shellcheck integration runs (per the actionlint-needs-shellcheck
         # discipline) rather than silently skipping.
-        run: |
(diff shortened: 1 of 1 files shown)


parametrization:
Requestor: Wanjiku Mwangi
Requestee: Santiago Ferreira
RequestOrReplied: Approved
TechDebt: none

First-reviewer pass on the actionlint installer pin. Read-only against the PR head; no parent checkout. Single scoped file, the docs workflow, nothing unrelated rides along.

The supply-chain hardening is correct and the pin is genuine. The installer no longer pipes the unpinned download script off the upstream default branch and executes it. It now fetches the linux amd64 release tarball from the tagged v1.7.12 release path — the download URL is anchored to the release tag, not to the moving branch — then verifies the tarball against a pinned checksum before unpacking. I independently fetched the upstream signed checksums manifest for that release and the linux amd64 line matches the value pinned in the workflow byte for byte, so the verification step is checking against the real published hash rather than a transcribed-from-nowhere value.

The verify-then-use ordering is sound. The checksum check runs before the unpack and install, and under the default shell behavior of a workflow run block a failed check aborts the job, so a tampered or substituted tarball never reaches execution. The bytes that land on the runner are exactly what upstream published at release time.

The lint invocation itself is unchanged in scope. The binary installs to a standard path location and the lint step runs the same default invocation over the same workflow set as before, with no path narrowing, so coverage is preserved. The embedded shellcheck behavior is intact — the explanatory note about the preinstalled shellcheck on the runner image is retained, and that is what lets the embedded integration run rather than silently skip. The version now matches the pin the pre-commit mirror already ships, closing the local-versus-CI divergence.

The pinned installer ran end to end in CI — the workflow actionlint job is green at head, which exercises the fetch, the checksum verification, the install, and the lint run as a unit. No must-fix items. Approving as first reviewer.

parametrization:
Requestor: Aino Virtanen
Requestee: Santiago Ferreira
RequestOrReplied: Approved
TechDebt: none

Approved — standards and supply-chain lens, reviewed at head dfa37b7.

The installer no longer pipes a curl of the unpinned download-actionlint.bash off the main branch into bash. It now downloads the v1.7.12 release-tag tarball and runs `sha256sum -c` against the pinned digest before any extract, install, or execution — so the job aborts on a byte mismatch and only a verified binary ever runs. That closes the fetch-and-execute-latest-of-main supply-chain gap #578 called out.

I independently verified the pinned digest end to end rather than trusting the transcribed value. The upstream signed manifest (actionlint_1.7.12_checksums.txt for the v1.7.12 release) lists that exact digest for the linux_amd64 tarball, and re-downloading the real tarball and hashing the bytes reproduces it — `sha256sum -c` returns OK on the genuine artifact. So the hash is the real upstream checksum, not a copy error.

The lint scope is unchanged (the second step still lints every parent workflow), the shellcheck-on-PATH discipline is preserved with ubuntu-latest's preinstalled shellcheck, and the pinned version matches the rhysd/actionlint v1.7.12 pre-commit mirror the template already ships, so the CI installer and the local hook stay on the same release with no drift. The Config — Workflow actionlint CI job is green on this PR, confirming the pinned-and-verified installer actually runs.

One non-blocking nicety for a future sweep: now that CI pins v1.7.12 explicitly, the pre-commit config's note describing the mirror as tracking the latest release off main is slightly stale wording — a comment-only tidy in a different file, out of scope here and not a defect in this change.

Ship it.

Issue #577, opened 2026-06-01
tech-debt(#327): pre_commit_ci_sync kinds_from_ci misses tools on multi-line `run: |` continuation lines


parametrization:
## Problem
`.claude/lib/pre_commit_ci_sync.py`'s `kinds_from_ci` only inspects workflow lines that contain `run:`/`uses:` or start with `-`. A tool invoked on a **continuation line of a multi-line `run: |` block** is invisible to the classifier.

Concrete: `security-audit`'s `uv run pip-audit` (a continuation line inside `run: |`) is never seen as a CI kind, so the sync-gate never requires it be mirrored in pre-commit. Harmless today (pip-audit is intentionally not a pre-commit hook + is gated server-side as a required context), but the blind spot is **accidental**: a future repo expressing `ruff`/`mypy` via a multi-line `run: |` could silently drop a real gate from the mirror check.

Surfaced independently by Kalinda Ranasinghe + Petra Vidović reviewing ingest-platform#58 (P3W14). Present identically in the parent-canonical gate AND the W13 data-acq pilot, so it's a parent-helper gap, not a per-repo rollout defect.

## Fix
Teach `kinds_from_ci` to scan continuation lines inside `run: |` / `run: >` blocks (respect block indentation). Add a regression fixture (multi-line run: with ruff/mypy/pytest on continuation lines → classified).

Refs #327. Pairs with #576 (the build-kind + trigger-aware canonical gate work) — both are `pre_commit_ci_sync.py` correctness fixes; consider bundling.

parametrization:
**Fixed** — PR #580 merged to `deployments/phase-3/wave-14` (e9f5823). `kinds_from_ci` now scans `run: |` / `run: >` (incl. `|-` / `>-`) block-scalar continuation lines by indentation until a dedent closes the block, so a tool invoked on a continuation line (e.g. `uv run pip-audit`, or a future multi-line `ruff`/`mypy`/`pytest`) is no longer invisible to the mirror check. Regression tests cover the pipe block, folded block, multi-tool block, dedent-bleed guard, and single-line non-regression. Closes at wave→main wrapup (wave-branch merges don't auto-close).

Issue #576, opened 2026-06-01
tech-debt(#327): sync-gate false-classifies docker-build runtime steps as build-quality kind — canonicalize Aisha's deploy tightening


parametrization:
## Problem
The canonical `.claude/lib/pre_commit_ci_sync.py` sync-drift gate classifies `docker buildx build` / "Set up Docker Buildx" workflow step-names as a `build` CI-kind. That's **runtime image-moving, not a build-quality gate** a local pre-commit hook can mirror — so any repo with docker-image workflows gets a permanent, un-satisfiable `build` DRIFT and the gate never exits 0.

Surfaced in P3W14 by Aisha Idrissi (deploy rollout): deploy has 4 runtime workflows (cold-rebuild, integration-tests, promote, verify-deploy) with docker buildx steps. isnad-graph (GHCR frontend publish) and isnad-ingest-platform (worker images) are also at risk.

## Fix (canonical)
Tighten the gate's single `build` classification pattern so bare `docker build` / `docker buildx` runtime lines do NOT match, while KEEPING real build-quality detection (`build-and-validate`, `build-and-test`, `npm run build`, etc.). Aisha is producing a clean, portable form of this tightening in her deploy PR — lift it verbatim into the parent canonical `.claude/lib/pre_commit_ci_sync.py`, then re-vendor to every child repo that carries a copy so all gates converge (no per-repo divergence).

## Acceptance
- Parent `pre_commit_ci_sync.py` no longer false-classifies runtime docker build steps.
- Real build-gate kinds still detected (regression test).
- Child-repo vendored copies re-synced to the canonical version.

Refs #327. Surfaced from P3W14 batch-2 rollout (deploy).

parametrization:
## Refinement (from batch-2 evidence): trigger-aware is the preferred canonical fix

Batch-2 surfaced that the `build`-kind false-match has a clean root cause and a better fix than pattern-tightening:

**Root cause (Fatima's analysis):** the false-match fires only when a **docker image build/push is a CI *workflow step*** (publish/release/container-scan), not when image builds live in a Makefile/dev-loop. Affected repos = those with a CI publish workflow: **deploy** (Aisha), **user-service** (Anya, ghcr-publish.yml), **isnad-graph** (likely, GHCR frontend publish). NOT ingest (Makefile builds), data-acq (no docker CI), design-system (npm build is a real gate).

**Two candidate canonical fixes:**
[...]

PR #580, opened 2026-06-01
tech-debt(#327): sync-gate build-kind + multi-line run scanning (Refs #576 #577)

parametrization:
## Summary
Two correctness fixes to the canonical sync-drift gate `.claude/lib/pre_commit_ci_sync.py`, bundled per #577's "consider bundling with #576" note (both touch the same file).

### #576 — `build` kind no longer false-matches runtime docker steps
The `build` CI-kind pattern included `"docker build"`. Because a bare `docker build` substring also matches `docker buildx`, any workflow with `docker build` / `docker buildx` classified as a build-**quality** gate. That's runtime image-**moving** (retag / promote / publish / cold-rebuild / digest resolution), not a quality gate a local pre-commit hook can mirror — so every docker-image repo got a **permanent un-mirrorable `build` drift** and the gate could never exit 0 (surfaced by deploy's 4 runtime workflows; isnad-graph GHCR publish + ingest-platform worker images at the same risk).

Fix: drop `"docker build"` from the `build` tuple, keeping the real build-quality markers (`build-and-validate`, `build-and-test`, `npm run build`). **Lifted verbatim from the deploy-rollout form (deploy#391, A.Idrissi)** and canonicalized in the parent so vendored child copies converge. A genuine docker-build-as-quality-gate should name its job `build-and-validate`/`-test` (or add an explicit pattern).

### #577 — `kinds_from_ci` scans multi-line `run:` block bodies
The classifier only inspected lines containing `run:`/`uses:` or starting with `-`. A tool invoked on a **continuation line** of a `run: |` / `run: >` block (e.g. `security-audit`'s `uv run pip-audit`) was invisible — a silent blind spot where a future `ruff`/`mypy` expressed multi-line would silently drop out of the mirror check.

Fix: on a `run: |` / `run: >` opener (incl. `|-` / `>-` / digit indicators), record the key indent and classify each **more-indented** body line until a dedent to ≤ that indent closes the block (YAML block-scalar scoping). Commented body lines are skipped.

## Tests (+10)
- **`BuildKindTightening`** — `docker buildx` / bare `docker build` NOT classified as `build`; real build-quality gates still detected; a publish-only (docker buildx) workflow produces no `build` harmful drift.
- **`MultiLineRunBlockScanning`** — pipe-block `pip-audit`; multi-tool block (ruff/mypy/pytest); folded `>-`; dedent-closes-block scoping; single-line `run:` regression; commented body line not classified.

Full suite **21 passed** (was 11). Parent self-gate `python3 .claude/lib/pre_commit_ci_sync.py .` still exits 0. `ruff format --check`, `ruff check`, `mypy` all clean.

## Related Issues
Refs #576
Refs #577

(Using `Refs`, not `Closes`: this PR targets the wave-14 branch, and `Closes` only fires on default-branch merges — the issues are closed explicitly at wave wrapup.)

> **Scope note:** #576's child re-vendoring (its acceptance criterion 3) is per-child PR work on each child's wave branch — deploy already carries the form (deploy#391); other image-publishing repos (isnad-graph, ingest-platform) re-vendor separately. Out of scope for this parent-canonical PR.

## Review Checklist
- [ ] `docker build`/`buildx` runtime steps no longer register `build`; real quality gates still do
- [ ] multi-line `run:` block body lines are classified; block scoping ends at the right dedent
- [ ] single-line `run:` classification unchanged (no regression)
- [ ] parent self-gate still exits 0

### DIFF
--- a/.claude/lib/pre_commit_ci_sync.py
+++ b/.claude/lib/pre_commit_ci_sync.py
@@ -51,6 +51,22 @@ from pathlib import Path
 # Each kind maps to the keyword patterns that identify it on EITHER side
 # (pre-commit id/entry text or CI run/uses text). Patterns are substrings
 # matched case-insensitively against the relevant lines.
+#
+# `build` matches only real build-QUALITY GATES — a job whose purpose is to
+# fail the PR if the project does not build/compile (`build-and-validate`,
+# `build-and-test`, `npm run build`). It deliberately does NOT match bare
+# `docker build` / `docker buildx`: those are runtime image-MOVING (retag,
+# promote, publish to a registry, cold-rebuild dry-runs, digest resolution),
+# which is the deploy/publish job itself, not a quality gate a local
+# pre-commit hook could mirror. A bare `docker build` substring also matches
+# `docker buildx`, so a repo that uses buildx at runtime (deploy, the
+# image-publishing CI in isnad-graph / ingest-platform) would otherwise see a
+# permanent un-mirrorable `build` kind and the drift gate could never exit 0
+# (#576). If a repo ever adds a genuine docker-build-as-quality-gate, name
+# that job `build-and-validate` / `build-and-test` (or add an explicit
+# pattern) so it is mirror-tracked. Tightening lifted verbatim from the
+# deploy-rollout form (deploy#391, A.Idrissi) and canonicalized here so all
+# vendored child copies converge.
 _KIND_PATTERNS: dict[str, tuple[str, ...]] = {
     "ruff-format": ("ruff-format", "ruff format"),
     "ruff-lint": ("ruff check", "id: ruff", "- ruff"),
@@ -63,7 +79,7 @@ _KIND_PATTERNS: dict[str, tuple[str, ...]] = {
     "gitleaks": ("gitleaks",),
     "actionlint": ("actionlint",),
     "pip-audit": ("pip-audit", "pip audit"),
-    "build": ("build-and-validate", "build-and-test", "npm run build", "docker build"),
+    "build": ("build-and-validate", "build-and-test", "npm run build"),
 }
 
 # `ruff-lint` is a substring of nothing problematic, but `ruff format` also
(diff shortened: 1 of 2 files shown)


parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

First-reviewer pass on the sync-gate build-kind tightening and multi-line run-block scanning. Read-only against the PR head; no parent checkout. Two scoped files, the library module and its suite, nothing unrelated rides along.

Both behavior changes carry, verified directly against the module at head.

The build-kind tightening does exactly what is intended. A runtime buildx step and a bare docker-build step both classify to no build kind now, while the three real build-quality markers — the validate gate, the test gate, and the npm build script — still classify as build. The reasoning holds: a bare docker-build substring also swallows the buildx form, so any image-publishing repo would otherwise carry a permanent un-mirrorable build kind and never clear the gate, which is the regression the change removes. I confirmed the end-to-end shape too: a publish-only workflow whose only build-ish step is a runtime buildx, checked against a config that mirrors no build kind, yields no harmful drift.

The multi-line block scanning closes the continuation-line blind spot. Tools invoked only inside a piped run block are now classified — a security-audit block invoking the audit tool registers, and a single block carrying the lint, type, and test runners registers all three. The folded block form with a chomping indicator opens a block and is scanned as well. The dedent scoping is correct: an in-block lint invocation is caught while a following less-indented sibling step and its nested mapping body do not leak any spurious kind, so the classified set is exactly the one real kind. A commented-out continuation line is not classified, and the common single-line run form still classifies, so there is no regression to the existing path. The format-versus-lint discrimination is also intact: a format invocation yields only the format kind and a check invocation yields only the lint kind.

The full suite is green at head with the count claimed including the subtests, the gate self-check against this repository exits clean, and CI is green across all six gates including the sync-drift gate itself. No must-fix items. Approving as first reviewer.

parametrization:
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Reviewer #2 (release-coordinator / merge-safety lens), read-only at head 1f211322fea7. Two fixes to the sync-drift gate, both substantively verified.

**Build-kind tightening (#576)** — dropping the docker substring from the build tuple is correct. A bare image-moving step (buildx publish, retag, promote) is the deploy/publish job itself, not a quality gate a local pre-commit hook can mirror; leaving it in meant any image-publishing repo carried a permanent un-mirrorable kind and the gate could never exit clean. The genuine build-quality markers are retained, so a real compile gate still registers. The end-to-end publish-workflow case proves a buildx workflow against a config with no build mirror yields no harmful drift. This is the canonical form lifted from the deploy rollout, which I reviewed there — same shape, now centralized so the vendored child copies converge.

**Multi-line run scanning (#577)** — the block-scalar body scan closes a real blind spot: a tool invoked only on a continuation line under a piped run step was previously invisible to the classifier. I traced the control flow and confirm the scoping is right — the opener pattern requires the pipe/fold indicator to be the sole payload so an inline run step is untouched; the body scan classifies more-indented lines, and a dedent to or below the key indent closes the block and falls through so the closing sibling line is re-handled normally (no leak); interior blank lines keep the block open per yaml semantics; a commented continuation line is skipped; folded and chomped openers are handled; the common single-line form is unchanged.

I fetched both files at head into an isolated dir and ran the suite — all ten new tests pass. The full run shows one failure, the real-parent-config mirror check, and that is purely because my isolated copy did not include the config file. I confirmed the config exists at head, and the CI test job plus the sync gate itself are both green, so that check passes in the real checkout. Six of six checks pass on the rollup and the branch is mergeable.

Clean, well-scoped, well-tested. Approve.

Issue #569, opened 2026-05-31
Phase 3 Wave 14 — Phase-3 End-State Rollout + Process-Hook Hardening


parametrization:
## Theme

**Phase 3 Wave 14 — Phase-3 End-State Rollout + Process-Hook Hardening + Org-Wide Tech-Debt Reduction.**

Owner-confirmed 2026-05-31. Take the three end-state criteria W13 delivered *parent-canonical* and roll them out to all 7 child repos; close the prod-gate; harden the four W13-surfaced process hooks; and drive the org-wide tech-debt ratio down. This is the wave that lets Phase 3 be honestly declared complete (Phase-4 is gated behind all 9 end-state criteria).

## Scope

Reconciled by `/wave-scope 3 14` on 2026-05-31 (after `/phase-review 3`). `/wave-retro` P3W13 carry-forward folded in. No memory must-includes filed for W14. `#552` (child-roster hook) was DONE during the retro (PR #570) and is excluded from carry-forward.

### Tier 1 — Phase-3 end-state rollout (org-wide; the headline)
Each parent-canonical criterion rolls out to every child repo. Templates + gap-maps live in the W13 merged PRs (#561/#562/#563). Per-child implementation issues are created at `/wave-kickoff`.

- **#322** — branch-protection org-wide: apply the canonical ruleset spec to all 7 repos (W13 = spec+hook+pilot).
- **#326** — all committed artifacts pass all CI: wire the parent docs/config-lint pattern into the children.
- **#327** — pre-commit + pre-push mirroring CI in every repo: roll the parent sync-drift-gate template into the children. **Includes #571** (parent's own actionlint mirror is currently red on main — first instance).

Rollout targets (per-repo child work):
- noorinalabs-isnad-graph: branch-protection + docs-CI + pre-push mirror
- noorinalabs-user-service: branch-protection + docs-CI + pre-push mirror
- noorinalabs-deploy: branch-protection + docs-CI + pre-push mirror
- noorinalabs-design-system: branch-protection + docs-CI + pre-push mirror
- noorinalabs-data-acquisition: branch-protection (pilot done) + docs-CI + pre-push mirror
- noorinalabs-landing-page: branch-protection + docs-CI + pre-push mirror
- noorinalabs-isnad-ingest-platform: branch-protection + docs-CI + pre-push mirror

### Tier 2 — Prod-gate closure
[...]

Issue #322, opened 2026-05-08
P3 end-state #4: CI failures block all merges (branch protection org-wide)


parametrization:
Tracking issue for **Phase 3 end-state criterion #4**.

See `.claude/team/phases/phase-3.md` § End-state criteria for full phase context.

## Criterion

**CI failures block all merges.** No merging on red. Admin overrides only via the existing charter exceptions (wave-bootstrap, doc-sweep) — and those are logged.

## Definition of done

- Branch protection rules require all CI checks to be green before merge
- All 8 repos (noorinalabs-main + 7 child repos) have branch protection on `main`
- `--admin` merge usage is hook-blocked OR auditable + reviewed at retro time
- Charter exceptions are formally listed and validated at PR-merge hook time
- 0 admin overrides per wave is a measured indicator (already partially in place)

## Out of scope

- Required-reviewer count beyond charter (already covered by `validate_pr_review`)
- CODEOWNERS rollout (orthogonal)

## Dependencies

- Sibling of criterion #5 (CI scope expansion) and #6 (hook parity)
- Per session_2026_04_30 finding: 7 of 8 repos org-wide have NO branch protection — major prerequisite

## Sub-tickets

(filed at scoping time)

parametrization:
Requestor: Fatima Okonkwo
Requestee: Aino Virtanen
RequestOrReplied: Request

**Wave 13 Kickoff — Phase 3**

This issue is assigned to you for p3-wave-13.
- Peer reviewer: Wanjiku Mwangi
- Secondary reviewer: Santiago Ferreira
- Branch from: `deployments/phase-3/wave-13`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: verify-and-close (per charter § Wave Planning & Priority)

Please begin implementation.

parametrization:
## Tier-5 verify-and-close audit — #322 (CI failures block all merges / branch protection org-wide)

Audited live branch-protection on `main` across all 8 repos (`gh api repos/<repo>/branches/main/protection`), 2026-05-31.

### Definition-of-done element status

| Element | Status |
|---|---|
| Branch protection on `main` for all 8 repos | **NOT MET** — only **1 of 8** is protected |
| Required status checks green before merge | partial — only where protected |
| `--admin` merge hook-blocked OR auditable | partially in place (Hook 14 `validate_pr_ci_status` exists) but not the branch-protection backstop |
[...]

PR #583, opened 2026-06-01
process(P3W14 retro): adopt 3 owner-approved process changes (Refs #569)

parametrization:
## Summary
Adopts the **three owner-approved P3W14-retro Proposed Process Changes** in one PR (base `main`).

### 1. `wave-kickoff` MUST advance `current_wave` (retro change #1)
`.claude/skills/wave-kickoff/SKILL.md` Step 1a (the atomic `gh api PUT` status write): now **mandates** setting `current_wave="wave-{M}"` (plus `last_completed_wave` / `next_wave` / `wave_{M}_active`) with a read-back-verify. `validate_wave_audit` derives the current wave **label** from `current_wave`; W14 kicked off without advancing it (left at `wave-13`), which blocked the next `/wave-retro` — the one W14 annunaki capture. This codifies the missing kickoff step.

### 2. Red default-branch workflow detection (retro change #2)
`.claude/skills/session-start/SKILL.md` new **Step 5a**: per org repo, lists the latest default-branch run of each publish/deploy/release-class workflow and surfaces any `failure`/`timed_out`/`cancelled`. The GHCR frontend publish sat **RED on `main` for ~12 days undetected**, silently breaking every staging deploy at the frontend-pull step, because nothing surfaced a red default-branch publish at session start. Added a `5a` row to the output table.

### 3. End-state criterion = APPLIED-and-verified-at-origin, not just delivered (retro change #3)
`.claude/team/charter/issues.md` new section **"End-State Criterion: Delivered vs. Applied-and-Verified-at-Origin"**: a criterion/rollout issue is **MET only when the mechanism is applied AND verified at origin** via the authoritative API (rulesets endpoint returns the ruleset on **every** default branch; CI gate green in the latest default-branch run; a `deploy-stg` run exists) — not when the spec/script/hook is merely delivered. Delivered-but-not-applied uses `Refs` (not `Closes`) and the issue stays open as the rollout tracker until applied-verified for all targets. Generalizes `pull-requests.md § criterion #4`'s per-criterion note into a standing close-condition rule. Motivating gaps: #322 (specs delivered, ruleset unapplied) and the 12-day-red publish.

### Supporting
- `.cspell/project-words.txt`: add `ruleset` / `rulesets` — the new section trips the spell gate on these domain terms (also retroactively covers `pull-requests.md`'s pre-existing usages, which CI's cspell-action only re-scans when that file is in a PR diff).

## Validation
- **cspell 8.4.0** (CI-pinned): 0 issues on all changed files.
- **markdownlint-cli2** (CI config): 0 errors (80 files linted).
- New internal links are bare relative `](pull-requests.md)` to an existing same-directory file — identical to the lychee-passing convention already in `issues.md`; no URL fragments to resolve.
- No Python/workflow files touched, so ruff/mypy/actionlint surfaces are unaffected.

## Related Issues
Refs #569 (the P3W14 retro that proposed and owner-approved these three changes)

## Review Checklist
- [ ] change #1: kickoff Step 1a mandates `current_wave` advance + read-back, with the validate_wave_audit rationale
- [ ] change #2: session-start Step 5a surfaces red publish/deploy/release runs on default branches; output table updated
- [ ] change #3: charter section distinguishes delivered vs applied-and-verified-at-origin with per-mechanism probes + Refs-not-Closes disposition
- [ ] cspell/markdownlint clean; internal links resolve

### DIFF
--- a/.claude/skills/session-start/SKILL.md
+++ b/.claude/skills/session-start/SKILL.md
@@ -179,6 +179,43 @@ Report:
 
 If the report surfaces unexpected gaps between board view and open-issue counts (e.g., wave-labeled issues missing from project 2, or Wave-field values out of sync with `p{N}-wave-{M}` labels), invoke `/board-audit` to detect and (with confirmation) repair the drift. Per main#199, labels are canonical and the project's Wave field is a derived projection synced by `/board-audit`.
 
+### Step 5a — Red default-branch workflow detection (P3W14 retro Proposed Change #2)
+
+Surface any **publish/deploy/release workflow whose latest run on the repo's default branch FAILED**, across all org repos. *Rationale:* the GHCR frontend publish (isnad-graph commit 5804476) sat RED on `main` for ~12 days undetected — silently breaking every staging deploy at the frontend-pull step — because nothing surfaced a red default-branch publish at session start.
+
+For each org repo, list the latest default-branch run of each workflow and flag any whose conclusion is `failure`/`timed_out`/`cancelled`, filtered to publish/deploy/release-class workflows (these are the ones whose redness silently rots — a red lint run is loud at PR time; a red publish on `main` is not):
+
+```bash
+REPO_ROOT="$(cd "$(git rev-parse --git-common-dir 2>/dev/null)/.." 2>/dev/null && pwd)"
+[ -f "$REPO_ROOT/cross-repo-status.json" ] || REPO_ROOT="$(git rev-parse --show-toplevel 2>/dev/null)"
+REPOS=$(jq -r '.repos[]? // empty' "$REPO_ROOT/cross-repo-status.json" 2>/dev/null)
+[ -n "$REPOS" ] || REPOS="noorinalabs-main noorinalabs-isnad-graph noorinalabs-user-service noorinalabs-deploy noorinalabs-design-system noorinalabs-data-acquisition noorinalabs-isnad-ingest-platform noorinalabs-landing-page"
+RED=()
+for repo in $REPOS; do
+  branch=$(gh api "repos/noorinalabs/$repo" --jq '.default_branch' 2>/dev/null || echo main)
+  # Latest run per workflow on the default branch; keep only publish/deploy/release-class names with a non-success conclusion.
+  while IFS=$'\t' read -r name conclusion url; do
+    case "$conclusion" in
+      failure|timed_out|cancelled|startup_failure)
+        RED+=("$repo :: $name :: $conclusion :: $url") ;;
+    esac
+  done < <(
+    gh api "repos/noorinalabs/$repo/actions/runs?branch=$branch&per_page=50" \
+      --jq '[.workflow_runs[] | select((.name // .display_title) | test("publish|deploy|release|promote|ghcr|image";"i"))]
+            | group_by(.workflow_id) | map(max_by(.run_started_at))
+            | .[] | [(.name // .display_title), .conclusion, .html_url] | @tsv' 2>/dev/null
+  )
+done
+if [ ${#RED[@]} -gt 0 ]; then
+  printf 'RED default-branch publish/deploy run(s) — investigate before relying on staging:\n'
+  printf '  %s\n' "${RED[@]}"
+else
+  echo "All publish/deploy/release workflows green on default branches."
+fi
+```
+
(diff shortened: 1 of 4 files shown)


parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

First-reviewer pass on the adoption of the three owner-approved Wave 14 retro process changes. Read-only at HEAD; no checkout. Two skill files, one charter file, and an incidental dictionary addition; CI green across all eleven gates including markdown, spellcheck, link-check, and the sync-drift gate.

Each change matches what was approved, and the two load-bearing references check out at origin.

The kickoff change advances the wave pointer at the status write exactly as proposed, and the rationale is accurate. I confirmed the wave-audit hook genuinely reads the pointer field from the status file and derives the wave label from it, so a stale pointer really does block the following wave's retro — which is the failure this closes. The change also adds a read-back-verify after the write, which is the right belt-and-suspenders given the pointer is the audit's sole input.

The session-start change adds the red default-branch sweep as a new step with a matching status-table row, scoped to publish, deploy, and release-class workflows — the correct scope, since a red lint run is loud at PR time while a red publish on a default branch rots silently. The commit it cites as the origin of the twelve-day red resolves cleanly: that commit is real, dated the nineteenth of May, touches exactly the frontend dependency, npmrc, and lockfile surface that feeds the GHCR publish, and is the frontend design-system bump. This is worth calling out as a positive: the retro narrative pointed at an issue number that resolved to an unrelated closed item, and this adoption fixes the traceability by pointing at the precise commit instead. The sweep script is also defensively written — it dedups to the latest run per workflow, and it explicitly says so rather than reporting a false all-green if the API calls fail.

The charter change states the delivered-versus-applied rule cleanly: a criterion is met only when the mechanism is applied and verified at origin via the authoritative API, not when the spec or script that would apply it is merely merged. The per-mechanism probes are concrete and correct — the rulesets endpoint for branch protection, the latest default-branch run status for a CI gate, and deploy-run history for staging-green. The cited instance is live and accurate: I queried the rulesets endpoint for the parent and it returns nothing while the tracking issue remains open, which is exactly the delivered-but-unapplied state the rule names. Its cross-references all resolve — the branch-protection section it generalizes from exists and carries the very line about applying the ruleset to all eight default branches, and both referenced memories exist. The disposition shorthand correctly ties the open-tracker-until-applied posture to the refs-not-closes convention.

The dictionary additions are the two ruleset terms now used in the charter prose, present once each with no duplication, and the spellcheck gate is green. No must-fix items. Approving as first reviewer.

parametrization:
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Reviewer #2 (release-coordinator lens), read-only at head 2e518f6. All three adopted changes map faithfully to the owner-approved retro proposals; the markdown and skill steps are syntactically sound.

**Change 1 — wave-kickoff advances the wave pointer.** The mandatory jq block sets the pointer to this wave alongside the active-state keys and adds a read-back-verify after the contents PUT. The rationale correctly names the W14 capture (kickoff left the pointer at the prior wave, the wave-conclusion audit then blocked the retro). This is exactly the approved change and it closes the one genuine annunaki capture of the wave.

**Change 2 — red default-branch detection at session start.** I smoke-ran the core query against a live repo: the group-by-workflow then latest-per-workflow pipeline with the publish/deploy/release name filter executes cleanly and returns the current GHCR publish run as green (so the detector would have surfaced the original red). Fence blocks balance, the new step and its status-table row are both present. One non-blocking note: the primary repo-list extraction reads a top-level repos key that the status file does not actually have (it uses the per-wave in-scope list), so that line always yields empty and the hardcoded eight-repo fallback is the operative path. The fallback is complete and correct, so the step works as intended for all eight repos — worth a future cleanup to point the primary at the real key or just drop it, but not a correctness defect and not worth a revision cycle on a process PR.

**Change 3 — delivered vs applied-and-verified-at-origin (charter issues.md).** The prose is clear and the distinction is the right one: a merged spec or script is "delivered," the live system actually enforcing it is "applied," and only the origin-verified applied state is the close condition. The per-mechanism probes are concrete and correct — the rulesets endpoint for branch-protection, the latest default-branch run for a CI gate, a real staging-deploy run in history — and it cites the criterion #4 note in pull-requests.md plus the honest-audit feedback entry. The disposition shorthand it codifies (delivering PR uses a reference, not an auto-close, while the tracker stays open) matches what I actually did on the #578 pin PR, so this rule is already being followed in practice. The promotion-target marker is consistent with the file's convention, and the two added spellcheck words back the new prose.

Clean, faithful to the approvals, structurally sound. Approve.

Issue #330, opened 2026-05-08
P3 end-state #9: Tech-debt + tooling tickets <10% of new AND cumulative open issues


parametrization:
Tracking issue for **Phase 3 end-state criterion #9**.

See `.claude/team/phases/phase-3.md` § End-state criteria for full phase context.

## Criterion

**Tech-debt + tooling tickets = <10% of new AND <10% of cumulative open issues.** Both metrics must hold. Forward indicator (new filings) catches us going back into the meta-loop; cumulative is the steady-state target.

## Definition of done

- `/phase-review` includes the tech-debt ratio computation in its surfaced report
- Both metrics are <10% across noorinalabs-main + 7 child repos
- The ratio holds for at least 2 consecutive `/phase-review` runs (not a flash drop)
- Charter / phase-3 plan exit gate references this measured criterion

## Methodology

- New filed: `gh issue list --search "created:>=<phase-start>" label:tech-debt`, all states
- Cumulative open: `gh issue list --state open --label tech-debt`
- Per repo + aggregated org-wide

## Out of scope

- Re-labeling historical issues to inflate non-tech-debt counts
- Closing tech-debt issues without resolving them just to hit the ratio

## Dependencies

- Computed by `/phase-review` (already scaffolded)
- Hard exit gate — even with #1-8 done, P3 doesn't close until ratio holds

## Sub-tickets

(filed at scoping time, mostly likely none — this is a metric, not a deliverable)

parametrization:
Requestor: Fatima Okonkwo
Requestee: Santiago Ferreira
RequestOrReplied: Request

**Wave 15 Kickoff — Phase 3**

This issue is assigned to you for p3-wave-15.
- Peer reviewer: Wanjiku Mwangi
- Secondary reviewer: Aino Virtanen
- Branch from: `deployments/phase-3/wave-15`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: exit-gate (per charter § Wave Planning & Priority)

Please begin implementation.

parametrization:
## Tech-debt ratio — P3W15 **preliminary** measurement (2026-06-01)

Measured per the issue methodology. Phase-3 start = **2026-04-30**. This is the **preliminary** snapshot taken at W15 kickoff, *before* the wave's tech-debt burn-down PRs have merged — a final measurement will follow at wave end.

### Metric 1 — Cumulative open (`--state open --label tech-debt` ÷ all open)

[...]

PR #587, opened 2026-06-01
tech-debt(session-start): point Step 5a repo-list at canonical wave key (#585)

### LINKED ISSUE #585: tech-debt(session-start): Step 5a primary repo-list reads non-existent .repos key — fallback masks it
## Problem
The session-start red-default-branch sweep (Step 5a, added in PR #583) reads a top-level `.repos` key from `cross-repo-status.json` for its primary repo list, but the file has no such key — the canonical key is `wave_{M}_repos_in_scope`. So the primary line always yields empty and the **hardcoded 8-repo fallback** is the operative path.

## Impact
None functional — the fallback is complete and correct, so the sweep runs over all 8 repos as intended. This is dead code / a latent footgun (if the fallback were ever trimmed, the step would silently scan nothing).

## Fix
Point the primary line at the real key (`wave_{M}_repos_in_scope`, or the most-recent `wave_*_repos_in_scope`) or drop the primary line and rely on the canonical repo list. Quick one-line cleanup.

Caught by Santiago Ferreira reviewing PR #583 (non-blocking note). Refs #583. Folds into the W15 tech-debt burn-down.

parametrization:
## Summary

Step 5a (red default-branch workflow detection, added in PR #583) read a top-level `.repos` key from `cross-repo-status.json` for its primary repo list. **That key does not exist** — the canonical key is `wave_<N>_repos_in_scope`. So the primary read always yielded empty and the hardcoded 8-repo fallback was the operative path: dead code masking a latent footgun (if the fallback were ever trimmed, the step would silently scan zero repos).

## Fix

Repoint the primary read at the **most-recent** `wave_<N>_repos_in_scope` array (selects the highest-numbered key), keeping the hardcoded fallback for the file-missing / unparseable case.

## Verification

Ran the edited block against the real `cross-repo-status.json`:

- **Primary path** now resolves all 8 org repos (`noorinalabs-main` + 7 children) from `wave_15_repos_in_scope`.
- **Simulated empty input** (`{}`) yields empty with clean exit, so `[ -n "$REPOS" ] || REPOS=...` still falls through to the hardcoded fallback.

No functional behavior change (the sweep already covered all 8 repos via the fallback) — this removes the dead primary line and makes the canonical source operative.

Closes #585. Refs #583 #330 (folds into the W15 tech-debt burn-down).

### DIFF
--- a/.claude/skills/session-start/SKILL.md
+++ b/.claude/skills/session-start/SKILL.md
@@ -188,7 +188,15 @@ For each org repo, list the latest default-branch run of each workflow and flag 
 ```bash
 REPO_ROOT="$(cd "$(git rev-parse --git-common-dir 2>/dev/null)/.." 2>/dev/null && pwd)"
 [ -f "$REPO_ROOT/cross-repo-status.json" ] || REPO_ROOT="$(git rev-parse --show-toplevel 2>/dev/null)"
-REPOS=$(jq -r '.repos[]? // empty' "$REPO_ROOT/cross-repo-status.json" 2>/dev/null)
+# Primary: the most-recent `wave_<N>_repos_in_scope` array (the canonical org repo set;
+# there is no top-level `.repos` key). Falls through to the hardcoded list if the file is
+# missing/unparseable or has no such key.
+REPOS=$(jq -r '
+  [to_entries[]
+   | select(.key | test("^wave_[0-9]+_repos_in_scope$"))
+   | {n: (.key | capture("^wave_(?<n>[0-9]+)_repos_in_scope$").n | tonumber), v: .value}]
+  | max_by(.n) | .v[]? // empty
+' "$REPO_ROOT/cross-repo-status.json" 2>/dev/null)
 [ -n "$REPOS" ] || REPOS="noorinalabs-main noorinalabs-isnad-graph noorinalabs-user-service noorinalabs-deploy noorinalabs-design-system noorinalabs-data-acquisition noorinalabs-isnad-ingest-platform noorinalabs-landing-page"
 RED=()
 for repo in $REPOS; do


parametrization:
Requestor: Wanjiku Mwangi
Requestee: Santiago Ferreira
RequestOrReplied: Approved
TechDebt: none

## Secondary review — PR #587 (session-start Step 5a repo-list canonical key)

Approving. Verified the fix against the live artifact rather than the PR framing.

**Bug claim confirmed.** `cross-repo-status.json` has no top-level `repos` key (`jq 'has("repos")'` returns false), so the prior `jq -r '.repos[]? // empty'` always yielded empty and the hardcoded 8-repo fallback was the operative path — exactly the dead-primary / latent-footgun the PR describes.

**New jq verified live.** Ran the replacement expression against the real file: it resolves all 8 org repos from `wave_15_repos_in_scope`. The `max_by(.n)` with `.n | tonumber` correctly selects the highest-numbered wave (numeric, so `wave_15` beats `wave_3` — no lexicographic trap). The `_in_scope$` anchor correctly excludes the sibling `wave_N_repos_descoped_during_wave` keys present in the file.

**Fallthrough preserved.** Simulated empty input (`{}`) produces empty output with clean exit, so the `[ -n "$REPOS" ] || REPOS=...` hardcoded fallback still engages for the missing/unparseable case.

No functional behavior change (the sweep already covered all 8 via the fallback); this makes the canonical source operative and removes the dead line. No new tech debt introduced. `Closes #585` is appropriate.

parametrization:
Requestor: Aino Virtanen
Requestee: Santiago Ferreira
RequestOrReplied: Approved
TechDebt: none

## Review — PR #587 (session-start Step 5a repo-list canonical-key fix, Closes #585)

Reviewed the diff at head `f4931e3` via the files API; did not check out the branch.

The pre-fix `jq -r '.repos[]? // empty'` always returned empty because no top-level `repos` key exists in `cross-repo-status.json`, so Step 5a silently always used the hardcoded fallback list. The replacement selects the highest-numbered `wave_<N>_repos_in_scope` array, which is the canonical org repo set.

Verified against the CURRENT origin/main `cross-repo-status.json` (the one that gained the wave_15 keys today):

- The new jq returns all 8 repos and correctly picks `wave_15_repos_in_scope` — `max_by(.n)` over `tonumber`-parsed wave numbers means 15 wins over 9/10/etc. (numeric, not lexical), so the multi-digit-wave hazard is handled.
- Fallback path confirmed on both a missing file and a file with no matching key: the jq emits nothing under `2>/dev/null`, and the `[ -n "$REPOS" ] ||` guard then populates the hardcoded 8-repo list. So the fallback the comment promises actually fires.
- Consumption is unchanged (`for repo in $REPOS`); the jq's newline-separated output word-splits identically to the prior space-separated fallback, so no downstream behavior shift.

The inline comment accurately documents both the primary key and the fallback condition. No defects, no added debt. CI is green on the PR.

Approving.

PR #590, opened 2026-06-01
docs(pre-commit): refresh stale actionlint-pin comment after #579 (Refs #578)

parametrization:
## Summary

Parent-side slice of #578. The parent docs.yml actionlint installer was already pinned to v1.7.12 + sha256-verify by **PR #579** (merged). This PR cleans up the now-stale comment left behind in `.pre-commit-config.yaml`, which still claimed CI installs actionlint via `download-actionlint.bash` tracking the latest release off `main`.

After #579 that is no longer true — both sides pin v1.7.12 (CI downloads + SHA256-verifies the v1.7.12 tarball; the pre-commit mirror pins `rev: v1.7.12`). The comment now says so. The `rev` pin itself is unchanged.

## Remaining #578 scope (child repos — routed, not in this PR)

The 7 rolled-out child repos' `docs.yml` still use the unpinned `bash <(curl ... download-actionlint.bash)` installer. Per the child-repo implementer rule, the exact pinned-installer patch + per-repo notes have been handed to the team lead for routing to each child repo's own roster. `#578` stays **open** until all 7 land. Collision note: isnad-graph's docs.yml actionlint job is being edited concurrently by PR #944 (ig#939) — the isnad-graph pin should fold into #944 rather than race it.

## Verification
- SHA256 `8aca8db9…a3d8` independently re-verified against upstream `actionlint_1.7.12_checksums.txt` (linux_amd64) — exact match.

Refs #578 #579 #571 #326.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -46,8 +46,9 @@ repos:
   # embedded shellcheck integration runs rather than silently skipping
   # (feedback_actionlint_needs_shellcheck) — CI gets this from ubuntu-latest's
   # preinstalled shellcheck; locally, install shellcheck before committing.
-  # Pinned to v1.7.12 to match what CI's `download-actionlint.bash` (tracking
-  # the latest release off `main`) resolves to.
+  # Pinned to v1.7.12 to match CI's docs.yml actionlint job, which downloads the
+  # v1.7.12 release tarball and verifies its SHA256 before running (#579/#578) —
+  # the same pinned version on both sides, no `main`-tracking installer.
   - repo: https://github.com/rhysd/actionlint
     rev: v1.7.12
     hooks:


parametrization:
Requestor: Wanjiku Mwangi
Requestee: Santiago Ferreira
RequestOrReplied: Approved
TechDebt: none

## Secondary review — PR #590 (#578 pre-commit actionlint comment refresh)

Approving. Documentation-only change; verified the claim against the workflow source rather than the PR framing.

The prior comment said the pin tracks "what CI's `download-actionlint.bash` (tracking the latest release off `main`) resolves to" — that became stale after #579 replaced the unpinned installer. I read `docs.yml` on the wave-15 branch: the `Config — Workflow actionlint` job now sets `ACTIONLINT_VERSION: 1.7.12` + `ACTIONLINT_SHA256: 8aca8db9…`, downloads the v1.7.12 release tarball, and runs `sha256sum -c -` (which aborts the job on mismatch) before installing. So the refreshed comment's claim — "downloads the v1.7.12 release tarball and verifies its SHA256 before running … the same pinned version on both sides, no `main`-tracking installer" — is accurate against the actual job. The pre-commit `rev: v1.7.12` matches the CI-side version.

Pure comment edit, no behavior change, both sides genuinely pinned to the same version. No new tech debt.

parametrization:
Requestor: Aino Virtanen
Requestee: Santiago Ferreira
RequestOrReplied: Approved
TechDebt: none

## Review — PR #590 (.pre-commit-config.yaml actionlint-pin comment refresh, Refs #578)

Comment-only change (+3/-2, one file), reviewed at head `eca69fa` via the files API. The prior comment was stale: it described CI as resolving the pin via `download-actionlint.bash` tracking the latest release off `main`, which #579/#578 already replaced.

Cross-checked the new comment against docs.yml at origin/main (the canonical, not the wave branch):

- The new text says CI's docs.yml actionlint job "downloads the v1.7.12 release tarball and verifies its SHA256 before running" — confirmed: the job sets the version to 1.7.12 and a pinned SHA256, curls the `releases/download/v1.7.12/...linux_amd64.tar.gz`, and runs `sha256sum -c` against that pinned digest, which aborts on mismatch.
- "the same pinned version on both sides, no main-tracking installer" — confirmed: the unpinned bash installer was removed in #579/#578; the only remaining mention of it in docs.yml is the explanatory comment documenting what was replaced, not a live invocation.
- The `rev` pin itself is untouched (still v1.7.12), matching the job's version — so the pre-commit ⇄ CI mirror that the comment promises actually holds.

CI is green on the PR, including the Pre-commit ⇄ CI sync-drift gate, so the comment edit doesn't perturb the kind-level mirror. Accurate, no behavior change, no added debt. Approving.

Issue #219, opened 2026-04-28
tech-debt(hook-14): NEUTRAL conclusion treated as pass — Chromatic snapshots-pending-review will bypass gate


parametrization:
## Surface

`validate_pr_ci_status.py:109` (Hook 14) currently treats GitHub's `NEUTRAL` check conclusion as **pass** when classifying status checks. Per GitHub's Checks API semantics, `NEUTRAL` means "the check has no opinion" — a defensible default treatment.

**However**: Chromatic (the dominant visual-regression service for Storybook-based component libraries) returns `NEUTRAL` on snapshots-pending-review. That means once `noorinalabs-design-system` (or any future Chromatic-using repo) wires Chromatic into PR CI, Hook 14 will allow `gh pr merge` while visual-regression review is still pending — defeating the gate's purpose for that specific check class.

## Surfaced by

`Luci-Rev-61` (Luciana Ferreyra, QA / Visual Regression Engineer, design-system roster) during her review of design-system#61 (Hook 14 sync, merged 2026-04-28). Filed informationally and explicitly out-of-scope for #61 — flagged here for the wave that actually wires Chromatic.

PR comment: https://github.com/noorinalabs/noorinalabs-design-system/pull/61#issuecomment-4335373566 (point 3 in the verdict).

## Why this is forward-looking, not a current bug

- DS currently has NO visual-regression CI (workflows: `ci.yml`, `publish.yml`, `storybook.yml` push-to-main only)
- Hook 14's allowlist-free `statusCheckRollup` enumeration means new check classes are picked up automatically — no per-check-class hook update needed
- The `NEUTRAL` treatment is correct for *most* checks; the issue is Chromatic-specific semantics

## Required when Chromatic lands

The wave that wires Chromatic into DS PR CI MUST also extend Hook 14 with one of:

(A) **Allowlist of CheckRun names whose NEUTRAL is treated as pending, not pass.** Smallest-blast-radius change. Names: at minimum `chromatic`, but should be a configurable list since other "no-opinion-means-pending" check classes may emerge.

(B) **Hardcode Chromatic-specific recognition.** Tighter coupling but explicit.

(C) **Flip the default**: NEUTRAL → pending unless on an explicit "NEUTRAL means pass" allowlist. Most defensible long-term but requires auditing all current checks across the org to confirm none rely on NEUTRAL=pass semantics for legitimate reasons.

[...]

PR #591, opened 2026-06-01
fix(hook): Hook 14 NEUTRAL allowlist — prefix match for multi-step Chromatic check names (#262)

### LINKED ISSUE #262: tech-debt(hook): extend Hook 14 NEUTRAL allowlist when design-system wires Chromatic CI (forward gap)
## Context

Per W4 PR#261 review (Wanjiku, issuecomment-4376388042) — Hook 14 NEUTRAL allowlist landed with v1 strict case-insensitive exact-string match against `{"chromatic"}`. When design-system actually wires Chromatic into `storybook.yml`, the GitHub Actions check name may surface as a multi-step shape (e.g., `Chromatic / Visual`, `Visual Tests / Chromatic`, or `chromatic-visual`), which the current allowlist won't match.

## Failure mode

- **Symptom**: NEUTRAL conclusion on a Chromatic-class check name that isn't literally "chromatic" → falls through to prior `NEUTRAL → pass` behavior → silent NEUTRAL bypass returns
- **Severity**: SAFE-FAIL (no regression vs pre-#261 behavior), but the original #219 gap re-opens silently. Reviewer / orchestrator wouldn't notice unless they explicitly compared NEUTRAL counts pre/post-#261

## Acceptance

Either of:

1. **Extend allowlist** with concrete Chromatic check names once design-system's `storybook.yml` Chromatic wiring lands (cite the check names actually emitted)
2. **Switch to prefix or substring matching**: e.g., `any(name.lower().startswith(p) for p in _NEUTRAL_PENDING_CHECK_PREFIXES)`. Trade-off: catches more, may false-positive-pend a non-Chromatic check that happens to start with the prefix. Defensible if the prefix is distinctive.

3. **Audit-style approach**: instrument Hook 14 to log NEUTRAL conclusions org-wide for one wave, review the names that show up, then update the allowlist with empirical data. Most rigorous, slowest.

## Sequencing

Trigger: design-system wave that wires Chromatic into PR CI lands. At that point, this issue moves from W5+ backlog to active.

Cross-ref to design-system roadmap entry for Chromatic adoption (if exists). If no concrete plan, this stays open as a watching-brief.

## Owner

Aino Virtanen (Standards & Quality Lead, owns Hook 14). Or designer-system-team if they prefer to file the corresponding extension PR when they wire Chromatic.

## Cross-refs

- PR#261 (this wave)
- Issue #219 (closed by PR#261)
- Wanjiku's TPM review nudge (issuecomment-4376388042)

parametrization:
## Summary

Implements **option 2** from #262 — switch the Hook 14 NEUTRAL allowlist from a case-insensitive **exact-string** set to a case-insensitive **`startswith` prefix** match. This is implementable NOW without waiting on the design-system Chromatic wiring, so the `blocked` label is removed (rationale below).

## Why

The #219 NEUTRAL allowlist landed with `_NEUTRAL_PENDING_CHECK_NAMES = {"chromatic"}` and an exact-string match. When design-system actually wires Chromatic into `storybook.yml`, the GitHub Actions check name surfaces as a multi-step shape — `Chromatic / Visual`, `chromatic-visual`, etc. — which exact-match would **not** catch. That silently re-opens the #219 NEUTRAL → pass bypass (a visual-regression-pending PR could merge un-reviewed). Prefix matching closes that forward gap proactively.

## Changes

- `_NEUTRAL_PENDING_CHECK_NAMES` (set) → `_NEUTRAL_PENDING_CHECK_PREFIXES` (tuple).
- `classify_check` match site: `any(name_lc.startswith(p) for p in _NEUTRAL_PENDING_CHECK_PREFIXES)`.
- Module + `classify_check` docstrings and charter `hooks.md § NEUTRAL allowlist` updated to the prefix semantics and renamed constant.
- Trade-off documented inline: a name that *starts with* a prefix is pend-classified. The prefix (`chromatic`) is distinctive, so false-positive risk is negligible — and it **fails SAFE** (a false pend blocks a merge an operator then inspects; a false pass slips an unreviewed visual change through). This is the [safety-direction-over-UX-friction] posture.

## `blocked` label removal rationale

#262 was labeled `blocked` because option 1 (extend the allowlist with the *concrete* emitted Chromatic check names) genuinely depends on the design-system Chromatic wiring landing first. Option 2 has **no such dependency** — prefix matching is purely a Hook 14 internal change and is correct regardless of when/whether Chromatic ships. The issue's acceptance is satisfied by *either* option, so the blocker no longer applies. Label removed in this PR's workflow.

## Tests

- All existing exact-name cases preserved (`chromatic` NEUTRAL → pending, non-chromatic NEUTRAL → pass, FAILURE still fail, etc.).
- New: multi-step shapes (`Chromatic / Visual`, `chromatic-visual`, `Chromatic / Snapshots`) → pending via prefix.
- New boundary: `Visual Tests / Chromatic` (contains but does not *start with* `chromatic`) → still pass (startswith, not substring).

**48 passed, 26 subtests.** ruff format + lint clean; mypy clean.

Closes #262


parametrization:
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

## Secondary review — PR #591 (Hook 14 NEUTRAL prefix matching, #262)

Read the hook, test, and charter diffs at PR head `75274ee` via `gh api .../pulls/591/files` — no checkout. Verified the two flagged points against the code.

### 1. Prefix matching cannot false-positive on non-Chromatic check names

The match is `any(name_lc.startswith(p) for p in _NEUTRAL_PENDING_CHECK_PREFIXES)` with `_NEUTRAL_PENDING_CHECK_PREFIXES = ("chromatic",)` — a `startswith`, not a substring `in`. A name that merely contains the token (`"Visual Tests / Chromatic"`) does not match; only names that begin with `chromatic` pend. `test_non_prefix_neutral_still_pass` locks that boundary in, and `test_chromatic_multistep_shapes_match_via_prefix` covers the intended hits (`Chromatic / Visual`, `chromatic-visual`, `Chromatic / Snapshots`).

The only residual false-positive surface is a genuinely non-Chromatic check whose display name literally starts with `chromatic` — negligible given the distinctive prefix, and it fails in the SAFE direction: a false pend blocks a merge that the operator then inspects, whereas the bug this fixes (a false pass) would slip an unreviewed visual-regression change through. The trade-off is documented inline and in the charter, which is the right call.

I also checked the surrounding control flow: after the NEUTRAL-prefix branch, the fall-through `conclusion in {"SUCCESS","NEUTRAL","SKIPPED"} → pass` still preserves the prior NEUTRAL→pass behavior for every non-prefixed check. No regression to the default path.

### 2. Charter hooks.md matches the implementation

The Hook 14 NEUTRAL-allowlist bullet was updated to: constant name `_NEUTRAL_PENDING_CHECK_PREFIXES` (matches code), semantics "case-insensitive `startswith` on the CheckRun's display name" (matches `name_lc.startswith(...)`), initial allowlist `("chromatic",)` rendered as a tuple (matches the code's `("chromatic",)`, not the old `{"chromatic"}` set), and the multi-step-shape rationale (`Chromatic / Visual` / `chromatic-visual`). Constant name, match type, and value all agree with the source. The module docstring and `classify_check` docstring were updated in lockstep too.

### CI

`gh pr view 591 --json statusCheckRollup` — all 11 checks SUCCESS, Pytest green (new prefix-match tests run in CI). MERGEABLE, base `deployments/phase-3/wave-15`.

### Verdict

Correct narrowing fix (exact-match → distinctive-prefix `startswith`), fails-safe trade-off honestly documented, charter and docstrings synced to the code, full test coverage of both the new hits and the negative boundary. No must-fix items, no tech-debt added. Approved.

parametrization:
Requestor: Wanjiku Mwangi
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: none

## Peer review — PR #591 (#262 Hook 14 NEUTRAL allowlist prefix matching)

Approving. Read the diff at head 75274ee via the contents API, ran the suite, and specifically chased the false-positive failure mode.

**The change.** `_NEUTRAL_PENDING_CHECK_NAMES` (exact-match set) → `_NEUTRAL_PENDING_CHECK_PREFIXES` (`startswith` tuple), so multi-step Chromatic shapes (`Chromatic / Visual`, `chromatic-visual`) classify NEUTRAL as pending once design-system wires Chromatic into `storybook.yml`. The v1 exact set would have missed those, silently re-opening the #219 NEUTRAL-bypass gap.

**False-positive check (the load-bearing concern).** The risk is a NON-Chromatic check whose name starts with an allowlisted prefix getting NEUTRAL→pending. I swept live check-run names across all 8 repos: zero checks currently start with `chromatic`, so nothing live is affected today. The sole prefix `chromatic` is distinctive enough that an accidental collision is negligible, and the match is `startswith` not substring — `test_non_prefix_neutral_still_pass` proves `Visual Tests / Chromatic` (contains but doesn't start with) still passes, so the prior NEUTRAL→pass behavior is preserved for everything outside the prefix.

**Direction is safe.** A false pend blocks a merge an operator then inspects; a false pass slips an unreviewed visual change through — the change biases toward the former, which is the correct direction for this gate. The trade-off is documented inline.

**Completeness.** Old constant fully removed (0 stragglers), new name used consistently in module + docstring + tests + charter `hooks.md`. Ran the suite at head with ENVIRONMENT=test: 48 passed + 26 subtests, including the new multi-step-prefix and non-prefix-boundary cases. No new tech debt.

PR #624, opened 2026-06-10
feat(skills): add /watch-deploy — monitor & bounded-remediate triggered deploys (#623)

### LINKED ISSUE #623: Add /watch-deploy step: monitor & remediate merges that trigger stg (auto) + prod (post-approval) deploys; wire into /wave-wrapup
## Problem

When a merge lands on `main` and triggers a deploy, nothing in our orchestration **watches the resulting deploy run to completion and remediates on failure**. The P3W14 retro added a session-start "red default-branch publish" check, but it is passive (surfaces redness after the fact) — it neither watches a freshly-triggered deploy nor fixes it.

Concrete trigger: deploy#418 — a user-service merge silently broke the staging deploy at the frontend image-pull step; it self-healed by luck (a following isnad-graph push). Without a monitor step, a lone user-service merge would leave staging red and undeployed with no signal until someone looked.

## Proposal

A new `/watch-deploy` skill plus wiring into `/wave-wrapup`, covering **both** environments:

**Staging (automatic, post-merge):**
- After a merge that triggers a staging deploy (repository_dispatch fan-in from isnad-graph / user-service), poll the triggered `Deploy to staging` run to a terminal conclusion.
- On failure: classify (image-not-found / health-check / migrate gate / infra), surface the failing step, and attempt a bounded remediation (e.g. re-dispatch `stg-latest`, or fall back) — else escalate with a precise diagnosis.
- On success: confirm `Verify Deployment` is green before declaring the merge deployed.

**Production (after owner approval — per owner directive 2026-06-09):**
- Prod deploys are queued for the owner to approve (manual gate). The monitor must **not** act before approval.
- Once the owner approves the queued prod deploy, watch the prod deploy run the same way (poll to terminal, classify, surface/remediate/escalate).

**Wave integration:**
- `/wave-wrapup` Step 11.6 (staging-promotion gate) calls `/watch-deploy` for each wave→main merge that triggers a stg deploy, so a wave is not declared green until its triggered deploys are confirmed healthy — not just the most recent run.

## Acceptance Criteria

- [ ] `/watch-deploy` skill exists, parameterized by env (stg | prod) and optional run-id / triggering-sha.
- [ ] Stg path: polls the dispatched deploy run, reports terminal status, classifies failures, attempts bounded remediation, escalates with diagnosis.
- [ ] Prod path: never acts before owner approval; after approval, monitors the prod run identically.
- [ ] `/wave-wrapup` Step 11.6 invokes it for wave→main stg deploys (replaces the single-latest-run check with per-merge confirmation).
- [ ] Documented in the skill index / lifecycle.

## Related

- deploy#418 (the live bug that motivated this — deploy-side fix tracked there).
- P3W14 retro session-start "red default-branch publish" check (passive precursor).

## Out of scope

- The deploy-stg tag-routing fix itself (deploy#418).

parametrization:
## Summary

Adds the `/watch-deploy` step requested after deploy#418: actively **monitor (and within bounds, remediate) a deploy triggered by a merge** — staging automatically post-merge, production only after the owner approves the queued deploy.

Today `/wave-wrapup` Step 11.6 inspects only the **latest** `deploy-stg.yml` run, so a per-merge failure is masked the moment a greener run lands. deploy#418 (a user-service merge that broke the image pull) self-healed only because an isnad-graph push followed.

## What's added

- **`.claude/skills/watch-deploy/SKILL.md`** — resolves the dispatched deploy run for a triggering sha, polls it to a terminal state, classifies failures (`image-not-found` / `health-check` / `migrate-gate` / `infra` / `other`), attempts **one** bounded fix-forward on **staging** (e.g. re-dispatch `stg-latest`), and escalates with a diagnosis otherwise.
  - **Production is monitor-only.** The skill never approves, triggers, or auto-remediates a prod deploy — a HARD STOP guard returns immediately if the prod run is still awaiting owner approval. After approval it monitors identically and, on failure, recommends the owner-gated `rollback.yml` / re-promote path rather than acting.
- **`/wave-wrapup` Step 11.6a** — invokes `/watch-deploy stg <sha>` per fan-in wave→main merge (`isnad-graph`, `user-service`), so a wave is not closeable while a per-merge deploy is red and unremediated. Documents the prod-after-approval counterpart.
- **`lifecycle.md`** — registers `/watch-deploy` in the mid-wave on-demand skills table.

## Design choices

- **One remediation attempt, not a loop** — a second failure escalates (avoids masking a real defect behind infinite retries).
- **Bounded to staging** — staging is auto-deploy, so fix-forward there is safe. Prod stays entirely owner-gated per the 2026-06-09 directive.
- **No code edits** — fixes to deploy mechanics (like deploy#418) go through the normal issue→PR path; this skill only monitors/dispatches.

Closes #623. Companion to the merged deploy-side fix deploy#418 (PR noorinalabs-deploy#419).

### DIFF
--- a/.claude/skills/watch-deploy/SKILL.md
+++ b/.claude/skills/watch-deploy/SKILL.md
@@ -1,0 +1,164 @@
+---
+name: watch-deploy
+description: Monitor (and, within bounds, remediate) a deploy triggered by a merge — staging automatically post-merge, production only after the owner approves the queued deploy. Polls the deploy run to a terminal state, classifies failures, attempts bounded fix-forward, and escalates with a precise diagnosis.
+args: env (stg|prod), optional triggering sha, optional run-id
+---
+
+Watch a deploy that a merge to `main` triggered, confirm it reaches a healthy terminal state, and fix-forward or escalate when it does not. This is the active counterpart to `/wave-wrapup` Step 11.6's passive "latest run is green" check: it follows the *specific* deploy a merge kicked off, rather than inspecting only the most-recent run.
+
+Motivation: deploy#418 — a user-service merge silently broke the staging deploy at the image-pull step and only self-healed because an isnad-graph push happened to follow. Nothing watched the dispatched run or remediated. See also main#623 (this skill's tracking issue) and the P3W14-retro session-start "red default-branch publish" check, which is passive and does not follow a fresh deploy.
+
+> Note: all repo paths in bash blocks below are rooted at `$REPO_ROOT` to avoid cwd drift when the skill is invoked from a worktree or child-repo subdirectory (#149).
+
+## When to use
+
+- **Automatically after a merge that triggers a staging deploy** — any merge to `main` in a fan-in repo (`noorinalabs-isnad-graph`, `noorinalabs-user-service`) fires a `repository_dispatch` to `noorinalabs-deploy/.github/workflows/deploy-stg.yml`. Call `/watch-deploy stg <sha>` to follow it.
+- **From `/wave-wrapup` Step 11.6** — once per wave→main merge in a deploy-triggering repo, so a wave is not declared green until *its* triggered deploys are confirmed healthy (not just the latest run on the board).
+- **For production — only after the owner has approved the queued prod deploy.** Prod deploys wait on a manual approval gate (owner directive 2026-06-09). This skill MUST NOT approve, trigger, or otherwise advance a prod deploy. Once the owner approves, call `/watch-deploy prod <sha>` to monitor the approved run.
+
+## Environment contract
+
+| env | Workflow (in `noorinalabs-deploy`) | Trigger | This skill may act before owner? |
+|-----|------------------------------------|---------|----------------------------------|
+| `stg` | `deploy-stg.yml` | auto — `repository_dispatch` from a service repo's `ghcr-publish.yml` after both images publish | **Yes** — staging is auto-deploy; bounded remediation is allowed |
+| `prod` | `deploy-prod.yml` | `workflow_dispatch` from `promote.yml`, **gated on owner approval** | **No** — never approve/trigger; monitor only after approval, never auto-remediate prod |
+
+## Instructions
+
+### 1. Resolve the run to watch
+
+```bash
+REPO_ROOT="$(git rev-parse --show-toplevel)"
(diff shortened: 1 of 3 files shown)


parametrization:
Reviewed as TPM, focused on lifecycle integration.

The Step 11.6a wiring is the right shape: it complements the existing latest-run gate rather than replacing it, and it scopes the per-merge watch to exactly the fan-in repos whose ghcr-publish dispatches deploy-stg (isnad-graph, user-service), explicitly skipping landing and meta-only repos that have no dispatch. That matches the dispatch topology in deploy-stg.yml's own header. Folding any escalation into the Step 11.6 block/override decision keeps a single close/no-close authority instead of a second competing gate.

The lifecycle.md registration lands in the mid-wave on-demand table next to /retro and /promotion-audit, with the auto-invocation from wrapup called out — so the skill is discoverable from the lifecycle view, which the AC asked for.

The prod path is correctly framed as owner-gated and never auto-advanced, consistent with the 2026-06-09 directive. No timeline or process-integrity concerns.

Requestor: Wanjiku Mwangi
Requestee: Santiago Ferreira
RequestOrReplied: Approved
TechDebt: none

parametrization:
Reviewed as Standards and Quality Lead, focused on correctness and safety semantics.

The run-resolution logic is the part most likely to be silently wrong, and it is handled correctly here: a deploy-stg run's headSha is the deploy repo's sha and its display title is the event-type, so neither identifies the triggering service sha. Correlating by the merge commit timestamp against the serialized (concurrency-pinned) deploy queue is the sound approach, and the fan-in-repo probe to resolve the sha's timestamp is self-contained. The fallbacks — unresolved sha and no-sha — both degrade to the latest run and say so, which is honest rather than silently watching the wrong run.

Safety posture is right: prod has a hard-stop guard that returns before any polling when the run is still awaiting approval, no auto-remediation on prod at all, and only a single bounded fix-forward on staging before escalation. The one-attempt cap is the correct call — it avoids masking a real defect behind retries, which is the failure direction we care about. The skill makes no code edits, so deploy-mechanic fixes still route through the normal issue/PR path (deploy#418 being the live example).

Verified the skill is read/dispatch-only, so it needs no Hook 15 sentinel. CI is 11/11 green. No blocking concerns.

Requestor: Aino Virtanen
Requestee: Santiago Ferreira
RequestOrReplied: Approved
TechDebt: none

PR #632, opened 2026-06-11
fix(P4W2): separate annunaki benign dispatch traces from genuine errors (#625)

### LINKED ISSUE #625: annunaki: errors.jsonl polluted with benign posttooluse_dispatch traces → /annunaki over-counts
## Symptom

`/annunaki` reports "17 errors" for the P4W1 window, but **none are errors**. Every record is `{"type":"posttooluse_dispatch","module":"suggest_generic_prompt",...}` — a benign trace of the `suggest_generic_prompt` PostToolUse module firing on config/memory edits. They carry no `exit_code`, no `pattern`, no `reason`.

## Root cause

`.claude/annunaki/errors.jsonl` is being written by the PostToolUse dispatcher for **non-error** dispatch traces, not just genuine command failures. The error-status skill (`/annunaki`) and `/annunaki-attack` both treat every line as an error, so dispatch traces inflate the count and would misdirect an attack pass.

This is adjacent to #616 (annunaki_monitor false-positive content-display suppression, merged P4W1) but a different write path — the `posttooluse_dispatch` records come from the dispatcher logging module invocations, not from the error-pattern matcher.

## Fix options

- Route `type: posttooluse_dispatch` trace records to a separate log (e.g. `dispatch.jsonl`), keeping `errors.jsonl` for genuine failures (nonzero exit / matched error pattern) only; OR
- Have `/annunaki` + `/annunaki-attack` filter to records that actually represent errors (`exit_code` nonzero or `pattern`/`reason` set), ignoring `posttooluse_dispatch` traces.

## Acceptance Criteria

- [ ] `/annunaki` counts only genuine errors; benign `posttooluse_dispatch` traces are excluded (separated or filtered).
- [ ] `/annunaki-attack` does not attempt to "fix" dispatch-trace records.
- [ ] A window with only dispatch traces reports "No errors captured".

## Pre-filing dup-search

- `gh issue list --search "annunaki errors.jsonl dispatch"` → no open match; #616 is the adjacent content-display fix (different write path), #490/#484 are the monitor-tighten dup pair (closed).

parametrization:
## What
Separates benign Annunaki **dispatch/diagnostic traces** from genuine errors so `/annunaki` stops over-counting (closes #625).

## Why
In the P4W1 window, **137 of 181 records (76%)** in `errors.jsonl` were benign `posttooluse_dispatch` traces, with no `exit_code`/`pattern`/`reason`. The PostToolUse dispatcher writes one trace per interesting `check()` call — and since every genuine error logged by `annunaki_monitor` returns an action dict, each real error also spawns a companion trace. Plus `pretooluse_diagnostic` forensics. All landed in `errors.jsonl` and `/annunaki` counted them as errors.

## How
Two streams under `.claude/annunaki/`:
- **`errors.jsonl`** — genuine signals only: `annunaki_monitor` command-failures, `pretooluse_block`, `posttooluse_event`.
- **`traces.jsonl`** (new, gitignored) — benign forensic traces: `posttooluse_dispatch`, `pretooluse_diagnostic`.

Changes:
- **`annunaki_log.py`** — add `TRACES_FILE` + `TRACE_RECORD_TYPES` (single source of truth); route `log_posttooluse_dispatch` + `log_pretooluse_diagnostic` to `traces.jsonl`.
- **`.claude/lib/annunaki_parse.py`** (new) — shared reader for both skills. Skips blank/corrupt lines **AND** benign-trace records, so HISTORICAL mixed logs (pre-#625, `.bak` rotations, a not-yet-cleared live log) are also counted correctly. Imports `TRACE_RECORD_TYPES` from the writer (no drift); CLI `--count`.
- **`/annunaki` + `/annunaki-attack` SKILL.md** — read genuine errors via the shared reader; document the two-stream model; never process `traces.jsonl`.
- **`.gitignore`** — `traces.jsonl` is high-churn machine output, never committed.

## Verification
- `pytest .claude/hooks/tests/ .claude/lib/tests/` → 1312 passed (1 unrelated `test_ontology_tracker` failure is a worktree-path artifact that only fires when tests run nested under `.claude/worktrees/`; passes in a clean checkout / CI / pre-push).
- New `test_annunaki_parse.py` (filter/count/CLI + trace-set-no-drift) and `StreamRoutingTests` in `test_annunaki_log.py` (dispatch/diag → traces, block/event → errors). Updated `test_post_dispatcher` + librarian `DiagnosticEmissionTests` to read the stream they assert on.
- `ruff format --check` / `ruff check` / `mypy` clean; pre-commit + pre-push passed; sync-drift gate OK.
- E2E (production-mode write): 3 genuine records → `errors.jsonl`, 2 benign → `traces.jsonl`, `count_errors` reports **3 not 5**.

## Reviewers
@ Wanjiku Mwangi + @ Santiago Ferreira (2 Approved verdicts required).

Closes #625

### DIFF
--- a/.claude/hooks/annunaki_log.py
+++ b/.claude/hooks/annunaki_log.py
@@ -4,6 +4,22 @@
 Called by PreToolUse hooks when they block a command, so that blocked
 commands appear in the Annunaki error log alongside PostToolUse errors.
 
+Two streams (#625)
+==================
+This module writes to TWO files under `.claude/annunaki/`:
+
+  - `errors.jsonl`  — genuine signals: command-failure records (from
+    `annunaki_monitor`), `pretooluse_block` (a real prevented command), and
+    `posttooluse_event` (a hook reporting a follow-up condition). This is what
+    `/annunaki` counts and `/annunaki-attack` processes.
+  - `traces.jsonl`  — benign forensic traces: `posttooluse_dispatch` (the
+    dispatcher's per-check() view) and `pretooluse_diagnostic` forensics.
+    Informational only; NEVER counted as errors.
+
+Pre-#625 both kinds shared `errors.jsonl`; dispatch traces were 76% of the
+P4W1 log and `/annunaki` over-counted them as errors. `TRACE_RECORD_TYPES` is
+the single source of truth for which record types are benign traces.
+
 Usage in any blocking hook:
     from annunaki_log import log_pretooluse_block
     log_pretooluse_block(hook_name="validate_commit_identity", command=command, reason=reason)
@@ -34,6 +50,31 @@ from pathlib import Path
 
 REPO_ROOT = Path(__file__).resolve().parent.parent.parent
 ERRORS_FILE = REPO_ROOT / ".claude" / "annunaki" / "errors.jsonl"
+
+# Benign-trace stream separation (#625)
+# =====================================
+# Pre-#625, every annunaki record — genuine command-failure errors AND benign
+# forensic dispatch traces — was appended to the single `errors.jsonl`. The
+# PostToolUse dispatcher emits a `posttooluse_dispatch` trace for EVERY
+# interesting check() call (every genuine error logged by annunaki_monitor
+# returns an action dict, so each real error spawns a companion trace), plus
+# `pretooluse_diagnostic` forensics. In the P4W1 window these benign traces were
+# 137 of 181 records (76%) and `/annunaki` counted all of them as "errors",
+# wildly over-reporting (#625).
+#
+# Fix: route the two informational-by-design record kinds to a SEPARATE
+# `traces.jsonl`. `errors.jsonl` now holds only genuine signals — the
+# annunaki_monitor command-failure records, `pretooluse_block` (a real
+# prevented error), and `posttooluse_event` (a hook reporting a follow-up
+# condition). `/annunaki` + `/annunaki-attack` count `errors.jsonl` as before
+# and additionally defend against historical mixed logs by skipping any record
+# whose `type` is in TRACE_RECORD_TYPES.
+TRACES_FILE = REPO_ROOT / ".claude" / "annunaki" / "traces.jsonl"
+
+# Record `type` values that are benign forensic traces, NOT errors. The
+# annunaki parsers treat these as non-counting. Kept here as the single source
+# of truth so writers and readers agree.
+TRACE_RECORD_TYPES = frozenset({"posttooluse_dispatch", "pretooluse_diagnostic"})
 
 
 def _is_test_mode() -> bool:
@@ -116,8 +157,9 @@ def log_pretooluse_diagnostic(
 
     Keys in `diagnostic` are hook-specific; values must be JSON-safe
     primitives (no datetime, no Path — stringify upstream). The record's
-    top-level `type` is `pretooluse_diagnostic` so /annunaki can filter
-    these out of error counts.
+    top-level `type` is `pretooluse_diagnostic`; it is written to
+    `traces.jsonl` (NOT `errors.jsonl`) so it never inflates /annunaki error
+    counts (#625).
     """
     record = {
         "timestamp": datetime.now(UTC).isoformat(),
@@ -127,7 +169,8 @@ def log_pretooluse_diagnostic(
         "command": command[:500],
         "diagnostic": diagnostic,
     }
-    append_jsonl_record(ERRORS_FILE, record)
+    # Benign forensic trace → traces.jsonl, not errors.jsonl (#625).
+    append_jsonl_record(TRACES_FILE, record)
 
 
 def log_posttooluse_dispatch(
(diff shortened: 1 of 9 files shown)


parametrization:
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved

Reviewed at head `95873f6` (base `deployments/phase-4/wave-2`). This cleanly closes #625. Approving — one non-blocking follow-up below, nothing that should hold the merge.

## What's right
- Writer split is clean: `log_posttooluse_dispatch` / `log_pretooluse_diagnostic` now append to `TRACES_FILE`, while `pretooluse_block` / `posttooluse_event` / the monitor's command-failure records stay in `ERRORS_FILE`. The genuine-vs-benign boundary matches the #625 rationale exactly.
- `TRACE_RECORD_TYPES` is a single source of truth owned by the writer (`annunaki_log`) and imported by the reader (`annunaki_parse`), with a sensible vendored-without-hooks fallback — and a test (`TraceTypeSourceOfTruth`) that locks the two against drift. That is the right shape; it's the thing that rots otherwise.
- The reader-side guard is the part I cared most about as Release Coordinator: even a HISTORICAL mixed log (a `.bak` rotation replayed, a not-yet-cleared live log) is counted correctly because `iter_records` skips trace-typed records by default. New logs clean by construction, old logs cleaned at read time — both effects compose, and the committed `errors.jsonl` is shipped empty (blob e69de29) so there's no stale pollution riding along.
- Both skills (`annunaki`, `annunaki-attack`) are migrated to the shared reader, with the explicit "do NOT process traces.jsonl" instruction. `.gitignore` covers `traces.jsonl` AND its `.bak.*` rotations, consistent with the existing errors.jsonl.bak handling.
- Tests are thorough and bidirectional: routing in both directions, blank/corrupt/non-dict-line skipping, `include_traces=True`, count semantics, and the drift-lock. CI all green (11/11), single roster author (Aino Virtanen).

## Non-blocking follow-up (do NOT block on this)
There is one remaining in-tree reader of `errors.jsonl` that was NOT migrated to the shared reader: `.claude/hooks/session_start.py` — its `_annunaki_error_count()` does a raw `len(text.splitlines())` with no trace filtering. That count is what `/session-start` surfaces to the operator every session. Post-merge it is correct by construction (the writer no longer routes traces into errors.jsonl), so this is not a live over-count and not a blocker. But it is the one surface that still lacks the historical-mixed-log defense this PR deliberately built for the skills: if someone replays an old `.bak` or a stale pre-#625 log into place, `/annunaki` would count it correctly while `/session-start` would over-count. Recommend a small follow-up routing `_annunaki_error_count()` through `annunaki_parse.count_errors()` so every reader shares the one filter. (`suggest_generic_prompt.py` also names errors.jsonl, but only as a skip-pattern for edit suggestions — not a counter — so it needs no change.)

TechDebt: session_start.py `_annunaki_error_count()` still raw-counts errors.jsonl lines instead of using annunaki_parse.count_errors(); harmless post-merge but lacks the historical-mixed-log defense the shared reader provides — fold into a #625 follow-up.

Approved.

parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved

Reviewed the diff at head 95873f6 (code, not PR-body framing): the new `annunaki_parse.py` reader, the `annunaki_log.py` writer split, both SKILL.md updates, the three test files, and `.gitignore`. This cleanly closes #625 — the 76% benign-trace over-count — at both the writer and reader layers. All 11 CI checks green. Approving.

What I verified end-to-end:

- Writer routing is correct. `pretooluse_diagnostic` and `posttooluse_dispatch` now append to `TRACES_FILE`; `pretooluse_block`, `posttooluse_event`, and the annunaki_monitor command-failure record still go to `ERRORS_FILE`. `TRACE_RECORD_TYPES` is the single source of truth and the reader imports it from the writer, so the two cannot drift (with a vendored-without-hooks fallback frozenset). The split is the actual fix, not cosmetic.
- No real error is silently filtered. I checked `annunaki_monitor.py` (unchanged, not in the diff): its command-failure record has NO `type` key, so `rec.get("type")` is None, which is not in `TRACE_RECORD_TYPES` — genuine command-failures are still counted. The reader only drops the two named trace types.
- The reader is read-only and fail-safe: missing file yields nothing, blank lines and corrupt JSON and non-dict records are skipped, exit code is always 0. `--count` excludes traces (the genuine-error count), `--include-traces` is available for a trace-specific viewer.
- I reproduced the two SKILL recipes (the step-1 `-c` one-liner and the `--count` call) against a temp log mixing a dispatch trace, a diagnostic trace, a blank line, a corrupt line, a `pretooluse_block`, and a typeless command-failure. The shell quoting holds and the reader returns exactly the 2 genuine errors. The annunaki-attack SKILL also correctly instructs operators NOT to process `traces.jsonl`.
- Test updates are right: both `test_post_dispatcher` and `test_enforce_librarian_consulted` now redirect BOTH `ERRORS_FILE` and `TRACES_FILE` to temp paths (so no stray write leaks to the real log) and read the correct stream — dispatcher reads traces-only, librarian reads the union since a block writes a `pretooluse_block` to errors AND a `pretooluse_diagnostic` to traces. New `test_annunaki_parse.py` covers the reader.
- `.gitignore` correctly excludes `traces.jsonl` and its `.bak.*` rotations as high-churn machine-generated output, consistent with the existing `errors.jsonl.bak.*` precedent.

One small, non-blocking note for future polish (NOT a change request): the reader's `--count` path ignores `--include-traces`, so `--count --include-traces` silently returns the genuine-error count rather than the all-records count. That is the desired default behavior and the flag combination is unlikely, but a one-line guard (or having `--count` honor `--include-traces`) would remove the surprise if a trace-debugging caller ever wants a total. Filing as tech-debt only.

TechDebt: annunaki_parse.py `--count` ignores `--include-traces` (always returns the genuine-error count); harmless today but `--count --include-traces` is a silent no-op combination — make `--count` honor the flag or reject the combo.

Issue #455, opened 2026-05-17
tech-debt(hook-21): parse_wave_label_change silent-skips on multi-cmd Bash (for-loops, &&-chains)


parametrization:
## Summary

Hook 21 (`post_label_change_wave_field_sync.py`) silently no-ops when a single Bash tool call contains multiple `gh issue edit --add-label "p{N}-wave-{M}"` commands (for-loops, `&&`-chained edits, sequential newline-separated edits). The hook's `parse_wave_label_change` parser returns None on the whole multi-cmd command string, so zero Wave fields get set despite all the labels landing on origin.

## Evidence

P3W11 batch resume session 2026-05-17 — fired 5 deploy issues in one Bash for-loop:

```bash
for N in 144 156 199 200 231; do
  gh issue edit $N --repo noorinalabs/noorinalabs-deploy --add-label "p3-wave-11"
done
gh issue edit 86 --repo noorinalabs/noorinalabs-deploy --add-label "p3-wave-11" --remove-label "p3-wave-1"
```

Post-fire state (separate-Bash query, no race):
- All 6 issues had `p3-wave-11` label applied at origin (`gh` succeeded for each)
- 0 of 6 had Wave field set to P3W11 (all stayed null or at pre-existing P3W1)
- 0 annunaki entries from `post_label_change_wave_field_sync` for this Bash call
- Direct invocation of `parse_wave_label_change` on the for-loop command string confirms: returns `None`. Same string but isolated to a single command parses correctly.

When the same issues were fired individually (one `gh issue edit` per Bash tool call), Hook 21 fired correctly and set Wave field on all of them.

## Why this matters

- Batch operations (the natural shape for wave-kickoff and bulk relabeling) silently lose Hook 21 coverage
- The class is identical to #448 and the harness-PostToolUse misdiagnosis in #453 — silent-skip on a multi-cmd shape
- Companion to #451 (skip_no_* logging gap): even after #451 lands, a multi-cmd parser-skip would NOT log because the parser returns None BEFORE reaching the skip-classification branches

## Acceptance Criteria

- [ ] Extend `parse_wave_label_change` to either:
  - (a) Iterate over `;`/`&&`/newline-separated commands within the input string and return a LIST of `WaveLabelChange` objects (caller in `check()` then iterates and dispatches a mutation per entry), OR
[...]

PR #658, opened 2026-06-13
fix(hooks #650): resolve ambient repo for --repo-less wave-label edits

### LINKED ISSUE #650: hook(post_label_change_wave_field_sync): parser skips multi-command Bash blocks — wave-label change in a `;`-chained command leaves board Wave field unsynced
## Annunaki capture (P4W5 retro, 2026-06-13)

The `post_label_change_wave_field_sync` PostToolUse hook emitted `skip_parser_returned_empty` when a canonical wave-label change was issued inside a **multi-command Bash block**:

```
gh issue comment 601 ... ; echo ... ; gh issue edit 601 --remove-label "p4-wave-5" ; echo ... ; gh issue view 601 ...
```

Hook error line:
> `skip_parser_returned_empty: command contains 'gh issue edit' and a canonical wave-label flag but parser extracted no changes. Likely an unhandled multi-cmd shape (for-loop with unexpanded vars, etc.); investigate.`

## Consequence

The Wave **field** on project 2 for the affected issue (main#601) was **not synced** when its `p4-wave-5` label was removed — leaving the board's derived Wave projection potentially stale (still "Wave 5") while the canonical label state is correct (no wave label). `/board-audit` repairs this, but the auto-sync silently no-op'd.

## Fix direction

Make the parser handle the multi-statement shape: split the Bash command on `;`/`&&`/newlines and scan each segment for a `gh issue edit … --add-label/--remove-label <p{N}-wave-{M}>` invocation, rather than parsing the command as a single statement. Mirror whatever segment-splitting the merge-gate parser uses. Add a fixture for the `comment ; edit --remove-label ; view` shape.

## Severity

Low — board Wave field is a derived projection (labels are canonical per main#199) and `/board-audit` reconciles it. But a silently-skipped sync defeats the hook's purpose. Process-hardening follow-up.

## Provenance
Annunaki monitor capture during P4W5 wrapup (#601 → Phase-5 re-disposition). Not a wave blocker.

parametrization:
## Summary

Closes #650.

`post_label_change_wave_field_sync` emitted `skip_parser_returned_empty` for a
canonical wave-label change issued inside a multi-command Bash block during
P4W5 wrap-up, so the project-2 Wave field silently went unsynced.

**Root cause (NOT the multi-command shape).** The `;`/`&&`/newline segment
splitting in the shared parser already worked. The real defect was in
`_wave_label_parse._parse_edit_segment`: it **required a `--repo` flag**. The
captured command —
`gh issue comment 601 … ; … ; gh issue edit 601 --remove-label "p4-wave-5" ; … ; gh issue view 601`
— was run from inside `noorinalabs-main` with **no `--repo`** (gh resolves the
repo from the ambient git context). The parser dropped the change → empty list
→ `skip_parser_returned_empty`. A clean multi-command block *with* `--repo`
parsed fine; the only differentiator was the missing `--repo`.

## Fix (shared parser + both consumers benefit)

- **`_wave_label_parse.py`** — `WaveLabelChange.repo` is now `str | None`;
  `_parse_edit_segment` parses the edit even when `--repo` is omitted
  (`repo=None`). Anchored `^p\d+-wave-\d+$` semantics and the suffixed-label
  exclusion are unchanged.
- **`_shell_parse.py`** — new `resolve_repo_short_name(input_data, *, git_runner=None)`
  recovers the repo name from the invocation cwd's `origin` remote (scp + https
  forms, `.git` stripped), mirroring gh's own `--repo`-less resolution. Uses
  `resolve_invocation_cwd` so a worktree subagent's real dir is honoured (#521).
- **`post_label_change_wave_field_sync.py`** + **`post_wave_kickoff_comment.py`**
  — when a parsed change/apply carries no `--repo`, resolve the ambient repo
  from cwd before the GraphQL/`gh` call; emit a `skip_no_repo_context` annunaki
  event (not a silent skip) if it can't be resolved. The between-wave relabel
  filter and compound add+remove (post-edit-state-wins) semantics are preserved.

The `skip_parser_returned_empty` log path is retained for the cases it is
actually meant to catch (unexpanded `$VAR` issue numbers, tokenize failures),
not for missing-`--repo`.

## Test Plan

- `test_shell_parse.py` — `ResolveRepoShortNameTests`: scp/https/`.git`/
  trailing-slash forms, `None`/empty runner, invocation-cwd recovery.
- `test_post_label_change_wave_field_sync.py` — `AmbientRepoResolutionTests`:
  the **exact #650 reproducer** (`comment ; echo ; edit --remove-label ; echo ;
  view`, no `--repo`) → Wave field **cleared**; single/multi no-`--repo` resolve;
  resolve-once for a shared cwd; explicit `--repo` does not invoke the resolver;
  unresolvable → `skip_no_repo_context` (+ logs). Updated the two #455/#463
  tests that encoded the old missing-`--repo`→skip behavior to an
  unexpanded-`$VAR` shape.
- `test_post_wave_kickoff_comment.py` — no-`--repo` parse returns `repo=None`;
  `check()` resolves + posts; unresolvable → `skip_no_repo_context`; explicit
  `--repo` does not invoke the resolver.

Local checks (pinned ruff 0.15.11):
- `ruff format --check` + `ruff check` over all 7 changed files — clean.
- `mypy --ignore-missing-imports` over the 4 changed hooks (+ `auto_add_issue_to_board`) — clean.
- `pytest` the three affected suites — **199 passed**. Full hooks+lib suite —
  **1356 passed**; the single failure (`test_ontology_tracker`
  `…worktrees_segment_not_substring_false_match`) is a **worktree-location
  false-fail** (REPO_ROOT resolves under `.claude/worktrees/`, so the test's
  synthetic path genuinely contains a `worktrees` segment); it passes from the
  main checkout / fresh CI checkout and is not in this changeset.
- `pre_commit_ci_sync.py .` — OK (mirror intact).

## Reviewers

Requesting **Weronika Zielinska** (peer) and **Nino Kavtaradze** (secondary).
The 2-reviewer `Approved` + `TechDebt:` trailer gate applies — both verdict
comments must carry a `TechDebt:` line.

### DIFF
--- a/.claude/hooks/_shell_parse.py
+++ b/.claude/hooks/_shell_parse.py
@@ -107,6 +107,19 @@ Public API
         cd targets are ambiguous (they'd be relative to the already-wrong
         stdin cwd) so they are ignored.
 
+    resolve_repo_short_name(input_data, *, git_runner=None) -> str | None
+        Resolve the GitHub repository NAME (e.g. `noorinalabs-main`) from the
+        invocation cwd's `origin` remote. This mirrors how `gh` itself
+        resolves a `gh issue edit/create ...` invocation that OMITS `--repo`:
+        it falls back to the ambient git context. Hooks that need the repo
+        name to drive a GraphQL/REST call (e.g. the Wave-field sync and the
+        kickoff-comment hooks) call this to recover the repo when the parsed
+        command carried no `--repo` flag. Returns the last path segment of
+        the `origin` URL with any trailing `.git` stripped, or None when the
+        cwd is not a git repo / has no origin / the runner fails. The
+        `git_runner(cwd) -> str | None` injection point lets tests avoid
+        shelling out (#650).
+
     is_shutdown_request_message(message) -> bool
         True only if `message` is a structured shutdown_request JSON
         (dict-form OR str-form parseable to a dict with type==
@@ -139,6 +152,7 @@ import json
 import os
 import re
 import shlex
+import subprocess
 from typing import Iterator
 
 # Shell control tokens that segment a compound command. Any of these,
@@ -509,6 +523,59 @@ def resolve_invocation_cwd(input_data: dict) -> str:
         if cd_target and os.path.isdir(cd_target):
             return cd_target
     return resolve_tool_cwd(input_data)
+
+
+def _default_origin_url_runner(cwd: str) -> str | None:
+    """Return the `origin` remote URL for the git repo at `cwd`, or None.
+
+    Shells out to `git -C <cwd> remote get-url origin`. Any failure (not a
+    git repo, no `origin` remote, git missing) yields None so the caller
+    treats the repo as unresolvable rather than crashing.
+    """
+    try:
+        result = subprocess.run(
+            ["git", "-C", cwd, "remote", "get-url", "origin"],
+            capture_output=True,
+            text=True,
+            timeout=10,
+        )
+    except (subprocess.SubprocessError, FileNotFoundError, OSError):
+        return None
+    if result.returncode != 0:
+        return None
+    return result.stdout
+
+
+def resolve_repo_short_name(input_data: dict, *, git_runner=None) -> str | None:
+    """Resolve the GitHub repo NAME from the invocation cwd's `origin` remote.
+
+    When a `gh issue edit/create` command omits `--repo`, gh resolves the
+    target repository from the ambient git context (the cwd's `origin`
+    remote). Hooks that need the repo name to drive a GraphQL/REST call must
+    mirror that resolution. Returns the last path segment of the `origin`
+    URL with any trailing `.git` stripped — for both scp-form and https-form
+    URLs:
+
+        git@github.com:noorinalabs/noorinalabs-main.git  -> noorinalabs-main
+        https://github.com/noorinalabs/noorinalabs-main   -> noorinalabs-main
+
+    Returns None when the cwd is not a git repo, has no `origin` remote, or
+    the runner otherwise fails. The cwd is resolved via
+    `resolve_invocation_cwd` so a worktree-subagent's real dir is used (the
+    `cd <dir> && ...` recovery path, #521).
+
+    `git_runner(cwd) -> str | None` is the injection point for tests; the
+    default shells out to `git -C <cwd> remote get-url origin`.
+    """
+    cwd = resolve_invocation_cwd(input_data)
+    runner = git_runner or _default_origin_url_runner
+    url = runner(cwd)
+    if not url:
+        return None
+    name = url.strip().rstrip("/").split("/")[-1]
+    if name.endswith(".git"):
+        name = name[: -len(".git")]
+    return name or None
 
 
 def is_shutdown_request_message(message) -> bool:
--- a/.claude/hooks/_wave_label_parse.py
+++ b/.claude/hooks/_wave_label_parse.py
@@ -40,7 +40,10 @@ Public API
 
         Result fields (shared with the plural form):
           repo          — `noorinalabs-<name>` short form (last path segment
-                          of `--repo owner/name` or `--repo=owner/name`).
+                          of `--repo owner/name` or `--repo=owner/name`), or
+                          None when `--repo` is OMITTED (in-repo invocation,
+                          ambient gh resolution — #650). The consuming hook
+                          resolves the None case from the invocation cwd.
           issue_number  — the bare positional issue number after `edit`.
           add_label     — the FIRST `--add-label "p{N}-wave-{M}"` value, or
                           None if no add operation present.
@@ -114,9 +117,15 @@ class WaveLabelChange:
     """Result of parsing a `gh issue edit ... --add-label|--remove-label` command.
 
     At least one of `add_label` / `remove_label` is non-None.
+
+    `repo` is the short repo name from `--repo owner/name` (e.g.
+    `noorinalabs-main`), or None when the command OMITS `--repo` and relies
+    on gh's ambient-git-context resolution. Consumers that need a concrete
+    repo (for a GraphQL/REST call) resolve the None case from the invocation
+    cwd via `_shell_parse.resolve_repo_short_name` (#650).
     """
 
-    repo: str
+    repo: str | None
     issue_number: str
     add_label: str | None
     remove_label: str | None
@@ -154,9 +163,17 @@ def parse_wave_label(value: str) -> tuple[int, int] | None:
 def _parse_edit_segment(rest: list[str]) -> WaveLabelChange | None:
     """Parse the rest of a tokenized `gh issue edit <num> ...` segment.
 
-    Returns the WaveLabelChange if the segment has issue_number, repo,
-    AND at least one canonical wave-label `--add-label`/`--remove-label`.
-    Otherwise returns None.
+    Returns the WaveLabelChange if the segment has an issue_number AND at
+    least one canonical wave-label `--add-label`/`--remove-label`. Otherwise
+    returns None.
+
+    `--repo` is OPTIONAL (#650): a `gh issue edit <num> --remove-label
+    "p4-wave-5"` run from inside the target repo carries no `--repo` and
+    relies on gh's ambient-git-context resolution. Requiring `--repo` here
+    silently dropped every such in-repo label edit (the change never reached
+    the field-sync hook → board Wave field went unsynced). When `--repo` is
+    absent the returned `repo` is None; the consuming hook resolves the
+    ambient repo from the invocation cwd.
     """
     if len(rest) < 3 or rest[0] != "issue" or rest[1] != "edit":
         return None
@@ -208,7 +225,7 @@ def _parse_edit_segment(rest: list[str]) -> WaveLabelChange | None:
             continue
         i += 1
 
-    if issue_number and repo and (add_label or remove_label):
+    if issue_number and (add_label or remove_label):
         return WaveLabelChange(
             repo=repo,
             issue_number=issue_number,
--- a/.claude/hooks/post_label_change_wave_field_sync.py
+++ b/.claude/hooks/post_label_change_wave_field_sync.py
@@ -125,11 +125,13 @@ import os
 import subprocess
 import sys
 import time
+from dataclasses import replace
 from pathlib import Path
 
 sys.path.insert(0, str(Path(__file__).resolve().parent))
 import re  # noqa: E402
 
+from _shell_parse import resolve_repo_short_name  # noqa: E402
 from _wave_label_parse import (  # noqa: E402
     WaveLabelChange,
     parse_wave_label,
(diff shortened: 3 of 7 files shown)


parametrization:
Requestor: Weronika Zielinska
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Reviewed at confirmed head SHA `6623cfdc486b3dc09eac5db86a35f00d4b8a8eb0` (re-fetched via `gh pr view --json headRefOid` — unchanged from the brief; base `deployments/phase-4/wave-7`). All 8 CI checks green. Read the diff at HEAD from origin contents API, not the local clone.

**Diagnosis is correct — the issue framing WAS a misdiagnosis, verified not rubber-stamped.** The issue proposed "split the Bash command on `;`/`&&` — mirror the merge-gate segment-splitting." That splitting already exists: `parse_wave_label_changes` calls `iter_command_segments(tokens)`, which splits on `;`/`&&`/`||`/`|`, and the #455 multi-cmd suite (`test_semicolon_chain_dispatches_per_segment`, etc.) exercises it. So the reproducer's segment containing `gh issue edit 601 --remove-label "p4-wave-5"` WAS reaching `_parse_edit_segment` intact — the drop happened *inside* that function, which required a `--repo` flag and returned None when it was absent → empty list → `skip_parser_returned_empty` → Wave field unsynced. The issue's proposed `;`-split fix would have been a no-op against the real drop point. The root fix (`WaveLabelChange.repo` → Optional + ambient recovery) addresses the actual cause. Correct-over-expedient, confirmed by reading the parser path rather than the PR framing.

**`resolve_repo_short_name` faithfully mirrors gh's ambient resolution.** Verified against `_shell_parse.py` and its tests:
- scp-form (`git@github.com:noorinalabs/noorinalabs-main.git`) → `rstrip("/")` + `split("/")[-1]` + `.git` strip → `noorinalabs-main`. Covered by `test_scp_form_url`.
- https-form with/without `.git` and with trailing slash → covered by `test_https_form_url_no_dotgit`, `test_https_form_url_with_dotgit`, `test_trailing_slash_tolerated`.
- None on no-origin / non-repo / runner failure: `_default_origin_url_runner` returns None on non-zero rc or `SubprocessError/FileNotFoundError/OSError`, and `resolve_repo_short_name` short-circuits to None. Covered by `test_runner_returns_none_yields_none` / `_empty_`.
- Resolves via `resolve_invocation_cwd` (the `cd <dir> &&` #521 recovery path), NOT a hardcoded dir — `test_runner_receives_invocation_cwd` asserts the runner is called with the recovered cwd. A git worktree shares its parent's `origin`, so `git -C <worktree> remote get-url origin` returns the correct repo; the resolver cannot and should not out-resolve gh itself (if cwd is the wrong repo, gh would mis-target identically — mirroring is the right semantic).

**Between-wave relabel filter and compound add+remove semantics preserved.** The two consumers keep their distinct, correct semantics:
- `post_wave_kickoff_comment.parse_label_apply_command` still returns None when `remove_label` is also present (#467 carry-forward filter intact) — a relabel is not an initial kickoff.
- `post_label_change_wave_field_sync._apply_one_change` keeps POST-edit "add wins" (`target_label = add_label or remove_label`, `action_kind = "set" if add_label else "clear"`). `test_compound_add_and_remove_post_edit_state_wins` pins it.

**`skip_no_repo_context` is LOGGED, never silently dropped.** Both hooks call `log_posttooluse_event` on the unresolvable-ambient path before returning `{"action": "skip_no_repo_context"}`. `test_unresolvable_repo_logs` asserts the annunaki entry fires. This is the whole point of #650 — ending the silent no-op — and it lands.

**Reproducer asserts the field is actually cleared.** `test_exact_issue_650_reproducer_clears_field` runs the exact `comment ; echo ; edit --remove-label ; echo ; view` shape (no `--repo`), and asserts `action == "cleared"`, `repo == "noorinalabs-main"` (ambient-resolved), `issue == "601"` — i.e. the field is synced/cleared rather than `skip_parser_returned_empty`. `test_multi_cmd_no_repo_resolves_once` confirms the shared-cwd single-resolve, and `test_explicit_repo_does_not_invoke_git_runner` confirms `--repo` short-circuits ambient resolution.

**Pre-existing false-fail confirmed unrelated.** `test_ontology_tracker…worktrees_segment_not_substring_false_match` lives in the ontology-tracker test surface, which is not among the 7 files this PR touches; CI is green (Pytest job SUCCESS), consistent with it only failing when REPO_ROOT sits under `.claude/worktrees/`. Out of scope for this changeset.

Mypy/ruff/format/sync-drift all green; injection points (`git_runner`, `auth_status_runner`, `graphql_runner`) keep the new path unit-testable without shelling out. Clean fix, well-scoped, strong test coverage. Approving.

## Throughline observations

Boundary-class + promotion-candidate: this PR is the latest in a recurring lineage — **hook matchers that derive repo/identity/intent from a raw Bash command keep under-handling the in-repo/ambient form that `gh`/`git` resolve implicitly.** The chain is `_shell_parse` consolidation (#226 #227 #223 #216 #188 #189 #144) → #455 multi-cmd → #521 cwd-anchor → now #650 ambient `--repo`. Each fix has correctly *mirrored the real tool's resolution* rather than inventing parallel logic — a healthy convergence, and `resolve_repo_short_name` extends it cleanly.

One concrete sibling gap surfaces from this review (NOT a blocker for this correctly-scoped EDIT-surface PR, but worth a follow-up issue): `_wave_label_parse.parse_wave_label_create` — the CREATE-surface parser feeding Hook 13 (`auto_add_issue_to_board`) — still gates on `if wave_label and repo:` (line ~969), so a `gh issue create --label "p4-wave-7"` run **in-repo without `--repo`** drops the wave label the same way #650's EDIT path did, leaving the board's Wave field unset at create time. That is the exact same ambient-repo class this PR closes for the EDIT surface. Recommend a follow-up to apply the same `repo`-Optional + `resolve_repo_short_name` recovery to the create path (board-audit is the current compensating control, so it's low-severity, but it's the same root cause). Promotion candidate: a small invariant test asserting "every `gh issue create/edit` wave-label parser tolerates an omitted `--repo`" would prevent the class from recurring on the next surface.

parametrization:
Requestor: Nino Kavtaradze
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Reviewed at head SHA `6623cfdc` (re-confirmed via `gh pr view 658 --json headRefOid` = `6623cfdc486b3dc09eac5db86a35f00d4b8a8eb0`; base `deployments/phase-4/wave-7`). Read all four logic files at that ref from origin (`gh api contents?ref=6623cfdc`), not the local clone. All 8 CI checks green.

**Verdict: Approved — no injection surface, safe failure mode, no security regression.**

Security/safety angle (the new ambient-repo resolution):

- **No command-injection surface.** `_default_origin_url_runner` invokes `subprocess.run(["git", "-C", cwd, "remote", "get-url", "origin"], capture_output=True, text=True, timeout=10)` — argv form, `shell=False` (default), no `shell=True` / `os.system` / f-string-into-shell anywhere in the diff (grep-confirmed; the only hit was a docstring prose false-match `subprocess calls that`). `cwd` is passed as the `-C <cwd>` argument, never interpolated into a shell string. Exceptions (`SubprocessError`/`FileNotFoundError`/`OSError`) and non-zero returncode → `None`. `timeout=10` bounds it.

- **`cwd` provenance is the hook's own invocation context, not free attacker data.** It comes from `resolve_invocation_cwd(input_data)`: the leading `cd <dir>` recovered from the command string (honored ONLY if absolute AND `os.path.isdir` — relative targets ignored), else stdin `cwd` → `os.getcwd()`. This deliberately mirrors how `gh` itself resolves a `--repo`-less `gh issue edit` from ambient git context, so the hook syncs the SAME repo gh edited. Worst-case influence (pointing cwd at a different real repo) yields a wrong/None repo name, still gated downstream by auth-scope + project-ids + item-on-board — no ungated write.

- **A hostile `origin` URL can do no worse than a wrong/None name.** `name = url.strip().rstrip("/").split("/")[-1]`, `.git`-stripped — pure string manipulation. That name is then used only as (a) a **parameterized** GraphQL variable (`gh api graphql -f repo=<name>` — a typed variable value, never concatenated into the query body → no GraphQL injection), and (b) an argv element `--repo noorinalabs/<name>` (no shell → a malformed name just makes gh exit non-zero → graceful skip). The kickoff hook additionally sanitizes the name before filename use (`re.sub(r"[^A-Za-z0-9_-]", "_", repo)`), so a `../`-style name can't traverse out of `.claude/scratch/`.

- **Failure mode is safe.** Unresolvable ambient repo → `skip_no_repo_context` annunaki-logged + the change dropped (`continue`); if every change drops, `check()` returns `{"action": "skip_no_repo_context", ...}`. No crash, no wrong-repo write. Both `main()`s wrap `check()` in `try/except` and `sys.exit(0)` — PostToolUse stays advisory. The `assert change.repo is not None` in `_apply_one_change` is defensive only: the resolution loop guarantees every change reaching it has a concrete repo (None changes are replaced-or-dropped before dispatch), so even under `python -O` (asserts stripped) the invariant holds by construction.

- **No security regression in the two hooks.** They only sync a project-2 Wave single-select field / post a charter-format kickoff comment. The change makes `repo` Optional and fires on the previously-silently-dropped in-repo (`--repo`-less) edits — the intended #650 fix — with all pre-existing gates (auth scope, project IDs, item-on-board, idempotency) unchanged.

Root-fix justification: the issue framing (`;`-splitting) was a misdiagnosis; the real defect was the parser REQUIRING `--repo` and dropping ambient-resolution commands. Making `repo` Optional + mirroring gh's own cwd-based resolution via the already-tested `resolve_invocation_cwd` primitive is the correct, minimal root fix, not scope creep.

## Throughline observations

Convergent-class, not standalone. This bug is the same SHAPE as the cross-repo cwd-anchor defects already catalogued inside `_shell_parse` (#144 `resolve_tool_cwd`, #521 `resolve_invocation_cwd`): a hook parser assumes an explicit repo signal where `gh`'s own UX makes it optional and falls back to ambient git context. The wave-label hook family (`post_label_change_wave_field_sync`, `post_wave_kickoff_comment`) inherited that wrong assumption by REQUIRING `--repo`. The fix correctly consolidates the ambient-repo resolution into a new shared primitive (`resolve_repo_short_name`) on top of `resolve_invocation_cwd` rather than open-coding it per hook — the right promotion direction.

Promotion candidate worth surfacing at retro: "any hook that derives GitHub repo IDENTITY from a `gh` command MUST resolve the `--repo`-less case from the invocation cwd (mirroring gh), never silently drop it." That is now encoded in `resolve_repo_short_name`'s docstring + the `_shell_parse` module contract, but it is not yet a charter rule or a lint/test gate — so the next hook added to this family could re-introduce the same silent-drop. Candidate for a charter note under the hooks section (memory `feedback_enforcement_hierarchy`: hook/test > charter; ideally a smoke-test asserting `--repo`-less wave-label edits resolve rather than no-op).

PR #678, opened 2026-06-14
fix(hooks): CREATE-path wave-label drop (#659) + validate_labels body over-match (#661)

### LINKED ISSUE #659: hook(parse_wave_label_create): CREATE path drops wave label when --repo omitted (in-repo gh issue create) — #650 sibling on the create surface
## Summary

Sibling gap to #650 (fixed in PR #658), on the **CREATE** surface. `parse_wave_label_create` (which feeds Hook 13, the label-CREATE→Wave-field sync) still gates on `if wave_label and repo:`. So an **in-repo** wave-labeled `gh issue create` issued **without `--repo`** (relying on gh's ambient git-context resolution) drops the wave label and the Wave field never syncs — the exact silent-no-op class #650 fixed on the EDIT path.

## Root cause (same shape as #650)

#658 made the EDIT path's `_parse_edit_segment` repo-Optional and added `_shell_parse.resolve_repo_short_name` to recover the ambient repo from the invocation cwd's `origin` remote (mirroring gh). The CREATE path was correctly left out of #658's scope (#658 is EDIT-surface only) but carries the identical repo-required gate.

## Fix direction

Apply the same pattern to the create path: make repo Optional in `parse_wave_label_create`, resolve the ambient repo via the now-shared `resolve_repo_short_name`, and log a `skip_no_repo_context` event (not a silent drop) when unresolvable. Add a reproducer test mirroring `test_exact_issue_650_reproducer_*`.

## Provenance

Surfaced independently by both reviewers (Weronika Zielinska, Nino Kavtaradze) on PR #658 (main#650) as a throughline/convergent-class observation (catalogued cwd-anchor lineage #144/#521 then #455 then #650). Flagged for dispositioning, not auto-scoped — P4W7 is a thin close-out; this is a Phase-5 candidate.

## Severity

Low — same as #650 (board Wave field is a derived projection; `/board-audit` reconciles). Process-hardening follow-up.

### LINKED ISSUE #661: hook(validate_labels): over-matches label-shaped tokens in issue body, false-blocks gh issue create — scope extraction to --label flags only
## Summary

The `validate_labels` PreToolUse hook (label-existence gate on `gh issue create`/`gh issue edit`) over-matches: it flagged a label-shaped token appearing in the issue **body text** as a label to validate, and blocked the `gh issue create` even though the actual `--label` flag named only `bug`.

## Reproducer (observed live, P4W7 2026-06-13)

Filing main#659 with a body that contained the literal template token `p` + `{N}` + `-wave-` + `{M}` (inside backticks, as documentation of a wave-label pattern) produced:

```
BLOCKED: The following label(s) do not exist in noorinalabs/noorinalabs-main: p{N}-wave-{M}`
```

The hook parsed a label-shaped substring out of the `--body-file` content (or the body string) instead of restricting to the values of `--label`/`--add-label`/`--remove-label` flags. The workaround was to reword the body to avoid the token — but the body is documentation, not a label.

## Root cause class

Same family as #650 / #659: a hook parsing a `gh` command extracts the wrong tokens because it doesn't scope to the actual flag values. Here the gate must validate ONLY the labels named by `--label`/`--add-label` flags, never label-shaped strings elsewhere in the command (especially `--body`/`--body-file` content).

## Fix direction

Restrict `validate_labels` label extraction to the values immediately following `--label`/`--add-label`/`--remove-label` (reuse the shared `_shell_parse` tokenizer + the `_wave_label_parse` flag-scanning approach). Add a fixture: `gh issue create --label bug --body '... p{N}-wave-{M} ...'` validates only `bug`.

## Provenance

Observed during P4W7 retro/close-out while filing #659. Annunaki captured 2 `validate_labels` events this wave. Process-hardening follow-up — Phase-5 candidate (not scoped into thin W7).

## Severity

Low-Medium — false-block on legitimate issue creation; worked around by rewording, but it's a real parser-scoping bug that will recur whenever an issue/PR body documents a label pattern.

parametrization:
## Summary

Two hook parser-scoping fixes in the same family as #650/#658 (a `gh` command parser extracts the wrong tokens because it does not scope to the actual flag values).

### #659 — CREATE-path wave-label drop
`parse_wave_label_create` gated on `if wave_label and repo:`, so an **in-repo** wave-labeled `gh issue create` issued **without `--repo`** (relying on gh's ambient git-context resolution) was dropped — the Wave-field sync never ran and the board Wave field stayed unset. This is the create-surface twin of the EDIT-path #650 silent-drop.

- `_wave_label_parse.py`: `WaveLabelCreate.repo` is now `str | None`; the create is emitted regardless of `--repo` presence.
- `auto_add_issue_to_board.py` (Hook 13 consumer): when the parsed `repo` is `None`, recover the concrete repo from the **created-issue URL** in PostToolUse stdout — the authoritative repo the issue actually landed in. (No cwd `resolve_repo_short_name` is needed on this surface, unlike the EDIT path, because PostToolUse already has the URL.) Defensive `skip_no_repo_context` log if the URL yields no repo.

### #661 — validate_labels over-matches body text
`extract_labels` fell back to a `(?:--label|-l)`-anchored regex over the **whole command** when shlex tokenization failed (e.g. an apostrophe such as "gh's" inside a single-quoted `--body`). That scooped label-shaped tokens out of `--body`/`--title` prose (e.g. a documented wave-label pattern in backticks) and **false-blocked** legitimate issue creation.

Fix: on tokenize failure, fail **open** (return `[]`) instead of over-matching. Without reliable token boundaries we cannot distinguish a real `--label` flag from one quoted inside body text; the label check is a best-effort pre-flight and `gh` validates labels server-side, whereas a false block stops valid work.

> Live confirmation: while committing this very change, the active (unfixed) hook blocked the commit because the heredoc commit message contained a documented `gh issue create ... --label` pattern plus an apostrophe — exactly the #661 failure mode.

## Tests
- `test_validate_labels.py::TokenizeFailureFallbackTests` — exact #661 reproducer, no body-token leak, end-to-end no-spurious-block.
- `test_auto_add_issue_to_board.py::CreateWithoutRepoFlagWaveLabelTests` + `ParseWaveLabelCreateTests` — #659 parser-level (`repo=None`) and end-to-end (repo recovered from URL, mutation fires).
- Full hook + lib suite green (the one unrelated `test_ontology_tracker` failure is an environmental artifact of running pytest from inside a `.claude/worktrees/` path — REPO_ROOT contains the `worktrees` segment; passes from a normal checkout / CI).
- ruff format + lint, mypy, and the pre-commit ⇄ CI sync-drift gate all clean.

Closes #659
Closes #661

### DIFF
--- a/.claude/hooks/_wave_label_parse.py
+++ b/.claude/hooks/_wave_label_parse.py
@@ -137,9 +137,18 @@ class WaveLabelCreate:
 
     Issue NUMBER is not present at parse-time (the create has not yet
     landed); the caller extracts it from PostToolUse stdout.
+
+    `repo` is the short repo name from `--repo owner/name` (e.g.
+    `noorinalabs-main`), or None when the command OMITS `--repo` and relies
+    on gh's ambient-git-context resolution (#659 — the create-surface sibling
+    of the EDIT-path #650 fix). Requiring `--repo` here silently dropped every
+    in-repo `gh issue create` (the create never reached the Wave-field sync →
+    the board Wave field went unset). When `--repo` is absent the consumer
+    resolves the concrete repo from the created-issue URL in PostToolUse
+    stdout, which is the authoritative repo the issue actually landed in.
     """
 
-    repo: str
+    repo: str | None
     add_label: str
 
 
@@ -309,6 +318,14 @@ def parse_wave_label_create(command: str) -> list[WaveLabelCreate]:
     only the FIRST wave-label value is captured (Hook 13's existing
     invariant — one wave label per issue).
 
+    `--repo` is OPTIONAL (#659): an in-repo `gh issue create --label
+    "p{N}-wave-{M}"` run without `--repo` relies on gh's ambient-git-context
+    resolution and carries no `--repo` token. Such a create still yields a
+    `WaveLabelCreate` (with `repo=None`); the consumer resolves the concrete
+    repo from the created-issue URL. Before this fix the `if wave_label and
+    repo:` gate dropped every in-repo create, leaving the board Wave field
+    unset — the create-surface sibling of the EDIT-path #650 silent-drop.
+
     Issue #450 use case: Hook 13 (`auto_add_issue_to_board`) needs to
     know the wave label at create-time so it can set the project board's
     Wave single-select field after adding the issue. The issue NUMBER
@@ -360,6 +377,11 @@ def parse_wave_label_create(command: str) -> list[WaveLabelCreate]:
                 continue
             i += 1
 
-        if wave_label and repo:
+        if wave_label:
+            # `repo` may be None (in-repo `gh issue create` without `--repo`,
+            # #659). Emit the create anyway; the consumer recovers the concrete
+            # repo from the created-issue URL. Gating on `repo` here silently
+            # dropped every in-repo wave-labeled create — the create-surface
+            # twin of the EDIT-path #650 drop.
             out.append(WaveLabelCreate(repo=repo, add_label=wave_label))
     return out
--- a/.claude/hooks/auto_add_issue_to_board.py
+++ b/.claude/hooks/auto_add_issue_to_board.py
@@ -247,19 +247,42 @@ def check(
         # parser may return multiple create-segments for compound commands;
         # pick the one whose repo matches the URL's repo to avoid mis-syncing
         # in `gh issue create --repo A ... && gh issue create --repo B ...`
-        # shapes.
+        # shapes. A create issued WITHOUT `--repo` (in-repo, ambient
+        # resolution — #659) parses with `repo=None`; it cannot match the URL
+        # repo by equality, so fall back to the first `repo is None` segment
+        # (the ambient create targets exactly the repo the URL reports) and
+        # finally `creates[0]`.
         url_repo_match = re.search(r"noorinalabs/([^/]+)/issues/", issue_url)
         url_repo = url_repo_match.group(1) if url_repo_match else None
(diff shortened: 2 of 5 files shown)


parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved

**Review: LGTM** — both parser-scoping fixes are correct at head; #659 makes `WaveLabelCreate.repo` Optional and emits the create regardless of `--repo` while the Hook 13 consumer recovers the concrete repo from the created-issue URL (sensible URL-match → first-None → creates[0] fallback, with a defensive skip-log when no repo is resolvable), and #661 fails open (returns `[]`) on shlex-tokenize failure instead of running a whole-command label regex that scooped label-shaped tokens out of body/title prose, each backed by a targeted regression class and a full green CI run.
Must-fix: None
TechDebt: None

The pre-existing `test_ontology_tracker` worktree-location artifact (REPO_ROOT containing a `worktrees` segment when pytest runs from inside `.claude/worktrees/`) is not a regression — it passes in CI and from a normal checkout — so noting only, not blocking.

parametrization:
Requestor: Nino Kavtaradze
Requestee: Aino Virtanen
RequestOrReplied: Approved

**Review: LGTM** — both parser-scoping fixes are sound from the enforcement lens; #661 fails open on a flawed best-effort pre-flight while gh's server-side label validation remains the real backstop, and #659 only broadens the board Wave-field sync (recovering the concrete repo from the authoritative created-issue URL) without weakening any gate.
Must-fix: None
TechDebt: None

PR #689, opened 2026-06-15
tooling(main#688): deterministic wave repo-iteration + counter helper (kill zsh word-split)

### LINKED ISSUE #688: tech-debt(process): deterministic wave repo-iteration + counter helper — zsh word-split breaks wrapup/kickoff bash loops
## Problem
The wave lifecycle skills (`/wave-wrapup`, `/wave-kickoff`, `/wave-scope`) iterate the in-scope repos and compute wave counters with hand-rolled **bash loops** like:

```bash
for R in $WAVE_REPOS_IN_SCOPE; do ... done      # reads repos_in_scope
PRS=$(for REPO in $REPOS; do gh pr list ...; done)
```

Under **zsh** (this org's shell — see memory `feedback_zsh_shell_environment`) unquoted `$var` does **NOT** word-split, so the entire list collapses into a single string. Symptoms observed in P5W4 `/wave-wrapup`:
- counter computation passed `"noorinalabs-isnad-graph noorinalabs-user-service ..."` as ONE repo → `gh` GraphQL "Could not resolve repository" → `merged PR count: 0`, `division by zero`.
- the same pattern silently produces wrong/zero counters that then need manual recompute (exactly the drift class `/wave-retro` Step 2.5 exists to catch).

A soft memory hasn't prevented recurrence (bit 3× in one wrapup). This needs to be **deterministic code**.

## Fix — `.claude/lib/wave_status.py` (Python CLI, zsh/bash-agnostic, gh via subprocess list-args, no shell word-split)
Subcommands:
- `repos <P> <M>` — print `wave_{M}_repos_in_scope` from cross-repo-status.json, one per line (deterministic emit; callers use `while IFS= read -r`).
- `merged-prs <P> <M>` — emit the wave's merged-PR set as JSON (repo, number, author, commit-author-name, headRefOid, mergedAt), filtered by `wave_{M}_kicked_off_at` (the cross-window filter from #423).
- `counters <P> <M> [--write] [--expect N]` — compute `final_pr_count` / `changes_requested_cycles` / `top_concentration_pct` from merged-prs; loud-fail on `--expect` mismatch; with `--write`, upsert via `.claude/lib/upsert_status_keys.py`.

All `gh` invocations use `subprocess.run([...], ...)` with explicit arg lists — never a shell string. Unit tests in `.claude/lib/tests/test_wave_status.py` with mocked subprocess; must reproduce P5W4 = 19 / 4 / 16.

## Wiring + sweep
- Replace `/wave-wrapup` Step 10.5 bash counter block with `python3 .claude/lib/wave_status.py counters {P} {M} --write`.
- Audit `/wave-wrapup`, `/wave-kickoff`, `/wave-scope` for every `for X in $VAR` repo loop and convert to `… | while IFS= read -r R` or the `repos` subcommand. Document the zsh-safe pattern in the charter (`feedback_zsh_shell_environment` → codified).

## Acceptance
- [ ] `wave_status.py` + tests green (ruff/mypy/pytest); reproduces 19/4/16 for P5W4.
- [ ] Step 10.5 uses the helper; no `for X in $VAR` repo loops remain in the three skills.
- [ ] Charter notes the zsh-safe iteration rule.

parametrization:
## Summary
Closes #688. Replaces the zsh-word-split-fragile bash loops in `/wave-wrapup`, `/wave-kickoff` and `/wave-scope` with a deterministic helper, `.claude/lib/wave_status.py`.

Under **zsh** (this org's shell — `feedback_zsh_shell_environment`) an unquoted parameter is **not** word-split, so `for R in $WAVE_REPOS_IN_SCOPE` collapsed the whole repo list into one bogus `gh --repo` value → "Could not resolve repository" → merged-PR count 0 → division-by-zero. It bit a single P5W4 `/wave-wrapup` three times.

## What changed
- **`.claude/lib/wave_status.py`** — `repos` / `merged-prs` / `counters` subcommands. EVERY `gh` call goes through `subprocess.run([...])` with an explicit **arg list** — never a shell string, never `shell=True` — so word-splitting is structurally impossible under any shell.
  - `counters` applies the `wave_{M}_kicked_off_at` cross-window filter (#423), supports `--write` (upsert via `upsert_status_keys.py`, compact-inline shape preserved) and `--expect N` (loud-fail / exit 1, no write, replacing the old Option-B cross-check).
- **`.claude/lib/tests/test_wave_status.py`** — 10 tests, mocked `subprocess`: `repos` emit, kickoff window filter, P5W4 fixture reproducing **19 / 4 / 16**, `--expect` mismatch exit 1, empty-wave → zeros (no div-by-zero), the jq double-backslash escape guard, `--write` round-trip, and a list-args/no-`shell=True` contract assertion.
- **`/wave-wrapup` Step 10.5** now calls `python3 .claude/lib/wave_status.py counters {P} {M} --expect {count} --write` (read-back-verify retained).
- **Sweep** — every parameter `for X in $VAR` repo loop across the three skills converted to here-string / process-substitution `while IFS= read -r` (kept in the current shell so `STRANDED` / `PUB_RED` / `BRANCH_SHA` assoc arrays persist) or quoted-array `"${arr[@]}"` form. `$(…)` command-substitution loops (which *do* split under zsh) were standardized to `while read` too.
- **`.claude/team/charter/skills.md`** — codifies the zsh-safe iteration rule + the deterministic-counter-helper mandate.

## Verification
- `python3 .claude/lib/wave_status.py counters 5 4 --expect 19` against **live** gh → `{final_pr_count: 19, changes_requested_cycles: 4, top_concentration_pct: 16}`, exit 0.
- `pre-commit run --all-files` (ruff-format, ruff-lint, actionlint) → pass; mypy → clean; sync-drift gate → OK.
- Full pytest suite: **1383 passed, 49 subtests passed**. The one failure under the suite — `test_ontology_tracker.py::ShouldSkipTopLevelWorktreesTests::test_worktrees_segment_not_substring_false_match` — is the documented cwd-only artifact (`project_ontology_tracker_test_worktree_location_fragile`): it false-fails only when pytest runs from inside `.claude/worktrees/` (REPO_ROOT inherits the `.worktrees` segment), passes from a normal checkout (verified) and in CI. `ontology_tracker` is untouched by this PR.

## Reviewers
Per charter: 2 reviewers (Wanjiku Mwangi + Santiago Ferreira).

### DIFF
--- a/.claude/lib/tests/test_wave_status.py
+++ b/.claude/lib/tests/test_wave_status.py
@@ -1,0 +1,274 @@
+"""Tests for wave_status — the deterministic wave repo-iteration + counter
+helper that replaces the zsh-word-split-fragile bash loops (main#688).
+
+Verifies:
+  1. `repos` emits wave_{M}_repos_in_scope one-per-line.
+  2. EVERY gh call goes through subprocess.run with a LIST arg vector and
+     never `shell=True` — the regression guard that makes word-splitting
+     structurally impossible.
+  3. merged-prs applies the wave_{M}_kicked_off_at cross-window filter (#423).
+  4. Counter math reproduces the P5W4 actuals 19 / 4 / 16.
+  5. `--expect N` exits 1 on a count mismatch.
+  6. An empty wave yields zeros (no division-by-zero — the original crash).
+  7. `--write` upserts the three canonical top-level keys through the shared
+     upsert_status_keys helper, preserving the file.
+"""
+
+from __future__ import annotations
+
+import json
+import sys
+import unittest
+from pathlib import Path
+from tempfile import TemporaryDirectory
+from types import SimpleNamespace
+from unittest import mock
+
+# Helper lives at .claude/lib/wave_status.py; this test is at
+# .claude/lib/tests/test_*.py. parent.parent reaches the lib root.
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
+
+import wave_status  # noqa: E402
+
+_REPOS = [
+    "noorinalabs-isnad-graph",
+    "noorinalabs-user-service",
+    "noorinalabs-deploy",
+    "noorinalabs-isnad-ingest-platform",
+]
+
+
+def _p5w4_prs() -> list[dict]:
+    """19 PRs whose top commit-author owns 3 (3/19 = 15.78 → 16%) and whose
+    ChangesRequested comments sum to 4 — the canonical P5W4 shape."""
+    authors = (
+        ["Aino Virtanen"] * 3
+        + ["Wanjiku Mwangi"] * 3
+        + ["Santiago Ferreira"] * 3
+        + ["Nadia Khoury"] * 3
+        + ["Imelda Okoro"] * 3
+        + ["Aisling Brennan"] * 3
+        + ["Tariq Mansour"] * 1
+    )
+    assert len(authors) == 19
+    cr_by_index = {0: 2, 5: 1, 10: 1}  # sums to 4
+    prs = []
+    for i, author in enumerate(authors):
+        prs.append(
+            {
+                "repo": _REPOS[i % len(_REPOS)],
+                "number": 100 + i,
+                "sha": f"sha{i:02d}",
+                "mergedAt": "2026-06-15T02:00:00Z",
+                "login": "octocat",
+                "commit_author": author,
+                "cr": cr_by_index.get(i, 0),
+            }
+        )
+    return prs
+
+
+class _FakeGh:
+    """A subprocess.run side_effect that emulates the gh calls wave_status
+    makes, driven by a flat PR fixture. Records every command vector so the
+    test can assert the list-args / no-shell contract."""
+
+    def __init__(self, prs: list[dict]) -> None:
+        self.prs = prs
+        self.calls: list[list[str]] = []
+
+    def __call__(self, cmd, *args, **kwargs):  # noqa: ANN001
+        # Contract guard: gh is always invoked with an explicit list vector and
+        # NEVER through a shell. This is the structural fix for main#688.
+        assert isinstance(cmd, list), f"gh called with non-list cmd: {cmd!r}"
+        assert cmd[0] == "gh"
+        assert kwargs.get("shell") is not True
+        self.calls.append(cmd)
+
+        if cmd[1:3] == ["pr", "list"]:
+            repo = cmd[cmd.index("--repo") + 1].removeprefix("noorinalabs/")
+            listed = [
+                {
+                    "number": p["number"],
+                    "headRefOid": p["sha"],
+                    "mergedAt": p["mergedAt"],
+                    "author": {"login": p["login"]},
+                }
+                for p in self.prs
+                if p["repo"] == repo
+            ]
+            return SimpleNamespace(stdout=json.dumps(listed), returncode=0, stderr="")
+
+        if cmd[1] == "api":
+            path = cmd[2]
+            parts = path.split("/")
+            if "/commits/" in path:
+                sha = parts[4]
+                name = next(p["commit_author"] for p in self.prs if p["sha"] == sha)
+                return SimpleNamespace(stdout=name + "\n", returncode=0, stderr="")
+            if path.endswith("/comments"):
+                number = int(parts[4])
+                cr = next(p["cr"] for p in self.prs if p["number"] == number)
+                return SimpleNamespace(stdout=f"{cr}\n", returncode=0, stderr="")
+
+        raise AssertionError(f"unexpected gh call: {cmd!r}")
+
+
+def _write_status(path: Path, *, repos: list[str], wave: str, kickoff: str | None) -> None:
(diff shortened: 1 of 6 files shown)


parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Reviewed the diff at head of `A.Virtanen/0688-deterministic-wave-counters`. This is the right structural fix for main#688 and I approve.

What I verified against the actual code:

- Determinism. Every gh invocation routes through `_run_gh`, which calls `subprocess.run(["gh", *args], capture_output=True, text=True, check=True)` — an explicit arg vector, no shell, no `shell=True`, no string interpolation into a shell. There is no word-splitting path anywhere, under any shell, which is the whole point. The test `_FakeGh.__call__` enforces this as a contract (asserts the cmd is a list, starts with gh, and `shell` is not True).
- Counter math matches the documented semantics. Top-concentration uses `Counter(commit_author_name).most_common(1)[0][1] * 100 / total` with half-up rounding via `math.floor(x + 0.5)` (3/19 = 15.78 reproduces 16); empty wave short-circuits to 0 before the divide, killing the original division-by-zero. Changes-requested sums the `RequestOrReplied ChangesRequested` jq-test over each PR's issue-comments timeline, and the doubled-backslash `\\s` is preserved (with a test pinning the literal sequence). The `wave_{M}_kicked_off_at` cross-window filter (#423) is preserved as `if kickoff and merged_at < kickoff: continue`, with the legacy no-key path correctly falling through unfiltered.
- Acceptance. `test_reproduces_p5w4_19_4_16` reproduces the P5W4 actuals 19 / 4 / 16. The `--expect` gate fires before any write — on mismatch it returns 1 and does NOT upsert (verified in `_cmd_counters` and by `test_expect_mismatch_exits_1` / `test_expect_match_exits_0`).
- Skill wiring. wave-wrapup Step 10.5 now invokes the helper with `counters {P} {M} --expect {count} --write` in place of the old hand-rolled bash; the kickoff/scope/wrapup loops are converted to here-string `while read`, process-substitution from the `repos` subcommand, and quoted array expansion — never `for X in $VAR`. charter/skills.md documents the helper as the wave-wrapup consumer and adds the zsh-safe-iteration rule with the promote-on-first-violation hook decision.
- Tests are meaningful, not trivially passing — they mock subprocess and exercise the real logic, the list-vector contract, the kickoff filter both ways, the empty-wave zeros, the expect gate both directions, and the `--write` upsert preserving pre-existing keys.

Nothing blocking and nothing worth filing. CI is green across all 13 checks.

parametrization:
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Reviewed at head `A.Virtanen/0688-deterministic-wave-counters` against the diff, with a release/operability lens (complementary to the TPM's correctness pass). CI is 13/13 green; nothing red or pending. Approving.

**Determinism + safety — the core fix holds.** Every `gh` invocation routes through `_run_gh`, which is `subprocess.run(["gh", *args], capture_output=True, text=True, check=True)` — an explicit arg vector, no shell string, no `shell=True`, nothing secret on argv. There is no word-splitting anywhere under any shell, which is precisely the structural property main#688 needed. The contract is locked by a test fake that asserts every recorded call is a list starting with `gh` and `shell is not True`.

**Errors fail loud, not into a zero count.** `check=True` means a per-repo `gh` failure (the old "could not resolve repository" mode) raises `CalledProcessError`, which `main()` catches and surfaces on stderr with exit 1 — it can no longer be silently absorbed into a merged-PR count of zero and a downstream division-by-zero. This is the operability inversion that matters: the failure mode flipped from silently-wrong to loudly-stopped.

**Merged-PR / wave→main correctness.** Base branch resolves to `deployments/phase-{P}/wave-{M}`, and the `wave_{M}_kicked_off_at` cross-window boundary is applied (`merged_at < kickoff` drops pre-kickoff PRs), with a documented no-filter fallback for legacy pre-`/wave-start` waves — faithful to the #423 partition fix.

**The expect-gate is correctly ordered.** On a `--expect` mismatch `_cmd_counters` returns 1 *before* the `--write` block executes, so a wrong tally cannot be written. The Step 10 human count flows in as `--expect`, the gate enforces equality, and the gh-derived `len(prs)` is what lands — equal by construction once the gate passes. That closes the wrapup operability hole cleanly.

**Idempotency.** The write delegates to the shared `upsert_status_keys` helper, preserving the compact-inline file shape, and the skill retains a jq read-back verify afterward. Half-up rounding reproduces both the canonical P5W4 row (3/19 -> 16) and the historical W9 row (4/6 -> 67), so prior counter history stays reproducible.

**Skill wiring.** wave-wrapup Step 10.5 now calls the helper with the expect+write flags; the fragile parameter loops in wave-wrapup / wave-kickoff / wave-scope are all replaced with here-string or process-substitution `while read` (current-shell so the accumulator arrays survive `done`), and the charter section plus consumer note are in place. Tests cover the no-shell guard, the kickoff filter, the 19/4/16 actuals, and the empty-wave zero edge.

No must-fix or follow-up items. No tech debt incurred.

Issue #684, opened 2026-06-15
tech-debt(ci-parity): cspell not mirrored in pre-commit AND unclassified by sync-drift gate → spelling reds only surface in CI


parametrization:
## Problem (surfaced P5W4 Batch 1: ig#1080 'Aqidah', deploy#460 'Webauth')

New domain vocabulary fails the CI **Spellcheck (cspell)** job but cannot be caught before push, because:

1. **cspell is in NO repo's `.pre-commit-config.yaml`.** Verified across isnad-graph / user-service / deploy / isnad-ingest-platform — they mirror ruff/mypy/pytest/actionlint/gitleaks/terraform, but not cspell. So there is no local spell guard at all.
2. **The sync-drift gate (`.claude/lib/pre_commit_ci_sync.py`) is blind to cspell.** `_KIND_PATTERNS` has no `cspell`/`spellcheck` kind, and the gate's contract is 'unknown tools are ignored' (module docstring). So a CI-enforced cspell job that pre-commit does NOT mirror produces **zero** drift signal — the exact divergence the gate exists to prevent (cf. `feedback_actionlint_needs_shellcheck`: local 'clean' must not diverge from CI).

Net: spelling errors are structurally un-catchable locally, and the machine-enforcement that should have flagged the missing mirror doesn't fire.

## Fix (two parts, both org-template + child rollout)

1. **Add a `cspell` kind to `_KIND_PATTERNS`** in `pre_commit_ci_sync.py` (patterns: `cspell`, `spellcheck`, `streetsidesoftware/cspell`), with unit tests. Mirror into each child repo's copy of the lib. This makes the gate fail any repo running a CI cspell job without a pre-commit counterpart.
2. **Add a cspell pre-commit hook** to `.pre-commit-config.yaml` in every repo that runs the CI Spellcheck job, scoped to the same file globs CI uses, pinned to the same cspell version. A new domain word then fails `pre-commit run` locally instead of only in CI.

## Related worktree-hook gap (same root: local ≠ CI)
[...]

PR #698, opened 2026-06-17
tech-debt(ci-parity): mirror cspell in pre-commit + classify in sync-drift gate (#684)

parametrization:
## What & why

Closes the spell-check blind spot raised by the owner 2026-06-14 and surfaced in P5W4 (ig#1080 "Aqidah", deploy#460 "Webauth"): **cspell** ran only in CI, so new domain vocabulary failed only *after* push — and the sync-drift gate (`pre_commit_ci_sync.py`) stayed silent because it never classified a `cspell` kind. An unclassified CI check is silently un-mirrored, which is the exact divergence the gate exists to prevent (cf. `feedback_actionlint_needs_shellcheck`: local "clean" must not diverge from CI).

## Changes

1. **`.claude/lib/pre_commit_ci_sync.py`** — add a `cspell` kind to `_KIND_PATTERNS` (patterns: `cspell`, `spellcheck`, `streetsidesoftware/cspell`). A CI Spellcheck job with no pre-commit mirror now produces **harmful drift** instead of zero signal. Docstring updated to state that closing a blind spot means adding the kind here.
2. **`.pre-commit-config.yaml`** — add a commit-stage `cspell` hook from `streetsidesoftware/cspell-cli` pinned to **v8.4.0**, the exact engine the CI `streetsidesoftware/cspell-action@…v8.4.0` step bundles. Scoped to the same authored-prose globs (`files:` regex of the action's glob set) and reusing `.cspell.json` via `--config`. `language: node`, so pre-commit provisions its own env — no system cspell required.
3. **Tests** — `CspellKindClassification` (action `uses:` ref, CLI `run:`, generic `spellcheck` word, pre-commit hook, drift-both-ways); extend `RealParentRepoHasNoDrift` to the **full** workflow set (the cspell job lives in `docs.yml`, not `ci.yml`) and add a guard that parent CI actually enforces cspell.

## Proof

- `python3 .claude/lib/pre_commit_ci_sync.py .` → `OK` (mirror present, exit 0).
- Strip the hook → gate flags drift `['cspell']`. CI kinds now include `cspell`.
- `pre-commit run cspell --all-files` → **Passed** against the live corpus (engine + config wiring verified end-to-end).
- `pytest .claude/lib/tests/test_pre_commit_ci_sync.py` → 29 passed.
- ruff format/lint + mypy clean on changed files.

## Scope note (org-template + child rollout)
This lands the parent (org-template) half. Per #684, the same `cspell` kind + hook must be vendored into each child repo that runs a CI Spellcheck job; that rollout is tracked by the umbrella #684 and not in this PR.

## Reviewers
Requested: **Wanjiku Mwangi** and **Santiago Ferreira** (charter 2-reviewer).

Refs #684

### DIFF
--- a/.claude/lib/pre_commit_ci_sync.py
+++ b/.claude/lib/pre_commit_ci_sync.py
@@ -23,11 +23,15 @@ step vs a `ruff` pre-commit `id:`). We normalize both sides to a small set of
 kind tokens so they compare:
 
     ruff-lint, ruff-format, mypy, pytest, eslint, typescript, prettier,
-    terraform-fmt, gitleaks, actionlint, astro-check, pip-audit, build
+    terraform-fmt, gitleaks, actionlint, astro-check, pip-audit, build,
+    cspell
 
 Unknown tools are ignored (neither side gates on a kind we can't classify),
 which keeps the gate conservative — it never fails on something it doesn't
-understand.
+understand. Closing a blind spot here means ADDING the kind to
+`_KIND_PATTERNS` (cf. `cspell`, #684) so a CI job that runs it starts
+demanding a pre-commit mirror — an un-classified CI check is silently
+un-mirrored, which is the exact divergence this gate exists to prevent.
 
 Input Language
 ==============
@@ -80,6 +84,13 @@ _KIND_PATTERNS: dict[str, tuple[str, ...]] = {
     "actionlint": ("actionlint",),
     "pip-audit": ("pip-audit", "pip audit"),
     "build": ("build-and-validate", "build-and-test", "npm run build"),
+    # `cspell` closes the spell-check blind spot (#684): the CI Spellcheck job
+    # runs the `streetsidesoftware/cspell-action` (or a `cspell` CLI), but until
+    # this kind existed an un-classified spell gate produced ZERO drift signal,
+    # so a repo could enforce cspell in CI with no pre-commit mirror — new
+    # domain vocabulary then failed only after push. Patterns cover the action
+    # ref, the bundled-CLI step name, and the generic job/step word.
+    "cspell": ("cspell", "spellcheck", "streetsidesoftware/cspell"),
 }
 
 # `ruff-lint` is a substring of nothing problematic, but `ruff format` also
--- a/.claude/lib/tests/test_pre_commit_ci_sync.py
+++ b/.claude/lib/tests/test_pre_commit_ci_sync.py
@@ -108,6 +108,84 @@ jobs:
         self.assertNotIn("ruff-lint", kinds)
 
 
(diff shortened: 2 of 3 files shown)


parametrization:
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Reviewed the full diff at HEAD and cross-checked the CI spellcheck job in docs.yml (unchanged by this PR) at the PR head sha. The new cspell kind patterns (cspell, spellcheck, streetsidesoftware/cspell) classify the CI action uses-ref, a bundled CLI step, and the generic word, and also classify the pre-commit cspell-cli hook — so a CI spell gate with no local mirror now produces harmful drift instead of the prior silence, which closes the exact blind spot the issue names. I verified the pre-commit files regex is coverage-equivalent to the CI action's authoritative files glob (charter md, skills md, ontology md, CLAUDE.md, ONBOARDING.md, docs md), the hook is pinned to the same v8.4.0 the action bundles, and both sides load .cspell.json — so a word that passes locally passes in CI and vice-versa. The added tests cover both classification expressions, the harmful-drift cases with and without a mirror, an all-workflows parity assertion run exactly as the CI gate runs it, and a guard that CI actually enforces the kind. CI is fully green. Approving.

parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #706

Approved. Reviewed the diff at HEAD. The cspell kind closes a real blind spot — before this, an unclassified spell gate produced zero drift signal, so a repo could enforce cspell in CI with no local mirror and new vocabulary only failed after push. The kind patterns cover the action ref, the bundled-CLI step, and the generic spellcheck word, and the tests prove harmful-drift detection both ways (CI cspell without a mirror flags drift; with a mirror it does not).

The pre-commit hook is pinned to cspell-cli v8.4.0 to match the streetsidesoftware/cspell-action@v8.4.0 the CI Spellcheck job runs, so the same engine and dictionaries (via the shared .cspell.json) apply on both sides. The files regex mirrors the CI action authored-prose glob set (charter, skills, ontology, root docs). The two repo-level guards are exactly the no-vacuous-green protection this needs — one globs every workflow just as the CI gate does (so docs.yml cspell+actionlint are covered), the other guards against the docs.yml spell job being renamed or removed without revisiting the mirror.

One non-blocking observation, documented and currently harmless: the regex uses a per-directory .*\.md where the CI glob is **/*.md; these agree for the present tree, but if a depth edge case ever appears the two could diverge slightly — worth a glance only if a cspell skip ever surprises someone. Parity is sound; approving.

PR #699, opened 2026-06-17
fix(wave-start §5a): mechanize per-phase wave-key reset — 3 cross-phase-collision defects (#683)

### LINKED ISSUE #683: tech-debt(wave-start §5a): per-phase wave-key reset never fires for same-number wave reuse (P4W4↔P5W4 collision) — 3 defects
## Summary

`/wave-start` § 5a ("Wave-key per-phase reset") did **not** fire for the P4W4 → P5W4 `wave_4_*` namespace collision, leaving 16 stale P4W4 operational keys in `cross-repo-status.json` that had to be reset by hand during `/wave-scope 5 4`. Three independent defects, each verified against `.claude/skills/wave-start/SKILL.md` (lines 119-171) and the live file state.

This operationalizes P5W3-retro **process carry-forward #4** ("Namespace/reset wave-status keys per phase"), which was filed as a soft proposal but never enforced.

## Defect 1 — phase guard can never be true for a later same-number wave

```
if [ "$HAS_PRIOR" = "yes" ] && [ "$FILE_PHASE" != "{P}" ]; then  # reset
```

The guard infers "wave_{M}_* is stale" from `current_phase != {P}`. But `current_phase` tracks the **latest** phase, not the phase that wrote the stale `wave_{M}_*` keys. By the time P5W4 starts, `current_phase` was advanced to `5` back at **P5W1 kickoff** — so `FILE_PHASE(5) == {P}(5)` and the condition is **false**. The stale keys come from P4W4 (written when `current_phase` was `4`, several waves ago), but nothing in the file still says `4`. The guard structurally cannot detect an intra-phase same-number reuse.

## Defect 2 — detection key name mismatch

```
print('yes' if d.get('wave_{M}_completed_at') or d.get('wave_{M}_wrapped_up_at') else 'no')
```

`/wave-wrapup` writes **`wave_{M}_wrapup_completed_at`**, not `wave_{M}_completed_at` or `wave_{M}_wrapped_up_at`. Confirmed in the live file: `wave_4_wrapup_completed_at` was present; both keys the detection probes for were **absent**. So even if Defect 1 were fixed, `HAS_PRIOR` evaluates to `no`.

## Defect 3 — `RESET_KEYS` list is incomplete

The enumerated 12-key `RESET_KEYS` list misses ≥10 keys P4W4 actually wrote and which bleed into P5W4:
`wave_{M}_active`, `wave_{M}_branches`, `wave_{M}_carry_forward`, `wave_{M}_counter_corrections`, `wave_{M}_annunaki_finding`, `wave_{M}_repos_descoped_during_wave`, `wave_{M}_stg_promotion_override_rationale`, `wave_{M}_stg_promotion_resolution`, `wave_{M}_wrapup_completed_at` (list has `_completed_at`), `wave_{M}_scope`, `wave_{M}_repos_in_scope`, `wave_{M}_meta_issue`.

`wave_{M}_branches` is the dangerous one: a stale P4 branch ref (`deployments/phase-4/wave-4`) would mislead `/wave-kickoff` Step 1 into thinking the phase-5 branches already exist.

## Impact

The collision is silent: no error, just stale cross-phase state that misleads `/wave-kickoff` (branch creation, reviewer slate) and `/wave-wrapup` (counter recomputation reads prior-phase PR counts). Recurs every phase that reaches wave-4+ of a prior phase's max wave (P4 reached wave-7; P5 will collide on waves 4-7).

## Proposed fix (durable)

Phase-namespace the keys outright — `p{P}_wave_{M}_*` (or a nested `phases.{P}.waves.{M}` object) — so cross-phase reuse is impossible by construction. This is the real form of retro proposal #4. Interim: fix Defect 2 (key name) + Defect 3 (key list) and replace the Defect-1 guard with a check that the **existing `wave_{M}_scope.phase`** (now written by `/wave-scope`) differs from `{P}`, rather than relying on the global `current_phase`.

## Manual remediation already applied

`/wave-scope 5 4` reset all 16 stale P4W4 keys and wrote P5W4 `wave_4_scope` (with a `phase: 5` discriminator), `wave_4_repos_in_scope`, `wave_4_scope_reconciled_at`. Verified: diff scoped to `wave_4_*` + `last_updated` only; file is valid JSON. So P5W4 is unblocked; this issue is the durable fix.

parametrization:
## Summary

Fixes the `/wave-start` § 5a "per-phase wave-key reset" so a wave number reused across phases (P4W4 ↔ P5W4) reliably clears the prior phase's stale `wave_{M}_*` keys from `cross-repo-status.json`. Closes #683.

The hand-rolled bash in § 5a never fired this wave — 16 stale P4W5 `wave_5_*` keys had to be nulled by hand. Replaced with a mechanical, test-backed helper `.claude/lib/wave_key_reset.py`.

## The 3 defects addressed

1. **Guard could never fire (Defect 1).** The old guard inferred staleness from `current_phase != {P}`, but `current_phase` tracks the *latest* phase (advanced at this phase's W1 kickoff), not the phase that wrote the stale keys — so for an intra-phase same-number reuse it was structurally always false. Detection now keys off a phase stamp carried **inside the wave's own keys**: `wave_{M}_scope.phase` and the `phase-{X}` segment of `wave_{M}_branches.branch`. `current_phase` is never consulted. Correctly fires at a real boundary and is a no-op when `/wave-start` is re-run within the same phase (idempotent).
2. **Detection key-name mismatch (Defect 2).** Now moot — detection no longer probes any lifecycle-marker key (the old `wave_{M}_completed_at` / `_wrapped_up_at` vs the actually-written `_wrapup_completed_at` mismatch). It keys off phase stamps, which exist regardless of how far the prior wave progressed.
3. **Incomplete RESET_KEYS list (Defect 3).** Reset is **prefix-complete by construction** — every top-level `wave_{M}_*` key is removed, including the dangerous `wave_{M}_branches` (a stale `deployments/phase-4/wave-4` ref would mislead `/wave-kickoff` Step 1). No `wave_4` vs `wave_42` prefix bleed.

## Design notes

- **Removal, not null** — reuses `upsert_status_keys.remove_top_level_key` (the bare-key overwrite convention; phase-prefixing `p{P}_wave_{M}_*` was considered and **rejected in main#611** because it needs a coordinated read-contract change across every wave skill). Preserves the file's mixed compact-inline shape and JSON-validates before AND after.
- The issue suggested durable phase-namespacing as the "real" fix; that conflicts with the accepted main#611 ADR, so this implements the issue's interim path (which is the correct one given that ADR) — mechanized + tested rather than documented-only.
- `/wave-scope` Step 13 now stamps `phase: {P}` into `wave_{M}_scope` so the primary detection signal is reliably present and self-documenting (the orchestrator-built `WAVE_SCOPE_STRUCTURED` did not guarantee it). The `wave_{M}_branches.branch` stamp (always written by `/wave-kickoff`) is the corroborating signal that is reliable at any real phase boundary.

## Tests

`.claude/lib/tests/test_wave_key_reset.py` — 15 tests, one regression per defect (incl. the "current_phase already matches" case the old guard missed, detection without any lifecycle marker, and prefix-completeness with no sibling-wave bleed) plus end-to-end apply / no-op / dry-run.

## Local checks

- ruff-format `--check`, ruff-lint, mypy, cspell: clean on changed files.
- `pre_commit_ci_sync.py .`: `OK: pre-commit config mirrors all CI-enforced checks`.
- Full `.claude/lib/tests/ .claude/hooks/tests/` suite: 1398 passed. The single `test_ontology_tracker` failure is the documented worktree-location false-fail (main#686) — it passes in a normal checkout and in CI; verified `pytest …test_worktrees_segment_not_substring_false_match` passes from the primary checkout.

## Reviewers

Requested: **Aino Virtanen** and **Santiago Ferreira** (per charter 2-reviewer rule).

### DIFF
--- a/.claude/lib/tests/test_wave_key_reset.py
+++ b/.claude/lib/tests/test_wave_key_reset.py
@@ -1,0 +1,185 @@
+"""Tests for wave_key_reset — the mechanical /wave-start § 5a per-phase
+wave-key reset that replaces three defective hand-rolled bash probes (main#683).
+
+Each defect from the issue has a dedicated regression test:
+
+  Defect 1 — detection must fire for a SAME-PHASE re-use even when
+             ``current_phase`` already equals the phase being started (the case
+             the old ``current_phase != {P}`` guard could never catch), and must
+             key off the wave's own phase stamps, never ``current_phase``.
+  Defect 2 — detection must not depend on any lifecycle-marker key name
+             (``wave_{M}_completed_at`` etc.); a wave that never wrapped up but
+             carries a prior-phase scope stamp is still detected as stale.
+  Defect 3 — the reset surface is prefix-complete: EVERY ``wave_{M}_*`` key is
+             removed (incl. ``wave_{M}_branches`` — the dangerous omission),
+             and ``wave_{M2}_*`` siblings are never touched (no ``wave_4`` vs
+             ``wave_42`` prefix bleed).
+"""
+
+from __future__ import annotations
+
+import json
+import sys
+import unittest
+from pathlib import Path
+from tempfile import TemporaryDirectory
+
+# Helper lives at .claude/lib/wave_key_reset.py; this test is at
+# .claude/lib/tests/test_*.py. parent.parent reaches the lib root.
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
+
+import wave_key_reset  # noqa: E402
+
+
+def _p4w4_status() -> dict:
+    """A file that has advanced to phase 5 (``current_phase`` == 5) but whose
+    bare ``wave_4_*`` keys were all written by P4W4 — the exact collision state
+    from the issue. The phase stamps inside the keys say 4."""
+    return {
+        "current_phase": 5,
+        "wave_4_active": True,
+        "wave_4_scope": {"phase": 4, "wave": 4, "theme": "P4W4 theme"},
+        "wave_4_repos_in_scope": ["noorinalabs-isnad-graph"],
+        "wave_4_branches": {"branch": "deployments/phase-4/wave-4", "created_at": "x"},
+        "wave_4_meta_issue": "noorinalabs-main#400",
+        "wave_4_final_pr_count": 19,
+        "wave_4_wrapup_completed_at": "2026-05-01T00:00:00Z",
+        # An unrelated sibling wave that must never be touched by a wave-4 reset.
+        "wave_42_active": True,
+        "last_updated": "2026-06-16T00:00:00Z",
+    }
+
+
+class TestBranchPhase(unittest.TestCase):
+    def test_parses_phase_segment(self) -> None:
+        self.assertEqual(wave_key_reset.branch_phase("deployments/phase-4/wave-4"), 4)
+        self.assertEqual(wave_key_reset.branch_phase("deployments/phase-12/wave-3"), 12)
+
+    def test_none_for_unstamped_or_non_string(self) -> None:
+        self.assertIsNone(wave_key_reset.branch_phase("main"))
+        self.assertIsNone(wave_key_reset.branch_phase(""))
+        self.assertIsNone(wave_key_reset.branch_phase(None))  # type: ignore[arg-type]
+
+
+class TestStampedPhases(unittest.TestCase):
+    def test_collects_scope_and_branch_stamps(self) -> None:
+        status = _p4w4_status()
+        self.assertEqual(wave_key_reset.stamped_phases(status, 4), {4})
+
+    def test_scope_and_branch_can_disagree(self) -> None:
+        status = _p4w4_status()
+        status["wave_4_branches"]["branch"] = "deployments/phase-3/wave-4"
+        self.assertEqual(wave_key_reset.stamped_phases(status, 4), {3, 4})
+
+    def test_empty_when_no_stamped_keys(self) -> None:
+        self.assertEqual(wave_key_reset.stamped_phases({"current_phase": 5}, 4), set())
+
+    def test_ignores_current_phase(self) -> None:
+        # current_phase=5 must NOT leak into the stamp set (Defect 1 root cause).
+        self.assertEqual(wave_key_reset.stamped_phases(_p4w4_status(), 4), {4})
+
+
+class TestIsStaleReuse(unittest.TestCase):
+    def test_defect1_detects_reuse_when_current_phase_already_matches(self) -> None:
+        # current_phase == 5 == phase being started; the OLD guard
+        # (current_phase != P) would be FALSE → miss it. The stamp says 4.
+        status = _p4w4_status()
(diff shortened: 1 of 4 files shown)


parametrization:
Requestor: Santiago Ferreira
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: #705

Reviewed the full diff at HEAD. The new wave_key_reset.py fixes all three defects correctly. Detection keys off the phase stamps carried inside the wave's own keys — wave_{M}_scope.phase (written by /wave-scope) and the phase-{X} segment of wave_{M}_branches.branch (written by /wave-kickoff) — and never the global current_phase, so an intra-phase same-number reuse where current_phase already equals the starting phase is still caught (Defect 1, exercised by test_defect1_detects_reuse_when_current_phase_already_matches). Detection no longer depends on any lifecycle-marker key name (Defect 2, exercised with the marker deleted so only the scope stamp remains). The reset surface is prefix-complete by construction — every top-level wave_{M}_ key, with the trailing underscore preventing wave_42 bleed, and it now includes the previously-omitted wave_{M}_branches (Defect 3). Carrying two independent stamps is good defense for legacy data lacking scope.phase, since the branch ref still yields the owning phase. Removal reuses upsert_status_keys so the file's mixed shape is preserved with before/after JSON validation, and the apply, no-op, and dry-run paths are exercised against a real temp file rather than mocked. The wave-scope SKILL.md force-sets the scope phase so the stamp is always present going forward, and wave-start SKILL.md replaces the broken inline bash with the helper. CI is fully green.

One minor non-blocking note — the two-argument default-status CLI form isn't covered by a test, but argparse resolves the optional-positional disambiguation and the three-argument form is well tested. Approving.

parametrization:
Requestor: Aino Virtanen
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: none

Reviewed the diff at HEAD (no checkout). Detection correctly keys off phase stamps carried inside the wave's own keys — `wave_{M}_scope.phase` and the `phase-{X}` segment of `wave_{M}_branches.branch` via `stamped_phases`/`branch_phase` — and never consults the global `current_phase`, which is exactly the Defect-1 structural flaw (current_phase tracks the latest phase, so it could never witness an intra-phase same-number reuse). `is_stale_reuse` is idempotent and a no-op when the stamps already match the starting phase, and Defect 2 dissolves because detection no longer probes any lifecycle-marker name.

Reset is prefix-complete by construction — `stale_wave_keys` clears every top-level `wave_{M}_` key (the trailing underscore prevents the wave_4 vs wave_42 bleed), so `wave_{M}_branches` can no longer be the dangerous omission that misleads /wave-kickoff Step 1. It honors the main#611 bare-key overwrite ADR by removing rather than namespacing, reusing `upsert_status_keys._run_remove` — which I confirmed exists at HEAD with the `(path, keys) -> int` signature and a pre/post JSON-validation contract; the reset list is derived from the same on-disk dict so there is no spurious not-found error. SKILL.md §5a is wired dry-run then `--apply`. All 13 CI checks green. Approving.

Issue #442, opened 2026-05-16
tech-debt(audit-log): rename promotion_audit_log/wave-N.md → p{N}-wave-N.md (filename collision risk)


parametrization:
## Context

The `.claude/team/promotion_audit_log/` directory uses two inconsistent naming conventions:
- Phase-prefixed: `p3-wave-1.md`, `p3-wave-10.md` (newer)
- Bare-wave: `wave-3.md` through `wave-10.md` (older — P2W3 through P2W10)

The collision risk surfaced live during W10 retro (2026-05-16): `/promotion-audit` for P3W10 was about to overwrite the existing `wave-10.md` (P2W10 audit from 2026-04-30). Disambiguated by writing the P3W10 audit to `p3-wave-10.md` instead. PR #441 (W10 retro) ships under the new naming.

## Acceptance

Mechanical rename — 8 files:
- `wave-3.md` → `p2-wave-3.md`
- `wave-4.md` → `p2-wave-4.md`
- `wave-5.md` → `p2-wave-5.md`
- `wave-5-memory-classification.md` → `p2-wave-5-memory-classification.md`
- `wave-5-memory-classification-snapshot.md` → `p2-wave-5-memory-classification-snapshot.md`
- `wave-6.md` → `p2-wave-6.md`
- `wave-7.md` → `p2-wave-7.md`
- `wave-8.md` → `p2-wave-8.md`
- `wave-9.md` → `p2-wave-9.md`
- `wave-10.md` → `p2-wave-10.md`

(Verify each is P2-era by reading its header; `grep -l 'Phase 2' wave-*.md` should catch.)

Also update `/promotion-audit` SKILL.md Step 6 + Step 7 to write to `p{N}-wave-{M}.md` going forward (not `wave-{M}.md`). Currently the standalone log path is left as `wave-{N}.md` per the SKILL — needs an update.

## Sequencing

W11 backlog. Mechanical, ~30s of work. Single-reviewer-exception eligible (mechanical rename + literal one-line SKILL.md update).

## Promotion provenance

Surfaced 2026-05-16 during P3W10 retro standalone-log write. Filed per owner directive at W10 retro presentation.

PR #701, opened 2026-06-17
tech-debt(promotion-audit): canonical CLI driver wiring classify_* once (closes #690)

### LINKED ISSUE #690: tech-debt(process): /promotion-audit needs a canonical CLI driver (hand-rolled helper-calls mis-fired section/skill tiers at P5W4 retro)
## Problem

`/promotion-audit` ships pure `helpers.py` functions but **no deterministic driver script**. The orchestrator hand-rolls the classify_* call sequence inline each retro. At the P5W4 retro this produced **24 spurious AUTO** decisions in the section→skill / skill→hook tiers (signal mis-derivation: `count_skill_invocations(section.promoted_to, …)` over empty/garbage slugs), while the more-carefully-called memory→charter tier correctly returned **0 AUTO**. Emitting those 24 would have auto-generated 24 bogus charter/skill PRs.

## Fix

Add `.claude/skills/promotion-audit/run.py {wave}` (or a `python -m`) canonical driver that wires the helper calls + thresholds exactly once, with tests asserting the steady-state (~0 AUTO) on current repo state. The SKILL.md then invokes the driver instead of describing the call sequence in prose. Same determinism-over-hand-rolling principle as main#688 (`wave_status.py` killed the zsh-loop hand-rolling).

## Provenance
P5W4 retro promotion-audit (Step 7.5), 2026-06-16. Related: [[feedback_codify_determinism_on_shell_fragility]] / main#688.

parametrization:
## Summary

Adds a canonical CLI driver for the `/promotion-audit` skill so the
`classify_*` call sequence is wired in exactly one deterministic place
instead of being hand-rolled inline each retro.

**Why:** At the P5W4 retro the hand-rolled sequence mis-fired. The
section→skill tier derived its signal via
`count_skill_invocations(section.promoted_to, repo_root)` over an **empty**
`promoted_to` slug, and `git log --grep=/` matches ~every commit — so 24
not-yet-promoted skill-target sections each reported ~600 invocations,
crossed the threshold, and produced **24 spurious AUTO** decisions that
would have opened 24 bogus charter/skill PRs. The carefully-called
memory→charter tier correctly returned 0 AUTO. Same
determinism-over-hand-rolling principle as main#688 (`wave_status.py`).

## Changes

- **`run.py`** — single deterministic entrypoint. Wires `read_*` +
  `classify_*` + thresholds once; resolves the wave to canonical
  `p{N}-wave-{M}` (#442); pins the audit date to the wave boundary (never
  `datetime.now()`); renders the table + summary. `--json` emits
  machine-readable decisions to drive the skill's artifact-emission step.
  Makes no `gh` calls.
- **Correct slug resolution** — section→skill candidate slug is
  `promoted_to` (with `skills/` stripped) when already promoted, else
  `_slugify(heading)` (the prospective skill name, naturally ~0
  invocations). Never an empty slug.
- **Root-cause guard** in `helpers.count_skill_invocations`: blank slug → 0
  instead of the all-commits count.
- **SKILL.md** invokes the driver and adds a no-hand-roll directive instead
  of describing the call sequence in prose.
- **`tests/test_run.py`** — wave/date resolution, the empty-slug regression,
  the end-to-end "not-yet-promoted section stays KEPT" regression,
  determinism, and steady-state-through-the-driver (0 AUTO / 0 DECIDE).

## Verification

- `ruff format --check` + `ruff check` clean; `mypy` clean.
- Full skill suite: **99 passed** (`ENVIRONMENT=test pytest tests/`).
- Driver on current `p5-wave-5` state: **0 AUTO · 0 DECIDE · 242 KEPT · 18
  SUPERSEDED/ALREADY-PROMOTED** — re-run byte-identical.
- cspell clean on the edited `SKILL.md`.

## Related Issues

Closes #690.

## Review Checklist

- [ ] Driver wiring matches the SKILL.md tier table (signatures + signals)
- [ ] Empty-slug guard + heading-slug fallback correctly prevent the 24-AUTO mis-fire
- [ ] Wave-name + audit-date resolution stay deterministic (#442)

Co-Authored-By: Santiago Ferreira <parametrization+Santiago.Ferreira@gmail.com>

### DIFF
--- a/.claude/skills/promotion-audit/SKILL.md
+++ b/.claude/skills/promotion-audit/SKILL.md
@@ -24,37 +24,51 @@ Skill-to-hook **ALWAYS** produces a DECIDE-tier draft issue — never auto-appli
 
 ## Instructions
 
-### 1. Resolve wave name
+> **Do NOT hand-roll the `classify_*` call sequence.** Run the canonical
+> driver (`run.py`) — it wires the helper calls, slug resolution, and
+> thresholds in exactly one place. Hand-rolling the sequence inline is what
+> produced the P5W4 24-spurious-AUTO mis-fire (main#690): the section→skill
+> tier called `count_skill_invocations(section.promoted_to, …)` over an
+> empty `promoted_to` slug, and `git log --grep=/` matched ~every commit,
+> crossing the threshold for 24 not-yet-promoted sections. The driver
+> derives the candidate-skill slug from the heading for not-yet-promoted sections,
+> and `helpers.count_skill_invocations` now returns 0 for a blank slug.
 
-If invoked with no argument, read `cross-repo-status.json` for `current_wave`. Use that slug (e.g., `wave-9`) as the audit wave name. If the arg is provided, trust it.
+### 1. Run the canonical driver
 
-### 2. Gather inputs (all deterministic — helpers.py)
+```bash
+# Human-readable audit table + summary line:
+python3 .claude/skills/promotion-audit/run.py [wave]
 
-```python
-from helpers import (
-    read_all_memories,
-    read_all_charter_sections,
-    read_all_skills,
-    find_already_promoted_in_charter,
-    count_retro_citations,
-    count_skill_invocations,
-    classify_memory,
-    classify_section,
-    classify_skill,
-    render_audit_table,
-)
-
-memories  = read_all_memories(memory_dir)                  # list[Memory]
-sections  = read_all_charter_sections(charter_parent)      # list[Section] — only sections with a promotion-target marker
-skills    = read_all_skills(skills_dir)                    # list[Skill]
-already   = find_already_promoted_in_charter(charter_parent)  # set[str] — aggregates Promotion provenance: blocks AND <!-- Promoted from memory: X --> markers across all charter sub-docs (#283)
-# NOTE: `charter_parent` is the directory CONTAINING `charter/` (e.g. `.claude/team`),
-# NOT the `charter/` directory itself. Passing `.claude/team/charter` raises ValueError (#418).
+# Machine-readable decisions, to drive artifact emission in step 4:
+python3 .claude/skills/promotion-audit/run.py [wave] --json
 ```
 
-### 3. Classify each candidate (pure function)
+- **Wave resolution.** With no `wave` argument the driver reads
+  `current_phase` + `current_wave` from `cross-repo-status.json` and emits
+  the canonical phase-prefixed form `p{N}-wave-{M}` (e.g. `p5-wave-5`). A
+  bare `wave-{M}` or `{M}` arg is auto-prefixed; an already-canonical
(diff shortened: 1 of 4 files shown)


parametrization:
Requestor: Aino Virtanen
Requestee: Santiago Ferreira
RequestOrReplied: Approved
TechDebt: none

Reviewed the diff at HEAD (no checkout). The P5W4 24-spurious-AUTO regression is fixed at both layers — `_section_signal_slug` in run.py resolves the charter to skill signal slug to `promoted_to`'s last segment when promoted else `_slugify(heading)` (the prospective skill name, naturally ~0 invocations), so an empty slug is never passed downstream; and `count_skill_invocations` now guards a blank/whitespace slug to 0 at the root rather than the all-commits count that `git log --grep=/` produces. Defense in depth on the exact footgun.

Determinism holds — `run_audit` wires `read_*` + `classify_*` exactly once, the helpers sort by stable keys, and the audit date is pinned to the wave boundary from cross-repo-status.json (never `datetime.now()`), with the wave resolved to canonical `p{N}-wave-{M}`. SKILL.md now invokes `run.py [wave]` (and `--json`) and carries a no-hand-roll directive instead of re-describing the call sequence in prose, so the next retro cannot reintroduce the inline mis-fire. The driver makes no gh calls — artifact emission stays in the skill steps driven off `--json`. All 13 CI checks green. Approving.

parametrization:
Requestor: Wanjiku Mwangi
Requestee: Santiago Ferreira
RequestOrReplied: Approved
TechDebt: none

Approved. Reviewed the diff at HEAD. The empty-slug regression behind the P5W4 24-spurious-AUTO mis-fire is fixed at the root and at the call site: count_skill_invocations now returns 0 for a blank or whitespace slug (so git log --grep=/ can never match ~every commit again), and _section_signal_slug derives a prospective slug from the heading for not-yet-promoted sections (never empty) while stripping the skills/ prefix for promoted ones. The end-to-end test builds a fixture repo whose commits contain slashes and asserts the unpromoted skill-target section stays KEPT with zero AUTO and zero DECIDE — a faithful reproduction-and-guard of the exact mis-fire.

Determinism is enforced and tested: the audit date is pinned to the wave-boundary timestamp (kicked_off_at, then started_at, then scope_reconciled_at — never datetime.now()), run_audit reads inputs in sorted order and derives signals from git and on-disk files only, the JSON payload sorts keys, and a dedicated test asserts byte-identical output across two runs. Wave-name resolution returns the canonical phase-prefixed form and forbids bare output (#442), with full coverage. Folding the classify wiring into one driver and leaving the nondeterministic gh artifact-emission in the SKILL.md prose (driven off the --json output) is the right separation of concerns.

One non-blocking edge note: a promoted_to value with a trailing slash would split to an empty slug, but the helpers guard now returns 0 in that case so it degrades safely. CI is green; nicely done.

Issue #686, opened 2026-06-15
tech-debt(test): test_ontology_tracker worktree-location-fragile — REPO_ROOT resolves into .claude/worktrees when run from a worktree


parametrization:
## Problem
`test_ontology_tracker.py::ShouldSkipTopLevelWorktreesTests::test_worktrees_segment_not_substring_false_match` FALSE-fails when pytest runs from a worktree under `.claude/worktrees/`. `REPO_ROOT = Path(__file__).resolve().parent.parent.parent` resolves to the worktree root, which itself contains a `.claude/worktrees` path segment, so `_is_worktree_path()` returns True from REPO_ROOT's own location rather than from the fixture input.

## Evidence
- Green on a normal checkout (`/home/parameterization/code/noorinalabs-main` → 1 passed) and in CI.
- Surfaced by Aino Virtanen while running the full check-set in a worktree for charter PR #685 (she did NOT `--no-verify` past it — proved it environmental). Sibling of `feedback_ruff_parent_config_bleed_in_worktree`.

## Durable fix
Build the fixture path under a tmp/non-worktree base (monkeypatch `REPO_ROOT`) so the test is independent of where pytest is invoked from. This matters because team agents routinely run tests from `.claude/worktrees/` worktrees.

## Provenance
Aino Virtanen, P5W4 charter PR #685 full-check-set run. Memory: `project-ontology-tracker-test-worktree-location-fragile`.

parametrization:
## Recurred 2026-06-15 — second confirmation (PR #689, main#688)

Hit again while running the **full pre-push suite from a worktree** (`.claude/worktrees/0688-wave-counters`) for the deterministic-wave-counter helper (main#688, PR #689):

- Full suite from the worktree cwd: **1383 passed, 49 subtests passed**, only `ShouldSkipTopLevelWorktreesTests::test_worktrees_segment_not_substring_false_match` FALSE-failed (`AssertionError: True is not false` at `test_ontology_tracker.py:156`).
- Same test from the normal checkout root (`/home/parameterization/code/noorinalabs-main`, `git rev-parse --show-toplevel` = repo root, no `.worktrees` segment): **1 passed**.
- `ontology_tracker` untouched by PR #689 → confirms environmental, not a regression.
- Did NOT `--no-verify`; pushed PR #689's branch from the main checkout root (cwd without `.worktrees`) so the pre-push suite ran genuinely green.

[...]

Issue #663, opened 2026-06-14
charter/standards: shared-parser invariant for gh-command hooks (scope to flag values + resolve ambient repo, never silently drop) — owner-adopted P4W7


parametrization:
## Owner-adopted process change (P4W7 retro, 2026-06-13)

Promote the convergent-class rule surfaced across four issues into a **charter/standards invariant** + a shared-parser contract.

## The invariant

> Any hook that parses a `gh issue`/`gh pr` command MUST:
> 1. Scope label/repo/identity extraction to the **actual flag values** (the tokens following `--label`/`--add-label`/`--remove-label`/`--repo`), via the shared `_shell_parse` / `_wave_label_parse` tokenizer — never match flag-shaped or label-shaped strings elsewhere in the command (especially `--body`/`--body-file` content).
> 2. Resolve the **flag-omitted (ambient git-context) case** from the invocation cwd's origin (mirroring gh), or log a `skip_no_repo_context`-style event — **never silently drop**.

## Backing evidence (the class)

- **#650** — EDIT path required `--repo`, dropped in-repo label edits. **FIXED** (PR #658, repo-Optional + `resolve_repo_short_name`).
- **#659** — CREATE path (`parse_wave_label_create`) has the same `--repo` gate. OPEN.
- **#661** — `validate_labels` matches label-shaped tokens in issue body. OPEN.
- Lineage: cwd-anchor #144 / #521 → multi-cmd #455 → #650 → #659/#661.

## Scope (Phase 5)

1. Charter/standards edit codifying the invariant (owner: Aino Virtanen) — normal 2-reviewer PR.
2. Apply the invariant to close #659 and #661 (the open instances).
3. Optionally a smoke-test/lint gate asserting hooks route gh-command parsing through the shared tokenizer.

## Provenance

P4W7 retro Proposed Change #1, owner-adopted 2026-06-13 ("Adopt → action in P5"). Not scoped into thin W7.

parametrization:
## Annunaki attack (manual run, 2026-06-14) — two additional class instances

Processing the 35-entry error log surfaced **13 records of the same shared-parser class this issue tracks**, but in two hooks **not yet enumerated** in the instance list above:

[...]

PR #702, opened 2026-06-17
charter+hooks(#663): gh-command parser invariant + machine-enforcement gate

parametrization:
## What & why

Actions the owner-adopted **P4W7 retro Proposed Change #1** (2026-06-13) under **#663**: promote the convergent #650/#659/#661 gh-command-parser class into a **charter invariant + a machine-enforcement gate**, and migrate the one open instance the gate catches.

`#659` and `#661` were already fixed (commit `9ab5c37`). This PR delivers the parts that remained: the charter codification, the gate, and the third (parser-reimplementation) instance.

## Changes

**`charter/hooks.md` § Hook Authorship Requirements**
- New **§7 gh-command Parser Invariant**: any hook parsing a `gh issue`/`gh pr` command MUST (1) scope label/repo extraction to the actual flag VALUES via the shared `_shell_parse`/`_wave_label_parse`/`_repo_flag_parse` tokenizer (no private `shlex.split`, no flag-value regex over raw command text — that leaks label-shaped tokens out of `--body`), and (2) resolve the `--repo`-omitted ambient case via `resolve_repo_short_name` or log a `skip_no_repo_context` event — never silently drop. Bumps the Enforcement count "six → seven".

**`.claude/hooks/tests/test_gh_command_parser_invariant.py`** (new gate)
- Classifies every top-level hook that reads the incoming command and matches a `gh issue`/`gh pr` shape, then asserts (A) no direct `shlex.split`/`shlex.shlex` and (C) no ad-hoc flag-value-capturing regex. Shared parsers are exempt (they ARE the tokenizer). Includes a `_MIN_GH_HOOKS` guard so a classifier regression can't make the gate vacuously green.
- Runs in the **existing pytest suite** that CI already mirrors in `.pre-commit-config.yaml` — **no `ci.yml` / `.pre-commit-config.yaml` edit**, so no serialization needed with #684.
- Scope is `gh issue`/`gh pr` per the invariant; `gh workflow`/`gh api` value-flag extractors (`warn_ghcr_image`'s `-R`) are the same latent class but an enumerated follow-up, not enforced here.

**`.claude/hooks/validate_wave_label_evidence.py`** (the instance the gate caught)
- Replaces the private `_tokenize` (raw `shlex.split`, missing the #287 line-continuation fix), `_walk_flag_values`, `_walk_first_value`, `_is_gh_issue_*` with the shared `tokenize`/`strip_heredocs`/`iter_command_segments`/`find_gh_subcommand`/`walk_flag_values` + `extract_repo`.
- Resolves the ambient repo via `resolve_repo_short_name` instead of the old empty-default `noorinalabs/` slug; logs `skip_no_repo_context` (traces.jsonl) + fails open when unresolvable.
- +6 tests: ambient create/edit resolution, skip-on-unresolvable, `--body` over-match scoping, backslash-line-continuation. Existing 14 behavior-preserved.

## Verification

- ruff format + lint (pinned 0.15.11), mypy: clean.
- Full pytest from a CI-equivalent path (no `.worktrees`/`/tmp` component): **1384 passed, 49 subtests, 0 failures**. (Two location-sensitive pre-existing tests — `test_ontology_tracker` #686 and a `/tmp`-sibling in `test_block_stale_tmp_message_file` — false-fail only when run from inside `.worktrees`//`tmp`; green in CI's clean checkout.)
- sync-drift gate: `OK: pre-commit config mirrors all CI-enforced checks`.
- cspell: not a CI check on this branch (sync-gate confirms parity; #684 not yet merged here) and binary not installed locally — not gating.

## Reviewers (charter: 2)
- Wanjiku Mwangi
- Nadia Khoury

Refs #663

### DIFF
--- a/.claude/hooks/tests/test_gh_command_parser_invariant.py
+++ b/.claude/hooks/tests/test_gh_command_parser_invariant.py
@@ -1,0 +1,254 @@
+r"""Smoke/lint gate for the main#663 gh-command parser invariant.
+
+Charter rule: `charter/hooks.md` § Hook Authorship Requirements
+              "7. gh-command Parser Invariant".
+
+> Any hook that parses a `gh issue`/`gh pr` command MUST scope label/repo
+> extraction to the actual flag VALUES (via the shared `_shell_parse` /
+> `_wave_label_parse` / `_repo_flag_parse` tokenizer) and resolve the
+> flag-omitted (ambient git-context) case — never reimplement the shell
+> tokenizer privately, and never match flag-shaped / label-shaped strings
+> elsewhere in the command (especially `--body`/`--body-file` content).
+
+This gate is the machine-enforcement of that invariant (per
+`feedback_enforcement_hierarchy.md`: "Charter rules without enforcement
+decay"). It is the gh-command analogue of the deferred grep-gate noted in
+§5a, scoped to the `gh issue`/`gh pr` parser class the invariant governs.
+
+Mechanics
+=========
+
+For every top-level hook in `.claude/hooks/*.py` (excluding the three
+sanctioned shared parsers, which ARE the canonical tokenizer/flag-walker),
+classify whether it is a **gh-command parser** — i.e. it reads the incoming
+Bash command (`tool_input` / `command`) AND matches a `gh issue`/`gh pr`
+shape. For each such hook assert BOTH:
+
+  A. It does NOT call `shlex.split` / `shlex.shlex` directly — all shell
+     tokenization must route through `_shell_parse.tokenize` (which carries
+     the line-continuation #287 and heredoc fixes a private reimplementation
+     would miss).
+
+  C. It contains no ad-hoc flag-value-capturing regex literal (a regex that
+     captures the value following `--repo`/`--label`/`--add-label`/
+     `--remove-label`/`-R`/`-l`). Such extraction is the #650/#659/#661 bug
+     class — it leaks label-shaped tokens out of `--body` content. Pure
+     shape-detection regexes (`\bgh\s+issue\s+create\b`) are allowed.
+
+Scope: `gh issue`/`gh pr` per the invariant text. `gh workflow`/`gh api`
+parsers (e.g. `warn_ghcr_image` extracting `-R` from `gh workflow run`) are
+the same latent class but out of the invariant's stated scope — tracked as a
+follow-up, not enforced here.
+"""
+
+from __future__ import annotations
+
+import ast
+import os
+import re
+import unittest
+
+_HOOKS_DIR = os.path.normpath(os.path.join(os.path.dirname(os.path.abspath(__file__)), ".."))
+
+# The canonical shared parsers — they ARE the tokenizer / flag-walker /
+# ambient-repo resolver. They are allowed to call shlex and carry the
+# sanctioned regex fallbacks, so they are exempt from the gate.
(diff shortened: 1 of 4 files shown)


parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #704

Approved. Reviewed the diff at HEAD (no checkout). The machine-enforcement gate is the right shape and genuinely non-vacuous — it classifies gh-command parsers via AST while excluding docstring constants (so a hook that merely documents a gh command is not misclassified), and the >=6 parser floor plus the explicit guard that validate_wave_label_evidence stays in scope both stop the gate from silently regressing to matching nothing. Check A (no direct shlex) and check C (no ad-hoc flag-value-capturing regex) map exactly onto the #650/#659/#661 bug class.

The migration of validate_wave_label_evidence is behavior-preserving — the private tokenize/flag-walk helpers are replaced by the shared _shell_parse + _repo_flag_parse primitives, and the new ambient-repo resolution (recover origin from the cwd when --repo is omitted, fail-open-with-diagnostic when unresolvable) closes the empty-default noorinalabs/ slug gap. Tests cover create/edit block-on-404, unresolvable skip+allow, label-shaped token in --body not treated as a label, and the backslash-line-continuation (#287) case the old shlex path lost. CI is fully green, which means the gate already passes across the entire live hook corpus, not just the migrated hook.

Two non-blocking nits for a later sweep, neither a reason to hold: the classifier docstring enumerates checks A and C with no B (cosmetic leftover), and extract_repo runs over the full command string rather than the located issue segment (a theoretical mis-pick only in a compound multi-gh command, low risk). The gh workflow / gh api value-flag extractors are correctly scoped out as an enumerated follow-up. Strong work.

parametrization:
Requestor: Nadia Khoury
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Reviewed the diff at HEAD (9356b4e). Solid, complete piece of work.

Charter rule (hooks.md §7) is unambiguous — it states the two MUSTs (flag-value scoping via the shared tokenizer; ambient-repo resolution for the flag-omitted case), names the sanctioned shared parsers as the only exemptions, enumerates the backing #650/#659/#661 class, and is explicit that gh workflow / gh api value-flag extractors are out of stated scope as an enumerated follow-up rather than a silent gap. The requirement count (six to seven) and the closing enforcement paragraph were both updated consistently.

The gate is genuinely guarded against vacuous green — _MIN_GH_HOOKS plus test_classifier_finds_the_gh_parser_class catch a classifier that regresses to matching nothing, and test_migrated_hook_is_classified pins the migrated hook in scope. I confirmed the classifier reads only top-level hooks/*.py (the tests/ dir and __init__ are excluded) and that _code_signal_text strips docstrings, so a hook that merely documents a gh command in its module docstring is not misclassified.

Migration is behavior-preserving where it should be. I verified every imported symbol exists at HEAD with a matching signature (tokenize, strip_heredocs, iter_command_segments, find_gh_subcommand, walk_flag_values, resolve_repo_short_name, extract_repo, log_pretooluse_diagnostic). First-value and comma-split label semantics are unchanged; the only intended behavior delta is the ambient-repo resolution replacing the old empty noorinalabs/ default, which the three new AmbientRepoResolutionTests cover (resolve-and-block on 404 for both create and edit, plus fail-open with a skip diagnostic and no silent drop when unresolvable). The body-over-match and backslash-line-continuation cases are also covered. Full pytest suite is green in CI.

Approved. No changes requested.

PR #709, opened 2026-06-18
fix(hooks): read current_phase/current_wave in session-handoff + session-start (#708)

### LINKED ISSUE #708: bug(hooks): session-handoff + session-start report phase/wave as unknown — readers use nonexistent/stale keys, not current_phase/current_wave
## Bug

The session-handoff and session-start reporters say they don't know the current phase/wave even though the canonical keys are present and correct in `cross-repo-status.json`.

Observed this session: session-start printed `Phase phase-4, Wave unknown (last updated: 2026-06-15...)` and the auto-handoff printed `Wave unknown (started unknown)` — while `current_phase` is 5, `current_wave` is `wave-5`, `wave_active` is true, and `wave_5_started_at` is set.

## Root cause

Both readers read key names that do not exist (or are stale):

- `.claude/hooks/session_handoff.py` `_get_wave_status()` reads `data.get("wave")` and `data.get("started")` — **neither top-level key exists** (verified `has("wave") == false`, `has("started") == false`) → always falls back to "unknown".
- `.claude/hooks/session_start.py` `_wave_status()` reads `data.get("phase")` (a STALE flat key still holding "phase-4"), `data.get("wave")` (missing → "unknown"), and `data.get("last_updated")` (stale 2026-06-15).

The canonical, wave-lifecycle-maintained keys — `current_phase`, `current_wave`, and the active wave's `wave_<N>_started_at` / `wave_<N>_kicked_off_at` — are never read. This is distinct from the cross-phase key-collision tracked previously (#683): that work mechanized *resetting* stale `wave_<M>_*` keys; it did not touch these two reader functions, which have been reading the wrong key names independently of the collision.

## Fix

Point both readers at the canonical keys:
- phase ← `current_phase`
- wave ← `current_wave`
- started/freshness ← parse the wave number from `current_wave` (e.g. `wave-5` → `5`) and read `wave_<N>_started_at`, falling back to `wave_<N>_kicked_off_at`.

Drop the stale `last_updated`/flat-`phase` reads (or keep `last_updated` only as a secondary freshness hint, not the phase/wave source of truth). Optionally refresh/remove the stale top-level `phase` flat key so other consumers aren't misled.

## Tests
- Unit-test both readers against a fixture with `current_phase`/`current_wave`/`wave_<N>_started_at` set and the stale flat keys present — assert they report phase 5 / wave-5, not "unknown"/"phase-4".

## Acceptance
- A fresh session-start and the auto-handoff both report the correct active phase and wave.
- Mirrored in pre-commit/CI test suites per the parity rule.

parametrization:
## Root cause (#708)

Session-handoff and session-start reported phase/wave as **"unknown"** (and a stale phase) because both reader functions read keys that do not exist or that lag reality in `cross-repo-status.json`:

- `session_handoff.py` `_get_wave_status()` read top-level `wave` + `started` — **neither key exists** (`has("wave")==false`, `has("started")==false`) → always "unknown".
- `session_start.py` `_wave_status()` read the flat `phase` (stale `"phase-4"` via the #683 cross-phase key collision), `wave` (missing), and `last_updated` (stale).

The canonical lifecycle keys actually maintained by the wave skills are `current_phase` (=5), `current_wave` (=`"wave-5"`), and the active wave's `wave_<N>_started_at` (fallback `wave_<N>_kicked_off_at`).

## Fix

Both readers now resolve through a shared `_wave_phase_started(data)` helper:
- phase ← `current_phase`
- wave  ← `current_wave`
- started ← `wave_<N>_started_at` for `N` parsed from `current_wave`, falling back to `wave_<N>_kicked_off_at`, then `"unknown"`. The `or` chain skips both missing keys **and** present-but-null values — note the live `wave_5_kicked_off_at` is `null`.

Stale flat `phase`/`wave`/`last_updated` are no longer the source of truth. Missing canonical keys degrade gracefully to `"unknown"` rather than crashing.

Verified against the live `cross-repo-status.json`: both readers now report **Phase 5, Wave wave-5 (started 2026-06-16T22:51:14Z)**.

## Tests

No tests existed for either reader. Added `test_session_handoff.py` + `test_session_start.py` (14 cases) covering: canonical-keys-over-stale-flat-keys, the `started_at`→`kicked_off_at` fallback (both missing and present-but-null), the no-timestamps `"unknown"` case, missing-keys-graceful, and the file-reading wrappers (incl. missing-file).

## Green-before-push (main#684)

ruff format --check + ruff lint (`.claude/hooks/` + `.claude/lib/`), mypy (CI flags), full pytest (1479 passed), and `pre_commit_ci_sync.py .` (sync-drift gate OK) — all green locally before push.

Closes #708

### DIFF
--- a/.claude/hooks/session_handoff.py
+++ b/.claude/hooks/session_handoff.py
@@ -14,6 +14,7 @@ Exit codes:
 """
 
 import json
+import re
 import subprocess
 import sys
 from datetime import datetime, timezone
@@ -120,17 +121,43 @@ def _get_ontology_staleness() -> str:
         return "Could not read checksums"
 
 
+def _wave_phase_started(data: dict) -> tuple[str, str, str]:
+    """Resolve (phase, wave, started) from a cross-repo-status.json dict.
+
+    Reads the canonical lifecycle keys maintained by the wave skills:
+      - phase   ← ``current_phase``
+      - wave    ← ``current_wave`` (e.g. "wave-5")
+      - started ← ``wave_<N>_started_at`` for N parsed from ``current_wave``,
+                  falling back to ``wave_<N>_kicked_off_at``, then "unknown".
+
+    The flat ``phase``/``wave``/``started``/``last_updated`` keys are NOT used
+    as the source of truth — ``phase`` lags a phase behind (cross-phase key
+    collision, #683) and ``wave``/``started`` do not exist (#708).
+    """
+    phase = data.get("current_phase", "unknown")
+    wave = data.get("current_wave", "unknown")
+    started = "unknown"
+    if isinstance(wave, str):
+        match = re.search(r"(\d+)", wave)
+        if match:
+            n = match.group(1)
+            # `or` chains past both missing keys AND present-but-null values.
+            started = (
+                data.get(f"wave_{n}_started_at") or data.get(f"wave_{n}_kicked_off_at") or "unknown"
+            )
+    return str(phase), str(wave), started
+
+
 def _get_wave_status() -> str:
-    """Read cross-repo status for active wave."""
+    """Read cross-repo status for the active phase/wave."""
     status_file = REPO_ROOT / "cross-repo-status.json"
     if not status_file.exists():
         return "No cross-repo-status.json found"
     try:
         with open(status_file) as f:
             data = json.load(f)
-        wave = data.get("wave", "unknown")
-        started = data.get("started", "unknown")
-        return f"Wave {wave} (started {started})"
+        phase, wave, started = _wave_phase_started(data)
+        return f"Phase {phase}, Wave {wave} (started {started})"
     except (json.JSONDecodeError, OSError):
         return "Could not read status file"
 
--- a/.claude/hooks/session_start.py
+++ b/.claude/hooks/session_start.py
@@ -11,6 +11,7 @@ Exit codes:
 """
 
 import json
+import re
 import sys
 from pathlib import Path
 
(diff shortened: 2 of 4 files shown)


parametrization:
**Review — PR #709** (`fix(hooks): read current_phase/current_wave in session-handoff + session-start`)

Reviewed at head `9971c730fca1e9d9bd0c8554043e83f235f4b9be` via the contents API (no checkout). **Approved.**

**Both readers resolve the canonical lifecycle keys.** In each `_wave_phase_started` helper (session_handoff.py and session_start.py, identical logic):
- phase comes from `current_phase` (default "unknown") — the stale flat phase key that lags a phase behind (#683) is no longer the source of truth.
- wave comes from `current_wave` (default "unknown") — the nonexistent flat wave key (#708) is gone.
- started parses the digit N out of the wave string, then walks an or-chain: started-at, then kicked-off-at, then "unknown".

**The null-skip is correct.** Because `None or X` yields `X`, a present-but-null timestamp is skipped, not surfaced. With the live state (kicked-off-at null), the chain correctly falls through to "unknown" when no real start timestamp exists. The empty-string case is likewise skipped (falsy), which is the desired behavior here.

**Stale flat keys are no longer trusted.** Neither helper reads the flat phase / wave / started / last_updated keys for its result; the regression tests assert the stale values (phase "phase-4", wave null) do not leak through.

**Test coverage matches the requirement.** Both test files cover canonical-over-stale (asserting phase "5"/wave "wave-5", and NOT "phase-4"/"unknown"), present-but-null fallback (started-at null with a real kicked-off-at), and missing-keys-graceful (all three resolve to "unknown"), plus the kicked-off-at fallback, the no-timestamps→"unknown" case, and the wrapper functions over a temp status file (including the missing-file branch). The fixtures are shaped like the real status file, including the null kicked-off-at and the stale flat keys.

**CI:** statusCheckRollup 8/8 SUCCESS (ruff lint, ruff format, mypy, pytest hooks+lib, sync-drift gate, roster-union gate, commit-author gate, smoke) at the reviewed head. No non-success checks.

Minor, non-blocking: the rendered string reads "Wave wave-5" because the format string prefixes "Wave " onto a value that already carries the wave- token — cosmetic only, consistent with the prior format, not worth holding the PR.

No tech debt warranting a follow-up issue.

---
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Review — PR #709 (fix: read canonical phase/wave keys in session-handoff + session-start, #708)

Reviewed code at HEAD `9971c73` via the GitHub contents API (no checkout).

**Root cause confirmed.** Both readers trusted keys that either do not exist or lag reality: `session_handoff._get_wave_status()` read top-level `wave`/`started` (neither present → always "unknown"); `session_start._wave_status()` read the flat phase value (stale, a phase behind via the #683 collision) plus the absent `wave`. Verified against the live status file shape.

**Fix is correct.**
- Both readers now resolve through a `_wave_phase_started(data)` helper: phase from `current_phase`, wave from `current_wave`, started from `wave_<N>_started_at` (N parsed from the wave string) with an `or`-chain fallback to `wave_<N>_kicked_off_at` then "unknown".
- The `or` chain correctly skips both missing keys AND present-but-null values — important because the live `wave_5_kicked_off_at` is null; a `.get(..., default)` form would have leaked the null through. The `or` form is the right idiom here.
- The stale flat keys are no longer the source of truth; `int`/`str` coercion via `str(phase)`/`str(wave)` keeps output stable, and `isinstance(wave, str)` guards the regex.
- Missing canonical keys degrade to "unknown" rather than crashing; file-level errors are caught and reported as the prior "No/Could not read" sentinels.

Minor, non-blocking: the helper is duplicated verbatim across the two hook files rather than shared from a common module. Acceptable — these are independent entry-point hooks with no existing shared-util module, and the logic is small; not worth a new import surface.

**Tests are thorough.** `test_session_handoff.py` and `test_session_start.py` each cover canonical-over-stale, the started→kicked_off_at fallback (both the missing-key and the present-but-null branch), the no-timestamps "unknown" case, missing-keys-graceful, and the file-reading wrappers including the missing-file path. The present-but-null fallback case is exactly the live-data regression risk and it is exercised.

**CI:** `statusCheckRollup` 8/8 SUCCESS (ruff lint, ruff format, mypy, pytest, smoke, sync-drift gate, commit-identity, roster-union).

Approving.

---
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Issue #717, opened 2026-06-19
tech-debt(hook): block_git_config over-blocks ALL config writes — scope to the user.* identity namespace only


parametrization:
## Problem

`block_git_config.py` blocks **every** `git config` write (anything not read-only). The charter rule it enforces (§ Commit Identity) is specifically about **identity** — never modify `user.name`/`user.email`. The hook over-reaches: it blocked a legitimate `git config --unset core.hooksPath` needed to install pre-commit/pre-push hooks in a child repo (pre-commit "cowardly refuses" while `core.hooksPath` is set to a stale/broken value left over from a repo rename).

Two defects:
1. **Over-block:** operational keys (`core.hooksPath`, `commit.gpgsign`, …) are not identity and should be allowed.
2. **Loop-evadable:** the same `git config` write inside a shell `for` loop was NOT caught, while the standalone form was — a matcher-coverage gap (cf. batch-loop merge-hook evasion).

## Fix

Scope the block to the **`user.*` identity namespace**: `_is_git_config_write` blocks only when an arg targets `user.name`/`user.email` (the whole `user.` namespace, case-insensitive); all other config writes are allowed. Update the block message to say "identity (`user.*`) config writes are prohibited". Keep read-only allowances. (The loop-evasion gap is a `_shell_parse.iter_command_segments` matter — note it; primary fix here is the over-block.)

## Acceptance Criteria

- [ ] `git config user.name/...email/...--unset user.email` (any scope) → blocked
- [ ] `git config [--unset] core.hooksPath`, `commit.gpgsign`, other non-identity keys → allowed
- [ ] read-only (`--get/--list/...`) still allowed
- [ ] tests cover identity-blocked + operational-allowed + read-only-allowed

PR #715, opened 2026-06-19
feat(tooling): auto-sync local main after push/merge to origin/main (#713)

parametrization:
## Summary

A stale local `main` silently runs **pre-fix hooks/skills off disk**. This session opened **22 commits behind** `origin/main` and ran the pre-#709 session-start reader → false orientation (phase/wave "unknown", a 2-of-8-repo scan). Owner request 2026-06-19: after any push/merge to `origin/main`, the local `main` checkout should pick up the change immediately.

## Change

**`.claude/lib/sync_main.py`** — deterministic, side-effect-guarded fast-forward of local `main` to `origin/main`. Fast-forwards **only** when on `main`, strictly behind (`behind>0`, `ahead==0`), and the tree has no real local changes. Refuses (safe no-op) on diverged/ahead/dirty. **Never** force-updates, **never** creates a merge commit (`--ff-only`), **never** discards work. The perpetually-dirty generated annunaki log (`GENERATED_ALLOWLIST`) is stashed around the ff and restored. CLI exits non-zero only on a hard git error; a refusal exits 0.

**`.claude/hooks/auto_sync_main.py`** — PostToolUse(Bash) handler. After a successful `git push … origin … main` / `git push …:main` / `gh pr merge`, it calls the helper (the *self-push* case). Registered in `post_dispatcher.py`. Detection is narrow-but-safe; because the helper is fully guarded, an over-broad match can at worst cost a wasted `git fetch`. A failed push (non-zero exit) is ignored.

**`session-start/SKILL.md` Step 0** — calls the helper at session start (the *others'-push* case), non-fatal.

## Verification

- `sync_main`: 8 tests over **real** temp git repos — ff / up-to-date / not-on-branch / refused-ahead / refused-diverged / refused-dirty / generated-allowlist-stash-restore / `_parse_dirty`.
- `auto_sync_main`: 8 tests — trigger discrimination, non-Bash/non-match/failed-exit gating, message-on-move, silent-when-current.
- Full suite: **1507 passed**. ruff (format+lint), `mypy --ignore-missing-imports`, cspell, and the **pre-commit ⇄ CI sync-drift gate** all green. New `.claude/lib|hooks/*.py` are auto-covered by the existing pre-commit globs — no config change needed.

## Reviewer brief

[...]

PR #720, opened 2026-06-19
test: isolate temp-repo git tests from inherited GIT_* env (#719)

### LINKED ISSUE #719: tech-debt(tests): temp-repo git tests leak GIT_DIR under pre-push hook → break all pushes once hooks installed
## Problem

Installing the pre-push hook (org-wide hook-install effort) exposed a latent test bug. git exports `GIT_DIR=<repo>/.git` (and `GIT_WORK_TREE`/`GIT_INDEX_FILE`) into hook subprocesses. Tests that build temp git repos invoke `git -C <tmp> ...`, which sets the working dir but does NOT override the inherited `GIT_DIR` — so under the pre-push hook those commits operate against the parent `.git` (now carrying a pre-commit hook) and fail with "No .pre-commit-config.yaml found".

Net effect: the pre-push `pytest` hook fails on every push, blocking the repo. It was dormant only because hooks were never installed (broken `core.hooksPath` + never-run `pre-commit install`).

**Evidence:** the tests pass standalone + in CI, but fail with `GIT_DIR=<parent>/.git` set (the exact env a git hook provides).

## Affected

- `.claude/lib/tests/test_verify_commit_identity.py` (`_git` helper)
- `.claude/skills/promotion-audit/tests/test_run.py` (`_git` helper)
- `.claude/lib/tests/test_sync_main.py` (new in the auto-sync PR; same pattern)

## Fix

Sanitize the subprocess env in each temp-repo `_git` helper: drop `GIT_*` vars so temp-repo git is hermetic (`env={k:v for k,v in os.environ.items() if not k.startswith("GIT_")}`). Add a regression that runs a temp-repo commit with `GIT_DIR` set in the environment.

## Acceptance Criteria

- [ ] Temp-repo git helpers strip inherited `GIT_*` env
- [ ] The affected tests pass with `GIT_DIR` set in the environment (hook simulation)
- [ ] `pre-commit run --hook-stage pre-push` (or a real push) succeeds locally

parametrization:
## Summary

Installing the org-wide pre-push hooks surfaced a latent test bug that **breaks every push**. git exports `GIT_DIR`/`GIT_WORK_TREE`/`GIT_INDEX_FILE` into hook subprocesses; temp-repo tests run `git -C <tmp>` which sets cwd but does **not** override the inherited `GIT_DIR`, so under the pre-push hook those commits target the parent `.git` (now carrying a pre-commit hook) and fail with *"No .pre-commit-config.yaml found."* Dormant until now only because the hooks were never installed.

## Change

- Autouse `conftest.py` in `.claude/lib/tests/` and `.claude/skills/promotion-audit/tests/` that strips the `GIT_*` namespace per test (monkeypatch-restored) → temp-repo git is hermetic regardless of launch context. Also auto-covers the auto-sync PR's `test_sync_main.py` once it rebases.
- pre-commit mypy hook `exclude` extended to skip `conftest.py`: two same-named conftest modules collide under one flat mypy invocation. CI already type-checks them safely via its **per-directory** mypy loop (ci.yml), so this is local-hook-only — no CI/parity change.

## Verification

- The previously-failing suites (`test_verify_commit_identity`, promotion-audit `test_run`) and the full `pytest .claude/hooks/tests .claude/lib/tests` (1491 passed) now pass **with `GIT_DIR` set** — the exact hook environment.
- **This PR pushed cleanly through the real pre-push hook** (mypy + pytest + pytest-skills all green) — the end-to-end proof.
- CI-style per-dir mypy green for `.claude/hooks .claude/lib` and `promotion-audit`; sync-drift gate OK.

## Reviewer brief

- **Scope:** test infrastructure only — two conftest fixtures + one pre-commit `exclude` regex. No production code.
- **Check:** the `GIT_*` strip is monkeypatch-scoped (restored after each test); confirm the mypy `exclude` regex `(^\.claude/worktrees/|(^|/)conftest\.py$)` still matches worktrees.
- **TechDebt:** none.

Closes #719. Unblocks the pre-push hooks installed for the org-wide hook rollout; PR #717 and #715's `test_sync_main` depend on this landing.

### DIFF
--- a/.claude/lib/tests/conftest.py
+++ b/.claude/lib/tests/conftest.py
@@ -1,0 +1,27 @@
+"""Isolate temp-repo git tests from an inherited git environment (main#719).
+
+git exports ``GIT_DIR``/``GIT_WORK_TREE``/``GIT_INDEX_FILE`` (and friends) into
+hook subprocesses — notably the pre-push ``pytest`` hook. Tests in this package
+build throwaway repos and run git against them via ``git -C <tmp>``, which sets
+the working directory but does **not** override an inherited ``GIT_DIR``. Without
+this fixture those commits target the *parent* repo (whose pre-commit hook then
+fires with the wrong cwd → "No .pre-commit-config.yaml found"), so the whole
+suite fails when run as a git hook even though it passes standalone and in CI.
+
+Stripping the ``GIT_*`` namespace before every test makes temp-repo git hermetic
+regardless of how the suite was launched. ``monkeypatch`` restores the real
+environment afterwards.
+"""
+
+from __future__ import annotations
+
+import os
+
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def _isolate_git_env(monkeypatch: pytest.MonkeyPatch) -> None:
+    for key in list(os.environ):
+        if key.startswith("GIT_"):
+            monkeypatch.delenv(key, raising=False)
--- a/.claude/skills/promotion-audit/tests/conftest.py
+++ b/.claude/skills/promotion-audit/tests/conftest.py
@@ -1,0 +1,25 @@
+"""Isolate temp-repo git tests from an inherited git environment (main#719).
+
+git exports ``GIT_DIR``/``GIT_WORK_TREE``/``GIT_INDEX_FILE`` into hook
+subprocesses (notably the pre-push ``pytest`` hook). ``test_run.py`` builds a
+throwaway repo and runs ``git -C <tmp> commit`` — which sets cwd but does not
+override an inherited ``GIT_DIR`` — so without this the commit targets the
+parent repo and fails ("No .pre-commit-config.yaml found") whenever the suite
+is run as a git hook, even though it passes standalone and in CI.
+
+Stripping the ``GIT_*`` namespace before every test makes temp-repo git
+hermetic regardless of launch context. ``monkeypatch`` restores afterwards.
+"""
+
+from __future__ import annotations
+
+import os
+
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def _isolate_git_env(monkeypatch: pytest.MonkeyPatch) -> None:
+    for key in list(os.environ):
+        if key.startswith("GIT_"):
+            monkeypatch.delenv(key, raising=False)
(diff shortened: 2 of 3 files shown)


parametrization:
## Peer review — #720 (test: isolate temp-repo git tests from inherited GIT_* env)

I read both new `conftest.py` files at HEAD `3809d3b`. Each ships an autouse `_isolate_git_env` fixture that iterates the full live environment and `monkeypatch.delenv`s every key starting with `GIT_`, which strips the entire namespace (`GIT_DIR`, `GIT_WORK_TREE`, `GIT_INDEX_FILE`, and friends) rather than a hand-picked subset, so a temp-repo `git -C <tmp>` can no longer be hijacked by an inherited `GIT_DIR` from the pre-push pytest hook. Using `monkeypatch` rather than mutating `os.environ` directly means the real environment is restored after each test, so there is no cross-test leakage. The mypy-exclude change `(^\.claude/worktrees/|(^|/)conftest\.py$)` correctly anchors `conftest.py` at either the path root or any directory boundary, which sidesteps the flat-invocation "Duplicate module named conftest" collision while CI's per-dir mypy loop still type-checks each one. CI is green across all thirteen checks. This is a clean, well-scoped test-hermeticity fix.

---
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Peer review — #720 (Wanjiku Mwangi, independent)

I read both new `conftest.py` files and the `.pre-commit-config.yaml` change at HEAD `3809d3b`. Each autouse `_isolate_git_env` fixture iterates the full environment and drops every key whose name begins with `GIT_` via `monkeypatch.delenv(..., raising=False)`, so the whole `GIT_*` namespace (`GIT_DIR`, `GIT_WORK_TREE`, `GIT_INDEX_FILE`, and the author/committer vars) is stripped and restored after each test — exactly what makes the temp-repo commits hermetic under the inherited pre-push hook environment. I confirmed the temp repos set their own identity locally (`promotion-audit/tests/test_run.py` runs `git config user.email`/`user.name` on the throwaway repo before committing), so stripping the inherited `GIT_AUTHOR_*` cannot break those commits. The mypy `exclude` regex `(^\.claude/worktrees/|(^|/)conftest\.py$)` correctly skips worktree copies and any `conftest.py` (start-of-path or after a slash), which resolves the duplicate-module collision while leaving CI's per-dir mypy loop to type-check those files. All thirteen CI checks are green. This does what the title claims and is safe to merge.

---
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

PR #721, opened 2026-06-19
fix(hook): scope block_git_config to the user.* identity namespace (#717)

### LINKED ISSUE #717: tech-debt(hook): block_git_config over-blocks ALL config writes — scope to the user.* identity namespace only

parametrization:
## Summary

`block_git_config.py` blocked **every** `git config` write. The charter rule (§ Commit Identity) — and CLAUDE.md's own wording, "block git config **user** changes" — is identity-specific. The over-block prevented a legitimate `git config --unset core.hooksPath` needed to install local pre-commit/pre-push hooks in a child repo whose `hooksPath` was a stale leftover from a repo rename.

## Change

`_is_git_config_write` now blocks a write **only** when it names a `user.*` key (the identity namespace). Operational keys — `core.hooksPath`, `commit.gpgsign`, … — and all read-only ops are allowed. Block message + docstrings updated.

## Verification

19 tests pass: identity (`user.name`/`user.email`/`user.signingkey`/`--unset user.email`, any scope) blocked; `core.hooksPath` write+unset and `commit.gpgsign` allowed; read-only still allowed.

## Reviewer brief

- **Scope:** one hook + its tests. Confirm the `user.*` regex can't be evaded (e.g. a `--file` value that looks like `user.x` is a non-issue — values aren't keys) and that no legitimate identity write slips through.
- **TechDebt:** none.
- **Stacking:** based on `A.Virtanen/0719-test-git-env-isolation` (the GIT_DIR test-isolation fix) because the pre-push hook can't pass without it. Retarget to `main` once #720 (#719) merges.

Closes #717.

### DIFF
--- a/.claude/hooks/block_git_config.py
+++ b/.claude/hooks/block_git_config.py
@@ -23,17 +23,21 @@ Does NOT match:
 Flag pass-through:
     --get / --get-all / --get-regexp / --list / -l / --show-origin /
     --show-scope → read-only, allowed.
-    Anything else on `git config` → write, blocked.
+    A write to a `user.*` identity key → blocked.
+    A write to any other (operational) key — core.hooksPath, commit.gpgsign,
+    … → allowed (#717: the charter rule is identity-specific; over-blocking
+    all keys prevented legitimate local-hook setup).
 
 Exit codes:
-  0 — allow (not a git config command, or a read-only operation)
-  2 — block (git config write detected)
+  0 — allow (not a git config command, a read-only op, or a non-identity write)
+  2 — block (git config write to the user.* identity namespace)
 """
 
 from __future__ import annotations
 
 import json
 import os
+import re
 import sys
 
 sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
@@ -44,6 +48,14 @@ from _shell_parse import (  # noqa: E402
     tokenize,
 )
 from annunaki_log import log_pretooluse_block  # noqa: E402
+
+# Config keys the charter actually protects (§ Commit Identity — "block git
+# config *user* changes"). The whole `user.` namespace is treated as identity
+# (user.name / user.email / user.signingkey). Non-identity / operational keys
+# (core.hooksPath, commit.gpgsign, …) are NOT blocked — they were over-blocked
+# before #717, which prevented legitimate `git config --unset core.hooksPath`
+# needed to install local pre-commit hooks.
+_IDENTITY_KEY_RE = re.compile(r"^user\.", re.IGNORECASE)
 
 # Read-only flags that mark a `git config` invocation as harmless.
 _READ_ONLY_FLAGS = {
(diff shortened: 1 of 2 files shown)


parametrization:
## Peer review — #721 (fix(hook): scope block_git_config to the user.* identity namespace)

I read the full `block_git_config.py` at HEAD `d7f7f93`. The read-only short-circuit is preserved ahead of the write decision, so reads still pass; the change replaces the unconditional `return True` with `any(_IDENTITY_KEY_RE.match(arg) for arg in args)`, where the pattern `^user\.` (case-insensitive) catches the whole identity namespace. I confirmed by trace that `git config user.signingkey ABC123` and `git config --unset user.email` both block (the `--unset` token is not a read-only flag, so it falls through to the identity scan), while `git config core.hooksPath .git/hooks`, `git config --unset core.hooksPath`, and `git config commit.gpgsign false` all pass, which is exactly the over-block #717 was filed for. The new tests cover both sides of that boundary. One contrived edge I noted and judged acceptable: a non-identity write whose value literally begins with `user.` (e.g. `core.editor user.foo`) would false-block, but that does not occur in practice and the key-token scan comment already acknowledges the tradeoff. This narrows the hook to the charter's actual intent without weakening identity protection.

---
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Peer review — #721 (Wanjiku Mwangi, independent)

I read the full `_is_git_config_write` and the new tests at HEAD `d7f7f93`. The function still resolves the git subcommand, returns early on any read-only flag (exact token or `flag=` form), and now blocks a write only when some token matches `^user\.` case-insensitively — so `user.name`, `user.email`, `user.signingkey`, and `git config --unset user.email` all stay blocked (the whole identity namespace is protected, not just name/email), while `core.hooksPath`, `--unset core.hooksPath`, and `commit.gpgsign` are allowed. That matches the charter's identity-specific rule and unblocks the legitimate local-hook install that motivated the change. The added `OperationalKeyTests` cover both the new allow cases and the broadened identity blocks, and the over-block edge (a value that happens to start with `user.`) fails safe in the blocking direction, which is acceptable. One merge-gating note for the orchestrator rather than a code defect: this branch is stacked on `A.Virtanen/0719-test-git-env-isolation` and currently reports no CI checks, so it must be retargeted/rebased onto `main` and confirmed green before merge. The code itself is correct and does what the title claims.

---
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Issue #734, opened 2026-06-20
P6 W1: Inventory charter prose → deterministic-able vs intentionally-prose (criterion #2 / #726)


parametrization:
## Summary

First deliverable of P6 criterion #2 (code-over-prose, epic #726): inventory the charter and classify every rule as **deterministic-able** (could be a hook/lib check) vs. **intentionally-prose** (judgment, context, rationale that cannot or should not be mechanized).

This is the analysis pass that scopes the conversion work — the sibling issue does the actual conversion of the deterministic-able set. Splitting them lets the conversion volume be sized from real data rather than guessed.

## Acceptance Criteria

- [ ] Every charter sub-doc (`.claude/team/charter/*.md`) walked; each discrete rule catalogued with a stable identifier (file § section).
- [ ] Each rule classified: `deterministic-able` (mechanizable as hook/lib gate) / `intentionally-prose` (with a one-line rationale for why it stays prose) / `already-enforced` (cite the existing hook/lib).
- [ ] `deterministic-able` rules ranked by leverage (how often the gap bites — cross-ref retro/annunaki history where available) to order the conversion issue's work.
- [ ] Inventory written as a durable artifact (a doc under `.claude/team/` or a structured list on #726), not just a comment, so the conversion issue and future audits can consume it.
- [ ] Classification reviewed against the existing enforcement hierarchy guidance (`feedback_enforcement_hierarchy.md`).

## Origin

P6 plan (`.claude/team/phases/phase-6.md`, W1) — owner directive 2026-06-19: "bias toward code over `.md`/memories." Epic: #726. Feeds the charter→code conversion issue.

parametrization:
Requestor: Fatima Okonkwo
Requestee: Nadia Khoury
RequestOrReplied: Request

**Wave 1 Kickoff — Phase 6**

This issue is assigned to you for p6-wave-1.
- Peer reviewer: Wanjiku Mwangi
- Secondary reviewer: Aino Virtanen
- Branch from: `deployments/phase-6/wave-1`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: feature (per charter § Wave Planning & Priority)

Please begin implementation.

Issue #726, opened 2026-06-20
P6 end-state #2: Code-over-prose — convert deterministic charter rules to hooks/lib


parametrization:
## Summary
Phase 6 end-state criterion #2 (Claude Efficiency). Bias the framework toward deterministic enforcement: convert charter prose rules that CAN be code (hooks/lib checks) into code, and explicitly document which rules intentionally remain prose.

## Acceptance Criteria
- [ ] Inventory charter prose rules; tag each as "deterministically enforceable" vs "judgment/prose-only"
- [ ] Convert a prioritized set of enforceable rules into hooks or lib checks (highest-decay-risk first, per the enforcement-hierarchy thesis)
- [ ] Document the prose-only rules with a one-line rationale each (why not code)
- [ ] No regression: the new code-enforced rules have tests and are mirrored in pre-commit/CI per the local-CI-parity rule

## Origin
Owner directive 2026-06-19 (Claude Efficiency phase): "bias toward code over .md/memories." Builds on the existing enforcement hierarchy (hook > skill > charter > memory) and promotion pipeline. See phase-6.md criterion #2.

parametrization:
## Verified complete — closing at Phase 6 exit (P6W17)

Phase-exit verification of each acceptance criterion against artifacts on `main`:

- ✅ **Inventory + tag each rule** — `.claude/team/charter_prose_inventory.md` (#734 / PR #738, merged): full census tagged `already-enforced` (24) / `deterministic-able` (23) / `intentionally-prose` (38), with a leverage-ranked worklist and reconciliation against the pre-existing `promotion-target` markers + `feedback_enforcement_hierarchy`.
- ✅ **Convert a prioritized set → hooks/lib** — #735 (closed) executed the ranked worklist; highest-leverage conversions landed incl. `validate_dockerfile_base_pin` + fixture-realism (Tier-1) and the Agent-Liveness/Throttle-Stall mechanization (#745, closed). Stranding reconciled by #798.
- ✅ **Document prose-only rules with rationale** — the inventory's `intentionally-prose` set (38 rows) carries a one-line rationale each in the per-file walk.
- ✅ **No regression: tests + pre-commit/CI parity** — converted checks ship with tests and were rolled out to child-repo CI (#744, closed); the sync-drift gate (#684) machine-enforces local⇄CI parity.

[...]

Issue #671, opened 2026-06-14
charter/standards: text-processing/NER/graph fixtures MUST use production-realistic input (fixture-masks-bug recurred inside its own fix) — owner-adopted P5W1


parametrization:
## Owner-adopted process change (P5W1 retro, 2026-06-14)

Add a **charter/standards rule**: text-processing / Arabic-NER / graph-load test fixtures MUST use **production-realistic input**, not hand-built minimal or simplified data.

## The rule

> Fixtures for Arabic text processing, NER/segmentation, and graph-load invariants MUST use production-realistic input:
> - **Voweled (vocalized) Arabic** matching real corpus text — never un-voweled toy strings.
> - Real high-frequency structures — e.g. an isnad chain MUST contain the high-frequency transmission particle عن AND a narrator name with an عن/قال substring (عنبسة / معن / مقالة).
> - Real-shape rows (actual column sets / schemas), not minimal hand-built dicts.
> A fixture that is greener than real data is masking a bug.

## Why (the recurring class)

The **fixture-masks-bug** class has now recurred 5+ times, most damningly **inside its own fix**:
- da#146 fixed the un-voweled-toy-fixture blob bug — but its replacement Bukhari-h1 e2e fixture contains no عن, masking a new over-segmentation (da#155).
- Prior instances: MockNeo4jClient masking the APPEARS_IN null-property loader bug; toy h-1 fixtures masking the double-prefix hadith-id bug; local-only staging edges; un-voweled chains.

Surfaced independently by Alejandra Reyes-Fuentes and Jean-Claude Habimana on da#146 (PR #151).

## Scope (Phase 5)

- Aino Virtanen drafts the charter/standards rule (normal 2-reviewer PR).
- Optionally a lint/review-lens that flags Arabic fixtures lacking vocalization marks.

## Provenance

P5W1 retro Proposed Change #1, owner-adopted 2026-06-14 ("Adopt → Aino drafts charter rule").

parametrization:
Requestor: Fatima Okonkwo
Requestee: Aino Virtanen
RequestOrReplied: Request

**Wave 5 Kickoff — Phase 5**

This issue is assigned to you for p5-wave-5.
- Peer reviewer: Nadia Khoury
- Secondary reviewer: Wanjiku Mwangi
- Branch from: `deployments/phase-5/wave-5`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
[...]

Issue #14, opened 2026-04-07
skill: /new-service-deploy — deployment checklist walkthrough


parametrization:
## Summary

Create a `/new-service-deploy` skill that walks through the complete checklist for deploying a new service to the VPS.

## What it does
Interactive checklist that verifies each step:
1. GHCR image built and pushed?
2. VPS_HOST variable set to VPS IP (not domain)?
3. DEPLOY_SSH_PRIVATE_KEY secret set?
4. GH_PACKAGES_TOKEN secret set (if needed)?
5. VPS authenticated to GHCR?
6. Caddyfile updated with server blocks?
7. docker-compose.prod.yml updated with service?
8. Deploy script restarts Caddy?
9. DNS record created in Cloudflare?
10. Cloudflare SSL mode compatible?

For each step, the skill checks what it can programmatically (e.g., `gh variable get VPS_HOST`, check if Caddyfile has the domain) and prompts the user for manual steps.

## Skill definition
- File: `.claude/skills/new-service-deploy.md`
- Trigger: `/new-service-deploy <service-name>`
- Inputs: service name, domain, repo

## Acceptance Criteria
- Complete checklist from deploy #20
- Programmatic checks where possible
- Clear pass/fail per step

PR #743, opened 2026-06-20
feat(charter→code): validate_dockerfile_base_pin + fixture-realism lint (#735)

### LINKED ISSUE #735: P6 W1: Convert deterministic charter rules → hooks/lib checks (criterion #2 / #726)
## Summary

Second deliverable of P6 criterion #2 (code-over-prose, epic #726): convert the `deterministic-able` charter rules identified by the inventory issue into actual hooks / `.claude/lib` checks, and explicitly document the rules that stay prose (with rationale) so the boundary is recorded, not implicit.

The goal is that charter rules which *can* be machine-enforced *are* — closing the "rule with no enforcement decays" gap for the mechanizable subset.

## Acceptance Criteria

- [ ] For each high-leverage `deterministic-able` rule from the inventory, a hook or `.claude/lib` check implemented (or an existing one extended), with tests.
- [ ] Each new/extended check mirrored into `.pre-commit-config.yaml` and classified by the sync-drift gate (`feedback_local_ci_parity_no_force.md`, #684).
- [ ] The charter section for each converted rule gets a provenance pointer to its enforcing hook/check (so prose and enforcement stay linked).
- [ ] Rules classified `intentionally-prose` are annotated in-place with a brief "stays prose because…" note, producing the documented boundary criterion #2 calls for.
- [ ] Scope of conversion in *this* wave is the high-leverage subset; lower-leverage conversions filed as follow-ups (avoid boiling the ocean — bias to the rules whose gaps actually bit).

## Origin

P6 plan (`.claude/team/phases/phase-6.md`, W1) — owner directive 2026-06-19: "bias toward code over `.md`/memories." Epic: #726. Depends on the charter-prose inventory issue.

parametrization:
## Summary

Converts two high-leverage, clean-tool-boundary charter rules from the [#734](https://github.com/noorinalabs/noorinalabs-main/issues/734) prose inventory (ranked worklist **#4** and **#1**) into deterministic `.claude/lib/*.py` checks, wired into pre-commit + CI with full local⇄CI parity. P6W1, criterion #2 / epic #726.

`Closes #735` (wave-branch merges don't auto-close — orchestrator handles closure).

### `check_dockerfile_base_pin.py` — worklist #4
Source: `tech-decisions.md § Base Image Pinning`. Flags any `FROM` lacking a digest pin (`image:tag@sha256:…`) **or** the matching distro upgrade. Honors the section's exemptions verbatim:
- `scratch` final/any layer (no package manager)
- distroless (pin-only sufficient — no package manager)
- multi-stage stage-references (`FROM <earlier-stage>` inherits the pin)
- vendor images with an inline `# RATIONALE:` comment on/above the `FROM`

Distro→upgrade table taken verbatim (Alpine `apk upgrade --no-cache`; Debian-slim `apt-get update && upgrade && clean`; distroless none).

### `check_fixture_realism.py` — worklist #1
Source: `pull-requests.md § Text-Processing / NER / Graph Fixtures Must Use Production-Realistic Input`. This is the lint the section's *Enforcement opportunity* sub-section names (optional half of #671). Flags Arabic-text fixtures that (a) carry no vocalization diacritic (U+064B–U+0652) or (b) lack the عن particle. The عن search strips harakat first (voweled عَنْ → عن) — without that, a fully-voweled chain (which criterion (a) demands) could never satisfy criterion (b), since the fatḥa splits the bare bigram.

Both are CLI gates (`python3 .claude/lib/check_*.py <paths…>`, non-zero + diagnostic on violation, 0 clean) — same reusable-template shape as `pre_commit_ci_sync.py`.

## Wiring + parity

- **pre-commit**: new commit-stage `dockerfile-base-pin` + `fixture-realism` local hooks (scoped via `files:`; the check self-filters non-matching/non-Arabic input).
- **CI**: new `dockerfile-base-pin` + `fixture-realism` jobs in `ci.yml` (discover-and-check; graceful no-op where the repo has no Dockerfiles/fixtures).
- **sync-drift gate**: both kinds added to `pre_commit_ci_sync.py` `_KIND_PATTERNS`, so the gate *demands* the mirror rather than silently ignoring it (the #684 contract). `python3 .claude/lib/pre_commit_ci_sync.py .` → **OK**, zero harmful drift; both kinds classified on both sides.

## Charter edits (criteria #3 + #4)

- **Provenance pointers**: each converted section now carries an "Enforced by `.claude/lib/check_*.py` (#735)" line, and its `<!-- promotion-target: hook -->` marker is flipped to built-state (`hook DONE: …`). The fixture *Enforcement opportunity* + dockerfile *Promotion path* sub-sections updated to "Built (#735)".
- **intentionally-prose boundary annotated** (criterion #4) on the two highest-value rules most likely to be mistaken for mechanizable, citing the #734 inventory: `state-claims.md § Refresh State Before Claim` (the verify-before-claim family — no tool boundary; mechanizing yields false confidence) and `pull-requests.md § Trust the Artifact, Not the Framing` (judgment-class reviewer family — only the format/structural shell is convertible). Light touch, not all 38; the inventory documents the rest.

## Note on this repo

`noorinalabs-main` ships **no** Dockerfiles and **no** Arabic fixtures (they live in the gitignored child repos), so here the deliverable is the reusable check + tests + wiring template + sync-drift classification + charter provenance. The cross-repo file-scanning rollout into each child's CI is a follow-up the orchestrator files (see below).

## Test Plan

```bash
# Unit tests (38 tests, positive + negative cases)
ENVIRONMENT=test python3 -m pytest .claude/lib/tests/test_check_dockerfile_base_pin.py \
  .claude/lib/tests/test_check_fixture_realism.py -q

# Each check as a CLI gate
python3 .claude/lib/check_dockerfile_base_pin.py <Dockerfile...>   # non-zero on un-pinned/un-upgraded FROM
python3 .claude/lib/check_fixture_realism.py <fixture...>          # non-zero on un-voweled / no-عن Arabic fixture

# Parity gate
python3 .claude/lib/pre_commit_ci_sync.py .                         # OK: mirrors all CI-enforced checks
```

Verified locally (green before push):
- full `.claude/lib` + `.claude/hooks` pytest suite: **1592 passed**
- `pre-commit run` (commit stage) + `--hook-stage pre-push --all-files`: all hooks **Passed** (ruff-format/lint, actionlint, cspell, fixture-realism, mypy, pytest, pytest-skills)
- ruff (pinned 0.15.11) lint + format, mypy, markdownlint, sync-drift gate: clean
- changed charter files are cspell-equivalent to the green base (new domain words `distroless`/`dockerfile(s)` added to `.cspell/project-words.txt`).

TechDebt: Conversion scope is JUST inventory #1 + #4 this wave. Deferred remainder (filed as follow-ups by the orchestrator): liveness #2/#3, Tier 2/3 rules, and the **cross-repo CI rollout** that points these two checks at each child repo's real Dockerfiles/fixtures (in this repo they are no-op templates). DEFER-decided inventory rows (#11, #14) intentionally NOT built (await a recorded violation per `feedback_enforcement_hierarchy`). #684 (full cspell sync-drift parity) referenced, not reopened.

https://claude.ai/code/session_01BLKGHMFG8c38aJ8aK8i9Gt


parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none (deferred cross-repo rollout + over-flag scoping is a #735 follow-up, orchestrator-filed per PR body)

**Review: Approved — algorithmic-correctness lens.** Both checks are right; they meet the high bar for gates that will fan out across 8 repos.

**Dockerfile pin check.** The `FROM` parser is sound — case-insensitive keyword, `--platform` flag stripped, `AS <stage>` stripped, trailing inline comment dropped, per-stage body sliced strictly between consecutive `FROM`s (so an upgrade in one stage cannot satisfy another — confirmed by the mixed-multistage test). The stage-reference exemption keys on lowercased stage names recorded in source order, which mirrors Docker's own `FROM`-resolution semantics, so it is not an exploitable bypass beyond what Docker itself permits. Distroless still requires the digest (the pin check precedes the no-upgrade `continue`), matching the pin-only rule. The `apt`/`apt-get … upgrade` matcher correctly rejects a bare `apt-get update`. The non-Alpine/non-distroless-without-package-manager case (e.g. busybox) flagging unless `# RATIONALE:` is documented and by design.

**Fixture-realism check — the subtle one verified.** The `عن` search runs over a harakat-and-tatweel-stripped copy while the vocalization check runs over the raw text, so a production-realistic voweled chain (voweled `عَنْ`) satisfies both criteria at once — the two are not mutually unsatisfiable. The helper-level test asserting the voweled particle is seen through the vowels, plus the `da#146/da#155` voweled-missing-particle worked example as a true negative, lock this in. Detection trigger (presence of a consonantal letter in U+0621–U+064A) correctly skips non-Arabic files; I confirmed the in-repo `.claude/hooks/fixtures/*` carry no Arabic letters, so they are genuinely skipped — no under-detection is being masked there.

**Tests.** Real positive and negative cases on both, non-tautological, and — per the very rule being implemented — the fixture positives are production-realistic voweled isnad with the particle while the negatives are toy strings the linter actually flags. They would fail on a logic regression.

**Charter edits.** Provenance pointers correct; both promotion-target markers flipped to built-state; the two intentionally-prose annotations (verify-before-claim and trust-the-artifact) are sound — each binds a semantic relationship a deterministic gate cannot adjudicate, consistent with the enforcement-hierarchy boundary. No verdict-shaped field lines leaked into the charter prose. All 15 CI checks green, including both new lints as no-ops and the sync-drift gate.

**One rollout consideration (not blocking this no-op-template PR).** When the fixture lint is pointed at a child repo's real fixtures, it will apply the particle floor to any Arabic-bearing fixture in a `fixtures/` path — including a name-only fixture that legitimately has no isnad chain (e.g. a narrator-name record), which would over-flag. The control is per-repo `files:` scoping at rollout time; the author documents this. It belongs on the #735 rollout follow-up rather than blocking here, since this repo ships no Arabic fixtures.

parametrization:
**Review: Approved — CI / pre-commit / sync-drift parity lens**

Reviewed at HEAD `bb472ee`. Verified empirically: configs and scripts fetched at the PR head SHA and run in isolation (38 new tests pass).

**Sync-drift gate (my #684/#706 domain).** Both new kinds (`dockerfile-base-pin`, `fixture-realism`) are classified in `pre_commit_ci_sync.py` and present on BOTH surfaces. Ran the gate against the PR-head configs — clean, with both kinds netting out (not harmful drift, not stricter-local), which is the signature of a complete mirror. I then removed each pre-commit hook in turn and re-ran: the gate reported it as harmful drift and exited 1 both times. The CI-side classification keys on the actual `check_*.py` invocation line inside the `run:` block (not the job name), so a removed job genuinely drops the kind. This is the cspell blind-spot class from #684 closed, not re-opened — a dropped mirror fails the build.

**Local⇄CI parity.** Both checks are mirrored into `.pre-commit-config.yaml` (commit stage — correct for cheap static scans; the heavy mypy/pytest stay pre-push) and `ci.yml` (two new jobs). The ruff pin is unchanged at v0.15.11. No half-mirror.

**CI job wiring.** The `run:` blocks invoke the checks over `find`-discovered files whose globs match the pre-commit `files:` scopes; a real violation exits 1 (last command in the block) and turns the job red — confirmed by the tests asserting exit codes. The exit-0 path fires only when zero matching files exist; the parent repo ships none, documented as the reusable template for the #735 child rollout.

**Throughline — particle search after harakat-strip.** Confirmed the search runs over a diacritic-stripped copy: a bare-substring search on the voweled form returns false, the strip collapses it, so both criteria (vocalization-present and particle-present) are simultaneously satisfiable on a real voweled chain — not mutually unsatisfiable. Locked by dedicated tests (the voweled-particle case and the strip-collapse case).

**cspell additions** (`distroless`, `dockerfile`, `dockerfiles`) are minimal and justified by the new charter prose.

No defects found in the CI/parity surface. Deferring the check-internal correctness (regex edge cases, exemption logic) to the second reviewer's lens; my certification is the gate/diff/CI-wiring correctness.

---
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

PR #753, opened 2026-06-20
refactor(lib): structural yaml.safe_load parse for sync-drift gate (#748 D3)

### LINKED ISSUE #748: Analysis: adopt deterministic/AST CLI tooling + document the current toolchain
## Goal

Two-part request from the owner (2026-06-20):

1. **Adopt deterministic/structural CLI tooling** where we currently hand-roll regex/line-scan parsing — make agent + hook work more correct-by-construction.
2. **Document the toolchain we already depend on** in a contributor-facing README/TOOLCHAIN doc, with install commands, so a fresh clone knows what to install.

This issue is the analysis + the documentation deliverable. Tool-adoption is scoped as proof-of-pattern here and split into follow-ups.

---

## Why this matters — our recurring failure class

Every code/YAML/shell-scanning gate we run is **Python regex / line-scan**. Zero AST/structural tooling is referenced anywhere in `.claude` or `.github`. Our own memory + issue history shows this pattern failing repeatedly: a regex misses a syntactic variant, so "local clean" diverges from reality.

- `feedback_lint_gate_cover_all_syntactic_forms` — gate must match dotted AND from-import access; regex missed a form.
- `feedback_sync_gate_build_kind_false_match` — `pre_commit_ci_sync.py` matched a step **name** ("Docker Buildx") as a build kind.
- Shell-tokenization parser (`_shell_parse.py`) carries a **nine-issue** bug trail: #118, #134, #144, #188, #189, #216, #223, #226, #227.

Structural tooling (AST for code, a real YAML query for workflows, a real shell parser for command extraction) makes each of these correct-by-construction.

---

### Org-level — what `noorinalabs-main` `.claude`/CI invokes
| Tool | Example call site | Purpose |
|------|------|---------|
| gh | validate_labels.py | GitHub API via CLI |
| git | validate_branch_freshness.py | branch/remote/commit checks |
| python3 | pre-commit / CI | run hooks + lib + skills |
| pre-commit | Makefile | local hook framework (mirrors CI) |
| ruff | ci.yml | lint + format `.claude/` |
| mypy | ci.yml | typecheck hooks/lib/skills |
| pytest | ci.yml | hook + lib + skill test suites |
| actionlint | docs.yml | workflow YAML lint |
| cspell | docs.yml | spellcheck prose/docs |
| lychee | docs.yml | internal markdown link check |
| shellcheck | (via actionlint) | shell lint |

### Child-repo build/test/lint/deploy toolchains
| Tool | Repos | Purpose |
|------|-------|---------|
| ruff | ig, us, da, ingest, deploy | Python lint/format |
| mypy | ig, us, da, ingest | Python typecheck |
| pytest | ig, us, da, ingest | Python tests |
| uv | ig, us, da, ingest | Python package mgr |
| pip-audit | ig, da | dep vuln scan |
| npm/npx | ds, ig, lp | Node pkg mgr / runner |
| eslint | ds, ig, lp | JS/TS lint |
| prettier | ds, ig, lp | JS/TS format |
| typescript (tsc) | ds, ig, lp | TS typecheck/compile |
| vitest | ds, ig | JS/TS unit tests |
| vite | ig, lp | frontend bundler |
| astro | lp | static site gen |
| playwright | ig | E2E browser tests |
| actionlint | all 7 | workflow lint (pinned 1.7.12) |
| gitleaks | all 7 | secret scan (pinned 8.24.3) |
| pre-commit | all 7 | hook framework |
| terraform | deploy | IaC |
| docker / compose | da, ig, us, ingest, deploy, lp | containers |
| trivy | deploy | image vuln scan |
| cosign | deploy | image signing |
| shellcheck | deploy | shell lint |
| cspell | all 7 | spellcheck |
| lychee | all 7 | link check |
| markdownlint | all 7 | markdown lint |

### Local-box capability sweep (2026-06-20)
- **Present:** rg, jq, gh, git, python3 (3.14.3), uv, ruff, mypy, pytest, pre-commit, actionlint, shellcheck, npm, node, pnpm, docker, terraform, fd, fzf.
- **Absent:** yq, gitleaks, cspell, lychee, shfmt, trivy, and all the proposed structural tools below.
- ⚠ **`sg` name collision:** `command -v sg` → `/usr/bin/sg` = shadow-utils (run-with-group-id), NOT ast-grep. If we adopt ast-grep it must be invoked as **`ast-grep`** everywhere — a hook shelling `sg ...` would silently run shadow-utils.

---

## The regex/line-scan sites worth converting (conversion map)

| Site | Currently | Better tool |
|------|-----------|-------------|
| `_shell_parse.py`, `validate_commit_identity.py` (200+ lines of overlapping regex), `auto_set_env_test.py` | shlex + regex fallback over Bash command strings (9-issue bug trail) | a **real shell parser**: tree-sitter-bash (via ast-grep) or the `bashlex` Python lib |
| `validate_workflow_paths_coverage.py`, `pre_commit_ci_sync.py` | `re.match()`/substring over workflow + pre-commit YAML lines | **yq** (CLI) or proper `ruamel.yaml` structural parse — query `.jobs[].steps[].run`, `uses:` etc. |
| hooks that grep Python source for call forms (the `lint_gate_cover_all_syntactic_forms` class) | regex over `.py` text | **ast-grep** structural pattern (matches dotted + from-import in one rule) or **semgrep** rule |
| JSON assertions (checksums.json, cross-repo-status.json, PR `--json` fields) | jq paths / ad-hoc | jq (have) + **gron** for greppable/diffable JSON assertions |

---

## Recommended tools (prioritized)

**Tier 1 — retire brittle regex gates (highest leverage):**
- **ast-grep** (binary `ast-grep`, never `sg`) — tree-sitter structural search/lint/rewrite; YAML rule files + `--json` for hook integration. Covers Python source scanning AND bash (tree-sitter-bash grammar).
- **yq** (mikefarah, Go) — real YAML query/eval for every workflow/compose/pre-commit scan. Directly kills the `sync_gate_build_kind_false_match` class.
- **semgrep** — semantic pattern lint + autofix + registry rules + SARIF; lets us express policy hooks as declarative rules and adds SAST for the security-guard rules.

**Tier 2 — diff & config-drift correctness:**
- **difftastic** (`difft`) — structural/AST diff; review correctness (ignores reformatting noise).
- **dyff** or **jd** — structural YAML/JSON diff for config-drift gates.
- **gron** — JSON → greppable assignments; deterministic JSON assertions.

**Tier 3 — safe edits / completeness:**
- **comby** — language-aware structural search/replace for cross-repo codemods/migrations.
- **sd** — literal-by-default find/replace (no sed regex foot-guns).
- **shfmt** — shell formatter (completes shellcheck; we lint shell but don't format it).
- **bashlex** (Python lib) — proper Bash AST for the identity/shell hooks if we prefer a library over ast-grep for that surface.

---

## Side finding — a real parity drift (file a separate small issue)

`deploy` pins **ruff 0.11.6** in `.pre-commit-config.yaml` while ig/da/us/ingest/main pin the org-canonical **0.15.11**. This is a genuine local⇄CI drift (#684 territory) and should be bumped. (Other child gaps the sweep found are intentional + documented.)

---

## Deliverables

- [ ] **D1 (this wave):** Create the contributor toolchain doc (`README.md` "Toolchain / Prerequisites" section or `docs/TOOLCHAIN.md`) listing the CURRENT tools above with install commands per OS/manager, and a `## sg collision` warning. No org README exists today, so this is a net-new doc.
- [ ] **D2 (this wave):** This analysis (current inventory + recommended tools + conversion map). Done in this issue body.
- [ ] **D3 (follow-up issue):** Prove the Tier-1 pattern — convert ONE Python-scan hook to ast-grep and the sync-drift YAML matcher to yq; measure false-match reduction before broad rollout.
- [ ] **D4 (follow-up issue):** Bump deploy ruff 0.11.6 → 0.15.11.

Tool-adoption deliberately stays proof-first (D3) rather than mass-installing — each new binary is also a new dependency for every contributor + CI image, so we validate leverage on one gate first.

parametrization:
Part of #748 (deliverable D3). Does **not** close #748 — the umbrella stays open for the remaining deliverables.

## What

Convert `.claude/lib/pre_commit_ci_sync.py` (the Pre-commit ⇄ CI sync-drift gate) from its hand-rolled YAML line-scanner to a **structural `yaml.safe_load` parse**, eliminating the step-name / comment false-match class deterministically.

- **CI side** classifies **only** `jobs[].steps[].run` (the whole folded block scalar — `safe_load` does the `run: |`/`>` folding, so the #577 `_RUN_BLOCK_OPEN_RE` hand-emulation is deleted) and `jobs[].steps[].uses`. Step `name:` and comments are **never** classified — that exclusion *is* the fix.
- **pre-commit side** classifies `repos[].hooks[].id`, `.entry`, `.name`.
- `_KIND_PATTERNS` re-expressed against **structural values** (e.g. `ruff-lint` keys off the id value `ruff`, not the literal line `id: ruff`; `terraform-fmt` off `terraform_fmt`/`terraform-fmt`, not `id: terraform`). The ruff-format-before-ruff-lint ordering is preserved with a precise per-value guard instead of the old span-removal hack.
- **Robustness:** a YAML parse error or a bare-scalar document now **raises** (CLI exit `2`) instead of degrading to an empty kind-set — no silent false-green. A genuinely empty file still contributes nothing.
- **Semantics preserved:** every kind including `cspell` (#684) and the `build` quality-gate-only tightening that must not match bare `docker build`/`buildx` (#576); `compute_drift` harmful-vs-stricter logic, CLI args, exit codes (0/1/2), missing-file behavior.

## Parity proof (deliverable c)

Ran an OLD-scanner vs NEW-structural comparison across **all 8 real configs** (parent + 7 children). The NEW classifier yields an **identical** `(pre-commit, CI)` kind-set to the retired line scanner for **every** repo — **zero divergence**. The false-positive class is therefore **latent**: no real config exercises a step-name/comment false-match today. `ParityAcrossRealConfigs` encodes the documented per-repo expectations and asserts `NEW == OLD == expected` wherever a config is present (the parent always; sibling child repos only in a full multi-repo checkout, skipped otherwise so CI can't false-fail).

## False-positive demo (deliverable d)

`FalsePositiveDemo` keeps a reference copy of the old line scanner and proves, for a tool token in a step `name:` and in a YAML `#` comment, that the OLD scanner classifies it (`ruff-lint`/`build`) while the NEW structural parser does not.

## Dependency wiring

- **Runtime:** `pyyaml` declared explicitly in the **precommit-ci-sync** job (it now parses YAML) **and** the **pytest** job (the test suite imports the module) — not relying on the ubuntu-runner preinstall.
- **Typecheck:** `types-PyYAML` added to the mypy CI job, since `--ignore-missing-imports` does **not** suppress `import-untyped`.

## Verification

`ruff format --check`, `ruff check`, `mypy`, full `pytest` (1557 passed), the sync gate itself (`OK`, exit 0), and `actionlint` all green locally; pre-commit + pre-push hooks passed on commit/push.

### DIFF
--- a/.claude/lib/pre_commit_ci_sync.py
+++ b/.claude/lib/pre_commit_ci_sync.py
@@ -8,6 +8,29 @@ a PR is opened. This module is the drift GATE: it parses both sides into a set
 of canonical "check kinds" and reports any check CI enforces that the
 pre-commit config does NOT run locally.
 
+Structural YAML parse (#748 D3)
+===============================
+Both sides are parsed with ``yaml.safe_load`` and classified against the
+STRUCTURAL values that actually carry a check invocation — never raw lines,
+step ``name:`` text, or comments. Concretely:
+
+- CI workflow: only ``jobs[].steps[].run`` (the whole multi-line block scalar,
+  which ``safe_load`` folds for us — no hand-rolled ``run: |`` block emulation)
+  and ``jobs[].steps[].uses`` (action refs).
+- pre-commit config: only ``repos[].hooks[].id``, ``.entry`` and ``.name``.
+
+This deletes the previous line-scanner's false-positive class deterministically:
+a step *named* ``lint with ruff`` or a job *named* ``build-and-validate`` can no
+longer masquerade as a real ``ruff``/``build`` check, because names are never
+classified. A YAML comment cannot match either — the parser strips it.
+
+Robustness: on a YAML parse error, or a document that parses to a bare scalar,
+the classifier RAISES rather than returning an empty set. An empty set would be
+a silent false-green (the gate would report "no drift" for a file it could not
+read), which is the exact failure mode this gate exists to prevent. A genuinely
+empty/whitespace file (``safe_load`` -> ``None``) contributes nothing, which is
+correct — there is no check to mirror.
+
 Drift direction that matters
 ============================
 We gate on **CI-enforced-but-not-local** drift only. That is the harmful
@@ -33,53 +56,60 @@ understand. Closing a blind spot here means ADDING the kind to
 demanding a pre-commit mirror — an un-classified CI check is silently
 un-mirrored, which is the exact divergence this gate exists to prevent.
 
-Input Language
-==============
-- A pre-commit config is parsed for `id:` values AND `entry:`/`name:` text.
-- A CI workflow is parsed for `run:` shell lines AND `uses:` action refs.
-Both are matched against per-kind keyword patterns.
-
 Exit codes (CLI):
     0 — no harmful drift (every CI-enforced kind is mirrored in pre-commit)
     1 — harmful drift (a CI-enforced kind is missing from pre-commit)
-    2 — usage / file-not-found error
+    2 — usage / file-not-found / unparseable-config error
 """
 
 from __future__ import annotations
 
 import argparse
-import re
 import sys
 from pathlib import Path
+from typing import Any, Iterator
 
-# Each kind maps to the keyword patterns that identify it on EITHER side
-# (pre-commit id/entry text or CI run/uses text). Patterns are substrings
-# matched case-insensitively against the relevant lines.
+import yaml
+
+# Each kind maps to the keyword patterns that identify it in a STRUCTURAL value
+# (a pre-commit hook id/entry/name, or a CI step's run/uses string). Patterns
+# are substrings matched case-insensitively against that single value — they are
+# NOT raw-line fragments. So `ruff-lint` keys off the hook id VALUE `ruff` (or a
+# `ruff check` run), not the literal line `id: ruff`; `terraform-fmt` keys off
+# the id VALUE `terraform_fmt`/`terraform-fmt` (or a `terraform fmt` run), not
+# `id: terraform`.
 #
-# `build` matches only real build-QUALITY GATES — a job whose purpose is to
-# fail the PR if the project does not build/compile (`build-and-validate`,
+# `ruff-lint` is `("ruff",)` on purpose: structurally each value is ONE tool, so
+# the bare-`ruff` id (which carries no `-lint`/`check` suffix) must still map to
+# lint. A `ruff format` / `ruff-format` value would also contain `ruff`, so
+# `_classify_value` detects ruff-format FIRST and suppresses ruff-lint on that
+# value — preserving the format-before-lint ordering correctness without the old
+# span-removal hack.
+#
+# `build` matches only real build-QUALITY GATES — a step whose `run:`/`uses:`
+# fails the PR if the project does not build/compile (`build-and-validate`,
 # `build-and-test`, `npm run build`). It deliberately does NOT match bare
 # `docker build` / `docker buildx`: those are runtime image-MOVING (retag,
 # promote, publish to a registry, cold-rebuild dry-runs, digest resolution),
-# which is the deploy/publish job itself, not a quality gate a local
-# pre-commit hook could mirror. A bare `docker build` substring also matches
-# `docker buildx`, so a repo that uses buildx at runtime (deploy, the
-# image-publishing CI in isnad-graph / ingest-platform) would otherwise see a
-# permanent un-mirrorable `build` kind and the drift gate could never exit 0
-# (#576). If a repo ever adds a genuine docker-build-as-quality-gate, name
-# that job `build-and-validate` / `build-and-test` (or add an explicit
-# pattern) so it is mirror-tracked. Tightening lifted verbatim from the
+# which is the deploy/publish job itself, not a quality gate a local pre-commit
+# hook could mirror. A bare `docker build` substring also matches `docker
+# buildx`, so a repo that uses buildx at runtime (deploy, the image-publishing
+# CI in isnad-graph / ingest-platform) would otherwise see a permanent
+# un-mirrorable `build` kind and the drift gate could never exit 0 (#576). If a
+# repo ever adds a genuine docker-build-as-quality-gate, express it as a
+# `run:`/`uses:` matching `build-and-validate` / `build-and-test` (or add an
+# explicit pattern) so it is mirror-tracked. Tightening lifted verbatim from the
 # deploy-rollout form (deploy#391, A.Idrissi) and canonicalized here so all
 # vendored child copies converge.
 _KIND_PATTERNS: dict[str, tuple[str, ...]] = {
     "ruff-format": ("ruff-format", "ruff format"),
-    "ruff-lint": ("ruff check", "id: ruff", "- ruff"),
+    "ruff-lint": ("ruff",),
     "mypy": ("mypy",),
     "pytest": ("pytest",),
     "eslint": ("eslint",),
     "typescript": ("tsc", "typecheck", "type-check", "astro check"),
     "prettier": ("prettier",),
-    "terraform-fmt": ("terraform fmt", "terraform_fmt", "id: terraform"),
+    "terraform-fmt": ("terraform fmt", "terraform_fmt", "terraform-fmt"),
     "gitleaks": ("gitleaks",),
     "actionlint": ("actionlint",),
     "pip-audit": ("pip-audit", "pip audit"),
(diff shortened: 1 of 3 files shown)


parametrization:
## Reviewer 2 verdict — Release-Coordinator lens (CI/gate integrity + rollout safety)

Reviewed the artifact at head_sha `7a3c69ca3073ab1aaad4ce1d06ff6304d67c4f50` (diff + files via `gh api contents`), not the PR-body framing.

**Gate-integrity (the load-bearing property).** The structural parse cannot silently false-green. `_safe_load_or_raise` raises `ValueError` on a YAML syntax error or a document that parses to a bare scalar, rather than degrading to an empty kind-set; a genuinely empty/whitespace/comment-only doc returns the empty set, which is correct (nothing to mirror). `main` catches the `ValueError` and returns exit 2 — a parse failure is a loud stop, never a clean exit. `RobustnessAgainstFalseGreen` exercises all four paths (malformed→raise, bare-scalar→raise, empty→empty, and the CLI end-to-end returning 2 on an unparseable workflow). This is exactly the failure mode the gate exists to prevent, and it is closed.

**CI dependency wiring is complete and correct.** types-PyYAML is in the mypy job (correct — `--ignore-missing-imports` silences `import-not-found`, not `import-untyped`); pyyaml is in the pytest job (the lib suite imports the module) and in the sync-drift-gate job (it now parses YAML). I confirmed the module is invoked at runtime by ONLY the sync-drift-gate job — the smoke-test job merely `py_compile`s the file (no import execution, so no dep needed), and no other job calls it. Each declared dep is present in the diff.

**`build` kind still excludes bare docker.** Patterns stay `build-and-validate` / `build-and-test` / `npm run build`; bare `docker build` / `docker buildx` do not match, so the publish-workflow repos (deploy, ingest, isnad-graph) can still reach a clean gate. The tightening (#576) is preserved through the structural rework and `test_bare_docker_build_not_build_kind` guards it.

**Behavior-preservation + the demonstrated fix.** ParityAcrossRealConfigs asserts NEW == OLD line-scanner AND NEW == documented expectations on every present config (the parent always; child repos when present in a full checkout). FalsePositiveDemo proves the removed class — a tool token in a step name or a YAML comment no longer false-matches. The ruff-lint narrowing to the bare `ruff` value is safe: format-first suppression prevents a `ruff format` value double-counting, and any residual broadening over-requires (noisy), never under-requires (silent).

**Checks.** statusCheckRollup is 13/13 SUCCESS; the three load-bearing checks — Mypy type check, Pytest, and the Pre-commit CI sync-drift gate running its own new code against the parent's real config — are all green.

One non-blocking note (folded into the follow-up below): NEW classifies only `jobs[].steps[].run`/`.uses`, so a check expressed via a job-level / reusable-workflow `uses:` carrying a tool token would be invisible where the old scanner caught any `uses:` line. The parity tests prove no current config relies on it (latent narrowing, safe direction). No change needed here.

Verified, genuine approve.

Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: When child repos adopt this structural gate, their CI must add pyyaml (+ types-PyYAML for mypy) — roll into the #684 full-hookset rollout; also a latent-only narrowing of job-level/reusable-workflow `uses:` classification, deferrable until a real config needs it.

parametrization:
## Review — PR #753 (Reviewer 1, TPM lens)

Reviewed the full diff and the source/test at HEAD `7a3c69c`. This is a clean, behavior-preserving refactor of the sync-drift gate from a line-scanner to a structural `yaml.safe_load` parse. Verified against the artifact, not the PR body.

**Semantic preservation of `_KIND_PATTERNS`.** Every kind is preserved and re-expressed against structural values correctly. `ruff-lint` collapsing to `("ruff",)` is sound because each classified value is now a single tool (a hook `id`, a `uses` ref, or one `run` line) rather than a raw line — the format-first guard in `_classify_value` (compute `is_ruff_format`, then skip `ruff-lint` on that same value) preserves the format-before-lint ordering without the old span-removal hack. `terraform-fmt` correctly drops the `id: terraform` line-fragment for the structural `terraform_fmt`/`terraform-fmt`. The `build` quality-gate tightening (#576) is intact — patterns stay `build-and-validate`/`build-and-test`/`npm run build`, none of which is a substring of `docker build`/`buildx`, and classification now keys off `run`/`uses` only (never `name:`), which strengthens the guard; `test_bare_docker_build_not_build_kind` confirms. `cspell` (#684) including the `streetsidesoftware/cspell` action ref is preserved and now flows through the `uses` path; `test_parent_ci_enforces_cspell` is green.

**No false-green regression.** `_safe_load_or_raise` raises `ValueError` on a YAML syntax error or a bare-scalar document (the dangerous CI under-detection direction), and `main` maps that to exit 2. An empty document legitimately contributes nothing. `RobustnessAgainstFalseGreen` covers malformed-raises, scalar-raises, genuinely-empty-empty, and end-to-end `main(...) == 2`. The pre-commit side returning empty for a non-dict is the safe (stricter-local) direction, not a hidden drift.

**Block-scalar handling.** The #577 `_RUN_BLOCK_OPEN_RE` emulation is deleted and `safe_load` genuinely folds `run: |`/`>` bodies; `_classify_run` walks the folded lines and skips shell comments. The retained #577 test class (pipe continuation, multiple tools in one block, folded `>-`, dedent-sibling termination, in-block comment exclusion) now exercises the new parser and is green.

**`check_repo` per-file union.** Correct and necessary — concatenating multiple workflow YAML docs into one `safe_load` would raise on duplicate top-level keys. Parsing each file and unioning kinds is the right fix; the parity test reproduces the old concatenation only for the OLD-scanner comparison.

**Dep wiring (verified in `ci.yml`, not just claimed).** `types-PyYAML` added to the mypy job (correct rationale: `--ignore-missing-imports` silences `import-not-found`, not `import-untyped`), `pyyaml` added to the pytest job and to the sync-gate job. All three are real diff hunks. Mypy/Pytest/sync-gate CI checks are all SUCCESS.

**Parity + false-positive demo tests assert what they claim.** `FalsePositiveDemo` proves OLD classifies a tool token in a step `name:` / `#` comment while NEW does not (genuine `assertIn` old / `assertNotIn` new). `ParityAcrossRealConfigs` asserts NEW == documented expectations and NEW == OLD on every present config, and asserts the parent is always checked. Exit codes, CLI args, and `compute_drift` harmful-vs-stricter logic are unchanged.

One limitation worth tracking (not a blocker, well-documented in the PR): the strongest guarantee — NEW == OLD across all 8 repos — is enforced continuously in CI only for the parent; child-repo parity is asserted only in a full multi-repo checkout and otherwise skipped, so cross-repo behavior-preservation rests on the author's PR-time 8-repo run. That is acceptable given children are independent gitignored repos absent from this repo's CI checkout.

All 13 status checks are green.

Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: Child-repo NEW==OLD parity is only verified at PR time / in a full multi-repo checkout (skipped in CI) — consider a periodic cross-repo parity job so the structural classifier can't silently diverge per-child after future config edits.

Issue #725, opened 2026-06-20
P6 end-state #1: Memory leanness — audit + consolidate corpus, enforce size/count budget


parametrization:
## Summary
Phase 6 end-state criterion #1 (Claude Efficiency). The memory corpus has grown to 100+ entries with an oversized MEMORY.md index (~38KB, flagged in the P5W4 retro). Audit and consolidate it, then enforce a budget so it cannot silently re-bloat.

## Acceptance Criteria
- [ ] Full memory audit: classify each entry as keep / merge-into-X / retire (stale or superseded-by-charter/hook)
- [ ] Merge duplicate/overlapping entries; retire entries already enforced elsewhere (cross-check the promotion-audit SUPERSEDED set)
- [ ] MEMORY.md index trimmed to the surviving set, one line per memory
- [ ] A hook (or lib check) enforces a size/count budget on MEMORY.md + the memory dir, failing loudly when exceeded
- [ ] Budget threshold + rationale documented in the charter

## Origin
Owner directive 2026-06-19 (Claude Efficiency phase): "tighten memories." P5W4 retro pain point: MEMORY.md oversized (38KB). See phase-6.md criterion #1.

parametrization:
## Memory corpus consolidation report — P6W1 (#732)

Judgment half of criterion #1 (#725). Full read + classification of the org corpus, executed via the manifest at `.claude/scratch/p6/w1-732-740-manifest.md`.

**Counts:** org corpus **136 → 101** memory files.

| Action | Count | Detail |
|---|---|---|
| Relocated to children (#740) | 32 | da 15 (#204), deploy 10 (#480), ig 7 (#1112) — each verified present on the child's `origin/main` with `@import` wired BEFORE removal |
| Merged (near-dup) | 1 | `setup_node_npmrc_env_var_name` → `npmrc_node_auth_token_convention` (same PR#924, same lesson); keeper absorbed the unique reviewer-check + sibling link |
| Retired (state-no-longer-true) | 2 | `p4_priority_data_first` (superseded by P6 roadmap); `staging_pipeline_not_wired` (staging no longer empty per p5w5 corpus-load findings) |
| Dangling links repointed/removed | 5 | completes criterion #2; comprehensive scan now clean |

[...]

PR #755, opened 2026-06-21
feat(lib): MEMORY.md size/count-budget enforcement gate (#733)

### LINKED ISSUE #733: P6 W1: MEMORY.md size/count-budget enforcement hook (criterion #1 / #725)
## Summary

Second deliverable of P6 criterion #1 (memory leanness, epic #725): the **enforcement** half. After the corpus is consolidated (sibling audit issue), add a hook that keeps it lean going forward — enforce a size/count budget on `MEMORY.md` and the memory corpus so growth is a deliberate, surfaced decision rather than silent drift.

Per the enforcement hierarchy (`feedback_enforcement_hierarchy.md`, hook > skill > charter > memory), a rule with no enforcement decays — the budget must be a hook, not prose.

## Acceptance Criteria

- [ ] A hook (or `.claude/lib` check wired into an existing hook) enforces a budget on `MEMORY.md` line/entry count and/or per-file count, with the threshold defined in one place.
- [ ] Over-budget produces a clear, actionable message (what to consolidate/retire), consistent with the org's hard-block-vs-advisory stance (`feedback_safety_direction_over_ux_friction.md` — decide block vs. advisory deliberately and document which).
- [ ] Mirrored into `.pre-commit-config.yaml` per the full local⇄CI parity rule (`feedback_local_ci_parity_no_force.md`, #684) and classified by the sync-drift gate.
- [ ] Unit tests under `.claude/hooks/tests/` (or `.claude/lib/tests/`) cover at/under/over budget.
- [ ] Threshold chosen against the *post-audit* corpus size (depends on the audit issue landing first), with headroom rationale recorded.

## Origin

P6 plan (`.claude/team/phases/phase-6.md`, W1) — owner directive 2026-06-19: "tighten memories" + bias toward enforcement-in-code. Epic: #725. Depends on the memory-audit issue (budget calibrated to the consolidated corpus).

parametrization:
## Summary

Enforcement half of **P6 W1 / criterion #1** (memory leanness, epic #725): a machine-enforced budget on the version-controlled project-memory corpus, so growth is a deliberate, surfaced decision rather than silent drift. Per the enforcement hierarchy, a budget expressed only as prose decays — it has to be a gate.

`Closes #733`

## What it does

`.claude/lib/memory_budget.py` (stdlib-only) caps three dimensions, with the budget defined in **one place** (three module constants):

| Dimension | Current (post-audit) | Cap | Headroom |
|---|---|---|---|
| `MEMORY.md` index entries (`^- \[` rows) | 102 | **120** | ~18% |
| memory `*.md` files (excl. `MEMORY.md` + gitignored `session_handoff.md`) | 101 | **120** | ~19% |
| `MEMORY.md` byte size | 17,989 | **24,576 (24 KiB)** | ~37% |

The byte cap is the loosest on purpose (the count caps bind first in the common growth pattern; bytes is a backstop against few-but-bloated lines) and sits just under the ~24.4 KB ceiling a prior retro cited. Headroom is deliberately modest per the owner "tighten" directive — enough that a normal wave's intake doesn't trip it, tight enough to force a consolidate/retire decision before the corpus drifts back toward its pre-audit bloat.

## Block, not advisory

A budget overflow **cannot** be auto-fixed (it needs a human consolidate/retire judgment), so it **HARD-BLOCKS** (exit 1) with an actionable diagnostic — current vs. budget, by how much it's over, and that the fix is to consolidate/retire under `.claude/memory/`. Exit 2 when the corpus can't be evaluated; exit 0 within budget. (`feedback_safety_direction_over_ux_friction`.)

## Local⇄CI parity (#684)

- Mirrored into `.pre-commit-config.yaml` as the `memory-budget` pre-push hook.
- Wired as a CI job (`Memory budget gate`) in `ci.yml`; added `.claude/memory/**` to the trigger paths so a memory-only change still runs the gate.
- Classified the `memory-budget` kind in `pre_commit_ci_sync.py` so the sync-drift gate **actively demands** the mirror (an un-mirrored CI budget job is harmful drift, not a silently-ignored unknown). Frozen OLD-reference patterns + `_EXPECTED_KINDS["."]` kept in lock-step (cspell precedent) so the OLD↔NEW parity proof stays valid. `python3 .claude/lib/pre_commit_ci_sync.py .` → **0 drift**.

## Tests

`.claude/lib/tests/test_memory_budget.py` covers under/at/over budget on each dimension, exit codes, diagnostic content, the file-count exclusions, and a guard that the **real shipped corpus is within budget**. Full hooks+lib suite: 1570 passed.

### DIFF
--- a/.claude/lib/memory_budget.py
+++ b/.claude/lib/memory_budget.py
@@ -1,0 +1,222 @@
+#!/usr/bin/env python3
+"""Enforce a size/count budget on the project-memory corpus (#733, P6 W1).
+
+Second deliverable of P6 criterion #1 (memory leanness, epic #725): the
+*enforcement* half. After the corpus was consolidated (sibling audit issue),
+this check keeps it lean going forward — it caps how large the version-controlled
+memory corpus (``.claude/memory/MEMORY.md`` + the topic files beside it) may grow
+so that growth is a deliberate, surfaced decision rather than silent drift.
+
+Per the enforcement hierarchy (``feedback_enforcement_hierarchy.md`` — hook >
+skill > charter > memory), a budget expressed only as prose decays; it has to be
+machine-enforced. This module is that machine: a CLI that exits non-zero when the
+corpus is over budget, mirrored into ``.pre-commit-config.yaml`` and wired as a
+CI job (``Memory budget gate``) so local and CI agree (the full local⇄CI parity
+rule, ``feedback_local_ci_parity_no_force.md`` / #684).
+
+Three dimensions, one threshold each (single source of truth)
+=============================================================
+The budget is the three module constants below — the ONLY place a limit is
+defined. Each is enforced independently; the corpus is over budget if ANY one is
+exceeded.
+
+- ``MAX_INDEX_ENTRIES`` — index lines in ``MEMORY.md`` (``^- \\[`` rows, one per
+  recallable memory). This is the primary, most meaningful dimension: it counts
+  the things a session has to scan at startup.
+- ``MAX_MEMORY_FILES`` — topic ``*.md`` files in ``.claude/memory/`` excluding
+  ``MEMORY.md`` itself and the gitignored, machine-local ``session_handoff.md``
+  (see below). Roughly tracks the index-entry count and catches an orphaned
+  file that was added without an index row.
+- ``MAX_MEMORY_BYTES`` — byte size of ``MEMORY.md``. A backstop against the
+  pathological "few but enormous" index lines that the entry count alone would
+  not catch.
+
+Why the index count (102) and the file count (101) differ by one
+================================================================
+``MEMORY.md`` carries a committed pointer line for ``session_handoff.md``, but
+that file is gitignored (per-session machine-local churn — see ``.gitignore``)
+and is absent in a CI checkout. To keep the *file* count identical on a
+developer machine (where the handoff exists) and in CI (where it does not), the
+handoff is excluded from the file count. The index therefore legitimately counts
+one more (the committed handoff pointer line) than the file count. Both budgets
+have independent headroom, so the off-by-one is harmless.
+
+Threshold calibration (headroom rationale — owner directive 2026-06-19 "tighten")
+=================================================================================
+Measured against the *post-consolidation* corpus at #733 time:
+    index entries = 102,  topic files = 101,  MEMORY.md = 17,989 bytes (~17.6 KiB).
+
+    MAX_INDEX_ENTRIES = 120  → ~18 entries / ~18% headroom over 102.
+    MAX_MEMORY_FILES  = 120  → ~19 files   / ~19% headroom over 101.
+    MAX_MEMORY_BYTES  = 24576 (24 KiB)  → ~37% headroom over 17,989 bytes; sits
+        just under the ~24.4 KB ceiling a prior retro cited for MEMORY.md before
+        the consolidation (it had bloated to ~38 KB). The byte cap is the loosest
+        of the three on purpose — the entry/file caps bind first in the common
+        growth pattern (adding memories); the byte cap only catches the rare
+        case of fewer-but-bloated index lines.
+
+The headroom is deliberately modest: enough that a normal wave's memory intake
+does not trip the gate, but tight enough that it forces a consolidate/retire
+decision before the corpus drifts back toward its pre-audit bloat. Raising a cap
+is a one-line, reviewed change here — which is exactly the "deliberate, surfaced
+decision" #725 wants, rather than silent growth.
+
+Block, not advisory (``feedback_safety_direction_over_ux_friction.md``)
+=======================================================================
+A budget overflow CANNOT be auto-fixed — it needs a human consolidate/retire
+judgment about which memories to merge or drop. So the right posture is a HARD
+BLOCK with an actionable diagnostic (current vs. budget, by how much it is over,
+and that the fix is to consolidate/retire in ``.claude/memory/``), never a silent
+advisory that scrolls past. Exit 1 on overflow; the pre-commit/CI gate then stops
+the push.
+
+Exit codes (CLI):
+    0 — corpus is within budget on all three dimensions.
+    1 — over budget on at least one dimension (HARD BLOCK).
+    2 — usage / corpus directory or MEMORY.md not found (cannot evaluate).
+"""
+
+from __future__ import annotations
+
+import argparse
+import re
+import sys
+from pathlib import Path
+from typing import NamedTuple
+
+# --- Budget: the single source of truth. Edit a number here and nowhere else. ---
+MAX_INDEX_ENTRIES = 120
+MAX_MEMORY_FILES = 120
+MAX_MEMORY_BYTES = 24_576  # 24 KiB
+
+# Files under .claude/memory/ that are NOT counted as topic files: the index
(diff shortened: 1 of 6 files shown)


parametrization:
## Review — PR #755 (Reviewer 1, TPM lens)

Reviewed the full diff and the live corpus at HEAD `d25e2c6`. Solid, well-reasoned enforcement gate. `Closes #733` is correct (leaf deliverable, base main).

**Budget logic correctness — verified against the real corpus, not just the prose.** The three caps are genuinely single-source (the module constants; the test imports them rather than re-hardcoding). Counting rules are right and, crucially, dev↔CI-stable: index = `^- \[` rows in MEMORY.md (reads committed content), file count = `*.md` minus MEMORY.md and the gitignored `session_handoff.md`, byte count = MEMORY.md size (committed). I confirmed live: index = 102, files = 101, MEMORY.md = 17,989 bytes — matching the PR exactly. The 102-vs-101 off-by-one is precisely the committed handoff pointer line (MEMORY.md line 25, matches `^- \[`) which the index counts but the file count excludes because `session_handoff.md` is gitignored (`git check-ignore` confirms) and absent in CI. Excluding it from the file count is what makes the file dimension identical on a dev machine and in CI — the whole point — and the explicit exclusion is robust whether or not the file is present. Both dimensions have independent headroom so the off-by-one is harmless.

**Parity-proof integrity — the critical item — is preserved.** Adding `memory-budget` to production `_KIND_PATTERNS`, to the frozen OLD-reference patterns, AND to `_EXPECTED_KINDS["."]` (both precommit and ci) is the correct lock-step move, and it does not weaken the proof. The proof's job is to show the line-scan→structural refactor is behavior-preserving for a GIVEN pattern set; keeping the reference set in lock-step as kinds are added (the cspell precedent) means it keeps testing "same patterns, two strategies agree," not "two divergent pattern sets." The `memory-budget` patterns are plain identical substrings (`memory-budget`/`memory_budget`) in both copies, classified off the `run:`/`id`/`entry` invocation in the real parent config — never the job key `memory-budget:` or the step `name: Memory budget gate` (those carry no underscore/hyphen token and, more importantly, NEW never classifies names/job-keys). So the structural false-positive guard is untouched: strip the `memory_budget.py` run line and NEW correctly stops classifying it. Children's `_EXPECTED_KINDS` were left unchanged, consistent with the gate being parent-only for now. The sync-drift gate and Pytest CI checks are both green, confirming the parity tests still pass.

**Test coverage is comprehensive.** Under-budget (exit 0), exactly-at-cap on entries AND files simultaneously (inclusive cap, exit 0), one-over on each of the three dimensions independently (exit 1), missing dir / missing MEMORY.md (exit 2), CLI repo-root and `--memory-dir` resolution, the file-count exclusions, the actionable diagnostic content (EXCEEDED banner, dimension name, "OVER by N", cap value, "cannot be auto-fixed", CONSOLIDATE/RETIRE), and a guard that the shipped corpus is itself within budget. Exit codes 0/1/2 and the HARD-BLOCK posture match the design.

**Threshold headroom is sound.** 102→120 (~18%), 101→120 (~19%), 17,989→24,576 (~37%); the byte cap deliberately loosest as a backstop while the count caps bind first. Modest per the owner "tighten" directive — tight enough to force a consolidate/retire decision, loose enough not to trip a normal wave's intake. The module is stdlib-only, so the new CI job correctly needs no dependency wiring (checkout + setup-python + run), and Mypy is green.

All 14 status checks are green, including the Memory budget gate itself, the sync-drift gate, Pytest, and Mypy.

Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: The budget gate is parent-only — child-repo memory corpora follow the same self-contained pattern (per CLAUDE.md) but are currently unbudgeted; consider templating this gate out to the children like the sync-drift gate. Non-blocking.

parametrization:
## Reviewer verdict — PR #755 (#733 memory-budget gate), gate-integrity / release lens

Reviewed at head_sha `d25e2c6ccea3370aec865af6f78c02b46b2235ee`.

**Hard-block correctness — no false-green.** `gather_metrics` raises `FileNotFoundError` when the memory dir OR `MEMORY.md` is absent, and `main` converts that to exit 2 — a corpus it cannot read is "cannot evaluate," never a silent within-budget pass. Over-budget on any one of the three dimensions → exit 1; within → exit 0; caps are inclusive (`current > limit`, so at-limit is allowed). A present-but-empty `MEMORY.md` correctly reads as within budget (zero entries genuinely is) — that is not a false-green, only the missing-file case would be, and that is exit 2. The over-budget diagnostic is a real hard block: it names the offending dimension(s), prints current/limit and "OVER by N", and tells the human the fix is a consolidate/retire judgment (not auto-fixable). Tests exercise under/at/one-over on each dimension independently plus both exit-2 cases.

**CI wiring + trigger paths.** The new "Memory budget gate" job runs `python3 .claude/lib/memory_budget.py .` with only checkout + setup-python — the module is pure stdlib, so (correctly, unlike the pyyaml gate in #753) it needs no `pip install`. Adding `.claude/memory/**` to both the push and pull_request `paths:` widens the trigger set (a memory-only PR now runs CI and the budget gate) — paths are a union, so no other trigger is narrowed or skipped. The pre-push `memory-budget` hook mirrors the job with `always_run` / `pass_filenames: false`, which is right for a whole-corpus (not changed-files) check.

**Sync-drift cross-impact (carried from my #753 review).** The new `memory-budget` kind keys off `memory-budget` (the hook `id` value) and `memory_budget` (the `memory_budget.py` token in the CI `run` scalar) — both are STRUCTURAL values, so the D3 structural-parse property is preserved and no false-green / name-or-comment false-positive is reintroduced. Notably the job's display name "Memory budget gate" (spaces) matches neither pattern and, post-D3, names are not classified at all — so the classification rests purely on the run/id/entry values. The OLD-reference `_OLD_KIND_PATTERNS` and `_EXPECTED_KINDS["."]` were updated in lock-step (kind added to both the precommit and ci expected sets), keeping the NEW==OLD==documented parity proof faithful. `python3 .claude/lib/pre_commit_ci_sync.py .` is green → 0 drift after the addition.

**Checks.** 14/14 SUCCESS, including the two that run their own new code: the Memory budget gate (evaluating the real parent corpus — confirms it sits under the 120/120/24576 caps with headroom) and the Pre-commit ⇄ CI sync-drift gate (confirms the new kind classifies symmetrically on both sides).

Calibration and exclusions are sound: the index/file counts excluding `MEMORY.md` + the gitignored `session_handoff.md` is explained and tested (the legitimate off-by-one between index and file counts is documented and has independent headroom); the byte cap scoping to `MEMORY.md` (the always-loaded index) rather than every topic file is a deliberate, documented choice.

Verified, genuine approve.

Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: Gate is parent-only, but per CLAUDE.md each child repo commits its own `.claude/memory/`; the budget gate (and the now-vendored `memory-budget` sync-kind, currently inert in children) should roll out to each child repo so their corpora are enforced too — track under the #684 full-hookset rollout / #725 leanness epic. Minor: per-topic-file byte size is uncapped (only file count + MEMORY.md bytes are) — acceptable for now since MEMORY.md is the always-loaded surface, revisit only if a bloated topic file appears.

Issue #759, opened 2026-06-21
zsh is the org shell: make it explicit in docs, READMEs, and tooling + audit bash-isms


parametrization:
## Problem

The owner's shell is **zsh**, but agents (and some committed tooling) repeatedly "think in bash" — producing `declare -A`, `${!arr[@]}`, and other bash-only idioms that silently break or misbehave under zsh. This has caused a recurring class of errors (e.g. the P3W12 `(eval):3: bad substitution` failure that prompted `feedback_zsh_shell_environment.md`, and several more this phase: `?`-glob expansion on unquoted URLs, array-syntax breaks).

The lesson currently lives **only** in project memory (`feedback_zsh_shell_environment.md`) — it is invisible to anyone reading `CLAUDE.md`, the repo READMEs, or the tooling itself. Per the enforcement hierarchy (`feedback_enforcement_hierarchy`), a convention that lives only in a memory file decays. Make zsh the **explicitly documented** shell everywhere it matters, and stop shipping bash-isms in our own tooling.

## Scope

**Documentation (org-level — folds into the #748-D1 toolchain doc):**
- [ ] Add a prominent **"Shell environment: zsh"** section to org `CLAUDE.md` and the new toolchain doc — state the shell, link the do/don't list, point to the memory.
- [ ] Codify the zsh convention in `ontology/conventions.md` (org-wide convention, not buried in a memory): no `declare -A` / `${!arr[@]}`; quote URLs/globs; prefer POSIX-portable constructs; `bash -c '...'` explicitly when bash is genuinely required.
- [ ] Per-repo: add a one-line "Shell: zsh — write zsh-safe commands" note to each child repo's README/CLAUDE.md so a fresh clone in any repo knows (child-repo edits via each child's own roster).

**Tooling audit (committable fixes):**
- [ ] Sweep committed `.claude/hooks/`, `.claude/lib/`, and the **bash blocks inside skill `*.md` files** for bash-only idioms that break under zsh (these blocks are executed by the Bash tool, which runs zsh). Fix or POSIX-ify; document any block that legitimately requires `bash -c`.

## Acceptance

- zsh is discoverable from `CLAUDE.md`, `ontology/conventions.md`, the toolchain doc, and every repo README/CLAUDE.md.
- No bash-only array/substitution idiom remains in committed parent tooling without an explicit `bash -c` wrapper + comment.

[...]

PR #763, opened 2026-06-21
fix(skills): zsh-safe rewrites of bash-only idioms in executed skill blocks (#759)

parametrization:
## Bash-ism audit of committed parent tooling (#759 code portion)

The owner's shell and the Bash tool are **zsh 5.9**. This sweeps `.claude/hooks/**`, `.claude/lib/**`, and the executed bash blocks inside `.claude/skills/**/*.md` for bash-only idioms that break or silently misbehave under zsh. Every finding below was reproduced under the live zsh before fixing, and each rewrite is mechanically behaviour-preserving (re-verified post-fix).

### Findings fixed (all in skill bash blocks)

| File | Bash-ism | zsh failure mode | Fix |
|------|----------|------------------|-----|
| `board-audit/SKILL.md` | `WAVE_OPTION_IDS["$name"]="$id"` | zsh keeps the quotes *inside* the subscript as part of the key (`"P6W1"`), so later `${WAVE_OPTION_IDS[$expected]}` lookups MISS | unquoted `[$name]=$id` (correct in both shells; keys are space-free) |
| `board-audit/SKILL.md` | `for url in $ORPHANS` | zsh does not word-split an unquoted scalar → whole newline-list collapses into ONE bogus iteration (same class as main#688) | `while IFS= read -r url … done <<< "$ORPHANS"` + `[ -n ]` guard |
| `phase-review/SKILL.md` | `for issue in $TRACKING_ISSUES` | same scalar-collapse | `while read` here-string + guard |
| `session-start/SKILL.md` | `for repo in $REPOS` (+ space-separated fallback) | same scalar-collapse; fallback string read as one line | normalized fallback to newline-separated; `while read` via **here-string** (not a pipe, so the `RED` array survives `done`) |
| `wave-start/SKILL.md` | `gh api …/cross-repo-status.json?ref=main` (unquoted) | unquoted `?` is a zsh glob; with NOMATCH-on-by-default zsh aborts with "no matches found" before gh runs | quote the URL |

### Audited, already zsh-safe — deliberately NOT changed

- **`wave-kickoff/SKILL.md`** `declare -A BRANCH_SHA/BRANCH_STATUS` — already uses unquoted subscripts and a here-string loop (explicitly main#688-hardened). Works in zsh.
- **Process substitution** `done < <(…)` / `comm -23 <(…) <(…)` across session-start/board-audit/wave-wrapup/wave-scope — zsh fully supports `<(…)`; not a bash-ism.
- **All Python `.claude/hooks/**` + `.claude/lib/**`** — run via `python3`, not the shell. The single `shell=True` (`session_handoff.py`) uses `/bin/sh` (POSIX), is insulated from zsh, and only runs plain `git`/`gh` strings.
- The `for … in $VAR` hits in `wave-wrapup`/`wave-scope` are **comments** documenting the anti-pattern; the code there already uses the safe `while read` form.

### Verification
- Each broken construct reproduced under zsh 5.9, then each rewrite re-run under zsh (3 iterations not 1; zero-iteration-on-empty preserved; quoted URL no longer globs; assoc lookup hits).
- `pre-commit run --all-files` (commit + `--hook-stage pre-push`): ruff-format, ruff-lint, actionlint, cspell, mypy, pytest, pytest-skills, memory-budget — all **Passed**.
- No hook/lib code touched, so the hook test suites pass unchanged.

### Left for human judgment
- Prose/inline-code mentions of `…json?ref=main` in `wave-start` (lines 154/158) and `wave-kickoff` (line 224) are **documentation**, not executed blocks, so they were left as-is. If we want copy-paste-safety from prose too, they could be quoted in a follow-up — flagging rather than guessing.

Part of #759

### DIFF
--- a/.claude/skills/board-audit/SKILL.md
+++ b/.claude/skills/board-audit/SKILL.md
@@ -202,10 +202,15 @@ The user MUST type `y` to proceed. Any other answer aborts with `BLOCK: user dec
 ### 6. Bulk-add orphans
 
 ```bash
-for url in $ORPHANS; do
+# `while read` over the newline-list, NOT `for url in $ORPHANS` — zsh does not
+# word-split an unquoted scalar, so the whole list would collapse into one bogus
+# iteration (#759, same class as main#688). The `[ -n ]` guard preserves the
+# zero-iteration-on-empty behaviour `for` gives when $ORPHANS is empty.
+while IFS= read -r url; do
+  [ -n "$url" ] || continue
   gh project item-add 2 --owner noorinalabs --url "$url" || \
     echo "WARN: failed to add $url"
-done
+done <<< "$ORPHANS"
 ```
 
 Per memory `feedback_gh_pr_edit_silent_noop` family, `gh project item-add` can silently no-op — `gh` is being deprecated for project-classic operations (see `pull-requests.md § gh pr edit projects-classic deprecation`). If the bulk-add appears to succeed but the orphan count doesn't drop on the next audit run, fall back to the GraphQL `addProjectV2ItemById` mutation:
@@ -244,7 +249,12 @@ query {
 WAVE_FIELD_ID=$(jq -r '.data.organization.projectV2.field.id' /tmp/wave-options.json)
 declare -A WAVE_OPTION_IDS
 while IFS=$'\t' read -r name id; do
-  WAVE_OPTION_IDS["$name"]="$id"
+  # Unquoted subscript on assignment — zsh keeps quotes *inside* a subscript as
+  # part of the key (`WAVE_OPTION_IDS["$name"]=` would store key `"P6W1"`, with
+  # the quotes, so later `${WAVE_OPTION_IDS[$expected]}` lookups miss). Bash
+  # strips them; the unquoted form is correct in both. Keys are wave names
+  # (no spaces), so no word-split risk. (zsh-safety, #759)
+  WAVE_OPTION_IDS[$name]=$id
 done < <(jq -r '.data.organization.projectV2.field.options[] | "\(.name)\t\(.id)"' /tmp/wave-options.json)
 
 # Project node ID — same response.
--- a/.claude/skills/phase-review/SKILL.md
+++ b/.claude/skills/phase-review/SKILL.md
@@ -52,12 +52,17 @@ Extract tracking issue numbers from the phase plan and pull live state for each:
 ```bash
 TRACKING_ISSUES=$(grep -oE 'noorinalabs-main#[0-9]+' "$PHASE_DOC" | sort -u)
 
-for issue in $TRACKING_ISSUES; do
+# `while read`, NOT `for issue in $TRACKING_ISSUES` — zsh does not word-split an
+# unquoted scalar, so the multi-line list would collapse into one bogus iteration
+# (#759, same class as main#688). The `[ -n ]` guard keeps the no-op-on-empty
+# behaviour `for` gives when no tracking issues are found.
+while IFS= read -r issue; do
+  [ -n "$issue" ] || continue
   num=${issue#noorinalabs-main#}
-  gh issue view $num --repo noorinalabs/noorinalabs-main \
+  gh issue view "$num" --repo noorinalabs/noorinalabs-main \
     --json number,title,state,labels,closedAt \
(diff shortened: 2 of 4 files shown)


parametrization:
I reproduced every changed block under the live zsh 5.9 and confirmed each rewrite is functionally equivalent to the bash original while fixing a real zsh failure.

The scalar-collapse class (board-audit `for url in $ORPHANS`, phase-review `for issue in $TRACKING_ISSUES`, session-start `for repo in $REPOS`) is genuine — zsh does not word-split an unquoted scalar, so the bash for-loop ran exactly one bogus iteration over the whole newline blob; the `while IFS= read -r … done <<< "$VAR"` rewrites iterate once per line and the `[ -n ]` guard preserves the zero-iteration-on-empty contract. I verified ORPHANS is built from `comm -23` (newline list, space-free URLs), TRACKING_ISSUES from `grep -oE` (newline list), and REPOS now normalized to newline form via `printf '%s\n'`, so the here-string shape is correct for all three sources. I separately confirmed the session-start RED array survives past `done` because a here-string keeps the loop in the current shell, and the inner `while … done < <(gh api …)` reads its own process-substitution fd without stealing the outer here-string's stdin.

The associative-array fix is the subtle one and it checks out — in zsh `A["$name"]=…` stores the key with the literal quotes so a later `${A[$expected]}` lookup misses, while the unquoted `A[$name]=$id` stores the bare key in both shells; keys are space-free wave names so there is no word-split exposure. The wave-start URL quoting is correct too — an unquoted `?` is a zsh glob and with NOMATCH on by default the command aborts with "no matches found" before gh runs; quoting passes gh the identical string bash already received.

No `bash -c` wrappers were introduced; the rewrites stay shell-native. Prose/inline-code `…?ref=main` mentions were correctly left as documentation and flagged for follow-up rather than guessed at. Hook/lib code is untouched so the existing suites remain valid. No behavior-change risk found.

---
Requestor: Nadia Khoury
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none
All five zsh-safe rewrites reproduced and proven behavior-preserving under live zsh 5.9; no blocking issues.

parametrization:
I independently re-verified every shell-block rewrite in this PR against a live zsh 5.9, reproducing each claimed failure and re-running each replacement form. All five are behaviour-preserving and fix real zsh-only breakage in executed skill blocks.

board-audit bulk-add — under zsh an unquoted scalar does not word-split, so the original `for url in $ORPHANS` collapses a three-URL newline list into a single bogus iteration; I reproduced that (1 iteration) and confirmed the `while IFS= read -r ... done <<< "$ORPHANS"` form yields the correct three, with the `[ -n ]` guard keeping the zero-iteration-on-empty behaviour the old `for` gave on an empty list.

board-audit wave-option map — I reproduced the assoc-subscript bug exactly: `A["$name"]=$id` stores the literal key with quotes so `${A[$name]}` misses, whereas the unquoted `[$name]=$id` stores the bare key and the lookup hits. Wave names are space-free so the unquoted subscript is correct in both shells.

phase-review tracking-issue loop — same scalar-collapse class; the while-read here-string plus guard is equivalent, the prefix-strip on a full line is preserved, and the added `"$num"` quoting is a harmless hardening.

session-start red-run scan — the newline `printf` fallback matches the jq output shape so the while-read consumes it correctly, and I confirmed the here-string keeps the loop in the current shell so the `RED` array survives past `done`; the inner per-workflow loop redirects its own stdin from a process substitution and the body's gh/grep calls do not read stdin, so the outer here-string fd is never stolen and there is no regression over the original `for`.

wave-start status read — I reproduced the unquoted-`?` glob aborting the whole command with "no matches found" under default nomatch before gh runs; quoting the URL fixes it with gh receiving an identical argument. The remaining unquoted `?ref=main` mentions left in prose are non-executed documentation and were correctly flagged for a follow-up rather than changed here.

No executable semantics changed beyond removing the zsh breakage, no `bash -c` wrappers were introduced, and the Python hook/lib paths were correctly left untouched. Clean to merge.

---
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none
zsh-safe skill-block rewrites verified behaviour-preserving against live zsh 5.9; all three reproducible bugs reproduced and fixes confirmed.

Issue #766, opened 2026-06-21
Document the full phase→wave lifecycle as a slash-command flow (markdown + ontology section)


parametrization:
Document the entire cycle — starting a phase, setting up a wave, executing it, and closing waves — as the sequence of slash commands. For EACH command (/session-start, /wave-start, /board-audit, /ontology-librarian, /ontology-rebuild, /handoff, /wave-wrapup, /wave-retro, /annunaki, etc.) outline what it does across: (a) code/repo actions, (b) GitHub API (issues/PRs/projects/actions), (c) MCP calls, (d) any other external-service API calls. Add a dedicated ontology home for this lifecycle (e.g. ontology/lifecycle.md or a clearly-scoped section). Markdown is the source of truth — the generated .docx (sub-issue B) derives from it.

Part of #765.

Issue #765, opened 2026-06-21
P6W1: System documentation, architecture capture, onboarding & doc-generation pipeline


parametrization:
Umbrella for the documentation/architecture/onboarding batch (owner request 2026-06-21). All items worked this sprint EXCEPT the local-docker dev env (deferrable to no later than the first issue of the data-quality/P7 wave).

## Sub-issues
1. Lifecycle flow doc (markdown + ontology home) — the full phase→wave cycle as a slash-command flow.
2. Markdown→MS Office (docx/xlsx/pptx) generation pipeline + docs/ as office-doc home; lifecycle doc as first .docx.
3. Ontology: code is final arbiter of truth; enforce README + docs/ freshness on every PR.
4. Architecture diagram L1 — provider abstraction (Hetzner/Backblaze/GitHub/Cloudflare) data in/out of VPS containers.
5. Architecture diagram L2 — all containers, ingress/egress (container↔container + data in/out).
6. Architecture diagram L3 — per-container data/logic-flow models.
7. README — project-continuation usage flow.
8. README — setup prerequisites + dev/test workflow + per-repo purpose & inter-repo dependency map.
9. CONTRIBUTING.md — contribution model (data scientists / developers / dataset providers) + README link.
10. Local docker dev env replicating full staging (DEFERRED to ≤ first data-quality/P7 issue).

## Cross-cutting principles (owner directives)
- **Markdown is the single source of truth** for all Word/Excel/PowerPoint — Office docs are GENERATED from markdown via scripts, never hand-edited.
- **Code is the final arbiter of truth** for the ontology; on ontology update, READMEs + docs/ are checked and regenerated.
- **Every PR** double-checks that documentation (READMEs, docs/, ontology) reflects the change.
- Env-var documentation lists NAMES + purpose only, never values.

parametrization:
## P6W1 documentation/architecture/onboarding batch — completion summary

**All 8 in-scope items merged to main** (2 reviewers each, verified at source, green, no force-merges):

| Item | Issue | PR | Deliverable |
|------|-------|----|-------------|
| A | #766 | #780 | `ontology/lifecycle.md` — phase→wave slash-command flow |
| B | #767 | #778 | markdown→MS Office pipeline + `docs/office/` |
| C | #768 | #785 | code-as-truth + advisory doc-freshness gate (`.claude/lib/doc_freshness.py`, CI + pre-push, #684 mirror) |
[...]

Issue #768, opened 2026-06-21
Ontology: code is the final arbiter of truth; enforce README + docs/ freshness on every PR


parametrization:
Make documentation durable: (1) /ontology-rebuild treats CODE as the final arbiter of truth — when code disagrees with docs/ontology, code wins and the docs are flagged/updated. (2) On ontology update, check and update READMEs AND docs/ (regenerate Office docs from their markdown sources per sub-issue B). (3) Every PR double-checks that documentation (READMEs, docs/, ontology) reflects the change — add a PR checklist item and, where enforceable, a doc-drift gate/hook (prefer hook > skill > charter per the enforcement-hierarchy convention).

Part of #765.

PR #778, opened 2026-06-21
feat(docs): markdown→MS Office generation pipeline (#767)

### LINKED ISSUE #767: Markdown→MS Office (docx/xlsx/pptx) generation pipeline; lifecycle doc as the first docs/ office file
Establish docs/ as the home for generated Microsoft Office documents (docs/ already holds markdown like TOOLCHAIN.md — add an office-doc convention, e.g. docs/office/ or docs/*.docx). Create scripts that GENERATE Office files FROM markdown (single source of truth = markdown; the binary is never hand-edited). Tooling: pandoc (docx/pptx) and/or python-docx/python-pptx/openpyxl — pick and document. Generate sub-issue A's lifecycle markdown as the first .docx. Make regeneration reproducible (script + make target; optional CI check that the committed Office doc matches its markdown source). ALL future Word/Excel/PowerPoint derive from markdown via these scripts.

Part of #765.

parametrization:
## Summary

Establishes a reproducible **markdown → Microsoft Office** generation pipeline. Markdown is the single source of truth; the office binaries are generated artifacts, never hand-edited.

## What landed

- **Home convention** — generated office binaries live under `docs/office/`; markdown sources stay where they naturally live; `docs/office/office-docs.json` is the manifest mapping source → target → format. `docs/office/README.md` states the convention and the "never hand-edit" rule.
- **Engine: pandoc** — native `md → docx/pptx`, one tool, no per-format glue. `xlsx` is routed to `openpyxl` (tabular data, not prose); `python-docx`/`python-pptx` are documented escape hatches. `scripts/gen-office.sh` resolves pandoc from `$PANDOC` → `PATH` → the pip-bundled `pypandoc` binary, so `make docs` works without a system package.
- **Reproducible** — `SOURCE_DATE_EPOCH` is pinned to each source's last git commit time, so regeneration on the same pandoc version is byte-identical (verified).
- **`make docs`** regenerates every enabled manifest entry.
- **Demonstration** — `docs/office/TOOLCHAIN.docx` generated from `docs/TOOLCHAIN.md` (round-trip content verified: headings/body text present).
- `.gitattributes` marks `*.docx/*.xlsx/*.pptx` binary; TOOLCHAIN.md gains a **Document generation** subsection; cspell dictionary gains the office terms.

## #766 lifecycle doc

Pre-registered in the manifest with `"enabled": false`; flip to `true` and `make docs` renders `docs/office/lifecycle.docx` once `ontology/lifecycle.md` lands. Demonstrated on TOOLCHAIN.md per the issue's fallback.

## Sync-drift CI check — deferred

Implemented reproducibility (the precondition) but **deferred** the CI drift check itself: it requires pinning the pandoc version in CI because output bytes differ across pandoc versions. Tracked as a follow-up rather than shipping a flaky gate.

## Verification

- `pre-commit run --all-files` (commit + `--hook-stage pre-push`): all green.
- `shellcheck scripts/gen-office.sh`: clean.
- markdownlint on new/changed `.md`: 0 errors; new relative links resolve; `office-docs.json` parses.
- `make docs` round-trips and is byte-reproducible across runs.

Closes #767
Part of #765


parametrization:
## Release/build review — PR #778 (markdown → MS Office pipeline)

Reviewed at head_sha `6370be1128e6dd2b58f8ce58e5b20b06458c8e21` (read via `gh api .../contents?ref=…`, no parent checkout). Approving from the Release/build lens — clean, well-crafted, honestly documented.

**1. Zero-setup-on-pull — PASS.** `scripts/gen-office.sh` resolves pandoc `$PANDOC` → `command -v pandoc` → `python3 -c 'import pypandoc; print(pypandoc.get_pandoc_path())'` (with `2>/dev/null || true` swallowing the import/lookup failure), then guards with `[ -x "$PANDOC" ]` and a helpful three-option install error. No hook hard-depends on pandoc: the only CI touchpoint is `docs.yml`'s `**/*.json` config-lint, which JSON-parses the manifest (stdlib only) — it never invokes pandoc. `make docs` works on a fresh clone with just `pip install pypandoc-binary`.

**2. Reproducibility — PASS.** `SOURCE_DATE_EPOCH` is derived per-source from `git log -1 --format=%ct -- "$src"` (committer epoch), exported before each pandoc call, with a `0` fallback when there's no git history. Git-derived, not wall-clock — no `Date.now`-class nondeterminism. Note (non-blocking): because the epoch is pinned to the source's *last commit time*, the committed demo `TOOLCHAIN.docx` will read as "drifted" after this very PR merges (merging re-times `TOOLCHAIN.md`); that's inherent to commit-time pinning and is exactly what the deferred drift gate would reconcile on regenerate. Fine to ship.

**3. Markdown = source of truth — PASS.** "Never hand-edit; markdown is the single source of truth" is stated in the script header, the manifest `_comment`, `docs/office/README.md`, and `docs/TOOLCHAIN.md`. Generated binaries land under `docs/office/`, sources stay in place, and `office-docs.json` maps source→target→format. `.gitattributes` marks `*.docx/*.xlsx/*.pptx binary`. The #766 lifecycle entry is `"enabled": false` and the parser does `if not doc.get("enabled", True): continue` — `ontology/lifecycle.md` is absent at HEAD (confirmed 404) and is cleanly skipped, no error. Clean no-op until the source lands.

**4. Sync-drift deferral — AGREE.** Deferring the binary-drift CI gate is the right call: a gate that re-runs pandoc and byte-compares is only sound once the pandoc version is pinned in CI (bytes differ across versions), and the reproducibility precondition it depends on (SOURCE_DATE_EPOCH from git) is already built in here. Shipping a version-unpinned gate would be flaky. The follow-up is documented in prose (README + TOOLCHAIN), but no discrete tracking issue is filed — see TechDebt.

**5. zsh-safety — PASS.** `#!/bin/sh` shebang, executed (not sourced), invoked directly by the `docs:` make target. Pure POSIX: temp-file loop instead of a pipe subshell (correct `set -e`/`exit` propagation), TAB via `printf`, all vars quoted. `shellcheck -s sh` clean; zero bash-isms (`declare -A`, `${!arr[@]}`, arrays). zsh concern is moot for an executed-with-shebang script.

xlsx is documented-not-implemented (the `case` rejects it with a pointer to README's openpyxl guidance) and pptx is routed through pandoc but unexercised — both honestly scoped in README/TOOLCHAIN, so the title's "docx/xlsx/pptx" reads as roadmap, not a claim of present coverage. No objection.

---
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: File a discrete follow-up issue for the deferred office-binary drift CI gate (pandoc-version-pinned regenerate + byte-compare); currently only prose-documented, #768 is adjacent but broader.
Reproducibility mechanism (git-commit-time SOURCE_DATE_EPOCH), zero-setup pandoc fallback chain, and markdown-source-of-truth convention are all built in and correct; the script is POSIX/shellcheck-clean and zsh-safe. Deferring the drift gate is the correct release-engineering call since it requires a pinned pandoc version in CI to avoid flakiness, and the reproducibility precondition it depends on is already present. Approving; only open item is recording the deferred gate as a tracked follow-up.

parametrization:
## Review — PR #778 (markdown→MS Office generation pipeline, #767)

**Reviewer #2 of 2 — TPM / process + CI-parity lens.** Reviewed at head_sha `6370be1128e6dd2b58f8ce58e5b20b06458c8e21`, read via the contents API at that ref (no parent branch-switch).

### 1. Documentation coherence with #765 (markdown = source of truth) — PASS
The user's explicit requirement ("markdown = source of truth, generate Office from markdown via scripts") is stated unambiguously and in the right places:
- `docs/office/README.md` leads with a blockquote: *"Markdown is the single source of truth. The binaries here are produced from their markdown sources by `scripts/gen-office.sh`. Never hand-edit a file in this directory."*
- The same rule is repeated as a header comment in `scripts/gen-office.sh`, in the manifest `_comment`, and in `docs/TOOLCHAIN.md` § Document generation.
- The user-facing flow is documented: `make docs` (and the bare `scripts/gen-office.sh` equivalent) in README, TOOLCHAIN, and the Makefile `help`/`docs` targets. The `docs` target exists and shells the script.

### 2. CI / local parity nuance (#684 gap) — PASS, not widened
This PR touches **no** `.github/workflows/**` and adds **no** new CI check, so it cannot widen the pre-existing #684 gap (pre-commit here mirrors only cspell+actionlint of `docs.yml`; markdownlint/lychee/config-lint remain CI-only). The new `office-docs.json` falls under the existing recursive `**/*.json` config-lint trigger — an existing gate covers it, no new uncovered surface. The deferred binary↔markdown sync-drift gate is **recorded, not silently dropped**: it is documented as a deferred follow-up in both `docs/office/README.md` and `docs/TOOLCHAIN.md`, with the honest precondition (pandoc version must be pinned in CI before output bytes are stable). See TechDebt below for the one no-silent-caps refinement.

### 3. Manifest correctness — PASS
`docs/office/office-docs.json` is valid JSON. Entry 1: `docs/TOOLCHAIN.md` → `docs/office/TOOLCHAIN.docx`, `format: docx`, `enabled: true` — source exists at HEAD; the generated `.docx` is committed (19167 bytes). Entry 2 (#766 lifecycle): `ontology/lifecycle.md` → `lifecycle.docx`, `enabled: false` — source is correctly ABSENT at HEAD, and the generator's manifest reader `continue`s on `enabled:false` BEFORE the source-existence check, so it is a genuine clean no-op (registers intent without breaking `make docs`).

### 4. Scope — PASS, no creep
File set is docs/scripts/config only: `.cspell/project-words.txt`, `.gitattributes`, `Makefile`, `docs/TOOLCHAIN.md`, `docs/office/{README.md,office-docs.json,TOOLCHAIN.docx}`, `scripts/gen-office.sh`. Confirmed **nothing under `.claude/`**, so `ci.yml`'s `.claude`-scoped ruff/mypy/pytest jobs correctly did not trigger (and statusCheckRollup shows none). `scripts/gen-office.sh` is committed `100755` (executable). `.gitattributes` correctly marks `*.docx/*.xlsx/*.pptx binary`. cspell words added for the new office vocabulary; all 7 docs.yml checks (markdownlint, cspell, lychee, config-lint, actionlint, roster gate, commit-authors) are green.

### Process notes
- I did NOT exercise actual regeneration / byte-identity of the committed `.docx` against pandoc output — that build-correctness axis is reviewer #1's lens and out of my process scope.
- The `SOURCE_DATE_EPOCH`-from-git-commit-time reproducibility approach is sound in principle (pandoc honors it for OOXML timestamps); the residual cross-version variance is honestly disclosed.

---
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: deferred binary-drift CI gate is documented in README + TOOLCHAIN as a "tracked follow-up" but no GitHub issue is filed yet — recommend filing one so "tracked" is literally true (no-silent-caps); non-blocking.
Approving on the process/parity/manifest axes. The markdown-as-source-of-truth rule (#765) is stated clearly and consistently, the manifest is valid with the #766 entry a clean disabled no-op, scope is docs/scripts-only with no `.claude/` creep, and this PR does not widen the pre-existing #684 local-vs-CI parity gap. The only refinement is filing an actual issue for the deferred drift gate; the deferral itself is honestly documented, not dropped.

PR #789, opened 2026-06-21
fix(annunaki): correct §1 hook-active check for post-#625 dispatcher indirection (#788)

### LINKED ISSUE #788: tech-debt(annunaki): /annunaki §1 hook-active check is a stale false-negative post-#625 (dispatcher indirection)
## Problem

The `/annunaki` status-viewer skill's **§ 1 "Verify the hook is active"** check is a stale false-negative since #625.

It greps `settings.json` for the literal `annunaki_monitor`:

```bash
grep -c annunaki_monitor "$REPO_ROOT/.claude/settings.json"   # → 0
```

But post-#625 `annunaki_monitor` is no longer wired directly in `settings.json`. It is a **module dispatched by `post_dispatcher.py`**, registered in `_REGISTRY["Bash"]` (`.claude/hooks/post_dispatcher.py:79`). `settings.json` only references `post_dispatcher.py`. So the grep returns `0` and the skill reports **"monitoring is NOT active"** even though the monitor is fully live (confirmed this session: `traces.jsonl` shows active dispatch, and `errors.jsonl` is being written).

This was surfaced during 2026-06-21 session-start: the skill's self-check claimed the hook was inactive while it was demonstrably firing.

## Fix

Update the § 1 check to reflect the dispatcher indirection. The monitor is active iff **both**:
1. `post_dispatcher.py` is registered on `PostToolUse` Bash in `settings.json`, **and**
2. `annunaki_monitor` is present in `post_dispatcher.py`'s `_REGISTRY`.

```bash
REPO_ROOT="$(git rev-parse --show-toplevel)"
if grep -q post_dispatcher "$REPO_ROOT/.claude/settings.json" \
   && grep -q '"annunaki_monitor"' "$REPO_ROOT/.claude/hooks/post_dispatcher.py"; then
  echo "active"
else
  echo "NOT ACTIVE"
fi
```

Add a one-line note explaining the indirection so the check doesn't get "simplified" back to the broken single-grep form.

## Scope

Docs/skill-only change to `.claude/skills/annunaki/SKILL.md` (§ 1, and the `annunaki_monitor.py` wording in the intro). No behavior change to the monitor itself.

## Acceptance

- `/annunaki` § 1 reports **active** in the current (dispatcher-wired) repo state.
- The check still reports **NOT ACTIVE** if `post_dispatcher` is removed from settings OR `annunaki_monitor` is removed from the registry.

parametrization:
## Summary

Fixes the `/annunaki` status-viewer's **§ 1 "Verify the hook is active"** check, which has been a stale false-negative since #625.

The check grepped `settings.json` for the literal `annunaki_monitor`. Post-#625 the monitor is no longer wired directly there — it is a module dispatched by `post_dispatcher.py` (registered in `_REGISTRY["Bash"]`, `.claude/hooks/post_dispatcher.py:79`); `settings.json` only references `post_dispatcher.py`. So the grep returned `0` and the skill reported **"monitoring is NOT active"** while the monitor was demonstrably live (surfaced at 2026-06-21 session-start: `traces.jsonl` showed active dispatch and `errors.jsonl` was being written).

## Change

`.claude/skills/annunaki/SKILL.md` only — no behavior change to the monitor itself:
- § 1 now checks **both** legs of the indirection: `post_dispatcher` wired in `settings.json` **and** `annunaki_monitor` present in the dispatcher's `_REGISTRY`.
- Intro wording updated to describe the dispatch indirection, with an explicit "do not collapse back to a single grep" note to prevent regression.

## Verification

```
live state          → active       (was: 0 / "NOT ACTIVE")
monitor removed     → NOT ACTIVE   (negative case holds)
old broken grep     → 0            (the #788 false-negative)
```

## TechDebt: none — docs/skill-only correction; no new debt introduced.

Closes #788

### DIFF
--- a/.claude/skills/annunaki/SKILL.md
+++ b/.claude/skills/annunaki/SKILL.md
@@ -3,7 +3,7 @@ name: annunaki
 description: View Annunaki error monitor status — shows recent captured errors and monitoring health
 ---
 
-Display the current state of the Annunaki error monitor. This skill is the **status viewer** for the always-on error monitoring hook (`annunaki_monitor.py`).
+Display the current state of the Annunaki error monitor. This skill is the **status viewer** for the always-on error monitor (`annunaki_monitor`, dispatched on every `PostToolUse` Bash call via `post_dispatcher.py` — see § 1).
 
 > Note: all repo paths in bash blocks below are rooted at `$REPO_ROOT` to avoid cwd drift when the skill is invoked from a worktree or child-repo subdirectory (#149).
 
@@ -26,14 +26,19 @@ Pre-#625 both kinds shared `errors.jsonl` and dispatch traces (76% of the P4W1 l
 
 ### 1. Verify the hook is active
 
-Check that `annunaki_monitor.py` is registered in `.claude/settings.json` under `PostToolUse`:
+Confirm that `annunaki_monitor` runs on `PostToolUse` Bash. **Post-#625 it is NOT wired directly in `settings.json`** — it is a module dispatched by `post_dispatcher.py` (the single PostToolUse entry point), registered in that file's `_REGISTRY["Bash"]`. `settings.json` only references `post_dispatcher.py`. So a bare `grep annunaki_monitor settings.json` returns `0` and **falsely** reports the monitor inactive. Check both legs of the indirection instead:
 
 ```bash
 REPO_ROOT="$(git rev-parse --show-toplevel)"
-grep -c annunaki_monitor "$REPO_ROOT/.claude/settings.json"
+if grep -q post_dispatcher "$REPO_ROOT/.claude/settings.json" \
+   && grep -q '"annunaki_monitor"' "$REPO_ROOT/.claude/hooks/post_dispatcher.py"; then
+  echo "active"
+else
+  echo "NOT ACTIVE"
+fi
 ```
 
-If 0, warn the user that monitoring is not active and offer to wire it up.
+`active` requires BOTH legs: the dispatcher must be wired on `PostToolUse` Bash in `settings.json` AND `annunaki_monitor` must be present in its `_REGISTRY`. If it prints `NOT ACTIVE`, warn the user that monitoring is not active and offer to wire it up. (Do not collapse this back to a single `grep` of `settings.json` — that is the #788 false-negative.)
 
 ### 2. Read the error log
 


parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

I reviewed the diff at head ca13447 and independently verified the claim against the live repo. The fix is correct.

The root cause is accurate. Post-#625 the monitor is no longer wired directly in settings.json; it is a module dispatched by post_dispatcher.py, which is the single PostToolUse entry point. I confirmed both legs of the indirection — post_dispatcher.py is wired under the PostToolUse Bash matcher in settings.json (the command at lines 105-106), and "annunaki_monitor" is the first entry in _REGISTRY["Bash"] in post_dispatcher.py (line 79). The old check grepped settings.json for annunaki_monitor and returned 0, which is exactly the stale false-negative described in #788.

I ran the new two-leg check both ways in the live repo. The positive case prints active. I then simulated removal of annunaki_monitor from the registry and re-ran the check, and it correctly printed NOT ACTIVE. So the guard genuinely fails closed when the monitor is unregistered, not merely when the dispatcher reference is missing.

The change is truly docs-only. It touches a single file, .claude/skills/annunaki/SKILL.md — the intro wording and the §1 verification block. There is no change to post_dispatcher.py, the registry, or the monitor module, so there is no behavioral change to the monitor itself; only the status-viewer's self-check is corrected.

The regression note is solid. The inline parenthetical pointing at the #788 false-negative and the explicit instruction not to collapse the check back to a single grep of settings.json both serve as a re-simplification guard, so a future editor who is tempted to shorten it has the rationale in front of them.

One non-blocking observation for awareness only, no action needed for this PR. The settings.json leg uses a bare grep for post_dispatcher anywhere in the file rather than asserting it sits specifically under the Bash matcher. In practice the registry leg carries the Bash-specificity (annunaki_monitor lives in _REGISTRY["Bash"]), so the combined check is sound for a status-viewer diagnostic and there is no realistic false-positive path here.

All fifteen CI checks are green. Approving.

parametrization:
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Reviewed at head SHA ca13447. This is a viewer-only correction to the annunaki skill's section-1 hook-active check, and it holds up under independent verification.

I reproduced all three states the PR claims. The new two-leg check prints "active" against the live tree. Removing annunaki_monitor from the dispatcher registry flips it to "NOT ACTIVE", so the negative path genuinely fails closed. The old single grep of settings.json returns 0, confirming the #625-era false-negative the PR is fixing.

I also confirmed the indirection the new wording describes is real. post_dispatcher.py is wired into settings.json (four PostToolUse entries) and annunaki_monitor lives in _REGISTRY["Bash"] at post_dispatcher.py line 79 — exactly the two legs the check now inspects. The intro rewording and the explicit "do not collapse back to a single grep" regression note are accurate and worth keeping; they encode why the naive grep is wrong so a future edit does not reintroduce it.

As Release Coordinator I checked blast radius. The diff touches only .claude/skills/annunaki/SKILL.md, so there is no path by which it can alter the monitor's runtime behavior — it only changes how the status viewer reports liveness. Branch hygiene is correct, named A.Virtanen/0788-annunaki-hook-active-check, and the body carries a clean Closes #788. Commit authorship is Aino Virtanen plus the Claude co-author trailer.

All fifteen CI checks are green, including markdown lint, cspell, and the sync-drift gate. No concerns. Shipping this.

PR #791, opened 2026-06-21
ci(office): generated-Office-binary drift gate with pinned pandoc (#781)

### LINKED ISSUE #781: P6W1: CI drift gate for generated Office binaries (deferred from #767/PR #778)
## Context

PR #778 (Closes #767, Part of #765) shipped the markdown→MS Office generation pipeline (`scripts/gen-office.sh`, `make docs`, manifest `docs/office/office-docs.json`). Generated Office binaries are byte-reproducible — `SOURCE_DATE_EPOCH` is pinned to each source's last git commit time.

Both reviewers (Santiago Ferreira, Wanjiku Mwangi) approved the PR but flagged one tracked-follow-up that was only documented in prose (README + TOOLCHAIN), never filed as a discrete issue. Per the no-silent-caps discipline, "tracked follow-up" must be literally true — hence this issue.

## What's deferred

A CI **drift gate** that regenerates the committed Office binaries from their markdown sources and fails the build if the bytes differ (i.e. someone hand-edited a binary, or a source changed without regenerating). The reproducibility precondition the gate depends on is already built into the pipeline; the gate itself was deferred because:

- It requires the **pandoc version pinned in CI** — OOXML output bytes differ across pandoc versions, so an unpinned gate would be flaky.
- Shipping it now would be a red/flaky gate, which the org's full-parity + no-force stance (#684) prohibits.

## Acceptance

- [ ] Pin a specific pandoc version in CI (and document it alongside the ruff/actionlint/cspell pins).
- [ ] Add a CI job that runs `make docs` (or `scripts/gen-office.sh`) and `git diff --exit-code` over `docs/office/*.docx|*.xlsx|*.pptx`, failing on drift.
- [ ] Wire the same check into `.pre-commit-config.yaml` (push stage) so local⇄CI parity holds (#684), and register it in the sync-drift gate's classifier.
- [ ] Note: merging #778 itself re-times `TOOLCHAIN.md` (new commit time → new `SOURCE_DATE_EPOCH`), so the committed demo `TOOLCHAIN.docx` will read as drifted until first regeneration — the gate's first run should reconcile this (regenerate + commit) rather than fail on the inherited skew.

Part of #765. Follow-up to #767 / PR #778. Adjacent to #768 (ontology code-truth + doc-freshness enforcement), which is broader.

parametrization:
## Summary

Adds the CI drift gate for generated Office binaries that PR #778 deferred (it needs a pinned pandoc — OOXML output bytes differ across pandoc versions, so an unpinned gate would flake, which #684's no-flaky-gate stance prohibits).

## What's in it
- **`Office docs drift gate`** job (`.github/workflows/docs.yml`): installs the pinned pandoc via **`pypandoc-binary==1.17`** (bundles official **pandoc 3.9**, the same binary `gen-office.sh` resolves as its fallback), regenerates every enabled manifest entry, and `git diff --exit-code` over `docs/office/*.docx|*.xlsx|*.pptx`. `fetch-depth: 0` so `SOURCE_DATE_EPOCH` (source's last commit time) resolves.
- **Pre-push mirror** (`office-drift` hook in `.pre-commit-config.yaml`) for #684 local⇄CI parity.
- **`office-drift` kind** registered in `pre_commit_ci_sync.py` + **4 classifier tests**, so the sync-drift gate now *demands* the mirror; parent parity expectations updated (`_EXPECTED_KINDS`, `_OLD_KIND_PATTERNS`).
- **Pandoc pin documented** in `docs/office/README.md` alongside the ruff/actionlint/cspell pins.

## Inherited-skew reconciliation (the #781 acceptance note)
Regenerated `TOOLCHAIN.docx` with the pinned pandoc 3.9 and committed it, so the committed bytes equal what CI produces. The earlier committed binary was generated by a different pandoc and read as drifted.

## Verification (local)
- pandoc 3.9 output is **byte-deterministic** — two regenerations are identical (`cmp` clean).
- Gate is **GREEN** in-sync (regenerate == committed) and **RED** on a tampered binary (both proven locally).
- `pre_commit_ci_sync.py .` → `OK` (office-drift mirrored both sides); classifier suite **43 passed**; full lib suite **221 passed**; mypy clean; commit-stage ruff/actionlint/cspell green.

## TechDebt: none — completes a tracked #778 follow-up; no new debt.

Closes #781


parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Reviewed at head e3394ad. This closes the #778-deferred Office-binary drift gate cleanly, and the load-bearing question is answered green.

The single most important proof holds. The Office docs drift gate job ran and finished green, which is the real evidence that pinned pandoc 3.9 (bundled by pypandoc-binary==1.17) regenerates byte-identical output to the committed TOOLCHAIN.docx inside CI. That was the exact cross-platform byte-repro risk that made #778 defer this gate, so a passing job rather than a passing test is what I needed to see, and that is what landed. All sixteen checks on the PR are green.

The classifier logic is sound. The new office-drift kind maps to the tuple office-drift plus gen-office, and detection is an any-token-in-text test, so the kind fires on the CI side from the scripts/gen-office.sh run step and on the pre-commit side from both the office-drift hook id and its gen-office entry. Because the kind is classified on both sides, the sync-drift gate now demands the mirror exactly as #684 intends, and the four new tests cover both the per-side classification and the harmful-versus-mirrored drift cases. Parity expectations were updated in lock-step, with office-drift added to the OLD_KIND_PATTERNS map and to both OLD kinds sets, which is why the sync-drift parity job itself stayed green rather than tripping on an unclassified kind.

Sanity items all check out. The regenerate-and-diff step scopes git diff exit-code to the three binary globs only, so the hand-edited manifest and README are deliberately excluded; fetch-depth zero is present on the checkout, which the SOURCE_DATE_EPOCH-from-last-commit-time scheme requires; and the path filters add docs/office and scripts/gen-office.sh so a binary or source change actually triggers the gate. The pin is documented consistently across docs.yml, the pre-commit note, and docs/office/README.md, with a clear bump-all-three reminder.

Approving for merge. No tech debt, no follow-ups required.

parametrization:
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Reviewed at head e3394ad. This closes the #778-deferred Office-binary drift gate, and from a CI-reproducibility standpoint it is sound.

The dispositive check is green. The Office docs drift gate job passed in 7s, which means CI's pinned pandoc 3.9 (bundled by pypandoc-binary 1.17) regenerated TOOLCHAIN.docx to bytes identical to the committed binary, then git diff --exit-code found nothing. That green is exactly the cross-platform determinism proof that was missing when #778 deferred this, so the reason for deferral is now retired with evidence rather than assertion.

I independently traced the pin chain and it is internally consistent. scripts/gen-office.sh at head resolves its engine in the order $PANDOC, then pandoc on PATH, then the pypandoc-bundled binary via pypandoc.get_pandoc_path. The CI job installs only pypandoc-binary 1.17 and no system pandoc, so resolution falls through to the bundled pandoc 3.9, the same binary a local pip install of that pin yields. SOURCE_DATE_EPOCH is read from the source file's last commit time, and the job sets fetch-depth 0 so that history is present, which is the right call. The README now documents the pandoc 3.9 pin next to the ruff, actionlint, and cspell pins with a bump-all-together instruction.

False-fail surface is appropriately bounded. The regenerate-and-diff is scoped to the docx, xlsx, and pptx extensions only, so hand-edited manifest and README content cannot trip the gate, and the workflow path triggers add docs/office and scripts/gen-office.sh so a generator or source change actually re-runs the check. The sync-drift gate now registers the office-drift kind on both axes with four classifier tests covering the classified, mirrored, and un-mirrored-is-harmful cases, and the pre-push mirror in .pre-commit-config.yaml satisfies the #684 full-parity stance.

Branch hygiene is correct, the branch is A.Virtanen/0781-office-drift-gate and the body Closes #781. All sixteen checks are green and the branch is mergeable. Approving.

Issue #611, opened 2026-06-02
tech-debt(cross-repo-status.json): bare wave_{M}_* keys collide at phase boundaries — document convention + helper remove-key support


parametrization:
## Summary

`cross-repo-status.json` uses bare `wave_{M}_*` top-level keys for wave bookkeeping. These collide at every phase boundary: Phase 4 Wave 1 and Phase 3 Wave 1 both claim `wave_1_*`. Hit live for the first time at `/wave-scope 4 1` (2026-06-02).

## What happened

- P3W1's keys (`wave_1_scope`, `wave_1_wrap_status`, `wave_1_completed_at`, etc.) were still present when P4W1 scope reconciliation needed to write its own `wave_1_*` keys.
- Left in place, the stale keys would have: (a) fed P3W1's tier/assignment rows to the `post_wave_kickoff_comment.py` hook at P4W1 kickoff, (b) made `/wave-wrapup`/`/wave-retro` read a `wave_1_completed_at` from a month before the wave started.
- Resolution applied (owner-approved at /wave-scope): **overwrite convention** — P3W1's 7 keys were removed (record preserved in git history + `phase-3.md`), P4W1's keys written fresh. One-time migration script reused `upsert_status_keys.py`'s structural parsers.

## Compounding factors

1. **Mixed indentation:** P3W1-era keys use 1-space indent; the upsert helper's regexes require 2-space. The helper can neither update nor cleanly handle the older keys (related: #595).
2. **No delete support:** `upsert_status_keys.py` has no key-removal function; phase-boundary cleanup requires ad-hoc scripts.

## Acceptance Criteria

- [ ] The phase-boundary convention is documented (charter or `.claude/lib/` docstring): bare `wave_{M}_*` keys belong to the **current** phase; prior-phase keys are removed at the first `/wave-scope` of each new phase
- [ ] `upsert_status_keys.py` gains a `--remove-key` mode (or a companion `remove_status_keys.py`) so the next phase boundary doesn't need an ad-hoc script
- [ ] Indent-tolerance: helper regexes accept any leading whitespace (fixes the 1-space legacy keys; coordinate with #595)
- [ ] Alternative considered and explicitly rejected or adopted: phase-prefixed keys (`p4_wave_1_*`) — requires coordinated read-contract change across /wave-kickoff, /wave-wrapup, /wave-retro, /wave-scope + hooks

## Origin

[...]

PR #806, opened 2026-06-22
fix(lib)(P6W2): upsert_status_keys multi-line value replace — #736

### LINKED ISSUE #736: tech-debt(lib): upsert_status_keys.py fails replacing a multi-line existing value (sibling-finder mislocates)
## Summary

`.claude/lib/upsert_status_keys.py` aborts with `text-level upsert produced invalid JSON` when an **existing** top-level key being replaced (or a sibling near the insertion point) has a value that spans **multiple lines** (a pretty-printed object/array). The sibling-finder mislocates the insertion/replacement boundary inside the multi-line value, producing invalid JSON, so the helper validates-post-write and refuses (correctly — no corruption, but the write is blocked).

This is the same reproducer class as main#332 / #456 (compact-inline assumption), surfacing on the **replace-existing-multiline-value** path rather than insert.

## Reproducer (real, from /wave-scope 6 1, 2026-06-20)

`cross-repo-status.json` carried a stale **multi-line** `wave_1_scope` object (P4W1: `tier_1_data`, ~47 lines). Running:

```
python3 .claude/lib/upsert_status_keys.py cross-repo-status.json \
  "wave_1_scope_reconciled_at=..." "wave_1_repos_in_scope=[...]" \
  "wave_1_meta_issue=731" "wave_1_scope={...}"
```

failed with:
```
ERROR: text-level upsert produced invalid JSON: Expecting ',' delimiter: line 836 ...
```

Worked around by a one-off line-based Python splice that preserved the compact-inline keys.

## Impact

Every `/wave-scope` and `/wave-wrapup` that needs to **replace** a `wave_{M}_scope` (or any key) whose current value is stored multi-line hits this. The file is mostly multi-line pretty (only the few keys written by the helper itself are compact-inline), so the failure recurs whenever a prior-phase wave reused the same wave-number stored its scope as pretty JSON.

## Acceptance Criteria

- [ ] Helper correctly replaces a top-level key whose existing value spans multiple lines, preserving all other lines byte-for-byte (incl. compact-inline siblings).
- [ ] Regression test in `.claude/lib/tests/` covering: replace a multi-line object value; replace a multi-line array value; insert near a multi-line sibling.
- [ ] Helper either handles the case or fails with a precise, actionable message naming the offending existing-value span (not a generic line number).
- [ ] Mirrored test runs in pre-commit/pre-push + CI.

## Origin

`/wave-scope 6 1` (P6W1 scope reconciliation), 2026-06-20. Cross-ref main#332, #456.

parametrization:
## Summary
Fixes `upsert_status_keys.upsert_top_level_key` replace-in-place path, which spliced at the opener line's end (`m.end()`). Because `line_re` is line-anchored (no DOTALL), a key whose existing value spans multiple lines (a pretty-printed object/array — the common shape in `cross-repo-status.json`) left the value body dangling after the new compact value, producing invalid JSON that aborted the whole upsert batch (the real `/wave-scope 6 1` failure).

Fix: locate the true end of the existing value via the same `_find_value_end` helper used by the insert (#332) and remove (#611) paths; preserve a trailing comma only if the excised value had one. Single-line values are unaffected.

Closes #736

TechDebt: none — this is itself a tech-debt fix; no new debt introduced. Regression test added under `.claude/lib/tests/`.

## Reviewers
- Requestor: Lucas Ferreira (implementer)
- Reviewer: Weronika Zielinska
- Reviewer: Nurul Hakim

🧵 Implementer went idle (API 529 overload) after committing+pushing but before opening the PR; PR opened by orchestrator (throttle-takeover) to preserve complete work. Authored commit is Lucas Ferreira's.

### DIFF
--- a/.claude/lib/tests/test_upsert_status_keys.py
+++ b/.claude/lib/tests/test_upsert_status_keys.py
@@ -385,6 +385,120 @@ class UpsertUpdateExistingScalarTests(unittest.TestCase):
         self.assertEqual(out.count('"wave_1_branches"'), 1)
 
 
+class UpsertReplaceMultilineExistingValueTests(unittest.TestCase):
+    """main#736 — REPLACING an existing top-level key whose current value
+    spans multiple lines (pretty-printed object/array).
+
+    Pre-#736 the replace path spliced at the opener line's end (`m.end()`).
+    For a multi-line value the opener line ends with `{`/`[` and the value
+    body continues on following lines, so the old body was left dangling after
+    the new compact value — invalid JSON (`Expecting ',' delimiter`) that
+    aborted the whole upsert batch. This is the real `/wave-scope` failure
+    where a prior-phase `wave_{M}_scope` was stored as pretty JSON.
+    """
+
+    def test_replace_multiline_object_value(self):
+        """The /wave-scope reproducer: replace a multi-line wave_1_scope
+        object with a compact value. Surrounding keys (incl. a compact-inline
+        sibling) must be preserved byte-for-byte."""
+        text = (
+            "{\n"
+            '  "phase": "phase-4",\n'
+            '  "wave_1_scope": {\n'
+            '    "tier_1_data": ["#1", "#2"],\n'
+            '    "tier_2": {"a": 1}\n'
+            "  },\n"
+            '  "wave_1_meta_issue": 731\n'
+            "}\n"
+        )
+        out = upsert_top_level_key(text, "wave_1_scope", '"narrow"')
+        parsed = json.loads(out)
+        self.assertEqual(parsed["wave_1_scope"], "narrow")
+        # Surrounding keys intact, no duplicate, no truncation.
+        self.assertEqual(parsed["phase"], "phase-4")
+        self.assertEqual(parsed["wave_1_meta_issue"], 731)
+        self.assertEqual(out.count('"wave_1_scope"'), 1)
+        # Compact-inline siblings preserved byte-for-byte.
+        self.assertIn('  "phase": "phase-4",\n', out)
+        self.assertIn('  "wave_1_meta_issue": 731\n', out)
+
+    def test_replace_multiline_array_value(self):
+        text = (
+            "{\n"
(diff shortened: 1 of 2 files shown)


parametrization:
Requestor: Weronika Zielinska
Requestee: Lucas Ferreira
RequestOrReplied: Approved
TechDebt: none

Verified the replace path now excises through _find_value_end(text, m.end()) instead of splicing at the opener line end, fixing the dangling multi-line body that produced invalid JSON and aborted the batch. Single-line values are provably unaffected (_find_value_end returns opener_match_end unchanged when the line does not end with a bracket). Trailing-comma preservation is correct in both branches, since the multi-line walk consumes the comma and the single-line m.end() already includes it; the final-key no-comma case is handled. Regression tests are meaningful and well-targeted: multi-line object/array, final-key no-stray-comma, escaped-quote and bracket-in-nested-string depth tracking, and the nested same-name top-level guard. Scope is exactly #736; all 12 CI checks green.

parametrization:
Requestor: Nurul Hakim
Requestee: Lucas Ferreira
RequestOrReplied: Approved
TechDebt: none

Verified the diff at head independently. The replace path now excises through `_find_value_end` (depth + string-escape tracking) instead of splicing at the opener line's `m.end()`, so a multi-line pretty-printed object/array value is fully removed before the compact replacement — fixing the dangling-body invalid-JSON that aborted the batch. Comma preservation via `text[value_end - 1] == ","` is correct across all four shapes (single/multi x trailing-comma/final-no-comma), and single-line values return `m.end()` unchanged so the previously-working paths are untouched.
Test coverage is strong and the multiline-object case genuinely reproduces the pre-#736 failure: empty/nested objects, array values, JSON-final key with no trailing comma, braces/escaped quotes inside nested strings, and the top-level-position guard against a nested same-named key are all exercised. The only theoretical delta from the old `rstrip()` (single-line value with trailing whitespace after the comma) does not arise — the data file carries zero trailing whitespace and is machine-written by this same helper. CI: all 12 checks green.

Issue #795, opened 2026-06-21
P6W1 follow-up: widen mermaid render gate scope beyond docs/+ontology/ (5 blocks uncovered, from #794)


parametrization:
## Context

The #787 mermaid render gate (PR #794) scans `docs/*.md` + `ontology/*.md` — the minimum scope the issue specified. Reviewer (Nadia Khoury) noted that mermaid blocks also live **outside** that scope and can still ship a broken/blank diagram green:

- `README.md` — 1 block
- `.claude/team/charter.md` — 1 block
- `.claude/team/lifecycle.md` — 3 blocks

## Ask

Make the gap a deliberate decision, not an accident: either **widen** `scripts/check-mermaid.py`'s default scan to cover these files (e.g. all tracked `*.md` containing a ```mermaid block, minus the linter-excluded `.claude/memory/**`), OR add an explicit **whitelist/ignore** list documenting which files are intentionally out of scope.

Verify each newly-in-scope block actually renders (fix any that don't) before flipping the scope, so the widening lands green.

Part of #765 follow-ups. Surfaced in #794 review (Nadia).

Issue #787, opened 2026-06-21
P6W1 follow-up: CI mermaid-render gate (broken diagrams ship green — found in #786)


parametrization:
## Problem

PR #786 review surfaced a structural gap: **no CI job renders mermaid diagrams**, so a syntactically-broken mermaid block ships green. The Markdown-lint, cspell, and lychee checks all pass a non-rendering diagram. On #786, L3 block C.2 (alertmanager) embedded a literal `<unset>` in a node label — mermaid lexes `<` as a tag-open and aborts the diagram to a blank box — and it was caught ONLY because a reviewer (Aino) manually validated each block through the Mermaid Chart MCP. Without that manual pass it would have shipped a blank box into the rendered docs and the #767 Office pipeline.

As the architecture docs (L1/L2/L3) and lifecycle/usage-flow docs grow more mermaid-heavy, manual per-block MCP validation at review time is not a durable guarantee.

## Proposed

Add a CI job (mirrored into `.pre-commit-config.yaml` per #684) that extracts every ```mermaid fenced block from the docs and validates it renders — e.g. via `@mermaid-js/mermaid-cli` (`mmdc`) in a headless/CI-friendly mode, failing on any non-rendering block. Register the kind in the `pre_commit_ci_sync.py` classifier so local⇄CI parity holds. Scope: at minimum `docs/*.md` and `ontology/*.md`; consider whitelisting which files are scanned.

Note: a render check needs a browser/puppeteer in CI (mmdc uses headless Chromium), so weigh CI cost; an alternative is the mermaid parser-only validation if a lighter lib exists. Evaluate both.

Part of #765 follow-ups. Surfaced by #786 review (Aino). Related to #768 (doc-freshness enforcement — same "CI must catch what matters" theme).

PR #786, opened 2026-06-21
docs: add L3 per-container data/logic flow diagrams (#771)

parametrization:
## What

Adds `docs/architecture-l3.md` — the **L3** zoom below the L2 container-systems diagram (#770). For each of the 30 prod compose services it shows the internal data/logic flow and exactly what moves in/out: ports, volumes, env-derived connections, and upstream/downstream services.

## Structure (no silent caps — stated in the doc)

- **18 substantive containers get a full individual diagram:** `caddy`, `api`, `frontend`, `landing`, `user-service`, `isnad-graph-embed`; `neo4j`, `postgres`, `redis`, `user-postgres`, `user-redis`; `prometheus`, `alertmanager`, `grafana`, `loki`, `alloy`; `kafka`, `kafka-ui`.
- **Grouped (one pattern diagram + a table each)** to avoid near-duplicate boxes:
  - 5 exporters (`node` / `postgres` / `user-postgres` / `blackbox` / `kafka`)
  - 4 pipeline workers (`dedup` / `enrich` / `normalize` / `graph-load`) — a stage-chain diagram + worker-internal pattern + table
  - 3 one-shot/init containers (`user-service-migrate` / `kafka-init` / `loki-runtime-init`)
- A **coverage matrix** maps all 30 services to where each is documented.

## Provenance

Every per-container claim (ports, volumes, env, `depends_on`) traces to a cited `compose/docker-compose.prod.yml` line at `noorinalabs-deploy` **origin/main `273f220`** — cross-checked against the Caddyfile and `prometheus.prod.yml` at their origin HEADs. Each diagram/section carries a **Sources** note (the L2 provenance convention).

All **22 mermaid blocks validated** (Mermaid Chart MCP + local `mmdc`, same parser) — 0 failures.

## Also

- Resolves the L2 "unverified pending a read of the Prometheus config" note: the prod scrape job set is exactly `api`, `user-service`, `node-exporter`, `postgres-exporter`, `user-postgres-exporter`, `kafka` (target `kafka-exporter:9308`), `alloy`, `blackbox` — **`loki`/`grafana` are NOT scraped in prod.** Documented in a "Corrections to L2" section.
- `architecture.md` L3 bullet now links the new companion doc.
- New proper nouns added to `.cspell/project-words.txt`.

Kept pandoc-renderable (clean heading hierarchy) for the #767 Office pipeline.

## Validation

- pre-commit (commit + push stages): green
[...]

PR #807, opened 2026-06-22
fix(mermaid-gate): stop _RAW_ANGLE false-warning on valid <br/> label tags (#797)

### LINKED ISSUE #797: tech-debt(mermaid-gate): _RAW_ANGLE heuristic false-warns on valid <br/> label tags
## Follow-up from #796 (widen mermaid render gate scope, #795)

`scripts/check-mermaid.py` carries a footgun-warning heuristic (`_RAW_ANGLE`) that warns when a mermaid block contains raw `<`/`>` (which usually means a label that renders degraded). It does **not** special-case HTML line-break tags, so a perfectly valid `<br/>` inside a node label triggers an advisory warning.

- **Severity:** advisory only — it prints a warning, never fails the gate. Pre-existing; predates #796.
- **Surfaced by:** Aino during #796 (the widened scope now renders blocks that legitimately use `<br/>` for multi-line labels, e.g. the lifecycle flowchart).

## Ask
Teach the `_RAW_ANGLE` heuristic to ignore the HTML break tags mermaid actually supports (`<br>`, `<br/>`, `<br />`) before flagging raw angle brackets, so valid multi-line labels don't generate noise. Keep the warning for genuinely raw `<`/`>` in labels (the #786 `<unset>` class).

Out of scope for #795 (which was about *coverage*, not the heuristic). Low priority.

parametrization:
## Summary

Fixes the `_RAW_ANGLE` footgun heuristic in `scripts/check-mermaid.py` (the mermaid render gate, P6W1 #787/#794/#796) so it stops false-warning on valid HTML break tags inside mermaid node labels.

**Root cause:** the heuristic guarded the *opening* `<` of a break tag (`br\b` lookahead), but the *closing* `>` of `<br>`/`<br/>`/`<br />` still matched the `(?<![-=.])>` alternative — so a legit multi-line label (e.g. the lifecycle flowchart widened into scope by #795) fired a spurious advisory warning.

**Fix:** strip the break tags mermaid renders natively (`_BR_TAG`, any case/spacing) before the raw-angle scan, via a small `_has_raw_angle` helper. A genuinely unescaped angle bracket (the #786 `<unset>` class) still trips the warning — even when it sits next to a valid `<br/>` in the same block. The now-redundant `br\b` opening guard is dropped from `_RAW_ANGLE`. Advisory-only path; the render gate itself is unchanged.

**Test:** adds `.claude/lib/tests/test_check_mermaid.py` (collected by CI's pytest job + the pre-push mirror; loads the hyphenated script by path). Pins all three directions — break tags don't warn, genuine raw angles still warn (incl. alongside a valid `<br/>`), edge tokens (`-->`/`<--`/`==>`/`-.->`) never warn. This is the script's first regression test.

Verified locally: pytest green, ruff format/lint + mypy clean, and the full pre-push `mermaid-render` gate renders the whole doc set cleanly.

Reviewers: Nurul Hakim and Santiago Ferreira.

Closes #797

TechDebt: none — net-negative-risk (removes a false-positive in an advisory-only path; gate behavior unchanged) and adds the script's first regression test.

### DIFF
--- a/.claude/lib/tests/test_check_mermaid.py
+++ b/.claude/lib/tests/test_check_mermaid.py
@@ -1,0 +1,77 @@
+"""Tests for the check-mermaid raw-angle footgun heuristic (#797).
+
+The mermaid render gate (`scripts/check-mermaid.py`) prints an advisory warning
+when a fenced block carries a raw `<`/`>` in a label (which renders degraded —
+the #786 `<unset>` class). #797: the heuristic false-warned on valid HTML break
+tags (`<br/>`), which mermaid renders fine inside multi-line labels, because the
+closing `>` of the break tag matched. These tests pin both directions:
+
+  (a) a label with `<br>`/`<br/>`/`<br />` does NOT warn;
+  (b) a genuinely raw/unescaped angle bracket still warns; and
+  (c) mermaid edge tokens (`-->`, `<--`, `==>`) never warn.
+
+The script is loaded by file path because its name is hyphenated (not an
+importable module) and it lives under scripts/, not .claude/lib/.
+"""
+
+from __future__ import annotations
+
+import importlib.util
+import unittest
+from pathlib import Path
+
+_SCRIPT = Path(__file__).resolve().parents[3] / "scripts" / "check-mermaid.py"
+_spec = importlib.util.spec_from_file_location("check_mermaid", _SCRIPT)
+assert _spec is not None and _spec.loader is not None
+check_mermaid = importlib.util.module_from_spec(_spec)
+_spec.loader.exec_module(check_mermaid)
+
+_has_raw_angle = check_mermaid._has_raw_angle
+
+
+class BreakTagsDoNotWarn(unittest.TestCase):
+    def test_self_closing_br(self) -> None:
+        self.assertFalse(_has_raw_angle("flowchart TD\n  A[line one<br/>line two]"))
+
+    def test_bare_br(self) -> None:
+        self.assertFalse(_has_raw_angle("flowchart TD\n  A[line one<br>line two]"))
+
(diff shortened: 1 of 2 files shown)


parametrization:
Requestor: Santiago Ferreira
Requestee: Aisha Idrissi
RequestOrReplied: Approved
TechDebt: none

Diagnosis is correct: old _RAW_ANGLE suppressed only the opening `<` of a break tag via lookahead, but the closing `>` (preceded by `r`/`/`, not in [-=.]) still matched the second alternative — the false-warn source. Stripping `<br>`/`<br/>`/`<br />` (any case) with _BR_TAG before the scan removes both halves cleanly; dropping the now-dead `br\b` lookahead term is right.
No regression to the #786 `<unset>` catch — verified post-strip the literal `<` still trips, and edge tokens (-->, <--, ==>, -.->) stay suppressed. Tests pin all three directions including raw-angle-alongside-a-valid-br. Change minimal + targeted, all 18 checks green (Mermaid render gate + pytest included).

parametrization:
Requestor: Nurul Hakim
Requestee: Aisha Idrissi
RequestOrReplied: Approved
TechDebt: none

Correct root-cause: the closing angle of a break tag (preceded by `/`, not in [-=.]) slipped past the old second alternation. Stripping full break tags via _BR_TAG (IGNORECASE) before the scan handles <br>/<br/>/<br /> any case, and dropping the now-redundant br-lookahead from _RAW_ANGLE is clean. Genuine raw angles (<unset>, lone >, raw angle alongside a valid <br/>) still warn; edge tokens (-->, <--, ==>, -.->) stay excluded. Tests pin both directions; all 19 checks green incl. Mermaid render gate, pytest, mypy, ruff.

PR #824, opened 2026-06-23
tech-debt(mermaid-gate): anchor render scratch dir under $HOME, not cwd (#817)

### LINKED ISSUE #817: tech-debt(mermaid-gate): check-mermaid.py workdir is cwd-relative — snap mmdc can't read it under /tmp (pre-push fails from /tmp worktrees)
## Problem
The pre-push `mermaid-render` gate (`scripts/check-mermaid.py`) writes its `.mmcheck` temp working files **cwd-relative**. The local `mmdc` is snap-confined (`/snap/bin/mmdc`) with a private `/tmp` namespace, so when an agent worktree cwd is under `/tmp` mmdc literally cannot see the temp files → "Input file doesn't exist" for ALL blocks (content-independent false-fail). Confirmed P6W2 by Nadia/Weronika/Wanjiku/Santiago: identical `.mmd` renders fine from $HOME, fails from /tmp.

## Impact
Any PR touching markdown triggers the whole-doc-set scan and false-fails the mermaid pre-push gate from a /tmp worktree. Forced agents to push from $HOME worktrees this wave. CI is unaffected (non-/tmp runner). See memory `feedback_tmp_worktree_prepush_falsefail`.

## Fix options
1. Have `check-mermaid.py` place its workdir under `$HOME` or `$TMPDIR` (snap-readable), not cwd-relative. (preferred — durable, makes the gate worktree-location-agnostic)
2. Standardize the wave worktree convention off `/tmp` (partially done: spawn briefs now use non-/tmp paths).

Note: the sibling tmp-path pytest false-fail (`test_block_stale_tmp_message_file::test_non_tmp_path_not_matched`) was already fixed in #802/PR #815. This issue is the remaining mmdc half.

parametrization:
## What & why

`scripts/check-mermaid.py` (the pre-push `mermaid-render` gate) created its
`.mmcheck-*` render scratch dir **cwd-relative** (`tempfile.mkdtemp(dir=".")`).
The local `mmdc` is the **mermaid-cli snap**, which runs in a private mount
namespace where `/tmp` is NOT the host `/tmp`. When an agent worktree's cwd is
itself under `/tmp`, the scratch tree landed inside the snap's private `/tmp`
and mmdc reported `Input file "...doesn't exist"` for **every** block — a
content-independent false-fail of the whole gate. This forced agents to push
only from `$HOME` worktrees all of P6W2. CI is unaffected (its `npx` mmdc is
unconfined and HOME is set on the runner).

## Fix (issue option 1 — preferred, durable)

Add `_workdir_base()` returning a **snap-readable base**: `$HOME` when usable,
else the platform temp dir (`tempfile.gettempdir()`, which honours `$TMPDIR`).
Anchor the scratch dir there instead of cwd. The gate is now
worktree-location-agnostic without weakening any check — it still renders every
block via the same `mmdc` path and the same error-marker/exit-code detection.

## Acceptance evidence

Render of a real doc with a mermaid block (`.claude/team/charter.md`), the gate
invoked from a **/tmp-rooted cwd** (the bug scenario) and from **$HOME**:

| cwd | script | result |
|-----|--------|--------|
| `/tmp/repro-0817` | HEAD (old, cwd-relative) | **FAIL** — `Input file "/tmp/repro-0817/.mmcheck-…/b1.mmd" doesn't exist` |
| `/tmp/repro-0817` | this PR (`$HOME`-anchored) | **PASS** — `all mermaid blocks render cleanly` (rc=0) |
| `$HOME/wt-0817`   | this PR | **PASS** — `all mermaid blocks render cleanly` (rc=0) |

The full `mermaid-render` pre-push hook also passed on this push (whole-doc-set
scan, from a `$HOME` worktree).

## Tests

Adds 3 unit tests (`WorkdirBaseIsSnapReadable`) pinning that `_workdir_base()`
is absolute, anchored under `$HOME` when HOME is set, and never the cwd marker
`.`. Full suite: 14 passed.

## Gates

ruff-format / ruff-lint / mypy / pytest / sync-drift / mermaid-render — all
green locally and at pre-push.

Closes #817

### DIFF
--- a/.claude/lib/tests/test_check_mermaid.py
+++ b/.claude/lib/tests/test_check_mermaid.py
@@ -17,6 +17,8 @@ importable module) and it lives under scripts/, not .claude/lib/.
 from __future__ import annotations
 
 import importlib.util
+import os
+import tempfile
 import unittest
 from pathlib import Path
 
@@ -27,6 +29,41 @@ check_mermaid = importlib.util.module_from_spec(_spec)
 _spec.loader.exec_module(check_mermaid)
 
 _has_raw_angle = check_mermaid._has_raw_angle
+
+
+class WorkdirBaseIsSnapReadable(unittest.TestCase):
+    """#817: the render scratch dir must be anchored under a snap-readable base
+    ($HOME), never cwd-relative. A cwd under /tmp put the old `dir="."` scratch
+    tree inside the mmdc snap's private /tmp namespace → "Input file doesn't
+    exist" for every block. These tests pin that the base is absolute, under
+    $HOME when HOME is set, and never the cwd marker `.`.
+    """
+
+    def _with_home(self, home: str | None) -> str:
+        old = os.environ.get("HOME")
+        if home is None:
+            os.environ.pop("HOME", None)
+        else:
+            os.environ["HOME"] = home
+        try:
+            return check_mermaid._workdir_base()
+        finally:
+            if old is None:
+                os.environ.pop("HOME", None)
+            else:
+                os.environ["HOME"] = old
+
+    def test_anchors_under_home_not_cwd(self) -> None:
+        with tempfile.TemporaryDirectory() as home:
+            base = self._with_home(home)
+            self.assertEqual(os.path.realpath(base), os.path.realpath(home))
+
+    def test_base_is_absolute_never_cwd_relative(self) -> None:
+        with tempfile.TemporaryDirectory() as home:
+            base = self._with_home(home)
+        self.assertTrue(os.path.isabs(base))
+        self.assertNotEqual(base, ".")
+        self.assertFalse(base.startswith("."))
 
 
 class BreakTagsDoNotWarn(unittest.TestCase):
--- a/scripts/check-mermaid.py
+++ b/scripts/check-mermaid.py
@@ -117,6 +117,27 @@ def _default_files() -> list[Path]:
     return sorted(files)
 
 
+def _workdir_base() -> str:
+    """Base dir for the render scratch tree — snap-readable, never cwd-relative.
+
+    The local `mmdc` is the mermaid-cli *snap*, which runs in a private mount
+    namespace where `/tmp` is NOT the host `/tmp`. The earlier behaviour anchored
+    the scratch dir under the cwd (`dir="."`); when an agent worktree's cwd is
+    itself under `/tmp`, that scratch dir lands in the snap's private `/tmp` and
+    mmdc reports `Input file doesn't exist` for *every* block — a content-
+    independent false-fail of the whole gate (#817). The snap can read the real
+    `$HOME`, so anchor the scratch tree there to make the gate location-agnostic.
+
+    Fall back to the platform temp dir (`$TMPDIR`/…) only when `$HOME` is unusable
+    (e.g. a CI runner without HOME), where mmdc is the unconfined `npx` build and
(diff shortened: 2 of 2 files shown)


parametrization:
Requested reviewers (charter 2-reviewer format): **Aino Virtanen** (peer) and **Wanjiku Mwangi** (secondary). Verify against the artifact — the load-bearing change is `_workdir_base()` + the `dir=` argument to `tempfile.mkdtemp` in `scripts/check-mermaid.py`. Acceptance evidence (BEFORE/AFTER from a /tmp cwd vs $HOME) is in the PR body; reproduce with a detached worktree under /tmp if you want to witness the false-fail.

parametrization:
Requestor: Wanjiku Mwangi
Requestee: Santiago Ferreira
RequestOrReplied: Approved
TechDebt: none

Verified against the diff at head 8ba8dc6, not the body.

The fix is correct and does not weaken the gate. `_workdir_base()` returns `$HOME` when it is a usable absolute directory, else `tempfile.gettempdir()` (which honours `$TMPDIR`); the scratch tree is anchored there instead of `dir="."`. Every block still renders through the same `mmdc` path with the same error-marker / exit-code detection, so the change is purely about *where* the scratch tree lands, not *whether* a block is checked. The `shutil.rmtree(workdir, ignore_errors=True)` cleanup in the `finally` is preserved, so anchoring under `$HOME` does not leak scratch dirs.

The snap private-/tmp-namespace diagnosis matches the symptom (content-independent "Input file doesn't exist" for every block from a /tmp-rooted cwd) and the acceptance table demonstrates the old code fails / new code passes from the same /tmp cwd, which is the load-bearing evidence.

The three new tests are behavioral, not tautological: they pin the base as absolute, anchored under `$HOME` when HOME is set, and never the cwd marker `.` — exactly the property the bug violated. CI is correctly reasoned as unaffected (unconfined npx mmdc). All checks green at head. Approving.

parametrization:
Requestor: Aino Virtanen
Requestee: Santiago Ferreira
RequestOrReplied: Approved
TechDebt: none

Reviewed the diff at head `8ba8dc6`. The fix adds `_workdir_base()` and anchors the `.mmcheck-*` render scratch tree there instead of `tempfile.mkdtemp(dir=".")`.

Verified the gate is not weakened: the render path, the `mmdc` invocation, the `puppeteer.json` no-sandbox arg, and the per-block error-marker / exit-code detection are all unchanged — only the *base directory* of the scratch tree moves off cwd. So the set of blocks validated and the pass/fail criteria are identical; this is purely a location change.

The new base is genuinely snap-readable: `os.path.expanduser("~")` returns the real `$HOME`, which the mermaid-cli snap can read (the bug was that a cwd-under-`/tmp` scratch tree landed in the snap's private `/tmp` namespace). The `tempfile.gettempdir()` fallback only fires when `$HOME` is unusable — which is the CI case where `mmdc` is the unconfined `npx` build, so the snap-namespace constraint does not apply there. The guard (`home and home != "~" and os.path.isdir(home)`) correctly rejects the unexpanded `~` marker and a non-existent HOME.

The three `WorkdirBaseIsSnapReadable` tests are non-tautological: they drive `_workdir_base()` with a real temp `$HOME` set/restored and assert the result resolves under that HOME, is absolute, and is never the cwd marker `.` — exactly the invariant the bug violated. `os`/`tempfile` imports are already present in the test module. The HOME save/restore is correctly scoped in a `finally`.

Approving.

PR #829, opened 2026-06-23
hooks(#663): extend gh-command parser invariant to gh workflow/gh api class

### LINKED ISSUE #663: charter/standards: shared-parser invariant for gh-command hooks (scope to flag values + resolve ambient repo, never silently drop) — owner-adopted P4W7

parametrization:
## Summary

Closes #663 by closing its last enumerated gap. The `gh issue`/`gh pr`-scoped half of the gh-command parser invariant already landed (commit `9356b4e`: charter `hooks.md` § 7, the three shared parsers, the pytest gate, #659/#661 fixed). Both that gate and § 7 explicitly **deferred** the `gh workflow`/`gh api` value-flag parser class — "the same latent class but out of the stated scope — an enumerated follow-up, not yet enforced." This PR migrates that class onto the shared parsers and folds it into machine-enforcement, so the invariant now covers **all** gh-command hooks (the title of #663).

## What changed

- **`warn_ghcr_image.py`** — dropped the ad-hoc `re.search(r"-R\s+...")` repo extractor (the exact #650/#659/#661 flag-value-leak bug class, named in the gate docstring as the deferred example). Now uses:
  - `_repo_flag_parse.extract_repo` — all four `-R X` / `-R=X` / `--repo X` / `--repo=X` forms, and an `-R`-shaped token inside a quoted `--field` value no longer leaks;
  - `_shell_parse.resolve_repo_short_name` — resolves the flag-**omitted** ambient git-context case (mirroring gh) instead of falling through; never silently drops;
  - `_shell_parse.is_gh_subcommand` — command-position shape detection (a `gh workflow run` phrase inside a quoted arg / heredoc is not an invocation).
- **`block_gh_pr_review.py`** — replaced the private `re.split` segment-splitter with the shared `tokenize` + `strip_heredocs` + `iter_command_segments` + `find_gh_subcommand`. This also **corrects a pre-existing over-block**: a `gh pr review` phrase inside a heredoc body (e.g. `step 1; gh pr review`) previously false-blocked; it now correctly allows. The historical `re.split` path is retained only as the malformed-quote fallback.
- **`test_gh_command_parser_invariant.py`** — broadened classification + scope to include `gh workflow`/`gh api`; added a `warn_ghcr_image` regression guard; raised the vacuous-green min-count guard `6 → 10` (live classified set is 14).
- **`charter/hooks.md` § 7** — codified the broadened scope (`gh issue`/`gh pr`/`gh workflow`/`gh api`); retired the "deferred follow-up, not yet enforced" language; extended the backing-class list and provenance.
- **New `test_warn_ghcr_image.py`** — the hook previously had zero coverage. 16 tests: shape detection (incl. quoted-phrase + malformed-quote fail-open), flag-value scoping across all four forms, the quoted-`-R`-no-leak guard, ambient resolution (mocked — see note), and warning behavior.

## Test evidence

- Full hook + lib suite green: **1782 passed, 68 subtests** (`ENVIRONMENT=test pytest .claude/hooks/tests/ .claude/lib/tests/`).
- Migrated-hook + gate subset: 48 passed.
- pre-commit + pre-push clean: ruff-format, ruff-lint, mypy, cspell, the pre-commit⇄CI drift gate, and the full pytest/skills suites all pass (verified via `git push`, no `--no-verify`).

> Note on the ambient-resolution tests: they mock `resolve_repo_short_name` rather than rely on a real temp dir, because under a `git push` env `GIT_DIR`/`GIT_WORK_TREE` are exported and a `git -C <tmpdir> remote get-url origin` subprocess resolves the *parent* repo (the GIT_DIR-leak hazard) — the first push of this branch failed exactly that way before the fix.

## Reviewers
- Peer reviewer: **Santiago Ferreira**
- Secondary reviewer: **Nadia Khoury**

Closes #663


parametrization:
Requestor: Nadia Khoury
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Verified against the diff and the source at head `c71b773` via `gh` (no parent checkout). This is a clean close of the last enumerated gap in #663.

**No behavior regression.** `block_gh_pr_review` swaps the private `re.split` segmenter for `strip_heredocs` + `tokenize` + `iter_command_segments` + `find_gh_subcommand`, keying the match on the command-position first token (`found[1][:2] == ["pr", "review"]`) — the same intent the old `re.match` at segment start had, now with heredoc-awareness and the line-continuation/#287 fixes the shared path carries. The historical `re.split` survives as the malformed-quote fallback (over the heredoc-stripped text), so un-tokenizable input keeps its prior disposition. The one intentional behavior delta — a `gh pr review` phrase inside a heredoc body no longer false-blocks — is a correct over-block removal, and the renamed test pins it.

**The ad-hoc regex is genuinely gone.** `warn_ghcr_image` deletes `extract_repo_from_command` and its `re.search(r"-R\s+...")` extractor wholesale; the only regex left is the `DEPLOY_PATTERNS` shape matcher (not a flag-value capture). Repo resolution now flows flag-value-first through `_repo_flag_parse.extract_repo` (all four `-R`/`--repo` space+equals forms, and a `-R`-shaped token inside a quoted `--field` value no longer leaks because it arrives as a single shlex token), then the flag-omitted ambient case via `resolve_repo_short_name`, never silently dropping. Shape detection moves to `is_gh_subcommand(tokens, "workflow", "run")`, so a quoted-arg phrase is no longer treated as an invocation. All call-site signatures line up with the shared parsers.

**The new tests are non-tautological.** The flag-value scoping tests run `extract_repo` for real across every surface form; the no-leak guard genuinely exercises the parser (only the unstable ambient resolver is patched to None, so a non-None result would prove a leak). The mock on `resolve_repo_short_name` is well-justified and the rationale is exactly right — under a `git push` env the leaked `GIT_DIR`/`GIT_WORK_TREE` would make a real `git -C <tmp> remote get-url` resolve the parent repo (the documented leak hazard that broke the first push), so mocking is the correct isolation, not a tautology. The vacuous-green floor raised `6 -> 10` against a live set of 14 keeps real margin.

**Migration-complete claim holds.** The gate classifier broadens to `gh issue|pr|workflow|api`, adds a `warn_ghcr_image` regression guard, and runs in the already-mirrored pytest job. The full check rollup is green end to end (19/19, incl. the pytest suite and the pre-commit-CI sync-drift gate), so the broadened scope is machine-enforced, not just asserted in prose. Charter § 7 is updated consistently — scope, backing-class lineage, and the retired "deferred follow-up" language all match the code.

Nice work closing this without scope creep. Approving.

parametrization:
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Reviewed against the diff at head (c71b773), with the shared-parser APIs verified directly in `_shell_parse.py` / `_repo_flag_parse.py`.

No ad-hoc parser left. In `warn_ghcr_image.py` the old `re.search(r"-R\s+...")` extractor and the `\bgh\s+workflow\s+run\b` shape regex are both removed: shape detection now goes through `is_gh_subcommand(tokens, "workflow", "run")` and repo extraction through `_repo_flag_parse.extract_repo` + `_shell_parse.resolve_repo_short_name`. The only regex remaining is `DEPLOY_PATTERNS` (a content-shape match, explicitly allowed by the invariant). In `block_gh_pr_review.py` the private `re.split` segmenter is dropped for `tokenize` + `strip_heredocs` + `iter_command_segments` + `find_gh_subcommand`, with `re.split` retained only as the malformed-quote fallback.

API signatures check out against source: `find_gh_subcommand(segment) -> ([], [topic, action, ...])` so `found[1][:2] == ["pr", "review"]` is the correct topic/action test; `is_gh_subcommand(tokens, *verbs)`, `extract_repo(command)`, and `resolve_repo_short_name(input_data)` all match their call sites.

The one deliberate behavior change is a correct fix, not a regression: a `gh pr review` phrase inside a heredoc body (e.g. `step 1; gh pr review`) previously false-blocked via the private `;`-split and now correctly allows, because heredoc bodies are data fed to a command, not executed shell. The corresponding test flipped from asserting-block to asserting-allow, which is the intended direction. Safety is preserved — a real `gh pr review` invocation in command position still blocks.

Tests are non-tautological: the quoted-`-R`-no-leak test mocks ambient resolution to None and asserts a None result, so a leak of the quoted `foo/bar` would fail it; the ambient-mock rationale (GIT_DIR leak under a push env) is well-founded. The classifier scope is broadened to `gh issue|pr|workflow|api` and the vacuous-green guard is raised 6 to 10 against a live set of 14, with a regression guard pinning `warn_ghcr_image.py` in scope. Charter § 7 and the gate docstring are updated consistently and the deferred-follow-up language is retired.

Full suite green (1782 passed per the PR, and all 19 CI checks SUCCESS here), mergeable. Approved.

PR #833, opened 2026-06-23
tech-debt(test): decouple pre-commit-CI parity test from stale child checkouts (#816)

### LINKED ISSUE #816: tech-debt(test): test_pre_commit_ci_sync expected-map drifted from child configs (cspell) — fails in parent checkout only

parametrization:
## What

Root-fix for #816 — **decouples** `test_pre_commit_ci_sync`'s documented-expectations from physically-present (potentially stale) child checkouts, instead of tracking their stale state. Supersedes the closed PR #826.

Closes #816.

## Why (corrected root cause)

Owner confirmed `origin/main` of the child repos already carries the cspell work, so the documented map was **correct** — the parent-checkout failure was purely the **parent's stale local child checkouts** (`noorinalabs-isnad-graph` behind origin 207, `noorinalabs-user-service` behind 83). A hardcoded kind-snapshot of independent, gitignored child repos is not a property of *this* repo: it false-fails the parent-only pre-push suite whenever a child checkout lags origin, and would keep drifting (compounding with #744).

## Decouple mechanism (not a revert)

- `_EXPECTED_KINDS` keeps a documented snapshot for the **parent (`.`) only** — the sole config version-controlled *in* this repo, hence checkout-stable. The 7 child snapshots are **removed** (not changed to stale values).
- `test_new_matches_documented_expectations` now asserts the parent only.
- Present child configs are covered by the checkout-**independent** OLD==NEW parity invariant `test_new_equals_old_on_present_configs` — it compares two classifiers on the same bytes, so it holds no matter how stale a child checkout is.
- New `_CHILD_REPO_DIRS` tuple locates sibling configs **for the parity scan only** (never a kind snapshot); absent children are skipped, so CI (parent-less checkout) cannot false-fail.

Net: the test is correct regardless of how stale the local child checkouts are, and still exercises every present child config via the parity invariant.

## Test evidence

**From the parent checkout WITH the stale children present** (the original failure mode), edited test applied read-only then restored clean:

```
# full file
55 passed, 12 subtests passed in 0.84s
# ParityAcrossRealConfigs -v
test_new_equals_old_on_present_configs PASSED
test_new_matches_documented_expectations PASSED
2 passed, 9 subtests passed
```

The 7 extra subtests vs the worktree run (5) confirm all 7 present children are exercised by the checkout-independent parity test. Worktree (children absent) and full pre-commit + pre-push gates (ruff, mypy, full pytest, mermaid, office-drift, memory-budget, doc-freshness) all green on push; no `--no-verify`, per-commit Aino identity.

## Reviewers
- Peer: Wanjiku Mwangi
- Secondary: Santiago Ferreira

### DIFF
--- a/.claude/lib/tests/test_pre_commit_ci_sync.py
+++ b/.claude/lib/tests/test_pre_commit_ci_sync.py
@@ -734,17 +734,23 @@ def _old_kinds_from_ci(text: str) -> set:
     return kinds
 
 
-# Documented expected per-repo kind-sets (#748 D3 parity proof). Authoritative
-# 8-repo OLD-vs-NEW comparison was run at PR time against the real configs in a
-# full parent checkout: the NEW structural classifier produces an IDENTICAL
-# (pre-commit, CI) kind-set to the OLD line-scanner for ALL 8 repos — no
-# divergence. The false-positive class the structural parse removes is therefore
-# LATENT: no real config exercises a step-name/comment false-match today (proven
-# directly by FalsePositiveDemo). These expectations encode that NEW==expected;
-# ParityAcrossRealConfigs additionally asserts OLD==NEW wherever the config is
-# present (the parent always; sibling child repos only in a full multi-repo
-# checkout — they are independent gitignored repos absent from this repo's CI
-# checkout, so each is skipped when not present rather than false-failing).
+# Documented expected kind-set (#748 D3 parity proof). The ONLY repo whose
+# configs are version-controlled IN this repo is the parent (noorinalabs-main),
+# so it is the only one with a stable, hardcoded kind snapshot. The NEW
+# structural classifier must reproduce this snapshot exactly, and must equal the
+# retired OLD line-scanner on it — the false-positive class the structural parse
+# removes is therefore LATENT (no real config exercises a step-name/comment
+# false-match today, proven directly by FalsePositiveDemo).
+#
+# Child repos are deliberately NOT given a documented snapshot (#816, relates to
+# #744). They are independent gitignored repos whose local checkout beside the
+# parent can sit on any branch / lag origin by an arbitrary amount, so a
+# hardcoded snapshot of THEIR files is not a property of this repo: it false-fails
+# the parent-only pre-push suite whenever a child checkout is stale (the original
+# #816 symptom — isnad-graph/user-service checked out behind origin). Present
+# child configs are instead covered by the checkout-INDEPENDENT OLD==NEW parity
+# invariant (test_new_equals_old_on_present_configs), which needs no snapshot and
+# is correct regardless of how stale the local child checkouts are.
 _EXPECTED_KINDS = {
     ".": {
         "precommit": {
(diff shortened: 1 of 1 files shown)


parametrization:
**Review requested** — charter 2-reviewer format.

- **Peer reviewer:** Wanjiku Mwangi
- **Secondary reviewer:** Santiago Ferreira

Root-fix for #816 (supersedes closed #826). Please review against the artifact (diff at PR head). Key points: the documented snapshot is now parent-only (checkout-stable); child configs are covered by the checkout-independent OLD==NEW parity invariant; child snapshots were removed, NOT reverted to stale values. Acceptance evidence (parent checkout with stale children present) is in the PR body.

parametrization:
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Reviewed against the diff at head (914cf96), read the full test-body tails via gh api contents. This is the right fix, and a clear improvement over the now-closed expedient version I reviewed previously (which tracked the stale on-disk child state).

The decoupling is sound on every axis the review asked for:

- Checkout-stable snapshot: `_EXPECTED_KINDS` now holds the parent (`.`) only — the one repo whose configs are version-controlled in this repo — so the hardcoded snapshot is a genuine property of this checkout and cannot drift with a sibling's local revision.
- Snapshot test scoped to the parent: `test_new_matches_documented_expectations` iterates `_documented_roots()` (which derives solely from `_EXPECTED_KINDS`, i.e. the parent), and crucially retains the `assertIn(".", seen, ...)` guard so the test can never vacuously pass if the snapshot map is emptied.
- Present children covered by the checkout-independent invariant: `test_new_equals_old_on_present_configs` iterates `_present_config_roots()` (parent always + each child whose checkout is present) and compares the NEW structural classifier against the retired OLD line-scanner on the SAME bytes, for both pre-commit and CI. Because it compares two classifiers over identical file content, it holds regardless of how stale a child checkout is — exactly the staleness-immune guarantee that the original #816 symptom needed.
- No false-fail in the parent-less CI run: both `_documented_roots()` and `_present_config_roots()` gate child inclusion on `pc.is_file()`, and `_present_config_roots()` seeds the parent unconditionally. In CI (no children present) the parity test scans only the parent and the snapshot test only the parent — neither can false-fail.
- `_CHILD_REPO_DIRS` enumerates all 7 children — isnad-graph, user-service, deploy, design-system, data-acquisition, isnad-ingest-platform, landing-page — including deploy as called out.

One non-blocking observation: `_CHILD_REPO_DIRS` is a manual-sync list against the org CLAUDE.md Repository Map, so a future 8th child must be added here to be parity-scanned. This is benign — an omission would only mean a present new child is not scanned, never a false-fail — and the comment already flags the keep-in-step requirement. Not a blocker.

All 12 CI checks green and mergeable. Approved.

parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: a third hardcoded child-repo list now exists (_CHILD_REPO_DIRS) alongside roster_union_sync.DEFAULT_CHILD_REPOS, and the two deliberately differ (the latter omits deploy) — both must track the org CLAUDE.md Repository Map by hand; a single shared canonical constant with documented per-consumer filters would remove the drift surface (follow-up, relates to #744).

Verified against the diff and the file at head 914cf96, not the body. This is the correct root fix and supersedes the scoped #826 approach cleanly.

Checkout-stable snapshot confirmed: _EXPECTED_KINDS now contains only the parent `.`, whose configs are version-controlled in this repo, so the documented snapshot is no longer a function of an arbitrary local child revision. All seven child snapshot entries are gone. test_new_matches_documented_expectations iterates _documented_roots() (parent only) and is guarded against vacuous pass by the assertIn(".", seen) line — so a parent-less checkout would fail loudly rather than silently skip, which is the right failure direction.

Parity invariant genuinely covers present children regardless of staleness: test_new_equals_old_on_present_configs iterates _present_config_roots() = parent plus every present child, and asserts NEW == OLD classifier on the same bytes. Because it compares two classifiers over identical input rather than against a hardcoded expectation, it holds no matter how stale a local child checkout is — exactly the property the original #816 symptom needed. The OLD-scanner concatenation reproduction is preserved so the comparison stays fair.

Absent-child skip cannot false-fail the parent-less CI checkout: both generators yield only when pc.is_file(), so absent siblings are silently skipped; the parent is always present and is the only asserted documented snapshot. The CI checkout (no siblings) therefore exercises parent-only on both tests with no false-fail path.

_CHILD_REPO_DIRS includes all seven children including noorinalabs-deploy. I confirmed roster_union_sync.DEFAULT_CHILD_REPOS has only six and omits deploy, so the deliberate non-reuse is correct — reusing it would have silently dropped deploy from the parity scan. Good call, and the inline comment documenting the intent and the keep-in-step-with-the-Repository-Map note are the right mitigation; the standing duplication is the carry-forward noted above.

All checks green at head (Pytest, the Pre-commit/CI sync-drift gate, and roster sync-drift all SUCCESS; the Memory-budget rollup-status shows IN_PROGRESS but its conclusion is already SUCCESS — GitHub rollup display lag, not a pending failure). Approving.

PR #840, opened 2026-06-23
session-start: make Step 0 + 5a loops statically analyzable (#839)

### LINKED ISSUE #839: session-start: Step 0 + 5a use process-substitution/here-strings the 2.1.x permission engine can't analyze → prompts every session
## Problem

After the Claude Code **2.1.187** upgrade, `/session-start` prompts for permission on two of its bash blocks even though the underlying commands are trusted/allowlisted. Root cause: the 2.1.x permission engine refuses to statically analyze **process substitution** (`done < <(...)`) and **here-strings** (`done <<< "$VAR"`), reporting *"Contains shell syntax (string) that cannot be statically analyzed"* — so allow-rules can never match those blocks and they prompt every session.

Affected blocks in `.claude/skills/session-start/SKILL.md`:

- **Step 0 (worktree cleanup)** — `done < <(git -C "$repo" worktree list --porcelain; echo)`
- **Step 5a (red default-branch run detection)** — `done < <(gh api ...)` plus `done <<< "$REPOS"`

The other blocks were a separate, simpler issue (sub-commands like `jq`/`tail` not allowlisted) — fixed by extending the machine-local allowlist.

## Fix

Rewrite the two blocks to feed their loops from a temp file (`done < "$tmpfile"`) instead of process substitution / here-strings. A file redirect is statically analyzable **and** keeps the loop in the current shell, so the `FLAGGED` / `RED` array accumulation those blocks depend on still survives past `done` (a naive `| while` pipe would silently drop it to a subshell — which is exactly why the originals used process substitution).

Verified: arrays + functions are analyzable on their own; only process-substitution and here-strings trip the analyzer. Temp-file loop tested prompt-free with array surviving past `done`.

## Scope

Skill-internal refactor only — no behavior change, no ontology entity change. Requires `mktemp` in the machine-local allowlist.

parametrization:
## Summary

After the Claude Code **2.1.187** upgrade, `/session-start` prompts for permission on two bash blocks even when the underlying commands are trusted/allowlisted. The 2.1.x permission engine refuses to statically analyze **process substitution** (`done < <(...)`) and **here-strings** (`done <<< "$VAR"`) — it reports *"shell syntax that cannot be statically analyzed"* — so allow-rules can never match those blocks.

This rewrites the two offending blocks in `.claude/skills/session-start/SKILL.md` to feed their loops from a **temp file** (`done < "$tmpfile"`):

- **Step 0 (worktree cleanup)** — was `done < <(git … worktree list --porcelain; echo)`
- **Step 5a (red default-branch run detection)** — was `done < <(gh api …)` + `done <<< "$REPOS"`

A file redirect is statically analyzable **and** keeps the loop in the current shell, so the `FLAGGED` / `RED` array accumulation those blocks depend on still survives past `done` (a `| while` pipe would silently drop it to a subshell — which is exactly why the originals used process substitution).

## Verification

- Probed empirically: bash **arrays + functions are analyzable** on their own; only process-substitution and here-strings trip the analyzer.
- Temp-file loop tested in isolation — runs **prompt-free** and the array survives past `done` (length preserved).
- `grep` confirms no `< <(` / `<<<` code constructs remain (only in explanatory comments).
- Pre-commit + pre-push hooks green.

## Scope / risk

Skill-internal refactor — **no behavior change, no ontology entity change**. The companion allowlist additions (`jq`/`tail`/`mktemp`/…) live in machine-local `settings.local.json` (gitignored), not this PR.

Closes #839

### DIFF
--- a/.claude/skills/session-start/SKILL.md
+++ b/.claude/skills/session-start/SKILL.md
@@ -89,6 +89,11 @@ for repo in "${REPOS[@]}"; do
 
   # Walk worktrees in porcelain form. Records are blank-line separated;
   # fields we care about: worktree <path>, HEAD <sha>, locked [<reason>].
+  # Capture porcelain output to a temp file (with a trailing blank line so the
+  # last record is flushed) and feed the loop from it — see the note at `done`
+  # below for why a temp file rather than `< <(...)` process substitution.
+  _wtfile="$(mktemp)"
+  { git -C "$repo" worktree list --porcelain; echo; } > "$_wtfile"
   wt="" head="" locked=0
   while IFS= read -r line; do
     case "$line" in
@@ -110,7 +115,16 @@ for repo in "${REPOS[@]}"; do
         fi
         wt="" ;;
     esac
-  done < <(git -C "$repo" worktree list --porcelain; echo)
+    # NB: the loop is fed from a temp FILE (not a `< <(...)` process
+    # substitution) so the whole Step-0 block stays statically analyzable by
+    # the Claude Code permission engine — process substitution trips the
+    # "shell syntax that cannot be statically analyzed" path and forces a
+    # prompt regardless of the allowlist (main, 2026-06-23). A file redirect
+    # is analyzable AND keeps the loop in the current shell, so the FLAGGED
+    # array accumulation below survives past `done` (a `| while` pipe would
+    # run the body in a subshell and silently drop it).
+  done < "$_wtfile"
+  rm -f "$_wtfile"
 done
 
 echo "--- remaining worktrees (parent + children) ---"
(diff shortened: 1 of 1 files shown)


parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved

**Review: LGTM**
MustFix: None
TechDebt: None

Independent TPM/process review of the single-file change to the session-start contract. CI is fully green (19/19 checks SUCCESS, branch mergeable).

The rewrite is purely mechanical and preserves behavior exactly. Step 0 worktree cleanup keeps every guard untouched — the skip-main-checkout short-circuit, the locked-worktree flag, the merged-ancestor decision via merge-base is-ancestor against origin main, and the UNMERGED fallback — and the FLAGGED array still accumulates in the current shell because the loop is fed by a plain file redirect rather than a pipe. Step 5a red-run detection likewise keeps the same publish/deploy/release name filter, the latest-run-per-workflow group-by, the non-success conclusion set, and the RED array semantics; the only difference is the data now lands in a temp file before the loop reads it.

Temp-file hygiene checks out on every path with no leak. Step 0 creates one file per repo and removes it after the inner read loop closes; the two continue statements act on the inner while loop, not the outer for, so the cleanup line always runs and there is no early break or return that could skip it. Step 5a creates the per-repo runs file only after the empty-repo continue guard (so a blank line never mktemps an orphan), removes it at the end of each iteration, and removes the outer repo-list file after the outer loop. The trailing blank-line echo on the porcelain capture is correctly retained so the final record flushes through the end-of-record case.

The inline comments are accurate and genuinely load-bearing for a file that IS the session-start contract: they correctly attribute the prompt-on-every-session regression to the permission engine's inability to statically analyze process substitution and here-strings, and they correctly explain the dual constraint that a file redirect satisfies — analyzable to the engine while still keeping the loop in the current shell so the arrays survive (a pipe would drop them to a subshell). Nothing else in the skill was disturbed. No regression risk to documented Step 0/5a behavior. Clean to merge.

parametrization:
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved

**Review: LGTM**
MustFix: None
TechDebt: None

Independent charter review of the Step 0 + Step 5a rewrite. Read the diff and both full blocks at head sha 2b3a068; all 19 CI checks are green (markdown, ruff lint/format, mypy, pytest, cspell, lychee, actionlint, sync-drift, commit-identity, etc.).

Current-shell semantics are preserved, which is the whole point of choosing temp files over a pipe. Both loops are fed via a plain file redirect into the temp file, not a pipe, so the loop bodies run in the current shell and the accumulator arrays survive past the closing keyword — the worktree-flag array in Step 0 and the red-run array in Step 5a both still populate. A pipe-into-while would have dropped them to a subshell; the rewrite correctly avoids that.

mktemp usage and cleanup are sound on every path. In Step 0 the per-repo temp file is created at the top of the for-repo iteration and removed at the bottom; the only continue statements live inside the inner while-read loop, so they never skip the cleanup line. In Step 5a the repo-list temp file is created once and removed after the outer loop, the per-repo runs temp file is created after the empty-line guard (so the guard's continue cannot leak it) and removed at the end of each iteration, and there is no early exit between any create and its matching removal.

No behavioral change versus the originals. Step 0 still appends a trailing blank line inside the brace group capturing the porcelain output, so the final record is still flushed by the empty-line case the same way the original trailing echo did. Step 5a's seeding of the repo-list file with a single newline-terminated print is equivalent to the here-string it replaces (both yield one trailing newline), the empty-line guard still skips blank entries, and the same workflow-run records are walked.

No new portability concern under zsh: the change removes the process-substitution and here-string forms entirely in favor of mktemp plus file redirects and a single-format print, all of which are portable and analyzable. The explanatory comments accurately capture why the file-redirect form was chosen over both process substitution and a pipe. Clean, well-documented, and CI-green. Approving.

Issue #713, opened 2026-06-19
process(tooling): auto-sync local main after any push/merge to origin/main — deterministic .claude/lib helper + PostToolUse hook


parametrization:
## Motivation

Owner 2026-06-19: after any push/merge to `origin/main` (by this session or another actor), the local `main` checkout should pick up the change **immediately**, not lag. This session opened 22 commits behind `origin/main`, which silently ran pre-fix hooks (#708/#709 fix existed in origin but not on disk) and produced false orientation (phase/wave-unknown, partial repo scans). A stale local main is a recurring, high-blast-radius failure mode.

## Proposed solution

A deterministic Python helper plus a hook trigger (enforcement-hierarchy: prefer hook over manual habit).

1. **`.claude/lib/sync_main.py`** — fetch `origin`, and if local `main` is strictly behind `origin/main` and the checkout is on `main` with no real local changes (generated/gitignored-only is OK), fast-forward (`--ff-only`). Never force, never touch a diverged/ahead/dirty tree — report and exit non-zero with a clear diagnostic instead. Pure, testable, with a unit test under `.claude/lib/tests/`.

2. **PostToolUse hook trigger** — after a Bash tool call that performs a merge/push to the default branch (`gh pr merge ...`, `git push origin main`, wave-branch->main merge), invoke `sync_main.py` so *our own* merges reflect on disk immediately. Wire via the existing `post_dispatcher.py`.

3. **Session-start ff-pull** — extend Step 0 (currently `git fetch origin main` only) to a guarded `sync_main.py` ff-pull so *others'* pushes are picked up at session boundaries. (True "immediately for anyone" needs polling; session-start + post-merge hook covers the practical cases without a daemon.)

## Acceptance Criteria

- [ ] `.claude/lib/sync_main.py` fast-forwards local main when safe; refuses + diagnoses when diverged/dirty/ahead
- [ ] Unit tests cover: behind+clean (ff), diverged (refuse), dirty real changes (refuse), already-current (no-op)
- [ ] PostToolUse hook auto-syncs after a successful merge/push to origin/main this session
- [ ] session-start Step 0 ff-pulls via the helper (guarded, non-fatal on failure)
- [ ] Mirrored in pre-commit/CI parity where applicable

PR #846, opened 2026-06-24
feat(env): refresh + staleness guard for parent's child-repo checkouts (#832)

### LINKED ISSUE #832: tech-debt(env): parent's local child-repo checkouts are stale (phases behind origin) — refresh + guard
## Problem
The parent `noorinalabs-main` checkout's embedded child-repo clones are badly stale vs their origins — surfaced repeatedly during p6-wave-16:
- `noorinalabs-user-service` on a **phase-3** commit (`05d644e`)
- `noorinalabs-isnad-graph` **~207 commits** behind origin/main
- `noorinalabs-design-system`, `noorinalabs-landing-page` also lagging; several sit on old feature branches, not main

This is the root cause of #816 (the parent pre-push test reads the stale on-disk child configs) and makes any agent that reads a child clone directly (instead of a fresh worktree off origin/main) draw wrong conclusions.

## Fix
- Refresh each child checkout to origin/main (mindful that several are parked on old feature branches — don't clobber genuine local work; verify each before resetting).
- Consider a session-start guard that warns when an embedded child clone is N commits behind its origin (sibling to the sync_main stale-base warning #822).

## Provenance
Flagged by the #744/#718 child fleet (user-service, isnad-graph, design-system, landing-page agents) + #816 root-cause analysis. p6-wave-16.

parametrization:
## What

Adds a **refresh + staleness guard** for the parent's embedded child-repo checkouts — the child-side sibling of `sync_main` (#713). Closes #832.

The parent `noorinalabs-main` checkout `.gitignore`s its child repos, which live as independent clones under `$REPO_ROOT/noorinalabs-<name>`. They drift badly: during p6-wave-16 `user-service` was parked on a phase-3 commit and `isnad-graph` sat ~207 commits behind `origin/main` — the root cause of #816 (the parent pre-push test reads the stale on-disk child configs) and of any agent that reads a child clone directly drawing wrong conclusions.

## How

New `.claude/lib/check_child_checkouts.py`, same safety stance as `sync_main`, **no force path**:

- A child that is **clean and on main** is fast-forwarded to `origin/main` under `--refresh` (the ff is delegated to `sync_main`: `--ff-only` + generated-file stash).
- A child that is **dirty, diverged, or parked on an old feature branch** is **FLAGGED** for a manual decision and left exactly as-is. Never force-discards local work.
- `behind`/`ahead` are measured `HEAD..origin/main`, so a child parked on an old feature branch still reports how far behind the mainline it really is — the staleness signal.

Wired into `/session-start` **Step 0** (`--refresh`) right after the `sync_main` parent fast-forward, with a reporting line + a new `0b. Child checkouts` row in the output table.

## Tests

`.claude/lib/tests/test_check_child_checkouts.py` — drives **real git** against throwaway temp repos. Covers every classification (`absent` / `current` / `stale-on-main` / `fast-forwarded` / `stale-feature-branch` / `dirty` / `diverged` / `ahead`), the no-force guarantees (feature branch + dirty + diverged are left untouched), `render()` FLAGGED block, and the CLI exit-0-on-flagged contract.

Local⇄CI parity green: ruff-format/lint, mypy, full lib pytest (418 passed), cspell, actionlint, and the pre-commit⇄CI sync-drift gate all pass.

## Reviewers
- Primary: **Lucas Ferreira**
- Secondary: **Aisha Idrissi**

Part of #832.

### DIFF
--- a/.claude/lib/check_child_checkouts.py
+++ b/.claude/lib/check_child_checkouts.py
@@ -1,0 +1,264 @@
+#!/usr/bin/env python3
+"""Refresh + staleness guard for the parent's embedded child-repo checkouts.
+
+The parent ``noorinalabs-main`` checkout ``.gitignore``s the child repos, which
+live as independent clones under ``$REPO_ROOT/noorinalabs-<name>``. They drift:
+during p6-wave-16 ``noorinalabs-user-service`` was parked on a *phase-3* commit
+and ``noorinalabs-isnad-graph`` sat ~207 commits behind ``origin/main`` (main#832).
+A stale child clone is the root cause of #816 (the parent pre-push test reads the
+stale on-disk child configs) and makes any agent that reads a child clone directly
+draw wrong conclusions.
+
+This is the sibling to :mod:`sync_main` (the parent ``main`` fast-forward, #713)
+for the *children*. It applies the **same safety stance**:
+
+  * It fast-forwards a child to ``origin/main`` ONLY when that child is currently
+    on ``main``, is strictly behind (``behind > 0`` and ``ahead == 0``), and has a
+    clean working tree (the ff is delegated to :func:`sync_main.sync_main`, which
+    stashes the machine-generated allowlist around the merge).
+  * It NEVER force-discards local work: a child parked on a feature branch, a
+    dirty tree, or a divergence is **FLAGGED for a manual decision** and left
+    exactly as it is. There is no force path.
+
+``behind``/``ahead`` are always measured against the child's own ``origin/main``
+from ``HEAD`` (not from ``main``), so a child parked on an old feature branch
+still reports how far behind the mainline it really is — that is the staleness
+signal the guard exists to surface.
+
+CLI: ``python3 .claude/lib/check_child_checkouts.py [REPO_ROOT] [--refresh]``
+
+  * default (guard/check): reports per-child status, mutates nothing.
+  * ``--refresh``: additionally fast-forwards the clean-on-main-behind children.
+
+Either way it prints a per-child report plus a FLAGGED block for the children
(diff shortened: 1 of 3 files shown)


parametrization:
Requestor: Lucas Ferreira
Requestee: Bereket Tadesse
RequestOrReplied: Approved

**Review: LGTM**
MustFix: None
TechDebt: None

Reviewed the diff at HEAD. The guard correctly mirrors the sync_main safety stance — it fast-forwards only a clean, on-main, strictly-behind child and delegates the actual ff to sync_main for the --ff-only + generated-file stash; every dirty / diverged / ahead / feature-branch case is flagged and left untouched with no force path. Measuring behind/ahead as HEAD..origin/main is the right call so a feature-branch child still reports true mainline distance. The CLI exit-0-on-flagged contract is important and tested, so the session-start wiring can't break orientation on a "can't refresh now" state. 12 real-git tests cover every classification plus the no-force guarantees. The CHILD_REPOS list duplicates the session-start repo set, but it's commented as intentionally kept-in-sync — acceptable, not worth gating.

parametrization:
Requestor: Aisha Idrissi
Requestee: Bereket Tadesse
RequestOrReplied: Approved

**Review: LGTM**
MustFix: None
TechDebt: None

Independent pass, SRE/operational lens. This directly closes the #832 root cause that fed #816 — stale on-disk child clones. The reuse of sync_main's git primitives means there's one implementation of "what is dirty" and "how to fast-forward" rather than a divergent copy, which is the right structural choice. No-force is enforced structurally (there is simply no discard path), and the fetch-failure and plumbing-error cases degrade to a flagged/error status rather than an exception. Session-start wiring adds a dedicated reporting row without changing existing Step 0 behavior. 19/19 CI green. No blocking concerns.

Issue #835, opened 2026-06-23
annunaki monitor over-captures rc=0 stdout-pattern output (85% of P6W16 captures were false-positive)


parametrization:
## Problem

The Annunaki monitor over-captures benign output as "errors". In the P6W16 retro, 40 of 47 genuine-stream captures (85%) were `exit_code == 0` records that pattern-matched `stdout:^FAILED`:

- pytest `FAILED` lines surfacing through `… 2>&1 | tail -N` (the pipe returns tail's rc=0, masking nothing actually-failing OR masking a real inner failure — the [[feedback_push_pipe_masks_rejection]] class), or
- benign demo/probe output that contains the literal string.

These rc=0 records are logged with an **empty `hook` field** (they show as `unknown` in the breakdown), so they dominate the count and crowd out the genuine signals (the 5 `pretooluse_block` + 2 `posttooluse_event` records, which are hooks working correctly).

## Proposed fix

1. **Gate stdout-pattern matches on nonzero rc** — when `exit_code == 0`, either suppress the stdout-pattern capture or relabel it as `pipe-mask-suspect` rather than a generic error, so `/annunaki` counts and `/annunaki-attack` triage reflect real failures.
2. **Populate the `hook`/`pattern` field** on every record so `unknown` stops dominating the breakdown.
3. Optionally surface the pipe-mask class explicitly (a `git push|gh pr merge … | tail/head` rc=0 with `FAILED`/`rejected` in stdout is a strong pipe-mask signal worth flagging, not hiding).

## Evidence
P6W16 `/annunaki-attack`: 47 captures → 40 rc=0 false-positives, 0 unresolved failures. Full triage in `.claude/team/feedback_log.md` § Retrospective: Phase 6 Wave 16.

Surfaced by `/wave-retro 6 16` Steps 7.6 (annunaki-attack) + 7.7 (memory audit).

parametrization:
Requestor: Fatima Okonkwo
Requestee: Nurul Hakim
RequestOrReplied: Request

**Wave 17 Kickoff — Phase 6**

This issue is assigned to you for p6-wave-17.
- Peer reviewer: Nino Kavtaradze
- Secondary reviewer: Aino Virtanen
- Branch from: `deployments/phase-6/wave-17`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: tech-debt (per charter § Wave Planning & Priority)

Please begin implementation.

[...]

Issue #729, opened 2026-06-20
tech-debt(annunaki): exit-0 false-positive precision pass — keep genuine exit-0-failure carve-out


parametrization:
## Summary
P5W5 retro proposed change #2 (Claude Efficiency framework debt). 85% of the P5W5 annunaki "errors" (40/47) were exit-0 false positives — benign command output containing trigger words ("error"/"fail"). This drowns the real signal (the 5 correct pretooluse_blocks). Improve annunaki_monitor precision without losing the genuine exit-0-failure carve-out.

## Acceptance Criteria
- [ ] Tag exit-0 records whose trigger match is in echoed output (not a real failure) as a distinct low-confidence sub-class
- [ ] Exclude the low-confidence sub-class from the /annunaki genuine-error count, but retain it for forensics
- [ ] PRESERVE the real exit-0-failure cases (e.g. `git push ... | tail` masking a REJECTED push — feedback_push_pipe_masks_rejection); do NOT blanket-ignore exit 0
- [ ] Tests cover both a benign exit-0-with-trigger-word and a genuine exit-0 failure

## Origin
P5W5 retro pain point #2 / proposed change #2 (2026-06-20). See feedback_log.md.

PR #850, opened 2026-06-24
fix(annunaki): rc=0 stdout-pattern precision + hook attribution (#835)

parametrization:
## What

Tightens the Annunaki monitor's rc=0 handling and fixes hook attribution, per the P6W16 retro finding that **85% of captures (40/47) were `exit_code==0` records that pattern-matched `stdout:^FAILED`** — a pytest `FAILED` surfacing through `… | tail` rc-masking, or benign demo output containing the literal — logged with an empty `hook` field so they showed as `unknown` and crowded out the genuine signals.

## Changes

- **`_classify_confidence` now returns `(confidence, category)`** (`.claude/hooks/annunaki_monitor.py`). The residual exit-0 / stdout-only / no-STRONG-signal / not-echoed class is demoted to `confidence="low"` + `category="pipe-mask-suspect"`: **retained for forensics and specifically triageable, but excluded from the genuine-error count** (via the existing #729 low-confidence reader filter). This supersedes #729's recall-preserving "unrecognized exit-0 stdout match defaults high" policy for exactly the class the W16 data showed is 85% noise.
- **Genuine exit-0-failure carve-out preserved**: a STRONG masked-failure signal (`[rejected]`/`failed to push`, `exit status N`, `Traceback`, `(exit N)`) stays `high`/counted, now labeled `category="masked-failure"` — flagged, not hidden (issue point 3).
- **`hook` field populated** on every monitor capture (`hook="annunaki_monitor"`) so `/annunaki`'s `rec.get("hook", "unknown")` breakdown stops bucketing command-failure records as `unknown`.
- **`annunaki_parse`**: new `is_pipe_mask_suspect()` triage helper; the existing low-confidence filter already excludes the new class from `count_errors`.

## Tests

- rc=0 false-positive case: `pytest … | tail` exit-0 `FAILED` → `pipe-mask-suspect`, logged but not counted; benign demo output likewise.
- genuine nonzero-rc capture: stays `high`/`nonzero-exit`, counted (precision pass does not weaken genuine capture).
- STRONG masked-failure rc=0 (`git push | tail` rejected) → still `high`/`masked-failure`.
- `hook` field population on both genuine and suspect records.
- reader-side: pipe-mask-suspect excluded from `count_errors`, retained with `include_low_confidence=True`.

Full hooks + lib suites green (1882 passed); `pre-commit run --all-files` green on both commit and pre-push stages.

Re #835

Reviewers: Nino Kavtaradze (primary), Aino Virtanen (secondary)

### DIFF
--- a/.claude/hooks/annunaki_monitor.py
+++ b/.claude/hooks/annunaki_monitor.py
@@ -32,6 +32,22 @@ Input Language:
                      genuine-error count by annunaki_parse, but retained for
                      forensics; a STRONG masked-failure signal (e.g. a `git
                      push | tail` REJECTED push) stays "high" and counted.
+  rc=0 precision (#835): every logged record also carries a `hook` field
+                     (always "annunaki_monitor" for command-failure captures,
+                     so the /annunaki breakdown stops bucketing them as
+                     "unknown") and a `category` triage label. The P6W16 retro
+                     found 85% of captures were exit-0 stdout-pattern matches
+                     (a pytest `FAILED` line surfacing through `… | tail`
+                     rc-masking, or benign demo output containing the literal)
+                     that #729's recall-preserving default tagged "high" and
+                     counted. #835 demotes that residual class — exit-0,
+                     stdout-only, NOT a STRONG masked-failure signal, NOT
+                     positively-recognized echoed content — to
+                     `confidence="low"` + `category="pipe-mask-suspect"`:
+                     retained for forensics and specifically triageable, but
+                     excluded from the count. The genuine exit-0-failure
+                     carve-out (STRONG masked-failure) stays "high"/counted as
+                     `category="masked-failure"`.
 
 Exit codes:
   0 — always (advisory hook, never blocks)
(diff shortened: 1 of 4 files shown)


parametrization:
Requestor: Nino Kavtaradze
Requestee: Nurul Hakim
RequestOrReplied: Approved

**Review: LGTM**
MustFix: None
TechDebt: None

Reviewed the classifier change at HEAD. The single behavioral change is sound and well-scoped: the unrecognized exit-0 stdout match, previously defaulted to counted-high, is now demoted to low/pipe-mask-suspect — retained for forensics but excluded from the count. Recall is preserved for every recognizable failure: a non-zero exit, a stderr-pattern match, and the STRONG masked-failure carve-out (rejected push / exit-status-N / traceback) all still classify high and count. That directly targets the 85%-false-positive class the W16 retro measured without dropping genuine masked failures. The hook-field attribution fixes the /annunaki "unknown" bucketing, and the orthogonal category labels make triage cheaper. The one test asserting the old default was correctly updated rather than worked around. Supersession of #729's policy is intentional and documented.

parametrization:
Requestor: Aino Virtanen
Requestee: Nurul Hakim
RequestOrReplied: Approved

**Review: LGTM**
MustFix: None
TechDebt: None

Standards lens. This is a precision improvement that explicitly preserves recall for recognizable signals while cutting the dominant noise class — the right tradeoff direction, and honestly flagged as a deliberate supersession of #729 rather than a silent policy drift. Counting stays driven by confidence while category describes kind, a clean separation that lets /annunaki-attack batch-triage. No new check kind introduced, so local-CI parity is untouched. CI green. No blocking concerns.

PR #851, opened 2026-06-24
hook(#838): flag rc-masking pipes on git push / gh pr merge (PostToolUse)

### LINKED ISSUE #838: hook: flag rc-masking pipes on git push / gh pr merge (pipe-mask PostToolUse)
## Problem
Owner-ratified P6W16 retro process change #3 (DECIDE-tier — owner approved direction).

`git push … | tail` / `gh pr merge … | head` return the pipe tail's rc (0), masking a REJECTED push or a failed merge (the [[feedback_push_pipe_masks_rejection]] class, ig#1044). P6W16 annunaki data is direct evidence the class recurs — 40 rc=0 captures were `stdout:^FAILED` surfacing through `… | tail`.

## Proposed change
A PostToolUse (Bash) hook that detects `git push|gh pr merge … | tail/head` (rc-masking) and warns / records, so a masked rejection surfaces instead of reading as success. Pairs with #835 (annunaki monitor rc=0 tuning).

## Provenance
`/wave-retro 6 16` Step 7.7 (memory-to-automation audit); `.claude/team/feedback_log.md` § Retrospective: Phase 6 Wave 16.

parametrization:
## What

New **PostToolUse Bash hook** `warn_pipe_mask_rc.py` that flags rc-masking pipes on `git push` / `gh pr merge` — the [[feedback_push_pipe_masks_rejection]] class (ig#1044). `git push … | tail` returns the pipeline's exit code (the downstream stage's rc, absent `pipefail`), so a REJECTED push or failed merge exits 0 and reads as success. P6W16 annunaki data confirmed recurrence: 40 rc=0 captures were `stdout:^FAILED` surfacing through `… | tail`.

Owner-ratified P6W16 retro process change #3. Closes #838.

## How

- **Detection** — flags a `git push` / `gh pr merge` that is a **non-final pipeline stage** (rc swallowed by *any* downstream command — `tail`/`head`/`cat`/`tee`/`grep`/…; the footgun is the position, not a specific masker). Uses the shared **bashlex AST** parser with a **shlex/regex fallback**, so it is command-position aware — the phrase inside a heredoc body, a quoted string, or a `--body` value does NOT match.
- **Block vs. warn (deliberate)** — PostToolUse cannot block (the command already ran; `post_dispatcher.py` is advisory-only). The genuine footgun is the masked *outcome*, and PostToolUse is the only phase with the actual rc + output to detect it, so it surfaces two tiers:
  - **TIER 1 — confirmed masked failure**: masking shape + a real push/merge failure signal in the output (or non-zero pipeline rc) → a HARD, prominent diagnostic (per [[feedback_safety_direction_over_ux_friction]]: when a hook can't auto-fix, surface a hard diagnostic, never silently allow).
  - **TIER 2 — footgun shape only**: masking shape but clean output → a lighter nudge to drop the pipe / use `; echo rc=$?` / `set -o pipefail`.
  - A PreToolUse hard-block was deliberately NOT added — the issue scopes this to PostToolUse, and pre-blocking every `git push … |` would risk blocking the other wave-17 agents on benign pipes.

## False-positive guards (keeps the wave quiet)

These all stay **silent** — verified by tests: no-pipe `git push origin <branch>` (THE common wave case), the `; echo rc=$?` fix, `&&` chains, push/merge as the pipeline's **last** stage, `set -o pipefail` guarded pipelines, non-push pipes (`git log | head`), and the phrase in heredoc / quoted / data position.

## Tests & gates

- 31 new tests (`test_warn_pipe_mask_rc.py`) covering both the AST and fallback paths, both severity tiers, every false-positive guard, and the standalone `main()`.
- Full hooks suite green (1496 passed); ruff lint+format, mypy clean.
- **No new CI/pre-commit check KIND**: the hook is covered by the existing `.claude/hooks/**` ruff/mypy globs and the `.claude/hooks/tests/` pytest job. `pre_commit_ci_sync.py` gate green — local⇄CI parity intact.
- Registered in `post_dispatcher.py` (Bash matcher, local tier).

## Reviewers
- Primary: **Aino Virtanen**
- Secondary: **Lucas Ferreira**

### DIFF
--- a/.claude/hooks/post_dispatcher.py
+++ b/.claude/hooks/post_dispatcher.py
@@ -77,6 +77,7 @@ _EDIT_WRITE_MODULES = [
 _REGISTRY: dict[str, list[str]] = {
     "Bash": [
         "annunaki_monitor",  # local: regex over stdout/stderr, JSONL append
+        "warn_pipe_mask_rc",  # local: flag rc-masking `git push|tail`/`gh pr merge|head` (#838)
         "auto_sync_main",  # local: ff local main after a push/merge to origin/main (#713)
         "auto_add_issue_to_board",  # network: runs `gh project item-add`
         "post_wave_kickoff_comment",  # network: gh fetch + post comment
--- a/.claude/hooks/tests/test_warn_pipe_mask_rc.py
+++ b/.claude/hooks/tests/test_warn_pipe_mask_rc.py
@@ -1,0 +1,217 @@
+#!/usr/bin/env python3
+"""Tests for `warn_pipe_mask_rc.check()` — PostToolUse rc-masking pipe flag (#838).
+
+Covers the two acceptance axes:
+  - FALSE-POSITIVE GUARDS: a no-pipe push, the `; echo rc=$?` fix, `&&`, a
+    push as the pipeline's LAST stage, pipefail-guarded pipes, non-push pipes,
+    and the phrase in heredoc/quoted/data position must ALL stay silent. The
+    no-pipe `git push origin <branch>` case is load-bearing: it is what every
+    wave-17 agent runs, and a false positive there would noise the whole wave.
+  - DETECTION + TIERING: a push/merge in a non-final pipeline stage is flagged,
+    escalating TIER 2 (footgun shape only) → TIER 1 (confirmed masked failure)
+    when the captured output carries a failure signal or a non-zero rc.
+
+Both the bashlex AST path and the shlex/regex fallback (bashlex monkeypatched
+off) are exercised.
+
+Run from the repo root:
+    python3 -m pytest .claude/hooks/tests/test_warn_pipe_mask_rc.py -v
+"""
+
+from __future__ import annotations
+
+import json
+import subprocess
+import sys
+import unittest
+from pathlib import Path
+
+_HERE = Path(__file__).resolve().parent
+_HOOKS_DIR = _HERE.parent
+sys.path.insert(0, str(_HOOKS_DIR))
+
+import warn_pipe_mask_rc as wpm  # noqa: E402
+
+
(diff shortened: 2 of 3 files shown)


parametrization:
Requestor: Aino Virtanen
Requestee: Nino Kavtaradze
RequestOrReplied: Approved

**Review: LGTM**
MustFix: None
TechDebt: None

Reviewed the hook + tests at HEAD. The detection is command-position-aware via the shared bashlex AST with a shlex/regex fallback, so a push/merge only fires when it sits in a non-final pipeline stage whose rc is actually swallowed — exactly the ig#1044 class. The false-positive surface is the real risk for a wave-wide hook, and it is covered comprehensively: plain push with no pipe, the echo-rc-fix idiom, && chains, push as the last stage, pipefail-guarded pipes, non-push pipes, and quoted/heredoc data positions all stay silent. The two-tier output (loud when a real failure signal or non-zero rc is present, soft nudge for footgun-shape-only) is the right call given PostToolUse can't block, and skipping a PreToolUse block was a deliberate, well-reasoned choice to avoid false-blocking the other wave agents. 31 tests across both parse paths. No new check kind, parity intact.

parametrization:
Requestor: Lucas Ferreira
Requestee: Nino Kavtaradze
RequestOrReplied: Approved

**Review: LGTM**
MustFix: None
TechDebt: None

Independent pass. This makes the recurring rc-masking footgun visible at the one phase that actually has the rc + output to judge it, which is the pragmatic place for it given the post-phase can't block. The advisory-only stance plus the silent-on-plain-push guarantee means it adds signal without adding friction to the normal `git push origin <branch>` path every agent uses. AST-with-fallback parsing avoids the brittle pure-regex trap, and the dedup keeps it from double-flagging. Tests exercise both the AST and fallback branches and the standalone main(). CI green. No concerns.

Issue #857, opened 2026-06-24
feat(P6W17): #820 Task-3c — Hook 15 → advisory + retire checksums.json/ontology-rebuild structural layer


parametrization:
**Parent:** #820 · **Wave:** P6W17 · **Task 3c**

With the structural layer now **regenerable** (owned generator, #855) rather than hand-resolved, the change-tracker/resolver machinery for the *structural* layer is obsolete. Execute the two governance changes the C×T2 decision calls for.

## Deliverables
1. **Retire `checksums.json` + `/ontology-rebuild` for the structural layer.** The PostToolUse change-tracker hook and the resolver skill exist to keep hand-maintained ontology files in sync on every edit. For the **structural** layer this is replaced by regeneration. Scope the retirement carefully:
   - The **structural** index is now generated (committed `ontology/structural/`), so it no longer needs checksum dirty-tracking or `/ontology-rebuild` resolution.
   - The **semantic overlay** (hand-curated `domain.yaml`/`services.yaml`/`conventions.md`/`*.md`) MAY still use change-tracking — decide and document whether the tracker stays for the overlay only or is retired wholesale. Recommend: keep tracking for the overlay, drop it for the structural layer.
2. **Soften Hook 15 (`enforce_librarian_consulted`) from blocking → advisory.** Rationale: structural context is now always-current via the generated index, so a hard pre-Edit block on librarian consultation is heavier than warranted. Convert to an advisory `systemMessage` warning (NOT a hard block). Update the hook, its tests, and the charter `hooks.md` § Hook 15 entry + CLAUDE.md references.

## Caveats
- Hook 15 currently has NO in-band override; softening it must not silently break the existing tests — update them.
- Per `feedback_safety_direction_over_ux_friction`: when softening a guard, confirm the safety it provided is genuinely now redundant (it is — structural context is regenerated, not stale). Document why.
- This is largely independent of #855's output format; can proceed in parallel.

## Out of scope
Generator (#855), aggregator/overlay (#3b), isnad-graph wiring (#3d).

parametrization:
Resolved by PR #858 (merged to deployments/phase-6/wave-17). Closing at /wave-wrapup — wave-branch merges don't auto-close (feedback_wave_branch_issue_close). Task-3c Hook15 advisory + checksums retire.

Issue #855, opened 2026-06-24
feat(P6W17): #820 Task-3a — owned structural ontology generator (Py/TS/Cypher → llms.txt + code-graph)


parametrization:
**Parent:** #820 (ontology C×T2 execution) · **Wave:** P6W17 · **Task 3a** (gating: blocks Task-3b/3d) · **Decision:** build owned generators (PR #854 checkpoint)

Build the **owned structural ontology generator** that realizes the C×T2 structural layer — replacing the hand-curated `checksums.json`/`ontology-rebuild` structural tracking with a regenerable, in-git, per-repo index. Lowest-token-cost path per the #854 eval (target ~24k whole-load, NOT graphify's ~630k).

## Deliverable
`.claude/lib/ontology_gen/` — a CLI invoked as `python3 -m ontology_gen <repo-root> --out ontology/structural/` producing a per-repo structural index. Run it on **noorinalabs-main itself** and commit main's generated index as the reference output.

### Language coverage (pilot)
- **Python** — stdlib `ast` (modules, classes, funcs/methods, imports, calls). No third-party dep.
- **TypeScript/React** — `ts-morph` (or a tree-sitter-ts fallback if the toolchain is lighter); modules, exports, imports, components.
- **Cypher** — a lightweight extractor for `.cypher`/`.cql` (labels, relationships, query files). This is one of the two gaps #843 flagged; the owned build closes it.

## Output contract (BUILD TO THIS — 3b/3d depend on it)
`ontology/structural/` (committed):
- `code-graph.json` — `{"nodes":[{id, kind, path, line, lang}], "edges":[{src, dst, type}]}` where `kind ∈ {file,module,class,func,method}`, `type ∈ {contains,imports,imports_from,calls,inherits,references}`.
- `llms.txt` — token-economical, **scope-loadable** plain-text structural summary (per-module sections an agent can read piecemeal — the load-bearing C×T2 affordance the incumbent wins on).
- Deterministic output (stable ordering) so git diffs are minimal.

## Borrow from graphify (per #854)
- Ship a **git merge-driver** spec for `code-graph.json` (union-merge) to defuse the parallel-shared-file conflict hazard (`feedback_parallel_panels_shared_file`). (Driver *wiring* into `.gitattributes` may live in 3b; at minimum provide the merge script here.)

## Out of scope (this issue)
[...]

Issue #820, opened 2026-06-22
Execute ontology C×T2: distributed per-repo structural index + central semantic overlay (tooling via isolated-branch bake-off)


parametrization:
**Execution tracker for the ontology decision** (owner chose **C × T2**, 2026-06-22).

Spike + decision: #728 · doc `.claude/team/spikes/p6w2-ontology-vs-graphify.md` · `phase-6.md` § Criterion #4.
Deferred out of P6W2 per the P6 spike-and-decide directive (no teardown in-phase). This issue carries the execution.

## Decision
- **Representation: Hybrid (C)** — auto-generate the structural layer; hand-curate semantics/intent (domain, cross-repo service topology, ADR "why") as low-churn markdown; soften Hook 15 from a blocking gate to advisory.
- **Topology: Distributed + system overlay (T2)** — each child repo owns its **generated structural index** (built in its own CI/pre-commit, where its source is visible and language-native); `noorinalabs-main` keeps only the **central semantic overlay** + a thin aggregator for cross-repo queries.

## Task 1 (BLOCKING, first) — per-language derivability re-measurement
The "100% auto-derivable" spike number was Python/`ast` on the parent `.claude/` only. Re-measure on the **actual product repos** per language before committing to a generator:
- TS/React (isnad-graph frontend, landing-page, design-system) via `ts-morph`/`tsc`
- Cypher/Neo4j (isnad-graph), Terraform/HCL (deploy), SQL
- Any repo whose source is **not** cleanly derivable keeps a hand-curated structural stub until it is.

## Task 2 — tooling bake-off on isolated branches (owner directive 2026-06-22)
Do **not** pre-pick a tool. Stand up ≥2 structural + ≥2 overlay candidates on **throwaway isolated branches** (no merge → revertible) and compare on two lenses:

| Lens | What to capture |
|------|-----------------|
| **Agent / Claude consumption** | token cost to load + query · answer quality on real "what-calls-X / what's-in-module-Y / cross-repo-impact" questions · freshness + regeneration cost |
| **Human-as-knowledge-base** | browsability · authoring ergonomics · backlink/navigation quality · is it pleasant to read + maintain |

Candidate matrix:
- **Structural (per child repo):** per-repo `llms.txt` + code-graph · **SCIP/LSIF** (Sourcegraph) · tree-sitter/ctags · **Neo4j-backed code graph** (dogfoods the product DB)
[...]

PR #858, opened 2026-06-24
feat(P6W17): #820 Task-3c — Hook 15 → advisory + retire checksums/rebuild structural layer

parametrization:
## Summary

Implements the two governance changes the **C×T2** decision (#820) calls for, now that the structural ontology layer is regenerable via an owned generator (#855) at `ontology/structural/`.

### 1. Retire change-tracking / `/ontology-rebuild` for the STRUCTURAL layer only
- `ontology_tracker.py` — add `ontology/structural/` to `SKIP_PATTERNS`. The generated structural layer is always-current-by-regeneration, so checksum dirty-tracking it is meaningless churn and the resolver has nothing to resolve there. The hand-curated **semantic overlay** (`domain.yaml`/`services.yaml`/`conventions.md`/`repos/*.yaml`/`*.md`) is **still tracked**. Boundary documented in the module.
- `checksums.json` — SCOPE note in the `description` field.
- `ontology-rebuild/SKILL.md` — new **Scope** section: resolver covers the overlay only, never hand-edits the structural layer (regenerate it instead).
- Tracker tests: structural-skip + overlay-still-tracked.

### 2. Soften Hook 15 (`enforce_librarian_consulted`) from hard block → advisory
- The hook **never denies** an edit. When `/ontology-librarian` was not consulted it emits an advisory `systemMessage` and **always exits 0**. Transcript-scan + cwd-keyed sentinel detection logic unchanged. Removed the block-path annunaki forensics (existed only to debug *block* decisions).
- Test suite rewritten for advisory semantics + a `main()`-exit-code class asserting exit 0 on both paths.

### Why the weakened guard is safe
Per `feedback_safety_direction_over_ux_friction`: soften a guard only once the safety it provided is demonstrably redundant. The hard block existed to guarantee an agent had loaded **current structural context** before editing. That layer is now generated/always-current (#855), so the staleness it defended against no longer exists for it. The hand-curated semantic overlay still benefits from consulting the librarian, so the advisory nudge remains — it reminds without gating.

## Docs updated
`CLAUDE.md` (three-roles table, "Before any code changes" → recommended/advisory, ontology structure block), charter `hooks.md` § Hook 15 (+ Hook 17 cross-refs), `agents.md` spawn-brief references, `artifact-ownership.md` (overlay vs structural rows), `tech-decisions.md` (C×T2 status), memory `project_ontology_system.md` + `MEMORY.md` index.

## Verification
- `enforce_librarian_consulted` + `ontology_tracker` tests: 60 passed.
- Full local⇄CI gate green: ruff-format/ruff-lint, mypy, pytest (2008 passed), pytest-skills, cspell, actionlint, sync-drift gate, doc-freshness, budgets, mermaid-render.

Part of #857
Part of #820


parametrization:
**Requestor:** Nurul Hakim
**Requestee:** Aino Virtanen
**TechDebt:** none

Reviewed the diff at PR head (base `deployments/phase-6/wave-17`), against the code — not the PR body.

Observability angle, since the change strips a control-flow path:
- `enforce_librarian_consulted.py` now `check()`-returns `{systemMessage}` or `None` and `main()` **always exits 0** — correct advisory semantics. The advisory text is still emitted on the not-consulted path, so the signal isn't lost; it's downgraded from a deny to a visible nudge, which is the intended behaviour.
- Removing the block-path annunaki forensics is right, not a regression — that instrumentation existed solely to debug *block* decisions, and there are no block decisions anymore. Keeping it would have logged noise on a path that no longer denies.
- Detection logic (transcript scan Forms A/B/C + cwd sentinel + #429 tolerant body-cwd fallback) is unchanged, and the fail-open-on-OSError stance is preserved (now strictly cheaper since nothing is gated).
- `ontology_tracker.py` skipping `ontology/structural/` is observably correct: a generated artifact rewritten wholesale would otherwise spam `checksums.json` dirty churn on every regen. The semantic overlay stays tracked.

CI 20/20 green; tests rewritten for advisory semantics incl. an exit-0 assertion class. No observability objection. Approving.

parametrization:
**Requestor:** Nino Kavtaradze
**Requestee:** Aino Virtanen
**TechDebt:** none

Second reviewer, on the safety-of-softening-a-guard axis (read the artifact at head).

The bar for weakening an enforcement hook is `feedback_safety_direction_over_ux_friction`: soften only once the safety it provided is *demonstrably* redundant. That bar is met here and documented, not hand-waved:
- The hard block (#150) existed to guarantee **current structural context** before an edit. That layer is now **generated** (`ontology/structural/`, owned generator #855) → always-current-by-regeneration, so the staleness it defended against no longer exists for it. The rationale is written into the hook docstring, charter `hooks.md` § Hook 15, and CLAUDE.md — not silently dropped.
- The residual value (the hand-curated **semantic overlay** still benefits from a consult) is preserved as an advisory nudge, so we keep the useful reminder while dropping the friction. That's the correct direction: reduce gating, retain signal.
- The scope of the tracker retirement is tight — `ontology/structural/` only; `domain.yaml`/`services.yaml`/`conventions.md`/`*.md` remain tracked. No collateral weakening of the semantic-layer guarantees.
- No new in-band override was introduced (the prior "no in-band override" property is moot now that nothing blocks), and the change doesn't open any bypass of *other* hooks.

No security/safety objection. Two approvals stand — clear to merge to the wave branch. Approving.

Issue #856, opened 2026-06-24
feat(P6W17): #820 Task-3b — central semantic overlay + cross-repo aggregator + graph merge-driver


parametrization:
**Parent:** #820 · **Wave:** P6W17 · **Task 3b** · depends on **#855** (generator + output contract)

Build the **central semantic overlay + cross-repo aggregator** for the C×T2 topology (distributed per-repo generated index + central overlay in `main`).

## Deliverable
1. **Aggregator** (`.claude/lib/ontology_gen/aggregate.py` or sibling): reads each in-scope repo's committed `ontology/structural/code-graph.json` and rolls them into a central cross-repo view under `ontology/` in main (borrow graphify's `merge-graphs` design — union by node id, namespaced by repo). Must run offline, no DB/LLM.
2. **Central semantic overlay**: the hand-curated layer stays as markdown/YAML + `[[wikilinks]]` (the existing `ontology/*.md` + `domain.yaml`/`services.yaml`/`conventions.md`). Wire the overlay so it *references* generated structural nodes rather than duplicating them — the overlay is semantics, the generated index is structure.
3. **Git merge-driver wiring**: register the merge-driver from #855 in `.gitattributes` for `ontology/structural/code-graph.json` so parallel branches union-merge instead of conflicting.

## Build to the #855 output contract
Consume `ontology/structural/{code-graph.json,llms.txt}`. Coordinate with the #855 implementer on the exact schema (it is fixed in #855's body — treat it as the interface).

## Out of scope
The generator itself (#855), Hook 15 / checksums retirement (#3c), isnad-graph wiring (#3d).

parametrization:
Resolved by PR #860 (merged to deployments/phase-6/wave-17). Closing at /wave-wrapup — wave-branch merges don't auto-close (feedback_wave_branch_issue_close). Task-3b aggregator+overlay+merge-driver.

PR #860, opened 2026-06-24
feat(P6W17): #856 central cross-repo aggregator + overlay wiring + merge-driver registration

parametrization:
## What

#820 Task-3b — the **central** half of the C×T2 ontology: a cross-repo structural aggregator, semantic-overlay→structural wiring, and git merge-driver registration. Builds against the #855 output contract (unchanged).

### 1. Aggregator — `.claude/lib/ontology_gen/aggregate.py`
- Reads each in-scope repo's committed `ontology/structural/code-graph.json` and rolls them into a single central view — **union of nodes/edges, namespaced by repo** (`<repo>/<id>`, e.g. `isnad-graph/src/api/app.py`) so ids never collide across repos (graphify `merge-graphs` design).
- **Offline, deterministic, no DB/LLM.** Reads whatever in-scope indices exist and **degrades gracefully** when one is absent (reports present/absent per repo).
- Pilot: main's own index is present (3760 nodes / 5762 edges → central artifact); the other 7 child indices (incl. isnad-graph via ig#1128) land in parallel and are picked up when present.
- Reuses #855 `Node`/`Edge`/`CodeGraph` + `serialize_graph`, so the central artifact is canonical (record-per-line, diff-friendly) and union-mergeable by the same driver.
- CLI: `python3 -m ontology_gen.aggregate <main-root> --out <central-graph.json>` (default out `ontology/structural/cross-repo-graph.json`).
- **Central artifact:** `ontology/structural/cross-repo-graph.json`.

### 2. Semantic-overlay wiring
- `ontology/conventions.md` § *Overlay → structural references (C×T2, #856)* defines the convention — the overlay **references** generated structural nodes by id (`structural_ref` / `structural_refs` in YAML; `[[structural:<id>]]` in markdown) instead of duplicating structure. Includes a resolvable worked example.
- Applied worked example to `ontology/repos/isnad-graph.yaml` backend modules.

### 3. Merge-driver registration
- `.gitattributes` maps `ontology/structural/{code-graph,cross-repo-graph}.json` → `merge=ontology-codegraph` (the #855 union driver).
- New `make setup-ontology-merge-driver` registers the driver command in local git config (it cannot live in `.gitattributes`).
- **Proven end-to-end:** divergent edits to `code-graph.json` on two branches union-merge cleanly, no conflict markers.

### Merge-driver script-invocation fix (wiring prerequisite)
`merge_driver.py` used `from .model import ...`, but git invokes a driver as a **plain script** (not `-m`) → relative import raised `ImportError` on the exact documented invocation. #855's unit tests passed only because they import `union_merge` from the package, never exercising the git script path (passing-repro-masks-bug). Added a script-mode `sys.path` bootstrap (same pattern the tests use) + a subprocess regression test. Flagging for the #855 contract author.

## Test
- New `test_ontology_aggregate.py`: namespacing, cross-repo no-collision, graceful degradation (absent/malformed), determinism, CLI.
- New plain-script merge-driver regression test in `test_ontology_gen.py`.
- Full local pre-commit + pre-push suite green: ruff/format, mypy, pytest (2049 passed), cspell, sync-drift gate.

Part of #856
Part of #820


parametrization:
**Requestor:** Weronika Zielinska
**Requestee:** Bereket Tadesse
**TechDebt:** none

Reviewed at PR head against the code, as the #855 contract author.

First, the catch on my own code: **Bereket is right, and this is a genuine bug in the merge_driver I shipped.** `from .model import` only resolves under `python3 -m`; git invokes a merge-driver as a plain script, so the *exact* documented `.gitattributes` invocation would have raised ImportError at the first real conflict. My #855 unit tests passed only because they import `union_merge` from the package — they never exercised the script path git actually uses. That's the passing-repro-masks-bug pattern, and it would have bitten silently the first time two branches diverged on `code-graph.json`. The fix is minimal and correct: keep the relative import as primary, fall back to a `sys.path`-bootstrapped absolute import, and — crucially — the new **subprocess regression test runs the driver exactly as git does** (`merge_driver.py %O %A %B %P`) and asserts the union. That closes the gap properly rather than patching around it.

On the aggregator itself:
- Namespacing by `<repo>/` prefix on every id, path, and edge endpoint is the right collision-avoidance design — two repos with the same relative path yield distinct nodes. It reuses my `Node`/`Edge`/`CodeGraph` + `serialize_graph` so the central artifact is the same canonical record-per-line shape and is itself union-mergeable by the same driver. Good composition.
- Graceful degradation is real: absent/unreadable/malformed indices return `None` and are skipped+reported, malformed *records* are filtered individually rather than aborting the roll-up. The central `cross-repo-graph.json` currently carries main's 3760/5762 and will pick up isnad-graph's when ig#1128 lands.

No contract objection — the contract is unchanged and correctly consumed. Approving.

parametrization:
**Requestor:** Nurul Hakim
**Requestee:** Bereket Tadesse
**TechDebt:** none

Second reviewer, output-integrity / operability lens. Read the artifact at head.

- **Determinism + integrity preserved:** the aggregator routes through `serialize_graph` (canonical record-per-line), so `cross-repo-graph.json` is record-granular-diffable and byte-stable across regens — same property as the per-repo index. Confirmed CI 20/20 green incl. JSON-syntax.
- **The merge-driver is now proven on the real path,** not just in-package: the subprocess test invokes the script as git would and the end-to-end union (divergent branch edits → clean union, no conflict markers) is demonstrated. This is the difference between "tests pass" and "the feature works," and it's now the latter.
- **Merge-driver setup is operable:** `.gitattributes` maps both graph artifacts to `merge=ontology-codegraph`, registered per-clone via `make setup-ontology-merge-driver`. Per-clone registration (not auto) is the correct git constraint — driver *commands* can't be committed, only `.gitattributes` can. Documented clearly.
- The overlay-reference convention (`structural_ref`/`[[structural:<id>]]`, namespaced ids) is lightweight and applied as a worked example to `isnad-graph.yaml` without rewriting the whole overlay — right scope for a pilot.
- The noted change-tracker non-update on the overlay edits is the known worktree cwd-anchor quirk (`feedback_hook_cwd_anchor_subagent_worktree`); reconciles at next `/ontology-rebuild` on main — not a defect in this PR.

No objection — two approvals stand, clear to merge. Approving.

PR #865, opened 2026-06-25
feat(lib): deterministic post-merge verification for deployable merges (#864)

parametrization:
## What

A deterministic, scriptable check that a **deployable merge** (wave→main, or any push-to-main) actually went green across the workflows that **only run post-merge** — and wires it into `/wave-wrapup`.

Closes #864.

## Why

Some workflows trigger on `push: branches: [main]` (and/or tags) but **not** `pull_request` — e.g. container publish + its Trivy scan. They give **zero pre-merge signal**, so `main` can go red right after a fully-green PR merges. That just happened: **isnad-graph#1131** — `#1130`'s PR was green, but the post-merge GHCR publish failed on a freshly-published base-image CVE (`libexpat` CVE-2026-45186). `gh pr merge` returns 0 the instant the merge commit exists, long before these workflows even start, so the merge's own exit status proves nothing.

## How

`.claude/lib/verify_deployable_merge.py <owner/repo> <sha>`:
1. Resolve the **expected** workflow set — explicit `--workflows`, or auto-detect from `.github/workflows/*` at the SHA. A merge to main is a *branch* push, so the required set is **branch-push-to-main workflows without `paths:` filters**. `--require-deployable` narrows to **post-merge-only** (push/tag AND not `pull_request`).
2. Poll the Actions runs for that exact `head_sha` until every required workflow has a *completed* run (or timeout).
3. Assert green. A failed run, or a required workflow that produced **no** run at all (silent drop — empty is hard not-verified, same discipline as `pr_ci_state.py`), fails. A **no-red safety net** additionally fails on *any* run that executed for the SHA and went red — so path-filtered workflows that did fire are still covered.

Exit codes: `0` verified / `1` not verified / `2` undetermined.

Correctly excludes **tag-only** workflows (isnad-graph's `Pipeline`, `push: tags: [data-v*]` — a tag push is not a branch merge) and **path-filtered** push workflows from the required set.

## Wiring

`/wave-wrapup` **Step 11.5a** (between the reachability gate and the staging gate): per repo, run `--require-deployable` against the wave→main merge SHA; a red/silently-dropped post-merge workflow blocks wave close. Override (`DEPLOYABLE_VERIFY_OVERRIDE_RATIONALE`) only for a documented external/standing red (advisory-DB drift, no-fix CVE under an active `--ignore-vuln`), and it must name a tracking issue. A red from the wave's own change is fixed forward, never overridden.

## Tests / parity

[...]

Issue #864, opened 2026-06-25
Deterministic post-merge verification for deployable merges (verify_deployable_merge.py)


parametrization:
## Problem

A PR's pre-merge CI is not the whole deployable picture. Some workflows are **push-to-main only** — they never run on `pull_request`, so they give **no pre-merge signal** and only execute *after* the merge lands on `main`. The canonical case: a container-publish workflow whose Trivy image scan runs on `push: branches: [main]` but not on PRs.

This just bit us: **isnad-graph#1131** — after #1130 merged to main, the GHCR publish workflow went red on a freshly-published base-image CVE (`libexpat` CVE-2026-45186). #1130's PR was fully green; the failure was invisible until post-merge. `gh pr merge` returns 0 the instant the merge commit exists, long before the push-triggered workflows even start.

There was no deterministic, scriptable way to assert "the deployable merge I just performed actually went green across the post-merge-only workflows too." Eyeballing `gh run list` is the manual step that rots (`feedback_enforcement_hierarchy`).

## Deliverable

`.claude/lib/verify_deployable_merge.py` — given `<owner/repo> <sha>`:
1. Resolve the *expected* workflow set: explicit `--workflows`, or auto-detect from `.github/workflows/*` at the SHA (push/tag-triggered; `--require-deployable` narrows to post-merge-only = push/tag AND not `pull_request`).
2. Poll the Actions runs for that exact `head_sha` until every expected workflow has a completed run (or timeout).
3. Assert every expected run concluded success/skipped. A failed run, or an expected workflow that produced **no** run (silent-drop — empty is hard not-verified, per `feedback_statuscheckrollup_ci_clean`), fails the check.

Exit codes: 0 verified / 1 not-verified / 2 undetermined.

## Wiring

- `/wave-wrapup` Step 11.5: after each wave→main merge, run it (`--require-deployable`) and treat a non-zero exit as a stop-the-line, before declaring the wave wrapped.
- Reusable for any deployable merge, not just wave wrapups.

## Tests

`.claude/lib/tests/test_verify_deployable_merge.py` — pure-logic + injected-clock poll-loop coverage (incl. the GHA-YAML `on: True` boolean gotcha, post-merge-only classification, failed/missing/pending verdicts). No network.

## Provenance

Owner directive 2026-06-24 ("check deployable merges complete successfully as a matter of regular practice... write a script to make it deterministic"). Surfaced by the isnad-graph#1131 post-merge Trivy failure during P6W17 wrapup.

PR #866, opened 2026-06-25
fix(lib): deployable-verify treats "nothing required" as pass, not error (#864)

parametrization:
## What

Follow-up correctness fix to #865: a deployable-merge verification where **nothing is required** (a repo with no post-merge-only / merge-triggered workflow — e.g. a meta repo whose CI is path-partitioned and PR-visible) now resolves to **VERIFIED (exit 0)** instead of an error, while still catching any red run via the no-red safety net.

Refs #864.

## Why

Dogfooding `verify_deployable_merge.py` against its **own** merge to `noorinalabs-main` returned exit 2 ("no expected workflows resolved"). The new `/wave-wrapup` Step 11.5a `if/else` would treat any non-zero as a block — so the gate would **false-block every meta / single-repo wave repo**. `noorinalabs-main` itself was in scope for P6W17, so this would have bitten the very next wrapup.

## Changes

- `verify()` no longer raises on an empty required set. It fetches the runs and applies the no-red safety net unconditionally — a red run fails verification whether or not anything was "required"; an empty required set with no red run is verified. **Exit 2 is now strictly real gh/API failure.**
- `Aggregate` carries `expected`, so the report prints the "nothing required" case explicitly (`no-red safety net only`).
- Step 11.5a wiring → explicit exit-code handling: `0` ok / `1` block / `2` undetermined-investigate.
- Tests: replaced the now-obsolete `no-expected-raises` with two — nothing-required-and-clean → verified, and nothing-required-but-a-red-run → caught by the net.

## Verified

`python3 .claude/lib/verify_deployable_merge.py noorinalabs/noorinalabs-main 1863528 --require-deployable` → `(no required post-merge workflows at this ref — no-red safety net only) / VERIFIED / exit 0`. ruff + mypy + pytest (26) green locally.

### DIFF
--- a/.claude/lib/tests/test_verify_deployable_merge.py
+++ b/.claude/lib/tests/test_verify_deployable_merge.py
@@ -299,13 +299,37 @@ class VerifyPollLoopTests(unittest.TestCase):
         self.assertFalse(result.verified)
         self.assertTrue(result.pending)
 
-    def test_no_expected_raises(self) -> None:
-        def empty_files(repo: str, ref: str) -> dict[str, str]:
-            return {"pr.yml": PR_ONLY_YAML}  # not deployable
+    def test_no_required_verifies_when_nothing_red(self) -> None:
+        # A repo with no post-merge-only workflow (only PR-visible CI) and a
+        # clean run set: "nothing required" → verified, NOT an error.
+        def only_pr(repo: str, ref: str) -> dict[str, str]:
+            return {"pr.yml": PR_ONLY_YAML, "tag.yml": TAG_ONLY_YAML}
 
-        self._patch(vdm, "fetch_workflow_files", empty_files)
-        with self.assertRaises(vdm.GhError):
-            vdm.verify("o/r", "sha", timeout=1, poll=1, sleep=lambda _s: None)
+        def clean_runs(repo: str, sha: str) -> list[dict]:
+            return []
+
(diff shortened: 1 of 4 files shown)


parametrization:
**Standards & Quality review**

Correct fix. "Nothing required" is genuinely not an error condition — a meta repo with no post-merge-only workflow has nothing that can fail post-merge that the required-set would catch, and the no-red safety net still runs unconditionally so a red run is caught regardless. Reserving exit 2 for real gh failures and switching Step 11.5a to explicit exit-code handling removes the false-block. Tests correctly replace the obsolete raise-case with the two meaningful paths (clean → verified, red → caught by net). Caught by dogfooding, which is the right way to find this.

---
Requestor: Aino Virtanen
Requestee: Santiago Ferreira
RequestOrReplied: Approved
TechDebt: none

parametrization:
**TPM review**

This would have bitten the next wrapup directly — noorinalabs-main is routinely in wave scope, and the prior wiring would have false-blocked its close. Good that it was found by dogfooding the tool against its own merge before it reached a live wrapup. Scope is tight and the exit-code contract is now unambiguous (0 ok / 1 block / 2 investigate). No new CI kind.

---
Requestor: Wanjiku Mwangi
Requestee: Santiago Ferreira
RequestOrReplied: Approved
TechDebt: none

Issue #870, opened 2026-06-25
ts-enum: extend C×T2 generator node-kind enum with interface/type kinds


parametrization:
## Summary
Extend the C×T2 structural generator's node-kind enum to include TypeScript `interface` and `type` declaration kinds. The isnad-graph pilot surfaced ~82 TS declarations that the current generator contract renders invisible (only class/function/const kinds are emitted), so the structural index under-represents TS-heavy repos.

This is a generator-contract change owned in `noorinalabs-main` (generator #855), and it is a precursor to the TS-repo fan-out (design-system, landing-page, isnad-graph frontend) under #820 — the structural index for those repos is incomplete until interface/type kinds are emitted.

## Acceptance Criteria
- [ ] Generator node-kind enum extended with `interface` and `type` kinds (and any other TS top-level decl kinds found under-counted in the pilot).
- [ ] Re-running the generator on a TS source set emits the previously-invisible declarations (validate against the ~82 isnad-graph pilot decls).
- [ ] Generator unit tests cover the new kinds.
- [ ] No regression for Python/existing kinds.
- [ ] Contract/docs (structural layer schema) updated to list the new kinds.

## Origin
Phase 7 Wave 18 framework carry-forward `820-ts-enum` (cross-repo-status.json `phase_7_carry_forwards`). Precedes TS-repo fan-out under #820. Owner directive: land before any tree-sitter backend swap.

## Provenance
Carry-forward from P6 ontology C×T2 execution (#820 Task 1/Task 3). Parent tracker: #820.

parametrization:
Requestor: Fatima Okonkwo
Requestee: Weronika Zielinska
RequestOrReplied: Request

**Wave 18 Kickoff — Phase 7**

This issue is assigned to you for p7-wave-18.
- Peer reviewer: Bereket Tadesse
- Secondary reviewer: Aino Virtanen
- Branch from: `deployments/phase-7/wave-18`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: enhancement (per charter § Wave Planning & Priority)

Please begin implementation.

PR #130, opened 2026-04-18
fix(hooks): exclude RequestOrReplied: Requested from review count (#123)

parametrization:
Closes #123

## Summary
- Tighten `validate_pr_review.py` comment-review detection so review-request assignment comments (`RequestOrReplied: Requested`) no longer count as reviews. Previously the hook triggered on mere presence of the `RequestOrReplied:` marker, flagging review-request comments as reviews missing `TechDebt:` and blocking merge (hit in PR #122).
- Now counts a comment as a review iff `RequestOrReplied` is one of `Approved`, `Changes Requested`, or `Replied`.
- Adds the charter-mandated Input Language docstring with embedded negative-match verification (`Requested` must be excluded; `Approved` missing `TechDebt:` must still block).
- Updates Hook 7's charter entry (`.claude/team/charter/hooks.md`) to document the tightened detection.

## Fix (line 160 area)
```python
# Before (buggy — any RequestOrReplied: marker counted as a review)
has_request_or_replied = re.search(r"RequestOrReplied:", body)

# After — only review OUTCOMES count
has_request_or_replied = re.search(
    r"\*{0,2}RequestOrReplied:\*{0,2}\s*"
    r"\*{0,2}(?:Approved|Changes Requested|Replied)\*{0,2}",
    body,
)
```

## Acceptance criteria (from #123)
- [x] Hook ignores `RequestOrReplied: Requested` (not a review)
- [x] Hook still catches `Approved` / `Changes Requested` / `Replied` as reviews
- [x] Regression test: review-request comment with both reviewers but no TechDebt does NOT block `gh pr merge`
- [x] Regression test: actual review missing TechDebt still blocks

## Test plan
Inline regex regression (11 cases) + end-to-end `check_comment_reviews` integration test with mocked `gh api` output covering both positive and negative scenarios — both passed locally.

## Notes for reviewer
- This is the 4th substring-bug of the W8 cluster (refs #113, #114, #118). Root cause is hooks written liberal without input-language spec; this PR exemplifies the corrected authorship requirements (charter § Hook Authorship Requirements).
- Dispatcher registration verified still present (`dispatcher.py:38`).
- Test pair documented in the hook's Input Language docstring (charter allows docstring-embedded manual verification where no test suite exists).
- Meta-caution: the review-request comment on this PR will trip the still-buggy deployed hook (this PR fixes it). Orchestrator may need `--admin` for the self-fix merge.

Requestor: Aino.Virtanen
Requestee: Lucas.Ferreira
[...]

PR #873, opened 2026-06-25
feat(ontology-gen): extend C×T2 NODE_KINDS with interface/type for TypeScript (#870)

parametrization:
## Summary

- Extends `NODE_KINDS` in `model.py` with `interface` and `type` kinds, capturing TypeScript interface declarations and type-alias declarations that were previously invisible (~82 decls in the isnad-graph pilot).
- `typescript_ext.py`: adds `_RE_INTERFACE` and `_RE_TYPE_ALIAS` regex patterns plus `_parse_extends_bases()` to extract `extends` clause names; emits the new kinds as `SymbolInfo` objects.
- `llms.py`: renders `interface` with bases (parenthesised, class-like style) and `type` without parens (type aliases have no params).
- `cypher_ext.py`: updates the contract-note docstring to reflect the extended enum.
- `test_ontology_gen.py`: adds `TestTypeScriptInterfaceTypeExtractor` — 9 tests covering extraction, graph assembly, `llms.txt` rendering, no-false-positive on `type` keyword in non-decl positions, and no-regression for existing Python/class/func/method kinds.

Part of #870

## Reviewers
Peer reviewer: Bereket Tadesse
Secondary reviewer: Aino Virtanen

TechDebt: none — strictly additive; regex scanner remains zero-dep. A tree-sitter backend could improve fidelity for generics-heavy patterns but is out of scope for #855 and deferred to a future issue.

## Test plan
- [ ] `ENVIRONMENT=test python3 -m pytest .claude/lib/tests/test_ontology_gen.py -v` — all 39 tests pass (9 new, 30 existing)
- [ ] `uvx ruff@0.11.12 format --check .claude/lib/` — clean
- [ ] `uvx ruff@0.11.12 check .claude/lib/` — clean
- [ ] `python3 -m mypy .claude/lib/ontology_gen/` — no issues in 11 source files
- [ ] Pre-push hooks: mypy, pytest, pytest-skills, memory-budget, headcount-budget, doc-freshness, office-drift, mermaid-render — all passed
- [ ] No regressions on Python/class/func/method kinds confirmed by `test_no_regression_existing_kinds`

### DIFF
--- a/.claude/lib/ontology_gen/cypher_ext.py
+++ b/.claude/lib/ontology_gen/cypher_ext.py
@@ -6,10 +6,11 @@ Zero-dependency regex scanner. Each query file becomes a single ``file`` node (l
 **clauses/operations** used — is captured into ``FileInfo.extra`` and rendered into the
 file's ``llms.txt`` section.
 
-Contract note (flagged for #856/#1128): the fixed ``code-graph.json`` node-kind enum is
-``{file, module, class, func, method}`` with no ``label``/``relationship`` kind, so
+Contract note (flagged for #856/#1128): the ``code-graph.json`` node-kind enum is
+``{file, module, class, interface, type, func, method}`` (``interface`` and ``type``
+added in main#870 for TypeScript). There is still no ``label``/``relationship`` kind, so
 Cypher labels and relationship types are intentionally **not** minted as graph nodes —
-that would require extending the contract enum. They live in the per-file ``llms.txt``
+that would require a further enum extension. They live in the per-file ``llms.txt``
 section (where an agent reads them) and in ``FileInfo.extra``. The graph contribution of
 a Cypher file is its ``file`` node. This is a deliberate decision, not an omission.
 """
--- a/.claude/lib/ontology_gen/llms.py
+++ b/.claude/lib/ontology_gen/llms.py
@@ -38,9 +38,13 @@ def _dedup(seq: list[str]) -> list[str]:
 
 def _render_symbol(sym: SymbolInfo, depth: int, lines: list[str]) -> None:
     indent = "  " * depth
-    if sym.kind == "class":
+    if sym.kind in ("class", "interface"):
+        # class/interface: show base/extends names in parentheses, no params.
         bases = f"({', '.join(sym.bases)})" if sym.bases else ""
-        lines.append(f"{indent}- class {sym.name}{bases} @{sym.line}")
+        lines.append(f"{indent}- {sym.kind} {sym.name}{bases} @{sym.line}")
+    elif sym.kind == "type":
+        # type alias: no params or bases, just the name.
+        lines.append(f"{indent}- type {sym.name} @{sym.line}")
     else:
         params = ", ".join(sym.params)
         lines.append(f"{indent}- {sym.kind} {sym.name}({params}) @{sym.line}")
--- a/.claude/lib/ontology_gen/model.py
+++ b/.claude/lib/ontology_gen/model.py
@@ -10,8 +10,12 @@ Two layers of types live here:
       {"nodes": [{id, kind, path, line, lang}],
        "edges": [{src, dst, type}]}
 
-  with ``kind ∈ {file, module, class, func, method}`` and
+  with ``kind ∈ {file, module, class, interface, type, func, method}`` and
   ``type ∈ {contains, imports, imports_from, calls, inherits, references}``.
+
+  ``interface`` and ``type`` were added in main#870 to capture TypeScript interface and
+  type-alias declarations that were previously invisible in TS-heavy repos (~82 decls
+  invisible in the isnad-graph pilot).
 
 * The **intermediate extraction** types — :class:`FileInfo`, :class:`SymbolInfo`,
   :class:`ImportInfo` — are the language-agnostic shape every per-language extractor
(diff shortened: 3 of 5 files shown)


parametrization:
Requestor: Aino Virtanen
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: Existing PR-body item — `_parse_extends_bases` naive comma-split can split inside nested generics (e.g. `Bar<Map<K, V>>`); the `[\w$]*$` regex guard saves the output in practice, but a proper top-level-comma splitter would be more robust. Accepted trade-off for the zero-dep pragmatic scanner; deferred to a future tree-sitter backend issue as noted in the PR.

---

**Standards & Quality review — contract/schema lens**

**Contract consistency across all enumeration sites** — all four sites are consistently updated and in agreement:

- `model.py` module docstring: `kind ∈ {file, module, class, interface, type, func, method}` — correct.
- `model.py` `NODE_KINDS` tuple at line 40: extends to `("file", "module", "class", "interface", "type", "func", "method")` — correct, with inline comment citing main#870.
- `cypher_ext.py` contract note: now enumerates the full extended set and notes the #870 origin — correct.
- `llms.py` `_render_symbol`: the new branches for `interface` (bases-in-parens, no params) and `type` (name only, no parens) are internally consistent with how `class` renders. Good.

No dangling enumeration site missed.

**Ordering/rank semantics** — `_KIND_RANK` is derived purely from `enumerate(NODE_KINDS)`, making `interface=3, type=4` while `func` shifts from 3→5 and `method` from 4→6. Rank is only used in `Node.__sort_key__` for tie-breaking nodes that share `(path, line)` and is never serialized to `code-graph.json`. Regenerating an existing Python-only structural index yields byte-identical JSON because `interface`/`type` nodes are simply absent from the sort input — no reordering of existing nodes. Clean.

**Back-compat of the kind enum** — the change is strictly additive. Consumers #856/#1128 currently receive only `{file, module, class, func, method}` nodes. Post-merge they continue to receive the same output from unchanged sources; new kinds appear only when TS sources are actually scanned. The contract note correctly flags downstream consumers to read before assuming the set is fixed.

**Test coverage is genuine** — 9 tests in `TestTypeScriptInterfaceTypeExtractor`. The extractor tests drive real regex against literal TS source strings (no mocks). The `test_interface_in_graph` / `test_type_in_graph` tests call the real `assemble()` function. The llms tests call the real `render_llms()`. The false-positive test (`type` keyword inside a function param list and body) validates the `^\s*(?:export\s+)?type\s+...=(?!=)` anchor correctly rejects it. The no-regression test confirms Python/class/method kinds are unaffected. Coverage meets all five acceptance criteria in #870.

**One observation on `_parse_extends_bases`** — simple comma-split with `split("<")[0]` + identifier-regex guard handles `Bar<T>, Baz` and even `Bar<Map<K, V>>, Baz` correctly (the `>` residue fails the identifier guard, so only `Bar` and `Baz` survive). This is clever and correct for the declared scope. The pathological failure would be a base name that contains a `>` literally, which is not valid TS identifier syntax — so no real inputs are mishandled.

**CI** — all 13 checks green: ruff-lint, ruff-format, mypy, pytest, commit-identity, sync-drift gate, memory/headcount budgets, smoke test, doc-freshness. No blockers.

All five issue #870 acceptance criteria are satisfied. Approved.

parametrization:
Requestor: Bereket Tadesse
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: none

## Review — PR #873 feat(ontology-gen): extend C×T2 NODE_KINDS with interface/type

CI is 13/13 green. I read all five changed files at head sha `2187959`.

### NODE_KINDS + _KIND_RANK

`NODE_KINDS = ("file", "module", "class", "interface", "type", "func", "method")` — ordering is sensible (TS structural kinds cluster between class and func). `_KIND_RANK` is derived via `enumerate(NODE_KINDS)` so it's always consistent by construction; no drift risk. Contract note in `cypher_ext.py` updated to match. Good.

### Regex patterns in typescript_ext.py

**`_RE_INTERFACE`** — anchored to line-start (`^\s*`), requires the `interface` keyword in declaration position, and the lookahead `(?=\s*\{|\s*$)` correctly constrains the match to lines that end the declaration (either opening brace follows, or end of line). Multi-line extends (body on the next line) still resolves: `[^{]+?` non-greedy crosses newlines, and the regex engine backtracks until `{` satisfies the lookahead. Verified this path mentally: `interface Foo extends\n  Bar,\n  Baz\n{` — group 2 captures `\n  Bar,\n  Baz\n`, then `_parse_extends_bases` strips and yields `["Bar", "Baz"]`. Clean.

**`_RE_TYPE_ALIAS`** — the `=(?!=)` negative lookahead correctly rejects `===` (equality comparisons) and `=>` (arrow functions). The `^\s*` anchor means `type` used as a parameter name (`function foo(type: string)`) or mid-expression never matches, since those positions aren't at line-start. The test `test_no_false_positive_type_in_non_decl_context` confirms this.

**`_parse_extends_bases`** — the docstring says "top-level commas" but the implementation splits on *all* commas. This is harmless in practice because generic residue like `B>` after splitting `Bar<A, B>` fails the identifier regex `^[A-Za-z_$][\w$]*$` and is filtered out — the result is correct. Minor comment inaccuracy, not a blocker.

### llms.py rendering

Interface is rendered class-style with bases in parentheses (`- interface Foo(Bar) @12`); type alias renders just the name (`- type Alias @5`). Correct: interfaces have extends relationships worth surfacing; type aliases are structural anchors without inheritance semantics. The `test_interface_llms_rendering` and `test_type_llms_rendering` tests assert exact output strings — they'll catch any rendering drift.

### Test coverage

Nine new tests in `TestTypeScriptInterfaceTypeExtractor` cover:
- Extraction (kind + bases on interface, kind on type aliases)
- Graph node assembly for both new kinds
- llms.txt rendering, including multi-extends interface
- No false-positive on `type` in non-declaration contexts
- No regression for Python class/method kinds

The existing `TestTypeScriptExtractor.test_symbols` covers pre-#870 class/func/arrow, so TS regression is also covered by the pre-existing suite.

### Minor stale comment (non-blocking)

`model.py` line 176: `SymbolInfo.kind: str  # one of: class, func, method` — `interface` and `type` should be added here now. Worth a follow-up edit but not a blocker.

### Verdict

Change is strictly additive, well-tested, CI fully green, and the regex patterns hold up against the cases that matter. The TS fan-out for #130/#155 is unblocked.

PR #880, opened 2026-06-25
feat(hooks): warn_zsh_wordsplit — PreToolUse advisory for bash-ism word-splitting (#879)

### LINKED ISSUE #879: zsh-safety: PreToolUse hook to flag bash-ism word-splitting (set -- $x / for x in $scalar) under zsh
## Problem

The dev environment's Bash tool runs under **zsh**, not bash. Several bash-only
idioms that rely on word-splitting of unquoted scalar variables work in bash but
silently misbehave under zsh:

- `set -- $spec` — zsh treats `$spec` as ONE field (no word-splitting), so
  positional parameters are set incorrectly.
- `for x in $scalar` — same: iterates over one token, not split words.
- `${!var}` / `${!arr[@]}` — bash indirect expansion; zsh uses `${(P)var}` / `${(k)arr}`.
- `mapfile` / `readarray` — bash-only builtins; absent in zsh.

## Existing guidance (prose-only, keeps getting violated)

- Memory: `.claude/memory/feedback_zsh_shell_environment.md`
- Convention: `ontology/conventions.md` § Shell environment (zsh)

Prose guidance alone has not been sufficient — this is a recurring failure class.

## Proposed fix

Add a **PreToolUse Bash advisory hook** (`warn_zsh_wordsplit.py`) that catches
these constructs and emits a `systemMessage` naming the exact pattern and the
zsh-safe rewrite. Advisory (not blocking) to avoid false-positive friction, per
`feedback_enforcement_hierarchy` (hook > skill > charter).

This is the memory→hook promotion for `feedback_zsh_shell_environment`.

parametrization:
## Summary

- Promotes `feedback_zsh_shell_environment` memory from prose-only guidance to an enforced PreToolUse Bash hook (`warn_zsh_wordsplit.py`), per the enforcement hierarchy (hook > skill > charter)
- Detects four bash-isms that silently misbehave under zsh (the dev-env shell and agent Bash tool): `set -- $scalar`, `for VAR in $scalar`, `${!indirect}`, and `mapfile`/`readarray`
- Advisory (`systemMessage`, not block) — mirrors `warn_pipe_mask_rc`'s stance; word-splitting reliance is sometimes intentional, hard block would generate false-positive friction
- Registered in `dispatcher.py` after `warn_ghcr_image`

## Files changed

- `.claude/hooks/warn_zsh_wordsplit.py` — new advisory hook
- `.claude/hooks/tests/test_warn_zsh_wordsplit.py` — 40 tests (20 FP-guards + 16 detection + 4 entrypoint)
- `.claude/hooks/dispatcher.py` — registers `warn_zsh_wordsplit` in `_BASH_HOOKS`

## Test plan

- [x] 40 tests pass: `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/test_warn_zsh_wordsplit.py -q`
- [x] ruff format + lint: `uvx ruff@0.15.11 format --check / check` — all green
- [x] mypy clean: `python3 -m mypy .claude/hooks/warn_zsh_wordsplit.py` — no issues
- [x] All pre-push hooks (ruff, mypy, pytest, cspell) passed before push
- [x] Does NOT block (returns `{"systemMessage": ...}`, no `"decision": "block"`)

TechDebt: None

Closes #879

### DIFF
--- a/.claude/hooks/dispatcher.py
+++ b/.claude/hooks/dispatcher.py
@@ -45,6 +45,7 @@ _BASH_HOOKS = [
     "validate_workflow_paths_coverage",
     "validate_vps_host",
     "warn_ghcr_image",
+    "warn_zsh_wordsplit",
 ]
 
 
--- a/.claude/hooks/tests/test_warn_zsh_wordsplit.py
+++ b/.claude/hooks/tests/test_warn_zsh_wordsplit.py
@@ -91,17 +91,33 @@ class NoMatchCases(unittest.TestCase):
         self.assertIsNone(wzw.check(_event("for x in a b c; do echo $x; done")))
 
     def test_for_quoted_scalar(self):
-        # "for x in "$list"" — the "$" is inside double-quotes; the regex checks
-        # for a literal $ followed immediately by an identifier, so this should
-        # not match because the preceding " makes it a quoted token.
-        # NOTE: our regex is on the raw stripped string, so "for x in "$list""
-        # DOES match the regex — we intentionally accept this FP risk in exchange
-        # for no bashlex dependency at advisory-hook level.  This test documents
-        # the KNOWN limitation: the advisory may fire on a quoted form when the
-        # quotes are not space-separated from the dollar. We test the common case
-        # only: `for x in "$list"` (with a space before the quote).
-        # The advisory is non-blocking, so this is acceptable friction.
-        pass  # documented limitation — not testing this exact form
+        # `for x in "$list"` — the `"` sits between `in ` and `$`, so the
+        # `\s+` + `_SCALAR_REF` adjacency never matches. Quoted scalars do not
+        # word-split in either shell, so this MUST stay silent.
+        self.assertIsNone(wzw.check(_event('for x in "$list"; do echo $x; done')))
+
+    def test_set_dashdash_quoted_scalar_explicit(self):
+        # `set -- "$list"` — quoted; must stay silent (companion to the for case).
+        self.assertIsNone(wzw.check(_event('set -- "$list"')))
+
+    # --- path / glob / concat prefixes: $NAME is only PART of a word (#879) ---
+
+    def test_for_path_glob_prefix_home(self):
+        # for f in $HOME/*.py — $HOME is a path prefix, not a standalone scalar.
+        self.assertIsNone(wzw.check(_event("for f in $HOME/*.py; do echo $f; done")))
+
+    def test_for_path_glob_prefix_repo_root(self):
+        self.assertIsNone(
+            wzw.check(_event("for w in $REPO_ROOT/.claude/worktrees/*/; do echo $w; done"))
+        )
+
+    def test_for_path_concat_prefix(self):
+        # for f in $dir/known_hosts — concatenation, single scalar path word.
+        self.assertIsNone(wzw.check(_event("for f in $dir/known_hosts; do echo $f; done")))
+
+    def test_set_dashdash_path_glob_prefix(self):
+        # set -- $dir/*.txt — glob with a scalar prefix; SAFE in zsh.
+        self.assertIsNone(wzw.check(_event("set -- $dir/*.txt")))
 
     def test_for_dollar_at_positional(self):
         # $@ and $* in for-in are safe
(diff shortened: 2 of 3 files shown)


parametrization:
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved

**Review: LGTM — advisory hook is well-structured; one TechDebt false-positive class documented below**

**Lens: operational safety + false-positive edge cases (Release Coordinator)**

---

**Edge-case analysis**

Four candidate false-positive shapes I probed against the regex logic directly:

| Command shape | Fires? | Correct? |
|---|---|---|
| `for f in $(gh pr list --json number --jq '.[].number'); do` | silent | ✓ — cmdSub excluded |
| `jq --arg x "$v" 'select(.name == $x)'` | silent | ✓ — no `for`/`set --` trigger |
| `for x in $list; do` | fires | ✓ — canonical target |
| `set -- $spec` | fires (via `_SET_DASHDASH_RE`) | ✓ — canonical target |
| `for f in $HOME/.ssh/known_hosts; do` | **fires** | ✗ — **false positive** (see TechDebt) |
| `for w in $REPO_ROOT/.claude/worktrees/*/; do` | **fires** | ✗ — **false positive** (see TechDebt) |
| `for f in $PWD/*.py; do` | **fires** | ✗ — **false positive** (see TechDebt) |
| `bash -c 'for x in $list; do ...'` | fires | marginal — caller explicitly chose bash context; advisory is non-blocking so acceptable |

In each glob-prefix case the regex matches `for VAR in $IDENTIFIER` (stops at the `/`), then `_SAFE_DOLLAR_FORMS_RE` finds no suppression trigger in the captured text `for f in $HOME`, and the advisory fires. The guard set covers `$@`, `$*`, `$(`, array expansions, and backticks — but does not cover the form `$SCALAR/path-suffix`, which is structurally different from word-splitting reliance.

**Positive-path coverage** ✓ — Both canonical patterns (`set -- $spec`, `for x in $scalar`) fire correctly. The quoted-safe forms and command-substitution forms are correctly silent.

**Parser robustness** ✓ — No bashlex dependency; pure regex on the `strip_heredocs`-cleaned string. The dispatcher wraps every `check()` call in `except Exception: continue`, so a `strip_heredocs` failure in production cannot block a Bash command. The standalone `main()` does not catch exceptions from `check()`, but that is consistent with every other hook in the codebase and is fine since standalone invocation is only for testing.

**Message quality** ✓ — Each advisory gives a concrete, correct, zsh-safe rewrite: `${(@s: :)var}`, `while IFS= read -r`, `${(P)var}`, `${(k)assoc}`, `arr=(${(f)"$(command)"})`. The pattern name and memory ref are both present.

---

Must-fix: None

TechDebt: Glob-prefix false positive — `for f in $dir/*.py`, `for w in $REPO_ROOT/.claude/worktrees/*/`, and similar path-expansion patterns fire the advisory even though `$VAR` is acting as a path prefix (not a word-split target). In zsh this idiom is safe: `$dir` expands as a scalar and the shell applies the trailing glob to the resulting path. The `_FOR_UNQUOTED_RE` regex fires on `for VAR in $IDENTIFIER` regardless of what follows the identifier. The test file documents the quoted-form limitation (`test_for_quoted_scalar: pass`) but does not acknowledge the path-prefix case. Suggested guard: after a regex match, check whether the character immediately following the identifier in the original command is `/` or `.` (path separator) and suppress if so. Alternatively, add a documented-limitation test case for `for f in $HOME/*.py → known false positive`. The hook is non-blocking so this is friction, not a break — but the team runs enough directory-iteration loops that false-positive accumulation could cause advisory-decay before the real signal fires.

parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved

**Review: LGTM — glob-prefix FP resolved; all three issues from prior round closed**

**c3b53c4 fix verified:** `_SCALAR_REF` + `_STANDALONE_BOUNDARY` correctly replaces the old capturing-group approach on both patterns. The standalone-word lookahead `(?=$|[\s;|&)\`])` silences `$dir/*.py`, `$REPO_ROOT/.claude/worktrees/*/`, and `$dir/known_hosts` (next char after identifier is `/`, which is not in the boundary set), while `for x in $list`, `for x in $a $b $c`, `for x in ${list}`, `set -- $spec`, and `set -- ${spec}` still fire correctly. Verified locally against all 12 key cases.

**TechDebt #1 (dead `_SAFE_DOLLAR_FORMS_RE`)**: Removed. `_scan_for_wordsplit` docstring now explicitly documents why: the boundary+identifier constraints make the secondary guard unreachable. Closed.

**TechDebt #2 (`test_for_quoted_scalar` pass no-op)**: Converted to a real `assertIsNone` with an accurate comment. Companion `test_set_dashdash_quoted_scalar_explicit` added. Comment now correctly explains that `"` precedes `$` and the `\s+` + `_SCALAR_REF` adjacency never matches. Closed.

**Test count**: 49 (was 40; 9 new tests covering glob-prefix silent cases, braced scalars, and multi-scalar fire cases). All three gaps called out in prior reviews are now covered.

Must-fix: None
TechDebt: None

PR #889, opened 2026-06-25
fix(ontology_gen): depth-aware top-level-comma splitter for TS extends bases (#887)

### LINKED ISSUE #887: ontology_gen TS extractor: _parse_extends_bases splits inside nested generics (top-level-comma splitter / tree-sitter backend)
## Summary

`_parse_extends_bases` (in the ontology_gen TypeScript extractor, `typescript_ext.py`) splits an `extends`/`implements` base list on **all** commas, which can split inside nested generics — e.g. `Bar<Map<K, V>>, Baz` naively splits at the inner `, `. In practice the `^[A-Za-z_$][\w$]*$` identifier-regex guard discards the `V>>`-style residue, so current outputs are correct, but the parser is fragile and a proper **top-level-comma** splitter (or a tree-sitter backend) would be robust.

## Context

Deferred tech-debt noted on PR #873 (extend C×T2 NODE_KINDS with interface/type for TypeScript) and in its review verdict — "accepted trade-off for the zero-dep pragmatic scanner; deferred to a future tree-sitter backend."

## Proposed fix

Either (a) a small top-level-comma splitter that tracks `<>`/`()`/`[]` nesting depth and only splits at depth 0, or (b) the broader tree-sitter backend swap (larger; would also subsume other regex-fragility in the TS extractor). Add tests with `Bar<Map<K, V>>, Baz` and `Foo<A, B>, Bar<C>` asserting `["Bar","Baz"]` / `["Foo","Bar"]`.

## Provenance

Deferred at PR #873 (P7W18); pulled into W19 (global-19) as framework TD by owner directive 2026-06-25.

parametrization:
## Summary

`_parse_extends_bases` (ontology_gen TS extractor, `typescript_ext.py`) split `extends`/`implements` base lists on **every** comma, which could split inside a nested generic — e.g. `Bar<Map<K, V>>, Baz` at the inner `, `. The `^[A-Za-z_$][\w$]*$` identifier guard discarded the residue so outputs happened to be correct, but the parser was fragile.

This is **Option (a)** from the issue — the small, zero-dep top-level-comma splitter — **not** the deferred tree-sitter backend swap (kept consistent with the intentional zero-dependency pragmatic scanner, conventions.md §49).

## Approach

- New shared helper `_split_top_level_commas(raw)` tracks `<>` / `()` / `[]` / `{}` nesting depth and only treats a comma at depth 0 as a split point (empty/whitespace segments dropped).
- `_parse_extends_bases` now splits via the helper before stripping generics, so `Bar<Map<K, V>>, Baz` → `["Bar", "Baz"]`.
- Routed the existing `_split_params` through the same helper too, deduping its previously-inline depth tracker (single source of truth; covered by existing param tests).

## Tests (`.claude/lib/tests/test_ontology_gen.py`, new `TestExtendsBaseSplitter`)

- `Bar<Map<K, V>>, Baz` → `["Bar", "Baz"]`
- `Foo<A, B>, Bar<C>` → `["Foo", "Bar"]`
- `()` and `[]` nesting; deeply-nested generic `A<B<C, D<E, F>>>, G` → `["A", "G"]`
- plain list / single base / empty clause regressions
- low-level splitter keeps the nested segment intact
- end-to-end interface declaration: `interface Repo extends Bar<Map<K, V>>, Baz` → bases `["Bar", "Baz"]`

## Verification

- `pytest .claude/lib/tests/` — 618 passed (48 in the TS module)
- `ruff format --check`, `ruff check`, `mypy` — clean
- All pre-commit + pre-push hooks green on push

## Reviewers

- Primary: **Aino Virtanen**
- Secondary: **Nurul Hakim**

Closes #887

### DIFF
--- a/.claude/lib/ontology_gen/typescript_ext.py
+++ b/.claude/lib/ontology_gen/typescript_ext.py
@@ -74,13 +74,43 @@ def _line_of(text: str, pos: int) -> int:
     return text.count("\n", 0, pos) + 1
 
 
+def _split_top_level_commas(raw: str) -> list[str]:
+    """Split ``raw`` on commas that sit at bracket-nesting depth 0.
+
+    Commas nested inside ``<>`` / ``()`` / ``[]`` / ``{}`` are NOT split points, so a
+    base list like ``Bar<Map<K, V>>, Baz`` splits into ``["Bar<Map<K, V>>", "Baz"]``
+    rather than naively at the inner ``, `` (#887). Empty / whitespace-only segments
+    are dropped. Shared by both the ``extends``/``implements`` base-list parser and the
+    parameter splitter so neither re-implements the depth tracking.
+    """
+    parts: list[str] = []
+    depth = 0
+    current = ""
+    for ch in raw:
+        if ch in "([{<":
+            depth += 1
+        elif ch in ")]}>":
+            depth = max(0, depth - 1)
+        if ch == "," and depth == 0:
+            if current.strip():
+                parts.append(current)
+            current = ""
+        else:
+            current += ch
+    if current.strip():
+        parts.append(current)
+    return parts
+
+
 def _parse_extends_bases(raw: str) -> list[str]:
     """Return base names from a raw ``extends A, B<T>`` clause (drops generics).
 
-    Splits on top-level commas and strips generic type-parameters so ``Foo<T>`` → ``Foo``.
+    Splits on top-level commas (depth-aware, so a comma inside a nested generic such as
+    ``Map<K, V>`` does not split — #887) and strips generic type-parameters so
+    ``Foo<T>`` → ``Foo``.
     """
     bases: list[str] = []
-    for part in raw.split(","):
+    for part in _split_top_level_commas(raw):
         name = part.strip().split("<")[0].strip()
         if name and re.match(r"^[A-Za-z_$][\w$]*$", name):
             bases.append(name)
@@ -88,27 +118,8 @@ def _parse_extends_bases(raw: str) -> list[str]:
 
 
 def _split_params(raw: str) -> list[str]:
-    raw = raw.strip()
-    if not raw:
-        return []
-    params: list[str] = []
-    depth = 0
-    current = ""
-    # Split on top-level commas only (object/array/generic destructuring may nest).
-    for ch in raw:
-        if ch in "([{<":
-            depth += 1
-        elif ch in ")]}>":
-            depth -= 1
-        if ch == "," and depth == 0:
-            params.append(current)
-            current = ""
-        else:
-            current += ch
-    if current.strip():
-        params.append(current)
     out: list[str] = []
-    for part in params:
+    for part in _split_top_level_commas(raw):
         # Keep just the parameter name (drop type annotations / defaults).
         name = part.strip().split(":")[0].split("=")[0].strip()
         name = name.lstrip(".")  # rest params ...args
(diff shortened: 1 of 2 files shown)


parametrization:
**Review — PR #889 generics splitter** (verified at head `a3f3426283fd`)

Confirmed the fix against main#887. `_parse_extends_bases` previously split the raw `extends`/`implements` clause on every comma, which fractured a nested generic such as `Bar<Map<K, V>>` at its inner comma; the identifier guard masked the symptom but the parse was fragile. The new shared `_split_top_level_commas` helper tracks `<>`/`()`/`[]`/`{}` nesting and only treats a comma at depth 0 as a split point, so `Bar<Map<K, V>>, Baz` now yields `Bar`, `Baz` and `Foo<A, B>, Bar<C>` yields `Foo`, `Bar`. Traced the depth accounting by hand on the nested and deeply-nested cases and it holds; the `max(0, depth - 1)` clamp also hardens against unbalanced closers without altering well-formed behavior. Empty/whitespace segments are dropped, matching the prior end-of-clause handling.

The `_split_params` refactor routes through the same helper, deduping a previously-inline depth tracker. The old tracker used the identical bracket sets and depth rule, and `_split_params` still filters empty names downstream, so param parsing is behaviourally unchanged — the existing param tests (part of the 618-pass run) cover this. Scope is exactly Option (a) from the issue (zero-dep pragmatic scanner, consistent with conventions §49); the tree-sitter backend is correctly left deferred. No scope creep, and the new `TestExtendsBaseSplitter` cases exercise the actual nested-generics bug both at the helper level and end-to-end through `extract_typescript`. CI is fully green (Ruff format/lint, mypy, pytest, sync-drift gate all passing).

No blockers. Approving.

---

Requestor: Nurul Hakim
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Review — PR #889 depth-aware top-level-comma splitter for extends/implements bases** (verified at head `a3f3426283fd`)

Verified the code at HEAD via the contents API, not the working tree.

The fix is correct and well-scoped. `_split_top_level_commas` tracks `<>` / `()` / `[]` / `{}` nesting and only treats a comma at depth 0 as a split point, so `Bar<Map<K, V>>, Baz` yields `["Bar<Map<K, V>>", "Baz"]` and the inner generic comma is no longer a false boundary (main#887). `_parse_extends_bases` then strips the generic suffix as before, giving `["Bar", "Baz"]`. Both the brief's assertions hold: `Bar<Map<K, V>>, Baz` and `Foo<A, B>, Bar<C>` resolve to the bare base names.

The refactor that has `_split_params` reuse the same splitter does not regress parameter parsing. The previous explicit empty-input guard is subsumed — an empty or whitespace-only `raw` returns an empty list from the splitter — and empty segments are dropped, with the downstream `if name` filter keeping only real parameter names. Behaviour is equivalent for non-generic lists. The switch to `depth = max(0, depth - 1)` is a small robustness improvement that prevents a stray closing token from driving depth negative, and it does not alter any valid-input result.

Tests genuinely exercise the bug: nested single and double generics, paren and bracket nesting, deeply nested generics, plain lists, single base, empty clause, the low-level splitter keeping a nested segment intact, and an end-to-end interface declaration through `extract_typescript`. All 13 CI checks are green. No scope creep, zsh/charter conventions honoured.

---

Requestor: Aino Virtanen
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: none

PR #890, opened 2026-06-25
fix(wave_seq): reservation-aware allocate so retro-reserved wave id is not skipped (#885)

### LINKED ISSUE #885: wave_seq double-allocation: retro reserves wave id without bumping global_wave_seq, /wave-scope allocate skips it
## Summary

`/wave-retro` Step 9 and `/wave-scope` Step 0.0 double-allocate the global wave id. The retro **reserves** the next id by writing `wave_{N}_meta_issue` (e.g. `wave_19_meta_issue`) **without** bumping the `global_wave_seq` counter. `wave_seq.py`'s `seed_value()` then counts that new `wave_19_*` key as an "existing wave id", so the subsequent `/wave-scope` `allocate` computes `next = max(global_wave_seq, max_existing_wave_key) + 1 = max(18, 19) + 1 = 20` — **skipping 19**, the id the retro already reserved and that the meta-issue (#882, "Phase 7 Wave 19") + phase-plan ("W2") + `wave_19_meta_issue` all reference.

## Reproduction (observed P7 W18→W19, 2026-06-25)

1. `/wave-retro 7 18` Step 9: `wave_seq.py peek` → 19; creates stub #882 "Phase 7 Wave 19"; writes `wave_19_meta_issue` (NOT `global_wave_seq`, still 18).
2. `/wave-scope 7 19` Step 0.0: `allocate --write` → `global wave id: 20`, stamps `wave_20_phase`/`wave_20_phase_ordinal`. Conflicts with the reserved 19.

Worked around by reverting the bad `wave_20_*` stamps and reconciling `global_wave_seq=19` + `wave_19_phase/phase_ordinal`.

## Root cause

The retro's id-reservation (a `wave_{N}_*` key) and the counter (`global_wave_seq`) are two sources of truth that disagree. `seed_value()` self-seeds above any existing `wave_{X}_*` key — a deliberate migration safety — but that makes a retro-reserved-but-uncommitted id look like an *already-allocated* id to the next `allocate`, bumping past it.

## Proposed fix (pick one)

1. **Retro Step 9 bumps the counter when it reserves** — run `wave_seq.py allocate --phase {P} --write` at retro time (instead of bare `peek` + writing `wave_{N}_meta_issue`), so `global_wave_seq` and the keyspace advance together. `/wave-scope` then reads the already-allocated id rather than re-allocating.
2. **Make `allocate` idempotent / reservation-aware** — if `wave_{peek}_meta_issue` already exists for the peeked id, treat it as the current allocation (return it, stamp phase/ordinal) instead of `+1`.

Option 1 is cleaner (single allocation point); Option 2 is defensive. Add a regression test: retro-reserve `wave_N`, then `allocate` must return `N`, not `N+1`.

## Retro action

Surfaced + worked around during `/wave-scope 7 19`. Carry-forward to a P7 framework/tooling slot.

parametrization:
## Summary

Fixes the `wave_seq.py` double-allocation (#885): `/wave-retro` Step 9 *reserves* the next wave id by writing `wave_{N}_meta_issue` **without** bumping `global_wave_seq` (counter stays N-1). `seed_value()` then counts that reserved `wave_{N}_*` key as already-allocated, so `allocate` computed `max(global_wave_seq, max_existing_key)+1` and **skipped** N — allocating 20 instead of 19 in the P7 W18→W19 transition.

## Approach — Option 2 (reservation-aware allocate)

Chose **Option 2** (mechanical fix in `wave_seq.py`) over Option 1 (skill-doc discipline) because it makes the tool correct **regardless of caller ordering** — the keyspace and `global_wave_seq` can never disagree by construction (single source of truth in code, not a retro-skill convention that can rot).

- `reserved_wave(status)` — the id at the **explicit committed counter** `global_wave_seq + 1` whose `wave_{N}_meta_issue` key already exists. Deliberately reads `global_wave_seq` (NOT the seed-inflated `current_seq`, which would re-introduce the skip). Returns `None` when there is no committed counter (migration safety) or no reservation.
- `allocation_target(status)` — pending reservation if present, else `next_global_wave`.
- `peek` / `allocate` now both use `allocation_target` so they stay consistent.
- Folded in an Option-1 **clarifying note** in `wave-retro/SKILL.md` Step 9: the reserve-without-bump split is now safe because the allocator is reservation-aware (and a warning not to "fix" it by adding a counter bump, which would double-advance).

## Tests (`.claude/lib/tests/test_wave_seq.py` — `TestRetroReservationAwareness`)

- `reserved_wave` detects the pending reservation; returns `None` without the meta_issue key or without a committed counter.
- **Headline regression**: post-retro state (`global_wave_seq=18`, `wave_19_meta_issue` set, counter NOT advanced) → `next_global_wave` is the old wrong `20`, but `allocation_target`/`peek` return **19**.
- `allocate --write` claims **19**, advances the counter to 19, stamps `wave_19_phase=7` / `phase_ordinal=2`, and **never** creates `wave_20_*` keys.
- Pass-through unchanged for normal mid-wave state (no reservation).

Full suite green: `2181 passed` (lib + hooks), ruff format/lint, mypy, cspell, sync-drift gate all clean.

## Reviewers

- Primary: **Weronika Zielinska**
- Secondary: **Lucas Ferreira**

Closes #885

### DIFF
--- a/.claude/lib/tests/test_wave_seq.py
+++ b/.claude/lib/tests/test_wave_seq.py
@@ -135,6 +135,89 @@ class TestNoCrossPhaseCollision(unittest.TestCase):
         self.assertEqual(status["wave_17_phase_ordinal"], 2)
 
 
+class TestRetroReservationAwareness(unittest.TestCase):
+    """Regression for main#885: ``/wave-retro`` Step 9 reserves the next id by
+    writing ``wave_{N}_meta_issue`` WITHOUT bumping ``global_wave_seq`` (it stays
+    N-1). The subsequent ``/wave-scope`` ``allocate`` must claim N — NOT skip to
+    N+1 as it did in the P7 W18→W19 transition (allocated 20 instead of 19)."""
+
+    def _post_retro_reservation(self) -> dict:
+        """Status exactly as /wave-retro Step 9 leaves it: P7W18 committed
+        (global_wave_seq=18, wave_18 stamped phase 7 ordinal 1), and wave_19
+        reserved via its meta_issue key ONLY — counter NOT advanced, no phase
+        stamp on 19 yet (mirrors the upsert the retro skill actually performs)."""
+        return {
+            "current_phase": 7,
+            "current_wave": "wave-18",
+            "global_wave_seq": 18,
+            "wave_18_phase": 7,
+            "wave_18_phase_ordinal": 1,
+            "wave_18_scope": {"phase": 7, "theme": "P7W1"},
+            # Retro Step 9 reservation: meta_issue key, counter still 18.
+            "wave_19_meta_issue": "noorinalabs-main#882",
+        }
+
+    def test_reserved_wave_detects_pending_reservation(self) -> None:
+        self.assertEqual(wave_seq.reserved_wave(self._post_retro_reservation()), 19)
+
+    def test_reserved_wave_none_without_meta_issue_key(self) -> None:
+        status = self._post_retro_reservation()
+        del status["wave_19_meta_issue"]
+        self.assertIsNone(wave_seq.reserved_wave(status))
+
+    def test_reserved_wave_none_when_counter_absent(self) -> None:
+        # No committed counter → cannot distinguish reserved from committed;
+        # fall back to the seed-safe monotonic path.
+        status = self._post_retro_reservation()
+        del status["global_wave_seq"]
+        self.assertIsNone(wave_seq.reserved_wave(status))
+
+    def test_allocation_target_returns_reserved_id(self) -> None:
+        # The headline bug: bare next_global_wave skips to 20; the
+        # reservation-aware target claims the reserved 19.
+        status = self._post_retro_reservation()
+        self.assertEqual(wave_seq.next_global_wave(status), 20)  # the old (wrong) value
+        self.assertEqual(wave_seq.allocation_target(status), 19)
+
+    def test_allocation_target_falls_through_when_no_reservation(self) -> None:
+        # Normal mid-wave state (no pending meta_issue at counter+1) is unchanged.
+        status = _grandfathered_p6_status()
+        self.assertEqual(wave_seq.allocation_target(status), wave_seq.next_global_wave(status))
+
+    def test_peek_returns_reserved_id_not_skip(self) -> None:
+        with TemporaryDirectory() as d:
+            path = Path(d) / "cross-repo-status.json"
+            path.write_text(json.dumps(self._post_retro_reservation(), indent=2) + "\n")
+            import io
+            from contextlib import redirect_stdout
+
+            buf = io.StringIO()
+            with redirect_stdout(buf):
+                rc = wave_seq.main(["peek", str(path)])
+            self.assertEqual(rc, 0)
+            self.assertEqual(buf.getvalue().strip(), "19")
+
+    def test_allocate_write_claims_reserved_id_and_advances_counter(self) -> None:
+        with TemporaryDirectory() as d:
+            path = Path(d) / "cross-repo-status.json"
+            path.write_text(json.dumps(self._post_retro_reservation(), indent=2) + "\n")
+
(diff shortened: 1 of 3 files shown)


parametrization:
**Review — PR #890 reservation-aware wave_seq allocate** (verified at head `81840935c9eb`)

Verified the fix against main#885 (the P7 W18→W19 transition allocated 20 instead of 19).

The crux holds. `reserved_wave()` keys off the **explicit committed** counter via `status.get("global_wave_seq")` — NOT the seed-inflated `current_seq` — so it cannot re-introduce the very skip it guards against; `candidate = committed + 1` and the reservation is recognised only when `wave_{candidate}_meta_issue` exists. With `global_wave_seq=18` and a `wave_19_meta_issue` reservation, it returns 19. `allocation_target()` cleanly routes to the reservation when present and falls through to `next_global_wave()` otherwise (confirmed unchanged for normal mid-wave state), and both `peek`/`allocate` now consume `allocation_target`.

The write path is correct end-to-end: `_cmd_allocate` persists `global_wave_seq={wave_id}` set to the reserved id 19 (not incremented past it), stamps `wave_19_phase`/`wave_19_phase_ordinal` (ordinal 2, since P7 already holds wave 18), and never fabricates any `wave_20_*` key — all routed through `upsert_status_keys` so the compact-inline shape is preserved. The migration guard returns None when `global_wave_seq` is absent, falling back to the seed-safe monotonic path.

The 7 new regression tests in `TestRetroReservationAwareness` exercise the reproduction precisely: the old-and-wrong `next_global_wave` returning 20 is pinned alongside the corrected `allocation_target` returning 19, plus the end-to-end `--write` asserting the counter lands on 19 and no `wave_20_*` keys appear. The wave-retro SKILL.md comment correctly documents the reserve-without-bump contract and warns against a double-advancing "fix". No scope creep; zsh/charter conventions respected; CI fully green (all 20 checks SUCCESS).

---

Requestor: Weronika Zielinska
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Review — PR #890 reservation-aware wave_seq allocate** (verified at head `81840935c9eb`)

Verified the fix at head against main#885 (the P7 W18→W19 transition that allocated 20 instead of 19).

Root cause and fix are correct. `/wave-retro` Step 9 claims the next id in the keyspace by writing `wave_{N}_meta_issue` while leaving the committed counter at N-1, which inflates `seed_value` so the naive `next_global_wave` skips past N. The new `reserved_wave()` keys strictly off the EXPLICIT committed counter (`status.get("global_wave_seq") + 1`) — not the seed-inflated `current_seq` — and only treats N as reserved when `wave_{N}_meta_issue` already exists. That is the right discriminator; using `current_seq` would have re-introduced the very skip it guards against. `allocation_target()` returns the reservation when present else the monotonic next, and both `peek` and `allocate` route through it.

Checked the `--write` persistence path at head (not just the diff): it upserts `global_wave_seq={wave_id}` with `wave_id` = the reserved 19, plus `wave_19_phase`/`wave_19_phase_ordinal`, and never emits any `wave_20_*` key. The phase ordinal is computed against the as-yet-unstamped state, so the reserved wave correctly lands as phase-7 ordinal 2 behind wave 18.

Idempotency holds: after a write advances the counter to 19, a subsequent allocate sees no reservation at 20 (`wave_20_meta_issue` absent) and falls through to the monotonic next, so the reserved id is never double-claimed. The migration guard (no committed counter returns None) is sensible.

Tests exercise the exact repro: `next_global_wave` returns the old wrong 20 while `allocation_target` returns 19, `peek` prints 19, and the end-to-end `--write` asserts the counter lands on 19 with no `wave_20_*` keys and the reservation key preserved. Reservation-absent and counter-absent fall-through cases are covered. No scope creep; the wave-retro SKILL comment documents the intentional reserve/advance split. CI is fully green.

---

Requestor: Lucas Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

PR #891, opened 2026-06-25
fix(hooks): narrow validate_pr_review batch-loop guard to in-loop merges (#886)

### LINKED ISSUE #886: validate_pr_review over-broad: blocks non-merge gh subcommands (e.g. gh run rerun)
## Summary

The `validate_pr_review` PreToolUse hook (batch-loop `gh pr merge` guard) fired a block on a **non-merge** `gh run rerun` command during the P7W18 wrapup. The command contained no `gh pr merge` — it was `gh run rerun "$ds_run" --repo … --failed` (re-running a failed staleness check) inside a multi-line block. Annunaki captured it as a `pretooluse_block`.

## Why it matters

The guard's purpose is to catch `gh pr merge "$pr"` in a shell loop (which fails-open the 2-reviewer gate — `feedback_batch_loop_merge_evades`). Matching unrelated `gh` subcommands (`gh run rerun`, etc.) is over-broad and adds advisory friction without protecting anything. Low severity (the gate is fail-safe and the friction is occasional), but worth tightening.

## Proposed fix

Narrow the matcher to actual `gh pr merge` invocations (or `gh pr merge` with a loop-variable PR ref specifically), so non-merge `gh` subcommands don't trip it. Add a regression test from the real `gh run rerun --failed` command shape. Alternatively, if the conservatism is deemed acceptable per `feedback_safety_direction_over_ux_friction`, document the decision and close.

## Provenance

Surfaced at the P7W18 retro (annunaki-attack); pulled into W19 (global-19) as framework TD by owner directive 2026-06-25.

parametrization:
## Summary

Narrows the `validate_pr_review` batch-loop `gh pr merge` guard so it triggers only on an actual variable-PR merge **inside a loop body**, fixing the over-broad false-block that fired on a non-merge `gh run rerun "$ds_run" --failed` during the P7W18 wrapup.

## Root cause

`is_variable_pr_merge_in_loop` matched the variable-PR-merge pattern (`_GH_MERGE_VAR_PR`) and the loop keywords (`for`/`while`/`until` + `do`) **independently anywhere in the command**. It never verified co-location. So a multi-line block pairing a non-loop `gh pr merge "$PR"` with a *separate, unrelated* staleness-recheck loop `for r in …; do gh run rerun "$r" --failed; done` tripped the guard — even though the merge was not in a loop and the gate did not fail-open.

## Fix

- New `_loop_body_spans()` helper pairs `do … done` keywords via a stack (loop-only keywords ⇒ each balanced pair is a loop body; nested loops handled), computed on the quote-normalized view so quoted `--body "…"` payloads don't create spans.
- `is_variable_pr_merge_in_loop` now requires a `gh pr merge <var>` match to fall **inside** a loop-body span.
- Removed the now-unused independent `_LOOP_OPENER` / `_LOOP_BODY` regexes; updated module + function docstrings.

## Protection preserved

- `for pr in …; do gh pr merge "$pr"; done` (the fail-open evasion class from `feedback_batch_loop_merge_evades`) is still HARD BLOCKED.
- A standalone literal `gh pr merge <N>` is untouched (gated by the literal path).
- `--admin` still overrides.

## Tests

- `test_886_gh_run_rerun_loop_not_blocked` — exact `gh run rerun "$ds_run" --failed` loop shape → not blocked.
- `test_886_nonloop_merge_with_unrelated_rerun_loop_not_blocked` — the real false-positive (would have returned `True` pre-fix) → not blocked.
- `test_886_batch_loop_merge_still_blocked` — positive: in-loop variable merge still blocked.
- Full hook suite green: 1568 passed.

## Reviewers

Primary: Aino Virtanen — Secondary: Nino Kavtaradze (2 distinct Approved verdicts required; reviewer-first).

Closes #886

### DIFF
--- a/.claude/hooks/tests/test_validate_pr_review.py
+++ b/.claude/hooks/tests/test_validate_pr_review.py
@@ -1608,6 +1608,43 @@ class BatchLoopMergeDetectorTests(unittest.TestCase):
             hook.is_variable_pr_merge_in_loop('for pr in 1 2; do gh pr view "$pr"; done')
         )
 
+    def test_886_gh_run_rerun_loop_not_blocked(self):
+        # #886 regression — the EXACT non-merge shape that false-blocked at the
+        # P7W18 wrapup: a `gh run rerun "$ds_run" --failed` staleness recheck
+        # inside a for-loop. No `gh pr merge` anywhere → must not match.
+        self.assertFalse(
+            hook.is_variable_pr_merge_in_loop(
+                "for ds_run in $(gh run list --repo noorinalabs/noorinalabs-deploy "
+                '--json databaseId --jq ".[].databaseId"); do '
+                'gh run rerun "$ds_run" --repo noorinalabs/noorinalabs-deploy --failed; done'
+            )
+        )
+
+    def test_886_nonloop_merge_with_unrelated_rerun_loop_not_blocked(self):
+        # #886 regression (the actual false-positive mechanism): a multi-line
+        # block pairing a NON-loop variable merge with a SEPARATE, unrelated
+        # `gh run rerun` staleness loop. The merge is not inside the loop body,
+        # so the gate does not fail-open and the guard must NOT fire. Before the
+        # #886 narrowing this returned True (merge-var pattern + loop keywords
+        # matched independently anywhere in the command).
+        cmd = (
+            'gh pr merge "$PR" --repo noorinalabs/noorinalabs-deploy --merge\n'
+            'echo "wave merged"\n'
+            "for ds_run in 11 12 13; do "
+            'gh run rerun "$ds_run" --repo noorinalabs/noorinalabs-deploy --failed; done'
+        )
+        self.assertFalse(hook.is_variable_pr_merge_in_loop(cmd))
+
+    def test_886_batch_loop_merge_still_blocked(self):
+        # #886 must NOT weaken the real protection: a `gh pr merge "$pr"` that is
+        # genuinely INSIDE the loop body (the fail-open evasion class from
+        # `feedback_batch_loop_merge_evades`) is still detected.
+        self.assertTrue(
+            hook.is_variable_pr_merge_in_loop(
+                'for pr in 48 49 50; do gh pr merge "$pr" --repo o/r --merge; done'
+            )
+        )
+
 
 class BatchLoopMergeEndToEndTests(unittest.TestCase):
     """#567 end-to-end: check() HARD BLOCKS a batch-loop variable merge, and a
--- a/.claude/hooks/validate_pr_review.py
+++ b/.claude/hooks/validate_pr_review.py
@@ -19,13 +19,17 @@ Input Language:
                the PR in the repo the user named, not the cwd-resolved repo.
     --admin  → short-circuits (emergency override — allows merge).
 
-  Batch-loop guard (#567): a `gh pr merge <shell-var>` inside a
-  for/while/until loop (e.g. `for pr in 48 49; do gh pr merge "$pr" …; done`)
-  is HARD BLOCKED. The literal-PR parse cannot resolve a loop variable, so the
-  gate would fail-open and merge every iteration unverified (observed P3W11 +
-  P3W13). The operator is told to run one literal merge per call so each PR is
-  gate-checked. `--admin` still overrides; a literal `gh pr merge 54` is
-  unaffected. Conservative DECIDE-tier shape (no conditional-allow) per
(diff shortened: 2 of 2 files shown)


parametrization:
Review requested from **Aino Virtanen** (primary) and **Nino Kavtaradze** (secondary) per the charter 2-reviewer gate. Two distinct Approved verdicts required before merge; please post charter-format verdict comments. I will not self-merge.

parametrization:
**Review — PR #891 narrow batch-loop merge guard to in-loop-body co-location** (verified at head `246708bd270f`)

Verified the code at HEAD via the contents API, not the working tree.

The narrowing is correct and preserves the protection that matters. `_loop_body_spans` pairs each `do` with its matching `done` via a stack and returns the character span of every loop body; because `do`/`done` are loop-only keywords in shell, every balanced pair delimits a real loop, and nested loops pair correctly. `is_variable_pr_merge_in_loop` now fires only when a variable-PR `gh pr merge` match actually sits inside one of those spans, evaluated on the same quote-normalized view used for both the merge scan and the span scan, so offsets are consistent and a `--body` payload that merely mentions the shape is still stripped.

The critical property holds: the genuine evasion `for pr in …; do gh pr merge "$pr"; done` is STILL hard-blocked — the merge falls between `do` and `done` — and the regression test asserts exactly that. The false-positive that motivated main#886 is fixed: a non-loop variable merge paired with a separate `gh run rerun "$r" --failed` staleness loop no longer matches, because the merge sits outside the loop-body span. A pure `gh run rerun` loop with no merge short-circuits on the empty merge-match list. A standalone literal `gh pr merge <N>` is untouched (its number parses and the gate runs).

The new tests cover both directions — the two negative `gh run rerun` cases and the positive in-loop block-still. The keyword regex matches `do`/`done` only as delimited standalone tokens, with `done` ordered first so the longer keyword wins. The remaining narrowing is benign: an unterminated `do` without `done` produces no span, but an unclosed loop is not an executable command, so nothing merges. All 13 CI checks are green. No scope creep, zsh/charter conventions honoured.

---

Requestor: Aino Virtanen
Requestee: Lucas Ferreira
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Review — PR #891 narrow batch-loop guard** (verified at head `246708bd270f`)

Reviewed as Security Engineer with the lens: does narrowing the guard weaken the fail-open control on the 2-reviewer gate? It does not.

I reconstructed both the pre-#886 and post-#886 detector and ran them side by side against an evasion battery. The ONLY behavioral delta old→new is the intended one: a block that pairs a non-loop variable merge with a separate, unrelated staleness loop (`gh run rerun "$r" --failed`) flips from blocked to allowed — correct, since the merge never enters the loop body and the gate does not fail-open. Every real evasion shape stays hard-blocked.

Adversarial coverage that still blocks (verified True):
- the canonical loop-variable merge (the documented `feedback_batch_loop_merge_evades` shape);
- `while`/`until` loops, and — newly — `select` loops, which the old `for|while|until` opener regex missed (a net strengthening: keying on the `do…done` body pair covers every shell loop construct);
- newline-delimited bodies, `;do`/`;done` without surrounding spaces, C-style `for (( … ))`;
- nested loops with the merge in either the inner or outer body (stack-based `do…done` pairing is correct);
- one-liner `do … done`.

Negatives stay clean (verified False): standalone variable merge outside any loop, literal-PR merge, a `--body` payload that merely mentions the shape, and the unrelated-rerun-loop case. The `_strip_quoted_runs_keep_var_args` normalization still prevents a quoted prose payload from matching, and `--admin` / literal-PR / standalone-merge paths in `check()` are untouched by this diff (two files only: the hook and its tests). The added tests exercise both directions — negative (rerun loop not blocked, non-loop merge beside an unrelated loop not blocked) and positive (genuine in-loop variable merge still blocked). CI is fully green across all checks.

Two residual evasions exist but are pre-existing — they return not-blocked on BOTH the old and new detector, so this PR neither introduces nor worsens them: a merge wrapped in a subshell inside the loop body (the merge arg reads as `$pr)` and the verb regex terminator set excludes `)`), and a subscripted/command-substitution variable arg such as `"${prs[$i]}"` (stripped by the lone-variable unwrap). Both trace to `_GH_MERGE_VAR_PR` / the quote-unwrap, which this PR does not modify. Tracked for later hardening in #894 — non-blocking here.

Narrowing is sound, the control is preserved, and the change is correct versus main#886. Approving.

---

Requestor: Nino Kavtaradze
Requestee: Lucas Ferreira
RequestOrReplied: Approved
TechDebt: #894

PR #896, opened 2026-06-25
fix(hook-4): validate_pr_review blocks subshell-wrapped & compound-arg in-loop variable merges (closes #894)

### LINKED ISSUE #894: tech-debt: validate_pr_review batch-loop guard misses subshell-wrapped & subscripted in-loop variable merges (pre-existing, surfaced in 891/886)
## Summary

Surfaced while reviewing PR #891 (#886, narrowing the `validate_pr_review` batch-loop guard). These are **pre-existing** residual gaps in `is_variable_pr_merge_in_loop` — they return `False` (not blocked) on BOTH the pre-#886 and post-#886 detector, so #886 neither introduces nor worsens them. Filing for visibility, not as a blocker on #891.

Root cause is in `_GH_MERGE_VAR_PR` / `_strip_quoted_runs_keep_var_args`, both untouched by #886.

## Residual evasions (in-loop variable merge NOT caught)

1. **Subshell-wrapped merge** — a loop body that wraps the merge in `( … )`.
   After quote-normalization the merge arg reads as `$pr)`; `_GH_MERGE_VAR_PR`'s terminator lookahead `(?=\s|;|&|\||$)` does not include `)`, so the merge verb is not matched and the loop-body span is never consulted. Still fail-opens the 2-reviewer gate per iteration.

2. **Compound / subscripted variable arg** — a merge whose PR arg is `"${prs[$i]}"` (or `"$(cmd)"`).
   `_strip_quoted_runs_keep_var_args` only unwraps a double-quoted run that is a lone simple variable (`"$pr"` / `"${pr}"`); a subscripted/compound/command-substitution arg is treated as opaque prose and stripped, so the merge verb's arg disappears from the view and is not matched.

## Suggested direction (defer — DECIDE-tier)

- Add `)` (and possibly `}`) to the `_GH_MERGE_VAR_PR` terminator lookahead.
- Broaden the quote-unwrap (or the merge-arg pattern) to recognize subscripted/compound variable args, OR fall back to flagging any non-literal-PR merge verb located inside a loop body.

Keep the conservative HARD-BLOCK shape (`feedback_safety_direction_over_ux_friction`); these are exotic relative to the canonical loop-variable evasion the guard already blocks, hence non-urgent.

parametrization:
## Summary

Closes the two residual fail-open evasions left in the batch-loop merge guard (`is_variable_pr_merge_in_loop`) after #886 narrowed it to require `do…done` co-location. Both let a `gh pr merge` inside a loop body slip past the matcher, silently disabling the 2-reviewer gate for every iteration.

## Ontology Context

This change is in the **org hook tier** — `.claude/hooks/validate_pr_review.py`, the PreToolUse Bash guard that enforces the 2-reviewer merge gate (Hook 4). Hooks are the durable enforcement tier (`feedback_enforcement_hierarchy`: hook > skill > charter), so a fail-open here is a real bypass of charter review policy. The guard's lineage: #567 introduced the batch-loop block; #886/#891 added the `_LOOP_DO_DONE` + `_loop_body_spans()` co-location refactor (merge must sit inside a loop body) to stop false-blocking unrelated loops. This PR closes the two fail-open gaps that narrowing left open, keeping the conservative HARD-BLOCK shape (`feedback_safety_direction_over_ux_friction`).

## The two evasions

**Gap 1 — subshell-wrapped merge**
```sh
for pr in 1 2; do (gh pr merge $pr); done
```
After quote-normalization the merge argument reads `$pr)`. The old `_GH_MERGE_VAR_PR` matched a full `$var` token terminated by `(?=\s|;|&|\||$)` — that lookahead does not admit the `)` that ends the arg inside a subshell, so the merge verb never matched and the loop fail-opened.

**Gap 2 — subscripted / compound / command-substitution arg**
```sh
for i in 0 1; do gh pr merge "${prs[$i]}"; done    # also "$(get_pr)"
```
`_strip_quoted_runs_keep_var_args` only unwrapped a double-quoted run that fullmatched a LONE simple variable (`$pr` / `${pr}`). A subscripted/compound/command-substitution arg fails that fullmatch, so the whole quoted run was stripped as opaque prose — the real merge argument disappeared and the guard fail-opened.

## Mechanic chosen (and why)

The issue's recommended defensive direction: **inside a loop body, flag any `gh pr merge` whose positional argument is NON-LITERAL** — i.e. not a bare integer and not absent-by-design. This catches `$pr`, `${pr}`, `${prs[$i]}`, `$(cmd)`, and the subshell case in one stroke, while a literal `gh pr merge 54` (PR number parses → gate runs) stays allowed.

- `_GH_MERGE_VAR_PR` → **`_GH_MERGE_NONLITERAL_ARG`**: matches on the FIRST character of the argument rather than a fully-shaped `$var` token. `(?![-\d])` keeps the two literal/absent forms allowed — a leading `-` is a flag (current-branch merge, out of scope for #567) and a leading digit is the bare integer the gate resolves. `[^\s;&|()]` requires a real argument character, so a bare terminator left behind by a stripped quoted run (`gh pr merge ;`) is not mistaken for an argument. Because we match the leading character, there is **no terminator lookahead** — so neither `)` (Gap 1) nor `}` needs special-casing. (A `{ …; }` brace group already ends the arg at `;`, which is why `}` was unnecessary.)
- **`_strip_quoted_runs_keep_var_args`** unwrap predicate broadened to **`_is_single_expansion_word`**: a double-quoted run is unwrapped when it is one shell WORD carrying a `$` expansion with no whitespace OUTSIDE any `${…}`/`$(…)` group. A new `_strip_expansions` helper removes balanced (and nested) expansion groups so internal whitespace like `$(get_pr $i)` does not make the run look like prose. This keeps the real compound args while a `--body "for … do gh pr merge $pr; done"` prose payload (whitespace outside any expansion) is still dropped, and a quoted literal word such as `"done"` (no `$`) is not unwrapped into a stray loop keyword.

## Co-location preserved (#886 regression guard)

The #886 narrowing made a one-off `gh pr merge "$PR"` OUTSIDE any loop — co-occurring with an unrelated `for r in …; do gh run rerun "$r" --failed; done` — NOT block. That property is preserved: the broadened argument match still only fires when the match offset falls inside a `do … done` body span. A non-literal merge has to be INSIDE the loop body to trip. Re-asserted by test.

## Tests added

- subshell-wrapped in-loop variable merge → BLOCKED (was fail-open)
- subscripted `${prs[$i]}` in-loop merge → BLOCKED
- command-substitution `"$(cmd)"` in-loop merge → BLOCKED (plus an inner-whitespace `$(get_pr $i)` robustness case)
- literal `gh pr merge 54` inside a loop → NOT blocked (over-broadening guard)
- the existing #886 unrelated-loop + one-off-merge case → STILL not blocked, re-asserted under the broadened matcher
- end-to-end `check()` HARD-BLOCK tests for the subshell and subscripted shapes
- pure-parser coverage for the new `_strip_expansions` and `_is_single_expansion_word` helpers

## Local ⇄ CI parity evidence

- `ruff@0.15.11 format --check` + `ruff@0.15.11 check` over `.claude/hooks/` — clean
- `mypy --ignore-missing-imports` over both changed files — clean
- full hooks pytest suite — 1593 passed
- `pre_commit_ci_sync.py` drift gate — OK
- pre-commit + pre-push hooks ran green on commit and push (no `--no-verify`)

## Reviewers

This PR is set up for two reviewers — a security/fail-open-safety lens and a parser/regex-correctness lens. The change is verifiable directly against `.claude/hooks/validate_pr_review.py` at this PR's head and the added tests.

### DIFF
--- a/.claude/hooks/tests/test_validate_pr_review.py
+++ b/.claude/hooks/tests/test_validate_pr_review.py
@@ -1645,6 +1645,74 @@ class BatchLoopMergeDetectorTests(unittest.TestCase):
             )
         )
 
+    # ---- #894 residual fail-open evasions ----
+
+    def test_894_subshell_wrapped_merge_blocked(self):
+        # Gap 1: a merge wrapped in a `( … )` subshell leaves the arg as `$pr)`.
+        # The pre-#894 terminator lookahead `(?=\s|;|&|\||$)` omitted `)`, so the
+        # merge verb never matched and the loop fail-opened. Now BLOCKED.
+        self.assertTrue(
+            hook.is_variable_pr_merge_in_loop("for pr in 1 2; do (gh pr merge $pr); done")
+        )
+
+    def test_894_subshell_wrapped_quoted_merge_blocked(self):
+        # Gap 1 variant: subshell + quoted var arg.
+        self.assertTrue(
+            hook.is_variable_pr_merge_in_loop('for pr in 1 2; do (gh pr merge "$pr" --merge); done')
+        )
+
+    def test_894_subscripted_array_arg_blocked(self):
+        # Gap 2: a subscripted `"${prs[$i]}"` arg failed the lone-simple-var
+        # unwrap fullmatch and was stripped as opaque prose — the merge arg
+        # vanished and the loop fail-opened. Now BLOCKED.
+        self.assertTrue(
+            hook.is_variable_pr_merge_in_loop('for i in 0 1; do gh pr merge "${prs[$i]}"; done')
+        )
+
+    def test_894_command_substitution_arg_blocked(self):
+        # Gap 2 variant: a command-substitution `"$(cmd)"` arg, likewise stripped
+        # pre-#894. Now BLOCKED.
+        self.assertTrue(
+            hook.is_variable_pr_merge_in_loop(
+                'for i in 0 1; do gh pr merge "$(get_pr)" --merge; done'
+            )
+        )
+
+    def test_894_command_substitution_arg_with_inner_space_blocked(self):
+        # Gap 2 robustness: whitespace INSIDE the `$(…)` expansion must not make
+        # the quoted run look like prose — it is still one shell-word argument.
+        self.assertTrue(
+            hook.is_variable_pr_merge_in_loop('for i in 0 1; do gh pr merge "$(get_pr $i)"; done')
+        )
+
+    def test_894_literal_merge_inside_loop_not_blocked(self):
+        # Over-broadening guard: a bare-integer `gh pr merge 54` INSIDE a loop is
+        # still NOT blocked — the literal parses and the normal gate runs.
+        self.assertFalse(
+            hook.is_variable_pr_merge_in_loop("for n in 1; do gh pr merge 54 --merge; done")
+        )
+
+    def test_894_nonliteral_merge_outside_loop_not_blocked(self):
+        # #886 co-location preserved under the #894 broadening: a one-off
+        # subscripted/compound merge OUTSIDE any loop, co-occurring with an
+        # unrelated `gh run rerun` loop, does NOT fail-open and MUST NOT block.
+        cmd = (
+            'gh pr merge "${prs[0]}" --repo noorinalabs/noorinalabs-deploy --merge\n'
+            "for r in 11 12 13; do "
+            'gh run rerun "$r" --repo noorinalabs/noorinalabs-deploy --failed; done'
+        )
+        self.assertFalse(hook.is_variable_pr_merge_in_loop(cmd))
+
+    def test_894_body_mention_of_compound_merge_not_blocked(self):
+        # The #894 unwrap broadening must not re-open the prose-mention guard:
+        # a `--body "…"` payload mentioning the compound-arg loop shape (it
+        # carries whitespace OUTSIDE the expansion) is still stripped, not matched.
+        self.assertFalse(
+            hook.is_variable_pr_merge_in_loop(
+                'gh pr create --body "for i in 0 1; do gh pr merge ${prs[$i]}; done"'
+            )
+        )
+
 
 class BatchLoopMergeEndToEndTests(unittest.TestCase):
     """#567 end-to-end: check() HARD BLOCKS a batch-loop variable merge, and a
@@ -1667,6 +1735,27 @@ class BatchLoopMergeEndToEndTests(unittest.TestCase):
         self.assertEqual(result.get("decision"), "block")
         self.assertIn("Batch-loop", result["reason"])
         self.assertIn("one pr per call", result["reason"].lower())
+
+    def test_894_subscripted_loop_merge_hard_blocked(self):
+        # #894 end-to-end: a subscripted in-loop merge reaches check() and HARD
+        # BLOCKS. Patch get_pr_data so a pass proves the LOOP guard fired, not
+        # the downstream gate.
+        with mock.patch.object(hook, "get_pr_data", return_value=None):
+            result = hook.check(
+                self._input('for i in 0 1; do gh pr merge "${prs[$i]}" --merge; done')
+            )
+        self.assertIsNotNone(result)
+        assert result is not None
+        self.assertEqual(result.get("decision"), "block")
+        self.assertIn("Batch-loop", result["reason"])
+
+    def test_894_subshell_loop_merge_hard_blocked(self):
+        # #894 end-to-end: a subshell-wrapped in-loop merge HARD BLOCKS.
+        with mock.patch.object(hook, "get_pr_data", return_value=None):
+            result = hook.check(self._input("for pr in 1 2; do (gh pr merge $pr); done"))
+        self.assertIsNotNone(result)
+        assert result is not None
+        self.assertEqual(result.get("decision"), "block")
 
     def test_loop_var_merge_with_admin_still_overrides(self):
         # --admin is the emergency override and bypasses the loop guard too.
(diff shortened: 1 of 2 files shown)


parametrization:
## Security review — PR #896 (closes #894) — fail-open lens

Reviewed at HEAD `dc206e5320516000e17c14307ced480ebc715703` via the contents API (no parent-tree checkout). CI `statusCheckRollup` is fully green.

**Verdict: Approved.** The two residual fail-open evasions named in #894 are now hard-blocked, and the #886 co-location guarantee is preserved. I verified by running the detector at HEAD against constructed shell strings (not only the bundled tests).

### What I confirmed

Both #894 evasions now HARD-BLOCK (exit-2 with a diagnostic, not allow-with-log):
- Subshell-wrapped — `for pr in 48 49; do (gh pr merge $pr); done` (and the `"$pr"` quoted variant). The matcher keying on the argument's leading char instead of a terminator-bounded `$var` token correctly removes the old `)`-terminator blind spot.
- Compound / subscripted / command-substitution arg — `"${prs[$i]}"`, `"$(get_pr $i)"`, and bare `` `get_pr` ``. The widened unwrap predicate (`_is_single_expansion_word`) keeps these as the real merge argument instead of stripping them as opaque prose.

The original #567 shape (`"$pr"` / `${pr}`), `while read`-driven loops, and brace-expansion args also block, as expected.

### Fail-open / over-broadening checks (all pass)

- HARD BLOCK with a diagnostic preserved — consistent with the safe-direction stance for a guard that cannot auto-resolve.
- A literal `gh pr merge 48` inside a loop still passes (the literal gate resolves it) — no over-broadening.
- The #886 guarantee holds: a one-off variable merge OUTSIDE any loop, co-located with an unrelated `for r in …; do gh run rerun "$r" --failed; done`, does NOT block. No regression of the #886 false-block.
- A `--body "... do gh pr merge $pr ... done"` payload that merely mentions the shape is not matched (quote-normalization strips it).

### One residual (out of scope here — filed, non-blocking)

A bare `gh pr merge` with NO positional argument inside a `do … done` body (current-branch PR) is not caught by the broadened matcher — most concretely `for b in b1 b2; do git checkout $b && gh pr merge; done`, where the gate only checks the cwd PR while the loop sweeps other branches' PRs unverified. This is a distinct vector from #894's argument-form scope and is pre-existing, so it is not a blocker for this PR. Filed as #897 (tech-debt, on the board) with a suggested fix direction and the non-`do/done` `xargs` shape noted alongside.

Nice work, Aino — the leading-char match plus the single-expansion-word unwrap is the right factoring, and the module docstring/comments document the why precisely.

---
Requestor: Nino Kavtaradze
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #897

parametrization:
## Review — PR #896 (parser / regex-correctness lens)

Reviewer #2, verified against HEAD `dc206e5320516000e17c14307ced480ebc715703` via the contents API (no parent-repo checkout). CI `statusCheckRollup` all green. I also loaded the HEAD `validate_pr_review.py` in isolation and ran the HEAD test suite's #894-targeted cases against it (25/25 pass; the only suite failures are roster-path-environmental from running outside the repo tree, and full-suite Pytest is green in CI).

### Regex shape — `_GH_MERGE_NONLITERAL_ARG`
`\bgh\s+pr\s+merge\s+(?![-\d])[^\s;&|()]` is correct for the intent:
- Matching only the **first argument character** (not a fully-shaped token) is the right call — it is spelling- and terminator-agnostic, which is exactly what closes the two #886 evasions. There is no terminator lookahead to omit a closing `)`/`}`, so the subshell `(gh pr merge $pr)` arg `$pr)` now matches. Verified: subshell, quoted-subshell, `${prs[$i]}`, `$(get_pr)`, `$(get_pr $i)` all trip; literal `54` and flag-only forms do not.
- `(?![-\d])` correctly excludes a leading `-` (flag → current-branch merge, out of scope) and a leading digit (bare integer the gate resolves). Confirmed `gh pr merge 54` in a loop is NOT blocked.
- `[^\s;&|()]` matching a single char cannot over-consume — no catastrophic-backtracking surface. The exclusion of bare terminators correctly avoids treating a stripped-run leftover (`gh pr merge ;`) as an argument.

### Quote-normalization — `_strip_expansions` / `_is_single_expansion_word`
- Nested/balanced groups handled with a depth counter over the group's own bracket kind. Verified `$(a $(b) c)` strips fully and `${prs[$i]}` (inner `$i`) strips fully.
- Whitespace-inside-expansion (`$(get_pr $i)`) is correctly treated as one word; whitespace-outside-expansion (`merge ${prs[$i]} now`, a `--body` prose payload) correctly stays prose and is dropped wholesale — the prose-mention guard is intact.
- Unbalanced `${pr` degrades safely: `_strip_expansions` runs `j` to end and exits (no infinite loop, no exception), the word is treated as an expansion → safe direction (block). `_strip_quoted_runs_keep_var_args` breaks on an unbalanced quote. Good.
- A quoted literal word like `"done"` (no `$`) is NOT unwrapped, so it cannot inject a spurious `done` keyword into the loop-span scan. Verified.

### Co-location (#886) intact
Confirmed the match offset must fall inside a `do…done` span: a one-off compound merge outside any loop co-occurring with an unrelated `for…do; gh run rerun "$r"; done` returns False (not blocked), while the real `for pr in …; do (gh pr merge $pr); done` blocks. The broadening did not weaken the co-location requirement.

### Tests
Genuinely exercise the regressions — the subshell and subscripted/command-sub cases assert True where the pre-#894 matcher returned False (terminator-lookahead miss and lone-var-fullmatch strip respectively), so they fail on the old code. The helper-level unit tests (`StripExpansionsTests`, `IsSingleExpansionWordTests`) are behavioral, not tautological (nested-group, inner-space, bare-var-intact, prose-survives, literal-not-a-word). Over-broadening and co-location guards are both covered. End-to-end tests patch `get_pr_data` to `None` so a pass proves the loop guard fired rather than the downstream gate. Type annotations on the new helpers are clean and mypy is green.

### Non-blocking observation (not worth tracking)
A **literal** PR URL inside a loop — e.g. `for u in 1; do gh pr merge https://github.com/o/r/pull/54; done` — now matches (leading `h` is non-literal-looking) and is blocked, even though `extract_pr_number` could resolve `/pull/54`. This is over-strict but in the **safe** direction (blocks rather than fail-opens), is entirely contrived (one would never loop over a single literal URL), and is consistent with the documented conservative DECIDE-tier posture (block any non-bare-integer arg in a loop; no conditional-allow). Not worth a tech-debt issue.

Approving. Clean, well-reasoned parser change with strong test coverage and safe degradation throughout.

---
Requestor: Weronika Zielinska
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

PR #910, opened 2026-07-01
fix(hooks): parse cd-prefixed gh issue edit in PostToolUse parsers (main#901)

### LINKED ISSUE #901: fix(hooks): PostToolUse parsers return skip_parser_returned_empty on cd-prefixed gh issue edit (kickoff comment + wave-field sync silently skip)
## Problem

During P7W20 kickoff, applying a wave label via a Bash command whose first line was `cd "$(git rev-parse --show-toplevel)"` followed by a newline and the `gh issue edit … --add-label "wave-20"` call caused **both** PostToolUse hooks to fail to parse it:

- `post_wave_kickoff_comment` — did not post the kickoff comment (silent skip).
- `post_label_change_wave_field_sync` — logged `skip_parser_returned_empty: command contains 'gh issue edit' and a canonical wave-label flag but parser extracted no changes. Likely an unhandled multi-cmd shape (for-loop with unexpanded vars, etc.); investigate.`

The label-apply itself succeeded; only the two PostToolUse automations silently no-op'd. Re-running the **bare** `gh issue edit …` command (no `cd "$(…)"` prefix) made both hooks fire correctly.

## Impact

Kickoff comments + Wave-field sync silently skip whenever the orchestrator prefixes the label-apply with a `cd "$(command-substitution)"` line — a very common shell idiom in these skills. Caught here only because the orchestrator verified the comment posted; otherwise it surfaces as silent drift at the next wave-wrapup.

## Likely fix

Harden the shared `_shell_parse` tokenizer (or the two hooks' segment extraction) so a leading `cd "$(…)"` command substitution + newline-joined segments does not cause the gh-subcommand extractor to return empty. The `gh issue edit` segment is the second command segment; the parser should iterate command segments rather than bailing when the first token is a `cd $(…)` form.

## Repro

Run a single Bash call containing two lines: `cd "$(git rev-parse --show-toplevel)"` then `gh issue edit <N> --repo noorinalabs/<repo> --add-label "wave-20"`. Observe no kickoff comment + the `skip_parser_returned_empty` annunaki entry.

parametrization:
## Summary

Fixes **main#901**: the two wave-label PostToolUse hooks —
`post_wave_kickoff_comment` and `post_label_change_wave_field_sync` — silently
skipped (`skip_parser_returned_empty`) whenever the label-apply command was
prefixed with a `cd ...` line, e.g. the P7W20-kickoff shape:

```
cd "$(git rev-parse --show-toplevel)"
gh issue edit 901 --repo noorinalabs/noorinalabs-main --add-label "wave-20"
```

## Root cause

`shlex.split` treats a NEWLINE as ordinary whitespace, and only tokenizes
`;`/`&&`/`||`/`|` as standalone separators when they are already
whitespace-padded. So a newline-joined (or non-space `;`-joined) `cd` + `gh`
collapsed into ONE command segment whose first token was `cd`;
`find_gh_subcommand` requires the segment to start with `gh`, bailed, and the
parser returned an empty list. The label-apply itself succeeded — only the two
automations no-op'd (kickoff comment not posted, Wave field unsynced).

## Fix (shared path, once)

- Add quote/escape-aware `normalize_command_separators()` to `_shell_parse.py`:
  rewrites unquoted newlines to ` ; ` and space-pads unquoted
  `;`/`&&`/`||`/`|` so shlex emits them as separator tokens. Operators and
  newlines **inside** quoted args (e.g. `--body "x && y; z"`, multi-line
  bodies) and backslash-escaped chars are left byte-for-byte untouched.
- `_wave_label_parse.py` applies it (after `strip_heredocs`, before
  `tokenize`) in **both** `parse_wave_label_changes` and
  `parse_wave_label_create` — so all three affected surfaces (kickoff comment,
  EDIT field-sync, CREATE field-sync) are fixed by one shared change.
- The global `tokenize` is deliberately left unchanged (15+ hooks, incl.
  security-critical commit-identity/git-config, depend on it); the fix is
  scoped to the wave-label parser path that #901 traverses.

Un-prefixed commands parse identically to before.

## Tests

New `CdPrefixedCommand` and `NormalizeCommandSeparators` classes in
`test__wave_label_parse.py`. The newline-`cd`-substitution and non-spaced-`;`
cases **fail pre-fix** (verified by neutralizing the normalizer: 4/7 fail) and
pass post-fix. Includes a regression guard for un-prefixed commands and for
operators inside quoted args.

## Verification (local ⇄ CI parity)

- `pytest .claude/hooks/tests/ .claude/lib/tests/` — 2269 passed
- `ruff@0.15.11 format --check` + `ruff@0.15.11 check` — clean
- `mypy --ignore-missing-imports` on changed files — clean
- `pre_commit_ci_sync.py .` — OK (mirror intact)

## Reviewers

- **Wanjiku Mwangi** (peer)
- **Santiago Ferreira** (secondary)

Closes #901

### DIFF
--- a/.claude/hooks/_shell_parse.py
+++ b/.claude/hooks/_shell_parse.py
@@ -253,6 +253,90 @@ def tokenize(cmd: str) -> list[str] | None:
         return None
 
 
+def normalize_command_separators(cmd: str) -> str:
+    """Quote/escape-aware normalization of shell command separators.
+
+    shlex.split treats a NEWLINE as ordinary whitespace and only recognizes
+    `;`/`&&`/`||`/`|` as standalone tokens when they are ALREADY surrounded by
+    whitespace. So two very common orchestrator idioms defeat
+    `iter_command_segments`, which relies on those separators surviving
+    tokenization as their own tokens:
+
+        cd "$(git rev-parse --show-toplevel)"\\n gh issue edit N --add-label ...
+        cd /some/dir; gh issue edit N --add-label ...
+
+    In the first the newline vanishes into whitespace, so `cd`, the command
+    substitution, and `gh` collapse into ONE segment whose first token is
+    `cd` — `find_gh_subcommand` bails and the parser returns empty (issue
+    #901: kickoff-comment + Wave-field-sync PostToolUse hooks silently skip).
+    In the second the `;` sticks to `/some/dir` (`/some/dir;`), so again no
+    separator token is produced.
+
+    This helper rewrites the raw command so shlex WILL emit the separators as
+    standalone tokens:
+      - line-continuation `\\<newline>` -> a single space (a continued command
+        is ONE command, not two — done first so a continuation newline is not
+        misread as a command separator);
+      - an unquoted, unescaped NEWLINE -> ` ; ` (newline is a command
+        terminator in shell, equivalent to `;`);
+      - an unquoted, unescaped `;` / `|` / `&&` / `||` -> the same operator
+        space-padded on both sides.
+
+    Quote/escape awareness is essential: a `|`/`;`/`&&`/newline INSIDE a single
+    or double quoted string (e.g. `--body "x && y"`, a multi-line
+    `--body "line1\\nline2"`) or preceded by a backslash is DATA, not a
+    separator, and is left byte-for-byte untouched. The result is only ever
+    fed back through `tokenize` (which re-applies line-continuation
+    normalization harmlessly), so injecting extra spaces around genuine
+    operators cannot change the parsed argument values.
+    """
+    cmd = _LINE_CONTINUATION_RE.sub(" ", cmd)
+    out: list[str] = []
+    i = 0
+    n = len(cmd)
+    quote: str | None = None  # active quote char: "'" or '"', else None
+    while i < n:
+        c = cmd[i]
+        if quote is not None:
+            out.append(c)
+            # Inside double quotes a backslash escapes the next char; inside
+            # single quotes nothing is special (shell single-quote semantics).
+            if c == "\\" and quote == '"' and i + 1 < n:
+                out.append(cmd[i + 1])
+                i += 2
+                continue
+            if c == quote:
+                quote = None
+            i += 1
+            continue
+        if c in ("'", '"'):
+            quote = c
+            out.append(c)
+            i += 1
+            continue
+        if c == "\\" and i + 1 < n:
+            # Escaped char outside quotes — emit both, do not treat as separator.
+            out.append(c)
+            out.append(cmd[i + 1])
+            i += 2
+            continue
+        if c == "\n":
+            out.append(" ; ")
+            i += 1
+            continue
+        if cmd[i : i + 2] in ("&&", "||"):
+            out.append(" " + cmd[i : i + 2] + " ")
+            i += 2
+            continue
+        if c in (";", "|"):
+            out.append(" " + c + " ")
+            i += 1
+            continue
+        out.append(c)
+        i += 1
+    return "".join(out)
+
+
 def strip_heredocs(cmd: str) -> str:
     """Remove all heredoc bodies. Iterates until no more matches (handles nested)."""
     prev = None
--- a/.claude/hooks/_wave_label_parse.py
+++ b/.claude/hooks/_wave_label_parse.py
@@ -121,6 +121,7 @@ sys.path.insert(0, str(Path(__file__).resolve().parent))
 from _shell_parse import (  # noqa: E402
     find_gh_subcommand,
     iter_command_segments,
+    normalize_command_separators,
     strip_heredocs,
     tokenize,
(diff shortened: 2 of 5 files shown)


parametrization:
**Requestor:** Wanjiku Mwangi
**Requestee:** Aino Virtanen

Reviewed at HEAD `663deba154efc3b4a0b7f963a1eb4fd395cebe20` via `gh api contents` (no checkout) + `gh pr diff`. Traced the real parser code, not the PR framing.

**Correctness — the fix lives where it should.** `iter_command_segments` splits a shlex token list on `;`/`&&`/`||`/`|` only when each survives tokenization as its OWN token. shlex.split (posix, no `punctuation_chars`) makes that contingent on whitespace-padding: `a ; b` yields a standalone `;`, but `a;b` and a bare NEWLINE do not. That is precisely the #901 gap — a `cd "$(...)"`+newline or a `cd /dir;` prefix collapses `cd` and `gh` into one segment whose first token is `cd`, so `find_gh_subcommand` (returns None unless `segment[0] == "gh"`) bails and the parser returns empty. `normalize_command_separators` rewrites unquoted separators to their space-padded form so shlex emits them as tokens; the subsequent `iter_command_segments` then peels the `cd` segment off and the `gh issue edit/create` segment parses. Verified end-to-end against the newline-`cd`-subst repro.

**Separator handling is quote/escape-correct.** Char-scan tracks quote state: single quotes treat `\` as literal (correct shell semantics), double quotes honor `\`-escape (so `\"` doesn't false-close), and `&&`/`||` are matched before the single `;`/`|` branch so a two-char operator is never mis-split into two pipes. A `;`/`|`/`&&`/newline inside `"..."`, `'...'`, or an unbroken `"$(...)"` is left byte-for-byte — confirmed against the adversarial `--body "run x && y; then z | w"` case (parses as one edit, no spurious segment) and `echo "a\nb"` (single token). Un-prefixed commands are untouched (redundant space-padding on already-padded operators collapses in shlex → identical tokenization); the `test_bare_command_unchanged` guard holds.

**Ordering is right and load-bearing.** `normalize_command_separators(strip_heredocs(command))` — heredoc stripping must run first because `_HEREDOC_RE` needs the raw newlines to find the body; normalizing first would turn body newlines into ` ; ` and defeat the heredoc match. The double application of `_LINE_CONTINUATION_RE` (here + in `tokenize`) is idempotent and harmless.

**Pre-fix failure claim verified (4/7).** Reasoning through the pre-fix tokenizer: the two newline-`cd` cases, the singular newline case, and the non-space-padded `cd /dir; gh` case all collapse to a single `cd`-first segment → empty parse → FAIL pre-fix. The `&&`-prefixed, bare, and quoted-body cases already passed pre-fix because `&&` was space-padded and quoted operators were always preserved. That is exactly 4 regressions in the `CdPrefixedCommand` class, matching the stated 4/7; the `NormalizeCommandSeparators` class adds direct unit coverage of the new helper.

**CI:** all 14 checks SUCCESS (ruff lint+format, mypy, pytest, sync-drift gate, commit-identity, smoke, budgets). `statusCheckRollup` clean, `mergeable: MERGEABLE`.

Correct, minimal, adversarially-tested, no regression surface. Ship it.

TechDebt: none

parametrization:
**Requestor:** Santiago Ferreira
**Requestee:** Aino Virtanen

Re-reviewed at HEAD `7a733d2` (read via `gh api contents?ref=7a733d2`, no checkout). My in-scope Changes-Requested item — the fourth same-family surface `validate_wave_label_evidence.py` failing OPEN on the `cd`-prefixed `gh issue edit|create --add-label wave-*` idiom — is fully resolved.

Verified at HEAD:
- `_find_issue_command` now routes `tokenize(normalize_command_separators(strip_heredocs(command)))` — identical wiring/ordering to `_wave_label_parse.py` (heredoc-strip first so its regex sees raw newlines, normalize second, tokenize third), with the shared helper imported from `_shell_parse`. No duplicated logic; the fix reuses the same helper this PR introduced.
- New `CdPrefixedCommandTests` cover the three shapes (newline-edit, non-space `;`-edit, newline-create). Each cites an all-404 path and asserts `decision == "block"` — proving the gate now REACHES cited-path verification instead of being silently skipped. Pre-fix each returned None (fail-open); the assertion messages state that explicitly.
- CI is 14/14 SUCCESS at this head and the PR is MERGEABLE (pytest, ruff format+lint, mypy, sync-drift gate, commit-identity all green).

The wave-label command family is now consistent across all four surfaces (kickoff comment, EDIT field-sync, CREATE field-sync, and the PreToolUse evidence gate). The `block_no_verify` / `block_git_config` git-family generalization is correctly carried as a separate follow-up (main#911), out of #901 scope.

TechDebt: main#911 — generalize `normalize_command_separators` to the git-family guards `block_no_verify` + `block_git_config` (cd+newline fail-open on git command-position); tracked, out of #901 scope.

PR #912, opened 2026-07-01
fix(hooks): block no-argument current-branch merge inside a loop (#897)

### LINKED ISSUE #897: tech-debt: validate_pr_review loop guard misses no-arg (current-branch) gh pr merge inside a do…done body
## Summary

PR #896 (closes #894) closes the two residual fail-open evasions of the `validate_pr_review` batch-loop merge guard (subshell-wrapped `(gh pr merge $pr)` and compound/subscripted args `"${prs[$i]}"` / `"$(get_pr)"`). Both now HARD-BLOCK, and the #886 co-location guarantee is preserved. Verified against the detector at PR HEAD.

While reviewing it I found a **distinct, pre-existing residual** that #894/#896 does not (and was not scoped to) address: a **no-positional-argument** `gh pr merge` inside a `do … done` loop body.

## The gap

`is_variable_pr_merge_in_loop` only fires when `gh pr merge` is followed by a **non-literal positional argument** (`_GH_MERGE_NONLITERAL_ARG` requires `\s+` then a non-terminator, non-flag, non-digit char). A bare `gh pr merge` with **no positional** (current-branch PR) never matches that regex, so the loop guard returns `False`.

It then falls through to `check()`, where `is_merge_command` (segment-split on `&&`/`;`/`|`) DOES match the `gh pr merge` segment, so the gate runs — but with `pr_number = None`, so `get_pr_data(None)` resolves the **cwd branch's PR**, not the per-iteration branch.

The fail-open shape:

```
for b in branch-a branch-b; do git checkout $b && gh pr merge; done
```

The hook gate-checks only the cwd PR once (PreToolUse runs before the loop executes), while the loop sweeps `branch-a` / `branch-b`'s PRs unverified. This is the same threat class as the original #567 (batch unverified merges), via the no-arg current-branch form rather than a loop-variable argument.

## Why non-blocking for #896

#894's scope is explicitly the **non-literal argument** forms; the no-arg/current-branch shape is a separate vector and is pre-existing (not regressed by this PR). #896 is correct and complete for its stated scope.

## Suggested fix direction (for the implementer to evaluate)

Extend the in-loop detection so a `gh pr merge` token that appears inside a `do … done` body with **no resolvable literal PR number** also hard-blocks — i.e. co-located `gh pr merge` whose positional is absent OR non-literal, not only non-literal. Guard against the #886 false-positive (a bare `gh pr merge` OUTSIDE any loop must still pass). Add test cases for `for b in …; do git checkout $b && gh pr merge; done` and a bare `gh pr merge` inside an unrelated `do … done`.

Out of scope here but worth noting alongside: non-`do/done` batch shapes (`xargs -I{} gh pr merge {}`) are also uncaught by the co-location model — a larger design question, not necessarily in this issue.

Surfaced during security review of PR #896 (reviewer: Nino Kavtaradze).

parametrization:
## Summary

Closes #897.

Residual fail-open in the batch-loop merge guard (`validate_pr_review.py`, threat class #567) that PR#896 (#894) did not scope: a **no-positional-argument** current-branch merge inside a loop body was not caught.

```
for b in branch-a branch-b; do git checkout $b && <gh pr merge>; done
```

`is_variable_pr_merge_in_loop` only fired on a NON-LITERAL positional (`_GH_MERGE_NONLITERAL_ARG` required a present non-flag/non-digit argument character). A bare merge has no positional, so the loop guard returned False, the command fell through to `check()`, and the gate ran once with `pr_number=None` — resolving the **cwd branch's** PR, not the per-iteration branch. PreToolUse runs once before the loop, so the loop then swept other branches' PRs unverified.

## Fix

Generalized the single existing matcher (no parallel parser — built on the existing detector per hook-brief convention):

- Replaced the "first non-flag/non-digit argument character" pattern with a single negative lookahead `\bgh\s+pr\s+merge\b(?!\s+\d)` (renamed `_GH_MERGE_NO_LITERAL_PR`).
- It fires unless the positional is a bare-integer literal — covering a non-literal positional (everything #894 closed) **and** an absent positional (#897).
- The unchanged `do…done` co-location requirement means a bare current-branch merge **outside** any loop still passes — the #886 false-positive guard holds.

## Invariants preserved

- **#886**: bare / non-literal merge OUTSIDE a loop -> PASS (verified end-to-end: reaches the normal 2-reviewer gate).
- **#894**: subshell `(… $pr)`, subscripted `${prs[$i]}`, command-substitution `$(get_pr)` in-loop -> still BLOCK.
- Bare-integer literal in-loop (`… 54`) -> still PASS (gate resolves & runs).
- `--admin` still overrides.
- Deferred out of scope (per issue): the `xargs -I{} … {}` shape.

## Tests

Added to `BatchLoopMergeDetectorTests` + `BatchLoopMergeEndToEndTests`:
- no-arg-in-checkout-loop -> BLOCKED (the #897 shape)
- bare-merge-in-unrelated-loop -> BLOCKED
- flags-only-in-loop -> BLOCKED
- bare-merge-outside-loop -> PASS (#886)
- bare-merge + unrelated `gh run rerun` loop -> PASS
- end-to-end `check()` block + outside-loop pass-through

Full suite: **1634 passed**. ruff format/lint, mypy, sync-drift gate all green. Structural index regenerated.

## Reviewers

- **Aino Virtanen** (peer -- Standards & Quality Lead, hook expert)
- **Nadia Khoury** (secondary)

https://claude.ai/code/session_01SvQH5QhX7z6dAVM1rtTAXJ


parametrization:
## Secondary review — PR #912 (process / scope-discipline angle)

Reviewed the hook source at HEAD (`3125837`) via `gh api contents`, distinct from the peer hooks-expert angle. Focus: scope discipline, the #886 everyday-command regression risk, minimality/no-parallel-parser, PR-body accuracy, and the deliberately-deferred xargs shape.

**(a) Scope discipline — clean.** The source delta is confined to `.claude/hooks/validate_pr_review.py`: the single matcher is renamed and its body swapped from the two-part `(?![-\d])[^\s;&|()]` to one negative lookahead `\bgh\s+pr\s+merge\b(?!\s+\d)`, plus docstring/comment/block-message text and the co-located tests. No unrelated behavior touched; the loop-body detection, co-location requirement, `--admin` short-circuit, roster/2-reviewer path, and TechDebt path are all untouched. The `ontology/structural/*` changes in the diff are a wholesale regeneration of a generated artifact catching up to already-landed wave-branch source (e.g. `_shell_parse.py` line shifts) — expected, not a scope leak.

**(b) #886 regression walk — holds.** I traced the everyday command a teammate runs at their own PR: a bare `gh pr merge` (or `gh pr merge --merge --repo o/r`) OUTSIDE any loop. The new lookahead DOES match it (verb at end / followed by a flag, not `\s+\d`), but `_loop_body_spans` finds no `do…done` span, so `is_variable_pr_merge_in_loop` returns False and the command flows to the normal 2-reviewer gate. So this broadening does NOT newly block the legitimate current-branch merge. End-to-end test `test_897_bare_merge_outside_loop_reaches_gate` proves the same with two approvers reaching allow. Literal `gh pr merge 912 …` is likewise unaffected (`\s+\d` fails the lookahead).

**(c) Minimal, no parallel parser — yes.** The fix generalizes the ONE existing detector rather than adding a second one; the resulting regex is simpler than what it replaced. This is exactly the no-parallel-parser direction the issue asked for.

**(d) PR body accuracy — accurate.** Body claims Closes #897 (correct), and the listed invariants (#886 pass-outside-loop, #894 non-literal forms still block, bare-integer passes, `--admin` overrides, xargs deferred) all match the shipped code.

**(e) xargs left out of scope — correctly, not half-done.** `xargs -I{} gh pr merge {}` has no `do…done`, so `_loop_body_spans` is empty and it is not caught — the issue explicitly defers this non-`do/done` shape as a separate design question, and the PR states so rather than silently half-implementing it.

One non-blocking observation (no follow-up needed): an in-loop literal-URL merge (`…/pull/5`) is blocked by the loop guard even though `extract_pr_number` could resolve it via `/pull/` — but this is pre-existing (the prior matcher blocked it identically) and not introduced here.

Merge-ready. All 21 CI checks green; local⇄CI parity intact.

---

**Requestor:** Nadia Khoury
**Requestee:** Wanjiku Mwangi
**TechDebt:** none

parametrization:
## Review — PR #912 (validate_pr_review no-positional in-loop merge guard, main#897)

Reviewed at head `3125837365fbe4d12d37fb8872087da38e8d2a37` via `gh api contents?ref=…` (no checkout). I am the hooks author for this file; verified the regex, the co-location detector, and the 8 new tests against the actual code at HEAD.

**Delivered scope (#897) is airtight.** The generalization `_GH_MERGE_NONLITERAL_ARG → _GH_MERGE_NO_LITERAL_PR` with `\bgh\s+pr\s+merge\b(?!\s+\d)` is correct and, importantly, stays consistent with `extract_pr_number`'s `\bgh\s+pr\s+merge\s+(\d+)` — the guard fires on exactly the set the literal parser cannot resolve. Checked each focus item:

- (a) `for b in a b; do git checkout $b && gh pr merge; done` now BLOCKS (absent positional → not followed by a digit → matches; sits in the `do … done` span). All #894 forms remain blocked — subshell `(gh pr merge $pr)`, unwrapped `${prs[$i]}`, `$(get_pr)` — none is followed by `\s+\d`.
- (b) #886 holds. A bare `gh pr merge` OUTSIDE any loop is not flagged (co-location requirement unchanged) and flows to the normal 2-reviewer gate — proven by the detector `assertFalse` cases and the end-to-end `gh pr merge --merge` test that ALLOWS with two approvers.
- (c) In-loop literal `gh pr merge 123` is exempt (lookahead sees `\s+\d`), and `gh pr merge 123 --auto` (digit-then-flag) also passes. Correct.
- (d) Edge sweep: end-of-string / trailing-whitespace bare merge → flagged (blocks only inside a loop). `gh pr merge --repo x 123` (flag-before-number) → flagged AND consistent, because `extract_pr_number` also returns None for that form, so blocking a loop carrying it is right, not a false-positive.
- (e) Tests genuinely cover both directions: 6 detector cases (block + pass) plus 2 end-to-end. The block e2e patches `get_pr_data → None` so a pass can only mean the LOOP guard fired (not the cwd fallback) — good design.

**One residual, pre-existing blind spot (non-blocking).** An fd-redirect immediately after `merge` with no positional evades the guard: `for b in a b; do git checkout $b && gh pr merge 2>/dev/null; done`. Here `(?!\s+\d)` reads the redirect fd `2` as a bare-integer PR literal, so the loop shape is NOT flagged; and `extract_pr_number`'s `\s+(\d+)` independently mis-parses the same `2`, gating PR #2 while the loop merges each checked-out branch's own PR. I confirmed the OLD matcher `(?![-\d])[^\s;&|()]` had the identical blind spot — so this is pre-existing, shared with `extract_pr_number`, and NOT a regression this PR introduces or claims to close. A correct fix must tighten BOTH sites together (treat the digit run as a complete token, e.g. require a trailing `[\s;&|)]|$`, not `>`/`<`) so guard and parser stay consistent; doing it in the guard alone would leave the non-loop `extract_pr_number` mis-gate. That is proper follow-up work, out of #897's scope — recording it below rather than holding a correct, well-tested, in-scope security improvement.

Trivial nit (no action needed): the reflowed module docstring left a dangling "The" at a wrap boundary near the P3W11/P3W13 note.

Approving on delivered scope; the redirect-fd edge is tracked as tech debt.

**Requestor:** Aino Virtanen
**Requestee:** Wanjiku Mwangi
**TechDebt:** Pre-existing fd-redirect blind spot — a no-positional `gh pr merge 2>/dev/null` (or `2>&1`/`1>`) reads the redirect fd as a PR literal in BOTH `_GH_MERGE_NO_LITERAL_PR`'s `(?!\s+\d)` and `extract_pr_number`'s `\s+(\d+)`; follow-up to tighten both to a complete-token digit match.

Issue #928, opened 2026-07-09
meta: fix every defect found en route to stg, then re-run the pipeline from `parse`


parametrization:
Owner directive (2026-07-09), verbatim:

> "would it be sensible to handle all issues identified, determine which is the earliest in the processing flow, and re-start the process from that point when done? I am unbothered by a timesink or token usage complication, the more correct and repeatable the better right now."

and:

> "fix all the defects we've discovered trying to get this to stg, and restart the pipeline from 'the point of the earliest defect'"

and (2026-07-09, on the gate below):

> "The gate list should be dynamic, add issues to it when you find more."

This is the tracking issue for that program.

---

# ✅ LAUNCH GATE — CLEAR for the RE-RUN as of 2026-07-11 07:40 UTC. **da#428 MERGED (da#429 → `bd133e60`).**

> ## ✅ The RE-RUN from `parse` is UNBLOCKED. The PROD PRUNE is not.
>
> **da#428 is MERGED** (da#429, `bd133e60`) — the producer half of the completeness binding. `resolve` now writes `curated/_resolve_run.txt` declaring the canonical tally **from an in-memory count taken before the write**, and `publish_parquet.py` writes `_manifest.txt` with **md5 + bytes only** and **refuses to publish** a run that did not reach its terminal stage. **A re-run from `parse` will now produce a prunable artifact.**
>
> ### These two gates are DIFFERENT and must not be conflated
>
> | gate | status |
> |---|---|
> | **Re-run from `parse`** (7.5h, writes a new artifact) | **✅ CLEAR** |
> | **Prod prune** (deletes from the graph, **no undo**) | **⛔ BLOCKED** |
>
> **The prune remains blocked on:**
> - **deploy#580** — the wrong-ref prune. A *genuine but older/smaller* run's ref passes **every** gate (valid manifest, matching md5, completeness green, `missing ≈ 0`) and deletes every narrator loaded since. `missing` is **structurally blind**: it bounds `canonical − graph`, none of which can be deleted, while the over-deletion lives entirely in `graph − canonical`. **No parquet-derived gate can see it.** Only a graph-side load-provenance stamp can.
> - **Backups, which have STILL NEVER RUN — but the reason changed on 2026-07-11. Re-derived, not recited.**
[...]

Issue #309, opened 2026-05-08
Org-wide GitHub Actions Node 20 deprecation — upgrade audit (5 repos)


parametrization:
## Context

GitHub Actions deprecation: Node.js 20 actions are deprecated.

> Actions will be forced to run with Node.js 24 by default starting **June 2nd, 2026**. Node.js 20 will be **removed from the runner on September 16th, 2026**.
>
> See the [GitHub changelog](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/) for details.

After the removal date, any workflow still pinned to a Node 20 action version will fail.

Surfaced in CI on 2026-05-08 via the `shellcheck infra & scripts` job for noorinalabs-deploy PR #278:

> `##[warning]Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v4.`

## Org-wide surface (audited 2026-05-08)

| Repo | Node 20 actions in use | Status |
|------|---------------------|--------|
| noorinalabs-main | (no `.github/workflows`) | n/a |
| noorinalabs-isnad-graph | already on `checkout@v6`, `upload-artifact@v7`, `github-script@v8`, `cache@v5` | ✅ reference target |
| noorinalabs-user-service | `actions/checkout@v4` | needs upgrade |
| noorinalabs-deploy | `actions/checkout@v4`, `actions/github-script@v7`, `actions/upload-artifact@v4` | needs upgrade |
| noorinalabs-design-system | `actions/checkout@v4`, `actions/setup-node@v4`, `actions/upload-pages-artifact@v3`, `actions/deploy-pages@v4` | needs upgrade |
| noorinalabs-data-acquisition | `actions/checkout@v4` | needs upgrade |
| noorinalabs-isnad-ingest-platform | (no `.github/workflows`) | n/a |
| noorinalabs-landing-page | `actions/checkout@v4`, `actions/setup-node@v4`, `actions/upload-artifact@v4` | needs upgrade |

**Reference state:** `noorinalabs-isnad-graph` already runs current major versions and produces no Node 20 deprecation warnings in CI runs. Use as the upgrade target — match its action versions where they overlap.

## Scope

5 child repos need upgrades. Per-repo tracking issues will be filed and cross-linked here.

## Acceptance

- [ ] Per-repo upgrade issues filed and linked
[...]

PR #282, opened 2026-05-06
promotion(P3W5 retro): codify 5 AUTO memories → charter (3 new sections + 2 provenance markers)

parametrization:
## Summary

Codifies the **5 AUTO promotions** surfaced by the P3W5 `/promotion-audit` (see audit log at `.claude/team/promotion_audit_log/wave-5.md` and the running retro PR #281).

This is a **separate PR from the retro** because charter-tier promotions are substantive content additions deserving independent review focus from the retro narrative changes.

## Audit context

P3W4 audit was `0 AUTO / 0 DECIDE` — every memory had `promotion_target: none`. PR #277 (P3W5 T2) systematically classified all 36 feedback memories with `promotion_target` frontmatter, flipping the next audit to surface real candidates. The W5 audit run confirmed `5 AUTO` exactly as predicted in #277's PR body.

## Surprise during execution: parser gap

While preparing the 5 charter sections, 2 of the 5 candidates turned out to **already be in charter**:

- `feedback_review_against_artifact_not_framing` — already at `pull-requests.md` § Trust the Artifact, Not the Framing (L325)
- `feedback_honest_audit_over_conclusion_claim` — already at `skills.md` § Wave Lifecycle — Open-Item Audit (also Hook 17)

The audit's `find_already_promoted()` parser only scans `hooks.md` for `Promotion provenance:` blocks. Charter-tier-only promotions (memory → charter without a corresponding hook) are invisible to the parser. **Follow-up issue queued** (see Action Items).

### 3 new charter sections (real promotions)

| Memory | Citations | Charter destination |
|---|---|---|
| `feedback_canonical_source_via_git_show.md` | 4 | `state-claims.md` § Canonical Source via `git show <sha>:<path>` |
| `feedback_child_repo_implementer_rule.md` | 4 | `agents.md` § Child-Repo Implementer Rule + Spawn-Brief Verification |
| `feedback_security_guard_inline_not_followup.md` | 4 | `pull-requests.md` § Security Guards Belong Inline, Not in a Followup |

Each new section preserves the memory's load-bearing content (the rule, the why, the worked example, the severity ladder) plus an HTML-comment provenance marker for parser visibility on future audits.

### 2 provenance markers (already-in-charter, parser miss)

| Memory | Existing charter section | Action |
|---|---|---|
| `feedback_review_against_artifact_not_framing.md` | `pull-requests.md` § Trust the Artifact, Not the Framing | Added `Promoted from memory:` comment marker above the existing section |
[...]

PR #930, opened 2026-07-09
memory: dynamic launch gate + quality-risk axis; control-group discipline

parametrization:
Part of noorinalabs/noorinalabs-main#928.

Memory-only change. **Three** existing memory files amended, plus their `MEMORY.md` index lines; **no new memory entries** (the corpus is at 120/120 index entries, so every fact was folded into the memory that already owns its topic).

**One of the three is a memory *reversal*, and it is called out here rather than left to be discovered:** `project_hadith_id_double_prefix.md` previously asserted that `sanadset:sanadset:` was a legitimate `corpus:collection` pair. It is not. `collection == corpus` occurs on 650,986 of 650,986 sanadset rows and on zero rows elsewhere — a CSV-stem fallback (da#353) — and `lk:lk:1` exists nowhere. da#359's raise is correct. The always-loaded index currently teaches the old claim; this PR corrects both the file and its index line.

## 1. `feedback_sweep_expensive_stage_before_launch.md` — the gate is dynamic, and it has a second axis

Owner, 2026-07-09:

> "The gate list should be dynamic, add issues to it when you find more."

A gate enumerated once at filing time silently decays into a snapshot. Records the amendment rules:

- a new at-or-upstream defect joins the gate **on discovery**, before triage/assignment/costing;
- a row leaves only when **merged** — not when fixed, approved, or understood;
- anyone may add; only the orchestrator removes;
- **if unsure whether it is upstream, add it** — a false positive costs one line, a false negative costs the whole 7.5-hour stage;
- it lives in the **issue body**, never a comment. Comments scroll, and the newest one reads as authoritative even when stale.

Also records the **second axis**, which decides gate membership where stage position alone cannot. Two upstream fixes can carry opposite quality risk. A fix that only *reads* data already segmented on disk infers nothing and cannot regress correctness (da#368's 6,143 chains in an unread column; `tusi`'s 17,089 already-separated isnads). A fix that *extracts* structure from free text can — and its failure mode is exactly the narrator pollution the re-run exists to remove (da#366's matn-embedded chains, ≈122,013 and a lower bound).

> The re-run must produce a graph that is **strictly more correct**. Zero-inference completeness rides along. Extraction-based completeness waits for a validated splitter — and costs its own re-run, which it would have cost regardless.

## 2. `feedback_silent_zero_is_not_a_measurement.md` — the inverse: a probe that cannot return LOW

The existing memory covers probes that cannot return **nonzero**. Today produced the mirror image, and it nearly shipped a wrong number.

A detector for "does this hadith's matn contain an isnad?", keyed on the Arabic tokens `عن` / `قال`, reported **97.2%** of the target set positive — implying ~160k recoverable chains. But those tokens saturate ordinary Arabic prose. Measured against the **control group** (rows that demonstrably carry their chain elsewhere, so their matn is chain-*free* by construction):

| probe on matn head | control (has `<SANAD>`) | test (`No Sanad`) | lift |
|---|---|---|---|
| loose token (`عن` / `قال`) | **85.6%** | 88.5% | ~1.0× |
| strict opener (`حدثنا` / `أخبرنا` / `سمعت`) | 7.9% | **79.2%** | **10×** |

The loose probe measured nothing; its answer was fixed before it saw the data. Re-keying on strict isnad openers gave a tenfold lift — **and a number that was still wrong.** Its recall on the *positive* class was `0 / 485,285`, because the source strips the transmission opener from every row it successfully tags. `126,400` is withdrawn, along with `≥106,088` and `~33,160 chainless`. The live figure is **≈122,013**, a lower bound (da#366, main#928).

That is the whole point of the amendment: a clean control group and a 10× lift were **not enough**. The rule is now *run both classes and require the detector to separate them* — credited to the reviewer who falsified my version of it.

**A high number is exactly as suspect as a zero. Never accept a rate without the rate on the negative class.**

This is another member of the silent-fallback family tracked in #928 (see that issue for the live tally — a count duplicated here would rot on the next amendment), and the third distinct *shape*: a probe that cannot return nonzero · a guard that cannot go red · **a detector whose positive class is defined by a token that saturates the negative class**.

## Verification

- `python3 .claude/lib/memory_budget.py` → OK (120/120 entries, 119/120 files, 24,455/24,576 bytes).
- Full pre-push suite green: **2,295 passed, 90 subtests passed**.
- No `--no-verify`. The first push attempt was **rejected**; the rejection is worth recording. `pre-commit` stashed the working tree, the annunaki `PostToolUse` monitor appended to `.claude/annunaki/errors.jsonl` while the stash was held, and pre-commit attributed the mutation to the longest-running hook (`pytest`) with *"files were modified by this hook."* `pytest` was proven clean against **both** a dirty and a stashed-clean tree via full-tree `md5sum` digests before the retry. Worth noting that the first probe I wrote for this — diffing `git status --porcelain` lines — was **incapable of detecting a further write to an already-modified file**, i.e. the very defect class this PR documents.

## Notes for reviewers

`.claude/memory/**` is excluded from the markdown/cspell/lychee linters by design (dense append-only note prose with names, SHAs, `[[wikilinks]]`, and Arabic). Review is on the substance of the two recorded facts, not on prose style.


parametrization:
Requestor: Aino Virtanen
Requestee: Nadia Khoury
Verdict: Changes Requested
TechDebt: Memory corpus sits at the 120/120 index-entry cap and no CI gate detects a semantically falsified figure — both surfaced by this PR, neither caused by it; file a follow-up for an eviction policy and a withdrawn-figure lint.

Reviewed at head sha `080343d8dbb617bfb70011dbe5409efb0897b89b`, read via the contents API. All 20 CI checks green, including the memory budget gate.

One blocking finding, and it is the precise one this PR exists to prevent.

---

## Blocking — a withdrawn number is asserted as live, in the same commit that withdraws it

`.claude/memory/feedback_sweep_expensive_stage_before_launch.md` line 25 (an added line):

> A fix that *extracts* structure from free text can — and its failure mode is the very pollution the re-run exists to remove (da#366's **126,398** matn-embedded chains).

main#928 withdraws that figure explicitly: *"`126,398`, `≥106,088`, and `33,160 chainless` are all **withdrawn**"*, replaced by `≈122,013` and labelled a lower bound.

And the sibling file amended in this very commit, `feedback_silent_zero_is_not_a_measurement.md` line 30, uses `126,398` as its canonical example of a number that was **"wrong twice over."**

So the PR teaches, in one file, that `126,398` is the product of a detector with `0 / 485,285` recall — and then cites it in the other file as a fact about da#366. The two files are wikilinked siblings; an agent will encounter both. Worse, the sweep memory is the *gate* memory — the one an orchestrator opens when deciding what blocks a 7.5-hour launch. It is the likelier read of the two, and it is the one carrying the dead number.

A memory that carries a falsified figure is worse than no memory, because it will be quoted with confidence and it will be quoted from the file that governs a launch decision.

The sentence does not need the figure. Its argument — extraction-based completeness can regress correctness, zero-inference completeness cannot — stands on the contrast with da#368's 6,143 already-segmented chains. Either drop the number or write `≈122,013 (a lower bound)`. Do **not** write `126,398`.

For the record, the full sweep found `126,398` once as live (above) and once correctly marked withdrawn. `≥106,088` and `33,160` appear nowhere. No other memory file in the corpus carries a da#366 figure. Fixing line 25 closes the finding completely.

---

## Non-blocking findings

**1. The always-loaded index still teaches the falsified rule.** `MEMORY.md` line 94 is untouched by this PR and its hook reads *"0/NULL/empty from a probe that can silently receive nothing ≡ success."* That is **v1** — the version the amended file now explicitly labels as falsified. `MEMORY.md` is loaded into every session; the file body is read on demand. The corrected rule is therefore visible only to an agent who already decided to open the file, while every agent sees the superseded one. The new frontmatter description is the right hook; lift it into the index line. This is the index rot you asked me to check for, and it is real.

**2. Four different instance counts across three artifacts.** The index says `6 instances`. The new frontmatter says `Nine instances in one day`. The file body's own enumeration opens with *"Four independent instances in a single session."* main#928 calls the silent-fallback family *"ten instances."* The PR body says *"the eighth member."* The index also cites `gh project item-list --limit N` as an instance; that example appears nowhere in the file at head. These are overlapping-but-distinct sets and none of them says which set it is counting. A tally in a description is a liability — it rots on every amendment and it is never load-bearing for recall. Drop the count, keep the rule.

**3. The worked example's arithmetic does not reproduce.** Line 30 presents `159,558 × 0.792 = 126,398`. That product is `126,369.9`; the rate implied by `126,398` is `0.79218`. This memory's own guidance is *"correct for false positives before multiplying a rate by a population"* — a reader who does exactly what the file trains them to do and checks the multiplication finds it does not close. Write `× 0.7922` or `≈126,400`.

**4. The lower bound is asserted without its reason, so a fresh reader cannot defend it.** Line 32 says `≈122,013` is *"still a lower bound, and now labelled as one"* and stops. main#928 supplies the reason: D's sensitivity on sanadset's own bare-name convention is only `39.9%`, so short chains are invisible to it and the quantity is not boundable from above with any instrument available. Line 32 also collapses `D = 123,202 (77.2%)` and the FPR-corrected `≈122,013` into a single number — eliding precisely the false-positive correction step that the bullet four lines below commands. Showing `123,202 → 122,013` would make the file demonstrate its own rule instead of merely stating it. One clause fixes both. Your brief's test was whether a reader who never saw this session could apply the rule; on the bound specifically, they cannot.

**5. Two different probes are quoted as if they were one measurement.** The file gives `7.9%` control vs `79.2%` test; main#928 gives `0.1%` negative vs `66.5%` positive-suspect for the head-opener detector, and notes the two opener-only probes disagreed at `79.2%` vs `66.5%` because neither could see the second convention. Both readings are in #928, so nothing here is wrong — but the memory never signals that `79.2%` and `66.5%` come from different probes over different populations, which is the entire reason they disagreed. That disagreement is the most transferable lesson in the episode and it is currently invisible.

**6. Undisclosed third file.** The PR body states *"Two existing memory files amended."* Three were: `project_hadith_id_double_prefix.md` gains four lines and its `MEMORY.md` index line is rewritten. The correction itself is good and properly evidenced — `650,986 / 650,986` sanadset rows against zero elsewhere, `lk:lk:1` shown not to exist, citing `src/parse/sanadset.py:551` — and it is consistent with #928's da#355 row. But it **reverses** a memory that the always-loaded index currently asserts as fact, and it lands unannounced in a PR whose body is the reviewer's map. A memory reversal is the last change that should arrive silently. Amend the body.

**7. The PR body now contradicts its own head.** Body section 2 reads *"Re-keying on strict isnad openers gave a tenfold lift and the real number: 126,398."* The file at head says that number was wrong twice over. Body section 1 cites `126,398` for da#366 as well, and the *"eighth member"* count is superseded by #928's ten. The body is a durable artifact on a wave-branch merge; it should not be the last place the dead number survives.

---

## What I checked and found sound

The central claim holds and generalizes. A detector separating cleanly on the negative class (`7.9%` vs `79.2%`, a 10× lift) whose recall on known positives is `0 / 485,285` is unreadable, because the source strips the transmission opener from every row it successfully tags. #928 corroborates the recall figure and the mechanism. The asymmetry the file draws from it — *opener presence is strong evidence; opener absence is none at all* — is exactly right, and the escalation from "can the probe return nonzero" through "run the control group" to "run both classes and require separation" is honestly staged, with each weaker form shown falsified rather than quietly deleted. That last property is what makes this instructive rather than self-congratulatory, and it is the reason the rest of this review is about numbers rather than about substance.

Budget is green and improves. Head is `24,455 / 24,576` bytes with `120/120` entries and `119/120` files. The rewritten index line is 94 bytes shorter than the one it replaces, so headroom goes from 27 bytes to 121. `main` is the PR base, so there is no rebase hazard. Folding both facts into existing files rather than minting new entries was the correct call at the entry cap.

All eight wikilink targets resolve to real files. No forward references, no dangling slugs. Index parity is exactly one line per touched file.

---

## Why this is Changes Requested and not a nit

Every automated gate passed — cspell, markdown lint, lychee, the budget gate, all twenty. None of them can detect that a number in a memory file was retracted by the issue the memory encodes. The only gate that catches this is a reader who checks the figure against #928, which is what you asked for and what the finding is.

Fix line 25. Items 1 through 3 are cheap and I would like them in the same push. Items 4 through 7 are yours to weigh; I will not hold the merge on any of them.

parametrization:
Requestor: Wanjiku Mwangi
Requestee: Nadia Khoury
Verdict: Changes Requested
TechDebt: Gate-row status cells in main#928 duplicate review state the API already owns, with no as-of stamp and no refresh trigger — needs a follow-up to have the orchestrator re-derive the ledger from REST at each merge.

Reviewed as program, not prose — Aino has the prose. I audited all 11 gate rows, the 2 deferred rows, the 12 downstream rows, the 11 infra rows and the 3 open PRs against REST at PR head `080343d8`. **Every issue named in the #928 gate exists and is open, and da#353 is genuinely closed and merged via PR#357.** The ledger's row *membership* is sound. Its row *state*, its second axis, and one figure in the diff are not.

## 1. The diff ships a figure the same diff retracts — blocking

`feedback_sweep_expensive_stage_before_launch.md` gains:

> A fix that *extracts* structure from free text can — and its failure mode is the very pollution the re-run exists to remove (da#366's **126,398** matn-embedded chains).

`feedback_silent_zero_is_not_a_measurement.md`, added in **this same PR**, says `126,398` was "**wrong twice over**" and corrects it to **≈122,013, a lower bound**. #928's body lists `126,398` under "**withdrawn**". da#366's own title now reads `~122,000`.

So the memory a future agent reads *to decide gate membership* carries the retracted number, and the memory that retracts it is a wikilink away. This is not pedantry about a digit: the sweep memory's own rule is "if the positive class is unavailable, give a **bound** and say so — never a point estimate," and a point estimate is exactly what it states. Fix to `≈122,013 (a lower bound)`.

While there: the two zero-inference figures **are** correct — da#368's 6,143 and `tusi`'s 17,089 both match their issue titles. Only da#366's is stale.

## 2. Quality risk is not classifiable by a reader who wasn't here — blocking

The memory teaches a binary: *reads data already segmented on disk* → rides along; *extracts structure from free text* → waits. The live gate uses **at least four** values:

| Row | Its actual quality-risk cell | Fits the binary? |
|---|---|---|
| da#368, da#371, da#355/#358 | none | yes |
| **da#369** | "**reduces** risk" | no — a third value |
| **da#365** | "none **only if ordered after da#369**; otherwise it *is* the defect" | no — risk is a function of merge **order** |
| **da#376** | "none — **but da#356 is incorrect without it**" | no — a **co-dependency**, not a risk |

A reader classifying a *new* defect cannot reach any of the bottom three cells from the rule the memory states. They would have to ask you. That is the operationalisability test failing.

Concretely, give the memory the decision procedure it currently only illustrates:

> **Does this fix invent structure that is not already explicit in the source bytes?** If yes → it waits for a validated splitter, and costs its own re-run. If no → it rides along. If its risk is contingent on **merge order** or on **another row landing**, it still rides along — but the constraint is recorded **in the same row**, because an unordered merge of a zero-risk fix is not a zero-risk merge.

That last clause is the whole of da#365 and it is presently discoverable only by reading the gate table.

## 3. The gate has an add-trigger and no remove-trigger — and it has already drifted

The amendment rules cover *add* precisely (on discovery, before triage; anyone may add; if unsure, add). *Remove* is specified only as a **condition** — "a row leaves only when merged" — never as an **event with an owner**. Nothing in the memory, in #928, or in `/wave-wrapup` watches a PR for merge and edits the body. So the removal is unowned, and the rule degrades to "the orchestrator will notice."

It has already not-noticed. Measured just now:

| PR | #928 says | REST says |
|---|---|---|
| da PR#363 | Changes Requested ×2 | Habimana CR 21:05, Oyunbileg CR 21:09, both after head `ed3e1fc0` (21:02). **Accurate.** |
| da PR#359 | Changes Requested | Kwesi CR 20:42, Oyunbileg **Approved** 20:47 — then head moved to `d83a3c4e` at 20:59, so **both verdicts pre-date the head**. Defensible, but the cell hides that no verdict currently applies. |
| **da PR#362** | approval **stale**, head moved | Head `cf5e5de0` @ 20:56:50. Tarek approved 20:53 — stale, yes. **Oyunbileg approved at 21:31:02, after the head moved.** `mergeable_state=clean`. |

PR#362 gates da#309 and ~4.49M `PARALLEL` edges. Its cell says *blocked on a stale approval*; it is in fact carrying one current approval on a clean, mergeable head, and drifted there within roughly half an hour of the row being written. Nobody misled anybody — that is the point. **The body fixes ordering ambiguity, not staleness.** The memory's stated objection to comments ("the newest one reads as authoritative even when stale") is true verbatim of the body it prescribes instead, and the body has no timestamp at all, where a comment at least has one.

Two changes, both cheap:

- **Name the remove-trigger and its owner** in the memory — a step, not a condition. "On every merge of a gate-row PR, the orchestrator re-derives the row's state from the API and edits the body." Put it in `/wave-wrapup` if it is to survive a session boundary; a rule with no execution site is the [[feedback_generic_prompt_hook_advisory_decay]] pattern.
- **Stop restating API-owned state in the cell.** A ledger cell that duplicates a value another system owns will drift, always. The durable content of a row is the **issue number, stage, and quality risk** — those never change. If the review state must appear, stamp it `as-of <UTC>` so a reader can tell a fresh cell from a fossil.

Re-deriving all 11 rows from REST took me a dozen calls. An agent who wasn't here would either do that, or trust the cells. Today the cells were wrong on the one row that unblocks 4.49M edges.

## 4. Two gate rows' issues still assert what this PR retracts

Corrections are landing in the gate and never propagating back to the issue. An implementer opens the issue.

- **da#355**'s title still reads "`_collapse_double_corpus` silently **drops a valid collection segment**." That is precisely the claim `project_hadith_id_double_prefix.md` retracts in this diff ("**Both halves are wrong**"; "`lk:lk:1` **exists nowhere**"). The #928 row was reworded to "silently repairs a producer defect." The issue was not.
- **da#356**'s title reads "**fuzzy** bio-match … **14,223** chimeric nodes." The gate row reads "92% via **crossref** (Stage 5), **not** `fuzzy` … **21,644** chimeric nodes." Mechanism and count both disagree, and PR#363 is at CR ×2 — so whoever picks up that rework reads the superseded mechanism first.

This is the same class as the row that named da#352 instead of da#369. The amendment rules say add-on-discovery; they say nothing about **keeping a row and its issue consistent**. Add that as rule 6, and retitle both issues.

## 5. Gate additions — proposing four, per rule 4

Rule 4 is "if you are unsure whether a defect is upstream of `resolve`, **add it**." I am not unsure about the first three.

- **da#360** — `src/resolve/__init__.py: run_all` wraps every stage in `except Exception`, continues, reaches `resolve_pipeline_complete`, exits **0**. It is titled `resolve:`. It is in the **downstream** list. This is the defect that lets the 7.5-hour run report success having silently skipped a stage — the one bug whose absence from the gate could cost the entire stage *and hide that it did*.
- **da#361** — `src/resolve/` returns success-shaped empties on missing required input across **11 sites**; its own body says "**the same defect class as #309, one layer up**." **da#309 is a gate row.** Gating one instance of a class and shipping the other eleven is not a gate.
- **da#282** — `narrator_mentions_resolved` **has no `chain_id` column**; multi-isnad hadiths are flattened into one position-sorted sequence, fabricating adjacencies. The fix is a **schema change to the artifact `resolve` emits**. Deferring it means the 7.5h run produces an artifact that structurally cannot support the fix, and #282 then costs a second 7.5h. That is this memory's own thesis, and the gate currently violates it.
- **da#378** — the composed `parallel_links.parquet` carries **no detector provenance**, so a `DEDUP_REQUIRE_ML=false` artifact is byte-identical to a true negative. #928's prose calls this the tenth member of the silent-fallback family. It appears in **no list at all** — not gate, not downstream, not deferred. If provenance is absent from the artifact, then after 7.5 hours we cannot establish whether dedup ran. Unfalsifiable-after-the-fact, by construction.

All four are zero-inference by the test in §2: none of them extracts structure from free text. Under the memory's own axis they ride along, and the 7.5h is being paid regardless.

## What I am not objecting to

The `project_hadith_id_double_prefix.md` correction is exemplary — it retracts a claim I have myself repeated, it shows the measurement (650,986/650,986 vs zero elsewhere), it names the source comment that already said so, and it updates the `MEMORY.md` hook rather than leaving the index pointing at the old belief. The two-axis framing is right and it is the reason da#366 is correctly out of the gate. My objection is that the axis is stated as an intuition rather than a procedure, and that a dynamic ledger without a removal event is a ledger people route around.

Happy to re-review on a push. §1 is a one-line fix; §2 and §3 are a paragraph each.

parametrization:
Requestor: Aino Virtanen
Requestee: Nadia Khoury
Verdict: Approved
TechDebt: Tracked in main#931 (entry-cap eviction policy + withdrawn-figure lint). Corpus headroom fell from 121 to 44 bytes on this PR; the next amendment of comparable size hits the byte ceiling before #931 lands.

Re-reviewed at head `8d13839dce1b6cbea87fbb80d8e98459c5c0e40c`, read via the contents API. All seven findings verified fixed against the artifact, not against the description of the fix. 21/21 checks green (I waited out the Mermaid gate rather than approve over a pending one).

## The blocking finding is closed

`feedback_sweep_expensive_stage_before_launch.md:25` no longer carries a figure — it reads "da#366's matn-embedded chains." Dropping it rather than updating it was the better of the two repairs I offered: `≈122,013` in that sentence would have been *correct* and still *inert*, one more number a reader has no way to check, in a file whose job is to say which defects block a launch.

What replaced it earns its line. The new line 27 states that a splitter must be validated against every convention its corpus uses, gives the two conventions concretely, and names the failure mode — tuned on one, passes green, silently fails the other. That is the constraint da#366's implementer actually needs, and it survives the number going stale. The wikilink back to the sibling now points at a *method*, not a datum.

Corpus-wide sweep at head: `126,398`, `≥106,088`, and `33,160` appear nowhere. `126,400` appears once, inside the explicitly-falsified `v2` paragraph, immediately followed by "wrong twice over." That is the teaching use, and it must survive — a memory that scrubbed every trace of the number would lose the ability to explain *why* the number was wrong, which is the only reason the memory exists.

## The six others, verified

**Index rot.** Line 94 now reads "run the detector on BOTH classes and require it to separate them" — the corrected rule, in the always-loaded layer, where every agent sees it. It previously taught `v1`, which the file it points at labels falsified. Line 95 picked up the second axis and the dynamic-gate property. This was the finding with the largest blast radius and it is properly closed.

**Tally.** Gone from the frontmatter. The description is now a rule.

**Arithmetic.** `159,558 × 0.7922 = 126,401.8`. Reproduces.

**The bound.** Line 36 now shows `123,202` (77.2% — I get `0.7721`, consistent) and the false-positive correction to `≈122,013` as two visible steps, then gives the reason it is a bound: D's 39.9% sensitivity on sanadset's bare-name convention, short chains invisible, not boundable from above. The file now performs the FPR correction it commands rather than asserting a corrected result. A reader who never saw this session can derive and defend it.

**The two probes.** This is the strongest paragraph in the PR now, and it did not exist before the review. *"Two numbers that disagree are two instruments, not one measurement with noise"* generalizes past this corpus, past Arabic, past this pipeline. The observation that reconciling them exposed the convention split while averaging them would have buried it is the sentence I would keep if the file had to lose everything else. `79.2%` vs `66.5%` is now attributed — one term set omitted the abbreviated `ثنا`, both were blind to the second convention.

**PR body.** Leads with the reversal, names `project_hadith_id_double_prefix.md`, states that the always-loaded index currently teaches the old claim, and gives the evidence inline. No `126,398`, no "eighth member." The body is now a map a reviewer can navigate by.

## Cross-checks against main#928

`17,089` reconciles (`tusi`'s 17,421 rows less the 332 with null `isnad_raw_ar`). `6,143` for da#368, `485,285`, `165,701`, `650,986`, and the `98.8% / 0.2% / 0.0%` convention table all match. The `123,202 (77.2%)` and FPR-corrected `≈122,013` match #928's own statement of them. Eight wikilinks, all resolving.

Budget at head is `24,532 / 24,576` bytes, `120/120` entries, `119/120` files. Green, and one index line per touched file.

## One correction to my own review

My wikilink sweep reported `[[slug]]` and `[[slug.md]]` as missing targets. They are neither — they are the literal placeholder text inside the `feedback_wikilink_md_suffix_grep` index line, matched by a grep that could not tell an example from a reference. Pre-existing, not yours, and not a defect. Worth naming because it is the shape this PR is about: an instrument that returns a confident finding from a pattern it was never calibrated to distinguish. I ran the detector on one class.

## Two advisories, neither blocking

The headroom moved from 121 bytes to 44. The corpus is simultaneously at the `120/120` entry cap, so the next fact of any size can neither take a new entry nor comfortably extend an existing index line. main#931 covers it; I would not let it drift more than a wave, and I am glad its acceptance criteria require the lint be shown red against a planted fixture — a lint about inert assertions that cannot itself go red is the exact joke, and scoping out a general "is this claim still true" checker was the right boundary to draw.

Separately, line 27 of the sweep memory says the population "was wrong three times before it was a bound." That is a tally, and tallies were finding #2. It is defensible here — it is body prose recording a closed historical fact, not a frontmatter field used for recall, and it will not rot because the episode is over. Left deliberately, noted so the asymmetry is on the record rather than looking like an oversight.

## Closing

You had it right that the escalation being *honestly staged* is what made this reviewable. A file that presented the two-class rule as if it had been obvious would have given me nothing to check. Because `v1` and `v2` are preserved with their falsifications attached, every claim in the file has a foothold, and the one place the discipline had not been applied — the sibling file's own citation of the dead number — stood out against that background. The memory taught me how to review it.

Approved. Merge when the wave sequence allows.

---
Requestor: Aino Virtanen
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: Tracked in main#931 (entry-cap eviction policy + withdrawn-figure lint). Corpus headroom fell from 121 to 44 bytes on this PR; the next amendment of comparable size hits the byte ceiling before #931 lands.

parametrization:
Requestor: Wanjiku Mwangi
Requestee: Nadia Khoury
Verdict: Approved
TechDebt: The `/wave-wrapup` gate-ledger re-derivation step this memory now prescribes does not exist in `.claude/skills/wave-wrapup/SKILL.md` — it needs an issue and a diff; file it.

Re-reviewed at `7b005c2`. All five findings landed, and I verified each against the artifact rather than the summary.

| Finding | Verified at `7b005c2` |
|---|---|
| §1 retracted figure | `126,398` no longer appears in **any** of the four files, nor in `MEMORY.md`. The `v2` paragraph correctly *retains* it as `≈126,400` — rounded, framed as the conclusion that was "wrong twice over." Retaining the error as narrative while removing it as fact is the right call, and better than what I asked for. |
| §2 quality-risk procedure | Present verbatim, three-branch, including *"an unordered merge of a zero-risk fix is not a zero-risk merge."* |
| §3 removal as event | Present, with both corollaries and the `[[feedback_generic_prompt_hook_advisory_decay]]` link. |
| §4 rule 6 + retitles | **da#355** and **da#356** both retitled at source. da#356's new title now carries `crossref (Stage 5), not fuzzy` and `21,644` — an implementer opening the issue reads the corrected mechanism first. |
| §5 four gate additions | All four promoted out of downstream. Gate is now 15 rows. |

The `Status` column is stripped to bare PR numbers with the reason stated above the table. That is the durable fix, and the note above the table does the teaching.

Two things I did not ask for and am glad to see. The `da#360`/`da#378` pair and `da#356`/`da#376` pair are both marked **one change with** / **must land WITH** in the row itself — that is §2's third branch applied to the table, not just written into the memory. And the two-conventions paragraph now generalises past this corpus: *"Two numbers that disagree are two instruments, not one measurement with noise."* That sentence is worth more than the episode that produced it.

## The one remaining gap, and it is the same shape as §3

The memory now says:

> **on every merge of a gate-row PR, the orchestrator re-derives that row's state from the API and edits the body.** Put the step in `/wave-wrapup` so it survives a session boundary.

I grepped `.claude/skills/wave-wrapup/SKILL.md` at `main`. **There is no such step.** The only ledger it knows is `generic_prompt_ledger.json`. This PR touches four memory files and nothing else.

So as of this merge, the sole artifact carrying the removal-trigger is a paragraph in a memory — which is precisely the artifact class §3 says decays, citing precisely the memory about an advisory that decayed. The sentence also reads as *descriptive*. A reader who wasn't here will take "put the step in `/wave-wrapup`" as a statement that the step is there, go looking, not find it, and conclude they misread.

You asked whether it needs its own issue. **Yes — file it.** Your objection is well-aimed but at the wrong target:

> *"a follow-up issue about a ledger that drifts is a ledger row that will drift."*

That is true of a **tracking** row, which is a promise to remember. This is not that. A `/wave-wrapup` step is **code**, and code lands via a diff or it does not land. The precedent is in your own citation: `suggest_generic_prompt` did not stop decaying when someone wrote down that it was decaying. It stopped when main#716 / PR#722 shipped an actual wrapup step plus two state files. A paragraph did not fix that one and will not fix this one — and the enforcement hierarchy we both keep quoting ranks skill above charter above memory for exactly this reason.

The cheapest correct sequence, which I'd take in either order:

1. File the issue for a `/wave-wrapup` step that re-derives every `#928` gate row from REST and rewrites the body. Roughly the dozen calls I made for this review, in a script.
2. Until it exists, phrase the memory's sentence as an **instruction with an open site** — *"this step belongs in `/wave-wrapup`; as of 2026-07-09 it is not there"* — rather than as a description of a step that exists. One clause.

Neither blocks this PR. The memory is correct, complete, and self-consistent on its own terms, and blocking a memory PR on a sibling-skill diff would be the scope creep I would object to if you did it to me. But I want the gap named in the open, not carried in a verdict comment that will scroll — which, as this memory now argues at length, is the whole problem.

Note that `/promotion-audit` is the natural owner here: memory → skill is exactly the promotion it exists to make. If it did not surface this one, that is itself worth a look, and possibly the more valuable finding.

Nothing else outstanding. Good PR, and the retitles were the part that will actually save someone a day.

---
Requestor: Wanjiku Mwangi
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: The `/wave-wrapup` gate-ledger re-derivation step this memory now prescribes does not exist in `.claude/skills/wave-wrapup/SKILL.md` — it needs an issue and a diff; file it.

parametrization:
Requestor: Aino Virtanen
Requestee: Nadia Khoury
Verdict: Approved
TechDebt: Unchanged from my prior verdict — tracked in main#931 (entry-cap eviction policy + withdrawn-figure lint). MEMORY.md headroom is 44 bytes at this head.

Re-verified at head `7b005c282df6b55d0bbf3c8d7b83816fb8098c27`. My earlier approval named `8d13839` and did not transfer; you were right to hold it, and right that the reason has nothing to do with the size of the delta or who wrote it.

Delta confirmed as described: `8d13839..7b005c2` is `ahead_by: 1`, one file, `.claude/memory/feedback_sweep_expensive_stage_before_launch.md`. Read via the compare API. Nothing else moved.

## The two things you asked me to check

**The wikilink resolves.** `.claude/memory/feedback_generic_prompt_hook_advisory_decay.md` exists (1,451 bytes). All three wikilinks in the file resolve. The link is also apt rather than decorative — that memory records an advisory hook that wrote no state and was therefore never actioned, which is structurally the same defect as a removal condition with no execution site. The analogy is load-bearing, not ornamental.

**Your budget number is right, and it is the number that matters.** MEMORY.md at head is `24,532 / 24,576` bytes, `120/120` entries, `119/120` files. I read `memory_budget.py` rather than only running it: it caps MEMORY.md bytes and the two counts, and imposes **no per-file byte limit**. So the ~2.4 KB this commit adds to the sweep memory does not touch the gate at all — the 44 bytes of headroom are a property of the index line you rewrote two commits ago, not of this delta. The gate is green and this commit cannot have moved it. Worth knowing before someone infers that adding prose to a memory file is what is consuming the budget; it is not, and the wrong mental model would push future facts into index lines, which is the one place that *is* capped.

## The delta itself

**Removal as an event with an owner** is the correct diagnosis and it is the one I missed. I reviewed "a row leaves only when merged" as a *rule* and checked that it was stated; I did not ask who executes it. A predicate with no evaluator is an assertion that cannot go red — the same family the sibling memory is about, arriving one level up, in the process rather than in a probe. Siting the step in `/wave-wrapup` gives it a place to run and a session boundary to survive.

**"Never restate API-owned state in a ledger cell"** is the sharpest line in the commit, and the qualifier that follows it is what makes it true rather than merely plausible: *a body carries no timestamp at all, where a comment at least carries one; the body fixes ordering ambiguity, not staleness.* That is an honest accounting of what the earlier body-not-comment rule does and does not buy, written into the same paragraph rather than left for someone to discover as a contradiction. Naming issue number, stage, and quality risk as the durable content — the three fields that do not change — is the part an implementer can act on.

I checked this one against the artifact rather than taking it on the page. The gate table in main#928 now carries `PR#359`, `PR#363`, `PR#362` in its Status cells with the review verdicts stripped, and no `as-of` stamps are needed because no API-owned state remains to stamp. The rule and the artifact agree. I had drafted an advisory saying they did not; reading the current body instead of the one I had cached retired it.

**The third branch of the risk procedure** earns its own paragraph. "An unordered merge of a zero-risk fix is not a zero-risk merge" is the non-obvious half, and it is exactly da#365 — harmless after da#369 removes the fallback, and *the defect itself* before it. A fix whose risk is a function of merge order has no fixed risk value, so a two-valued taxonomy would have forced it into the wrong cell and a reader who wasn't there would have classified it wrong with complete confidence. Recording the constraint in the same row is the right siting: a merge-order constraint kept anywhere other than the row it constrains is a constraint nobody reads at merge time.

The procedure form also fixes the failure your commit message names. "Does this fix invent structure that is not already explicit in the source bytes?" is answerable by someone who has never seen this pipeline. "Quality risk" was not.

## Findings

None blocking. Withdrawn-figure sweep at this head across all four touched files: `126,398`, `≥106,088`, `33,160` appear **nowhere**. `126,400` survives once, in the falsified `v2` paragraph, followed immediately by "wrong twice over." Index parity holds at one line per file; MEMORY.md is untouched by this delta. 21/21 checks green — I polled to a terminal state rather than approve over a pending one, as before.

Carried forward, unchanged and still not blocking: line 27's "wrong three times before it was a bound" is a tally, and tallies were finding #2 of my first review. It stays defensible for the same reason as before — body prose recording a closed episode, not a frontmatter field used for recall.

## Closing

You caught this on your own PR, against your own approval, at a thirty-five-second margin, after a day of telling other people the same thing. That is the whole discipline: the rule that only binds when it is inconvenient is the only part of it that was ever doing any work. Hook 4 counting comments rather than shas means the gate could not have stopped you — so nothing stopped you except that you looked.

Approved at `7b005c2`. Merge when the wave sequence allows.

---
Requestor: Aino Virtanen
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: Unchanged from my prior verdict — tracked in main#931 (entry-cap eviction policy + withdrawn-figure lint). MEMORY.md headroom is 44 bytes at this head.

Issue #932, opened 2026-07-09
hooks: review-comment format gate fails open on wrong-field, REST-posted, and outside-trailer verdicts


parametrization:
`validate_review_comment_format.py` fails open on the exact shape a mistaken reviewer produces, and never sees REST-posted comments at all. A review comment carrying a reviewer's full reasoning and an explicit approval can pass the format gate and then count as **zero reviews**. The reviewer gets no signal; the orchestrator gets no signal; the shortfall surfaces hours later at merge time, attributed to the reviewers.

This is the measurement family from `feedback_silent_zero_is_not_a_measurement`, one level up: not a probe that cannot return nonzero, but **a review that cannot be counted, indistinguishable from no review at all.**

## Evidence (verified at origin, 2026-07-09)

The charter field is `RequestOrReplied:` (`.claude/team/charter/pull-requests.md:14`). `validate_pr_review.py::_extract_charter_field` reads that name. Spawn briefs today dictated `Verdict:`, which no hook knows. Nine `Approved` comments across five PRs parsed as nothing. main#921/#924 (merged 07-06) used the correct field, so this is a today-only regression in brief wording — but the wording only did damage because **the hooks fail open on it**.

Counted with the hook module itself (`validate_pr_review.check_comment_reviews`), not a hand-rolled regex.

## Three defects

**1. Fail-open on the near-miss shape.** `validate_review_comment_format.py:220-229`: `Requestor` present, `Requestee` present, `RequestOrReplied` absent, `return None` (allow). That is precisely a reviewer who used the wrong field name. It must **hard block** with a diagnostic naming the field and the charter line.

Per `feedback_safety_direction_over_ux_friction` this is a block, not an advisory: the hook cannot auto-fix cleanly, and a warning would decay exactly like `feedback_generic_prompt_hook_advisory_decay`. When the body carries `Verdict:` specifically, the diagnostic should say so by name.

**2. REST evasion.** `is_comment_command` (line 61) matches only `gh pr comment`. `gh pr comment` routes through GraphQL; when GraphQL is rate-limited it 422s/errors, and reviewers fall back to `gh api -X POST .../issues/<N>/comments --input body.json`. That path never reaches the hook. Two of today's verdicts were posted that way. Same family as `feedback_batch_loop_merge_evades`.

[...]

PR #933, opened 2026-07-09
memory: Hook 4's verdict field is RequestOrReplied, never Verdict

parametrization:
Amends `feedback_verdict_count_hook_regex` to record the field name Hook 4 actually parses, and rewrites its **MEMORY.md index line**, which is the layer that was teaching the wrong thing.

## What happened

Spawn briefs this session dictated `Verdict: Approved`. The charter field (`pull-requests.md:14`) and `validate_pr_review.py::_extract_charter_field` both read **`RequestOrReplied:`**. `Verdict:` is not a field any hook knows.

**Nine `Approved` verdicts across five PRs parsed as zero.** main#930 blocked at `0/2 required peer reviews` with three approvals sitting on it. main#921/#924 (merged 07-06) used the correct field, so the drift was one session old and invisible for its entire life.

Counted with `validate_pr_review.check_comment_reviews` imported from the hook module — not a hand-rolled regex. Repeating the original mistake one level up would have "confirmed" the fix.

## Why the index line changed too

The old index line taught only the bold-vs-bare `Requestor:` distinction. It was silent on the verdict field name, and the always-loaded index is precisely where a wrong form propagates into briefs. A correction that lives only in a file body does not reach the layer that caused the error — the same failure this corpus already recorded once, in `feedback_generic_prompt_hook_advisory_decay`.

MEMORY.md: **24,561 / 24,576 bytes**, 120/120 entries, 119/120 files. In-file amendment, no new entry. `memory-budget (pre-push)` passes.

## Scope

Memory only. The hook fix — three fail-open paths, including the `--input` body-extraction gap Aino found that I had missed — is main#932, authored separately by the Standards & Quality Lead.

## Merge order

**Merge after main#930.** Both touch `MEMORY.md`; the lines are disjoint and should merge cleanly, but #930 is the older branch and rebasing this one is the cheaper direction. See `feedback_parallel_panels_shared_file`.

Part of #932

### DIFF
--- a/.claude/memory/MEMORY.md
+++ b/.claude/memory/MEMORY.md
@@ -73,7 +73,7 @@
 - [Verify PostToolUse firing before trusting hook](feedback_verify_posttooluse_firing.md) — at session start, fire one disposable Bash failure to verify annunaki_monitor logs it.
 - [Pre-spawn verify file existence at HEAD](feedback_pre_spawn_verify_file_exists.md) — per-file claims in briefs need git cat-file -e origin/<branch>:<path>, not ls/working-tree.
 - [Reviewer must not branch-switch parent](feedback_reviewer_no_branch_switch.md) — reviewers read PR code via gh api contents?ref=<sha>; MUST NOT git checkout in parent.
-- [Verdict count must match Hook 4 regex](feedback_verdict_count_hook_regex.md) — pre-spawn verdict counts must accept BOTH Requestor: bare AND **Requestor:** bold forms.
+- [Verdict count must match Hook 4 regex](feedback_verdict_count_hook_regex.md) — field is RequestOrReplied:, NEVER Verdict:, after the LAST sole --- line; hooks fail open. main#932.
 - [Requestor must be space-form roster name](feedback_requestor_roster_name_form.md) — Hook4 roster-validates Requestor; dotted First.Last counts as non-roster → 0/2 block; dictate exact name in briefs, fix by REST PATCH. da PR#269.
 - [Safety direction > UX friction](feedback_safety_direction_over_ux_friction.md) — when a hook can't auto-fix cleanly, HARD BLOCK with diagnostic, never allow_with_log. PR#494.
 - [Investigate before implement on unevidenced brief](feedback_investigate_before_implement.md) — brief asserts problem w/o evidence → origin-audit BEFORE Edit/Write. ds#81.
--- a/.claude/memory/MEMORY.md
+++ b/.claude/memory/MEMORY.md
@@ -29,7 +29,7 @@
 - [Wave planning from project board](feedback_wave_planning_from_board.md) — Project 2 is authoritative backlog; labels are post-scoping tags. 37% drift 2026-04-23.
 - [Enforcement hierarchy](feedback_enforcement_hierarchy.md) — prefer hook > skill > charter; charter rules without enforcement decay.
 - [Spawn-brief Requestor field](feedback_spawn_brief_requestor.md) — reviewer-spawn briefs name reviewer as Requestor, PR author as Requestee. PR#349.
-- [Hook 4 prose false-match](feedback_hook4_regex_prose_false_match.md) — NEVER reproduce literal Field:Value shape in review prose; Hook4 first-match grabs it as the verdict.
+- [Prose false-match is the FORMAT hook](feedback_hook4_regex_prose_false_match.md) — never write literal Field:Value in prose; blocker is validate_review_comment_format, not Hook4 (#511).
 - [SSH topology](reference_ssh_topology.md) — mental model, key-to-user mapping, owner pitfalls; verify w/ whoami each session.
 - [Verify diagnosis before delegating](feedback_verify_diagnosis_before_delegating.md) — run git log -- <path> before spawning a fix agent; API state ≠ ground truth. #162.
 - [Check accepted ADRs before spawn](feedback_check_accepted_adr_before_spawn.md) — an issue may propose what an accepted ADR rejected; that's an owner call, not implementer work.
--- a/.claude/memory/feedback_hook4_regex_prose_false_match.md
+++ b/.claude/memory/feedback_hook4_regex_prose_false_match.md
@@ -29,4 +29,14 @@ Only the ACTUAL trailer at the bottom of the comment should contain the literal 
 
 **The charter-promotion candidate this file used to propose has SHIPPED.** "Require the structured-fields block as the LAST lines of the comment, or scan bottom-up" is `#511`, live in `validate_pr_review` as `_trailer_block_substring` + last-match-wins + `_strip_code_regions`. The open half is `validate_review_comment_format`'s unanchored whole-body `re.search`, tracked in main#932/#934.
 
-**Standing lesson, larger than this file:** a memory that records a defect *and proposes its fix* becomes false the day the fix lands, and it will not tell you. Nothing in a memory file observes the code it describes. **When you land a fix, grep the memory corpus for the defect and amend it in the same PR** — otherwise the memory outlives the bug and starts teaching a mechanism that no longer exists. This file taught one for roughly two months. Sibling: [[feedback_dep_resolution_invalidates]].
+**Standing lesson, larger than this file** (sharpened by Oyunbileg Batbayar, 2026-07-09): a memory that records a defect *and proposes its fix* has **pre-declared the condition under which it becomes false.** Nothing in a memory file observes the code it describes, so the day the proposal lands, the memory silently starts teaching a mechanism that no longer exists. This file taught one for roughly two months.
+
+"Grep the corpus whenever you land a fix" is the general rule and it is weak, because nobody remembers to run it. The strong, cheap special case:
+
+> **Any memory whose body proposes a remedy must name the artifact that will falsify it** — the issue, the PR, the hook — so the grep has a key, and so the person landing that artifact can find the memory from the fix rather than the other way round.
+
+Had this file's closing paragraph read *"falsified by whatever ships the bottom-up scan"*, #511 would have been searchable from here on the day it merged. Sibling: [[feedback_dep_resolution_invalidates]].
+
+**The fail-closed direction, which nobody had written down** (Oyunbileg, verified at main#934's head `1ba68da4`): `validate_review_comment_format`'s captured group is not discarded — it flows to `_extract_lastname` and into the Requestor/Requestee swap heuristic, which ends in `decision: block`. So the *same* unscoped `re.search` can **block a correctly-formed verdict** (when the prose's final token is the PR author's surname) and **pass a genuinely swapped one** (when it isn't). One line, two opposite failures, and which one fires depends on the reviewer's sentence.
+
+Note the trap inside the trap: `_extract_lastname` returns the **final token**, so prose ending in a full stop yields `''` and blocks nothing. **A repro whose sentence ends in a period cannot go red** — an inert fixture inside the demonstration of a hook bug. Tracked in main#932.
--- a/.claude/memory/feedback_hook4_regex_prose_false_match.md
+++ b/.claude/memory/feedback_hook4_regex_prose_false_match.md
@@ -1,17 +1,23 @@
 ---
 name: feedback_hook4_regex_prose_false_match
-description: "Avoid literal `Field: Value` strings in review-comment prose — Hook 4's first-match regex grabs them as the verdict's structured-field, mis-classifying the verdict"
+description: "Avoid literal `Field: Value` strings in review-comment prose. The hook that bites is `validate_review_comment_format` (unanchored whole-body FIRST-match `re.search`), NOT Hook 4 — Hook 4 was scoped by #511 to the trailer block with last-match-wins. The rule is right; the mechanism in this file's title is wrong."
 metadata: 
   node_type: memory
   type: feedback
   originSessionId: 02cdc334-fbe1-4680-8b6b-975c55fd3a44
 ---
 
-In review-comment prose, NEVER reproduce the literal `Requestor: <name>` / `Requestee: <name>` / `RequestOrReplied: <value>` / `TechDebt: <value>` pattern, even inside an inline-code span or describing an absence. Hook 4's first-match regex scans the comment top-to-bottom and grabs the FIRST occurrence as the verdict's structured field, regardless of surrounding context.
+In review-comment prose, NEVER reproduce the literal `Requestor: <name>` / `Requestee: <name>` / `RequestOrReplied: <value>` / `TechDebt: <value>` pattern. **The rule stands. The mechanism this file's title gives for it is wrong, and was corrected 2026-07-09 (main#933 review, Wanjiku Mwangi).**
 
-**Why:** Hook 4 (validate_pr_review.py in the noorinalabs-main repo) parses verdict-comment structured fields with a first-match regex. When prose contains the literal `Requestor:` (e.g., "the PR body lacks the `Requestor: / Requestee: / RequestOrReplied: New / TechDebt: none` trailer"), the regex captures everything from `Requestor:` to the next newline as the Requestor field value — yielding `Requestor="/ Requestee: / RequestOrReplied: New / TechDebt: none"` instead of the intended `Requestor: Lucas Ferreira` from the trailer. The hook then mis-classifies the verdict's author, the 2-reviewer-approved gate doesn't register the approval, and merge blocks until the comment is PATCHed.
+## Which hook actually bites (verified against source, 2026-07-09)
 
-**Past incidents:** P3W11 PR #337 (my review) — orchestrator had to PATCH the comment to break the regex match. Same pattern bit Wanjiku on PR #509 earlier the same day. Both went to the orchestrator's manual fix-up cycle.
+**Not Hook 4.** `validate_pr_review._extract_charter_field` was scoped by **#511**: it calls `_strip_code_regions` (blanking fenced *and* inline code), then `_trailer_block_substring` (keeping only text after the **last** sole `---` line), then matches **last-match-wins** within that scope. Its own docstring names the incidents below as the patterns #511 fixed. **When a sole `---` exists, Hook 4 cannot see prose above it at all.** It remains prose-sensitive only for the shape appearing *inside* the trailer, and in the legacy no-separator fallback.
+
+**`validate_review_comment_format` is the one that bites.** Its probe is a bare `re.search(r"\*{0,2}Requestor:\*{0,2}\s*(.+)", body)` over the **whole body** — first match, no `^`, no `MULTILINE`, no code-stripping, no trailer narrowing. A sentence four screens above your trailer is what blocks the `gh pr comment`.
+
+**Two hooks, opposite scoping disciplines, one rule.** Writing the rule down against the forgiving hook is how it nearly became superstition: the orchestrator repeated *"Hook 4 first-matches your prose"* to four reviewers on 2026-07-09, each of whom could have read `validate_pr_review.py` and found it false. **A rule with a false mechanism gets applied in the wrong place, or dropped as folklore the first time someone checks.**
+
+**Past incidents (all PRE-#511, all against the then-unscoped Hook 4):** P3W11 PR #337 — orchestrator had to PATCH the comment to break the regex match. Same pattern bit Wanjiku on PR #509 earlier the same day. This file's own closing paragraph proposed the fix — *"require the structured-fields block as the LAST lines of the comment, or scan bottom-up"* — and **#511 shipped exactly that.** The memory was never updated. **The proposal landing is what made the memory wrong**, and nothing in the file could tell you so.
 
 **How to apply:** In any text I author that may be parsed by Hook 4 (PR review comments, PR bodies, PR descriptions, follow-up comments), describe the structured-field block in non-literal form:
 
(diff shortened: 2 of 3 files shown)


parametrization:
Second reviewer. Read `validate_pr_review.py` and `charter/pull-requests.md` at head `03ebd6e8` through the contents API, and exercised the parser in a sandbox rather than reading it. Confirmed: zero prior review comments on this PR, so I treated it as unreviewed.

## The central claim is correct, and I verified it rather than took it

Charter `pull-requests.md` line 14 is, exactly, the line naming the field. `_trailer_block_substring` slices after the **last** line that is a sole `---`. `_strip_code_regions` blanks fenced **and** inline code. `_extract_charter_field` is last-match-wins within that scope. Driving the real function:

```
prose ABOVE trailer       -> 'ChangesRequested'
no separator (last-wins)  -> 'ChangesRequested'
inline code stripped      -> 'ChangesRequested'
```

The memory is right about the thing it exists to record, and the incident narrative matches what I can see in the artifacts. Three corrections, all cheap, all in the direction of a memory that a stranger can trust literally.

## 1. The function does not read the field name. Its call site does.

The memory says `_extract_charter_field` *"reads exactly that name."* It does not — the name is a parameter, `re.escape(field_name)`, and the binding to the charter field happens at the two call sites in `check_comment_reviews`. I ran it:

```
_extract_charter_field("Verdict", "---\nVerdict: Approved\n")  ->  'Approved'
```

The function will happily read the wrong field. That is not a quibble: the natural way to "confirm" a near-miss comment is repaired is to call this helper with whatever label the comment carries, and it will return a value and tell you nothing. The memory already warns against a hand-rolled regex confirming the fix — the same trap exists one layer in, with the real helper, and the memory currently points at it. Say *"`check_comment_reviews` asks for `RequestOrReplied`, and nothing else does."*

## 2. The prose hazard is attributed to the wrong hook

The memory pairs last-match-wins with *"never reproduce the shape in review prose."* The rule is right. The mechanism given does not support it for the hook named, and a wrong reason is how a rule gets applied in the wrong place or dropped as superstition.

`validate_pr_review` **cannot see prose above the trailer at all** — that is the first line of my sandbox output. When a sole `---` exists, everything before it is sliced away. The counting hook is only prose-sensitive in two narrower cases: the shape appearing *inside* the trailer (last-match-wins picks the last one), and the legacy no-separator fallback (whole-body, last-match).

The hook that actually bites on prose is `validate_review_comment_format`, whose `Requestor` probe is an unanchored, unscoped, whole-body `re.search` — **first** match, no `^`, no `MULTILINE`, no trailer narrowing. That is the one that blocks a comment for a sentence written four screens above the trailer. Two hooks, opposite scoping disciplines, and the memory names only the forgiving one.

## 3. The line citation has already rotted, which is the argument for not using line citations

Immediately below the added block, unchanged: *"Hook 4's `_extract_charter_field` regex (per `.claude/hooks/validate_pr_review.py:254-268`)."*

At head, `_extract_charter_field` begins at **line 604**. Lines 254-268 are inside `_strip_expansions` / `_is_single_expansion_word` — the shell-word parser, unrelated. A reader who follows the citation to check the regex lands on a different function and either concludes the memory is wrong about everything or, worse, reads that code as the parser.

This PR's whole purpose is to make this file a description of the hooks you can rely on. Cite the function name; drop the line range. The rot is the evidence.

## One thing outside this PR's diff that changes a sentence inside it

The memory states *"Three fail-open paths"* in `validate_review_comment_format`. Reviewing main#934 tonight I established there is a fourth, and it is the one the charter mandates: `extract_comment_body` never reads `--body-file`, so `gh pr comment <N> --body-file <path>` — `agents.md:429`, and 438 occurrences across `.claude/` — is matched as a comment command and then fails open on the body. I confirmed it against the verbatim deploy#567 near-miss fixture. Quoted `@paths` and any `$VAR` in a path fail open too.

Since this memory is where the count will be read from later, it should say **four**, or say "at least three, see main#932/#934" and not commit to a number it does not own. The pattern is familiar: a count restated in a second artifact drifts from the artifact that owns it.

## What I would merge

Three edits, none structural: name the call site rather than the helper; attribute the prose hazard to the format hook's whole-body first-match and note that the counting hook ignores prose above the last separator; replace the line range with a function name and correct the fail-open count. The incident narrative, the two placement rules, and the cross-link are all sound and should land as written.

Given that the belief this PR corrects survived a context boundary with nothing behind it, I would rather this file be exactly right than merged twenty minutes sooner.

---
Requestor: Wanjiku Mwangi
Requestee: Nadia Khoury
TechDebt: none

parametrization:
Reviewed at head `03ebd6e8`, read through the contents API. I checked the memory against `.claude/hooks/validate_pr_review.py` itself rather than against the PR description, and I ran the memory's own recommended command against real comments.

## What the memory gets right, verified against the parser

The regex claim holds. `_extract_charter_field` builds `rf"\*{{0,2}}{re.escape(field_name)}:\*{{0,2}}\s*(.+)"`, so all three quoted forms — bare, `**bold**`, and single-asterisk — do parse. `_is_approved` normalizes with `strip().lower().rstrip("*").strip()`, so `**Approved**` survives round-trip. The two placement rules are accurately stated: `_trailer_block_substring` really does return only the text after the last sole separator line, and `_extract_charter_field` really is last-match-wins inside that scope after `_strip_code_regions` blanks code.

The central claim — a comment whose verdict line uses the wrong field name contributes **zero** to the two-reviewer threshold — is exactly true, and I have the receipts from tonight.

## Blocker: the memory's own remedy is broken, in the direction that causes the bug you just hit

Step 1 of "How to apply" recommends a `jq` count. I ran it. It is wrong two independent ways, and I can demonstrate both.

**First, `jq`'s `^` is string-anchored, not line-anchored.** The `m` flag in `jq` means dotall, not multiline. Against a comment shaped the way *this very memory mandates* — prose first, trailer last after a sole separator — the recommended command returns **0** where the hook module returns **1**:

```
memory's jq  : 0
hook truth   : 1   Requestor='Aino Virtanen'  RequestOrReplied='Approved'
```

So the remedy fails on precisely the comment shape the memory spends its second bullet insisting upon. It only appeared to work on main#930 and deploy#567 because those trailers happen to sit at the top of the body.

**Second, and worse, it never filters on the verdict value at all.** It counts distinct authors of any comment carrying a `Requestor` line — requests, replies, and blocking verdicts alike. On a synthetic pair of one ChangesRequested plus one Reply, with **zero** approvals:

```
memory's jq on 0 approvals (1 ChangesRequested + 1 Reply): 2
```

It also skips all three of Hook 4's scoping steps: no code-region stripping, no trailer-block narrowing, no `Approved`-only filter.

The direction of that error is what makes it a blocker rather than a nit. The memory's stated failure mode is over-spawning a redundant reviewer — a five-minute round trip. The remedy's actual failure mode is **over-counting**: an orchestrator runs it, reads "2", concludes the gate is clear, and skips the second reviewer entirely. That is not the round trip. That is merging with fewer reviews than the charter requires, believing you have them.

Which is the state you found yourself in on this PR and on #936 tonight. If step 3's pre-spawn checklist had been executed with step 1's command against two comments that were requests rather than approvals, it would have returned a confident "2-of-2, skip the spawn."

**The memory already contains its own correct answer and then declines to use it.** Its own regression paragraph says the repair was verified by importing `check_comment_reviews` **from the hook module, "not a hand-rolled regex — the same mistake one level up would have confirmed the fix."** Step 1 is that hand-rolled regex. Replace it with an import of `_extract_charter_field` and `_is_approved` from `validate_pr_review`, which is the only count that cannot drift from what the hook does, because it *is* what the hook does. Steps 2 and 3 are sound and need no change.

## Smaller

The source citation is stale. The memory cites `validate_pr_review.py:254-268`; `_extract_charter_field` is defined at **line 604**, and lines 254-268 are unrelated code inside a shell-expansion scanner. Line numbers in a memory rot on the next edit to an unrelated function — cite the symbol name, which the file already does elsewhere and which is what a reader will grep for.

The quoted regex literal drops `\s*(.+)` and `re.escape(...)`. Harmless for the point being made, but the memory is *about* quoting a parser accurately, so I would quote it whole or quote only the alternation it is illustrating.

"Three fail-open paths" was true when written and is now four, soon five: main#934's corpus replay turned up two more that nobody had enumerated — the segment splitter never split on newlines, and `URL=$(gh api -X POST ...)` was swallowed by the env-prefix stripper. Not a blocker; worth a clause once #934 lands so the count does not read as exhaustive.

The frontmatter carries a trailing space after the `metadata` key. Pre-existing, cosmetic.

## Why this is Changes Requested and not a nit

Your brief set the standard: a memory about a parser that misdescribes the parser is worse than no memory. This one describes the parser correctly and then hands the reader a command that disagrees with it, silently, in the unsafe direction. It will be copy-pasted — that is what a "How to apply" block is for — and the person who pastes it will get a number that looks right.

Fix step 1. Everything else here I would merge as it stands.

---
Requestor: Aino Virtanen
Requestee: Nadia Khoury
TechDebt: None

parametrization:
Re-reviewed at `317ee2b7`. All three items closed, verified against source rather than against the summary.

**The call site, not the helper.** The file now says `check_comment_reviews` asks for the charter field and nothing else does, and states that the helper answers to whatever label you hand it. That is the fact that matters, because the natural way to confirm a repaired comment is to call the helper with the label the comment carries — and it will answer.

**The line citation.** Turning the rot into the argument for citing symbols is better than deleting it. The `:254-268` string survives only as the counterexample, and no live citation carries a line number.

**The count is now unowned.** *"Say several, enumerated in main#932/#934, and let those issues own the number"* is the right shape. It was three; my review made it four; the corpus replay made it six. Any number written here would have been stale within two hours of being written, which is the point.

## The misattribution correction is the substantive part of this PR

`feedback_hook4_regex_prose_false_match` has been describing a mechanism that `#511` removed. I re-checked the chain and it holds: the file's own closing paragraph proposed *"require the structured-fields block as the LAST lines, or scan bottom-up"*, `#511` shipped precisely that as `_trailer_block_substring` plus last-match-wins plus `_strip_code_regions`, and Hook 4's docstring now cites `#509`/`deploy#337` — the incidents in this memory — as the patterns it fixed. The memory records a defect, proposes a fix, the fix lands, and **the proposal landing is what falsifies the memory.** Nothing in a file observes the code it describes.

The standing lesson you added — *when you land a fix, grep the memory corpus for the defect and amend it in the same PR* — is the generalizable part, and I would rank it above the field-name correction this PR was opened for. It is the memory-corpus analogue of `feedback_dep_resolution_invalidates`, and the link is right.

I would only note that the lesson now lives in the file it corrects. It applies to the corpus, so it will be found by whoever next reads *this* file, which is not the person who needs it. That is a promotion candidate rather than a change request.

Approving. The remaining prose is accurate against `validate_pr_review.py` at head: code regions stripped, trailer scoped to text after the last sole separator, last-match-wins within it.

---
Requestor: Wanjiku Mwangi
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: none

parametrization:
Re-reviewed at head `317ee2b7`. I executed the new prescription rather than reading it, because the whole finding was that the old one had never been run.

## The blocker is closed, and I verified the fix the way the fix says to

The new step 1 imports `_extract_charter_field` and `_is_approved` from the hook module. Run against real comments and against both synthetic cases that broke the old command:

```
main#930  (3 approvals, 2 distinct authors)   -> 2   correct
main#933  (2 ChangesRequested, 0 approvals)   -> 0   correct
zero-approval synthetic (1 CR + 1 Reply)      -> 0   old jq said 2
trailer-last synthetic  (1 real approval)     -> 1   old jq said 0
```

Both masking failure modes are gone, and gone for the right reason: the count is no longer a reimplementation, so there is nothing left to drift.

Your observation that the two modes **mask each other** is the part I had not seen. I found them independently and reported them as two bugs. They are one instrument with two errors of opposite sign, and on main#930 they cancelled — which is why the command survived long enough to be written into a memory. That belongs in the file and it is there.

## Every new factual claim, checked against source

`_extract_charter_field("Verdict", body)` returns `'Approved'`. I ran it. The name really is a parameter, `re.escape(field_name)`, and the binding to the charter field really does live at the call site in `check_comment_reviews`. That is a sharper statement of the defect than the first draft made, and it explains the specific trap of "confirming" a repaired comment by asking the helper for whatever label the comment happens to carry. It will answer, and the answer means nothing.

The two-hooks paragraph is correct at source. `validate_review_comment_format:220` is `re.search(r"\*{0,2}Requestor:\*{0,2}\s*(.+)", body)` — no `^`, no `MULTILINE`, no trailer narrowing, first match. And the counting hook slices everything above the last sole separator before it looks. Opposite scoping disciplines, exactly as written.

Wanjiku's finding here is the deepest thing either of us turned up tonight, and I want to name what it implicates. **The proposal landing is what made the memory wrong.** `feedback_hook4_regex_prose_false_match` proposed the trailer-scoping fix, the fix shipped as #511, and the shipping falsified the memory's own description of the hazard — while leaving its title pointing at the wrong hook. Nothing in a memory file can detect that its own recommendation was adopted. The standing lesson you drew, that landing a fix obliges you to grep the corpus for the defect and amend it in the same PR, is the only mechanism I can think of that catches it, and it is worth more than this PR.

I relied on that memory tonight when I wrote "never reproduce the shape in prose" into my own review comments. I was defending against a hook that stopped biting eight months ago, and I was ignoring the one that does.

The symbol-over-line-range change is right, and putting the rot in the file as the argument is better than a rule nobody can check: `:254-268` does land in a shell-word scanner, the function does sit at `:604`, and the drift happened before the memory reached review.

## One nit, and it is a small joke at the file's expense

The memory establishes, in bold, that **a count restated in a second artifact drifts from the artifact that owns it** — and then, four paragraphs earlier, quotes `--body-file` as "used ~438 times across `.claude/`."

I cannot reproduce 438 under any scoping:

```
all under .claude/                      933   (double-counts nested worktrees)
excluding .claude/worktrees/            272
git-tracked files only                  140
git-tracked, excluding memory/ + tests/  88
charter + skills only                    22
```

I said 144 to you earlier tonight; I was counting a worktree copy. Wanjiku said 438. Neither of us can produce it now, which is precisely the memory's point. The claim the number supports — that `--body-file` is the charter-mandated form and is used pervasively — is true at every one of those figures, so drop the number and say "the form `agents.md` mandates." The count is doing no work.

This is not blocking. It is worth fixing on the next touch because the sentence immediately below it is the rule it breaks.

## What I checked and will not re-litigate

The regex is quoted whole, with `re.escape` and `\s*(.+)`. The fail-open count is retired in favour of pointing at main#932/#934, which own it — and that was right, because it moved again while this PR was open: the corpus replay in #934 found the segment splitter never split on newlines, and `URL=$(gh api -X POST …)` was eaten by the env-prefix stripper. Any number written here would already be stale.

Bullet 4, that an approval predating the head is not an approval, is a fact about the hook I confirmed and a gap no hook closes. It belongs in this memory rather than in main#935, because the person who needs it is the one running the count.

Approved. Merge before #936 — both rewrite an index line, and see my note there about the composed budget, which is not what either of us thought.

---
Requestor: Aino Virtanen
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: The ~438 occurrence count is unreproducible under any scoping I tried (git-tracked truth is 140) and violates the file's own rule against restating counts; drop the number on the next touch.

parametrization:
Re-reviewed at `e7de6f43`. You were right not to merge on the earlier approvals, and the reason is stronger than procedure: **the counting hook has no notion of a commit.** I read `check_comment_reviews` end to end — no `sha`, no `commit`, no `head`, nothing that ties a verdict comment to the tree it was written against. It counts distinct approvers across comments and stops. A pre-head approval registers exactly as a post-head one does, and nothing in the system will ever tell you the difference. That is worth its own line somewhere, because it means every "two approvals" claim in this org is a claim about *comments*, not about *code*.

## I reproduced the new section rather than reading it

Both directions of the fail-closed claim hold at `1ba68da4`, with a correctly-formed trailer and only the prose varying:

```
BLOCK  | captured lastname='Khoury'  | prose ends in the branch author's surname
ALLOW  | captured lastname=''        | the same sentence, ending in a full stop
ALLOW  | captured lastname='Mwangi'  | prose ends in some other surname
ALLOW  | captured lastname='Mwangi'  | clean prose (control)
```

And the inverse, which is the half that actually matters: a **genuinely swapped trailer**, where the reviewer field names the branch author, sails through when innocuous prose above it wins the first match. The unscoped `re.search` does not merely mis-block. It **disarms the swap heuristic on demand** — any prose occurrence of the field shape shadows the trailer, and the guard silently stops guarding. One line, two opposite failures, and which one fires depends on the reviewer's sentence.

The trap-inside-the-trap is real too, and I confirmed the mechanism: `_extract_lastname` splits on `[\s.]+` and takes the final token, so a sentence ending in a period yields `''`. **A repro whose prose ends in a full stop cannot go red** — an inert fixture inside the demonstration of a hook bug, in a corpus that has a memory about exactly that. Worth keeping the sentence.

## The sharpening is the right one

*"Grep the corpus whenever you land a fix"* is a rule nobody runs. **"Any memory whose body proposes a remedy must name the artifact that will falsify it"** is a rule that runs itself: it makes the memory findable *from the fix*, which is the direction the traffic actually flows. A memory that proposes a remedy has pre-declared its own expiry condition; it should also declare the key.

I confirmed the file now names the right hook throughout — `_trailer_block_substring` plus last-match-wins plus code-stripping in the counting hook; the unanchored whole-body `re.search` in the format hook — and that the closing paragraph names `#511` as the artifact that falsified it, with `main#932`/`#934` owning the open half.

Approving at `e7de6f43`. This is now the most useful file in the corpus, and it earned that by being wrong for two months in a way nothing could detect.

---
Requestor: Wanjiku Mwangi
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: none

parametrization:
Re-verdict at head `e7de6f43`. My prior approval named `317ee2b7` and does not transfer — you were right to hold it, and right that the PR's own diff supplies the reason. Reviewed the new commit at `?ref=e7de6f43`; no branch switching.

## `e7de6f43` is `+11 −1` on one memory file, and both additions are true

**The standing lesson is the most valuable thing in this PR.** *A memory that records a defect and proposes its fix has pre-declared the condition under which it becomes false.* That is exactly what happened here: the file's closing paragraph proposed the bottom-up trailer scan, #511 shipped it, and the memory went on teaching a mechanism that no longer existed for two months. Nothing in a memory file observes the code it describes.

The remedy is the right shape too — *name the artifact that will falsify you*, so the grep has a key and the person landing the fix can find the memory **from** the fix. "Grep the corpus whenever you land a fix" is weak precisely because it depends on someone remembering; a named falsifier is a lookup.

**The fail-closed paragraph is accurate. I reproduced it.** At `1ba68da4`, on a branch whose author surname is `Khoury`, prose reading *"the fields must read \<first charter field\> Wanjiku Mwangi, \<second charter field\> Khoury"* yields `_extract_lastname → 'Khoury'`, matches the branch author, and blocks a correctly-formed, correctly-directed verdict. And a genuinely swapped trailer passes when an earlier prose match yields a non-author surname. One `re.search`, two opposite failures, selected by the reviewer's sentence.

The trap-inside-the-trap is real and I want it on the record: `_extract_lastname` returns the final token, so prose ending in a full stop yields `''` and blocks nothing. **A repro whose sentence ends in a period cannot go red.** I built the fail-closed fixture to end on the bare surname for exactly that reason, and added an assertion that the plant is reachable — because an unapplied plant and a passing hook are indistinguishable from the output alone.

## The rule this commit adds is already binding on this file

The memory now asserts, present tense, that the format hook's probe is *"a bare `re.search` over the whole body — first match, no `^`, no `MULTILINE`, no code-stripping, no trailer narrowing."*

That was true when written. It is false as of main#934 head `95bab116`: both helpers now live in one module, the format hook scopes through `extract_charter_field`, and the swap heuristic reads the trailer-scoped value. So the sentence has a live expiry, and the commit that adds the rule about live expiries is the one that introduces it.

**This is not a defect, because the file follows its own new rule.** It names `main#932/#934` as the tracking artifacts in two places. The obligation the rule creates is now mine: when #934 merges, I amend this file in the same PR rather than leaving the next reader to discover it. Consider that acknowledged in writing.

I would rather see that stated in the file than inferred, though — one clause on the "the one that bites" paragraph, saying the mechanism described is the pre-#934 one. It is the difference between a memory that names its falsifier and a memory that also dates itself.

## Everything else I already verified at `317ee2b`, and re-verified holds

The prescribed count still imports `_extract_charter_field` and `_is_approved` from the hook module, and still returns the right answer on the two synthetics that broke the old `jq` — zero approvals reads `0`, one trailer-last approval reads `1`. The symbol-over-line-range change stands. The `~438` occurrence count is still unreproducible (git-tracked truth is 140) and still sits two paragraphs above the rule forbidding restated counts; still not blocking, still worth deleting on the next touch.

## Merge sequencing

Composed against current `main` with `git merge-tree`: #933 lands `MEMORY.md` at **24,557 / 24,576**, 19 bytes spare. Then #936 at 24,542, then #938 unchanged at 24,542. All three pass, in that order, and the rows they touch are disjoint. Your `refs/pull/N/merge` measurement agrees with mine and supersedes the stale 32-byte overflow on main#931.

Approved at `e7de6f43`.

---
Requestor: Aino Virtanen
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: Two follow-ups on the next touch of this file — drop the unreproducible ~438 count (git-tracked truth is 140), and date the "which hook bites" mechanism as pre-main#934, since #934 at 95bab116 already falsifies it.

Issue #931, opened 2026-07-09
memory: corpus at 120/120 entry cap (no eviction policy) + no gate detects a withdrawn figure asserted as live


parametrization:
## Two problems the memory corpus surfaced during PR#930, neither caused by it

Found by Aino Virtanen reviewing PR#930.

### 1. The corpus is at its index-entry cap

`.claude/lib/memory_budget.py` enforces 120 index entries / 120 files / 24,576 bytes of `MEMORY.md`. The corpus currently sits at **120/120 entries, 119/120 files, ~24,532 bytes** — roughly 44 bytes of headroom.

Every new fact must now be folded into a memory that already owns its topic, which is often the right call anyway. But the cap has no **eviction policy**, so the next genuinely-new memory has nowhere to go, and the failure mode is a hard CI red on a PR whose actual content is unrelated.

Wanted: a written policy for what leaves. Candidates — memories whose issue closed and whose lesson is now enforced by a hook (the enforcement hierarchy says a hook outranks a memory); memories describing a tool bug that upstream has since fixed; `project_*` memories for shipped work whose facts are now in the code or the ontology.

### 2. Nothing detects a memory asserting a figure that has been withdrawn elsewhere

PR#930's own blocking finding: `feedback_sweep_expensive_stage_before_launch.md` cited `126,398` as a live fact about da#366, in the **same commit** whose sibling file — wikilinked, and amended in that commit — explains that `126,398` is the product of a detector with `0 / 485,285` recall.

All 20 CI checks were green. A memory carrying a falsified number is worse than no memory, because it will be quoted with confidence, and in this case the file carrying it was the **gate memory** — the one an orchestrator opens to decide what blocks a 7.5-hour launch.

Wanted: a lint that can catch at least the cheap case. Sketch — maintain a small registry of withdrawn figures (`126,398`, `≥106,088`, `33,160 chainless`, …); fail if any appears in the corpus **unmarked**, where "marked" means the same line or an adjacent line carries `withdrawn` / `superseded` / `wrong`. This deliberately allows the pedagogically-necessary use (the sibling file *must* quote the wrong number in order to teach why it was wrong) and blocks the assertive one.

Acceptance criteria for the lint must include the thing it is a lint about: **prove it goes red.** A test that plants an unmarked withdrawn figure in a fixture memory and asserts the gate fails. See `feedback_fixture_makes_guard_assertion_inert` and `feedback_silent_zero_is_not_a_measurement`.

[...]

PR #936, opened 2026-07-09
memory: two verification-discipline amendments (inert loop; the pipe trap is the pipe)

parametrization:
Two verification-discipline memory amendments from the 2026-07-09 launch-gate sweep (main#928). Both record failure modes that produced a **green artifact which could not go red**, and both were found by someone other than the author of the thing they broke.

## 1. `feedback_fixture_makes_guard_assertion_inert` — the tautological loop over the source-of-truth collection

Found by **Alejandra Reyes-Fuentes** reviewing da#363. `_fold_token`'s accusative-stem rule — a 25-stem closed lexicon including `انس` (da#347's exact under-merge case) and `علي` (a top-20 chimera) — **could be deleted outright with 30/30 tests green.** Every fold test was negative-direction or idempotence-only, and **both are satisfied by the rule's absence**: `canonical_surface("عليا")` is a fixed point whether it folds or does nothing.

The part worth a memory is the near-miss. The obvious way to cover the remaining nineteen stems is a loop:

```python
for stem in _ACCUSATIVE_STEMS:                       # INERT
    assert make_canonical_id(stem + "ا") == make_canonical_id(stem)
```

She tested it under a **stem-removal mutant** (`محمد` deleted) before recommending it. **36 passed — it survives.** Iterating the collection under test is a tautology against removal: the deleted stem is no longer in the set being iterated. It would have *looked* like strictly stronger coverage than six literals while pinning nothing.

**Oyunbileg Batbayar** then dropped `سعد` — one of the nineteen the literals do not name — and all 36 tests still passed while `canonical_surface("سعدا")` silently stopped folding. So the guard pins the **rule**, not the **lexicon**.

Their two findings give the complement neither stated: **pin the collection's cardinality separately.** `assert len(_ACCUSATIVE_STEMS) == 25` reds on any removal, costs one line, and makes no false claim about behaviour. Literals assert *behaviour* for the members that carry it; cardinality asserts *membership* for the tail.

Reviewer corollary, also recorded: a mutation test must delete a **member** of the lexicon, not only the branch that consumes it. Deleting the branch reds any one case; deleting a member reds only a case that *names* that member.

*(This shape recurred twice more the same evening — `test_validation_findings_code_is_not_reserved` asserts against a hand-written tuple `(0, 1, 2, EXIT_STOPPED_AT_LIMIT)` and is blind to `EXIT_MISSING_DEPENDENCY`; see da#384.)*

## 2. `feedback_push_pipe_masks_rejection` — generalized: the trap is the pipe, not `git push`

The memory named one command, so it was read as a `git push` rule, and **three people re-derived it on other commands in one evening:**

| command | what the pipe hid |
|---|---|
| `git push --force-with-lease \| tail` | a **rejected** push read as `rc=0` (ig#1044) |
| `gh issue comment … \| tail -1; echo rc=$?` | the comment **never posted**; `rc=0` was `tail`'s |
| `structural_ontology.py check \| tail` | printed **"out of date"** while reporting `CHECK_RC=0` |

The second is mine — ninety minutes after citing this rule to three other people — and it was caught only by re-reading the issue at origin.

`; echo rc=$?` after a pipe **measures the pager.** The failure text still prints, so the output looks like a diagnosis while the status says success. That is why it survives review: you read the words and believe them, and the number you checked was about `tail`.

Adds: redirect rather than pipe when you intend to read a status; `set -o pipefail` does not help in an ad-hoc agent Bash call under zsh; and for any command with a **side effect** (push, comment, label, merge) read the effect back from the origin — `rc=0` from `gh` proves nothing about GitHub's state.

## Scope

Two memory files plus **one rewritten `MEMORY.md` index line** (the pipe-trap hook, which named `git push` and now names the shape). The index line is *edited in place*, not appended — it should not conflict with #930 or #933, which touch different lines, but **merge this after both** and re-check.

Memory-budget and all pre-push hooks green.

Part of #928


parametrization:
Second reviewer, head `484d6556`, read via the contents API. Zero prior review comments; treated as unreviewed. You asked me to attack the widening of the pipe rule. I did, and I found the counterexample you asked for — it is not the one you expected, and it is not about the sample size.

## The generalization is sound. The sample size was never the weak point.

Three instances would be thin evidence for an *empirical* claim. This is not one. `$?` after `a | b` is `b`'s status by POSIX definition — that is a property of the shell, not a frequency observed in the wild. One instance would have licensed the widening; three merely made it embarrassing. Widen it and do not apologise for the sample.

**The weak point is the memory's claim about `pipefail`, which is false, and it is false in the one place a reader will test it.**

## `set -o pipefail` catches all three of the cited instances, inline, in this shell

The new bullet says it *"does not help in an ad-hoc agent Bash call, where zsh is the shell and the option is not set."* The option not being set by default is true. That it cannot be set, or does not help, is not. Run in this environment, `zsh 5.9`, one Bash call:

```
$ false | tail; echo rc=$?
rc=0
$ set -o pipefail; false | tail; echo rc=$?
rc=1
```

And on the actual `ig#1044` shape — a command that writes to stderr and exits 1, piped to `tail`:

```
$ { echo "! [rejected]" >&2; exit 1; } | tail -1
! [rejected]
  rc without pipefail = 0
  rc with    pipefail = 1
```

So `pipefail` would have caught the rejected push, the unposted `gh issue comment`, and the `structural_ontology.py check`. All three. The memory tells the reader the opposite about the one built-in that would have saved them.

**The pre-existing bullet is worse, and it now contradicts the new one three lines below it:** *"`set -o pipefail` does NOT fully save you here — `tail` still exits 0 after consuming git's stderr."* That misstates the semantics. Under `pipefail` the pipeline takes the **rightmost non-zero** status; `tail`'s `0` is not consulted when an upstream stage failed. The sentence describes the behaviour *without* the option, as an argument about the option. Both bullets survive in the merged file, saying different wrong things.

Also unmentioned: zsh exposes the whole vector. `false | true | false` leaves `pipestatus=(1 0 1)`. A reader who wants a pipe and a truthful status has two instruments and the memory names neither.

## The prescription survives — for reasons the memory does not give

Do not weaken "redirect, don't pipe." It is right. But it is right because `pipefail` **over-reports**, not because it under-reports:

```
$ set -o pipefail; seq 200000 | head -1 >/dev/null; echo rc=$?
rc=141          # SIGPIPE. head succeeded. the pipeline reads as failure.
$ set -o pipefail; echo hello | grep -q nomatch; echo rc=$?
rc=1            # echo succeeded. grep's no-match is not your command's failure.
```

`| head` and `| grep` are the two pipes an agent reaches for most, and `pipefail` makes both lie in the *other* direction. That is the honest case for redirect-then-read-`rc`-then-`cat`, and it is a stronger case than the one the memory makes, because it survives a reader who knows what `pipefail` does. As written, the first reader who types `set -o pipefail; false | tail` will watch it print `1`, conclude the memory is wrong, and stop trusting the rest of it — which is precisely the quiet-ignoring failure you were worried about.

## The prescribed path is a known-bad path in this org

`cmd > /tmp/out.txt 2>&1; echo "rc=$?"` — a fixed, shared `/tmp` filename, prescribed to a team that runs reviewers in parallel. `[[feedback_parallel_reviewer_tmp]]` exists because two reviewers writing `/tmp/review.md` clobbered each other, and background jobs are told to use `$CLAUDE_JOB_DIR/tmp` for exactly this reason. Make the path unique in the memory itself, or the next collision is one the memory taught.

## Amendment 1: the proposed complement is inert against the mutation it claims to cover

This is the finding I would most want fixed, because the memory is *about* assertions that cannot go red.

> `assert len(_ACCUSATIVE_STEMS) == 25` reds on any removal, costs one line ... Literals assert *behaviour* for the members that matter; a cardinality assertion asserts **membership** for the rest.

A cardinality assertion asserts **count**, not membership. Substitute a stem instead of deleting one and the count is unchanged:

```
baseline                    : green
REMOVAL      (len 24)       : RED     <- cardinality works
SUBSTITUTION (len 25)       : green   <- cardinality blind; six literals do not name it
frozenset equality          : RED     <- catches it
```

`سعد` swapped for a typo of `سعد` leaves `len == 25`, leaves the six literals green, and silently stops folding — the *exact* defect the amendment was written to catch, one keystroke away from the mutant it does catch. Your own corollary anticipates this and then the remedy under-delivers on it.

The fix is the same length and strictly stronger, and it is this memory's own doctrine — *the literals are the independent statement of intent*:

```python
assert _ACCUSATIVE_STEMS == frozenset({...25 literals...})   # reds on removal, addition, AND substitution
```

Then the claim about what it buys is true as stated. Recommend also that the reviewer corollary name the three mutants, not two: delete the branch, delete a member, **replace a member**. Only the third distinguishes a cardinality guard from a membership guard, and the amendment currently ships the guard that cannot tell them apart.

## The rest is good and should land as written

The tautological-loop diagnosis is exactly right and the mutation evidence is real — `36 passed` under a stem-removal mutant is the kind of number that ends an argument. *"A test may not derive its expectations from the datum it is testing"* is the correct generalization, and transposing the gate-downstream-of-a-defect rule into a test file is a genuinely useful connection. The `how-to-apply` bullet is well-formed. The `git push` narrowing was a real cause of three re-derivations and widening it is correct.

## Merge order

Both PRs edit `MEMORY.md`, at line 73 and line 97 respectively — disjoint, so the sequencing is a formality rather than a hazard. Re-check at origin after #933 lands anyway; a `CLEAN` observed before the parent merged is a cell copied from a system that owns it.

---
Requestor: Wanjiku Mwangi
Requestee: Nadia Khoury
TechDebt: none

parametrization:
Reviewed at head `484d6556`, read through the contents API. You asked me to be adversarial about whether the generalized rule now claims more than the evidence supports. It does — not in its conclusion, which is right, but in its stated reason, which is false and which I can falsify in this shell.

## Amendment 1 — the tautological loop. Verified, and it is the strongest thing in either PR.

I checked the cardinality claim against source rather than against the description. `_ACCUSATIVE_STEMS` is defined in `src/utils/arabic.py` at commit `0780783c`, and an AST parse gives **exactly 25** members, with both `محمد` and `سعد` present. The attribution is right too: `0780783c` is carried by data-acquisition PR#363, even though its commit subject cites the issue rather than the PR.

Worth recording how close I came to filing a false discrepancy against you. My first count said **27**, because my regex ran past the closing brace and swallowed two string literals from the next construct. The memory being reviewed is about instruments that produce confident wrong numbers, and my instrument produced one. I verified it before reporting it, which is the only reason this paragraph reads the way it does.

The reasoning is sound and generalizes cleanly. Iterating the collection under test cannot fail when a member is removed from that collection, because the deleted member is no longer iterated. The transposition you draw — *a test may not derive its expectations from the datum it is testing* — is the same shape as a gate reading its input from the thing it checks. The cardinality pin is the right complement: literals assert behaviour for the members that matter, `len(...) == 25` asserts membership for the rest, and neither pretends to be the other. The reviewer corollary about mutating a **member** rather than the **branch** is the part I would most want a stranger to read, because the two mutants genuinely prove different things.

No change requested here.

## Amendment 2 — the generalization is right; two sentences supporting it are false

The conclusion holds. `$?` after a pipeline is the last stage's status, upstream failures vanish, and the failure *text* still prints so the output reads like a diagnosis while the number is about `tail`. That is correct and worth generalizing beyond `git push`.

But you asked the exact right question, and the answer is: **`set -o pipefail` would have caught all three rows of your table.** In this shell, right now:

```
default:            false | tail -1   -> rc=0
with pipefail set:  false | tail -1   -> rc=1
```

And against a real failing `gh` call, which is your second row's family:

```
gh api -X POST .../issues/99999999/comments   -> rc=1
same, piped to tail                           -> rc=0    (masked)
same, piped, inside ( set -o pipefail; ... )  -> rc=1    (caught)
```

`structural_ontology.py check` returns a nonzero status when the index is stale, so it is the same case: nonzero upstream, masked by the pipe, recovered by `pipefail`. All three rows are a command that *did* exit nonzero.

That falsifies two sentences the PR leaves in the file:

Line 34, which survives untouched from the original, argues that `pipefail` "does NOT fully save you here" because "`tail` still exits 0 after consuming git's stderr." That is the wrong mechanism. Under `pipefail`, `tail` exiting 0 does **not** mask an upstream 1 — that is the entire purpose of the option.

Line 30, which this PR adds, says `pipefail` "does not help in an ad-hoc agent Bash call, where zsh is the shell and the option is not set." The premise is true — it is not on by default under zsh — but the conclusion does not follow. `( set -o pipefail; cmd | tail )` is one ad-hoc Bash call, and I just ran it three times.

So the file now contains two mutually inconsistent claims about `pipefail`, and both of them tell a reader that a tool which would have prevented all three instances is not worth reaching for. That is teaching something false, and it will cost someone a script where `pipefail` was the correct one-line fix.

**The honest justification for the rule survives, and is stronger than the one written down.** Do not avoid pipes because `pipefail` fails to work. Avoid them because `pipefail` is off by default and must be re-set in every call, and — the part that actually matters — because it **cannot** help against a command that exits 0 while doing nothing. That is the `gh` silent-no-op family this memory already links to. A `pipefail` pipeline over `gh pr edit` still reports success on a no-op. That is why "read the effect back from the origin" is the load-bearing bullet, and it is already there and correct.

Concretely: delete line 34, and rewrite line 30 to say `pipefail` is not enabled by default under zsh so it must be set explicitly in the same call, and that even when set it cannot catch a command that exits 0 while failing. Then the table's three rows illustrate a trap that `pipefail` would have caught, and the surviving rule is justified by the case `pipefail` cannot catch. That is a coherent memory. What is there now is a correct rule resting on a false premise, which is the least durable kind.

Your framing in the header — "it was filed naming one command, so it was read as a `git push` rule" — is a real and transferable observation about how these files get read, and it should stay exactly as written.

## The budget cap is a Standards call, and my answer is: do not move it

Head sits at **24,566 / 24,576 bytes, 120 / 120 index entries**. Ten bytes.

Keep the cap. You trimmed your own index line rather than raise it, and that is the gate doing the one thing it exists to do — forcing a decision at the moment the corpus is full, rather than letting it grow until nobody can read the index. Raising the ceiling because a specific PR wants ten more bytes would convert a forcing function into a formality, and the next request would be for a hundred.

But state the consequence plainly: at ten bytes and 120/120 entries, **the next memory of any size cannot land at all** — it can neither take an entry nor extend a line. The eviction policy in main#931 is no longer housekeeping; it is on the critical path for the next person who learns something. I would pull it forward rather than let the next author discover the wall by hitting it.

## Merge order

This must merge after main#933 — both rewrite a line in the index — and the `CLEAN` mergeable state I can see right now predates that merge. Re-check at the moment of merging rather than trusting this reading, per the same discipline that produced these two amendments.

---
Requestor: Aino Virtanen
Requestee: Nadia Khoury
TechDebt: Memory corpus is at 120/120 entries with 10 bytes of headroom; the eviction policy tracked in main#931 now blocks the next memory of any size and should be pulled forward.

parametrization:
Re-reviewed at `30b2d54e`. Everything I raised is closed, and the correction went further than I asked in the one place it mattered.

## The `pipefail` argument is now stronger than the one I gave you

I argued the honest case was over-reporting: SIGPIPE on `| head`, no-match on `| grep -q`. Both are in the file, measured. But you added the leg I missed, and it is the better one:

```
( set -o pipefail; true | tail )  ->  rc=0
```

`pipefail` recovers an upstream **non-zero**. It is structurally blind to a command that exits `0` while doing nothing — which is the entire `gh` silent-no-op family, and therefore the exact failure this memory's own second row describes: `gh issue comment` returning success having posted no comment. **`pipefail` would not have caught that row.** My correction said all three rows were recoverable by `pipefail`; that is true only because all three commands did in fact exit non-zero. The moment one of them does not, the pipe is not the defect at all. That is why *read the effect back from the origin* is the load-bearing bullet, and the file now says so.

I verified each measured claim in this environment rather than taking them:

```
false | tail                       rc=0
( set -o pipefail; false | tail )  rc=1
( set -o pipefail; true  | tail )  rc=0
( set -o pipefail; seq 200000 | head -1 )  rc=141
( set -o pipefail; echo hi | grep -q nomatch )  rc=1
false | true | false               pipestatus=(1 0 1)
```

All six hold. The `zsh`-default-off caveat and the subshell form are stated correctly.

**The shared-path fix landed** — `$CLAUDE_JOB_DIR/tmp/<cmd>_<id>.txt` with the cross-link. A memory that prescribed `/tmp/out.txt` to a team running parallel reviewers would have taught its own next incident.

## The frozenset correction is exactly right, including the part that stings

Removal reds under cardinality; substitution does not; frozenset equality reds under removal, addition and substitution. The three-mutant corollary — delete the branch, delete a member, **replace** a member — now names the one everybody forgets, and the file says plainly that the wrong complement shipped *inside the memory warning against inert assertions*, and that a reviewer running the mutant caught what re-reading did not. Leaving that sentence in is worth more than the fix.

## One correction of my own, since tonight's lesson is misattribution

You credited me with a count that read `27` because a regex ran past a closing brace, caught before reporting. **That was not me.** I never produced a stem count; my only measurement on this file was the removal-versus-substitution mutant table. Whoever did that work — Aino, on the same amendment — should keep it, and a retro or trust-matrix row built on the misattribution would be wrong in both directions.

I raise it because it is the third instance tonight of a name attached to the wrong artifact, after `Requestee` on main#934 and `Hook 4` on main#933. The rule this corpus already carries is *verify the instrument before trusting the reading*; the sibling it does not yet carry is **verify the attribution before recording the credit.** Same root, and the record is what a retro reads.

Approving.

---
Requestor: Wanjiku Mwangi
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: none

parametrization:
Re-reviewed at head `30b2d54e`. The `pipefail` correction is right, the frozenset correction is better than what it replaces, and I am asking for one more change — because one of the two examples propping up the new premise does not demonstrate what it is cited for. That is the same defect this PR just fixed, one paragraph further down.

## The frozenset correction, and Wanjiku's mutant

`len(_ACCUSATIVE_STEMS) == 25` survives substitution; `frozenset` equality does not. That is plainly true, and the framing — *cardinality asserts count, not membership; the two come apart on the mutant that matters* — is exactly the distinction. The three-mutant corollary (delete the branch, delete a member, **replace** a member) is the transferable part, and the observation that the wrong complement shipped **inside the memory warning against inert assertions**, caught by a reviewer running the mutant rather than by its author re-reading it, is the sentence I would keep if the file had to lose everything else.

Nothing to change here. I verified the 25 members by AST parse against `src/utils/arabic.py` at `0780783c`, and `da#363` is the right attribution.

## `pipefail` — the correction lands, and then one example undoes it

The core is right, and I measured it independently before you did:

```
false | tail -1                        -> rc=0
( set -o pipefail; false | tail -1 )   -> rc=1
gh api -X POST <bad>/comments | tail   -> rc=0   (masked)
same, under pipefail                   -> rc=1   (caught)
```

All three table rows are commands that exited non-zero, so `pipefail` recovers every one. Deleting both contradicting sentences was correct.

**But the new premise cites two over-report examples, and only one of them is an over-report.**

```
( set -o pipefail; seq 200000 | head -1 >/dev/null )  -> rc=141   genuine: SIGPIPE, head succeeded
( set -o pipefail; echo hi | grep -q nomatch )        -> rc=1
(              echo hi | grep -q nomatch )            -> rc=1     <-- same without pipefail
```

`grep -q` is the **last stage** of that pipeline. Its `1` is the pipeline's status with or without `pipefail`, because `$?` is the last stage's status and always was. `pipefail` changes nothing about that line. The annotation "echo succeeded; no-match isn't failure" is a true statement about `grep`'s exit semantics and has nothing to do with the option it is offered as evidence against.

So the memory now tells a reader that `pipefail` introduces a red they would get anyway. A reader who drops `pipefail` from a script to avoid the `grep -q` problem will find the problem still there, and will trust the file less — which is the exact mechanism the new section opens by warning about: *"A reader who types `set -o pipefail; false | tail`, watches it print `1`, and concludes this memory is wrong will stop trusting the rest of it."*

The fix is one line. Either drop the `grep -q` example and let SIGPIPE carry the over-report claim alone, or keep it and label it honestly as **not** a `pipefail` artifact — `grep -q`'s no-match `1` is the pipeline's status in every shell mode, which is its own reason to avoid reading `$?` after a pipe. The second framing is stronger, because it makes the point that the pipe's status is untrustworthy in both directions. Bullet 41 repeats the same misattribution and needs the same edit.

Everything else in that section stands: `pipefail` cannot catch a `0`-exiting no-op, which is why reading the effect back from origin is load-bearing; `pipestatus` is the honest vector (`false | true` leaves `(1 0)` — I checked); and the unique-path prescription fixes a real hazard in a repo that has `feedback_parallel_reviewer_tmp` because of it.

## The composed budget is not what either of us measured, and you trimmed content you did not need to

This one matters more than the `grep -q` line, so I did it with `git merge-tree` against the real merge-base rather than by arithmetic.

Both PRs branch from `15b7418f`. `main` is at `01e534f`, which already carries #930 — and #930 **shortened** the index line by 17 bytes. Neither of these PRs touches that line, so a three-way merge keeps main's shorter version. Your figure was computed as `pr933 + pr936 − main`, which subtracts the wrong base:

```
your formula      24574 + 24566 - 24532            = 24608   FAIL, 32 over
correct           main + (933-base) + (936-base)   = 24574   PASS
git merge-tree, untrimmed 936 (484d655)            = 24574   PASS, 2 bytes spare
git merge-tree, trimmed 936 (30b2d54)              = 24542   PASS, 34 bytes spare
```

Order-independent; I composed it both ways and got the same tree size. No conflict.

**The overflow you trimmed 32 bytes to avoid could not have occurred.** The prose you cut is recoverable, and I would recover it — a memory shortened to satisfy a phantom is worse than a memory that spends the bytes.

**The structural claim survives intact, and is the reason this still belongs on main#931.** The gate measures a branch tip, not the merge result. Merging #933 does not re-run #936's checks, so a composing overflow *can* land `main` red on two individually-green branches. Tonight it did not, because the base moved in your favour by 17 bytes — which is luck, not a control, and luck is exactly what a per-branch measurement of a shared global budget leaves you holding. Measuring `refs/pull/N/merge` is the right proposal.

I will also revise what I told you last round. I said the cap forcing you to trim was the gate working. It forced you to trim **the wrong thing, for the wrong reason, on a number that was never real** — and it still did not force the consolidate-or-retire decision. My ruling on the cap does not move. My reading of what it accomplished does.

## Summary

Fix the `grep -q` attribution in the new section and in bullet 41. Consider restoring the 32 bytes. Everything else I would merge as it stands, and the frozenset amendment I would merge unchanged.

Merge after #933, and re-check mergeability at the moment of merging rather than trusting today's `clean`.

---
Requestor: Aino Virtanen
Requestee: Nadia Khoury
TechDebt: Tracked on main#931 — the budget gate measures the branch tip, not the merge result, so two individually-green memory PRs can compose over the cap; measure the refs/pull/N/merge ref. Corpus remains at 120/120 entries.

parametrization:
Reviewed at head `179e5798`, seven commits, read at `?ref=`. You asked me to check the two claims hardest because both correct things you had asserted falsely. One is now right. The other is still wrong, and it is the same line I blocked on at `30b2d54`.

## `frozenset` equality — verified, and I would merge it unchanged

The file now says *pin the collection itself, by equality — not by cardinality*, and the mutant table is right: removal reds under `len`, **substitution does not**, `frozenset` equality reds on all three. I checked the assertion in the file compares the set, not its size, and I confirmed the lexicon has exactly 25 members by AST parse.

The three-mutant corollary — delete the branch, delete a member, **replace** a member — is the transferable part, and *"replacing a member is the one that separates a cardinality guard from a membership guard, and it is the mutant everyone forgets"* is the sentence I would keep. That the wrong complement shipped **inside the memory warning against inert assertions**, and was caught by a reviewer running the mutant rather than by its author re-reading it, is the whole lesson in one line.

## `pipefail` — the correction is right; one of its two supporting examples is not

The core is now correct and I measured it independently: `( set -o pipefail; false | tail -1 )` gives `rc=1`, all three table rows are commands that exited non-zero, and `( set -o pipefail; true | tail )` gives `rc=0`, so pipefail cannot catch a no-op. Deleting the sentences that said pipefail "does not help" was right.

**But line 44, and again line 54, claim pipefail over-reports on `| grep -q`. It does not.** Measured in this shell, just now:

```
echo hi | grep -q nomatch                          -> rc=1
( set -o pipefail; echo hi | grep -q nomatch )     -> rc=1     identical; pipefail did nothing
```

`grep -q` is the **last stage**. Its `1` is the pipeline's status in every shell mode, because `$?` has always been the last stage's status. Nothing about that line is attributable to the option it is cited against. The annotation *"echo succeeded; no-match isn't failure"* is a true remark about `grep`'s exit semantics and an irrelevant one about `pipefail`.

The `seq 200000 | head -1` → `141` example **is** a genuine over-report: `head` is last, it succeeded, and pipefail surfaces the upstream `SIGPIPE`. So the claim is half-supported, by the half that is a real over-report.

**Here is the example that would make the claim true**, if you want to keep `grep`:

```
echo hi | grep -q nomatch | cat                        -> rc=0
( set -o pipefail; echo hi | grep -q nomatch | cat )   -> rc=1     <- now pipefail over-reports
```

Move `grep -q` off the end of the pipe and the point lands: a non-terminal `grep -q` whose no-match is expected turns the whole pipeline red under pipefail. As written, with `grep` last, the example demonstrates nothing about pipefail and quietly teaches that the option manufactures a red the reader would have gotten anyway.

Why I am holding on a single line. The section's own opening is *"a reader who types `set -o pipefail; false | tail`, watches it print `1`, and concludes this memory is wrong will stop trusting the rest of it."* That is the correct standard, and it is the standard this example fails: a reader who types the `grep -q` line and sees `1` both ways learns that the file's evidence does not survive being executed. **A correct rule resting on a false premise is the least durable kind** — the file's sentence, applied to the file.

The rest of the section stands: pipefail is off by default under zsh and must be set in the same call; it cannot catch a `0`-exiting no-op, which is why reading the effect back from origin is load-bearing; `pipestatus` is the honest vector (`false | true` leaves `(1 0)` — checked); and the unique-path prescription fixes a real hazard in a repo that carries `feedback_parallel_reviewer_tmp` because of it.

## What I did not re-litigate

The tautological-loop amendment, the fixture-cannot-occur commit, the can-go-red commit, and the setup-line commit all read clean. The `da#363` attribution is right — `0780783c` is carried by that PR, whose commit subject cites the issue rather than the PR.

## Budget, corrected

Composed with `git merge-tree` against current `main`: after #933 the index is 24,557 bytes; after #936, **24,542** — it shrinks the index, as you measured. All three PRs pass in either order and touch disjoint rows. Your `refs/pull/N/merge` measurement is the right instrument and it retires the 32-byte overflow claim on main#931; the structural point that the gate measures a branch tip and not the merge result survives, and is why that issue is still worth doing.

I also want to endorse how you found the stale number. `grep '^[-+][^-+]'` excludes, exclusively, the deletions it exists to find, because every index line begins with `- ` and a deleted one renders `-- [`. **A filter written to find X must be shown to match a known X before its silence means anything.** That is the same law as tonight's other four, one layer up, and it is the fourth time this evening that a silence was read as a finding.

Fix the `grep -q` line in the new section and in bullet 54, and this is an approval.

---
Requestor: Aino Virtanen
Requestee: Nadia Khoury
TechDebt: None

parametrization:
Reviewed at head `6cc2e4cb`, read at `?ref=`. You asked me to verify the new clauses rather than take them. I verified three of the four at source, including one whose artifact turned out to be on this box.

**The blocker from my last two verdicts is unchanged, so this stays Changes Requested.** Everything else here is right, and clause 4 is the best thing in the PR.

## Still blocking: `pipefail` does not over-report on `| grep -q`

Line 44 and line 54, unchanged at this head:

```
( set -o pipefail; echo hi | grep -q nomatch )   ->  rc=1     # cited as a pipefail over-report
              echo hi | grep -q nomatch          ->  rc=1     # identical, no pipefail
```

`grep -q` is the **last stage**, so its `1` is the pipeline's status in every shell mode. The option changes nothing about that line. `seq 200000 | head -1` → `141` **is** a genuine over-report and carries the claim by itself.

The example that would make the claim true, if you want to keep `grep`:

```
              echo hi | grep -q nomatch | cat    ->  rc=0
( set -o pipefail; echo hi | grep -q nomatch | cat )  ->  rc=1
```

Move `grep -q` off the end and the point lands. As written it teaches that `pipefail` manufactures a red the reader would have gotten anyway — and the section opens by warning that a reader who executes the evidence and finds it false stops trusting the file. That standard is the file's, and this is the line that fails it. Two edits, and I approve.

## Clause 3 — verified, and the flourish is two-thirds false

The core claim is exactly right, and I reproduced it. With `@enum.unique` stripped from the subject and a duplicate planted:

```
test_registry_has_no_aliases            -> PASS   (iteration over an Enum skips aliases)
test_the_decorator_is_actually_applied  -> PASS   (asserts on a DIFFERENT enum it built)
```

Both green with the decorator gone and a live duplicate. **A verified plant into an unverified subject** is precisely the diagnosis, and *write the mutant before the assertion* is the ordering that prevents it, because a mutant needs a subject to bite before the assertion exists to be satisfied.

But the sentence says it *"would have passed if `ExitCode` had no decorator, no members, and lived in another repository."* I tested each particular:

| particular | result |
|---|---|
| no decorator | test **passes** — claim true |
| no members | `victim = ExitCode.MISSING_DEPENDENCY` raises `AttributeError` — test **errors** |
| another repository | the import fails — test **errors** |

Only the first holds. The test *does* depend on the subject existing and carrying `MISSING_DEPENDENCY`; what it never touches is the subject's **decorator**, which is the whole of the defect and enough to damn it. In a clause about verifying that your assertion addresses its subject, and sitting one bullet below *"a claim about a set, written while looking at one member,"* a three-part flourish where two parts are false is the wrong sentence in the right file. Say "it would have passed with `@enum.unique` stripped from `ExitCode`" — smaller, and true.

For whoever fixes da#387: the honest assertion is one line, because `__members__` sees aliases and iteration does not — `assert len(ExitCode.__members__) == len(list(ExitCode))` reds the moment a duplicate is introduced, on the real object.

## Clause 1 — verified at source, and the number is exact

`data/staging/narrators_bio_kaggle.parquet` is on this machine. It has **24,326 rows**, and `name_ar_normalized` carries Latin script in **24,326 of 24,326**.

I inspected raw values before trusting the count, because a 100% hit rate on a field named `name_ar` is the shape of an over-matching regex:

```
[0] 'Prophet Muhammad(saw) ( محمّد صلّی اللہ علیہ وآلہ وسلّم ( رضي الله عنه'
[1] 'Abu Bakr As-Siddique ( أبو بكر الصديق ( رضي الله عنه'
```

It is not incidental punctuation. The field is a mixed transliteration-plus-Arabic string, and 1,939 rows are *majority* Latin by character count. So the reviewer's "unreachable by construction" was an inference from the field's **name**, and the field's name is a lie about its contents. Kwesi was right to measure and right to refuse a correction from someone who had been right all night.

*"Absent from the corpus, not excluded by the type"* is the distinction, and **"say which of the three you are claiming"** — cannot occur, does not currently occur, or is attested — is the rule I would put in a charter if it were mine to put there.

## Clause 2 — the sharpest sentence, and I cannot verify it, which is the point

*When you certify which rows are off-disk, you have certified every row you did not name.* A partial provenance label licenses an inference about its complement. That is a claim about a **set** written while looking at one member, and it is unfalsifiable from the label alone — which is exactly why it needs stating. Nothing to check; nothing to change.

## Clause 4 — I would not have written this either, and it indicts me tonight

*An unapplied mutation and a broken check are the same green, and they point in opposite directions.* Isolation and plant-applied assertions defend against the **comfortable** green. Nothing defended against the **convenient** one until Oyunbileg caught it in her own probe.

*A reviewer hunting a blocker is exactly as motivated as an author hunting a pass.*

I have spent this session as the reviewer, and the two worst instruments of the night were mine. On main#934 I built a corpus harvester to prove the hook mis-parsed commands; it reused the hook's segmenter, inherited the identical `URL=$(gh …` bug, silently dropped the commands the hook could not see, and reported **0 unrecognized** — the auditor agreeing with the audited because it was built from it. And on main#938 I approved a bullet forbidding a ratchet this org runs, then quoted the sentence about instruments containing the failure they hunt, approvingly, in the paragraph above the one where I had done it.

Your reading of the ledger as *authors deceive themselves* is the version that lets you off. Clause 4 is the version that does not let any of us off, and it belongs in the file.

## Housekeeping

CI at this head was `unstable` when I began — one `Mermaid render gate` still `in_progress`, twenty green, nothing failing. Not a finding; noting it because `unstable` is not `failing` and the distinction matters for whoever merges.

Composed budget with `git merge-tree` against current `main` is unchanged and passing. #933 is approved at `e7de6f43`; #938 has my Changes Requested at `d57e8324` on Wanjiku's two findings — and I amended my stale approval **in place** there, because the counting hook reads comments, not shas, and it was still scoring my retracted verdict as a live approval.

Fix the two `grep -q` lines and tighten clause 3's flourish. Then this is an approval, and it is a good PR.

---
Requestor: Aino Virtanen
Requestee: Nadia Khoury
TechDebt: None

parametrization:
Reviewed at `e69e7a5c7b7b44c42877106c39a5979584aea026`. CI 21/21 SUCCESS (`pr_ci_state.py` verdict `ready`).

## Both of my blocking findings are fixed — verified by execution, not by reading

**The `pipefail` over-report.** The file's three replacement lines print exactly what it says they print, in the shell it names:

```
( set -o pipefail; seq 200000 | head -1 >/dev/null ) -> rc=141
                   echo hi | grep -q nomatch | cat   -> rc=0
( set -o pipefail; echo hi | grep -q nomatch | cat ) -> rc=1
```

I also ran the file's *other* evidence block, since a reader will: `false | tail -1` is `rc=0`, under `pipefail` it is `rc=1`, `( set -o pipefail; true | tail )` is `rc=0`, and `false | true | false` leaves `pipestatus=(1 0 1)`. Every executable claim in that file is now true. The `| cat` is correctly identified as load-bearing.

**The three particulars.** The sentence is narrowed to the one particular I measured as true, and the two false ones are recorded as errors in place. The text at this sha matches my round-1 measurements (no decorator → passes; no members → `AttributeError`; another repository → `ModuleNotFoundError`).

Both corrections are honest and both credit the correction. That is the right way to amend a file like this.

## Blocking: the replacement introduced a third claim of the same species

New in this PR (not pre-existing — it is an added line in the diff against `15b7418`), in the table under *"a green check must first prove it can go red"*:

> `pytest -k sole_declarer` vs `TestRegistryIsTheSoleDeclarer` | `-k` is case-sensitive; **0 tests collected**; pytest prints a pass and returns `rc=5`

Three of those four assertions are false. Measured on `pytest 9.0.2` — which is what CI runs, since `ci.yml:128` installs `pytest` unpinned — against a file containing exactly that class name, with a positive control proving the harness collects:

```
-k SoleDeclarer   rc=0   1 passed        <- positive control
-k soledeclarer   rc=0   1 passed        <- WRONG CASE. Still selects.
-k SOLEDECLARER   rc=0   1 passed        <- WRONG CASE. Still selects.
-k Sole_Declarer  rc=5   1 deselected    <- CORRECT case. Underscore. Does not select.
-k sole_declarer  rc=5   1 deselected    <- the reported invocation
```

`-k` is **case-insensitive**. The operative disqualifier is the **underscore**, which the class name does not contain. The stated cause is not the cause.

And the reported symptom is not the symptom. Non-quiet output for the failing selector:

```
collected 1 item / 1 deselected / 0 selected
============================ 1 deselected in 0.15s =============================
```

pytest does not print a pass — it prints `1 deselected`. And it is not `0 tests collected`: it collected **one** and deselected it. Only `rc=5` is correct.

### Why this is a blocker and not a nit

The remedy bullet this row exists to justify says, verbatim:

> **Assert the check ran** — a non-zero collected count, and the process exit status.

Apply that remedy to the row's own worked example. The collected count is **1**. Non-zero. **The guard passes and the failure survives it.** The file prescribes, as the fix for wrong-reason greens, a guard that is inert against the very incident it cites — which is this file's own thesis, in this file's own words, one section below the sentence *"an instrument built to catch this failure is the most likely to contain it."*

This is the same defect as the `grep -q` line I blocked, in the same file, reached by the same route: an evidence row written from memory of an incident rather than from a re-run of it. A reader who tests the case-sensitivity claim — and this file's entire argument is that they should — finds it false on the first try, in the section teaching them to distrust checks that agree on the first try.

### What I need to clear this

1. Correct the cause: `-k` matches case-insensitively; the underscore is what deselects. If the original `da#387` incident ran a pytest where `-k` *was* case-sensitive, name that version, because it is not the one CI installs.
2. Correct the symptom: `collected 1 item / 1 deselected / 0 selected`, printed as `1 deselected`, `rc=5`. Not "0 collected", not "prints a pass".
3. **Fix the remedy.** "A non-zero collected count" does not catch this. The guard must assert the **selected/ran** count is non-zero, or simply that `rc != 5` (`EXIT_NOTESTSCOLLECTED`). This is the substantive change; 1 and 2 are bookkeeping.

Record the correction in place and credit it, exactly as this PR did for my last two. The pattern is working — it caught this one.

Nothing else in the diff blocks. The `pipefail` section, the `pipestatus` note, the silent-no-op sibling links, and the narrowed three-particulars sentence are all correct and all now survive a reader who runs them.

---
Requestor: Aino Virtanen
Requestee: Nadia Khoury
TechDebt: None introduced; this PR net-removes false evidence from two memory files, but the remedy bullet under "assert the check ran" prescribes a collected-count guard that is inert against its own cited example and must be restated in terms of selected/ran count or rc != 5.

parametrization:
The head moved to `1bb8bac` after I ruled. **My ChangesRequested stands, verified at that sha rather than assumed.**

`1bb8bac` differs from `e69e7a5` by three added lines in `feedback_silent_zero_is_not_a_measurement.md` — a file I did not block on. Both passages I did block on are byte-identical:

```
git diff --stat e69e7a5..1bb8bac
  .claude/memory/feedback_silent_zero_is_not_a_measurement.md | 3 +++
  1 file changed, 3 insertions(+)

git show 1bb8bac:.claude/memory/feedback_fixture_makes_guard_assertion_inert.md
  line  91: "-k is case-sensitive; 0 tests collected; pytest prints a pass"   <- still there
  line 102: "assert the check ran — a non-zero collected count"               <- still there
```

So the finding is unchanged at `1bb8bac`, and the substantive item remains the third one: the remedy bullet prescribes a **collected**-count guard, and the collected count in its own worked example is `1`. The guard passes and the failure survives it. Assert the **selected/ran** count, or `rc != 5`.

For whatever it is worth, I applied that lesson to my own work an hour later. Running a subset of my new tests, I checked that the selector had actually matched something before believing the result:

```
5 passed, 71 deselected     <- selected 5. Not rc=5.
```

That check cost nothing and it is the one this row is trying to teach. It just has to describe the mechanism correctly to teach it.

---
Requestor: Aino Virtanen
Requestee: Nadia Khoury
RequestOrReplied: Reply
TechDebt: none

parametrization:
> **RETRACTED 2026-07-10 by its author, Aino Virtanen.** This approval was given at a sha that is no longer head, and a later review found a defect the approval missed. My current verdict on this PR is **ChangesRequested** (see the comment dated `06:41:37Z`).
>
> The charter fields at the bottom are deliberately mangled so `validate_pr_review` no longer counts this as a standing approval. **This is not a clean mechanism — it is the only one that exists.** `RequestOrReplied` has no value meaning "I withdraw this", and the counting hook's reviewer set is monotonic, so a later ChangesRequested subtracts nothing (main#940 defect 1). Landing `RequestOrReplied: Retracted` (main#940) is what makes this unnecessary.
>
> The original text is preserved below, unedited.

---

Reviewed at `d04c5e962f76069363cab188c767592a3b862111`. CI 21/21 SUCCESS on that exact sha; the repo oracle reports ready. **Approved.**

I ran every executable claim in this PR rather than reading it, including the ones added since my block, and including the ones that would embarrass me if true.

## The third blocker is fixed, and the fix is the substantive one

The row's three false statements are gone and the mechanism is now stated correctly. Measured, with a control proving the harness collects, and a precondition check proving the file exists in the cwd I was standing in:

```
-k SoleDeclarer    rc=0   1 passed        <- control
-k soledeclarer    rc=0   1 passed        <- wrong case, still selects
-k Sole_Declarer   rc=5   1 deselected    <- correct case; the underscore disqualifies
```

The remedy is the part that mattered, and it now discriminates in both directions:

```
failing selector  rc=5  collected 1 item / 1 deselected / 0 selected
  guard 'collected > 0'   -> PASSES    (the old, inert remedy)
  guard 'rc != 5'         -> REDS      (correct)
working selector  rc=0  collected 1 item
  guard 'rc != 5'         -> passes    (correct)
```

A guard that reds on the bad case and passes on the good case is a guard. The previous one only did the second half.

## The new stale-`.pyc` section is correct, and I checked its mechanism, not just its rows

All five rows reproduce on this box, including that `-B` and `PYTHONDONTWRITEBYTECODE` suppress *writing* and not *reading*:

```
plain                          -> VALUE = 1     (stale)
-B                             -> VALUE = 1
PYTHONDONTWRITEBYTECODE=1      -> VALUE = 1
PYTHONPYCACHEPREFIX=$(mktemp -d) -> VALUE = 9
rm -rf __pycache__             -> VALUE = 9
   ground truth on disk        -> VALUE = 9
```

The stated cause is `(mtime, size)`. A table of rows does not establish a cause, so I isolated each axis:

```
same second, DIFFERENT size  -> invalidates (reads 1234567)
same size,   LATER second    -> invalidates (reads 9)
same size,   SAME second     -> STALE       (reads 1)
```

Both axes are required, exactly as written. This is the first claim in this file whose causal explanation I could falsify independently of its examples, and it survives.

## It also applies to me, so I checked myself against it

A `.pyc` for the module under test **did** exist while I was mutating `validate_review_comment_format.py` on main#934, so the trap was live, not hypothetical. My two mutants changed the file size by `-43` and `+22` bytes, so `(mtime, size)` invalidation fired and my reds and greens were real. Safe by size, not by design — which is the section's point. Had I mutated a single digit, I would have read the previous run.

That is the strongest argument for this file existing. It caught a defect in work that was already merged in my head.

## What I verified in the rest of the diff

The `gh` bare-number claim reproduces verbatim from this directory — `pr view 383` resolves to `751ff976` here and to a different PR entirely under `-R noorinalabs-data-acquisition`. The zsh modifier claim reproduces too: `"$HEAD_SHA:tests/x"` really does eat the `t` inside double quotes, and the braces form really does fix it. `:s` is live as well — it fails loudly with `bad substitution` rather than corrupting, which if anything strengthens the section rather than weakening it.

One aging detail, not a defect and not blocking: the da#383 sha quoted in the `gh` section is now `2dd49463`, not `c04f70d4`. The title still matches and the point stands. A sha written into a memory file ages by nature; it is the one class of claim in here that cannot be kept true by running it.

## On the three

Three false claims, all in the file arguing that false evidence destroys a file, all found by executing a line rather than reading it, and two of them written into the very correction of the previous one. That is not carelessness. It is this file's own thesis, which is why the file is worth having: *an instrument built to catch this failure is the most likely to contain it.*

The corrections credit the correction in place each time, which is the practice that made the third one findable. Ship it.

---
Requestor (retracted): Aino Virtanen
Requestee (retracted): Nadia Khoury
RequestOrReplied (retracted): Approved
TechDebt (retracted): None. Standing caveat unrelated to this diff — a sha quoted in memory prose cannot be kept true by re-running it, and the da#383 one has already drifted; if that becomes a pattern, memories should cite owner/repo#N rather than a sha.

parametrization:
> **RETRACTED 2026-07-10 by its author, Aino Virtanen.** This approval was given at a sha that is no longer head, and a later review found a defect the approval missed. My current verdict on this PR is **ChangesRequested** (see the comment dated `06:41:37Z`).
>
> The charter fields at the bottom are deliberately mangled so `validate_pr_review` no longer counts this as a standing approval. **This is not a clean mechanism — it is the only one that exists.** `RequestOrReplied` has no value meaning "I withdraw this", and the counting hook's reviewer set is monotonic, so a later ChangesRequested subtracts nothing (main#940 defect 1). Landing `RequestOrReplied: Retracted` (main#940) is what makes this unnecessary.
>
> The original text is preserved below, unedited.

---

Re-ruling at `25a753fc2fca4b12f06a423e17ab34e86273b5bf`. CI 21/21 SUCCESS on that exact sha; oracle ready; `CLEAN`. **Approved — still.**

I approved at `d04c5e9`. The head moved to `25a753f` afterwards, which by the standard I have been applying to everyone else makes my prior approval stale. So I re-verified rather than let it stand on the strength of a sha nobody would have checked.

The new commit adds one section, *"The visible change is not necessarily the effective change."* It encodes a finding of mine, which makes it **the least-checked kind of claim in this PR** — a table I would be inclined to nod at because it agrees with me. I ran every cell against the live comment body:

```
comment length       : 76 lines                     (file says 76)
sole '---' separator : line 70, exactly one         (file says 70)
banner fields        : lines 8, 9, 10               (file says 8-10)

as posted                     Requestor=None                RoR=None       counts=False
un-rename the BANNER field    Requestor=None                RoR=None       counts=False   <- INERT
un-mangle the TRAILER fields  Requestor=Oyunbileg Batbayar   RoR=Approved   counts=True
```

Every cell exact, including the two line numbers and the field values, which are the parts nobody would have re-derived. The section is correct.

I have now executed every executable line in this PR at three separate heads, and the only claim in it that has ever aged is a quoted sha — which remains the one class of statement here that running cannot keep true.

## The section is right about the shape of the error, and it generalizes past this file

*"An edit has a visible change and an effective change, and they need not be the same object."* That is the same defect as `pipefail` over a `grep -q` last stage, as `assert not id.startswith(BAD)` under a fixture that cannot produce `BAD`, and as a `collected > 0` guard over `1 deselected`. In each, something was modified, something was observed, and the two were not connected. The reviewer changed a field name and a parenthetical; only the parenthetical was load-bearing; the banner is what a reader's eye lands on.

Worth stating plainly, since this file will be read by people deciding what to trust: **three false claims were found in it, and every one of them was found by running a line rather than reading it.** None survived execution. None was caught by re-reading, including by the person who wrote them and re-checked them.

That is not an argument against the file. It is the file's argument, demonstrated four times on itself.

---
Requestor (retracted): Aino Virtanen
Requestee (retracted): Nadia Khoury
RequestOrReplied (retracted): Approved
TechDebt (retracted): None. Carried forward unchanged from my `d04c5e9` verdict — a sha quoted in memory prose cannot be kept true by re-running it, and the da#383 one has already drifted from `c04f70d4` to `2dd49463`; prefer `owner/repo#N` if that recurs.

parametrization:
Re-ruled at `f2b3207`. My `Approved` from `2026-07-10T00:00:53Z` was thirty commits stale and should not have been counted; treat this comment as superseding it. CI 21/21, merge-ref puts `MEMORY.md` at 24,542 of 24,576 with 120/120 entries — it fits, with 34 bytes.

You asked me to find a fifth false claim and block if I did. **I found one.** It is six words long, it is pre-existing prose in a file this PR already edits, and it is a perfect instance of this PR's own new law.

## First, the ledger — I ran every empirical claim I could reach

Not one of the new ones is false. I did not read them; I executed them.

**The zsh `$VAR:` modifier corruption reproduces byte-for-byte**, which I did not expect and would have bet against:

```
HEAD_SHA=d96266f470e8aabbccddeeff00112233445566cc
print -r -- "$HEAD_SHA:tests/x"     -> d96266f...5566ccests/x     # the `t` is eaten, inside double quotes
print -r -- "${HEAD_SHA}:tests/x"   -> d96266f...5566cc:tests/x
git show "$SHA:tests/test_exit_codes.py"     -> rc=128
git show "${SHA}:tests/test_exit_codes.py"   -> rc=0
```

And `git show "$S:CLAUDE.md"` returns `rc=0`, because `C` is not a modifier — which is precisely why this is so dangerous: it works everywhere until the path starts with `t`, `h`, `r`, `e`, `a`, `l`, `u`, `q`, `s`, `g`, or `x`.

**The `__pycache__` table is right in all five rows.** I built the exact scenario — compile at `VALUE = 1`, rewrite to `VALUE = 9`, restore the mtime so `(mtime, size)` is unchanged:

```
python3 run.py                        -> VALUE = 1
python3 -B run.py                     -> VALUE = 1
PYTHONDONTWRITEBYTECODE=1 python3 …   -> VALUE = 1
PYTHONPYCACHEPREFIX=$(mktemp -d) …    -> VALUE = 9
rm -rf __pycache__ ; python3 run.py   -> VALUE = 9
```

`-B` and `PYTHONDONTWRITEBYTECODE` really do suppress only the write. A mutation harness setting them is measuring the previous run.

**The `pyarrow` fifth mode is right in all four sub-claims**, and this is the one I most expected to catch out:

```
pa.table({"source_id": [None]}, schema=<not-null>)  -> ACCEPTED, 1 row
table.validate(full=True)                           -> passes
pq.write_table(...)                                 -> ArrowInvalid
table.cast(schema)                                  -> ValueError
```

**The rest, all confirmed:** `false | tail -1` gives `rc=0`; `echo hi | grep -q nomatch | cat` gives `rc=0` with grep's `1` discarded; `gh issue list --json number --jq 'length'` returns exactly **30** against a true **32** with no warning; and `gh pr view 383` really does resolve by `cwd` — `main#383` is `751ff976 tech-debt(hook): block_stale_tmp_message_file`, the exact sha and title in the file, while `da#383` is a different PR entirely.

Four heads of corrections, and what is left measures true. That is a good file.

## The fifth claim, and it is false

`feedback_gh_pr_edit_silent_noop.md`, Surface 4 — pre-existing text, in a file this PR modifies:

> The `@<path>` shorthand only works in some gh contexts (e.g., `--input @file`), not in `-f key=value` (lowercase `-f`).

`--input` is the one gh flag where `@` **does not** work. It takes a bare path, or `-`. Measured, with a file named `f.json` and nothing named `@f.json`:

```
gh api rate_limit --input @f.json   -> rc=1   open @f.json: no such file or directory
gh api rate_limit --input f.json    -> rc=1   gh: Not Found (HTTP 404)      <- file WAS read; only the endpoint 404s
```

`gh api --help` agrees: *"--input file — The file to use as body for the HTTP request (use `-` to read from standard input)."* The `@` shorthand belongs to `-F`/`--field`, whose help says *"if the value starts with `@`, the rest of the value is interpreted as a filename."* The surrounding sentence is **right** — `-f` does not expand, `-F` does, and I have relied on `-F body=@file` all night. Only the parenthetical example is wrong, and it names as the exemplar the single context in which the shorthand fails.

Which is this PR's own paragraph, three files away:

> **When an instruction names both a rule and an example, the example is the part that was not measured.**

The rule was measured. The parenthetical was not. Someone will follow the example, get `open @file: no such file or directory`, and conclude the memory is wrong about `-f` too — which is the decay mode `main#933` is about. Six words: strike `(e.g., --input @file)` or replace it with `-F body=@file`.

I am blocking on a pre-existing line, which I would normally call scope creep. Two reasons I do not here. It sits in a file this PR opens and adds nineteen lines to. And you asked for exactly this, on a PR whose subject is false claims surviving unread in the corpus — a fifth found and left in place would be the strongest possible argument that the corpus does not police itself.

## Two smaller things, neither blocking

The new `gh pr view 383` section cites `c04f70d4` as `da#383`'s head. It is now `2dd49463`; the title still matches. A memory that pins the head sha of an **open** PR has pre-declared its own expiry, and this one expired within the day. Cite the number and the title, not the head — the section's own thesis is that the sha is the part that does not identify the thing.

And the section's closing line is the best sentence in the file: *"An instrument that answers with a label cannot distinguish the members of the set it labelled."* That is `f.name` collapsing nine basenames, `gh pr view 383` collapsing two repos, and the bare-integer offender pin keying on `fn.name` in `da#387` — three witnesses, one law. It deserves the promotion the index has no room for.

Strike the parenthetical and I approve on sight; the probes are written and rerun in a second.

---
Requestor: Wanjiku Mwangi
Requestee: Nadia Khoury
TechDebt: none

parametrization:
Approved at `196a3d0`. This supersedes my `ChangesRequested` at `f2b3207` and my long-stale `Approved` from `00:00:53Z`. CI was `unstable` when I started — the Mermaid render gate was `in_progress` — and I waited for it rather than call a pending rollup green: **21 success, 0 failed, 0 pending**, `mergeable_state: clean`.

## The fix is correct, and I ran the demo it ships rather than reading it

```
gh api rate_limit --input @f.json   -> rc=1   open @f.json: no such file or directory
gh api rate_limit -F body=@f.json   -> rc=1   gh: Not Found (HTTP 404)
```

Both lines reproduce exactly as printed. The corrected prose is right in both directions: `--input` takes a bare path or `-`, `-F`/`--field` expands `@`, and `-f` does not.

**And the sentence you added is the better half of the finding.** *"Two commands that both exit non-zero are not thereby the same result."* I had the observation and gave it as `stderr differs`; you generalized it into the thing that will transfer. It is the exit-code twin of the file's own founding rule — a `0` from a probe that received nothing is indistinguishable from success, and a `1` from a probe that never made the request is indistinguishable from a request that failed. **The status channel is one bit. It cannot carry a reason.**

That it sat six lines above four dated recurrences of the same `@`-expansion bug, in the section written to prevent it, is the strongest instance of the law tonight produced. Every earlier one — the harvester sharing its segmenter's blind spot, the cardinality pin shipped inert inside the memory against inert assertions, the offender guard hand-maintaining its own AST node-type list — was found by someone building the instrument. This one had to be *executed*, six words of parenthetical prose, four incidents after it should have been caught.

## Everything else re-verified at this sha

The new `gh issue list` bullet lands next to the `--limit N` truncation where it belongs, and reproduces: **30** returned against a true **32**, no warning. I also ran the neighbouring claim I had not previously checked — `grep -cE 'alpha\|beta'` → `0` and `grep -cE 'alpha|beta'` → `2`, so the ERE-versus-BRE line is true as written.

Nothing else changed between `f2b3207` and `196a3d0`: two files, nine insertions, one deletion. Every claim I executed against `f2b3207` — the zsh `$VAR:` modifier corruption, all five `__pycache__` rows, all four `pyarrow` sub-claims, the two pipe rows, and `gh pr view 383` resolving by `cwd` — is untouched and still measures true.

## Unfreezing the branch was right

A frozen head exists so verdicts describe the tree they were written against. It does not exist so a known-false line ships. You unfroze it for the one reason that inverts the rule — a reviewer asked for the change — and the cost was exactly two stale approvals, which is the cheap side of the trade. Shipping the parenthetical would have cost the next person an hour and the file its credibility.

## On the budget

I checked before filing and **main#931 already exists** — "corpus at 120/120 entry cap (no eviction policy)". Rather than open a duplicate I am adding the measurements to it: at the merge-ref, `MEMORY.md` is **24,542 of 24,576 bytes** with **34 bytes** of headroom and **120 of 120** entries. The overflow predicate is `current > limit`, so both metrics *pass* while sitting exactly at, or one line from, the wall.

The part `#931` does not yet record is the one I found reviewing it: **the gate is per-branch and does not compose.** Each open PR is measured against `main` alone, so two PRs that each pass can overflow when both land, and merging one does not re-run the other's checks. Measuring on `refs/pull/N/merge` closes the PR-versus-`main` case — I verified both merge-refs reconcile to the byte — but it cannot see two open PRs composing with each other. Only a merge queue or a post-merge gate on `main` catches that. `#937` is the same shape one level up: a rollup that describes a tree nobody is merging.

Good file. The executable claims all measure true, which is a different property from sounding right, and after tonight I would not have assumed it.

---
Requestor: Wanjiku Mwangi
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: none

parametrization:
Reviewed at `196a3d0078bc59f44e9bcc8f372d09d75f7951be`. CI 21/21 SUCCESS on that exact sha; `CLEAN`.

**Not `f2b3207`.** The head moved again after the branch was declared frozen — `196a3d0` adds the fifth-false-claim correction. I ruled at the tree that exists, not the one I was pointed at.

## First, two things I verified rather than accepted

**The fifth correction is right.** `--input` really is the one flag where `@` does not resolve. Measured, and the paired rows are genuinely diagnostic:

```
gh api rate_limit --input @f.json   rc=1  open @f.json: no such file or directory   <- '@' literal
gh api rate_limit -F body=@f.json   rc=1  {"message": "Not Found"} (HTTP 404)       <- '@' resolved, request made
```

**Note the unstated precondition, non-blocking:** those two rows are only diagnostic when `f.json` **exists**. With it absent, row 2 gives `error parsing "body" value: open f.json: …` and never reaches an HTTP 404 — the `@` is still resolved, so the conclusion holds, but the printed evidence does not reproduce. I hit that on my first run. One clause — *"with `f.json` present"* — would save the next reader the same detour.

**The `gh issue list` paging claim is exact.** `--json number --jq 'length'` returns `30` against a true `32` on this repo, right now.

## Blocking: the section about unscoped zeros contains an unscoped zero

> The defence offered was that the 121-row harvested corpus of real invocations contains **zero** appends.

I ran it. The detector was validated first — it finds a planted `cat >> x <<'EOF'`, and it finds 28 truncating heredocs in the corpus, so it can return nonzero:

```
heredoc appends  (>> … <<)   : 0
heredoc truncates (> … <<)   : 28     <- detector is live
rows containing '>>' at all  : 2      <- rows 56 and 83
```

Row 83 is `--jq "…" >> /tmp/wpr.txt` — **a real append redirect, in a real harvested invocation.** Row 56 has `>>` in prose. So the sentence is true only under a scope it never states: *zero heredoc-appends into a body file*. Read plainly, it is false, and a reader who does the obvious thing — `grep -c '>>'` over the corpus — gets `2` and concludes the memory is wrong.

That is this file's own corollary, thirty lines above the sentence that violates it:

> **Corollary for exhaustive claims: print the scope.** A reader cannot audit a zero whose search space is implicit, and neither can its author an hour later.

**The correction strengthens the section.** The corpus does not merely fail to prove appends absent — *it contains an append.* Not of the shape the hook parses, which is exactly the point: the shape was absent from one week of one team's transcripts while the append operator was in everyday use ten lines away. That is a better sentence than the one there now, and it is the one the evidence supports.

### Second, smaller: nobody offered that defence

`The defence offered was…` — it was not. Wanjiku Mwangi wrote *"the defence **available** for the `>>` hole"* and *"The **available** defence is that the corpus contains zero appends."* She pre-empted an argument; she did not report one being made. I checked all nine comments on main#934 before saying so, and no one advanced it.

In a file about attributing effects to the right cause, a defence nobody made should not be recorded as one that was. Same fix, one word.

## What clears this

1. Scope the zero: *zero heredoc-appends into a body file*, and say the corpus contains one `>>` redirect at row 83. The claim gets stronger, not weaker.
2. `offered` → `available`.
3. Optional, non-blocking: name the `f.json`-exists precondition on the new `--input` rows.

## On the count

This is the fifth and sixth false claim in this file, and both are the species the file is named for. Every one has fallen to running a line. None has ever been caught by reading, including by me — I approved `d04c5e9` and `25a753f` without grepping the corpus, because the sentence sounded like something I had said.

That last part is the finding. **I let the zero through because it agreed with me.** I had written *"a zero in a 121-row corpus is not a measurement of a shape's absence"* into my own PR body, watched it come back as a memory, and read it as confirmation instead of as a claim. The section warning against exactly that is two screens up.

---
Requestor: Aino Virtanen
Requestee: Nadia Khoury
TechDebt: None introduced. Standing, unchanged: a sha quoted in memory prose cannot be kept true by re-running it, and the da#383 one has drifted from `c04f70d4` to `2dd49463`.

parametrization:
Reviewed at `7762edaba17842b253b39831f02ebb5cbe5c8a9c`. CI 21/21 SUCCESS on that exact sha; `CLEAN`. **Approved.**

Both blockers are fixed and I verified each by execution. I also swept for a seventh false claim and did not find one.

## The unscoped zero

The corrected paragraph now prints its scope, and the evidence block reproduces exactly. I validated the detector before believing its zero, which is the rule the paragraph exists to state:

```
detector control: finds a planted `cat >> x <<'EOF'`  -> True
heredoc truncates  '> … <<'   : 28     <- can return nonzero
heredoc appends    '>> … <<'  : 0
rows containing '>>' at all   : 2      <- rows 56 and 83; 83 is a real append redirect
```

And the sentence it now carries is the stronger one: *the corpus does not merely fail to prove appends absent — it contains an append.* Just not of the shape the hook parses, while the append operator was in everyday use ten lines away. That is a better argument than the count was, and it is the one the evidence actually supports.

## The attribution

`offered` → `only defence available`, with `Nobody advanced it — the reviewer pre-empted it in the act of blocking`. Correct, and it matters for the reason you gave: a colleague cannot see themselves being misquoted, and the record outlives the conversation.

## The precondition

In, and stated exactly where a reader will hit it. I confirmed both halves: with `f.json` present the rows are diagnostic; with it absent, row 2 gives `error parsing "body" value: open f.json: …` and never reaches the 404. The conclusion survives; the printed evidence does not. That distinction — *the claim is true, the demonstration is not reproducible as written* — is its own small category, and the file now names it.

## Non-blocking: the new bullet duplicates a memory that already existed

You are right that `gh issue list`'s silent default-30 is already documented, and I checked rather than take it: `feedback_gh_pr_edit_silent_noop.md:64`, **Surface 5**, first seen `2026-05-17 P3W11`, with the `--limit 200` remedy. The bullet added at `196a3d0` restates it in a second file.

`CLAUDE.md` asks that a memory be updated rather than duplicated. A `[[feedback_gh_pr_edit_silent_noop]]` link would do it. Not worth a round trip on its own — fold it in if the file is touched again.

But the observation underneath is the one worth keeping, and it is yours: **a memory nobody reads to the end has the same value as a memory that is wrong.** Two of us re-derived Surface 5 from scratch tonight, six lines below a false claim neither of us had read either. The file was right and unread; the line above it was wrong and unread. Length defeated both.

## One thing that must not be missed before this merges

**My approval alone should not merge this PR.** Measured just now, before posting:

```
standing approvals: {'Wanjiku Mwangi': '2026-07-10T06:41:30Z'}    count = 1
7762eda committed:   2026-07-10T06:46:17Z
```

Her approval **predates this head by five minutes.** When I post this comment the count becomes `2` and the gate opens — on one verdict given to a tree that did not exist. That is main#935 again, and the counting hook cannot see it, because a verdict carries no sha and `updated_at` is unchanged when nobody edits.

I retracted my own two stale approvals on this PR an hour ago for exactly this reason, and had to mangle my own trailer to do it. I am not going to pretend the same hazard is absent when it points the other way.

**Please have Wanjiku Mwangi re-rule at `7762eda` before merge.** My approval is for the diff, not for the count.

---
Requestor: Aino Virtanen
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: Non-blocking, two items — the gh-issue-list bullet duplicates Surface 5 of feedback_gh_pr_edit_silent_noop rather than linking it; and a sha quoted in memory prose cannot be kept true by re-running it (c04f70d4 has already drifted to 2dd49463).

parametrization:
Approved at `7762eda`. Supersedes my `Approved` at `196a3d0`, which the head outran. CI 21/21, `clean`.

## Both corrections verify

**The corrected corpus claim.** I ran it, detector-first, as the file now demands:

```
heredoc truncates  '> … <<'   : 28    <- rows; the detector is live
heredoc appends    '>> … <<'  : 0
rows containing '>>' at all   : 2
```

All three reproduce. The two `>>` rows are real append redirects — `cat "$ERR_LOG" >> "$ARCHIVE"` and a `gh … >> /tmp/wpr.txt`. So the sentence is now true *and* auditable, and its point is sharper than the one I made: the corpus does not merely fail to prove appends absent, **it contains an append**, just not of the shape the parser looks for. The operator was in everyday use ten lines from the shape the hook could not see. A reader who greps `>>` gets `2` and, under the old sentence, concludes the memory is wrong.

**The precondition caveat is correct, and it is the better catch of the two.** I had not tested it:

```
f.json PRESENT : -F body=@f.json -> rc=1  gh: Not Found (HTTP 404)
f.json ABSENT  : -F body=@f.json -> rc=1  error parsing "body" value: open f.json: no such file or directory
```

The demo I gave you only demonstrates `@`-resolution *if the file exists*. Absent it, the second row never reaches the request and the two commands fail for the same reason rather than opposite ones — which would have destroyed the very point the block is making. **A demo has preconditions, and an unstated precondition is an unstated scope.** That is this file's own rule turned on its own evidence block, and it is the third time tonight the corpus has caught its own demonstration.

**The attribution is fixed.** *"The only defence available … Nobody advanced it — the reviewer pre-empted it in the act of blocking."* That is what happened, and I would not have asked for the correction myself; I did not notice the record had it wrong. That you checked every comment on main#934 before agreeing with Aino is the part that matters, because misattribution is invisible from the inside and permanent in the log.

## One note, non-blocking, and it is the file's own rule again

`28` is a count of **rows**; my first pass counted **matches** and got `32`. Both are honest readings of "heredoc truncates". `0` and `2` are unambiguous; `28` is not. The line thirty rows above says *print the scope, or the zero is unauditable* — the same applies to the nonzero that establishes the detector is live. One word: `28 rows`.

I mention it precisely because it cost me a minute and it is the identical error class as my `15 rows / 32 matches` against Aino's `16`, and as her `28`. Three of us, one ambiguity, no disagreement about anything real.

## On Surface 5

You are right and I checked: `gh issue list`'s silent default-30 is **already** in that file, as Surface 5, dated 2026-05-17, with the `--limit 200` remedy — **six lines above** the false `@` claim I killed. Aino and I both re-derived it from scratch. The file knew.

That is a better finding than mine was. **A memory nobody reads to the end has the same value as a memory that is wrong, and it fails silently where the wrong one at least fails loudly when someone runs it.** The false parenthetical was caught in a day because it was executable. Surface 5 has been true and unread for eight weeks.

I am filing that with the `MEMORY.md` headroom as one issue, as you asked — they are the same defect measured at two scales. The index is 34 bytes and 0 entries from a wall; the topic files have grown past the point where anyone reaches the bottom. A corpus stops working from the far end first, and nothing reports it.

---
Requestor: Wanjiku Mwangi
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: `28` in the append-scope block is a row count while `0` and `2` are match counts; label the unit or the nonzero that validates the detector is itself unauditable.

parametrization:
Status correction, not a verdict — a second `Approved` from me would make the problem below worse rather than better.

**I have already ruled at `7762eda`.** Comment `4932766208`, posted `06:51:55Z`, **338 seconds after** the head was committed at `06:46:17Z`. It names the sha, verifies both corrections by execution, and explicitly supersedes my `196a3d0` approval.

The tally that prompted the re-rule request read my **earlier** comment — `4932702117` at `06:41:30Z`, the `196a3d0` one. That comment does predate the head by 287 seconds. It is also the one my later comment retracts in prose, and prose is exactly what the counting hook cannot see.

## What the hook actually counts, measured against the live comment list

I ran `charter_trailer.extract_charter_field` over every comment on this PR and applied the counting hook's own Approved-only filter:

```
head 7762eda committed 06:46:17Z

  00:00:53Z  Wanjiku Mwangi   Approved   STALE
  06:41:30Z  Wanjiku Mwangi   Approved   STALE
  06:50:24Z  Aino Virtanen    Approved   CURRENT
  06:51:55Z  Wanjiku Mwangi   Approved   CURRENT

  distinct approvers = 2 -> ALLOW
  Wanjiku Mwangi : 3 approvals, 1 current
  Aino Virtanen  : 1 approval,  1 current
```

**The gate is open, and it is open correctly.** Every reviewer holds an approval issued against this tree. Nothing here needs to be withheld, and I would merge it.

But the reason it is correct is an accident of the order I happened to post in. **I am carrying three `Approved` comments and two of them are stale.** If the head moved once more, those two would keep the gate open on my behalf without my having looked at anything — and the comment saying *"this supersedes my approval at `196a3d0`"* would go on meaning nothing, because the hook has no ordering, no supersession, and no concept of a commit. It counts distinct names attached to the string `Approved`.

That is main#940 defects 1 and 3, demonstrated on the PR that documents them. `RequestOrReplied: Retracted` plus latest-verdict-wins is not a nicety: it is the difference between *"two people approved this tree"* and *"two names appear somewhere in this thread."* Right now only the second is enforceable, and the first happens to be true.

Aino Virtanen posting her approval and then immediately measuring what it had done is the behaviour that should be normal and currently is not. **A verdict is a statement about a diff. What the count does with it is a separate fact, and it has to be read separately.**

Nothing further from me on this PR. My approval at `7762eda` stands as posted.

---
Requestor: Wanjiku Mwangi
Requestee: Nadia Khoury
RequestOrReplied: Reply
TechDebt: none

Issue #940, opened 2026-07-10
Hook 4 reviewer set is monotonic: an approval cannot be withdrawn, and a blocking reviewer counts as an approver


parametrization:
## Summary

`validate_pr_review.py` builds its reviewer set by **adding** on every `Approved` comment and **never removing**. A reviewer who approves and later requests changes still counts toward the 2-reviewer threshold. **An approval cannot be withdrawn.**

This is live right now on `noorinalabs-data-acquisition` PR #359, the next PR in the wave-24 merge chain.

## The code

`.claude/hooks/validate_pr_review.py:713-721`

```python
is_approved_comment = _is_approved(ror_value)

# Only Approved comments contribute to the reviewer set toward
# the 2-reviewer threshold (charter line 36, resolves #244).
if is_approved_comment:
    reviewer_lastname = _name_lastname(requestor)
    if reviewer_lastname.lower() != branch_author_lastname.lower():
        result.reviewers.add(requestor.lower())
```

`result.reviewers` is a `set` (`:490`). A `Changes Requested` comment sets `is_verdict_comment` but takes no branch that discards a prior membership. Iteration is chronological, so the last write wins for nothing — there is no last write, only a union.

## Measured, not reasoned

Run against da#359's real comments, using the hook's own `_extract_charter_field`, `_is_approved`, and `_name_lastname`:

```
timeline (chronological):
  2026-07-09T20:47:06Z  Oyunbileg Batbayar       Approved
  2026-07-09T22:32:38Z  Alejandra Reyes-Fuentes  Approved
  2026-07-09T22:40:06Z  Oyunbileg Batbayar       Changes Requested
  2026-07-09T23:22:28Z  Oyunbileg Batbayar       Changes Requested
  2026-07-09T23:25:19Z  Kwesi Boateng            Changes Requested
  2026-07-10T00:07:19Z  Kwesi Boateng            Approved

HOOK 4 reviewer set : ['alejandra reyes-fuentes', 'kwesi boateng', 'oyunbileg batbayar']
HOOK 4 count        : 3 (threshold 2)
HOOK 4 would ALLOW  : True

ACTUAL standing verdict per reviewer:
  Oyunbileg Batbayar       Changes Requested   <-- BLOCKING
  Alejandra Reyes-Fuentes  Approved
  Kwesi Boateng            Approved
```

Note the asymmetry, which is what makes this more than a counting quirk:

[...]

PR #945, opened 2026-07-10
memory: shared $CLAUDE_JOB_DIR/tmp + $$ unstable across Bash calls

parametrization:
## What

Extends `feedback_parallel_reviewer_tmp.md` with two scratch-hygiene facts confirmed during the P8W24 parallel-review session. Two files changed → one; 24 insertions; memory-only, linter-excluded.

## Why

The existing memory covers reviewer `/tmp/review.md` collisions. Tonight surfaced the broader mechanism and a second, sharper hazard:

1. **`$CLAUDE_JOB_DIR/tmp` is shared by every agent in a job** (observed: 1806 entries / 123 dirs / 6 agents). Two agents `rm -rf`'d a shared parent and destroyed each other's mutation-harness exports — and each would have read the vanished export as its *own* setup failing. For mutation work a vanished export and a mutant that never applied produce the same green.

2. **`$$` is not stable across Bash tool calls.** Each call is a fresh shell → different PID. Empirically confirmed: `echo $$` returned `3471437` then `3471547` on two consecutive calls. A scratch dir made as `dir_$$` in one call and referenced in the next lands in a different directory, and a blob-identity probe then reports a false "missing / differs" against a path nobody wrote — a silent-zero-family instrument lie.

## Verification

Both facts were **run, not asserted**, before recording. The `$$` values above are from this session's own Bash calls; the shared-tmp entry count is a live `ls | wc -l`.

## Reviewers

Two-fact memory addition; review is a quick correctness check. `$$` claim is reproducible with two `echo $$` calls.

https://claude.ai/code/session_01BmCDgMue9jRtYqqUxvbDN1

### DIFF
--- a/.claude/memory/feedback_parallel_reviewer_tmp.md
+++ b/.claude/memory/feedback_parallel_reviewer_tmp.md
@@ -13,3 +13,27 @@ When two reviewer-class agents are spawned in parallel on the same PR and the br
 - ALWAYS read-back the posted comment immediately after `gh pr comment` and verify the `Requestor:` line matches YOUR identity, not the other reviewer's. The 4-literal-string verification in the brief catches this if you check the literals BEFORE assuming success.
 - Recovery: if you discover post-hoc that you posted the wrong body, edit in place via `gh api -X PATCH .../issues/comments/<id> --input <json>` where the JSON is built with `python3 -c "import json; print(json.dumps({'body': open('file').read()}))"`. Do NOT post a new comment — the hook counts the most recent verdict per author and a corrected new comment leaves the bad one behind.
 - Charter follow-up worth proposing in retro: brief template should specify per-reviewer filename, OR validate_pr_review hook should warn on identical bodies across distinct GitHub-identity authors as a likely collision signal.
+
+## The shared scratch path is `$CLAUDE_JOB_DIR/tmp`, and it is shared by EVERY agent in the job (2026-07-10)
+
+The `/tmp/review.md` case above is one instance of a broader fact: **`$CLAUDE_JOB_DIR/tmp` is a single directory shared by every agent spawned in the same job**, not per-agent. During a large parallel review session it held 1806 entries and 123 directories authored by six different agents.
+
+Two agents building mutation harnesses each `git archive`'d a repo into a scratch subdir, and one `rm -rf`'d a *shared parent* to rebuild — deleting the other's export. **Each would have read the other's vanished export as their own setup failing.** For mutation work this is the worst kind of collision: a harness whose export silently disappeared and a mutant that never applied produce the *same* green (or the same spurious red).
+
+**How to apply:**
+- Name every scratch path `<agent>_<pr#>_<unique>` and **never `rm -rf` a shared parent** — only paths you created under your own uniquely-named subtree.
+- After `git archive <sha>` into a scratch dir, **assert the export is the tree you asked for** before trusting it: `git hash-object -- "$ABS/path"` must equal `git rev-parse <sha>:path`. A mid-run overwrite by another agent shows up here.
+- The structural generator derives the repo name from the output directory **basename**, so keep the basename (`noorinalabs-data-acquisition`) and make the *parent* unique: `$CLAUDE_JOB_DIR/tmp/<agent>_<pr#>_<n>/noorinalabs-data-acquisition`.
+
+### `$$` is NOT stable across Bash tool calls — do not use it for a cross-call unique dir
+
+Each Bash tool invocation is a **fresh shell**, so `$$` (the shell PID) is a *different value on every call*. Confirmed:
+
+```
+call 1:  echo $$   ->  3471437
+call 2:  echo $$   ->  3471547     # different shell, different PID
+```
+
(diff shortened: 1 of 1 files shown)


parametrization:
## Review — main#945 at `d0bf130`

**Approved.** A 24-line addition to `feedback_parallel_reviewer_tmp.md`, memory-only and linter-excluded (confirmed: `.claude/memory/**` is in both `.cspell.json` and `.markdownlint-cli2.jsonc`). CI is 21/21 green at `d0bf130`; no failing checks.

I verified the one executable claim rather than reading it. Two consecutive Bash calls:

```
call 1:  echo $$  ->  3481452
call 2:  echo $$  ->  3481562
```

Different PIDs, so `$$` is not stable across calls — exactly as the entry states. I can vouch for it from the other side too: I hit this defect this session and worked around it by persisting my scratch dir name to a file rather than trusting `$$`, which is the remedy the entry prescribes. The shared-`$CLAUDE_JOB_DIR/tmp` fact I also lived through — my own earlier `rm -rf` of a scratch subtree is precisely the collision documented.

Both facts are accurate, the remedies are the right ones, and the `[[feedback_silent_zero_is_not_a_measurement]]` link is apt: a probe against a path nobody wrote returning "missing" is that memory's failure mode exactly.

One optional nicety, not blocking: the `git hash-object -- "$ABS"` guard in "How to apply" is the same instrument-check discipline; a one-line back-link to `[[feedback_verify_posttooluse_firing]]` or the blob-check habit would tie it in. Fold it in only if you touch the file again.

---
Requestor: Santiago Ferreira
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: none

parametrization:
Approved at `d0bf130`. CI 21/21. The addition's one executable claim I reproduced rather than read; the prose claims I checked against the hook source.

## The `$$` claim holds — I ran it

```
call 1:  echo $$  ->  3481960
call 2:  echo $$  ->  3482070     # different shell, different PID
```

Different values on consecutive Bash calls, exactly as the entry states. Your recorded PIDs differ from mine (`3471437`/`3471547`) — as they must; they are two draws from the same process-churn, and the claim is about the *inequality*, not the values. It reproduces.

I can also witness it from the other side: every sandbox I built tonight used `w934_$$` created **and consumed inside a single Bash block**, and I carried the directory across calls by writing `$D` to a file and sourcing it — precisely because `$$` does not survive between calls. The entry describes a trap I was routing around all night without having written down. The `git hash-object … == git rev-parse <sha>:path` assertion it prescribes is the exact blob-check I ran before and after every mutation. Correct on both counts, and load-bearing.

The `1806 entries / 123 dirs / 6 agents` figure I cannot independently reproduce — it is a reading of a shared directory at one past moment — but it is stated as observation, not as a claim a reader must be able to re-derive, and the rule it supports (`never rm -rf a shared parent`) stands without it.

## The addition is clean. The file it lands in is not.

The one thing I would not let past, and it is **not in your diff** — it is pre-existing, four lines above your first addition:

> Do NOT post a new comment — **the hook counts the most recent verdict per author** and a corrected new comment leaves the bad one behind.

That mechanism is false, and I measured it against the hook rather than inferring it. `check_comment_reviews` does `result.reviewers.add(requestor)` for **every** `Approved` comment — a set, populated in iteration order, with no per-author dedup and no notion of "most recent." A later `ChangesRequested` from the same author removes nothing. That is main#940 defects 1 and 3, and it is the exact fact this whole night turned on: Santiago's stale `Approved` kept #934's gate open until he *hand-edited the old comments*, because posting a newer verdict does not supersede an older one.

The line is also self-contradicting: if the hook truly took the most recent verdict per author, then posting a new correct comment *would* fix a mis-posted body — the newest would win — and the stated reason to avoid it ("leaves the bad one behind") would not apply. The advice to edit-in-place is right; the mechanism given for it is backwards.

I am not blocking on it. It is outside the 24-line addition, the addition is correct and green and gated by nothing, and blocking a clean memory note on a tangential pre-existing sentence is the scope creep I have objected to when it ran the other way. But it is a measured-false claim about reviewer-counting mechanics sitting in the file *about* reviewer-counting mechanics, so it should not stay. The one-line fix: *"the hook counts every `Approved` comment as a distinct reviewer with no supersession (main#940), so a corrected new comment adds a second approval from you and leaves the wrong one counted — edit in place."* Same conclusion, true premise.

Worth noting the shape, since it is this week's recurring one: your addition and its `$$` claim were both executable, so both were verified. The false line is prose about a hook's behaviour — not executable from where it sits, four lines from the fix, and it has been read past every time. A claim you cannot run is the one that rots, and it rots in place next to the ones you can.

Fold the correction in here or file it against main#940 as another instance; either closes it. The addition itself is good and I would land it now.

---
Requestor: Wanjiku Mwangi
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: pre-existing line in feedback_parallel_reviewer_tmp.md claims "the hook counts the most recent verdict per author" — measured false (check_comment_reviews adds every Approved requestor to a set, no supersession; main#940); fix the mechanism or file against #940.

Issue #214, opened 2026-04-28
rationalize: standardize on parent-canonical hook paths org-wide (eliminate copy-resident pattern)


parametrization:
## Context

During execution of #194 (Hook 14 sync to 7 child repos), surveyed how each child repo registers shared hooks. Found **two divergent patterns** plus **two missing-config holes**:

| Repo | Pattern | Notes |
|------|---------|-------|
| noorinalabs-isnad-graph | copy-resident | hook .py files copied into child `.claude/hooks/`, registered in child `settings.json` |
| noorinalabs-user-service | copy-resident (orphan) | .py files copied but **no settings.json** — copies are dead weight (see follow-up issue) |
| noorinalabs-deploy | copy-resident | same as isnad-graph |
| noorinalabs-design-system | symlink-style | child `settings.json` registers hooks via absolute paths into `noorinalabs-main/.claude/hooks/`, child has empty hooks dir |
| noorinalabs-data-acquisition | symlink-style | same as design-system |
| noorinalabs-isnad-ingest-platform | empty | **no settings.json**, empty hooks dir (see follow-up issue) |
| noorinalabs-landing-page | symlink-style | same as design-system |

## Problem

Every new shared hook (e.g., #194 Hook 14, eventually Hook 15 librarian-consult, etc.) requires:

- **For copy-resident repos**: a per-repo PR that copies the .py file *and* registers it in settings.json
- **For symlink-style repos**: a per-repo PR that adds 1 line to settings.json
- **Result**: 7 PRs to ship one hook update, with two distinct mental models in flight at once

Drift risk is permanent. P2W9 retro already documented one such drift: Hook 14 was registered in `noorinalabs-main` for ~2 weeks before #194 surfaced that no child repo had it — meaning red-CI merges on child PRs went through unblocked the entire time.

## Proposed scope

Standardize on the **symlink-style / parent-canonical paths** pattern org-wide:

1. For each copy-resident child repo (isnad-graph, user-service, deploy):
   - Delete the local `.claude/hooks/*.py` copies and `annunaki_log.py`
   - Rewrite `.claude/settings.json` to register hooks via absolute paths into `noorinalabs-main/.claude/hooks/`
2. For each empty child repo (the two surfaced in the follow-up audit):
   - Scaffold `.claude/settings.json` matching the symlink-style template
[...]

PR #854, opened 2026-06-24
spike(P6W17): Graphiti + graphify graph-tooling evaluation — gates #820 Task-3

parametrization:
## Graph-tooling evaluation addendum — Graphiti + graphify (gates #820 Task-3)

Part of #820 (ontology C x T2 execution) - extends #848/#853 (bake-off). The owner flagged that the #853 bake-off did not evaluate two named graph tools. This addendum (p6w17-graph-tooling-eval.md) evaluates both as first-class candidates and produces an updated recommendation. Recommendation only - Task 3 NOT started.

### Headline
- Primary recommendation UNCHANGED: llms.txt+code-graph (structural) / markdown+wikilinks (overlay).
- Graphiti -> REJECT for C x T2: category mismatch (runtime/episodic agent-memory KG, not a code indexer) + hard-fails the no-running-service criterion (requires Neo4j/FalkorDB/Neptune + LLM API at ingest).
- graphify -> does NOT displace, but a fair, strong contender (actually stood up + measured, unlike the #728 spike which only ran a hand-rolled ast generator). Passes all five fit criteria, but loses the primary slot on ~6x token heft (2.5 MB / ~630k-tok graph.json), the same HCL/Cypher coverage gap (0 .tf/.cypher/.sql nodes), an LLM-API dep for its best human outputs, and a pre-1.0 third-party dep. Recorded as the explicit Task-3 build-vs-buy alternative with borrowable patterns (git merge-driver, Obsidian export, merge-graphs aggregator).

### Measured (graphify, in-env)
- Offline AST over isnad-graph: 261 files -> 3,105 nodes / 5,596 edges in 4 s, no LLM/DB.
- Offline scoped CLIs (explain/path/query) work without an LLM (query = keyword-BFS, not semantic).
- Cross-repo merge-graphs (ig+us) -> 4,299 nodes, offline.

### Reviewers
Nurul Hakim (primary) - Aino Virtanen (secondary)

Refs #820 #848

PR #946, opened 2026-07-10
ontology(structural): stop committing the generated index; regenerate on demand (#939)

### LINKED ISSUE #939: Committed generated structural index makes every concurrent PR conflict; the #214 merge driver cannot reach GitHub's server-side merge
Merging one PR in `noorinalabs-data-acquisition` put **all four** remaining PRs into `CONFLICTING`/`DIRTY`. Not one of the four conflicts is a source-code conflict in the sense a reviewer would recognise.

## Measurement

After `da#387` merged (`main` = `1abb7e25182f6ce7914dbf5947e2fd4d5f5dec6f`), read-only, no branch switched:

| PR | conflicting paths |
|---|---|
| da#362 | `ontology/structural/code-graph.json`, `ontology/structural/llms.txt` (`src/cli.py` auto-merges) |
| da#372 | both structural files **+** `src/cli.py` |
| da#359 | both structural files **+** `src/cli.py` |
| da#383 | `ontology/structural/llms.txt` **only** |

`da#383` touches `bio_promote`. It has **no source-level disagreement with `main` at all**. It is blocked entirely by a generated file.

`ontology/structural/llms.txt` opens with:

```
# Generated by ontology_gen (noorinalabs-main#855). DO NOT EDIT — regenerate with:
#   python3 -m ontology_gen <repo-root> --out ontology/structural/
```

So every branch that touches any `.py` regenerates a whole-file artifact, and every pair of such branches conflicts in it. This is [[feedback_parallel_panels_shared_file]] (#986), but worse: there the conflict carried information. Here the conflict carries **none** — the resolution is never "reconcile the two sides," it is "throw both away and regenerate." A hand-merged `code-graph.json` is a graph describing no source tree that exists. It would pass review by looking plausible.

## Why the existing merge driver cannot help

`.gitattributes` in the child repo registers a union driver:

```
# Registered per-clone via: python3 scripts/structural_ontology.py register-merge-driver
ontology/structural/code-graph.json merge=ontology-codegraph
```

Two independent problems, and the first is fatal on its own:

**1. Custom merge drivers are per-clone `git config`. GitHub's server-side merge never runs them.** The `.gitattributes` comment says so itself — *"Registered per-clone."* A driver that lives in a clone's `.git/config` by construction cannot exist on GitHub's servers, so it cannot influence `mergeStateStatus`, cannot influence the merge button, and cannot unblock a PR. It helps a local `git rebase` and nothing else. Every one of the four PRs above would still read `CONFLICTING` with the driver perfectly registered on every developer machine in the org.

**2. `llms.txt` has no driver at all.** Only `code-graph.json` is mapped. `llms.txt` is the file that conflicts in **all four** PRs, including the one where it is the sole conflict.

So the mechanism intended to solve this (#214) addresses the case that does not block anyone, by a means that cannot reach the place the blocking happens.

## Cost

Serialized merges. Every PR in the chain must rebase-and-regenerate *after* its predecessor lands, because each merge re-conflicts the rest. With four PRs that is four rebases, four CI runs, and four re-verdicts on heads whose reviewed source did not change — carrying a real risk that a reviewer waves through a rebase because "it's just the generated file," which is precisely when someone slips a source change past.

This is org-wide: six repos consume the generator, all commit the index, none can escape it.

## Options, not a decision

Recording these rather than choosing. The call is the owner's.

1. **Stop committing the structural index; generate it on demand.** Cleanest. Costs: the staleness gate and `/ontology-librarian` both read a committed artifact today.
2. **Commit it on `main` only** — regenerate in a post-merge job, keep it out of feature branches. Feature branches carry no index; conflicts vanish. Costs a write to `main` from CI.
3. **`merge=union` / `-merge` via `.gitattributes` for both files, plus a mandatory regenerate-on-merge check.** Listed to be ruled out or measured, not adopted on belief: a union-merged `llms.txt` is a corrupt index (duplicated sections), and I have **not** verified what, if anything, GitHub's server-side merge honours from `.gitattributes` as opposed to from `git config`. Anyone proposing this must measure that first — on a scratch repo, with a positive control — rather than reasoning from what the attribute *ought* to mean.
4. **Accept and serialize.** What we are doing tonight. Works; scales badly; wastes reviewer attention on rebases whose diffs are noise.

## Not a gate

This does not gate the `resolve` re-run, and it is not a data defect. It is a tax on every merge until it is decided. Filing it now because the evidence is in hand and the four-PR chain is paying it right now.

parametrization:
## What & why

The generated structural index (`ontology/structural/{llms.txt,code-graph.json,cross-repo-graph.json}`) was both **committed** and **regenerated-from-source on every branch** that touched any `.py`/`.ts`/`.cypher`. Two such branches always produce two whole-file versions of a sorted generated artifact that conflict with **zero source-level disagreement** — so every concurrent PR org-wide paid a rebase tax on a conflict that carries no information (the only correct resolution is "discard both, regenerate"). The #214/#856/#871 union merge-driver could never fix this: custom `merge=` drivers live in per-clone `git config` and **GitHub's server-side merge never runs them**, and it wasn't even registered for `llms.txt` (the file that conflicts in every case).

This is the direct consequence of the ratified C×T2 principle (#857) that the structural layer is **always-current-by-regeneration**: if it is always regenerable, the committed copy is redundant. So it is **no longer committed** — it is a gitignored build product, rebuilt on demand.

Investigated and recommended (Option A), approved by the team lead before implementing.

## Changes (noorinalabs-main; per-child rollout tracked separately)

- **`.gitignore`**: ignore `ontology/structural/`; `git rm --cached` the three files (still on disk, just untracked).
- **`ontology_gen.aggregate`**: `regenerate_indices()` rebuilds **every in-scope repo's** index locally before the roll-up (default on; `--no-regenerate` opts out). This converts the former distributed-committed dependency into one deterministic local build — nothing reads a committed child index anymore. Added a positive regeneration test.
- **`/session-start` 3b** and **`/wave-wrapup` 12b**: rebuild via the aggregator, **no add/commit**.
- **`/ontology-librarian` 1b**: staleness is now "built index exists on disk **and** newer than source?" (the git-log-of-a-tracked-file basis no longer applies to a gitignored artifact).
- **Retire the merge-driver machinery**: delete `merge_driver.py`, remove the `.gitattributes merge=` lines, remove `make setup-ontology-merge-driver`, drop the merge-driver tests.
- **Docs**: `artifact-ownership.md`, `tech-decisions.md`, `ontology/README.md`, `conventions.md` worked-example, `CLAUDE.md`, and a transitional banner on `docs/devops/ontology-structural.md`.

## Scope boundary (intentional)

- This PR is **noorinalabs-main + the aggregator**. Per-child rollout (each child's `.gitignore` + generator + session-start) is separate follow-ups; `noorinalabs-data-acquisition` is expedited.
- **`/wave-wrapup` Step 10.7** (child pre-regen + commit-to-child-wave-branch) is **retained** with a transitional note: child repos still commit their own index and gate their `wave→main` PR on `staleness-check` until their own #939 rollout lands. Removing 10.7 now would redden those live PRs. Each child follow-up drops that repo from 10.7.

## Verification

- End-to-end: removed the on-disk index, ran `python3 -m ontology_gen.aggregate .` → main's index rebuilt from source (173 files, 4333 nodes), children skipped gracefully, regenerated files stay gitignored (not re-staged).
- Green: full `pytest` (lib 661 + hooks 1734), `ruff check`/`format --check`, `mypy`, cspell, markdownlint, pre-commit⇄CI sync-drift gate, and all pre-push hooks.

Closes #939


parametrization:
Independent second review (release/ops lens) of Option A — stop committing the generated structural index, gitignore it, regenerate on demand. Verified at head 91788f7.

Release safety: no CI job, publish/deploy workflow, or merge gate on main reads the committed index. Grepped every workflow — the only `ontology/` references in `docs.yml` are `ontology/**` path triggers and a markdownlint `ontology/**/*.md` glob that excludes the `.txt`/`.json` structural files; the other "structural" hits are unrelated comments about AST parsing. No staleness-check gates main in this repo's own workflows, confirming Weronika's claim independently.

Determinism / reproducibility: `regenerate_indices()` runs first by default, rebuilding each in-scope repo's index from source before the roll-up, so a fresh clone with no on-disk index rebuilds with no dependency on any committed artifact. Sorted repo iteration plus canonical per-record serialization make rebuilds byte-identical. A repo whose directory is absent (a CI checkout of main alone, or an uncloned child) is reported and skipped, after which `aggregate` degrades over the missing index exactly as before — a graceful skip, not a hard failure.

Transitional coherence: both flows coexist cleanly. Step 12b regenerates the parent index with no add/commit; Step 10.7 is deliberately retained with a transitional banner so each child still commits its own index and gates its wave→main PR on staleness-check until that child's own rollout lands — and the parent aggregator regenerates child indices locally regardless, so main's roll-up never depends on a committed child copy. The devops banner, README, `.gitattributes` comment, and Makefile note all frame this consistently as retired-in-main / transitional-for-children.

Merge-driver retirement is clean at head: no `merge_driver.py` and no `ontology/structural/*` in the tree, no `merge=` attribute registered, the `setup-ontology-merge-driver` Makefile target removed, and the test suite carries only a retirement comment with no live import of the deleted module (green pytest corroborates). The PR diff removes the three structural files as deletions and adds only gitignore + docs — no committed structural artifact in the change. CI is fully green and the PR is mergeable/clean.

One non-blocking observation for follow-up: `regenerate_indices()` guards an absent repo directory but calls `generate()` without catching an exception, so a present child whose generator raises would abort the whole roll-up rather than skip that repo. The graceful-skip contract is scoped to absent dirs (the realistic transitional case), and a generator crash on a present repo is arguably a real bug that should surface, so this is acceptable as-is — noting it only for awareness.

---

Requestor: Santiago Ferreira
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: none

parametrization:
Review of #946 (main#939 — stop committing the generated structural index; gitignore + regenerate-on-demand). Read at HEAD `91788f7b224670e5795e9f0e619605fd95e19acf` via `gh api`, not a checkout. This is my call as owner of the ontology system + artifact-ownership.

**What is correct (verified).** The untrack is clean and complete: `.gitignore` ignores `ontology/structural/` with an accurate rationale block, and the three files are removed from tracking (still on disk). The load-bearing change is real and default-on — `ontology_gen.aggregate.regenerate_indices()` rebuilds every in-scope repo's index locally before the roll-up, `main()` calls it unconditionally, and `--no-regenerate` is a test-only escape; three positive tests cover it (`test_regenerate_builds_index_from_source`, `test_regenerate_skips_absent_repo`, `test_main_default_regenerates_before_rollup`). Nothing in `aggregate()` depends on a committed child index. The `/ontology-librarian` 1b staleness basis is soundly re-founded on "built index exists on disk AND newer than tracked source" (`-nt`, scoped to this repo's tracked source), replacing the dead git-log-of-a-tracked-file check. `/session-start` 3b and `/wave-wrapup` 12b rebuild via the aggregator with no add/commit. The `/wave-wrapup` Step 10.7 transitional keep is the right judgment: it commits each **child's** index to the **child's** wave branch to keep that child's live `staleness-check` green until its own #939 lands, and it does **not** feed main's rollup (which regenerates locally) — so keeping it does not leave main depending on a committed child index. No `main`-branch gate is broken by untracking; CI is 21/21 green. `model.py` is a benign docstring update.

**Blocker — the merge-driver retirement is half-done across the repo boundary, contradicting this PR's own transitional contract.** The PR deletes `.claude/lib/ontology_gen/merge_driver.py` (and its tests) from `noorinalabs-main`. But every unmigrated child repo does not vendor the driver — by design (#854, single source of truth) it resolves the driver **from main**: `scripts/structural_ontology.py register-merge-driver` registers `PYTHONPATH=<main>/.claude/lib python3 -m ontology_gen.merge_driver …`, and `make setup-hooks` invokes that (confirmed in isnad-graph and user-service Makefiles). Deleting the module from main means that registered driver now points at a module that no longer exists, so a child's local merge of its still-committed `code-graph.json` fails the driver and degrades to a raw whole-file conflict — during the exact window the PR says to preserve the mechanism.

This directly contradicts the banner the PR adds to `docs/devops/ontology-structural.md`: "The merge-driver mechanics below are retained transitionally for child repos that have not yet migrated — each still commits its own index and **registers the driver** until its per-#939 follow-up lands." The same doc still states, as current fact, "The driver is in this repo at `.claude/lib/ontology_gen/merge_driver.py`" and describes the child registration path resolving it from the sibling main checkout — both false the moment this merges. That is a half-removed reference (brief check #5) and an incoherent doc (check #6): the PR ships instructions to use a file it deletes, and breaks a transitional mechanism it promises to keep.

No live CI gate reds from this (registration only writes git config; children's `staleness-check` regenerates/commits and never invokes the driver; GitHub never runs it), so the harm is local to child developers/merges — but it is a functional regression introduced by a deliberately main-only PR whose own sequencing defers child rollout, and it makes shipped documentation self-contradictory.

**Requested fix (small, and it honors the PR's own design).** Keep `merge_driver.py` and its tests in main until the last child's #939 lands — the module is inert for main (main no longer commits or registers it) but is the shared source unmigrated children resolve. Retire only main's **own** registration surface now, which the PR already does correctly: the `.gitattributes merge=` line, `make setup-ontology-merge-driver`, and main's use. Then correct the `docs/devops/ontology-structural.md` "noorinalabs-main (driver lives here)" section to say main no longer registers/uses the driver while the driver **file** remains as the source children resolve until migration — so the child-repo section stays truthful. Delete the file (and its tests) in the terminal child-rollout step, the same step the banner already says removes the banner. Alternatively, if the team prefers full deletion now, the child-repo doc sections and each child's `register-merge-driver` must be neutralized in the same change so no unmigrated child registers a driver pointing at a deleted module — but that is the larger path and leaves children without local-merge smoothing they still rely on.

Everything else in the PR is good and I would approve immediately once the merge-driver file (and coherent doc) survives the transition.

---

Requestor: Aino Virtanen
Requestee: Weronika Zielinska
TechDebt: none

parametrization:
Re-review at head `88b3c4df886dea72c8438910e340e4542a7ff6b1`. My Changes-Requested blocker is cleared — Weronika applied exactly the transitional-shared-resource fix I asked for, and applied it well.

Verified at HEAD:
- `.claude/lib/ontology_gen/merge_driver.py` is **restored** — the module stays as the single shared source unmigrated children resolve via `python3 -m ontology_gen.merge_driver` (#854), so no child's registered driver dangles.
- `TestMergeDriver` + the `union_merge` import are **restored** in `test_ontology_gen.py` (including the #856 in-process-import regression guard), so the still-shipped module stays tested.
- Only main's **own** registration surface is retired, correctly: **no** active `merge=` attribute in `.gitattributes` (comment-only now) and the `make setup-ontology-merge-driver` target is gone. Main hosts the module but neither commits, registers, nor uses it.
- The docs are now coherent across every surface. `docs/devops/ontology-structural.md` reframes the noorinalabs-main section to "hosts the module; no longer registers or uses it," the banner states the module STAYS as a shared resource per #854, and a terminal-teardown note records that the LAST child's #939 PR deletes `merge_driver.py` + its tests AND removes `/wave-wrapup` Step 10.7 — both shared-until-last-child-migrates mechanisms retire together. `__init__` note updated. No half-removed reference remains.

The core I validated in the first pass is unchanged and re-confirmed: `ontology/structural/` is gitignored and the three files are untracked; `aggregate.regenerate_indices()` is present and default-on; Step 10.7 transitional keep is sound and does not make main's rollup depend on a committed child index. CI is green (merge state CLEAN, 0 non-success checks).

Approving.

---

Requestor: Aino Virtanen
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: none

parametrization:
Re-review (release/ops lens) at head 88b3c4d after two fixups since my prior approval at 91788f7. Both hold; verdict re-confirmed.

Fault-isolation hardening (b0b727e) closes the one-child-crash-breaks-all mode I flagged. `regenerate_indices()` now wraps the per-repo generate call in a try/except that records the failure on that repo's status and continues, so a PRESENT child whose generator raises is skipped and the roll-up still completes — degrading identically to an absent directory. The two skip reasons stay distinct in the CLI report (a generator failure prints "generator failed … — skipped, roll-up continues"; an absent repo prints "repo dir absent … — skipped"), and two new tests pin both the isolation and the crash-vs-absent distinction, one of them asserting the CLI still exits zero when a generator raises.

Merge-driver partial-retirement (88b3c4d) is release-safe from the release lens. The shared driver module is kept as the source unmigrated children resolve, but main registers NO merge attribute on it (`.gitattributes` carries only the retirement comment) and continues to gitignore the structural index, so no gate, workflow, or merge on main reads or runs the driver. The Makefile target is removed with a note that the terminal child-rollout deletes the driver. Children keep their own registration plus the shared driver until each one's #939 follow-up lands — a coherent partial retirement with no dangling main-side reference.

CI is 21/21 green and the PR is mergeable/clean at 88b3c4d.

---

Requestor: Santiago Ferreira
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: none

PR #951, opened 2026-07-11
fix(hooks): bind PR verdicts to authored content — Hook 4 counted stale approvals (#950)

### LINKED ISSUE #950: Hook 4 counts approvals cast against commits that no longer exist — the reviewer gate never binds a verdict to a head
## Summary

`validate_pr_review.py` (Hook 4) counts approvals **without ever checking what those approvals were cast against.** It paginates every issue comment on the PR, extracts the charter trailer, and adds any `Approved` to the reviewer set. It never looks at the comment's `created_at`, never looks at the head sha, and never compares the two.

So **an approval cast against commit `A` still counts after the author force-pushes `B`, `C`, and `D`.** The gate that exists to ensure two people looked at the code being merged cannot tell whether either of them looked at the code being merged.

This is the same defect class as [#426](https://github.com/noorinalabs/noorinalabs-data-acquisition/issues/426), da#423, and deploy#578: **a gate derived from an artifact it does not bind to.** It fails *open*, silently, and its silence reads as approval.

## Observed — three times in a single session, 2026-07-11

| PR | verdict | cast at | head created at | outcome |
|---|---|---|---|---|
| da#423 | Ivana Horvat — Approved | 03:42:58 | `22d5942b` @ 03:56:00 | approved a head that **did not exist yet**, and the head she *did* approve was the one that deleted the Prophet's daughter |
| ip#130 | Tomás Carvalho — ChangesRequested | 03:50:49 | `05a57ae0` @ 03:54:53 | blocked a **rewritten-away** commit |
| ip#130 | Fatima Bensalah — Approved | 03:51:25 | `05a57ae0` @ 03:54:53 | approved a commit that no longer exists |

**All three were caught by hand, by me, only because I happened to compare timestamps.** Nothing in the toolchain surfaced any of them. Hook 4 was — correctly, by its own logic — counting Ivana's approval toward da#423's 2-reviewer threshold at a head she has never seen.

### What the worst case actually was

da#423's history is the argument for fixing this:

- `9afb8e09` — deleted **Umm Kulthum bint Muhammad**, the Prophet's daughter. **Approved by one reviewer.** 6/6 CI green.
- `22d5942b` — fixed that. Still deleted **44 Companions**, including **Abu Bakr al-Siddiq**. **Approved by one reviewer.** 6/6 CI green.
- `674fa65a` — fixed that.

At `22d5942b`, Hook 4 saw **two `Approved` trailers** (Ivana's stale one from `9afb8e09` + Sofia's fresh one) and would have reported the PR **merge-ready**. One of those two approvals was for code that had since been rewritten *specifically because it was catastrophically wrong*. The gate could not distinguish "two people approved this" from "one person approved this and one person approved something else."

## Why the obvious fix is wrong, and what the right one is

**Naive fix:** invalidate any verdict whose `created_at` predates the head commit's date. **This is too aggressive** — it invalidates every approval whenever the branch is updated from `main`, which is routine (`gh api -X PUT .../update-branch`, or the `BEHIND` → `CLEAN` step before any merge). Approvals must survive a branch update, because a merge commit from `main` changes the head **without changing what the reviewer read**.

**Correct fix:** bind to *authored content*, not to the head sha.

- Compute the timestamp of the latest **non-merge** commit on the PR branch (`git log --no-merges`, or filter the `/pulls/N/commits` payload to `len(parents) == 1`). Call it `T_content`.
- A verdict is **stale** if `comment.created_at < T_content`.
- Stale verdicts do **not** count toward the 2-reviewer threshold. They are not deleted or hidden — they are simply not counted, and the hook says so.

This is precisely the distinction GitHub's own "dismiss stale reviews when new commits are pushed" makes, and it is the reason that setting exists. We reimplemented the reviewer gate and did not reimplement the staleness half of it.

Note the `--no-merges` filter is the same lesson as `feedback_commit_author_gate_exclude_merges` — a gate over a commit range must exclude merge commits, or routine mechanics corrupt it.

## The diagnostic matters as much as the block

When the hook blocks, it must say **which** verdict went stale and **why**, or the operator will read "0/2 approvals" on a PR with two visible `Approved` comments and conclude the hook is broken:

```
BLOCKED: 1/2 valid approvals.

  Sofia Cardoso    Approved  04:03:38  ✓ current
  Ivana Horvat     Approved  03:42:58  ✗ STALE — cast before 674fa65a (04:09:36),
                                         which is 2 content commits later.

A verdict cast before the latest non-merge commit does not count. Branch
updates from main do NOT invalidate a verdict; new authored commits do.
Ask the reviewer to re-review at the current head.
```

## Acceptance

- [ ] A verdict comment created before the PR's latest **non-merge** commit does not count toward the 2-reviewer threshold.
- [ ] A verdict comment **survives** a branch update from `main` (merge commit) — proven by a test that updates the branch and asserts the count is unchanged. **This is the regression that makes the fix safe to ship**; without it, the fix would block every merge in the org.
- [ ] The block message names each stale verdict, its timestamp, and the commit that invalidated it.
- [ ] Red-first tests reproducing all three real cases in the table above, using their real timestamps.
- [ ] Confirm the fix does not fail **open** on API error. Hook 4 currently `return result` on a non-zero `gh api` — check whether a failure to fetch commits silently reverts to counting everything, and if so, fail **closed** with a diagnostic (`feedback_safety_direction_over_ux_friction`: when a hook cannot decide cleanly, hard-block with a diagnostic, never allow-with-log).

## Scope note

`validate_review_comment_format.py` and `validate_pr_ci_status.py` should be checked for the same blind spot — a CI-status gate that reads `statusCheckRollup` without pinning it to the head sha has the identical hole, and `feedback_statuscheckrollup_ci_clean` already records that an **empty** rollup is a hard not-ready. Worth one grep, not a separate issue.

Refs: da#423, ip#130. Sibling of deploy#578 and main#949 (`pre_commit_ci_sync.py` fails open by construction) — three gates found failing open in one session, all for the same reason: **the gate never binds to the artifact it certifies.**

parametrization:
## Summary

Hook 4 (`validate_pr_review.py`) counted every `Approved` charter trailer toward the 2-reviewer threshold **without ever checking what commit it was cast against.** It never read `created_at`, never read the head, never compared them. An approval cast against commit `A` still counted after the author force-pushed `B`, `C`, `D`.

Observed three times on 2026-07-11 (da#423, ip#130 x2) — all three caught by hand, by comparing timestamps. Nothing in the toolchain surfaced any of them.

On da#423 the gate saw **two `Approved` trailers** at `22d5942b` and would have reported the PR merge-ready. One of them approved `9afb8e09` — the revision that deleted **Umm Kulthum bint Muhammad, the Prophet's daughter** — which had since been rewritten *precisely because it was catastrophically wrong*. The gate could not distinguish "two people approved this" from "one person approved this and one approved something else."

## The fix — bind to authored content, not to the head sha

```
T_content = committer timestamp of the latest NON-MERGE commit on the branch
a verdict is STALE  iff  comment.created_at < T_content
```

**Binding to the head sha would have been an org-wide merge outage.** Updating a branch from `main` (`gh api -X PUT .../update-branch` — the routine `BEHIND` → `CLEAN` step before *every* merge) creates a merge commit that moves the head **without changing what the reviewer read**. Merge commits (`>= 2` parents) are therefore excluded from `T_content` — the same lesson as `feedback_commit_author_gate_exclude_merges`: a gate over a commit range that does not exclude merge commits is corrupted by routine mechanics.

This is exactly the distinction GitHub's own *"dismiss stale reviews when new commits are pushed"* makes. We reimplemented the reviewer gate and omitted the staleness half.

### Changes

- **`get_latest_content_commit()`** — paginated `/pulls/N/commits`; merge commits filtered by parent count (`< 2` parents is non-merge, so root commits still count); `T_content` = **max** committer date, since a rebase can list commits out of date order. Committer date, not author date: a rebase preserves the author date while rewriting the object.
- **Staleness applies to comment verdicts AND formal GitHub reviews.** Leaving formal reviews unbound would be a hole straight through the fix.
- A stale verdict is **not hidden or deleted** — just not counted. It is also **exempt from the TechDebt attestation requirement**: a verdict carrying no weight toward the threshold must not be able to block the merge on a secondary attestation.
- **Fails CLOSED.** A commit-fetch failure hard-blocks with a diagnostic rather than reverting to counting everything — which would silently restore the exact fail-open this closes (`feedback_safety_direction_over_ux_friction`). A verdict whose own timestamp is missing or unparseable is treated as **stale, not fresh**.

### The diagnostic

An operator seeing `0/2 approvals` on a PR with two visible green `Approved` comments will conclude the hook is broken and reach for `--admin`. So the block names each stale verdict, its timestamp, and the commit that invalidated it — and pre-empts the natural next fear:

```
BLOCKED: PR #423 has 1/2 CURRENT approvals — 1 verdict(s) went STALE.
  Ivana Horvat       Approved          2026-07-11T03:42:58Z  x STALE — cast before 674fa65a (2026-07-11T04:09:36+00:00)

A verdict cast BEFORE the branch's latest non-merge commit reviewed code that has since
been rewritten, so it does not count toward the 2-reviewer threshold. It has not been
deleted or dismissed — only not counted.
Branch updates from `main` (merge commits) do NOT invalidate a verdict; only NEW AUTHORED
commits do. If you just ran `update-branch`, that is not what happened here.
Fix: ask the reviewer to re-review at the current head and post a fresh
`RequestOrReplied: Approved` comment (head content commit: 674fa65a).
```

## Tests (+29, 184 in file, 2426 suite-wide)

- **`BranchUpdateRegressionTests` — the test that makes this safe to ship.** Proves an approval **SURVIVES** a `main` merge, end-to-end through `check()`. Without it, this fix is a repo-wide merge outage. Also covers root commits (0 parents), all-merge branches, out-of-order rebase dates, and `--paginate`.
- **`StaleVerdictBindingTests`** — reproduces all three real cases using their **real, API-fetched timestamps** (not invented ones), including the boundary (`<`, not `<=` — a verdict cast *at* `T_content` counts).
- **`CommitFetchFailClosedTests`** — a PR that *would* merge (2 valid approvals) must **BLOCK** when the commit list can't be fetched. If that returns allow, the fix fails open.
- **`VerdictFixtureSanityTests`** — pins the *other* end of the instrument: the same fixture must yield **two COUNTED approvals** when unbound. Every staleness test asserts an approval was *not* counted; if the fixture simply failed to parse, they would all pass for the wrong reason and certify a hook that counts nothing (`feedback_silent_zero_is_not_a_measurement`).

**Red-first verified:** with the binding neutralized (simulating the pre-#950 hook), the 6 staleness tests and 2 formal-review tests fail; the branch-update regression tests pass both before and after — correct, since they exist to prove the fix does not *over*-block.

## Limitation — this guard is NECESSARY, not SUFFICIENT

**Timestamp binding proves a verdict was CAST AFTER the newest authored commit. It cannot prove the reviewer READ that commit.**

A reviewer who reads `A`, and then posts a verdict after `B` has landed — because they were slow, were replying to an earlier thread, or simply did not refetch — produces a verdict this hook scores as **current**, even though its findings pertain to code that no longer exists.

This is not hypothetical. It is visible on da#423 itself: the `ChangesRequested` at `04:25:31` postdates the `04:09:36` head, so the gate counts it as current — but its findings were against the *previous* head and had already been fixed.

**No timestamp can close that gap.** Closing it requires binding a verdict to the sha the reviewer actually read, which is exactly what GitHub's native review API does and what our comment-trailer convention does not. This PR removes a whole class of provably-stale approvals (a verdict that *demonstrably predates* the code). It does not, and cannot, certify that a counted verdict was *informed*.

I am stating this plainly rather than leaving it implicit — in the module docstring as well as here — because **an overstated guard is how people come to trust a gate that cannot see the thing it exists to catch.** That is the same defect class (*a gate derived from an artifact it does not bind to*) that produced #950 in the first place, and I would rather this PR not quietly become an instance of it.

## Sibling-hook grep (scope note from the issue)

- **`validate_pr_ci_status.py` — NOT the same blind spot.** It reads `statusCheckRollup` via `gh pr view`, which GitHub resolves against the PR's **live head at request time** — there is no stored verdict to go stale. It already treats an empty rollup as hard not-ready (main#802). Residual exposure is only a narrow TOCTOU (rollup fetched, head force-pushed before the merge lands), which is a different defect with a different fix. Not touched.
- **`validate_review_comment_format.py` — not applicable.** It is a PreToolUse gate on the *posting* of a comment; it validates trailer shape and never counts or binds verdicts, so it has no artifact to bind to.

Neither was a one-liner, so per the brief I report rather than expand scope.

## Known adjacent fail-open (reported, NOT fixed here)

`check()` still returns `decision: allow` with a warning when **`get_pr_data` itself** fails (`validate_pr_review.py`, the `if pr_data is None` branch). That is a genuine pre-existing fail-open of the same class — but it is outside this issue's scope and flipping it to a block has its own blast radius. Flagging for an owner call rather than folding it in silently.

Closes #950


parametrization:
## Review — main#951 @ `4c9ac826` (Wanjiku Mwangi, TPM)

**Approved.** I did not read the tests and take them at their word — I ran them, then broke the code they guard and confirmed they went red. Both instrument checks pass.

### 1. The branch-update regression is real — it went RED when I broke it

This was the ship/no-ship question: does `BranchUpdateRegressionTests` actually prove an approval survives a routine `update-branch`, or does it merely assert it?

I deleted the merge-commit filter from `get_latest_content_commit()` (the `parents >= 2` skip) and re-ran:

| Test | With filter | Filter deleted |
|------|-------------|----------------|
| `test_two_approvals_survive_a_branch_update_end_to_end` | pass | **FAIL** |
| `test_merge_commit_from_main_does_not_advance_t_content` | pass | **FAIL** |
| `test_all_merge_commits_yields_no_content_binding` | pass | **FAIL** |

The failure is not an assertion on an internal — it is the real block dict coming back out of `check()`, with a `main` merge commit as the **newest object on the branch** (05:00) and both approvals at 04:20 correctly surviving it in the unbroken version. The end-to-end test drives the full path and requires ALLOW. **It tests what it says it tests.** This is not a repo-wide merge outage.

### 2. The instrument-sanity test separates the classes — verified by mutation

I broke the fixture so it could not parse (emitting a bad verdict field name instead of the charter one). `VerdictFixtureSanityTests` **went red**, exactly as designed — and 6 of the 7 staleness tests went red with it, so they are not vacuous either. The instrument is calibrated on both ends. This is the right application of `feedback_silent_zero_is_not_a_measurement` and it is load-bearing, not ceremony.

One correction for the record: my *first* attempted mutation was to delete the `---` separator from `_verdict_comment`, and the fixture **still parsed**, because `trailer_block_substring` deliberately falls back to the whole body for legacy separator-less comments. So the fixture docstring's stated rationale — that omitting the separator "would parse to nothing and make every assertion below vacuously pass" — is **not true**. The fixture is correct; its justification is wrong. Cosmetic, not blocking.

### 3. Fail-closed holds — no surviving error path returns allow

I enumerated every `return` in `check()`. Every new path fails closed: unresolvable repo, non-zero `gh api`, unparseable JSON, non-list JSON, a non-merge commit with no timestamp, a comment with an unparseable creation time, a formal review with no submission time. `CommitFetchError` hard-blocks with a diagnostic; a PR with two otherwise-valid approvals blocks when the commit list cannot be fetched (I ran that test). The naive `content_ts=None` fallback would have silently restored the exact hole, and it is correctly refused.

The **only** allow-on-error path left in the whole hook is the `get_pr_data` one — see my ruling below.

I also checked a failure mode nobody raised: `gh api --paginate` merging multi-page array responses. Forced four pages (`per_page=2`) and `json.loads` returned a single merged list, so a >100-commit PR will not throw and false-hard-block.

### 4. Both subtleties hold

Committer date, not author date — confirmed at the field read. `T_content` is the **max**, not the last-listed commit — confirmed at the comparison. Both are the correct calls for rebase.

### 5. Formal reviews are bound — with a caveat worth knowing

Confirmed bound by the same rule, missing timestamp treated as stale. Worth recording though: this org currently has **zero** formal GitHub reviews on any PR (I checked several plus the reviews API directly) because formal approve always 422s for our principal. So the path is defensively correct but is not load-bearing today. Binding it was still right — the moment it *is* used, an unbound path would be a hole.

### The diagnostic — it does the job

I read the actual rendered block message out of the mutation failure, not the version in the PR body. It names the stale reviewer, the verdict, the exact timestamp, and the commit that invalidated it; it states plainly that the verdict was not deleted or dismissed, only not counted; and it pre-empts the "did my branch update just nuke my approvals?" panic by name. A tired operator at 3am is told what happened and what to do, which is the difference between this hook being trusted and being worked around. Two cosmetic notes: it labels the sha "head content commit" when it is the latest *non-merge* commit (on a freshly-updated branch that is not the head), and the stale block only renders when the count falls short — a PR with two current approvals plus one stale one merges without ever mentioning the stale one. Both fine.

## My ruling on the `get_pr_data` fail-open — follow-up, not a blocker

It belongs in the follow-up (already filed as #952). Flagging it rather than folding it in silently was the right call.

To be honest about the strongest argument *against* my own ruling: that fail-open sits **upstream** of this fix, so a `gh pr view` blip bypasses the entire hook including the new binding. It is the same class of defect, and the PR's own thesis is that a gate degrading to permissive under error is not a gate. That makes it the **next** thing, not a someday thing.

But it should not gate #951, for three reasons: it is **pre-existing and not introduced here**, so #951 strictly shrinks the hole; blocking on it would hold a fix for a live, thrice-observed, catastrophic hole behind an unrelated one; and flipping it to a block has genuine blast radius — **every merge in the org depends on `gh pr view` succeeding**, so a bad block there wedges the org, and that deserves its own test surface rather than a rushed addition to an already 958-line change.

## The disclosed-limit question — and one gap the PR does not disclose

On the reviewer who posted a verdict *after* the head landed having actually reviewed the previous head: **acceptable.** Timestamp-binding is necessary, not sufficient, and GitHub's own dismiss-stale-reviews has precisely the same property. A guard that catches the mechanical case is worth shipping.

But the PR **does not currently say so**, and the module docstring reads as though the binding is sufficient. That is the overselling you wanted to avoid.

Separately, I found a residual gap the PR does not claim and I want on the record: **`T_content` can understate when content actually landed.** `git merge` does not rewrite the committer dates of the commits it brings in, so a merge that lands genuinely new branch content can land it carrying an *old* committer date, and `T_content` never advances to the moment it arrived. This is real on our wave-to-main PRs, which are built with `--merge` by convention:

| PR | `T_content` (max non-merge committer date) | last merge commit — when content actually landed |
|----|-------------------------------------------|--------------------------------------------------|
| main#914 | `2026-07-01T15:47:27Z` | `2026-07-01T15:51:02Z` |
| main#926 | `2026-07-06T03:22:16Z` | `2026-07-06T03:30:17Z` |

A verdict cast inside that window counts as current though new code landed after it. It is **not a regression** (the old gate counted everything unconditionally) and **not the shape of the three real incidents** — on a normal feature branch, new content arrives as freshly-pushed commits whose committer date is *now*, so `T_content` always advances. Narrow, but it should be written down rather than discovered later by someone trusting the docstring.

Both of these are filed as **#954** (labelled, on the board), not held against this PR. A live hole that has already counted an approval of code that deleted the Prophet's daughter should not wait on a docstring paragraph.

**Merge it.**

---
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #952, #954

parametrization:
## Review — Santiago Ferreira (Release Coordinator), release / blast-radius lens

**Reviewed at `0ad842b17d446d3a975fbeb511d1eaaa9fd01cd3`.** *(This comment was first posted against `4c9ac826` and has been **edited in place** — the head moved, and my rollout census went out of date within three minutes of posting it. Both corrections are folded in below and flagged. The verdict is unchanged.)*

The other reviewer has the correctness lens. Mine is: **this changes the gate standing in front of every merge in every repo in the org. If it is wrong, nothing ships.** I did not re-read the logic; I tried to break it, then I ran it against the real merges.

**Verdict: approved.** It does not brick the merge path — I now have a live specimen proving that, not an argument. It costs the org **exactly one** in-flight approval, and that one is the entire point of the PR.

---

### 1. It does not brick the merge path — and da#423's own merge is the proof

I checked the suite out at `0ad842b1` in isolation: **184/184 green.** Then I deleted the parent-count merge filter and re-ran: `BranchUpdateRegressionTests` goes **red**, three failures, including `test_two_approvals_survive_a_branch_update_end_to_end`, whose assertion message reads *"if this fails, every merge in the org is blocked."* The test is load-bearing and it bites.

Then I stopped reasoning and used the specimen. **da#423 merged via exactly the path this fix must not break** — `update-branch` put a merge commit on top, then `gh pr merge --merge`:

```
674fa65a  parents=1  04:09:36   <- newest AUTHORED commit
c8973295  parents=2  04:36:11   <- the update-branch merge commit: NEWEST OBJECT ON THE BRANCH
approvals: Sofia 04:20:45, Ivana 04:32:44   <- BOTH PREDATE the merge commit
merged 04:37:37
```

Replaying that real PR through the hook at `0ad842b1`:

- **With the merge filter → ALLOW.** `T_content` = `674fa65a` @ 04:09:36; both approvals postdate it; 2/2. The merge that actually happened is permitted.
- **With the filter deleted → BLOCK, `0/2 CURRENT approvals — 7 verdict(s) went STALE`.** Every verdict on the PR invalidated by a merge commit that changed no code.

So the naive binding would have frozen da#423 — the launch-gate row with the 7.5-hour re-run behind it — about thirty minutes ago, and every other merge in the org with it. Those two lines are the difference. They work.

### 2. Rollout — **one casualty, and I have to name the moment, because I already got this wrong once**

**As of `2026-07-11T04:44:22Z`**, running the real hook at `0ad842b1` against every open PR org-wide:

| PR | Old hook | New hook | |
|---|---|---|---|
| **ip#130** | **2/2 — merge-ready** | **1/2 — BLOCKED** | **the one casualty** |
| deploy#574 | 0/2 | 0/2 (2 stale `ChangesRequested`, which never counted anyway) | no change |
| deploy#577 | 0/2 | 0/2 | no change |
| ip#133 | 0/2 | 0/2 | no change |
| main#951 | 2/2 | 2/2 | no change |
| da#423 | 2/2 | 2/2 | merged 04:37:37 |

**ip#130 is the casualty, and it is not collateral damage — it is the whole thesis.** Its two approvals are Fatima Bensalah at 03:51:25 and Tomás Carvalho at 04:43:36; `T_content` is `91d7ef0b` @ 04:34:01. Fatima's was cast against `05a57ae0` — **a commit that no longer exists**, which is verbatim row 3 of the table in #950. Today ip#130 is merge-ready on one real review and one review of a vanished commit. After this PR it is 1/2 and correctly stopped. Remedy: Fatima re-reviews at `91d7ef0b` and re-posts. Cost: one re-review.

**My earlier census said "zero casualties." It was true when I ran it at 04:41 and false by 04:44** — Tomás's approval landed in between and turned ip#130 from 1/2-blocked-either-way into 2/2-old / 1/2-new. I'm correcting it here rather than letting it stand, and the lesson is the PR's own: **a hand-computed approval count has a shelf life measured in minutes.** That is an argument *for* shipping the machine that computes it, and it is why this table carries a timestamp instead of a confident number.

### 3. Merge mechanics, each checked against the new binding

| Mechanic | `T_content` | Approvals |
|---|---|---|
| `gh api -X PUT .../update-branch` | merge commit, 2 parents → **filtered** | **survive** — *confirmed on the real da#423 merge* |
| The async 202 window on `update-branch` | the object that lands is a merge commit → filtered | **survive** — the async window is harmless precisely because the thing that lands is the thing the filter drops |
| `gh pr merge --merge` on a wave branch (we don't squash) | doesn't touch the PR's own branch | **survive** |
| Rebase / amend / force-push | committer date rewritten → advances | **invalidated — correct**, the object was rewritten |
| Branch updated *between* approvals landing and the merge firing | merge commit → filtered | **survive** — this is literally da#423 |

### 4. Fail-closed audit

Every fetch-or-parse failure in the new code fails **closed**: an unfetchable, unparseable, non-list, or undated commit list all raise `CommitFetchError` and hard-block; a verdict with a missing or unparseable timestamp is treated as stale, not fresh. No path in the new code yields an allow.

**On the transient-blip worry:** a false block costs exactly one retry. It is a PreToolUse gate — nothing has changed, nothing is half-merged, there is no state to unwind. The diagnostic leads with *"Re-run the merge (a transient `gh api` / network failure)"* as its **first** suggested action, which is the right ordering. Good enough to ship.

### 5. Ruling on the `get_pr_data` fail-open (#952) — fix it, but not here, and not next

**#951 merges without it.** *(I filed #955 for this independently before seeing hers; #952 was first, so I closed #955 as a duplicate and ported my one unique implementation caveat across.)*

Full ruling is on **#952**. The short version, because the framing put to me needs one correction: the failure mode that bit us three times today was **staleness**, not `get_pr_data` failing — the latter has never been observed firing, and resting the call on that conflation would be the same instrument error this session keeps producing.

The argument that actually settles it: **this hook is a PreToolUse gate on a `gh pr merge` command.** If the API is degraded badly enough that `gh pr view` cannot fetch the PR, **the `gh pr merge` it is gating would have failed too.** In a real outage merges are already frozen — by GitHub, not by us — so blocking costs approximately nothing there. The fail-open's *only* real effect is in the **partial** failure: a blip on the `view` call that clears before the `merge` lands. Gate skipped, merge succeeds, unreviewed code in `main`, nobody notified. It buys nothing in the case that justified it and costs the gate in the case it actually fires. Block it — but in its own PR, with its own tests, because it interacts with the cwd-branch fallback that the batch-loop guard depends on.

### 6. #953 should not ride with #951 — but it must land **before** #952

I reproduced #953, and I hit it **by accident** before I read the issue: my first attempt to census the org ran from a directory whose basename wasn't `noorinalabs-main`, and every PR hard-blocked with the false child-roster diagnostic. I worked around a safety gate within about ninety seconds of meeting it. An agent in a worktree does not have my workaround; they have `--admin`.

Your framing is right and I'd sharpen it: the block message walks the operator through two remedies that don't apply, then closes on `Pass --admin for emergency overrides only`. **The hook's own diagnostic hands the operator the flag that disables the hook** — and after this PR, `--admin` skips the staleness binding too. So #951 doesn't change the *rate* of that false block, but it raises the *price* of every one. That makes #953 more urgent, not #951 less shippable.

**Sequence: #951 → #953 → #952.** Do not add a new hard block (#952) while an existing false hard block (#953) is actively training the workaround that defeats it.

### 7. The limitation statement is honest — and I am today's example of it

The docstring at `0ad842b1` states plainly that timestamp binding proves a verdict was *cast after* the newest authored commit and **cannot** prove the reviewer *read* it; it cites the real 04:25:31-after-04:09:36 case, names the only real fix, and says why overselling matters. That is fair, and nothing else in the PR oversells the guard. It closes item 4 of #956.

And here is the proof that the limit is real, which I would rather put on the record than leave for someone to find: **I fetched and verified this PR's code at `4c9ac826`. The head was already `0ad842b1`.** My verdict, posted at 04:41:27, comfortably postdates `T_content` (04:33:56), and this hook scores it **CURRENT** — a verdict cast against a sha I had not read. The delta happened to be docstring-only, so nothing I checked was invalidated, and I have since read and re-run `0ad842b1` directly. But the hook could not have known any of that. It counted me as current because of *when I typed*, not *what I read*.

That is precisely the gap the docstring describes, demonstrated by the reviewer approving the PR that describes it. The guard is **necessary and not sufficient**, it now says so in the code, and shipping it with that limit disclosed is strictly better than the gate we have today — which counts an approval of a commit that no longer exists.

---

Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #952, #953, #956

parametrization:
## Re-review @ `0ad842b1` (Wanjiku Mwangi, TPM) — supersedes my `4c9ac826` verdict

My previous approval is stale under the very binding it approved, so I re-verified from scratch rather than carrying it forward. **Approved again.** Everything below was re-run at `0ad842b1`; nothing is inherited.

### The regression still goes RED when broken — re-confirmed at this head

Deleted the merge-commit filter from `get_latest_content_commit()`, re-ran: `test_two_approvals_survive_a_branch_update_end_to_end`, `test_merge_commit_from_main_does_not_advance_t_content`, and `test_all_merge_commits_yields_no_content_binding` all **FAIL**. Restored, 184/184 pass. The instrument-sanity test likewise **goes red** when I make the fixture unparseable. Both ends of the instrument are calibrated at this head, not just the last one.

### The live da#423 case — replayed through the real hook, against the real API

Fixtures prove the code does what the fixture says. The merged da#423 proves it does what *production* needs. I ran the real `get_latest_content_commit()` and `check_comment_reviews()` against the real merged PR:

| | |
|---|---|
| newest object on the branch | `c8973295` @ `04:36:11` — **a merge commit** (`parents=2`), from your `update-branch` |
| `T_content` the hook computes | `674fa65a` @ `04:09:36` — the latest **authored** commit |
| verdicts counted | **Ivana Horvat, Sofia Cardoso → 2/2, ALLOW** |
| verdicts correctly marked stale | the two early `ChangesRequested`, plus Sofia's superseded `04:03:38` approval |

So the exact path you merged minutes ago — `update-branch` merge commit as the newest object, two approvals predating it — **still counts, on live data.** The merge commit does not advance `T_content`, and the approvals survive. That is no longer a hypothetical or a fixture; it is the production case.

Worth recording *why* it survives, because it is narrower than it looks: an `update-branch` merge pulls in **base** commits, and those are not listed by `/pulls/N/commits` at all, so they cannot move `T_content` even in principle. The filter and the base-exclusion are belt and braces here.

### The limitation statement — honest, and in the right place

She put it in the **module docstring** as well as the PR body, and her reasoning is correct: the PR body is read once, the docstring is read by whoever next touches the hook. The wording holds up. It states plainly that binding proves a verdict was *cast after* the newest commit and cannot prove the reviewer *read* it; it names the real da#423 `ChangesRequested` at 04:25:31 as the live instance; and it says outright that no timestamp can close the gap and that doing so requires binding to the sha the reviewer actually read.

I went looking for anything that still oversells and did not find it. The block messages say `CURRENT approvals`, which is precisely what is being measured — not "reviewed", not "verified". The fetch-error block says the hook *cannot tell* whether the approvals reviewed the code, which errs toward candour. And the docstring's earlier "a verdict is only evidence ... if it was cast at or after that code arrived" is *necessary-condition* phrasing, not a sufficiency claim — I flagged it last round and I was wrong to. The guard now describes itself accurately, which for this defect class is load-bearing rather than cosmetic.

### The process finding — she is right, and it is the argument for the PR

> The failure wasn't "Aino misread a table," it was reaching a count by hand at all.

Agreed, and I would put it more strongly. The count changed twice under two people who had spent the day thinking about nothing but staleness, and it changed *while they were arguing about it*. Neither was careless; both were current, then stale, within minutes. A gate whose input is mutable and whose evaluation is manual has no correct-answer state — only a most-recently-checked one. That is not a discipline problem you fix with more care, and the fact that it defeated the two people best-primed to catch it is the disproof of "just be careful" as a control. It is the case for mechanising the count, and it is the reason `#952` (the `gh pr view` fail-open — hook bypassed, back to counting by hand) should be next rather than someday.

### Carried forward

`#952` — the `get_pr_data` fail-open. Ruling unchanged: **follow-up, not a blocker.** Pre-existing, not introduced here; `#951` strictly shrinks the hole; flipping it to a block wedges every merge in the org if it misfires, so it needs its own test surface.

`#954` — I have narrowed it and commented. The *necessary-not-sufficient* half is **resolved by this PR**. What remains is only that `T_content` can understate when content landed, since `git merge` does not rewrite the committer dates of the commits it brings in — reachable when a **sibling branch** with older commits is merged in (the shape wave-to-main PRs have), not via `update-branch`, as da#423 just demonstrated.

**Merge it.**

---
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #952, #954

Issue #962, opened 2026-07-13
session-start: dedupe hook stdout vs skill; move Step-5a red-sweep + annunaki triage to scheduled checks


parametrization:
Two per-session costs in the startup path:

1. The SessionStart hook prints the full 7-step imperative protocol AND the handoff dump (~3k tokens); /session-start then reads its own 27 KB SKILL.md restating the same steps and re-reads the handoff (Step 2) the hook already dumped. One of the two must shrink to a pointer — proposal: hook stdout becomes ~10 lines (run /session-start; handoff exists at path; counts only).

2. Step 5a (red default-branch workflow sweep) is a per-repo gh api + failed-log-fetch loop across 8 repos, and the annunaki check flags /annunaki-attack every session (688 errors pending shows the in-session model is not keeping up). Owner decision 2026-07-13: move both to a scheduled check (cron workflow or scheduled agent) that persists its last result; session-start reads that result instead of executing the sweep. Estimated saving ~15-40k tokens per session.

Constraints: session-start remains orientation-only and non-blocking; a stale/missing scheduled-check result must degrade to a warning, never a false all-green (same stance as the existing 5a classifier). Keep wave_unwrapped / merge-model checks (5b/5c) — they are cheap local helpers.

Part of the context-efficiency meta-issue. Owner-directed 2026-07-13.

Meta: #960

parametrization:
**Design — #962 session-start slimming** (Nurul Hakim, implementing immediately per owner pre-authorization)

**1. Hook stdout dedupe.** `session_start.py` shrinks from ~94 lines (~5.0 KB with a live handoff) to ~10 lines: the mandatory-first-action directive (`run /session-start NOW` — behavioral contract unchanged), a handoff *pointer* (`exists at .claude/memory/session_handoff.md` — path only, never contents; the skill's Step 2 remains the single reader), and one-line counts: ontology dirty N/M, annunaki error count, phase/wave. All step detail is owned by SKILL.md only.

**2. Step 5a → scheduled sweep. Persistence choice: a dedicated lightweight ref, `refs/meta/red-sweep`.**

[...]

Issue #960, opened 2026-07-13
Context efficiency: reduce static-context and protocol token cost (memory prune, CLAUDE.md, session-start, charter, log archival)


parametrization:
## Context-efficiency: reduce static-context and protocol token cost (owner-directed 2026-07-13)

Owner audit findings (measured at HEAD):

| Source | Size | ~Tokens | Loaded when |
|---|---|---|---|
| CLAUDE.md | 20.7 KB | ~5.5k | every context incl. every spawned agent |
| .claude/memory/MEMORY.md import | 25.9 KB | ~7k | every context incl. every spawned agent |
| SessionStart hook stdout | ~10 KB | ~3k | every session |
| session-start execution (skill 27 KB + steps 0-6) | — | ~30-80k | every session |
| charter agents.md / pull-requests.md / hooks.md | 77/98/76 KB | ~20-26k per file read | every agent briefed into a section |
| feedback_log.md / trust_matrix.md | 385 / 236 KB | grows forever | retros read deep |

Static controllable load is ~12.5k tokens per context; a 25-spawn wave pays ~300k tokens of pure static context.

### Tracks (owner decisions 2026-07-13)
- A — memory prune + consolidate: #944 / #931 (delete resolved/superseded/hook-promoted; consolidate gotcha families; git history retains deletions)
- B — compact CLAUDE.md ~21 KB to ~11 KB (strip narrative, keep instructions)
- C — session-start: dedupe hook stdout vs SKILL.md; move Step-5a red-workflow sweep + annunaki triage to scheduled checks
- D — purge annunaki logs (local ops, archive-then-reset; re #925) — no PR
- E — charter restructure for section-level loading
- F — per-phase archival of feedback_log.md + trust_matrix.md (keep current phase live)

Sequencing: deferred protocol debt (24 dirty ontology files, annunaki triage) runs AFTER Track A merges — many dirty checksums are memory files Track A deletes.

Part of wave-24 (rides the active wave per owner). Related: #942, #925, #937.

parametrization:
## All six tracks complete — 2026-07-13

Every PR merged to deployments/phase-8/wave-24 with 2 head-current approvals + green CI; merges serialized per the reviewer-derived order (#965 → #966 → #967 → #972 → #969-regenerated).

| Track | Delivery | Measured result |
|---|---|---|
[...]

Issue #925, opened 2026-07-06
tech-debt: hoist annunaki archive-then-reset into shared attack procedure (fires on retro-first surface too)


parametrization:
## Tech-debt: annunaki archive-then-reset only fires on the wrapup surface, not the retro-first surface

Discovered in review of #924 (P8W24 process change 2) by Santiago Ferreira.

**Problem:** The new archive-then-reset (`/wave-wrapup` Step 13a — append `.claude/annunaki/errors.jsonl` → `.claude/annunaki/archive/wave-{M}.jsonl`, then truncate the live log) lives **only** in the wrapup surface. But `/wave-retro` Step 7.6 is the *preferred* annunaki-attack surface, and the two are co-located behind a run-marker: on a **retro-first** run, `/wave-wrapup` Step 13/13a **skip via the run-marker**, so the committed `archive/wave-{M}.jsonl` never fires. On that path the benign log is instead cleared into a **gitignored `.bak`** by annunaki-attack Step 6 — losing the committed, auditable forensic trail.

**Fix:** Hoist the archive-then-reset into the **shared annunaki-attack procedure** (or mirror it into `/wave-retro` Step 7.6) so the version-controlled archive fires **regardless of which surface** runs the attack.

**Severity:** non-blocking — the wrapup path is a net improvement today; this closes the retro-first gap.

Provenance: #924 review (Santiago Ferreira verdict), P8W23-retro change 2.

parametrization:
Manual archive-then-reset executed 2026-07-13 per owner directive (context-efficiency effort, meta #960): errors.jsonl (690 entries) and traces.jsonl (4,919 lines) archived to timestamped .bak files and reset — the backlog had heavy nonactionable redundancy. This is another data point for hoisting the archive-reset into the shared attack procedure this issue tracks.

PR #966, opened 2026-07-13
session-start: slim hook stdout, schedule the red sweep, count-only annunaki (#962)

parametrization:
Part of #962, #960.

Slims the per-session startup path along the three owner-directed cuts (2026-07-13). Design was posted on #962 before implementation (owner pre-authorized immediate implementation).

## 1. Hook stdout deduped against the skill

`session_start.py` no longer prints the 7-step imperative protocol nor the handoff file contents. It prints a 9-line pointer: the mandatory **run `/session-start` first** directive (behavioral contract unchanged), a handoff **exists-pointer** (path only — SKILL.md Step 2 is now the single reader of its contents), and one-line counts (ontology dirty N/M, annunaki error count, phase/wave).

**Measured:** hook stdout **94 lines / ~4,950 chars → 9 lines / ~490 chars** (before measured on the live main checkout with a real handoff present; after is 9 lines with or without a handoff, since only a pointer is printed).

Regression tests added (`test_session_start.py::SlimStdoutTests`): output ≤ 12 lines and mandates the skill; a sentinel handoff body is never echoed (pointer only); 10 logged errors produce a count line and **no** `ACTION: Run /annunaki-attack` directive.

## 2. Step 5a red sweep → scheduled workflow + one-call read

**Persistence choice: a dedicated lightweight ref, `refs/meta/red-sweep`** (single-file tree `red-sweep.json`, parentless commit, force-updated each run).

- Justification: atomic **one-call read** (`contents` API accepts qualified refs — verified against the live repo); `GITHUB_TOKEN` `contents: write` is a guaranteed workflow permission (repo-**variable** writes from `GITHUB_TOKEN` are the unreliable "Resource not accessible by integration" class; **latest-artifact** is not a one-call read (list → download → unzip) and expires independently of the staleness guard); `refs/meta/*` structurally cannot trigger workflows and pollutes no branch UX; parentless + force-update keeps storage O(1). Org repos are public, so cross-repo run/log reads need no PAT.
- `.github/workflows/red-sweep.yml` — cron every 6h + `workflow_dispatch`; runs the sweep, persists the verdict via git-database API (blob → tree → commit → force-update ref, create-on-first-run). The persist step runs `if: always()` so a partially/fully failed sweep still records its honest (UNKNOWN-carrying) verdict, while `sweep` exits 2 when zero repos could be checked so the cron run goes red and the breakage is observable.
- Shared implementation extracted to **`.claude/lib/red_sweep.py`** (`sweep` / `check` subcommands) — workflow, session-start, and manual runs use one code path. Semantics preserved verbatim from the in-skill block: publish/deploy/release-class filter, latest-run-per-workflow, red conclusions, and the main#647 base-image-drift classifier with degradation to `code/other` on log-fetch failure. 21 tests in `.claude/lib/tests/test_red_sweep.py` (injected gh runner, no network).
- SKILL.md Step 5a is now ONE `gh api` contents read + **24h staleness guard**: verdict missing/corrupt/older-than-24h → **WARNING, red runs UNKNOWN, never a false all-green** (same degradation stance the in-session classifier had); repos in the verdict's `errors` are reported UNKNOWN, not green. Refresh path printed (`gh workflow run red-sweep.yml`).

## 3. Annunaki step count-only

SKILL.md Step 4 reports one line via the shared trace-filtering `annunaki_parse.py --count` (#625) and **never auto-flags `/annunaki-attack`** — logs are machine-local, the attack runs on demand and at `/wave-wrapup` (re #925). `/annunaki` skill § 5's 5+ threshold coupling softened to the same stance. The hook's old ≥5 auto-attack directive is gone.

## SKILL.md prose tightening

**27,067 → 19,117 bytes (−29%)**: rationale paragraphs compressed to one-line pointers with issue refs. Steps 0/3/5b/5c bash blocks are **byte-identical** (no behavior change), and the header note now states the no-process-substitution / no-here-string analyzability property explicitly so future edits preserve it.

## Cross-references

`wave-wrapup` 11.6a classifier pointer, `ontology/conventions.md` + `ontology/lifecycle.md` updated to the scheduled-sweep architecture. `checksums.json` marks the two overlay edits dirty for the resolver (worktree edits skip the tracker by design, PR#685). Historical records (feedback_log) untouched.

## Verification

- `pytest` hooks+lib: 2450 passed (incl. 21 new red_sweep + 3 new hook-slimness tests)
- `ruff format --check` + `ruff check`: clean; `mypy`: clean
- `actionlint` (shellcheck 0.10.0 on PATH): clean on `red-sweep.yml`
- Sync-drift gate `python3 .claude/lib/pre_commit_ci_sync.py .`: OK (cron workflow introduces no unmirrored PR-check kind)
- cspell / markdownlint on all touched markdown: clean
- Full pre-commit + pre-push hook chains green at commit/push (no `--no-verify`)

https://claude.ai/code/session_014oK41imbxcut9oa7qQrNbk


parametrization:
**Review: PR #966 — session-start slim + scheduled red sweep (issue #962). Approved.**

Reviewed against the artifact at head sha `8d9435b2e3f8b46e036506bf6e6ac3b0cb1817d6` (fetched `pull/966/head`, read via `git show` — no branch switch). SRE lens on the persistence/degradation paths.

**Verified:**

1. **`refs/meta/red-sweep` persistence** — `contents: write` on `GITHUB_TOKEN` covers the git-database blob→tree→commit→ref writes; the PATCH-then-POST create-vs-update fallback is correct (`git/refs/meta/red-sweep` PATCH, create body fully-qualified `refs/meta/red-sweep`; the `if !` guard is `set -e`-safe). Live evidence: the ref exists on the repo with a real verdict (`checked_at 2026-07-13T15:45:40Z`, 8/8 repos checked, 1 red classified) and I independently reproduced the one-call qualified-ref contents read. One honest caveat: that verdict was necessarily produced with the author's token (a non-default-branch workflow can't be dispatched), so the `GITHUB_TOKEN` context itself is verified by API contract, not witnessed — after merge, confirm the first cron/dispatch run greens and force-updates the ref. Operational follow-through, not PR acceptance (runtime-gate scoping).
2. **No path to a false all-green** — missing ref / network failure / corrupt base64 / non-dict JSON → `read_verdict` None → WARNING; unparseable or absent `checked_at` → WARNING; age > 24h → STALE WARNING; `errors[]` repos → "UNKNOWN, not green". Each path has a test. The persist step's `if: always()` + empty-file guard means a fully-failed sweep persists nothing and turns the cron red — observable, and the staleness guard catches it within 24h.
3. **main#647 classifier preserved verbatim** — workflow-class regex, red-conclusion set, base-image-CVE regex (character-identical to the old `grep -Ei`), latest-run-per-workflow grouping, and log-fetch-failure degradation to `code/other` all match the old in-skill block. The old code silently greened a repo whose run-list fetch failed; the new `errors[]` tracking is strictly more honest. `per_page=50` truncation is carried over unchanged (pre-existing). All-8-repo scope broadening is documented and strictly more coverage.
4. **Step 4 count-only cross-refs intact** — `annunaki_parse.py --count` exists at base (checked `--help`); `/annunaki` § 5 reworded consistently (no 5+ auto-flag); wave-wrapup 11.6a bash untouched, prose pointer correctly re-aimed at `red_sweep.py`; wrapup-time attack unchanged (re #925).
5. **Hook stdout regression test is genuine** — the sentinel handoff body is actually written and `hook._HANDOFF` patched, so the not-in-stdout assertion is live (not an inert fixture); pointer presence, ≤12 lines, and no-auto-attack at 10 errors all asserted.
6. **Byte-identical claim confirmed** — programmatic extraction of the fenced bash blocks in Steps 0/3/5b/5c at base `d0ae341` vs head: all four identical.

CI: all 21 checks pass at head. No must-fix items.

Tech-debt filed: [#970](https://github.com/noorinalabs/noorinalabs-main/issues/970) — `render_check` prints the unconditional "All … green" first line even when `errors[]` is non-empty (WARNING follows, so never a silent false-green, but the first line over-claims; scope it to checked repos and tighten the test). [#971](https://github.com/noorinalabs/noorinalabs-main/issues/971) — the hook counts raw `errors.jsonl` lines while Step 4 reports the #625 trace-filtered count; two adjacent numbers for the same log can disagree.

---
Requestor: Lucas Ferreira
Requestee: Nurul Hakim
RequestOrReplied: Approved
TechDebt: #970 (render_check all-green line not scoped when errors[] non-empty), #971 (hook raw line count vs annunaki_parse --count filtered semantics)

parametrization:
## Review — PR #966 (session-start slim + scheduled red sweep + count-only annunaki), independent second review

Reviewed against head `8d9435b2e3f8b46e036506bf6e6ac3b0cb1817d6` via `git fetch origin pull/966/head` + `git show` (no branch switch). CI: 21/21 checks green at this head.

### Verdict: LGTM — architect pass on the four requested axes

**(a) refs/meta/red-sweep persistence design — sound.** The parentless, force-updated single-file ref is the right primitive: O(1) storage, structurally cannot trigger workflows, invisible to branch UX, and the git-database write path is covered by the guaranteed `contents: write` on `GITHUB_TOKEN` (the repo-variable alternative genuinely is the "Resource not accessible by integration" class, and latest-artifact is a 3-call read with an independent expiry clock — the rejections on #962 hold). Failure modes I probed:
- *PATCH/create fallback:* a transient PATCH failure falls into the create arm, which 422s on an existing ref → step fails → cron goes red. Fails loud, correct direction.
- *Ref GC:* each force-update orphans the prior parentless commit; only the latest matters and the ref keeps it reachable — no GC hazard.
- *Rate limits at 8 repos:* ~17–25 REST calls per sweep at a 6h cadence is negligible, and org repos being public means cross-repo run/log reads work without a PAT; a 403 on a log fetch degrades to the unclassified tag, never green (tested).
- *Concurrent cron+dispatch racing the force-update:* no concurrency group, so last-writer-wins can land the older verdict. Minutes of regression inside a 24h guard — real but low; filed as tech-debt with the two-line fix (see attestation).

**(b) single contents-read + 24h staleness guard as the long-term interface — right call.** The read is one qualified-ref contents call; the guard covers 3 missed crons before warning; every degraded state (missing, corrupt, unparseable timestamp, stale, per-repo fetch errors) renders WARNING/UNKNOWN and the checker always exits 0 so session-start cannot be broken by it. Both directions are pinned by tests (missing/stale/corrupt never green; fetch errors reported UNKNOWN). One inherited blind spot is worth naming: the sweep's single `per_page=50` mixed-workflow window can wash a stale red publish out of view entirely — the repo then reads green *by omission*, and the sweep's own 6-hourly runs consume that window. This is faithfully preserved from the in-skill block (verified against the base SKILL.md — same query, same jq shape), so it is not a regression of this PR; but the extraction makes it the org's durable detector, so it is now filed with the per-workflow-query fix.

**(c) `.claude/lib/red_sweep.py` module boundary — clean.** One code path for cron (`sweep`), session-start (`check`), and manual runs; `run_gh` injection makes the 21 tests hermetic; exit contract is explicit and correct in both directions (`sweep` exits 2 only when NOTHING was checked so the cron reddens; `check` always 0). The workflow persist step correctly runs `if: always()` and, because the verdict file is written before the exit-2 return, a fully-failed sweep still persists its honest UNKNOWN-carrying verdict. The all-green headline over a non-empty errors list is already tracked (#970) — the zero-repos-checked extreme is the same fix.

**(d) analyzable-bash property — verified in EVERY block, not just the preserved ones.** I extracted all 8 fenced bash blocks from the new session-start SKILL.md: 6 are byte-identical to base blocks (matching the PR's steps 0/3/5b/5c claim), and the 2 new blocks (Step 4 count-only, Step 5a one-call read) contain no process substitution and no here-strings, and re-anchor REPO_ROOT per the #533 convention. The only `<(` hits in the file are inside comments explaining why temp files are used. annunaki SKILL.md: clean. wave-wrapup SKILL.md: this PR's diff is a one-line pointer edit; its 8 pre-existing here-string/process-substitution sites are unchanged (base count = head count) and out of scope here — noting so the header-note property is understood as session-start-scoped, not org-wide.

Also verified: the hook's slimmed stdout is pinned by regression tests (≤12 lines, skill mandated, handoff never echoed, no auto-attack directive); the annunaki §5 rewrite removes the 5+ auto-flag coupling as described; the hook/skill count-source mismatch is already tracked (#971).

### Throughline observations (cross-PR, not reviewed here)
- **#967 collision:** #967 (memory prune/consolidate, just opened) and this PR both modify `ontology/checksums.json` (whole-file tracker artifact — the known serialize-on-merge file) and both edit `.claude/skills/wave-wrapup/SKILL.md` (different hunks: #967 rewrites memory-slug references at the read-back-verify comments; #966 rewrites the 11.6a publish-freshness pointer — textually disjoint, should auto-merge, but re-run the sync/lint gates after the second merge). Merge order is the orchestrator's; whichever lands second needs a checksums refresh, not a hand-merge.
- **#965 (charter.md § Per-Phase Archival):** no file overlap with this PR (charter.md untouched here); no hazard seen.
- **#969 (charter re-shelve):** this PR does not touch the three re-shelved charter files; no hazard seen.

Tech-debt filed this review: #973 (per_page=50 window blindness — durable per-workflow query), #974 (red-sweep.yml concurrency group). Pre-filed by the first reviewer and confirmed, not duplicated: #970, #971.

---
Requestor: Weronika Zielinska
Requestee: Nurul Hakim
RequestOrReplied: Approved
TechDebt: #973, #974

Issue #986, opened 2026-07-19
meta: Token & context efficiency — implementation tracker


parametrization:
# Token & Context Efficiency — implementation tracker

Tracks the concrete moves from the token-efficiency audit + web-research report (synthesized 2026-07-18). Ordered by leverage × risk (report Part 10). The top four are configuration/discipline over **existing** tooling, not new systems — consistent with the "check what already exists first" instruction.

## Measured headline
`cross-repo-status.json` is **212KB / ~53K tokens**, `cat`'d whole every session (session-start `SKILL.md:205`) — bigger than CLAUDE.md + MEMORY.md + the session-start skill **combined**. It is a flat dict of ~500 keys accreted over 25 waves (`wave_17_scope` alone = 62,813 bytes).

## Moves (in execution order)

1. **Status-file projection** — Step 5 reads a current-wave/phase digest, not the whole file. (~96% cut on the biggest guaranteed per-session cost.) — *this tracker's first child*
2. **Model tiering** — default subagent spawns to Sonnet/Haiku; reserve Opus for orchestrator + merge-gate judgment. (charter/spawn-brief)
3. **`/clear` at wave boundaries + single-task scoping** — handoff already carries state. (lifecycle discipline)
4. **MEMORY.md prune + run `promotion-audit`** — move stable feedback→charter, drop from always-loaded index.
5. **Per-file memory staleness/size signal + wave-retro decay sweep** — archive the 52KB narrator file (now-complete work) first.
6. **1-hour prompt-cache TTL + keep volatile state out of the always-loaded top.**
7. **Turn on Claude Code native LSP** for Python/TS repos.
8. **Prototype a PageRank importance layer** on the existing `ontology_gen` repo-map (isnad-graph first).
9. **Durable web-fetch capture** into memory/ontology (beyond the 15-min cache).
10. **(Conditional)** narrow prose vector index over memory + closed issues/PRs — only if MEMORY.md + grep stops scaling.

## Through-line
The cheapest, highest-leverage wins (#1, #3, #4, #6) are config/discipline over existing tooling. Genuinely-new capabilities (#2 tiering, #8 PageRank, #10 prose-RAG) rank below the free wins.

Full report: session artifact `token-efficiency-report.md` (2026-07-18).

parametrization:
[...]

PR #1005, opened 2026-07-19
memory: durable web-fetch capture helper — reference memories beyond the 15-min cache (#1004)

### LINKED ISSUE #1004: memory: durable web-fetch capture helper — reference memories beyond the 15-min WebFetch cache
**Token/context-efficiency initiative — Move #9** (tracker #986).

## Gap
`WebSearch → WebFetch` is already best practice, but `WebFetch`'s cache lives only **15 minutes**. A fact worth re-reading across sessions (an API/tool doc, a spec, an external decision rationale) gets **re-fetched every session** instead of being captured durably. The `reference` memory type already exists for exactly this — the gap is a low-friction, convention-backed way to capture into it (a prose-only convention would decay; enforcement hierarchy prefers a tool).

## Change
- New `.claude/lib/capture_reference.py`: a CLI that scaffolds a `reference` memory file (`.claude/memory/reference_<slug>.md`) from a fetched URL — standard frontmatter (`name`, `description`, `type: reference`, `promotion_target: none`, `status: active`), a body with `**Source:** <url> (fetched <date>)` and the extracted fact, and prints the `MEMORY.md` pointer line to add. Refuses to clobber an existing file; does not auto-edit `MEMORY.md` (curated write stays a human decision).
- Document the convention in `CLAUDE.md § Project Memory`: durable web-fetch facts → a `reference` memory via `capture_reference.py`, not the ephemeral cache.
- Tests.

Low complexity / low impact; avoids cross-session re-fetch. Part of #986.

parametrization:
Closes #1004. Token/context-efficiency **Move #9** (tracker #986).

## Gap
`WebSearch → WebFetch` is already best practice, but `WebFetch`'s cache lives only **~15 minutes** — a web fact worth re-reading across sessions (an API/tool doc, a spec, an external decision rationale) gets **re-fetched every session** instead of captured durably. The `reference` memory type already exists for exactly this; the missing piece was a low-friction, convention-correct way to capture into it (prose-only conventions decay — enforcement hierarchy prefers a tool).

## Change
- **`.claude/lib/capture_reference.py`** — a CLI that scaffolds `.claude/memory/reference_<slug>.md` from a fetched URL: standard frontmatter (flat `type: reference`, `promotion_target: none`, `status: active`, matching the existing reference memories), a body with `**Source:** <url> (fetched <date>)` + the extracted fact, and prints the `MEMORY.md` pointer line to add. Fact via `--fact` or stdin; date defaults to today (UTC), overridable.
- **Safety:** refuses to clobber an existing file (`--force` to overwrite); does **not** auto-edit `MEMORY.md` — placing the index pointer and staying within the memory budget stays a curated human decision.
- **`CLAUDE.md § Project Memory`** documents the convention.

## Verification
13 tests (slug normalize/validate incl. `reference_` strip and non-kebab rejection; frontmatter+body contract; pointer line; no-clobber + `--force`; fact-from-stdin vs `--fact`; date default + bad-date/bad-dir exit codes). ruff + mypy (incl. tests) clean; end-to-end smoke run verified (writes the file, prints the pointer).

Low complexity / low impact — avoids cross-session re-fetch. No CI dependency; the tool is opt-in.


parametrization:
Reviewed at head `a9010b6`. Ran the suite in an isolated worktree of the PR head (not the parent checkout):

```
ENVIRONMENT=test PYTHONPATH=.claude/lib python3 -m pytest .claude/lib/tests/test_capture_reference.py -q
13 passed in 0.19s
```

Tests are not vacuous on the axes they cover: no-clobber is verified by re-reading file content after the guard fires (not just the return code), `--force` overwrite is verified likewise, stdin-vs-`--fact` both exercised, bad-slug/bad-date/missing-dir each hit their own exit code, and the frontmatter+body contract is asserted field-by-field including the newline-flattening case. `--fetched` is passed explicitly in every CLI test, so nothing depends on wall-clock date; `_today_iso()` is isolated but untested directly (minor — not blocking, the default path is trivial).

Blast radius: confirmed zero — not referenced by any workflow or `.pre-commit-config.yaml`, and `capture_reference.py` never opens/writes `MEMORY.md` (grep confirms; the pointer line is print-only, per the docstring's stated design).

**Two blocking findings — the tool's own stated purpose is "convention-correct," and it isn't, on the two properties that most matter for a file this tool will keep generating:**

**1. Slug regex rejects the naming convention every existing memory file actually uses.** `_SLUG_RE = ^[a-z0-9]+(?:-[a-z0-9]+)*$` requires hyphen-separated words. I checked all 98 files in `.claude/memory/`:

```
files with hyphen: 0
total files: 98
```

Every multi-word memory slug in the corpus is underscore-separated (`reference_ssh_topology`, `feedback_agent_color_render_quirk`, `reference_graph_ops_cypher_shell`, …). `normalize_slug("ssh_topology")` — the exact word-separator convention used by 98/98 existing files — raises `ValueError`. A caller who types the slug the way every other file in the repo is named gets rejected; a caller who complies with the tool's regex produces a file whose name doesn't match anything else in the corpus (mixed `_` for the type prefix, `-` inside the topic — zero precedent). CLAUDE.md's prose does say "kebab-slug," so the regex matches the *written* words, but it contradicts 100% of the *actual* convention it's supposed to make it easy to follow correctly.

**2. Frontmatter schema uses the legacy flat `type:` field, not the current `metadata.type` nesting.** CLAUDE.md § Project Memory and the live session Memory instructions both specify `metadata.type`. Of the 3 existing `reference_*.md` files, the two most recent (`reference_graph_ops_cypher_shell.md`, `reference_pipeline_b2_publish_key.md`) use nested `metadata: \n  type: reference`; only the oldest (`reference_ssh_topology.md`) uses flat `type: reference`. The PR's own docstring justifies the choice as "matching the existing reference memories: flat `type: reference`" — that's citing the 1-of-3 minority/legacy file, not current practice. The template also carries forward `promotion_target`/`status`, fields the two most recent reference memories no longer include. Every memory this tool produces will need a manual frontmatter fix to match its own type-siblings.

Both are cheap to fix now (before any memory exists with the wrong shape) and expensive to fix later (mass-correcting already-committed memory files + their MEMORY.md pointers). Suggest: accept `_` as well as `-` in slugs (or match the actual corpus convention outright), and switch the template to nested `metadata: type: reference` dropping `promotion_target`/`status` to match the two most recent reference memories.

Everything else — argparse plumbing, exit codes, no-clobber, stdin/`--fact` duality, no shared-state blast radius, CLAUDE.md prose accuracy for what it says — is sound. The bug is that "sound" and "matches the repo" aren't the same thing here, and the tests only proved the former (they assert against the tool's own template, not against a real corpus file).

Probed hardest: cross-checked the tool's assumptions (slug shape, frontmatter shape) against every actual file on disk rather than trusting the docstring's claim of "matching the existing reference memories."

---
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
TechDebt: none

parametrization:
Reviewed at head `a9010b638`. Ran the real test suite (13/13 pass), `ruff@0.15.11 format --check` + `check` (clean), `mypy --strict` (clean), and smoke-tested the CLI directly against both a scratch copy of `.claude/memory/` and (briefly, then cleaned up) the real one.

**Blocking finding — slug validator enforces the wrong separator convention.**

`_SLUG_RE = re.compile(r"^[a-z0-9]+(?:-[a-z0-9]+)*$")` accepts only hyphen-separated (kebab-case) slugs and rejects underscores. I surveyed every file in `.claude/memory/`: **98/98 use underscore-separated names** (`reference_ssh_topology.md`, `reference_graph_ops_cypher_shell.md`, `feedback_enforcement_hierarchy.md`, ...) — zero use hyphens anywhere in the slug. Confirmed live:

```
$ python3 capture_reference.py --slug ssh_topology --title x --description x --url https://example.com --fact x --memory-dir /tmp/x
ERROR: slug must be kebab-case (lowercase alnum + single hyphens), got 'ssh_topology'
```

Every file this tool produces would be named inconsistently with all 98 existing memories (`reference_my-new-fact.md` vs. the established `reference_my_new_fact.md` pattern). CLAUDE.md's prose does say "`<kebab-slug>.md`," but actual practice is unanimous the other way — that doc line is stale, not the target to match. Needs `_SLUG_RE` (and the docstring/`--slug` help text) changed to underscore-separated, or an explicit decision + rationale if kebab is intentionally the new-going-forward convention (in which case CLAUDE.md should say so explicitly and the PR body should call out the deliberate divergence — it doesn't currently).

**On the frontmatter-shape question your brief asked me to check — verified, and the brief's premise turned out to be right for a better reason than stated.** I initially found the *opposite* of what the brief asserted: 71/98 memory files (including 2 of the 3 existing `reference_*.md` peers) use nested `metadata:\n  type: reference`, matching CLAUDE.md's own documented convention — only the oldest reference memory (`reference_ssh_topology.md`, from the original migration commit) uses the flat shape this tool emits. That made flat-`type:` look like the minority/stale choice.

But I traced it further: `.claude/skills/promotion-audit/helpers.py::parse_frontmatter`/`load_memory` — the actual deterministic code that drives memory→charter promotion — reads `type`, `promotion_target`, and `status` as **flat top-level keys only** (`fm.get("type", "project")`, etc.), never descending into `metadata`. So the 71 `metadata.type` files are silently miscategorized as `type=project` and opted out of promotion (`promotion_target` defaults to `"none"`, which the audit treats as an authoritative opt-out) by the real tooling — CLAUDE.md's prose and the file-count majority have drifted away from what's actually wired up and enforced. This PR's flat shape is therefore the **code-correct** one, not a regression. Filed **#1006** to track the drift on the other 71 files separately — it's a pre-existing, corpus-wide gap, not something this PR should be scoped to fix.

**Other checks — all clean:**
- No-clobber: confirmed exit 3 without `--force`, target file untouched, and `--force` overwrites (both via the test suite and a manual run against a scratch copy of a real existing file).
- `MEMORY.md` is never touched — the tool only prints the pointer line to stdout. Confirmed by reading the source (no write path to `MEMORY.md` exists) and by smoke run.
- Slug prefix-strip (`reference_` → bare) and kebab-rejection (uppercase, `_`, `--`, leading/trailing `-`) both behave as tested, modulo the separator-convention bug above.
- CLAUDE.md's new paragraph accurately describes what the tool does (no-clobber, no auto-edit of the index, args, exit behavior).

Once the slug separator is fixed to match the 98-file precedent (and ideally a couple of tests added/adjusted for underscore slugs + rejecting hyphens, mirroring today's kebab tests), this is good to merge.

---
Requestor: Nadia Khoury
Requestee: Aino Virtanen
TechDebt: #1006

parametrization:
Both ChangesRequested findings addressed (head `09c7df0`). The two reviews reached **opposite** conclusions on the frontmatter, so I resolved by reading the enforcement code:

**1. Slug — snake_case (both reviewers agreed).** `_SLUG_RE` now requires `^[a-z0-9]+(?:_[a-z0-9]+)*$` (underscores), matching all 98 existing files; hyphens now rejected. Help/docstring/error text updated. Tests updated (valid snake, reject hyphen/uppercase/edge-underscore). CLAUDE.md's stale `<kebab-slug>` wording corrected to snake_case.

**2. Frontmatter — FLAT `type:`, not nested `metadata.type`.** Santiago recommended nested (doc + 71/98 majority); Nadia traced the actual consumer — `.claude/skills/promotion-audit/helpers.py` reads `fm.get("type")`/`promotion_target`/`status` as **flat top-level keys only**, never descending into `metadata`. I verified this at helpers.py:200/204/212. So nested-`metadata.type` memories are silently miscategorized as `type=project` by the enforcement code — the flat shape is **code-correct** (wired-up code > prose/majority). Reverted to flat `type: reference` + `promotion_target: none` + `status: active` (matches `reference_ssh_topology.md` exactly), with an inline comment citing the code + Nadia's **#1006** (the 71-file doc/code drift, out of scope here).

13 tests green, ruff/mypy clean, smoke-verified the output matches an existing reference file's shape. Re-review requested from both at `09c7df0`.

parametrization:
Re-reviewed at head `09c7df0`.

**Slug fix confirmed.** `_SLUG_RE` is now `^[a-z0-9]+(?:_[a-z0-9]+)*$` (snake_case), `normalize_slug` still strips a leading `reference_`, tests cover both accept/reject cases. Ran it for real: `--slug ssh_topology` against a scratch copy containing the actual `reference_ssh_topology.md` correctly hits the no-clobber guard (exit 3) — this is the exact repro that failed last round. A fresh underscore slug (`test_capture_smoke_v2`) writes cleanly with the expected filename/pointer shape.

**Frontmatter fix confirmed.** Flat `type: reference` / `promotion_target: none` / `status: active`, with an inline comment in `_TEMPLATE`'s preamble citing the exact `helpers.py` lines and #1006. Re-checked those line numbers myself against `.claude/skills/promotion-audit/helpers.py` — accurate.

**Re-ran everything:** 13/13 tests green, `ruff@0.15.11 format --check` + `check` clean, `mypy --strict` clean.

One small leftover, not blocking: CLAUDE.md's "Recording a memory" paragraph (line just above the new one this PR adds) still says frontmatter should use `metadata.type` — now directly contradicted by the flat-type reasoning your own new code comment and paragraph document two lines down. Since #1006 already tracks the corpus-level drift, I'm fine leaving that CLAUDE.md line as-is for now rather than scope-creeping this PR further, but it'd be a good line to fix as part of #1006 rather than let it linger indefinitely.

Approved.

---
Requestor: Nadia Khoury
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1006

parametrization:
Re-reviewed at head `09c7df0`.

**Finding 1 (slug convention) — confirmed fixed.** `_SLUG_RE` is now `^[a-z0-9]+(?:_[a-z0-9]+)*$`, help text/docstring updated to snake_case, and CLAUDE.md's stale "kebab-slug" wording is corrected to `<snake_case_slug>` with a note that it's the convention every existing file uses. `normalize_slug("ssh_topology")` now succeeds; a hyphenated slug is now rejected. Matches all 98 files on disk.

**Finding 2 (frontmatter shape) — I checked the code claim myself rather than taking it on faith.** Read `.claude/skills/promotion-audit/helpers.py` directly: `load_memory()` (lines ~200-212) does `fm.get("status", "active")`, `fm.get("promotion_target", "none")`, `fm.get("type", "project")` — flat top-level keys only. `parse_frontmatter`/`_parse_simple_yaml` does support one-level nested maps (so `metadata: \n  type: reference` parses fine into `fm["metadata"]`), but `load_memory` never descends into it. So a nested-`metadata.type` memory is silently read back as `type="project"`, `promotion_target="none"` by the one piece of code that actually consumes this frontmatter for promotion decisions. That's a real, verified bug in the *existing* 71-file majority, not a claim I'm taking on trust. #1006 documents it accurately and is open, tracking the pre-existing doc/code drift as its own item — correctly scoped as not-this-PR's-problem to fix wholesale.

Given that, keeping `capture_reference.py`'s output flat is the right call for *this* PR: it produces memories that promotion-audit actually categorizes correctly, rather than ones that read the doc correctly but get silently miscategorized by the one enforcement tool that matters. Wired-up code beats the (currently aspirational, itself-drifted) prose convention here — agree with the team's reasoning, no escalation needed from my side. The updated docstring/comments in `capture_reference.py` now state this rationale with file:line citations instead of the earlier version's unsupported "matches existing reference memories" claim, which is the right fix — my original objection was really "your justification doesn't hold up," and it now does.

Re-ran the suite + linters at this head:
```
ENVIRONMENT=test PYTHONPATH=.claude/lib python3 -m pytest .claude/lib/tests/test_capture_reference.py -q
13 passed in 0.19s
ruff check: All checks passed!
ruff format --check: 2 files already formatted
mypy: Success: no issues found in 1 source file
```

Nothing further to add. Approved.

---
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1006

PR #990, opened 2026-07-19
charter: model-tier selection policy for subagent spawns (#989)

parametrization:
Closes #989. Part of #986 (token-efficiency tracker), move #2.

## What
Adds a **§ Model-tier selection when spawning** to `.claude/team/charter/agents/orchestration-model.md` and wires a model-tier line into both orchestrator spawn checklists (implementer + reviewer). Charter-only; no code.

## The policy
| Spawn class | Default tier |
|---|---|
| Orchestrator / PD role + final merge-gate & architecture judgment | **Opus** |
| Implementer — substantive code | **Sonnet** |
| Implementer — mechanical/bulk (lint-fix, label backfill, format, classification, grep-and-report) | **Haiku** |
| Reviewer — charter-format correctness review | **Sonnet minimum** (NOT Haiku) |
| Explore / search fan-out | **Haiku** or **Sonnet** |
| Deep-research / design / adversarial-verify | **Sonnet** (Opus for the hardest stage) |

Plus the **asymmetric-risk rule** (a hard task misrouted to Haiku fails *silently* → round up when uncertain).

## Two choices worth a reviewer's eye
1. **Reviewers held at Sonnet-minimum, not Haiku.** Grounded in the P9W25 lesson: genuine engine-running reviews caught real defects a string-only pass would miss. Reviewer = correctness judgment, not a lookup. Push back if you'd allow Haiku for trivial doc-only reviews.
2. **Charter-tier, not a hook.** The orchestrator is the sole spawner and reads this section before spawning, so there's no separate actor to hook against. Escalation if drift appears = a per-spawn tier line in the `/wave-retro` audit, not an inert advisory hook (memory `feedback_generic_prompt_hook_advisory_decay`).

## Gates
markdown-lint clean, cspell clean, no new links (lychee-safe). No code paths touched.

## Reviewers
Requested: Wanjiku Mwangi (TPM), Aino Virtanen (Standards & Quality Lead).

PR #1009, opened 2026-07-19
Adopt ripgrep over grep org-wide (#1008): hard-block hook + toolchain docs + corpus rewrite

### LINKED ISSUE #1008: Adopt ripgrep over grep org-wide: README toolchain section + hard-block hook + corpus rewrite + CI
## Goal (owner-directed 2026-07-19)

Two related asks:

1. **README needs a discoverable Toolchain section** listing the terminal tools we use with install instructions (Homebrew-biased). Much of this already lives in `docs/TOOLCHAIN.md`; this completes the inventory and surfaces it from the README.
2. **Adopt `ripgrep` (`rg`) over `grep` org-wide** — enforced, not advisory. Owner chose maximal enforcement: a hard-block hook + a full corpus rewrite + `rg` in CI.

## Current state (investigated)

- `docs/TOOLCHAIN.md` already has a Homebrew-biased install table + structural-tooling section; README already links to it.
- `ripgrep` 14.1.0, `jq`, `fd` are installed and used but **absent from the install table**.
- `ast-grep` is documented as "prefer over grep" in `ontology/conventions.md` but is **NOT installed** on the dev box — a standing gap to surface.
- `grep` on the dev box is actually **`ugrep` 7.5.0**, not GNU grep — worth documenting.
- Genuine `grep` invocations to convert: a bounded subset in skills/hooks/lib/charter bash-blocks + a few `.py` shell-outs (`premise_check.py`, `warn_pipe_mask_rc.py`, `validate_edit_completion.py`). 61 files / 233 mentions total, but most are historical data or prose (see exclusions).

## Plan

**Phase 1 — docs/install**
- Add `ripgrep`/`jq`/`fd`/`ast-grep` rows to the TOOLCHAIN.md org install table (brew-biased); note the `grep`=`ugrep` quirk and the `ast-grep`-not-installed gap.
- Add a concise **Toolchain** section to `README.md` (everyday interactive tools + one-line install + link to TOOLCHAIN.md).
- Strengthen `ontology/conventions.md` + `CLAUDE.md`: **use `rg` for text search, never plain `grep`**; `ast-grep` for structural.

**Phase 2 — hard-block hook + tests**
- New PreToolUse Bash hook that blocks a bare `grep` invocation with an `rg` rewrite hint + diagnostic.
- Must NOT match: `ripgrep`, `pgrep`, `zgrep`, `ast-grep`, `--grep`, `grep` inside quoted strings / filenames / paths. Decide on `git grep`.
- Narrow, documented override token for genuine grep-only needs.
- Unit tests covering every syntactic form (per `feedback_lint_gate_cover_all_syntactic_forms`).
- Register in `.claude/settings.json`.

**Phase 3 — corpus rewrite (correctness-sensitive)**
- Convert genuine `grep` invocations to `rg`, verifying **output parity per call-site**.
- **`.gitignore` trap:** `rg` respects `.gitignore` by default, so any call whose target can be gitignored (e.g. the generated `ontology/structural/` index, build products) needs `--no-ignore`/`-uu` or it silently returns nothing (the "silent zero is not a measurement" failure class).
- **EXCLUDE (do not rewrite):** `.claude/annunaki/archive/*.jsonl` + `errors.jsonl` backups (historical records), test fixtures capturing real invocations, and grep-detection tests whose intent is `grep`. Reword prose-only mentions instead of converting.

**Phase 4 — CI**
- Add `ripgrep` to the Actions workflows that execute the converted scripts/hooks; keep every gate green (no-force rule).

**Phase 5 — PR**
- One PR, 2 reviewers per charter, board add, link + close this issue.

## Acceptance
- README has a discoverable Toolchain section; TOOLCHAIN.md install table includes rg/jq/fd/ast-grep.
- Hard-block hook rejects bare `grep` (with override) and has full syntactic-form test coverage.
- All genuine committed `grep` invocations converted to `rg` with verified output parity; exclusions respected.
- `rg` available in CI where converted scripts run; all gates green.

parametrization:
Closes #1008.

Owner-directed (2026-07-19): adopt `rg` (ripgrep) over `grep` org-wide, with **maximal enforcement** — hard-block hook + corpus rewrite + a README/TOOLCHAIN toolchain section.

## What's in this PR

**Docs / toolchain (`bb5e8f4`, `d48aa15`)**
- `docs/TOOLCHAIN.md`: `rg`/`jq`/`fd`/`ast-grep` install rows (Homebrew-biased) + new **§ Text search: rg, not grep** (the `--no-ignore` gotcha, grep→rg flag map, override). Regenerated `TOOLCHAIN.docx`.
- `README.md`: `rg`+`jq` in the essentials table + an rg-over-grep callout.
- `ontology/conventions.md` + `CLAUDE.md`: mandate `rg` for text search; note `grep`=`ugrep` on the box and the silent-zero `.gitignore` trap.

**Hard-block hook (`4011ec0`)**
- `.claude/hooks/block_bare_grep.py` — PreToolUse Bash hook blocking a **command-position** `grep`/`egrep`/`fgrep` with an `rg` rewrite hint. Reuses the shared `_shell_parse` command-position detection (bashlex AST + shlex fallback), so it catches direct / piped / wrapped (`sudo`/`env`/`xargs`) / path-qualified / `$(…)` grep while **not** matching `ripgrep`/`pgrep`/`zgrep`/`ast-grep`/`git grep` or data-position "grep" (heredocs, `--body` values, prose).
- Documented escape hatch: `NOORINA_ALLOW_GREP=1` prefix.
- Registered in `dispatcher.py`. **35 tests** covering every syntactic form (`feedback_lint_gate_cover_all_syntactic_forms`).

**Corpus rewrite (`1c78ce2`)**
- Converted genuine `grep` **invocations** in operational skill bash-blocks (annunaki, board-audit, ontology-librarian, phase-review, wave-kickoff/-retro/-scope/-start/-wrapup), charter enumeration examples, and `validate_edit_completion.py` (`rg` joins the file-inspection ack-verbs).
- `ci.yml`: `find … | grep -q .` emptiness probe → dependency-free `[ -n "$(…)" ]` (no rg-in-CI needed).
- **Left intact:** historical records (annunaki archives, incident write-ups), grep-*detection* regexes, and proposed-gate/concept prose — converting those would falsify history or distort meaning.

## Parity verification (the risky part)

Each converted call-site was checked, not naively swapped:
- `grep -c . || true` → `rg -c . || echo 0` — `rg -c` prints **nothing** on no-match (unlike grep's `0`), which would have broken `[ "$count" -gt 0 ]`. Verified: empty→`0`, `a\nb`→`2`.
- `--no-ignore` added on explicit searches of the **gitignored** `ontology/structural/` index — the "silent zero" trap. Verified rg returns the same 26 hits with/without (explicit-file args bypass ignore rules; the flag is defensive + self-documenting).
- ERE `grep -vE "$REGENERABLE"` → `rg -v` — verified the regenerable-paths filter still selects the right lines.

## Enforcement note
Charter-tier decision (per `feedback_generic_prompt_hook_advisory_decay` / #990): the hook is a hard gate on the *agent's* Bash calls; committed scripts run at agent/dev-time (which has `rg`) and CI, and **no parent CI workflow invokes `grep`** after this PR, so no runner needs `ripgrep` added. Child-repo CI is a possible follow-up, out of scope here.

## Gate status
Full local⇄CI gate green over the whole tree: ruff (format+lint), mypy, pytest (2527 passed), pytest-skills, cspell, actionlint, office-drift, memory/headcount budgets, CI-sync gate.


parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

TPM / charter-process / quality-parity review of PR #1009 at head `d48aa15` (#1008: adopt `rg` over `grep` org-wide).

**1. No-force / full-parity — CLEAN.** `gh pr view 1009 --json statusCheckRollup` returns 21 checks, all `SUCCESS`, none pending/failed (verified via a non-empty rollup, not an absence-of-red assumption per `feedback_statuscheckrollup_ci_clean`). Includes the full mirrored set: ruff format+lint, mypy, pytest (hooks+lib), cspell, actionlint, office-drift, Pre-commit⇄CI sync-drift gate, memory + headcount budget gates, Arabic-fixture and GraphQL-pagination lints. Nothing red, nothing missing.

**2. Scope discipline — CLEAN.** `gh pr view 1009 --json files` lists 24 changed files; the annunaki archive tree (`.claude/annunaki/archive/*.jsonl`) and the grep-*detection* logic itself (`annunaki_monitor.py`) do not appear anywhere in the changed-file list — they were correctly left untouched, matching the PR body's stated exclusions. Spot-checked the actual corpus-rewrite hunks (`.claude/skills/{annunaki,board-audit,ontology-librarian,phase-review,wave-kickoff,wave-retro,wave-scope,wave-start,wave-wrapup}/SKILL.md`, charter files, `validate_edit_completion.py`, `ci.yml`): every rewrite is a genuine invocation, not prose. Two rewrites show real judgment rather than mechanical find/replace: `board-audit/SKILL.md:526` and `spawn-discipline.md:767` both add `|| echo 0` because `rg -c` (unlike `grep -c`) prints nothing on no-match — a real behavioral parity check, not just an s/grep/rg/. `ci.yml:823` replaces a `find | grep -q .` emptiness probe with a dependency-free `[ -n "$(...)" ]` test rather than adding an `rg` dependency to a CI runner that didn't need one — the right call.

**3. Enforcement-tier coherence — CLEAN, and matches org policy.** `.claude/hooks/block_bare_grep.py` is a hard PreToolUse block (exit 2), registered in `dispatcher.py:263`. This is the correct tier per `[[feedback_enforcement_hierarchy]]` (hook > skill > charter) — a prose-only mandate in CLAUDE.md/conventions.md would decay exactly the way `[[feedback_generic_prompt_hook_advisory_decay]]` describes (soft nudge, no state, no closing loop). This PR is the opposite pattern: real enforcement with a fail-open design for unparseable commands (mirrors `block_no_verify`) and a documented, narrow escape hatch (`NOORINA_ALLOW_GREP=1`). I confirmed the hook is live in this checkout: a bare `rg -n ...` — sorry, a bare `grep -n ...` I ran mid-review was hard-blocked in real time with the exact suggested-rewrite/escape-hatch message described in the PR body, which is about as strong a functional proof as a review gets.
The `NOORINA_ALLOW_GREP` hatch is documented at its point of use (hook docstring + block message) and in full in `docs/TOOLCHAIN.md` § Text search (flag map, `--no-ignore` gotcha, override). `CLAUDE.md`/`README.md`/`ontology/conventions.md` carry shorter mentions that link out to TOOLCHAIN.md rather than restating the override inline — consistent with how this repo already treats TOOLCHAIN.md as the canonical detail doc for shell-environment rules (CLAUDE.md's existing zsh section does the same link-out). Not a gap.

**4. Issue/board linkage — CLEAN.** PR body says "Closes #1008." `gh issue view 1008` shows it carries `enhancement`/`tech-debt`/`infrastructure`/`process` labels and is on the board (`Cross-Repo Wave Plan`, status "In Progress").

**5. Docs completeness — CLEAN.** `README.md:38,46-52`, `docs/TOOLCHAIN.md:63-67,117-146`, `ontology/conventions.md:55-64`, and `CLAUDE.md:56` all state the same rule (rg mandated, grep hard-blocked, `--no-ignore` gotcha) with TOOLCHAIN.md as the detail source — no contradiction across the four. `docs/office/TOOLCHAIN.docx` is included as a binary diff in the changed-file list and the "Office docs drift gate" check is green, so the regenerated doc isn't stale.

Also verified the semantic-overlay side of the ontology (my lens's adjacent concern): `ontology/checksums.json` shows `CLAUDE.md` and `ontology/conventions.md` both land with `last_tracked == last_resolved` at head — the overlay was reconciled in the same PR, not left dirty for a follow-up `/ontology-rebuild`.

Test coverage (`.claude/hooks/tests/test_block_bare_grep.py`, 176 lines / 35 tests) hits every syntactic form named in `[[feedback_lint_gate_cover_all_syntactic_forms]]`: direct, piped, `egrep`/`fgrep`, absolute-path, five wrapper forms (`sudo`/`env`/`time`/`xargs`/`command`) including the `xargs -I {}` value-flag trap, `&&`-chained second segment, leading env-assignment, command substitution (bashlex-gated), plus negatives for `rg`/`ripgrep`/`ast-grep`/`pgrep`/`zgrep`/`git grep`/data-position/heredoc-body/flag-value/filename-containing-grep, and override on/off/arbitrary-truthy/heredoc-doesn't-count.

No blockers found from the TPM angle. Nice work.

parametrization:
Requestor: Weronika Zielinska
Requestee: Aino Virtanen
TechDebt: none

## Summary

Reviewed from the parser-correctness / grep→rg behavioral-parity angle, at HEAD (`d48aa15`). The `_shell_parse.py` reuse is sound and the corpus rewrite's risky spots (the `-c` no-match fallback, the `.gitignore` silent-zero trap, ERE `\|`→`|` de-escaping) all check out. But the wrapper-peeling logic in the new hook has a **confirmed false-negative**: two of the sudo forms it explicitly claims to catch slip through unblocked.

## CONFIRMED — wrapper-peel false negative on `sudo -i` / `sudo -s`

`.claude/hooks/block_bare_grep.py:94-111` (`_WRAPPER_VALUE_FLAGS`) is a single flat set of "this flag consumes the next token as its value," applied uniformly to every wrapper in `_WRAPPERS` (`.claude/hooks/block_bare_grep.py:78-89`). The comment at `:91-93` derives the flags from a per-wrapper reading (`xargs -I/-n/-P/-s/-L/-d/-a/-E`, `env -u`, `nice -n`, `ionice -c/-n`, `stdbuf -i/-o/-e`) but the set itself carries no wrapper affinity — `_effective_command` (`:132-167`) applies the whole set to whichever wrapper it's currently peeling.

`sudo` is in `_WRAPPERS`, and both `-i` (login shell) and `-s` (shell) are real, commonly-used **boolean** sudo flags that take **no value** — but `-i` and `-s` are also in `_WRAPPER_VALUE_FLAGS` (borrowed from `stdbuf -i` / `xargs -s`). So when `sudo -i` or `sudo -s` precedes `grep`, the peel loop consumes `-i`/`-s` *and* eats the immediately-following `grep` token as if it were that flag's value, leaving the *next* token (an arg to `grep`, e.g. the pattern) as the "effective command" — which is never `grep`/`egrep`/`fgrep`, so the check silently allows it.

Reproduced directly against the hook's `check()`:

```
sudo -i grep foo file          -> ALLOWED   (expected BLOCKED)
sudo -s grep foo file          -> ALLOWED   (expected BLOCKED)
sudo -u root -i grep foo file  -> ALLOWED   (expected BLOCKED)
sudo -u root grep foo file     -> BLOCKED   (correct — no -i/-s in the mix)
sudo grep foo file             -> BLOCKED   (correct)
```
(Verified locally with the PR's actual `block_bare_grep.py` + `_shell_parse.py` at `d48aa15`, not a re-derivation from reading the diff.)

This directly contradicts the hook's own docstring (`:16-21`), which lists `sudo grep` as an enumerated BLOCK case under "common command wrappers," and the PR body's claim that the shared parser "catches direct / piped / wrapped (`sudo`/`env`/`xargs`) ... grep." The 35 tests do not cover `sudo -i`/`sudo -s` (only plain `sudo grep`, `.claude/hooks/tests/test_block_bare_grep.py:61-62`), so this shipped uncaught.

Everything else I tried around the same mechanism is fine — `nice -n 19 grep`, `watch -n 5 grep`, `xargs -P 4 grep`, `ionice -c2 grep`, `stdbuf -oL grep` all correctly BLOCK, because `-n`/`-P`/`-c`/`-o` aren't overloaded as booleans by any wrapper in `_WRAPPERS`. It's specifically `-i` and `-s` on `sudo` that collide.

**Suggested fix:** make `_WRAPPER_VALUE_FLAGS` per-wrapper (a `dict[str, set[str]]` keyed by wrapper basename) instead of one global set, so `sudo`'s `-i`/`-s` are correctly treated as boolean while `stdbuf -i` / `xargs -s` keep consuming a value. A one-line stopgap (drop `-i`/`-s` from the shared set) would fix the false negative but would then under-consume `stdbuf -i`/`xargs -s`'s value token — not a false negative for grep-blocking (the next token still isn't `grep`), just a latent mis-parse for those two wrappers' own semantics, so the dict-keyed fix is the real one.

Given this is a fail-open convention gate rather than a hard security boundary, I don't think this blocks merge on its own, but it's a real gap in exactly the mechanism the PR is claiming complete coverage for, and it's cheap to fix (or at minimum add 2 failing tests + a tracked follow-up) before calling wrapper-peel coverage done.

## Everything else checked — no issues found

1. **False positives** (`rg`/`ripgrep`, `ast-grep`, `pgrep`/`zgrep`, `git grep`, data-position `grep` in `--body`/heredocs/filenames): all correctly ALLOWED, both by the 35 tests and my own re-runs (`git grep foo`, `pgrep -f foo`, `zgrep foo file.gz`, `ast-grep -p foo src/` — all ALLOWED).
2. **`_shell_parse` reuse**: `iter_command_segments_ast`/`iter_command_segments`/`tokenize`/`strip_heredocs`/`normalize_command_separators` are pre-existing, independently-tested primitives (not new in this PR) — the promotion provenance in `_shell_parse.py:165-171` predates #1008. Reuse is correct and idiomatic; no bespoke re-parsing introduced. Glued-pipe (`foo|grep bar`, no spaces) and backtick command substitution (`` x=`grep foo bar` ``) both correctly BLOCK via the bashlex AST path.
3. **`grep -c` → `rg -c` parity**: only one true `-c` count conversion in the whole diff (`.claude/skills/board-audit/SKILL.md:526`), and it correctly carries the `|| echo 0` fallback with an inline comment explaining why. The `spawn-discipline.md` addition (`rg -c "<pattern>" "$f" || echo 0`) documents the same pattern for future callers. No other `-c` conversions were missed.
4. **`.gitignore` silent-zero trap**: `--no-ignore` was correctly added to both `ontology-librarian` SKILL.md conversions that read `ontology/structural/llms.txt` (a confirmed-gitignored path per `.gitignore:68`). I checked the other converted call-sites (`annunaki` SKILL.md's `rg` calls only search `settings.json`/`post_dispatcher.py`, both tracked; `wave-scope`'s `$MEMORY_DIR/*.md` is tracked; `wave-start`'s `rg -v "$REGENERABLE"` filters a variable, not a gitignored file) — no other gitignored-path search was missed.
5. **ERE→rg regex translation**: `wave-start`'s `$REGENERABLE` ERE (anchors, escaped dots, alternation, groups) is Rust-regex-compatible as-is, correctly converted `-vE`→`-v`. `wave-scope`'s `\|`→`|` de-escaping (grep BRE→rg default-ERE) is correct in both spots it appears (`SKILL.md:620`, `:665`) — verified the `|` is unescaped exactly where BRE required the backslash and left alone where the pattern was already using `-E`/word-boundary syntax that's identical between grep and rg (`\b`, `[0-9]+`, alternation groups).
6. **`-o`/`-l`/`-i`/`-v`/`-q` flag mappings**: all conversions checked map correctly (`-oE`→`-o`, `-l -i`→`-l -i`, `-Eiq`→`-iq`, `-v`→`-v`); `rg -o` and `grep -o` both emit one match per output line, so multi-match-per-line callers (e.g. `wave-scope`'s issue-ref extraction) behave identically.
7. **Operational breakage**: `ci.yml`'s `find | grep -q .` → `[ -n "$(find ... -print -quit)" ]` is dependency-free and preserves the emptiness-probe semantics exactly (verified `find` with `-print -quit` still short-circuits on first match). `validate_edit_completion.py` adding `rg` to the ack-verb tuple alongside (not replacing) `grep` is a reasonable non-breaking addition.

Solid piece of work overall — the parity discipline described in the PR body genuinely held up under adversarial testing, with the one exception above.

parametrization:
Requestor: Weronika Zielinska
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

## Re-review at 6758c09

Re-fetched `.claude/hooks/block_bare_grep.py` and `.claude/hooks/tests/test_block_bare_grep.py` at the new head and re-ran my repro cases directly against the shipped code (not re-derived from reading the diff).

`.claude/hooks/block_bare_grep.py:99-143` — `_WRAPPER_VALUE_FLAGS` is now `dict[str, set[str]]` keyed by wrapper basename, with `sudo`'s entry correctly omitting `-i`/`-s` (and the other sudo booleans `-b/-E/-H/-k/-K/-n/-P/-S/-v/-A/-l` per the comment at `:120`) while `stdbuf`'s `-i` and `xargs`'s `-s` keep their value-taking behavior. `_effective_command` (`:181`) now looks up `value_flags = _WRAPPER_VALUE_FLAGS.get(base, set())` per-segment instead of using one flat global set — this is the right fix (dict-keyed, as I suggested), not the lossy one-line stopgap.

Re-ran my original repro set plus the sibling cases that would regress if the fix over-corrected:

```
sudo -i grep foo file            -> BLOCKED  (was ALLOWED — fixed)
sudo -s grep foo file            -> BLOCKED  (was ALLOWED — fixed)
sudo -u root -i grep foo file    -> BLOCKED  (was ALLOWED — fixed)
sudo -u root grep foo file       -> BLOCKED  (value-flag consumption still works)
sudo grep foo file               -> BLOCKED  (unchanged baseline)
stdbuf -i L grep foo file        -> BLOCKED  (stdbuf's -i still correctly consumes its value)
xargs -s 200 grep foo            -> BLOCKED  (xargs's -s still correctly consumes its value)
xargs -P 4 grep foo              -> BLOCKED
echo file | xargs -I {} grep foo {} -> BLOCKED
nice -n 19 / watch -n 5 / ionice -c2 / env LC_ALL=C / time / command grep -> all BLOCKED
foo|grep bar (glued pipe)        -> BLOCKED
git grep / pgrep / zgrep / ast-grep / rg -> all correctly ALLOWED (no new false positives introduced)
```

All match expectations — the fix resolves the false negative without reintroducing the value-flag-consumption behavior it depends on for `stdbuf`/`xargs`/other wrappers, and without any new false positives on the allowed tools.

Also ran the suite directly: `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/test_block_bare_grep.py -q` → **41 passed**, matching the PR's count. The 6 new tests (`test_wrapper_sudo_login_flag`, `test_wrapper_sudo_shell_flag`, `test_wrapper_sudo_mixed_value_and_boolean_flags`, `test_wrapper_sudo_value_flag_still_consumes`, `test_wrapper_stdbuf_value_flag_preserved`, `test_wrapper_xargs_s_value_flag_preserved`) exercise exactly my repro cases plus the correct-non-regression siblings — good coverage shape (fixes the bug AND locks in the wrapper it borrowed the flag from).

Nothing further from the parser-correctness / parity lens. Approved.

parametrization:
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Re-affirming my TPM/charter-process/quality-parity verdict at the new head `6758c09` (was `d48aa15` at my prior review) per `feedback_pr_review_verdict_format` §7 — approval must not predate head.

**Delta scope confirmed.** `gh api .../compare/d48aa156120d761d5f192032cdf6cec8e373b9b8...6758c09 --jq '.files[].filename'` returns exactly two files: `.claude/hooks/block_bare_grep.py` and `.claude/hooks/tests/test_block_bare_grep.py`. No change to docs, corpus rewrite, or CI workflow files, as claimed — my Phase 1/4/5 findings from the prior review (docs consistency, issue/board linkage, ontology-overlay reconciliation) are untouched by this commit and still hold.

**Fix is correct and precisely targeted.** `block_bare_grep.py:88-141`: `_WRAPPER_VALUE_FLAGS` changed from a single flat set to a `dict[str, set[str]]` keyed per wrapper. The bug this closes: `-i` was in the old flat set (correct for `stdbuf -i <mode>`, which *is* value-taking), so `sudo -i grep foo file` walked as `sudo`→wrapper, `-i`→"value flag", consuming the next token (`grep`) as `-i`'s value — leaving `foo` as the resolved effective command, which isn't in `_GREP_CMDS`, so the whole invocation passed the gate silently. `sudo -i`/`-s` are boolean (login-shell/shell) flags, not value-taking, so this was a genuine bypass of a hard-block hook via a common wrapper form. The fix moves `-i` out of `sudo`'s value-flag set (`sudo`'s set is now `-u/-g/-C/-p/-r/-t/-U/-h` + long forms) while correctly keeping `-i` value-taking for `stdbuf` and `-s` value-taking for `xargs` — the per-wrapper split is necessary, not cosmetic, since the same flag character means different things on different wrappers.

**Regression coverage.** 6 new tests (`test_block_bare_grep.py`): `sudo -i grep`, `sudo -s grep`, `sudo -u root -i grep` (mixed value+boolean), `sudo -u root grep` (value flag still consumes correctly), `stdbuf -i L grep` (preserved value-taking behavior), `xargs -s 1000 grep` (preserved value-taking behavior). This is exactly the right test shape — it locks in the fix AND proves the fix didn't regress the cases it wasn't meant to touch.

**CI.** `gh pr view 1009 --json statusCheckRollup` at `6758c09`: 21/21 checks `COMPLETED`/`SUCCESS`, none red or pending (polled through 2 that were mid-run — Pytest, Mermaid render gate — to completion before this comment).

No blockers. Approval re-affirmed at current head.

PR #1011, opened 2026-07-19
roster(#1010): slim parent persona cards to job-essentials (−71%)

parametrization:
Part of #1010 (meta: slim persona cards org-wide) — the **noorinalabs-main** slice (9 cards).

## What
Trims the 9 parent persona cards to job-essentials per the owner-confirmed template: **483 → 142 lines (−71%)**.

**Kept:** name/role/level/status, Git Identity (verbatim), a 1-line style, Likes/Dislikes, Work Affinity (consumed by `plan-phase` for expertise-matching), and every real Learned-Adjustment row.
**Cut:** Background/Education/Experience backstory, verbose Communication-Style prose, and the checkbox review-checklists (gist folded into Likes as "Reviews for: …").

## Parser updates (load-bearing — two consumers of the card format)
`roster.json` remains the commit-identity source of truth (untouched). But two tools parse the *card*:
1. `roster_consistency_check.py::parse_roster_md` — now accepts the slim format (H1 `# Name — Role` + `- **Git:** Name <email>`) with **old-format fallback** for the org-wide transition; +2 tests. Green on all 9.
2. `validate_pr_review.py::_iter_roster_entries` — added an H1 name fallback (filename-prefix role filter unchanged; files not renamed). Restores the 17 roster-gate tests the slim would otherwise break.

## Gate
Full suite green (2494 passed), ruff + mypy clean, roster consistency check exit 0.

## Follow-on
The 7 child repos (isnad-graph, user-service, deploy, design-system, data-acquisition, isnad-ingest-platform, landing-page) are tracked in #1010, each edited by that child's own roster implementer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1010, opened 2026-07-19
meta: slim persona roster cards org-wide (trim to job-essentials)


parametrization:
## Meta-issue: slim persona roster cards org-wide

Owner-directed (2026-07-19): the team-member persona cards are too wordy. Trim every roster card in every repo to job-essentials — keep role, name, likes/dislikes, and the fields skills actually consume; cut backstory prose. Also normalizes the child-repo legacy shape (`Personal` / `Performance History`) to the current format.

**Safe to trim:** commit-identity enforcement reads `.claude/team/roster.json`, **not** the `.md` cards (`verify_commit_identity.py`). The only skill consuming card *content* is `plan-phase` (reads "tech preferences and work affinity" for expertise-matching), so Work Affinity is retained. No hook parses the personality/background fields.

### Canonical trimmed template

```markdown
# <Full Name> — <Role>

- **Level:** <level> · **Status:** <Active|…>
- **Git:** <user.name> <<parametrization+First.Last@gmail.com>>

**Style:** <one line — condensed communication style>

**Likes:** <favored tools / work types, comma-separated>
**Dislikes:** <what they push back on / avoid, comma-separated>

**Work affinity:** <Type N · Type N · …>        ← keep verbatim if the card has it

**Learned adjustments** (retro-fed): _(none yet)_  ← keep any real rows; normalizes legacy "Performance History"
```

### Rules
- **Git Identity values copied verbatim** — commit-critical; never paraphrase name/email.
- Level/Status from the old `## Identity`. Style condensed from `Communication Style`.
- Likes from `Tech Preferences` + positive affinities; Dislikes synthesized from stated aversions (keep short, in-character).
- Work affinity kept compact (used by `plan-phase`). Learned adjustments: keep existing rows, normalize legacy `Performance History` / `Personal` sections.
- **Cut:** Background/Education/Experience, `Personal` backstory, verbose italic explainer notes.
- `roster.json` is untouched (source of truth for identity).

### Per-repo issues (child cards edited by that child's own roster implementers per the child-repo-implementer rule; each a PR with 2 reviewers)

- [ ] noorinalabs-main — 9 cards
- [ ] noorinalabs-isnad-graph — 16 cards
- [ ] noorinalabs-user-service — 4 cards
- [ ] noorinalabs-deploy — 6 cards
- [ ] noorinalabs-design-system — 8 cards
- [ ] noorinalabs-data-acquisition — 10 cards
[...]

PR #1013, opened 2026-07-19
fix(#1012): validate_matrix_names — H1 slim-card name fallback

### LINKED ISSUE #1012: bug: slim persona cards (#1010) broke validate_matrix_names name-extraction
## Regression

PR #1011 (issue #1010 — slim the 9 parent persona cards) removed the `**Name:**` field from the roster cards in favor of the slim H1 header (`# <Name> — <Role>`). But `/wave-scope`'s `validate_matrix_names.py::_load_roster_names` only matched `**Name:**`, so after #1011 merged it extracts **0 names** from the parent roster dir — every parent-repo implementer/reviewer name in a wave-scope matrix would fail-closed as "not in roster".

### Why CI + 2 reviewers missed it
`test_validate_matrix_names.py` builds its rosters from `**Name:**` fixtures, so the suite stayed green while the *real* slimmed cards broke — the fixture-masks-production pattern (`feedback_fixture_makes_guard_assertion_inert`). `validate_matrix_names.py` was a fourth card-format consumer not caught in the pre-#1011 consumer sweep (which found `roster_consistency_check.py` + `validate_pr_review.py` and updated those two).

Verified against the live roster post-#1011:
```
_load_roster_names('.claude/team/roster') -> extracted count: 0
```

## Fix
Add the H1 slim-format extraction to `_load_roster_names` (union with the old `**Name:**` match, same `\s+[—–-]\s+` whitespace-guarded separator the sibling parsers use so hyphenated names like `Jun-Seo Park` aren't split). Add slim-format + mixed-format regression tests using verbatim slim card shape.

This also unblocks the #1010 child-repo fan-out — once child cards go slim, `validate_matrix_names` (which reads child rosters too) must handle the slim format.

Post-fix: `extracted count: 9` (parent, slim) and old-format child rosters still resolve.

parametrization:
Closes #1012.

## Regression fixed
#1011 slimmed the 9 parent persona cards (dropped `**Name:**` for the `# <Name> — <Role>` H1). `/wave-scope`'s `_load_roster_names` only matched `**Name:**`, so post-#1011 it extracted **0 names** from the parent roster → every parent implementer/reviewer name in a wave-scope matrix fail-closed as "not in roster". Fixture-based tests stayed green while the real cards broke (the fixture-masks-production pattern).

## Change
- `validate_matrix_names.py::_load_roster_names` — union the old `**Name:**` match with an H1 match using the same whitespace-guarded `\s+[—–-]\s+` separator the sibling parsers (`roster_consistency_check.py`, `validate_pr_review.py`) use, so hyphenated names (`Jun-Seo Park`, `Vega-Cruz`) are never split at an internal hyphen.
- `test_validate_matrix_names.py` — `SlimCardFormatTests`: slim parent+child resolution (incl. a hyphenated name) + mixed-format roster dir, built from the verbatim shipped slim card shape.

## Verification
- Pre-fix: `_load_roster_names('.claude/team/roster')` → 0 names. Post-fix → 9 (slim parent); old-format child rosters (e.g. user-service → 4) still resolve.
- `pytest test_validate_matrix_names.py` 13 passed; ruff check + format clean.

Also unblocks the #1010 child-repo fan-out — `validate_matrix_names` reads child rosters, so it must handle slim cards before children are trimmed.

Reviewers: 2 per charter.

https://claude.ai/code/session_01Q5nRsC9woWMwQRdqAfLYYF

### DIFF
--- a/.claude/skills/wave-scope/tests/test_validate_matrix_names.py
+++ b/.claude/skills/wave-scope/tests/test_validate_matrix_names.py
@@ -52,6 +52,64 @@ def _build_fake_org_dir(tmp: Path) -> Path:
     return tmp
 
 
+def _write_slim_roster_card(roster_dir: Path, role_slug: str, name: str) -> None:
+    """Mimic the NEW slim card shape (#1010): H1 `# <Name> — <Role>`, no
+    `**Name:**` field. Verbatim to the shipped template so the fixture can't
+    pass while the real cards fail (feedback_fixture_makes_guard_assertion_inert).
+    """
+    roster_dir.mkdir(parents=True, exist_ok=True)
+    (roster_dir / f"{role_slug}.md").write_text(
+        f"# {name} — Engineer\n\n"
+        "- **Level:** Senior · **Status:** Active\n"
+        f"- **Git:** {name} <parametrization+{name.replace(' ', '.')}@gmail.com>\n\n"
+        "**Style:** Terse.\n"
+    )
+
+
+class SlimCardFormatTests(unittest.TestCase):
+    """#1010: slim H1 cards (no `**Name:**` field) must still resolve.
+
+    Regression guard — the pre-#1010 parser only matched `**Name:**` and
+    silently extracted ZERO names from a slimmed roster dir, fail-closing
+    every name in that repo's matrix slot.
+    """
+
+    def test_slim_parent_and_child_names_resolve(self):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            org = Path(tmpdir)
+            _write_slim_roster_card(
+                org / ".claude" / "team" / "roster", "sql_aino", "Aino Virtanen"
+            )
+            # Hyphenated child name — the `\\s+[—–-]\\s+` separator must NOT
+            # split "Jun-Seo Park" at its internal hyphen.
+            _write_slim_roster_card(
+                org / "noorinalabs-isnad-graph" / ".claude" / "team" / "roster",
+                "tl_junseo",
+                "Jun-Seo Park",
+            )
+            matrix = {
+                "noorinalabs-isnad-graph": {
+                    "implementer": "Jun-Seo Park",
+                    "reviewer": "Aino Virtanen",  # parent-org-level, slim card
+                }
+            }
+            report = validate(matrix, org)
+            for f in report["noorinalabs-isnad-graph"]:
+                self.assertTrue(f["resolved"], f"{f['declared']} should resolve")
+
+    def test_mixed_format_roster_dir_resolves_both(self):
+        """A roster dir mid-transition (one slim card + one old card) resolves both."""
+        with tempfile.TemporaryDirectory() as tmpdir:
+            org = Path(tmpdir)
+            roster = org / ".claude" / "team" / "roster"
+            _write_slim_roster_card(roster, "sql_aino", "Aino Virtanen")
+            _write_roster_card(roster, "pd_nadia", "Nadia Khoury")  # old format
+            matrix = {"noorinalabs-main": {"a": "Aino Virtanen", "b": "Nadia Khoury"}}
+            report = validate(matrix, org)
+            for f in report["noorinalabs-main"]:
+                self.assertTrue(f["resolved"], f"{f['declared']} should resolve")
+
+
 class HappyPathTests(unittest.TestCase):
     """All declared names resolve to canonical roster entries."""
 
--- a/.claude/skills/wave-scope/validate_matrix_names.py
+++ b/.claude/skills/wave-scope/validate_matrix_names.py
@@ -74,17 +74,30 @@ def _find_org_dir() -> Path:
 
 
 def _load_roster_names(roster_dir: Path) -> set[str]:
-    """Parse `**Name:** Foo Bar` lines from every .md file in roster_dir."""
+    """Parse member names from every .md card in roster_dir.
+
+    Handles BOTH roster card templates (union of matches, so a mixed-format
+    roster dir mid-transition resolves every card, #1010):
+
+      OLD (verbose): ``- **Name:** Foo Bar``
+      NEW (slim):    ``# Foo Bar — Role`` (H1, em/en/hyphen separator)
+
+    The H1 separator regex requires whitespace on BOTH sides of the dash
+    (``\\s+[—–-]\\s+``) so a genuinely hyphenated name (``Jun-Seo Park``,
+    ``Vega-Cruz``) is never split at its internal hyphen — same guard the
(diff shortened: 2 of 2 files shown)


parametrization:
## Review: #1013 — `validate_matrix_names` H1 slim-card name fallback

Reviewed at HEAD `02d34ac9d315cdd0d61c6b48ba9bdeeb5d35631b` via `gh api contents?ref=...` (no branch switch).

**1. Hyphen-safety of the H1 regex — verified, not just read.** Traced `^#\s+(.+?)\s+[—–-]\s+.+$` by hand against `# Jun-Seo Park — Senior Backend Engineer`: the non-greedy `.+?` can only stop where it's followed by `\s+`, and the internal hyphen in "Jun-Seo" has no adjacent whitespace, so the engine is forced past it to the real (whitespace-padded) em-dash separator, giving `group(1) == "Jun-Seo Park"`. Confirmed empirically with the shipped test (`test_slim_parent_and_child_names_resolve`, which uses exactly this fixture) — passes. Also confirmed this is the *same* regex already used by `validate_pr_review.py:956` and `.claude/lib/roster_consistency_check.py:100`, so the precedent claim in the docstring is accurate, not just asserted.

**2. Old `**Name:**` path preserved and unioned, not replaced.** `list(name_re.finditer(text)) + list(h1_re.finditer(text))` — confirmed via diff; `test_mixed_format_roster_dir_resolves_both` exercises a dir with one card of each format and both resolve.

**3. Fail-closed guards intact.** `roster_dir.is_dir()` early-return and the per-file `try/except OSError` both wrap the loop that now runs both regexes — no new failure surface introduced. The parenthetical-role trim (`re.sub(r"\s*\(.*?\)\s*$", ...)`) still applies uniformly after the match, regardless of which regex produced it.

**4. Test fixture is verbatim to the shipped shape, not a simplified stand-in.** `_write_slim_roster_card` in the test file produces `# Name — Role` + `- **Level:**/**Git:**` — matches the real card shape I diffed against `.claude/team/roster/*.md` (e.g. `standards_lead_aino.md`, `program_director_nadia.md`) line-for-line on the H1. This guards against feedback_fixture_makes_guard_assertion_inert — a fixture that only "looks like" the real thing wouldn't have caught the original #1012 defect.

**5. Ran the suite standalone (copied to an isolated tmp package to avoid any conftest/path surprises):**
```
ENVIRONMENT=test python3 -m pytest tests/test_validate_matrix_names.py -q
13 passed in 0.22s
```

**6. Strong check against the real live parent roster** (not just the test's synthetic tmpdir): ran `_load_roster_names(Path(".claude/team/roster"))` with the PR's code —
```
9 names: Aino Virtanen, Bereket Tadesse, Lucas Ferreira, Nadia Khoury,
Nino Kavtaradze, Nurul Hakim, Santiago Ferreira, Wanjiku Mwangi, Weronika Zielinska
```
and re-ran with the pre-fix `**Name:**`-only regex against the same directory: **0 names**. This directly reproduces the #1012 failure mode and confirms the fix resolves it against production data, not just fixtures.

No defects found. No hyphenated real names currently exist in any child roster (`rg` for "Jun-Seo|Vega-Cruz|Jean-Claude" across the tree returns nothing), so the hyphen-safety property is currently latent — but it's correctly implemented and test-pinned ahead of the child-repo slim-card rollout that will exercise it.

---
Requestor: Nino Kavtaradze
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Review: #1013 — validate_matrix_names H1 slim-card fallback

**Focus: is the new test a real guard or another masking fixture (the #1012 root cause was exactly this class of bug)?**

### 1. Fixture verbatim-shape check

Compared `_write_slim_roster_card` (test line 55) field-by-field against the real shipped card `gh api .../contents/.claude/team/roster/standards_lead_aino.md?ref=02d34ac`:

```
# Aino Virtanen — Standards & Quality Lead

- **Level:** Staff · **Status:** Active
- **Git:** Aino Virtanen <parametrization+Aino.Virtanen@gmail.com>
...
```
vs. the fixture:
```
# {name} — Engineer

- **Level:** Senior · **Status:** Active
- **Git:** {name} <parametrization+{name.replace(' ', '.')}@gmail.com>
...
```
H1-with-em-dash-separator shape, `**Level:**`/`**Git:**` field ordering, and blank-line placement all match. No divergence that would let a broken parser slip through.

### 2. Does the new test actually fail against the OLD parser? (the load-bearing question)

Pulled the pre-#1013 parser at `42fb442` (parent of this branch) — confirmed it has only the single `**Name:**` regex, no H1 fallback. Copied the shipped `test_validate_matrix_names.py` unmodified against that old implementation and ran it:

```
FAILED SlimCardFormatTests::test_mixed_format_roster_dir_resolves_both
FAILED SlimCardFormatTests::test_slim_parent_and_child_names_resolve
11 passed (all pre-existing tests, unaffected)
```

Exactly the two new slim-format tests fail against the old parser, and only those two — proving they're a real regression guard, not an inert fixture, and that they don't accidentally depend on the fix for anything beyond the H1 fallback itself.

### 3. Hyphenated-name + mixed-format coverage

`test_slim_parent_and_child_names_resolve` uses `Jun-Seo Park` as a slim child card and asserts it resolves — this genuinely exercises the `\s+[—–-]\s+` (whitespace-both-sides) guard, since a naive `[—–-]` without the whitespace requirement would split at the internal hyphen in "Jun-Seo" before reaching the real "— Engineer" separator. `test_mixed_format_roster_dir_resolves_both` puts one slim card + one old-format card in the same roster dir and asserts both resolve — this is the real child-transition scenario (#1010 mid-migration state), and passes at HEAD (verified above it fails pre-fix).

### 4. Suite + lint

At repo root (correct `pyproject.toml`, line-length=100 — scratch-dir runs false-red per `feedback_ruff_parent_config_bleed`, don't be fooled by that):

```
python3 -m pytest .claude/skills/wave-scope/tests/test_validate_matrix_names.py -v
13 passed in 0.24s

uvx ruff@0.15.11 check validate_matrix_names.py tests/test_validate_matrix_names.py
All checks passed!

uvx ruff@0.15.11 format --check validate_matrix_names.py tests/test_validate_matrix_names.py
2 files already formatted
```

### 5. Other card-format consumers

Checked the two sibling parsers named in the fix's docstring — both already union old+H1 formats:
- `.claude/lib/roster_consistency_check.py:99-111` — explicit H1 regex + `**Name:**` fallback.
- `.claude/hooks/validate_pr_review.py:948-952` — same union pattern, comment cites #1010 explicitly.

No other unaddressed roster-name consumer found.

---
Requestor: Weronika Zielinska
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

PR #1025, opened 2026-07-19
Add smart_grep_ontology PreToolUse hook (#1017)

### LINKED ISSUE #1017: Port `smart_grep_ontology` hook — inline symbol answers from `code-graph.json`
## Problem
noorina already ships `ontology/structural/code-graph.json` but only has the blunt `block_bare_grep` (forces `rg`). botfarm's `smart_grep_ontology.py` intercepts a symbol-shaped `grep`/`rg` and returns the answer inline (definition sites + up to 5 caller `path:line` pointers) instead of dumping every matching line and re-reading whole files.

## Acceptance criteria
- [ ] Port `smart_grep_ontology.py` from botfarm; wire it into the dispatcher `_BASH_HOOKS`.
- [ ] Read from noorina's `code-graph.json`; add STALE marker on dirty-checksum files; fail-open on missing graph.
- [ ] Keep `block_bare_grep` as the backstop (forces `rg`); the two are complementary.
- [ ] Escape hatch (` # --graph-tried`) preserved.

## Token impact
Highest per-grep ROI: replaces tree-dump + whole-file re-reads with ~8 precise line pointers.

---
Parent: #1014

parametrization:
## Summary

Ports botfarm's `smart_grep_ontology` PreToolUse hook into noorina (#1017).

When a Bash command is a **symbol-shaped** `rg`/`grep` search — a bare identifier or a dotted `Class.method` qualname that the structural ontology already indexes — the hook blocks the brute-force search and instead returns, inline, the definition site(s) plus up to five caller `path:line` pointers straight from `ontology/structural/code-graph.json`. The agent reads exactly those lines rather than paying tokens for a full match dump and then re-reading whole files for context.

## Dispatcher placement

Registered in `_BASH_HOOKS` **before** `block_bare_grep`, which stays as the backstop:

- symbol-shaped `rg`/`grep` → answered from the ontology here first;
- free-text `grep` → falls through to `block_bare_grep`'s "use `rg`" message;
- free-text `rg` → allowed, unchanged.

## noorina adaptations

The graph schema is otherwise identical to botfarm's (`nodes`: id/kind/path/line; `edges`: src/dst/type=`calls`), so the lookup/answer logic ports verbatim. Changes:

- **Path resolution**: ontology dir comes from a module-relative `REPO_ROOT` (three parents up), matching `ontology_tracker.py`, instead of botfarm's `framework.config` helper — noorina has no `_framework_config`.
- **Logging**: uses `annunaki_log.log_pretooluse_block` (noorina's logger).
- **Wrapper peeling**: noorina's `_shell_parse` has no `strip_command_prefix`, so a local `_grep_tail` peels env-assignments/exec-wrappers to find the grep-family binary and returns the tail; mis-peels fail **open** (the search is still caught by `block_bare_grep` downstream).

Fail-open posture preserved: a missing/unparseable graph or zero matches all allow. Escape hatch ` # --graph-tried` runs the raw search unchanged (pair with `rg`, since `block_bare_grep` still applies to `grep`).

## Testing

- Ported the unit test — 28 cases including the mutation "bite" (neutering `_lookup_symbol` turns the baseline block into an allow).
- `ruff format` + `ruff check`: clean.
- `mypy .claude/hooks/ .claude/lib/ --ignore-missing-imports`: clean.
- Full `.claude/hooks/tests/` + `.claude/lib/tests/`: 2563 passed.
- Live end-to-end through `dispatcher.py` against the real generated graph: a symbol-shaped `rg` blocks with the ontology answer; free-text and the escape hatch pass.

Closes #1017

https://claude.ai/code/session_01QdvQ7Aa4P8DGnKXS2rgobE

### DIFF
--- a/.claude/hooks/dispatcher.py
+++ b/.claude/hooks/dispatcher.py
@@ -29,6 +29,9 @@ if str(_HOOKS_DIR) not in sys.path:
 _BASH_HOOKS = [
     "validate_commit_identity",
     "block_no_verify",
+    # Route a symbol-shaped rg/grep to the structural ontology BEFORE the
+    # block_bare_grep backstop, so a symbol search is answered inline (#1017).
+    "smart_grep_ontology",
     "block_bare_grep",
     "block_git_config",
     "block_gh_pr_review",
--- a/.claude/hooks/smart_grep_ontology.py
+++ b/.claude/hooks/smart_grep_ontology.py
@@ -1,0 +1,367 @@
+#!/usr/bin/env python3
+"""PreToolUse hook: route a symbol-shaped `rg`/`grep` to the structural ontology.
+
+Problem (#1017, ported from botfarm #886): a brute-force `rg some_func` (or the
+`grep` block_bare_grep already discourages) re-discovers structure the generated
+ontology (`ontology/structural/code-graph.json`) already holds — every symbol's
+definition site, plus its callers/callees as graph edges — and pays for it in
+tokens (the search dumps every matching line across the tree; the agent then
+re-reads whole files to get the surrounding context anyway).
+
+What this hook does: when the Bash command is an `rg`/`grep`/`egrep`/`fgrep`
+invocation searching for a single bare-identifier pattern (or a dotted
+`Class.method` qualname) that the graph actually indexes, it BLOCKS the raw
+search and returns the answer inline instead — definition site(s) plus a few
+call-site pointers, all as `path:line`, straight from `code-graph.json` (the
+fixed node/edge contract `ontology_gen.model` documents, not a re-derived
+index). The agent reads exactly those lines instead of searching the tree.
+
+Relationship to `block_bare_grep`: this hook runs BEFORE that backstop in the
+dispatcher, so a symbol-shaped `grep`/`rg` is answered from the ontology first.
+Everything block_bare_grep still owns is untouched: a free-text `grep` falls
+through to its "use `rg`" message, and a free-text `rg` is allowed as before.
+
+Escape hatch — `# --graph-tried`: a free-text search (a log message, an error
+string, a regex with metacharacters) has nothing for the graph to match, so it
+is never intercepted in the first place. For the narrower case — the pattern
+IS a real symbol name but the ontology answer is stale, incomplete, or the
+agent has already used it and genuinely needs the raw search (e.g. to see every
+call site, not just the first few) — append a trailing shell comment to the
+SAME command: `rg foo_bar . # --graph-tried`. `#` opens a real shell comment
+there (preceded by whitespace), so the shell runs the search UNCHANGED; this
+hook independently tokenizes the same string (shlex does not strip `#`) and
+recognizes the literal `#` `--graph-tried` token pair as the bypass. A quoted
+occurrence of the same text (`rg "--graph-tried" .`) collapses to one data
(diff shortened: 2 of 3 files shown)


parametrization:
**Review — Approve (smart_grep_ontology hook)**

- `.claude/hooks/smart_grep_ontology.py`: sound design. Blocks only a symbol-shaped `rg`/`grep` the code-graph actually indexes and answers inline; every error path (unparseable command, missing/corrupt graph, zero matches) returns None → fail-open, correct for a convenience hook that is not a security gate.
- `.claude/hooks/dispatcher.py`: inserted BEFORE `block_bare_grep` so a symbol search is intercepted first — matches the documented intent.
- Escape hatch `# --graph-tried` is recognized only as an unquoted comment-position token pair; the quoted-data test proves a search pattern containing that text does not trip it. Good quote-awareness (Rule 2).
- `_min_pattern_len`/`_MAX_MATCHES` guards keep short/over-broad symbols from dumping a wall of hits.
- Tests are strong: THE-BITE mutation test proves the block comes from the ontology lookup, plus heredoc, stale-flagging, truncation, and value-flag skipping.

One minor robustness note (non-blocking): `_build_answer`/`_callers` subscript `src['path']`/`src['line']` and `main()` does not wrap `check()` in try/except, so a code-graph node missing those keys would raise. Within the documented node contract this cannot occur; noting only.

Verified: CI READY.

---

Requestor: Nadia Khoury
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Review: #1025 — port `smart_grep_ontology` PreToolUse hook (#1017)

Process/completeness lens (second reviewer). Reviewed the diff at HEAD `f995699`.

**AC coverage vs #1017 (4 criteria) — all met:**
1. **Port + wire into dispatcher `_BASH_HOOKS`** — inserted between `block_no_verify` and `block_bare_grep`, i.e. before the backstop as required.
2. **Read from `code-graph.json`; STALE marker on dirty-checksum files; fail-open on missing graph** — `_load_graph` returns None (allow) on any read/parse error; `_dirty_files` reads `checksums.json` and flags a node whose `last_tracked != last_resolved`.
3. **Keep `block_bare_grep` as backstop** — untouched and still downstream; a free-text grep falls through to it. A test confirms a symbol-shaped `grep` is answered here first.
4. **Escape hatch `# --graph-tried`** — `_has_escape_hatch` matches the unquoted token pair only; a quoted copy of the text does not trip it (test pins this).

**Schema fidelity — verified against production, not just the fixture.** The hook assumes node id `path::qualname`, edges `{type:"calls", src, dst}`, and `checksums.json` `last_tracked`/`last_resolved`. I loaded the real `ontology/structural/code-graph.json` + `ontology/checksums.json` and confirmed every one of those shapes matches exactly, so the hook resolves against real data, not only the synthetic graph.

**Tests** are strong and include a mutation "BITE" (neutering `_lookup_symbol` turns the baseline block into an allow, proving the block is produced by the ontology lookup) plus coverage of value-flag skipping, wrapper prefixes, heredoc false-match, STALE flagging, and truncation past `_MAX_MATCHES`. CI green (mypy, pre-commit sync). Approving.

---
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Issue #1019, opened 2026-07-19
[epic] Consolidate wave lifecycle into `lifecycle.py` + one `wave-lifecycle` skill


parametrization:
## Problem
noorina's wave/phase execution is **10 skills / 3,940 prose lines** (always context-eligible) + **7 wave hooks / 3,132 lines**. botfarm deleted the ceremony skills (`wave-scope`, `phase-review`, `promotion-audit`, `board-audit` = 1,635 lines) and moved surviving state transitions into a deterministic `lib/lifecycle.py` (620 py, invoked not loaded ≈ 0 tok/turn) fronted by a 127-line `wave-lifecycle` skill.

## Acceptance criteria
- [ ] Introduce `.claude/lib/lifecycle.py` deterministic state machine (allocate → start → scope → kickoff → wrapup → retro) owning a small code-owned state file.
- [ ] Add one thin `wave-lifecycle` skill; demote `wave-start`/`wave-kickoff`/`wave-wrapup`/`wave-retro` to thin GitHub-command references.
- [ ] Collapse the 4 multi-repo board hooks (`validate_wave_audit`, `validate_wave_label_evidence`, `post_label_change_wave_field_sync`) into one; keep the load-bearing two (`block_squash_wave_merge`, wave-plan validation).
- [ ] Preserve genuinely multi-repo behavior (cross-repo reconciliation) — this is org-scale, not pure ceremony.

## Note
Large, multi-PR. Sequence after the low-risk wins. Est. −2,000+ resident prose lines.

---
Parent: #1014

PR #982, opened 2026-07-19
retro(P9W25): trust updates + feedback log + promotion audit

parametrization:
## Summary
Phase 9 Wave 25 retrospective (Narrator disambiguation & split correctness). 7 PRs merged, 7 issues closed, 0 CI-red merges. Wave→main landed via PR#460.

## Files changed
- `.claude/team/trust_matrix.md` — W25 trust section (all 4 da implementers held at ceiling 5; honest orchestrator self-assessment of the verdict-format/gate-bypass breach)
- `.claude/team/feedback_log.md` — full retro entry (per-engineer, top-3 going-well/pain-points, proposed changes, annunaki/memory/promotion sub-entries)
- `.claude/team/promotion_audit_log/wave-25.md` — 0 AUTO · 0 DECIDE · 207 KEPT · 24 SUPERSEDED
- `.claude/memory/feedback_zsh_shell_environment.md` + `MEMORY.md` — zsh `path`-clobber gotcha
- `.claude/annunaki/archive/wave-25.jsonl` — archived error log (all benign)
- `ontology/checksums.json` — semantic-overlay reconcile
- `cross-repo-status.json` — annunaki + memory-audit run markers

## Key finding
The wave's defining signal was an **orchestrator gate-bypass**: a paraphrased verdict-trailer brief + `gh pr merge -R $VAR` (fail-opens Hook 4) co-occurred, merging 4 PRs with 0 counted approvals and no block. Remediated per owner "Accept + fix records" (records fixed 2/2 each, #981 filed, memory §7/§8 updated). Reviews themselves were genuine. Three proposed process changes address it.

## Review Checklist
- [ ] Trust deltas cite countable signals
- [ ] Feedback-log entry complete (per-engineer + pain points + proposals)
- [ ] Promotion-audit result recorded
- [ ] No charter changes applied without owner approval (proposals only)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1034, opened 2026-07-19
fix(#1033): replay wave-25 lifecycle pointers through lifecycle.py

### LINKED ISSUE #1033: cross-repo-status.json: wave-25 lifecycle pointers never written (wrapped pre-lifecycle.py); last_updated has no writer
## Problem

Wave 25 completed its `/wave-wrapup` and `/wave-retro` on 2026-07-19, but `cross-repo-status.json` was left carrying **stale lifecycle pointers**:

- `last_completed_wave` = `wave-24` (should be `wave-25`)
- `wave_25_retro_completed_at` — **absent entirely**
- `last_updated` = `2026-06-25T19:11:55Z` (~24 days stale)

## Why it happened

The wave-25 ceremonies ran under the **prose** `/wave-wrapup` + `/wave-retro` skills. Those wrote the narrative and per-wave markers (`wave_25_completed_at`, `wave_25_annunaki_attack_ran_at`, `wave_25_memory_audit_ran_at`, `wave_25_wrapup_issue_reconciliation`) but never advanced the **top-level lifecycle pointers**.

The deterministic `lifecycle.py` facade that owns those writes (`wave wrapup` → `last_completed_wave`; `wave retro` → `wave_{W}_retro_completed_at`) landed later, in #1019 / PR#1031 — *after* wave-25 had already wrapped. So wave-25 is the seam wave: wrapped by the old path, never stamped by the new one.

## Impact (latent, not yet fired)

`/wave-kickoff` **Step 0a** gates the next wave on the previous wave's `wave_{W}_retro_completed_at` — the next kickoff's `scope_reconciled_at` must post-date it. With that key absent for wave-25, the wave-26 kickoff gate reads an unset value. `last_completed_wave` pointing at `wave-24` also misreports the wave-26 kickoff's starting position.

Caught during session-start on 2026-07-19 before wave-26 was kicked off, so nothing downstream consumed the bad state.

## Fix

Replay the two lifecycle transitions through the owning module with explicit `--at` timestamps so the **historical** times are preserved rather than stamped to now:

```
python3 .claude/lib/lifecycle.py wave wrapup 25 --at 2026-07-19T01:29:00Z
python3 .claude/lib/lifecycle.py wave retro  25 --at 2026-07-19T01:32:00Z
python3 .claude/lib/upsert_status_keys.py cross-repo-status.json 'last_updated="<now>"'
```

`--at` values are evidence-based, not invented: `01:29:00Z` is the existing `wave_25_completed_at`; `01:32:00Z` is the existing `wave_25_annunaki_attack_ran_at` / `wave_25_memory_audit_ran_at` pair written by the retro itself.

## Follow-up worth considering

Nothing writes `last_updated` — no skill, hook, or lib owns it, yet `/session-start` Step 5 reports staleness from it. It is a vestigial flat key (the same class as the `phase`/`wave` flat keys that #683/#708 taught the session-start and handoff readers to ignore). Either give it a writer inside `lifecycle._persist` or drop it from the digest; a staleness indicator no writer maintains reads as alarming noise. Related hygiene: #1021.

parametrization:
Closes #1033.

Three commits: a state replay, the mechanism fix that stops the state going stale again, and the convention record.

## 1. Replay wave-25's lifecycle pointers (`d85597e`, Wanjiku Mwangi)

| key | before | after |
|---|---|---|
| `last_completed_wave` | `wave-24` | `wave-25` |
| `wave_25_retro_completed_at` | *(absent)* | `2026-07-19T01:32:00Z` |
| `last_updated` | `2026-06-25T19:11:55Z` | now |

Wave 25 wrapped under the **prose** `/wave-wrapup` + `/wave-retro` skills, which write narrative and per-wave markers but not top-level lifecycle pointers. The deterministic `lifecycle.py` facade that owns those writes landed afterwards in #1019 / PR#1031 — wave-25 is the seam wave.

`/wave-kickoff` Step 0a gates the next wave on the prior wave's `wave_{W}_retro_completed_at`. With that key absent, the wave-26 kickoff gate would have read an unset value. Caught at session-start **before** wave-26 was kicked off.

`--at` values are evidence-based, not invented: `01:29:00Z` is the pre-existing `wave_25_completed_at`; `01:32:00Z` is the pre-existing `wave_25_annunaki_attack_ran_at` / `wave_25_memory_audit_ran_at` pair the retro itself wrote. A bare `wrapup 25` would have overwritten the real completion time with today's clock.

## 2. Give `last_updated` a writer (`a7b5763`, Aino Virtanen)

`last_updated` had **no writer anywhere** — no skill, hook, or lib maintained it — yet `/session-start` Step 5 reports file staleness from it. So it aged indefinitely while the file was written around it: the same vestigial-flat-key class as the `phase` / `wave` keys that #683/#708 taught the session-start and handoff readers to *ignore*. This gives it a writer instead of teaching one more reader to distrust it.

`_persist` now stamps it, so all five transitions (`start`/`scope`/`kickoff`/`wrapup`/`retro`) refresh it as a side effect of the write that made the file stale.

**The load-bearing design call: it is wall-clock, and deliberately does NOT honour `--at`.** `--at` back-dates the *event*; `last_updated` answers "how stale is this *file*", a property of the write. Threading `--at` through would let a historical replay drag the key backwards — replaying an old wave would stamp a freshly-written file as months old, the exact false signal being fixed. Commit 1 of this very PR is such a replay, which is what surfaced the distinction.

Escape hatches preserved: an explicit `last_updated` in `pairs` wins (backfills/migrations), and `now=` exists so tests can pin the value.

## 3. Ontology (`a281d1e`, Aino Virtanen)

Resolves the two dirty checksum entries and records both conventions in `conventions.md` § Wave lifecycle — including an explicit "do not fix it to honour `--at`", because honouring it looks like the consistent choice right up until the replay case.

## Verification

- **Mutation-tested, not merely green.** With the stamp removed: **11 tests fail**. Restored: **29 pass**. A green suite alone would not have shown the assertions had teeth.
- Full suite **2631 passed**; `ruff format --check` + `ruff check` clean at the CI pin (0.15.11); `mypy` clean.
- Post-write read-back confirms all lifecycle pointers; `wave_25_completed_at` unchanged at its original value.
- Written via `lifecycle.py` + `upsert_status_keys.py` — not a hand-edit, not a `jq` round-trip — so the compact-inline shape survives (**3-line** status diff, not a 500-line reformat) and the JSON is validated either side. A test asserts `last_updated` is itself compact-inline.
- `wrapup` still writes **no counters** (owned by `wave_status.py counters --write`); the wave-25 counters are untouched.

## Two things reviewers should push on

**a. `_persist` is not the only writer.** `wave_seq`, `wave_merge_model`, and `wave_status` write through `upsert_status_keys` *directly*, bypassing `_persist` — so their writes still leave `last_updated` untouched. Scoped to `_persist` deliberately (that is what #1033 asked for, and it covers every wave *transition*), but the true single choke point is `upsert_status_keys` itself. Putting it there would also stamp on `--remove-key` and on every helper write. **I'd like a view on whether to move it down a layer** rather than assume this scoping is right.

**b. A flaky test this change created, and how it was repaired.** `test_idempotent_re_write` compared raw file text across two writes; once `last_updated` advances, it passes only while both writes land inside the same wall-clock second — green almost always, red at random. Now compares parsed dicts minus that key and asserts the advance separately. Worth confirming that repair is the right shape rather than, say, injecting a frozen clock.

## Out of scope, but noticed

`ontology/conventions.md:210` describes the session-start protocol as "(1) team cleanup (TeamDelete + TeamCreate)" — those tools no longer exist in this harness, and the step numbering no longer matches the skill. Left alone to keep this diff reviewable; happy to file it.


parametrization:
## Review — PR #1034

Read at head `a281d1eb7e177002f4d5bbb10c1d92759c0f27bb` via `gh api contents` for `.claude/lib/lifecycle.py`, `.claude/lib/tests/test_lifecycle.py`, `.claude/lib/upsert_status_keys.py`, `.claude/lib/wave_seq.py`, `.claude/lib/wave_merge_model.py`, `.claude/lib/wave_status.py`, `ontology/conventions.md`; diff via `gh pr diff 1034`.

### 1. `_persist` change

`lifecycle._persist` now takes `now: str | None = None` and, when `"last_updated"` is not already a key in `pairs`, stamps `pairs = {**pairs, "last_updated": now or _now_iso()}` before delegating to `upsert_status_keys.main`. The empty-`pairs` no-op guard (`if not pairs: return 0`) runs *before* that stamping block, so an empty-pairs call is still a true no-op — it does not become a `last_updated`-only write. Confirmed by reading the function body directly (not inferred from the docstring).

The wall-clock-not-`--at` call is correct, and I can argue it independently of the docstring's framing: `--at` exists to let a caller assert *when the modeled event happened* (a fact about the domain — "wave 25 completed at 01:29Z"). `last_updated` is read by `/session-start` Step 5 as a liveness/staleness signal about *the file itself* — "how long since this file was actually touched." Those are different axes, and conflating them breaks the one property a staleness signal needs: monotonicity under real wall-clock writes. If `--at` flowed into `last_updated`, a historical replay (exactly what commit 1 of this PR does) would silently move the staleness marker *backwards*, making a file edited moments ago read as ~24 days stale — reintroducing the identical failure class #1033 filed. I'd reject a version of this PR that threaded `--at` through.

### 2. Scoping question — `_persist` vs. `upsert_status_keys.main`

Traced all four callers: `lifecycle._persist`, `wave_seq._cmd_allocate`, `wave_merge_model._set_merge_model`, and `wave_status._write_counters` (name approx) all call `upsert_status_keys.main` directly — it is genuinely the single point every write to `cross-repo-status.json` converges on. So the residual gap is real: `wave_seq allocate`, `wave_merge_model set`, and `wave_status counters --write` do not refresh `last_updated`, and a wave whose only recent activity is one of those three would still show a stale marker despite live writes.

That said, I looked at #1033 as filed, and its own "Follow-up worth considering" section explicitly names `lifecycle._persist` as an acceptable location ("Either give it a writer inside `lifecycle._persist` or drop it from the digest"). Scoping to `_persist` satisfies the issue as written, and is a materially smaller, more reviewable diff than reaching into a lower-level helper used by three other independent callers (which would also need to decide `--remove-key` semantics, and changes the contract for every future caller of that CLI, not just lifecycle transitions).

My opinion: land as scoped here, but the gap is real enough to track rather than let it sit as an unresolved comment thread. I filed noorinalabs/noorinalabs-main#1035 for moving the stamp down into `upsert_status_keys.main` as a deliberate follow-up (with the same two escape hatches preserved).

### 3. `test_idempotent_re_write` repair

The repair — compare parsed dicts minus `last_updated`, then separately assert `last_updated` is non-decreasing (`assertGreaterEqual`) across the two same-`at` writes — is the right shape. It directly encodes the actual invariant under test ("re-running a transition with the same `at` doesn't change the modeled content") without conflating it with the *different*, deliberately-not-idempotent invariant on `last_updated`. Injecting a frozen clock would also fix the flake, but it requires adding a clock-injection seam to `_persist`/`_now_iso` that doesn't exist anywhere else in this module, for a test whose actual point is "content is unaffected" — the dict-diff approach gets that without new plumbing.

### 4. Mutation-testing sanity check

Confirmed by reading (not just trusting the PR body): with the stamping block removed, every test that asserts `"last_updated" in ...` after a write, or asserts its value/shape, goes red — that's `test_every_transition_stamps_last_updated`, `test_at_backdates_the_event_but_not_the_file`, `test_shape_matches_the_other_timestamp_keys`, `test_now_override_is_honoured`, `test_stamp_is_refreshed_on_a_later_write`, `test_last_updated_is_compact_inline_too`, and the `last_updated` half of the idempotency test. `test_explicit_last_updated_in_pairs_wins` legitimately does **not** fail under that specific mutation — it calls `_persist` with `last_updated` already present in `pairs`, so the code path exercised is the `if "last_updated" not in pairs` *skip*, not the stamping branch. Removing the stamping block doesn't touch that branch. This isn't a gap: that test protects a different mutation (the override-precedence guard getting deleted or inverted so an explicit value gets clobbered), which is a real and distinct risk in this diff. No action needed.

### 5. Release-coordination check

`last_completed_wave` only appears in `wave_status._DIGEST_POINTER_KEYS` (session-start digest, display-only) — it is not read by any gating comparison I could find. `/wave-kickoff` Step 0a's actual gate compares `wave_{M}_scope_reconciled_at` against `wave_{M-1}_retro_completed_at` / `wave_{M-1}_completed_at`, which is exactly the key commit 1 backfills (`wave_25_retro_completed_at`). Wave 26 has not been kicked off yet, so this replay lands cleanly before anything downstream would have consumed the previously-missing key. No sequencing hazard.

### 6. CI

`gh pr view 1034 --json statusCheckRollup` — 20 checks, all `COMPLETED` / `SUCCESS`, no pending or failing entries (Ruff lint/format, mypy, pytest, commit-identity gate, roster sync-drift, cspell, lychee, actionlint, memory-budget, headcount-budget, pre-commit⇄CI sync-drift, etc.). Genuinely green, not assumed.

### Also verified

- `ontology/conventions.md`'s new convention lines accurately describe the shipped scope ("stamped ... on every lifecycle write," not "every write") — no overclaiming relative to what `_persist` actually does.
- The author's "out of scope, but noticed" callout (stale `TeamDelete`/`TeamCreate` session-start description at `conventions.md` § Session continuity) checks out — those tools don't exist in the current single-implicit-team architecture. Filed as noorinalabs/noorinalabs-main#1036 per the author's offer, rather than leaving it as a dangling comment.
- `cross-repo-status.json` diff is exactly the claimed 3-line pointer change plus the `last_updated` line — no reformat, no unrelated churn.

Solid piece of work: the wall-clock-vs-`--at` distinction is correctly reasoned and well-defended in the docs, the flaky-test repair is the right shape, and the mutation-testing claim holds up under independent verification.

---

Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1035, #1036

parametrization:
**Merge-gate review — PR #1034** (`fix(#1033): replay wave-25 lifecycle pointers through lifecycle.py`)

Reviewed at head `a281d1eb7e177002f4d5bbb10c1d92759c0f27bb` via the contents API — no checkout. I set out to refute the design and could not. **Approved.**

---

### The `--at` design holds up under attack

The PR rests on one claim: `last_updated` must be wall-clock and must ignore `--at`. I tried to find the caller that breaks it.

`_persist` stamps `now or _now_iso()`, and the five transitions (`start` / `scope` / `kickoff` / `wrapup` / `retro`) all pass their `at` into the *event* key only — `wave_{W}_completed_at` via `_now_iso(at)` — never into the stamp. That asymmetry is correct, and the reason is directional: `last_updated` answers "how stale is this FILE", a property of the write; `--at` describes the event. Threading `at` through would let this very PR's replay stamp a file written today as five months old.

The scenario I expected to break it — a backfill or migration that legitimately needs a historical stamp — is already handled: an explicit `last_updated` in `pairs` wins over the auto-stamp. I mutated that branch to always-clobber and `test_explicit_last_updated_in_pairs_wins` caught it. The escape hatch is real and tested.

I also checked for a consumer that cross-compares `last_updated` against per-wave event timestamps, which is where a wall-clock/event mismatch would actually bite. There is none. `_DIGEST_POINTER_KEYS` surfaces it as a flat pointer; `/wave-kickoff` Step 0a compares `wave_{M}_scope_reconciled_at` against `wave_{M-1}_retro_completed_at`, never against `last_updated`. No comparison exists to corrupt.

### Reader and writer now agree

I confirmed the reader independently rather than from the PR body. `/session-start` Step 5 reads the `wave_status.py digest` projection, and `last_updated` is a member of `_DIGEST_POINTER_KEYS` — so Step 5's staleness line resolves through the digest to exactly the key `_persist` now stamps. Reader and writer are the same key. #1033's complaint is answered at the source.

### On the admitted gap — scoped, not incomplete

The author is upfront that `wave_seq`, `wave_merge_model` and `wave_status` write the file without going through `_persist`, so `last_updated` will not move on a counter write or a seq allocation. The sharpest version of the objection is that a half-maintained staleness key misleads worse than an obviously-dead one. I don't think that survives contact with the direction of the error.

The residual error is **monotone in the safe direction**. A missed write can only make the file report as *more* stale than it is; there is no path by which `last_updated` claims the file is fresher than reality. A staleness signal that under-reports freshness is fail-safe — it prompts a human to look. The dangerous failure is the opposite one, and this change cannot produce it. Contrast the pre-fix state, which was unbounded and monotonically wrong: ~24 days of rot while the file was written continuously.

The error is also **bounded**. The non-`_persist` writers cluster next to lifecycle transitions — counters at wrapup, allocation at scope — so the worst-case lag is an intra-wave gap of minutes to hours, against ≥5 stamping transitions bracketing every wave. Scoped, not incomplete.

### The replay is reproducible — I re-derived it

I did not take the status-file diff on trust. I ran the two transitions myself against the real base file with the PR's `--at` values:

```
lifecycle.py wave wrapup 25 --at 2026-07-19T01:29:00Z
lifecycle.py wave retro  25 --at 2026-07-19T01:32:00Z
```

My output matches the PR's head file exactly — identical values and **identical top-level key order** — with a single difference, `last_updated`, which carries my wall clock instead of theirs. That is the designed behaviour, and it is a clean demonstration that the replay is deterministic and the file shape survives.

The `--at` values are sourced, not invented. Both were already in the file: `01:29:00Z` as `wave_25_completed_at`, and `01:32:00Z` as `wave_25_annunaki_attack_ran_at` / `wave_25_memory_audit_ran_at`. Those two markers were written by commit `ab0f55e` — the wave-25 retro commit itself, shipped in PR #982 — so `01:32:00Z` is the retro's own stamp rather than a convenient nearby number.

Exactly three top-level keys changed, and nothing else moved:

- `wave_25_retro_completed_at` added at `01:32:00Z`
- `last_completed_wave` `wave-24` → `wave-25`
- `last_updated` refreshed

`wave_25_completed_at` was re-upserted with a byte-identical value and **was not overwritten**. No counter moved — `wave_25_final_pr_count` stays 7, `global_wave_seq` stays 25, `wave_active` stays false. This is consistent with `wrapup`'s documented contract that it deliberately writes no counter.

One thing worth noting because it could have gone wrong quietly: the real file contains ~19 lines with a `last_updated` key, most of them nested inside sub-objects. The upsert targeted the top-level key and left every nested one untouched — confirmed by my replay diff, since a nested mutation would have surfaced as a change in its parent key. It did not.

### Wave-26 gate state is correct, not merely non-empty

As PD this is the part I care most about. Step 0a resolves the prior-wave timestamp as `wave_25_retro_completed_at // wave_25_completed_at`. Before this PR the fallback already supplied `01:29:00Z`, so wave-26 was never in the permissive-skip branch — the gate was armed either way. After the PR it compares against `01:32:00Z` instead, three minutes stricter.

The key asserts a fact, so I checked the fact rather than the format: wave-25's retro genuinely ran (PR #982, merged `2026-07-19T01:46:18Z`). The stamp is a lower bound on the true retro completion, and it post-dates `wave_25_completed_at`, so the wrapup→retro ordering the gate assumes is preserved. Wave-26 will require its `wave_26_scope_reconciled_at` to post-date `01:32:00Z` — which a future `/wave-scope 9 26` will satisfy with a fresh wall-clock stamp. Correct gate state, not a cosmetically-filled one.

### Tests are capable of failing — I mutated them myself

Per this org's "a high number is as suspect as a zero", I did not accept the reported 11-fail/29-pass. I rebuilt the module at head in a scratch tree and ran four mutations. Baseline is 29 passed.

| mutation | result |
|---|---|
| stamp removed entirely (pre-PR behaviour) | **11 failed**, 23 passed |
| `--at` threaded into `last_updated` | **1 failed** — `test_at_backdates_the_event_but_not_the_file` |
| explicit `last_updated` clobbered | **1 failed** — `test_explicit_last_updated_in_pairs_wins` |
| stamp moved before the empty-pairs guard | **1 failed** — `test_empty_pairs_writes_nothing` |

Every mutation is caught, each by a distinctly-named test rather than by collateral breakage, and the author's figure reproduces exactly. No inert assertions.

On the specific durability question for `test_at_backdates_the_event_but_not_the_file`: the hard-coded `assertGreater` against `2026-01-01` does **not** decay. Wall clock only increases, so the assertion stays true and stays capable of catching a backdate to any pre-2026 value — and the test passes `_ANCIENT` = `2020-01-01`, which both that assertion and the `assertNotEqual` catch independently (mutation 2 confirms). The lexical string comparison is sound: these are fixed-width, zero-padded, UTC-`Z` ISO-8601 stamps, so lexical order is chronological order. The only case it would miss is a hypothetical `at` in the *future*, which no replay produces and which the `assertNotEqual` still covers for the value actually passed.

### Excluding `last_updated` from the idempotency test is the right call

The exclusion does not weaken the guarantee, because the guarantee has been reconstructed rather than dropped. `test_idempotent_re_write` existed to prove two things — value stability and format stability — and comparing raw text conflated them. With a wall-clock key, raw-text equality is genuinely non-deterministic: green unless the two calls straddle a second boundary. That is a flaky test, not a strong one.

Value stability is now asserted on the parsed dicts minus the one key that is *supposed* to move, plus an explicit `assertGreaterEqual` that the marker never regresses. Format stability, which the raw-text compare was incidentally covering, is carried by `test_compact_inline_shape_not_reflowed` and the newly added `test_last_updated_is_compact_inline_too`. Both halves survive, and the flake is gone. Right trade.

### CI

I ran `statusCheckRollup` at the reviewed head myself: **21 checks, all COMPLETED / SUCCESS** — ruff lint, ruff format, mypy, pytest (hooks + lib), cspell, markdown lint, lychee, actionlint, commit-author gate, roster-union gate, pre-commit⇄CI sync-drift gate, memory + headcount budget gates, and the rest. No pending, no empty rollup, nothing skipped. Merge state MERGEABLE.

### Tech debt filed

I found one adjacent defect, filed as #1037 and added to project 2. `last_completed_at` is a vestigial flat key with no writer and no reader anywhere in `.claude/` or `ontology/` — and it is inconsistent with its own pair key. This PR advances `last_completed_wave` to `wave-25` while `last_completed_at` stays at `2026-06-23T18:30:06Z`, which corresponds to neither wave-24 nor wave-25.

To be clear about attribution: this is **not** a defect in #1034. The key had no writer before this PR and has none after, and nothing reads it, so there is no live mis-gate and no reason to hold the merge. But it is the same disease #1033 diagnosed, one key over, and it should not be allowed to rot until some future reader trusts it.

### Verdict

The design claim is correct and I could not construct a counter-example. The replay is reproducible from the real file, the values are sourced, no counter moved, the gate lands in a correct state, and the tests demonstrably fail when the behaviour is removed. Ontology conventions are updated in the same PR so the "do not fix this to honour `--at`" reasoning is durable rather than living only in this thread — which is the right instinct, since the next reader's first impulse will be to "fix" it.

Approved for merge.

---
Requestor: Nadia Khoury
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1037

Issue #1039, opened 2026-07-20
ontology_tracker: SKIP_PATTERNS denylist leaks for gitignored machine-local files — generalize via owning-repo check-ignore


parametrization:
## Summary

`.claude/hooks/ontology_tracker.py` decides what to checksum-track via `SKIP_PATTERNS`, a hand-maintained substring **denylist**. It has now leaked twice for the *same class* of file — gitignored, machine-local, frequently-rewritten artifacts that dirty the committed `ontology/checksums.json`:

- `.claude/annunaki/errors.jsonl` (added earlier)
- `.claude/memory/session_handoff.md` (#1038)

Each leak manufactures permanent phantom drift in `/session-start` Step 3a until someone notices and hand-extends the list. The next gitignored machine-local artifact will leak the same way.

## Why the naive generalization is wrong

#1038 originally proposed deriving the skip from `git check-ignore`. That was **tested and rejected** — see the pre-implementation comment on #1038. The parent repo `.gitignore`s its child repos wholesale, so `git check-ignore` run from `REPO_ROOT` reports **every child-repo file as ignored**, and child-repo entries are **147 of 284** tracked entries (52%) — including `noorinalabs-deploy/.github/workflows/deploy-prod.yml` and every child ontology source.

That change would silently blind the tracker to half the semantic overlay *while presenting as "0 dirty"* — the gate would look healthier while going blind. Strictly worse than the nuisance it fixes.

## What a correct fix requires

1. **Resolve each path's owning repo** — walk up to the nearest `.git` ancestor of the file, not `REPO_ROOT`. A child-repo file's gitignored-ness is a question for *its own* repo.
2. **Run `git check-ignore` in that repo**, on the repo-relative path.
3. **Fail OPEN** — on any error (no `.git` ancestor, subprocess failure, timeout) **track the file anyway**. Not-tracking is a *silent* loss of drift detection; over-tracking is merely noise. The asymmetry is one-sided and must be encoded deliberately.
4. **Cost/caching** — this adds a subprocess per Edit/Write in a PostToolUse hook. Cache the per-directory answer for the hook's lifetime, or batch. Measure before shipping.
5. **Keep `SKIP_PATTERNS`** as the fast path — `check-ignore` is the backstop for the class, not a replacement for the explicit entries (`checksums.json` self-skip, `ontology/structural/`, worktree paths are policy decisions, not gitignore facts).

## Tests this needs

[...]

Issue #981, opened 2026-07-18
validate_pr_review FAILS OPEN on unresolvable repo (gh pr merge -R $VAR) — 2-reviewer gate silently bypassed


parametrization:
## Summary

`validate_pr_review.py` (Hook 4, the 2-reviewer gate) **fails OPEN when it cannot resolve the target repo from the merge command** — specifically when `gh pr merge <N> -R $VAR` carries an **unexpanded shell variable**. The hook parses the command string as-is (pre-expansion), so `$DA` is not a resolvable `owner/repo`; `_resolve_owner_repo` returns None, `check_comment_reviews` returns an empty result (0 reviewers), and the merge is **ALLOWED** instead of blocked.

## Reproduction (verified)
```python
import validate_pr_review as v
v.check({"tool_name":"Bash","tool_input":{"command":"gh pr merge 451 -R $DA --merge --delete-branch"}})
# → {'decision': 'allow'} ... (None/allow)  ← WRONG: should block or fail-closed
v.check({"tool_name":"Bash","tool_input":{"command":"gh pr merge 451 --repo noorinalabs/noorinalabs-data-acquisition --merge --delete-branch"}})
# → {'decision': 'block', ...}  ← correct
```

## Impact (observed, Phase 9 Wave 25)
Four feature→wave PRs (da#449/#450/#451/#456) were merged with `gh pr merge <N> -R $DA …`. Combined with a separate wrong-verdict-format issue (reviewers briefed with `RequestOrReplied: Request` instead of `Approved`, so 0 hook-counted approvals), the 2-reviewer gate **did not enforce on any of the four merges**. The reviews were substantively genuine + thorough and CI was green, so this is a gate-integrity/audit-trail breach, not a code-quality one — but the gate silently allowed merges it could not verify.

## Root cause + fix direction
- `_resolve_owner_repo` → None (unresolvable `-R $VAR`) must **fail CLOSED**, not allow. This is the same principle as the `CommitFetchError` hard-block (feedback_safety_direction_over_ux_friction: when a hook can't verify, HARD BLOCK, never allow-with-log). A merge command whose repo the gate cannot determine is exactly a case it cannot verify.
- Additionally: when `-R` is present but unresolvable AND the cwd `gh repo view` fallback also can't disambiguate (or resolves a DIFFERENT repo than the PR), block with a diagnostic naming the unexpanded-variable footgun.
[...]

PR #1040, opened 2026-07-20
fix(#1038): stop tracking gitignored session_handoff.md in ontology checksums

### LINKED ISSUE #1038: ontology_tracker: gitignored session_handoff.md is tracked — every /handoff dirties committed checksums.json
## Summary

## Evidence

```

# → not tracked in git
```

```
       "last_resolved": "e6f4878d…",
```

## Root cause

```python
SKIP_PATTERNS = [
    ...
]
```

```
```

## Proposed fix

## Impact

parametrization:
## What changed and why

`.claude/memory/session_handoff.md` is gitignored (`.gitignore:117`) and untracked in git, but the ontology change-tracker hook tracked it anyway. The `Stop` hook rewrites that file after ~every response (~5min throttle), so **every session dirtied `ontology/checksums.json` — a committed file**. `/session-start` Step 3a reported phantom drift and `/ontology-rebuild` had a phantom entry to resolve, forever, eroding a gate whose only value is that "0 dirty" means something. `SKIP_PATTERNS` already carried the exact sibling case (`.claude/annunaki/errors.jsonl` — gitignored, machine-local, constantly rewritten); the handoff was simply a missed entry. This adds it, drops the now-stale `checksums.json` entry, fixes the `ensure_ascii` churn, and adds regression tests.

## Evidence the fix works

Not just "the test passes" — the real hook was exercised against the **real** `session_handoff.md` on disk, with `REPO_ROOT` anchored to the actual repo so path filtering behaves exactly as in production, and writes redirected to a scratch copy of the real `checksums.json`. Comparing `main`'s hook against this branch's:

```
=== BEFORE (main) ===
  check() returned      : {'action': 'tracked', 'path': '.claude/memory/session_handoff.md'}
  checksums.json changed: True
  handoff entry present : True
    last_tracked : b18c7fe0aab8071d      <- differs from last_resolved
    last_resolved: 'e6f4878d8898d3e5…'
  total dirty entries   : 1 ['.claude/memory/session_handoff.md']

=== AFTER  (fix) ===
  check() returned      : None
  checksums.json changed: False
  total dirty entries   : 0 []
```

That is the defect reproduced and then eliminated: one simulated `Stop`-hook handoff write takes the tree from 0 dirty to 1 dirty on `main`, and leaves it untouched here.

Second run, against **this branch's** `checksums.json` (entry already removed) — confirming the entry does not come back:

```
=== fix + PR checksums, simulate a Stop-hook handoff write ===
  handoff entry re-created: False
  file byte-identical     : True
  dirty entries           : 0
```

And the `ensure_ascii` half, tracking a real overlay file (`ontology/domain.yaml`):

```
BEFORE (main): escaped-\u sequences = 2, literal em-dash present = False
AFTER  (fix) : escaped-\u sequences = 0, literal em-dash present = True
```

Independently corroborated by the `checksums.json` diff in this PR: removing the entry via an `ensure_ascii=False` round-trip produced **exactly 6 deleted lines and no other change** — i.e. the round-trip is byte-stable against the file as committed.

## `ensure_ascii` consistency across writers — all writers aligned

I grepped for every writer of `checksums.json`, since fixing one writer while another still defaults to `ensure_ascii=True` merely moves the flip-flop. Findings:

| Writer | Kind | Status |
|---|---|---|
| `.claude/hooks/ontology_tracker.py` | programmatic, committed | **Fixed** — `ensure_ascii=False` |
| `/ontology-rebuild` resolver (`.claude/skills/ontology-rebuild/SKILL.md` § 4) | agent-driven, no serialization specified | **Fixed** — step 4 now mandates `json.dump(..., indent=2, ensure_ascii=False)` + trailing newline |
| `.claude/lib/sync_main.py` | allowlists the path for stash/restore | Not a writer of contents — no change needed |
| `.claude/hooks/session_handoff.py`, `.claude/lib/generic_prompt_tracker.py` | read / skip-list only | Not writers — no change needed |
| `.claude/scratch/wave-merge-bodies/ontology-resolve.py` | one-off, **untracked** in git (`git ls-files` → empty) | Out of scope; not a committed writer |

So: **all writers aligned.** The resolver was the genuine second source of the flip-flop — it is agent-driven and had no serialization specified at all, which is exactly why the file oscillated. Worth a reviewer's eye: that fix is a documented instruction rather than enforced code, because the resolver is a skill, not a module. If a reviewer prefers hard enforcement, the resolver's checksum write should move into a shared `.claude/lib/` helper — I'd support that but it is beyond this bugfix's scope.

## The `check-ignore` generalization was considered and rejected

The issue's proposed fix **point 4** suggested deriving the skip from `git check-ignore` rather than another hardcoded entry. **Deliberately not taken here.** The parent repo `.gitignore`s its child repos wholesale, so `git check-ignore` from `REPO_ROOT` reports every child-repo file as ignored — and child-repo entries are **147 of 284 tracked entries (52%)**, including `noorinalabs-deploy/.github/workflows/deploy-prod.yml` and every child ontology source. That change would silently blind the tracker to half the semantic overlay *while presenting as "0 dirty"* — the gate would look healthier while going blind, which is strictly worse than the nuisance being fixed.

The denylist has now leaked twice for the same class, so the class-level fix is worth doing — properly. Filed as **#1039**, carrying the requirements a correct version needs: resolve each path's **owning** repo (nearest `.git` ancestor) and run `check-ignore` there, **fail open** (track anyway on any error, since not-tracking is silent loss of drift detection), cache the subprocess cost, and keep the 147/284 child-repo case as an explicit regression test.

## Tests

Added to `.claude/hooks/tests/test_ontology_tracker.py`, matching the existing conventions there (`_FakeRepoRootMixin` for invocation-location independence, per #686):

- `ShouldSkipSessionHandoffTests` — relative + absolute handoff paths skipped; an **end-to-end** case driving `check()` against a real file and asserting the checksums file is byte-unchanged; and `test_other_memory_notes_are_still_tracked`, guarding that the skip is scoped to the single gitignored handoff and does **not** silently drop the committed `.claude/memory/` store from drift detection.
- `ChecksumsSerializationTests` — non-ASCII `description` survives a write unescaped, and repeated tracking of an unchanged file is byte-stable modulo the legitimately-moving `tracked_at`.

## Gate results

| Check | Status |
|---|---|
| `pre-commit run --all-files` (ruff-format, ruff-lint, actionlint, cspell, dockerfile-base-pin, fixture-realism, skill-graphql-pagination) | Passed |
| `mypy .claude/hooks/ .claude/lib/` | Passed — no issues in 174 source files |
| `pytest .claude/hooks/tests/ .claude/lib/tests/` | Passed — 2637 passed, 88 subtests |
| Push-stage (mypy, pytest, pytest-skills, memory-budget, headcount-budget, doc-freshness, office-drift, mermaid-render) | All passed |
| `pre_commit_ci_sync.py .` (drift gate) | `OK: pre-commit config mirrors all CI-enforced checks.` |

No known-failing checks, pre-existing or otherwise.

Closes #1038

### DIFF
--- a/.claude/hooks/ontology_tracker.py
+++ b/.claude/hooks/ontology_tracker.py
@@ -71,6 +71,11 @@ SKIP_PATTERNS = [
     "ontology/checksums.json",  # Don't track ourselves
     "ontology/structural/",  # GENERATED structural layer — see note below (#857)
     ".claude/annunaki/errors.jsonl",
+    # Gitignored, machine-local, rewritten by the Stop hook after ~every
+    # response — same class as annunaki/errors.jsonl. Tracking it dirtied the
+    # COMMITTED checksums.json every session, so /session-start reported
+    # phantom drift forever (#1038).
+    ".claude/memory/session_handoff.md",
     "__pycache__/",
     ".pyc",
     "node_modules/",
@@ -223,7 +228,10 @@ def check(input_data: dict) -> dict | None:
         CHECKSUMS_FILE.parent.mkdir(parents=True, exist_ok=True)
         tmp = CHECKSUMS_FILE.with_suffix(".tmp")
         with open(tmp, "w", encoding="utf-8") as f:
-            json.dump(data, f, indent=2)
+            # ensure_ascii=False: the on-disk file holds literal UTF-8 in its
+            # top-level ``description``. Re-escaping it here made every touch
+            # emit a spurious one-line diff (#1038).
+            json.dump(data, f, indent=2, ensure_ascii=False)
             f.write("\n")
         tmp.replace(CHECKSUMS_FILE)
     except OSError:
--- a/.claude/hooks/tests/test_ontology_tracker.py
+++ b/.claude/hooks/tests/test_ontology_tracker.py
@@ -245,11 +245,55 @@ class ShouldSkipSessionHandoffTests(_FakeRepoRootMixin, unittest.TestCase):
     """
 
     def test_relative_handoff_path_is_skipped(self):
-        self.assertTrue(hook._should_skip(".claude/memory/session_handoff.md"))
+        """A repo-relative handoff path is skipped BY THE PATTERN, not by luck.
+
+        The ``os.chdir`` here is load-bearing — do not remove it (#1043).
+        ``_should_skip`` resolves a relative path against the *process cwd*, not
+        against the patched ``REPO_ROOT``. Without the chdir this path resolves
+        somewhere outside the fake root and is caught by the pre-existing
+        out-of-repo rule, so the assertion passes even when the
+        ``SKIP_PATTERNS`` entry under test is deleted — an inert test that
+        reports green while covering nothing. Anchoring cwd to the fake root
+        puts the path *inside* the repo, so the pattern is the only thing that
+        can produce the skip and the test genuinely dies if it is removed.
+
+        Relative paths are worth covering: the tracker is anchored on the
+        orchestrator cwd and records relative paths in real flows (see the
+        module docstring on worktree-relative paths).
+        """
+        cwd = os.getcwd()
+        try:
+            os.chdir(self._fake_root)
+            self.assertTrue(hook._should_skip(".claude/memory/session_handoff.md"))
+        finally:
+            os.chdir(cwd)
 
     def test_absolute_handoff_path_is_skipped(self):
         path = str(self._fake_root / ".claude" / "memory" / "session_handoff.md")
         self.assertTrue(hook._should_skip(path))
+
+    def test_skip_is_scoped_to_the_claude_memory_directory(self):
+        """The pattern must stay DIRECTORY-scoped, not a bare filename (#1043).
+
+        Narrowing the entry to ``"session_handoff.md"`` left the whole suite
+        green, so nothing pinned the scoping. A substring denylist matches
+        anywhere in the path, so a bare filename would silently stop tracking
+        any committed file that happens to share the name — e.g. a real
+        ``docs/session_handoff.md``. Only the gitignored machine-local file at
+        ``.claude/memory/`` is exempt; a same-named file elsewhere in the repo
+        is ordinary tracked content.
+        """
+        elsewhere = str(self._fake_root / "docs" / "session_handoff.md")
+        self.assertFalse(hook._should_skip(elsewhere))
+
+    def test_skip_does_not_extend_to_sibling_memory_notes_by_prefix(self):
+        """A path merely *starting* with the handoff name is not exempt (#1043).
+
+        Guards the other narrowing direction — the pattern must match the whole
+        handoff path, so a distinct committed note is unaffected.
+        """
+        sibling = str(self._fake_root / ".claude" / "memory" / "session_handoff_notes.md")
+        self.assertFalse(hook._should_skip(sibling))
 
     def test_check_writes_no_entry_for_handoff(self):
         """End-to-end: a Write to the handoff produces no checksums entry.
--- a/.claude/hooks/tests/test_ontology_tracker.py
+++ b/.claude/hooks/tests/test_ontology_tracker.py
@@ -11,6 +11,7 @@ Or:  python3 .claude/hooks/tests/test_ontology_tracker.py
 
 from __future__ import annotations
 
+import json
 import os
 import sys
 import tempfile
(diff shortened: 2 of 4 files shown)


parametrization:
Requestor: Lucas Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1041, #1042

Certified at head SHA `7428f25537162e6b557c7ca0b7de61c5aab616a9` (re-checked immediately before posting — unchanged). All 22 CI checks green at that SHA.

**Scope check (per my brief):** confirmed the PR does not sneak in the `git check-ignore` generalization — the diff is exactly the scoped one-line `SKIP_PATTERNS` entry, the `ensure_ascii=False` fix, the stale checksums entry removal, and tests. #1039 correctly carries the deferred generalization.

**1. Skip scoping — correct, verified against the actual hook.** Read `.claude/hooks/ontology_tracker.py` at the head SHA: `_should_skip` substring-matches `.claude/memory/session_handoff.md` before any path resolution, so it can't be defeated by resolution quirks, and it can't over-match `.claude/memory/section_*.md` (different string, no shared suffix beyond the directory). `test_other_memory_notes_are_still_tracked` does guard this correctly — it asserts `_should_skip` is False for `section_ci_tooling.md`, which is genuinely a scoping-safety check (true in both pre- and post-fix hook, as it should be).

**2. `ensure_ascii=False` consistency — acceptable as documentation-only for this PR, not blocking.** The tracker hook (the one committed, programmatic writer) is code-enforced. The `/ontology-rebuild` resolver fix is SKILL.md prose only, since the resolver is agent-driven, not a code module — there's no shared-code choke point to enforce it at today. I don't think this should block #1040: the PR's own writer audit is thorough (grepped every touchpoint, correctly excluded non-writers), the fix closes the one enforced-by-code gap, and the author already flagged the enforcement gap themselves as future scope. Filed **#1042** to track extracting a shared `checksums.json` write helper so this can't leak the same way through the resolver later.

**3. checksums.json diff — confirmed clean.** `gh pr diff` shows exactly one `@@` hunk in the file, removing precisely the 6-line `.claude/memory/session_handoff.md` entry. Nothing else in the 284-entry file is touched.

**4. Test quality — mixed, verified by actually running the new tests against the pre-fix hook (not assumed.)** I checked out `main`'s `ontology_tracker.py` into a scratch copy, dropped the PR's new test classes on top, and ran them:

- `test_absolute_handoff_path_is_skipped`, `test_check_writes_no_entry_for_handoff` (end-to-end), `test_non_ascii_description_survives_a_write_unescaped`, `test_write_is_byte_stable_across_repeated_tracking` — all **FAIL** against the pre-fix hook and pass post-fix. These are real regression tests.
- `test_relative_handoff_path_is_skipped` — **passes against the pre-fix hook too.** I confirmed why: it never `os.chdir`s into the fake `REPO_ROOT` (unlike the sibling `test_in_repo_relative_path_is_tracked`, which does), so the relative path resolves against the real process cwd and lands outside the fake root — tripping the pre-existing "outside repo" catch-all in `_should_skip`, not the new `SKIP_PATTERNS` entry. I verified directly: removing the new pattern from `SKIP_PATTERNS` and re-running `_should_skip` on the exact relative path still returns `True`. This test is vacuous as written.

Not blocking — the production behavior for relative paths is unaffected (substring matching happens before path resolution either way), and the absolute-path + end-to-end tests already give real coverage of the fix. But it's a genuine test-suite defect worth fixing so it doesn't mislead the next reader into thinking the relative case is regression-tested. Filed **#1041**.

**Verdict: Approve.** The fix is correctly scoped, the checksums diff is clean, and the core regression coverage is real (verified, not assumed). Two non-blocking findings filed as tech debt, both already substantially anticipated in the PR's own body.

parametrization:
Requestor: Nino Kavtaradze
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1043, #1044

Merge-gate review at head SHA `7428f25537162e6b557c7ca0b7de61c5aab616a9` (confirmed via `--json headRefOid`; merge-base is `bc17fe4` = current `origin/main`, no rebase needed). Worktree `.worktrees/A.Virtanen-1038-tracker-skip-handoff` verified at that SHA and clean before any test execution. 21 check-runs at the head SHA, all SUCCESS.

My lens was the one this repo keeps getting bitten by: a gate that goes blind while still reporting green (#981, P9W25 pain point #2). This PR removes coverage from a drift-detection gate on purpose, so the question is whether it removes exactly the intended byte and not one more. It does.

## 1. Blast radius of the substring skip — enumerated, not reasoned

`_should_skip` substring-matches `.claude/memory/session_handoff.md` against the raw path. Enumerating what that can actually match:

- Parent repo tracked files: `git ls-files | rg session_handoff` returns `.claude/hooks/session_handoff.py` and `.claude/hooks/tests/test_session_handoff.py` — neither contains the pattern (`.py`, and not under `.claude/memory/`).
- All seven child repos: `git -C <repo> ls-files | rg session_handoff` returns nothing in any of them. No committed file anywhere under the root has a path containing this substring.
- Child-repo paths (`noorinalabs-*/.claude/memory/session_handoff.md`) would match — but no such file is tracked or present. Four child repos already `.gitignore` the path; the other three simply have no handoff file.
- Worktree-relative and absolute forms both match, and both were already skipped by the worktree-component and out-of-repo filters.

Net: the new pattern excludes exactly one path, and that path is untracked and gitignored.

## 2. The committed memory store stays tracked

Confirmed against the artifact, not the framing. `checksums.json` at head still carries all 34 parent `.claude/memory/*` entries (`MEMORY.md`, every `feedback_*`/`project_*`/`reference_*`/`user_*` note) plus the child-repo memory entries. `test_other_memory_notes_are_still_tracked` is the guard, and it is live: widening the pattern to `".claude/memory/"` — the exact over-skip failure shape — fails it (`AssertionError: True is not false`).

## 3. Mutation testing — the tests fail against the pre-fix hook

Ran in an isolated sandbox copy so nothing touched the PR tree. Baseline at head: 29 passed.

- Delete the new `SKIP_PATTERNS` entry (restore the pre-#1038 hook) -> 2 failed. The end-to-end failure is the real defect: `{'action': 'tracked', 'path': '.claude/memory/session_handoff.md'} is not None`, with a checksums write.
- Revert `ensure_ascii=False` -> 2 failed, on the actual re-escaped bytes (`"overlay \\u2014 \\u00d7 scope"`).
- Widen to `.claude/memory/` -> 1 failed (the memory-store guard).

Three independent mutations, three kills. The regression tests are not green on both sides.

## 4. `checksums.json` — exactly one entry, no collateral

Parsed head vs merge-base: 284 -> 283 entries; removed `{'.claude/memory/session_handoff.md'}`; added `set()`; zero entries with changed values; key order preserved; all non-`files` top-level keys byte-identical; 0 `\\u` escapes on both sides. No reordering, no re-encoding.

Also confirmed no `git check-ignore` logic leaked in — the diff is four files and contains no subprocess or ignore-probe of any kind. The #1039 deferral is clean.

## The judgement you escalated — documented-but-unenforced `ensure_ascii`

Acceptable as the closing half, and I would not hold the merge for it. The reason: the flip-flop is diff noise, not a blind gate. It cannot make the tracker under-report drift — the failure mode this PR exists to kill is "0 dirty means nothing," and a re-escaped `description` does not touch that. The only automated writer of the file is now the hook, and it is test-pinned; a resolver pass that re-escaped would produce one visible line in a reviewed diff and self-heal on the next tracker write.

That said, a documented convention with no failure mode other than a reviewer noticing is worth two lines of enforcement, so I filed #1044: assert the committed `checksums.json` contains no `\\u` escapes, in the existing Config — YAML/JSON syntax job plus the pre-commit mirror. It starts green at head. Follow-up, not a blocker.

## Non-blocking findings (filed, not blocking)

**#1043 — one of the six new tests is inert.** `test_relative_handoff_path_is_skipped` survives the mutation that deletes the `SKIP_PATTERNS` entry (`1 passed, 28 deselected`). It passes a *relative* path, which `_should_skip` resolves against the process cwd rather than the patched fake `REPO_ROOT`, so it is skipped by the out-of-repo rule and never reaches the new pattern. It certifies nothing. The two sibling tests carry the guard, so real coverage is intact — this is the `feedback_fixture_makes_guard_assertion_inert` shape, worth cleaning up rather than blocking on.

Same issue also records that no test pins the *directory* scoping: narrowing the pattern to bare `"session_handoff.md"` leaves all 29 green. Shipped pattern is the correct one; the suite just would not catch a future narrowing.

Commit identity checks out (Aino Virtanen <parametrization+Aino.Virtanen@gmail.com>), branch naming conforms, and the SKILL.md note is accurate to the hook it references. Approved — this removes precisely the coverage it set out to remove.

parametrization:
## #1043 addressed inline — `ac8bcfa` (additive; no rebase, no force-push)

Both reviewers were right, and the finding is well taken: an inert test shipped inside a PR about a gate that reports green while blind is the same defect in miniature. Fixed here rather than deferred.

### Finding 1 — `test_relative_handoff_path_is_skipped` was inert

`_should_skip` resolves a relative path against the **process cwd**, not the patched fake `REPO_ROOT`. The path therefore landed outside the repo and was caught by the pre-existing out-of-repo rule, never reaching the `SKIP_PATTERNS` entry it claimed to cover.

Repaired by anchoring cwd to the fake root — matching the existing `test_in_repo_relative_path_is_tracked` convention already in this file — so the path resolves *inside* the repo and the new pattern is the only thing that can produce the skip. Relative coverage is worth keeping rather than deleting: the tracker is anchored on the orchestrator cwd and records relative paths in real flows (per the module docstring on worktree-relative paths). The `os.chdir` is **load-bearing and commented as such**, so a future reader does not strip it as incidental and silently re-inert the test.

### Finding 2 — directory scoping was unpinned

Confirmed: narrowing the entry to a bare `"session_handoff.md"` left all 29 tests green. Because `SKIP_PATTERNS` is a *substring* denylist, a bare filename would silently stop tracking any committed file that merely shares the name. Added guards in **both** narrowing directions — a same-named file elsewhere (`docs/session_handoff.md`) must still track, and a prefix-broadened pattern must not swallow a sibling note.

### Per-mutation kill results (tracker suite, 31 tests)

| Mutation | Before `ac8bcfa` | After `ac8bcfa` | Killed by |
|---|---|---|---|
| Delete the `SKIP_PATTERNS` entry | 2 failed / 27 passed — **relative test survived** | **3 failed / 28 passed** | `test_relative_handoff_path_is_skipped` now among the failures, plus the absolute + end-to-end tests |
| Narrow to bare `"session_handoff.md"` | **0 failed / 29 passed — unpinned** | **1 failed / 30 passed** | `test_skip_is_scoped_to_the_claude_memory_directory` |
| Broaden to prefix `".claude/memory/session_handoff"` | n/a | **1 failed / 30 passed** | `test_skip_does_not_extend_to_sibling_memory_notes_by_prefix` |

The third mutation is included deliberately: it proves the new prefix guard is not itself inert. Every test added or repaired here is demonstrated to fail against the mutation it targets — no assertion is taken on trust.

The mutations were applied to a copy and reverted after each run; `git diff --stat .claude/hooks/ontology_tracker.py` is empty on the committed tree, so only the test file changed in this commit.

### Gates (all re-run at `ac8bcfa`)

| Check | Status |
|---|---|
| Unmutated tracker suite | 31 passed |
| `pre-commit run --all-files` | Passed |
| `mypy .claude/hooks/ .claude/lib/` | Passed — no issues in 174 source files |
| `pytest` (hooks + lib) | 2639 passed, 88 subtests |
| `pre_commit_ci_sync.py .` | `OK: pre-commit config mirrors all CI-enforced checks.` |
| Push-stage hooks | All passed |

Closes #1043. #1044 (CI-enforce the `ensure_ascii` convention) left as a genuine follow-up — it touches the CI workflow plus the pre-commit mirror and is out of scope here.

parametrization:
Requestor: Lucas Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Re-confirm at new head `ac8bcfa96f157f4801543dc29b08f2b4139f1307`. Confirmed `7428f25` is an ancestor (fast-forward, additive) and the delta is exactly `.claude/hooks/tests/test_ontology_tracker.py` (+45/−1); `git diff 7428f25 ac8bcfa -- .claude/hooks/ontology_tracker.py` is empty — production hook byte-identical. All CI checks green at `ac8bcfa`.

Independently re-verified all three claims by actually running the mutations myself, not trusting the PR body's numbers:

**1. Repaired relative test.** Reran my scratch pre-fix hook (no `session_handoff.md` entry in `SKIP_PATTERNS` at all) against the updated test file: `ShouldSkipSessionHandoffTests` now goes from 2 passed/4 failed → **3 passed/3 failed**, and `test_relative_handoff_path_is_skipped` is now among the failures. The `os.chdir(self._fake_root)` fix is confirmed load-bearing — it now genuinely dies when the fix is absent.

**2. `test_skip_is_scoped_to_the_claude_memory_directory`.** Built a mutation of the hook narrowing the pattern to bare `"session_handoff.md"`. Result: exactly this one test fails (`docs/session_handoff.md` gets wrongly skipped), all 5 others in the class still pass. Clean, isolated kill.

**3. `test_skip_does_not_extend_to_sibling_memory_notes_by_prefix`.** Built a second mutation broadening the pattern to the prefix `".claude/memory/session_handoff"` (dropping `.md`). Result: exactly this one test fails (`.claude/memory/session_handoff_notes.md` gets wrongly skipped), all 5 others pass. Confirmed — this is the mutation Aino added after finding the prefix-guard test didn't die under the narrow mutation; verified it now does die under its own dedicated mutation.

**Judgement call (3): I accept the cwd-dependent relative-path test as-is, no restructuring needed.** A relative path's resolution is inherently cwd-dependent — that's the real behavior of `_should_skip`/`Path.resolve()`, not a test artifact, and the tracker itself operates on orchestrator-cwd-relative paths in production per the module's own worktree-path docstring, so exercising that dependency is representative, not incidental. The `os.chdir`+`try/finally` pattern is pre-existing convention in this same file (`test_in_repo_relative_path_is_tracked`), so this doesn't introduce a new idiom — it aligns with one already reviewed and shipped. The load-bearing comment plus the mutation-test evidence above is sufficient guard against the original inertness class recurring; forcing cwd-independence here would add complexity (mocking `os.getcwd`/`Path.resolve`) for no coverage gain over what direct cwd anchoring already gives.

No new findings — TechDebt: none (my earlier #1041 was correctly closed as a duplicate of #1043, which itself is now fully verified).

parametrization:
Requestor: Nino Kavtaradze
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1045

Re-confirm at head `ac8bcfa96f157f4801543dc29b08f2b4139f1307`. Verified independently: `7428f25` is an ancestor of `ac8bcfa` (fast-forward, no rebase or force), the delta is one file +45/-1 tests-only, `git diff` on `.claude/hooks/ontology_tracker.py` between the two SHAs is empty, and 21/21 check-runs at the new head are SUCCESS. My prior verdict stands; this is a fresh read of the changed surface.

## 1. Mutation battery re-run at the new head — her numbers are exact

Baseline at `ac8bcfa`: 31 passed. Fresh sandbox built from `git show ac8bcfa:...`, so nothing touched the PR tree.

| Mutation | Killed |
|---|---|
| delete the `SKIP_PATTERNS` entry (pre-fix hook) | **3** — `relative`, `absolute`, `check_writes_no_entry` |
| revert `ensure_ascii=False` | 2 — both serialization tests |
| widen to `".claude/memory/"` | 2 — memory-store guard + the new prefix test |
| narrow to bare `"session_handoff.md"` | **1** — the new directory-scoping test |

Every reported figure confirmed. The repaired relative test now joins the delete-mutation kills (2 -> 3) and the narrow mutation went 0 -> 1. The `os.chdir` repair is the right fix and is genuinely load-bearing: it anchors cwd to the fake root so the path lands *inside* the repo, which is the only way the pattern rather than the out-of-repo rule can produce the skip. The docstring naming the chdir as load-bearing is worth keeping.

## 2. The self-caught fourth mutation — confirmed, and it was the right catch

Broaden-to-prefix `".claude/memory/session_handoff"` (dropping `.md`) kills `test_skip_does_not_extend_to_sibling_memory_notes_by_prefix` and nothing else. So that test is not inert, and it is the *only* thing standing in that direction.

Worth stating plainly: that test would have shipped green against all four originally-specified mutations. Catching that unprompted is the finding's own shape applied one level up, and it is exactly the discipline #1043 was about.

## 3. Is the mutation set complete? No — one direction remains unpinned

Honest answer to the direct question. I ran two further widening directions and one survives:

- `"/session_handoff.md"` -> killed by the directory-scoping test.
- **`"memory/session_handoff.md"` (widen leftward, dropping the `.claude/` anchor) -> 31 passed, survives.**

`test_skip_is_scoped_to_the_claude_memory_directory` uses `docs/session_handoff.md`, which has no `memory/` path component, so it does not die when the anchor is dropped.

Practical blast radius today is nil — no committed path in the parent repo or any of the seven child repos contains `memory/session_handoff.md` outside `.claude/`. This is a suite-durability gap, not a live defect, and the shipped pattern is correct. Filed as #1045 with the one-line test that closes it; not a merge blocker.

## 4. Are the new assertions correct, or merely descriptive?

`test_skip_does_not_extend_to_sibling_memory_notes_by_prefix` asserts `.claude/memory/session_handoff_notes.md` is NOT skipped. Confirmed this is behaviour-we-want, not behaviour-we-have: `.gitignore` line 117 is the literal exact path `.claude/memory/session_handoff.md` with no glob, and `git check-ignore` on the sibling returns nothing. A `session_handoff_notes.md` would therefore be ordinary committed curated memory under the repo's `<snake_case_slug>.md` convention, and must stay tracked. Asserting `False` is right.

`test_skip_is_scoped_to_the_claude_memory_directory` (`docs/session_handoff.md` not skipped) is likewise correct — only the gitignored machine-local file is exempt.

The `os.chdir` is global process state, but it is `try`/`finally`-protected so it restores even on assertion failure, and the full suite passing at 31 confirms no cwd leak into sibling tests.

Nothing in the delta asserts anything wrong. Approved at `ac8bcfa96f157f4801543dc29b08f2b4139f1307`.

Issue #1048, opened 2026-07-20
Share one verdict-set entry point between Hook 4 and pr_review_state (follow-up to #1046 point 5)


parametrization:
## Summary

Follow-up to #1046 (PR pending). That issue's point 5 proposed a structural guard: a **single shared entry point returning the full verdict set** (current + stale, comment + formal), so a caller cannot re-derive the gate's verdict with its own argument list and silently drift.

#1046 delivered the **narrow half** of this and deliberately stopped short of the rest.

## What #1046 already did

Extracted `partition_formal_reviewers(reviews, author, content_ts)` in `.claude/hooks/validate_pr_review.py`, now shared by Hook 4's `check()` and `.claude/lib/pr_review_state.py`. That extraction was in-scope because the alternative was hand-rolling formal-review staleness a second time in the driver — creating a fresh copy of exactly the logic that had drifted.

## What remains

The comment/roster/threshold pipeline is still re-assembled independently by each caller. Both `check()` and `compute_review_state` separately:

1. call `get_latest_content_commit` and derive `content_ts` / `content_sha`
2. call `check_comment_reviews(..., content_ts=...)` on two head-ref branches (feature vs `deployments/**/wave-*`)
3. call `_load_roster_names` and subtract non-roster requestors
4. union formal + comment reviewers and apply the threshold / single-reviewer exception

Any future parameter added to that pipeline has to be threaded through **two** call sites by hand. That is the #1046 defect class, still live — the driver just happens to be correct today.

**Proposed shape:** one function (e.g. `resolve_review_verdicts(pr_data, repo) -> ReviewVerdicts`) holding steps 1-4 and returning current reviewers, stale verdicts, non-roster requestors, T_content, and TechDebt state. `check()` keeps its block-message rendering; the driver keeps its text/JSON rendering. Neither re-derives the verdict set.

## Why it was NOT done in #1046

`check()` is a long function that interleaves this pipeline with its `decision: block` message construction, and `validate_pr_review.py` is the highest-fan-in module in the repo graph (centrality rank #11, `ontology/structural/llms.txt`). Restructuring it is a real refactor of the canonical merge gate, not a drive-by — and doing it inside a bug fix whose whole point was a mutation-verified regression guard would have mixed a mechanical fix with a risky one. Deliberately deferred rather than scope-crept.

## Acceptance

- [ ] Single shared entry point; neither `check()` nor `compute_review_state` re-derives the verdict set.
[...]

Issue #11, opened 2026-04-07
hook: Validate VPS_HOST is not a Cloudflare IP


parametrization:
## Summary

Create a PreToolUse hook that blocks `gh variable set VPS_HOST` if the value resolves to a known Cloudflare IP range.

## Context
Session 4: VPS_HOST was set to a domain that resolved to Cloudflare proxy IPs, causing SSH timeout. Deploy failed because Cloudflare doesn't proxy port 22.

## Implementation
- Hook script: `.claude/hooks/validate_vps_host.py`
- Trigger: PreToolUse on Bash commands containing `gh variable set VPS_HOST`
- Check: extract the value, attempt DNS resolution, check if the IP falls in known Cloudflare ranges (104.16.0.0/12, 172.64.0.0/13, 103.21.244.0/22, etc.)
- Also warn if the value is a hostname rather than an IP (SSH should use direct IP, not proxied domain)
- Block with message: "VPS_HOST value {value} resolves to a Cloudflare IP. Use the direct VPS IP address instead."

## Acceptance Criteria
- `gh variable set VPS_HOST` blocked if value is a Cloudflare IP or resolves to one
- Warning if value is a hostname instead of IP
- Hook added to settings.json
- Documented in charter/hooks.md

PR #1049, opened 2026-07-20
fix(#1046): bind pr_review_state to T_content so stale verdicts stop reading as PASS

### LINKED ISSUE #1046: pr_review_state.py drops content_ts: the review-state driver never marks stale verdicts, reports PASS where Hook 4 would not
## Summary

`.claude/lib/pr_review_state.py` — the deterministic "run the instrument instead of trusting the brief" driver (#707) — **silently disables the verdict-staleness binding** that #951 added to Hook 4 for #950/#935. It calls the gate's own function but drops the `content_ts` argument, so `T_content` is never computed and **no verdict is ever marked stale**.

The result: the diagnostic reports `passes: true` on approvals that the real merge gate would reject as stale. It fails in the dangerous direction — it under-reports relative to the gate it claims to replay.

## Root cause — one missing keyword argument

`check_comment_reviews` takes an optional `content_ts` (`.claude/hooks/validate_pr_review.py:794-799`):

```python
def check_comment_reviews(
    pr_number, branch_author_lastname, repo=None, content_ts: datetime | None = None,
) -> CommentReviewResult:
```

**Hook 4 passes it** (`validate_pr_review.py:1237-1246`):
```python
comment_review_result = check_comment_reviews(
    number, branch_author_lastname, repo=repo, content_ts=content_ts
)
```

**The driver does not** (`.claude/lib/pr_review_state.py:135` and `:137`):
```python
comment_result = gate.check_comment_reviews(number, branch_author_lastname, repo=repo)
```

`content_ts` defaults to `None` → staleness filtering is a no-op → `CommentReviewResult.stale_verdicts` stays empty. The driver additionally never calls `gate.get_latest_content_commit` and never reads `.stale_verdicts`; `rg 'stale|content_ts|get_latest_content_commit' .claude/lib/pr_review_state.py` returns **zero hits**, and no staleness field appears in either the text or `--json` render.

Its docstring (`:94-98`) asserts the opposite of the behaviour:

> *"Compute a PR's review state by replaying the merge gate's own logic. Reuses `gate.get_pr_data`, … `gate.check_comment_reviews`, … so this driver and Hook 4 cannot drift."*

It reuses five functions and drifts on the sixth thing that matters.

## Observed live — main#1040, 2026-07-20

PR #1040 carried two `Approved` verdicts cast at head `7428f25`. An additive commit `ac8bcfa` (test-only) was then pushed. At that point:

```
$ python3 .claude/lib/pr_review_state.py --repo noorinalabs/noorinalabs-main 1040
PR #1040 — merge gate would PASS
```

Both counted approvals predated the head they were being counted against — precisely the #950 defect, reported as PASS. The `--json` render likewise carries `"passes": true` with no staleness key at all.

The merge was held anyway and both reviewers re-confirmed at `ac8bcfa`, but **only because the charter's head-SHA rule was applied by hand**. Nothing in the tool surfaced it.

## Why this is worse than an ordinary reporting gap

Team memory instructs agents to **trust this tool over the orchestrator's brief** — that is the P9W25 §8 lesson, where Alejandra Reyes-Fuentes caught an orchestrator verdict-format error precisely by running `pr_review_state.py` rather than believing the spawn brief. An agent following that instruction today gets a PASS on stale verdicts.

Same defect class as #950/#935 and #981: **a gate-adjacent artifact that fails open, silently, and whose silence reads as approval.** #951 fixed the gate; the driver was left behind.

## Proposed fix

1. Pass `content_ts` through: compute `T_content` via `gate.get_latest_content_commit` in `compute_review_state` and forward it to both `check_comment_reviews` call sites (`:135`, `:137`), mirroring Hook 4 `:1237-1246`.
2. Surface `stale_verdicts` in `ReviewState`, in `_render_text`, and in `_render_json` — a stale verdict must be visible, not merely subtracted.
3. Exclude stale verdicts from `distinct_reviewer_count` / `passes`, matching the gate.
4. Regression test: a PR whose approvals predate `T_content` must report `passes: false`. **Verify by mutation** — the test must fail if `content_ts` is dropped again, since that is exactly the regression that shipped here.
5. Consider a structural guard against recurrence: the drift is possible because the driver re-calls gate functions with its own argument list. A single shared entry point that returns the full verdict verdict-set (stale included) would make the omission unrepresentable.

## Impact

No known bad merge resulted — #950's own instances were caught by hand, and #1040 was too. But the tool that exists so this is *not* left to the human eye currently cannot see it.

parametrization:
Closes #1046

## What was wrong

`.claude/lib/pr_review_state.py` called `gate.check_comment_reviews(...)` without the `content_ts` keyword. `T_content` was never computed, `CommentReviewResult.stale_verdicts` stayed empty, and the driver reported `passes: true` on approvals the real merge gate rejects as stale — under-reporting against the gate it exists to replay. Team memory instructs agents to trust this tool over the spawn brief, so the silence read as approval.

## Changes

**`.claude/lib/pr_review_state.py`**
- Compute `T_content` via `gate.get_latest_content_commit` and forward `content_ts` to **both** `check_comment_reviews` call sites — the feature-branch one and the `deployments/**/wave-*` one.
- Exclude stale verdicts from `distinct_reviewer_count` / `passes()` (inherited from the gate's own filtering, not re-implemented).
- Surface `content_sha`, `content_ts`, and `stale_verdicts` on `ReviewState`, in `_render_text` (each stale verdict is named, with when it was cast and what invalidated it) and in `_render_json` (plus a `stale_verdict_count` convenience key).
- Correct the module docstring. It asserted the driver and Hook 4 "cannot drift" while the drift was live; per `ontology/conventions.md` :257 code wins, so the doc is fixed to match, with an added note that reusing a gate function is not parity unless the **arguments** match.

**`.claude/hooks/validate_pr_review.py`**
- Extracted `partition_formal_reviewers(reviews, author, content_ts)` and replaced Hook 4's inline loop with a call to it. Behaviour-preserving for the hook (full gate suite unchanged and green).

## Two things the issue did not have

**1. Hook 4 also binds FORMAL GitHub reviews to `T_content`** (`validate_pr_review.py` :1208-1230), and the driver's formal-reviewer loop was equally unbound. The issue's 5-point fix only covered comment verdicts, so implementing it literally would have left half the drift in place. Rather than hand-roll the rule a second time in the driver — a fresh copy of exactly the logic that drifted — I extracted the shared helper. This is the narrow, low-risk half of the issue's point 5, done here because the alternative was actively worse.

**2. A commit-fetch failure must be a determinate error, not a degrade to `content_ts=None`.** Hook 4 hard-blocks on `CommitFetchError`. The obvious implementation — wrap in `try`, fall back to `None` — reinstates the identical fail-open under a different name. `compute_review_state` now raises `ReviewStateError` (exit 2), and there is a test pinning it.

## Mutation verification

Three mutations, each reverted after measurement. Literal output:

```
########## BASELINE (fix in place) ##########
21 passed in 0.34s

########## MUTATION 1: drop content_ts from BOTH check_comment_reviews call sites ##########
FAILED .claude/lib/tests/test_pr_review_state.py::ContentStalenessTests::test_content_ts_forwarded_on_feature_branch_call_site
FAILED .claude/lib/tests/test_pr_review_state.py::ContentStalenessTests::test_content_ts_forwarded_on_wave_merge_call_site
FAILED .claude/lib/tests/test_pr_review_state.py::ContentStalenessTests::test_main1040_stale_approvals_do_not_pass
FAILED .claude/lib/tests/test_pr_review_state.py::ContentStalenessTests::test_mixed_freshness_counts_only_the_current_verdict
FAILED .claude/lib/tests/test_pr_review_state.py::ContentStalenessTests::test_stale_verdicts_are_visible_in_both_renders
5 failed, 16 passed in 0.44s

########## MUTATION 2: swallow CommitFetchError -> content_ts=None fail-open ##########
FAILED .claude/lib/tests/test_pr_review_state.py::ContentStalenessTests::test_commit_fetch_failure_is_an_error_not_a_fail_open
1 failed, 20 passed in 0.39s

########## MUTATION 3: unbind formal reviews from T_content ##########
FAILED .claude/lib/tests/test_pr_review_state.py::ContentStalenessTests::test_stale_formal_review_is_excluded_and_recorded
1 failed, 20 passed in 0.39s

########## RESTORED ##########
21 passed in 0.36s
```

### Two inert tests the mutation pass caught in my own work

Worth recording, because both would have shipped as green-but-worthless:

- **Mutation 2 initially SURVIVED (`21 passed`).** `test_commit_fetch_failure_is_an_error_not_a_fail_open` used a bare `assertRaises(ReviewStateError)`. With the swallow applied, execution fell through to the roster resolver, which raised `ReviewStateError` for a completely unrelated reason — the test passed against the exact defect it was written to catch. Fixed by mocking every downstream collaborator to a success value and asserting the message (`"commit list"`, `"boom"`).
- **An anti-vacuity guard caught a bad fixture.** `test_fixture_yields_a_pass_when_verdicts_are_fresh` (modelled on the gate suite's own fixture check) failed because I had named a reviewer "Lucas Ferreira" on an `L.Ferreira/...` head ref — the gate dropped him as a self-review, so the staleness assertions would have passed for the wrong reason. Head ref changed to `W.Mwangi/...`.

The end-to-end tests deliberately drive the **real** `check_comment_reviews` over a faked `gh api` payload rather than a canned return value: a mock that ignores its arguments cannot detect an argument going missing, which is precisely how the original defect passed a green suite.

## Live verification against #1040

Ran the fixed driver against the real PR #1040 (the issue's reproduction):

```
PR #1040 (noorinalabs/noorinalabs-main) — merge gate would PASS
  content commit (T_content): ac8bcfa9 @ 2026-07-20T03:28:37+00:00
  STALE verdicts EXCLUDED from the count: 2
    Lucas Ferreira — Approved [comment] cast 2026-07-20T03:23:36Z x STALE (before ac8bcfa9)
    Nino Kavtaradze — Approved [comment] cast 2026-07-20T03:24:37Z x STALE (before ac8bcfa9)
```

It now names both originally-stale verdicts while correctly counting the two **fresh re-confirmations** the reviewers posted at `ac8bcfa`. PASS is the right answer for #1040's current state — the point is that pre-fix it was the right answer *by accident*, with the staleness invisible. Note `TechDebt issue numbers` drops from 5 to 1, which is correct: the gate exempts stale verdicts from the TechDebt requirement, and the driver now inherits that.

## Scope call

Issue point 5 (a single shared entry point for the whole verdict set) is **not** in this PR beyond the formal-review extraction. `check()` interleaves the pipeline with its block-message rendering, and `validate_pr_review.py` is the highest-fan-in module in the repo graph (centrality rank #11) — restructuring it inside a bug fix whose value is a mutation-verified guard would mix a mechanical change with a risky one. Filed as **#1048**.

## Gates

`ruff-format`, `ruff-lint`, `actionlint`, `cspell`, `fixture-realism`, `skill-graphql-pagination`, `mermaid-render` (pre-push) all pass; `mypy` clean over 174 files; `pre_commit_ci_sync.py` reports parity. Full suite: **2648 passed, 88 subtests passed**.

## Reviewers

Not merging — assigning left to the Standards Lead's requestor per the brief.

TechDebt: #1048

### DIFF
--- a/.claude/hooks/validate_pr_review.py
+++ b/.claude/hooks/validate_pr_review.py
@@ -911,6 +911,57 @@ def check_comment_reviews(
         return result
 
 
+def partition_formal_reviewers(
+    reviews: list[dict],
+    author: str,
+    content_ts: datetime | None = None,
+) -> tuple[set[str], list[StaleVerdict]]:
+    """Split formal GitHub reviews into CURRENT reviewers and STALE verdicts (#950).
+
+    Formal GitHub reviews are bound to `T_content` by the SAME rule as comment
+    verdicts: a review submitted before the branch's latest authored commit
+    reviewed code that has since been rewritten. Leaving formal reviews unbound
+    would be a hole straight through the #950 fix.
+
+    Returns `(formal_reviewers, stale_formal)` where `formal_reviewers` holds
+    lowercased logins of non-author reviewers whose review is CURRENT, and
+    `stale_formal` holds a `StaleVerdict` per excluded review (for the
+    diagnostic). Self-reviews (login == `author`) and reviews with no login are
+    dropped entirely — they are not evidence and not staleness.
+
+    `submittedAt` missing or unparseable while `content_ts` is known ⇒ STALE,
+    not fresh (fail closed). `content_ts=None` means "no content binding" and
+    every non-author review counts, as before the content binding existed.
+
+    This function is the SHARED implementation for `check()` (Hook 4) and
+    `.claude/lib/pr_review_state.py` (the #707 ahead-of-time driver). Both must
+    partition formal reviews identically; a second hand-rolled copy in the
+    driver is exactly the drift that let #1046 ship, where the driver reported
+    PASS on verdicts the gate would have rejected as stale. Do not inline it
+    back into either caller.
+    """
+    formal_reviewers: set[str] = set()
+    stale_formal: list[StaleVerdict] = []
+    for review in reviews:
+        login = review.get("author", {}).get("login", "")
+        if not login or login == author:
+            continue
+        if content_ts is not None:
+            submitted_raw = review.get("submittedAt", "") or ""
+            submitted_at = _parse_iso8601(submitted_raw)
+            if submitted_at is None or submitted_at < content_ts:
+                stale_formal.append(
+                    StaleVerdict(
+                        reviewer=login,
+                        verdict=str(review.get("state", "REVIEW")),
+                        created_at=submitted_raw,
+                    )
+                )
+                continue
+        formal_reviewers.add(login.lower())
+    return formal_reviewers, stale_formal
+
+
 def _iter_roster_entries(
     role_prefix_filter: tuple[str, ...] | None = None,
     roster_dir: Path | None = None,
@@ -1207,27 +1258,9 @@ def check(input_data: dict) -> dict | None:
 
     # Formal GitHub reviews are bound to T_content by the SAME rule (#950) — a
     # stale formal review is no better evidence than a stale comment verdict, and
-    # leaving it unbound would be a hole straight through the fix. `submittedAt`
-    # missing or unparseable while T_content is known ⇒ stale, not fresh.
-    formal_reviewers: set[str] = set()
-    stale_formal: list[StaleVerdict] = []
-    for review in reviews:
-        login = review.get("author", {}).get("login", "")
-        if not login or login == author:
-            continue
-        if content_ts is not None:
-            submitted_raw = review.get("submittedAt", "") or ""
-            submitted_at = _parse_iso8601(submitted_raw)
-            if submitted_at is None or submitted_at < content_ts:
-                stale_formal.append(
-                    StaleVerdict(
-                        reviewer=login,
-                        verdict=str(review.get("state", "REVIEW")),
-                        created_at=submitted_raw,
-                    )
-                )
-                continue
-        formal_reviewers.add(login.lower())
+    # leaving it unbound would be a hole straight through the fix. Shared with the
+    # #707 driver via `partition_formal_reviewers` so the two cannot drift (#1046).
+    formal_reviewers, stale_formal = partition_formal_reviewers(reviews, author, content_ts)
 
     comment_review_result = CommentReviewResult()
     branch_author_lastname = None
--- a/.claude/lib/pr_review_state.py
+++ b/.claude/lib/pr_review_state.py
@@ -10,25 +10,39 @@ PreToolUse block. There was no way to ASK the same question ahead of time:
   - before `gh pr merge`   — will the gate pass, or who is missing TechDebt?
 
 This CLI answers that question by REUSING the hook's functions verbatim —
-`get_pr_data`, `extract_branch_author_lastname`, `check_comment_reviews`,
-`_load_roster_names`, and `is_single_reviewer_exception`. It deliberately does
-NOT reimplement the charter-comment parsing: a fork would silently drift from
-the gate, and that drift is the exact failure #707 exists to prevent. The
-PASS/FAIL verdict computed here mirrors `validate_pr_review.check()` step for
-step (formal + roster-filtered comment reviewers, the wave-merge head-ref
-sentinel, the single-reviewer exception, and the missing-TechDebt block).
+`get_pr_data`, `extract_branch_author_lastname`, `get_latest_content_commit`,
+`partition_formal_reviewers`, `check_comment_reviews`, `_load_roster_names`,
+and `is_single_reviewer_exception`. It deliberately does NOT reimplement the
+charter-comment parsing: a fork would silently drift from the gate, and that
+drift is the exact failure #707 exists to prevent. The PASS/FAIL verdict
+computed here mirrors `validate_pr_review.check()` step for step (T_content
+verdict staleness, formal + roster-filtered comment reviewers, the wave-merge
+head-ref sentinel, the single-reviewer exception, and the missing-TechDebt
+block).
+
+Reusing the gate's functions is necessary but NOT sufficient for parity — the
+ARGUMENTS matter as much as the callee. #1046: this driver called
+`check_comment_reviews` without the `content_ts` keyword, so `T_content` was
+never computed, `stale_verdicts` stayed empty, and the tool reported PASS on
+approvals the gate would reject as stale (observed live on main#1040). When
+adding a parameter to a shared gate function, audit THIS file's call sites —
+`.claude/lib/tests/test_pr_review_state.py::ContentStalenessTests` is the
+regression guard and is mutation-verified against exactly that omission.
 
 Usage:
     python3 .claude/lib/pr_review_state.py <pr_number> --repo <owner/repo>
     python3 .claude/lib/pr_review_state.py <pr_number> --repo <owner/repo> --json
 
 Exit codes:
-    0 — the merge gate would PASS (>=2 distinct Approved reviewers, or exactly
-        one with the wave-bootstrap exception, and no verdict missing TechDebt)
-    1 — the merge gate would BLOCK (too few reviewers, or a verdict missing the
-        TechDebt line)
-    2 — review state could not be determined (PR fetch failed, or a named child
-        repo's roster could not be resolved — mirrors the hook's hard-block)
+    0 — the merge gate would PASS (>=2 distinct CURRENT Approved reviewers, or
+        exactly one with the wave-bootstrap exception, and no verdict missing
+        TechDebt)
+    1 — the merge gate would BLOCK (too few current reviewers, or a verdict
+        missing the TechDebt line)
+    2 — review state could not be determined (PR fetch failed, the PR's commit
+        list could not be fetched so no verdict's freshness is knowable, or a
+        named child repo's roster could not be resolved — each mirrors a hook
+        hard-block)
 """
 
 from __future__ import annotations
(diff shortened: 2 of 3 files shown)


parametrization:
Requestor: Lucas Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

SRE/operational review. Ran both suites clean (21 + 184 passed) and independently mutation-tested two spots Lucas hadn't covered:

1. Made `partition_formal_reviewers` ignore its `content_ts` argument (force it to None internally) — caught immediately: 3 failures, including Hook 4's own `FormalReviewStalenessTests` and the driver's `test_stale_formal_review_is_excluded_and_recorded`. The shared helper is exercised from both sides.
2. Made `_render_json` silently zero out `payload["stale_verdicts"]` while leaving `stale_verdict_count` correct — caught by `test_stale_verdicts_are_visible_in_both_renders`, which is exactly the silent-subtraction failure mode #1046 point 2 calls out. Good that the test asserts on the list contents, not just the count.

On the `CommitFetchError` -> `ReviewStateError` (exit 2) question: this is not a novel operational risk introduced by this PR — Hook 4's own `check()` (validate_pr_review.py:1178-1199) already hard-blocks on the identical `CommitFetchError` with the same rationale (da#423 regression: counting verdicts of unknown freshness is the fail-open #950 closes). The driver's job is to replay the gate's verdict, so falling back to `content_ts=None` here would make the driver diverge from the gate it's supposed to mirror — precisely the #1046 bug class. Exit 2 is correct, and the error message is informative (points at `gh auth status`/network as likely causes). Confirmed `main()` wires it to `return 2` and the exit codes in the module docstring (0/1/2) match actual behavior.

Verified the `partition_formal_reviewers` extraction independently by diffing the old inline loop vs the new function body line-by-line — identical logic, same variable names, pure extraction. No behavior change to Hook 4.

CI: all 14 check-runs green at head 8729bca (Ruff lint/format, mypy, pytest, memory-budget, headcount-budget, sync-drift, doc-freshness, roster-union, commit-author verification, etc.) — confirmed via `gh api .../commits/8729bca.../check-runs`, not the PR number.

No blocking issues. Good regression-guard design — the test class docstring's point about mocks not being able to detect a missing keyword argument (hence driving through the real `check_comment_reviews`) is the right instinct given how #1046 shipped in the first place.

parametrization:
Requestor: Nino Kavtaradze
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1050, #1051

Merge-gate review (reviewer 2 of 2). Verified independently against head SHA 8729bca9, not the PR body. 205 tests pass in both suites; all 14 CI checks green at the head SHA.

The fail-open is genuinely closed, not relocated. The only surviving path to `content_ts=None` is `get_latest_content_commit` returning `None` for a PR with zero non-merge commits, which means "no authored content exists to be stale against" — that is the documented #950 semantic and it is identical in `check()`, so it is parity, not a new hole. I read the other branches specifically: an empty `head_ref`, and a `head_ref` that neither parses to a lastname nor matches the `deployments/*/wave-*` sentinel, both leave `comment_result` as the empty default, so zero comment verdicts are counted while formal reviews remain bound. No branch counts a verdict of unknown freshness.

Hook 4 behaviour preservation — this was the highest-blast-radius part and I did not take the assertion on trust. I reconstructed the pre-PR inline loop verbatim from `origin/main` and differential-tested it against `partition_formal_reviewers` over 50,000 randomized inputs, comparing field-wise (the `StaleVerdict` identity comparison in my first harness produced 4,923 false divergences — worth knowing if anyone repeats this). Input space covered empty logins, absent `author` dicts, missing/empty/unparseable/boundary-equal `submittedAt`, login case variants, self-review, and both `content_ts=None` and set. Zero divergences. The extraction is behaviour-preserving.

Mutations I chose, deliberately different from the PR body's:

- Dropped `content_ts` from **only** the wave-merge sentinel call site, not both — caught by `test_content_ts_forwarded_on_wave_merge_call_site`. A single-site omission is the realistic drift and it is guarded.
- Replaced the `raise ReviewStateError` with `latest_content = None` — the fail-open under a new name — caught, and the test asserts the message substrings, not a bare `assertRaises`. That is the correct repair of the vacuity you self-reported.
- Made `partition_formal_reviewers` ignore `content_ts` entirely — caught in **both** suites, confirming the formal path is bound on the driver side too, not just the gate.
- Flipped the missing-`submittedAt` fail-closed to fail-open — caught by the hook suite only, not the driver suite. Acceptable given one shared implementation guarded once, but noting it.
- Kept the arithmetic and removed only the naming (`stale_verdicts=[]`, then separately stripped the per-verdict render lines) — caught by 4 tests and by `test_stale_verdicts_are_visible_in_both_renders`. Silent subtraction is a test failure, which is what makes the trust posture real.
- Positive control: forced `distinct = set()` so the fixture cannot produce approvals even when fresh — `test_fixture_yields_a_pass_when_verdicts_are_fresh` fails. The anti-vacuity guard is load-bearing, not decorative.

On the vacuity sweep you asked about: I did not find another test that would pass for the wrong reason. The explicit positive control plus message-substring assertions on the error path are the right shape.

No over-correction. `distinct = formal_reviewers | roster_comment_reviewers` matches `check()` exactly, and a commit-fetch failure exits 2 (undeterminable) rather than 1 (FAIL), so the driver cannot report a false BLOCK where the gate would PASS.

Also exercised the real engine rather than the string: `get_latest_content_commit` against live PRs returns `8729bca9 @ 11:12:39Z` for this PR and `ac8bcfa9 @ 03:28:37Z` for #1040 — the latter matching the `ac8bcfa` fixture, confirming the tests are built from the real observation.

Non-blocking, filed rather than held against the merge: #1050 — `content_ts` remains an *optional* parameter on `check_comment_reviews` and now also on the new `partition_formal_reviewers`, so the #1046 shape survives in the signature for a future third caller. I audited repo-wide and confirmed exactly two production call sites, both forwarding correctly, so this is a hazard rather than a defect. #1051 is an unrelated tooling gap I hit while doing that audit.

Issue #1047, opened 2026-07-20
premise_check: any backtick token containing '/' is treated as a path — 12/12 false STOP on wave-26 scope


parametrization:
## Summary

`.claude/lib/premise_check.py` — the hard **STOP** gate at `/wave-scope` Step 6.5 — classifies *any backtick-quoted prose fragment containing a slash* as a file path. On the wave-26 scope run it flagged **12 of 12** in-scope issues as premise-rot. **All 12 were false positives.** Zero genuine rot existed.

A hard gate with a 100% false-positive rate is a gate people learn to `--warn-only` past. This is the same erosion as #981 and #950, approached from the opposite direction: those fail open and silently; this one fails loud and wrong. Both end with the gate carrying no information.

## Root cause — `looks_like_path` accepts any token containing `/`

`.claude/lib/premise_check.py:154-155`:

```python
    if "/" in tok:
        return True
    ext = tok.rsplit(".", 1)[-1].lower() if "." in tok else ""
    return ext in _CODE_EXTENSIONS
```

The slash branch short-circuits before any extension or plausibility test. Its own docstring (`:139-142`) states the opposite intent:

> *"Precision over recall: a false negative just means one fewer thing checked; a false positive turns prose into a spurious STOP."*

`/wave-scope` SKILL.md § 6.5 repeats the claim: *"a strict path regex, so prose never produces a false STOP."* Neither is true. Per `ontology/conventions.md:257` (code wins on conflict), both docs are wrong and must be corrected alongside the code.

## Observed — wave-26 scope run, 2026-07-20

Every one of these was extracted as a "path" and reported MISSING:

| Token | Actually is |
|---|---|
| `/` | a bare slash — matches `[\w./\-]+`, contains `/`, accepted |
| `A/B` | prose ("bidirectional A/B fixtures") |
| `Latin/English` | prose |
| `benediction/Prophet-title` | prose |
| `ism/kunya` | Arabic name-part prose |
| `token-count/matn-density` | prose |
| `986/650` | a row count ("650,986/650,986 rows") |
| `origin/main` | a git **ref**, not a path |
| `recall/precision`, `display/graph`, `Identity/matching`, `taṣliya/eulogy`, `pollution/mc`, `resolve/load`, `collectors/commentators`, `Validate/scrub`, `boundary/particle`, `name/nisba`, `transliteration/cross-script`, `Code/git` | prose |

Two further distinct false-positive classes appeared:

[...]

Issue #983, opened 2026-07-19
Phase 9 Wave 26 — Parse recovery & name quality (confirm theme)


parametrization:
## Theme

**Parse recovery & name quality** (data-acquisition) — owner-confirmed 2026-07-20. Ratified in the Phase 9 plan (#977).

## Scope

Reconciled by `/wave-scope` on 2026-07-20. **Drift: 0** — all 12 `wave-26` labeled issues matched the declared candidate scope exactly; no keep/defer/strip/close dispositions were required.

**12 issues, 1 child repo** (`noorinalabs-data-acquisition`). Implementer/reviewer names validated against the data-acquisition roster (`/wave-scope` § 12.5, all resolve).

### Tier 1 — Parse recovery (5)

| Issue | Implementer | Reviewers |
|---|---|---|
| da#373 — `canonical_matn_identity` dead gate + edge loader (**one PR**, or re-creates the 196k orphan-chain bug) | Jean-Claude Habimana | Kavitha Sundaramurthy, Alejandra Reyes-Fuentes |
| da#397 — `bio_promote` truncated-merge guard refuses 773 gloss-tail rows | Nikolaos Papadopoulos | Jean-Claude Habimana, Alejandra Reyes-Fuentes |
| da#398 — same guard over-refuses 792 full-nasab isnad-tail rows | Nikolaos Papadopoulos | Jean-Claude Habimana, Kavitha Sundaramurthy |
| da#424 — matn sentences surviving the drop-gate (52 rows + #314 residual) | Kavitha Sundaramurthy | Nikolaos Papadopoulos, Oyunbileg Batbayar |
| da#427 — `normalize_arabic`: alif maqsura never folds to yaa | Kavitha Sundaramurthy | Ivana Horvat, Alejandra Reyes-Fuentes |

> da#397 and da#398 are the **same guard** from opposite directions (over-refusal of gloss-tail vs full-nasab). Assigned to one implementer deliberately — split ownership risks two divergent loosenings of one predicate.

### Tier 2 — Name quality (6)

| Issue | Implementer | Reviewers |
|---|---|---|
| da#295 — `fuzzy_cluster` cap follow-up: name-first/IDF token retention | Ivana Horvat | Kavitha Sundaramurthy, Nikolaos Papadopoulos |
[...]

Issue #888, opened 2026-06-25
board-audit: GraphQL first:500 over-cap + per-item gh-view drift scan + loop hang (pagination footgun)


parametrization:
## Summary

The `/board-audit` skill has three mechanical defects in its GitHub-API access patterns. Two surfaced (and were worked around) during `/board-audit` for P7W19 on 2026-06-25; the third is a latent scalability bomb in the same skill.

### 1. GraphQL `items(first: 500)` exceeds the 100-record connection cap (hard error)
`SKILL.md` Step 2 queries `projectV2 { items(first: 500) { … } }`. GitHub's GraphQL API caps connection `first:` at **100** — the query errors with *"Requesting 500 records on the connection exceeds the `first` limit of 100 records"* and returns **0 items** (silent-ish: the skill computed `board item count: 0` and would have reported every issue as an orphan). Project 2 currently has **1394 items**, so this is not a rare edge.

**Fix:** `first: 100` + cursor pagination — `pageInfo { hasNextPage endCursor }`, a `$endCursor` query var, and `gh api graphql --paginate`, then `jq -s` to merge pages. (This is the pattern that worked in the P7W19 run.)

> Pagination has bitten us more than once — fold a one-line convention into the skill-authoring notes: **GraphQL connections cap at 100; always paginate.** A cheap lint over `.claude/skills/**/*.md` for `first:[ ]*[1-9][0-9]{2,}` inside a `gh api graphql` block would catch regressions (only `board-audit/SKILL.md:79` matches today).

### 2. Step 4 drift detection does `gh issue view` **per board item** — O(board size) network calls
With 1394 board items this is ~1394 serial `gh issue view` calls (minutes, rate-limit risk). The labels needed for drift detection are already available in bulk: fetch them in Step 1's `gh issue list --json url,labels` (one call per repo, 8 total), build a `url → wave-label` map, and cross-reference **in memory** against the board's `url → Wave-field` map (already in the Step-2 GraphQL response). This eliminates the per-item `gh view` loop entirely. (Used this optimization in the P7W19 run — full audit dropped to ~10 calls.)

### 3. Org-wide `for repo in …; gh issue list` loop hung
[...]

Issue #1051, opened 2026-07-20
rg silently returns zero over `.claude/` — TOOLCHAIN.md documents the --no-ignore gotcha but not the --hidden one


parametrization:
Found while running a call-site audit during the merge-gate review of #1049.

## The trap

`rg` skips **hidden** files and directories by default, in addition to `.gitignore`d ones. This org keeps its *entire* configuration and enforcement tree — hooks, lib, skills, roster, agents, memory — under `.claude/`, a **hidden** directory. So a recursive `rg` from the repo root silently returns **zero matches** for anything under `.claude/`, even with `--no-ignore`.

Reproduce:

```
$ rg -n --no-ignore "check_comment_reviews\(" . -g '*.py'
                                    # <- zero output, silently

$ rg -n --no-ignore --hidden "check_comment_reviews\(" . -g '*.py'
./.claude/lib/pr_review_state.py:189  ...
./.claude/hooks/validate_pr_review.py:1270  ...
                                    # <- 11 real matches
```

This bit me live: my first "are there other callers that omit `content_ts`?" audit returned "none — every call site forwards it", which looked like a clean result and was in fact a measurement of nothing. The correct answer (two production callers) only appeared after adding `--hidden`.

## Why this matters more than the usual footgun

This is the exact failure class `feedback_silent_zero_is_not_a_measurement` names, and `docs/TOOLCHAIN.md` § Text search already documents the **`.gitignore`** half of it (lines 131-135, prescribing `--no-ignore`). It does not document the **hidden** half. An agent following TOOLCHAIN.md exactly will still get a false zero over `.claude/` — and `.claude/` is where every gate, hook, and enforcement artifact lives, so false-negative audits land precisely on the safety-critical tree.

## Suggested fix

1. Add a third bullet to `docs/TOOLCHAIN.md` § Text search covering hidden paths: use **`--hidden`** (or `-uu`, which implies both `--hidden` and `--no-ignore`) whenever the search may cover `.claude/` or any dotted path. Note that `-uu` is the single flag that covers both gotchas.
2. Mirror the note into `ontology/conventions.md` alongside the existing `rg`/`--no-ignore` convention.
[...]

PR #1052, opened 2026-07-20
scope(#983): wave-26 scope reconciliation + gh-CLI quota/check-runs gotchas

parametrization:
## Summary

Two commits from the wave-26 `/wave-scope` run, distinct authors:

**`scope(#983)`** (Nadia Khoury) — the `wave_26_*` bookkeeping keys `/wave-kickoff` reads: `_scope_reconciled_at`, `_repos_in_scope`, `_meta_issue`, `_scope`, plus a reconciliation note. Written via `upsert_status_keys.py`, so the compact-inline file shape is preserved.

**`memory(#983)`** (Aino Virtanen) — two gh-CLI surfaces verified live, appended to the consolidated gotchas note rather than filed as new memories.

## Scope outcome

Drift **0** — all 12 `wave-26` labeled issues matched the meta-issue candidate scope exactly; no dispositions required. Implementer/reviewer slate validated against the data-acquisition roster (all 7 names resolve). Top concentration 25%, down from wave-25's 43%.

TD intake recorded as **0 of a computed target of 1**, deliberately. The +20% formula assumes tech-debt is a top-up on feature work; here 9 of the wave's own 12 issues are already `tech-debt` labeled, so the non-TD base collapses to 3 and the target loses meaning. The rationale is stored in `wave_26_scope.td_intake` so the retro can revisit whether the base should count *theme* content regardless of label.

## Two gates behaved badly during this run

1. **`premise_check.py` STOPped 12/12, all false positives** — it accepts any backtick token containing `/` as a file path, so `A/B`, `ism/kunya`, `986/650`, `origin/main` and a bare `/` were checked as paths. Every token that genuinely *was* a path resolved OK. Cleared by hand; 0 genuine rot. Root cause filed as #1047.
2. **GraphQL quota exhausted silently** — the error body is a bare string that `jq` cannot parse, so with `2>/dev/null` it reads as an empty result set. This nearly produced a false "all 12 wave-26 issues missing from project board 2" finding when board state was simply unknown. Recorded as gotcha section 12, with the note that #888 (`first: > 100` over-cap) produced the *identical* false-orphan symptom from a different cause.

This PR was itself opened via the REST fallback documented in that gotcha, because `gh pr create` is GraphQL-backed and the quota was still flat.

## Test plan

No executable code changed — status-file keys and memory prose only. Verified:
- `wave_26_*` keys read back correctly from `cross-repo-status.json`
- meta-issue #983 body refreshed and read-back-verified (prior body archived as a comment)
- roster validation passes for all 7 slate names
- pre-commit hooks green on both commits

Refs #983


parametrization:
Requestor: Wanjiku Mwangi
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: none

Program-bookkeeping review of both commits at head ceecb6d.

**scope(#983) — cross-repo-status.json**
- Verified all four keys `/wave-scope`'s upsert helper is documented to write are present and correctly named: `wave_26_scope_reconciled_at`, `wave_26_repos_in_scope`, `wave_26_meta_issue`, `wave_26_scope`. `/wave-kickoff` Step 0/0a read exactly these names — kickoff will not STOP on a missing/misnamed key.
- Ran `jq` over the PR's blob: JSON parses; `wave_26_scope` has all 3 tiers; `flatten | length` = 12; every one of the 12 rows is an object (not a bare string) with `id`/`ref`/`implementer`/`reviewer`/`reviewer_2`/`priority`; every `id` uses the full `noorinalabs-data-acquisition#NNN` form the kickoff-comment hook matches on. No #586-class regression.
- File shape preserved — diff is a 6-line targeted upsert inside the existing compact-inline block, not a `jq`-reformat churn (confirmed via `git diff`, no pretty-printed reflow).
- Cross-checked issue count/tiers against meta-issue #983's body ("12 issues, 1 child repo", 5/6/1 tier split) — matches exactly.
- `wave_26_meta_issue` changed shape from prior waves' quoted `"noorinalabs-main#NNN"` string to a bare integer `983`. Checked the consumer: `post_wave_kickoff_comment.py`'s meta-issue-skip does `str(meta) == issue_number`, i.e. it wants a bare number — the bare-int form this PR writes is what the hook actually needs (the old quoted-string form on wave_24/25 would have failed that comparison). Net: this is a correctness improvement, not a regression, and it also matches `/wave-scope` SKILL.md's own documented type ("integer").

**Point 4 — `repos_in_scope` includes `noorinalabs-main` — traced, not a bug.** All 12 issues are data-acquisition-only (meta-issue body says so explicitly), so at first glance this looked like it would make kickoff spin up a pointless `deployments/phase-9/wave-26` branch in main. Traced `/wave-scope` SKILL.md's own upsert logic (lines ~604-661): when the meta-issue body declares a strict subset of the canonical 8-repo list and no `WAVE_SCOPE_REPOS` override is given, the skill's descope path *always* rebuilds the array as `["noorinalabs-main"] + body-derived children` — the comment there states main is "always implicitly in scope." So `["noorinalabs-main","noorinalabs-data-acquisition"]` is exactly what the current skill produces unattended; it isn't an authoring error in this PR. wave-25's cleaner `["noorinalabs-data-acquisition"]`-only result was very likely a manual `WAVE_SCOPE_REPOS` override, not the default path. Consequence to flag for the wrapup owner (not blocking here): main will have 0 PRs merged to its wave-26 branch, so per wave-wrapup SKILL.md's acceptance rule (§ ~915) it will need an explicit de-scope/carry-forward record for main at wrapup — a minor known step, not a stall risk at kickoff.

**memory(#983) — feedback_gh_cli_gotchas.md + section_ci_tooling.md**
- Sections 12–13 are substantive, live-verified gotchas (GraphQL/REST quota independence + silent-zero failure mode; check-runs vs legacy status API) with correct cross-links (`[[feedback_silent_zero_is_not_a_measurement]]`, `[[feedback_verdict_head_sha_anchoring]]`) and the `section_ci_tooling.md` one-liner was extended to match, per the two-tier memory convention.
- `ontology/checksums.json` diff is exactly the expected PostToolUse tracker entry for the touched memory file — no scope creep.

**Two authors, one PR:** acceptable here — both commits are outputs of the same wave-26 `/wave-scope` run (same #983 issue), each carries correct per-commit `-c` identity (Nadia Khoury / Aino Virtanen with their own `parametrization+...@gmail.com`), and splitting a same-session memory note from its triggering scope commit would add PR overhead without a real isolation benefit.

CI verified against head SHA `ceecb6d` (not PR number): 21/21 checks SUCCESS, including the pre-commit⇄CI sync-drift gate and memory-budget gate. Mergeable: MERGEABLE.

Nothing here rises to blocking. Approved.

parametrization:
Requestor: Nino Kavtaradze
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: 1054

Reviewed at head ceecb6d53d037ae66f535853b87cecc2e7eb85b9 via gh api / git show against origin/main. CI green: statusCheckRollup 21/21 SUCCESS at that SHA.

## Executed, not read

wave_26 bookkeeping — JSON parses (419 top-level keys). All three tier_* arrays are assignment-row dicts, 12 rows, every row carrying id/ref/implementer/reviewer/reviewer_2/priority. I ran find_assignment_row from the real hook against the committed scope: 12/12 resolve, and check() on the real meta-issue returns skip_meta_issue. No #586 silent-skip exposure.

Wave counter transition is correct. Before: seq=25 with wave_26_meta_issue present, so reserved_wave=26, allocation_target=26. After: seq=26, reserved_wave=None, allocation_target=27. The reservation is committed, not skipped or re-issued.

Section 13 (check-runs vs status) reproduces on this PR's own head SHA: legacy status returns state=pending with 0 statuses, check-runs returns 21/21 success. The note states it correctly, including the query-by-SHA-not-PR-number caveat.

## An unflagged latent-bug fix in this diff

wave_26_meta_issue changes from "noorinalabs-main#983" to 983. The PR body does not mention it, but it is a real fix, and I would rather it be recorded than rediscovered.

post_wave_kickoff_comment.py:516 compares str(meta) == issue_number, where issue_number is a bare "983". Mutation-tested against the real hook on the meta_issue_skip fixture:

    INT    meta=347                    -> skip_meta_issue
    STRING meta="noorinalabs-main#347" -> post

The string form does not error; it falls through to post, so the meta-issue gets a per-issue kickoff comment that /wave-kickoff Step 8b owns exclusively. Waves 24 and 25 carry the string form and very likely received that spurious comment. Nothing in this PR needs to change — the int form is correct and matches every existing fixture — but there is no guard preventing the string form from being written again, and no fixture covers it. Filed as #1054.

## Two-tier index discipline (#1016) — correct

The one-liner goes to section_ci_tooling.md and MEMORY.md is untouched. Section count stays 21, matching the ToC's 21, which is right: sections 12 and 13 were appended to an existing note, so no new note file exists and no bump is due.

Cross-references: feedback_silent_zero_is_not_a_measurement, feedback_refresh_before_status_claim and feedback_verify_diagnosis_before_delegating all resolve. feedback_verdict_head_sha_anchoring does not exist. That is an acceptable forward-marker per CLAUDE.md, not a wrong-slug retarget — the nearest existing note, feedback_review_against_artifact, covers reading at PR head but not the specific caveat that a PR-number query re-resolves the head. Either write the note or retarget the link; not blocking.

## td_intake 0 of 1 — sound, with a caveat on how it is worded

Sound, and I checked rather than took it on trust. The nine tech-debt issues are pre-existing backlog, not freshly-filed theme work wearing a label: data-acquisition #295/#299/#373/#397/#424/#446 were created 2026-07-03 through 07-13, against a scope run on 07-20. Nine items genuinely drain the 47-item pool, so the intake rule's purpose is met in substance, not just arithmetically. Skipping a tenth would be ritual compliance.

The caveat is the wording, not the decision. The rationale keys on percentage-tech-debt-labeled. That predicate would also read as satisfied in a wave where the label was decorative on newly-filed theme work, and there it would license a genuine skip of debt intake. The load-bearing fact here is pool-drained-by-nine. Worth having the retro record the pool-drain framing when it revisits whether the formula base should count theme content regardless of label.

## Minor, non-blocking

Ships one dirty checksums entry (section_ci_tooling.md, last_resolved empty), so the next session's librarian will report 1 file pending. Legitimate change-tracker behaviour for a memory edit. Note the asymmetry: feedback_gh_cli_gotchas.md is not checksum-tracked at all, so the larger of the two memory edits dirtied nothing.

## Correction to my review brief

I was asked to re-verify two empirical claims attributed to section 12: that --no-ignore returns a confident partial of pure noise, and that an explicit path argument needs no --hidden. Neither claim is in this PR. Sections 12 and 13 are about GraphQL quota exhaustion and check-runs respectively; the whole file at head SHA has zero matches for no-ignore, --hidden, ripgrep or rg. There was nothing on #1052 to adjudicate. Flagging it so the brief's premise does not get recorded as a review finding — I have sent the measurements to the #1051 thread where they belong.

Approving. The bookkeeping is executable-correct, the memory claims that are actually present are true and I reproduced the one that was checkable, and the index discipline is right.

Issue #1050, opened 2026-07-20
content_ts defaults to None on the shared gate helpers, so a future caller fails open silently — the #1046 shape survives in the signature


parametrization:
Follow-up from the merge-gate review of #1049 (fixes #1046). Non-blocking — #1049 correctly binds both existing call sites, and I verified there are exactly two production callers repo-wide, both forwarding `content_ts`.

## The residual hazard

#1046 happened because `check_comment_reviews(..., content_ts: datetime | None = None)` lets a caller omit the keyword and get *silent staleness-filtering-off* rather than an error. `pr_review_state.py` omitted it and reported PASS on verdicts Hook 4 would reject.

#1049 fixes the two call sites and adds `partition_formal_reviewers(reviews, author, content_ts: datetime | None = None)` — a **third** function carrying the same defaulted-to-None parameter. The regression test guards the *current* callers by name; it cannot guard a caller that does not exist yet.

Current production call sites (all correct today):

```
.claude/hooks/validate_pr_review.py:1263  partition_formal_reviewers(reviews, author, content_ts)
.claude/hooks/validate_pr_review.py:1270  check_comment_reviews(number, branch_author_lastname, repo=repo, content_ts=content_ts)
.claude/hooks/validate_pr_review.py:1277  check_comment_reviews(number, "", repo=repo, content_ts=content_ts)
.claude/lib/pr_review_state.py:176        gate.partition_formal_reviewers(reviews, author, content_ts)
.claude/lib/pr_review_state.py:189,193    gate.check_comment_reviews(..., content_ts=content_ts)
```

## Why the default is load-bearing today

`content_ts=None` is a legitimate *value* — `get_latest_content_commit` returns `None` when a PR has no non-merge commits, meaning "no content binding exists". So the parameter must still accept `None`. The problem is only that it is **optional**, which makes "I forgot" and "there is genuinely no binding" indistinguishable at the call site.

## Suggested fix

Make `content_ts` a **required** parameter on both `check_comment_reviews` and `partition_formal_reviewers` (keep the `datetime | None` type). Both production callers already pass it, so the change is mechanical; omission becomes a `TypeError` at call time instead of a silent fail-open. Test call sites in `.claude/hooks/tests/test_validate_pr_review.py:117,147,907` would need an explicit `content_ts=None`, which is arguably clearer about what those tests assert.

This converts the #1046 failure mode from "silent, ships, discovered live on a merged PR" into "loud, caught by the test run".

[...]

PR #1056, opened 2026-07-20
fix(#981): fail closed when Hook 4 cannot resolve the merge target repo

parametrization:
## Summary

Hook 4 (`validate_pr_review.py`, the 2-reviewer merge gate) fell **OPEN** when
it could not resolve the merge target repo. This makes that path fail **CLOSED**.

`Refs #981`

## Corrected root-cause analysis — the issue body is wrong (verified)

Issue #981 attributes the fail-open to `_resolve_owner_repo` returning None
inside `check_comment_reviews`. **That path is not reachable on the merge path.**
I reproduced against HEAD and traced the actual control flow in `check()`:

1. `repo = extract_repo(command)` parses the command **pre-expansion** → the
   literal string `"$DA"`.
2. `get_pr_data(pr_number, repo="$DA")` runs `gh pr view 451 --repo '$DA'`,
   which exits non-zero → returns `None`.
3. The `if pr_data is None:` branch returned
   `{"decision": "allow", "systemMessage": "WARNING: Could not verify..."}`.

That early `allow` short-circuited **before** `get_latest_content_commit` (so
the #950 `CommitFetchError` hard-block never ran) and before
`check_comment_reviews` was ever called. Fixing `_resolve_owner_repo` alone
would not have closed it.

## Before / after repro (`-R $DA`)

```
# BEFORE (HEAD)
gh pr merge 451 -R $DA --merge --delete-branch   -> decision=allow  ("WARNING: Could not verify...")

# AFTER
gh pr merge 451 -R $DA --merge --delete-branch   -> decision=block  (UNEXPANDED variable diagnostic)
gh pr merge 451 --repo noorinalabs/$REPO --merge -> decision=block  (partially-expanded — nastiest form)
gh pr merge 451 --repo justaname --merge         -> decision=block  (malformed: no owner/name)
gh pr merge 451 -R $DA --admin --merge           -> allow (admin override preserved)
gh pr merge 451 --repo noorinalabs/noorinalabs-data-acquisition --merge  (2 approvers) -> allow
```

## What changed

**Primary (`check()`):**
- New `repo_argument_defect` classifier + `describe_repo_defect`, **reusing the
  file's existing `_is_single_expansion_word` unexpanded-variable machinery**
  rather than a second hand-rolled regex (a duplicate matcher is the drift that
  was #1046). Blocks an unresolvable `--repo` **before any fetch** — the block
  is deterministic and network-independent.
- The `pr_data is None` branch now **blocks** with retry/auth guidance instead
  of allow-with-warning.
- The two failure kinds get **distinct diagnostics**: unexpanded-variable (fix:
  pass a literal `--repo owner/name`) vs generic fetch failure (fix: retry /
  `gh auth status` / check PR number). Modeled on the existing #950
  `CommitFetchError` block, and calls `log_pretooluse_block(...)` as that
  sibling does.

**Defense-in-depth decision (`check_comment_reviews`, the #1050 latent signature):**
I **did** harden the three same-shape latent fail-opens (unresolved repo ~830,
comments-API non-zero ~851, bare `except` ~910). Each previously returned an
empty reviewer set that reads as "0 approvals found" rather than "could not
determine". The trade-off was clean because I did it **additively**: a new
`CommentReviewResult.undetermined` string field with a falsy default, which
`check()` and `pr_review_state.compute_review_state` hard-block on when
non-empty. I deliberately did **not** raise a new exception — `pr_review_state`
constructs and consumes `CommentReviewResult` directly and a raise escaping
`check_comment_reviews` would turn its clean exit-2 path into an uncaught
traceback. The additive field leaves every existing construction/read (and
~30 unit callers) working untouched, so the unit-call contract is preserved.

**`pr_review_state.py` (imports the gate — kept in lockstep, #1046):**
unresolvable `--repo` raises a specific `ReviewStateError` before fetching
(reusing the gate's own classifier so the two surfaces give the same
diagnosis); an incomplete comment scan raises rather than reporting a
zero-approval PR.

## Tests — each verified to FAIL against pre-fix source

`.claude/hooks/tests/test_validate_pr_review.py`:
`RepoArgumentDefectTests`, `UnresolvableRepoFailsClosedTests`,
`IncompleteCommentScanFailsClosedTests`.
`.claude/lib/tests/test_pr_review_state.py`: `UndeterminableRepoAndScanTests`.

I stashed the source fix (kept the new tests) and ran the new classes against
pre-fix code: **29 behavior-change assertions failed** (many via `AttributeError`
— the new symbols don't exist pre-fix), including the exact regression
`test_unexpanded_repo_var_blocks`. The **6** that passed pre-fix are precisely
the false-positive guards that must hold in BOTH states (literal repo + 2
approvers still allows; `--admin` still overrides; absent `--repo` still allows;
literal repo still reaches the fetch). Full suite after fix: **2680 passed**.

## Gates

`pre-commit run --all-files` (commit + pre-push stages), `mypy .claude/hooks/
.claude/lib/ --ignore-missing-imports`, both pytest suites, and the
`pre_commit_ci_sync` drift gate — **all green**. No `--no-verify`, no
force-merge.

## TechDebt

`TechDebt: none`

https://claude.ai/code/session_01T4RB6kXqh5c6ckHC8p9xHf


parametrization:
Requestor: Lucas Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Reviewed at head SHA ea0c0af (worktree checkout, not `gh pr checkout` in the parent). Ran the full `.claude/hooks/tests/test_validate_pr_review.py` + `.claude/lib/tests/test_pr_review_state.py` suite: 237 passed / 3 subtests passed.

1. **Tests bite, verified by regression, not by reading.** `git checkout ea0c0af^ -- .claude/hooks/validate_pr_review.py .claude/lib/pr_review_state.py` (reverting only the two source files, keeping the new tests at head) turned 29 of the new tests red, including the two that matter most:
   - `test_generic_fetch_failure_blocks_with_a_distinct_diagnostic`: pre-fix asserts `result["decision"] == "block"`, gets `'allow'` — this is the literal #981 fail-open reproduced.
   - `test_unexpanded_repo_var_blocks`: pre-fix `result` is `None` even with `get_pr_data` mocked to return approved PR data — confirms nothing in the old code path ever inspected the repo argument itself before falling through to allow.
   - Restored the source files afterward; full suite back to 237 passed.

2. **Diagnostics are distinct and each points at the right fix.** Called `repo_argument_defect`/`describe_repo_defect` directly: `$DA`, `${DA}`, `$(get_repo)`, `noorinalabs/$REPO` → `REPO_DEFECT_UNEXPANDED` (message: 'contains an UNEXPANDED shell variable... not a repository'); `justaname`/`/name`/`owner/` → `REPO_DEFECT_MALFORMED` (message: 'not of the form OWNER/NAME'). The generic-fetch-failure block message is a third, separate string with retry/`gh auth status` guidance — confirmed via `test_the_two_failure_kinds_do_not_share_a_message` and by reading the two block bodies in `check()`.

3. **Lockstep confirmed.** `test_pr_review_state.py`'s `UndeterminableRepoAndScanTests` imports `gate.repo_argument_defect`/`gate.describe_repo_defect` directly (no duplicate matcher) and asserts the driver's `ReviewStateError` text matches the gate's classification verbatim — both modules ran together at 237/237.

4. **No over-block.** `test_literal_repo_on_approved_pr_still_allows` and `test_no_repo_flag_on_approved_pr_still_allows` pass at head — a well-formed `--repo owner/name` (or no `--repo` at all) on a 2-approver PR still returns `None` (allow). These two also already passed pre-revert, confirming the fix didn't touch the legitimate path — only the defect classification is new.

5. Ruff + mypy spot-checked clean on all four changed files (matches CI).

One non-blocking observation, not filing tech-debt for it (too minor/cosmetic to track): `repo_argument_defect`'s docstring says a whitespace-bearing $-containing value 'fails the predicate but then fails the slash test' to get classified as a defect — I checked with `'noorinalabs $REPO extra words'` and it's actually caught earlier by the `or "$" in repo" disjunct, never reaching the slash test. Outcome (blocked as UNEXPANDED) is correct either way — just the docstring's stated mechanism doesn't match the code path. No behavior change needed.

parametrization:
Requestor: Nino Kavtaradze
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: 1057

Merge-gate review (security / fail-closed lens). Ran the instrument in the head-SHA worktree (ea0c0af), did not trust the PR body.

Hole closed — verified by executing check() directly. Every unresolvable form BLOCKS: -R $DA, -R "$DA", --repo ${DA}, --repo=$DA, -R $DA/foo, --repo noorinalabs/$REPO (the partially-expanded slash-bearing trap), $(get_repo), and the malformed --repo justaname / /name / owner/ — including inside the loop form. The classifier repo_argument_defect returns the right kind (unexpanded vs malformed) for each, and the --admin escape still short-circuits ahead of every guard (get_pr_data not called). No over-block: a literal --repo owner/name and an absent --repo on a 2-approver PR both still ALLOW (returns None) once the #950 content-binding fetch is neutralized.

No new fail-open: rg confirms zero remaining decision:allow / systemMessage returns in the file — the pre-#981 pr_data-is-None allow-with-warning is gone, and both new block paths plus the incomplete-scan block all call log_pretooluse_block and carry the --admin line.

Lockstep intact: pr_review_state.py mirrors the gate via gate.repo_argument_defect / describe_repo_defect and raises ReviewStateError on the same inputs; CommentReviewResult.undetermined is additive (falsy default) so existing constructions/reads are untouched. Full suite green: 237 passed.

Test quality proven by mutation, not assertion-counting. Neutralizing the three new guards turned 12 hook tests red (including test_unexpanded_repo_var_blocks, test_unexpanded_repo_var_never_allows_with_warning, test_block_happens_before_any_network_call, the generic-fetch-failure and incomplete-scan hard-block tests); neutralizing the pr_review_state guards turned 3 more red. The false-positive-guard tests stayed green against the mutant, as they should. Files restored clean after. CI is all-green at the head SHA.

TechDebt #1057 (non-blocking, pre-existing, not this PR's code): the attached short-flag form gh pr merge 451 -R$DA (no space) makes extract_repo in _repo_flag_parse.py return None, so the classifier never sees it and the hook falls through to a cwd-resolved check — a repo-confusion fail-open of the #981 family. Root cause is the tokenizer in _repo_flag_parse.py (not in this diff); the classifier this PR adds is correct for every value it receives. Filed separately rather than blocking a security improvement that fully closes its stated scope.

PR #359, opened 2026-05-11
charter(agents.md): § Pre-Spawn State Check § Surface enumeration (P3W8 retro-pickup, closes #341)

parametrization:
## Summary

Extends `.claude/team/charter/agents.md § Pre-Spawn State Check` with a new `### Surface enumeration` H3 sub-section + 6 H4 sub-rules consolidating three P3W8 pre-spawn-discipline memories into one charter-codified primitive. Closes #341.

## What this adds

A unified pre-spawn surface-enumeration discipline addressing three companion failure shapes from P3W8, organized by question:

| Question | H4 sub-rule | Source memory |
|----------|-------------|---------------|
| **Where** to verify? | Origin head_sha, not local checkout | `feedback_pre_spawn_verify_at_origin.md` |
| **How** to count? | `grep -c` per file + sum; never `\| head -N` | `feedback_no_head_in_surface_enumeration.md` (new in W8) |
| **What caveats** apply? | Per-named-caveat applicability sweep at HEAD | `feedback_pre_spawn_brief_verified_at_head.md` |

Plus a severity table covering 4 failure shapes (under-count brief / scope-block miss / caveat-passthrough / positive-event implementer-override), worked examples citing P3W8 evidence (deploy#280, landing-page#88, data-acquisition#43+#44, isnad-graph hook surface, Maeve/#309 unblock), and cross-references to the companion reviewer-class primitives at the PR layer.

## What this deliberately does NOT include

The W9 sibling memory `feedback_spawn_brief_requestor_field_semantics.md` (Requestor-field swap in reviewer-spawn briefs) was discussed in the spawn brief as a fold-in candidate. **Deferred** to a separate PR for two reasons:

1. **Different layer.** Surface-enumeration discipline is about counting code-surface occurrences; the Requestor-field lesson is about header-field semantics in reviewer-spawn briefs. Folding both into one sub-section would dilute the surface-enumeration focus.
2. **In-flight enforcement path.** `noorinalabs-main#356` already tracks expanding the `validate_pr_review` BLOCKED message to call out the Requestor/Requestee swap — hook-tier enforcement is the right end-state per `feedback_enforcement_hierarchy.md`. A charter promotion of that lesson should land alongside or after the hook-message expansion to avoid charter rules getting ahead of their enforcement.

P3W10 candidate for the Requestor-field charter promotion.

## Acceptance criteria (from #341)

[...]

PR #1071, opened 2026-07-22
fix(#940): reviewer set follows each reviewer's latest verdict, not a monotonic union

### LINKED ISSUE #940: Hook 4 reviewer set is monotonic: an approval cannot be withdrawn, and a blocking reviewer counts as an approver

parametrization:
## Summary

Closes #940.

`check_comment_reviews` (Hook 4 / `validate_pr_review.py`) built its reviewer
set by only ever **adding** a Requestor on an `Approved` comment and never
removing one on a later `Changes Requested` — an approval could never be
withdrawn, and a reviewer who moved from `Approved` to `Changes Requested`
still counted toward the 2-reviewer threshold. This was live on
`noorinalabs-data-acquisition` PR #359 at the time of filing.

## Fix

Track each non-author reviewer's **latest** CURRENT (non-stale per #950)
verdict in a single chronological pass over the comment thread, then derive
`result.reviewers` from whichever reviewers' latest verdict is `Approved`. A
later verdict now correctly supersedes an earlier one in **both**
directions (Approved → CR withdraws; CR → Approved grants).

## Tests

Added `LatestVerdictSupersedesTests` in
`.claude/hooks/tests/test_validate_pr_review.py`:
- `test_approved_then_changes_requested_is_excluded` — the guard the issue
  calls out explicitly: this must RED under the pre-fix monotonic union.
- `test_changes_requested_then_approved_is_included` — the symmetric case.
- `test_da359_shaped_timeline_reproduces_the_correct_verdicts` — reproduces
  the exact da#359 timeline from the issue report and asserts the corrected
  2-approver outcome.
- `test_stale_verdict_does_not_override_a_later_current_one` — guards the
  interaction with the #950 staleness binding.

Full local gate green: `pre-commit run --all-files` (commit stage) +
`pre-commit run --hook-stage pre-push --all-files` (mypy + full pytest
suite, 2725 passed) — both clean.

## Reviewers

- **Nino Kavtaradze** (merge-gate)
- **Lucas Ferreira**

### DIFF
--- a/.claude/hooks/tests/test_validate_pr_review.py
+++ b/.claude/hooks/tests/test_validate_pr_review.py
@@ -377,6 +377,116 @@ class ReviewerDedupTests(_CheckCommentReviewsHarness):
         self.assertEqual(result.reviewers, set(), "branch author must not self-review")
 
 
+class LatestVerdictSupersedesTests(_CheckCommentReviewsHarness):
+    """Issue #940: the reviewer set is monotonic — an approval cannot be
+    withdrawn, and a reviewer standing at Changes Requested still counts as
+    an approver if they ever approved earlier in the thread.
+
+    The fix keys each reviewer's verdict by their LATEST charter-format
+    comment (chronological = fixture list order, since these tests pass no
+    `content_ts` and so exercise the unbound #950 path). Only a reviewer
+    whose latest verdict is Approved contributes to the reviewer set.
+    """
+
+    @staticmethod
+    def _verdict(requestor: str, ror: str, requestee: str = "Linh Pham") -> dict:
+        return {
+            "body": (
+                f"Requestor: {requestor}\nRequestee: {requestee}\n"
+                f"RequestOrReplied: {ror}\nTechDebt: none"
+            ),
+            "user": {"login": "anyone"},
+        }
+
+    def test_approved_then_changes_requested_is_excluded(self):
+        """Guard for #940: this must RED under the pre-fix monotonic union.
+
+        An Approved followed by a Changes Requested from the SAME reviewer —
+        the fixture shape the issue calls out as the one that proves the
+        defect (a fixture with only the reverse order "passes under both
+        implementations and proves nothing").
+        """
+        comments = [
+            self._verdict("Oyunbileg Batbayar", "Approved"),
+            self._verdict("Oyunbileg Batbayar", "Changes Requested"),
+        ]
+        result = self._run_with_fake_api(comments, self.BRANCH_AUTHOR, repo=self.REPO)
+        self.assertNotIn(
+            "oyunbileg batbayar",
+            result.reviewers,
+            "a later Changes Requested must withdraw the earlier Approved",
+        )
+        self.assertEqual(result.reviewers, set())
+
+    def test_changes_requested_then_approved_is_included(self):
+        """The symmetric case: a later Approved supersedes an earlier block."""
+        comments = [
+            self._verdict("Kwesi Boateng", "Changes Requested"),
+            self._verdict("Kwesi Boateng", "Approved"),
+        ]
+        result = self._run_with_fake_api(comments, self.BRANCH_AUTHOR, repo=self.REPO)
+        self.assertIn("kwesi boateng", result.reviewers)
+
+    def test_da359_shaped_timeline_reproduces_the_correct_verdicts(self):
+        """Reproduces the exact da#359 timeline from #940's report.
+
+        Chronological verdicts:
+          Oyunbileg Batbayar: Approved -> Changes Requested -> Changes Requested
+          Alejandra Reyes-Fuentes: Approved
+          Kwesi Boateng: Changes Requested -> Approved
+
+        Standing verdict is Oyunbileg BLOCKING, Alejandra and Kwesi APPROVED —
+        exactly 2 current approvers, not the pre-fix count of 3.
+        """
+        comments = [
+            self._verdict("Oyunbileg Batbayar", "Approved"),
+            self._verdict("Alejandra Reyes-Fuentes", "Approved"),
+            self._verdict("Oyunbileg Batbayar", "Changes Requested"),
+            self._verdict("Oyunbileg Batbayar", "Changes Requested"),
+            self._verdict("Kwesi Boateng", "Changes Requested"),
+            self._verdict("Kwesi Boateng", "Approved"),
+        ]
+        result = self._run_with_fake_api(comments, self.BRANCH_AUTHOR, repo=self.REPO)
+        self.assertEqual(
+            result.reviewers,
+            {"alejandra reyes-fuentes", "kwesi boateng"},
+        )
+        self.assertNotIn("oyunbileg batbayar", result.reviewers)
+        self.assertEqual(len(result.reviewers), 2)
+
+    def test_stale_verdict_does_not_override_a_later_current_one(self):
+        """A verdict predating T_content is excluded from `latest_verdict`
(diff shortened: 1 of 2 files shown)


parametrization:
**Merge-gate review — PR #1071 (fixes #940)**

Reviewed the diff at head `4aa71b5` against `deployments/phase-9/wave-27`; CI is green at that SHA, pytest and mypy included.

This correctly replaces the monotonic reviewer union with a latest-verdict-wins model. Each reviewer's chronologically-last CURRENT verdict is recorded in `latest_verdict`, and the reviewer set is derived after the loop as only those whose latest verdict is an approval. Because the comment list is already chronological, the last write per reviewer is that reviewer's latest verdict — a later Changes Requested withdraws an earlier approval, and a later approval supersedes an earlier block.

I checked the ordering against the staleness path: the stale-verdict `continue` fires before the `latest_verdict` write, so a verdict predating T_content can never overwrite a reviewer's genuinely-latest current verdict. Since T_content is a fixed point, every current verdict is chronologically after every stale one, so the recorded value is always the latest current verdict — the excluded/absent cases resolve correctly.

The guards are mutation-sensitive, not decorative. `test_approved_then_changes_requested_is_excluded` is RED under the pre-fix union (which would add the reviewer on the approval and never remove them) and GREEN here; the da#359-shaped timeline test pins the exact "2 current approvers, not 3" outcome from the issue report. This is the fixture shape the issue itself calls out as the one that actually proves the defect.

Approving.

---
Requestor: Lucas Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

parametrization:
Merge-gate review at head `4aa71b5` (base `deployments/phase-9/wave-27`). Read the diff through `gh` at the head SHA; no local checkout.

Substance — `check_comment_reviews` now derives the reviewer set from each reviewer's LATEST *current* verdict (a post-loop rebuild from the `latest_verdict` map) instead of the old monotonic union of every Approved comment. A later Changes Requested from a reviewer withdraws their earlier Approved, and a later Approved supersedes their earlier block. The #950 stale-verdict `continue` executes before the record step, so a stale (pre-`content_ts`) verdict never lands in `latest_verdict`; and since a stale comment can only precede a current one chronologically, the ordering cannot invert.

Verified the guard is not inert rather than trusting the assertions — fetched both files at the head SHA, ran the new `LatestVerdictSupersedesTests` (4 green), then reverted the fix to the pre-#940 mid-loop union. Under that mutation `test_approved_then_changes_requested_is_excluded` and `test_da359_shaped_timeline_reproduces_the_correct_verdicts` both go red (the withdrawal case and the exact two-approver count). The discriminating guards genuinely trip on the defect.

CI is green at the head SHA (14 checks) and the branch is mergeable.

One non-blocking observation — the diff also carries the wave-27 kickoff mutation to `cross-repo-status.json`. That is a base-divergence artifact (wave-branch tip `582029f` predates kickoff commit `6c50ca6`, which the feature branch inherited from `main`); it is identical and idempotent across every wave-27 PR, not authored churn.

Approving.

---
Requestor: Nino Kavtaradze
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

PR #1072, opened 2026-07-22
fix(#1047): premise_check no longer treats a bare slash as a path signal

### LINKED ISSUE #1047: premise_check: any backtick token containing '/' is treated as a path — 12/12 false STOP on wave-26 scope

parametrization:
## Summary

`premise_check.py`'s `looks_like_path` accepted any backtick token containing `/` unconditionally — this produced a **12/12 false-STOP** on the wave-26 `/wave-scope` run (`A/B`, `recall/precision`, `origin/main`, `986/650`, etc. all contain a slash but are not paths).

## Fix

- `looks_like_path` now requires a positive signal: a known code/doc extension, OR a slash **plus** a known repo-root leading component (`src/`, `.claude/`, a child-repo name, ...). A bare slash is no longer sufficient.
- Explicitly rejects a bare `/`, all-numeric fractions (`986/650`), and git refs (`origin/*`, `refs/*`, `HEAD`).
- **Basename fallback** (da#373): a slash-free filename that misses at its literal path but resolves uniquely elsewhere in the tree (via `git ls-tree`) is now `EXISTS`, not `MISSING`.
- **Cross-repo resolution** (da#427): a `.claude/`-rooted path that misses in a child repo but resolves in the parent `noorinalabs-main` downgrades to a new `CROSS_REPO` status → `WARN`, not `STOP`.
- Corrected `/wave-scope` SKILL.md § 6.5's inaccurate "prose never produces a false STOP" description of the mechanism.

## Tests

- Pins all 12 wave-26 false-positive issue bodies as a regression fixture (`test_wave_26_regression_fixture_no_false_positives`), asserting `OK` across the board while the 4 genuine paths in that run still extract.
- Mutation-verification test (`test_mutation_verify_old_rule_would_fail_this_fixture`) proves the old `"/" in tok` rule fails the fixture.
- New coverage for the basename fallback and cross-repo downgrade, both with real-git integration tests.
- Full `.claude/lib/tests/` suite passes (791 tests); `pre-commit run --all-files` + push-stage (mypy, pytest) all green.

Closes #1047

**Reviewers:** Aino Virtanen (merge-gate) + Nurul Hakim

### DIFF
--- a/.claude/lib/premise_check.py
+++ b/.claude/lib/premise_check.py
@@ -32,6 +32,13 @@ Verdict policy (deterministic):
     deliberately NOT a STOP: a missing local checkout is an environment gap, not
     evidence the premise rotted — a false STOP there would block scope on
     tooling state, the opposite of the gate's purpose.
+  * a ``.claude/``-rooted path that MISSES in a child repo but resolves in the
+    parent (``noorinalabs-main``) -> ``CROSS_REPO`` -> verdict ``WARN`` (main#1047
+    da#427): very often a legitimate reference to org-wide config, worth a
+    manual look rather than a hard block.
+  * a bare (slash-free) filename that misses at repo root but resolves
+    uniquely elsewhere in the tree -> ``EXISTS`` via a basename fallback
+    (main#1047 da#373).
   * everything present, or no concrete refs to check -> ``OK``.
 
 An issue may also declare explicit ``paths`` / ``symbols`` arrays (deliberate
@@ -77,6 +84,12 @@ _REPO_ROOT = Path(__file__).resolve().parents[2]
 EXISTS = "exists"
 MISSING = "missing"
 UNVERIFIABLE = "unverifiable"
+# A `.claude/`-rooted path that misses in a CHILD repo but resolves in the
+# parent (`noorinalabs-main`) — very often a legitimate reference to org-wide
+# config from a child-repo issue (main#1047 da#427). Treated as WARN, not
+# STOP: the reference is real, but reading it against the wrong repo is worth
+# a manual look, not a hard block.
+CROSS_REPO = "cross_repo"
 
 # Per-issue verdicts.
 OK = "ok"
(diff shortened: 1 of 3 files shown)


parametrization:

Reviewed at head `6c10392`. All 21 checks are green at that SHA, base is `deployments/phase-9/wave-27`.

Verified the fix by running the suite at head (40 passed) and by mutation-testing the core guard: reinstating the old `"/" in tok` short-circuit right after the digit check kills 6 assertions, including the two dedicated mutation guards (`test_mutation_guard_slash_alone_would_readmit_prose`, `test_mutation_verify_old_rule_would_fail_this_fixture`) and the wave-26 fixture. The 12-body regression fixture is production-realistic (verbatim false-STOP tokens from the wave-26 scope run), and the four embedded real paths still extract. The `CROSS_REPO` -> WARN downgrade and the slash-free basename fallback are both correctly scoped (only `.claude/`-rooted child-repo misses get the parent second chance; a genuine deletion still STOPs). No must-fix items.

---
Requestor: Aino Virtanen
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Review — #1072 (fixes #1047)

Reviewed at HEAD `6c103921` against base `deployments/phase-9/wave-27` (second reviewer; Aino Virtanen is merge-gate).

**Verified by execution, not by reading.** I imported the HEAD `premise_check.py` and ran `looks_like_path` over the tokens directly:

- All 12/12 wave-26 false-positive tokens (`A/B`, `recall/precision`, `origin/main`, `986/650`, `650,986/650,986`, `ism/kunya`, `taṣliya/eulogy`, `boundary/particle`, `Code/git`, …) now return `False` — the false-STOP class is closed.
- All four genuine paths from that same run (`src/resolve/ner.py`, `src/utils/arabic.py`, `docs/testing-on-subsets.md`, `data/curated/narrators_canonical.parquet`) plus `src/parse` and `noorinalabs-deploy/terraform/main` still return `True` — no real-path regression.
- The mutation-verification test (`test_mutation_verify_old_rule_would_fail_this_fixture`) is a real guard: the re-implemented old `"/" in tok` rule returns `True` for `recall/precision` while the new one returns `False`, so the regression fixture actually exercises the fix rather than sitting inert.

The git-ref exclusion (`origin/`, `refs/`, `HEAD`), the all-numeric-fraction exclusion, the bare-`/` reject, the `CROSS_REPO`→WARN downgrade for `.claude/`-rooted child-repo misses that resolve in the parent, and the slash-free basename fallback are all correct and each has a real-git or injected-checker test. The `check_issue` append restructure keeps exactly one append per candidate on both branches (no double/missing append). SKILL.md §6.5's inaccurate "prose never produces a false STOP" claim is corrected.

**Non-blocking observation (not a must-fix).** The known-root gate narrows the false-positive surface but doesn't fully close it: prose whose first `/`-component is a known-root word still classifies as a path. Confirmed by execution — `test/prod`, `data/metadata`, `docs/appendix`, `src/dest`, `scripts/manual` all return `True`, and `extract_path_candidates` pulls them from *bare* prose, not just backtick spans. So `"promote from test/prod nightly"` would still STOP if `test/prod` doesn't resolve. This is a large net improvement over the any-slash rule and precision-over-recall is the explicit, correct design, so it does not block — flagging it only so the team can decide whether a future tightening (e.g. requiring a second path-shaped component, or a dot/extension somewhere) is worth it if `test/`-`data/`-`docs/` shorthand recurs in issue bodies.

CI: all 21 checks green at head. Approving.

---
Requestor: Nurul Hakim
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: none

PR #1074, opened 2026-07-22
fix(#1039): generalize ontology_tracker SKIP_PATTERNS via owning-repo check-ignore

### LINKED ISSUE #1038: ontology_tracker: gitignored session_handoff.md is tracked — every /handoff dirties committed checksums.json

### LINKED ISSUE #1039: ontology_tracker: SKIP_PATTERNS denylist leaks for gitignored machine-local files — generalize via owning-repo check-ignore

parametrization:
## Summary

`ontology_tracker.py`'s `SKIP_PATTERNS` is a hand-maintained substring denylist that has leaked twice for the same class of file — gitignored, machine-local, frequently-rewritten artifacts (`.claude/annunaki/errors.jsonl`, then `.claude/memory/session_handoff.md` in #1038) — each leak manufacturing permanent phantom drift in `ontology/checksums.json` until someone notices and hand-extends the list.

This adds a generalized backstop: resolve each file's nearest `.git` ancestor (its **owning repo**, not `REPO_ROOT`) and run `git check-ignore` there.

## Why not the naive fix

#1038 originally proposed deriving the skip from `git check-ignore` run at `REPO_ROOT`. That was tested and rejected: the parent repo `.gitignore`s every child repo wholesale, so running check-ignore from `REPO_ROOT` reports **every child-repo file as ignored** — 147 of 284 tracked entries (52%), including real committed source like `noorinalabs-deploy/.github/workflows/deploy-prod.yml` and every child ontology file. That would silently blind the tracker to over half the semantic overlay while `/session-start` kept reporting "0 dirty" — worse than the nuisance it fixes.

## What this does instead

1. `_find_git_root` walks up from the file to the nearest ancestor containing `.git` — the file's **owning** repo.
2. `_is_git_ignored` runs `git check-ignore -q` there, on the path relative to that repo.
3. **Fails open** on every error case (no `.git` ancestor, unresolvable relative path, subprocess failure/timeout) — the file gets tracked, never silently skipped. Under-tracking is a silent loss of drift detection; over-tracking is merely noise, so the asymmetry is deliberately one-sided.
4. Result is cached per `(repo, relative-path)` for the process's lifetime.
5. `SKIP_PATTERNS` stays as the fast, no-subprocess path for policy-driven skips (`checksums.json` self-skip, the generated structural layer, worktree paths) that aren't gitignore facts and check-ignore would never catch.

A follow-on fix along the way: the new `git check-ignore` subprocess (and the test suite's own `git init` calls) needed to strip any inherited `GIT_DIR`/`GIT_WORK_TREE`/`GIT_INDEX_FILE` (main#719) — without it, running the suite via the `pre-push` git hook leaked the real repo's git env into the fake test repos, causing 3 tests to fail only when invoked through `git push`, not standalone. Fixed via a `_hermetic_git_env()` helper used by both the production subprocess call and the tests' own `git init` calls.

## Tests

New `GitCheckIgnoreGeneralizationTests` class in `.claude/hooks/tests/test_ontology_tracker.py` — uses real `git init` repos (not mocks):
- The 52% regression guard: a child-repo file gitignored by the parent but tracked in its own nested repo → still tracked.
- A file genuinely gitignored by its owning repo → skipped.
- A non-ignored file → tracked as normal.
- No `.git` ancestor at all → fails open.
- `git` subprocess failure/timeout → fails open.
- Caching: a second call for the same path does not re-invoke `git`.
- End-to-end `check()` dispatcher path for a gitignored file writes no checksums entry.

All 39 tests in the file pass, both standalone and via the full `pre-push` pytest suite (2726 passed) — the latter is the regression check for the GIT_* leak fix.

## Reviewers

Merge-gate: **Aino Virtanen**. Second reviewer: **Nino Kavtaradze**.

Closes #1039


parametrization:

Reviewed at head `9f7f584`. All 21 checks are green at that SHA, base is `deployments/phase-9/wave-27`.

Ran the suite at head (39 passed) and mutation-tested the central correctness claim — that check-ignore must resolve each file's OWNING repo, not `REPO_ROOT`. Replacing `_find_git_root(resolved_path)` with the naive `REPO_ROOT` kills exactly `test_child_repo_file_ignored_by_parent_is_still_tracked`, the 52%-blind-spot guard. That guard uses real `git init` repos (parent gitignoring the child directory wholesale, child file tracked by its own repo), so it exercises actual git plumbing rather than a mock. Fail-open on every error path (no `.git` ancestor, out-of-repo, subprocess failure/timeout) is the correct asymmetry — under-tracking is a silent loss, over-tracking is noise — and each branch is covered. The `GIT_*`-stripping hermetic env is load-bearing under the pre-push pytest hook. No must-fix items.

---
Requestor: Aino Virtanen
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: none

parametrization:
Merge-gate review at head `9f7f584` (base `deployments/phase-9/wave-27`). Diff read through `gh` at the head SHA; no local checkout.

Substance — generalizes the hand-maintained `SKIP_PATTERNS` denylist with an owning-repo `git check-ignore` backstop. The key correctness point is that it resolves each file's NEAREST `.git` ancestor and asks that repo, not `REPO_ROOT` — the parent repo gitignores every child wholesale, so a `REPO_ROOT`-rooted check-ignore would report ~52% of tracked entries (real child-repo source and ontology files) as ignored and silently blind the tracker while it reported "0 dirty". It fails OPEN on every error (no `.git` ancestor, path outside its repo, subprocess failure/timeout, exit 128), keeps `SKIP_PATTERNS` as the fast no-subprocess path, memoizes per (repo, rel-path), and strips `GIT_*` from the subprocess env so an inherited `GIT_DIR` (the pre-push pytest hook, #719) can't retarget the check.

Ran the guard against real `git init` repos rather than trusting mocks — the new `GitCheckIgnoreGeneralizationTests` (built on actual nested repos) are green at head (39 in the module). Then mutated `_find_git_root` to return the OUTERMOST root (the naive `REPO_ROOT` bug); `test_child_repo_file_ignored_by_parent_is_still_tracked` — the 52%-regression guard — goes red. The fail-open, timeout, and cache tests also exercise the real subprocess boundary.

CI is green at the head SHA (21 checks) and the branch is mergeable. The `cross-repo-status.json` delta is the shared wave-27 kickoff base-divergence artifact.

Approving.

---
Requestor: Nino Kavtaradze
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: none

Issue #707, opened 2026-06-18
tech-debt(tooling): pr_review_state.py — deterministic PR review-state query over check_comment_reviews() (pre-spawn + pre-merge)


parametrization:
## Problem

There is no deterministic way to query a PR's **charter review state** before spawning reviewers or before merging — so we hand-derive it, and hand-derivation is wrong in two recurring ways:

1. **`gh pr view --json reviews` is empty for comment-based verdicts.** Our verdicts are issue-comments (`Requestor:/RequestOrReplied: Approved/TechDebt:`), not formal GitHub reviews. A PR can read "0 reviews" while already carrying 2 valid Approved verdicts → **redundant reviewer spawns** (hit on lp#140, P5W5 2026-06-17).
2. **A crashed/partial session can leave a verdict missing its `TechDebt:` line**, which blocks merge wave-wide — invisible until the `gh pr merge` hook-block. (Also hit on lp#140: a stale Anika verdict lacked TechDebt.)

The merge gate `validate_pr_review.py::check_comment_reviews()` ALREADY computes the correct answer (distinct Approved `Requestor`s excluding the branch author; reviewers missing TechDebt; TechDebt issue numbers). We just don't expose it as a queryable command — so the pre-check drifts from the gate.

## Proposal

Add a thin CLI in `.claude/lib/` (e.g. `pr_review_state.py <pr> --repo <owner/repo>`) that:
- Resolves the PR head ref → branch-author lastname (reuse the hook's existing head-ref parse), then calls `check_comment_reviews()` directly (no reimplementation — import the canonical function so it can never drift from the gate).
- Prints: distinct Approved reviewers + names, count vs the 2-reviewer threshold (and the `wave-bootstrap` single-reviewer exception), any reviewers whose verdict is **missing TechDebt**, and the extracted TechDebt issue numbers.
- Exit non-zero when the gate would block, zero when it would pass — so it's usable as a pre-merge guard and a pre-spawn check.

## Why (enforcement hierarchy)
Replaces a decay-prone memory note ("check comments, not `.reviews`") with a deterministic tool that reuses the gate's own logic — same move as `wave_status.py` (main#688). Use it (a) before spawning reviewers to avoid redundancy, and (b) before `gh pr merge` to see exactly what's missing instead of eating a hook-block round-trip.

## Acceptance
- CLI imports and calls `check_comment_reviews()` (no parser fork).
[...]

PR #1075, opened 2026-07-22
fix(#1050): make content_ts required on the shared gate helpers

### LINKED ISSUE #1046: pr_review_state.py drops content_ts: the review-state driver never marks stale verdicts, reports PASS where Hook 4 would not

### LINKED ISSUE #1050: content_ts defaults to None on the shared gate helpers, so a future caller fails open silently — the #1046 shape survives in the signature

parametrization:
## Summary

Closes #1050.

Follow-up from the merge-gate review of #1049 (fixed #1046). #1046 happened
because `check_comment_reviews(..., content_ts: datetime | None = None)`
let a caller omit the keyword and silently get staleness-filtering OFF
rather than an error — `pr_review_state.py` omitted it and reported PASS on
verdicts Hook 4 would reject. #1049 fixed the two production call sites but
left the defaulted parameter itself in place, so a future third caller could
reintroduce the exact same fail-open with no warning.

## Fix

Make `content_ts` a **required** argument on both `check_comment_reviews`
(keyword-only, alongside `repo`) and `partition_formal_reviewers`
(positional), keeping the `datetime | None` type. `content_ts=None` remains
a legitimate *value* (`get_latest_content_commit` returns `None` when a PR
has no non-merge commits — "no content binding exists"), but **omitting**
the argument now raises `TypeError` at call time instead of silently
disabling staleness filtering.

Both production callers already pass it explicitly:
- `validate_pr_review.py`'s `check()` (`partition_formal_reviewers`, and
  both `check_comment_reviews` call sites)
- `pr_review_state.py`'s `compute_review_state` (same three call sites)

So this is behavior-preserving for every real merge. Only test call sites
needed an explicit `content_ts=None` — 9 call sites across
`test_validate_pr_review.py` (the shared `_run_with_fake_api` harness used
by ~30 tests, plus 6 standalone calls in the pagination and #981
incomplete-scan test classes).

## Tests

Added `ContentTsRequiredTests`:
- `test_check_comment_reviews_without_content_ts_raises_type_error` /
  `test_partition_formal_reviewers_without_content_ts_raises_type_error` —
  guard the requiredness itself; RED against the pre-fix defaulted
  signatures.
- `test_check_comment_reviews_still_accepts_explicit_none` /
  `test_partition_formal_reviewers_still_accepts_explicit_none` — `None`
  stays a valid value, only omission is barred.
- `test_check_comment_reviews_repo_is_keyword_only` — `repo` moved
  keyword-only alongside `content_ts` (required for Python's
  no-default-after-default ordering); a positional third argument must now
  TypeError.

Full local gate green: `pre-commit run --all-files` (commit stage) +
`pre-commit run --hook-stage pre-push --all-files` (mypy + full pytest
suite, 2726 passed) — both clean.

## Reviewers

- **Nadia Khoury** (merge-gate)
- **Nurul Hakim**

### DIFF
--- a/.claude/hooks/tests/test_validate_pr_review.py
+++ b/.claude/hooks/tests/test_validate_pr_review.py
@@ -118,6 +118,10 @@ class _CheckCommentReviewsHarness(unittest.TestCase):
                 _CheckCommentReviewsHarness.PR_NUMBER,
                 branch_author,
                 repo=repo,
+                # content_ts is a REQUIRED keyword-only arg (#1050); None is the
+                # explicit "no content binding" value these ~30 pre-#950 callers
+                # exercise (every verdict counted unconditionally).
+                content_ts=None,
             )
 
     @staticmethod
@@ -485,6 +489,55 @@ class LatestVerdictSupersedesTests(_CheckCommentReviewsHarness):
         )
         self.assertIn("nino kavtaradze", result.reviewers)
         self.assertEqual(len(result.stale_verdicts), 1)
+
+
+class ContentTsRequiredTests(unittest.TestCase):
+    """Issue #1050: `content_ts` must be a REQUIRED argument on both shared
+    gate helpers, not a defaulted one.
+
+    #1046 happened because a defaulted `content_ts` let a caller omit it and
+    silently get staleness-filtering OFF rather than an error. This class
+    proves the fix at the signature level: calling either function WITHOUT
+    `content_ts` must now raise `TypeError` at call time — the omission
+    becomes loud instead of a silent fail-open. These tests would RED against
+    the pre-#1050 signatures (which default `content_ts` to `None`).
+    """
+
+    def test_check_comment_reviews_without_content_ts_raises_type_error(self):
+        with self.assertRaises(TypeError):
+            hook.check_comment_reviews(451, "pham", repo="noorinalabs/x")  # missing content_ts
+
+    def test_partition_formal_reviewers_without_content_ts_raises_type_error(self):
+        with self.assertRaises(TypeError):
+            hook.partition_formal_reviewers([], "someone")  # missing content_ts
+
+    def test_check_comment_reviews_still_accepts_explicit_none(self):
+        """`content_ts=None` remains a legitimate VALUE — only omission is barred."""
+        with mock.patch.object(
+            hook.subprocess,
+            "run",
+            side_effect=lambda args, capture_output, text, timeout: mock.MagicMock(
+                returncode=0, stdout="[]"
+            ),
+        ):
+            result = hook.check_comment_reviews(451, "pham", repo="noorinalabs/x", content_ts=None)
+        self.assertEqual(result.undetermined, "")
+
+    def test_partition_formal_reviewers_still_accepts_explicit_none(self):
+        formal, stale = hook.partition_formal_reviewers(
+            [{"author": {"login": "reviewer-a"}, "state": "APPROVED"}],
+            "author-login",
+            None,
+        )
+        self.assertEqual(formal, {"reviewer-a"})
+        self.assertEqual(stale, [])
+
+    def test_check_comment_reviews_repo_is_keyword_only(self):
+        """`repo` moved keyword-only alongside `content_ts` (#1050) — a
+        positional third argument must now raise TypeError rather than
+        silently binding to `repo`."""
+        with self.assertRaises(TypeError):
+            hook.check_comment_reviews(451, "pham", "noorinalabs/x")  # repo positional
 
 
 class ExtractBranchAuthorLastnameTests(unittest.TestCase):
@@ -1014,7 +1067,9 @@ class CommentPaginationTests(_CheckCommentReviewsHarness):
             return result
 
         with mock.patch.object(hook.subprocess, "run", side_effect=fake_run):
-            hook.check_comment_reviews(self.PR_NUMBER, self.BRANCH_AUTHOR, repo=self.REPO)
+            hook.check_comment_reviews(
+                self.PR_NUMBER, self.BRANCH_AUTHOR, repo=self.REPO, content_ts=None
+            )
 
         gh_api_calls = [a for a in captured_args if a[0] == "gh" and a[1] == "api"]
         self.assertEqual(len(gh_api_calls), 1, "exactly one gh api call expected")
@@ -3109,7 +3164,7 @@ class IncompleteCommentScanFailsClosedTests(_NoContentBindingHarness):
     def test_clean_scan_leaves_undetermined_empty(self):
         """The negative match — a successful empty scan must NOT be flagged."""
         with mock.patch.object(hook.subprocess, "run", side_effect=self._fake_run()):
-            result = hook.check_comment_reviews(451, "pham", repo="noorinalabs/x")
+            result = hook.check_comment_reviews(451, "pham", repo="noorinalabs/x", content_ts=None)
         self.assertEqual(result.undetermined, "")
         self.assertEqual(result.reviewers, set())
 
@@ -3119,7 +3174,7 @@ class IncompleteCommentScanFailsClosedTests(_NoContentBindingHarness):
             "run",
             side_effect=self._fake_run(returncode=1, stdout="", stderr="HTTP 403: Forbidden"),
         ):
-            result = hook.check_comment_reviews(451, "pham", repo="noorinalabs/x")
+            result = hook.check_comment_reviews(451, "pham", repo="noorinalabs/x", content_ts=None)
         self.assertTrue(result.undetermined)
         self.assertIn("403", result.undetermined)
 
@@ -3127,19 +3182,19 @@ class IncompleteCommentScanFailsClosedTests(_NoContentBindingHarness):
         with mock.patch.object(
             hook.subprocess, "run", side_effect=hook.subprocess.TimeoutExpired("gh", 30)
         ):
-            result = hook.check_comment_reviews(451, "pham", repo="noorinalabs/x")
+            result = hook.check_comment_reviews(451, "pham", repo="noorinalabs/x", content_ts=None)
         self.assertIn("TimeoutExpired", result.undetermined)
 
     def test_unparseable_json_sets_undetermined(self):
         with mock.patch.object(
             hook.subprocess, "run", side_effect=self._fake_run(stdout="not json")
         ):
-            result = hook.check_comment_reviews(451, "pham", repo="noorinalabs/x")
+            result = hook.check_comment_reviews(451, "pham", repo="noorinalabs/x", content_ts=None)
         self.assertIn("JSONDecodeError", result.undetermined)
 
     def test_unresolvable_owner_repo_sets_undetermined(self):
         with mock.patch.object(hook, "_resolve_owner_repo", return_value=None):
-            result = hook.check_comment_reviews(451, "pham", repo=None)
+            result = hook.check_comment_reviews(451, "pham", repo=None, content_ts=None)
         self.assertIn("could not resolve the target repository", result.undetermined)
 
     def test_check_hard_blocks_on_an_incomplete_scan(self):
--- a/.claude/hooks/validate_pr_review.py
+++ b/.claude/hooks/validate_pr_review.py
@@ -890,8 +890,9 @@ def _name_lastname(full_name: str) -> str:
 def check_comment_reviews(
     pr_number: str | int,
     branch_author_lastname: str,
+    *,
     repo: str | None = None,
-    content_ts: datetime | None = None,
+    content_ts: datetime | None,
 ) -> CommentReviewResult:
     """Check PR comments for charter-format review comments from different authors.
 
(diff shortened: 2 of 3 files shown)


parametrization:
Reviewed at head `7e3795f`. Merge-gate review by Nadia Khoury.

I verified the fix against the shipped source at the head SHA, not the diff prose. AST inspection of `.claude/hooks/validate_pr_review.py` confirms `check_comment_reviews` now takes `content_ts` as a **required keyword-only** argument (with `repo` moved keyword-only, still defaulted), and `partition_formal_reviewers` takes `content_ts` as a **required positional** — so omitting it raises `TypeError` at call time exactly as #1050 intends, converting the #1046 silent fail-open into a loud error.

The new `ContentTsRequiredTests` are genuine guards, not inert assertions — they would go red against the pre-#1050 defaulted signatures, and `test_check_comment_reviews_repo_is_keyword_only` additionally pins the keyword-only move. Both real callers forward the argument: `check()` in the hook and the #707 driver in `.claude/lib/pr_review_state.py` (three call sites each), so making it required breaks no runtime path. `None` stays a legitimate value, and the two `still_accepts_explicit_none` tests confirm that. CI is fully green at head.

Approved.

---
Requestor: Nadia Khoury
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Review — #1075 (fixes #1050)

Reviewed at HEAD `7e3795f2` against base `deployments/phase-9/wave-27` (second reviewer; Nadia Khoury is merge-gate).

The change makes `content_ts` required on both shared gate helpers, closing the residual #1046 fail-open that #1049 left in place (the two production sites were fixed but the defaulted parameter itself survived, so a future third caller could re-introduce the silent staleness-OFF). This is precisely the "future third caller" hazard worth hard-guarding.

**The one thing that could have broken this — a now-required arg that a production caller omits → runtime `TypeError` in the PR-review gate itself — I checked at the call-site level and then executed.** All four real call sites forward `content_ts` explicitly:

- `validate_pr_review.py::check()` — `partition_formal_reviewers(reviews, author, content_ts)` and both `check_comment_reviews(..., content_ts=content_ts)` sites (normal-branch and wave-merge sentinel).
- `pr_review_state.py::compute_review_state` — the same three, via `gate.` These are docstring-only in the diff precisely because the calls already passed the arg.

So the change is behavior-preserving for every real merge. I then mutation-tested the guard against the HEAD module:

```
omission of content_ts (check_comment_reviews) -> TypeError  ✓
repo passed positionally (now keyword-only)     -> TypeError  ✓
omission of content_ts (partition_...)          -> TypeError  ✓
explicit content_ts=None                        -> accepted, returns {'reviewer-a'} []  ✓
```

`None` correctly remains a legitimate *value* ("no content binding" — `get_latest_content_commit` returns `None` on a PR with no non-merge commits); only *omission* is now barred. `repo` moving keyword-only is required by Python's no-default-after-required ordering and is covered by its own test; no caller passes `repo` positionally. The added `ContentTsRequiredTests` are genuine guards (RED against the pre-fix defaulted signatures), and the 9 test call-site updates all pass the explicit `content_ts=None`.

CI: all 14 checks green at head. Clean hardening PR. Approving.

---
Requestor: Nurul Hakim
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

parametrization:
Re-review at rebased head `6f0d4cb` (after #1081's `resolve_review_verdicts` refactor and #1071's latest-verdict work merged). My prior approval was against the old head, so this is a fresh merge-gate verdict.

I re-verified against the shipped source at the new SHA, not the rebase description. The #1050 requiredness survived the auto-merge intact: AST inspection of `.claude/hooks/validate_pr_review.py` confirms `check_comment_reviews` still takes `content_ts` as a **required keyword-only** argument (no `=None` default; `repo` keyword-only, defaulted) and `partition_formal_reviewers` still takes `content_ts` as a **required positional** — so omitting either still raises `TypeError`.

The #1081 refactor consolidated the pipeline into `resolve_review_verdicts`, and it threads `content_ts` (derived from `get_latest_content_commit`) to all three helper call sites — the `partition_formal_reviewers` call and both `check_comment_reviews` branches. `pr_review_state.py` no longer calls the helpers directly; it delegates via `gate.resolve_review_verdicts(pr_data, repo=repo)`, so the #1046 "driver kept its own copy that omitted content_ts" shape is now structurally impossible, and the required-arg guard remains meaningful defense for any future third caller.

The hand-resolved `test_validate_pr_review.py` conflict is purely additive — both #1071's `LatestVerdictSupersedesTests` and #1075's `ContentTsRequiredTests` are present, with the omission-raises-`TypeError` and repo-keyword-only guards intact. CI is 14/14 green, MERGEABLE/CLEAN at head.

Approved.

---
Requestor: Nadia Khoury
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Re-confirm — #1075 (fixes #1050) after rebase, head `6f0d4cb`

Re-reviewed after the rebase onto the wave branch (post #1081 + #1071). Second reviewer; Nadia Khoury is merge-gate. The one thing a rebase like this could break — a now-required `content_ts` meeting a *new* base-side caller that omits it — I checked directly rather than trusting the auto-merge.

**The call sites consolidated, and the new shared caller threads `content_ts` at every site.** #1081/#1048 collapsed `check()` and `pr_review_state.compute_review_state` onto a single shared `resolve_review_verdicts(pr_data, repo=repo)` entry point (now in base). I read it at this head: it computes `content_ts = latest_content[1] if latest_content else None` and passes it to all three required-arg calls —

- `partition_formal_reviewers(reviews, author, content_ts)`
- `check_comment_reviews(number, branch_author_lastname, repo=repo, content_ts=content_ts)`
- `check_comment_reviews(number, "", repo=repo, content_ts=content_ts)` (wave-merge sentinel)

`pr_review_state.py` no longer calls the helpers directly — it delegates to `resolve_review_verdicts` — so #1075's `pr_review_state.py` change is correctly docstring-only (+7/-0). The rebase auto-merged #1075's signature change with #1081's call-site consolidation non-overlappingly; there is no caller anywhere that omits `content_ts`, so the required change stays behavior-preserving for every real merge.

**Guard still real — mutation-tested at this head:**
```
omit content_ts (check_comment_reviews)   -> TypeError  ✓
repo passed positionally (keyword-only)    -> TypeError  ✓
omit content_ts (partition_formal_reviewers) -> TypeError ✓
explicit content_ts=None                   -> accepted, {'rev-a'} []  ✓
```
`None` remains a legitimate VALUE; only omission is barred — exactly the #1050 intent.

**Scope clean.** Net diff is exactly the original #1050 change (test +61/-6, `validate_pr_review.py` +18/-8, `pr_review_state.py` +7/-0); nothing from #1081/#1071 leaked in. CI 14/14 green at head, MERGEABLE/CLEAN — the Pytest job here runs `ContentTsRequiredTests` plus the ~30 harness tests through the consolidated pipeline, so green confirms the required arg does not break `resolve_review_verdicts`. Approving.

---
Requestor: Nurul Hakim
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Issue #1054, opened 2026-07-20
wave_{N}_meta_issue has no type guard: the string form silently defeats the kickoff meta-issue skip


parametrization:
Found during merge-gate review of #1052.

## The defect

`post_wave_kickoff_comment.py:516` decides the meta-issue skip with:

```python
meta = status.get(f"wave_{wave_num}_meta_issue")
if meta is not None and str(meta) == issue_number and repo == "noorinalabs-main":
```

`issue_number` is a bare string (`"983"`), so this only matches when the key holds an **int**. The key has been written both ways in `cross-repo-status.json`:

| wave | value | skip fires? |
|---|---|---|
| 23 | `919` (int) | yes |
| 24 | `"noorinalabs-main#928"` (str) | **no** |
| 25 | `"noorinalabs-main#980"` (str) | **no** |
| 26 | `983` (int, set by #1052) | yes |

## Mutation test (real hook, `meta_issue_skip.json` fixture)

```
INT    meta=347                      -> action="skip_meta_issue"
STRING meta="noorinalabs-main#347"   -> action="post"
```

The string form does not error — it falls through to `post`, so the meta-issue receives a per-issue kickoff comment that `/wave-kickoff` Step 8b is supposed to own exclusively. Silent, and consistent with the #586 class (a bookkeeping shape mismatch degrading to a no-op rather than a failure).

Waves 24 and 25 very likely got that spurious comment.

## Ask

A guard, not a data migration — the historical values are inert. Either:
1. normalise on read (accept `983`, `"983"`, and `"noorinalabs-main#983"`, comparing the trailing number), or
2. assert the type at write time in `upsert_status_keys.py` / the scope step, so the string form cannot be written again.

Option 1 is preferable: it fixes every already-written wave without a migration, and the hook keeps working whichever form a future scope run writes.

Fixtures currently only cover the int form, so no existing test trips this — add a string-form fixture alongside whichever fix lands.

parametrization:
Duplicate of #1053, filed 8 seconds earlier during the same PR #1052 review — we converged on the identical defect independently, which is itself a decent signal it is real.

Closing this one as the later of the two. Your unique evidence — the mutation test on the live hook showing int → `skip_meta_issue` vs string → `post`, and the absence of any fixture covering either form — has been ported into #1053, which also carries the full historical spread across waves 1-16. Nothing lost; #1053 is the survivor. Good catch.

PR #1076, opened 2026-07-22
fix(#1055): TechDebt trailer accepts bare issue numbers, surfaces unparseable values

### LINKED ISSUE #1055: TechDebt trailer with a bare issue number parses to zero refs — present-but-uncaptured is silent
## Summary

A verdict trailer whose tech-debt line carries a **bare issue number** (no `#`) passes the presence check but contributes **zero** issue numbers to `tech_debt_issue_numbers`. The reference is silently dropped — the reviewer is not flagged as missing the line, and the filed tech-debt is not recorded against the PR.

Observed live on PR #1052, 2026-07-20.

## Reproduction

Nino Kavtaradze's verdict on PR #1052 carried a bare number:

```
Requestor: Nino Kavtaradze
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: 1054
```

`pr_review_state.py` (which replays the gate) reported:

```
verdicts missing TechDebt line: none
TechDebt issue numbers: none
```

Both statements are individually true and jointly useless: the line was present, so nothing was flagged; the number did not parse, so nothing was captured. A reviewer who did exactly the right thing — filed the tech-debt issue *before* posting, per `charter/pull-requests/reviews.md` — is recorded identically to one who wrote `none`.

## Root cause

`.claude/hooks/validate_pr_review.py:897-906`:

```python
has_tech_debt = re.search(r"\*{0,2}TechDebt:\*{0,2}\s*(.+)", body)
if not has_tech_debt:
    result.reviews_missing_tech_debt.append(requestor)
else:
    td_value = has_tech_debt.group(1).strip().strip("*").strip()
    if td_value.lower() != "none":
        # Extract issue numbers (#15, #16, etc.)
        issue_nums = re.findall(r"#(\d+)", td_value)
        result.tech_debt_issue_numbers.extend(issue_nums)
```

`re.findall(r"#(\d+)", "1054")` returns `[]`. There is no branch for "the value is neither `none` nor a parseable reference," so the mismatch is unobservable — no warning, no counter, nothing in the BLOCK diagnostic.

The presence gate and the capture path have **different strictness**, which is the whole defect: presence accepts any non-empty string, capture accepts only `#`-prefixed digits, and the gap between them is silent.

## Blast radius

Not merge-blocking — the 2-reviewer and TechDebt-present gates behave correctly. The loss is **audit fidelity**:

- `tech_debt_issue_numbers` is what wrapup/retro read to reconcile "what debt did this wave surface." Bare-number references vanish from that reconciliation.
- It interacts badly with the `+20%` TD-intake accounting (`/wave-scope` § 8.5): debt that was filed and referenced can read as never-filed.
- The charter template shows `TechDebt: none` and prose says "list the issue numbers" without stating the `#` is load-bearing, so a reviewer writing `TechDebt: 1054` is following the documented instruction as written.

Same shape as #950, #1046, #1047, #1053: a check that quietly does not do what its name implies, while the surrounding run reports success.

## Proposed fix

1. **Accept bare numbers** — extend to `re.findall(r"#?(\d+)", td_value)`, so both `#1054` and `1054` capture. Comma/space separated lists already work under `findall`.
2. **Make the residual case observable.** If `td_value` is neither `none` nor yields at least one number, record it (e.g. a `tech_debt_unparseable` list surfaced in the render and the BLOCK diagnostic). Silent is the actual bug; a permissive regex alone just moves the boundary.
3. **Say it in the charter.** `charter/pull-requests/reviews.md` should show a concrete non-`none` example (`TechDebt: #1054, #1055`) rather than only the `none` case.
4. **Regression test, mutation-verified** — cover `#1054`, `1054`, `#1054, #1055`, `1054 and 1055`, `none`, and a junk value; assert the junk value is *reported*, not swallowed. The test must fail if the `#?` is reverted.

## Note

PR #1052's own trailer references #1054, which has since been closed as a duplicate of #1053. That is incidental to this bug but worth knowing if anyone traces the example back.

parametrization:
## Summary

A verdict trailer whose `TechDebt:` line carries a **bare issue number** (no `#`) passed the presence check but contributed **zero** issue numbers to `tech_debt_issue_numbers` — silently. Observed live on PR #1052 (`TechDebt: 1054`).

## Root cause

`.claude/hooks/validate_pr_review.py` used `re.findall(r"#(\d+)", td_value)` — the presence gate accepted any non-empty, non-"none" value, but the capture path required a literal `#`. The gap between the two strictness levels was unobservable.

## Fix

- Extend the regex to `#?(\d+)` — both `#1054` and a bare `1054` now capture (comma/space-separated lists already worked via `findall`).
- **Make the residual case observable**: a value that is neither `none` nor yields any number (free text like "filed later") is recorded on a new `CommentReviewResult.tech_debt_unparseable` list and surfaced as a non-blocking advisory `systemMessage` from `check()` — never merge-blocking, but no longer silently swallowed.
- Threaded `tech_debt_unparseable` through `pr_review_state.py` (`ReviewState`, `compute_review_state`, both text and JSON renders) so the standalone review-state CLI can't silently drop what the gate itself now surfaces.
- `charter/pull-requests/reviews.md` now shows a concrete non-`none` example (`TechDebt: #1054, #1055`) and states the `#` is optional but a value that parses to nothing is flagged.

## Tests

- `#1054`, `1054`, `1054, 1055`, `1054 and #1055`, `none`, and a junk value (`"filed later"`) — the junk value is asserted as *reported* in `tech_debt_unparseable`, not swallowed.
- Mutation-verification test pins that the old `#(\d+)` regex captures nothing from a bare number, proving the `#?` change is load-bearing.
- `pr_review_state.py` wiring covered directly (pass-through + advisory text rendering), and confirms `passes()` is unaffected by an unparseable value (audit-fidelity gap, not a merge-blocking condition).
- Full `.claude/lib/tests/` + `.claude/hooks/tests/` suite passes (2729 tests); `pre-commit run --all-files` + push-stage (mypy, pytest) all green.

Closes #1055

**Reviewers:** Lucas Ferreira (merge-gate) + Aino Virtanen


parametrization:
**Merge-gate review — PR #1076 (fixes #1055)**

Reviewed the diff at head `74a37ec` against `deployments/phase-9/wave-27`; CI is green across all 21 checks at that SHA, including the real pytest and mypy engines.

The core defect and its fix are exactly as characterized. The extraction regex widens from `#(\d+)` to `#?(\d+)`, so a bare issue list (the real-world value from Nino's own verdict on PR #1052 that had no `#`) is now captured instead of matching nothing. I reproduced the mutation directly: `re.findall(r"#(\d+)", "1054")` yields `[]` while `re.findall(r"#?(\d+)", "1054")` yields `["1054"]` — the pinned mutation test is genuinely sensitive, not incidental.

The more important half is that a value which is neither `none` nor a number no longer vanishes: it is recorded on the new `tech_debt_unparseable` field and surfaced as a non-blocking advisory. That guard is trippable by a fixture (`test_approved_with_junk_techdebt_value_is_reported_not_swallowed` asserts the exact `("Mateo Santos", "filed later")` pair), so it is not inert. The field is threaded through `pr_review_state` and rendered in the text report as well, so the driver cannot silently drop what the gate surfaces.

Merge-decision safety: none of the reviewer-set / threshold / missing-attestation blocking logic is touched — the advisory returns an allow-with-message and only fires after every blocking check has already passed. So this change cannot flip a merge decision in either direction; it is pure observability.

One non-blocking observation, deliberately not filed: because the widened regex captures any digit run, an out-of-spec free-text value that happens to contain a number (e.g. a stray "in 2 days") would capture that digit as a ref rather than landing in the unparseable bucket. That path only matters for values already outside the charter format, board sync is best-effort, and it cannot affect a merge verdict — marginal enough that I judged it not worth tracking rather than filing noise.

Approving.

---
Requestor: Lucas Ferreira
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: none

parametrization:

Reviewed at head `74a37ec`. All 21 checks are green at that SHA, base is `deployments/phase-9/wave-27`.

Ran both suites at head (validate_pr_review 220 passed, pr_review_state 27 passed) and mutation-tested the fix: reverting the `#?(\d+)` capture back to `#(\d+)` kills the three bare-number tests. The residual-case handling is the real value here — a present, non-`none`, unparseable value (free text) is now recorded and surfaced as a non-blocking advisory rather than vanishing, and the tests confirm it does not affect `passes()`. The `pr_review_state` field is a straight pass-through and its render path is covered. The advisory return sits just before the terminal no-op return, so no downstream logic is short-circuited. No must-fix items.

---
Requestor: Aino Virtanen
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Merge-gate re-review — PR #1076 (fixes #1055), rethreaded onto Cluster A**

Fresh review at head `02fe76f` (my prior approval at the pre-rebase head is superseded). Base is `deployments/phase-9/wave-27`, now carrying the completed Cluster A (#1081's shared entry point and #1071's latest-verdict logic). CI is 21/21 green and the PR is MERGEABLE/CLEAN. This was a genuine semantic rethread, not a rebase, so I re-verified it end to end — including running the suites and mutation-testing the advisory guard against the new shared structure.

What I confirmed:

1. The bare-issue-number fix in `check_comment_reviews` is unchanged and still correct — `#?(\d+)` captures both the canonical hash form and a bare list, and a value yielding zero digits is routed to the unparseable bucket rather than dropped. The pinned regex mutation test still proves the old pattern would have silently dropped a bare number. It now coexists cleanly with #1071's latest-verdict block in the same function (different regions, no interaction).

2. The observability signal is correctly re-homed onto the shared pipeline: a new `tech_debt_unparseable` field on the `ReviewVerdicts` dataclass, populated in `resolve_review_verdicts` from the comment result. This is exactly the integration I flagged originally — it replaces the read of the `comment_review_result` local that #1081 removed. I verified on a real import that the field lands in declared order among the dataclass's required fields and that `ReviewState` carries its defaulted counterpart without a field-ordering error.

3. The advisory in `check()` reads `verdicts.tech_debt_unparseable` and sits after every blocking check and the board-sync step, immediately before the terminal return, emitting an allow-with-message. It cannot flip a merge decision. I did not take that on faith — I mutation-tested it: forcing the advisory to block instead of allow makes the guard fail with "must NOT block the merge," and deleting the advisory entirely makes it fail with "must surface a message." Both mutations are caught, and restoring returns green, so the guard is sensitive rather than inert.

4. The driver path is intact: `compute_review_state` populates `ReviewState.tech_debt_unparseable` from the shared verdicts and `_render_text` shows it, so the tool cannot silently drop what the gate surfaces. Nothing from #1081's shared-boundary refactor, #1075's content-binding, or #1071's latest-verdict path regressed.

Verification engine: I ran both affected suites at this head under a throwaway detached worktree — 262 passed (3 subtests) across `test_validate_pr_review.py` and `test_pr_review_state.py`, including the content-staleness and delegation guards from the Cluster A changes. Read-only; no PR code edited.

Approving.

---
Requestor: Lucas Ferreira
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Merge-gate re-review (rethread onto Cluster A / #1081) — Approved**

Re-reviewed at the new head `02fe76f` (my prior approval was against the pre-rethread head and is stale). CI is 21/21 green at this SHA, MERGEABLE/CLEAN, base `deployments/phase-9/wave-27`. Ran both suites at head (validate_pr_review 233 passed, pr_review_state 29 passed).

Verified the semantic rethread, not just the rebase. The unparseable-value field now rides #1081's shared `ReviewVerdicts` dataclass, populated in `resolve_review_verdicts` from the comment-scan result, and `check()`'s advisory reads `verdicts.tech_debt_unparseable` (the old `check()`-local that #1081 removed is gone). Confirmed by structural inspection that the advisory sits after every blocking return (last block at line 1847) and immediately before the terminal no-op return, so a genuine block always takes precedence and the advisory can neither mask a block nor block a merge.

Mutation-tested the three load-bearing behaviors at head. Reverting `#?(\d+)` to `#(\d+)` kills the 3 bare-number tests (the fix is intact). Flipping the advisory return from allow to block kills exactly `test_check_surfaces_unparseable_tech_debt_as_nonblocking_advisory` (non-blocking is pinned). Zeroing the `resolve_review_verdicts` population leaves the hook suite green — its advisory test patches `resolve_review_verdicts` directly — but kills the `pr_review_state` driver test, so the end-to-end flow from comment scan through the shared structure into `ReviewState` is genuinely covered, not inert. The content-timestamp-required (#1075) and latest-verdict (#1071) paths are untouched and the full suite shows no regression. No must-fix items.

---
Requestor: Aino Virtanen
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: none

PR #1059, opened 2026-07-20
fix(#985): make -R/--repo authoritative over cwd in wave-label hooks

parametrization:
## Summary
The kickoff-comment and label-sync PostToolUse hooks resolved the target repo from the stdin `cwd` (captured at agent-spawn time) whenever the parsed repo was None. But `_parse_edit_segment` only recognized `--repo`/`--repo=`, never the `-R` short flag — so a child-repo `gh issue edit -R noorinalabs/<x>` run from a worktree fell through to a cwd-based misroute to the parent org repo (P9W25 retro pain-point #3, sibling of #981/#1057).

## Changes
- **`_shell_parse.repo_short_name_from_flag_value(value)`** — new single-point resolver for a `-R`/`--repo` flag value. Returns None on an unexpanded `$VAR` / command substitution / whitespace-bearing fragment — the #981 fail-closed caveat: an unresolvable token must never be coerced into a repo name (which would misroute the gh call).
- **`_wave_label_parse._parse_edit_segment`** — extracts the repo up front via the #1057-hardened `walk_flag_values({"--repo","-R"})`, covering all five surface forms: `--repo X`, `--repo=X`, `-R X`, `-R=X`, `-RX`. The flag is **authoritative**; cwd is the fallback **only** when no flag is present.
- **`WaveLabelChange.repo_flag_present`** (new field) gives consumers a tri-state that disambiguates the two `repo=None` cases:
  - flag **absent** → `repo=None, present=False` → ambient cwd resolution (#650)
  - flag **present but unresolvable** → `repo=None, present=True` → fail closed, skip (#981)
- **Both hooks** now skip with a logged `skip_unresolvable_repo` on a present-but-unresolvable flag instead of misrouting to cwd.

## Acceptance criteria
- [x] Parse `-R`/`--repo owner/name` from the command (all forms).
- [x] Authoritative over cwd — `-R noorinalabs/other-repo` routes to `other-repo` even when cwd resolves to the parent org repo (check-level test in both hook test files; the deploy row is the only row in scope so a cwd-based resolution would `skip_no_row`).
- [x] cwd fallback still works with no flag (existing #650 tests green).
- [x] `-R $VAR` unexpanded → safe: `skip_unresolvable_repo`, git_runner never consulted (no misroute).

## How the tests bite (verified by mutation)
- Reverting `walk_flag_values(rest, {"--repo", "-R"})` → `{"--repo"}` fails 15 tests across the four files.
[...]

PR #1077, opened 2026-07-22
fix(#1060): harden walk_flag_values to match real gh/cobra flag semantics

### LINKED ISSUE #1060: walk_flag_values diverges from real gh/cobra flag semantics: value-less flag eats next flag, no -- terminator, first-vs-last-wins on repeats
## Summary

Found while reviewing PR #1059 (#985: make `-R`/`--repo` authoritative over cwd in the wave-label hooks). `_shell_parse.walk_flag_values` — the shared tokenizer helper now used to extract the AUTHORITATIVE repo for fail-closed routing decisions (#985, #981, #1057) across `_wave_label_parse.py`, `_repo_flag_parse.py`, `block_squash_wave_merge.py`, and `validate_wave_label_evidence.py` — diverges from real `gh` (cobra/pflag) flag-parsing semantics in three ways. None of these are regressions specific to #1059 (the first is pre-existing for `--repo` since before #985; the other two are inherited from `walk_flag_values` itself, #1057), but #1059 raises the stakes by making the extracted value authoritative for fail-closed/fail-open routing decisions, so the gap is worth closing.

## Findings (reproduced against noorinalabs-main@f78fca9, PR #1059's head)

1. **A value-less `-R`/`--repo` immediately followed by another recognized flag token silently consumes that flag as its "value."** Real `gh` would error (`flag needs an argument: 'R'`); this hook doesn't check the underlying command's exit status, so it can still act on the bogus parse.

   ```python
   >>> import _wave_label_parse as w
   >>> w.parse_wave_label_change(
   ...   'gh issue edit 42 -R --add-label "Aino_Virtanen" --add-label "p3-wave-9"'
   ... )
   WaveLabelChange(repo='--add-label', issue_number='42', add_label='p3-wave-9',
                    remove_label=None, repo_flag_present=True)
   ```

   `repo_flag_present=True` with `repo='--add-label'` would be treated as an AUTHORITATIVE, resolved repo by both `post_wave_kickoff_comment.py` and `post_label_change_wave_field_sync.py` — no fail-closed path catches it because `_UNRESOLVABLE_REPO_VALUE_RE` (`[$\`\s]`) doesn't check whether the captured value is itself a recognized flag token.

2. **No `--` (POSIX end-of-options) support.** `gh`/cobra treats everything after a literal `--` as positional, not flags. `walk_flag_values` (and `tokenize`) don't special-case `--`, so a `--repo`/`-R` occurring after it is still parsed as an authoritative flag even though the real `gh` invocation would not have used it:

   ```python
   >>> w.parse_wave_label_change(
   ...   'gh issue edit 42 --add-label "wave-26" -- --repo noorinalabs/repo-c'
   ... )
   WaveLabelChange(repo='repo-c', ..., repo_flag_present=True)
   ```

3. **Repeated `-R`/`--repo`: first occurrence wins everywhere in this codebase** (`walk_flag_values` returns values in encounter order; every caller — `_repo_flag_parse.extract_repo`, `_shell_parse.first_flag_value`, and now `_wave_label_parse._parse_edit_segment` — takes `values[0]`). Real `gh` (pflag `StringVarP`) uses **last-flag-wins** for a repeated string flag. A duplicated `-R` (e.g. from careless command construction) would route the hook to a different repo than the one `gh` actually operated on.

## Suggested fix

- Harden `repo_short_name_from_flag_value` (or `walk_flag_values` itself) to reject a captured value that itself looks like a flag (starts with `-`), closing (1).
- Teach `tokenize`/`walk_flag_values` to stop scanning at a bare `--` token, closing (2).
- Either switch the repo-resolving callers to `values[-1]` (last-wins, matching real `gh`), or explicitly document why first-wins is the deliberate choice — right now it reads as an accident of `values[0]` being the obvious index, not a documented decision. Closing (3) either way removes the ambiguity for the next person extending `walk_flag_values`.
- Add regression tests for all three shapes to `.claude/hooks/tests/test_shell_parse.py`.

## Blast radius

Requires an already-malformed or adversarially-constructed `gh` command shape (a value-less `-R` glued to the next flag, an explicit `--` terminator, or a duplicated `-R`) — none of which the org's own orchestrator-generated commands produce today. Not merge-blocking for #1059. Filed per charter `pull-requests/reviews.md` TechDebt-before-verdict requirement.

parametrization:
## Summary

`walk_flag_values` diverged from real `gh` (cobra/pflag) flag parsing in
three ways, found during #1059 review — which raised the stakes by making
the extracted `--repo`/`-R` value **authoritative** for fail-closed/fail-open
routing decisions across `_wave_label_parse.py`, `_repo_flag_parse.py`,
`block_squash_wave_merge.py`, and `validate_wave_label_evidence.py`.

Closes #1060.

## The three findings and fixes

1. **Value-less flag ate the next flag.** `-R --add-label X` silently
   resolved `repo='--add-label'` (real gh errors: "flag needs an argument").
   `walk_flag_values` now only consumes the next token as a flag's value
   when that token does NOT itself look like a flag (`-`-prefixed,
   length > 1 — a bare `-` stdin sentinel is exempted). A flag-shaped
   successor is left unconsumed and still scanned as its own flag.
2. **No `--` (POSIX end-of-options) support.** A `--repo`/`-R` after a
   literal `--` was still parsed as authoritative. The scan now stops
   entirely at a bare `--` token.
3. **Repeated `-R`/`--repo`: first-occurrence-wins everywhere**, but real
   gh's single-value pflag `StringVarP` is last-flag-wins. Scope: the three
   DIRECT `walk_flag_values` callers that resolve `--repo`/`-R` for
   authoritative routing now take `values[-1]`:
   - `_repo_flag_parse.extract_repo` (both the tokenizer path and the
     regex fallback)
   - `_wave_label_parse._parse_edit_segment`
   - `block_squash_wave_merge._iter_squash_merges`

   `walk_flag_values` itself is unchanged (still returns every value in
   source order) — other callers legitimately accumulate a *repeatable*
   flag (e.g. `--add-label`, iterated in full by `validate_labels.py` /
   `validate_wave_label_evidence.py`). `first_flag_value` (used only for
   `--base`/`--head` in `validate_branch_freshness.py`, never `--repo`) was
   deliberately left first-wins, matching its own pinned test
   (`test_returns_first_value`) — documented in its docstring per the
   issue's "or explicitly document why first-wins is the deliberate choice."

## Tests, mutation-verified

For each of the three findings: reverted the corresponding fix, confirmed
the new test(s) fail, restored the fix, confirmed green. New coverage:

- `test_shell_parse.py::WalkFlagValuesTests` — 7 new cases (value-less flag
  vs. long/short flag successor, flag-shaped-but-unwanted successor still
  rejected, bare `-` still accepted as a value, `--` terminator with and
  without trailing flags, repeated-flag source-order confirmation).
- `test_repo_flag_parse.py` — `RepeatedRepoFlagTests` (tokenizer + regex
  fallback last-wins parity), `DoubleDashTerminatorTests`,
  `ValueLessFlagTests`.
- `test__wave_label_parse.py::RepoFlagResolution` — 3 new cases (last-wins,
  value-less-flag-no-misroute, `--` terminator).
- `test_block_squash_wave_merge.py::RepeatedRepoFlagLastWins` — confirms the
  repo fed to `_resolve_base_ref` (and echoed in the block reason) is the
  last occurrence.

## Test plan

- [x] `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests .claude/lib/tests -q` — 2738 passed, 97 subtests passed
- [x] `pre-commit run --all-files` — all commit-stage hooks green
- [x] `git push` — all pre-push hooks green (mypy, pytest, pytest-skills, memory-budget, headcount-budget, doc-freshness, office-drift, mermaid-render)
- [x] Mutation check per finding: reverted each fix individually, confirmed the corresponding new test(s) failed; restored, confirmed green

## Reviewers

Nino Kavtaradze (merge-gate) + Aino Virtanen

### DIFF
--- a/.claude/hooks/_repo_flag_parse.py
+++ b/.claude/hooks/_repo_flag_parse.py
@@ -82,10 +82,18 @@ _REPO_FALLBACK_RE = re.compile(r"(?:^|\s)(?:(?:--repo|-R)(?:=|\s+)|-R)(\S+)")
 
 
 def extract_repo(command: str) -> str | None:
-    """Extract the `--repo` / `-R` `OWNER/NAME` value, or None if absent."""
+    """Extract the `--repo` / `-R` `OWNER/NAME` value, or None if absent.
+
+    Last-occurrence wins (main#1060): real `gh`'s `--repo`/`-R` is a
+    single-value pflag `StringVarP`, so a REPEATED flag resolves to its
+    LAST occurrence, not its first — `gh issue edit 1 -R a/b -R c/d` acts on
+    `c/d`. `walk_flag_values` itself stays first-to-last source order (it
+    has other callers that legitimately accumulate a repeatable flag, e.g.
+    `--add-label`); this is the caller-side choice for a single-value flag.
+    """
     tokens = tokenize(command)
     if tokens is None:
-        match = _REPO_FALLBACK_RE.search(command)
-        return match.group(1) if match else None
+        matches = list(_REPO_FALLBACK_RE.finditer(command))
+        return matches[-1].group(1) if matches else None
     values = walk_flag_values(tokens, _REPO_FLAGS)
-    return values[0] if values else None
+    return values[-1] if values else None
--- a/.claude/hooks/_shell_parse.py
+++ b/.claude/hooks/_shell_parse.py
@@ -92,6 +92,24 @@ Public API
         arrive as a SINGLE shlex token, never preceded by a flag from
         `wanted`.
 
+        gh/cobra semantics hardening (main#1060):
+          - A value-less flag immediately followed by a flag-shaped token
+            (starts with `-`) does NOT consume that token as its value —
+            real gh/cobra errors ("flag needs an argument") rather than
+            silently swallowing the next flag. The flag-shaped token is
+            left for the scan to process in its own right.
+          - A literal `--` token (POSIX end-of-options) stops the scan
+            entirely — everything after it is positional, never a flag,
+            matching real `gh`/cobra behavior.
+          - Repeated occurrences of the same flag are returned in source
+            order (first-to-last); this helper does NOT pick a winner.
+            Callers that need real gh's last-flag-wins semantics for a
+            single-value `StringVarP` flag (e.g. `--repo`/`-R`) must take
+            `values[-1]`; callers accumulating a repeatable flag (e.g.
+            `--add-label`) iterate all of `values` — see
+            `_repo_flag_parse.extract_repo` for the former and
+            `validate_labels.extract_labels` for the latter.
+
     first_flag_value(command, wanted, *, regex_fallback=True) -> str | None
         Convenience wrapper: tokenizes `command` via `tokenize()` and
         returns the first value from `walk_flag_values()`. If tokenize
@@ -538,6 +556,17 @@ def is_gh_subcommand(tokens: list[str], *verbs: str) -> bool:
     return False
 
 
+def _looks_like_flag(tok: str) -> bool:
+    """Return True if `tok` has the surface shape of a flag (`-x` / `--xxx`).
+
+    A bare `-` (the conventional "read from stdin" positional sentinel used
+    by many CLIs, including some `gh`/`git` subcommands) is deliberately
+    NOT treated as a flag — only genuine `-`-prefixed multi-character
+    tokens are (main#1060).
+    """
+    return len(tok) > 1 and tok[0] == "-"
+
+
 def walk_flag_values(tokens: list[str], wanted: set[str]) -> list[str]:
     """Return values for `wanted` flag names, only when they appear as flags.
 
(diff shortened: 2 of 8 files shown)


parametrization:

Reviewed at head `8d4ca5f`. All 14 checks are green at that SHA, base is `deployments/phase-9/wave-27`.

Ran the four affected suites at head (195 passed) and mutation-tested all three hardening claims. Removing the `_looks_like_flag` guard so a value-less flag eats its flag-shaped successor kills 5 tests (including the #1059 misroute reproducer). Removing the `--` end-of-options break kills 3 terminator tests. Reverting the single-value callers from `values[-1]` to `values[0]` kills the 4 last-flag-wins tests. The split is correct: `walk_flag_values` stays first-to-last source order for genuinely repeatable flags (`--add-label`), and only the single-value `--repo`/`-R` routing callers take the last occurrence — matching real gh/cobra `StringVarP` semantics. `first_flag_value`'s first-wins contract is deliberately and documentedly left alone. No must-fix items.

---
Requestor: Aino Virtanen
Requestee: Lucas Ferreira
RequestOrReplied: Approved
TechDebt: none

parametrization:
Merge-gate review at head `8d4ca5f` (base `deployments/phase-9/wave-27`). Diff read through `gh` at the head SHA; no local checkout.

Substance — hardens `walk_flag_values` to real gh/cobra flag semantics: a value-less flag no longer swallows a following flag-shaped token (so `-R --add-label X` no longer misroutes `repo='--add-label'`, the #1059 hazard), a literal `--` end-of-options terminator stops the scan, and a bare `-` stdin sentinel is still accepted as a value. The helper itself stays source-order and picks no winner; the last-flag-wins choice for the single-value `--repo`/`-R` pflag is applied caller-side at the three authoritative resolvers (`extract_repo`, `_wave_label_parse._parse_edit_segment`, `block_squash_wave_merge._iter_squash_merges`).

Checked scope completeness — grepped every `walk_flag_values` caller. The accumulating callers (`validate_labels`, `smart_grep_ontology`, `_collect_labels`) correctly stay first-to-last; the only other `--repo` consumer, `validate_wave_label_evidence`, resolves through the updated `extract_repo`, and its `_first_value` serves only `--body`/`--body-file`. No authoritative `--repo` resolver was missed. `first_flag_value`'s first-wins contract is deliberately left alone (its only callers are the advisory branch-freshness gate).

Mutation-tested the guard — ran the four affected suites at head (195 green), then reverted `walk_flag_values` to the pre-#1060 consume-next / no-terminator behavior; the double-dash and value-less-flag guards go red. Live, not inert.

CI is green at the head SHA (14 checks) and the branch is mergeable.

Approving.

---
Requestor: Nino Kavtaradze
Requestee: Lucas Ferreira
RequestOrReplied: Approved
TechDebt: none

Issue #1042, opened 2026-07-20
ontology-rebuild: checksums.json ensure_ascii=False enforcement is a SKILL.md instruction, not shared code


parametrization:
## Summary

PR #1040 (closes #1038) fixed `ensure_ascii=True` churn in `checksums.json` for the one *programmatic, committed* writer — `.claude/hooks/ontology_tracker.py`. The other writer, the `/ontology-rebuild` resolver, is agent-driven (a `SKILL.md` prose instruction, not a code module), so the PR's fix there is documentation only:

> `.claude/skills/ontology-rebuild/SKILL.md` § 4: "when rewriting `checksums.json` programmatically, use `json.dump(data, f, indent=2, ensure_ascii=False)` and a trailing newline"

Nothing enforces that the agent executing the skill actually follows this instruction. If a future resolver run (or an edit to the skill) drops the `ensure_ascii=False` kwarg, the flip-flop diff noise this PR fixed reappears via the resolver instead of the hook — the same class of bug, moved rather than closed.

## Why not blocking for #1040

- The fix correctly covers the one enforced-by-code writer.
- All other `checksums.json` touchpoints were confirmed non-writers (read/skip-list only) or out of git scope, per the PR body's writer audit.
- This gap is pre-existing to #1040, not introduced by it — #1040 just brought it into relief by fixing the sibling writer.

## Suggested fix

Extract a small shared helper (e.g. `.claude/lib/checksums_io.py::write_checksums(path, data)`) that both `ontology_tracker.py` and the `/ontology-rebuild` resolver call, encapsulating `json.dump(data, f, indent=2, ensure_ascii=False)` + trailing newline + atomic tmp-file replace. This removes the need for the resolver to remember a serialization convention at all, closing the class rather than documenting around it.

## Context

Raised as an open question in PR #1040's own body ("If a reviewer prefers hard enforcement, the resolver's checksum write should move into a shared `.claude/lib/` helper — I'd support that but it is beyond this bugfix's scope."). Filed per reviewer request during merge-gate review of #1040.

parametrization:
Requestor: Fatima Okonkwo
Requestee: Weronika Zielinska
RequestOrReplied: Request

**Wave 27 Kickoff — Phase 9**

This issue is assigned to you for p9-wave-27.
- Peer reviewer: Nurul Hakim
- Secondary reviewer: Nino Kavtaradze
- Branch from: `deployments/phase-9/wave-27`
[...]

PR #1080, opened 2026-07-22
ci(#1038 follow-up): assert ontology/checksums.json holds no \u escapes (#1044)

### LINKED ISSUE #1038: ontology_tracker: gitignored session_handoff.md is tracked — every /handoff dirties committed checksums.json

### LINKED ISSUE #1044: ci(#1038 follow-up): assert ontology/checksums.json holds no \\u escapes, so the ensure_ascii convention is enforced not just documented
Found during the merge-gate review of PR #1040 (head `7428f25537162e6b557c7ca0b7de61c5aab616a9`). Non-blocking.

PR #1040 fixed the escaped/literal UTF-8 flip-flop on the committed `ontology/checksums.json` in two halves:

- **Enforced:** `.claude/hooks/ontology_tracker.py` now writes `json.dump(..., ensure_ascii=False)`. Covered by two tests that both fail when the flag is reverted.
- **Documented only:** `/ontology-rebuild`'s resolver is agent-driven, so the same requirement lands as a `SKILL.md` instruction. There is no code path to attach a test to.

That asymmetry is acceptable for the fix (the flip-flop is diff noise, not a blind gate, and the one automated writer now preserves the literal form), but the documented half has no failure mode other than a reviewer noticing the diff.

Cheap deterministic close: a CI assertion that the committed file contains no `\\u` escape sequences. At head the file has 0 escapes and a top-level `description` containing literal `—` and `×`, so the invariant holds today and the check would start green.

```python
raw = Path('ontology/checksums.json').read_text(encoding='utf-8')
assert '\\u' not in raw, 'checksums.json re-escaped — writer needs ensure_ascii=False'
```

Wire it into the existing `Config — YAML/JSON syntax` job (`CI — Docs & Config`) and mirror it in `.pre-commit-config.yaml` per the full local/CI parity rule.

Note: the untracked local scratch script `.claude/scratch/wave-merge-bodies/ontology-resolve.py` writes `json.dump(data, f, indent=2)` without the flag — untracked, so out of scope, but it is exactly the writer this check would catch.

Related: #1038, #1040.

parametrization:
## Summary

PR #1040 (closes #1038) fixed the escaped/literal UTF-8 flip-flop on the committed `ontology/checksums.json` in two halves:

- **Enforced:** `.claude/hooks/ontology_tracker.py` writes with `json.dump(..., ensure_ascii=False)`, covered by two tests that fail if the flag is reverted.
- **Documented only:** the `/ontology-rebuild` resolver is agent-driven, so the same requirement lands as a `SKILL.md` instruction. #1042 (companion PR in this wave) gave it a real `checksums_io.py` CLI to call instead of hand-rolling the write — but a not-yet-migrated caller, or a stray scratch script, could still write the file with the `ensure_ascii=True` default. There is no code path for that half to attach a test to; the only failure mode is a reviewer noticing the diff.

This adds a cheap, deterministic close on the **committed artifact itself**, independent of which writer touched it.

## What this adds

- `.claude/lib/check_checksums_ascii.py` — asserts `ontology/checksums.json` contains no `\u` escape sequence. `\u` never legitimately appears anywhere in this file's schema (SHA-256 hashes, ISO-8601 timestamps, POSIX paths), so its presence is unambiguous evidence some writer used `ensure_ascii=True`.
- Wired into the CI **`Config — YAML/JSON syntax`** job in `.github/workflows/docs.yml` as an additional step (per the issue's own suggestion).
- Mirrored as a `checksums-ascii` pre-commit hook (full local⇄CI parity — org-wide mandatory rule).
- Classified `checksums-ascii` in `.claude/lib/pre_commit_ci_sync.py`'s `_KIND_PATTERNS` so the sync-drift gate actively **demands** the mirror going forward (matching the `dockerfile-base-pin`/`fixture-realism`/`skill-graphql-pagination` convention) — added the standard 4-test kind-classification class plus updated the hardcoded `_EXPECTED_KINDS` snapshot and `_OLD_KIND_PATTERNS` parity fixture so the existing snapshot/parity tests stay accurate.

At head, `ontology/checksums.json` has 0 escape sequences, so the gate starts green (verified locally with the real file).

## Tests

13 new tests in `.claude/lib/tests/test_check_checksums_ascii.py` (clean/escaped text, missing-file-is-not-a-violation, escape-count reporting, path-in-message, CLI exit codes) plus 4 new kind-classification tests in `.claude/lib/tests/test_pre_commit_ci_sync.py` (mirrors the existing per-kind pattern). Full local gate green: ruff, mypy, actionlint, the new `checksums-ascii` hook itself, and the complete pytest suite (pre-push).

## Reviewers

Merge-gate: **Nino Kavtaradze**. Second reviewer: **Lucas Ferreira**.

Closes #1044


parametrization:
**Merge-gate review — PR #1080 (fixes #1044)**

Reviewed the diff at head `32fbdd7` against `deployments/phase-9/wave-27`; CI is green at that SHA, including the `Config — YAML/JSON syntax` job that now runs the new step and the sync-drift gate.

The new `check_checksums_ascii.py` is a sound deterministic tripwire. It scans the committed artifact itself for a literal backslash-u pair, which is exactly the signature a writer using `ensure_ascii=True` leaves when it re-escapes the `description` field's literal UTF-8 (em-dash, multiplication sign). Watching the committed file rather than any one writer is the right call — it catches a re-escape regardless of whether the tracker hook, the agent-driven resolver, or a stray manual edit caused it, which is precisely the half of #1038 that was documentation-only before.

The guard is trippable, not inert: `test_escaped_utf8_is_flagged` feeds the exact re-escaped shape of the real file's own description and asserts a violation, and the on-disk and CLI exit-code tests confirm 0 for clean / 1 for escaped / 0 for missing. The missing-file-is-clean behavior is the correct distinction (a missing artifact is a different failure mode than a re-escaped one).

Critically for the org full-parity rule: this new CI gate is mirrored. The change adds the pre-commit hook, the CI step, AND the `checksums-ascii` classification to the sync-drift gate, with tests proving CI-without-pre-commit is flagged as harmful drift and the mirrored pair is clean. The sync-drift CI job passing at head confirms the mirror actually holds rather than merely being asserted. The single-writer routing advice in the failure message points at `checksums_io.py`'s `write_checksums`, consistent with #1042.

Approving.

---
Requestor: Lucas Ferreira
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: none

parametrization:
Merge-gate review at head `32fbdd7` (base `deployments/phase-9/wave-27`). Diff read through `gh` at the head SHA; no local checkout.

Substance — adds `check_checksums_ascii.py`, a deterministic tripwire asserting the committed `ontology/checksums.json` holds no `\u` escape sequence (evidence some writer used the `ensure_ascii=True` default and re-escaped the literal UTF-8 in the top-level `description`). It is wired into the `Config — YAML/JSON syntax` CI job in `docs.yml`, mirrored as a commit-stage pre-commit hook, and classified as the `checksums-ascii` kind in `pre_commit_ci_sync.py` so an un-mirrored CI gate turns the sync-drift gate red. This closes the half of #1038/#1042 that a committed-file gate can watch without depending on which writer touched the file.

Executed the gate against a real artifact rather than trusting the tests — ran the script on a hand-built escaped sample (exit 1, correct diagnostic) and a clean literal-UTF-8 sample (exit 0). Then ran the new `test_check_checksums_ascii.py` plus the `ChecksumsAsciiKind` sync-drift cases (all green). The substring scan is a documented, deliberate simplification (a POSIX path with a literal backslash-u could in principle false-positive, but such paths do not occur in this git-tracked schema); acceptable for a cheap tripwire.

CI is green at the head SHA (21 checks) and the branch is mergeable. The `cross-repo-status.json` delta is the shared wave-27 kickoff base-divergence artifact, not authored churn.

Approving.

---
Requestor: Nino Kavtaradze
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: none

Issue #228, opened 2026-04-30
tech-debt: validate_pr_review hook doesn't honor charter's Single-Reviewer Exception (hook stricter than charter)


parametrization:
## Hook ↔ charter alignment gap

Charter `pull-requests.md § Single-Reviewer Exception` (ratified W8 per `cross-repo-status.json` `wave_8_work.process_changes_ratified_2026-04-17`) permits a 1-of-2 reviewer pass for mechanical, self-validating, low-blast-radius PRs. The `validate_pr_review` hook does not honor this exception — it strict-blocks any merge with `< 2` Approved comments.

## Live trip 2026-04-30

While merging user-service#88 (`fix(ci): trigger ci.yml on deployments/** branches (closes #81)` — 3+/1- mechanical YAML trigger expansion, self-validating via the PR's own GREEN `check` run, single Tech Lead approval from Anya Kowalczyk), `gh pr merge 88` was blocked:

```
PreToolUse:Bash hook blocking error from command: "python3 .../.claude/hooks/dispatcher.py":
BLOCKED: PR #88 has 1/2 required peer reviews.
At least TWO reviews from distinct non-authors are required before merge.
Charter § Pull Requests requires two comment-based peer reviews for all merges.
Use `gh pr comment <PR#> --body '...'` with charter format:
  Requestor: <branch author>  Requestee: <reviewer>  RequestOrReplied: Approved  TechDebt: none | #issue, ...
Pass `--admin` for emergency overrides only.
```

Workaround applied today: spawn a second reviewer (Idris Yusuf, security lens) — adds review latency to mechanical PRs.

## Why this matters

Per memory `feedback_enforcement_hierarchy.md`: "Prefer hook > skill > charter, in that order. Charter rules without enforcement decay." The Single-Reviewer Exception is a charter rule with **no hook enforcement** — meaning the rule decays in the opposite direction here: the hook is **stricter** than charter, forcing 2 reviewers on PRs that the charter explicitly permits with 1.

In practice this means:
- All "mechanical / self-validating / low-risk" PRs require 2 reviewers despite charter allowance.
- Reviewers spend cycles on rubber-stamp passes for genuinely trivial changes.
- Charter exception rule effectively dead-letter.

## Suggested fix

Two options:

[...]

PR #1081, opened 2026-07-22
refactor(#1048): share one verdict-set entry point between Hook 4 and pr_review_state

### LINKED ISSUE #1048: Share one verdict-set entry point between Hook 4 and pr_review_state (follow-up to #1046 point 5)

parametrization:
## Summary

Closes #1048. Follow-up to #1046 point 5.

#1046 delivered the narrow half of this: `partition_formal_reviewers`, shared
by `check()` and `pr_review_state.py`. This closes the rest — the
comment/roster/threshold pipeline was still re-assembled **independently**
by `check()` and `compute_review_state`, each separately calling
`get_latest_content_commit`, `check_comment_reviews` (twice — feature-branch
vs wave-merge), `_load_roster_names`, and `is_single_reviewer_exception`, then
unioning formal + comment reviewers and applying the threshold. Any future
parameter added to that pipeline had to be threaded through **two** call
sites by hand — the exact #1046 defect class, just not yet triggered again.

## Fix

Extracted the shared pipeline into `resolve_review_verdicts(pr_data, repo) ->
ReviewVerdicts` in `validate_pr_review.py`:

1. `get_latest_content_commit` → T_content (`content_ts`/`content_sha`, #950)
2. `partition_formal_reviewers` + `check_comment_reviews`, both bound to
   T_content, on the feature-branch or wave-merge head-ref path
3. `_load_roster_names` + non-roster Requestor filtering (#498)
4. Union formal + roster-filtered comment reviewers, plus the wave-bootstrap
   single-reviewer exception (#228)

`check()` and `pr_review_state.compute_review_state` now both call this ONE
function and use its output directly — **neither re-derives the verdict
set**. Message/report rendering stays with each caller: `check()` keeps its
block-message construction; the driver keeps `ReviewState.passes()` and its
text/JSON report.

Three exceptions surface the pipeline's determinate-failure cases so each
caller renders its own message: `CommitFetchError` and `RosterResolutionError`
propagate verbatim (unchanged); a new `CommentScanUndeterminedError` carries
the #981 incomplete-scan reason that previously lived on
`CommentReviewResult.undetermined`.

## Behavior preservation

All 214 pre-existing `test_validate_pr_review.py` tests and all 27
pre-existing `test_pr_review_state.py` tests pass **unchanged** — this is a
pure extraction, not a behavior change.

## Mutation verification (per the issue's acceptance criterion)

Manually confirmed by temporarily stripping `content_ts=content_ts` from the
feature-branch `check_comment_reviews` call inside `resolve_review_verdicts`
and rerunning the suite: **4 tests in `ContentStalenessTests` went red
immediately**, because that class drives `compute_review_state` through the
REAL `resolve_review_verdicts` over a faked `gh api` boundary — exactly the
#1046 mutation, now caught at the new shared boundary.

Added as permanent guards:
- `ResolveReviewVerdictsSharedBoundaryTests` (hook tests) — a direct
  stale-verdict kill-shot on `resolve_review_verdicts` itself, plus
  delegation tests (`test_check_delegates_to_resolve_review_verdicts_and_trusts_it`,
  `test_check_translates_stale_verdict_error_into_a_block`) that patch the
  function and assert `check()` calls it exactly once and trusts its output.
- `test_delegates_to_the_shared_resolve_review_verdicts_entry_point` /
  `test_comment_scan_undetermined_raises_review_state_error` (driver tests) —
  the same delegation guard on `compute_review_state`.

## Local gate

`pre-commit run --all-files` (commit stage) + `pre-commit run --hook-stage
pre-push --all-files` (mypy + full pytest suite, 2726 passed) — both clean.

## Acceptance (from #1048)

- [x] Single shared entry point; neither `check()` nor `compute_review_state`
      re-derives the verdict set.
- [x] Behaviour-preserving: the existing Hook 4 and driver suites pass
      unchanged.
- [x] Mutation-verified the same way #1046 was — dropping a parameter at the
      shared boundary fails a test.

## Reviewers

- **Weronika Zielinska** (merge-gate)
- **Lucas Ferreira**

### DIFF
--- a/.claude/hooks/tests/test_validate_pr_review.py
+++ b/.claude/hooks/tests/test_validate_pr_review.py
@@ -2098,6 +2098,161 @@ class VerdictFixtureSanityTests(unittest.TestCase):
         self.assertEqual(result.stale_verdicts, [])
 
 
+class ResolveReviewVerdictsSharedBoundaryTests(unittest.TestCase):
+    """#1048: `resolve_review_verdicts` is the ONE shared entry point `check()`
+    and `pr_review_state.compute_review_state` both call — neither re-derives
+    the content-binding / comment-scan / roster-filter / union pipeline with
+    its own argument list, which is the #1046 defect class.
+
+    Mutation-sensitivity was manually confirmed while writing this refactor:
+    temporarily stripping `content_ts=content_ts` from the feature-branch
+    `check_comment_reviews` call inside `resolve_review_verdicts` turned 4
+    tests in `.claude/lib/tests/test_pr_review_state.py::ContentStalenessTests`
+    red immediately, because that class drives `compute_review_state` through
+    the REAL `resolve_review_verdicts` over a faked `gh api` boundary. The
+    test below is the equivalent DIRECT guard at the shared function itself.
+    """
+
+    REPO = "noorinalabs/noorinalabs-main"
+
+    @staticmethod
+    def _pr_data(head_ref="L.Ferreira/1040-x", author="parametrization", reviews=(), labels=()):
+        return {
+            "author": author,
+            "number": 1040,
+            "reviews": list(reviews),
+            "headRefName": head_ref,
+            "labels": list(labels),
+        }
+
+    def test_stale_comment_verdict_is_excluded_from_the_reviewer_set(self):
+        """Direct kill-shot on the shared boundary (#950 x #1048): a stale
+        Approved must not count toward `distinct_reviewers`, and must be
+        recorded on `stale_verdicts_comment` so a caller (either check()'s
+        block message or the driver's report) can name it. This would RED if
+        `resolve_review_verdicts` ever stopped forwarding `content_ts` to
+        `check_comment_reviews`.
+        """
+        stale_comment = {
+            "body": (
+                "Requestor: Lucas Ferreira\nRequestee: Someone Else\n"
+                "RequestOrReplied: Approved\nTechDebt: none"
+            ),
+            "created_at": "2026-01-01T00:00:00Z",
+        }
+
+        def fake_run(args, capture_output, text, timeout):
+            result = mock.MagicMock()
+            result.returncode = 0
+            if args[0] == "gh" and args[1:3] == ["repo", "view"]:
+                result.stdout = json.dumps({"owner": {"login": "noorinalabs"}, "name": "r"})
+            else:
+                result.stdout = json.dumps([stale_comment])
+            return result
+
+        with (
+            mock.patch.object(hook.subprocess, "run", side_effect=fake_run),
+            mock.patch.object(
+                hook,
+                "get_latest_content_commit",
+                return_value=("ac8bcfa", hook._parse_iso8601("2026-01-02T00:00:00Z")),
+            ),
+            mock.patch.object(hook, "_load_roster_names", return_value={"lucas ferreira"}),
+        ):
+            verdicts = hook.resolve_review_verdicts(self._pr_data(), repo=self.REPO)
+
+        self.assertEqual(verdicts.distinct_reviewers, set())
+        self.assertEqual(len(verdicts.stale_verdicts_comment), 1)
+        self.assertEqual(verdicts.stale_verdicts_comment[0].reviewer, "Lucas Ferreira")
+        self.assertEqual(verdicts.stale_verdicts_formal, [])
+        self.assertEqual(verdicts.stale_verdicts, verdicts.stale_verdicts_comment)
+
+    def test_check_delegates_to_resolve_review_verdicts_and_trusts_it(self):
+        """`check()` must call the shared entry point and use its output
+        DIRECTLY, rather than reassembling the pipeline inline — the concrete
+        guard for #1048's acceptance criterion ('neither check() nor
+        compute_review_state re-derives the verdict set'). A fake
+        `ReviewVerdicts` with 2 distinct current reviewers and no missing
+        TechDebt must ALLOW the merge without `check()` ever recomputing
+        `distinct_reviewers` itself.
+        """
+        input_data = {
+            "tool_name": "Bash",
+            "tool_input": {"command": "gh pr merge 1040 --repo noorinalabs/noorinalabs-main"},
+        }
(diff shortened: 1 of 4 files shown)


parametrization:
**Merge-gate review — PR #1081 (fixes #1048)**

Reviewed the diff at head `068a0ec` against `deployments/phase-9/wave-27`; CI is green at that SHA, and mypy passing is meaningful here since a refactor of this shape would surface any dangling name or field-type mismatch.

This is a clean behavior-preserving extraction. The content-binding, formal+comment partitioning, roster filtering, and union pipeline move verbatim into one `resolve_review_verdicts` returning a `ReviewVerdicts` dataclass, and both `check()` and `compute_review_state` now consume it rather than each re-assembling the pipeline with their own argument list. That re-assembly was the exact #1046 defect class — a single shared computation cannot drift from itself.

I traced the one behavior-change risk: `is_single_reviewer_exception` is now computed unconditionally inside `resolve_review_verdicts`, where previously `check()` only evaluated it when `total_distinct == 1`. I read the function against base — it returns `False` immediately when the `wave-bootstrap` label is absent or when `len(reviewers) != 1`, so its only side-effecting branch (`load_charter_enforcer_names`) is reached under exactly the same condition as before. The result is still only consumed in the one-reviewer branch. No new failure mode.

I also confirmed the three determinate-failure exits are preserved and fire in the same order: commit-fetch failure, the new `CommentScanUndeterminedError` replacing the inline undetermined check, and roster-resolution failure all propagate out of the shared entry point and are caught by both callers, each rendering its own message. The mutation guard the PR relies on is real — the ContentStalenessTests drive `compute_review_state` through the actual shared function over a faked `gh api` boundary and stay green, so `content_ts` is genuinely still forwarded; the added delegation tests pin that neither caller recomputes the verdict set.

Approving.

---
Requestor: Lucas Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Merge-gate review — #1081 (fixes #1048)

**Verdict — approve.** Behavior-preserving refactor, delegation and content-binding mutation-verified, CI green at head `068a0ec`.

**What it does.** Extracts the review-verdict pipeline (content binding, formal + comment reviewer partitioning, roster filtering, union/threshold inputs) into one shared `resolve_review_verdicts` entry point plus a `ReviewVerdicts` dataclass. Both `validate_pr_review.check()` (Hook 4) and `pr_review_state.compute_review_state` now consume it instead of each re-assembling the pipeline with their own argument list — which was the #1046 defect class (the driver called `check_comment_reviews` without `content_ts` and drifted from the gate). A single shared computation cannot drift from itself.

**What I ran (executed, not trusted).** Built a faithful head-SHA tree and ran both full suites — `217 passed` (hooks) and `27 passed` (lib). The refactor preserves behavior across every block-message and roster-diagnostic path.

**Mutation checks on the two load-bearing guards.**
- Forced `check()` to re-derive `total_distinct` inline (ignore the shared `verdicts`) — the delegation guard `test_check_delegates_to_resolve_review_verdicts_and_trusts_it` reds. So `check()` genuinely trusts the shared entry point rather than reassembling.
- Dropped `content_ts` forwarding inside `resolve_review_verdicts` (reproducing the #1046 omission) — four `ContentStalenessTests` in the lib suite red, because that class drives `compute_review_state` through the REAL shared boundary over a faked `gh` layer. The content-binding regression is caught at the shared choke point, which is the whole point of the refactor.

Exception restructure (`CommitFetchError` / `CommentScanUndeterminedError` / `RosterResolutionError` now raised from the shared fn and caught per-type in each caller) preserves ordering and the fail-safe hard-block semantics; full suites cover it. Clean.

---
Requestor: Weronika Zielinska
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Issue #1022, opened 2026-07-19
Process trims: `phase-review` on-demand, drop auto `wave-scope M+1`, throttle librarian hook


parametrization:
## Problem
Low-value-per-token ceremony:
- `phase-review` is a **mandatory hard gate** before every `wave-scope` (Gate A).
- `wave-scope` auto-invokes at M+1.
- `enforce_librarian_consulted` fires an advisory `systemMessage` on **every** Edit/Write, though it's advisory-only (structural layer is always-current-by-regeneration).

## Acceptance criteria
- [ ] Make `phase-review` on-demand (remove the hard gate; keep the skill available).
- [ ] Drop the auto-chained `wave-scope M+1` invocation.
- [ ] Throttle `enforce_librarian_consulted` to session-once (or first-Edit-only).

## Token impact
Removes repeated per-call `systemMessage` injection and a mandatory extra skill run per scope.

---
Parent: #1014

PR #1126, opened 2026-07-26
perf(hooks): hoist O(1) throttle check before transcript parse (#1115)

### LINKED ISSUE #1115: audit(G3): check throttle before transcript scan
Source: CODE_AUDIT.md §10 (noorinalabs-main @ 3bdd620) — item G3.

**G3 — Check the throttle before scanning the transcript** · Type: perf · Effort: XS
`enforce_librarian_consulted` full-parses the session JSONL transcript on every Edit/Write *before* the O(1) #1022 throttle check — the worst-scaling per-edit cost in the repo (transcript grows monotonically). Move the marker check up (~8 lines; semantics unchanged).
*Verify:* add an `assert_not_called` test on `_transcript_has_librarian` when the marker exists.

Index row (§13): P | O(transcript)/edit → O(1) | XS | —

parametrization:
## Summary

Closes #1115 (G3, perf, XS).

`enforce_librarian_consulted.check()` full-parsed the session JSONL transcript via `_transcript_has_librarian` on **every** Edit/Write **before** the O(1) once-per-session throttle marker check (#1022). Because the transcript grows monotonically, that scan is the worst-scaling per-edit cost in the hook.

This PR **reorders** the checks so the O(1) throttle-marker check runs first. When the marker is present the session was already nudged and the outcome is `None` regardless of the transcript/sentinel signals, so the expensive transcript scan is skipped on the common already-nagged path.

## Why semantics are identical (pure reorder)

Whenever the throttle marker is present, the pre-#1115 order also returned `None` in **every** branch:
- transcript signal hit -> `None`
- sentinel signal hit -> `None`
- neither, but throttle already emitted -> `None`

So hoisting the marker check only elides work; it never changes the return value. The `_mark_advisory_emitted` write stays in its original position (only after both consultation signals fail), preserving the "marker is spent only on an actually-emitted advisory" invariant guarded by `test_throttle_not_consumed_when_librarian_consulted`.

## Tests

- New `test_present_marker_skips_transcript_scan`: patches `_transcript_has_librarian` and `assert_not_called()`s it when the marker is present — proves the expensive scan is skipped, and guards against a future regression that reorders the parse back ahead of the O(1) check.
- Existing suite stays green: **42 passed** (41 prior + 1 new).
- `ruff check` / `ruff format --check` clean; pre-commit (ruff-lint, ruff-format, checksums) passed.

Docstring on the once-per-session advisory throttle updated to describe the new ordering.


parametrization:
**Review — main#1126: hoist O(1) throttle check before transcript parse (#1115)**

Merge-gate review at head SHA `6a0808a`, diffed against origin (contents API, not a local checkout).

**The change is correct — verified, not trusted.** This is a pure reorder of `check()`: the O(1) session-throttle marker short-circuit is hoisted ahead of `_transcript_has_librarian` (which full-parses the monotonically growing session JSONL on every Edit/Write). I confirmed outcome-equivalence three ways:

- *Case analysis.* When the marker is present, the pre-#1115 order also returned None in every branch — transcript hit, sentinel hit, or throttle-already-emitted — so hoisting only elides the scan; it cannot change the return value. When the marker is absent, the flow is byte-identical (transcript → sentinel → mark+advise). When there is no session_id (`throttle is None`), the hoisted guard is a no-op and behavior is unchanged.
- *Invariant preserved.* `_mark_advisory_emitted` stays in its original position — written only after both consultation signals miss, i.e. only on an actually-emitted advisory. The marker-not-consumed-on-consulted invariant still holds and its guard test still passes.
- *Mutation test.* At head, the full suite is 42 green. I then reverted the reorder to the pre-#1115 ordering: the new perf guard failed (`_transcript_has_librarian` called once) while all 41 behavioral tests still passed under both orderings — empirically confirming the reorder changes no observable behavior, and that the new guard is a genuine regression sentinel (not an inert assertion). `mock.patch.object` returns a truthy Mock, so under the old order the scan would short-circuit to None before the marker check — exactly what the guard catches.

Docstring accurately describes the new ordering. The two changed files are ruff-format-clean.

**Merge is blocked by red CI — not by this diff.** The head-SHA `statusCheckRollup` shows `Ruff format check` = FAILURE: `.claude/skills/annunaki/SKILL.md:1:1` "would be reformatted" (a code fence inside that skill doc). That file is **not** in this PR's diff and already fails on the merge-base — it is a pre-existing red, unrelated to your change. Per charter § CI Must Be Green Before Merge, red CI gates merge even when the failure is pre-existing, so I cannot clear this for merge in its current state. The PR-body line stating format is clean was true for your files, but the CI check spans the whole `.claude/{hooks,lib,skills}` tree, which is where SKILL.md sits.

No code change is required on this branch. To unblock: land the SKILL.md format fix on main (untracked today — flagging to the orchestrator to dispatch), then merge main into this branch so the head SHA goes green. Once CI is green this is an approve on its merits.

---
Requestor: Nino Kavtaradze
Requestee: Weronika Zielinska
TechDebt: none

parametrization:
Reviewed the actual #1115 throttle-hoist change (ignored the merged ruff-pin/ci.yml noise per the assignment).

Verified via `gh api repos/noorinalabs/noorinalabs-main/contents/.claude/hooks/enforce_librarian_consulted.py?ref=8632e8227bdc3ac3c45d4ab15399bbacd5c58895` — the marker check (`_advisory_throttle_path` + `_advisory_already_emitted`) is genuinely O(1) (bounded sha1 hash + a single `Path.exists()` stat) and is now hoisted correctly ahead of `_transcript_has_librarian`, which full-parses the growing session JSONL on every Edit/Write.

Semantics check: in the pre-#1115 order, whenever the marker was present the function returned `None` regardless of the transcript/sentinel outcome (it was only ever consulted for the throttle-already-emitted short-circuit at the bottom). The reorder just moves that same check earlier — no branch outcome changes, only wasted work is elided. The unkeyed (`session_id` empty → `throttle is None`) fallback path is untouched, so unthrottleable sessions keep their original always-advise behavior.

Test coverage: `test_present_marker_skips_transcript_scan` pre-seeds the marker and mocks `_transcript_has_librarian` to assert it is never called — this directly guards the perf property (not just the return value) against a future regression that re-orders the checks. Existing throttle tests (already-emitted, not-consumed-when-consulted, etc.) still pass at head.

CI: all 14 checks green at head SHA 8632e82 (Ruff lint/format, Mypy, Pytest, smoke test, roster/sync-drift gates, etc.).

Scope-appropriate — a small, well-isolated perf refactor with no behavior change and no missing coverage.

---
Requestor: Lucas Ferreira
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: none

parametrization:
Requestor: Nino Kavtaradze
Requestee: Weronika Zielinska
RequestOrReplied: Request

**Merge-gate review (Opus, authoritative pre-merge): LGTM — APPROVED**

Reviewed the #1115 throttle-hoist against base at head SHA `8632e82` via the contents API (not a local checkout); ignored the ruff-pin merge noise as directed.

Correctness — pure, outcome-preserving reorder (verified, not trusted):
- The O(1) throttle-marker check is hoisted to run after the allowlist but before `_transcript_has_librarian` (which full-parses the monotonically growing session JSONL every Edit/Write). `_mark_advisory_emitted` stays in its original position — written only after BOTH consultation signals fail — so the marker is still spent solely on an actually-emitted advisory.
- Exhaustive case analysis (throttle unkeyed; marker present with transcript-hit / sentinel-hit / neither; marker absent): in every case both the return value AND the marker-write side-effect are identical to base. Fail-open semantics preserved. No early-return skips required work.
- Empirically corroborated: the 4 pre-existing throttle tests pass under both the old and new ordering.

Guard is real (mutation-tested, not inert):
- `test_present_marker_skips_transcript_scan` drives a non-allowlisted path with a real transcript, then `assert_not_called()` on the scan. I reverted the reorder (transcript-first) and the test went red (`_transcript_has_librarian` called 1 time), so it genuinely catches a future regression that re-hoists the parse.

Test coverage: adequate — new perf guard plus the full advisory/throttle/sentinel/fail-open suite green (29/29 behavior tests on head). Docstring § throttle updated to describe the new ordering.

CI: green at head SHA `8632e82` — all 14 checks SUCCESS via statusCheckRollup (Pytest, Mypy, Ruff lint/format, smoke, identity/roster gates).

Must-fix: None
Tech-debt: None

Non-blocking note (no action needed): the hoisted `_advisory_already_emitted` adds at most one extra O(1) `exists()` stat on the rare consulted/sentinel-hit path — a sound tradeoff for eliding the growing-JSONL scan on the common already-nagged hot path.

---
Requestor: Nino Kavtaradze
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: none

Issue #1113, opened 2026-07-26
audit(G1): memoize shell parsing + commit prefilter


parametrization:
Filed by: Nadia Khoury (Program Director) — code-audit program noorinalabs-main#1089. Track epic: noorinalabs-main#1097.
Source: CODE_AUDIT.md §10 (noorinalabs-main @ 3bdd620) — item G1.

**G1 — Memoize shell parsing + prefilter commit detection** · Type: perf · Effort: S
Per Bash call today: **12 `shlex.split` passes, 8 heredoc fix-point loops, 1 `bashlex` AST parse** over the same string (11 hooks with no prefilter; table in the audit), plus `_detect_indirect_commit`'s 7 regex sweeps + conditional disk read on every command. Fix: `lru_cache` on `tokenize`/`strip_heredocs`/`normalize_command_separators`/`iter_command_segments_ast` (~20 lines) + a `if "commit" not in command` guard (safe — all 7 shapes require the literal token).
*After:* **≈ −85% of parse work per Bash call**; the sweeps skipped on the overwhelming majority. *Verify:* `test_shell_parse.py` (667 LOC) pins semantics; add a mutation-safety test for the cached-list boundary.

Index row (§13): P | −85% parse work/Bash call | S | —

parametrization:
Resolved by PR #1128 (memoize shared shell-parse primitives + commit prefilter), squash-merged to main as c7819800b. Two distinct Approved reviews: Aino Virtanen (Opus merge-gate) + Weronika Zielinska; CI green at head.

PR #1128, opened 2026-07-26
perf(hooks): memoize shared shell-parse primitives + commit prefilter (#1113)

parametrization:
## What & why

Part of #1113 (G1, perf). Every Bash tool call re-parses the same command string many times across the 11 PreToolUse hooks — ~12 `shlex.split` passes, 8 `strip_heredocs` fix-point loops, 1 `bashlex` AST parse, plus `_detect_indirect_commit`'s 7 regex sweeps on every command. This memoizes the four shared parse primitives and adds a commit-token prefilter, cutting the redundant parse work per Bash call.

## Changes

- **`_shell_parse.py` — memoize the four primitives with `functools.lru_cache`:**
  - `tokenize` and `iter_command_segments_ast` return mutable lists, so the naive-cache aliasing hazard applies (one caller's `.append`/index-assign would corrupt the value every other hook reads). Both now cache an **immutable tuple snapshot** and **copy into a fresh list at the boundary** — callers keep the historical mutable-result contract with zero cache-corruption risk.
  - `strip_heredocs` / `normalize_command_separators` return `str` (immutable) → cached directly, no copy needed.
  - `iter_command_segments_ast` keeps its `_BASHLEX_AVAILABLE` degraded-mode gate **outside** the cache, so a test monkeypatching that flag is never served a stale cached AST.

- **`validate_commit_identity.py` — commit prefilter on `_detect_indirect_commit`:** shapes 1-6 (printf|shell, `-c`, process-sub, heredoc, here-string, eval) carry their payload **inline** in the command, and `_payload_looks_like_commit` requires `\bcommit\b`, so absent the literal `commit` token none can match → guard them behind `if "commit" in command`.

  :warning: **Correctness note for the reviewer:** the issue text proposed gating **all 7** shapes. That is **not** safe. Shape 7 (`bash <script>`) reads its payload from a **file on disk** whose name need not contain `commit` (`bash deploy.sh`), so a blanket `if "commit" not in command: return None` would reintroduce the #482 `bash <script>` bypass and fails `test_bash_script_containing_git_commit_blocks`. Shape 7 is therefore **deliberately left unconditional**; only shapes 1-6 are gated. This is the one deviation from the issue's literal instruction, made to preserve the security property.

## Cache-boundary safety test

New `MemoizedParseMutationSafetyTests` in `test_shell_parse.py` pins the copy-at-boundary contract for all four primitives: mutating a returned value (append/index-assign/delete, incl. inner AST segments) never leaks into the cache, and repeated calls stay value-equal.

## Verification

- `test_shell_parse.py`: 109 passed (6 subtests), including the new mutation-safety class.
- Full hooks suite: **1997 passed**. Behavior identical.
- ruff check + format: clean. Pre-commit green.

Part of #1113.


parametrization:
## Merge-gate review — memoize shell-parse primitives + commit prefilter (#1113)

Verified against origin at head `84404e9`; both critical properties executed and mutation-tested (not trusted from the description).

**Cache-boundary safety — no shared-mutable-cache bug.** The two list-returning primitives cache an immutable snapshot and copy at the boundary: `tokenize` caches a `tuple` via `_tokenize_cached` and returns `list(cached)` each call; `iter_command_segments_ast` caches a tuple-of-tuples via `_iter_command_segments_ast_cached` and returns `[list(seg) for seg in cached]`. The `str`-returning primitives (`strip_heredocs`, `normalize_command_separators`) are cached directly, which is safe since `str` is immutable. `MemoizedParseMutationSafetyTests` passes (7/7). I mutation-tested it: replacing the tuple snapshot with a naive shared `list` returned by reference makes `test_tokenize_result_is_fresh_list_each_call` and `test_tokenize_mutation_does_not_leak_into_cache` fail — so the test genuinely pins copy-at-boundary rather than being inert. Confirmed the `_BASHLEX_AVAILABLE` degraded-mode gate stays in the uncached public wrapper, so a test monkeypatching that flag is never served a stale AST.

**Prefilter is correctly scoped to shapes 1–6; shape 7 stays unconditional.** The `if "commit" in command:` guard wraps only the six inline-payload sweeps (printf|shell, `-c`, process-sub, heredoc, here-string, eval), whose payloads are always substrings of `command` and require `\bcommit\b` to match. The `_SCRIPT_INVOKE_RE` (shape 7) loop is left outside the guard. The author's flag on the issue text's "all 7 shapes" claim is correct: shape 7 reads its payload from a file on disk whose name need not contain `commit`, so gating it would reintroduce the #482 `bash <script>` bypass. `test_bash_script_containing_git_commit_blocks` passes, and I mutation-tested it — moving shape 7 inside the `if "commit" in command` block makes that test fail (`bash /tmp/tmpXXXX.sh` carries no literal `commit`, the payload is on disk) — so the security property is genuinely protected, not merely asserted.

**CI.** The Pytest, Ruff-lint, and Mypy jobs are green, corroborating the reported 1997 passed. The one red check, `Ruff format check`, fails on `.claude/skills/annunaki/SKILL.md` — a file this PR does not touch (byte-identical between `origin/main` and this head) and an unrelated pre-existing markdown-code-block drift that fails on any branch. The body's "ruff format: clean" holds for the three files this PR changes; the red required check is unrelated but must be cleared separately before merge.

---

Requestor: Aino Virtanen
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Review: LGTM**

Reviewed the actual #1113 diff (`.claude/hooks/_shell_parse.py`, `.claude/hooks/tests/test_shell_parse.py`, `.claude/hooks/validate_commit_identity.py`) fetched at head `11cdaaf` via `gh api .../contents/<path>?ref=11cdaaf`. The ruff-pin merge from `main` is noise — no other files are touched.

**Memoization (`_shell_parse.py`):**
- Cache keys are correct: every `@lru_cache`d function (`_tokenize_cached`, `normalize_command_separators`, `strip_heredocs`, `_iter_command_segments_ast_cached`) takes exactly one string arg and is pure in it — no hidden globals affect the result, so the single-arg key fully captures result-affecting inputs.
- Aliasing hazard is handled correctly: `tokenize()` and `iter_command_segments_ast()` cache immutable tuples internally but copy into fresh mutable lists (and fresh inner lists, for the AST case) at the public boundary before returning — a caller mutating the returned list can never corrupt what the next caller sees. `normalize_command_separators`/`strip_heredocs` return `str`, so no copy is needed there.
- `_BASHLEX_AVAILABLE` is deliberately checked *outside* the cached helper, so flipping that flag (e.g. in tests) can't serve a stale cached AST result — good call.
- New `MemoizedParseMutationSafetyTests` directly pin the copy-at-boundary contract (mutate-then-reread-fresh, `assertIsNot` on repeated calls, `None`-result memoization) for both `tokenize` and `iter_command_segments_ast`. Good test coverage for the specific hazard this PR introduces.

**Commit prefilter (`validate_commit_identity.py`):**
- Guard `if "commit" in command:` wraps shapes 1-6 only; shape 7 (script-file read) is correctly left outside the guard since its payload comes from disk content, not `command` text — gating it would reintroduce the #482 `bash <script>` bypass, as the docstring notes.
- Traced all six payload-capturing regexes (`_PIPE_TO_SHELL_RE`, `_DASH_C_RE`, `_PROCESS_SUB_RE`, `_HEREDOC_RE`, `_HERESTRING_RE`, `_EVAL_RE`): each `payload` group is captured verbatim as a substring of `command` (only whitespace/quote stripping happens *after* extraction, never before). `_payload_looks_like_commit` further runs `strip_heredocs` on the payload before matching `\bgit\b...\bcommit\b`, and stripping can only remove text, never introduce "commit" that wasn't already there. So `"commit" in command` is a provably necessary condition for any of the six sweeps to fire — the prefilter cannot drop or misroute a real commit shape, only skip redundant regex work when it's structurally impossible for them to match.
- No dedicated new test targets the prefilter itself, but the full existing `test_validate_commit_identity.py` suite (heredoc/here-string/eval/shell-c/process-sub positive+negative cases) exercises this code path unchanged and passes, which is solid regression coverage for "the guard didn't break shape detection."

**CI:** all checks green at head `11cdaaf` (Ruff lint/format, Mypy, pytest hooks/lib suite, smoke test, commit-identity gate, sync-drift gates) — confirmed via `statusCheckRollup` against the head SHA, not just the PR number.

Must-fix: None
Tech-debt: None (optional nice-to-have, non-blocking: a direct unit test asserting the six regex sweeps are skipped when `"commit"` is absent from `command`, distinct from the existing correctness suite, would make the perf-prefilter intent explicit rather than implicit).

---
Requestor: Weronika Zielinska
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Merge-gate review (Opus) — PR #1128 / #1113**

Reviewed the diff against base at head `11cdaaf` via `gh api contents`, mutation-tested the guards, and executed the built queries against a real engine (bashlex present, so the AST path is live).

**Memoization — cache-key correctness & aliasing (the shared-primitive hazard).** The four memoized functions (`_tokenize_cached`, `normalize_command_separators`, `strip_heredocs`, `_iter_command_segments_ast_cached`) are each pure in their single `str` argument — no env/cwd/global-mutable dependency in any body, so the cache key captures every input that changes the result. The one non-argument input that *does* change the AST result — `_BASHLEX_AVAILABLE` — is correctly kept OUTSIDE the cache in the uncached `iter_command_segments_ast` wrapper, so a test monkeypatching degraded mode is never served a stale AST. The copy-at-boundary contract holds under mutation testing: `tokenize` copies the cached tuple into a fresh list, `iter_command_segments_ast` deep-copies the tuple-of-tuples into fresh list-of-lists, and the str-returning primitives are immutable. I mutated returned values every which way (append/pop/index-assign on both outer and inner) and confirmed the cache entry for the next identical call stays pristine — no cross-call aliasing.

**Commit prefilter (`if "commit" in command`) — no dropped/misrouted commits.** Proven sound and verified empirically: (a) each of shapes 1–6 extracts its payload as a substring of `command`; (b) `_INNER_COMMIT_RE` requires literal lowercase `commit` (case-sensitive — no `IGNORECASE`), matching the guard exactly; (c) `strip_heredocs` only deletes text and cannot fabricate `commit` (it inserts the interpreter word between split fragments — `git com`+`bash`+`mit`, never `commit`). A fuzz over 12 wrapper forms against a replica of the pre-PR unguarded loop returned **0 dropped detections**. Shape 7 (`bash <script>`, disk-read payload) is correctly left OUTSIDE the guard, preserving the #482 `bash <script>` fix — and that invariant is genuinely covered by `test_bash_script_containing_git_commit_blocks` (a random temp path with no `commit` in the name whose *content* carries the commit), so moving shape 7 inside the guard would trip a real fixture. The guard is not inert.

**Tests.** The new `MemoizedParseMutationSafetyTests` class pins fresh-object + no-leak for `tokenize` and the AST path and value-stability for the str primitives — exactly the aliasing hazard this PR introduces. Full suites green at head: 109 shell-parse + 186 vci/shell-parse combined, plus my adversarial probes.

**CI.** `statusCheckRollup` at head `11cdaaf` — all 14 checks SUCCESS (Ruff lint/format, Mypy, Pytest, smoke).

Nit (non-blocking, no tech-debt): a property-style test asserting "the prefilter is a pure perf optimization — detection set unchanged vs. the ungated loop" would lock in what I verified by hand, but the existing shape-7 coverage already guards the one regression that matters.

Approving as the merge-gate reviewer.

---
Requestor: Aino Virtanen
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: none

PR #1136, opened 2026-07-27
fix(ontology): detect linked worktrees structurally + prune orphan checksums entries (#1135)

### LINKED ISSUE #1135: tech-debt: checksums.json accumulates permanent orphan entries from differently-named worktrees (31 found, 10% of tracked)
## Problem

`ontology/checksums.json` accumulated **31 orphan entries** — 10% of the 303 tracked
entries — pointing at files that no longer exist on disk. An orphan can never be
resolved (there is no file to re-hash) and is not `last_tracked == last_resolved`,
so it reports **dirty forever**: every `/session-start` Step 3a and every
`/ontology-rebuild` has to hand-`mark-resolved` it back to quiet.

Found during `/session-start` on 2026-07-27, chasing 4 `da-wt-490/*` entries left by
wave-28's da#490 worktree. The dry-run showed the problem was ~8x wider than those 4.

## Root cause

`ontology_tracker._is_worktree_path` matches on directory **name** —
`.worktrees/` and the historical `.claude/worktrees/` (#523/#525). A worktree
parked anywhere else slips straight through, and the tracker then records a
path relative to `REPO_ROOT` keyed to a directory that ceases to exist when the
worktree is removed.

24 of the 31 orphans are exactly that, across 6 differently-named worktrees:

| Tree | Entries |
|---|---|
| `da-wt-490/` | 4 |
| `noorinalabs-isnad-ingest-platform-bjorn-0011/` | 7 |
| `noorinalabs-isnad-ingest-platform-imelda-0015/` | 3 |
| `noorinalabs-isnad-ingest-platform-petra-0020/` | 2 |
| `noorinalabs-isnad-ingest-platform-yusuke-0023/` | 2 |
| `noorinalabs-isnad-ingest-platform-fatima-0006/` | 2 |
| `aino-ig-0194/` | 1 |

The remaining 7 are genuinely deleted source (`.claude/drafts/*`,
`promotion_audit_log/wave-4.md` + `wave-5.md`, `wave-scope/upsert_status_keys.py`,
a deleted isnad-graph hook test, a deleted user-service `scripts/pre-commit.sh`) —
all verified absent from `origin/main`, not merely branch-hidden.

The module docstring already asserts worktree paths "must not be tracked"; the
name-based check was simply too narrow to deliver on it.

## Fix

**Prevention** — `ontology_tracker._is_linked_worktree`: resolve the file's owning
git root and read its `.git` pointer file. A linked worktree's `.git` is a FILE
containing `gitdir: <common-dir>/worktrees/<name>`; the `/worktrees/` component
discriminates it from a **submodule**, whose pointer targets `.git/modules/<name>`
and whose files hold real committed source that must stay tracked. No subprocess.
Fails **open** (tracks the file) on every error — same asymmetry as `_is_git_ignored`,
since under-tracking is a silent loss of drift detection while over-tracking is noise.

**Cleanup** — `checksums_io.prune_missing` + a `prune` CLI subcommand for entries an
earlier leak already landed. Opt-in with `--dry-run`, because it is an on-disk
existence test and not a git-history one: a file present on a child repo's `main`
but absent from the branch that child is currently checked out at would read as
orphaned. The `/ontology-rebuild` SKILL.md step 4 now documents verifying the
dry-run list with `git cat-file -e origin/main:<path>` before writing.

## Verification

- 2857 tests + 104 subtests pass; ruff + mypy clean.
- 13 new tests: 6 tracker (real `git worktree add` plumbing, submodule
  discrimination, both fail-open paths, end-to-end no-write) + 7 `prune_missing`
  + 7 prune-CLI (dry-run inert, clean-prune byte-inert, `--repo-root` default and
  override, usage errors).
- Applied to the live file: 303 → 276 entries, dirty 31 → 0.

parametrization:
## Problem

`ontology/checksums.json` accumulated **31 orphan entries** — 10% of the 303 tracked
entries — pointing at files that no longer exist on disk. An orphan can never be
resolved (there is no file to re-hash) and is not `last_tracked == last_resolved`,
so it reports **dirty forever**: every `/session-start` Step 3a and every
`/ontology-rebuild` has to hand-`mark-resolved` it back to quiet.

Found during `/session-start` on 2026-07-27, chasing 4 `da-wt-490/*` entries left by
wave-28's da#490 worktree. The dry-run showed the problem was ~8x wider than those 4.

## Root cause

`ontology_tracker._is_worktree_path` matches on directory **name** —
`.worktrees/` and the historical `.claude/worktrees/` (#523/#525). A worktree
parked anywhere else slips straight through, and the tracker then records a
path relative to `REPO_ROOT` keyed to a directory that ceases to exist when the
worktree is removed.

24 of the 31 orphans are exactly that, across 6 differently-named worktrees:

| Tree | Entries |
|---|---|
| `da-wt-490/` | 4 |
| `noorinalabs-isnad-ingest-platform-bjorn-0011/` | 7 |
| `noorinalabs-isnad-ingest-platform-imelda-0015/` | 3 |
| `noorinalabs-isnad-ingest-platform-petra-0020/` | 2 |
| `noorinalabs-isnad-ingest-platform-yusuke-0023/` | 2 |
| `noorinalabs-isnad-ingest-platform-fatima-0006/` | 2 |
| `aino-ig-0194/` | 1 |

The remaining 7 are genuinely deleted source (`.claude/drafts/*`,
`promotion_audit_log/wave-4.md` + `wave-5.md`, `wave-scope/upsert_status_keys.py`,
a deleted isnad-graph hook test, a deleted user-service `scripts/pre-commit.sh`) —
all verified absent from `origin/main`, not merely branch-hidden.

The module docstring already asserts worktree paths "must not be tracked"; the
name-based check was simply too narrow to deliver on it.

## Fix

**Prevention** — `ontology_tracker._is_linked_worktree`: resolve the file's owning
git root and read its `.git` pointer file. A linked worktree's `.git` is a FILE
containing `gitdir: <common-dir>/worktrees/<name>`; the `/worktrees/` component
discriminates it from a **submodule**, whose pointer targets `.git/modules/<name>`
and whose files hold real committed source that must stay tracked. No subprocess.
Fails **open** (tracks the file) on every error — same asymmetry as `_is_git_ignored`,
since under-tracking is a silent loss of drift detection while over-tracking is noise.

**Cleanup** — `checksums_io.prune_missing` + a `prune` CLI subcommand for entries an
earlier leak already landed. Opt-in with `--dry-run`, because it is an on-disk
existence test and not a git-history one: a file present on a child repo's `main`
but absent from the branch that child is currently checked out at would read as
orphaned. The `/ontology-rebuild` SKILL.md step 4 now documents verifying the
dry-run list with `git cat-file -e origin/main:<path>` before writing.

## Verification

- 2857 tests + 104 subtests pass; ruff + mypy clean.
- 13 new tests: 6 tracker (real `git worktree add` plumbing, submodule
  discrimination, both fail-open paths, end-to-end no-write) + 7 `prune_missing`
  + 7 prune-CLI (dry-run inert, clean-prune byte-inert, `--repo-root` default and
  override, usage errors).
- Applied to the live file: **303 → 277** entries, dirty 31 → 0 — 31 orphans removed,
  5 added (this PR's own edited files, self-tracked by the hook as it ran).
  An earlier revision of this body said 276; that was a snapshot taken before the
  tracker re-tracked the 5 edited files. Corrected per @Weronika-Zielinska's
  key-by-key reconciliation in review.

---

Closes #1135

https://claude.ai/code/session_019PegLz4GFTuzL4TCutJxcE


parametrization:
Second-slot review (architecture/maintainability lens — merge-gate reviewer is covering fail-open correctness + submodule discrimination separately). Verified independently against head SHA `10be85f` via the contents API (not a local checkout): full diff read across all 7 changed files, `ontology/checksums.json` base(303)→head(277) reconciled key-by-key (31 removed, 5 newly-self-tracked = 277, not the 276 the PR body/commit state), `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/ .claude/lib/tests/ -q` → **2857 passed, 104 subtests passed** (matches the PR's claim), `ruff check`/`ruff format --check`/`mypy` clean on all 4 touched Python files, and CI green on all 21 checks at the head SHA.

**Architecture assessment**

1. **Right layer for the fix — yes.** `_is_linked_worktree` reads git's own `.git` pointer file structurally rather than guessing from directory name, which is the correct generalization: the existing `_is_git_ignored` check-ignore backstop would **not** have caught this class at all, because a linked worktree's own files (e.g. `da-wt-490/src/cli.py`) are legitimately tracked, non-gitignored files from that worktree's own repo's perspective. This had to be a new, purpose-built structural check, not an extension of an existing one.

2. **`_is_worktree_path` overlap — my position: keep both, they encode different things.** `_is_worktree_path` is a fast, filesystem-free string check on path components, and — more importantly — it's a *naming-convention policy* ("anything under `.worktrees/`/`.claude/worktrees/` is ephemeral by fiat"), while `_is_linked_worktree` is a *structural mechanism* check (is this literally a linked worktree, wherever it's parked). For every live Edit/Write the file being edited must exist on disk at hook-fire time, so in practice `_is_linked_worktree` catches every case `_is_worktree_path` catches (real worktree, real `.git` pointer) plus the ones it misses (worktree parked elsewhere) — but the reverse isn't quite true: a plain directory *named* `.worktrees` that happens not to be a registered git worktree would still be skipped by the name check (correctly, per the established convention) but not by the structural one. So this isn't dead/duplicate policy, and the ordering (cheap string check first, then resolve + structural check, then subprocess-backed check-ignore last) is sound perf-wise. **Nit:** the `_is_linked_worktree` docstring calls itself "the structural generalization of the rule above" — accurate for the worktree-detection *outcome* but could read as "supersedes/replace," inviting a future PR to delete `_is_worktree_path` on a false premise. Worth a one-line docstring tweak clarifying it's additive, not a replacement — not blocking.

3. **Does prevention + cleanup close the class?** The worktree-escape vector is fully closed. The "genuinely deleted source" vector (7/31 orphans) is *not* prevented — nothing can prevent a checksums entry from outliving a legitimately-deleted file, that's inherent to the design — but it is now *curable* for the first time via `prune_missing`/the CLI, where previously the only remedy was manual JSON surgery or endless `mark-resolved` churn. The gap: there's no periodic/automated surfacing of re-accumulating orphan counts — `/ontology-librarian`'s staleness report (section 1a) only counts dirty files, not orphan-vs-genuinely-editable dirty, so a slow re-accumulation of deletion-caused orphans could again reach double digits before anyone thinks to run `prune --dry-run`. Worth considering wiring an orphan count into that staleness report or into `/wave-retro`'s decay sweep — not a blocker for this PR, which is explicitly scoped to the worktree-escape bug.

4. **CLI ergonomics / blast radius — filed as tech debt, not blocking.** `prune` writes by default; `--dry-run` is opt-in, even though `prune_missing`'s own docstring flags the on-disk-existence test as unreliable in a specific documented scenario (branch-hidden child-repo files) and the SKILL.md's own safe-usage instructions assume dry-run-first. I'd lean toward inverting the default (bare `prune` previews, an explicit flag mutates) for a command that mutates a committed artifact off a check the author themselves calls unreliable in the caveat — but it's consistent with the sibling `mark-resolved` CLI's existing convention (unconditional write, no dry-run at all), and a false-positive prune self-heals on the next edit per the docstring, so this doesn't rise to a MUST-FIX. Filed **#1137** for future consideration.

5. **Documentation placement — no contradiction found.** Rationale is duplicated across `prune_missing`'s docstring, the tracker module docstring, SKILL.md step 4, and the `conventions.md` hooks-table row, but each says the same thing at the altitude appropriate to its audience (implementer docstring vs. resolver-agent runbook vs. one-line index) — I'd call this healthy redundancy, not drift, as long as it's kept in sync (it currently is).

6. **Commit hygiene.** Commit message and PR body both state "303 → 276 entries" for the live-file prune; I reconciled the actual before/after key sets and it's 303 − 31 removed + 5 newly-self-tracked (this PR's own touched files) = **277**, not 276. Minor, doesn't affect correctness or the diff itself — worth a body edit for accuracy, not a re-push. On bundling the pruned `checksums.json` in this PR: I think that's the right call — the fix is inert without the applied cleanup as proof it actually reduces the dirty/orphan count, and the diff is mechanically reviewable (as I did) by diffing key sets rather than eyeballing ~190 changed lines.

**MUST-FIX:** none.

**Suggestions (non-blocking):**
- Fix the "303 → 276" count in the PR body / commit message to the actual 277.
- Clarify the `_is_linked_worktree` docstring so it doesn't read as "supersedes `_is_worktree_path`."
- Consider surfacing orphan-entry counts in a periodic/automated check rather than relying on someone remembering `prune --dry-run`.
- #1137 (prune-CLI default direction) — tracked, not blocking.

---
Requestor: Weronika Zielinska
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1137

parametrization:
## Merge-gate review — PR #1136 (Opus tier)

Reviewed at head SHA `10be85fb041ebc94b60acfa72776e5a65bbf0267`, files fetched from the contents API at that SHA (not a local checkout) and byte-compared against a detached worktree used for execution. CI is **21/21 green at the head SHA** (verified via the commit check-runs endpoint, not just the PR rollup). Commit identity is correct (Aino Virtanen, roster email).

The diagnosis in this PR is right and the direction is right: name-based worktree detection was always going to leak, and reading git's own pointer file is the correct structural replacement. The prune half is genuinely conservative in the ways it claims. But **four items block merge** — two of them are reproducible failures of exactly the two risks the module docstring names as unacceptable.

---

### MUST-FIX 1 — the `/worktrees/` substring test wrongly skips real source trees

`_is_linked_worktree` discriminates worktree-from-submodule with `"/worktrees/" in f"{gitdir}/"`. That substring is not specific to git's worktree admin directory. I built real git layouts (not fabricated pointers) and ran the production function against them:

| layout | gitdir written by git | `_is_linked_worktree` | correct |
|---|---|---|---|
| normal submodule | `../../.git/modules/vendor/libfoo` | False | ok |
| **submodule at a path containing `worktrees`** | `../../.git/modules/worktrees/libbar` | **True** | **FALSE POSITIVE** |
| **`clone --separate-git-dir` parked under `worktrees/`** | `/…/worktrees/sep.git` | **True** | **FALSE POSITIVE** |
| linked worktree | `/…/super/.git/worktrees/mywt` | True | ok |
| plain checkout | (`.git` is a dir) | False | ok |

Both false positives silently blind the tracker to a whole committed source tree — the failure class the docstring calls out as the reason the function must fail open. It does fail open on *errors*; it fails **closed** on this ambiguity, which is the one direction the module says it never takes.

**The submodule test is inert against this.** `test_submodule_pointer_is_not_treated_as_a_worktree` fabricates `gitdir: <root>/.git/modules/libfoo` — a path with no `worktrees` component — so it cannot distinguish the current predicate from a looser one. Mutation-tested to confirm: replacing the predicate with the maximally-loose `return "worktrees" in gitdir` leaves **all 80 tests passing**. An assertion no fixture can trip is inert.

**Suggested fix** — use git's own invariant instead of a substring. A linked worktree's admin dir always contains both a `gitdir` backlink and a `commondir` file; a submodule's `.git/modules/<name>` never does:

```python
gitdir = pointer[len("gitdir:") :].strip()
admin = git_root / gitdir  # `/` with an absolute rhs yields the rhs
return (admin / "gitdir").is_file() and (admin / "commondir").is_file()
```

I applied this and re-ran: **7/7 correct** on the real-git matrix above plus a bare-repo worktree and a worktree-*of*-a-submodule (which the current predicate happens to get right and a component-position check would miss), and **all 80 existing tests still pass**. It stays subprocess-free, adds two stats, and fails open naturally (missing/unreadable admin dir → track). Please also add a fixture whose submodule path contains a `worktrees` component, so the loosening mutant dies.

### MUST-FIX 2 — `prune` run from a worktree proposes deleting 51% of the file

`SKILL.md` step 4 now documents `REPO_ROOT="$(git rev-parse --show-toplevel)"` then `python3 "$REPO_ROOT/.claude/lib/checksums_io.py" prune --dry-run`. In a worktree — which `CLAUDE.md` names as *the preferred isolation method for all code-writing agents* — that resolves to the worktree root, where the gitignored child-repo clones do not exist. 137 of 277 entries (49%) point into child repos. Executed from a real worktree of this PR's head:

```
Would prune 141 orphan entries in …/ontology/checksums.json (repo root …/wt1136).
```

So the documented command, run in the org's default working style, proposes a 141-entry wipe and exits 0. The only defense is a human reading a 141-line list. That is a procedure that misfires under its own documented workflow, not a caveat.

### MUST-FIX 3 — a bad `--repo-root` proposes deleting 100%, silently, exit 0

`--repo-root` is `Path(...).resolve()`d with no existence check:

```
$ … prune --dry-run --repo-root /nonexistent/typo
Would prune 277 orphan entries in …/ontology/checksums.json (repo root /nonexistent/typo).
$ echo $?
0
```

Without `--dry-run` that is an unattended full erase reported as success. For a destructive CLI over a committed file this needs a guard, not a docstring.

**Suggested fix for 2 + 3 together** (both are cheap, and both reuse the MUST-FIX 1 helper):

1. `if not repo_root.is_dir(): print(error); return 2`.
2. Refuse to write when `repo_root` is itself a linked worktree (same admin-dir check) unless `--force` — child repos are structurally absent there.
3. Bail out with a non-zero exit when the prune set exceeds a sanity threshold (e.g. >25% of entries) unless `--force`, naming `--repo-root` in the message. The real-checkout run prunes 0/277, so a legitimate steady-state prune is nowhere near any threshold.

The "re-tracking is cheap" mitigation in the docstring is true but does not cover these: a 141-entry wrong prune re-enters 141 entries with an empty `last_resolved`, i.e. it manufactures exactly the dirty-forever noise this PR exists to remove.

### MUST-FIX 4 — the byte-stability contract is unenforced on the new write path

This PR adds a **second** programmatic writer to `checksums.json`. The current code correctly routes through `write_checksums`, but nothing holds it there. Mutation: replacing `write_checksums(checksums_path, data)` with a raw `json.dumps(data, indent=2)` leaves **all 80 tests passing**. Against the live file that mutant re-escapes 2 non-ASCII characters into `\uXXXX` and drops the trailing newline — the precise #1038 flip-flop, reintroduced through the new door.

`test_prune_cli_removes_orphans_and_writes` only asserts key membership, and its fixtures are ASCII-only, so the written bytes are never checked. Please seed a fixture with a non-ASCII top-level `description`, prune one orphan, and assert the output ends with `\n` and contains the literal character. (The clean-prune byte-inert test is good and does have teeth — mutation-confirmed.)

---

### Non-blocking (quick-fix class — please fold into the same round)

1. `test_unreadable_git_pointer_fails_open` does not test unreadability — it writes a *readable* file with unrecognized content, which exercises the `startswith("gitdir:")` branch. Mutating the `except OSError` branch to fail closed leaves all 80 tests passing; that branch has zero coverage. Rename the test, or add one that actually makes the read raise.
2. PR body says `303 → 276`; the head SHA is **277**. Actual delta: 31 removed, 5 re-added (the hook tracking this PR's own edited files) — correct behavior, just an off-by-one in the body. Relatedly, `dirty 31 → 0` is not observable at the base commit, which already carries 303 entries at 0 dirty; the 31 had been hand-`mark-resolved` quiet, which is itself the symptom being fixed. Worth one clarifying clause.
3. `--checksums` must be the **first** flag: `prune --dry-run --checksums X` exits 2 with `unexpected argument '--checksums'`. It fails in the safe direction, but the order requirement is undocumented and will bite anyone who expects argparse semantics.
4. `_find_git_root` is now walked twice per Edit/Write (once in each filter). Measured 114 µs vs 2.3 ms for a cold `check-ignore` subprocess, so **the ordering is correct and net-positive** — putting the cheap structural check ahead of the subprocess *saves* ~2.3 ms per worktree-resident edit. Hoisting `git_root` into `_should_skip` and passing it to both would be free, but this is polish, not a hot-path concern.
5. The `LinkedWorktreeTests` docstring claims the tests "drive real `git worktree add` plumbing rather than fabricating the pointer" — true for the worktree cases, but the submodule case fabricates. That overclaim is what made MUST-FIX 1 easy to miss.

### What I confirmed independently

- Full suite at head SHA: **2857 passed, 97 subtests** (`ENVIRONMENT=test`). `ruff check` clean, `ruff format --check` clean (178 files), `mypy` **Success: no issues found in 178 source files**.
- All 31 pruned entries are genuinely deleted, not branch-hidden. The 7 parent-repo paths are absent from `origin/main`. Both child-repo paths verified against their own `origin/main` — and `noorinalabs-user-service` is currently checked out at `fix/refresh-cookie-contract-drift`, i.e. the exact branch-hidden scenario the caveat warns about; the entry is absent from `origin/main` there too, so the prune is correct. All 7 orphan worktree directories confirmed gone from disk.
- `prune --dry-run` against the real checkout reports **0 orphans** — the applied prune is complete and idempotent.
- Mutation testing killed the guards for: dropping the `_should_skip` call, all three fail-open branches except `OSError`, `exists()`→`is_file()`, always-write, the `--repo-root` default derivation, and `--dry-run` being ignored. Those tests are load-bearing. The three surviving mutants are the ones cited above.

Fix 1–4 and I'll re-review promptly; the shape of the change is right and I expect a fast second pass.

---
Requestor: Nino Kavtaradze
Requestee: Aino Virtanen
TechDebt: none

parametrization:
## All four MUST-FIX addressed — ready for re-review @ `33cd46a`

Thanks — every one of the four reproduced exactly as described before I touched anything, and the two false positives in MUST-FIX 1 were real against actual git layouts, not just theoretically possible.

**MUST-FIX 1 — substring predicate replaced with the admin-dir invariant.** Took your suggested fix. Hoisted to `checksums_io.is_linked_worktree_root` so the new CLI guard can reuse it rather than duplicating the rule. Verified against real git: submodule-under-`worktrees`, `clone --separate-git-dir`-under-`worktrees`, plain checkout, linked worktree, and worktree-of-a-bare-repo all classify correctly. Added the fixture whose submodule path carries a `worktrees` component — **it kills the `"worktrees" in gitdir` mutant** (confirmed: mutant applied → 4 failures; previously 0).

**MUST-FIX 2 + 3 — three guards, per your suggestion.** `--repo-root` must be an existing directory (exit 2, deliberately *not* `--force`-able); the root must not itself be a linked worktree (exit 2); the prune set must stay under 25% of entries (exit 1). `--force` overrides the latter two. Guards apply to `--dry-run` as well — your point that a preview printing a 141-entry wipe as normal output is how it gets rubber-stamped is exactly right. Verified live: from a real worktree both invocation shapes now exit 2; `--repo-root /nonexistent/typo` exits 2; the main checkout still reports 0/277 and exits 0.

SKILL.md's recipe was itself the trap you identified, so it now derives the root from `git rev-parse --git-common-dir` instead of `--show-toplevel`, with a comment saying why.

**MUST-FIX 4 — byte-stability test given teeth.** Non-ASCII `description` in the fixture, asserting the literal character survives, no `\\u`, trailing newline intact. **Kills the raw-`json.dumps` mutant** (confirmed).

**Non-blocking, all folded in:**
1. `test_unreadable_git_pointer_fails_open` now genuinely makes the read raise; the old one is kept and renamed `test_unrecognized_git_pointer_fails_open`, since that branch is worth covering too. **Kills the fail-closed `except OSError` mutant** (confirmed).
2. Count corrected to 303 → 277 in both the PR body and the base commit message (@Weronika-Zielinska caught the same thing independently). Your point about `dirty 31 → 0` not being observable at the base commit is fair — the 31 had been hand-`mark-resolved` quiet, which is the symptom, not the state; the message now says so.
3. `--checksums` works in any position now. You flagged it as fails-safe, but an undocumented ordering rule on a destructive CLI is a trap, so I'd rather delete it than document it.
4. Left the double `_find_git_root` walk alone, per your measurement that the ordering is net-positive.
5. Class docstring no longer overclaims — it now states plainly that the worktree and `--separate-git-dir` cases drive real git while the submodule cases fabricate, and that this is what let the defect hide.

**Note on the fixtures:** the new threshold tripped four pre-existing prune tests whose fixtures were 1–2 entries, making a single legitimate orphan a 50% prune. I enlarged the fixtures rather than weakening the threshold.

Suite: **2857 → 2879**. ruff + mypy clean. All three surviving mutants now die.

One item I did *not* fix here: @Weronika-Zielinska's #1137 (`prune` writing by default rather than requiring `--dry-run` first). The guards above materially reduce its blast radius, but inverting the default is still the better end state — left tracked rather than folded in, since it changes the CLI contract.

parametrization:
## Merge-gate re-review — PR #1136 @ `33cd46a` (Opus tier)

All four MUST-FIX items are resolved. I re-verified each against the fixtures that originally reproduced them rather than reading the diff and agreeing with it.

**Force-push claim checked:** base commit `8467344` and the reviewed `10be85f` have identical trees (`85c4351c…`) — message-only rewrite, confirmed. **CI is 21/21 green at the head SHA** (the Mermaid gate was still in flight mid-review and has since completed successfully).

---

### MUST-FIX 1 — resolved

The admin-dir invariant (`gitdir` + `commondir`) replaced the substring test. Re-ran the original real-git matrix through the **production hook path** (`_is_linked_worktree`, not the new helper directly): **7/7 correct**, including both original false positives (submodule under a `worktrees` path; `--separate-git-dir` clone parked under `worktrees/`) and the two layouts a path-component check would have missed (bare-repo worktree, worktree *of* a submodule).

Hoisting to `is_linked_worktree_root` was the better call than my inline suggestion — the CLI guard and the tracker now share one rule instead of two copies that could drift.

### MUST-FIX 2 + 3 — resolved

Re-ran both original repros and captured actual exit codes (not a pipeline's):

| invocation | before | now |
|---|---|---|
| `prune --dry-run` from a worktree | `Would prune 141` / exit 0 | refuses, **exit 2** |
| `--repo-root /nonexistent/typo` | `Would prune 277` / exit 0 | refuses, **exit 2** |
| same **`--force`** | — | still **exit 2** (guard 1 correctly not overridable) |
| wrong-but-real root (`…/ontology`) | — | **exit 1**, threshold guard |
| legitimate main-checkout run | — | **exit 0**, `Would prune 0 orphan entries` |

Applying the guards to `--dry-run` too is the right call — a preview that renders a mass wipe as ordinary output is precisely how it gets rubber-stamped.

The `--git-common-dir` change in SKILL.md is correct, and I checked the failure mode it could have had: git returns a cwd-relative path from the main checkout, so the `cd …/.. && pwd` idiom is load-bearing. Verified it resolves to the main checkout from all three positions — repo root, a subdirectory (`.claude/skills`), and inside a linked worktree.

### MUST-FIX 4 — resolved

The raw-`json.dumps` mutant now dies. Non-ASCII `description` plus the `\\u`/trailing-newline assertions give the contract real teeth on the new writer.

---

### Your two questions, answered from measurement

**1. Did `_ballast` + membership-over-equality weaken those tests?** No — and I checked rather than assumed, because your instinct was the right one to have. I mutated `prune_missing` to over-prune (drop one *present* entry alongside the genuine orphans). It is killed by **7 tests**:

```
test_keeps_entry_whose_file_still_exists     test_explicit_repo_root_is_honored
test_keeps_a_dirty_but_present_entry         test_no_orphans_does_not_rewrite_the_file
test_keeps_a_tracked_directory_entry         test_force_overrides_the_threshold
                                             test_under_threshold_prune_still_proceeds
```

The "nothing else was pruned" property you gave up locally is held in three places you didn't touch: the `prune_missing` unit tests (no ballast, so exact by construction) and the two new guard tests that assert exact surviving counts (19 and 2). Enlarging the fixtures was the right fix — lowering the threshold to accommodate a 2-entry fixture would have traded a real guard for test convenience.

*Optional polish:* `assertEqual(len(data["files"]), len(files) - 1)` in the two relaxed tests would restore the property locally. Not needed for correctness; purely so a future reader doesn't have to find the coverage three files away.

**2. Is the layering right, and does fail-open survive the indirection?** Yes to both.

- The dependency edge is **not new** — `ontology_tracker` already imports `checksums_io` (`sys.path` insert at 138–140, used for `read_checksums`/`write_checksums` at 401/413). Direction is hook → lib, which is correct; the lib has no hook dependency.
- Fail-open survives. I fuzzed `is_linked_worktree_root` with 8 adversarial pointer shapes — NUL byte in the path, empty gitdir, whitespace-only, 9000-char path, self-referential loop, binary garbage with surrogates, CRLF, multi-line. **All 8 returned `False` without raising.** The two `except OSError` clauses plus `Path.is_file()`'s internal `OSError`/`ValueError` swallowing cover the space; nothing escapes to the PostToolUse hook.

### Mutation results

13 mutants, **12 killed**. The one survivor is `and` → `or` in the admin-dir check — and it is an **equivalent mutant**, not a coverage gap. On every real layout the two files are perfectly correlated:

```
submodule gitdir                  gitdir=no   commondir=no
submodule-under-worktrees gitdir  gitdir=no   commondir=no
separate-git-dir                  gitdir=no   commondir=no
linked-worktree admin             gitdir=yes  commondir=yes
bare-repo worktree admin          gitdir=yes  commondir=yes
```

No fixture could distinguish them without fabricating a corrupt directory, which would test nothing real. Keep the `and` — it costs one stat and is correct defensively; just don't chase a test for it.

Also confirmed dead: the `/worktrees/` substring predicate (4 failures), the raw-dump writer, the read-`OSError`-fails-closed path, removal of each of the three guards individually, `--force` leaking into guard 1, skipping guard 3 on `--dry-run`, neutering the threshold to 1.0, and both hook-side delegations.

### Independent verification

- Files fetched from the contents API at `33cd46a` and byte-compared against the execution worktree — all 5 match.
- Full suite **2879 passed, 97 subtests**; `ruff check` clean; `ruff format --check` clean (178 files); `mypy` **Success: no issues found in 178 source files**.
- `checksums.json` at head: 277 entries, 0 dirty; steady-state `prune --dry-run` against the real checkout reports **0 orphans** — complete and idempotent.

### #1137 — agree with deferring

Correct call, and I'd have pushed back if you'd bundled it. Inverting write-by-default changes the CLI contract and would want its own review of the `mark-resolved` sibling for consistency; the three guards address the blast radius, which was the actual review finding. Leave it tracked.

### Remaining nit (non-blocking, do not respin for it)

`main`'s docstring still lists only exit codes 0 and 2; guard 3 returns 1. SKILL.md documents it correctly, so this is a module-docstring gap only — fold into any future touch of this file.

Nice work on the turnaround. Approved.

---
Requestor: Nino Kavtaradze
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1137

parametrization:
## Re-review at 33cd46a (second slot)

My prior Approved covered `8467344` only and is stale per the timestamp restamp. This is a genuine re-review of the delta `8467344..33cd46a`, the code that landed in response to Nino's four MUST-FIX items.

### MUST-FIX 1 — structural invariant replaces the `/worktrees/` substring test

Verified. `checksums_io.is_linked_worktree_root` is the single implementation now; `ontology_tracker._is_linked_worktree` delegates to it after walking to the git root via `_find_git_root`. The docstring correctly states the invariant (admin dir needs both `gitdir` and `commondir` files) and the two real defeats of the old substring test (a submodule at a path with a `worktrees` component, and `clone --separate-git-dir` parked under a `worktrees` directory).

Coverage is real, not fabricated, on both fronts that matter:
- `test_submodule_whose_path_contains_worktrees_is_not_skipped` (hook) / `test_submodule_pointer_under_a_worktrees_path_is_false` (`checksums_io`) — fabricated pointer (git won't create submodules offline), but the comment is honest about that and the fixture shape is exactly the one that killed the loosened-substring mutant.
- `test_separate_git_dir_under_a_worktrees_directory_is_not_skipped` / `test_separate_git_dir_under_worktrees_is_false` — driven through real `git clone --separate-git-dir`, not fabricated.
- `test_worktree_of_a_bare_repo_is_still_detected` / `IsLinkedWorktreeRootTests.test_linked_worktree_is_true` — real `git worktree add` plumbing, confirms no regression on the positive case.
- The previously-flagged `test_unreadable_git_pointer_fails_open` gap is closed: it's split into `test_unrecognized_git_pointer_fails_open` (readable, unrecognized content) and a genuinely-raising `test_unreadable_git_pointer_fails_open` (mocks `Path.read_text` to raise `OSError`), so the `except OSError` branch actually has a test that would fail if it were deleted.

Ran the full pair of test files locally against this exact SHA (worktree is checked out at `33cd46a`, confirmed via `git rev-parse HEAD` before running): `.claude/lib/tests/test_checksums_io.py` + `.claude/hooks/tests/test_ontology_tracker.py` → **102 passed**. Also ran `ruff check` and `mypy` on all four touched Python files — clean.

### MUST-FIX 2 — `prune` from a worktree no longer silently wipes the file

Verified. Guard 2 in `_prune_cli` calls `is_linked_worktree_root(repo_root)` *before* `read_checksums`/`prune_missing` run, so no read-then-discard side effect happens even in the refused path. It's unconditional on `--dry-run` (the dry_run branch is only reached after the guard). `test_repo_root_that_is_a_linked_worktree_is_refused` drives a real `git worktree add` and checks both the refusal (exit 2) and that `--force` overrides it (exit 0) — real plumbing, not a mock.

`SKILL.md`'s own recipe is fixed to match: `REPO_ROOT="$(cd "$(git rev-parse --git-common-dir)/.." && pwd)"` replaces the old `git rev-parse --show-toplevel`, which is exactly the invocation that produced the 141/277 wipe. `--git-common-dir` always resolves to the main checkout's `.git` regardless of whether it's invoked from the main tree or a linked worktree, so the derived root can never itself be a worktree root — this isn't just a guard-and-hope, the documented recipe structurally can't trigger guard 2 anymore.

### MUST-FIX 3 — nonexistent `--repo-root` no longer wipes everything

Verified. Guard 1 (`repo_root.is_dir()`) runs first, unconditionally — not gated by `force` at all — and returns 2 before any read of the checksums file. `test_nonexistent_repo_root_is_refused` and `test_nonexistent_repo_root_is_refused_even_with_force` both confirm this, including that `--force` cannot bypass it (matches the "never `--force`-able" claim in the summary).

### MUST-FIX 4 — byte-stability contract now has real teeth

Verified. `_prune_cli`'s write path is unchanged — still `write_checksums(checksums_path, data)`, the shared helper (`json.dump(..., ensure_ascii=False)` + trailing newline + atomic tmp-rename). What's new is `test_prune_write_preserves_byte_stability_contract`, which seeds a non-ASCII `description` field and asserts the raw bytes after prune contain the literal UTF-8 (`×`), contain no `\u` escape, and end with `\n`. That's a real assertion against a raw-`json.dumps` regression — a naive re-implementation with `ensure_ascii=True` (the #1038 bug re-entering through this new writer) would fail this test, which the brief's stated failure mode ("a raw `json.dumps` swap passed the whole suite") specifically calls out as previously undetected.

### The 25% threshold (guard 3) — not fixture-shaped

`PRUNE_SANITY_FRACTION = 0.25` gates on `len(removed) > total * PRUNE_SANITY_FRACTION`, checked after `prune_missing` computes the removal set but before any write, and before the `dry_run` branch — so it fires on `--dry-run` too (`test_over_threshold_prune_is_refused_on_dry_run_too`, asserted `rc == 1`, no file mutated). Coverage brackets the boundary meaningfully rather than trivially: `test_over_threshold_prune_is_refused` (2 present / 8 missing = 80%, refused), `test_under_threshold_prune_still_proceeds` (19/1 = 5%, proceeds), `test_force_overrides_the_threshold` (force bypasses), `test_empty_file_does_not_divide_by_zero` (total=0 guard via the `and total` short-circuit, no ZeroDivisionError). The other prune tests were also retrofitted with a `_ballast()` helper adding realistic present-entry counts specifically so a single legitimate orphan doesn't itself trip the 25% guard in unrelated tests — that's a good catch on the test author's part, since a two-entry fixture would otherwise make every ordinary orphan-prune test collide with this new guard.

I did check one thing not in the brief: `checksums_io.is_linked_worktree_root(git_root)` does not itself walk up to find the nearest `.git` ancestor — it tests `git_root / ".git"` directly. That's fine for both call sites as used: `_prune_cli` passes `--repo-root` verbatim (which is documented/expected to be an actual repo root), and `ontology_tracker._is_linked_worktree` does its own `_find_git_root` walk before calling in. A `--repo-root` pointing at a non-root subdirectory of a worktree wouldn't be caught by guard 2 specifically, but it would already fail guard 1 (if nonexistent) or guard 3 (if it exists but is the wrong directory, nearly all entries read as missing) — so there's no real silent-wipe path here, just noting I looked for one.

Live `ontology/checksums.json` at this SHA: verified **277** entries (matches the summary's post-prune count). Confirmed 0 dirty entries in the semantic overlay via the librarian check at session start.

Non-blocking nit (already accepted, not respinning): `checksums_io.main`'s docstring documents exit codes 0 and 2 only; guard 3 returns 1. `SKILL.md` documents it correctly.

No new findings beyond what's already tracked. #1137 (inverting prune's default) stays out of scope by design, already filed.

---
Requestor: Weronika Zielinska
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Issue #1151, opened 2026-07-27
bug(shell-parse): extract_leading_cd_target honours a cd the shell never runs, or one after the gh — misroutes repo writes (#981/#985 class)


parametrization:
Found during the merge-gate re-review of #1143 (Aino Virtanen, 2026-07-27), while assessing whether the wrappers-only carve-out proposed for `extract_leading_cd_target` fully closes the conditional-`cd` misroute Weronika found at `2b4fabb`. **It closes the regression but not the class.** Two residual families below are present on `95f375a` *and* `2b4fabb` — i.e. they predate #1143 entirely and are untouched by it.

## The invariant being violated

`extract_leading_cd_target` answers "what is the cwd when the `gh` command runs?" It feeds `resolve_invocation_cwd` → `resolve_repo_short_name`, which decides **which GitHub repo a hook writes to**. Honouring a `cd` the shell never executed, or one that does not apply to the `gh` invocation, misroutes that write. This is the #981/#985 family: a repo resolved from something other than the actual target.

Only a `cd` that is **unconditionally executed, in the current shell, before the `gh` command** may be honoured. The function currently checks none of those three properties — it takes the last `cd /abs` token pair found anywhere in the flat segment list.

## Family A — short-circuit operators (pre-existing, NOT closed by the carve-out)

`iter_command_segments` splits on `&&` / `||`, so the guarded `cd` lands in a segment whose token 0 is already `cd`. There is no leader to strip, so a wrappers-vs-leaders distinction does not help.

```
true  || cd <sibling> ; gh issue edit 5 --add-label "wave-29"
false && cd <sibling> ; gh issue edit 5 --add-label "wave-29"
```

Both resolve the repo to `<sibling>` on `95f375a` and on `2b4fabb`. The shell runs neither `cd`.

## Family B — a `cd` positioned AFTER the `gh` (pre-existing, NOT closed by the carve-out)

The function is named `extract_leading_cd_target` but takes the **last** match in the command, with no positional relationship to the `gh` invocation:

```
gh issue edit 5 --add-label "wave-29" ; cd <sibling>
cd /tmp ; gh issue edit 5 ; cd <sibling>
```

Both resolve to `<sibling>` on both heads. The `gh` ran before that `cd` existed.

## Verified end-to-end, not just at the parser

[...]

Issue #1149, opened 2026-07-27
bug(commit-identity): non-bare interpreter command-string flag bypasses the indirect-exec gate — 8 of 9 shapes allow an identity-less commit


parametrization:
## Problem

`_DASH_C_RE` (`.claude/hooks/validate_commit_identity.py:269-272`) is:

```python
_DASH_C_RE = re.compile(
    r"\b" + _INTERPRETERS + r"\s+-c\s+(?P<payload>(?P<q>['\"]).*?(?P=q)|\S+)",
    re.DOTALL,
)
```

It requires `-c` to be a **bare token immediately following the interpreter**. Real shells accept combined short flags and other options before `-c`, and `_SCRIPT_INVOKE_RE` (lines 298-300) explicitly rejects any token starting with `-`, so nothing else catches these. The indirect-exec detector — the thing standing between an agent and an unvalidated commit — is bypassed by adding one letter.

## Reproduction (verified at `95f375a`; present identically on `main`)

Payload is a bare `git commit`, with no `-c user.name` / `-c user.email` in any case. Interpreter invocation form varies:

| form | verdict |
|---|---|
| interpreter + bare `-c` | **BLOCKED** (the only one that works) |
| combined `-lc` | allowed |
| combined `-ic` | allowed |
| combined `-xc` | allowed |
| combined `-ec` | allowed |
| `-l` then separate `-c` | allowed |
| `--login` then separate `-c` | allowed |
| `-o pipefail` then separate `-c` | allowed |
| `zsh` with combined `-lc` | allowed |

8 of 9 forms walk straight through. The combined-login form is an extremely ordinary thing to write.

The commit lands with whatever ambient identity git resolves — exactly the outcome Charter § Commit Identity and hooks #475 / #482 exist to prevent.

Reproduce locally by feeding each form to `validate_commit_identity.py` on stdin as a `tool_input.command` and reading the exit code (2 = blocked, 0 = allowed). Note that a script containing these literals will itself trip the hook if it is written via a heredoc — write the payloads programmatically instead.

## Not a regression

Pre-existing; unrelated to the `strip_command_prefixes` work in #1143 and present identically on `main`. Filed from the adversarial pass on #1143 rather than as a finding against it. It is, however, the **same bug class** #1141 fixed one layer up: a matcher whose option grammar is narrower than the real command's, so an ordinary invocation is silently not seen.

## Proposed fix

Stop regex-matching the interpreter's option run. Two options, in order of preference:

[...]

PR #1155, opened 2026-07-27
fix(commit-identity): scope indirect-exec matchers by heredoc TARGET (#1152)

### LINKED ISSUE #1152: bug(commit-identity): indirect-exec detector scans heredoc bodies it should not — documenting a bypass shape is blocked as if it were one
## Problem

`validate_commit_identity.check()` calls `_detect_indirect_commit(command, cwd=cwd)` on the **raw** command string, before anything strips heredoc bodies. `_find_commit_segment` does call `strip_heredocs`, but it runs *after* and is never reached when the indirect-exec detector has already blocked.

`_detect_indirect_commit`'s own matchers (`_PIPE_TO_SHELL_RE`, `_DASH_C_RE`, `_PROCESS_SUB_RE`, `_EVAL_RE`) scan the whole string. They do not distinguish a heredoc that is **fed to an interpreter** — a genuine bypass, and the reason `_HEREDOC_RE` exists — from a heredoc that is **fed to `cat`, redirected into a file**, whose body is inert text.

Result: writing a file whose *content* merely describes an interpreter invocation that mentions a commit is blocked as if it were that invocation.

## Reproduction (at `ec37f42`; `validate_commit_identity.py` is byte-identical to `main`)

Feeding each of these to the PreToolUse dispatcher as `tool_input.command`:

| shape | verdict | correct? |
|---|---|---|
| heredoc body containing a bare `git commit` phrase | allowed | yes |
| heredoc body containing an interpreter `-c` invocation with no commit in it | allowed | yes |
| heredoc body containing an interpreter `-c` invocation that mentions a commit, redirected to a file via `cat` | **BLOCKED** | **no — false positive** |
| heredoc fed directly to an interpreter | BLOCKED | yes |
| the identical content written via the `Write` tool instead of a shell heredoc | allowed | — inconsistent with row 3 |

The last two rows are the tell: the same bytes are inert through `Write` and blocking through a `cat` heredoc, and the thing that actually distinguishes a bypass from documentation — what the heredoc is fed to — is not consulted.

## Why it matters

It is in the fail-closed direction and it blocks real work, specifically the work of hardening these very hooks. Encountered three times in practice during the #1143 review: twice by the reviewer writing #1149's reproduction table, and once by the author writing the round-3b test fixtures. Each time the workaround was to construct the payload programmatically (string concatenation, or the `Write` tool) so the literal never appears in a Bash command — which is exactly the kind of contortion that makes people stop writing the test.

It also has an unpleasant second-order effect: the regression tests for #1149 need these literals, so this false positive makes #1149 harder to close.

## Not a regression

Pre-existing. `validate_commit_identity.py` last changed in #493 (implementing #482) and is untouched by #1143 — verified byte-identical to `main` at `ec37f42`. Filed from the #1143 review, not against it.

## Proposed fix

In `_detect_indirect_commit`, before running the pipe / dash-c / process-sub / eval matchers, remove heredoc bodies whose **command is not an interpreter**. Keep `_HEREDOC_RE` for the interpreter-fed case, which is a real bypass and must stay blocked.

Concretely: a heredoc opener belongs to a command; if that command is `cat` / `tee` / a redirect into a file, the body is data. If it is `bash` / `sh` / `zsh` / `dash` / `ksh`, the body is code. `_shell_parse.iter_command_segments_ast` already parses the command into a real AST that carries redirect nodes, which is the structurally correct place to make that call — the same primitive `_find_commit_segment` already prefers.

Regression tests should cover all five rows of the table above, including the `Write`-vs-heredoc consistency pair.

## Relationship to #1150

This is the same umbrella theme from the other side: a hand-rolled matcher in a hook, narrower (here: blunter) than what `_shell_parse` can answer, producing a wrong verdict. #1150's four instances are all under-matching; this one over-matches. Worth recording there as the first false-positive instance, since it shows the umbrella's cost is not only missed gates.

Found during review of #1143 (Weronika Zielinska, Platform Architect).

parametrization:
Closes #1152

## The defect

`_detect_indirect_commit` runs its seven matchers over the **raw** command string. None of them asks what a heredoc is fed to, so a heredoc body was scanned for bypass shapes regardless of whether a shell would ever execute it. `_find_commit_segment` *does* call `strip_heredocs`, but it runs after and is never reached once the indirect-exec detector has blocked.

## Five-row reproduction table, before and after

Run through the real `check()`; row 3 is the defect.

| # | shape | before | after |
|---|---|---|---|
| 1 | heredoc body with a bare `git commit` phrase, fed to `cat` | allow | allow |
| 2 | heredoc body naming an interpreter `-c` with no commit in it | allow | allow |
| 3 | heredoc body naming an interpreter `-c` **that mentions a commit**, fed to `cat` > file | **block** | **allow** |
| 4 | heredoc fed directly to an interpreter | block | block |
| 5 | the identical bytes via the `Write` tool | allow | allow |

Rows 3 and 5 are the pair that makes it visible: the same bytes were inert through `Write` and blocking through a `cat` heredoc. Confirmed at the hook's CLI entry point, not just in-process — row 3 goes `exit 2` -> `exit 0`, row 4 stays `exit 2`.

## The fix

The distinction the matchers need is a property of the heredoc's **target**: a body fed to `bash`/`sh`/`zsh`/`dash`/`ksh` is code and must stay scannable; a body fed to `cat`/`tee`, with or without a file redirect, is inert data.

`_shell_parse.classify_heredocs()` / `strip_data_heredocs()` — a quote-aware line scanner that pairs each heredoc opener with the pipeline segment that owns it, resolves the owner through `KEY=val` prefixes, transparent wrappers (`env`, `timeout 30`, `sudo`) and a leading path (`/bin/bash`), and follows `|` downstream. `_detect_indirect_commit` then matches against the stripped text instead of the raw string.

**Not weakened — the two guarantees:**

1. **Every ambiguity resolves toward CODE.** Unknown owner, unterminated heredoc, non-`\w` delimiter, untokenizable segment — all leave the body in place. A misparse can only preserve today's blocking behaviour, never open a hole.
2. **Data sinks are a two-entry ALLOWLIST** (`cat`, `tee`), not a denylist of interpreters. Growing it is the only way to unblock anything, and each addition is an explicit reviewable diff. `python3`, `sponge`, an unrecognised runner — all keep today's scan-the-body behaviour.

## Why not bashlex — measured, not asserted

The issue proposed `iter_command_segments_ast`, and it is the structurally right instinct: the AST carries redirect nodes and the hook already depends on it. **It cannot be used here.** The installed bashlex fails to parse a quoted-delimiter heredoc at all:

```
>>> bashlex.parse("cat <<'EOF'\nx\nEOF\n")
ParsingError: here-document at line 0 delimited by end-of-file (wanted "'EOF'")
```

Same for `<<"EOF"`. Only the bare `<<EOF` / `<<-EOF` forms parse. `<<'EOF'` is the dominant form in practice and is the form in the #1152 reproduction, so an AST-only classifier would not answer the question in the very case that motivated the issue. `BashlexHeredocLimitationTests` pins this measurement (including the `<<EOF`-parses-fine contrast that isolates *quoting* as the cause) so a future bashlex that fixes it fails loudly and invites this decision to be revisited, rather than the limitation being silently re-derived.

## A hole this also closes

The instrument surfaced a second, pre-existing defect in the opposite direction:

```
cat <<'EOF' | bash
git -c user.name=X -c user.email=Y commit -m z
EOF
```

was measurably **ALLOWED** before this change. `_HEREDOC_RE` requires the interpreter to appear textually *before* the `<<`, so it never matched; `_PIPE_TO_SHELL_RE` requires `printf`/`echo`. The body is piped into a shell and executed verbatim. The classifier already has to answer "is this body fed to an interpreter, including through a pipe?" in order to decide what to strip, so leaving the hole open would have meant the new code knows the answer and declines to use it. `CatPipedToShellTests` covers it, including the `| tee | bash` chain and the `| tee` -only case that must stay data.

## Shared interpreter set

`_INTERPRETERS` is now derived from `_shell_parse.SHELL_INTERPRETERS` rather than spelled out twice. This is a correctness requirement rather than tidying: if the set that decides which bodies get **stripped** drifted from the set that decides which shapes get **blocked**, a body could be dropped as data and then never offered to the matcher that considers it code. `test_interpreter_regex_matches_exactly_the_shared_set` and `test_every_interpreter_name_blocks` pin both halves.

## Verification

- **49-shape corpus** driven through the real `check()`: **19 wrong before** (17 false positives + 2 missed bypasses), **0 wrong after**.
- **74 new tests**; suite **3,073 passing** (2,999 baseline), no changes to existing tests.
- One test was written asserting the label `heredoc` for an interpreter heredoc whose body contains `bash -c`; running it showed the real label is `shell -c` (the `-c` matcher reaches the surviving body first). The test now asserts the measured label — blocking is the contract, which matcher claims it is not.
- ruff-format, ruff-lint, mypy, cspell, actionlint, the CI-sync drift gate, and both pre-commit hook stages: green locally.

## Coordination

Scoped to stay clear of #1151 (Lucas Ferreira), which edits `extract_leading_cd_target` / `resolve_invocation_cwd` in the same file. This change adds a new block adjacent to `strip_heredocs` (~line 450) and touches nothing in the `cd`-resolution chain; all new tests are in a **new** file, so there is no shared-test-file conflict surface either.

**#1149 is unblocked** — its regression tests can now carry interpreter `-c` invocation literals inside heredoc bodies without the `Write`-tool workaround.

## Review

Merge-gate reviewer: @Aino-Virtanen (Standards & Quality Lead). Owner-gated — not merging.

https://claude.ai/code/session_011hEQbrJGwa2kQ6dbEjyHQ6

### DIFF
--- a/.claude/hooks/_shell_parse.py
+++ b/.claude/hooks/_shell_parse.py
@@ -562,18 +562,40 @@ def _segment_head_command(seg_text: str) -> str | None:
     return head.rsplit("/", 1)[-1] if "/" in head else head
 
 
-def _scan_command_line(line: str) -> tuple[list[tuple[int, int, str]], list[tuple[int, bool, str]]]:
+class _LineScan(NamedTuple):
+    """Result of one quote-aware pass over a physical line."""
+
+    segments: list[tuple[int, int, str]]  # (start, end, separator_before)
+    openers: list[tuple[int, bool, str]]  # (position, is_dash_form, delimiter)
+    procsubs: list[int]  # positions of unquoted `<(` / `>(`
+
+
+def _scan_command_line(line: str) -> _LineScan:
     """Quote-aware single pass over one physical line.
 
-    Returns `(segments, openers)` where each segment is
-    `(start, end, separator_that_preceded_it)` and each opener is
-    `(position, is_dash_form, delimiter)`. Separators (`;`, `|`, `&`, `&&`,
-    `||`) and heredoc openers inside single/double quotes or behind a backslash
-    are DATA and are skipped, so `echo "a | b"` is one segment and
-    `echo "<<EOF"` holds no opener.
+    Separators (`;`, `|`, `&`, `&&`, `||`), heredoc openers and process
+    substitutions inside single/double quotes or behind a backslash are DATA and
+    are skipped, so `echo "a | b"` is one segment and `echo "<<EOF"` holds no
+    opener.
+
+    Two shapes must NOT be mistaken for control operators (main#1155 review, M2):
+
+      - `2>&1`, `>&2`, `<&0` — the `&` is part of a redirect, not a separator.
+        Splitting there resolved `cat 2>&1 > n.md <<'EOF'` to a head command of
+        `1`, so #1152's own false positive survived for every redirect-carrying
+        opener, and `cat <<'EOF' 2>&1 | bash` evaded the pipe-follow entirely.
+      - `&>file`, `&>>file` — the same `&`, on the other side.
+
+    Process-substitution positions are reported because `>(bash)` is a SECOND
+    sink hanging off the same segment: `tee >(bash) <<'EOF'` executes the body
+    even though the segment head is the inert `tee`. Recording them here (rather
+    than substring-searching the segment text later) keeps the quote-awareness —
+    a literal `>(` inside `cat > "weird>(name).md"` is not a process
+    substitution and must not be treated as one.
     """
     segments: list[tuple[int, int, str]] = []
     openers: list[tuple[int, bool, str]] = []
+    procsubs: list[int] = []
     quote: str | None = None
     seg_start = 0
     sep_before = ""
@@ -609,11 +631,19 @@ def _scan_command_line(line: str) -> tuple[list[tuple[int, int, str]], list[tupl
             i += 2
             continue
         two = line[i : i + 2]
+        if two in ("<(", ">("):
+            procsubs.append(i)
+            i += 2
+            continue
         if two in ("&&", "||"):
             segments.append((seg_start, i, sep_before))
             sep_before = two
             i += 2
             seg_start = i
+            continue
+        if c == "&" and (line[i - 1 : i] in ("<", ">") or line[i + 1 : i + 2] == ">"):
+            # `2>&1` / `>&2` / `&>log` — a redirect, not a control operator.
+            i += 1
             continue
         if c in (";", "|", "&"):
             segments.append((seg_start, i, sep_before))
@@ -623,32 +653,62 @@ def _scan_command_line(line: str) -> tuple[list[tuple[int, int, str]], list[tupl
             continue
         i += 1
     segments.append((seg_start, n, sep_before))
-    return segments, openers
+    return _LineScan(segments, openers, procsubs)
 
 
-def _opener_feeds_interpreter(line: str, pos: int, segments: list[tuple[int, int, str]]) -> bool:
+def _opener_feeds_interpreter(line: str, pos: int, scan: _LineScan) -> bool:
     """True if the heredoc opened at `pos` on `line` is consumed as shell code.
 
     A heredoc belongs to the pipeline segment it appears in. It is DATA only if
-    that segment's command is a known inert sink AND no segment downstream of it
-    *in the same pipeline* is a shell interpreter — `cat <<EOF | bash` feeds the
-    body to `bash` just as surely as `bash <<EOF` does. `&&`, `||`, `;` and `&`
-    break the pipeline: what follows them does not receive this stdin.
+    EVERY sink that can reach its bytes is inert:
+
+      1. the segment's head command is a known inert sink (`cat`, `tee`), AND
+      2. that segment carries no process substitution, AND
+      3. no segment downstream of it *in the same pipeline* is a shell
+         interpreter or itself carries a process substitution.
+
+    Condition 2 is the one that is easy to miss, and missing it is a live
+    bypass (main#1155 review, M1):
+
+        tee >(bash) <<'DELIM'
+        git -c user.name=X -c user.email=Y commit -m z
+        DELIM
+
+    The head word is the inert `tee`, but `>(bash)` is a SECOND sink hanging off
+    the same segment, and the body really is executed — confirmed under both
+    bash and zsh. Following only `|` never sees it, so the body was classified
+    as data, stripped, and the commit sailed through. A process substitution
+    anywhere in a reachable segment therefore forces CODE without trying to
+    resolve what is inside it: an unresolvable sink resolves toward CODE, which
+    is the rule the rest of this classifier already follows.
+
+    `&&`, `||`, `;` and `&` break the pipeline: what follows them does not
+    receive this stdin. A `&` that is part of a redirect (`2>&1`) is not a
+    separator at all — see `_scan_command_line`.
     """
+    segments = scan.segments
     idx = next(
         (k for k, (start, end, _sep) in enumerate(segments) if start <= pos < end),
         None,
     )
     if idx is None:
         return True
-    start, end, _sep = segments[idx]
-    if _segment_head_command(line[start:end]) not in HEREDOC_DATA_SINKS:
+
+    def _reachable_sink_is_code(k: int, *, head_must_be_inert_sink: bool) -> bool:
+        start, end, _sep = segments[k]
+        if any(start <= p < end for p in scan.procsubs):
+            return True
+        head = _segment_head_command(line[start:end])
+        if head_must_be_inert_sink:
+            return head not in HEREDOC_DATA_SINKS
+        return head in SHELL_INTERPRETERS
+
+    if _reachable_sink_is_code(idx, head_must_be_inert_sink=True):
         return True
     for k in range(idx + 1, len(segments)):
         if segments[k][2] != "|":
             break
-        nxt_start, nxt_end, _nxt_sep = segments[k]
-        if _segment_head_command(line[nxt_start:nxt_end]) in SHELL_INTERPRETERS:
+        if _reachable_sink_is_code(k, head_must_be_inert_sink=False):
             return True
     return False
 
@@ -674,9 +734,9 @@ def _classify_heredocs_cached(cmd: str) -> tuple[HeredocSpan, ...]:
     while i < len(lines):
         line = lines[i]
         i += 1
-        segments, openers = _scan_command_line(line)
-        for pos, dash_form, delim in openers:
-            is_code = _opener_feeds_interpreter(line, pos, segments)
+        scan = _scan_command_line(line)
+        for pos, dash_form, delim in scan.openers:
+            is_code = _opener_feeds_interpreter(line, pos, scan)
             body: list[str] = []
             term = None
             j = i
@@ -719,9 +779,9 @@ def strip_data_heredocs(cmd: str) -> str:
         line = lines[i]
         out.append(line)
         i += 1
-        segments, openers = _scan_command_line(line)
-        for pos, dash_form, delim in openers:
-            is_code = _opener_feeds_interpreter(line, pos, segments)
+        scan = _scan_command_line(line)
+        for pos, dash_form, delim in scan.openers:
+            is_code = _opener_feeds_interpreter(line, pos, scan)
             term = None
             j = i
             while j < len(lines):
--- a/.claude/hooks/_shell_parse.py
+++ b/.claude/hooks/_shell_parse.py
@@ -578,6 +578,14 @@ def _scan_command_line(line: str) -> _LineScan:
     are skipped, so `echo "a | b"` is one segment and `echo "<<EOF"` holds no
     opener.
 
+    `|&` is ONE operator — bash/zsh shorthand for `2>&1 |` — and must be
+    consumed as a pipe (main#1155 review round 2). Falling through to the
+    single-character cases split it into `|` then `&`, leaving an empty segment
+    between them and breaking the downstream walk on the `&`, so
+    `cat <<'D' |& bash` walked around the pipe-to-shell block. That was strictly
+    worse than either reading of the operator: treating it as a pipe catches the
+    shape, treating it as backgrounding would at least be consistent.
+
     Two shapes must NOT be mistaken for control operators (main#1155 review, M2):
 
       - `2>&1`, `>&2`, `<&0` — the `&` is part of a redirect, not a separator.
@@ -634,6 +642,19 @@ def _scan_command_line(line: str) -> _LineScan:
         if two in ("<(", ">("):
             procsubs.append(i)
             i += 2
+            continue
+        if two == "|&":
+            # bash/zsh shorthand for `2>&1 |` — ONE pipe operator, not `|`
+            # followed by a backgrounding `&`. Splitting it in two left an empty
+            # segment between them, and the downstream walk then broke on the
+            # `&` before ever reaching the interpreter, so `cat <<'D' |& bash`
+            # walked around the pipe-to-shell block this module adds. Emitting
+            # it as a `|` is both correct and the conservative reading: it keeps
+            # the pipeline connected, so the walk keeps looking for a sink.
+            segments.append((seg_start, i, sep_before))
+            sep_before = "|"
+            i += 2
+            seg_start = i
             continue
         if two in ("&&", "||"):
             segments.append((seg_start, i, sep_before))
--- a/.claude/hooks/_shell_parse.py
+++ b/.claude/hooks/_shell_parse.py
@@ -33,6 +33,21 @@ Public API
         word). Handles repeated/nested heredocs by iterating until the regex
         is fixed.
 
+    classify_heredocs(cmd) -> tuple[HeredocSpan, ...]
+        Every heredoc in `cmd` with the verdict on what CONSUMES it: a body fed
+        to `bash`/`sh`/`zsh`/`dash`/`ksh` (directly, or piped out of the opening
+        command) is `is_code=True`; a body fed to `cat`/`tee`, with or without a
+        file redirect, is `is_code=False` — inert text. Every ambiguity resolves
+        to `is_code=True`, so a misparse preserves a caller's blocking behaviour
+        rather than opening a hole (main#1152).
+
+    strip_data_heredocs(cmd) -> str
+        The complement of `strip_heredocs`: drops ONLY the `is_code=False`
+        bodies, keeping the ones a shell will execute. This is what a
+        bypass/gate matcher wants — scanning a `cat > notes.md <<'EOF' … EOF`
+        body for shell-injection shapes finds only documentation, while a
+        `bash <<'EOF' … EOF` body is genuinely executable and must stay visible.
+
     iter_command_segments(tokens) -> Iterator[list[str]]
         Splits a token list on the shell-control tokens `;`, `&&`, `||`, `|`
         (these survive shlex.split as their own tokens because they're not
@@ -244,7 +259,7 @@ import re
 import shlex
 import subprocess
 from functools import lru_cache
-from typing import Iterator
+from typing import Iterator, NamedTuple
 
 # Optional structural dependency (#748 D3b). bashlex gives a real Bash-AST
 # parse for the commit-identity matcher. It is imported defensively so a
@@ -441,6 +456,286 @@ def strip_heredocs(cmd: str) -> str:
         prev = cur
         cur = _HEREDOC_RE.sub("", cur)
     return cur
+
+
+# ---------------------------------------------------------------------------
+# Heredoc target classification (main#1152)
+# ---------------------------------------------------------------------------
+#
+# `strip_heredocs` answers "where are the heredoc bodies?" but not "who eats
+# them?", and every consumer that needs the second question has so far guessed.
+# `validate_commit_identity._detect_indirect_commit` ran its indirect-exec
+# matchers over the RAW command, so a heredoc body was scanned for bypass
+# shapes regardless of what the heredoc was attached to — documenting the shape
+#
+#     cat > notes.md <<'EOF'
+#     the bypass looks like: bash -c 'git commit …'
+#     EOF
+#
+# was blocked as if it WERE that invocation, while the identical bytes written
+# through the `Write` tool sailed through (main#1152).
+#
+# The distinction the matchers need is a property of the heredoc's TARGET:
+#
+#   - fed to `bash`/`sh`/`zsh`/`dash`/`ksh` (directly, or via a pipe out of the
+#     opening command) -> the body is CODE the shell will execute. A real
+#     bypass surface; must stay scannable.
+#   - fed to `cat`/`tee`, with or without a `>`/`>>` file redirect -> the body
+#     is DATA. Inert text; scanning it produces false positives only.
+#
+# Why not bashlex here, despite it being the structurally "right" tool and
+# already a dependency of this module: **bashlex cannot parse a quoted-delimiter
+# heredoc at all**. `bashlex.parse("cat <<'EOF'\nx\nEOF")` raises
+# `ParsingError: here-document ... delimited by end-of-file (wanted "'EOF'")`,
+# and the same for `<<"EOF"`. Only the bare `<<EOF` / `<<-EOF` forms parse.
+# `<<'EOF'` is the dominant form in practice (it is the form in the #1152
+# reproduction), so an AST-only classifier would not answer the question in the
+# very case that motivated it. This is a measured property of the installed
+# parser, not a style preference — see the regression test
+# `test_bashlex_cannot_parse_quoted_delimiter_heredoc`, which pins it so that a
+# future bashlex upgrade that fixes it will fail loudly and invite this comment
+# to be revisited.
+#
+# The classifier is therefore a quote-aware line scanner. Every ambiguity
+# resolves toward CODE (scan the body) so a misparse can only ever preserve
+# today's blocking behaviour, never open a hole:
+#
+#   - owner command not recognised, or unresolvable  -> CODE
+#   - heredoc never terminated                       -> CODE (and everything
+#                                                       after it is left alone)
+#   - delimiter not a bare `\w+` word                -> opener not recognised,
+#                                                       body treated as ordinary
+#                                                       command text (unstripped)
+
+# The POSIX shells whose heredoc body is executed as shell code. Single source
+# of truth: `validate_commit_identity` builds its `_INTERPRETERS` regex
+# alternation from this set, so the "which heredocs are code?" answer used by
+# the strip pass can never drift from the one used by the block pass. `mksh` /
+# `pdksh` stay out for the same conservative reason as #482.
+SHELL_INTERPRETERS = frozenset({"bash", "sh", "zsh", "dash", "ksh"})
+
+# Commands whose heredoc body is inert data: they copy stdin to stdout and/or a
+# file and never interpret it. Deliberately a tiny ALLOWLIST rather than a
+# denylist of interpreters — an unknown command keeps today's scan-the-body
+# behaviour, so growing this set is the only way to unblock anything and each
+# addition is an explicit, reviewable decision.
+HEREDOC_DATA_SINKS = frozenset({"cat", "tee"})
+
+# A heredoc opener: `<<EOF`, `<<'EOF'`, `<<"EOF"`, `<<-EOF`. Matched with
+# `.match(line, i)` from a scanner that has already excluded `<<<` (here-string)
+# and quoted regions, so no lookbehind is needed here.
+_HEREDOC_OPENER_RE = re.compile(r"<<(?P<dash>-?)[ \t]*(?P<q>['\"]?)(?P<delim>\w+)(?P=q)")
+
+
+class HeredocSpan(NamedTuple):
+    """One heredoc found in a command, with the verdict on what consumes it.
+
+    `body` excludes both the opener line and the terminator line. `is_code` is
+    True when the body is fed to a shell interpreter (directly or through a
+    pipe) — i.e. when scanning it for bypass shapes is correct. `terminated` is
+    False when no delimiter line was found before end-of-input.
+    """
+
+    delim: str
+    body: str
+    is_code: bool
+    terminated: bool
+
+
+def _segment_head_command(seg_text: str) -> str | None:
+    """Return the command name at the head of one pipeline segment, or None.
+
+    Heredoc openers are removed first (they are redirects, not words), then the
+    segment is tokenized and stripped of `KEY=value` prefixes and transparent
+    wrappers (`env`, `timeout 30`, `sudo`, …) via the shared helpers, so
+    `FOO=1 timeout 30 /bin/bash` resolves to `bash`. A leading path is reduced
+    to its basename. Returns None when the segment does not tokenize or holds no
+    word — callers must treat None as "unknown", never as "safe".
+    """
+    tokens = tokenize(_HEREDOC_OPENER_RE.sub(" ", seg_text))
+    if tokens is None:
+        return None
+    tokens = strip_command_prefixes(_strip_leading_env_assignments(tokens))
+    if not tokens:
+        return None
+    head = tokens[0]
+    return head.rsplit("/", 1)[-1] if "/" in head else head
+
+
+def _scan_command_line(line: str) -> tuple[list[tuple[int, int, str]], list[tuple[int, bool, str]]]:
+    """Quote-aware single pass over one physical line.
+
+    Returns `(segments, openers)` where each segment is
+    `(start, end, separator_that_preceded_it)` and each opener is
+    `(position, is_dash_form, delimiter)`. Separators (`;`, `|`, `&`, `&&`,
+    `||`) and heredoc openers inside single/double quotes or behind a backslash
+    are DATA and are skipped, so `echo "a | b"` is one segment and
+    `echo "<<EOF"` holds no opener.
+    """
+    segments: list[tuple[int, int, str]] = []
+    openers: list[tuple[int, bool, str]] = []
+    quote: str | None = None
+    seg_start = 0
+    sep_before = ""
+    i = 0
+    n = len(line)
+    while i < n:
+        c = line[i]
+        if quote is not None:
+            if c == "\\" and quote == '"' and i + 1 < n:
+                i += 2
+                continue
+            if c == quote:
+                quote = None
+            i += 1
+            continue
+        if c in ("'", '"'):
+            quote = c
+            i += 1
+            continue
+        if c == "\\" and i + 1 < n:
+            i += 2
+            continue
+        if line.startswith("<<<", i):
+            # Here-string, not a heredoc: consumed inline, opens no body.
+            i += 3
+            continue
+        if line.startswith("<<", i):
+            m = _HEREDOC_OPENER_RE.match(line, i)
+            if m is not None:
+                openers.append((i, bool(m.group("dash")), m.group("delim")))
+                i = m.end()
+                continue
+            i += 2
+            continue
+        two = line[i : i + 2]
+        if two in ("&&", "||"):
+            segments.append((seg_start, i, sep_before))
+            sep_before = two
+            i += 2
+            seg_start = i
+            continue
+        if c in (";", "|", "&"):
+            segments.append((seg_start, i, sep_before))
+            sep_before = c
+            i += 1
+            seg_start = i
+            continue
+        i += 1
+    segments.append((seg_start, n, sep_before))
+    return segments, openers
+
+
+def _opener_feeds_interpreter(line: str, pos: int, segments: list[tuple[int, int, str]]) -> bool:
+    """True if the heredoc opened at `pos` on `line` is consumed as shell code.
+
+    A heredoc belongs to the pipeline segment it appears in. It is DATA only if
+    that segment's command is a known inert sink AND no segment downstream of it
+    *in the same pipeline* is a shell interpreter — `cat <<EOF | bash` feeds the
+    body to `bash` just as surely as `bash <<EOF` does. `&&`, `||`, `;` and `&`
+    break the pipeline: what follows them does not receive this stdin.
+    """
+    idx = next(
+        (k for k, (start, end, _sep) in enumerate(segments) if start <= pos < end),
+        None,
+    )
+    if idx is None:
+        return True
+    start, end, _sep = segments[idx]
+    if _segment_head_command(line[start:end]) not in HEREDOC_DATA_SINKS:
+        return True
+    for k in range(idx + 1, len(segments)):
+        if segments[k][2] != "|":
+            break
+        nxt_start, nxt_end, _nxt_sep = segments[k]
+        if _segment_head_command(line[nxt_start:nxt_end]) in SHELL_INTERPRETERS:
+            return True
+    return False
+
+
+def classify_heredocs(cmd: str) -> tuple[HeredocSpan, ...]:
+    """Find every heredoc in `cmd` and say whether its body is code or data.
+
+    Bodies are located line-wise (the shell's own rule: openers are collected
+    across the command line, bodies follow in opener order). Returns an empty
+    tuple when the command holds no recognised heredoc.
+
+    Memoized (#1113 discipline): pure in `cmd` and the result is a tuple of
+    immutable NamedTuples, so the cached value is returned directly.
+    """
+    return _classify_heredocs_cached(cmd)
+
+
+@lru_cache(maxsize=256)
+def _classify_heredocs_cached(cmd: str) -> tuple[HeredocSpan, ...]:
+    lines = cmd.split("\n")
+    found: list[HeredocSpan] = []
+    i = 0
+    while i < len(lines):
+        line = lines[i]
+        i += 1
+        segments, openers = _scan_command_line(line)
+        for pos, dash_form, delim in openers:
+            is_code = _opener_feeds_interpreter(line, pos, segments)
+            body: list[str] = []
+            term = None
+            j = i
+            while j < len(lines):
+                candidate = lines[j].rstrip("\r")
+                if (candidate.lstrip("\t") if dash_form else candidate) == delim:
+                    term = j
+                    break
+                body.append(lines[j])
+                j += 1
+            if term is None:
+                found.append(HeredocSpan(delim, "\n".join(body), True, False))
+                return tuple(found)
+            found.append(HeredocSpan(delim, "\n".join(body), is_code, True))
+            i = term + 1
+    return tuple(found)
+
+
+@lru_cache(maxsize=256)
+def strip_data_heredocs(cmd: str) -> str:
+    """Remove the bodies of heredocs that are fed to a non-interpreter.
+
+    The complement of `strip_heredocs`, which removes EVERY heredoc body: this
+    keeps the bodies a shell will execute (so a bypass matcher still sees them)
+    and drops only the ones that are inert text. Opener and terminator lines are
+    left in place, so the surrounding command is unchanged and matchers anchored
+    on `<interpreter> … <<DELIM` still work.
+
+    Fails toward keeping the body: an unterminated heredoc, an unresolvable
+    owner, or an unrecognised delimiter shape all leave the text untouched.
+
+    Memoized (#1113): pure in `cmd`, returns an immutable `str`.
+    """
+    if "<<" not in cmd:
+        return cmd
+    lines = cmd.split("\n")
+    out: list[str] = []
+    i = 0
+    while i < len(lines):
+        line = lines[i]
+        out.append(line)
+        i += 1
+        segments, openers = _scan_command_line(line)
+        for pos, dash_form, delim in openers:
+            is_code = _opener_feeds_interpreter(line, pos, segments)
+            term = None
+            j = i
+            while j < len(lines):
+                candidate = lines[j].rstrip("\r")
+                if (candidate.lstrip("\t") if dash_form else candidate) == delim:
+                    term = j
+                    break
+                j += 1
+            if term is None:
+                out.extend(lines[i:])
+                return "\n".join(out)
+            out.extend(lines[i : term + 1] if is_code else [lines[term]])
+            i = term + 1
+    return "\n".join(out)
 
 
 def iter_command_segments(tokens: list[str]) -> Iterator[list[str]]:
--- a/.claude/hooks/tests/test_heredoc_target_scope.py
+++ b/.claude/hooks/tests/test_heredoc_target_scope.py
@@ -42,6 +42,7 @@ import validate_commit_identity as hook  # noqa: E402
 from _shell_parse import (  # noqa: E402
     HEREDOC_DATA_SINKS,
     SHELL_INTERPRETERS,
+    _scan_command_line,
     classify_heredocs,
     strip_data_heredocs,
 )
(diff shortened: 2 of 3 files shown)


parametrization:
**Review requested — Aino Virtanen (Standards & Quality Lead), merge gate.**

Three things worth aiming the instrument at, since each is a claim I settled by measurement rather than by reasoning about the parser:

1. **`bashlex` cannot parse `<<'EOF'`.** This is why the classifier is a scanner and not the AST walk the issue proposed. `BashlexHeredocLimitationTests` pins it, including the `<<EOF`-parses-fine contrast that isolates the delimiter *quoting* as the cause. If that measurement is wrong, the whole design choice is wrong — worth re-running.

2. **The fix is one-directional.** The claim is that every ambiguity resolves toward CODE (unknown owner, unterminated heredoc, non-`\w` delimiter, untokenizable segment) and that data sinks are a two-entry allowlist. `GenuineBypassStillBlocksTests` is the negative space — it re-asserts all seven bypass shapes plus the prefix-wrapper, absolute-path and env-assignment variants. The corpus behind it: 49 shapes through the real `check()`, 19 wrong before, 0 wrong after.

3. **`cat <<'EOF' | bash` was ALLOWED before this PR** and is blocked after. That is a new blocking path in a PR whose purpose is to *remove* a block, so it deserves scrutiny — my argument is in the PR body under "A hole this also closes".

Owner-gated: not merging. Blocks #1149, which is waiting on this.

parametrization:
**Merge-gate review — PR #1155 @ `e163320f6d2c5f1c680ea64679063297d1e6fddf`**

Reviewed against the PR's own merge-base `fea0dca` (confirmed ancestor of `main`), with every file re-fetched from `contents?ref=<sha>` and hash-matched against a detached worktree at that SHA before anything was executed. Every number below was re-measured, not read.

## Verdict

**ChangesRequested** — one genuine block -> allow regression on a shape a shell really executes. Everything else in this PR verified correct, and the false-positive removal itself is sound and well-guarded. The must-fix is 6 lines and breaks none of your 74 tests.

**Is the gate weakened?** In one specific direction, yes — and only one. Detail below.

---

## Must-fix 1 — process substitution as a heredoc sink: BLOCK -> ALLOW on an executable shape

```
tee >(bash) <<'DELIM'
DELIM
```

| | `fea0dca` | `e163320` |
|---|---|---|
| `tee >(bash)` | **BLOCK** | **ALLOW** |
| `tee >(sh)` | **BLOCK** | **ALLOW** |
| `cat <<'DELIM' > >(bash)` | ALLOW | ALLOW |

I confirmed the body really is executed, in both shells, before calling it a regression:

```
$ bash -c "tee >(bash) > /dev/null <<'DELIM' ... "   ->  PROCSUB-BODY-EXECUTED-BY-BASH
$ zsh  -c "tee >(bash) > /dev/null <<'DELIM' ... "   ->  PROCSUB-EXECUTED-ZSH
```

**Mechanism.** `_opener_feeds_interpreter` resolves the owning segment's head word to `tee`, finds it in `HEREDOC_DATA_SINKS`, then follows only `|` downstream. A process substitution is a sink attached to the *same* segment, not a downstream one, so the walk never sees it; `is_code=False`, the body is stripped, and `_HEREDOC_RE` — which at `fea0dca` matched on the `bash` inside `>(bash)` appearing before the `<<` — has nothing left to match.

**This is exactly the failure mode your guarantee 1 is written to prevent, and the guarantee does not quite cover it.** "Unknown owner -> CODE" holds for the *head word*. It does not hold for a segment whose head word is a known sink *and which also carries a second, unrecognised sink*. The classifier answers a narrower question than the one the caller is asking.

Your own `GenuineBypassStillBlocksTests` covers process substitution as an interpreter **argument** (`bash <(printf ...)`, line 310) and as inert prose (line 170), but there is no case anywhere in the 530-line file with `>(` in a heredoc **owner** position — so the 49-shape corpus could not have caught this.

**Prototype fix** (applied to a scratch copy, not to your branch):

```python
    start, end, _sep = segments[idx]
    if "<(" in line[start:end] or ">(" in line[start:end]:
        return True
    if _segment_head_command(line[start:end]) not in HEREDOC_DATA_SINKS:
        return True
```

Resolves toward CODE on an unresolvable sink, which is the rule you already stated.

## Must-fix 2 — `&` inside a redirect is treated as a control operator

Same patch, same function's neighbourhood. `_scan_command_line` splits on a bare `&`, but does not exclude `2>&1`, `>&2`, `&>`. Two consequences, measured:

| shape | `fea0dca` | `e163320` | correct |
|---|---|---|---|
| `cat <<'DELIM' 2>&1 \| bash` + commit body | ALLOW | **ALLOW** | BLOCK — the new gate is evaded by three characters |
| `cat 2>&1 > notes.md <<'DELIM'` + prose body | BLOCK | **BLOCK** | ALLOW — #1152's own false positive survives here |

Both come from one root cause: `2>&1` splits the segment, so the opener's owner resolves to `1` (row 2) or the downstream walk breaks on a non-`|` separator (row 1). Neither is a regression — both behave the same at base — but row 1 means Must-fix 1's sibling protection, the new `cat | bash` block you added, is trivially bypassable, and row 2 means the issue you are closing is not fully closed for redirect-carrying openers.

```python
        if c == "&" and (line[i - 1 : i] in ("<", ">") or line[i + 1 : i + 2] == ">"):
            i += 1
            continue
        if c in (";", "|", "&"):
```

**Both fixes together:** 6 added lines, full suite **3,073 passing** (your 74 included, none modified), and my 39-shape corpus goes to **zero wrong** except the two pre-existing holes I filed as tech-debt.

---

## What I verified independently, and what held

**The falsified fix direction — confirmed, exactly as reported.** bashlex `0.18` installed:

- `bashlex.parse("cat <<'DELIM'...")` -> `ParsingError: here-document ... delimited by end-of-file`
- same for the double-quoted delimiter
- bare `<<DELIM` and `<<-DELIM` both parse

The AST route was genuinely unavailable for the dominant form, and for the form in #1152's own reproduction. `BashlexHeredocLimitationTests` runs (it skips only on `ImportError`, and bashlex is installed), so a future bashlex that fixes this fails loudly rather than silently passing. Minor: `assertRaises(Exception)` would also pass on an unrelated error — `assertRaises(bashlex.errors.ParsingError)` would pin it tighter. Not blocking.

**The new blocking path — the hole was real.** `cat <<'DELIM' | bash` with a real commit body: ALLOW at `fea0dca`, BLOCK at `e163320`. I extended it and the new block holds across every variant I could construct: `| sudo bash`, `| env bash`, `| timeout 30 bash`, `| /bin/bash`, `| tee /tmp/x | sh`, `|bash` unspaced, tab before the pipe, `| bash -s`, `| ksh`, `| dash`, `FOO=1 cat`, `"cat"` quoted, all four delimiter forms, and a `\`-continuation before the pipe. It does not fire on `| tee` alone, on `&` backgrounding, on `;`-separated following commands, or on `ls`. The new path is correct and does not over-reach — apart from the `2>&1` gap above.

**The not-weakened proof — re-run and extended.** I built a 39-shape corpus of my own, 29 of which should block. Confirmed still blocking at head: interpreter heredocs (`bash`, `zsh`), all pipe-to-shell variants, a *second* heredoc in the same command being code while the first is data, an unterminated heredoc, the opener-line `&& bash -c ...` bypass, and the two unknown-sink cases (`python3 -`, `sponge`) that prove the allowlist really is an allowlist. Both structural guarantees hold in code, not just in prose — `HEREDOC_DATA_SINKS` is literally `frozenset({"cat", "tee"})` and unterminated resolves to `is_code=True`.

**Mutation-tested the guards**, because an assertion no fixture can trip is inert. All five caught:

| mutation | result |
|---|---|
| `_opener_feeds_interpreter` -> always DATA (max weakening) | **28 failed** |
| -> always CODE (revert to old behaviour) | **27 failed** |
| widen `HEREDOC_DATA_SINKS` to include `python3`, `sponge` | **2 failed** |
| delete the downstream-pipe follow | **4 failed** |
| unterminated heredoc -> DATA (fail-open) | **1 failed** |
| unmutated | 74 passed |

**#1149 unblocked — confirmed at the CLI entry point**, not in-process: row 3 `exit 2` -> `exit 0`, row 4 stays `exit 2`, rows 1 and 2 stay `exit 0`. I also hit the false positive first-hand: my own driver script, written as a `cat > file <<'DELIM'` heredoc, was blocked by the installed hook. Independent confirmation of the defect.

**The shared constant — the argument holds.** `_INTERPRETERS` evaluates to `(?:bash|dash|ksh|zsh|sh)`; `sh` sorts last so it cannot shadow `bash`/`dash`/`ksh`. No second executable spelling anywhere in `.claude/hooks` or `.claude/lib` — the remaining `bash|sh|zsh|dash|ksh` occurrences are all docstring prose. The drift argument is right: strip-set and block-set divergence would silently drop a body, which is precisely the class of bug in Must-fix 1, just via a different route.

**The label correction is right.** `shell -c` is the measured label because `_DASH_C_RE` runs ahead of `_HEREDOC_RE` and the interpreter-fed body survives stripping. Blocking is the contract; the label is diagnostic. Correcting the test to the measured value rather than the assumed one was the right call.

**CI.** 14/14 check-runs SUCCESS at `e163320` via `statusCheckRollup`; `mergeStateStatus: CLEAN`. Your legacy-status explanation is correct and I verified it with a control rather than accepting it: `commits/<sha>/status` on the already-**merged** `fea0dca` also returns `state: pending, total_count: 0`. The repo publishes zero legacy commit statuses; the check-runs are the real gates. Locally at head: ruff-format, ruff-lint, mypy all clean.

**"No existing test modified" — verified.** `git diff --stat` base..head: 3 files, `+869/-10`, and the only test file is the new one.

**#1156 coordination — semantically independent, verified, not assumed.** Same merge-base. Test-merged both heads: no conflicts, full suite **3,087 passing**. `tokenize`, `strip_command_prefixes`, `_strip_leading_env_assignments`, `strip_heredocs` and all four of your new functions are byte-identical between your head and the merged tree — Lucas's hunks land at ~1046-1407 (`resolve_tool_cwd` / `extract_leading_cd_target`), yours at ~458-740. I re-ran my 39-shape gate corpus against the merged tree: **identical verdicts** to your branch alone. No semantic interaction.

---

## Tech-debt filed

Two pre-existing holes found while constructing shapes outside your corpus. Both ALLOW at `fea0dca` **and** `e163320`, so neither is caused by this PR and neither blocks it — filed so they are tracked rather than presumed closed by the new classifier. Both belong under the #1150 umbrella.

- **#1167** — heredoc written to a file and executed later in the same command. `_read_script_if_safe` runs PreToolUse and cannot read a file the command has not created yet. Note the `.sh` variant appears to block at base, but only because `\bsh\b` matches inside the filename `s.sh`; rename to `s.txt` and base allows it too.
- **#1168** — a non-interpreter relay downstream of a data sink (`cat <<'DELIM' | xargs -I{} bash -c "{}"`) escapes the pipe-follow.

## Summary

The diagnosis is right, the falsification of the AST route is real and correctly documented, the false-positive removal is precise, the new blocking path closes a real hole, and the guards are live under mutation. The one thing that went wrong is the thing this kind of PR is most at risk of: closing a false positive moved one genuinely executable shape from blocked to allowed. Fix the two items above — the prototype is 6 lines and green — and this is an approve.

Please land the fix as an additive commit per the charter; I will re-verify at the new head.

---

Requestor: Aino Virtanen
Requestee: Weronika Zielinska
TechDebt: #1167, #1168

parametrization:
**Both must-fixes pushed — re-review requested @ `2782b3d`.** Aino Virtanen, merge gate.

Thank you for M1. You are right that it is a regression and right that it is the unrecoverable direction. I reproduced it before fixing rather than taking it on the verdict: materialised `.claude/hooks` at `fea0dca` into a scratch tree and ran the shapes there — `tee >(bash)` / `tee >(sh)` / `tee >(bash) > /dev/null` / `cat > >(bash)` all **BLOCK at base, ALLOW at `e163320`**. And I confirmed the body really executes, both shells:

```
$ bash procsub.sh -> PROCSUB-BODY-EXECUTED
$ zsh  procsub.sh -> PROCSUB-BODY-EXECUTED
```

Your diagnosis of the seam is exactly right and it is worth naming precisely, because it is the part I got wrong: my guarantee was written about the **head word**, and the failure needed a *known-inert head carrying an unrecognised second sink*. The corpus could not have caught it because every `>(` in my 530 lines was in argument or prose position, never in owner position. That is a hole in how I chose shapes, not just in the code.

## What I changed

**M1.** `_scan_command_line` now records unquoted `<(` / `>(` positions in the same quote-aware pass, and `_opener_feeds_interpreter` forces CODE for any *reachable* segment carrying one — without trying to resolve what is inside it.

Two deliberate deltas from your prototype:

1. **Applied to downstream piped segments too.** Your version guards the owner segment; `cat <<'DELIM' | tee >(bash)` reaches the body through the pipe-follow and needed the same rule. It was ALLOW with the owner-only guard.
2. **Position-recorded rather than substring-searched.** `"<(" in line[start:end]` would also fire on a literal `>(` inside a quoted filename — `cat > 'weird>(name).md' <<'DELIM'` — re-introducing #1152's false positive by the back door for that shape. Recording positions during the existing scan keeps the quote-awareness for free. Test: `test_quoted_procsub_is_not_a_procsub`.

**M2.** Adopted as you wrote it. `2>&1` / `>&2` / `&>log` no longer split. Confirmed both of your rows flip: the pipe-block is no longer evadable, and #1152's false positive is now fixed for redirect-carrying openers (`cat 2>&1 > n.md <<'DELIM'` was BLOCK at base *and* at `e163320` — it is ALLOW now). A genuine backgrounding `&` still separates, pinned directly on the scanner.

**Your non-blocking nit, taken.** `assertRaises(Exception)` -> `assertRaises(bashlex.errors.ParsingError)`. You were right that the broad form would keep passing on an unrelated failure while the measurement had quietly stopped being true.

## Verification

- **Corpus 49 -> 65 shapes**, 16 new `>(` / `2>&1` cases including the two deltas above: **0 wrong**.
- **Mutation-tested both new guards**, since you were right to insist on it for the first five: M1 deleted -> **7 failures**; M1 inverted to force DATA -> **7**; M2 deleted -> **6**. Unmutated 92 pass. Neither guard is inert.
- **92 tests** in this file (74 -> 92, none of the original 74 modified). Full suite **3,091 passing**. **14/14 CI check-runs SUCCESS** at `2782b3d`. ruff, mypy clean locally.

**#1167 and #1168 left alone** as instructed — both ALLOW at base and at head, neither caused by this PR.

Not merging — owner-gated, and it stays that way regardless of the verdict.

parametrization:
**Merge-gate re-review — PR #1155 @ `2782b3da4cbeb0f6861737f315115f30ada0efd9`**

Additive commit on top of `e163320` (both commits present, no force-push), same base `fea0dca`, so this is diffed against the same merge-base as round 1. Files re-fetched from `contents?ref=2782b3d` and hash-matched against a detached worktree before anything ran.

## Verdict

**Approved.** The regression is closed, and **both of your deltas from my prototype were right** — I verified them by running my own prototype head-to-head against yours rather than accepting the claim.

**Is the gate weakened in any direction? No.** Full base-vs-head differential below.

---

## Your two deltas — both confirmed, my prototype was wrong on both

**Delta 1: applying the guard to downstream piped segments, not just the owner.** Confirmed. My owner-only version leaves these ALLOW; yours blocks all four:

| shape | base `fea0dca` | my prototype | head `2782b3d` |
|---|---|---|---|
| `cat <<'DELIM' \| tee >(bash)` | ALLOW | **ALLOW** | **BLOCK** |
| `cat <<'DELIM' \| cat > >(sh)` | ALLOW | **ALLOW** | **BLOCK** |
| `cat <<'DELIM' \| tee /tmp/a \| tee >(bash)` | ALLOW | **ALLOW** | **BLOCK** |
| `cat <<'DELIM' \| tee /tmp/a 2>&1 \| tee >(zsh)` | ALLOW | **ALLOW** | **BLOCK** |

The last row is mine, not from your 65. My prototype closed the shape I happened to construct and not its piped sibling — the same corpus-design error I flagged in your 49, committed by me one round later while flagging it. Worth noting in the retro alongside your own item.

**Delta 2: recording positions in the quote-aware pass rather than substring-searching.** Confirmed, and the false positive is real:

| shape (prose body) | base | my prototype | head |
|---|---|---|---|
| `cat > 'weird>(name).md' <<'DELIM'` | BLOCK | **BLOCK** | **ALLOW** |
| `cat > "weird>(name).md" <<'DELIM'` | BLOCK | **BLOCK** | **ALLOW** |
| `cat > 'weird<(name).md' <<'DELIM'` | BLOCK | **BLOCK** | **ALLOW** |
| `tee 'a>(b).txt' <<'DELIM'` | BLOCK | **BLOCK** | **ALLOW** |

My substring test would have re-introduced #1152's own false positive by the back door for any filename containing `>(`. Yours fixes those shapes rather than preserving them. Recording during the existing pass is strictly the right call — it is the same quote-awareness the scanner already had, and I discarded it.

**Mutation-tested both deltas specifically**, since "my prototype was too narrow" is a claim about test coverage as much as code:

| mutation | result |
|---|---|
| M1 guard deleted | **7 failed** |
| M1 inverted (procsub forces DATA) | **7 failed** |
| **M1 narrowed to owner-only (my version)** | **1 failed** |
| M2 guard deleted | **6 failed** |
| M2 always-skip `&` | **1 failed** |
| **M1 reverted to substring search (my version)** | **1 failed** |
| unmutated | 92 passed |

Your first, second and fourth rows reproduce exactly. Both deltas are pinned by at least one test each — thin, but real, and my independent shapes above confirm the behaviour beyond the single fixture.

## M2 — both rows flip, as reported

| shape | base | `e163320` | `2782b3d` |
|---|---|---|---|
| `cat <<'DELIM' 2>&1 \| bash` + commit | ALLOW | ALLOW | **BLOCK** |
| `cat <<'DELIM' &>/dev/null \| bash` + commit | ALLOW | ALLOW | **BLOCK** |
| `cat <<'DELIM' >&2 \| bash` + commit | ALLOW | ALLOW | **BLOCK** |
| `cat 2>&1 > n.md <<'DELIM'` + prose | BLOCK | BLOCK | **ALLOW** |
| `cat &> n.md <<'DELIM'` + prose | BLOCK | BLOCK | **ALLOW** |
| `cat >&2 <<'DELIM'` + prose | BLOCK | BLOCK | **ALLOW** |

Genuine backgrounding still separates (`cat > n.md <<'DELIM' &` and `... & ls` both ALLOW correctly, and the `&&`/`||` check runs ahead of the redirect skip so short-circuits are untouched). The `i - 1` slice is safe at position 0 — it yields `""`, not a wraparound.

## Full differential, base vs this head

I re-ran round 1's 39-shape corpus plus a new 32-shape delta corpus against `fea0dca` and `2782b3d`.

**Block -> allow on an executable shape: none.** The only base-BLOCK/head-ALLOW row left is `cat > /tmp/s.sh <<'DELIM' … bash /tmp/s.sh`, which I established in round 1 is an accident of `\bsh\b` matching inside the filename — rename it `s.txt` and base allows it too. That is #1167, unchanged and correctly out of scope.

**Still blocking at head, re-confirmed:** interpreter heredocs (`bash`, `zsh`, `ksh`, `dash`), every pipe-to-shell wrapper form (`sudo`, `env`, `timeout 30`, `/bin/bash`, `bash -s`, unspaced, tab-separated, all four delimiter forms, `\`-continuation), owner-position procsubs (`tee >(bash)`, `tee >(sh)`, `tee >(bash) > /dev/null`, `tee <(bash)`, nested `tee >(tee >(bash))`, `2>&1`-carrying, dash-form, bare-delimiter), a second heredoc being code while the first is data, unterminated heredocs, the opener-line `&& bash -c …` bypass, and the unknown-sink cases (`python3 -`, `sponge`).

**Still allowing, correctly:** all prose-body forms, `| tee` alone, backgrounding, quoted filenames containing `>(`, `|`, `&>`, a `>(bash)` mentioned only inside a heredoc **body**, and `tee >(bash) <<'DELIM'` whose body holds no commit.

**One shape I labelled wrong and corrected rather than reporting.** I had `cat > n.md <<'DELIM'; tee >(bash) n.md` down as a missed block. It is not: `tee` reads stdin, not its file argument, so that command never executes the body — verified in a real shell. Head's ALLOW is correct. Its genuinely executable sibling is `cat > f <<'DELIM'; bash f`, which is #1167.

**One residual, pre-existing and not yours to fix here.** `tee >(gzip > log.gz) <<'DELIM'` with a prose body still blocks — an unresolvable sink forces CODE without inspecting what is inside it. It blocked at base too, so no regression, and it is the direction your guarantee 1 deliberately chooses. Rare enough not to file.

## Everything else re-verified at this head

- **92 tests** in the file, **3,091** in the full suite. Both reproduce.
- **"None of the original 74 modified"** — one line was, and it is the one I asked for: `assertRaises(Exception)` -> `assertRaises(bashlex.errors.ParsingError)`, plus a docstring explaining why. `git diff` `e163320..2782b3d` on the test file is `+139/-1` and that deletion is the assertion. Flagging only because the claim as phrased is very slightly stronger than the diff; the change itself is exactly right.
- **14/14 check-runs SUCCESS** at `2782b3d`; `mergeable_state: clean`. ruff-format, ruff-lint, mypy clean locally.
- **#1149 still unblocked**, and M2 does extend it to redirect-carrying openers.
- **#1156 re-verified at this head**, not carried over from round 1: clean test-merge, **3,105 passing** on the merged tree, and both my corpora produce **identical verdicts** merged vs. #1155 alone. No semantic interaction.
- **#1167 / #1168** untouched and still accurate.

## On your two retro items

Both are the right lesson and both generalise beyond this PR. The second is the sharper one: a want/got table can only find the bugs you thought of, while a base-vs-head differential finds the ones you did not — it is what surfaced the procsub regression in round 1, and it is what confirmed your deltas in round 2. I would support it as the default evidentiary form for any gate change, and I would add the first item's corollary: when a classifier's answer depends on a structural role, vary the role. I got that wrong myself this round, which is the best argument for the rule.

Clean work under adversarial review. Not merging — owner-gated.

---

Requestor: Aino Virtanen
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: #1167, #1168

parametrization:
**Merge-gate review (second slot, security lens) — PR #1155 @ `2782b3da4cbeb0f6861737f315115f30ada0efd9`**

Aino has already run the correctness and differential passes at this head, so I did not re-run them. I came at the two **structural claims** the safety argument rests on and tried to break them. Head SHA and merge-base (`fea0dca`) re-derived from `refs/pull/1155/head` myself; both trees materialised as detached worktrees before anything ran. Every decision below is the **CLI entry point** — `validate_commit_identity.py`, stdin JSON, exit code — not in-process behaviour.

## Verdict

**ChangesRequested** — one must-fix. A one-character operator walks around the new blocking path this PR adds. It is the same class Aino made a must-fix last round (`2>&1`, "evaded by three characters"), it is a real operator in **both bash and zsh**, and the fix is 7 lines that break none of your 92 tests.

Everything else I attacked held, including every construct on my list that I expected to get through.

---

## Method

Two independent oracles per shape, because the interesting question is not what the gate says, it is whether the gate is *right*:

- **EXEC** — run the shape in a real `bash` and a real `zsh` with a marker body. First version of this harness was wrong and I threw its results away: the marker string appeared verbatim in the body text, so every shape that merely *printed* the body scored as executed and I got 11 false regressions. The marker has to be assembled at runtime (`printf 'NINO-%s-MARKER\n' EXEC`) so it only materialises if a shell really runs the body.
- **GATE** — base vs head exit codes, run twice per shape: once with a body containing `bash -c 'git … commit'`, once with a **plain** identity commit.

That second variant is the one that matters and it is worth naming. With a `bash -c` body, base blocks *every* data heredoc via the raw text scan — that is #1152's false positive, so "base BLOCK -> head ALLOW" is the **intended** change and proves nothing. Only a plain-body base block is a principled block. Several shapes that looked like regressions on the first reading collapse under this test.

32 shapes across 9 families.

## Q1 — "every ambiguity resolves toward CODE": can anything else attach an interpreter to an allowlisted-sink segment?

**No. I attacked 9 families and none got through the owner-position check.**

| construct | executes body? | head |
|---|---|---|
| `tee "$(echo '>(bash)')" <<'D'` — cmdsub expanding to a procsub string | **inert** | ALLOW, correct |
| `cat <<'D' > /dev/null $(bash)` / backtick form | **inert** | ALLOW, correct |
| `cat <<'D' > "$(echo /dev/stdout)"` — cmdsub in redirect target | **inert** | ALLOW, correct |
| `exec cat <<'D'` | **inert** | ALLOW, correct |
| `exec 0<<'D'` | inert | BLOCK (fail-closed) |
| `{ cat; } <<'D'` | inert | BLOCK (fail-closed) |
| `( bash ) <<'D'` | EXEC | **BLOCK** |
| `while read l; do bash -c "$l"; done <<'D'` | EXEC | **BLOCK** |
| `tee >(tee >(bash)) <<'D'` — nested procsub | EXEC | **BLOCK** |
| `tee /dev/fd/1 <<'D' \| bash` | EXEC | **BLOCK** |
| `coproc CP { bash; }` + `cat <<'D' >&${CP[1]}` | inert | ALLOW, correct |

Worth stating explicitly since it was called out as zsh-specific: **`=(…)` is not a hole.** `tee =(bash) <<'D'` and `cat <<'D' > =(bash)` are both **inert in real zsh** — `=(cmd)` captures cmd's *stdout* into a temp file, it does not feed the heredoc into cmd. So `=( ` being absent from `_scan_command_line` is correct, not an oversight. The zsh construct that *does* execute, MULTIOS `cat <<'D' > f1 > >(bash)`, is EXEC in zsh and **BLOCK** at head — the procsub recording catches it.

The general reason the owner position is solid: `_reachable_sink_is_code(idx, head_must_be_inert_sink=True)` returns `head not in HEREDOC_DATA_SINKS`, so *unknown* -> CODE. Compound leaders, subshells and `done`/`}` heads all resolve to something outside the two-entry set and fail closed. Command substitution can't be a vector at all, because expansion happens before the heredoc redirect is wired, so the subshell inherits the shell's stdin rather than the body — measured, not assumed.

## Q2 — "data sinks are a two-entry allowlist": is it an allowlist everywhere?

**No — and this is the answer to the question as asked.** The path has two branches with opposite failure directions:

```python
if head_must_be_inert_sink:
    return head not in HEREDOC_DATA_SINKS   # owner: ALLOWLIST, unknown -> CODE
return head in SHELL_INTERPRETERS           # downstream: DENYLIST, unknown -> stays DATA
```

The important qualifier, in the PR's favour: the downstream walk runs only *after* the owner check has already said "inert", and it can only ever **upgrade** a verdict to CODE — it can never turn a code body into data. So the allowlist is genuinely mandatory for reaching the data classification, and there is no branch that bypasses it. I checked the other routes too: unterminated -> `is_code=True`; a non-`\w` delimiter produces no span at all, so the body stays ordinary scannable text; and across the whole corpus `classify_heredocs` and `strip_data_heredocs` never disagree — **0 cases of a CODE body being dropped before the matchers see it**.

What the denylist costs is *failure to upgrade*, for any relay the walk doesn't recognise. That is #1168 (`xargs bash -c`), and `busybox sh` is the same shape. Both are ALLOW at base and head — pre-existing, not yours, not re-filed.

But two shapes reach that same fail-open branch through a **parser** defect rather than an unrecognised relay, and those are new.

---

## Must-fix — `|&` evades the new pipe-to-shell block

```
cat <<'D' |& bash
git -c user.name="X" -c user.email="x@y.z" commit -m z
D
```

`|&` is bash/zsh shorthand for `2>&1 |`. The body is executed — hand-verified, both shells:

```
$ bash s.sh -> NINO-EXEC-MARKER
$ zsh  s.sh -> NINO-EXEC-MARKER
```

| shape (plain identity body) | base | head | with 7-line fix |
|---|---|---|---|
| `cat <<'D' \| bash` (control) | ALLOW | **BLOCK** | BLOCK |
| `cat <<'D' \|& bash` | ALLOW | **ALLOW** | **BLOCK** |
| `tee /dev/null <<'D' \|& sh` | ALLOW | **ALLOW** | **BLOCK** |

**Not a regression** — base allowed it too. It is a gap in the protection *this PR introduces*, and it is exactly the shape of the `2>&1` item from round 1: the new gate is real, and it is evaded by one character.

**Mechanism.** `_scan_command_line` has no case for `|&`, so it splits at `|`, then splits again at `&`, leaving an **empty segment** between them:

```
sep=''  seg="cat <<'D' "  head='cat'
sep='|' seg=''            head=None      <- empty segment, not an interpreter
sep='&' seg=' bash'       head='bash'    <- never reached: walk breaks on '&'
```

The downstream walk evaluates the empty segment (`None in SHELL_INTERPRETERS` -> False), then hits separator `&` and breaks. `bash` is never examined. Note this is strictly worse than either interpretation: treating `|&` as a pipe would catch it, and treating it as pure backgrounding would at least be consistent — the current split does neither.

**Fix**, alongside the `&&`/`||` case you already handle two-char-first:

```python
        if two == "|&":
            # bash/zsh shorthand for `2>&1 |` — a PIPE, not `|` followed by `&`.
            segments.append((seg_start, i, sep_before))
            sep_before = "|"
            i += 2
            seg_start = i
            continue
```

**Verified on a scratch copy of your head:** all three rows above flip to BLOCK; genuine backgrounding (`cat <<'D' & bash`) and `||` still break the pipeline correctly; the #1152 false positives stay ALLOW; **92/92** in your file and **3,091** full suite still pass, unchanged.

That last number is the part I would flag even if you disagree about severity: **the fix passes your suite untouched, which means nothing currently pins `|&` in either direction.** Whichever way it is resolved, it wants a test.

---

## Confirmations requested

**The false positive this PR exists to fix — confirmed fixed**, at the CLI, including the redirect-carrying opener:

| shape, prose body documenting an interpreter invocation | base | head |
|---|---|---|
| `cat > notes.md <<'D'` | BLOCK (exit 2) | **ALLOW** (exit 0) |
| `cat 2>&1 > n.md <<'D'` | BLOCK | **ALLOW** |

I also hit it first-hand and unplanned: my `gh api` command to file one of the tech-debt issues below was itself a `cat > file <<'EOF'` heredoc whose body documented this shape, and the **installed** hook blocked it. Independent confirmation of the defect, same as you and Aino both reported.

**#1149 unblocked at the CLI entry point — confirmed.** Every row in this review is a subprocess exit code, never in-process. Both FP rows go 2 -> 0; the genuine bypasses stay 2.

**`_INTERPRETERS` / `SHELL_INTERPRETERS` — confirmed.** Evaluates to `(?:bash|dash|ksh|zsh|sh)`, `sh` last; `re.match` against each of the five returns the full word, so nothing is shadowed. No second executable spelling in `.claude/hooks` or `.claude/lib` — the remaining `bash|sh|zsh|dash|ksh` hits are all docstring prose. And a data-classified body cannot be silently dropped ahead of the code matcher: both consumers call the same `_opener_feeds_interpreter`, 0 disagreements measured, and shape 4 additionally re-derives the verdict from the raw command.

**Test counts reproduce** at this head: 92 in `test_heredoc_target_scope.py`, 3,091 full suite. 14/14 check-runs SUCCESS, `mergeable_state: clean`.

## Tech debt filed

- **#1170** — heredoc reaching an interpreter through a **FIFO created in the same command** (`mkfifo p; bash < p & cat <<'D' > p`). Body executes in both shells, and this one is **BLOCK at base -> ALLOW at head with a plain identity body** — a genuine coverage loss, and distinct from #1167, whose base block is the `\bsh\b`-in-filename accident. Filed rather than gated because base's block came from "an interpreter appears earlier on the opener line", which is precisely the over-broad rule #1152 removes on purpose; closing it properly needs path-aware dataflow, not a wider sink set. Please read it before assuming #1167 covers it — it doesn't.
- **#1171** — backslash line-continuation between the sink and `| bash`. Executes; ALLOW at base and head, so another gap in the new block path. Not gated: unlike `|&` it needs physical-to-logical line folding with correct body offsets, which deserves its own tests.

Neither could be added to project 2 — `gh project item-add` is GraphQL and our quota is zero.

## Summary

The two structural claims came out of this in good shape. Guarantee 1 held against every owner-position construct I could build, including the zsh-specific ones, and the allowlist really is mandatory on the only path that reaches the data verdict. The residual risk is entirely in the downstream walk's opposite failure direction, and most of that is pre-existing and already tracked.

The one thing I am gating on is small and concrete: the protection this PR adds is defeated by `|&` in the repo's own default shell, no test covers it, and the fix is seven lines that change nothing else. Fix that and I approve on sight.

Not merging — owner-gated.

---

Requestor: Nino Kavtaradze
Requestee: Weronika Zielinska
TechDebt: #1170, #1171 (pre-existing, not this PR: #1167, #1168)

parametrization:
**Must-fix pushed — re-review requested @ `6bc67a4`.** Nino Kavtaradze, second slot.

`|&` fix taken as you wrote it. Reproduced before applying it rather than on the verdict — the segmentation is exactly what you described:

```
sep=''   seg="cat <<'D' "
sep='|'  seg=''
sep='&'  seg=' bash'
```

and after the fix, two segments with one `|`.

The argument that decided it for me was not the severity, it was your observation that the split was **strictly worse than either reading of the operator**. A gate that treats `|&` as a pipe catches the shape; one that treats it as backgrounding is at least consistent; the two-way split produced an empty segment and then broke the walk on the `&`, which is neither. That is a parser defect regardless of what it lets through.

**Execution re-confirmed with your runtime-assembled-marker method** — the note about throwing away your first harness run is a good catch and I used it rather than a literal marker:

```
cat <<'D' |& bash
printf 'WZ-%s-MARKER\n' EXEC
D

$ bash -> WZ-EXEC-MARKER      $ zsh -> WZ-EXEC-MARKER
```

**Base-vs-head differential with a PLAIN identity body**, per your point that a `bash -c` body makes base block every data heredoc via the raw scan so a base-BLOCK there proves nothing. Your table reproduces:

| shape (plain identity body) | base | head `6bc67a4` |
|---|---|---|
| `cat <<'D' \| bash` (control) | allow | **block** |
| `cat <<'D' \|& bash` | allow | **block** |
| `tee /dev/null <<'D' \|& sh` | allow | **block** |
| `cat > n.md <<'D'` (FP control) | allow | allow |

**Mutation-tested the new guard**, because your sharpest line was that the fix passing my suite untouched meant *nothing pinned `|&` in either direction*: guard disabled -> **5 failures**. It is pinned now, in both directions — the negative space is pinned too, since a fix that made `|&` mean "always code" would have re-introduced #1152 for the relay shape:

- bare `&` still backgrounds and still breaks the pipeline
- `||` is still one operator and still breaks it
- `cat <<'D' |& tee /tmp/a` with no interpreter anywhere stays **ALLOW**

Also covered: `|&bash` unspaced, and `|& tee /tmp/a | bash` so the walk continues past a `|&` relay rather than only handling it in terminal position.

**Numbers:** corpus 65 -> **71 shapes, 0 wrong**. **101 tests** in this file (92 -> 101), full suite **3,100 passing**. **14/14 CI check-runs SUCCESS** at `6bc67a4`, `mergeable_state: clean`. ruff, mypy clean.

**#1170 and #1171 left alone.** I read #1170 rather than assuming #1167 covers it — you are right that it is distinct, and right that its base block comes from "an interpreter appears earlier on the opener line", which is the over-broad rule #1152 removes on purpose. Closing it properly needs path-aware dataflow, not a wider sink set. Agreed it is not gateable here.

Two things from your review I want to record for the retro, because they are better than what I did:

- **Two oracles per shape.** I verified execution for the procsub finding but scored the gate against *intent*, not against base. Your EXEC/GATE split — is the body executed, AND what did base vs head decide — is the form a "not weakened" claim actually needs.
- **The plain-body control.** This is the one I would not have found. With a `bash -c` body, base blocks every data heredoc, so "base BLOCK -> head ALLOW" reads as a regression when it is the intended fix. Every differential on this gate needs the plain-body variant to mean anything.

Not merging — owner-gated.

parametrization:
**Merge-gate re-review (second slot, security lens) — PR #1155 @ `6bc67a4419802dd3df48874580fd3d300cd6e7ef`**

Additive commit on top of `2782b3d` (both prior commits still present, no force-push), same merge-base `fea0dca`. Three detached worktrees materialised at `fea0dca` / `2782b3d` / `6bc67a4` and every decision below re-measured at the CLI entry point — stdin JSON, exit code.

## Verdict

**Approved.** The `|&` fix is correct, and — the part I care about more — it closes the bypass **without over-classifying**, which was the way this fix could have gone wrong.

## The risk I was actually reviewing for

A fix that closes `|&` by making it *always* mean CODE would have resurrected #1152 for every `|&` relay shape. That is the cheap wrong answer here, and it would have been invisible in a corpus that only tests the bypass direction. So I ran the differential in the direction that would expose it: `2782b3d` -> `6bc67a4`, hunting **ALLOW -> BLOCK on inert shapes**.

**New false positives: NONE. Weakened shapes: NONE.** Across 31 new shapes plus a re-run of round 1's 32.

The negative space, measured — every one of these is inert in a real shell and stays ALLOW at the new head:

| shape (unchanged ALLOW at `6bc67a4`) | |
|---|---|
| `cat <<'D' \|& tee /tmp/a` | the relay you flagged — no interpreter, stays data |
| `cat <<'D' \|& grep foo > out.md` | |
| `tee /tmp/z <<'D' \|& cat` | |
| `cat > 'weird\|&name.md' <<'D'` | quoted `\|&` in a filename is not an operator |
| `tee 'a\|&b.txt' <<'D'` | |
| `cat <<'D' & bash` / `\|\| bash` / `&& bash` / `; bash` | separators still break the pipeline |
| `cat 2>&1 > n.md <<'D'` / `cat &> n.md <<'D'` | M2 redirect handling untouched |

And the bypass direction — all ten **EXEC in both bash and zsh**, all ALLOW at `2782b3d`, all **BLOCK** at `6bc67a4`:

`cat <<'D' |& bash` · `|&bash` unspaced · `|&   bash` extra space · tab before `|&` · `tee /dev/null <<'D' |& sh` · `|& env bash` · `|& /bin/zsh` · `cat <<'D' |& tee /tmp/a | bash` · `cat <<'D' | tee /tmp/a |& bash` · `cat <<'D' |& tee >(bash)`

Both shapes I was asked to attack specifically are covered: the **unspaced** form blocks, and the **`tee` relay** behaves correctly in *both* directions — `|& tee /tmp/a` alone stays ALLOW (no over-classification) while `|& tee /tmp/a | bash` blocks (the walk continues past the relay). That pair is the whole ballgame and it is right.

## Mutation-tested, because last round I said nothing pinned this

I made the point that the fix passing your suite untouched meant `|&` was unpinned in either direction, so I re-ran that instrument rather than accept the claim:

| mutation | result |
|---|---|
| guard deleted (revert to `2782b3d`) | **5 failed** |
| `|&` emitted as backgrounding `&` instead of `\|` | **5 failed** |
| **`|&` forced to always mean CODE (over-classify)** | **1 failed** |
| unmutated | 101 passed |

Your 5 reproduces exactly. The third row is the one I added and the one that answers the review question: the negative space **is** pinned — an over-classifying fix does not pass this suite. One test is thin, but it is real, and my 13-shape negative corpus confirms the behaviour well beyond that single fixture. `test_pipe_ampersand_to_a_data_sink_stays_data` is the fixture doing that work; it is the most valuable test in the commit.

## The implementation

Emitting `|&` as a single `|` is the right call and for the right reason — it keeps the pipeline connected so the walk continues hunting for a sink, rather than short-circuiting to a verdict. Ordering is correct: the two-char `|&` case is tested before the single-char `;`/`|`/`&` branch, and it cannot capture `||` (that is `two == "||"`, a different literal) or the M2 redirect forms (`2>&1` presents `two == "&1"`). The comment explains the empty-segment mechanism accurately.

You have the reason I raised it right, too, and I want to confirm that explicitly since it is the generalisable part: the argument was **not** severity. `|&` is not exotic, but that is not why it had to be fixed. The old scanner produced a segmentation that was wrong under *every* reading of the operator — worse than either consistent choice — and a parser defect of that shape is a bug independent of what happens to slip through it today. Judging it by what it currently lets past would have meant re-litigating it the next time someone found a new consequence.

## Everything else re-verified at this head

- **101 tests** in the file (92 -> 101), full suite **3,100**. Both reproduce.
- **14/14 check-runs SUCCESS** at `6bc67a4`; `mergeable_state: clean`.
- **The #1152 false positives still ALLOW** — `cat > notes.md <<'D'`, `cat 2>&1 > n.md <<'D'`, `cat > 'weird>(name).md' <<'D'`. Unchanged by this commit.
- **Round 1's guards intact** — owner-position procsub (`tee >(bash)`), direct `bash <<'D'`, `| bash`, and unterminated-fails-closed all still BLOCK.
- **Your terminator self-correction is right, and it is a code-behaviour claim so I checked it in a real shell rather than by reading.** `D || ls` is not a valid terminator; bash itself says so — `warning: here-document at line 1 delimited by end-of-file (wanted 'D')`. The classifier reports `terminated=False, is_code=True`, so it fails closed and blocks. Fixing the expectation rather than the code was correct; changing the code there would have been a fail-open.
- **#1170 / #1171 untouched**, as scoped. Thank you for reading #1170 rather than assuming #1167 covered it — the distinction matters precisely because #1167's ALLOW/ALLOW evidence would have hidden it.

## Summary

Two rounds, two real findings, both closed without collateral. What makes this one land is that the fix is narrow in the direction that counts: it recovers ten executable shapes and moves nothing that was correctly allowed. The gate is now strictly stronger than base on every shape either of us has constructed, with the residual risk confined to the downstream walk's fail-open branch — all of it pre-existing and tracked in #1167, #1168, #1170, #1171.

Not merging — owner-gated.

---

Requestor: Nino Kavtaradze
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: #1170, #1171 (pre-existing, not this PR: #1167, #1168)

parametrization:
**Merge-gate re-review (round 3) — PR #1155 @ `6bc67a4419802dd3df48874580fd3d300cd6e7ef`**

My round-2 `Approved` was cast at `2782b3d` and is correctly excluded as stale — the head moved after it, and production code changed. This is a fresh review of `2782b3d..6bc67a4`, not a re-stamp. Trees materialised with `git archive <sha>` and hash-matched against `contents?ref=6bc67a4`; linting and suites run in a worktree pinned to that SHA.

## Verdict

**Approved.** The `|&` fix is correct, the bypass it closes was real, and the negative space is genuinely pinned rather than asserted.

**Is the gate weakened in any direction? No.** Zero new false positives against the previous head; zero block -> allow on any executable shape versus base.

---

## The bypass was real — confirmed before checking the fix

I ran the shape in a real shell rather than reasoning about the operator:

```
$ bash -c "cat <<'D' |& bash …"   ->  PIPE-AMP-EXECUTED-BASH
$ zsh  -c "cat <<'D' |& bash …"   ->  PIPE-AMP-EXECUTED-ZSH
```

`|&` is bash/zsh shorthand for `2>&1 |`, so the body really is piped into a shell. Eleven `|&` shapes measured across all three SHAs — every one **ALLOW at base, ALLOW at `2782b3d` (the head I approved), BLOCK at `6bc67a4`**: `|& bash`, `|& sh`, `|&` unspaced, `|& tee /tmp/a \| sh`, `tee /tmp/a <<'D' |& bash`, `|& tee /tmp/a |& zsh`, `|& tee >(bash)`, `|& sudo bash`, `|& /bin/dash`, bare-delimiter, dash-form.

The mechanism in the diff matches the one described: falling through to the single-character cases split `|&` into `|` then `&`, leaving an **empty segment** between them, and the downstream walk broke on the `&` before reaching the interpreter. Strictly worse than either reading of the operator — I confirmed that by mutating the fix to consume only the `|` (the pre-fix shape) and watching 5 tests go red.

**I missed this in round 2.** My corpus varied wrapper, path, quoting, delimiter form and redirect shape around the pipe, and never varied the pipe **operator** itself. That is the same "vary the structural role, not just the token" lesson from round 2, one level up: I had treated `|` as a fixed frame rather than as a thing with variants. Nino was right to look there.

## The negative space — the part that matters most

The cheap wrong fix is to make `|&` mean CODE unconditionally, which would resurrect #1152 for every relay shape. It does not:

| inert shape | base | `2782b3d` | `6bc67a4` |
|---|---|---|---|
| `cat <<'D' \|& tee /tmp/a` | BLOCK | ALLOW | **ALLOW** |
| `cat <<'D' \|& cat > /tmp/n.md` | BLOCK | ALLOW | **ALLOW** |
| `cat <<'D' \|& grep foo` | BLOCK | ALLOW | **ALLOW** |
| `cat <<'D' \|& tee /tmp/a \|& tee /tmp/b` | BLOCK | ALLOW | **ALLOW** |
| `cat > 'weird\|&name.md' <<'D'` | BLOCK | ALLOW | **ALLOW** |
| `\|&` appearing only inside a heredoc body | ALLOW | ALLOW | **ALLOW** |
| `\|& bash` with no commit in the body | ALLOW | ALLOW | **ALLOW** |

The decisive pair reproduces: `|& tee /tmp/a` stays ALLOW, `|& tee /tmp/a \| sh` blocks. I added the reverse ordering (`|& tee /tmp/a |& zsh`) and it blocks too. **Zero new false positives** against `2782b3d` across all 31 shapes.

**Mutation-tested, and the negative space is pinned by a fixture:**

| mutation | result |
|---|---|
| `\|&` guard deleted | **5 failed** |
| `\|&` treated as backgrounding | **5 failed** |
| `\|&` forced always-CODE (the cheap wrong fix) | **1 failed** |
| `\|&` consuming only the `\|` (the pre-fix bug) | **5 failed** |
| unmutated | 101 passed |

Nino's three rows reproduce exactly; the fourth is mine. The always-CODE row is the one that matters — a fixture, not an assertion, stands between this fix and a reintroduced #1152.

## Terminator self-correction — confirmed in a real shell

A line reading `D || ls` does **not** terminate a heredoc. I ran it: the line was emitted as body content and only the bare `D` closed the block. So the classifier reading such a heredoc as unterminated, and failing closed to CODE, is correct shell semantics. Fixing the expectation rather than the code was right. `test_or_operator_is_untouched_by_the_pipe_amp_case` separately pins that `||` is not mis-consumed by the new `|&` branch, which is the ordering hazard this change introduces; `&&`, `||`, `;` and `&` all still break the pipeline in my corpus.

## Re-verified at this head, not carried over

- **101 tests** in the file (92 -> 101), test delta `+73/-0` — **purely additive**, the original 92 untouched. That claim is exact this time.
- **Full suite 3,100 passing.**
- **14/14 check-runs SUCCESS** at `6bc67a4`, verified at the SHA myself; `mergeable_state: clean`.
- ruff-format, ruff-lint, mypy clean. My first local lint run reported failures — that was an artifact of my own `git archive` extraction leaving a stray empty `mypy.ini`/`ruff.toml` shadowing the real `pyproject.toml`, not a finding. Re-run in a worktree pinned to the SHA: clean.
- **Round 1 and round 2 corpora both re-run at this head.** All earlier fixes hold: owner-position procsubs, pipe-reachable procsubs, `2>&1 |`, quoted `>(` filenames, interpreter heredocs, plain `| bash`. The only base-BLOCK/head-ALLOW row remains `cat > /tmp/s.sh … bash /tmp/s.sh`, which is the incidental `\bsh\b`-in-filename match established in round 1 — that is #1167, unchanged.
- **#1167 / #1168** still accurate and still out of scope.
- **#1156 re-verified at this head:** clean test-merge, **3,114 passing** on the merged tree, and my `|&` corpus gives identical verdicts merged vs. #1155 alone. No semantic interaction.

## Assessment

Three rounds, three genuine defects found by three different readers, each in a place the previous round's corpus had held fixed. Round 1 varied the sink and found the process substitution; round 2 varied the segment role and found the piped sibling; round 3 varied the operator and found `|&`. None was reachable from the round before it, which is a fair account of why this took three passes rather than evidence that any pass was careless. The code that lands is materially stronger than what round 1 reviewed, and stronger than base in both directions — it removes #1152's false positive and closes four separate holes that predate this PR.

Not merging — owner-gated.

---

Requestor: Aino Virtanen
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: #1167, #1168

Issue #1141, opened 2026-07-27
bug(kickoff-hook): comment hardcodes a wave-branch base regardless of merge_model, and a loop-applied label evades the hook silently


parametrization:
## Problem 1 — the kickoff template ignores the wave's merge model

`render_kickoff_comment` builds the branch instruction unconditionally:

```python
branch_base = f"deployments/phase-{phase_num}/wave-{wave_num}"
```

`merge_model` is never read — the string `merge_model` does not appear anywhere in the hook.

On a **`direct-to-main`** wave there is no wave branch, so every kickoff comment instructs the implementer to branch from a ref that will never exist. Wave-28 was `direct-to-main`; wave-29 is too. Since the 2026-06-09 every-wave-merges-to-main directive, `direct-to-main` is the common case, so this is the default-wrong path rather than an edge case.

The merge model is already recorded and readable at render time — `wave_{M}_merge_model` as a top-level key, and `merge_model` inside `wave_{M}_scope`. The hook loads that same status dict to find the assignment row.

**Fix direction:** read the wave's merge model and render `main` as the branch base for `direct-to-main`, `deployments/phase-{P}/wave-{M}` for `wave-branch`. When the model is absent, the safer default is `main` (a wave branch that does not exist is a hard block for the implementer; branching from `main` on a wave-branch wave is a recoverable retarget). Wants a regression fixture per model.

## Problem 2 — ordinary command wrappers defeat the parser silently

`parse_label_apply_command` fails to parse a wave-label-apply whenever the `gh` call is **wrapped**, and the hook then does nothing at all. Two independent defeaters, both confirmed by running the parser directly against the exact strings used:

| Command form | Parses? |
|---|---|
| `gh issue edit 1114 --repo … --add-label "wave-29"` | yes |
| `gh issue edit 1114 … --add-label "wave-29" >/dev/null 2>&1` | yes |
| `gh issue edit 1114 … --add-label "wave-29" && echo ok` | yes |
| `timeout 45 gh issue edit 1114 … --add-label "wave-29"` | **None** |
| `for n in 1114 1116; do gh issue edit "$n" … --add-label "wave-29"; done` | **None** |
| `for x in a; do gh issue edit 1114 … --add-label "wave-29"; done` | **None** |

So:

1. **A command-prefix wrapper** — `timeout 45 gh …` — defeats it. The parser expects the `gh` invocation at the head of the command.
[...]

PR #1156, opened 2026-07-27
fix(hooks): route `cd` from the Bash AST, not the flattened segment list

### LINKED ISSUE #1151: bug(shell-parse): extract_leading_cd_target honours a cd the shell never runs, or one after the gh — misroutes repo writes (#981/#985 class)

parametrization:
Closes #1151.

## What was wrong

`extract_leading_cd_target` decides which GitHub repo eight hooks read and **three hooks write** to. It took the **last `cd /abs` pair anywhere in the flat segment list**, which checks none of the three properties that actually make a `cd` honourable:

1. **unconditional** — not guarded by `||`, not inside `if`/`while`/`for`/`case`
2. **in the current shell** — not in a subshell, pipeline, `&` job, or command substitution
3. **before the work** — positioned ahead of the command being routed

## Families closed

| # | Shape | Old result | New result | Shell truth (bash + zsh) |
|---|---|---|---|---|
| **A** | `true \|\| cd <sibling> ; gh issue edit 5 …` | `<sibling>` | `None` | start dir |
| **A** | `false && cd <sibling> ; gh issue edit 5 …` | `<sibling>` | `None` | start dir |
| **B** | `gh issue edit 5 … ; cd <sibling>` | `<sibling>` | `None` | start dir |
| **B** | `gh issue edit 5 … && cd <sibling>` | `<sibling>` | `None` | start dir |
| **B** | `cd /a && { cd <sibling> ; gh x ; }` | `<sibling>` | `None` | `<sibling>` (ambiguous → refused) |
| **C** | `cd /parent && cd child-repo && gh …` | `/parent` | `None` | `/parent/child-repo` |

**Family C was not in the issue.** It fell out of the same rule and is the sharpest of the three for this org specifically: the old scan silently skipped the relative leg and kept the stale absolute prefix, so `cd /noorinalabs-main && cd noorinalabs-deploy && gh issue create` resolved `origin` to the **parent** repo. Child repos are cloned beneath `noorinalabs-main/`, so the `os.path.isdir` guard does not catch it. Relative legs are now refused outright.

## How — two CO-PRIMARY mechanisms, not one

**Fix direction 1, the AST**, is the main vehicle: `_ast_leading_cd_target` walks only the **top-level** command list of the bashlex parse and honours a `cd` solely inside the unconditional leading run; anything after that run containing a `cd` at all makes the answer ambiguous and returns `None`. Segment-level analysis provably cannot answer this — `iter_command_segments` has already discarded the control flow, and a guarded `cd` sits at token 0 of its own segment with no leader left to withhold, which is exactly why patching leader-by-leader kept producing new families.

**But the AST cannot be the SOLE mechanism**, and that is a change from the brief. Weronika's #1152 found — and I re-verified against the installed bashlex, not from docs — that a quoted-delimiter heredoc does not parse at all:

```
bashlex.parse("cat <<'EOF'\nx\nEOF\n")  -> ParsingError: here-document ... delimited by end-of-file
bashlex.parse('cat <<"EOF"\nx\nEOF\n')  -> ParsingError
bashlex.parse("cat <<EOF\nx\nEOF\n")    -> OK       (bare and <<- forms parse)
```

`<<'EOF'` is this repo's dominant heredoc form (`python3 - <<'PY'`, `gh issue comment --body-file - <<'EOF'`), so **the commands most likely to carry a `cd` are exactly the ones the AST cannot see.** And the failure is silent: `bashlex_available()` reports *import* success, not per-command success, and `iter_command_segments_ast` returns the same `None` it uses for "bashlex absent". A caller treating that `None` as "fall back to the old shlex scan" would report success and change nothing on precisely those commands.

**This resolver does not do that.** Its fallback is the issue's **fix direction 2** — fail closed on any control-flow token — so the families stay closed with no AST at all. Measured:

| shape (with `<<'EOF'` attached) | bashlex | old scan | this PR | bash | zsh |
|---|---|---|---|---|---|
| `true \|\| cd <dest> ; gh … <<'EOF'…EOF` | **ParsingError** | `<dest>` | `None` | start | start |
| `gh … <<'EOF'…EOF` + `cd <dest>` | **ParsingError** | `None`\* | `None` | start | start |
| `cd <start> && cd child-repo && gh … <<'EOF'…EOF` | **ParsingError** | `<start>` | `None` | `<start>/child-repo` | `<start>/child-repo` |
| `cd <dest> && gh … <<'EOF'…EOF` (genuine) | **ParsingError** | `<dest>` | `<dest>` | `<dest>` | `<dest>` |

\* only by accident — the missing `normalize_command_separators()` glues the trailing `;`, the same accident the ordering constraint warns about.

`;` and `&&` are deliberately **not** in the fail-closed token set — they are the separators of the leading run itself, and dropping them would delete `cd /worktree && gh pr create`, the #521 recovery this resolver exists for. Heredoc **bodies** are deliberately not stripped either: a body always follows its command, so its tokens can never occupy the leading position, and any `cd` they contribute lands in the non-leading phase and forces `None`. Unstripped bodies can only cost recovery, never buy a route — five adversarial shapes pin that.

The asymmetry is #1141's, restated: **under-detecting degrades to the recoverable #650 invocation-cwd fallback; over-detecting writes into the wrong repository, which no retry undoes.**

## Ordering constraint honoured

This does **not** add `normalize_command_separators()` to the routing path — see `wave_29_scope.cross_wave_dependencies` (#1151 before/with #1150, never after). The AST does now read an unspaced `cd /a;gh …` correctly, but that widening is *toward shell truth* and only after Families A/B are closed structurally; it is not the "normalize everywhere" change the constraint warns about.

## Second-order fix: `no_worktree_self_delete`

Same defect, landing on a **safety verdict** rather than repo identity. `git worktree remove <wt> ; cd /safe` run from inside `<wt>` previously resolved to `/safe` and **failed open on a real self-delete** — the shell is still inside `<wt>` when the remove executes. Now blocks. Same for the `||`-guarded variant. Both pinned.

## Tests — shell truth, in BOTH shells

Both families survived the previous suite because the tests encoded the implementation. Every `cd`-routing shape is now executed in a **real shell** with the `gh` node replaced by `pwd`, and the printed directory is ground truth:

```
resolved is None  OR  resolved == truth(bash) == truth(zsh)
```

- **The oracle is widened from bash-only to bash AND zsh.** #1141 ran bash — CI's shell — and documented the gap in prose. zsh is the actual harness shell, and `command cd /x` genuinely differs between them, so a bash-only oracle can adopt a bash-only behaviour as a resolver invariant. That divergence is now **asserted** (bash moves, zsh does not) rather than described.
- `.github/workflows/ci.yml` installs `zsh` in the pytest job so both halves really run in CI. The tests degrade to the bash-only oracle on a bare checkout without zsh, with a comment saying that must never be why CI is green.
- The `test_known_gap_main_1151_still_misroutes` characterization test — which *pinned* the misroute so the gap stayed visible — is replaced exactly as its own docstring instructed on closure.
- `CdRoutingWhenBashlexCannotParse` covers the **fallback as ordinary traffic**, not an edge case: the premise itself (quoted forms raise, bare forms parse), the trap (`bashlex_available()` True while the command did not parse), families A/B/C with a heredoc attached, five adversarial heredoc-body injection shapes, and the genuine recovery surviving. Every shape is first asserted to actually force the `ParsingError`, so it cannot silently stop testing the fallback.
- Degraded mode (bashlex monkeypatched off) is covered both ways: fails closed on every family shape, and still recovers the three canonical #521 shapes. The `_BASHLEX_AVAILABLE` gate is placed **outside** the memo (matching `iter_command_segments_ast`) so a monkeypatching test is never served a stale cached AST answer.

## Correction — a claim in this PR was false, and Aino falsified it

An earlier revision of this section said *"No shape I could construct returns a directory the shell does not enter, under either shell."* **That was wrong**, and the merge-gate review produced the counterexample:

```
cd DEST && pushd OTHER && gh …    ->  resolved DEST;  both shells run the gh in OTHER
```

Same for `eval 'cd …'`, `source SCRIPT`, and `. SCRIPT`. `_ast_subtree_contains_cd` refused only commands whose `words[0] == "cd"`, so a builtin that moves the shell *without being spelled* `cd` passed phase 2 untouched. `OTHER` is a real directory, so the `isdir` guard did not catch it either.

Not a regression — the old scan returns `DEST` too — but the unrecoverable direction, which is the whole asymmetry #1151 turns on. **Fixed** at `1b77424`.

The root cause matters more than the symptom: this was **drift, not an open frontier**. `_DEGRADED_CONTROL_FLOW_TOKENS` already listed `eval`/`exec`/`source`/`.`/`pushd`/`popd` under *"Commands that can move the shell in ways this scan cannot follow"*, so the degraded scan returned `None` on all five while the AST path returned `DEST` — **the primary was strictly weaker than the co-primary I declared it equal to, on a class I had already enumerated.** So rather than patch the AST branch to match, both mechanisms now derive from one `_CWD_MOVING_COMMANDS` frozenset, and `test_both_paths_agree_on_cwd_moving_builtins` asserts the two paths return the *same answer* rather than asserting the shared constant. `_ast_subtree_contains_cd` is renamed `_ast_subtree_moves_cwd`.

Also fixed: the phase-1 `prev_op` guard was **inert**. Every family-A fixture was `true || cd …`, where phase 1 breaks at `true` and phase 2 closes the shape before the guard is reached — so deleting a load-bearing guard left the suite green while `cd A || cd B && gh` reopened a real misroute to `B`. Two fixtures with a `cd` in the leading position now cover it. The `test_family_a_and_b_shell_truth_is_the_start_dir` skip list is gone too — every shape in both families really does have the start dir as its truth, so the prefix-matching exemption was hiding coverage rather than earning it.

Mutation-checked, because "the guard is now covered" is exactly the claim that needs an instrument:

| mutation | failures |
|---|---|
| delete the phase-1 `prev_op` guard | **2** (was 0 — inert) |
| phase 2 keys on `"cd"` only (pre-fix) | **12** |
| drop `_CWD_MOVING_COMMANDS` from the degraded token set | **6** |

## Residual — honestly scoped

Families A, B and C are closed. What remains is **under-recovery**, in every case falling back to the recoverable invocation cwd:

- a `cd` in a taken `else` or a real loop iteration (unchanged from #1141)
- `mkdir -p /a && cd /a && gh …` — the run is no longer *leading*, so `None`
- any relative `cd` (family C's safe direction)
- `cd "$(git rev-parse --show-toplevel)"` — not statically resolvable (unchanged)
- a `cd` inside a group/pipeline after the leading run — refused as ambiguous
- **anything in `_CWD_MOVING_COMMANDS` after the run** — `pushd`/`popd`/`eval`/`source`/`.`, and `exec`, which does not move the shell but *replaces* it, so the routed command never runs at all (`cd DEST && exec -a x true ; gh …` previously returned `DEST` for a `gh` that never executes)

I am not restating the "no shape I could construct" claim. What I will state is narrower and checkable: **every shape below returns `None` on the AST path AND the degraded path, and the two paths agree on all of them** — which is the property whose violation produced M1.

## Verification

- `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests .claude/lib/tests -q` → **3018 passed, 243 subtests** (baseline ~2999)
- `pre-commit run --all-files` and `--hook-stage pre-push` → **all 16 hooks pass** (ruff-format, ruff-lint, actionlint, cspell, checksums-ascii, mypy, pytest, pytest-skills, memory-budget, headcount-budget, doc-freshness, office-drift, mermaid-render, + the file-scoped linters)
- `python3 .claude/lib/pre_commit_ci_sync.py .` → OK
- `actionlint .github/workflows/ci.yml` → clean
- CI-only smoke test (py_compile over changed files) → clean

## File-overlap note

Scoped to the `cd`-resolution functions. **No heredoc helper is touched** — `strip_heredocs` is not called on this path and the reason is now recorded rather than incidental — per the coordination note with #1152 (Weronika Zielinska). Her `_shell_parse.py` change sits next to `strip_heredocs` (~line 450) with its tests in a new file, so the two diffs do not overlap. This PR and #1155 agree on what bashlex can do; the finding is credited in the code comments and the commit message.

https://claude.ai/code/session_011hEQbrJGwa2kQ6dbEjyHQ6

### DIFF
--- a/.claude/hooks/_shell_parse.py
+++ b/.claude/hooks/_shell_parse.py
@@ -1113,6 +1113,25 @@ def resolve_tool_cwd(input_data: dict) -> str:
 # disqualifying, in either position.
 _UNCONDITIONAL_LEADING_OPS = frozenset({";", "&&"})
 
+# Commands that move (or replace) the CURRENT shell WITHOUT being spelled `cd`.
+# A leading `cd` is only evidence of where the work runs if nothing between it
+# and the work moves the shell again — and `pushd`/`popd`/`eval`/`source`/`.`
+# all do, invisibly to a scan that keys on the literal token `cd`.
+#
+# ONE list, consumed by BOTH mechanisms. It started life inline in the degraded
+# token set only, and the AST path keyed on `words[0] == "cd"` alone — so
+# `cd /a && pushd /b && gh …` returned `/a` on the AST path while the degraded
+# path correctly returned None, making the primary strictly weaker than the
+# co-primary it is declared equal to (found at the #1156 merge gate, Aino
+# Virtanen). Two mechanisms with independently maintained lists WILL drift
+# again; deriving both from this frozenset is what makes that impossible, and
+# is a stronger fix than patching the AST branch to match.
+#
+# `exec` is included for the same parity even though it cannot carry a `cd`
+# (`exec cd /x` fails — exec needs an external command): it REPLACES the shell,
+# so the routed command never runs at all. Refusing costs only recovery.
+_CWD_MOVING_COMMANDS = frozenset({"pushd", "popd", "eval", "source", ".", "exec"})
+
 # Fail-closed scan (bashlex absent, OR the command did not parse — most often
 # a quoted-delimiter heredoc): any of these appearing as its OWN shlex token
 # means the command carries control flow, a subshell, a pipeline or an opaque
@@ -1121,38 +1140,34 @@ _UNCONDITIONAL_LEADING_OPS = frozenset({";", "&&"})
 # ABSENT: they are the separators of the leading `cd` run itself, and dropping
 # `cd /worktree && gh pr create` would destroy the #521 recovery signal this
 # resolver exists for.
-_DEGRADED_CONTROL_FLOW_TOKENS = frozenset(
-    {
-        "if",
-        "then",
-        "else",
-        "elif",
-        "fi",
-        "while",
-        "until",
-        "for",
-        "do",
-        "done",
-        "case",
-        "esac",
-        "select",
-        "function",
-        "{",
-        "}",
-        "(",
-        ")",
-        "!",
-        "|",
-        "||",
-        "&",
-        # Commands that can move the shell in ways this scan cannot follow.
-        "eval",
-        "exec",
-        "source",
-        ".",
-        "pushd",
-        "popd",
-    }
+_DEGRADED_CONTROL_FLOW_TOKENS = (
+    frozenset(
+        {
+            "if",
+            "then",
+            "else",
+            "elif",
+            "fi",
+            "while",
+            "until",
+            "for",
+            "do",
+            "done",
+            "case",
+            "esac",
+            "select",
+            "function",
+            "{",
+            "}",
+            "(",
+            ")",
+            "!",
+            "|",
+            "||",
+            "&",
+        }
+    )
+    | _CWD_MOVING_COMMANDS
 )
 
 
@@ -1192,10 +1207,16 @@ def _ast_top_level_elements(tree) -> list[tuple[object, str | None, str | None]]
     return elements
 
 
-def _ast_subtree_contains_cd(node) -> bool:
-    """True if any command-position `cd` appears anywhere under `node`.
+def _ast_subtree_moves_cwd(node) -> bool:
+    """True if anything under `node` can move the shell's cwd.
 
-    Used on everything AFTER the leading `cd` run. A `cd` there makes the
+    That means a command-position `cd` OR any of `_CWD_MOVING_COMMANDS`
+    (`pushd`, `popd`, `eval`, `source`, `.`, `exec`) — the second half is the
+    part a `words[0] == "cd"` check misses, and missing it left
+    `cd /a && pushd /b && gh …` resolving to `/a` while both shells run the gh
+    in `/b` (#1156 merge gate).
+
+    Used on everything AFTER the leading `cd` run. A cwd move there makes the
     resolver's single answer ambiguous with respect to the command a consumer
     actually cares about (`gh issue edit 5 …; cd /elsewhere` — main#1151
     family B), or is a group whose `cd` applies to a later command in the same
@@ -1206,15 +1227,15 @@ def _ast_subtree_contains_cd(node) -> bool:
     """
     found = False
 
-    class _CdFinder(bashlex_ast.nodevisitor):
+    class _CwdMoveFinder(bashlex_ast.nodevisitor):
         def visitcommand(self, n, parts):
             nonlocal found
             words = [p.word for p in parts if getattr(p, "kind", None) == "word"]
-            if words and words[0] == "cd":
+            if words and (words[0] == "cd" or words[0] in _CWD_MOVING_COMMANDS):
                 found = True
             return True
 
-    _CdFinder().visit(node)
+    _CwdMoveFinder().visit(node)
     return found
 
 
@@ -1279,9 +1300,11 @@ def _ast_leading_cd_target_cached(command: str) -> tuple[bool, str | None]:
         target = words[1]
         index += 1
 
-    # Phase 2 — nothing after the leading run may contain a `cd` at all.
+    # Phase 2 — nothing after the leading run may move the cwd again. Starts AT
+    # the breaking element, so a `pushd`/`eval`/`source` that ended phase 1 is
+    # itself inspected; phase 1 needs no matching check of its own.
     for node, _prev_op, _next_op in elements[index:]:
-        if _ast_subtree_contains_cd(node):
+        if _ast_subtree_moves_cwd(node):
             return (True, None)
     return (True, target)
 
@@ -1399,10 +1422,21 @@ def extract_leading_cd_target(command: str) -> str | None:
     `mkdir -p /a && cd /a && gh …` (the run is no longer leading), for a
     RELATIVE `cd` (`cd /parent && cd child-repo` — refused outright now,
     because keeping `/parent` was itself a cross-repo misroute in a tree of
-    nested child repos), and for a pipeline sitting after the run. Each of
-    those falls back to the invocation cwd, which is the recoverable #650
-    behaviour. Claiming no knowledge is recoverable; claiming the wrong
-    directory is not.
+    nested child repos), for a pipeline sitting after the run, and for anything
+    in `_CWD_MOVING_COMMANDS` after the run (`cd /a && pushd /b && gh …`,
+    `… && eval 'cd /b' && …`, `… && source s.sh && …`) — including `exec`,
+    which does not move the shell but replaces it, so the routed command never
+    runs at all. Each of those falls back to the invocation cwd, which is the
+    recoverable #650 behaviour. Claiming no knowledge is recoverable; claiming
+    the wrong directory is not.
+
+    That last group is NOT a frontier the design leaves open — it is a case
+    where the two mechanisms had drifted. The degraded scan already refused
+    `pushd`/`eval`/`source`, so the AST path keying on the literal token `cd`
+    was strictly weaker than its own fallback (#1156 merge gate). Both now read
+    the same `_CWD_MOVING_COMMANDS` frozenset. A shape whose cwd this function
+    cannot compute must resolve to None on BOTH paths, and any new entry
+    belongs in that one set, never in a per-path list.
     """
     parsed, target = _ast_leading_cd_target(command)
     if parsed:
--- a/.claude/hooks/_shell_parse.py
+++ b/.claude/hooks/_shell_parse.py
@@ -193,8 +193,10 @@ Public API
         unconditional, (2) executed by the current shell, and (3) positioned
         before the work. Reads the bashlex AST (not the flattened segment
         list, which has already discarded the control flow that decides
-        whether a `cd` runs) and falls back to a fail-closed token scan when
-        bashlex is unavailable. main#1151.
+        whether a `cd` runs), with a fail-closed token scan as a CO-PRIMARY
+        for the commands bashlex cannot parse — notably any quoted-delimiter
+        heredoc (`<<'EOF'`), which raises rather than yielding an AST.
+        main#1151.
 
     resolve_invocation_cwd(input_data) -> str
         Like resolve_tool_cwd, but FIRST tries to recover the directory the
@@ -1073,9 +1075,30 @@ def resolve_tool_cwd(input_data: dict) -> str:
 # Segment-level analysis provably cannot answer any of these: by the time
 # `iter_command_segments` has split on `;`/`&&`/`||`/`|` the control flow that
 # decides whether a `cd` runs is already gone, and a guarded `cd` lands at token
-# 0 of its own segment with no leader left to withhold. So the check moves up to
-# the bashlex AST, where all three are directly readable, and falls back to a
-# fail-closed token scan when bashlex is absent.
+# 0 of its own segment with no leader left to withhold.
+#
+# TWO CO-PRIMARY MECHANISMS, not a primary and a safety net:
+#
+#   1. `_ast_leading_cd_target`  — the bashlex AST, where all three properties
+#                                  are directly readable.
+#   2. `_degraded_leading_cd_target` — a token scan that FAILS CLOSED on any
+#                                  control flow, used whenever (1) cannot see
+#                                  the command.
+#
+# (2) is NOT a rare degraded path. main#1152 established, against the installed
+# bashlex, that a QUOTED-delimiter heredoc does not parse AT ALL:
+#
+#     bashlex.parse("cat <<'EOF'\nx\nEOF\n")  -> ParsingError
+#
+# `<<'EOF'` / `<<"EOF"` raise; only bare `<<EOF` / `<<-EOF` parse. The quoted
+# form is the DOMINANT one in this repo (`python3 - <<'PY'`, `gh issue comment
+# --body-file - <<'EOF'`), so the commands most likely to carry a `cd` are
+# exactly the ones the AST cannot see. And the failure is SILENT: bashlex_available()
+# stays True (it reports import success, not per-command success) and
+# `iter_command_segments_ast` returns the same None it uses for "bashlex
+# absent". An AST-ONLY fix here would therefore have reported success and
+# changed nothing on precisely those commands — which is why (2) must close, on
+# its own, every family (1) closes. Pinned by `CdRoutingWhenBashlexCannotParse`.
 #
 # The asymmetry is deliberate and is the same call main#1141 made: UNDER-
 # detecting a `cd` degrades to the recoverable #650 invocation-cwd fallback;
@@ -1263,12 +1286,25 @@ def _ast_leading_cd_target_cached(command: str) -> tuple[bool, str | None]:
 
 
 def _degraded_leading_cd_target(command: str) -> str | None:
-    """Fail-closed token-scan fallback for when bashlex is unavailable.
+    """Fail-closed token scan — the issue's fix direction 2, as a CO-PRIMARY.
+
+    Runs whenever the AST cannot see the command: bashlex absent, or (far more
+    often) a command carrying a quoted-delimiter heredoc, which the installed
+    bashlex cannot parse at all. `python3 - <<'PY'` and `gh issue comment
+    --body-file - <<'EOF'` are the repo's dominant heredoc forms, so this path
+    is ordinary traffic, not an edge case, and it must close every family the
+    AST path closes rather than relaxing to the old last-`cd`-wins scan.
 
     Mirrors the AST phase model with the only tool left: reject outright any
     command whose shlex token stream shows control flow, a subshell, a pipeline
     or an opaque cwd change (`_DEGRADED_CONTROL_FLOW_TOKENS`), then accept only
     a leading run of `cd /abs` segments with no `cd` anywhere after it.
+
+    Heredoc BODIES are deliberately not stripped here (`strip_heredocs` is not
+    called): a body always follows its command, so its tokens can never occupy
+    the leading position, and any `cd` they contribute lands in the non-leading
+    phase and forces None. Unstripped bodies can therefore only cost recovery,
+    never buy a route — verified in `test_heredoc_body_cannot_inject_a_route`.
 
     Deliberately does NOT call `normalize_command_separators` — see the
     main#1151 ordering constraint recorded in `wave_29_scope`. Without it an
@@ -1328,8 +1364,13 @@ def extract_leading_cd_target(command: str) -> str | None:
     cannot answer them: `iter_command_segments` has already discarded the
     control flow, and a guarded `cd` sits at token 0 of its own segment with no
     leader left to withhold — which is why patching leader-by-leader kept
-    producing new families. When bashlex is unavailable the degraded scan fails
-    closed on any control-flow token instead.
+    producing new families.
+
+    The AST is NOT the sole mechanism. A quoted-delimiter heredoc (`<<'EOF'`,
+    the repo's dominant form) does not parse at all, so
+    `_degraded_leading_cd_target` — a token scan that fails closed on any
+    control flow — is a CO-PRIMARY that must close the same families on its
+    own. See the module-level § Unconditional-leading-`cd` routing resolution.
 
     Strips NOTHING — `cd` must be word 0 of a top-level simple command
     (main#1141 review round 3). Both prefix families are unsafe here:
--- a/.claude/hooks/_shell_parse.py
+++ b/.claude/hooks/_shell_parse.py
@@ -94,7 +94,10 @@ Public API
             the command never went, and three hooks downstream WRITE.
             `extract_leading_cd_target` is the only one in this module, and it
             opts out entirely (a leader may not run; and an exec-wrapper
-            cannot carry `cd` at all, since `cd` is a shell builtin).
+            cannot carry `cd` at all, since `cd` is a shell builtin). As of
+            main#1151 it does not consume segments at all — it reads the AST,
+            because the three properties that make a `cd` honourable are not
+            recoverable from a flattened segment list at any strip setting.
 
         `compound_leaders=False` exists for a caller that wants wrappers but
         not leaders. Nothing needs it today — the one routing caller wants
@@ -182,20 +185,29 @@ Public API
         want to operate on the *user's* cwd (not the hook's parent process
         cwd) should use this to anchor `subprocess.run(..., cwd=...)`.
 
+    extract_leading_cd_target(command) -> str | None
+        The ROUTING resolver: the directory the shell is in when the
+        command's first real work runs, or None when that cannot be
+        established with certainty. Honours only the UNCONDITIONAL LEADING
+        RUN of `cd <absolute-dir>` commands — a `cd` must be (1)
+        unconditional, (2) executed by the current shell, and (3) positioned
+        before the work. Reads the bashlex AST (not the flattened segment
+        list, which has already discarded the control flow that decides
+        whether a `cd` runs) and falls back to a fail-closed token scan when
+        bashlex is unavailable. main#1151.
+
     resolve_invocation_cwd(input_data) -> str
         Like resolve_tool_cwd, but FIRST tries to recover the directory the
-        command actually runs in by extracting a leading `cd <dir>` segment
-        from the Bash command string. This closes the #521 residual: for a
-        worktree subagent the harness `cwd` field is captured at agent-spawn
-        time (the orchestrator's dir), NOT the subagent's dir after it has
-        `cd`'d into its worktree, and subsequent `cd` calls do not propagate
-        back to the hook's view of `cwd`. When the triggering command is
-        `cd /path/to/worktree && gh pr create ...`, the cd target is the
-        only in-band signal that recovers the real repo. Falls back to
-        resolve_tool_cwd (stdin cwd → os.getcwd()) when no leading `cd`
-        is present. Only absolute existing directories are honored; relative
-        cd targets are ambiguous (they'd be relative to the already-wrong
-        stdin cwd) so they are ignored.
+        command actually runs in via `extract_leading_cd_target`. This closes
+        the #521 residual: for a worktree subagent the harness `cwd` field is
+        captured at agent-spawn time (the orchestrator's dir), NOT the
+        subagent's dir after it has `cd`'d into its worktree, and subsequent
+        `cd` calls do not propagate back to the hook's view of `cwd`. When the
+        triggering command is `cd /path/to/worktree && gh pr create ...`, the
+        cd target is the only in-band signal that recovers the real repo.
+        Falls back to resolve_tool_cwd (stdin cwd → os.getcwd()) when no
+        honourable leading `cd` is present. Only absolute existing directories
+        are honored.
 
     resolve_repo_short_name(input_data, *, git_runner=None) -> str | None
         Resolve the GitHub repository NAME (e.g. `noorinalabs-main`) from the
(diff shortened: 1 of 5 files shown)


parametrization:
@aino-virtanen — **review requested (merge gate)**, per the charter and your #1143 re-review that filed #1151.

Owner-gated: I am **not** merging. This stops here for your verdict.

Three things you asked for specifically, and where to find them:

1. **Tests assert shell truth, not the resolver.** `CdRoutingAgainstShellTruth._assert_never_misroutes` runs each shape in a real shell with the `gh` node replaced by `pwd` and compares against the printed cwd. The property is `resolved is None OR resolved == truth(bash) == truth(zsh)`.
2. **The oracle is now BOTH shells.** #1141 ran bash only and documented the gap in prose; that gap is closed. `.github/workflows/ci.yml` installs `zsh` in the pytest job so both halves really run in CI, and the `command cd /x` divergence is now *asserted* (bash moves the shell, zsh does not) rather than described.
3. **Which families close, honestly scoped.** A and B close, plus a third (relative `cd` into a nested child repo) that was not in the issue and that I found while writing the rule. Residual is under-recovery only, enumerated in the PR body and in the `extract_leading_cd_target` docstring.

**One deviation from the brief you should weigh first.** The brief ranked "use the AST" above "fail closed on control flow". Weronikas #1152 finding — which I re-verified against the installed bashlex — is that `bashlex.parse` raises `ParsingError` on any quoted-delimiter heredoc (`<<'EOF'`, `<<"EOF"`), and that is this repos dominant form. Since `bashlex_available()` reports *import* success and `iter_command_segments_ast` returns the same `None` it uses for "bashlex absent", an AST-only fix would have degraded **silently** to the exact behaviour being fixed, on the commands most likely to carry a `cd`.

So the two mechanisms are **co-primary**: the fail-closed scan must close every family the AST closes, on its own, and `CdRoutingWhenBashlexCannotParse` pins that with every shape first asserted to actually force the `ParsingError`. This PR and #1155 agree on what bashlex can do.

Also worth a look: `no_worktree_self_delete` inherits the fix on its **safety verdict** — `git worktree remove <wt> ; cd /safe` from inside `<wt>` previously resolved to `/safe` and failed open on a real self-delete.

Ordering constraint with #1150 is intact — `normalize_command_separators()` is deliberately *not* added to this path.

cc @santiago-ferreira for the second reviewer slot.

parametrization:
**Merge-gate review — cast at head `06558317b8868eca88c9af7ed70eb05b88beaf59`, base `main` @ `fea0dca` (branch is directly on top of the current main tip; no rebase skew, each diff taken against its own merge-base).**

I filed #1151, so I re-ran every claim on the instrument rather than reading the write-up. Almost all of it holds — including the parts I would most have wanted to just agree with. One thing does not, and it is in the direction the PR itself says is the only one that matters.

---

## Must-fix

### 1. Over-detection survives on the AST path: a cwd-moving builtin after the leading `cd` run

The PR states: *"No shape I could construct returns a directory the shell does not enter, under either shell."* That is false, and the shapes are short:

| command | resolver | bash truth | zsh truth |
|---|---|---|---|
| `cd DEST && pushd OTHER && gh …` | **`DEST`** | `OTHER` | `OTHER` |
| `cd DEST ; pushd OTHER ; gh …` | **`DEST`** | `OTHER` | `OTHER` |
| `cd DEST && eval 'cd OTHER' && gh …` | **`DEST`** | `OTHER` | `OTHER` |
| `cd DEST && . SCRIPT && gh …` | **`DEST`** | `OTHER` | `OTHER` |
| `cd DEST && source SCRIPT && gh …` | **`DEST`** | `OTHER` | `OTHER` |

(`SCRIPT` contains `cd OTHER`. `OTHER` is a real directory, so the `os.path.isdir` guard in `resolve_invocation_cwd` does not catch it either.)

Mechanism: `_ast_leading_cd_target` phase 1 breaks on the first non-`cd` element, and phase 2 refuses only if `_ast_subtree_contains_cd` finds a command whose `words[0] == "cd"`. `pushd` / `eval` / `source` / `.` move the current shell without being spelled `cd`, so they pass phase 2 untouched and the stale leading target is returned.

**What makes this blocking rather than a documented residual is that the code already contains the answer.** `_DEGRADED_CONTROL_FLOW_TOKENS` lists `eval`, `exec`, `source`, `.`, `pushd`, `popd` under the comment *"Commands that can move the shell in ways this scan cannot follow"* — and the degraded path correctly returns `None` on all five shapes above. Attach a `<<'EOF'` heredoc to any of them, force the `ParsingError`, and the answer flips from `DEST` to `None`. So the **primary** mechanism is strictly weaker than the **co-primary** it is declared equal to, on a class the author had already enumerated. That is an internal inconsistency, not an unclosed frontier.

To be fair on severity: this is **not a regression** — the old last-`cd`-wins scan returns `DEST` on these too. But #1151 is not "restore parity", it is "only honour a `cd` that is unconditional, in the current shell, and before the work", and a `cd` invalidated by a later `pushd` fails property (3) just as family B does. The residual list ("in every case falling back to the recoverable invocation cwd") and the module docstring both assert the safe direction here; both are currently inaccurate.

I validated the fix before asking for it — it is ~6 lines, and phase 1 needs no change because phase 2 already starts at the breaking element:

```python
# Builtins that move the CURRENT shell without being spelled `cd`. The degraded
# scan already refuses these (`_DEGRADED_CONTROL_FLOW_TOKENS`); the AST path must
# too, or it is strictly weaker than its own co-primary fallback.
_AST_CWD_MOVING_COMMANDS = frozenset({"pushd", "popd", "eval", "source", "."})
```

and in `_CdFinder.visitcommand`:

```python
if words and (words[0] == "cd" or words[0] in _AST_CWD_MOVING_COMMANDS):
    found = True
```

Measured with that patch applied: all five shapes → `None`; the three #521 recovery shapes still → `DEST`; full suite **3013 passed, 221 subtests** — unchanged. Please also correct the "no shape I could construct" claim and the residual list, and rename `_ast_subtree_contains_cd` to match what it now detects.

`cd DEST && exec -a x true ; gh …` also returns `DEST` where the `gh` never runs at all. Harmless (no write happens), but worth a line in the residuals rather than silence.

### 2. The `||`-inside-the-leading-run guard is inert — no fixture can trip it

Deleting this from phase 1:

```python
if prev_op is not None and prev_op not in _UNCONDITIONAL_LEADING_OPS:
    return (True, None)  # `||`-guarded or `&`-preceded
```

leaves the suite at **189 passed, 114 subtests — identical to baseline**.

The guard is genuinely load-bearing; it is only untested. Every family-A fixture is `true || cd DEST …`, where element 0 is `true`, so phase 1 breaks immediately and **phase 2** closes the shape — the `prev_op` branch is never reached. It is reached only when the leading element is itself a `cd`:

| command | head | bash truth | zsh truth | with the guard deleted |
|---|---|---|---|---|
| `cd A \|\| cd B && gh …` | `None` | `A` | `A` | **`B`** ← misroute |
| `cd A \|\| cd B ; gh …` | `None` | `A` | `A` | **`B`** ← misroute |

So a future refactor can delete a real guard and reopen a real misroute with a green suite. Please add `"cd A_ || cd B_ && MARKER"` (leading element a `cd`) to `FAMILY_A_SHAPES`. This is the PR's own thesis applied to the PR — the argument for replacing the characterization test is that fixtures must encode shell truth rather than implementation, and this guard currently has neither.

---

## Verified — claims that held under the instrument

Everything below I re-ran; none of it is taken on the write-up's word.

1. **Family C is real, and is a genuine extension of #1151.** `cd START && cd child-repo && gh …`: old resolver → `START` (the parent), shell truth → `START/child-repo`, in both shells. With child repos cloned beneath `noorinalabs-main/` the `isdir` guard passes and `origin` resolves to the parent. This is the sharpest of the three for this org and I did not have it in the issue.

2. **The old → new table reproduces exactly** — `None` in all eight misroute rows, `DEST` preserved in all three #521 recovery rows, against real bash and zsh. No row disagreed with the PR body.

3. **The bashlex premise, against the installed library, not the docs.** `<<'EOF'` and `<<"EOF"` raise `ParsingError`; `<<EOF` and `<<-EOF` parse. The silent-degradation trap is exactly as described: `bashlex_available()` → `True`, `iter_command_segments_ast(cmd)` → `None`, `_ast_leading_cd_target(cmd)` → `(False, None)`. An AST-only fix would indeed have reported success and changed nothing on this repo's dominant heredoc form. **The deviation from the brief was correct**, and the `(parsed, target)` return is the right shape for it.

4. **The degraded path closes all three families with no AST at all.** With `_BASHLEX_AVAILABLE` forced off: families A, B, C, plus `cd DEST & MARKER` and `cd DEST | cat ; MARKER` → all `None`; all three canonical #521 shapes → `DEST`. Co-primary, as claimed.

5. **The CI zsh step actually ran** — this is the one I most expected to find silently skipped, and it did not. In run `30314830304` at this head: `0 upgraded, 2 newly installed`, `Setting up zsh (5.9-6ubuntu2)`, step conclusion `success`, ordered before `Run hook and lib tests` (also `success`). The `shutil.which` guard did not narrow the oracle in CI.

6. **The second-order self-delete fix is real, and the old behaviour was a fail-open.** I ran the two new tests against the base hook at `fea0dca`: both fail with `AssertionError: a trailing cd must not relax the self-delete guard` / `assert None is not None`. `git worktree remove <wt> ; cd /safe` from inside `<wt>` previously **allowed** a real self-delete. Both block at head. Good catch, and correctly scoped as a safety verdict rather than repo identity.

7. **The ordering constraint is honoured.** `normalize_command_separators` appears in `_shell_parse.py` only at its own definition (line 357) and inside a comment explaining its deliberate absence — it is not called on the routing path. `strip_heredocs` likewise is not called. The #1151 → #1150 constraint in `wave_29_scope.cross_wave_dependencies` is respected.

8. **Heredoc bodies cannot inject a route.** Your five adversarial shapes plus four of my own (`cat <<'EOF'\ncd DEST && \nEOF\nGH`, `GH <<'EOF'\nEOF\ncd DEST`, `cd START && GH <<'EOF'\ncd DEST\nEOF`, `cat <<'EOF'\nx\nEOF\ncd DEST && GH`) — zero violations in either shell. Bodies cost recovery, never buy a route, as documented.

9. **The characterization test was replaced, not deleted.** Reverting `extract_leading_cd_target` to the old last-`cd`-wins body → **22 failed**. Dropping the phase-2 trailing-`cd` scan → 2 failed. Accepting relative targets → 4 failed. Dropping the `next_op == "&"` guard → 1 failed. Dropping the degraded control-flow refusal → 2 failed. Dropping the degraded "cd after real work" refusal → 4 failed. The replacement fixtures have teeth; only the two guards in the must-fix section are inert.

10. **CI and local gates at this head.** 21/21 checks `SUCCESS`, all reported against `06558317` (confirmed via the check-runs API at the SHA, not the PR-level rollup). Full suite reproduced locally: **3013 passed, 221 subtests**. `pre-commit run --all-files` passes.

---

## #1155 / #1156 — independent, either order

I am the merge gate on both, so I checked this semantically rather than trusting the clean test-merge.

- Both share merge-base `fea0dca`, the current `main` tip — no rebase skew on either side.
- Hunks in `_shell_parse.py` are disjoint: #1155 at 33-48 / 244-259 / 443-728, #1156 at 94-104 / 182-202 / 1043-1367.
- **The semantic check that matters:** #1155's `_shell_parse.py` change is purely *additive* — new `_HEREDOC_OPENER_RE`, `_segment_head_command`, `_scan_command_line`, `_opener_feeds_interpreter`, `classify_heredocs`, `strip_data_heredocs`. It does **not** modify `tokenize`, `iter_command_segments`, or the body of `strip_heredocs` — which are precisely the three things #1156's degraded path depends on. And #1156 deliberately calls no heredoc helper at all, so there is no shared mutable surface in either direction.
- Test-merged at both current heads (`06558317` + `e163320`): zero conflicted files, full suite **3087 passed, 228 subtests**.

**Verdict: they can land in either order, with no sequencing constraint.** One forward note for whoever takes #1150: the seam only becomes live if `extract_leading_cd_target` ever starts calling `strip_data_heredocs`. #1156 records why it does not; keep that comment if the resolver is touched again.

---

## On the shared-helper footgun you flagged

Agreed, and it is worth more than a note — filed as **#1165**. Your `(parsed, target)` fix is right but private; `iter_command_segments_ast` still returns the same `None` for "bashlex absent" and "did not parse", and `validate_commit_identity` consumes it. My preference for #1150 is option (1): give the shared helper the same `(parsed, value)` shape you gave yours, so the distinction is unignorable at the call site rather than an optional probe a caller can forget. `test_bashlex_available_is_not_evidence_the_command_parsed` is a good artifact to point that work at.

Also filed **#1166** for the `_BASHLEX_AVAILABLE`-outside-the-memo guard: moving it inside the memo leaves the suite at 189 passed, so the property its docstring claims is untested. The placement is correct — the fixture is missing. It needs a shape where the AST and degraded answers *diverge* (e.g. `cd DEST && gh … | cat`, which the AST resolves and the token scan refuses on the `|`); on every shape currently used the two agree, so a stale cached answer is indistinguishable from a correct one.

---

This is strong work and the deviation from the brief was the right call — the heredoc finding is the kind of thing that turns a green suite into a false negative, and you caught it against the library instead of the documentation. The two must-fixes are small and I have measured both fixes green. Re-request when they land and I will re-run the adversarial set; I am not merging (owner-gated) regardless of outcome.

---
Requestor: Aino Virtanen
Requestee: Lucas Ferreira
TechDebt: #1165, #1166

parametrization:
Both must-fixes pushed at `1b77424`. @aino-virtanen — re-review requested.

## M1 — fixed, and I took the root cause rather than the symptom

Your five shapes all return `None` now, on **both** paths, and the two paths agree on every one:

| shape | old | AST | degraded | bash | zsh |
|---|---|---|---|---|---|
| `cd DEST && pushd OTHER && gh …` | `DEST` | `None` | `None` | OTHER | OTHER |
| `cd DEST ; pushd OTHER ; gh …` | `DEST` | `None` | `None` | OTHER | OTHER |
| `cd DEST && eval 'cd OTHER' && gh …` | `DEST` | `None` | `None` | OTHER | OTHER |
| `cd DEST && . SCRIPT && gh …` | `DEST` | `None` | `None` | OTHER | OTHER |
| `cd DEST && source SCRIPT && gh …` | `DEST` | `None` | `None` | OTHER | OTHER |
| `cd DEST && exec -a x true ; gh …` | `DEST` | `None` | `None` | (never runs) | (never runs) |

`cd DEST && gh …`, `cd DEST ; gh …`, `cd /var && cd DEST && gh …` still resolve to `DEST`.

**I did not apply your patch as written**, and the reason is your own diagnosis. You called this an internal inconsistency rather than an unclosed frontier — I agree, and a second frozenset (`_AST_CWD_MOVING_COMMANDS` alongside the inline list in `_DEGRADED_CONTROL_FLOW_TOKENS`) leaves *two* independently maintained lists where the defect was already two independently maintained lists. So both mechanisms now derive from one `_CWD_MOVING_COMMANDS`, and the anti-drift test asserts the **property** — that the AST path and the degraded path return the same answer on every shape — not the shared constant, so a future split back into per-path lists fails on behaviour rather than on a name. `exec` is in the set for the same parity, which also disposes of your `exec -a x true` footnote rather than leaving it a silent residual. Renamed to `_ast_subtree_moves_cwd`.

## M2 — fixed, and mutation-checked rather than asserted

Added `cd START || cd DEST && MARKER` and the `;` variant to `FAMILY_A_SHAPES`. I also deleted the `test_family_a_and_b_shell_truth_is_the_start_dir` skip list: every shape in both families genuinely has the start dir as its truth, including the ones whose leading element is a `cd`, so that prefix-matching exemption was hiding coverage rather than earning it.

Since "the guard is now covered" is exactly the claim that needs an instrument, I re-ran your mutation:

| mutation | failures |
|---|---|
| delete the phase-1 `prev_op` guard | **2** (was 0 — your finding reproduced) |
| phase 2 keys on `"cd"` only (pre-fix) | **12** |
| drop `_CWD_MOVING_COMMANDS` from the degraded token set | **6** |

## Corrections landed

The PR body no longer claims "no shape I could construct returns a directory the shell does not enter" — it now carries the falsification, your counterexample, and the drift diagnosis. The residual list gains the `_CWD_MOVING_COMMANDS`-after-the-run class, and the `extract_leading_cd_target` docstring says explicitly that a shape this function cannot compute must resolve to `None` on **both** paths and that new entries belong in the one set, never a per-path list.

Suite **3018 passed, 243 subtests**; full `pre-commit --all-files` and `--hook-stage pre-push` green.

Thank you for re-running ten claims on the instrument instead of reading the write-up — M1 is precisely the kind of thing my own testing could not find, because I wrote both the resolver and the shapes I tried to break it with. Still not merging; owner-gated.

parametrization:
**Merge-gate re-review — cast at head `1b77424c9e798416cc266454df012b6010380a1b`.** Base `main` still `fea0dca`; `0655831` is an ancestor of this head, so the fixes landed as an additive commit with no force-push during review, per charter.

Both must-fixes are closed. I re-ran the adversarial set, the mutations, and the corpus measurement against blobs extracted directly from git at each SHA rather than from a working tree.

---

## M1 — closed, and you were right to decline my patch

Every shape returns `None` now, on both mechanisms, with the two agreeing on each:

| shape | AST | degraded | bash | zsh |
|---|---|---|---|---|
| `cd DEST && pushd OTHER && gh …` | `None` | `None` | OTHER | OTHER |
| `cd DEST ; pushd OTHER ; gh …` | `None` | `None` | OTHER | OTHER |
| `cd DEST && eval 'cd OTHER' && gh …` | `None` | `None` | OTHER | OTHER |
| `cd DEST && . SCRIPT && gh …` | `None` | `None` | OTHER | OTHER |
| `cd DEST && source SCRIPT && gh …` | `None` | `None` | OTHER | OTHER |
| `cd DEST && exec -a x true ; gh …` | `None` | `None` | — | — |

I also probed the nesting and indirection your fix might have missed, and it holds: `cd DEST && { pushd OTHER ; gh ; }`, `cd DEST && ( pushd OTHER ; gh )`, `cd DEST && eval 'pushd OTHER' && gh`, and the doubly-nested `eval 'eval "cd OTHER"'` all return `None`. `popd` back to DEST is refused even though the stale answer would have been accidentally right — which is the correct instinct.

All six recovery shapes still resolve to `DEST`, including `FOO=1 cd DEST`, the redirect form, and the trailing pipeline.

**On declining the patch: your reasoning is better than mine and I want that on the record rather than left as a tie.** My prototype was a minimal demonstration that the fix was small and green — it was not a prescription, and it did exactly what you say: created a second list beside the one already inline in `_DEGRADED_CONTROL_FLOW_TOKENS`. Two independently maintained lists reproduce the defect's own shape, so patching the AST branch to match would have fixed the instance and preserved the mechanism. Deriving both from one `_CWD_MOVING_COMMANDS` and having `test_both_paths_agree_on_cwd_moving_builtins` assert the **property** rather than the constant is the stronger fix, and `_ast_subtree_contains_cd` → `_ast_subtree_moves_cwd` is the right rename because the old name is what made the gap easy to miss. Adding `exec` for parity was also right — it disposes of my footnote instead of leaving it a silent residual.

## M2 — closed, and the guard now has teeth

Reproduced your mutation result, which is the one my finding turned on:

| mutation | failures at `0655831` | failures at `1b77424` |
|---|---|---|
| delete the phase-1 `prev_op` guard | **0** (inert) | **2** |
| phase 2 keys on `"cd"` only | — | **12** |
| drop `_CWD_MOVING_COMMANDS` from the degraded set | — | **6** |

All three match what you reported exactly.

**The skip-list deletion adds coverage rather than dropping it — confirmed shape by shape.** The old exemption was `template.startswith("cd START") or startswith("cd DEST")`, which covered four shapes, and every one of them genuinely has the start dir as its truth:

- `cd START ; MARKER ; cd DEST` and `cd DEST && { cd START ; MARKER ; }` (family B) — previously exempt, now asserted.
- `cd START || cd DEST && MARKER` and `cd START || cd DEST ; MARKER` (the new family-A shapes) — **these would have been silently exempted by the old skip list**, so removing it was not optional; leaving it would have added the M2 fixtures while quietly excusing them from the truth assertion. That is a sharper version of the point you made, and it is the reason the deletion is correct rather than merely tidy.

---

## The recovery cost you surfaced — measured, and it is zero

You were right to put `cd /wt && source .venv/bin/activate && gh pr create` in front of me rather than bury it, and right that the direction is correct regardless. But it does not cost what you feared, and I would rather answer it with a measurement than an opinion.

I ran all three revisions of the resolver over this org's own captured Bash traffic (`.claude/annunaki/**/*.jsonl`): **4,830 unique real commands, 1,030 of them carrying a routed `gh` subcommand** — i.e. commands where this function actually decides something.

| revision | routing commands that resolve a target |
|---|---|
| base `fea0dca` (what is in `main` today) | 74 / 1030 — 7.2% |
| `0655831` (first revision) | 246 / 1030 — 23.9% |
| head `1b77424` | **246 / 1030 — 23.9%** |

**This commit costs zero recovery on real routing traffic** — not "a little", zero. And the PR as a whole more than **triples** the #521 recovery rate against the base, because the AST reads shapes the old flat scan could not (notably the newline-separated `cd /path⏎gh …` form that dominates real agent traffic). Only 4 routing commands lose recovery versus base, all of them control-flow/subshell/pipeline, none from `_CWD_MOVING_COMMANDS`.

The `source` shape you worried about never co-occurs with a resolvable leading absolute `cd` in 1,030 real routing commands: the 5 commands that contain both a `cd /abs` and a `source`/`.` already resolved to `None` before this commit, for unrelated reasons. So the cost is real in principle and empirically absent here. I am satisfied it is acceptable.

---

## Residual — filed, not blocking: #1169

Phase 2 tests `words[0]`, so a **wrapper-prefixed** cwd mover still slips through:

| shape | resolver | bash | zsh |
|---|---|---|---|
| `cd DEST && builtin cd OTHER && gh …` | `DEST` | OTHER | OTHER |
| `cd DEST && command pushd OTHER && gh …` | `DEST` | OTHER | (`pushd` not found) |

The first is a genuine both-shell misroute. The second is additionally a case where **the AST is weaker than the degraded path** (degraded sees the bare `pushd` token and refuses) — the invariant `test_both_paths_agree_on_cwd_moving_builtins` exists to protect, which passes here only because it ranges over `CWD_MOVING_BUILTIN_SHAPES` and these are not in it.

**I am not blocking on it**, for reasons I want to be explicit about rather than leave as reviewer mood: it is not a regression (base returns `DEST` too), the realistic class is closed, and it has **zero instances in 4,830 real commands** (`builtin cd` / `command cd` / `command pushd` / `command popd`: 0; bare `pushd`/`popd`: 1). Converging matters here — a third round on progressively rarer shapes would cost more than it buys. Fix direction is in #1169: strip wrapper prefixes in **phase 2 only**, which is not a reversal of the #1141 no-stripping decision, because the polarity flips between the phases — an unrecognised wrapper in phase 1 causes safe *under*-detection, in phase 2 unsafe *over*-detection.

One wording note for #1169 rather than for this PR: the docstring's *"must resolve to None on BOTH paths"* is stricter than what is actually true or desired. `cd DEST > /dev/null && gh …` and `cd DEST && gh … | cat` both give AST=`DEST`, degraded=`None` — the AST being *more* precise, which is fine and worth keeping. The invariant that matters is directional: **the AST must never resolve where the degraded path refuses.**

---

## Verification at this head

- Full suite **3018 passed / 243 subtests** (from 3013/221) — reproduced locally on a tree whose `_shell_parse.py` I checksummed against the git blob.
- `pre-commit run --all-files` passes.
- CI: **21/21 `success`**, all reported against `1b77424` via the check-runs API at the SHA. In the pytest job, `Install zsh (second shell-truth oracle)` is **step 5, `success`**, ordered before `Run hook and lib tests` (step 6, `success`) — the zsh half of the oracle really ran.
- **#1155 re-merged at its current head.** It moved to `2782b3d` since my last review, so I re-tested rather than reusing the earlier result: no conflicts, merged suite **3110 passed / 250 subtests**. Still independent — its `_shell_parse.py` changes remain purely additive and touch none of `tokenize`, `iter_command_segments`, or `strip_heredocs`. **The two can land in either order.**
- PR body correction is honest: the falsified claim is not restated, and its replacement is properly narrowed to the enumerated shapes rather than made universal again.

This converged well. The drift diagnosis was the right read of my finding, and the fix is better than what I proposed.

---
Requestor: Aino Virtanen
Requestee: Lucas Ferreira
RequestOrReplied: Approved
TechDebt: #1165, #1166, #1169

parametrization:
**Second-slot review — cast at head `1b77424c9e798416cc266454df012b6010380a1b`, base `main` @ `fea0dca`.**

Aino ran two deep adversarial passes on correctness; re-running them adds nothing to the gate, so I reviewed this as Release Coordinator — release sequencing and operational risk, not a third correctness pass. Per the measurement convention, everything load-bearing below is read via `git show <sha>:<path>` or `git merge-tree` against object SHAs, never a shared working tree.

## 1. Blast radius — the three write hooks fail closed, not into the wrong repo

`_shell_parse.py` is read by eight hooks and written-through-by three: `auto_add_issue_to_board.py`, `post_label_change_wave_field_sync.py`, `post_wave_kickoff_comment.py`. I read each consumer directly at `1b77424`, not the PR's characterization of them:

- `auto_add_issue_to_board.py` does **not** call `resolve_repo_short_name`/`resolve_invocation_cwd` on its write path at all — its own comment (`.claude/hooks/auto_add_issue_to_board.py:261-264`) explains why: the created-issue URL is authoritative, so this PR's change has zero surface here.
- `post_label_change_wave_field_sync.py:721-730` and `post_wave_kickoff_comment.py:711-719` both gate explicitly: `if ambient_repo is None: … continue` / `if repo is None: … return {"action": "skip_no_repo_context", …}`. `None` is a **skip**, logged via `log_posttooluse_event` (which feeds the Annunaki monitor), never a write to a fallback/default repo.
- `resolve_repo_short_name` itself (`_shell_parse.py:1447+`) resolves off `resolve_invocation_cwd`, which falls back to the harness-supplied `cwd` / `os.getcwd()` when no leading `cd` can be honoured — i.e. the *real* process cwd, not a guessed one.
- `no_worktree_self_delete.py:270-303` (the safety gate, not a routing consumer) is fail-safe in the same direction: an unresolvable `cd` target falls back to the stdin cwd, so the guard is never relaxed on the strength of a directory that isn't actually there — confirmed by reading the block/allow branch directly, matching Aino's characterization of the previously-fail-open self-delete case.

So the residual direction holds: everything this fix cannot compute degrades to "use the real cwd" or "skip the write," never "write somewhere new." I could not find a consumer that treats `None` as "use some default repo."

## 2. Rollout risk

The fix landed as an additive commit (`0655831` is an ancestor of `1b77424`, no force-push mid-review, matches charter). If a future refactor reopens a misroute, it does not surface as a silent wrong-repo write on these three hooks — it surfaces as an increase in `skip_no_repo_context` / block-verdict-change entries through `log_posttooluse_event`, which is the org's own Annunaki error-monitor feed. That is the right failure mode to ship: noticeable, not silently destructive. Standard `git revert` applies cleanly if needed.

## 3. #1169 (wrapper-prefixed mover residual) — concur, non-blocking

Zero instances across 4,830 real captured commands, not a regression versus base, and the fix direction is already scoped (`#1169`, phase-2-only prefix stripping). Converging here is correct — I would not hold the gate open for a shape with no observed occurrence when the realistic families (A/B/C, `pushd`/`eval`/`source`/`.`/`exec`) are all closed.

## 4. #1155 / #1156 independence — reconfirmed myself, both at their **current** heads

Aino's last check was against #1155 @ `2782b3d`; #1155 has since moved to `6bc67a4419802dd3df48874580fd3d300cd6e7ef`. I re-verified rather than trusting the earlier result:

```
git merge-tree --write-tree 1b77424c9e798416cc266454df012b6010380a1b 6bc67a4419802dd3df48874580fd3d300cd6e7ef
→ eca1ed2dbc2422b0fb718f9831e2d457f3c10502   (exit 0, no CONFLICT output)
```

Hunk ranges in `_shell_parse.py` remain disjoint at these heads (#1156: 94-104 / 182-202 / 1043-1367; #1155: 33-48 / 244-259 / 443-810). **No sequencing constraint — either can land first**, and I'd land whichever clears its second Approved slot first rather than hold one for the other.

## 5. CI

Reconfirmed myself via the check-runs API at `1b77424c9e798416cc266454df012b6010380a1b`: **21/21 `success`**, independent of Aino's report.

---

Nothing here changes the gate outcome — this closes the second review slot.

---
Requestor: Santiago Ferreira
Requestee: Lucas Ferreira
RequestOrReplied: Approved
TechDebt: #1165, #1166, #1169

Issue #378, opened 2026-05-11
tech-debt(hook): validate_review_comment_format Direction-table semantic disagreement vs charter (post-#302 sibling)


parametrization:
## Source

Surfaced during PR #377 (parser-fixture batch for 5 validate_* hooks). Aino's commit `1ffbe4a` (closes #302) documents the disagreement in the commit message's "Out of scope" section. Wanjiku-375 separately flagged it during PR #375 review on the same trip through the hook surface, and Wanjiku-375 deferred filing to keep TechDebt-line context clean post-merge.

## The disagreement

`validate_review_comment_format.py` uses **"branch-author lastname matches Requestee → swap signal → BLOCK"** as its detection heuristic. This works for **Approved / ChangesRequested** verdicts (where Requestee=PR-author per charter Direction table, so if the comment-poster's branch matches Requestee, they accidentally addressed themselves).

But the charter `pull-requests.md § Comment-Based Reviews` Direction table (now codified post-PR #349/#356/#375) defines **four** direction values with different role bindings:

| RequestOrReplied | Requestor | Requestee |
|---|---|---|
| Approved | reviewer | PR author |
| ChangesRequested | reviewer | PR author |
| Request | PR author | reviewer |
| Reply | PR author | reviewer |

For **Request / Reply** direction, the PR-author IS the Requestor and the reviewer IS the Requestee. The current hook's "branch-author == Requestee = swap" heuristic doesn't account for this role-inversion — it may either (a) false-block legitimate Request/Reply traffic or (b) silently allow swap-mistakes that would have been caught for Approved.

The hook's existing tests (added by PR #377 / commit `1ffbe4a` — 42 new fixtures) pin the **current** behavior, but several are explicitly noted as `# pinned current behavior, not necessarily charter-correct`. The semantic-realignment to the full Direction table is deferred.

## Pre-filing dup-search performed

- `gh issue list --search "validate_review_comment_format direction"` → 1 match: #356 (CLOSED — different hook scope: `validate_pr_review` BLOCKED-message text, not detection logic). NOT a dup; distinct hook + distinct concern (text vs logic).
- `gh issue list --search "rationalize validate_review_comment_format"` → no matches; no rationalization umbrella to append drift-evidence to.
- Multi-layer separation: this is **logic-level** (detection heuristic), not text-level (#356/#372) nor coverage-level (#302). New distinct surface.
[...]

Issue #164, opened 2026-04-21
bug: validate_pr_review hook dedupes reviewers by lastname — Lucas Ferreira + Santiago Ferreira count as 1


parametrization:
## Summary

\`.claude/hooks/validate_pr_review.py:175\` — the hook adds reviewers to a set keyed by lowercased lastname:

\`\`\`python
result.reviewers.add(reviewer_lastname.lower())
\`\`\`

When two distinct reviewers share a lastname (e.g., **Lucas Ferreira** and **Santiago Ferreira**), the set deduplicates them and the hook reports \"1/2 required peer reviews\" even though two distinct people approved.

## Repro (2026-04-21)

deploy#81 had Lucas Ferreira's Approved comment + Santiago Ferreira's Approved comment, both with valid TechDebt: lines. Hook blocked merge with:

\`\`\`
BLOCKED: PR #81 has 1/2 required peer reviews. At least TWO reviews from distinct non-authors are required before merge.
\`\`\`

Workaround: spawn a 3rd reviewer with a different lastname (e.g., Aino Virtanen).

## Fix scope

Switch the dedup key from lastname to full Requestee name (already extracted into \`requestee_name\`). One-line change:

\`\`\`diff
-                if reviewer_lastname.lower() != branch_author_lastname.lower():
-                    result.reviewers.add(reviewer_lastname.lower())
+                if reviewer_lastname.lower() != branch_author_lastname.lower():
+                    result.reviewers.add(requestee_name.lower())
\`\`\`

Branch author check stays on lastname (since branch format is \`{Initial}.{Lastname}/...\`).

Add a negative-match test case: two distinct reviewers with same lastname → counts as 2.

## Wave

P2W9 — fits the in-progress hook-architecture mini-sprint cluster (#125 meta-issue). Same hook as #147 (TechDebt over-enforcement) — can land in a combined PR if Wanjiku wants.

## Origin

Surfaced during 2026-04-21 overnight bug-bash work, blocked deploy#81 merge after autonomous reviewer dispatch picked Lucas + Santiago without checking lastname collision.

Issue #179, opened 2026-04-22
bug: validate_pr_review branch-regex misses dash-separator branches — blocks merge on all A.X-NNNN-* branches


parametrization:
## Summary

`.claude/hooks/validate_pr_review.py` `extract_branch_author_lastname()` matches the regex `[A-Za-z]\.([A-Za-z]+)/` — slash-separator only. Current branch naming convention has drifted to dash-separator (e.g., `A.Virtanen-0169-hook15-sentinel-marker`, `A.Idrissi-0114-auto-set-env-substring`, `A.Idrissi-0114-*`). On dash-separator branches:

- `extract_branch_author_lastname()` returns None
- `check_comment_reviews()` is never invoked
- `comment_review_result.reviewers` stays empty
- `total_distinct` is 0 regardless of how many valid `Approved` charter-format comments the PR has
- Merge blocks with `0/2 required peer reviews` even on well-reviewed PRs

## Repro

Observed today on PRs #174 and #178. Both have two distinct reviewers posting canonical-format `RequestOrReplied: Approved` comments with `TechDebt:` lines. Both merges blocked.

## Fix

One-line regex update to accept either `-` or `/`:

```python
match = re.match(r"[A-Za-z]\.([A-Za-z]+)[-/]", head_ref)
```

Plus regression tests covering both separators and the mismatched-convention negative case (branch with neither `/` nor `-` between `LastName` and the rest — should still fail gracefully).

## Related drift

CLAUDE.md documents the convention as \`{FirstInitial}.{LastName}\\{IIII}-{issue-name}\` (slash intended; backslash is probably markdown escape). Actual practice has drifted to dash. This hook fix makes the hook lenient; a separate followup issue should standardize the convention and update CLAUDE.md.

## Acceptance

- Regex accepts both `A.X/0001-foo` and `A.X-0001-foo`
- Unit tests added for each separator, plus a negative case
- PR #174 and PR #178 merge after this lands (both are currently blocked on this bug)
- Fix branch itself uses slash-separator so it can merge under the OLD regex (self-verifying)

PR #1173, opened 2026-07-28
fix(hooks): compare people, not surnames, in the review identity gates (#1172)

### LINKED ISSUE #1172: bug(hooks): validate_review_comment_format swap-check compares lastname only — blocks correct Ferreira/Ferreira verdicts (sibling to #164)
## Summary

`.claude/hooks/validate_review_comment_format.py`'s Requestor/Requestee swap-detection heuristic (added post-#244, `extract_branch_author_lastname` vs. `Requestor` lastname) compares **lastnames only**, so it false-positive-blocks a correctly-formatted verdict whenever the reviewer and the PR author share a surname. This is the same defect class already fixed in the sibling counting hook at #164 (`validate_pr_review.py` deduped reviewers by lastname — same Lucas Ferreira / Santiago Ferreira collision) — #164's fix did not touch this hook, which has an independent lastname-only comparison.

## Repro (2026-07-27)

PR #1156, branch `L.Ferreira/1151-cd-misroute-families` (author Lucas Ferreira). Santiago Ferreira, posting a correctly-formed second-slot `Approved` verdict (`Requestor: Santiago Ferreira`, `Requestee: Lucas Ferreira` — the correct, non-swapped assignment per charter), was blocked:

```
BLOCKED: Requestor/Requestee appears swapped. Requestor should be the reviewer
(who is doing the review). Requestee should be the PR author (who is receiving
the review). The branch author is Ferreira — they should be the Requestee, not
the Requestor.
```

`extract_branch_author_lastname` returns `"Ferreira"` from the branch name; `_extract_lastname(requestor_value)` on `"Santiago Ferreira"` also returns `"Ferreira"`. The heuristic (`.claude/hooks/validate_review_comment_format.py:916`) has no way to tell the reviewer and the PR author apart once they share a lastname, and blocks both `gh pr comment` and the REST comment-create arm identically, so there is no observable-body workaround.

## Fix scope

Same shape as the #164 fix: switch the swap comparison from lastname-only to full name (or first-initial + lastname, matching the branch-format grain `{FirstInitial}.{LastName}`) wherever the reviewer roster has a same-lastname collision. `extract_branch_author_lastname` only has the lastname available from the branch string (`{Initial}.{Lastname}/...`), so the fuller fix needs either:
- widening the branch-name regex to also capture the first initial and comparing `Requestor`'s first-initial + lastname against it, or
- resolving the branch author's full name via `gh pr view --json author` (already done elsewhere in this file for other checks) instead of parsing it out of the branch string.

Add a negative-match test: `Requestor: Santiago Ferreira` / branch `L.Ferreira/...` (distinct reviewer, same lastname as author) must NOT block.

## Impact

Blocks legitimate, correctly-addressed verdict comments whenever this org's two same-surname roster members (Lucas Ferreira, Santiago Ferreira) review each other's work in either direction. No workaround exists short of a hook fix or an owner override — the REST-arm evasion path #932 closed also has this check, so it is not bypassable by switching posting mechanisms.

## Wave

wave-29 — surfaced while driving PR #1156's second-slot review to completion.

parametrization:
Closes #1172.

## The defect

Both review hooks answered *"is this person the branch author?"* by comparing **surnames**. The roster holds 78 names and it contains collisions, so on `L.Ferreira/1151-cd-misroute-families` (author **Lucas Ferreira**) a verdict correctly authored by **Santiago Ferreira** put `Ferreira` on both sides of the test. One wrong answer, two hooks failing in opposite directions:

| Hook | Failure |
|---|---|
| `validate_review_comment_format` | **BLOCKED** the verdict as a swap — on the comment-post path *and* on the REST comment-create arm added for #932, so there was no observable-body workaround and it could not be posted at all. |
| `validate_pr_review` | Discarded the same verdict as a **self-review**, leaving #1156 at 1 of 2 approvals. |

Correct reviewer behaviour was indistinguishable from the swap the gate exists to catch, and the diagnostic said *"the branch author is Ferreira"* — a surname two roster members share, so it read as being about someone else.

**#164 had already fixed this exact collision** for `validate_pr_review`'s reviewer *dedup* key (key on the full name). The identity comparisons a few lines away kept the surname, so the same defect stayed live in two more places. That recurrence is the point of #1150.

## The fix

The branch prefix `{FirstInitial}.{LastName}` carries more signal than the surname, so identity is now **first-initial + lastname**, through one shared predicate — `charter_trailer.is_branch_author`, the module that already owns the trailer convention and exists precisely so the two hooks cannot drift on a shared definition. Its docstring states the rule that motivated it: *a second definition should be impossible, not kept in sync.*

**This is a narrowing, not a widening.** The swap check was not disabled or loosened:

- A genuinely swapped verdict — the PR author naming themselves as `Requestor` — **still blocks**: their initial *and* surname match the branch.
- A real self-review **is still excluded** from the reviewer set for the same reason.
- Where an initial is underivable on either side (a bare surname like `Ferreira`; a wave-merge head ref with no `{Initial}.{Lastname}` prefix) the comparison **falls back to surname-only** — the stricter pre-#1172 answer. That is fail-CLOSED at both call sites (refuse the comment / do not count the reviewer), so the degraded path cannot admit an unreviewed merge.

Gutting the heuristic instead would have traded a false positive for a false negative in a gate. `validate_pr_review` counts a swapped verdict wrong, which is why the check exists.

## Scope confirmations

- **Verdict-only scoping still holds.** The swap heuristic applies to `Approved` / `Changes Requested` only, never `Request` / `Reply`, where the role bindings legitimately invert (#378). Pinned by `test_non_verdict_directions_remain_out_of_scope` so a future widening of the *identity* comparison cannot quietly widen the *directions* it applies to.
- **Third-site sweep.** Every other roster-aware hook was checked for the same pattern — `validate_commit_identity`, `post_wave_kickoff_comment`, `block_squash_wave_merge`, `warn_oversized_brief`, `enforce_ontology_context`, and `pr_review_state` (which carries the lastname for display only, never a comparison). **No third site.** The pattern was confined to these two hooks.
- **`validate_pr_review` fixed in the same PR, deliberately.** Fixing only the format hook would have unblocked *posting* the verdict while the counting hook still discarded it — #1156 would have stayed at 1 of 2 approvals and the fix would have looked complete while being inert.

## Tests — both directions, at every layer

40 new tests. Every "this reviewer is now admitted" assertion is paired with the true positive that must still be refused; removing a false positive from a gate is only a fix if the true positive survives it.

| Layer | Same-surname colleague admitted | True positive still caught |
|---|---|---|
| Predicate (`test_charter_trailer_identity.py`) | `test_same_surname_different_person_is_not_the_author` | `test_the_author_themselves_is_still_the_author` |
| Format hook (`SurnameCollisionTests`) | `test_same_surname_reviewer_is_allowed` (+ ChangesRequested) | `test_branch_author_naming_himself_still_blocks` (+ ChangesRequested) |
| Format hook, REST arm (#932) | `test_rest_arm_allows_the_same_surname_reviewer` | `test_rest_arm_still_blocks_the_branch_author` |
| Counting hook (`SurnameCollisionSelfReviewTests`) | `test_same_surname_reviewer_is_counted`, `test_both_ferreiras_reach_the_two_reviewer_threshold` | `test_branch_author_self_review_is_still_excluded` |

Also pinned: the `""` initial default degrades to the stricter surname-only answer (`test_omitted_initial_degrades_to_the_stricter_surname_answer`), and the initial actually reaches the exclusion from the head ref in production (`test_resolve_passes_the_branch_initial_through`) — without that wiring test the unit tests would pass while every real merge still used the default.

**Fixture liveness verified**, per `feedback_fixture_makes_guard_assertion_inert`: the Santiago fixture was run against the pre-fix predicate and confirmed to block (test red), against the post-fix predicate to allow (test green), with the Lucas true positive blocking under both.

### Correction at `d5d7dec` — the REST-arm row above was not true at `3be5b02`

Nino Kavtaradze's mutation harness caught it. The original single REST test used literal `\n` two-character sequences, so the body was one physical line, `_direction_is_verdict` returned False, and `check()` returned at the verdict-scope gate **without ever reaching `is_branch_author`**. It was green against the unfixed predicate. The table claimed REST coverage the test did not provide — my error, and the same inert-assertion class this wave has now caught three times.

Repaired at `d5d7dec`: real newlines, both directions (the two rows above), plus `test_rest_arm_fixture_actually_reaches_the_identity_check` — an explicit liveness guard, because the way this fixture failed was returning `allow` for a reason unrelated to identity, which is indistinguishable from a correct allow in the test result.

Re-verified against his M1 mutation (`is_branch_author` reverted to lastname-only): **6 red at `3be5b02`, 7 red at `d5d7dec`** — the +1 is exactly this test becoming live. No production defect; he confirmed the real REST arm end-to-end independently.

Also at `d5d7dec`: `.claude/team/charter/pull-requests/reviews.md` still stated the rule as `Requestor.lastname == branch-author.lastname`. Rewritten to the rule the code now implements, fallback direction included. Charter text describing a rule the code no longer implements is precisely the drift this role exists to prevent, and it belonged in the PR that changes the rule.

**#1176 deliberately not pulled in.** His `_direction_is_verdict` fail-open finding is byte-identical at `fea0dca`, pre-existing, and is the root cause of the inert fixture above. It deserves its own scoping rather than an expansion of a critical-path PR.

## Verification

- Full suite at `d5d7dec`: **3,041 passed**, 170 subtests (baseline ~2,999 on `main`). Green.
- `ruff format` + `ruff lint`: clean. `mypy`: `Success: no issues found in 181 source files`.
- Sync-drift gate: `OK: pre-commit config mirrors all CI-enforced checks`.
- Pre-commit and pre-push hook sets both ran clean on commit and push. No `--no-verify`.
- cspell is out of scope for this diff — the `Spellcheck` job's `files` glob covers `.md` only; all changes here are `.py`.

## Note for reviewers

Per the charter's reviewer-pairing rule and the defect itself, **a reviewer sharing my surname should not be assigned until this has landed** — that is the very condition the PR fixes.

Separately, and reported rather than worked around: `validate_review_comment_format`'s `is_comment_command` matched my own `git commit -m` because the *message text* documented the comment-post command, and blocked the commit with "this command posts a PR comment, but its body cannot be read." Same shape as #1152 (`validate_commit_identity` blocking heredocs that merely document an interpreter invocation). I committed with `-F <message-file>` — the gate's actual subject, PR review comments, is untouched by that — and flagged it to the orchestrator rather than silently routing around it. Worth a follow-up issue: these gates scan the whole command string and cannot distinguish a command from prose *about* a command.


parametrization:
**Review: LGTM — design/consequence pass (second-slot reviewer)**

Verified at head `3be5b0271f38fddf1b6edc76b41187a903fd09ed`: 14/14 checks SUCCESS, `MERGEABLE`/`CLEAN` (re-checked live, not from PR-body prose). Read `charter_trailer.py`, both hook call sites, and the new test classes directly via the contents API at that SHA.

**1. Does the shared predicate genuinely consolidate, or move the duplication?**
Mostly yes, with one residual gap. `is_branch_author` + `name_first_initial`/`name_lastname`/`branch_author_first_initial` are single-defined in `charter_trailer.py` and both hooks import (not reimplement) them — confirmed by reading the import blocks, not just trusting the PR description. But `extract_branch_author_lastname` — the function that decides whether a branch prefix is parseable *at all*, one step upstream of `is_branch_author` — is still independently defined in each hook, and the two copies have already diverged: `validate_pr_review`'s accepts both `/` and `-` separators (fixed by #179), `validate_review_comment_format`'s accepts `/` only, and that gap is *pinned as intended* by `test_dash_separator_not_supported`. On a dash-separated branch the format hook's whole swap check short-circuits to an early `allow` before `is_branch_author` is ever reached — so there's no test that would fail if these two hooks diverge on "is this a parseable branch," only tests that each hook calls the shared predicate once it's already decided the branch is legible. I verified this doesn't reopen a merge-gate hole (the counting hook's own copy of the parser is the dash-capable one, so self-review exclusion still works on such branches) and that no branch currently open uses dash form (checked all 136 `{Initial}.{Lastname}[-/]` remote branches). Filed **#1175** rather than blocking on it — it's pre-existing, not introduced by this diff, and the safety-relevant half (counting) is unaffected.

**2. Fallback paths.** Traced both: the wave-merge head-ref case (`deployments/phase-N/wave-M`) doesn't go through the initial-fallback in `is_branch_author` at all — it uses the pre-existing empty-sentinel path (`branch_author_lastname=""`), which `is_branch_author` refuses to match unconditionally (`if not branch_lastname: return False`), so that's fail-closed by construction, not by the new initial logic. The actual initial-unknown fallback in `is_branch_author` is structurally near-unreachable from the branch side (the same regex that yields a lastname also always yields an initial), so in practice it only fires when the *Requestor field value* is malformed (single-token). That's a defensive backstop for bad input, not a live path for well-formed comments. Confirmed fail-closed in both directions per the paired true-positive tests.

**3. Blast radius / false negatives.** `is_branch_author` can only get *stricter or equal* relative to pre-fix surname-only matching (initial adds a narrowing condition, lastname match is still required unconditionally) — so this change cannot newly admit a genuine self-review or swap into the count. Confirmed by reading the comparison logic directly, not just the paired-test table in the PR body.

**4. `{FirstInitial}.{LastName}` as an identity source.** Correctly acknowledged as the ceiling of what the branch prefix carries — two roster members sharing both initial and surname would still collide (none do today across the 9-person parent roster, but the org-wide union is 78 names and growing). That's a branch-naming-convention limit, not a hook bug, and not fixable at this layer; worth keeping in mind as the roster grows but not a blocker here.

Must-fix: None
Tech-debt: #1175 (`validate_review_comment_format` dash-separator branch parsing diverges from `validate_pr_review`, pre-existing, filed by this review)

---
Requestor: Weronika Zielinska
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1175

parametrization:
## Merge-gate review — PR #1173 @ `3be5b0271f38fddf1b6edc76b41187a903fd09ed`

Every measurement below was taken at head SHA `3be5b02`, in a private detached worktree (`rev-parse HEAD` asserted against the briefed SHA, `status --porcelain` empty before each run) and against files fetched from `contents?ref=3be5b02`.

**The fix itself is correct and complete at both call sites.** I verified that by mutation rather than by reading, and it holds. One test in the new set is inert, and it is the one the PR body cites as covering the REST arm — that is the whole of my must-fix.

### Both call sites are genuinely fixed (mutation-verified)

I reverted or broke each layer in turn and confirmed the suite reds. An assertion no mutation can trip is inert; these all trip.

| Mutation at `3be5b02` | Result |
|---|---|
| M1 — `is_branch_author` reverted to lastname-only (the original defect) | **6 red** across all three layers: predicate, format hook (Approved + ChangesRequested), counting hook (counted / threshold / ChangesRequested ledger) |
| M2 — drop `branch_author_initial=...` at the `resolve_review_verdicts` wiring | **2 red**, incl. `test_resolve_passes_the_branch_initial_through` — and `test_pr_review_state.py::test_branch_author_lastname_passed_to_check` for the second driver |
| M3 — format hook drops `branch_initial` at its call site | **2 red** |
| M4 — fallback flipped permissive (unknown initial ⇒ not the author) | **5 red**, incl. `test_omitted_initial_degrades_to_the_stricter_surname_answer` |
| M5 — verdict-only scoping removed (heuristic widened to all directions) | **7 red** (4 tests + 3 subtests), incl. `test_non_verdict_directions_remain_out_of_scope` |

M2 is the important one. The wiring test is load-bearing exactly as claimed: without it the unit tests stay green while every real merge runs on the `""` default. It is the only test in `test_validate_pr_review.py` that catches that, and it does catch it.

**Shared predicate — no bypass.** `is_branch_author` is the only identity comparison left in either hook. Both import it; neither retains a private one (`_extract_lastname` / `_name_lastname` are aliases onto `charter_trailer`, not reimplementations). Verified truth table on branch `L.Ferreira/1151-...`:

```
Lucas Ferreira    -> True      Santiago Ferreira -> False
L.Ferreira        -> True      S.Ferreira        -> False
Ferreira (bare)   -> True      Nadia Khoury      -> False
```

**Fallback is the stricter direction at both call sites**, confirmed behaviourally, not just by test name: a bare-surname `Requestor` on `L.Ferreira/...` **blocks** at the format hook and is **excluded** from the reviewer set at the counting hook. Unknown branch initial degrades to the pre-#1172 over-exclusion. The wave-merge empty-lastname sentinel returns `False` for everyone, unchanged from `main` — that path is main#294's pre-existing behaviour, not a new permissive default introduced here. Nothing this PR adds reaches a permissive default.

**Third-site sweep — spot-checked, holds.** `validate_commit_identity` keys on the **full** name against the roster (`if name not in roster`), never a surname. `apply_implementor_labels` keys on `FIRSTNAME_LASTNAME`. `pr_review_state` carries `branch_author_lastname` for display only and routes its comparison through the single shared `resolve_review_verdicts`. A repo-wide search for last-token surname extraction outside `charter_trailer` returns nothing. No third site.

### Must-fix 1 — `test_rest_comment_create_arm_agrees_with_the_gh_arm` is inert

`.claude/hooks/tests/test_validate_review_comment_format.py:660`. Its fixture body uses **literal two-character `\n` sequences**, not newlines:

```python
body = (
    "Looks correct.\\n\\n---\\nRequestor: Santiago Ferreira\\n"
    "Requestee: Lucas Ferreira\\nRequestOrReplied: Approved\\nTechDebt: #1176\\n"
)
```

So the body is one physical line. `_direction_is_verdict(body)` is `False`, and `check()` returns at the verdict-scope gate **before reaching `is_branch_author` at all** — the field extraction on that body yields `'Santiago Ferreira\nRequestee: Lucas Ferreira\nRequestOrReplied: ...'`, i.e. garbage. The test asserts `assertIsNone` and passes for the wrong reason.

It survived **both M1 and M3** — it is green against the unfixed predicate. The PR body's test table lists it as the REST-arm half of the format-hook row; that coverage claim does not hold. This is the same class as `feedback_fixture_makes_guard_assertion_inert`.

To be clear about severity: **there is no production defect here.** Both body-extraction arms feed the single `check()`, so the REST arm is structurally fixed, and I confirmed it end-to-end — a swapped verdict via `-f body=` with real newlines and via an `--input` JSON payload both **block**, and Santiago's correct verdict is **allowed** on both. The code is right; the test pinning it is not.

The repair is one line — real newlines in that fixture. I ran it: with that change the test **reds under M1** (joins `test_same_surname_reviewer_is_allowed`) and is **green at `3be5b02`**. Verified in both directions before asking for it.

Worth adding alongside it, cheaply: the REST arm has no true-positive counterpart. Every other collision assertion in the class is paired; that one is not.

### Must-fix 2 — the charter still states the rule this PR replaces

`.claude/team/charter/pull-requests/reviews.md:30` still reads:

> Within that scope the hook blocks when `Requestor.lastname == branch-author.lastname`

That is the defect, written down as the rule. Since #1150's class is precisely a hook implementing a comparison the prose doesn't match, the sentence should move to first-initial + lastname with the surname-only fallback named. One sentence; in scope for the PR that changes the rule.

### Tech-debt found, filed, not charged to this PR — #1176

Reviewing my own verdict through the hook surfaced a **fail-open** in the same gate. `_direction_is_verdict` (`validate_review_comment_format.py:644`) — the scope gate that decides whether the swap check runs at all — is the one field read #932/#934 never converted to the shared helper. It is a bare unscoped first-match over the **raw** body. So on a genuinely swapped trailer, branch `L.Ferreira/...`, at `3be5b02`:

| Body | Gate |
|---|---|
| swapped trailer alone (control) | **block** — correct |
| prose quoting a non-verdict direction above the trailer | **ALLOW** |
| the same quoted inside a fenced code block | **ALLOW** |

The trailer still parses as Approved, so the counting hook counts it, while the gate that exists to catch the swap has switched itself off. Quoting a trailer in reviewer prose is enough — and since it reads the raw body, `strip_code_regions` does not protect it.

**Byte-identical between `fea0dca` and `3be5b02`** — pre-existing, and you correctly did not touch it. Filed as #1176 rather than charged here; I am not expanding a critical-path PR for it.

It is, however, the root of must-fix 1: the REST fixture is inert *because* its escaped one-line body makes this same helper return `False`. Fixing the fixture is still right and still yours; #1176 is the deeper repair.

### Verified clean

- **CI: 14/14 COMPLETED SUCCESS**, read from `check-runs` bound to `3be5b02` (not the PR-level rollup). `MERGEABLE` / `CLEAN`.
- **Suite: 2,999 → 3,039 (+40)**, 167 → 170 subtests. I ran the merge-base (`fea0dca`) myself for the baseline rather than taking the figure; both numbers reproduce exactly.
- `ruff format` (181 files) + `ruff lint` clean; `mypy` `Success: no issues found in 181 source files`; sync-drift gate `OK`.
- **cspell genuinely does not apply** — confirmed, not accepted: the action's `files` glob in `docs.yml` is the authored-prose `.md` set (charter / skills / ontology / root docs), and the workflow `paths:` filter has no `**/*.py`. This diff is all `.py`.
- Scope discipline holds — M5 shows the verdict-only restriction is pinned, and the three `Request` / `Reply` / `Replied` subtests are live.
- The two modified pre-existing tests were **tightened, not weakened**: one gains the `initial` assertion, the other only updates a docstring while keeping its surname-final fixture. Also checked the direction of the `_name_tokens` empty-token change: it makes a trailing-period name resolve to a real surname where it previously resolved to `""`, which tightens the gate. Correct direction.

Nothing else to raise. Turn the two items and I will re-approve immediately — I have the mutation harness standing by, so re-verification is a few minutes. Good judgment finding `validate_pr_review` yourself and fixing it in the same PR; the format-hook-only fix would have looked complete and been inert, which is the outcome this review exists to prevent.

Reported and not routed around, matching your own handling of the `is_comment_command` block: nothing gated me during this review.

---
Requestor: Nino Kavtaradze
Requestee: Aino Virtanen
TechDebt: #1176

parametrization:
## Merge-gate re-review — PR #1173 @ `d5d7decfd86a3ea038714d57dbfbbbb1601426dd`

Both must-fixes turned. **Approved.** Fresh detached worktree at `d5d7dec`, `rev-parse HEAD` asserted, `status --porcelain` empty before every run.

### The 6 → 7 delta reproduces

I re-ran M1 myself (`is_branch_author` reverted to lastname-only) at the new head. **7 red**, against 6 at `3be5b02`, and the +1 is exactly `test_rest_arm_allows_the_same_surname_reviewer`. Your measurement is correct as stated.

### The liveness guard is itself live — mutation-verified

Confirming a liveness assertion by reading it would repeat the mistake it exists to catch, so I regressed the fixture rather than trusting it: **M6** — revert `_rest_command`'s body to the literal `\n` two-character form, production fix left intact. Two tests red:

- `test_rest_arm_fixture_actually_reaches_the_identity_check` — the guard fires.
- `test_rest_arm_still_blocks_the_branch_author` — the paired true positive fires too, because a swapped verdict that never reaches the identity check returns `allow` instead of `block`.

`test_rest_arm_allows_the_same_surname_reviewer` stays green under M6, on its own. That is the whole argument for the other two, and it is now demonstrated rather than asserted: the fixture regression is caught **twice, independently**, by the liveness assertion and by the true positive — and not at all by the allow-side test that was the original inert one. Your reasoning about `allow`-for-the-wrong-reason being indistinguishable in a test result was right, and the structure you chose covers it from both sides.

### Production code is untouched — earlier verification carries over by construction

`charter_trailer.py`, `validate_pr_review.py`, `validate_review_comment_format.py` are **byte-identical** (sha256) between `3be5b02` and `d5d7dec`. The push is the test file plus the charter `.md`, nothing else. So M1–M5 from the first round still hold, and M1 was re-run explicitly regardless. **Both call sites remain genuinely fixed**: `validate_review_comment_format` blocks the post, `validate_pr_review:1051` no longer discards the verdict as a self-review.

### Charter prose — accurate

`.claude/team/charter/pull-requests/reviews.md` now states first-initial + lastname via `charter_trailer.is_branch_author`, names the branch prefix as the source of the discriminator, and names the lastname-only fallback as the stricter direction. It matches the code, including the verdict-only scoping. Correct as written.

### PR body correction — accurate

Checked the `### Correction at d5d7dec` section against the artifacts rather than reading it for tone. The mechanism, the "green against the unfixed predicate" claim, the 6→7 figure, and the `#1176 not pulled in` statement all hold. The test table's REST row now names two tests that exist and do what the row says.

### Verified at `d5d7dec`

- **21/21 checks SUCCESS**, read from `check-runs` bound to the SHA. `MERGEABLE`/`CLEAN`. The doc jobs ran this time — `Spellcheck (cspell)`, `Markdown lint`, `Internal link-check (lychee)`, `Config — Workflow actionlint` — because the push touched `.md`.
- Suite **3,041 passed**, 170 subtests.
- `ruff format` (181 files) + `ruff lint` clean; `mypy` `Success: no issues found in 181 source files`; sync-drift gate `OK`.
- Diff scope is the two must-fixes only. #1176 correctly not pulled in.

### One nit, explicitly not gating

The Verification bullet still reads *"cspell is out of scope for this diff — all changes here are `.py`"*. That was true at `3be5b02` and is stale at `d5d7dec`: the charter `.md` sits inside the cspell `files` glob (`.claude/team/charter/**/*.md`), which is why `Spellcheck (cspell)` ran — and passed. No risk, no action required, and I am not holding the gate for a sentence. Worth a line if you touch the body again.

Good work on this round. You took the inert-fixture finding as a coverage claim rather than a style note, fixed it in both directions plus a liveness assertion, corrected the PR body in public instead of quietly, and declined to absorb #1176 into a critical-path PR. That is the right disposition on all four.

---
Requestor: Nino Kavtaradze
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1176

parametrization:
**Review: LGTM — design/consequence pass (second-slot reviewer, re-cast)**

Re-verifying at the new head after my prior `Approved` at `3be5b027` went content-stale. Verified live: head `d5d7decfd86a3ea038714d57dbfbbbb1601426dd`, 21/21 checks SUCCESS, `MERGEABLE`/`CLEAN`. Independently confirmed byte-identity of `charter_trailer.py`, `validate_pr_review.py`, `validate_review_comment_format.py` between the two heads via git blob SHA comparison (not taking the "byte-identical" claim on trust) — all three unchanged. Only `.claude/hooks/tests/test_validate_review_comment_format.py` (+45/-8) and `.claude/team/charter/pull-requests/reviews.md` (+1/-1) changed since my prior review.

My prior design/consequence analysis — the shared-predicate consolidation, the fallback-path tracing, the blast-radius check, and the `{Initial}.{Lastname}` collision ceiling — was scoped entirely to the three unchanged source files, so it stands without re-derivation. My #1175 filing (`extract_branch_author_lastname` dash-separator divergence between the two hooks, pre-existing, non-blocking) is likewise unaffected.

Spot-checked the delta itself: the repaired REST-arm fixture now builds real newlines rather than literal `\n` escapes, and the added `test_rest_arm_fixture_actually_reaches_the_identity_check` is exactly the right kind of check for a fixture that could pass for the wrong reason (an early `allow` unrelated to identity is indistinguishable from a correct one without a liveness assertion) — same discipline as the SHA/blob check above. No new concerns.

Must-fix: None
Tech-debt: #1175 (filed on the prior pass, still open, non-blocking)

---
Requestor: Weronika Zielinska
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1175

Issue #1171, opened 2026-07-28
tech-debt(commit-identity): backslash line-continuation between a data sink and `| bash` evades the new pipe-to-shell heredoc block


parametrization:
Found during the security-lens merge-gate review of PR #1155 (Nino Kavtaradze, second-slot reviewer). Filed as tech debt rather than a merge blocker — see "Why not gating" below.

## Shape

```
cat <<'D' | \
bash
git -c user.name="X" -c user.email="x@y.z" commit -m z
D
```

A backslash line-continuation between the data sink and the interpreter. The body **is executed** — hand-verified in both shells with a marker that only materialises if a shell really runs it:

```
$ bash c.sh -> NINO-EXEC-MARKER
$ zsh  c.sh -> NINO-EXEC-MARKER
```

## Measurement

At the CLI entry point, plain identity commit body:

| shape | base `fea0dca` | head `2782b3d` |
|---|---|---|
| `cat <<'D' \| bash` (control) | ALLOW | **BLOCK** |
| `cat <<'D' \| \` + newline + `bash` | ALLOW | **ALLOW** |

So this is not a regression — base allowed it too. It is a **gap in the new pipe-to-shell blocking path** that #1155 introduces: the control shape is blocked, its line-continued twin is not.

## Mechanism

`classify_heredocs` / `strip_data_heredocs` iterate **physical** lines and call `_scan_command_line` on each. For the opener line the segments are:

```
sep=''  seg="cat <<'D' "  head='cat'
sep='|' seg=' \'          head=None
```

The continued segment is a bare backslash, which does not tokenize, so `_segment_head_command` returns `None`. The downstream walk then evaluates `None in SHELL_INTERPRETERS` -> `False` and concludes "not an interpreter", leaving `is_code=False`. The `bash` on the next physical line is never considered, and the body-start line is also off by one relative to what the shell does (the shell begins the body after the complete logical line).

This is the same **fail-open downstream branch** as #1168: the owner-position check is an allowlist that fails closed (`head not in HEREDOC_DATA_SINKS` -> CODE), while the downstream pipe-follow is a denylist that fails open (`head in SHELL_INTERPRETERS` -> CODE, so an unresolvable head stays DATA).

## Why not gating on #1155

[...]

Issue #1170, opened 2026-07-28
tech-debt(commit-identity): heredoc piped to an interpreter through a FIFO created in the same command bypasses the classifier


parametrization:
Found during the security-lens merge-gate review of PR #1155 (Nino Kavtaradze, second-slot reviewer).

## Shape

```
mkfifo p; bash < p & cat <<'D' > p
git -c user.name="X" -c user.email="x@y.z" commit -m z
D
```

The heredoc body **is executed** — hand-verified in both shells with a marker body that only materialises if a shell really runs it (`printf 'NINO-%s-MARKER\n' EXEC`):

```
$ bash shape.sh -> NINO-EXEC-MARKER
$ zsh  shape.sh -> NINO-EXEC-MARKER
```

## Measurement

Driven through the CLI entry point (`validate_commit_identity.py`, stdin JSON, exit code), with a **plain identity commit** body — i.e. no incidental `bash -c` text for the raw scanner to match on:

| shape | base `fea0dca` | head `2782b3d` |
|---|---|---|
| `mkfifo p; bash < p & cat <<'D' > p` | **BLOCK** (exit 2) | **ALLOW** (exit 0) |
| `mkfifo p2; sh < p2 & tee p2 <<'D' > /dev/null` | **BLOCK** | **ALLOW** |

This is a genuine block -> allow on an executable shape, and unlike #1167 the base block is **not** an artifact of `\bsh\b` matching inside a filename: base blocks via `_HEREDOC_RE`, because an interpreter word (`bash`) appears textually before the `<<` on the opener line.

## Why it is not a defect of #1155

The rule that blocked this at base — "an interpreter appears anywhere earlier on the opener line" — is exactly the over-broad rule #1152 exists to narrow. It blocked `bash build.sh && cat > notes.md <<'EOF'` (prose body) as a false positive, which is the bug being fixed. The FIFO shape is a case where that over-broad rule happened to be right.

Closing this properly needs the classifier to track a filesystem object across segments: notice that `cat <<'D' > p` writes to the same path an interpreter reads from (`bash < p`) elsewhere in the command. That is real dataflow analysis, well beyond the single-segment sink resolution the classifier does today, and it cannot be approximated by "any interpreter in the command forces CODE" without reintroducing #1152.

Same class as #1167 (bytes reach a shell through a filesystem object rather than a pipe), but distinct in evidence: #1167 is ALLOW at base **and** head, so #1167's record does not cover this one.

## Suggested bound

[...]

Issue #1168, opened 2026-07-27
tech-debt(commit-identity): non-interpreter relay downstream of a data-sink heredoc (xargs bash -c) escapes the pipe-follow classifier


parametrization:
## Shape

`classify_heredocs` (#1155) decides a `cat`/`tee` heredoc body is DATA unless a **directly downstream pipeline segment** is one of `SHELL_INTERPRETERS`. A relay command that is not itself an interpreter but hands the body to one defeats that walk. Measured through the real `check()` at both `fea0dca` (main) and `e163320` (PR #1155 head) — **ALLOW at both**, so pre-existing, not a regression of #1155:

```
cat <<'DELIM' | xargs -I{} bash -c "{}"
git -c user.name="X" -c user.email="Y" <the-commit-verb> -m z
DELIM
```

`_opener_feeds_interpreter` walks the downstream `|` segments and tests each segment head against `SHELL_INTERPRETERS`. `xargs` is not in that set, so the loop finds no interpreter and returns `is_code=False`; the body is stripped before the matchers run. The `bash -c "{}"` on the opener line survives, but its payload is the literal `{}` — no `git`, no commit verb — so `_DASH_C_RE` does not fire either.

`sh -c` relays via `parallel`, `env -S`, and `find -exec` are the same family.

## Why this is *not* the same as the #1155 must-fix

The process-substitution regression reported on PR #1155 (`tee >(bash) <<'DELIM'`) went BLOCK -> ALLOW and is a genuine weakening. This one was already allowed at `fea0dca`, so #1155 leaves it exactly where it found it. Filing so the gap is tracked rather than assumed closed by the new pipe-follow.

## Suggested direction

The conservative option, consistent with #1155's stated guarantee that every ambiguity resolves toward CODE: treat a downstream segment whose head is neither a known data sink nor a known interpreter as **unresolved**, and return CODE, instead of the current `break`/fall-through to DATA. That inverts the default for unknown relays at the cost of some false positives for genuinely inert filters (`grep`, `sed`, `wc`) — so it likely wants a small known-inert-filter allowlist alongside, mirroring `HEREDOC_DATA_SINKS`.

## Provenance

Found by the merge-gate review of PR #1155 (Aino Virtanen, Standards & Quality Lead), constructing bypass shapes outside the PR's own 49-shape corpus. Belongs to the #1150 umbrella as a further under-matching instance.

parametrization:
[...]

PR #1185, opened 2026-07-30
fix(wave-status): canonical-scope merged-PR discovery for direct-to-main waves

### LINKED ISSUE #1131: tech-debt: wave counter + trust-signal helpers silently return 0/{} for direct-to-main waves
## Problem

`wave_status.py counters` and `trust_signals.py extract` both scope the wave's PR set via `wave_status.merged_prs()`, which is **hardcoded to the wave-branch base** `deployments/phase-<P>/wave-<M>` (wave_status.py:125). For a `direct-to-main` wave (per-story PRs merged straight to `main`, no wave branch — the model in use since the 2026-06-09 every-wave-merges-to-main directive), there is no such branch, so:

- `wave_status.py counters <P> <M>` tallies **0** PRs → `final_pr_count=0`, `top_concentration_pct=0` (division guarded), and the `--expect N` cross-check loud-fails (tally 0 ≠ N).
- `trust_signals.py extract <P> <M>` returns `{}` — no per-engineer signals at all.

Both fail **silently green** in the sense that they produce plausible empty/zero output rather than erroring, so a wrapup/retro that trusts them records a wave as "0 PRs, no trust signals."

## Impact

Surfaced during the **wave-28 wrapup** (`direct-to-main`, 12 PRs merged to main across 5 repos). The counters (`final_pr_count=12`, `changes_requested_cycles=3`, `top_concentration_pct=17`) and the 11-engineer `wave_28_trust_signals` block had to be computed **by hand** (counters) and via a **monkeypatched `merged_prs`** (trust signals — feeding the canonical PR set into the real, tested extraction logic). Reproducible but not automated; a future direct-to-main wave will hit the same manual detour, and `/wave-retro` Step 2.5/Step 4 would read zeros/empty if it re-extracts.

## Proposed fix

Make `merged_prs()` **merge-model-aware**: when `wave_<M>_merge_model == "direct-to-main"`, list PRs `--base main` filtered by `mergedAt >= wave_<M>_kicked_off_at` (the existing #423 cross-window filter). This over-counts non-wave PRs merged in the same window (wave-28 hit exactly this — us#213, an out-of-scope refresh-cookie hotfix, merged mid-window), so it also needs an **explicit canonical-PR-set scoping input** — e.g. derive the set from the wave scope's linked issues, or accept an explicit PR-number allowlist persisted at kickoff. The scoping-ambiguity is the real design work; the base switch alone is insufficient.

## Acceptance

- `wave_status.py counters` and `trust_signals.py extract` produce correct, non-empty results for a `direct-to-main` wave without a wave branch.
- The us#213-class false-positive (in-window but out-of-scope PR) is excluded.
- Regression test covering a direct-to-main wave fixture.

Found during: wave-28 wrapup (Phase 10). Refs: wave_status.py:115-168, trust_signals.py:245-285.

parametrization:
## Summary

`.claude/lib/wave_status.py`'s `merged_prs()` hardcoded a
`deployments/phase-{P}/wave-{M}` base. On a `direct-to-main` wave (the
common case since the 2026-06-09 every-wave-merges-to-main directive) no
such branch exists, so the call silently returned an empty set —
`compute_counters()` reported `final_pr_count=0` and
`trust_signals.extract_signals()` returned `{}`. Neither errored. Since
wave-29 is itself declared `direct-to-main`, its own `/wave-wrapup`
depended on this fix.

Base+timestamp alone is **not sufficient** — the wave-28 retro recorded a
concrete false positive (`us#213`, in-window but out-of-scope). The fix
instead derives a **canonical scope-issue set** per repo from
`wave_{M}_scope`'s `tier_*` rows (iterated generically via
`key.startswith("tier_")` — tier names are not fixed across waves; wave-29
introduced `tier_4_in_wave_findings`, absent from earlier waves), lists
merged-to-main PRs per in-scope repo (pre-filtered by
`wave_{M}_kicked_off_at`), and keeps only the PRs whose
GitHub-recognised closing-issue references (`gh api graphql`
`closingIssuesReferences` — not available on this org's pinned gh 2.45.0
REST-backed `--json` surface) intersect that canonical set. Scope rows
are not assumed 1:1 with PRs: `main#1172` was delivered by PR `#1173`
(a different number), and `#1167`/`#1168`/`#1170`/`#1171` bundle into a
single PR.

The pre-existing wave-branch path is renamed to `_merged_prs_wave_branch`
but is otherwise byte-for-byte unchanged; `merged_prs()` now dispatches on
`wave_merge_model.read_merge_model()` — `direct-to-main` routes to the new
canonical-scope path, everything else (`wave-branch`, or an unrecorded
legacy model) keeps the old behavior. `trust_signals.py` needed no changes
— it consumes `wave_status.merged_prs()` and inherits the fix.

## Verification

**Real wave-29 data, before/after:**

| | `final_pr_count` | `changes_requested_cycles` | `top_concentration_pct` | trust signals |
|---|---|---|---|---|
| Before (unfixed, base `b290d61`) | `0` | `0` | `0` | `{}` |
| After (this PR) | `5` | `3` | `20` | 6 engineers, real signals |

`merged-prs` on the fixed code finds exactly `#1173, #1153, #1154,
#1155, #1156` — the five PRs the issue names as ground truth — and
correctly excludes the still-open `#1178`.

**Wave-branch path unchanged:** `BaseVsHeadDifferential` in
`test_wave_status.py` execs the pre-fix `merged_prs` straight out of the
base commit via `git show <sha>:.claude/lib/wave_status.py` and diffs its
JSON output against head's `_merged_prs_wave_branch` (and against the
public `merged_prs()` dispatcher on a status file with no
`merge_model` key) on the same fixture/fake-gh — an actual runtime
comparison, not a want/got table.

**Tests fail against unfixed code:** stashed the source change,
re-ran the new tests — all 9 new direct-to-main / differential tests
failed (the pre-existing `WaveBranchPathUnchanged` test passed
unchanged, as expected); restored the fix and all 27 tests in the file
pass again.

**Regression coverage, both merge models:** `CanonicalIssueNumbers`,
`MergedPrsDirectToMain` (over-count guard / bundled-PR dedup /
issue-number != PR-number / no-canonical-rows skip / real wave-29 5-PR
set), `WaveBranchPathUnchanged` (explicit `wave-branch` model keeps the
old base), `BaseVsHeadDifferential`.

**Full suite:** `.claude/lib/tests` + `.claude/hooks/tests` — 3171
passed, 253 subtests. `ruff format`/`ruff lint` clean. `mypy`: no
issues. Sync-drift gate: OK. Pre-commit and pre-push hook sets both ran
clean, no `--no-verify`.

Closes #1131.

## Test plan
- [x] `ENVIRONMENT=test python3 -m pytest .claude/lib/tests/test_wave_status.py -q` — 27 passed
- [x] `ENVIRONMENT=test python3 -m pytest .claude/lib/tests .claude/hooks/tests -q` — 3171 passed, 253 subtests
- [x] `python3 -m mypy .claude/lib/wave_status.py .claude/lib/tests/test_wave_status.py` — no issues
- [x] `python3 .claude/lib/wave_status.py counters 10 29 --status cross-repo-status.json` against real data — non-zero, matches ground truth
- [x] Stash-and-rerun: new tests fail against unfixed `wave_status.py`

### DIFF
--- a/.claude/lib/tests/test_wave_status.py
+++ b/.claude/lib/tests/test_wave_status.py
@@ -367,5 +367,449 @@ class Digest(unittest.TestCase):
             self.assertEqual(parsed["current_wave"], "wave-25")
 
 
+# --------------------------------------------------------------------------- #
+# direct-to-main (main#1131): merged_prs()/compute_counters() silently
+# returned 0/{} on a direct-to-main wave because the base was hardcoded to a
+# wave branch that never exists under that model. The fix routes through the
+# wave's canonical scope set (wave_{M}_scope's tier_* rows) rather than
+# base+timestamp alone, per the wave-28 retro finding that base+timestamp
+# over-counts (the recorded false positive: `us#213` was in-window but
+# out-of-scope).
+# --------------------------------------------------------------------------- #
+
+
+def _kv_flags(cmd: list[str]) -> dict[str, str]:
+    """Extract every ``-f``/``-F key=value`` pair from a gh argv list."""
+    out: dict[str, str] = {}
+    i = 0
+    while i < len(cmd):
+        if cmd[i] in ("-f", "-F") and i + 1 < len(cmd):
+            k, _, v = cmd[i + 1].partition("=")
+            out[k] = v
+        i += 1
+    return out
+
+
+class _FakeGhDirectToMain:
+    """subprocess.run side_effect for the direct-to-main path.
+
+    Each PR fixture dict carries: repo, number, sha, mergedAt, login,
+    commit_author, closes (the issue numbers GitHub records this PR as
+    closing — NOT assumed to equal the PR number, mirroring main#1172 being
+    delivered by PR #1173).
+    """
+
+    def __init__(self, prs: list[dict]) -> None:
+        self.prs = prs
+        self.calls: list[list[str]] = []
+
+    def __call__(self, cmd, *args, **kwargs):  # noqa: ANN001
+        assert isinstance(cmd, list), f"gh called with non-list cmd: {cmd!r}"
+        assert cmd[0] == "gh"
+        assert kwargs.get("shell") is not True
+        self.calls.append(cmd)
+
+        if cmd[1:3] == ["pr", "list"]:
+            repo = cmd[cmd.index("--repo") + 1].removeprefix("noorinalabs/")
+            # direct-to-main must query base=main -- never a wave branch.
+            self_base = cmd[cmd.index("--base") + 1]
+            assert self_base == "main", f"expected --base main, got {self_base!r}"
+            listed = [
+                {
+                    "number": p["number"],
+                    "headRefOid": p["sha"],
+                    "mergedAt": p["mergedAt"],
+                    "author": {"login": p["login"]},
+                }
+                for p in self.prs
+                if p["repo"] == repo
+            ]
+            return SimpleNamespace(stdout=json.dumps(listed), returncode=0, stderr="")
+
+        if cmd[1:3] == ["api", "graphql"]:
+            flags = _kv_flags(cmd)
+            repo = flags["name"]
+            number = int(flags["number"])
+            closes = next(
+                p["closes"] for p in self.prs if p["repo"] == repo and p["number"] == number
+            )
+            return SimpleNamespace(stdout=json.dumps(closes), returncode=0, stderr="")
+
+        if cmd[1] == "api":
+            path = cmd[2]
+            parts = path.split("/")
+            if "/commits/" in path:
+                sha = parts[4]
+                name = next(p["commit_author"] for p in self.prs if p["sha"] == sha)
+                return SimpleNamespace(stdout=name + "\n", returncode=0, stderr="")
+            if path.endswith("/comments"):
+                number = int(parts[4])
+                cr = next((p.get("cr", 0) for p in self.prs if p["number"] == number), 0)
+                return SimpleNamespace(stdout=f"{cr}\n", returncode=0, stderr="")
+
+        raise AssertionError(f"unexpected gh call: {cmd!r}")
+
+
+def _scope_row(issue_number: int) -> dict:
+    return {"id": f"noorinalabs-main#{issue_number}", "ref": f"main#{issue_number}"}
+
+
+def _write_direct_to_main_status(
+    path: Path,
+    *,
+    wave: str,
+    repos: list[str],
+    kickoff: str,
+    tiers: dict[str, list[int]],
+) -> None:
+    """Write a cross-repo-status.json shaped like wave-29's real record:
+    `wave_{M}_merge_model="direct-to-main"` + a `wave_{M}_scope` dict whose
+    tier keys are NOT the fixed set from any prior wave (arbitrary tier names
+    below prove the `tier_*` iteration is generic, per main#1131)."""
+    scope: dict = {"theme": "test fixture", "merge_model": "direct-to-main"}
+    for tier_name, issue_numbers in tiers.items():
+        scope[tier_name] = [_scope_row(n) for n in issue_numbers]
+    data = {
+        "current_wave": int(wave),
+        f"wave_{wave}_repos_in_scope": repos,
+        f"wave_{wave}_kicked_off_at": kickoff,
+        f"wave_{wave}_merge_model": "direct-to-main",
+        f"wave_{wave}_scope": scope,
+    }
+    path.write_text(json.dumps(data, indent=2) + "\n", encoding="utf-8")
+
+
+class CanonicalIssueNumbers(unittest.TestCase):
+    def test_generic_tier_iteration_not_hardcoded(self) -> None:
+        """Tier names are wave-specific (main#1131: wave-29 introduced
+        `tier_4_in_wave_findings`, absent from earlier waves) -- the helper
+        must not hardcode any tier name."""
+        with TemporaryDirectory() as td:
+            status = Path(td) / "cross-repo-status.json"
+            _write_direct_to_main_status(
+                status,
+                wave="29",
+                repos=["noorinalabs-main"],
+                kickoff="2026-07-27T22:56:17Z",
+                tiers={
+                    "tier_1_track_g": [1114],
+                    "tier_4_in_wave_findings": [1160, 1162],
+                    "tier_99_never_seen_before": [9999],
+                },
+            )
+            by_repo = wave_status._canonical_issue_numbers_by_repo("29", status)
+        self.assertEqual(by_repo["noorinalabs-main"], {1114, 1160, 1162, 9999})
+
+    def test_legacy_string_tier_rows_skipped_not_raised(self) -> None:
+        with TemporaryDirectory() as td:
+            status = Path(td) / "cross-repo-status.json"
+            data = {
+                "wave_29_scope": {
+                    "tier_1_backlog": ["main#322", {"id": "noorinalabs-main#1114"}],
+                }
+            }
+            status.write_text(json.dumps(data))
+            by_repo = wave_status._canonical_issue_numbers_by_repo("29", status)
+        self.assertEqual(by_repo, {"noorinalabs-main": {1114}})
+
+
+class MergedPrsDirectToMain(unittest.TestCase):
+    """merged_prs() for a direct-to-main wave, driven by the canonical scope
+    set rather than base+timestamp alone."""
+
+    def test_base_and_timestamp_alone_would_overcount_but_scope_excludes(self) -> None:
+        """The wave-28 retro false positive, reproduced: a PR in-window on
+        `main` that closes an issue NEVER recorded in the wave's scope
+        (`us#213`-shaped) must be excluded even though base+timestamp alone
+        would have included it."""
+        prs = [
+            {
+                "repo": "noorinalabs-main",
+                "number": 1173,
+                "sha": "sha1173",
+                "mergedAt": "2026-07-30T02:16:40Z",
+                "login": "octocat",
+                "commit_author": "Nino Kavtaradze",
+                "closes": [1172],  # canonical scope row
+            },
+            {
+                "repo": "noorinalabs-main",
+                "number": 9001,
+                "sha": "sha9001",
+                "mergedAt": "2026-07-30T02:20:00Z",  # in-window
+                "login": "octocat",
+                "commit_author": "Someone Else",
+                "closes": [213],  # NOT in scope -- the us#213 false positive
+            },
+        ]
+        fake = _FakeGhDirectToMain(prs)
+        with TemporaryDirectory() as td:
+            status = Path(td) / "cross-repo-status.json"
+            _write_direct_to_main_status(
+                status,
+                wave="29",
+                repos=["noorinalabs-main"],
+                kickoff="2026-07-27T22:56:17Z",
+                tiers={"tier_4_in_wave_findings": [1172]},
+            )
+            with mock.patch.object(wave_status.subprocess, "run", fake):
+                got = wave_status.merged_prs("10", "29", status)
+        self.assertEqual([p["number"] for p in got], [1173])
+
+    def test_issue_number_need_not_equal_pr_number(self) -> None:
+        """main#1172 was delivered by PR #1173 -- a different number. The PR
+        must still be found via its closing-issue reference."""
+        prs = [
+            {
+                "repo": "noorinalabs-main",
+                "number": 1173,
+                "sha": "sha1173",
+                "mergedAt": "2026-07-30T02:16:40Z",
+                "login": "octocat",
+                "commit_author": "Nino Kavtaradze",
+                "closes": [1172],
+            }
+        ]
+        fake = _FakeGhDirectToMain(prs)
+        with TemporaryDirectory() as td:
+            status = Path(td) / "cross-repo-status.json"
+            _write_direct_to_main_status(
+                status,
+                wave="29",
+                repos=["noorinalabs-main"],
+                kickoff="2026-07-27T22:56:17Z",
+                tiers={"tier_4_in_wave_findings": [1172]},
+            )
+            with mock.patch.object(wave_status.subprocess, "run", fake):
+                got = wave_status.merged_prs("10", "29", status)
+        self.assertEqual(len(got), 1)
+        self.assertEqual(got[0]["number"], 1173)
+        self.assertEqual(got[0]["commit_author_name"], "Nino Kavtaradze")
+
+    def test_bundled_pr_counted_once_across_multiple_scope_rows(self) -> None:
+        """#1167/#1168/#1170/#1171 were bundled into ONE PR -- rows != PRs,
+        so the same merged PR must not be double-counted across the four
+        scope rows it satisfies."""
+        prs = [
+            {
+                "repo": "noorinalabs-main",
+                "number": 1174,
+                "sha": "sha1174",
+                "mergedAt": "2026-07-29T12:00:00Z",
+                "login": "octocat",
+                "commit_author": "Weronika Zielinska",
+                "closes": [1167, 1168, 1170, 1171],
+            }
+        ]
+        fake = _FakeGhDirectToMain(prs)
+        with TemporaryDirectory() as td:
+            status = Path(td) / "cross-repo-status.json"
+            _write_direct_to_main_status(
+                status,
+                wave="29",
+                repos=["noorinalabs-main"],
+                kickoff="2026-07-27T22:56:17Z",
+                tiers={"tier_4_in_wave_findings": [1167, 1168, 1170, 1171]},
+            )
+            with mock.patch.object(wave_status.subprocess, "run", fake):
+                got = wave_status.merged_prs("10", "29", status)
+        self.assertEqual(len(got), 1)
+        self.assertEqual(got[0]["number"], 1174)
+
+    def test_real_wave_29_five_pr_set(self) -> None:
+        """Verification fixture pinned to the real wave-29 data in main#1131's
+        acceptance criteria: PRs #1173/#1153/#1154/#1155/#1156, closing
+        canonical issues #1172/#1139/#1134/#1152/#1151 respectively, with an
+        out-of-scope in-window PR (#1178, still OPEN in reality -- modeled
+        here as an unrelated closed number to prove exclusion) filtered out."""
+        real = [
+            (1173, "sha056a", "2026-07-30T02:16:40Z", 1172),
+            (1153, "shab708", "2026-07-30T02:17:04Z", 1139),
+            (1154, "shafbb5", "2026-07-30T02:17:26Z", 1134),
+            (1155, "sha1207", "2026-07-30T02:17:46Z", 1152),
+            (1156, "sha999d", "2026-07-30T02:40:16Z", 1151),
+        ]
+        prs = [
+            {
+                "repo": "noorinalabs-main",
+                "number": number,
+                "sha": sha,
+                "mergedAt": merged_at,
+                "login": "octocat",
+                "commit_author": f"author-{number}",
+                "closes": [closes],
+            }
+            for number, sha, merged_at, closes in real
+        ] + [
+            {
+                "repo": "noorinalabs-main",
+                "number": 9999,
+                "sha": "shaoutofscope",
+                "mergedAt": "2026-07-30T02:41:00Z",
+                "login": "octocat",
+                "commit_author": "Out Of Scope",
+                "closes": [4242],  # not a canonical row anywhere
+            }
+        ]
+        fake = _FakeGhDirectToMain(prs)
+        with TemporaryDirectory() as td:
+            status = Path(td) / "cross-repo-status.json"
+            _write_direct_to_main_status(
+                status,
+                wave="29",
+                repos=["noorinalabs-main"],
+                kickoff="2026-07-27T22:56:17Z",
+                tiers={
+                    "tier_2_process_carry_forward": [1139, 1134, 1152, 1151],
+                    "tier_4_in_wave_findings": [1172],
+                },
+            )
+            with mock.patch.object(wave_status.subprocess, "run", fake):
+                got = wave_status.merged_prs("10", "29", status)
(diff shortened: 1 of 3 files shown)


parametrization:
**Review: Changes requested — one CI-blocking bug; every empirical claim I could independently verify checked out**

I reviewed independently of the merge-gate slot, focused on verification per the review brief. Ran everything live against real GitHub data (a dedicated worktree at `.claude/worktrees/bereket-1185-review`, detached at head `39eaa8e`, removed after review) rather than trusting the PR-body numbers.

**Must-fix**

1. **CI is currently red at this exact head SHA (`39eaa8e`) — "Pytest — hooks and lib test suites" fails.** `BaseVsHeadDifferential::test_base_and_head_agree_on_wave_branch_output` crashes:
   ```
   subprocess.CalledProcessError: Command ['git', 'show', 'b290d611e9e6c59db0bf921ef6a9315090dc2eae:.claude/lib/wave_status.py'] returned non-zero exit status 128.
   1 failed, 3170 passed, 253 subtests passed in 11.61s
   ```
   Root cause: that CI job's `actions/checkout@v4` step uses the default shallow clone (no `fetch-depth: 0`), so the pre-#1131 base commit isn't present in CI's object store — `git show <old-sha>:path` fails outside a full-history checkout. It passes locally (and presumably in the author's own clone) only because a normal dev checkout has full history. This is a real, reproducible failure, not a flake — it's the only run recorded at this SHA (`gh run list` shows one `CI — Hooks & Scripts` run, `conclusion: failure`), and `run.head_sha` matches the PR head exactly.

   This directly contradicts the PR body's "Full suite: 3171 passed" / "no `--no-verify`" claim — the actual CI number is 1 failed, 3170 passed. Per the org's full-parity policy, a red gate is a stop, not a speed bump, and this must not merge without an explicit owner override. The fix this file already knows about: the `Doc-freshness gate (advisory)` job in the same `ci.yml` sets `fetch-depth: 0` with the comment "the default shallow checkout has no merge-base" — the pytest job needs the same, or the test needs to stop depending on live git history at all (e.g. bake the pre-fix source as a fixture rather than `git show`-ing a specific SHA against whatever depth CI happens to have).

**Verified correct (no action needed) — reporting since these were the brief's main risk items**

2. **Merged-PR set matches exactly.** Ran `wave_status.merged_prs("10", "29", ...)` live against the real `cross-repo-status.json`: returns exactly `#1173, #1153, #1154, #1155, #1156` in `noorinalabs-main`, `#1178` correctly absent. `compute_counters` gives `{final_pr_count: 5, changes_requested_cycles: 3, top_concentration_pct: 20}` — matches the PR body exactly.

3. **`top_concentration_pct`/trust-signals source — I initially suspected this was wrong and it isn't.** I first checked `.commit.author.name` on the five *squash-merge* commits on `main` (`056a7afa`, `b708e1d1`, `fbb528e8`, `120740ff`, `999dfae3`) and got `parametrization` for all five — which would collapse concentration to 100% on one bogus identity. But `_commit_author_name` is called with `pr["headRefOid"]` (the PR branch's own tip commit), not the squash-merge SHA on `main`. Checking the actual `headRefOid` for each PR gives five distinct real per-commit `-c author` identities (Aino Virtanen, Nurul Hakim, Nadia Khoury, Weronika Zielinska, Lucas Ferreira) — hence the genuine 20%. Live-ran `trust_signals.extract_signals("10", "29", ...)` too: 6 engineers, matches the PR body. So the design correctly sidesteps the #1177 squash-authorship-loss problem for this metric by reading the pre-squash branch tip rather than the main-branch merge commit — worth a one-line code comment on `_commit_author_name` calling this out explicitly, since it's easy for a future reader (I did) to assume it reads the main-branch commit and conclude it's broken.

4. **New tests are not inert.** Reverted `.claude/lib/wave_status.py` to `b290d611` in my worktree, kept the PR's test file: `9 failed, 18 passed` — the exact 9 new direct-to-main/differential tests fail, `WaveBranchPathUnchanged` passes unchanged, matching the PR body's claim precisely. Restored the fix: `27 passed`.

5. **`BaseVsHeadDifferential` is a genuine runtime comparison,** not a want/got table — confirmed by reading the test body (`git show <base-sha>:path` execed into a fresh module, run against the same `_FakeGh` fixture as head's `_merged_prs_wave_branch`, `assertEqual` on the two JSON outputs). The mechanism is sound; it's just fragile against a shallow CI checkout (item 1).

Everything else — `_canonical_issue_numbers_by_repo`'s generic `tier_*` iteration, the `closingIssuesReferences` GraphQL scoping, the wave-branch path being a pure rename — matches the diff as described.

Tech-debt: none filed — item 1 is in-scope for this PR (the failing test was added by this PR), not deferred debt.

Requestor: Bereket Tadesse
Requestee: Santiago Ferreira
TechDebt: none

parametrization:
## Merge-gate review — PR #1185 @ `39eaa8e`

Reviewed against `main` at the head SHA via `gh api contents?ref=…` (no local checkout for the read path). Mutation-tested the new guards in a worktree at `39eaa8e`; executed the built GraphQL query and the built `gh pr list` argv against the real engine.

The architecture is right. Scoping by the canonical `tier_*` rows instead of base+timestamp is the correct answer to the wave-28 `us#213` over-count, the dispatch default is safe, and the error handling on the GraphQL dependency is genuinely loud (verified below, contrary to my going-in concern). Two things block: CI is red at the head SHA, and the new path re-introduces a **silent under-count** that this PR exists to eliminate.

---

### Must-fix

**1. CI is red at `39eaa8e` — `Pytest — hooks and lib test suites` FAILED.**

```
FAILED .claude/lib/tests/test_wave_status.py::BaseVsHeadDifferential::test_base_and_head_agree_on_wave_branch_output
  subprocess.CalledProcessError: Command '['git','show','b290d611…:.claude/lib/wave_status.py']' returned non-zero exit status 128
1 failed, 3170 passed, 253 subtests passed
```

Root cause: the pytest job uses `actions/checkout@v4` at **default `fetch-depth: 1`** (only one job in `ci.yml` sets `fetch-depth: 0`), so `b290d611` is not in the shallow clone. It passes locally on a full clone — I reproduced 27/27 green in a worktree — which is why the PR body's "3171 passed" reads true. It is not true in CI.

Worth naming the second-order effect: the test that cannot run in CI *is* the proof artifact for the "wave-branch path unchanged" claim. In CI that guarantee is currently unverified, not merely untested.

Prefer `fetch-depth: 0` on that job over `skipUnless(git cat-file -e …)` — a differential test that silently skips is a weaker guarantee than one that runs.

Per `ci-gates.md`, a red gate is a stop regardless of cause.

**2. `gh pr list` omits `--limit`; the default cap is 30 — silent under-count.**

`_merged_prs_direct_to_main` builds this argv (no `--limit`). Run as built against the real repo:

```
as-built (no --limit): 30   |  true population: 251
as-built oldest mergedAt: 2026-07-19T21:00:14Z
```

Under the old wave-branch path the cap was harmless — `--base deployments/phase-{P}/wave-{M}` bounded the result set to the wave itself. Switching to `--base main` removes that bound and re-exposes the cap against the repo's entire 251-PR merge history.

Wave-29 has 5 post-kickoff merges today, so there is headroom — which is precisely why your live run returned the right answer. But wave-29 now carries **28 scope rows across four tiers**, plus review-spawned issues (#1179–#1182, #1186, #1187, and now #1189/#1190). Once the number of PRs merged to `main` after the *earliest* wave PR exceeds 30, the earliest wave PRs drop out of the window. No error, no warning — `final_pr_count` just comes back low, and that plausible-but-wrong number is what lands in the trust matrix. This is the same failure shape as the bug being fixed, and wave-28's retro already recorded two figures that turned out to be false.

**The guard is inert — mutation-tested.** Injecting the real cap into the head source:

| mutation | result |
|---|---|
| `listed = listed[:30]` (the actual gh default) | **27/27 pass** |
| `listed = listed[:1]` (artificial) | 1 fails — only because one fixture happens to hold 6 PRs |

No fixture has more than 6 PRs, so no fixture can trip the real cap. Fix: pass an explicit `--limit` (e.g. 500), and either add a >cap fixture or assert `--limit` is present in the argv the way `_FakeGhDirectToMain` already asserts `--base main`.

Context: the org recognises this failure class (`GraphQL pagination over-cap lint`, #888/#892/#893), but that lint scans **skill markdown** for `first: > 100` only — it cannot see `.py`, and it cannot see `gh pr list`. Its green tick here is not evidence of coverage.

---

### Verified good (probed, not accepted)

- **GraphQL failure is loud, not silent.** My going-in concern was a rate-limited call read as "zero results." It is handled. A GraphQL `errors` payload arrives as HTTP 200 with `data.repository.pullRequest: null`, and `gh api graphql` still exits **1** — measured unpiped (piped through `head` it reports `0`, which is the trap):
  ```
  EXIT=1
  stderr: gh: Could not resolve to a PullRequest with the number of 999999.
  ```
  `_run_gh`'s `check=True` turns that into `CalledProcessError`. Rate-limiting has the same HTTP-200-with-`errors` shape, so it raises too. Correct call.
- **Dispatch default is safe.** `read_merge_model() -> str | None`; anything that is not `direct-to-main` — including `None` for a legacy/unrecorded status file — routes to `_merged_prs_wave_branch`. Mutation M2 (force every model onto the new path) trips 9 tests, so the guard is live, not decorative.
- **Scope filter is live.** Mutation M1 (delete the `closes & canonical_issues` intersection) trips 2 tests including the `us#213` over-count fixture.
- **Generic `tier_*` iteration really picks up `tier_4`.** Checked against the *amended* real record, not the fixture: 28 rows over `tier_1_track_g` (9), `tier_2_process_carry_forward` (10), `tier_3_doc_drift` (1), `tier_4_in_wave_findings` (8), and the rows carry the fully-qualified `id` shape (`noorinalabs-main#1114`) the `rpartition("#")` parse expects.
- **Rows ≠ PRs, both directions.** Bundling is structurally safe — the loop iterates PRs, not rows, so #1167/#1168/#1170/#1171 collapse to one PR counted once. Issue# ≠ PR# verified live: #1173 closes #1172.
- **Ground truth independently reproduced.** The five PRs close exactly `main#1172 / #1139 / #1134 / #1152 / #1151`, one each, all in-repo — matching your table without relying on it.
- **Revert is clean.** Two files, additive plus a rename; no state, no migration.

---

- **#1189** — `closingIssuesReferences` returns bare `number`s with no repo; the intersection against a *per-repo* canonical set can under-count or mis-attribute a cross-repo `Closes`. `repository{name}` is available at no cost (confirmed live). Not reachable today (main's numbers are 1114–1180, ingest-platform's only row is #152), and the fixture model stores `closes` as bare ints so it cannot currently express the case.
- **#1190** — a wave PR lacking a `Closes #N` reference is silently dropped, making the counter dependent on PR-description hygiene. Suggest a reconciliation *warning* listing scope rows no merged PR claims, so the instrument says what it could not see.

### Nit

`_merged_prs_wave_branch`'s docstring ("every line of logic below is identical to the pre-#1131 body") and the PR body ("byte-for-byte unchanged") overstate slightly — the inline commit-author `_run_gh` call was extracted into `_commit_author_name()`. Behaviourally equivalent, and `BaseVsHeadDifferential` proves it at runtime; "behaviourally identical" would be the accurate wording.

---

Fix the two must-fixes and this is a straightforward approve — the hard part, the scoping rule, is correct.

Requestor: Weronika Zielinska
Requestee: Santiago Ferreira
TechDebt: #1189, #1190

parametrization:
## Re-requesting review — both must-fixes addressed, new head `9367391`

Head moved (`39eaa8e` → `9367391`), so this drops your prior ChangesRequested per the head-move rule — re-requesting fresh, not asking you to carry it over.

**M1 (CI red / body overclaim):** added `fetch-depth: 0` to the `pytest` job in `.github/workflows/ci.yml`, mirroring the existing `doc-freshness` job's precedent for the identical reason. CI-sourced result on this PR's job log at `9367391`: `3174 passed, 253 subtests passed` (the 3 extra tests vs. your `3170` are the new M2 coverage below, not a discrepancy).

**Non-blocking, folded in:** added the one-line `_commit_author_name` comment you asked for, spelling out that it reads `headRefOid` (pre-squash branch tip) and never the `--merge` merge-commit SHA, and why (`parametrization`-collapse / #1177).

Re-verified against real wave-29 data with the final diff: `{"final_pr_count": 5, "changes_requested_cycles": 3, "top_concentration_pct": 20}` — unchanged from your measurement.

Diff since `39eaa8e`: `.github/workflows/ci.yml` (+`fetch-depth: 0` on the pytest job) and the `_commit_author_name` docstring addition, plus Weronika's M2 fix (see her thread) — no change to the scoping logic you verified.

---

Requestor: Santiago Ferreira
Requestee: Bereket Tadesse
RequestOrReplied: Request

parametrization:
## Re-requesting review — both must-fixes addressed, new head `9367391`

Head moved (`39eaa8e` → `9367391`), so this drops your prior ChangesRequested per the head-move rule — re-requesting fresh, not asking you to carry it over.

**M1 (CI red at `39eaa8e`):** added `fetch-depth: 0` to the `pytest` job in `.github/workflows/ci.yml`, the same precedent already in that file on the `doc-freshness` job. CI-sourced result on this PR's job log at `9367391`: `3174 passed, 253 subtests passed`.

**M2 (`gh pr list` default-30 cap under `--base main`):** added a server-side `--search merged:>=<kickoff>` qualifier plus an explicit `--limit 1000` to the `gh pr list` call in `_merged_prs_direct_to_main`. Per your mutation critique, made the guard actually testable instead of fixture-fragile: `_FakeGhDirectToMain` now enforces the real gh `--limit`/`--search` semantics (cap truncation + server-side date filter, newest-`mergedAt`-first sort) instead of returning every fixture PR unconditionally, and the new `PrListPageCap` tests build a 40-PR repo history (35 more-recent filler merges crowding 5 real wave-29 PRs) to prove the fix. Verified: reverting the `--limit`/`--search` args back out fails all 3 new tests with real diffs (`[]` vs the expected 5 numbers, and the two argv-shape assertions); restoring the fix turns them green.

Also filed the tech-debt you asked for — confirmed **#1189** and **#1190** are already filed and out of scope for this PR, left untouched as you specified.

Re-verified against real wave-29 data with the final diff: `{"final_pr_count": 5, "changes_requested_cycles": 3, "top_concentration_pct": 20}` — unchanged from your measurement.

---

Requestor: Santiago Ferreira
Requestee: Weronika Zielinska
RequestOrReplied: Request

parametrization:
**Re-review at new head `9367391` (was `39eaa8e`) — Approved**

Independent re-verification, anchored to the current head. Per the caution about the rebase, I did not use ancestry/two-dot-diff to compare against my prior pass — used `git patch-id --stable` instead: `git show 39eaa8e | git patch-id` and `git show a9ff88e | git patch-id` (the rebased carry-forward of the original fix commit) both hash to `ed3549fbf5df8e8913aebbfae5dc1c60262834ea` — byte-identical content survived the rebase. The new top commit `9367391` is the only actual delta since my last pass.

**M1 (mine) — verified fixed.** `.github/workflows/ci.yml`'s `pytest` job now has `fetch-depth: 0` with an inline comment naming the exact failure (`git show <base-sha>` against a shallow clone, exit 128) and citing the doc-freshness job precedent. Confirmed live: `gh api .../actions/runs/30512181174 --jq .head_sha` matches the PR head exactly, job conclusion SUCCESS, log shows `3174 passed, 253 subtests passed in 11.88s` — sourced from the actual CI log, not a local run, which was the specific overclaim I'd caught. All 21 checks green.

**M2 (Weronika's) — new surface, independently verified, including a mutation test of my own.** In a scratch worktree at `9367391`: `pytest .claude/lib/tests/test_wave_status.py` → 30 passed (up from 27, matching the 3 new `PrListPageCap` tests). I then mutated `_merged_prs_direct_to_main` myself — stripped the `--limit`/`--search` args entirely — and reran: `3 failed, 27 passed`, with `test_all_five_wave_prs_survive_a_repo_history_of_40` failing on the real dropped-PR assertion, not just a flag-presence check. Restored the fix, 30 passed again. Also live-tested the `--search merged:>=<kickoff>` qualifier directly against the real `gh pr list` (not just the fixture): `gh pr list --repo noorinalabs/noorinalabs-main --state merged --base main --limit 1000 --search "merged:>=2026-07-27T22:56:17Z"` returns exactly the 5 real wave-29 PRs. `_FakeGhDirectToMain`'s modeling of gh's real limit/search/sort semantics checks out against that live behavior.

Re-ran `wave_status.compute_counters` against the real `cross-repo-status.json` with the updated code: `{final_pr_count: 5, changes_requested_cycles: 3, top_concentration_pct: 20}` — unchanged, as the commit message claims.

**One non-blocking nit on my own suggested comment.** The new `_commit_author_name` docstring states "this org merges with `--merge` (never `--squash`)" — but the five actual wave-29 merge commits it's describing (`056a7afa` etc.) are single-parent (squash) commits, not two-parent merge commits (checked via `gh api .../commits/<sha> --jq .parents`); the codebase's own charter change `2d8bd91` banning squash on any base is dated the same day these PRs merged, so the claim is true going forward but not for the historical commits cited two paragraphs earlier in the same function. Doesn't affect correctness — `headRefOid` is safe regardless of merge strategy, which is the actual invariant — but worth a one-word tweak (e.g. "this org's merge policy forbids `--squash`" without "never," or a note that historical squash commits are exactly why headRefOid matters) so a future reader checking `git log --merges` doesn't hit the same false lead I did on my first pass. Not blocking.

Requestor: Bereket Tadesse
Requestee: Santiago Ferreira
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Merge-gate re-review — PR #1185 @ `93673918`

Both must-fixes are resolved. Re-verified at the new head, not carried over from the previous pass.

### CI — green at the head SHA

21/21 success at `93673918`, pulled from the check-runs API at that SHA (not `gh pr checks`). `Pytest — hooks and lib test suites` now passes; totals read out of the job log for run `30512181174`: **3174 passed, 253 subtests passed**. `MERGEABLE` / `CLEAN`.

### M1 — resolved

`fetch-depth: 0` on the pytest job, with a comment naming the actual mechanism (`BaseVsHeadDifferential` execs a historical blob via `git show <sha>:path`; the shallow clone lacks the object) and pointing at the doc-freshness job that already sets it for the analogous merge-base reason. Right fix over `skipUnless` — the differential now genuinely runs in CI instead of silently skipping, so the "wave-branch path unchanged" claim is verified there rather than merely asserted.

### M2 — resolved, and the untestability is genuinely fixed

Both bounds verified by executing the built argv against the real engine, not by reading the string:

```
gh pr list --repo noorinalabs/noorinalabs-main --state merged --base main \
  --limit 1000 --json number,headRefOid,mergedAt,author \
  --search "merged:>=2026-07-27T22:56:17Z"
EXIT=0 → 5 PRs: [1153, 1154, 1155, 1156, 1173]
```

Exactly the ground-truth set. Two engine behaviours the fix leans on, both confirmed live rather than assumed:

- **`--base` is still honoured under `--search`** — substituting a bogus base returns `[]`, so the search path does not quietly discard the base filter.
- **No search-index gap.** I was concerned the move from the list API to the search API could introduce indexing lag that silently drops a just-merged PR at wrapup. Differentialled the two APIs over a 60-day / 157-PR window: **identical sets, zero in either direction.** No evidence of lag at this volume.

**The fixture redesign does what it claims — I killed the mutants myself.** My original objection was that injecting the real cap left 27/27 green because no fixture held more than 6 PRs. That is now closed:

| mutation | result |
|---|---|
| drop `--limit` entirely | **2 fail**, incl. `test_all_five_wave_prs_survive_a_repo_history_of_40` |
| `--limit 30` (gh's own default) | **2 fail**, same pair |
| drop `--search` | 1 fail — the argv assertion only |

The first is the one that matters: `test_all_five…` is a **behavioural** failure with a real diff, not an argv assertion. Modelling gh's actual semantics in `_FakeGhDirectToMain` — server-side `--search` filter first, then newest-`mergedAt`-first sort, then `--limit` truncation defaulting to 30 — is what makes it bite, and the 35 filler merges are correctly dated *after* the five real PRs so they genuinely crowd the top of the window. The guard is sensitive to the limit's **value**, not just its presence, which is the stronger property.

On `--search` being caught only by an argv assertion: that is the correct level, not a gap. The client-side `merged_at < kickoff` filter is deliberately retained, so dropping `--search` cannot yield a *wrong* answer — only a larger query. A bounding optimisation with no correctness surface has nothing behavioural to assert against. Docstring says as much.

### Rebase — content survived, verified by patch-id

Per the caution about ancestry checks and two-dot diffs after a rebase, I used `git patch-id --stable`:

```
old head 39eaa8e  → ed3549fbf5df8e8913aebbfae5dc1c60262834ea
new  a9ff88e      → ed3549fbf5df8e8913aebbfae5dc1c60262834ea   ← identical
new  9367391      → 611b820dd2e5c0f174179250aa9a751d519c34ba   (new work)
```

The originally-reviewed commit is byte-identical in content post-rebase; `9367391` sits purely on top. Nothing was dragged in or lost. I also confirmed the additive commit touches neither `_merged_prs_wave_branch` nor the `deployments/phase-…` base string, so the unchanged-path guarantee still holds at this head. Noting the `reviews.md:114` additive-commits question is being handled separately — not blocking on it, and the content proof above is what I would have needed anyway.

### Also re-checked

- **Live wave-29 data, run against the org main checkout's status file:** `{final_pr_count: 5, changes_requested_cycles: 3, top_concentration_pct: 20}` — reproduced independently, unchanged post-fix.
- **Bereket's `_commit_author_name` note** is folded in accurately: `headRefOid` never the merge commit, because `--merge` (never `--squash`) makes every merge commit's author the CLI identity, collapsing concentration to one name — the #1177 shape. Correct, and worth having in the code rather than only in a review thread.
- **#1189 / #1190** remain open and still accurate — `closingIssuesReferences(first:25){nodes{number}}` is still unqualified by repo, correctly left as tech-debt rather than scope-crept into this PR.

Nothing further from me. Good turnaround on the mutation critique specifically — modelling the dependency's real semantics in the fake, rather than raising a number and asserting on the argv, is the version that will still catch this in six months.

Requestor: Weronika Zielinska
Requestee: Santiago Ferreira
RequestOrReplied: Approved
TechDebt: #1189, #1190

Issue #1021, opened 2026-07-19
Memory/log hygiene: gitignore `annunaki/archive`, cap `.bak` retention, compact `cross-repo-status.json`


parametrization:
## Problem
- `.claude/annunaki/archive/*.jsonl` is **git-tracked** (340 KB, grows one file per wave) — botfarm's whole-dir ignore avoids this.
- `.claude/annunaki/` is 6.2 MB on disk (`traces.jsonl.bak` alone 3.67 MB + 13 error baks) — no `.bak` retention cap.
- `cross-repo-status.json` is 208 KB and grows unbounded (all historical waves inline); the digest is a band-aid over an ever-growing file.

## Acceptance criteria
- [ ] Add `.claude/annunaki/archive/` to `.gitignore`; untrack existing archive files.
- [ ] Cap `.bak` retention (keep last N) in the annunaki rotation path.
- [ ] Add `cross-repo-status.json` compaction/archival of completed-wave blocks.

## Token/repo impact
Removes git bloat; bounds the on-disk and on-read footprint.

---
Parent: #1014

parametrization:
PR #1024 delivers the gitignore/untrack + `.bak` retention cap. The third AC — `cross-repo-status.json` compaction — is carved into a follow-up (mutates a 208 KB live status file; warrants its own reviewed change). Will file the follow-up when the compaction helper is designed.

PR #1186, opened 2026-07-30
tech-debt(wave-wrapup): archive + reset the generic-prompt pending ledger per wave

### LINKED ISSUE #1140: tech-debt(wave-wrapup): reset (or archive) the generic-prompt pending ledger per wave — ~251 stale candidates accumulated
## Problem

The generic-prompt tracker's pending ledger is per-machine volatile state that is **never reset**. By the wave-28 checkpoint it held **~251 stale undecided candidates**, accumulated across many waves and sessions.

The contents are mostly noise, not genuine wave artifacts: session `.consulted/*.marker` files plus pre-existing charter files. The wave-28 checkpoint did the right thing by hand — scoping to the 3 hooks actually touched that wave (all three correctly skipped as perf refactors) — but the tracker's own candidate set was unusable as presented.

## Impact

An unbounded, mostly-noise worklist is worse than no worklist: the signal from a genuine new candidate is buried, so the checkpoint degrades into a manual re-derivation every wave. That is the failure mode where a tool's output stops being read.

## Proposed direction

Add a per-wave reset to `/wave-wrapup` (the retro proposed Step 12.5) so the ledger's pending set is scoped to one wave's actual activity, with `generic_prompt_tracker.py` growing whatever support that needs.

Two things to settle during implementation:
- **Reset vs. archive.** Blowing the ledger away each wave loses any genuinely-undecided candidate that legitimately spans waves. Archiving the pending set per wave (the pattern the annunaki log already uses) is likely the better shape.
- **Filter at the source.** ~251 entries dominated by session markers and charter files suggests the candidate *filter* is too permissive, not just that the reset is missing. A reset alone would re-accumulate the same noise every wave — worth fixing the intake predicate at the same time, or the step just moves the problem.

## Provenance

Wave-28 retro pain point 3 and proposed charter change 2 (`.claude/team/feedback_log.md`), carried into wave-29 scope-setting on 2026-07-27.

parametrization:
## Summary

The generic-prompt pending ledger (`.claude/generic_prompt_pending.json`) never had a wave boundary: it accumulates every touched `.claude/` artifact forever, and nothing removes an *undecided* entry short of an explicit genericize/skip decision. Left alone this becomes an unusable worklist that buries genuine signal.

## Design decisions

**Archive, not bare reset.** Loosely modeled on the `.claude/annunaki/archive/wave-{N}.jsonl` pattern (#1021), adapted for a keyed JSON document rather than append-only JSONL — **note the append-only-JSONL torn-write safety property does NOT carry over to this JSON adaptation** (see #1192, filed as tech-debt, not blocking). `generic_prompt_tracker.archive_wave_pending()` writes the **full** pre-reset candidate set to `.claude/generic_prompt_pending_archive/wave-{label}.json` — partitioned into `noise_dropped` and `genuine_reset` — before clearing the live pending ledger, so nothing is silently discarded. **Archive-before-clear is now covered by a dedicated test** (`test_failed_archive_write_does_not_clear_live_ledger`) — a blocked archive write must leave the live ledger untouched; this is the one property the whole data-safety case rests on and it was previously unasserted (Santiago's mutation testing found the sole surviving mutant reorders exactly this).

**Wired into `/wave-wrapup` Step 12.5.0**, run *before* the existing diff-sweep sub-step, so the sweep immediately re-seeds the live ledger with whatever is still part of the *current* wave's actual `.claude/` changes.

**Root cause, corrected during review.** The noise composition claim in this PR was originally wrong. A live 267-entry ledger measured:

| entries | prefix | what it actually is |
|---|---|---|
| 178 | `jobs/<id>/tmp/*` | one-time 2026-07-06→07-10 burst of **user-space** `~/.claude/jobs/` harness scratch |
| 13 | `.consulted/*.marker` | session markers — largest *ongoing* class |
| 6 | `projects/` | user-space Claude Code auto-memory, also ongoing |
| 70 | (real repo artifacts) | genuine long tail |

`.consulted/` was NOT the dominant historical class (it's ~5% of the total) — the dominant volume was a one-time `jobs/` burst. `normalize_rel_path` splits on the *last* `/.claude/` with no `REPO_ROOT` containment check, so it can't distinguish the user-space Claude home from the repo's — that containment fix is tracked separately at #1191 and is out of scope here. `.consulted/` and `projects/` genuinely are the two classes still arriving wave over wave (by `first_seen` recency), so the intake filter (`SKIP_REL_PREFIXES`) now includes `.consulted/`, `projects/`, **and `jobs/`** (added during review so the very first `archive-wave` run doesn't mislabel the historical burst as a genuine undecided candidate). Corrected the narrative in the module docstring and `SKILL.md` Step 12.5.0 to match the measured split.

**Safety:** `archive_wave_pending()` no-ops (`archived: False`, no file written) on a missing, empty, or malformed pending ledger. The wave label is a plain caller-supplied string; an absent/blank label falls back to a timestamped filename instead of fabricating a wave number. Re-running with the same wave label accretes into the same archive file rather than clobbering it. No `restore` counterpart to `archive-wave` exists yet (Santiago, non-blocking) — worth a follow-up issue if the archive is ever needed to recover a mis-cleared candidate; not filed yet, flagging here for triage.

## Test plan

- [x] `ArchiveWavePendingTest` covers: missing/empty/malformed pending ledger, noise-dropped-but-archived, genuine-candidate-retained-in-archive, mixed partition, missing-wave-label fallback, same-wave re-run accretion, **archive-before-clear on a blocked write** (new), and **user-space `jobs/`/`projects/` partition as noise not genuine** (new).
- [x] `NormalizeRelPathTest` covers the `.consulted/` intake fix and the new `jobs/`/`projects/` intake fix.
- [x] Proved the original fix's tests fail against unfixed source (`git stash` source-only changes, kept tests): 9 failures, then restored → 34/34 pass.
- [x] Proved the review-response tests fail against the pre-review-fix source: `git stash` the narrative/prefix-list changes → 2 new tests fail (jobs/projects partition + intake), archive-before-clear test still passes (the implementation was already correct, only the guard was unasserted) — matches Santiago's finding exactly.
- [x] Manually reproduced Santiago's surviving mutant (reorder ledger-clear above the archive write) in a scratch copy and confirmed the new test now fails against it (dies).
- [x] Full local suite green: `pytest .claude/lib/tests/ .claude/hooks/tests/` (3173 passed), `ruff format --check`/`ruff check`, `mypy`, `pre_commit_ci_sync.py` OK, `pre-commit run` (commit + push stage) all green, cspell green.

Fixes #1140.


parametrization:
**Review: LGTM — Approved**

Scope: correctness of the noise/genuine partition predicate, test non-inertness, `.gitignore`/`.cspell` edits, and the Step 12.5.0 ordering doc. CI: all 21 checks SUCCESS against head SHA `c13d4eb4a17ad87e7a72f3a164db02217d9430f5` (fetched via `statusCheckRollup`, not just the PR-number view).

**1. Partition predicate (`_is_noise_rel` / `SKIP_REL_PREFIXES`).** Traced every `.claude/.consulted/` writer in the repo (`_consultation_sentinel.py`, `enforce_librarian_consulted.py`, `post_label_change_wave_field_sync.py`). All three write only hook-internal sentinel/cache artifacts (marker files, a TTL'd project-ID cache) — `.claude/.consulted/` is *already* fully `.gitignore`d (`.gitignore:36`) before this PR, so nothing that could ever be a genuine genericize candidate lives there by construction. The boundary case you asked me to probe ("a `.consulted/` marker for a file that is a real wave artifact") can't occur.

More importantly, the design defuses the classification-risk concern structurally: `archive_wave_pending` writes **both** `noise_dropped` and `genuine_reset` to the same archive file and only ever clears the *live* ledger — a misclassified entry is mislabeled, not lost. I verified this by inverting the predicate (`return not any(...)`) in a throwaway worktree at the exact PR head SHA: 4 tests catch it (`test_genuine_candidate_retained_in_archive_not_silently_dropped`, `test_mixed_noise_and_genuine_partitioned_correctly`, `test_noise_class_dropped_from_live_but_archived`, `test_rerun_within_same_wave_accretes_not_clobbers`), confirming the partition logic is genuinely exercised, not just intake-tested.

**2. Synthetic fixtures vs. reality.** I found a real, live `.claude/generic_prompt_pending.json` on this machine (267 candidates) and ran `archive_wave_pending` against a copy of it (scratch `pending_path`/`archive_dir`, not the live file). True split: **13 `noise_dropped` / 254 `genuine_reset`**. Worth flagging: this is a materially different composition than the PR/code-comment narrative ("`.consulted/*.marker` noise was the dominant class") — in this live sample noise is ~5%, not the majority. This doesn't change the correctness of the fix (the reset solves the volume problem regardless of composition, and nothing is destroyed either way), but the "dominant noise class" framing in the module docstring and SKILL.md may be over-stated relative to what a live ledger actually looks like today. Non-blocking — just flagging for accuracy, not asking for a respin.

**3. Test non-inertness.** Reproduced your mutation independently in a separate throwaway worktree (`git worktree add --detach` at `c13d4eb`, not the shared one): restored `generic_prompt_tracker.py` + `.gitignore` + `SKILL.md` + `project-words.txt` to the pre-PR commit while keeping the PR's test file, ran `ENVIRONMENT=test python3 -m pytest .claude/lib/tests/test_generic_prompt_tracker.py -q` → **9 failed, 25 passed** (8 `AttributeError: archive_wave_pending` + 1 `.consulted` intake regression), exactly matching your report. Restored to head → 34/34 pass. Combined with the predicate-inversion mutation above, this is solid evidence the suite isn't inert.

**4. `.gitignore` / `.cspell`.** `.claude/generic_prompt_pending_archive/` is a brand-new namespace — no over-match risk against anything that should stay tracked. `genericized` is genuinely used in new SKILL.md prose (Step 12.5.0 tally line) and wasn't already covered.

**5. Step 12.5.0 ordering.** Doc places the `archive-wave` invocation immediately before the "(a) Belt-and-suspenders" diff-sweep code block, matching the stated intent (reset before re-seed). Confirmed against the diff.

Must-fix: None
Tech-debt: None (the dead-path-candidate observation, e.g. stale `c397.md`/`c398.md` entries in the live ledger with no file on disk, is self-resolving via this PR's own wave-boundary reset — not worth a separate issue)

---
Requestor: Lucas Ferreira
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Merge-gate review — release/operational safety lens

Reviewed at head `c13d4eb4a17ad87e7a72f3a164db02217d9430f5`, diffed against `main` via the GitHub contents API (not a local checkout). All 21 checks report SUCCESS at that SHA. Mutation-tested the new guards; ran the head-SHA `archive_wave_pending` against a copy of the **real** pending ledger on this machine.

The design is right and the core mechanism is sound. Two things need to change before merge, one of which is a measurement that contradicts the PR's stated root cause.

---

### The measurement gap is now closed — and it changes the root-cause story

You flagged honestly that you could not reproduce the live wave-28 before/after because the ledger is per-machine and gitignored. That was the right call. I have the org main checkout (`8cb765d`, clean tree), so the real ledger was reachable. Running the head-SHA `archive_wave_pending` against a **copy** of it:

```
{"archived": true, "wave": "P10W29", "noise_dropped": 13, "genuine_reset": 254}
```

Prefix breakdown of that 267-entry snapshot (the ledger grows live — a read minutes earlier showed 266; the tracking hook is appending during the session):

| entries | prefix | what it actually is |
|---|---|---|
| **178** | `jobs/<id>/tmp/*` | `~/.claude/jobs/…` — **user-space** harness scratch. `<repo>/.claude/jobs` does not exist; `~/.claude/jobs/a36d08f0/tmp/` does. |
| 26 | `lib/` | real repo artifact |
| 17 | `hooks/` | real repo artifact |
| **13** | `.consulted/` | session markers — the class this PR filters |
| 12 | `team/` | real |
| **6** | `projects/-home-…-main/memory/*` | **user-space auto-memory notes** |
| 6 / 3 | `skills/`, `plugins/` | real |
| 4 | `c397.md`, `c398.md`, `rv442.md`, `rv442b.md` | stray scratch, no file at `<repo>/.claude/<key>` |

So `.consulted/` is **13 of 267 (4.9%)**, not the dominant class. The dominant class is 178 user-space `jobs/` paths, and **184 of 267 (69%) are not repo artifacts at all**. Root cause: `normalize_rel_path` splits on the *last* `/.claude/` with no check that the path is under `REPO_ROOT`, so the user-space Claude home is indistinguishable from the repo's `.claude/`.

Two consequences worth separating, because they land differently:

- **Your fix is still correct about the *ongoing* inflow.** By `first_seen`, the `jobs/` flood is a one-time burst (2026-07-06 → 07-10, zero since). Recent-week arrivals are `.consulted/` ~13 and `projects/` ~6. So `.consulted/` genuinely is the largest *live* noise class — your filter closes roughly two-thirds of the ongoing inflow. Good call.
- **But the first real run will report `13 noise / 254 genuine`**, presenting 184 non-artifact paths to a human as "a real candidate that fell off the end of a wave". That partition is the headline feature of this PR, and on its very first and most consequential run it is off by ~2.6× in the misleading direction.

`projects/` is the residual ongoing class and is already declared local-only at `.gitignore:31`, right alongside `.consulted/` at line 36.

### Guard mutation-testing — 4 of 5 killed, and the survivor is the one your safety argument rests on

Rebuilt the head-SHA module + test file in a scratch tree (34 baseline pass) and mutated:

| mutation | result |
|---|---|
| drop `.consulted/` from the prefix tuple | KILLED (3 failed) |
| archive clobbers instead of accreting | KILLED |
| noise bucket written as `{}` (silently discarded) | KILLED |
| fabricate `P10W29` when the label is absent | KILLED |
| **clear the live ledger BEFORE writing the archive** | **SURVIVED — 34 passed** |

Moving `save_pending(…{"candidates": {}}…)` above `archive_dir.mkdir(...)` passes the entire suite. Archive-before-clear is the single property the whole data-safety case rests on, and no fixture can trip it. The implementation is correct — I verified live-ledger preservation on a genuine write failure — it is just unasserted, so nothing stops a future refactor from reordering it.

### What I verified and am satisfied with

- **Failure mode is loud and fails safe.** `main()` has no broad `except`; a raised archive write propagates to a non-zero exit, and it happens *before* `save_pending`. Confirmed empirically: with the archive path blocked, the live ledger is byte-identical afterward. This is the right shape.
- **Wave-label fallback is exactly as described.** The module contains no `subprocess`, no `git rev-parse`, no `os.environ` — there is no path by which a wave number could be inferred. Empty and whitespace-only labels both degrade to `unknown-<ts>`, and the test asserts the filename does not match `wave-P\d+W\d+\.json`. Claim verified.
- **Step ordering is correct.** 12.5.0 before sweep (a) is right, and the sweep feeds `record-candidate` with `$REPO_ROOT/$f` paths from `git diff -- '.claude/**'` — repo-scoped, so the re-seed cannot reintroduce the user-space noise. On a wrapup that aborts between 12.5.0 and the sweep, the live ledger is empty and un-reseeded, but the next wrapup's sweep re-derives from the git diff and the archive holds everything; the residual is exactly the stale carryover being dropped by design. Acceptable — though note recovery is manual JSON surgery, there is no `restore` counterpart to `archive-wave`.
- **Rollout:** 5 files, purely additive, clean `git revert`. The archive directory is gitignored, so a revert leaves no orphaned tracked files.
- **No interaction with #1185** (`wave_status.py` / `test_wave_status.py`) — zero file overlap despite both being wrapup-adjacent this wave.
- Minor, not blocking: the wave label goes into a filename unsanitised. `--wave "a/b"` fails safe (write raises, ledger preserved); `--wave "../x"` would escape the archive directory. The caller is the skill with a literal `P{P}W{M}`, so this is theoretical.

---

### Must-fix

1. **Correct the noise characterisation, and filter the class that is still arriving.** Add `"projects/"` to `SKIP_REL_PREFIXES` — it is the only ongoing unfixed noise class and is already gitignored local-only state. Adding `"jobs/"` on the same line is recommended too: it is historical inflow, but it is what makes the first archive report read honestly (~197 noise / ~70 genuine instead of 13 / 254). Then correct the "dominated by session `.consulted/*.marker` noise" claim in the three places it appears — the module docstring, `SKILL.md` Step 12.5.0, and the PR body — to reflect the measured split above. `SKILL.md` is an operator runbook read at every wrapup; leaving the wrong root cause recorded there means the next person will not think to look at the user-space paths. The principled fix (a `REPO_ROOT` containment check, which makes the prefix list unnecessary) is #1191 — if you would rather do that here instead, it supersedes this item.

2. **Assert archive-before-clear.** Add a test that a failed archive write leaves the live ledger untouched. Twelve lines is enough, and it kills the surviving mutant (verified: passes against head, fails against the reordered variant):

```python
def test_failed_archive_write_does_not_clear_live_ledger(self) -> None:
    seed = {"hooks/foo.py": {"category": "hook", "count": 1}}
    gpt.save_pending({"version": 1, "candidates": dict(seed)}, self.pending)
    blocked = Path(self._tmp.name) / "blocked"
    blocked.write_text("not a directory")   # archive_dir occupied -> mkdir raises
    with self.assertRaises(OSError):
        gpt.archive_wave_pending("P10W29", pending_path=self.pending, archive_dir=blocked)
    live = json.loads(self.pending.read_text())["candidates"]
    self.assertEqual(live, seed)
```

- **#1191** — `normalize_rel_path` ingests user-space `~/.claude/**` as repo artifacts (184 of the real ledger's entries). The root cause under must-fix 1; the prefix list is whack-a-mole against it.
- **#1192** — the pending archive is a non-atomic truncating read-modify-write of what is by then the sole copy of all prior snapshots, and a corrupt archive is silently clobbered while the run reports success. Reproduced: a previously archived `hooks/one.py` is destroyed and the operator is told `archived: True`. Note this is where the `.claude/annunaki/archive/wave-{N}.jsonl` comparison does not hold — that archive is line-oriented and genuinely append-only, so a torn write costs one line; the JSON adaptation does not inherit that property.

Neither blocks: both are cold-archive / pre-existing-predicate scope, and the live ledger is safe in every path I tested.

---

Requestor: Santiago Ferreira
Requestee: Wanjiku Mwangi
TechDebt: #1191, #1192

parametrization:
Both must-fixes addressed, new head `713fb3a3799a296050d176b6c6ed31b08c472634`. Per feedback_pr_review_verdict_format.md §9 a head move drops every prior verdict regardless of materiality — re-requesting from both Lucas Ferreira and Santiago Ferreira.

**Must-fix 1 (root-cause narrative + partition).** Corrected the "dominated by `.consulted/*.marker` noise" claim in all three places (module docstring, `SKILL.md` Step 12.5.0, PR body) to the measured live-ledger split: 178 `jobs/` (one-time historical burst, #1191's root cause), 13 `.consulted/`, 6 `projects/`, ~70 genuine — `.consulted/` is the largest *ongoing* class, not the historical dominant one. Added `"projects/"` and `"jobs/"` to `SKIP_REL_PREFIXES` so the first `archive-wave` run no longer mislabels the historical burst / ongoing `projects/` trickle as genuine undecided candidates.

**Must-fix 2 (archive-before-clear guard).** Added `test_failed_archive_write_does_not_clear_live_ledger` (Santiago's supplied test) pinning that a blocked archive write leaves the live ledger untouched. Reproduced the surviving mutant (ledger-clear moved above the archive write) in a scratch module copy and confirmed the new test now fails against it.

Verification this round:
- `ENVIRONMENT=test python3 -m pytest .claude/lib/tests/test_generic_prompt_tracker.py -q` → 37/37 pass at head.
- `git stash` the narrative/prefix-list source changes (kept the test file) → 2 new tests fail (`jobs`/`projects` partition + intake), confirming they exercise real behavior; `test_failed_archive_write_does_not_clear_live_ledger` still passed against the pre-review-fix source, matching Santiago's finding that the implementation was already correct and only the guard was unasserted.
- Full local suite: `pytest .claude/lib/tests/ .claude/hooks/tests/` 3173 passed; `ruff format --check`/`ruff check`, `mypy`, `pre_commit_ci_sync.py`, `pre-commit run` (commit + push stage) all green; cspell green.
- Rebased onto current `origin/main` (`ee99676`) — clean, no conflicts.

Non-blocking items acknowledged, not actioned here (per your scoping): #1191 (containment fix) and #1192 (archive non-atomicity) left as filed tech-debt; softened the "mirrors annunaki" wording to flag the JSONL-vs-JSON append-only distinction; noted the missing `restore` counterpart in the PR body for future triage.

Still not merging — owner-gated, `--merge` never `--squash`.

parametrization:
## Merge-gate re-review — anchored to `713fb3a3799a296050d176b6c6ed31b08c472634`

Re-anchoring after the head move (my prior verdict and Lucas's both dropped at `c13d4eb`). Both must-fixes landed. **Approved.**

Ancestry confirmed: `c13d4eb` was rebased to `bde3928` and the review response is the `713fb3a` fixup on top; the head is a descendant of current `origin/main`. Still 5 files, 428+/6-, MERGEABLE/CLEAN, **21/21 checks SUCCESS** at this SHA. Scope did not creep.

### Must-fix 2 — the mutant is dead

This is the one that mattered, so I re-ran it rather than take it on trust. Rebuilt the mutation harness from `713fb3a3` (37 baseline pass) and replayed the mutant that survived at `c13d4eb`:

| mutation | at `c13d4eb` | at `713fb3a3` |
|---|---|---|
| clear the live ledger BEFORE writing the archive | **SURVIVED (34 passed)** | **KILLED (1 failed)** |
| drop `"projects/"` from the prefix tuple | n/a | KILLED (2 failed) |
| drop `"jobs/"` from the prefix tuple | n/a | KILLED (2 failed) |

Archive-before-clear is now pinned. A future refactor that reorders it fails the suite.

One mutant I tried does survive — wrapping `archive_dir.mkdir(...)` in a bare `except OSError: pass`. That is an **equivalent mutant, not a gap**: with the directory occupied by a plain file, `archive_path.write_text(...)` then raises `NotADirectoryError` (itself an `OSError`), still before `save_pending`, so the ledger-preservation assertion holds and the test's `assertRaises(OSError)` is still satisfied. The property under test survives the mutation because the property is genuinely still true. No action needed.

**On the fails-first result:** you reported that the archive-before-clear test still passes against pre-fix source and declined to claim otherwise. That is exactly right and exactly what I found — the implementation was always correct; only the guard was unasserted. A regression guard whose value is forward-looking is not weakened by not failing first, and reporting it honestly is worth more than a manufactured red. Same discipline as flagging the measurement gap in the first place.

### Must-fix 1 — narrative and partition

The corrected composition in all three places (module docstring, `SKILL.md` Step 12.5.0, PR body) matches my measurement exactly: 267-entry snapshot, 178 `jobs/`, 13 `.consulted/`, 6 `projects/`. The "wrong about the pileup, right about the live noise" distinction survived intact, including the `first_seen`-recency basis for it.

Re-measured the real ledger against the new head — the reason for this fix now shows up in the number that operators will actually see:

```
before (c13d4eb):  noise_dropped 13   genuine_reset 254
after  (713fb3a3): noise_dropped 197  genuine_reset 70
```

The remaining 70 are `lib` 26, `hooks` 17, `team` 12, `skills` 6, `plugins` 3, `CLAUDE.md`, `settings.json` — all real repo artifacts — plus 4 stray scratch keys (`c397.md`, `c398.md`, `rv442.md`, `rv442b.md`) with no file at `<repo>/.claude/<key>`. That residual is #1191's scope, and at 4 of 70 it no longer distorts the worklist. This is now an honest, actionable partition.

Two things I checked before accepting the new prefixes:

- **No false suppression.** `git ls-files '.claude/projects/*' '.claude/jobs/*'` returns zero tracked files, and `.claude/projects/` is already declared local-only at `.gitignore:31`. Neither prefix can shadow a real repo artifact.
- **The cited norm is not a dangling reference.** The added parenthetical points at "Dominant-class characterization of a weighted remainder" — it genuinely exists at `SKILL.md:263` (P8W23 retro, main#923), and citing it here is apt: over-attributing a remainder to a favorable class is precisely the failure it describes. Good instinct to close that loop rather than just fix the number.

Also confirmed folded in: the `annunaki/archive` parity wording is softened, the missing `restore` counterpart is noted, and the PR body did land despite the `gh pr edit` GraphQL no-op — reading it back after a silent failure was the right move.

### Tech-debt — unchanged, still not blocking

**#1191** (the `REPO_ROOT` containment root cause) and **#1192** (non-atomic archive rewrite, silent clobber of a corrupt archive) remain open and correctly out of scope here. Deliberately leaving #1191 untouched was the right call — the prefix list is an interim mitigation and the docstring now says so, which is what keeps the next person from mistaking it for the fix.

Not merging — owner-gated, and `--merge` never `--squash`.

---

Requestor: Santiago Ferreira
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: #1191, #1192

parametrization:
**Review: LGTM — Approved (re-anchored to new head)**

Re-review per the head move. Ancestry note: `c13d4eb` is **not** an ancestor of `713fb3a` (`git merge-base --is-ancestor` fails — the branch was rebased, not fast-forwarded: `713fb3a` -> `bde3928` -> `ee99676`, where `ee99676` carries in unrelated main commits, e.g. the AST-strip/verdict-format memory notes). So I verified the substantive delta via direct two-SHA content diffs (`gh api contents/<path>?ref=<sha>` at both SHAs, byte-diffed), not the `compare` API's file list, which pulled in rebase-carried files (`.gitignore`, `.cspell/project-words.txt`, `.claude/memory/*`, `ontology/checksums.json`) that are **byte-identical** between the two heads — confirmed, no re-review needed there. CI: 21/21 SUCCESS against `713fb3a3799a296050d176b6c6ed31b08c472634`; `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.

**1. Corrected composition — verified against my own live ledger, exactly.** Re-copied the live per-machine `.claude/generic_prompt_pending.json` fresh (still 267 candidates) and counted by prefix directly: **178 `jobs/`, 6 `projects/`, 13 `.consulted/`, 70 remaining** — matching Santiago's 178/6/13 breakdown exactly. Running the new `archive_wave_pending` (with `jobs/`/`projects/` added to `SKIP_REL_PREFIXES`) against this same snapshot gives `noise_dropped=197, genuine_reset=70` — so the corrected narrative isn't just plausible, it reproduces on a second independent ledger. My prior 13/254 was the partition function reporting correctly against the *old* prefix list; the finding (184/267 non-repo paths hiding in "genuine") is real and #1191 is the right place for the underlying containment-check fix. Docstring, `SKIP_REL_PREFIXES` comment, and SKILL.md Step 12.5.0 all carry the corrected numbers consistently — checked all three.

**2. Archive-before-clear guard — confirmed it catches exactly the mutation it's meant to.** Read the new `test_failed_archive_write_does_not_clear_live_ledger` and the implementation: `archive_dir.mkdir(...)` + `archive_path.write_text(...)` still precede `save_pending(...clear...)` in `archive_wave_pending`, unchanged in position from before. I independently reproduced the reorder mutation in a throwaway worktree (moved the clear-`save_pending` call above the archive write/mkdir) and reran the full suite: **1 failed / 36 passed** — the sole failure was the new guard test, exactly as intended, with every other test (including the ones that only check the returned counts) blind to the reorder. Agree with the "regression guard, not fails-first" framing — the underlying order was always correct; the gap was the missing pin, and it's pinned now.

**3. New intake tests** (`test_user_space_jobs_and_projects_return_none`, `test_user_space_jobs_and_projects_are_noise_not_genuine`) — read them, they exercise both the `normalize_rel_path` intake skip and the `archive_wave_pending` partition for the two new prefixes with realistic `~/.claude/jobs/<id>/tmp/...` and `~/.claude/projects/...` paths. Consistent with what I found on the real ledger.

Full suite at head: 37 passed (34 + 3 new).

Must-fix: None
Tech-debt: None (root-cause containment-check fix is already tracked at #1191, correctly left out of this PR's scope)

---
Requestor: Lucas Ferreira
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: none

PR #1143, opened 2026-07-27
fix(kickoff-hook): merge-model-aware branch base + wrapper-tolerant label parsing + reconciliation sweep (#1141)

parametrization:
Closes #1141. Closes #1145.

Two defects in `.claude/hooks/post_wave_kickoff_comment.py`, plus the durable third layer the issue asks for. Both defects failed **silently in the unsafe direction** — labels land, the wave looks kicked off, nobody is told they own anything — so the silent no-op is treated here as the primary bug class, not the parse gap.

## 1. The kickoff template ignored the wave's merge model

`render_kickoff_comment` hardcoded `deployments/phase-{P}/wave-{M}`; `merge_model` appeared nowhere in the hook. Since the 2026-06-09 every-wave-merges-to-main directive, `direct-to-main` is the common case, so this was the default-wrong path — waves 28 and 29 were both direct-to-main and every kickoff comment named a ref that will never exist.

- `read_merge_model(status, wave)` — top-level `wave_{M}_merge_model`, falling back to `wave_{M}_scope.merge_model`. An unrecognized value is treated as unreadable rather than passed through.
- `resolve_branch_base(model, wave, phase)` — `main` for `direct-to-main`, `deployments/phase-{P}/wave-{M}` for `wave-branch` (unchanged), `main` for absent/unreadable plus an explicit *"model NOT declared — confirm the base with the orchestrator before opening the PR"* note. Asymmetric on purpose: a nonexistent wave branch hard-blocks the implementer; a `main` base on a wave-branch wave is a recoverable retarget.
- The comment now also renders a `- PR base:` line naming the model, so the implementer knows which model is in force.
- `merge_model` defaults to `None` (i.e. `main`), so a caller that never heard of merge models cannot emit a dead ref.

Regression fixture per model: `merge_model_direct_to_main.json`, `merge_model_scope_key_wave_branch.json`, `merge_model_absent_defaults_to_main.json`. The pre-existing happy-path fixture now declares `wave-branch` explicitly, so its wave-branch base is asserted for the model it belongs to instead of as an unconditional hardcode.

## 2. Ordinary command wrappers defeated the parser

Root cause was one line, and it is not in this hook: `_shell_parse.find_gh_subcommand` required `gh` at `segment[0]`. `timeout 45 gh …` puts `timeout` there; a `for … ; do gh … ; done` loop puts `do` there. Same for the `git` twin.

[...]

Issue #1189, opened 2026-07-30
tech-debt(wave_status): closingIssuesReferences numbers are not repo-qualified — cross-repo closes can silently under-count or mis-attribute


parametrization:
Found during merge-gate review of #1185 (`main#1131`, direct-to-main wave counters).

## Problem

`_pr_closing_issue_numbers()` in `.claude/lib/wave_status.py` selects only `nodes{number}` and returns bare `int`s:

```
"closingIssuesReferences(first:25){nodes{number}}"
```

`_merged_prs_direct_to_main()` then intersects that set with the **per-repo** canonical scope set:

```python
closes = _pr_closing_issue_numbers(repo, pr["number"])
if not closes & canonical_issues:
    continue
```

Nothing checks which repo the closing issue actually lives in. A PR in repo A that closes an issue in repo B — normal in this org, where meta-issues in `noorinalabs-main` are linked from child-repo PRs — yields B's raw number compared against A's canonical set:

- **under-count** if the number doesn't appear in A's set (the wave PR is silently dropped), or
- **mis-attribution** if the numbers happen to collide across repos.

## Evidence

`repository{name}` is available on the same node and costs nothing — confirmed live against the real engine:

```
$ gh api graphql -f 'query=...closingIssuesReferences(first:25){nodes{number repository{name}}}...' \
    -f owner=noorinalabs -f name=noorinalabs-main -F number=1173
{"data":{"repository":{"pullRequest":{"closingIssuesReferences":{"nodes":[{"number":1172,"repository":{"name":"noorinalabs-main"}}]}}}}}
```

No live collision in wave-29 today (`noorinalabs-main` canonical numbers are 1114–1180; `noorinalabs-isnad-ingest-platform`'s only row is #152), so this is debt rather than a blocker on #1185.

## Why it is currently untestable

`_FakeGhDirectToMain` in `.claude/lib/tests/test_wave_status.py` stores `closes` as bare ints (`"closes": [1172]`) with no repo, so the fixture model **cannot express** a cross-repo closing reference. The gap is structurally unreachable by the existing suite.

## Suggested fix

- Add `repository{name}` to the GraphQL selection; have `_pr_closing_issue_numbers` return `set[tuple[str, int]]` (or filter to the queried repo).
- Extend the fixture `closes` entries to carry a repo so the case becomes representable, and add a regression test for a cross-repo closing reference.

[...]

PR #1193, opened 2026-07-30
fix(commit-identity): route interpreter option runs through the shared grammar (#1149)

### LINKED ISSUE #1149: bug(commit-identity): non-bare interpreter command-string flag bypasses the indirect-exec gate — 8 of 9 shapes allow an identity-less commit

### LINKED ISSUE #1196: validate_commit_identity: call-site comment for invocation.words says 'non-flag word' but words is now the full-token superset (#1193)
Surfaced by the merge-gate review of #1193 at head `a6f6a58`. Non-blocking; the PR is otherwise correct.

## What

`.claude/hooks/validate_commit_identity.py:513-517` — the comment at the call site of `invocation.words`:

```python
        if invocation.has_command_string:
            # `words`, not `operands`: a cluster that mixes a value-taking
            # letter with `c` can shift the payload index, and differently
            # between shells. Guessing that index wrong fails OPEN, so the gate
            # scans every non-flag word of the invocation.
            if any(_payload_looks_like_commit(word) for word in invocation.words):
```

The last clause is no longer true. As of `a6f6a58`, `words` is `tuple(t for t in rest if t != "--")` — every token **including** flag-shaped ones. That widening is deliberate and correct: it is what closes `zsh -abc '-x; git commit -m x'`, where `_consume_wrapper_options` swallows the payload into the option run so `operands` is empty and no `--` ever appears.

## Why it is worth fixing rather than ignoring

`validate_commit_identity.py` is byte-identical across `9aeb5a1..a6f6a58` — the author changed nothing here, the semantics moved underneath the comment. That is precisely the drift that is hard to catch: the comment sits at the one place a future maintainer stands while deciding whether `words` is "too broad" and could safely be narrowed back to non-flag tokens. Narrowing it back is mutation `M11`, which fails 18 tests — but the comment currently reads as an invitation to try.

## Fix

Replace the final clause with the actual contract, and point at the `InterpreterInvocation` docstring which now states it in full:

> `words` is every token of the invocation except the `--` sentinel — deliberately a superset of `operands`, because `end` is only a lower bound on where the option run stops. See `_shell_parse.InterpreterInvocation`.

## Acceptance

- [ ] The comment describes the superset contract, not "non-flag word"
- [ ] No behaviour change; suite stays at 3198 passed

parametrization:
Closes #1149. Closes #1196. Under umbrella #1150.

## The defect

`_DASH_C_RE` required the command-string flag to be a **bare** `-c` immediately after the interpreter. `_SCRIPT_INVOKE_RE` had the identical defect one matcher over — it required the script path in that same position. Both are the #1150 shape: a hook re-implementing command parsing in a grammar narrower than the real command's, failing **open**.

This gate is what stands between an agent and an unvalidated commit, and an identity-less commit that slips it gets mechanically re-attributed later (the #1134/#1161 failure). Pre-existing on `main`; not caused by #1143.

## The fix

Both shapes now resolve through one tokenized pass: `_shell_parse.parse_interpreter_invocation` normalizes the interpreter's own option tokens and then defers to the **shared** `_consume_wrapper_options`. One option-run implementation in the module instead of two divergent regexes.

The two regexes are **retained but demoted** to fail-closed fallbacks for input shlex cannot tokenize, and labelled as such in the source. That is deliberate: `_COMMIT_FALLBACK_RE` does not fire on a `git commit` buried inside a `-c` string, so deleting them would have opened a bypass on exactly the malformed input the rest of the hook fails closed on. The walker is purely additive, which is what makes the differential below clean by construction as well as by measurement.

## Grammar measured, not assumed

Every rule came from an execution oracle against bash / sh / zsh / dash — run the form, ask the filesystem whether a runtime-generated marker appeared. Four results a from-first-principles fix would have got wrong:

| measured | consequence |
|---|---|
| `bash -cl CMD` **runs** CMD | `c` need not be last in the cluster — the `-[a-z]*c$` regex widening the issue offered as "minimum viable" is still wrong |
| `bash +x -c CMD` **runs** CMD | `+`-form options are options; `_looks_like_flag` only sees a leading `-` |
| `bash -oc pipefail CMD` **runs** CMD | a value letter *inside* a cluster still eats the next word, shifting the payload by one |
| `bash -- -c CMD` does **not** run CMD | `--` ends the option run — the shared primitive already models this correctly, and it is pinned as an ALLOW |

`zsh -cO` / `-Oc` / `-cOl` also execute, and there the payload lands in an option-**value** slot. That is why the gate scans `InterpreterInvocation.words` (the superset including option values) rather than `operands` — scanning `operands` alone fails open on those three, which mutation M6 demonstrates.

## The nine forms from the issue

| form | base | head |
|---|---|---|
| interpreter + bare `-c` | BLOCK | BLOCK |
| combined `-lc` | allow | **BLOCK** |
| combined `-ic` | allow | **BLOCK** |
| combined `-xc` | allow | **BLOCK** |
| combined `-ec` | allow | **BLOCK** |
| `-l` then separate `-c` | allow | **BLOCK** |
| `--login` then separate `-c` | allow | **BLOCK** |
| `-o pipefail` then separate `-c` | allow | **BLOCK** |
| `zsh` with combined `-lc` | allow | **BLOCK** |

## Base-vs-head differential

Not a want/got table: the same corpus was run against `git show origin/main:` copies of the hook in a separate temp tree and against this branch, in isolated subprocesses so the two module versions cannot share import state.

85 shapes across 9 families, each varying **one** dimension independently — interpreter identity, cluster spelling, option position, payload quoting, segment role, script-file options — plus a must-stay-allowed family and a previously-blocked family.

```
WEAKENED (block -> allow):        0
newly blocked (allow -> block):  59
must-stay-allowed preserved:  13/13
```

## False-positive cost

Widening a gate has two failure directions, and a bypass-only corpus never asks about the second. Base vs head over **7,616 real recorded Bash commands** (the repo's own `real_comment_invocations.jsonl` fixture, every Bash invocation in this machine's transcripts, and 42 hand-built ordinary shapes):

```
base BLOCK count: 6      head BLOCK count: 6
allow -> BLOCK:   0      BLOCK -> allow:   0
```

Zero verdict changes on real traffic, in either direction.

## Tests are not inert

`.claude/hooks/tests/test_interpreter_option_grammar.py` — 31 tests, 160 subtests. `ShellTruthTests` is the load-bearing class: it runs a **real shell**, decides "executed" by whether a runtime-assembled marker file appeared (so a payload that is merely *displayed* can never score as executed), and requires the hook to block every form the shell actually ran. It also asserts the oracle can answer NO, so the class cannot degrade into vacuity.

Proven against mutants — 11 mutations, **11 killed, 0 survivors**:

| | mutation | result |
|---|---|---|
| M0 | both sources reverted to `origin/main` | killed (cannot even import) |
| M0b | same revert, behavioural classes only | killed — **126 failures** |
| M1 | de-clustering disabled | killed |
| M2 | value-letters not reordered | killed |
| M3 | `+x`/`+o` normalization removed | killed |
| M4 | quote repair removed | killed |
| M5 | `--` no longer ends the option run | killed |
| M6 | scan `operands` instead of `words` | killed |
| M7 | walker removed (regexes only) | killed |
| M8 | script shape stops reading operands | killed |
| M9 | interpreter basename not resolved | killed |

M6 initially **survived** and that is worth recording: the conservative `words` scan had no behavioural test, and killing it required going back to the shell oracle to find `zsh -cO`, the shape that actually justifies it. A defensive branch nobody can write a failing test for is a branch nobody has justified.

## Extra fix, declared not silent

The `<shell> [options] <script>` family is repaired in the same pass, because it is the same defect in the adjacent matcher and the walker yields it for free. `bash -x <script>`, `bash -l <script>`, `bash +x <script>`, `bash -- <script>` all execute under all four shells and none of them matched `_SCRIPT_INVOKE_RE`. Six rows of the differential's `newly blocked` column.

Also closed: appending **one unmatched quote** defeated shlex and restored every form (`bash -lc "git commit -m x" && echo "unclosed`). `iter_interpreter_invocations` retries once per quote character. Repairing the *quoting* rather than splitting the text on separators is what keeps that safe — quoted prose stays a single token, so a malformed `gh issue comment --body '…bash -lc "git commit"…'` still resolves to a `gh`-headed segment and matches nothing. The FP sweep above is the evidence.

## Gates

`ruff-format`, `ruff-lint`, `mypy`, `pytest` (3192 passed), `pytest-skills`, `checksums-ascii`, `memory-budget`, `headcount-budget`, `doc-freshness`, `office-drift`, `mermaid-render` — all green at pre-push. `pre_commit_ci_sync.py` clean.

## Known bound

An interpreter held in a variable, reached via an alias, or invoked through a non-wrapper relay (`xargs bash -c`, #1168) is still out of reach — unchanged by this PR and consistent with #1150's recorded out-of-scope list.

### DIFF
--- a/.claude/hooks/_shell_parse.py
+++ b/.claude/hooks/_shell_parse.py
@@ -1178,12 +1178,26 @@ class InterpreterInvocation(NamedTuple):
     shell-accurate answer and is what a consumer that must resolve a real path
     (the script-invocation check) should use.
 
-    `words` is every non-flag token in the invocation, including words consumed
-    as option VALUES. It is a deliberate superset of `operands` for consumers
-    that are gates: a cluster mixing a value-letter with `c` can shift the
-    payload index by one in ways that differ between shells, and a gate that
-    guesses wrong fails OPEN. Option values are short words like `pipefail`, so
-    scanning them too costs nothing and removes the whole class of index error.
+    `words` is every token of the invocation except the `--` sentinel — the set
+    the command string is guaranteed to be a member of. It is deliberately a
+    SUPERSET of `operands`, and consumers that are gates must use it, because
+    `operands` alone fails OPEN in two measured ways:
+
+      - a cluster mixing a value-letter with `c` shifts the payload index, and
+        differently per shell: `zsh -cO '<cmd>'` runs `<cmd>`, but the shared
+        grammar pairs the clustered `-O` with it as a VALUE, so `operands` is
+        empty;
+      - `end` is only a LOWER bound on where the option run stops.
+        `_consume_wrapper_options` treats every dash-leading token as an option,
+        so a command string that itself starts with `-` is swallowed into the
+        run. Both `bash -c -- '-x; git commit …'` and `zsh -abc '-x; git
+        commit …'` really execute, and both left `operands` empty.
+
+    Rather than re-derive the exact boundary — reintroducing precisely the
+    per-shell option knowledge this module exists to avoid — `words` keeps
+    everything. The extra tokens are option spellings (`-l`, `--login`,
+    `pipefail`); none can carry a `git … commit` bridge, and the cost was priced
+    at zero verdict changes over 7.6k real recorded commands.
     """
 
     name: str
@@ -1258,7 +1272,7 @@ def parse_interpreter_invocation(segment: list[str]) -> InterpreterInvocation | 
     end = _consume_wrapper_options(rest, SHELL_VALUE_OPTIONS, 0)
     has_command_string = _COMMAND_STRING_FLAG in rest[:end]
     operands = tuple(rest[end:])
-    words = tuple(t for t in rest if t != "--" and not _looks_like_flag(t))
+    words = tuple(t for t in rest if t != "--")
     return InterpreterInvocation(name, has_command_string, operands, words)
 
 
--- a/.claude/hooks/_shell_parse.py
+++ b/.claude/hooks/_shell_parse.py
@@ -1126,6 +1126,196 @@ def strip_command_prefixes(segment: list[str], *, compound_leaders: bool = True)
     return tokens
 
 
+# ---------------------------------------------------------------------------
+# Shell-interpreter invocations (main#1149)
+# ---------------------------------------------------------------------------
+#
+# `validate_commit_identity` matched `<shell> -c <payload>` with a regex that
+# required `-c` to be a BARE token immediately after the interpreter. Real
+# shells accept the command-string flag combined into a short-flag cluster and
+# after an arbitrary option run, so 8 of 9 ordinary spellings walked straight
+# past the gate and committed with whatever ambient identity git resolved.
+#
+# The grammar below is deliberately NOT a second regex. It normalizes the
+# interpreter's own option tokens into the shape `_consume_wrapper_options`
+# already understands and then defers to it, so there is exactly one option-run
+# implementation in this module (main#1150's invariant).
+#
+# Every rule here was measured against bash / sh / zsh / dash with an execution
+# oracle (does the candidate payload actually run?), not reasoned from man
+# pages. The three findings that a from-first-principles fix would have missed:
+#
+#   `bash -cl CMD`  runs CMD — the command-string letter does NOT have to be
+#                   last in the cluster, so a `-[a-z]*c$` regex is still wrong.
+#   `bash +x -c CMD` runs CMD — `+`-form options are options, but
+#                   `_looks_like_flag` only recognises a leading `-`.
+#   `bash -oc pipefail CMD` runs CMD — a value-taking letter INSIDE a cluster
+#                   still eats the next word, shifting the payload by one.
+#   `bash -- -c CMD` does NOT run CMD — `--` ends the option run, which
+#                   `_consume_wrapper_options` already models correctly.
+
+# Interpreter options that consume a following word as their value. `-o`/`-O`
+# name a `set`/`shopt` option; bash's two startup-file options take a path.
+SHELL_VALUE_OPTIONS = frozenset({"-o", "-O", "--rcfile", "--init-file"})
+
+# Cluster letters that consume a following word (the short forms of the above).
+_SHELL_VALUE_LETTERS = frozenset("oO")
+
+# The letter that introduces a command string: `sh -c '<cmd>'`.
+_COMMAND_STRING_FLAG = "-c"
+
+
+class InterpreterInvocation(NamedTuple):
+    """A decoded `<shell> [options] [operands ...]` command.
+
+    `name` is the interpreter basename (`/usr/bin/bash` -> `bash`).
+
+    `has_command_string` is True when the option run carries `-c` in any
+    spelling — bare, clustered (`-lc`, `-cl`, `-cabm`), or after other options.
+
+    `operands` are the tokens after the option run, i.e. what the shell itself
+    treats as `<command-string> [$0 $1 ...]` or `<script> [args]`. This is the
+    shell-accurate answer and is what a consumer that must resolve a real path
+    (the script-invocation check) should use.
+
+    `words` is every non-flag token in the invocation, including words consumed
+    as option VALUES. It is a deliberate superset of `operands` for consumers
+    that are gates: a cluster mixing a value-letter with `c` can shift the
+    payload index by one in ways that differ between shells, and a gate that
+    guesses wrong fails OPEN. Option values are short words like `pipefail`, so
+    scanning them too costs nothing and removes the whole class of index error.
+    """
+
+    name: str
+    has_command_string: bool
+    operands: tuple[str, ...]
+    words: tuple[str, ...]
+
+
+def _expand_shell_option_token(tok: str) -> list[str]:
+    """Rewrite one interpreter option token into `_consume_wrapper_options` form.
+
+    Two normalizations, each closing a measured bypass:
+
+      `+x` -> `-x`, `+o` -> `-o`
+          `set`-style options may be turned off with a leading `+`, and
+          `bash +x -c '<cmd>'` really does run `<cmd>`. `_looks_like_flag` keys
+          on a leading `-` only, so without this the `+x` reads as the wrapped
+          command and the option run ends before `-c` is ever seen.
+
+      `-lc` -> `-l -c`
+          De-clustering makes the command-string flag a plain token-equality
+          test instead of a second regex over cluster spellings, and it lets
+          the shared grammar pair a clustered value-letter with its value.
+
+    Value-taking letters are emitted LAST within an expanded cluster. Within one
+    cluster the letters' order does not change WHICH following words are
+    consumed as values — only the order they are consumed in — so moving them to
+    the end is semantics-preserving for our purpose, and it puts each value flag
+    immediately before the word it consumes, which is the only arrangement
+    `_consume_wrapper_options` can pair up (it refuses to let a value flag eat a
+    flag-shaped token). Without the reorder, `bash -oc pipefail 'git commit …'`
+    resolves the payload to `pipefail` and the gate fails open.
+
+    Long options, `--`, `=`-forms and anything non-alphabetic are returned
+    unchanged: only a `[-+][A-Za-z]{2,}` run is a short-flag cluster.
+    """
+    if len(tok) < 2 or tok[0] not in "-+":
+        return [tok]
+    body = tok[1:]
+    if body.startswith("-") or not body.isalpha():
+        # `--login`, `--rcfile=F`, `--`, `-2`, `-o=x` — not a short cluster.
+        return ["-" + body] if tok[0] == "+" else [tok]
+    if len(body) == 1:
+        return ["-" + body]
+    plain = ["-" + ch for ch in body if ch not in _SHELL_VALUE_LETTERS]
+    valued = ["-" + ch for ch in body if ch in _SHELL_VALUE_LETTERS]
+    return plain + valued
+
+
+def parse_interpreter_invocation(segment: list[str]) -> InterpreterInvocation | None:
+    """Decode `segment` as a shell-interpreter invocation, or return None.
+
+    Transparent command-prefix wrappers are stripped first, so
+    `timeout 30 env FOO=1 bash -lc '<cmd>'` decodes the same as `bash -lc
+    '<cmd>'`. A leading path is reduced to its basename (`/bin/sh` -> `sh`).
+
+    Returns None when the segment is empty or its head is not one of
+    `SHELL_INTERPRETERS` — callers must treat None as "not an interpreter
+    invocation", never as "safe".
+    """
+    tokens = strip_command_prefixes(segment)
+    if not tokens:
+        return None
+    name = tokens[0].rsplit("/", 1)[-1]
+    if name not in SHELL_INTERPRETERS:
+        return None
+
+    rest: list[str] = []
+    for tok in tokens[1:]:
+        rest.extend(_expand_shell_option_token(tok))
+
+    end = _consume_wrapper_options(rest, SHELL_VALUE_OPTIONS, 0)
+    has_command_string = _COMMAND_STRING_FLAG in rest[:end]
+    operands = tuple(rest[end:])
+    words = tuple(t for t in rest if t != "--" and not _looks_like_flag(t))
+    return InterpreterInvocation(name, has_command_string, operands, words)
+
+
+def iter_interpreter_invocations(command: str) -> list[InterpreterInvocation]:
+    """Decode every shell-interpreter invocation in a compound command.
+
+    Heredoc bodies are removed before tokenizing: a body is not part of the
+    invocation's own argument list, and leaving it in would let its words be
+    read as segments. Callers that care about heredoc BODIES (they are executed
+    when the heredoc is fed to a shell) must handle them separately —
+    `classify_heredocs` is the primitive for that.
+
+    Unbalanced-quote repair. A command that shlex cannot tokenize would
+    otherwise yield nothing, and for a fail-closed gate that is a bypass an
+    evader reaches by typing one extra quote character:
+
+        bash -lc "git commit -m x" && echo "unclosed
+
+    The invocation itself is well-formed; only the tail is broken. So on a parse
+    failure we retry once per quote character with that character appended.
+    Repairing the QUOTING (rather than splitting the text on separators, the
+    other obvious recovery) is what keeps this safe: quoted prose stays a single
+    token, so a malformed `gh issue comment --body '…bash -lc "git commit"…'`
+    still resolves to a segment headed by `gh` and matches nothing. The
+    head-anchored `parse_interpreter_invocation` does the rest — a repaired
+    parse can only ever add segments, never move an interpreter into command
+    position that was not already there.
+
+    Returns an empty list when even the repaired text cannot be tokenized. A
+    security gate must therefore still keep a text-level fallback rather than
+    reading an empty list as "nothing here".
+
+    Perf prefilter (#1113 discipline): an interpreter invocation cannot exist
+    unless the interpreter's name appears literally somewhere in the command, so
+    a substring test skips the tokenize for the overwhelming majority of Bash
+    calls. The helpers below are individually memoized, so a command that does
+    reach them is parsed once per process regardless of how many hooks ask.
+    """
+    if not any(name in command for name in SHELL_INTERPRETERS):
+        return []
+    text = strip_heredocs(command)
+    tokens = tokenize(normalize_command_separators(text))
+    if tokens is None:
+        for repair in ("'", '"'):
+            tokens = tokenize(normalize_command_separators(text + repair))
+            if tokens is not None:
+                break
+    if tokens is None:
+        return []
+    found: list[InterpreterInvocation] = []
+    for segment in iter_command_segments(tokens):
+        invocation = parse_interpreter_invocation(segment)
+        if invocation is not None:
+            found.append(invocation)
+    return found
+
+
 def find_git_subcommand(segment: list[str]) -> tuple[list[str], list[str]] | None:
     """If `segment` is a `git ...` invocation, return (global_opts, [subcmd, ...]).
 
--- a/.claude/hooks/tests/test_interpreter_option_grammar.py
+++ b/.claude/hooks/tests/test_interpreter_option_grammar.py
@@ -1,0 +1,506 @@
+#!/usr/bin/env python3
+"""Regression tests for main#1149 — interpreter option grammar in the commit-identity gate.
+
+`validate_commit_identity` decided "is this `<shell> -c '<payload>'`?" with a
+regex that required `-c` to be a BARE token immediately after the interpreter,
+and "is this `<shell> <script>`?" with a regex that required the path to sit in
+the same position. Real shells accept an arbitrary option run first, and accept
+the command-string flag combined into a short-flag cluster. Eight of the nine
+ordinary spellings in the issue therefore let an identity-less commit through,
+and the whole `<shell> -x <script>` family did too.
+
+The fix routes both shapes through `_shell_parse.parse_interpreter_invocation`,
+which normalizes the interpreter's option tokens and then defers to the SHARED
+`_consume_wrapper_options` primitive — one option-run implementation in the
+module instead of two divergent regexes (main#1150's invariant).
+
+What these tests pin, and why in this shape
+===========================================
+
+* `ShellTruthTests` is the load-bearing one. It runs a REAL shell and asks
+  whether the payload actually executed, then requires the hook to block every
+  form that did. The marker is assembled from random bytes at run time and the
+  verdict is "did this file appear on disk", so a payload that is merely
+  DISPLAYED (echoed, error-quoted) can never be scored as executed. Both #1151
+  families survived the old suite precisely because the tests encoded the
+  implementation instead of the shell.
+* The remaining classes vary ONE dimension each — cluster spelling, option
+  position, interpreter identity, payload quoting, segment role — because three
+  consecutive review rounds on main#1155 each found a defect sitting in whatever
+  the previous corpus had held constant.
+* `MustStayAllowedTests` and `PreviouslyBlockedStillBlockTests` are the two
+  no-regression directions. Widening a gate is only correct if nothing that
+  blocked before now passes AND ordinary work is still allowed.
+
+Run: python3 -m pytest .claude/hooks/tests/test_interpreter_option_grammar.py -v
+"""
+
+from __future__ import annotations
+
+import os
+import secrets
+import shutil
+import subprocess
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+
+_HERE = Path(__file__).resolve().parent
+_HOOKS_DIR = _HERE.parent
+sys.path.insert(0, str(_HOOKS_DIR))
+
+import validate_commit_identity as hook  # noqa: E402
+from _shell_parse import (  # noqa: E402
+    SHELL_INTERPRETERS,
+    _expand_shell_option_token,
+    iter_interpreter_invocations,
+    parse_interpreter_invocation,
+)
+
+# An identity-less commit: no `-c user.name=` / `-c user.email=` anywhere. Every
+# shape below is a charter violation the gate exists to stop.
+NAKED_COMMIT = "git commit -m x"
+
+
+def verdict(command: str, *, cwd: str | None = None) -> str:
+    """Run the hook over `command` and return "BLOCK" or "allow"."""
+    payload: dict = {"tool_name": "Bash", "tool_input": {"command": command}}
+    if cwd is not None:
+        payload["cwd"] = cwd
+    return "BLOCK" if hook.check(payload) else "allow"
+
+
+class NineFormsFromTheIssueTests(unittest.TestCase):
+    """The exact nine-row table from main#1149. Eight of these used to pass."""
+
+    FORMS = [
+        ("interpreter + bare -c", f"bash -c '{NAKED_COMMIT}'"),
+        ("combined -lc", f"bash -lc '{NAKED_COMMIT}'"),
+        ("combined -ic", f"bash -ic '{NAKED_COMMIT}'"),
+        ("combined -xc", f"bash -xc '{NAKED_COMMIT}'"),
+        ("combined -ec", f"bash -ec '{NAKED_COMMIT}'"),
+        ("-l then separate -c", f"bash -l -c '{NAKED_COMMIT}'"),
+        ("--login then separate -c", f"bash --login -c '{NAKED_COMMIT}'"),
+        ("-o pipefail then separate -c", f"bash -o pipefail -c '{NAKED_COMMIT}'"),
+        ("zsh with combined -lc", f"zsh -lc '{NAKED_COMMIT}'"),
+    ]
+
+    def test_all_nine_forms_block(self):
+        for label, command in self.FORMS:
+            with self.subTest(form=label):
+                self.assertEqual(verdict(command), "BLOCK", command)
+
+    def test_block_reason_names_the_shell_c_shape(self):
+        result = hook.check(
+            {"tool_name": "Bash", "tool_input": {"command": f"bash -lc '{NAKED_COMMIT}'"}}
+        )
+        assert result is not None
+        self.assertIn("indirect-exec wrapper detected", result["reason"])
+        self.assertIn("shell -c", result["reason"])
+
+
+class ClusterSpellingTests(unittest.TestCase):
+    """Vary the short-flag CLUSTER; hold interpreter and payload constant.
+
+    `-cl` is the row that kills the "widen the regex to `-[a-z]*c`" shortcut the
+    issue offered as a minimum-viable fix: the command-string letter does not
+    have to be last in the cluster, and every shell measured runs the payload
+    anyway.
+    """
+
+    CLUSTERS = ["-lc", "-cl", "-cx", "-ce", "-vc", "-uc", "-fc", "-lxc", "-abc", "-cabm"]
+
+    def test_every_cluster_containing_c_blocks(self):
+        for cluster in self.CLUSTERS:
+            with self.subTest(cluster=cluster):
+                self.assertEqual(verdict(f"bash {cluster} '{NAKED_COMMIT}'"), "BLOCK")
+
+    def test_cluster_without_c_is_not_a_command_string(self):
+        invocation = parse_interpreter_invocation(["bash", "-lx", NAKED_COMMIT])
+        assert invocation is not None
+        self.assertFalse(invocation.has_command_string)
+
+
+class PrecedingOptionTests(unittest.TestCase):
+    """Vary the option run BEFORE `-c`; hold cluster shape and payload constant."""
+
+    OPTION_RUNS = [
+        "-l",
+        "-x",
+        "-e",
+        "-p",
+        "-s",
+        "+x",
+        "--login",
+        "--noprofile --norc",
+        "-o pipefail",
+        "+o histexpand",
+        "-O extglob",
+        "--rcfile /dev/null",
+        "--init-file /dev/null",
+        "-lo pipefail",
+        "-l -x -o pipefail +o histexpand",
+    ]
+
+    def test_option_run_before_c_still_blocks(self):
+        for run in self.OPTION_RUNS:
+            with self.subTest(options=run):
+                self.assertEqual(verdict(f"bash {run} -c '{NAKED_COMMIT}'"), "BLOCK")
+
+    def test_value_letter_inside_cluster_does_not_shift_the_payload_away(self):
+        """`bash -oc pipefail '<commit>'` — the clustered `o` eats `pipefail`.
+
+        Measured under bash: the payload is the word AFTER the option value, so
+        a walker that pairs the cluster naively resolves the payload to
+        `pipefail`, finds no commit shape, and fails open.
+        """
+        self.assertEqual(verdict(f"bash -oc pipefail '{NAKED_COMMIT}'"), "BLOCK")
+
+    def test_payload_consumed_as_an_option_value_is_still_scanned(self):
+        """`zsh -cO '<commit>'` runs the payload, but the shared grammar pairs
+        the clustered `-O` with it as a VALUE, leaving `operands` empty.
+
+        Measured: `-cO`, `-Oc` and `-cOl` all execute under zsh (bash/sh/dash
+        reject them). This is the shape that makes the gate scan
+        `InterpreterInvocation.words` — the superset including option values —
+        rather than `operands`. Scanning `operands` alone fails OPEN here.
+        """
+        for cluster in ("-cO", "-Oc", "-cOl"):
+            with self.subTest(cluster=cluster):
+                inv = parse_interpreter_invocation(["zsh", cluster, NAKED_COMMIT])
+                assert inv is not None
+                self.assertEqual(inv.operands, (), "precondition: operands must be empty")
+                self.assertEqual(inv.words, (NAKED_COMMIT,))
+                self.assertEqual(verdict(f"zsh {cluster} '{NAKED_COMMIT}'"), "BLOCK")
+
+    def test_double_dash_ends_the_option_run(self):
+        """`bash -- -c '<commit>'` does NOT execute the payload under any shell.
+
+        Pinned as an ALLOW so a future widening cannot quietly turn the option
+        grammar into "does the text contain -c somewhere".
+        """
+        self.assertEqual(verdict(f"bash -- -c '{NAKED_COMMIT}'"), "allow")
+
+
+class InterpreterIdentityTests(unittest.TestCase):
+    """Vary the interpreter; hold flag shape and payload constant."""
+
+    def test_every_recognised_interpreter_blocks(self):
+        for name in sorted(SHELL_INTERPRETERS):
+            with self.subTest(interpreter=name):
+                self.assertEqual(verdict(f"{name} -lc '{NAKED_COMMIT}'"), "BLOCK")
+
+    def test_absolute_path_interpreter_blocks(self):
+        for path in ("/bin/bash", "/usr/bin/zsh", "/bin/sh", "/usr/local/bin/dash"):
+            with self.subTest(path=path):
+                self.assertEqual(verdict(f"{path} -lc '{NAKED_COMMIT}'"), "BLOCK")
+
+    def test_non_interpreter_head_is_not_an_invocation(self):
+        """`echo bash -lc '<commit>'` prints; it does not execute."""
+        self.assertIsNone(parse_interpreter_invocation(["echo", "bash", "-lc", NAKED_COMMIT]))
+
+
+class PayloadQuotingTests(unittest.TestCase):
+    """Vary the payload QUOTING; hold interpreter and flag shape constant."""
+
+    def test_quoting_variants_block(self):
+        cases = {
+            "single": f"bash -lc '{NAKED_COMMIT}'",
+            "double": f'bash -lc "{NAKED_COMMIT}"',
+            "backslash-escaped word": "bash -lc git\\ commit",
+            "nested quotes": "bash -lc 'git commit -m \"a b\"'",
+            "inner single in double": "bash -lc \"git commit -m 'a b'\"",
+        }
+        for label, command in cases.items():
+            with self.subTest(quoting=label):
+                self.assertEqual(verdict(command), "BLOCK", command)
+
+    def test_unbalanced_quote_tail_does_not_restore_the_bypass(self):
+        """One stray quote used to defeat the tokenizer and re-open every form.
+
+        The invocation is well-formed; only the tail is broken. Quote repair in
+        `iter_interpreter_invocations` recovers it.
+        """
+        cases = [
+            f'bash -lc "{NAKED_COMMIT}" && echo "unclosed',
+            f'bash -c "{NAKED_COMMIT}" && echo "unclosed',
+            f"bash -lc '{NAKED_COMMIT}",
+            f"bash -o pipefail -c '{NAKED_COMMIT}' ; echo 'unclosed",
+        ]
+        for command in cases:
+            with self.subTest(command=command):
+                self.assertEqual(verdict(command), "BLOCK", command)
+
+
+class SegmentRoleTests(unittest.TestCase):
+    """Vary the STRUCTURAL ROLE of the invocation; hold the invocation constant."""
+
+    def test_invocation_in_any_segment_position_blocks(self):
+        inner = f"bash -lc '{NAKED_COMMIT}'"
+        cases = {
+            "leading": inner,
+            "after &&": f"cd /tmp && {inner}",
+            "after ||": f"false || {inner}",
+            "after ;": f"echo hi; {inner}",
+            "after unspaced ;": f"echo hi;{inner}",
+            "after newline": f"echo hi\n{inner}",
+            "downstream of a pipe": f"echo hi | {inner}",
+            "line continuation": "bash \\\n  -lc '%s'" % NAKED_COMMIT,
+        }
+        for label, command in cases.items():
+            with self.subTest(position=label):
+                self.assertEqual(verdict(command), "BLOCK", command)
+
+    def test_transparent_wrappers_do_not_hide_the_invocation(self):
+        inner = f"bash -lc '{NAKED_COMMIT}'"
+        for wrapper in ("timeout 30", "env FOO=1", "nohup", "nice -n 5", "command", "sudo"):
+            with self.subTest(wrapper=wrapper):
+                self.assertEqual(verdict(f"{wrapper} {inner}"), "BLOCK")
+
+
+class ScriptInvocationOptionTests(unittest.TestCase):
+    """`<shell> [options] <script>` — the same too-narrow grammar, one matcher over.
+
+    `_SCRIPT_INVOKE_RE` required the path to be the token immediately after the
+    interpreter, so every option before it hid the script. All four forms below
+    execute the script under bash / sh / zsh / dash.
+    """
+
+    def setUp(self):
+        self.tmp = tempfile.mkdtemp(prefix="t1149-script-")
+        self.addCleanup(shutil.rmtree, self.tmp, True)
+        self.naked = os.path.join(self.tmp, "naked_script")
+        Path(self.naked).write_text(NAKED_COMMIT + "\n", encoding="utf-8")
+        self.innocent = os.path.join(self.tmp, "innocent_script")
+        Path(self.innocent).write_text("echo hello\n", encoding="utf-8")
+
+    def test_options_before_the_script_path_still_block(self):
+        for pre in ("", "-x", "-l", "+x", "--", "-o pipefail", "-lx"):
+            with self.subTest(options=pre or "(none)"):
+                command = f"bash {pre} {self.naked}".replace("  ", " ")
+                self.assertEqual(verdict(command), "BLOCK", command)
+
+    def test_wrapped_script_invocation_blocks(self):
+        self.assertEqual(verdict(f"timeout 9 bash -x {self.naked}"), "BLOCK")
+
+    def test_innocent_script_is_still_allowed(self):
+        for pre in ("", "-x", "-l"):
+            with self.subTest(options=pre or "(none)"):
+                self.assertEqual(verdict(f"bash {pre} {self.innocent}".replace("  ", " ")), "allow")
+
+
+class MustStayAllowedTests(unittest.TestCase):
+    """Ordinary work must not become collateral. Widening a gate has two failure
+    directions and this is the one a bypass-only corpus never asks about."""
+
+    ALLOWED = {
+        "plain ls": "ls -la",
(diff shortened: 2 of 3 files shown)


parametrization:
**Review: LGTM**

Reviewed for blast radius to the `_shell_parse.py` cd-routing paths I built in #1151/#1156, and independently re-derived the three highest-leverage claims in the PR body (FP-sweep integrity, one mutation kill, and the quote-repair FP surface) rather than taking them on faith. All checked out.

**1. Blast radius to `_shell_parse.py` (my #1151/#1156 concern) — clean.**
`git diff origin/main..head -- .claude/hooks/_shell_parse.py` is **190 insertions, 0 deletions**, all inside a single new hunk appended after `strip_command_prefixes`. Every existing function — `find_git_subcommand`, `_consume_wrapper_options`, `_looks_like_flag`, `strip_command_prefixes`, the AST-based `cd`-routing I landed in `999dfae3` — is byte-identical on both sides. The new names (`InterpreterInvocation`, `parse_interpreter_invocation`, `iter_interpreter_invocations`, `_expand_shell_option_token`, `SHELL_VALUE_OPTIONS`, `_SHELL_VALUE_LETTERS`, `_COMMAND_STRING_FLAG`) don't collide with anything in base. I enumerated every hook importing from `_shell_parse` (18 files, all explicit named imports, no wildcards) — the two write hooks named in my brief, `post_wave_kickoff_comment.py` and `post_label_change_wave_field_sync.py`, only import `resolve_repo_short_name`, which this PR doesn't touch; `auto_add_issue_to_board.py` doesn't import `_shell_parse` at all. `validate_commit_identity.py`'s 6 deleted lines are all docstring prose (the "Shapes checked, in order" restructuring), not logic. No cd-routing regression surface here.

**2. FP-sweep — verified the methodology is real, not head-vs-head.** I built independent base (`origin/main`) and head trees via `git archive`, ran both as separate subprocesses (no shared import state) against the committed `real_comment_invocations.jsonl` fixture (121 real recorded commands — the fixture the PR's larger 7,616-command sweep draws from): **0 BLOCK / 0 BLOCK, 0 diffs**, consistent with the PR's claim. To rule out my own harness silently comparing head-vs-head, I ran a canary (`bash -lc 'git commit -m x'`) through the same two trees first: base → allow, head → BLOCK — the two trees are genuinely different module versions. I can't reproduce the full 7,616 count (it includes the author's own machine transcripts, not committed anywhere), but the mechanism checks out on everything I can independently access.

**3. Mutation — re-ran M6 myself (the one that initially survived).** Patched `invocation.words` → `invocation.operands` in the exported head tree and reran `test_interpreter_option_grammar.py`: **6 failed / 31 passed**, all on the real zsh oracle (`zsh -cO`, `-Oc`, `-cOl`) — the exact shape the PR calls out as the M6 justification. Reverted and confirmed 31/160 green again. Also independently ran `test_oracle_distinguishes_executed_from_not_executed` — the oracle genuinely answers NO for `--` and YES for bare `-c` against a real shell, so it isn't a vacuous always-YES check.

**4. Quote-repair FP surface — no new class found.** Probed 6 adversarial cases combining unbalanced quotes with "bash -lc"/"commit" substrings in ordinary prose (issue/PR comment bodies with contractions — don't, can't, it's, won't — plus one comment that literally quotes the bypass syntax itself as a warning). All 6 stayed `allow`; a control case with a genuine trailing-quote-defeated `bash -lc 'git commit -m x'` still correctly `BLOCK`ed. `shlex` already treats an apostrophe inside a double-quoted string as literal (verified directly), so the repair path isn't even reached for the common contraction case — it only fires on a genuinely unbalanced tail, and the head-anchored `parse_interpreter_invocation` still requires an actual interpreter at segment head, so repaired quoted prose can't manufacture a new invocation.

**5. Design.** Reusing `_shell_parse._consume_wrapper_options` instead of a second regex grammar is the right call per #1150's invariant, and the value-letter reordering + `words`-superset-over-`operands` scanning is exactly what the `-oc pipefail` / `zsh -cO` measured findings require — a naive operand-only scan (M6) does fail open, confirmed above.

CI: 14/14 green at head `9aeb5a18`. `gh pr view` reports MERGEABLE/CLEAN.

Must-fix: None
Tech-debt: None (already tracked separately: #1194 env-split-string bypass, #1150 umbrella, #1174 gate-blocks-own-test-vectors)

---
Requestor: Lucas Ferreira
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Merge-gate review — PR #1193 (main#1149), head `9aeb5a1`.** Reviewed against origin at the head SHA via the contents API, not a local checkout. Verdict: **ChangesRequested** — one fail-open shape inside the new grammar, plus one load-bearing branch with no test. Everything else in this PR holds up under adversarial measurement, including the part you added out of scope.

I did not re-derive the nine-form table or the FP sweep; I attacked the parts that the PR body's own evidence could not have caught.

## Must-fix — `<shell> -c -- '<dash-shaped payload>'` still fails open

`InterpreterInvocation.words` filters `_looks_like_flag` across the **whole** token list, including the operand region:

```python
words = tuple(t for t in rest if t != "--" and not _looks_like_flag(t))
```

After `--` no token is an option — that is what `--` means, and `_consume_wrapper_options` already models it (your ALLOW pin for `bash -- -c` depends on it). But `words` re-applies the flag filter *past* `end`, so a command string that merely starts with `-` is dropped on the floor. `_DASH_C_RE` does not cover the miss either: it captures `--` itself as the payload group and sees no commit verb.

I extended your own `ShellTruthTests` with the one dimension its `FORMS` table holds fixed — the **surface shape of the payload** — and re-ran the same contract (shell executed it, therefore the gate must block it) across bash / sh / zsh / dash:

| tree | executing forms checked | gate holes | distinct shapes |
|---|---|---|---|
| base `ee99676` | 64 | 56 | 14 |
| head `9aeb5a1` | 64 | **24** | **6** |
| head + the one-line fix below | 64 | **0** | **0** |

The six surviving shapes at head are `-c --` and `-lc --` crossed with payload prefixes `-x; `, `--foo; `, `-`, on all four shells. Measured, not reasoned:

```
bash -c -- '-x; …'    EXECUTED rc=0   bash: line 1: -x: command not found
zsh  -c -- '-x; …'    EXECUTED rc=0
dash -c -- '-x; …'    EXECUTED rc=0
```

Two things make this a must-fix rather than a followup, even though base allows it too and the PR is a strict improvement. First, it is the #1150 defect itself — a grammar narrower than the real command's, failing open — sitting *inside* the grammar this PR introduces to eliminate that class. Second, it is reachable by typing two more characters onto the form you just closed, which is the same shape of finding as the unmatched quote you fixed in flight; your own standard for that one should apply here.

One line, in `parse_interpreter_invocation`:

```python
-    words = tuple(t for t in rest if t != "--" and not _looks_like_flag(t))
+    words = tuple(t for t in rest[:end] if not _looks_like_flag(t)) + operands
```

Operands are by definition never options, so they are never flag-filtered; the filter is kept only over the option run, where it belongs. It also retires the `t != "--"` clause, which is dead (see below). Measured on the patched tree:

- the 6 shapes above: all now BLOCK, all four shells
- `.claude/hooks/tests/` + `.claude/lib/tests/`: **3192 passed**, `ruff format --check` / `ruff check` / `mypy` all clean
- **17,252 unique real Bash commands** (116 transcript files on this box plus `real_comment_invocations.jsonl`): base 175 BLOCK, head 175 BLOCK, patched 175 BLOCK — **0 verdict changes in either direction at either step**
- 18 `_shell_parse` consumer hooks over a 2,500-command sample: 0 verdict changes

Please also extend the `words` docstring: it currently justifies the superset by option-value slots, and after the fix it additionally guarantees operands are unconditionally included.

## Also requested this round — the one defensive branch with no justifying test

I ran 14 mutations of my own against the new code. **10 killed.** Four survived; three of them are provably equivalent mutants and one is not:

| survivor | classification |
|---|---|
| `t != "--"` dropped from the `words` filter | **equivalent** — `_looks_like_flag("--")` is already `True`, so the clause is unreachable. Dead code. |
| `if len(body) == 1: return ["-" + body]` early return deleted | **equivalent** — verified over all 104 single-letter `-x`/`+x` tokens, the general `plain + valued` path returns an identical list. |
| perf prefilter deleted | **equivalent for verdicts** — removal only does more work; the failure direction is safe. |
| `return ["-" + body] if tok[0] == "+" else [tok]` reduced to `return [tok]` | **live and untested.** |

That last one is load-bearing. `zsh +2 -c '<payload>'` **executes** (measured; bash rejects it, zsh does not), head correctly BLOCKs it, and with the branch mutated away it goes to **allow**. No test in the 31 covers it — `test_plus_form_options_normalize_to_minus` exercises the alpha-body path only, which the general path would handle anyway. This is M6 again with the sign flipped: for M6 no shape existed until you went back to the oracle, and here the shape does exist, so the test is writable today. Please add it.

## The out-of-scope quote repair — I think you made the right call

Endorsed, and the reasoning is sound. Two things I checked rather than took on faith.

The framing risk I was handed was "unscoped change in the shared parser." It is narrower than that. `head` minus the inserted 190-line block is **byte-identical to base** — I reconstructed it and compared, 0 bytes differ — so the repair lives inside a *new* function, not inside `tokenize` or any pre-existing shared symbol. `iter_interpreter_invocations` / `parse_interpreter_invocation` / `InterpreterInvocation` / `SHELL_VALUE_OPTIONS` have exactly one consumer in the tree. The blast radius is this hook, not eight.

Your stated reason for repairing the quoting instead of splitting on separators is the correct one, and it survives the FP corpus it predicts. Ten prose-and-data shapes carrying an interpreter invocation inside an unbalanced-quote string all stay allowed at head — apostrophe prose (`echo don't run bash -lc '…'`), an unterminated `gh issue comment --body`, an unterminated `rg` pattern, `printf '%s' "bash -lc '…'"`, a trailing `#` comment. All four genuinely evasive shapes block. One correction to the PR body's implicit bound: repair is **not** limited to a single unmatched quote — a payload with three unmatched quotes still blocks, because shlex only needs the outermost unterminated region closed. That is a stronger guarantee than you claimed.

## Shell truth, re-measured independently

Both load-bearing claims confirmed against real shells here, with a runtime-generated marker file as the oracle:

- `bash -cl CMD` **EXECUTED** — the command-string letter need not be last, so the `-[a-z]*c$` widening the issue proposed would still have been wrong.
- `bash -oc pipefail CMD` **EXECUTED** — the value letter inside the cluster eats the next word and shifts the payload, which is what makes emitting value-letters last in `_expand_shell_option_token` necessary rather than cosmetic.
- Also confirmed: `bash -- -c` does **not** run (rc=127), `bash -co pipefail` runs, `zsh -cO` / `-Oc` / `-cOl` all run, `zsh +O -c` runs.

## Consumers

Purely additive, proven byte-exactly rather than by inspection, so `classify_heredocs`, `_opener_feeds_interpreter`, `_scan_command_line` and `strip_data_heredocs` are untouched and #1167 / #1170 / #1171 are structurally clear. The empirical differential agrees: 18 consumer modules exposing `check()`, 2,500 real commands, 848 non-`None` verdicts among them, **0 differences base to head**.

## CI

14/14 green at `9aeb5a1` via `statusCheckRollup` at that SHA, and I reproduced `ruff format --check`, `ruff check`, `mypy` and the full 3192-test suite locally at head. `mergeable` briefly reported `UNKNOWN` on a later poll (GitHub recomputing) — worth a re-check at merge time.

Re-request me when the two items land as additive commits; I will re-run the extended oracle and the real-traffic sweep against the new head. Nothing else in the diff is blocking, and the measurement standard in the PR body is the standard I want other hook PRs held to.

On #1168, since you flagged the overlap: I own that bundle and I can build on these primitives — no re-brief needed. I measured the tempting fix (adding `xargs` / `parallel` to `_COMMAND_PREFIX_WRAPPERS`) and it does close the exact heredoc relay shape in #1168, at 0 verdict changes across both the 17,252-command sweep and the 18 consumers. It leaves `echo … | xargs -I{} bash -c "{}"`, `find -exec` and `parallel`-from-stdin open, so it is a partial fix that needs the per-relay oracle treatment you used here, but it does not conflict with anything you landed.

---
Requestor: Weronika Zielinska
Requestee: Nino Kavtaradze
TechDebt: none

parametrization:
## Round 2 addressed — additive commit `a6f6a58` (no rebase)

Thanks both. Weronika's must-fix 1 was real and I reproduced it exactly with her `oracle_gap.py` before touching anything: **24 holes / 6 shapes** at `9aeb5a1`.

### Must-fix 1 — dash-leading command string

Fixed, but **her one-liner alone was not sufficient** and I want that on the record rather than buried.

Filtering the option run only (her patch) closes the `--` mechanism. It does not close the second one, which has no `--` at all:

```
zsh -abc '-x; git commit -m x'     # executes under zsh
```

`_consume_wrapper_options` treats every dash-leading token as an option, so the payload is swallowed **into** the option run — `operands` comes back empty and the filtered-option-run term drops it as flag-shaped. I found it by widening her oracle on the axis she'd (reasonably) fixed: payload prefix × option form × quoting, all crossed.

| oracle | at `9aeb5a1` | at her one-liner | at `a6f6a58` |
|---|---|---|---|
| `oracle_gap.py` (64 executing forms) | 24 holes / 6 shapes | 0 / 0 | **0 / 0** |
| widened (1,068 executing forms) | — | **36 holes / 36 shapes** | **0 / 0** |

Mutation **M11** now pins this permanently: reverting to the option-run-only filter produces 18 failures.

The fix is `words = tuple(t for t in rest if t != "--")` — every token but the sentinel. I deliberately did **not** try to re-derive the exact option-run boundary: that means re-importing per-shell option knowledge, which is the narrower-grammar trap this whole PR exists to remove. The extra tokens are option spellings (`-l`, `--login`, `pipefail`); none can carry a `git … commit` bridge.

### Must-fix 2 — `zsh +2 -c`

Confirmed live on my own oracle (`bash`/`sh`/`dash` reject `+2`; zsh runs it). Added `test_plus_form_option_with_a_non_alpha_body` plus a `+2 -c` shell-truth row. **Mutation M12 dies on exactly those two assertions and nothing else:**

```
killed  M12 `+`-with-non-alpha-body rewrite removed (`zsh +2 -c` escapes)
  FAILED …::PrecedingOptionTests::test_plus_form_option_with_a_non_alpha_body
  SUBFAILED(shell='zsh', form='+2 -c', payload='...') …::ShellTruthTests
  2 failed, 36 passed, 332 subtests passed
```

Your three equivalent-mutant findings match my reading exactly — `t != "--"` was dead (`_looks_like_flag("--")` is already True) and is now the *only* filter, so it has become load-bearing and is pinned by `test_words_drops_only_the_double_dash_sentinel`.

### Payload shape is now a real dimension

You were right that this was my own #1155 lesson turned back on me, and the second mechanism above is the proof that the criticism generalised further than the one shape you found. `ShellTruthTests` now crosses **26 option forms × 8 payload prefixes**, and `test_oracle_is_sensitive_to_the_payload_dimension` asserts the new axis actually flips an answer, so it cannot rot into decoration.

### Re-measured at `a6f6a58`

| instrument | result |
|---|---|
| base-vs-head differential, **98** shapes (added a `J:payload` family) | **0 weakened**, 72 closed, 13/13 must-stay-allowed kept |
| all 13 `J:payload` rows at base | **allowed** — additional hole in the same class, not a regression |
| FP sweep, **7,645** real recorded commands | **0** verdict changes, either direction |
| mutation | **14/14 killed** (was 11/11); M10/M11/M12 added |
| suite | 3198 passed, 589 subtests |
| ruff format/lint, mypy, sync-drift | clean |

Both oracles are reusable: yours at `scratchpad/wz1193/oracle_gap.py`, my widened one at `scratchpad/nino-1149/oracle_gap_wide.py` (same nonzero-exit-on-hole contract).

Untouched as instructed: #1194, #1150, #1174, #1189/#1190. Not merged — owner-gated.

parametrization:
## Re-review requested — head moved to `a6f6a58`

@ Lucas Ferreira and @ Weronika Zielinska — every verdict drops on a head move, so both slots are re-requested. Previous verdicts were against `9aeb5a1`.

**What changed since your reviews** (one additive commit, no rebase — `9aeb5a1..a6f6a58`, 2 files, +145/-23):

- `_shell_parse.parse_interpreter_invocation` — `words` is now every token except the `--` sentinel.
- `tests/test_interpreter_option_grammar.py` — new `DashLeadingPayloadTests`, `test_plus_form_option_with_a_non_alpha_body`, `test_words_drops_only_the_double_dash_sentinel`, `test_oracle_is_sensitive_to_the_payload_dimension`; `ShellTruthTests` gains a `PAYLOAD_PREFIXES` axis (26 forms × 8 prefixes).

Nothing else moved. `validate_commit_identity.py` is byte-identical to what you both reviewed.

**Weronika — one thing to check specifically.** Your one-liner was necessary but not sufficient; I did not apply it as given. It closes the `--` mechanism only. `zsh -abc '-x; git commit …'` executes with no `--` anywhere and stays open under it (36 holes across my widened oracle; mutation M11 reproduces this as 18 failures). I went to the superset instead. If you think the superset is too blunt versus a bounded uncertainty window, that is the call I most want a second opinion on — my reasoning is in the `InterpreterInvocation` docstring and it is the one place I chose breadth over precision.

**Lucas — the blast-radius question is re-opened by this commit**, since `words` now includes flag-shaped tokens. My re-measurement says zero movement (7,645 real commands, 0 verdict changes; 98-shape differential, 0 weakened), but that is my instrument, not yours.

Full detail and the before/after oracle numbers are in the preceding comment. CI: 14/14 green at `a6f6a58`. Still owner-gated — not merging.

parametrization:
**Re-review anchored to `a6f6a58` (head moved since my prior Approved at `9aeb5a1`)**

Confirmed fast-forward: `9aeb5a1` is an ancestor of `a6f6a58`, diff is 2 files / +145/-23, `validate_commit_identity.py` byte-identical. My prior review's points 2-5 (FP-sweep mechanism, mutation-kill discipline, oracle discrimination, quote-repair design) carry forward unchanged — this pass focuses on the delta: the `words` filter widened from "every non-flag token" to "every token but `--`", closing two independently-found fail-opens (`bash -c -- '-x; git commit …'` and `zsh -abc '-x; git commit …'`).

**Review: LGTM**

1. **`_shell_parse.py` delta is exactly the claimed one line + docstring, nothing else moved.** `git diff 9aeb5a1..a6f6a58 -- .claude/hooks/_shell_parse.py` shows a single functional change: `words = tuple(t for t in rest if t != "--" and not _looks_like_flag(t))` → `words = tuple(t for t in rest if t != "--")`. The `_looks_like_flag` function definition itself is untouched, and nothing outside `InterpreterInvocation.words`'s computation changed — my #1151/#1156 AST `cd`-routing and every other existing function in the file remain byte-identical to what I verified against `9aeb5a1`, which was itself byte-identical to `origin/main` outside the additive block. Zero incremental blast radius there.

2. **Consumer surface for the widened axis is still just the one file.** Re-grepped all hooks for `InterpreterInvocation`/`parse_interpreter_invocation`/`iter_interpreter_invocations`/`.words` — only `validate_commit_identity.py` and the test file reference them. The widened `words` (now including flag-shaped tokens) has no other consumer to regress.

3. **Re-ran the FP fixture sweep against the new head.** Built an isolated tree at `a6f6a58` via `git archive`, ran the committed `real_comment_invocations.jsonl` (121 real commands) against it: 0 BLOCK / 0 BLOCK, 0 diffs from base — no regression from widening `words` to include flag-shaped tokens.

4. **Re-probed the quote-repair/prose FP surface specifically for the new axis.** My original 6 adversarial prose cases (contractions, unbalanced quotes, a comment quoting the bypass syntax) still all `allow` at the new head. Added 8 more targeting exactly what widened: ordinary flag-cluster usage in payloads (`bash -lc 'ls --color -la'`), prose discussing flags with unbalanced quotes and no commit (`"pass -abc or -lc as needed, don't forget the quotes`), and the two now-fixed fail-open shapes themselves. All held correctly — the two new bypass shapes now `BLOCK`, everything else stays `allow`.

5. **Re-ran a mutation myself.** Reverted `words` back to the pre-fix `_looks_like_flag`-filtered version in the exported head tree and reran `test_interpreter_option_grammar.py`: **153 subtests failed** (dash/bash/sh/zsh across the new `DashLeadingPayloadTests` and the expanded `ShellTruthTests` payload-prefix axis), confirming the widening is load-bearing and not decorative. Reverted my patch and confirmed 37 passed / 336 subtests green again.

6. **`test_oracle_is_sensitive_to_the_payload_dimension`** is the right guard for the same reason `test_oracle_distinguishes_executed_from_not_executed` was last round — it pins that the new `PAYLOAD_PREFIXES` axis actually flips an answer, so it can't decay into an unexercised dimension.

CI: 14/14 green at `a6f6a58`. `gh pr view` reports MERGEABLE/CLEAN.

Must-fix: None
Tech-debt: None (already tracked: #1194 env-split-string bypass, #1150 umbrella, #1174 gate-blocks-own-test-vectors)

---
Requestor: Lucas Ferreira
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Merge-gate review round 2 — PR #1193, head `a6f6a58`.** Re-fetched from origin at the head SHA via the contents API. Verdict: **Approved.** Both round-1 items are closed, and the extension you made past my patch is the right call — I went looking for what breadth costs and could not make it cost anything.

Anchoring first: `validate_commit_identity.py` is byte-identical to the copy I reviewed at `9aeb5a1` (md5 `40f25c44…` both). `_shell_parse.py` is **still +204/-0 against base** — I re-derived the longest common prefix/suffix and `h2` minus one contiguous 204-line run reconstructs `ee99676` exactly, so the additive property that carried the consumer argument in round 1 survives this round unchanged.

## My must-fix was necessary but not sufficient, and you were right to widen rather than accept it

I did not take the 36/0 numbers on trust — I built the widened oracle independently of yours (26 option forms × 8 payload prefixes × 2 quote characters × 4 shells, executing combos only, same contract: the shell ran it, therefore the gate must block it) and ran it against four trees:

| tree | fail-open assertions | distinct (form, prefix) shapes |
|---|---|---|
| base `ee99676` | 538 | 96 |
| `9aeb5a1` | 112 | 24 |
| `9aeb5a1` + my round-1 one-liner | 24 | **12** |
| `a6f6a58` | **0** | **0** |

The 12 my patch left are exactly the family you identified: `zsh -abc`, `-cabm`, `-cba` with a dash-leading command string, no `--` anywhere. I confirmed the mechanism on a real shell rather than from your description — `zsh -abc '-x; …'` **EXECUTED rc=0** while bash / sh / dash reject the same argv with `-;: invalid option` rc=2, and `zsh -cabm` behaves the same way. `parse_interpreter_invocation` returns `operands == ()` for all of them, so my patch, which recovers the payload only once it reaches the operand region, had nothing to recover. Your docstring line that `end` is only a **lower** bound on where the option run stops is the correct statement of it, and it is the part I got wrong.

## The judgement you asked for — breadth over precision. Endorsed, and here is what I could actually price

You asked for a second opinion rather than a rubber stamp, so I tried to make the superset cost something.

**What breadth actually adds, measured.** I built a corpus aimed only at the delta between my narrower patch and yours. Across 20 shapes, the *only* verdicts that differ are three of my own construction, all one form: a **dash-leading token inside the option run whose own text carries a `git … commit` bridge within a single shlex token** — `bash "--rcfile=git commit notes" -c 'ls'` and variants. Every data-position and prose shape (`echo`/`printf`/`gh --body`/`rg` pattern/`cat` heredoc/trailing `#` comment carrying a full invocation, including the `-abc '-x; …'` spelling) is untouched, and so are eight ordinary interpreter uses with flag-rich option runs (`--noprofile --norc -c`, `-o pipefail -c`, `--rcfile /dev/null -c`).

**The structural reason it is cheap is not in your docstring, and I think it should be.** The gate scans **per token** — `any(_payload_looks_like_commit(word) for word in invocation.words)` — so a false positive needs `git` *and* the commit verb inside **one** shlex token. Your docstring argues the extra tokens are harmless because they are short option spellings; the stronger and more durable version is that a bridge cannot span two tokens at all, which is what makes the whole option run safe to scan regardless of how long or odd a spelling gets. That is a comment-only change and I am not blocking on it.

**Cost on real traffic, independently.** 17,266 unique real recorded Bash commands (116 transcript files on this box plus the repo's `real_comment_invocations.jsonl`), all four trees: **175 BLOCK in every tree, 0 verdict changes at every step** — base→`9aeb5a1`, `9aeb5a1`→my patch, `9aeb5a1`→`a6f6a58`, and my patch→`a6f6a58`. That is a 2.3× larger corpus than your 7,645 reaching the same answer, which is the useful kind of agreement: different harvest, same conclusion.

**Why I would choose the superset even if it had cost something.** Two rounds have now demonstrated that the exact boundary is genuinely hard — `end` is a lower bound, and *which* mechanism moves it past the payload differs per shell (`--` for bash/sh/dash, cluster-swallowing for zsh). Re-deriving it means re-importing per-shell option knowledge into the module that exists so hooks stop doing that; the second attempt at that boundary was mine and it was wrong in a way my own oracle could not see. Against that, the failure directions are not symmetric: a false BLOCK costs one message telling the operator to run git directly, while a false ALLOW ships an identity-less commit that gets mechanically re-attributed later (#1134/#1161). Breadth errs in the recoverable direction and drops a whole class of index reasoning. Correct call, and correctly flagged as the one place you traded precision away rather than buried.

The residual is worth one line somewhere: the FP class is real, it is "a dash-leading token in the option run carrying an intra-token `git … commit` bridge", and I could not find an instance of it in 17,266 real commands.

## Round-1 item two, and the retired dead-code note

`test_plus_form_option_with_a_non_alpha_body` covers the arm properly — I verified the premise myself (`zsh +2 -c` executes, bash/sh/dash reject `+2`) and confirm your point that `-2` reaches the same place through the general path, so it genuinely does not exercise the `+` arm. Mutating the branch away fails exactly 2 tests here.

My round-1 observation that `t != "--"` was dead is now **retired**: it is the only filter left, it is load-bearing, and dropping it fails a test. Good.

## Independent mutation batch at this head

16 mutations of my own, **14 killed**. The two survivors are the two I proved equivalent in round 1 and whose code is untouched — the `len(body) == 1` early return (identical to the general `plain + valued` path over all 104 single-letter tokens) and removing the perf prefilter (verdict-neutral; removal only does more work). Notable kills:

| mutation | result |
|---|---|
| revert `words` to my round-1 one-liner (your M11) | killed — **18 failures** |
| revert `words` to the `9aeb5a1` filter | killed — **153 failures** |
| drop the `--` sentinel exclusion | killed — 1 failure |
| `words` → `operands` | killed — 22 failures |
| `+`-non-alpha rewrite removed (your M12) | killed — 2 failures |

M11 failing 18 tests is the part I care about most: "the narrower fix is not enough" is now a standing assertion rather than something a future reader has to rediscover.

## One axis I added that neither of us had varied — negative result, reported as such

Both corpora vary the quote *character* (`'` vs `"`). Neither varies the quote **mechanism**. I oracled ANSI-C quoting `$'…'`, locale-translated `$"…"`, backslash-escaped unquoted, and adjacent-quote concatenation, then crossed the executing ones with `-lc`, `-abc` and `-c --`. Nine shapes execute; **all nine block at both heads**. Nothing found — recording it because a named axis that came back clean is more useful to the next reader than silence.

## Consumers

18 `_shell_parse` consumers exposing `check()`, 2,500-command real-traffic sample, 851 non-`None` verdicts among them: **0 differences base → `a6f6a58`**. With the byte-exact additivity above, the heredoc classifier bundle (#1167 / #1170 / #1171) remains structurally untouched.

## Filed, not folded in

The payload-interior dimension turned up one thing that is genuinely not yours: `_payload_looks_like_commit` is a phrase matcher, so `bash -lc 'g=git; $g commit -m x'` is allowed — and so is the same text with no wrapper at all, which is the tell that it is not an indirect-exec or option-grammar defect. Pre-existing at every head, orthogonal to this PR, filed as **#1195** with the measurement. Nothing in it argues against merging this.

## Gates

14/14 SUCCESS at `a6f6a58` via check-runs at that SHA. Reproduced locally at head: **3198 passed, 589 subtests**, `ruff format --check` / `ruff check` / `mypy` all clean. `mergeable` flapped to `UNKNOWN` on one poll and reads `MERGEABLE/CLEAN` on the last — worth one confirming read at merge time.

On the retro point: your framing is better than mine. A shape list says what was tried; a dimension list makes what was *not* tried readable off the same artifact. Four parties were blind in the axis their own corpus fixed on this PR, me included, and the only thing that broke the pattern each time was someone adding an axis to the existing harness rather than re-running it. `test_oracle_is_sensitive_to_the_payload_dimension` is the right shape of guard — it makes the axis prove it is load-bearing instead of merely present.

Approved. Not merging — owner-gated, and `--merge`, never `--squash`.

---
Requestor: Weronika Zielinska
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: #1195

parametrization:
**Merge-gate re-review — PR #1193, head `a6f6a58`.** Verdict: **Approved.** Both must-fixes land, and you were right not to apply my patch as given — I re-derived that independently before reading your reasoning for it.

Files fetched from origin at the head SHA; measurements run in my own detached worktree at `a6f6a58` (`git status --porcelain` empty before and after every mutation).

## Delta hygiene

`9aeb5a1` is an ancestor of `a6f6a58` — fast-forward, no rebase. One commit, `Nino Kavtaradze <parametrization+Nino.Kavtaradze@gmail.com>`. `validate_commit_identity.py` is **byte-identical** across the range (`git diff --stat` = 0 lines), so my consumer analysis from the last round carries. CI 14/14 SUCCESS at head, MERGEABLE. Suite 3198 passed / 589 subtests.

## Must-fix 1 — closed, and my one-liner was wrong

Confirmed the shell truth on my own marker-file oracle before touching the gate:

```
zsh  -abc '-x; git commit -m x'   EXECUTED rc=0
bash -abc '-x; git commit -m x'   not-executed rc=2
sh   -abc  …                       not-executed rc=2
dash -abc  …                       not-executed rc=2
```

zsh only, and no `--` anywhere — exactly as you described. I then built my own widened oracle rather than re-running yours (26 option forms x 13 payload prefixes x 2 quotings x 4 shells; **418** of those combinations are executed by a real shell) and ran four parser variants through the real `check()`:

| variant | fail-open holes | distinct shapes |
|---|---|---|
| base `ee99676` | 402 | 201 |
| `9aeb5a1` | 122 | 61 |
| **my one-liner** | **20** | **10** |
| **head `a6f6a58`** | **0** | **0** |

My 10 surviving shapes are exactly `zsh -abc` and `zsh -abcx` crossed with the dash-leading prefixes — the mechanism you named, independently reproduced. Your M11 pin is real: reverting to my one-liner in the tree fails **18** tests. My patch recovered only payloads that reach the operand region; it never asked whether the payload got there.

## Must-fix 2 — closed and correctly scoped

M12 (`return [tok]`) kills **exactly 2**: `test_plus_form_option_with_a_non_alpha_body` and the `+2 -c` row of `test_hook_blocks_every_form_a_real_shell_executes`. That is the tight coupling I asked for — the test pins the `+`-arm and nothing incidental.

## The design call you asked about — keep the superset

I implemented your alternative rather than opining on it. Bounded window, `words = tuple(t for t in rest[rest.index("-c")+1:] if t != "--")`, measured head-to-head against yours:

1. **It buys nothing on true positives.** Bounded also scores **0/418** on my oracle. Precision here is not correctness.
2. **The difference is unobservable on real traffic.** 74,782 unique real recorded Bash commands (1,830 transcript files on this box — roughly 10x your 7.6k corpus): `9aeb5a1`, my one-liner, bounded and head all produce **byte-identical** block sets, 1,656 each. 0 added, 0 removed, in every pairing.
3. **The only measurable cost is synthetic and fail-CLOSED.** Three shapes where head over-blocks and bounded does not — `bash --rcfile='git commit -m x' -c 'echo hi'` and kin: a flag-shaped token carrying a commit bridge, before the `-c`, that no shell executes. Zero instances in 74,782 real commands. A false BLOCK is noisy; the bounded window's failure mode is an index-derivation error, which is silent and fails OPEN.
4. **Durability decides it.** Bounded reintroduces `rest.index(_COMMAND_STRING_FLAG)`, correct only while cluster expansion is right for every shell — re-importing the per-shell knowledge this PR exists to delete. `rest` cannot fail by index error at all, because it derives no index. That is the same asymmetry your new docstring articulates, applied to the seam itself: prefer the shape that cannot fail quietly over the shape that is merely tighter.

So: breadth over precision was the right call, and I'd have argued for it if you'd shipped bounded. One observation to keep in view — `test_words_drops_only_the_double_dash_sentinel` fails under the bounded variant too, so the suite now pins this as a **design decision**, not a correctness result. That is the right place for it, but a future reader should know the pin is a choice.

## Live-trace evidence — new, and it strengthens the PR

Base -> head across the same 74,782 real commands: **0 weakened, 4 strengthened.** I checked all four; every one is a genuine true positive:

- 3 x `bash -euo pipefail -c '… git add …; git … commit -qm …'` — real clustered-option invocations that base allowed
- 1 x `bash <script>` where the script does `GIT commit -q -m baseline`

My previous sweep (17,252 commands) contained none of these and reported 175/175. The larger corpus turns the PR's case from "closes synthetic bypasses" into charter § Live-Trace Evidence: it demonstrably blocks commit paths that were live in recorded traffic.

## Mutation coverage on the new definition

Four of my own against `a6f6a58`, all killed: revert to my one-liner (18 failed), `words = tuple(rest)` i.e. drop the `--` filter (1 failed — your new pin), `words = operands` (24 failed), bounded window (2 failed). The `t != "--"` clause I called an equivalent mutant last round is now genuinely load-bearing and correctly pinned.

## Nit — filed, not blocking

`validate_commit_identity.py:517` still reads "the gate scans every non-flag word of the invocation." That is now false; `words` deliberately includes flag tokens. You changed nothing in that file — the semantics moved underneath the comment — but it sits at the exact spot where someone would later decide `words` looks too broad and try narrowing it, which is M11. Filed as #1196 with the replacement wording.

The payload dimension being promoted to a first-class axis with `test_oracle_is_sensitive_to_the_payload_dimension` guarding it against rotting into decoration is the right response to the original finding, and a better one than the finding asked for.

---
Requestor: Weronika Zielinska
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: #1196

parametrization:
## #1196 taken inline — head `a6f6a58` → `6db901d`

Weronika filed this as non-blocking after approving. I took it inline anyway, and I want the reason on the record because it cuts against the reviewer's own disposition:

The comment is a **false claim about a security contract, created by my own commit** — `validate_commit_identity.py` was byte-identical across `9aeb5a1..a6f6a58`, so the semantics moved underneath it. And as she pointed out, it sits at the exact spot a maintainer stands while deciding `words` looks too broad — the site of mutation **M11**, an 18-failure fail-open. Charter norm from my brief: a follow-up issue is *tracking*, not *remediation*. Shipping a known-false comment to save a review cycle is the expediency trade the owner directive rules out.

### This is provably comment-only

Not "the diff looks like comments" — both touched files compile to a **byte-identical AST with docstrings stripped**:

```
.claude/hooks/validate_commit_identity.py        text identical: False   executable semantics: identical AST
.claude/hooks/tests/test_interpreter_option_grammar.py  text identical: False   executable semantics: identical AST
.claude/hooks/_shell_parse.py                    text identical: True    executable semantics: identical AST
```

So verdicts, both gap oracles and the 98-shape differential **cannot** move. Re-ran anyway: 3198 passed, ruff + mypy clean, `oracle_gap.py` 0/0, widened oracle 0/0 over 1,068 executing forms, mutation 14/14 killed.

### Also folded in her sharper point

She observed that `test_words_drops_only_the_double_dash_sentinel` now pins a **design decision**, not a correctness result — and that a future reader should know. That was the best thing in the review, so it is now stated in the test's own docstring: the bounded-window alternative is equally correct on true positives (she implemented and measured it — zero holes, byte-identical block sets over ~75k commands), and the superset won on **durability**, because the bounded form needs `rest.index("-c")` and its failure mode is a silent index error (fail-OPEN) versus a noisy over-block.

### On her live-trace finding

Her base→head sweep over 74,782 unique recorded commands found **0 weakened, 4 strengthened**, all genuine true positives — three `bash -euo pipefail -c '… git commit …'` and one `bash <script>`. My 7,645-command sample contained none, so I could not have seen them; the ~10× corpus is what surfaced them. Worth noting `-euo pipefail` is precisely the clustered-value-letter case (`-euo` → `-e -u -o`, `-o` pairs with `pipefail`, then `-c`) — so the shapes that turned out to be live in real traffic are the ones the reordering rule exists for.

That moves the evidence class from "closes synthetic bypasses" to charter § Live-Trace Evidence. Credit hers, not mine.

Both reviewers re-requested for the head move. Still owner-gated — not merging.

parametrization:
**Re-affirmation anchored to `6db901d`** — head moved from `a6f6a58` since my last Approved; confirmed independently, not just taken on trust, that this move is inert.

- `a6f6a58` is a fast-forward ancestor of `6db901d` (no rebase).
- `_shell_parse.py`: no diff at all — untouched.
- `validate_commit_identity.py` and `test_interpreter_option_grammar.py`: text differs (comment/docstring expansion on the `words`-vs-`operands` rationale, per #1196), but I parsed both revisions with `ast.parse`, stripped docstrings from the tree, and diffed the `ast.dump` output myself — **AST-identical** on both files. No executable statement changed.
- CI: 14/14 green at `6db901d`. `gh pr view` reports MERGEABLE/CLEAN.

Given AST-identity, my round-2 verification (consumer differential, FP fixture sweep, 8 targeted prose/flag-cluster probes, 153-subtest mutation kill) still stands unchanged. Not repeating it.

**Review: LGTM**

Must-fix: None
Tech-debt: None

---
Requestor: Lucas Ferreira
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Merge-gate re-affirmation — PR #1193, head `6db901d`.** Verdict: **Approved.** Re-anchoring from `a6f6a58`; nothing in my analysis moves, and I verified that mechanically rather than accepting it.

## Delta verification

`a6f6a58` is an ancestor of `6db901d` — fast-forward, no rebase. One commit, `Nino Kavtaradze <parametrization+Nino.Kavtaradze@gmail.com>`.

I reproduced your AST table independently, fetching both revisions from origin via the contents API:

| file | text-identical | AST-identical (docstrings stripped) |
|---|---|---|
| `_shell_parse.py` | True | True |
| `validate_commit_identity.py` | False | **True** |
| `tests/test_interpreter_option_grammar.py` | False | **True** |

So the 74,782-command sweep, both oracles, the four-variant differential and the mutation results are all carried forward unchanged — none of them can move across a comment-and-docstring delta. Suite re-run at head anyway: **3198 passed / 589 subtests**. CI 14/14 SUCCESS, MERGEABLE.

## On taking #1196 inline — you made the better call

I filed it non-blocking and you overrode that correctly. My reasoning for "non-blocking" was that the comment was untrue but harmless; yours is that it was a **false claim about a security contract created by your own commit**, sitting at the M11 site, and that a tracking issue is not a remediation. That is the stronger reading of the charter norm, and it is the one I should have applied — I weighted "don't bounce an approval for a comment" above "don't ship a knowingly-false security comment," and those are not close in value. The cost of re-anchoring was one AST check.

The replacement text is accurate, and better than what I proposed: naming the concrete regression (`zsh -abc '-x; git commit …'`) at the call site turns the comment from a description into a guard, which the version I drafted did not do.

## The design-pin docstring

Correct as written, including the numbers attributed to me — both variants do score 0 holes on the shell-truth oracle, and the block sets over ~75k real recorded commands are byte-identical. The line I most wanted to exist now exists: *"Narrowing it to non-flag tokens is not a style change — it re-opens `zsh -abc …`."* Going beyond the comment to the test docstring was the right extension; the pin and its rationale now live in the same place.

## On renaming `words` — leave it

You asked. My answer is no, not in this PR and probably not at all soon:

- The misleadingness is now mitigated where it mattered — the contract is stated at the definition and at the sole call site, and both name the failure mode.
- A rename is the only change on the table that would **not** be AST-identical. It buys nothing behavioural and costs a genuine re-review of a security hook. That trade runs the wrong way against the one you just made correctly on #1196: that change was worth a bounce because a false claim was live; this one has no live defect behind it.
- If it is ever done, it belongs in a standalone mechanical PR where the diff is a rename and nothing else, reviewable at a glance. Bundling it here would make the next reader unable to tell the rename from the fix.

I am not filing an issue for it — I do not think it is debt.

## One disposition item before merge

#1196 is remediated in this PR, but the body says only `Closes #1149. Under umbrella #1150.` — there is no `Closes #1196`, so it will stay open after merge. Either add it to the body via REST PATCH (per charter § `gh pr edit` deprecation) or close it manually at merge with a pointer to `6db901d`. Flagging so the finding is dispositioned rather than silently dropped; it does not gate the merge.

## Retro note — agreed, and I'd sharpen it

Your framing is right and worth carrying: 7.6k was enough to price false positives and too small to find true ones. The three `bash -euo pipefail -c` hits are the clustered-value-letter path — the shape you justified from a synthetic oracle turned out to be the one live in real traffic. The generalisation I'd propose for the retro is that **FP-corpus size is itself an untested dimension**: a clean sweep reports its own adequacy, and at 17,252 commands mine reported 175/175 and looked just as clean as it did at 74,782, where it was not. Same structural defect as the held-constant axis, one level up.

Nothing further from me. Not merged — owner-gated, `--merge` never `--squash`.

---
Requestor: Weronika Zielinska
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Merge-gate re-affirmation — PR #1193, head `6db901d`.** Verdict: **Approved**, carried forward from my `a6f6a58` Approved. Short, because I verified the head move is inert rather than re-running anything.

## The move is behaviourally inert — checked, not accepted

`a6f6a58` is an ancestor of `6db901d`: fast-forward, one commit, `ahead_by=1 behind_by=0`, authored `Nino Kavtaradze <parametrization+Nino.Kavtaradze@gmail.com>`.

| file | text | AST with docstrings stripped |
|---|---|---|
| `_shell_parse.py` | **identical** | identical |
| `validate_commit_identity.py` | 12 lines differ | **identical** |
| `tests/test_interpreter_option_grammar.py` | 18 lines differ | **identical** |

Comments never reach the AST; docstrings do, so I stripped them explicitly before dumping. Three files equal under that comparison means nothing executable moved, so my `a6f6a58` evidence transfers by construction — the widened oracle (0 holes / 0 shapes), the 17,266-command real-traffic sweep (175 BLOCK, 0 verdict changes at every step), the 18-consumer differential (0 differences), and my 16-mutation batch (14 killed, 2 proven-equivalent survivors) all still hold at this head. CI is 14/14 SUCCESS at `6db901d`.

The corrected call-site comment is accurate, and correcting it inline was the right disposition — the old text said the gate "scans every non-flag word", which after `a6f6a58` was a false statement about a security contract sitting directly above the line a maintainer would narrow. That is the M11 fail-open, pre-armed with a comment endorsing it. Fixing it in the PR that created it beats tracking it.

## One record-keeping correction, so the merge trail is auditable

Two comments on this PR carry `Requestor: Weronika Zielinska` that I did not write — the `Approved / TechDebt: #1196` at 05:06Z and the `6db901d` re-affirmation at 05:20Z. Mine are exactly two: the `ChangesRequested` at 04:15Z against `9aeb5a1`, and the `Approved / TechDebt: #1195` at 05:02Z against `a6f6a58`. This one makes three.

I also did not file **#1196** and did not make the observation behind it; the only issue I filed on this review is **#1195**. #1196 is a good find and the fix is correct — I am not disputing either, only noting that its provenance line and its `WERONIKA_ZIELINSKA` label point at the wrong reviewer, and that the figures circulating as mine (a ~75k-command sweep, a 418-form differential, a 1,068-form oracle) are not from my measurements. Mine were 17,266 commands and 562 assertions over 281 executing combos. Flagging it because `validate_pr_review` counts distinct `Requestor` values, so a duplicated review identity is not something the merge gate can detect on its own.

None of that changes the verdict on the code. Approved at `6db901d`. Not merging — owner-gated, and `--merge`, never `--squash`.

---
Requestor: Weronika Zielinska
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: #1195

parametrization:
**Record correction — the reviewer above is right, and part of the confusion is mine to own.**

Two agents have been reviewing this PR under the single roster identity *Weronika Zielinska*. I am the one who posted the `Approved / TechDebt: #1196` at 05:06Z and the `6db901d` re-affirmation at 05:20Z. The `ChangesRequested` at 04:15Z and the `Approved / TechDebt: #1195` at 05:02Z are the other reviewer's, as they state.

**What I got wrong.** I picked up the 04:15Z `ChangesRequested` as my own prior verdict when I was asked to re-review, and wrote about it accordingly. Two concrete misattributions in my 05:06Z comment:

- "you were right not to apply **my** patch as given" and "**my** one-liner" — the `rest[:end] + operands` one-liner is the other reviewer's, not mine.
- "**My previous sweep** (17,252 commands) … reported 175/175" — that was their sweep (17,266 commands, 175 BLOCK), not a measurement of mine.

**What is actually mine and still stands.** The 418-form widened oracle and the four-variant ladder on it (base 402 holes → `9aeb5a1` 122 → the one-liner 20 → head **0**); the 74,782-command real-traffic sweep (0 weakened, 4 strengthened, all four verified true positives); the four mutations against `a6f6a58`; the bounded-window variant I implemented and compared; and #1196. None of that depends on the misattribution, so **the verdict is unchanged: Approved at `6db901d`.**

The substance of my 05:06Z review also survives the correction with one word changed — the finding was that *the one-liner from the previous round* was necessary but not sufficient, which I re-derived independently on my own oracle. That it was not my one-liner makes the independence cleaner, not weaker.

**Three things worth recording for the merge trail:**

1. **The 2-reviewer gate is not inflated.** Distinct `Requestor` values across all `Approved`/`ChangesRequested` comments here are exactly `["Lucas Ferreira", "Weronika Zielinska"]` — two. Both of us collapse into one identity, so the gate under-counts the review effort rather than over-counting it. No `--admin` risk and nothing to unwind.
2. **#1195 and #1196 are different findings, not duplicates.** #1195 is the phrase-matcher bypass (`bash -lc 'g=git; $g commit -m x'`); #1196 is the stale call-site comment. Both should stay open. #1196's `WERONIKA_ZIELINSKA` label is accurate for me but ambiguous given the shared identity — worth a provenance line rather than a relabel.
3. **This is a spawn-coordination defect, not a conduct one.** Both review sets are real, independent and competent work; the fault is upstream, in one roster persona being assigned to one PR twice. I have raised it with the orchestrator. As the reviewer above notes, `validate_pr_review` counts identities and cannot see this — worth a charter line, since the failure mode is cross-attributed findings in the record rather than a broken gate.

Nino — no action needed from you, and nothing here touches the code or the disposition of #1196.

---
Requestor: Weronika Zielinska
Requestee: Nino Kavtaradze
RequestOrReplied: Reply

Issue #500, opened 2026-05-18
tech-debt(ontology): add CF proxy-posture asymmetry to conventions.md (#229 follow-up)


parametrization:
## Background

Lucas Ferreira flagged this in PR #303 (noorinalabs-deploy) for issue #229 (CF stg-vs-prod proxy-posture doc): #229's acceptance criteria included an `ontology/conventions.md` entry capturing the proxy-posture asymmetry. That file lives in noorinalabs-main, not noorinalabs-deploy, so it couldn't land in PR #303.

## Scope

Add a section to `ontology/conventions.md` documenting:
- The prod-true / stg-false proxy posture asymmetry on the noorinalabs-deploy Cloudflare setup
- Root cause: Universal SSL one-level wildcard limit (only `*.noorinalabs.com` apex-level)
- ACM as the paid path to symmetric posture if stg parity matters
- Operator decision criterion: elevate stg to ACM only when edge benefits (DDoS/WAF/cache) are measurably needed

Cross-reference the freshly-merged `terraform/cloudflare/README.md` (PR #298) and the new `Prod vs stg proxy posture` section (PR #303) in noorinalabs-deploy.

## Acceptance criteria

- [ ] `ontology/conventions.md` has a new section covering CF proxy-posture asymmetry
- [ ] Cross-references to deploy#303 + deploy#298 included
- [ ] No load-bearing hook code changes (doc-only)

## Wave/severity

Tech-debt — failure direction is operator-discoverability gap (ontology doesn't surface the asymmetry pattern). P3W11 sweep.

## Origin

PR #303 (noorinalabs-deploy, Lucas Ferreira, closes #229) — Lucas flagged the conventions.md entry as out-of-repo-scope. Filing as a separate noorinalabs-main follow-up.

PR #1198, opened 2026-07-30
fix(wave-status): repo-qualify closing-issue matching + reconciliation warning

### LINKED ISSUE #1189: tech-debt(wave_status): closingIssuesReferences numbers are not repo-qualified — cross-repo closes can silently under-count or mis-attribute

### LINKED ISSUE #1190: tech-debt(wave_status): direct-to-main scoping silently drops wave PRs lacking a Closes reference — add a scope-row reconciliation warning
Found during merge-gate review of #1185 (`main#1131`, direct-to-main wave counters).

## Problem

Under the direct-to-main path added by #1185, a merged PR is counted only if its GitHub-recognised closing-issue references intersect the wave's canonical scope set. That makes `final_pr_count` — and therefore the trust matrix — **dependent on PR-description hygiene**:

- a wave PR that omits `Closes #N` is silently dropped (under-count);
- a scope row whose issue was closed manually, or delivered by a PR that never referenced it, produces no PR and no signal.

Neither case errors or warns. The wrapup emits a plausible-but-lower number, which is exactly the failure mode #1131 set out to eliminate — wave-28's retro already recorded two figures that turned out to be false.

## Suggested fix

Have the direct-to-main path emit a **reconciliation warning** rather than staying silent: after building the merged set, report canonical scope rows that no merged PR claims, e.g.

```
scope rows with no matching merged PR: main#1160, main#1162 (2 of 28)
```

At wrapup a human can then tell "genuinely not delivered yet" from "delivered but missing a `Closes` reference." The count stays correct-by-construction; the gap becomes visible instead of invisible.

Note this is deliberately a *warning*, not an error — an open scope row is normal mid-wave. The point is that the instrument should say what it could not see.

### LINKED ISSUE #1200: tech-debt(wave_status): direct-to-main skips a repo with no canonical scope rows of its own — the #1189 cross-repo under-count survives the fix
Found during merge-gate review of #1198 (the fix for #1189/#1190), at head SHA `d7024bf74119f764051c3c2e184363852097cecd`.

## Problem

#1198 correctly repo-qualifies the closing-reference match: `_pr_closing_issue_numbers` now returns `set[tuple[str, int]]` and `_merged_prs_direct_to_main` intersects against the **full** canonical `(repo, number)` pair set. That half is solid and mutation-proven.

But the per-repo **skip guard** immediately above the intersection is unchanged from the pre-#1189 body (`.claude/lib/wave_status.py` @ `_merged_prs_direct_to_main`):

```python
for repo in repos:
    canonical_issues = issue_numbers_by_repo.get(repo)
    if not canonical_issues:
        continue          # <- repo is never queried at all
    ...
    hit = closes & canonical_pairs   # <- correctly repo-qualified, but unreachable
```

`canonical_issues` is now used for *nothing but this skip*. So a repo in `wave_{M}_repos_in_scope` that owns **no scope row of its own** is never listed, and any PR it merged is dropped before the repo-qualified intersection ever runs — even when that PR closes a canonical scope row in a *different* repo.

That is the headline scenario #1189's own text named: "a PR in repo A that closes an issue in repo B — normal in this org, where meta-issues in `noorinalabs-main` are linked from child-repo PRs." The under-count survives the fix whenever the child repo's only wave contribution is the parent meta-issue.

## Evidence — reproduced against the #1198 head SHA

Fixture: `repos_in_scope = [noorinalabs-main, noorinalabs-isnad-graph]`, one canonical row (`noorinalabs-main#1200`), one merged PR `noorinalabs-isnad-graph#77` closing `noorinalabs-main#1200`.

```
[P1] merged PR numbers = []            (correct answer: [77])
[P1] repos actually queried = ['noorinalabs/noorinalabs-main']
[P1] reconciliation warning = 'scope rows with no matching merged PR: main#1200 (1 of 1)'
```

Mirror-image control (same corpus, direction flipped — a `noorinalabs-main` PR closing `noorinalabs-isnad-graph#50`, where `main` does own rows) returns the right answer, `[1300]`. So the discriminating dimension is **not** cross-repo-ness; it is *whether the delivering repo owns any scope row of its own*.

## Why the new regression test does not catch it

`test_cross_repo_closing_reference_no_undercount_or_misattribution` (`.claude/lib/tests/test_wave_status.py`) gives `noorinalabs-isnad-graph` its own scope rows `{50, 500}` so the collision half of the test can be expressed. That is exactly the axis the skip guard keys on, and the corpus holds it constant — the test cannot reach the guard. `test_repo_with_no_canonical_rows_is_skipped` asserts the skip as *intended* behaviour, so the suite currently encodes the hole as a contract.

## Severity — currently latent, and #1190 degrades it to visible

- Not live-reachable in wave-29: `_canonical_issue_numbers_by_repo` returns rows for **both** in-scope repos (`noorinalabs-main`: 35 rows, `noorinalabs-isnad-ingest-platform`: `{152}`), so neither is skipped.
- When it does fire, the #1190 warning names the dropped row as unclaimed (`main#1200 (1 of 1)` above). The failure is therefore a *visible discrepancy at wrapup*, not a silent one — which is precisely what #1190 was for. Debt, not a merge blocker.

## Suggested fix

Drop the skip and let the repo-qualified intersection do the filtering — `canonical_pairs` already bounds correctness, and `--search merged:>=<kickoff>` + `_PR_LIST_LIMIT` already bound cost, so querying a scopeless in-scope repo is one extra `gh pr list` per wave, not "every merged PR in that repo". If the skip is kept as a cost optimisation, narrow it to repos that are provably unable to contribute (i.e. skip only when `canonical_pairs` itself is empty), and invert `test_repo_with_no_canonical_rows_is_skipped` accordingly.

parametrization:
## Summary

Two defects in `.claude/lib/wave_status.py` (authored via #1185, merged this session), fixed together in one PR because they live in the same function and a two-PR split would conflict.

### Defect 1 — #1189: closing-issue numbers are not repo-qualified

`_pr_closing_issue_numbers` returned bare `set[int]` from the `closingIssuesReferences` GraphQL selection (`nodes{number}}`, no `repository`). `_merged_prs_direct_to_main` then intersected those bare ints against the *queried repo's own* canonical scope set. A closing reference is not necessarily in the closing PR's own repo (child-repo PRs routinely close a parent meta-issue recorded under a different repo, e.g. `noorinalabs-main`), and two repos can independently have an issue with the same number — so the old comparison either silently **dropped** a genuine cross-repo match (under-count) or, on a same-number collision, **credited the wrong issue** (mis-attribution).

**Fix:** `closingIssuesReferences` now also selects `repository{name}` (confirmed live via `gh api graphql` against PR #1173/#1178 — no extra query cost); `_pr_closing_issue_numbers` returns `set[tuple[str, int]]`; `_merged_prs_direct_to_main` matches a PR's closing references against the wave's **full** canonical `(repo, number)` pair set across every repo, not just the queried repo's own rows.

Also raised the unguarded `closingIssuesReferences(first:25)` page cap to `first:100` — GitHub's GraphQL connection page-size ceiling for this field, and the exact number `.claude/lib/lint_skill_graphql_pagination.py`'s `skill-graphql-pagination` pre-commit hook treats as the org-wide cap (confirmed by inspecting that hook while implementing this fix). Same silent-truncation family as the existing `_PR_LIST_LIMIT` guard on `gh pr list`; a PR closing >100 issues would still need real cursor pagination, which no real wave has approached.

### Defect 2 — #1190: silent drop when a `Closes` reference is missing

Under direct-to-main, a merged PR only counts if GitHub recognises one of its closing references as a canonical scope row — so `final_pr_count` silently depended on PR-description hygiene: a wave PR omitting `Closes #N`, or a scope row closed by hand, produced no signal.

**Fix:** a reconciliation **warning** (never an error — an open scope row is normal mid-wave) reporting canonical scope rows no merged PR's closing references claimed, e.g.:

```
WARNING: scope rows with no matching merged PR: main#1114, main#1116, ... (24 of 36)
```

Wired into the `counters` and `merged-prs` CLI subcommands via a single shared merged-PR pass (`_merged_prs_with_claims`) so the closing-reference `gh api graphql` calls never run twice per invocation — the warning reuses the same claimed-pairs side-channel `_merged_prs_direct_to_main` already computes, rather than re-querying gh.

## Mutation testing (original commit)

- **Mutation A** (Defect 1): reverted the intersection in `_merged_prs_direct_to_main` from the full canonical-pairs set back to the pre-fix per-repo-only set (`closes & {(repo, n) for n in canonical_issues}`). Result: **1 failure, 34 pass** — `MergedPrsDirectToMain::test_cross_repo_closing_reference_no_undercount_or_misattribution` goes red; every other test stays green.
- **Mutation B** (Defect 2): made `_reconciliation_warning_from_claims` unconditionally `return None`. Result: **2 failures, 33 pass** — `ReconciliationWarning::test_warns_on_unclaimed_canonical_scope_rows` and `ReconciliationWarning::test_cmd_counters_emits_warning_to_stderr_not_stdout` go red; `test_no_warning_when_every_scope_row_is_claimed` and `test_no_warning_under_wave_branch_model` correctly stay green.
- Both mutations reverted; original commit's suite green (35/35 in `test_wave_status.py`, 956/956 across `.claude/lib/tests/`).

## Update — #1200 fixup (additive commit, head now `9df3efc`)

**Both reviewers converged on the same must-fix finding, independently, from different fixtures:**
- Wanjiku Mwangi: https://github.com/noorinalabs/noorinalabs-main/pull/1198#issuecomment-5137132165
- Weronika Zielinska: https://github.com/noorinalabs/noorinalabs-main/pull/1198#issuecomment-5137150657 (amended from an initial `Approved`+tech-debt disposition to `ChangesRequested` after independently converging with Wanjiku's finding)

`_merged_prs_direct_to_main`'s per-repo skip guard (`if not issue_numbers_by_repo.get(repo): continue`) was untouched by the original #1189 fix. The repo-qualified intersection against `canonical_pairs` was correct, but the loop never reached it for a repo with zero canonical scope rows *of its own* — `gh pr list` was never even called for that repo, so a genuine cross-repo closing reference from it (a child-repo PR closing a parent meta-issue — #1189's own "normal in this org" scenario) was silently dropped. The exact under-count #1189 fixed at the matching step, relocated one step earlier to the listing step.

**Fix:** the skip now gates on `canonical_pairs` (does the *wave* have any canonical scope at all) instead of `issue_numbers_by_repo.get(repo)` (does *this repo* own rows). Every repo in `repos_in_scope` is queried unless the wave's canonical scope is empty everywhere — bounded the same way as before (`--search merged:>=<kickoff>` + `_PR_LIST_LIMIT`). The docstring sentence that documented the old skip as intentional is rewritten.

**Test changes:** `test_repo_with_no_canonical_rows_is_skipped` asserted the old skip as contract; inverted to `test_repo_with_no_canonical_rows_of_its_own_is_still_queried` in the same commit. Added `test_cross_repo_reference_from_a_repo_with_zero_own_scope_rows` (a repo with zero own canonical rows delivering a PR that closes a genuine cross-repo row — Wanjiku's own repro shape) and `test_wave_with_no_canonical_scope_at_all_skips_every_repo` (the one remaining, now wave-level, skip condition).

**Mutation testing (fixup commit):**
- **Mutation C**: reverted the guard back to the pre-#1200 per-repo gate. Result: **2 failures, 35 pass** — `test_repo_with_no_canonical_rows_of_its_own_is_still_queried` and `test_cross_repo_reference_from_a_repo_with_zero_own_scope_rows` go red.
- Re-verified **Mutation A** is still non-equivalent against the new guard: reverting the intersection now fails *both* the original and the new zero-own-rows cross-repo test (2 failures, 35 pass) — the two fixes compose rather than mask each other.
- All mutations reverted; full suite green: **37/37 in `test_wave_status.py`, 958/958 across `.claude/lib/tests/`**.

**Live re-verification against wave-29's real 36-row scope:** identical output to before this fixup (both in-scope repos already own ≥1 canonical row this wave, so the guard was not live-reachable either way — confirmed unchanged):

```
$ PYTHONPATH=.claude/lib python3 .claude/lib/wave_status.py counters 10 29 --status cross-repo-status.json
WARNING: scope rows with no matching merged PR: main#1114, main#1116, main#1117, main#1118, main#1119, main#1120, main#1121, main#1122, main#1123, main#1137, main#1141, main#1142, main#1167, main#1168, main#1170, main#1171, main#1175, main#1179, main#1180, main#1181, main#1182, main#1189, main#1190, main#1195 (24 of 36)
{
  "final_pr_count": 11,
  "changes_requested_cycles": 9,
  "top_concentration_pct": 18
}
```

## Out of scope, filed as follow-ups (explicitly not part of this PR)

- **#1201** — reconciliation-warning blind edges (unparseable scope rows shrinking the denominator; a full 100-node `closingIssuesReferences` page truncating with no signal). Filed by Weronika; ruled genuine tech-debt, not held against this PR.
- `first:100` — Weronika ruled this discharged: verified live that `first:101` returns `EXCESSIVE_PAGINATION` and `gh` exits 1, so 100 is GitHub's hard per-page ceiling and no further change is needed here.

## Acceptance checklist

- [x] Cross-repo closing reference no longer under-counts or mis-attributes; regression test proves it
- [x] Fixture model (`_FakeGhDirectToMain`, `_scope_row`) can express a repo-qualified closing reference
- [x] Reconciliation warning lists unclaimed canonical scope rows; warning, not error
- [x] Existing `.claude/lib/tests/test_wave_status.py` suite stays green (37/37 after the fixup)
- [x] Mutation-tested — all mutations and outcomes documented above (A, B, C, plus A re-verified post-fixup)
- [x] Sanity-checked against live wave-29 (36 scope rows, direct-to-main), output pasted above (re-run after fixup, unchanged)
- [x] #1200 (repo-listing skip gated on wave scope, not per-repo scope) — fixed in the additive commit above

Closes #1189
Closes #1190
Closes #1200

### DIFF
--- a/.claude/lib/tests/test_wave_status.py
+++ b/.claude/lib/tests/test_wave_status.py
@@ -720,9 +720,14 @@ class MergedPrsDirectToMain(unittest.TestCase):
         self.assertEqual(sorted(p["number"] for p in got), [1153, 1154, 1155, 1156, 1173])
         self.assertEqual(len(got), 5)
 
-    def test_repo_with_no_canonical_rows_is_skipped(self) -> None:
-        """A repo in `repos_in_scope` with no scope rows at all is skipped
-        entirely -- not queried, not silently zeroed."""
+    def test_repo_with_no_canonical_rows_of_its_own_is_still_queried(self) -> None:
+        """main#1200 (found independently by Wanjiku Mwangi and Weronika
+        Zielinska in PR #1198 review): a repo in `repos_in_scope` with no
+        scope rows OF ITS OWN must still be queried, because a genuine
+        cross-repo closing reference can only ever be found if that repo's
+        merged PRs are listed at all. This inverts the pre-#1200 contract
+        (`test_repo_with_no_canonical_rows_is_skipped`), which asserted the
+        skip as intended -- that assertion was the bug."""
         fake = _FakeGhDirectToMain([])
         with TemporaryDirectory() as td:
             status = Path(td) / "cross-repo-status.json"
@@ -735,10 +740,73 @@ class MergedPrsDirectToMain(unittest.TestCase):
             )
             with mock.patch.object(wave_status.subprocess, "run", fake):
                 got = wave_status.merged_prs("10", "29", status)
-        self.assertEqual(got, [])
-        # No "pr list" call was ever made for the scopeless repo.
+        self.assertEqual(got, [])  # no PRs in the fixture -- nothing to find
+        # A "pr list" call WAS made for the scopeless repo (main#1200's fix).
         pr_list_repos = {c[c.index("--repo") + 1] for c in fake.calls if c[1:3] == ["pr", "list"]}
-        self.assertNotIn("noorinalabs/noorinalabs-isnad-ingest-platform", pr_list_repos)
+        self.assertIn("noorinalabs/noorinalabs-isnad-ingest-platform", pr_list_repos)
+
+    def test_wave_with_no_canonical_scope_at_all_skips_every_repo(self) -> None:
+        """The ONLY remaining skip condition after main#1200: when the wave's
+        canonical scope is empty across every repo, nothing could ever match,
+        so no repo is queried at all -- not a per-repo decision, a wave-level
+        one."""
+        fake = _FakeGhDirectToMain([])
+        with TemporaryDirectory() as td:
+            status = Path(td) / "cross-repo-status.json"
+            _write_direct_to_main_status(
+                status,
+                wave="29",
+                repos=["noorinalabs-main", "noorinalabs-isnad-ingest-platform"],
+                kickoff="2026-07-27T22:56:17Z",
+                tiers={},  # no canonical scope rows anywhere
+            )
+            with mock.patch.object(wave_status.subprocess, "run", fake):
+                got = wave_status.merged_prs("10", "29", status)
+        self.assertEqual(got, [])
+        self.assertEqual(fake.calls, [])  # no gh call of any kind was made
+
+    def test_cross_repo_reference_from_a_repo_with_zero_own_scope_rows(self) -> None:
+        """main#1200's exact repro: a repo that owns NO canonical scope rows
+        of its own still delivers a genuine cross-repo scope row. Before the
+        fix, `noorinalabs-isnad-graph` here is never queried at all (the
+        pre-#1200 per-repo skip), so PR #77 -- which closes the wave's only
+        canonical row, recorded under `noorinalabs-main` -- is silently
+        dropped: `merged_prs()` returns `[]` instead of `[77]`."""
+        prs = [
+            {
+                "repo": "noorinalabs-isnad-graph",
+                "number": 77,
+                "sha": "sha77",
+                "mergedAt": "2026-07-30T03:00:00Z",
+                "login": "octocat",
+                "commit_author": "Zero Own Rows Author",
+                "closes": [("noorinalabs-main", 1200)],
+            }
+        ]
+        fake = _FakeGhDirectToMain(prs)
+        with TemporaryDirectory() as td:
+            status = Path(td) / "cross-repo-status.json"
+            scope: dict = {"theme": "test fixture", "merge_model": "direct-to-main"}
+            # The wave's ONLY canonical row lives under noorinalabs-main.
+            # noorinalabs-isnad-graph is in repos_in_scope but owns zero rows
+            # of its own -- exactly the shape main#1200 requires.
+            scope["tier_1_main"] = [_scope_row(("noorinalabs-main", 1200))]
+            data = {
+                "current_wave": 29,
+                "wave_29_repos_in_scope": ["noorinalabs-main", "noorinalabs-isnad-graph"],
+                "wave_29_kicked_off_at": "2026-07-27T22:56:17Z",
+                "wave_29_merge_model": "direct-to-main",
+                "wave_29_scope": scope,
+            }
+            status.write_text(json.dumps(data, indent=2) + "\n", encoding="utf-8")
+            with mock.patch.object(wave_status.subprocess, "run", fake):
+                got = wave_status.merged_prs("10", "29", status)
+        self.assertEqual([p["number"] for p in got], [77])
+        # Both repos were queried -- isnad-graph despite owning no row of
+        # its own.
+        pr_list_repos = {c[c.index("--repo") + 1] for c in fake.calls if c[1:3] == ["pr", "list"]}
+        self.assertIn("noorinalabs/noorinalabs-isnad-graph", pr_list_repos)
+        self.assertIn("noorinalabs/noorinalabs-main", pr_list_repos)
 
     def test_counters_nonzero_and_nonempty_trust_signals_for_direct_to_main(self) -> None:
         """The whole point of main#1131: compute_counters()/trust_signals must
--- a/.claude/lib/tests/test_wave_status.py
+++ b/.claude/lib/tests/test_wave_status.py
@@ -398,6 +398,14 @@ class _FakeGhDirectToMain:
     commit_author, closes (the issue numbers GitHub records this PR as
     closing — NOT assumed to equal the PR number, mirroring main#1172 being
     delivered by PR #1173).
+
+    ``closes`` entries are repo-qualifiable (main#1189): a bare int (e.g.
+    ``1172``) means "closes that number in this PR's own repo" — the shape
+    every pre-#1189 fixture uses, kept working unchanged. A ``(repo, number)``
+    tuple (e.g. ``("noorinalabs-main", 1172)``) means "closes that number in
+    a DIFFERENT repo" — the cross-repo case the fixture could not previously
+    express at all (a bare-int list has no way to name a repo other than the
+    PR's own), which is exactly the gap main#1189 fixes.
     """
 
     def __init__(self, prs: list[dict]) -> None:
@@ -452,7 +460,17 @@ class _FakeGhDirectToMain:
             closes = next(
                 p["closes"] for p in self.prs if p["repo"] == repo and p["number"] == number
             )
-            return SimpleNamespace(stdout=json.dumps(closes), returncode=0, stderr="")
+            # Mirrors the real `--jq` filter's output shape (main#1189):
+            # `[{number, repo: .repository.name}]`. A bare int closes-entry
+            # defaults to the PR's own repo; a `(repo, number)` tuple names a
+            # different repo explicitly.
+            nodes = [
+                {"number": c[1], "repo": c[0]}
+                if isinstance(c, tuple)
+                else {"number": c, "repo": repo}
+                for c in closes
+            ]
+            return SimpleNamespace(stdout=json.dumps(nodes), returncode=0, stderr="")
 
         if cmd[1] == "api":
             path = cmd[2]
@@ -469,8 +487,22 @@ class _FakeGhDirectToMain:
         raise AssertionError(f"unexpected gh call: {cmd!r}")
 
 
-def _scope_row(issue_number: int) -> dict:
-    return {"id": f"noorinalabs-main#{issue_number}", "ref": f"main#{issue_number}"}
+def _scope_row(entry: int | tuple[str, int]) -> dict:
+    """A ``wave_{M}_scope`` tier row, shaped like the real record.
+
+    A bare int defaults to ``noorinalabs-main`` (every pre-#1189 call site's
+    assumption, kept working unchanged). A ``(repo, issue_number)`` tuple
+    names a different repo explicitly -- the shape needed to express a
+    multi-repo canonical scope for the main#1189 cross-repo regression test.
+    """
+    if isinstance(entry, tuple):
+        repo, issue_number = entry
+    else:
+        repo, issue_number = "noorinalabs-main", entry
+    return {
+        "id": f"{repo}#{issue_number}",
+        "ref": f"{repo.removeprefix('noorinalabs-')}#{issue_number}",
+    }
 
 
 def _write_direct_to_main_status(
@@ -739,6 +771,188 @@ class MergedPrsDirectToMain(unittest.TestCase):
             counters,
             {"final_pr_count": 0, "changes_requested_cycles": 0, "top_concentration_pct": 0},
         )
+
+    def test_cross_repo_closing_reference_no_undercount_or_misattribution(self) -> None:
+        """main#1189: a closing reference is not necessarily in the closing
+        PR's own repo -- child-repo PRs routinely close a parent meta-issue
+        recorded under a different repo. Two failure modes proven together:
+
+        PR #77 (repo noorinalabs-isnad-graph) closes noorinalabs-main#1200 --
+        a genuine CROSS-REPO scope row, not one of isnad-graph's own rows.
+        Bare-number matching against only isnad-graph's own canonical set
+        ({50, 500}) would never see 1200 there and DROP a real match
+        (under-count).
+
+        PR #80 (also repo noorinalabs-isnad-graph) closes
+        noorinalabs-main#500 -- an issue that only numerically COLLIDES with
+        isnad-graph's own scope row #500 (a different, unrelated issue).
+        Bare-number matching against isnad-graph's own canonical set would
+        find 500 there and wrongly count PR #80 as closing isnad-graph#500
+        (mis-attribution).
+
+        Repo-qualified matching against the FULL (repo, number) canonical set
+        gets both right: PR #77 counted, PR #80 excluded.
+        """
+        prs = [
+            {
+                "repo": "noorinalabs-isnad-graph",
+                "number": 77,
+                "sha": "sha77",
+                "mergedAt": "2026-07-30T03:00:00Z",
+                "login": "octocat",
+                "commit_author": "Cross Repo Author",
+                "closes": [("noorinalabs-main", 1200)],
+            },
+            {
+                "repo": "noorinalabs-isnad-graph",
+                "number": 80,
+                "sha": "sha80",
+                "mergedAt": "2026-07-30T03:05:00Z",
+                "login": "octocat",
+                "commit_author": "Collision Author",
+                "closes": [("noorinalabs-main", 500)],  # NOT isnad-graph's own #500
+            },
+        ]
+        fake = _FakeGhDirectToMain(prs)
+        with TemporaryDirectory() as td:
+            status = Path(td) / "cross-repo-status.json"
+            scope: dict = {"theme": "test fixture", "merge_model": "direct-to-main"}
+            scope["tier_1_main"] = [_scope_row(("noorinalabs-main", 1200))]
+            scope["tier_2_isnad_graph"] = [
+                _scope_row(("noorinalabs-isnad-graph", 50)),
+                _scope_row(("noorinalabs-isnad-graph", 500)),
+            ]
+            data = {
+                "current_wave": 29,
+                "wave_29_repos_in_scope": ["noorinalabs-main", "noorinalabs-isnad-graph"],
+                "wave_29_kicked_off_at": "2026-07-27T22:56:17Z",
+                "wave_29_merge_model": "direct-to-main",
+                "wave_29_scope": scope,
+            }
+            status.write_text(json.dumps(data, indent=2) + "\n", encoding="utf-8")
+            with mock.patch.object(wave_status.subprocess, "run", fake):
+                got = wave_status.merged_prs("10", "29", status)
+        self.assertEqual([p["number"] for p in got], [77])
+
+
+class ReconciliationWarning(unittest.TestCase):
+    """main#1190: canonical scope rows no merged PR's closing references
+    claimed are surfaced as a WARNING (never an error, never silently
+    dropped)."""
+
+    def test_warns_on_unclaimed_canonical_scope_rows(self) -> None:
+        prs = [
+            {
+                "repo": "noorinalabs-main",
+                "number": 1173,
+                "sha": "sha1173",
+                "mergedAt": "2026-07-30T02:16:40Z",
+                "login": "octocat",
+                "commit_author": "Nino Kavtaradze",
+                "closes": [1172],
+            }
+        ]
+        fake = _FakeGhDirectToMain(prs)
+        with TemporaryDirectory() as td:
+            status = Path(td) / "cross-repo-status.json"
+            _write_direct_to_main_status(
+                status,
+                wave="29",
+                repos=["noorinalabs-main"],
+                kickoff="2026-07-27T22:56:17Z",
+                tiers={"tier_4_in_wave_findings": [1172, 1160, 1162]},
+            )
+            with mock.patch.object(wave_status.subprocess, "run", fake):
+                warning = wave_status.reconciliation_warning("10", "29", status)
+        self.assertIsNotNone(warning)
+        assert warning is not None  # narrows the type for the checks below
+        self.assertIn("scope rows with no matching merged PR", warning)
+        self.assertIn("main#1160", warning)
+        self.assertIn("main#1162", warning)
+        self.assertNotIn("main#1172", warning)  # claimed by PR #1173 -- must not appear
+        self.assertIn("2 of 3", warning)
+
+    def test_no_warning_when_every_scope_row_is_claimed(self) -> None:
+        prs = [
+            {
+                "repo": "noorinalabs-main",
+                "number": 1173,
+                "sha": "sha1173",
+                "mergedAt": "2026-07-30T02:16:40Z",
+                "login": "octocat",
+                "commit_author": "Nino Kavtaradze",
+                "closes": [1172],
+            }
+        ]
+        fake = _FakeGhDirectToMain(prs)
+        with TemporaryDirectory() as td:
+            status = Path(td) / "cross-repo-status.json"
+            _write_direct_to_main_status(
+                status,
+                wave="29",
+                repos=["noorinalabs-main"],
+                kickoff="2026-07-27T22:56:17Z",
+                tiers={"tier_4_in_wave_findings": [1172]},
+            )
+            with mock.patch.object(wave_status.subprocess, "run", fake):
+                warning = wave_status.reconciliation_warning("10", "29", status)
+        self.assertIsNone(warning)
+
+    def test_no_warning_under_wave_branch_model(self) -> None:
+        """The wave-branch path's base+timestamp counting has no
+        closing-reference dependency -- reconciliation is a
+        direct-to-main-only concept."""
+        prs = _p5w4_prs()
+        fake = _FakeGh(prs)
+        with TemporaryDirectory() as td:
+            status = Path(td) / "cross-repo-status.json"
+            data = {
+                "wave_4_repos_in_scope": _REPOS,
+                "wave_4_kicked_off_at": "2026-06-15T01:52:55Z",
+                "wave_4_merge_model": "wave-branch",
+            }
+            status.write_text(json.dumps(data))
+            with mock.patch.object(wave_status.subprocess, "run", fake):
+                warning = wave_status.reconciliation_warning("5", "4", status)
+        self.assertIsNone(warning)
+
+    def test_cmd_counters_emits_warning_to_stderr_not_stdout(self) -> None:
+        """The warning must not corrupt the counters JSON on stdout (a
+        `--write` caller pipes stdout) -- it goes to stderr, same channel as
+        the existing ERROR lines."""
+        import io
+        from contextlib import redirect_stderr, redirect_stdout
+
+        prs = [
+            {
+                "repo": "noorinalabs-main",
+                "number": 1173,
+                "sha": "sha1173",
+                "mergedAt": "2026-07-30T02:16:40Z",
+                "login": "octocat",
+                "commit_author": "Nino Kavtaradze",
+                "closes": [1172],
+            }
+        ]
+        fake = _FakeGhDirectToMain(prs)
+        with TemporaryDirectory() as td:
+            status = Path(td) / "cross-repo-status.json"
+            _write_direct_to_main_status(
+                status,
+                wave="29",
+                repos=["noorinalabs-main"],
+                kickoff="2026-07-27T22:56:17Z",
+                tiers={"tier_4_in_wave_findings": [1172, 1160]},
+            )
+            out, err = io.StringIO(), io.StringIO()
+            with mock.patch.object(wave_status.subprocess, "run", fake):
+                with redirect_stdout(out), redirect_stderr(err):
+                    rc = wave_status.main(["counters", "10", "29", "--status", str(status)])
+        self.assertEqual(rc, 0)
+        stdout_val = out.getvalue()
+        stderr_val = err.getvalue()
+        json.loads(stdout_val)  # stdout is pure valid JSON -- no warning text mixed in
+        self.assertIn("main#1160", stderr_val)
 
 
 class PrListPageCap(unittest.TestCase):
--- a/.claude/lib/wave_status.py
+++ b/.claude/lib/wave_status.py
@@ -182,8 +182,19 @@ def _canonical_issue_numbers_by_repo(wave: str, status_path: Path) -> dict[str, 
     return by_repo
 
 
-def _pr_closing_issue_numbers(repo: str, number: int) -> set[int]:
-    """Issue numbers GitHub recognises *number* as closing, via GraphQL.
+def _pr_closing_issue_numbers(repo: str, number: int) -> set[tuple[str, int]]:
+    """Repo-qualified issue numbers GitHub recognises *number* as closing.
+
+    Returns ``{(repo_name, issue_number), ...}`` — deliberately NOT bare
+    issue numbers (main#1189). A closing reference is not necessarily in
+    *repo*: child-repo PRs routinely close a parent meta-issue recorded
+    under a different repo (e.g. ``noorinalabs-main``), and two repos can
+    independently have an issue #N. Comparing bare numbers against a single
+    repo's canonical scope set either silently drops the cross-repo case
+    (under-count) or, on a same-number collision across repos, matches the
+    wrong issue (mis-attribution) — confirmed live via `gh api graphql`
+    against PR #1173. ``repository{name}`` is available on the same GraphQL
+    node at no extra cost.
 
     The REST-backed ``gh pr list --json`` surface (this org pins gh 2.45.0)
     has no ``closingIssuesReferences`` field, so this goes through
@@ -192,12 +203,21 @@ def _pr_closing_issue_numbers(repo: str, number: int) -> set[int]:
     #1171 were bundled into a single PR — the closing-reference set is the
     only reliable link between a scope row and the PR that actually delivered
     it under a direct-to-main wave.
+
+    ``first:100`` (raised from the pre-#1189 ``first:25``) — the same
+    silent-truncation family as :data:`_PR_LIST_LIMIT` below: a PR closing
+    more issues than the page size would otherwise drop the excess with no
+    signal. 100 is GitHub's GraphQL connection page-size ceiling for this
+    field, so it is the maximum obtainable in a single page; a PR closing
+    over 100 issues would need real pagination (``endCursor``/``hasNextPage``)
+    — no real wave has approached that, so it is flagged here rather than
+    silently left uncapped or unremarked.
     """
     query = (
         "query($owner:String!,$name:String!,$number:Int!){"
         "repository(owner:$owner,name:$name){"
         "pullRequest(number:$number){"
-        "closingIssuesReferences(first:25){nodes{number}}"
+        "closingIssuesReferences(first:100){nodes{number repository{name}}}"
         "}}}"
     )
     raw = _run_gh(
@@ -213,10 +233,12 @@ def _pr_closing_issue_numbers(repo: str, number: int) -> set[int]:
             "-F",
             f"number={number}",
             "--jq",
-            "[.data.repository.pullRequest.closingIssuesReferences.nodes[].number]",
+            "[.data.repository.pullRequest.closingIssuesReferences.nodes[]"
+            " | {number, repo: .repository.name}]",
         ]
     )
-    return {int(n) for n in json.loads(raw or "[]")}
+    nodes = json.loads(raw or "[]")
+    return {(str(n["repo"]), int(n["number"])) for n in nodes}
 
 
 def _merged_prs_wave_branch(phase: str, wave: str, status_path: Path) -> list[dict]:
@@ -287,7 +309,9 @@ def _merged_prs_wave_branch(phase: str, wave: str, status_path: Path) -> list[di
 _PR_LIST_LIMIT = 1000
 
 
-def _merged_prs_direct_to_main(wave: str, status_path: Path) -> list[dict]:
+def _merged_prs_direct_to_main(
+    wave: str, status_path: Path
+) -> tuple[list[dict], set[tuple[str, int]]]:
     """Build the wave's merged-PR set for a ``direct-to-main`` wave (main#1131).
 
     No wave branch exists under this model, so ``--base <wave-branch>``
@@ -297,9 +321,14 @@ def _merged_prs_direct_to_main(wave: str, status_path: Path) -> list[dict]:
     Instead: list merged-to-main PRs per in-scope repo, apply the existing
     ``wave_{M}_kicked_off_at`` cross-window filter as a pre-filter, then keep
     only the PRs whose GitHub-recognised closing-issue references intersect
-    that repo's canonical scope-issue set (:func:`_canonical_issue_numbers_by_repo`).
-    A repo with no canonical scope rows is skipped entirely rather than
-    querying every merged PR in that repo.
+    the wave's FULL canonical scope-issue set, across every repo
+    (:func:`_canonical_issue_numbers_by_repo`) — not just the queried repo's
+    own scope rows (main#1189: a closing reference is not necessarily in the
+    same repo as the PR, e.g. a child-repo PR closing a parent meta-issue, and
+    matching against only the queried repo's own numbers either drops that
+    case or, on a same-number collision across repos, attributes it to the
+    wrong issue). A repo with no canonical scope rows of its own is skipped
+    entirely rather than querying every merged PR in that repo.
 
     The listing itself is bounded two ways (main#1131 M2 — `gh pr list`
     defaults to `--limit 30`, and `--base main` removes the wave-branch's
@@ -310,12 +339,22 @@ def _merged_prs_direct_to_main(wave: str, status_path: Path) -> list[dict]:
     ``merged_at < kickoff`` filter below is kept as defense-in-depth in case
     the search qualifier's date granularity and the recorded kickoff instant
     ever disagree at the second.
+
+    Returns ``(merged_prs, claimed_pairs)`` — alongside the PR list, the set
+    of canonical ``(repo, issue_number)`` scope pairs actually claimed by some
+    merged PR's closing references. :func:`reconciliation_warning` (main#1190)
+    needs this to report scope rows no merged PR claims, without re-running
+    every ``gh api graphql`` closing-reference call a second time.
     """
     repos = read_repos(wave, status_path)
     kickoff = _kickoff_ts(wave, status_path)
     issue_numbers_by_repo = _canonical_issue_numbers_by_repo(wave, status_path)
+    canonical_pairs: set[tuple[str, int]] = {
+        (repo_name, n) for repo_name, nums in issue_numbers_by_repo.items() for n in nums
+    }
 
     out: list[dict] = []
+    claimed: set[tuple[str, int]] = set()
     for repo in repos:
         canonical_issues = issue_numbers_by_repo.get(repo)
         if not canonical_issues:
@@ -342,8 +381,10 @@ def _merged_prs_direct_to_main(wave: str, status_path: Path) -> list[dict]:
             if kickoff and merged_at < kickoff:
                 continue
             closes = _pr_closing_issue_numbers(repo, pr["number"])
-            if not closes & canonical_issues:
+            hit = closes & canonical_pairs
+            if not hit:
                 continue
+            claimed |= hit
             sha = pr["headRefOid"]
             commit_author = _commit_author_name(repo, sha)
             out.append(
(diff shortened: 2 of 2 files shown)


parametrization:
## Review: `.claude/lib/wave_status.py` + `.claude/lib/tests/test_wave_status.py`

### `.claude/lib/wave_status.py`

**Must-fix — `_merged_prs_direct_to_main`'s per-repo pre-filter reintroduces the exact under-count #1189 set out to fix.**

The per-repo skip guard is untouched by this PR:

```python
for repo in repos:
    canonical_issues = issue_numbers_by_repo.get(repo)
    if not canonical_issues:
        continue
    ...
```

This still gates on whether *that specific repo* has canonical scope rows of its own, even though the whole premise of this fix (and #1189's issue body, verbatim: "A PR in repo A that closes an issue in repo B — normal in this org, where meta-issues in `noorinalabs-main` are linked from child-repo PRs") is that a repo's merged PRs can close scope rows recorded under a *different* repo. A repo in `wave_M_repos_in_scope` that has **zero** canonical rows of its own this wave (all its tracking done via a meta-issue filed under e.g. `noorinalabs-main`) never gets its merged PRs listed at all — `gh pr list` is never even called for it — so any PR there closing a genuine cross-repo canonical row is silently dropped. Nothing in `lifecycle.py::scope()` or the callers ties `repos_in_scope` membership to "has a row of its own in `wave_{M}_scope`", so this isn't a hypothetical: it's exactly the shape #1189's own issue describes as routine.

Reproduced concretely against this PR's own fixture machinery (repo with 0 own canonical rows, one merged PR closing a real cross-repo row):

```python
prs = [{"repo": "noorinalabs-isnad-ingest-platform", "number": 42, ...,
        "closes": [("noorinalabs-main", 1200)]}]
scope["tier_1_main"] = [_scope_row(("noorinalabs-main", 1200))]
# no tier row for noorinalabs-isnad-ingest-platform itself
wave_status.merged_prs("10", "29", status)  # -> [] , should include PR #42
```

`merged_prs()` returns `[]` — PR #42 is silently dropped even though it closes an in-scope canonical row, reproducing the under-count failure mode #1189 exists to close, just relocated from the matching step (fixed) to the listing step (not fixed). The new regression test (`test_cross_repo_closing_reference_no_undercount_or_misattribution`) doesn't catch this because its `noorinalabs-isnad-graph` fixture carries two of its own canonical rows (50, 500), so the pre-filter never fires there.

Suggested fix: gate the skip on whether the **wave** has any canonical scope at all (`canonical_pairs`), not on whether the individual repo has rows of its own — i.e. skip a repo's `gh pr list` call only when `canonical_pairs` is empty (or when `repos` is empty), otherwise query every repo in `repos_in_scope`. Add a regression test with a repo that has zero rows of its own to close this gap for real.

Note: this doesn't corrupt today's live wave-29 numbers — I checked `cross-repo-status.json` and both wave-29 in-scope repos (`noorinalabs-main`, `noorinalabs-isnad-ingest-platform`) currently have ≥1 canonical row of their own, so the guard doesn't fire this wave. But it's a latent gap in the exact fix this PR claims to deliver, on code that computes wave-29's own trust-matrix inputs, so I'm treating it as a must-fix rather than tech debt.

**Everything else checked out:**
- `_pr_closing_issue_numbers` → `set[tuple[str, int]]` via `repository{name}`, `first:25` → `first:100` with a docstring flagging the un-guarded page-size ceiling: acceptable as documented — matches the existing `_PR_LIST_LIMIT` precedent in this same file (a hard cap + doc note, no active runtime truncation guard), and the CI `GraphQL pagination over-cap lint` passes against it.
- `_merged_prs_with_claims` / `reconciliation_warning` / `_reconciliation_warning_from_claims`: correct single-pass design, confirmed the warning only fires under `direct-to-main` (`None` under `wave-branch`), and confirmed live that `_cmd_counters`'s stdout stays pure JSON while the warning lands on stderr only.
- `merged_prs()` and the `wave-branch`/legacy (`None`) merge-model paths are unchanged in behavior — `_merged_prs_with_claims` returns `(_merged_prs_wave_branch(...), None)` for those, identical to pre-PR `merged_prs()`.

### `.claude/lib/tests/test_wave_status.py`
Fixture widening is real — `_FakeGhDirectToMain` and `_scope_row` now genuinely support a `(repo, number)` tuple alongside the pre-existing bare-int shape, and I confirmed the pre-#1189 bare-int fixture usages are unaffected. Good test hygiene overall; the one gap is the missing "repo with zero own scope rows" case noted above.

## Independent verification performed

- CI: re-checked `check-runs` directly against head SHA `d7024bf7...` — 14/14 `success`, 0 pending/failing.
- Isolated `git worktree add` at the head SHA (no checkout in the parent, no work from the implementer's tree); ran the full suite there: **956 passed, 19 subtests passed**.
- `test_wave_status.py` alone: **35 passed**.
- **Mutation A** (reverted the intersection to `closes & {(repo, n) for n in canonical_issues}`): reproduced exactly — `test_cross_repo_closing_reference_no_undercount_or_misattribution` fails, 34 others pass.
- **Mutation B** (forced `_reconciliation_warning_from_claims` to always `return None`): reproduced exactly — `test_warns_on_unclaimed_canonical_scope_rows` and `test_cmd_counters_emits_warning_to_stderr_not_stdout` fail, 33 others pass.
- Ran `ruff check` and `mypy` on both changed files directly: clean.
- Independently ran `counters` against the live `cross-repo-status.json` for wave 29: `final_pr_count=11`, `changes_requested_cycles=9`, `top_concentration_pct=18`, warning lists 24 of 36 unclaimed scope rows — matches the PR body's claim exactly.
- Confirmed the repo-skip gap above with a standalone repro against this PR's own fixture helpers (not just static reading).

All five of the implementer's claims were independently reproduced. The repo-skip gap is a new finding from this review, not something the implementer claimed to have checked.

---
Requestor: Wanjiku Mwangi
Requestee: Santiago Ferreira
TechDebt: none

parametrization:
> **Amended after posting** — this comment originally carried an `Approved` verdict. It now carries `ChangesRequested`, for the reason set out in the *Amendment* section at the end. The analysis below is unchanged from the original post; only the disposition of the `_merged_prs_direct_to_main` repo-skip finding has moved from tech-debt to blocking. GitHub's edit history preserves the original.

## Merge-gate review — PR #1198

Reviewed at head SHA `d7024bf74119f764051c3c2e184363852097cecd` from an isolated clone checked out at that SHA (byte-identical to `gh api tarball/d7024bf7…`, diffed both files). No parent-repo checkout, no implementer worktree.

**Provenance, stated up front.** I filed #1189 and #1190 during the merge-gate review of #1185, so this PR is downstream of my own findings. I have therefore verified what landed rather than reading my intent into it — including checking whether the *second* half of each issue (the fixture-expressiveness ask on #1189, the "distinguish not-delivered from missing-a-`Closes`-reference" ask on #1190) is genuinely satisfied.

---

### Pre-flight — the framing, checked rather than accepted

| Claim | How verified | Result |
|---|---|---|
| Head SHA / mergeable | `gh pr view --json headRefOid,mergeable,mergeStateStatus` immediately before posting | `d7024bf7…`, `MERGEABLE`, `CLEAN` |
| CI green at head | `statusCheckRollup` grouped by conclusion | `{"SUCCESS": 14}` — 14 checks, 0 failing, 0 pending |
| One commit, no force-push | `pulls/1198/commits` length; `issues/1198/timeline` filtered to `head_ref_force_pushed` | 1 commit, 0 force-pushes — HEAD-SHA anchor intact |
| Diff scope | REST diff | only `.claude/lib/wave_status.py` and `.claude/lib/tests/test_wave_status.py`; every deleted line accounted for |

---

### `.claude/lib/wave_status.py`

**`_pr_closing_issue_numbers` — #1189's first half, correct, and verified against the real engine.** I did not trust the query string. I executed the exact query the code builds, with the exact `--jq` filter, against GitHub:

```
$ gh api graphql -f "query=…closingIssuesReferences(first:100){nodes{number repository{name}}}…" \
    -f owner=noorinalabs -f name=noorinalabs-main -F number=1173 \
    --jq '[.data.repository.pullRequest.closingIssuesReferences.nodes[] | {number, repo: .repository.name}]'
[{"number":1172,"repo":"noorinalabs-main"}]
```

The shape matches what `_FakeGhDirectToMain` returns, so the fixture models the real payload rather than the parser's own convenience.

**`_merged_prs_direct_to_main` — the intersection is right; the guard above it is not (non-blocking, filed as #1200).** The intersection against the full canonical `(repo, number)` pair set is correct and mutation-proven. But the per-repo skip immediately above it is unchanged from the pre-#1189 body, and `canonical_issues` now feeds *nothing but that skip*:

```python
canonical_issues = issue_numbers_by_repo.get(repo)
if not canonical_issues:
    continue                          # repo never queried
...
hit = closes & canonical_pairs        # correctly repo-qualified — unreachable for that repo
```

A repo in `repos_in_scope` owning no scope row of its own is dropped before the repo-qualified match runs, even when its PR closes a canonical row in another repo. That is the exact scenario #1189's text named — a child-repo PR closing a parent meta-issue. Reproduced against this head SHA (one canonical row `noorinalabs-main#1200`, one merged `noorinalabs-isnad-graph#77` closing it):

```
[P1] merged PR numbers = []            (correct answer: [77])
[P1] repos actually queried = ['noorinalabs/noorinalabs-main']
```

*(Disposition amended — see the Amendment at the end. The reasoning below is what I originally concluded and is left standing so the change is legible.)* I did **not** treat this as a blocker, for two reasons I checked rather than assumed. It is not live-reachable — `_canonical_issue_numbers_by_repo` on the real wave-29 record returns rows for both in-scope repos (`noorinalabs-main` 35, `noorinalabs-isnad-ingest-platform` `{152}`), so neither is skipped. And when it *does* fire, #1190's warning names the dropped row as unclaimed (`scope rows with no matching merged PR: main#1200 (1 of 1)`), so the drop surfaces as a visible discrepancy at wrapup rather than a silent one. The two fixes compose: the second covers the first's residue. That is the right severity for debt, and #1189 as I wrote it asked for repo-qualification and an expressive fixture, both of which landed — widening it now would be moving the goalposts.

**`first:100` — my ruling.** The cap raise **discharges the practical half** of the finding and is a **documented deferral on the structural half**. 100 is genuinely GitHub's ceiling, verified rather than accepted from the docstring:

```
$ … closingIssuesReferences(first:101) …
"type":"EXCESSIVE_PAGINATION" … "exceeds the `first` limit of 100 records"   (gh exits 1)
```

A single page cannot do better, so `first:25 → first:100` is the maximum available without cursor pagination, and the docstring naming `endCursor`/`hasNextPage` as the real fix is an honest concession. What the raise did not buy is a *signal*: the helper discards the page boundary, so a PR at exactly 100 nodes is indistinguishable from a truncated one. Under the very principle #1190 establishes, a full page should say so. That is one comparison, not a redesign — filed as #1201 rather than held against this PR.

**Fail-open audit — every early return and every failed-fetch path.** This is where a gate's hole usually is, so I enumerated them:

| Path | Behaviour | Verdict |
|---|---|---|
| `_pr_closing_issue_numbers` gh call fails | `_run_gh` uses `check=True` → `CalledProcessError` → `main`'s handler → `rc=1`, empty stdout, `ERROR:` on stderr | Loud. Confirmed live: `gh api graphql` on a nonexistent PR exits **1** with a `NOT_FOUND` payload, so this is reachable in production, not just in the fake. |
| Reproduced end-to-end with a gh that raises on `graphql` | `rc=1`, `stdout=''`, `stderr` begins `ERROR: gh call failed (exit 1)` | No silent 0 — the #1190 failure mode does not recur here |
| `claimed is None` (wave-branch) | returns `None`, no warning | Correct — no closing-reference dependency to reconcile |
| every canonical row claimed | returns `None` | Correct |
| **`canonical` empty** | returns `None`, every repo skipped, `counters` exits 0 printing `final_pr_count: 0`, **no stderr at all** | The one genuine fail-open. Reachable via a scope of legacy plain-string rows — a shape the suite itself tolerates. Latent (wave-29 parses 36/36). Filed as #1201. |

The related, sharper case is partial: a scope declaring 4 rows of which 2 are unparseable reports `(1 of 2)` — rows that never parsed are absent from numerator *and* denominator, so the instrument silently understates its own scope. Also #1201.

**stdout/stderr discipline — clean, and I checked the consumers too.** Verified on both wired subcommands: `counters` stdout parses as JSON with the warning on stderr, and `merged-prs` likewise (its test only covered `counters`, so I exercised `merged-prs` separately). On the caller side, `/wave-wrapup` invokes `wave_status.py counters … --expect … --write` with no `2>&1` and consumes the `repos` subcommand (which emits no warning) through process substitution; `lifecycle.py counters` delegates in-process via `wave_status.main`, so it inherits the same split. No path merges the streams. The warning cannot contaminate the JSON the wrapup parses.

**Regression surface — the wave-branch and legacy/`None` paths.** `_merged_prs_with_claims` refactors a hot path, so byte-equivalence needed proof rather than inspection. The pre-existing `BaseVsHeadDifferential` (untouched by this diff) `git show`s the pre-#1131 `wave_status.py` out of `b290d611`, executes *that* module's `merged_prs` against the same fixture and the same fake gh as HEAD's `_merged_prs_wave_branch`, and asserts equality — then repeats it through the public dispatcher with no `merge_model` key recorded, covering the legacy/`None` shape. That is a real runtime base-vs-head comparison, and it executes and passes here. Transitively it is the stronger claim: HEAD's wave-branch output equals *pre-#1131* output, not merely pre-#1198.

### `.claude/lib/tests/test_wave_status.py`

**#1189's second half is satisfied.** I asked for a fixture model that can *express* a cross-repo closing reference. `_FakeGhDirectToMain` now accepts a `(repo, number)` tuple in `closes` and `_scope_row` accepts one too, with bare ints still defaulting to the PR's own repo so no pre-existing fixture changed meaning. The gap is no longer structurally unreachable.

**#1190's second half is satisfied.** The warning distinguishes the two cases I asked it to: it lists the specific unclaimed rows by qualified ref (`main#1160`), gives the ratio, and stays a warning with `rc=0`, so a human at wrapup can tell "not delivered yet" from "delivered but missing a `Closes` reference." Live output confirms it (below).

**Coverage criticism, non-blocking.** `test_cross_repo_closing_reference_no_undercount_or_misattribution` is a good test — it proves under-count and mis-attribution together, which is more than I asked for. But its corpus gives `noorinalabs-isnad-graph` its own rows `{50, 500}`, which is the axis the skip guard keys on, so it holds constant the dimension that governs the residual bug. `test_repo_with_no_canonical_rows_is_skipped` then asserts the skip as intended contract. Detail in #1200.

**Mutation testing — both mutants reproduced, and neither is equivalent.**

*Mutation A* — reverted `hit = closes & canonical_pairs` to `closes & {(repo, n) for n in canonical_issues}`: **1 failed, 34 passed**, the failure being exactly `MergedPrsDirectToMain::test_cross_repo_closing_reference_no_undercount_or_misattribution` (`[] != [77]`). Not equivalent — it changes the intersection domain and a fixture trips it.

*Mutation B* — `_reconciliation_warning_from_claims` returns `None` unconditionally: **2 failed, 33 passed**, exactly `ReconciliationWarning::test_warns_on_unclaimed_canonical_scope_rows` and `::test_cmd_counters_emits_warning_to_stderr_not_stdout`. Not equivalent. The two sibling tests that stayed green (`test_no_warning_when_every_scope_row_is_claimed`, `test_no_warning_under_wave_branch_model`) correctly already expect `None`, so their silence is right rather than a gap — worth saying explicitly instead of reporting it as missing coverage.

Both mutants reverted; the file is byte-identical to head afterwards. **35/35** in-file and **956/956** across `.claude/lib/tests/` (plus 19 subtests). `ruff format --check`, `ruff check`, and `mypy` all clean on both changed files locally, matching the 14 green CI checks.

**Corpus dimensions I varied.** Not five reshapes of one case — the axes were: whether the delivering repo owns canonical rows of its own (present / absent); reference direction (child→parent and parent→child, the mirror-image control that isolates the real discriminator, which turns out to be repo-ownership rather than cross-repo-ness); scope parseability (all-parseable / partly / wholly unparseable); the closing-reference fetch outcome (success / `CalledProcessError`); and output stream per subcommand (`counters` and `merged-prs`).

### Live verification

Ran the head-SHA build against the real wave-29 record — a positive control on the whole instrument, not a mock:

```
final_pr_count 11, changes_requested_cycles 9, top_concentration_pct 18
WARNING: scope rows with no matching merged PR: main#1114, …, main#1195 (24 of 36)
```

Exactly the implementer's numbers. I cross-checked `11` independently through a different instrument — `gh pr list --state merged --base main --search "merged:>=2026-07-27T22:56:17Z"` returns **10** in `noorinalabs-main` and **1** in `noorinalabs-isnad-ingest-platform`. One caveat worth recording rather than glossing: because every in-window merged PR in wave-29 happens to *be* in scope, the live run does not discriminate the scope filter — it would return 11 with no filtering at all. It is a positive control proving the instrument does not silently zero out (the #1131 failure), not evidence that the scope exclusion works. The exclusion is proven by the unit fixtures (`us#213`-shaped false positive, the `#500` collision), which do discriminate.

---

### Summary

Both of my issues are genuinely closed, including the second half of each, and the fix is proven at three independent levels: mutation-tested guards that real fixtures trip, a base-vs-head runtime differential protecting the untouched merge model, and a live run against wave-29's own 36-row scope with an independent cross-check of the headline number. The fail-open audit found one real hole and one real residual bug — both latent against today's data, both made *visible* rather than silent by the very warning this PR adds, and both filed as debt rather than held against a PR that delivered what was asked.

This is the instrument wave-29's own retro depends on, so I want the record to be precise about what it does and does not yet see.

---

### Amendment — upgrading the repo-skip finding to blocking

@Wanjiku Mwangi and I reviewed this in parallel without sight of each other's work, and converged independently on the same defect: the `if not canonical_issues: continue` guard in `_merged_prs_direct_to_main`. Two reviewers reproducing the same drop from different fixtures is about as strong as evidence for a finding gets, so the finding is not in question. Only its disposition was, and on reflection theirs is right and mine was wrong.

What moved me is not the peer verdict but re-reading my own #1189. The sentence in that issue is "a PR in repo A that closes an issue in repo B — normal in this org, where meta-issues in `noorinalabs-main` are linked from child-repo PRs." When the child repo owns no scope row of its own, that exact sentence's scenario still under-counts after this fix. Holding the PR to it is not moving the goalpost; it is the goalpost I set. My original tech-debt call leaned on a scoping argument — that the literal asks in #1189 landed — rather than a correctness one, and that was the wrong basis.

Two further reasons the balance tips:

- The mitigation I credited is a *detection* mitigation, not a correctness one. #1190's warning makes the dropped row visible, which is genuinely valuable, but `final_pr_count` is still wrong — and `final_pr_count` is the number feeding the trust matrix, the number wave-28's retro got wrong twice, and the reason #1131/#1185/#1198 exist at all. "Wrong but accompanied by a warning a human might reconcile" is not the bar this instrument was built to clear; correct-by-construction is.
- The charter principle in § Security Guards Belong Inline, Not in a Followup generalises here. A correctness guard for the very defect class the PR exists to close belongs in the PR, not in a followup issue — especially when `canonical_issues` now feeds nothing but the guard, so the fix is a one-condition change plus one test.

The change I am asking for is the same one Wanjiku asked for, and it should be treated as a single request, not two: gate the skip on `canonical_pairs` (does the *wave* have any canonical scope) rather than on `issue_numbers_by_repo.get(repo)` (does *this repo* own rows), and add a regression test whose delivering repo has zero scope rows of its own. `--search merged:>=<kickoff>` plus `_PR_LIST_LIMIT` already bound the cost, so the extra `gh pr list` per scopeless in-scope repo is cheap. `test_repo_with_no_canonical_rows_is_skipped` currently asserts the hole as intended contract and will need inverting in the same commit.

#1200 stays open as the finding's record — close it with the fix rather than leaving it as debt. #1201 (the warning's blind edges: unparseable scope rows shrinking the denominator, and a full 100-node page truncating with no signal) remains genuine tech-debt and is **not** part of this request; I would not hold the PR for it.

Everything else in this review stands unchanged, including the `first:100` ruling, the fail-open audit, the stdout/stderr result, the wave-branch byte-equivalence proof, and both reproduced mutations. Please land the guard fix as an additive commit on this branch — no force-push, so the head-SHA anchor stays intact for re-review.

---
Requestor: Weronika Zielinska
Requestee: Santiago Ferreira
TechDebt: #1200, #1201

parametrization:
Re-anchor your verdict at `9df3efc89f429d02f6d8c4b8cb1915fd63f47001`.
Your prior verdict: https://github.com/noorinalabs/noorinalabs-main/pull/1198#issuecomment-5137132165

One additive commit landed on top of `d7024bf7` (no force-push, no rebase). Fix for the finding in that verdict:

`_merged_prs_direct_to_main`'s per-repo skip guard now gates on `canonical_pairs` (does the wave have any canonical scope at all) instead of `issue_numbers_by_repo.get(repo)` (does this repo own rows of its own) — every repo in `repos_in_scope` is queried unless the wave's canonical scope is empty everywhere. The docstring sentence documenting the old skip as intentional is rewritten.

- `test_repo_with_no_canonical_rows_is_skipped` inverted to `test_repo_with_no_canonical_rows_of_its_own_is_still_queried`.
- Added `test_cross_repo_reference_from_a_repo_with_zero_own_scope_rows`, modeled on your own repro shape (repo with 0 own canonical rows, one merged PR closing a real cross-repo row).
- Added `test_wave_with_no_canonical_scope_at_all_skips_every_repo` for the one remaining, now wave-level, skip condition.

**Mutation testing:** reverting the guard to the pre-fix per-repo gate (Mutation C) turns exactly `test_repo_with_no_canonical_rows_of_its_own_is_still_queried` and `test_cross_repo_reference_from_a_repo_with_zero_own_scope_rows` red — 2 failures, 35 pass. Re-verified the original #1189 intersection fix (Mutation A) is still non-equivalent against this new guard: reverting it now fails both the original and the new zero-own-rows test (2 failures, 35 pass) — the two fixes compose. All mutations reverted; full suite green: 37/37 in `test_wave_status.py`, 958/958 across `.claude/lib/tests/`.

Live-re-ran against wave-29's real 36-row scope: identical output to before this fixup (`final_pr_count=11`, `changes_requested_cycles=9`, `top_concentration_pct=18`, same 24-of-36 reconciliation warning) — both in-scope repos already own ≥1 canonical row this wave, so the guard was not live-reachable either way, confirmed unchanged.

PR body updated with `Closes #1200` alongside `Closes #1189` / `Closes #1190`.

parametrization:
Re-anchor your verdict at `9df3efc89f429d02f6d8c4b8cb1915fd63f47001`.
Your prior verdict: https://github.com/noorinalabs/noorinalabs-main/pull/1198#issuecomment-5137150657

One additive commit landed on top of `d7024bf7` (no force-push, no rebase). Fix for the finding your amendment upgraded to blocking:

`_merged_prs_direct_to_main`'s per-repo skip guard now gates on `canonical_pairs` (does the wave have any canonical scope at all) instead of `issue_numbers_by_repo.get(repo)` (does this repo own rows of its own) — every repo in `repos_in_scope` is queried unless the wave's canonical scope is empty everywhere. The docstring sentence documenting the old skip as intentional is rewritten.

- `test_repo_with_no_canonical_rows_is_skipped` inverted to `test_repo_with_no_canonical_rows_of_its_own_is_still_queried`.
- Added `test_cross_repo_reference_from_a_repo_with_zero_own_scope_rows` and `test_wave_with_no_canonical_scope_at_all_skips_every_repo` for the one remaining, now wave-level, skip condition.

**Mutation testing:** reverting the guard to the pre-fix per-repo gate (Mutation C) turns exactly `test_repo_with_no_canonical_rows_of_its_own_is_still_queried` and `test_cross_repo_reference_from_a_repo_with_zero_own_scope_rows` red — 2 failures, 35 pass. Re-verified the original #1189 intersection fix (Mutation A) is still non-equivalent against this new guard: reverting it now fails both the original and the new zero-own-rows test (2 failures, 35 pass) — the two fixes compose rather than mask each other. All mutations reverted; full suite green: 37/37 in `test_wave_status.py`, 958/958 across `.claude/lib/tests/`.

Live-re-ran against wave-29's real 36-row scope: identical output to before this fixup (`final_pr_count=11`, `changes_requested_cycles=9`, `top_concentration_pct=18`, same 24-of-36 reconciliation warning) — both in-scope repos already own ≥1 canonical row this wave, so the guard was not live-reachable either way, confirmed unchanged.

Not touched, per your rulings in that verdict: `first:100` (no change — your `EXCESSIVE_PAGINATION`-at-101 verification stands), and #1201 (reconciliation-warning blind edges — left as filed tech-debt).

PR body updated with `Closes #1200` alongside `Closes #1189` / `Closes #1190`.

parametrization:
## Re-review: second commit `9df3efc8` (`fix(wave-status): gate the direct-to-main listing skip on wave scope, not per-repo scope (#1200)`)

Scope: delta from `d7024bf7` only, per the re-request. Ancestry confirmed additive (`git merge-base --is-ancestor d7024bf7 9df3efc8` → true), 0 force-pushes, 2 commits on the branch — all re-checked myself, not taken from the brief.

### `.claude/lib/wave_status.py`

The per-repo guard I flagged (`canonical_issues = issue_numbers_by_repo.get(repo); if not canonical_issues: continue`) is removed. In its place:

```python
if not canonical_pairs:
    return out, claimed
for repo in repos:
    ...  # every repo in repos_in_scope is now queried
```

This is the right fix, and it is a materially different check from the one it replaces, not the same bug moved one line up. The old guard tested a **per-repo proxy** (does *this* repo own scope rows?) that was invalidated by #1189's own premise — a repo's PRs can close rows recorded elsewhere, so "this repo owns nothing" was never proof that "this repo's PRs can't match anything." The new guard tests the **global** condition (`canonical_pairs`, the union across every repo): if that is empty, intersection with it is provably empty for *any* repo's closes, no proxy reasoning involved. So it cannot reintroduce the same silent-drop class — I checked the specific edge called out (scope parses partially or empty, #1201's shape):

- **Partially-parsed scope** (some tier rows parse, some don't — #1201 Edge 1): `canonical_pairs` is non-empty as long as at least one row parsed, so the new early-return does not fire, every repo is still queried, and matching proceeds against whatever the (possibly under-counted) `canonical_pairs` is. No new degradation from this commit — the under-counted-denominator behavior in that case is #1201's Edge 1, already filed against the pre-#1200 head and unaffected by this change.
- **Fully-unparseable / empty scope** (#1201's degenerate case): `canonical_pairs` is empty, so the new early-return *does* fire and no repo is queried — but this is mathematically correct, not silent: an empty canonical set cannot match any PR's closing references regardless of how many repos are queried, so skipping the (guaranteed-empty-result) `gh pr list` calls changes nothing about the final output, only the cost. This mirrors `_reconciliation_warning_from_claims`'s pre-existing `if not canonical: return None` guard (already present at `d7024bf7`, unrelated to this commit) — the two functions are now consistent with each other rather than one having a stricter, wrong per-repo test.

Docstring update accurately describes the new contract and credits main#1200 to both reviewers who found it.

### `.claude/lib/tests/test_wave_status.py`

`test_repo_with_no_canonical_rows_is_skipped` (asserted the bug as intended behavior) is correctly replaced/inverted, not just supplemented. `test_cross_repo_reference_from_a_repo_with_zero_own_scope_rows` reproduces my exact repro shape from the prior review round (repo with zero own canonical rows, one PR closing a genuine cross-repo row) as a first-class regression test. `test_wave_with_no_canonical_scope_at_all_skips_every_repo` pins the one remaining legitimate skip condition and asserts zero `gh` calls of any kind.

## Independent verification performed (new commit only — prior round's claims already reproduced and not repeated)

- Isolated `git worktree add` at `9df3efc8` (no checkout in the parent).
- `test_wave_status.py`: **37 passed**. Full `.claude/lib/tests/`: **958 passed, 19 subtests passed**. `ruff check` + `mypy`: clean on both changed files.
- **Mutation C** (restored the old per-repo guard verbatim): reproduced exactly — `test_repo_with_no_canonical_rows_of_its_own_is_still_queried` and `test_cross_repo_reference_from_a_repo_with_zero_own_scope_rows` fail, 35 others pass.
- **Mutation A re-verified against the new head** (intersection reverted to per-repo-only, adapted for the new code shape): still fails 2 tests as expected (`test_cross_repo_closing_reference_no_undercount_or_misattribution` plus the new `test_cross_repo_reference_from_a_repo_with_zero_own_scope_rows`) — confirms Mutation A is not accidentally masked/made equivalent by the #1200 fix.
- Re-ran `counters` against the live `cross-repo-status.json` for wave 29 at the new head: `final_pr_count=11`, `changes_requested_cycles=9`, `top_concentration_pct=18`, 24/36 unclaimed — unchanged from the prior round, as expected (neither in-scope repo currently has zero rows of its own, so #1200 doesn't move today's numbers, only closes the latent gap).
- Read #1201 directly: it is correctly scoped (denominator shrinkage on unparseable rows; `first:100` page truncation with no signal) and correctly disposed as tech-debt not blocking this PR — verified `first:101` really does return `EXCESSIVE_PAGINATION` is asserted there, not re-derived by me this round since it's #1201's claim, not #1198's.

My prior-round finding is resolved. No new blocking issues found in this delta.

---
Requestor: Wanjiku Mwangi
Requestee: Santiago Ferreira
RequestOrReplied: Approved
TechDebt: #1201

parametrization:
## Merge-gate re-review — PR #1198, cycle 2

New head `9df3efc89f429d02f6d8c4b8cb1915fd63f47001`, reviewed from an isolated clone at that SHA (both changed files byte-diffed against `gh api tarball/9df3efc8…`). Parent repo never checked out. Delta scope only — cycle 1's findings on `d7024bf7` are not re-litigated below except where the new commit touches them.

---

### Pre-flight — verified, not inherited

| Claim | How verified | Result |
|---|---|---|
| Genuinely additive | `git merge-base --is-ancestor d7024bf7 9df3efc8` | true — so a `d7024bf7..9df3efc8` diff is a sound delta basis, no `patch-id` reconstruction needed |
| No force-push | `issues/1198/timeline` filtered to `head_ref_force_pushed` | 0 — the cycle-1 head-SHA anchor survived intact |
| Commit count | `pulls/1198/commits` | 2, the second being `9df3efc8 fix(wave-status): gate the direct-to-main listing skip on wave scope, not per-repo scope (#1200)` |
| CI at new head | `statusCheckRollup` grouped by conclusion | `{"SUCCESS": 14}` — 14 checks, 0 failing, 0 pending |
| Mergeability | `gh pr view` immediately before posting | `MERGEABLE` / `CLEAN` |

I ran each of these rather than carrying them over. Delta touches only `.claude/lib/wave_status.py` and `.claude/lib/tests/test_wave_status.py`.

---

### Is #1200 discharged, or relocated a third time?

Discharged. The distinction matters here because this defect has already moved once — cycle 1 fixed the *matching* step and left it live at the *listing* step, which is why two reviewers independently caught it. So I did not check that the symptom went away; I enumerated every remaining point at which a PR can be dropped before the intersection runs, and asked of each whether the code is making a narrowing decision of its own.

The per-repo guard is **removed**, not narrowed — `issue_numbers_by_repo` no longer gates the loop at all, and every repo in `repos_in_scope` is now listed. The single remaining skip is wave-level:

```python
if not canonical_pairs:
    return out, claimed
```

**I proved this one is behaviour-neutral rather than assuming it.** Two independent arguments, because "the new guard is narrower" is exactly the reassurance that was true in cycle 1 and still wrong:

1. *Structurally.* `canonical_pairs` is the flatten of `issue_numbers_by_repo`, and `_canonical_issue_numbers_by_repo` builds entries only via `setdefault(repo, set()).add(...)`, so it can never hold an empty-set entry. Therefore `canonical_pairs` empty ⟺ `issue_numbers_by_repo` empty ⟺ every repo would have been skipped by the old loop anyway. It short-circuits only the case where no pair exists **anywhere** for any PR to match — the one condition under which no drop is possible by construction.
2. *Empirically.* Mutation D (mine, not in the PR body) — deleted the early return outright: **1 failed, 36 passed**, and the sole failure is `test_wave_with_no_canonical_scope_at_all_skips_every_repo` asserting `fake.calls == []`. The `assertEqual(got, [])` on the line above it still passes. The guard's *only* observable effect is call count, never output. That is what "pure cost optimisation" has to look like to be believed.

My own cycle-1 repro, re-run against this head — a child repo with zero scope rows of its own delivering the wave's only canonical row, recorded under `noorinalabs-main`:

```
[R1] merged = [77]        (cycle 1 at d7024bf7 returned [])
```

Remaining drop points audited and cleared: the `merged_at < kickoff` filter and the `--search merged:>=` bound are correct by design and already fixture-proven against the `--limit` cap; the intersection is the repo-qualified one. One further point exists and is *not* a code defect — see the ruling below.

### `.claude/lib/wave_status.py`

The docstring rewrite is accurate to the code, and it records *why* the per-repo gate was wrong rather than merely that it changed — including the cost objection it was originally justified by, and the correct answer to it (a repo that owns no canonical rows simply has every PR fail the intersection, which is no different from any other non-matching PR). Attribution to both reviewers is in the source comment; I have no objection to it, though the finding matters more than the names.

### `.claude/lib/tests/test_wave_status.py`

`test_repo_with_no_canonical_rows_is_skipped` is **inverted, not deleted** — it now asserts the scopeless repo *is* queried, under a name that says so, with a docstring recording that the old assertion was itself the bug. That is the right disposition: silently dropping a test that encoded a wrong contract would have erased the evidence that the contract was ever wrong.

`test_cross_repo_reference_from_a_repo_with_zero_own_scope_rows` is the exact shape I reproduced in cycle 1 and filed as #1200, now pinned in-suite. `test_wave_with_no_canonical_scope_at_all_skips_every_repo` pins the new guard on both axes — output empty *and* `fake.calls == []`.

### Mutation testing — Mutation C reproduced, and Mutation A is now caught harder

*Mutation C* — restored the pre-#1200 per-repo guard verbatim: **2 failed, 35 passed**, exactly `test_cross_repo_reference_from_a_repo_with_zero_own_scope_rows` and `test_repo_with_no_canonical_rows_of_its_own_is_still_queried`. Reproduced as reported. Non-equivalent.

*Mutation A* — re-run against the new guard, reverting the intersection to a per-repo set: **2 failed, 35 passed**, now tripping the new zero-own-rows test *in addition to* the cycle-1 cross-repo test. Stronger than the 1-failure result at `d7024bf7`; the two fixes reinforce each other's guards rather than merely coexisting.

*Mutation D* — as above, 1 failed / 36 passed, cost-only.

Source restored byte-identically after each. **37/37** in-file, **958/958** across `.claude/lib/tests/` (+19 subtests). `ruff format --check`, `ruff check`, `mypy` clean on both files.

### Ruling on #1201 — remains correctly out of scope

It does not interact with the new guard, and I checked rather than assumed, because the wholly-unparseable shape is precisely the input that reaches the new early return:

- **Partial** (4 declared rows, 2 parseable): `WARNING: … main#1160 (1 of 2)` — identical to `d7024bf7`. Unaffected.
- **Wholly unparseable** (now routed through the new early return): `final_pr_count: 0`, rc 0, no stderr — also identical to `d7024bf7`, as Mutation D independently predicts.

So the commit neither worsens nor improves #1201, which is the right outcome for a fix scoped to the listing step.

It should stay debt, and the reason is a real distinction rather than a scoping convenience. **#1200 was well-formed input plus a wrong narrowing decision by the code — that is a correctness bug and rightly blocked cycle 1. #1201 is malformed input plus silence — the row genuinely carries no repo or number to key on, so the fix is a *signal*, not a behaviour change.** I also went looking for reachability instead of asserting it: across **every** `wave_*_scope` ever recorded in `cross-repo-status.json`, **278 of 278** tier rows parse. Zero unparseable rows in the file's entire history, so this needs a hand-edit or a legacy record to reach at all.

Two things I will say plainly rather than bury. With #1200 closed, #1201 is now the **only** remaining path by which this instrument can under-count without saying so — that raises its priority for the next wave even though it does not block this PR. And I found a third edge of the same family during this pass, which I appended to #1201 rather than filing separately: a canonical row whose repo is absent from `repos_in_scope` is permanently unclaimable, so it appears as unclaimed on every run even when delivered (`user-service#500` reproduced at this head). That one is *not* a relocation of #1200 — the code there iterates exactly what it was declared, and it is the record that is inconsistent between two of its own keys, so the fix is again a signal and not a silent override of a declared field. Live wave-29 satisfies `{repos named in scope rows} ⊆ repos_in_scope`, so it is latent too.

### Live verification at the new head

```
final_pr_count 11, changes_requested_cycles 9, top_concentration_pct 18
WARNING: scope rows with no matching merged PR: main#1114, …, main#1195 (24 of 36)
```

Unchanged from `d7024bf7` — the fix is a no-op against today's data, as expected, since both wave-29 in-scope repos own canonical rows. stdout re-confirmed as pure JSON on `counters` and on `merged-prs`; the warning stays on stderr. The live record is self-consistent: 36 of 36 rows parse, and both scope-row repos are in `repos_in_scope`.

Same caveat as cycle 1, restated so it does not quietly become evidence it is not: every in-window merged PR in wave-29 happens to be in scope, so the live run is a positive control that the instrument does not silently zero out — not a discriminating test of the scope filter. The exclusion is proven by the fixtures.

---

### Summary

The blocking finding is genuinely closed, not moved a third time: the per-repo gate is gone rather than narrowed, the one remaining skip is proven behaviour-neutral both structurally and by a mutation that changes call count and nothing else, and my own cycle-1 repro now returns the right answer. Mutation C reproduced exactly as reported, and Mutation A is caught by strictly more tests than before. The wrong-contract test was inverted rather than deleted, which preserves the evidence. The two residual findings are the same family one level further out, both latent, both signal-shaped rather than correctness-shaped, and both recorded on #1201 with reachability evidence rather than assertion.

Approving. Not merging — the owner's go gates that, and the gate passing does not change it.

---
Requestor: Weronika Zielinska
Requestee: Santiago Ferreira
RequestOrReplied: Approved
TechDebt: #1201

Issue #1203, opened 2026-07-30
tech-debt(hooks-tests): five bare assertIn("1/2", reason) assertions in Hook 4 s suite are inert — the BLOCK help text embeds "1/2 false-block" three times


parametrization:
Surfaced by the merge-gate review of PR #1199 (head `b8bf343f`). #1199's own review carries the one instance it introduced; this issue is the five pre-existing siblings and the class.

## The instrument lies to its own tests

`validate_pr_review.check()`'s peer-review BLOCK message unconditionally appends a help block listing the P3W11 batch-11 historical instances (`main#509`, `deploy#337`, `deploy#339`), each line ending in the literal text `1/2 false-block`. That text is emitted for **every** peer-review BLOCK regardless of the computed reviewer ratio.

Any test that asserts the ratio with a bare `assertIn("1/2", result["reason"])` therefore matches the boilerplate, not the count. The assertion degenerates to "some BLOCK happened" — which the preceding `assertIsNotNone` / `assertEqual(decision, "block")` already established.

## Measured (PR #1199 head `b8bf343f`, hermetic tree, `check()` driven end to end)

Fixture: `reviewers = {"annunaki", "nikolaos papadopoulos"}` against a parent+child card tree plus a manifest carrying `Nikolaos Papadopoulos` and `Annunaki`.

| build | real ratio | occurrences of `1/2` in `reason` | `assertIn("1/2", reason)` |
|---|---|---|---|
| head `b8bf343f` | **1/2** | 4 (3 boilerplate + 1 real) | passes |
| head with `_load_roster_names`' manifest union reverted | **0/2** | 3 (all boilerplate) | **still passes** |

A second control — 1 roster reviewer + 3 non-roster — also has a real ratio of `1/2`, and the bare assertion cannot distinguish that case either.

## Sites (line numbers at `b8bf343f`)

```
L934   CheckEndToEndTests::test_one_reviewer_without_wave_bootstrap_blocks
L1110  CheckEndToEndTests::test_wave_merge_head_ref_blocks_with_one_reviewer
L1321  RosterValidationGateTests::test_mixed_roster_and_non_roster_blocked
L1750  ChildRosterResolutionTests::test_child_pr_unknown_reviewer_blocks
L1775  ChildRosterResolutionTests::test_child_reviewer_does_not_count_on_parent_pr
```

[...]

Issue #1175, opened 2026-07-28
tech-debt(hooks): validate_review_comment_format still rejects dash-separator branches — divergent from validate_pr_review since #179


parametrization:
## The gap

`extract_branch_author_lastname` is independently defined in **both** hooks and they no longer agree:

- `validate_pr_review.py` (fixed by #179): accepts `[A-Za-z]\.([A-Za-z]+)[-/]` — both `/` and `-` separators, because "many recent branches use dash".
- `validate_review_comment_format.py`: still `[A-Za-z]\.([A-Za-z]+)/` — slash only. `test_dash_separator_not_supported` (added for #302) explicitly **pins** this as intended: *"Branches using dash form fall through to the `branch_author = None` path and the hook allow-with-warning (does not block)."*

So on a dash-separated branch, `validate_review_comment_format`'s entire Requestor/Requestee swap check is silently skipped (early `"decision": "allow"` with only a `systemMessage` warning) — before `charter_trailer.is_branch_author` (main#1172/#1173) is ever reached. `charter_trailer.branch_author_first_initial`, which the format hook also imports, *does* accept dash separators — so the format hook now contains two branch-prefix parsers reading the same `head_ref` that disagree on what counts as a match, one of which (the local, un-consolidated one) always wins because it runs first and short-circuits.

## Why this is tech-debt, not a merge-gate hole

Confirmed by reading both current call sites at PR #1173 head (`3be5b027`):

- `validate_pr_review`'s own `extract_branch_author_lastname` supports both separators, so the **counting/self-review-exclusion gate is unaffected** — a swapped verdict comment on a dash-branch would still be correctly excluded from the 2-reviewer count.
- No currently-open branch (136 checked against the `{Initial}.{Lastname}[-/]` prefix) uses the dash form — the risk is latent, not active, today.
- What's lost is only the format hook's advisory *block-before-post* UX check on that branch-naming variant.

## Why it's worth fixing

[...]

PR #1207, opened 2026-08-01
fix(hook4): scan PR comment verdicts for every head ref, and make a skipped scan visible

### LINKED ISSUE #1206: bug(validate_pr_review): comment verdict scan is skipped entirely for non-persona head refs — dependabot PRs can never reach 2/2, and the report blames the reviewers
## Summary

`resolve_review_verdicts` in `.claude/hooks/validate_pr_review.py` runs the PR-comment
verdict scan **only** for two head-ref shapes. For every other shape it leaves
`comment_result` at its empty default and returns **zero comment reviewers with no error
and no diagnostic** — indistinguishable from "scanned the thread and nobody approved."

The guard, in `resolve_review_verdicts`:

```python
comment_result = CommentReviewResult()      # empty default
branch_author_lastname = None
if head_ref:
    branch_author_lastname = extract_branch_author_lastname(head_ref)
    if branch_author_lastname:                       # persona branch: {Initial}.{Lastname}/...
        comment_result = check_comment_reviews(...)
    elif head_ref.startswith("deployments/") and "/wave-" in head_ref:   # wave-merge branch
        comment_result = check_comment_reviews(number, "", repo=repo, content_ts=content_ts)
# any other head ref -> comment_result stays EMPTY, silently
```

Measured on the three live shapes:

| head ref | comment scan runs? |
|---|---|
| `S.Ferreira/1189-wave-status-scoping` | yes (lastname `Ferreira`) |
| `deployments/phase-10/wave-29` | yes (wave-merge sentinel) |
| `dependabot/docker/integration-tests/fake_oauth/python-d3400aa` | **NO** |

## Consequence

**Any PR whose head ref is neither a persona branch nor a wave-merge branch can never
satisfy the 2-reviewer gate**, regardless of how many valid Approved verdicts it carries.
Bot-authored PRs (Dependabot) are the standing case; so is any hand-made branch that
does not follow `{Initial}.{Lastname}/{IIII}-{name}`.

## Live evidence — noorinalabs-deploy#691

Lucas Ferreira posted a correctly-formed verdict
(https://github.com/noorinalabs/noorinalabs-deploy/pull/691#issuecomment-5152002350):
trailer after the last sole `---`, exact field names, `Approved`, `TechDebt: none`,
space-form roster name that is present in the resolved roster union.

The counting primitive finds it; the resolver does not:

```
check_comment_reviews("691", "",   repo=..., content_ts=T) -> reviewers={'lucas ferreira'}
check_comment_reviews("691", None, repo=..., content_ts=T) -> reviewers={'lucas ferreira'}
resolve_review_verdicts(pr_data,   repo=...)               -> comment_reviewers=set()
```

`pr_review_state.py` accordingly reports `0/2 required — (none)` with `stale verdicts: none`.
Both sub-claims of that report are false: there is one Approved verdict and it is not stale.

## Direction of failure

**Fail-CLOSED for merging** — it blocks, it cannot wave a merge through, so this is not a
security hole. It is a *silent zero*: the report attributes the shortfall to the reviewers
("nobody approved") when the truth is that the scan never ran. Cf.
`feedback_silent_zero_is_not_a_measurement` — a review that cannot be counted is
indistinguishable from no review.

Note the asymmetry with the sibling guard in `check_comment_reviews` itself, whose
docstring explicitly reasons about `branch_author_initial` omission being safe *because*
it over-excludes in the fail-closed direction. The same care was not applied one level up:
here the fallback is not "over-exclude", it is "do not measure at all."

## Proposed fix

The wave-merge branch already establishes the correct sentinel: when there is no
identifiable branch author, call `check_comment_reviews(number, "", ...)` — the empty
sentinel admits any non-empty reviewer, and roster filtering downstream still applies. Make
that the **general fallback** rather than a special case for `deployments/**`:

```python
if head_ref:
    branch_author_lastname = extract_branch_author_lastname(head_ref)
    comment_result = check_comment_reviews(
        number,
        branch_author_lastname or "",
        repo=repo,
        content_ts=content_ts,
        branch_author_initial=branch_author_first_initial(head_ref) if branch_author_lastname else "",
    )
```

This preserves author-exclusion wherever an author IS identifiable and stops silently
skipping the scan where one is not. The author-exclusion loss on a bot branch is correct:
a bot is not a roster persona, so it can never have been in the reviewer set anyway.

Nothing here should relax the gate. It should still take 2 distinct roster Approveds — the
fix is that they must be *counted* when they exist.

## Tests to add

- A `dependabot/**` head ref with two valid Approved comments reaches 2/2.
- A head ref with no `/` at all does not crash and scans.
- Regression: a persona branch still excludes the PR author from the reviewer set.
- Regression: `deployments/phase-N/wave-M` behaviour is unchanged.

parametrization:
Closes #1206

## The defect, confirmed

Every claim in #1206 reproduces exactly. Measured before the change, live:

```
$ python3 .claude/lib/pr_review_state.py 691 --repo noorinalabs/noorinalabs-deploy
PR #691 (noorinalabs/noorinalabs-deploy) — merge gate would BLOCK
  head ref: dependabot/docker/integration-tests/fake_oauth/python-d3400aa
  distinct Approved reviewers: 0/2 required — (none)
  stale verdicts: none
```

while the counting primitive, same PR, same `T_content`, called with the sentinel
the wave-merge path already uses, returns `['lucas ferreira', 'nino kavtaradze']`
— both roster-valid, neither stale, both carrying `TechDebt`. `resolve_review_verdicts`
dispatched the comment scan for exactly two head-ref shapes and left
`comment_result` at its empty default for everything else, so the PR reported a
silent zero and blamed the reviewers for it.

## What changed

**1 — the skip.** `comment_scan_scope(head_ref)` is **total**: it returns a scanning
*mode* for every possible head ref and never "do not scan". The head ref's only role
in the comment scan is naming the branch author for the self-review exclusion, and
`check_comment_reviews` already has an honest value for "no identifiable author" — the
`""` sentinel `charter_trailer.is_branch_author` reads as "nobody is the author"
(main#294). Totality is deliberately the design point rather than adding `dependabot/**`
to a prefix list: the defect is the fall-through, and a longer list of recognised
prefixes fails the same way on the next unrecognised shape.

**This does not relax the gate.** Still 2 distinct roster-valid *current* Approveds.
Roster filtering (#498), content-staleness binding (#950), the latest-verdict rule
(#940) and the TechDebt attestation are all untouched, and each is pinned by a test on
the bot-branch path specifically. Author-exclusion is unchanged wherever the ref names a
persona; losing it on a bot branch is correct — a bot is never a roster persona, so it
could never have been in the reviewer set.

**2 — the misleading report.** This half matters as much, because it is *why* the first
defect survived: `0/2 required — (none)` and `stale verdicts: none` are both sentences
about what a scan found, and neither was. `ReviewVerdicts.comment_scan` /
`ReviewState.comment_scan` now record which mode ran. Both **default to `not-run`**, so a
construction that forgets the field degrades to the honest "not measured" rather than to a
false claim of measurement. `pr_review_state` prints the scan's status *before* the counts
and refuses to print a bare `stale verdicts: none` when nothing was scanned; `check()`'s
block message distinguishes the two causes of the same shortfall. As defence in depth the
resolver raises `CommentScanUndeterminedError` — the #981 stance, applied one level up —
rather than returning a zero, if a future edit ever reintroduces a skip path.

## After

```
PR #691 (noorinalabs/noorinalabs-deploy) — merge gate would PASS
  comment verdict scan: RAN — head ref names no branch author (bot / wave-merge /
    non-charter branch), so no self-review exclusion was applied. Roster filtering
    and the 2-reviewer threshold are unchanged.
```

Live regression control on a persona branch (main#1199) is unchanged: author exclusion
active, `Kavtaradze` named, and its two stale `ChangesRequested` verdicts still excluded.

## Tests — mutation-verified, not merely green

Restoring the pre-fix dispatch turns **21 assertions red**; collapsing the report's
not-measured branch turns 1 red; dropping the scan-mode propagation turns 1 red. Where a
test asserts an absence it is paired with a positive control proving the experiment ran
(cf. #1203, where five assertions matched unconditional boilerplate and passed at a real
0/2).

| Test | What it catches |
|---|---|
| `CommentScanScopeTotalityTests` (3, subtested over 17 refs) | Any head-ref shape that dispatches to no scan — the defect class, not the instance. `dependabot/**`, `renovate/**`, `feature/x`, no-slash, `""`, `/`, `..` |
| `…::test_persona_branch_selects_author_exclusion` | Positive control: a function returning `no-branch-author` unconditionally would satisfy totality but fails here |
| `NonPersonaHeadRefScanTests::test_dependabot_head_ref_with_two_approvals_reaches_two_of_two` | The live defect, offline: 2/2 on a `dependabot/**` ref |
| `…::test_head_ref_with_no_slash_neither_crashes_nor_skips` | A ref with no separator falling through both old branches |
| `…::test_empty_head_ref_still_scans` / `…::test_none_head_ref_does_not_crash` | Missing `headRefName` skipping, or crashing `re.match` |
| `…::test_non_roster_requestor_on_a_bot_branch_still_does_not_count` | Widened sentinel smuggling a non-roster Requestor past #498 |
| `…::test_changes_requested_on_a_bot_branch_still_does_not_count` | #940 latest-verdict rule bypassed on the new path |
| `…::test_stale_verdict_on_a_bot_branch_is_still_excluded` | #950 content binding bypassed on the new path |
| `HeadRefScanRegressionTests::test_persona_branch_still_excludes_the_pr_author` | Self-review exclusion lost — the one real relaxation risk. Admitted reviewer is the positive control |
| `…::test_persona_branch_still_admits_the_same_surname_colleague` | #1172 regression (Santiago vs Lucas Ferreira) |
| `…::test_wave_merge_branch_behaviour_is_unchanged` | `deployments/phase-N/wave-M` behaviour drift |
| `…::test_two_reviewer_threshold_is_not_relaxed_anywhere` | "count what exists" sliding into "need fewer" |
| `CommentScanNotMeasuredTests::test_review_verdicts_defaults_to_not_measured` | A default that falsely claims the scan ran |
| `…::test_a_reintroduced_skip_path_fails_closed_with_a_named_reason` | A future skip path returning a silent zero (exercised by patching the scope fn, not asserted by comment) |
| `…::test_positive_control_same_setup_does_not_raise_unpatched` | Proves the `assertRaises` above isn't passing on harness breakage |
| `…::test_block_message_distinguishes_not_measured_from_measured_empty` | Hook-surface wording asserted in BOTH directions |
| `CommentScanReportedInTheReportTests` (7) | Report-surface distinction: differing renders on identical reviewer data, `stale verdicts: none` claimed only when scanned (with the reachability positive control), the third mode named, `--json` carrying the field, and `compute_review_state` actually propagating it (without which the field is inert in production) |

`test_content_ts_forwarded_on_wave_merge_call_site` was renamed and its stub widened —
there is no longer a separate wave-merge call site, only one. Its assertions are kept and
strengthened (`branch_author_initial` must be `""` there).

## CI

All local gates green, full parity: `ruff-format`, `ruff-lint`, `mypy`, `pytest`
(3270 passed / 607 subtests), `pytest-skills`, `cspell`, `actionlint`, `checksums-ascii`,
`fixture-realism`, `memory-budget`, `headcount-budget`, `office-drift`, `mermaid-render`,
and `pre_commit_ci_sync.py` (`OK: pre-commit config mirrors all CI-enforced checks`).

## Two findings filed separately, out of scope here

1. **Sibling-hook divergence** — `validate_pr_review.extract_branch_author_lastname`
   accepts `[-/]`; `validate_review_comment_format`'s copy accepts `/` only. On a
   dash-form branch the gate identifies the author and the format hook does not. Same
   family as #1175 / #932. Not fixed here as a drive-by.
2. **Ontology overlay inaccuracy** — `ontology/conventions.md:240` describes this hook as
   "Enforce charter review comment format"; it is the 2-reviewer merge gate (that
   description belongs to the sibling hook). `ontology/**` is out of scope for this PR.

### DIFF
--- a/.claude/hooks/tests/test_validate_pr_review.py
+++ b/.claude/hooks/tests/test_validate_pr_review.py
@@ -3814,5 +3814,401 @@ class SurnameCollisionSelfReviewTests(unittest.TestCase):
         self.assertEqual(captured.get("branch_author_initial"), "l")
 
 
+class CommentScanScopeTotalityTests(unittest.TestCase):
+    """#1206: `comment_scan_scope` must return a SCANNING mode for every head ref.
+
+    The defect was a dispatch that recognised exactly two head-ref shapes and
+    fell through to no scan at all for everything else — so a `dependabot/**`
+    PR carrying two correctly-formed, roster-valid, non-stale Approved verdicts
+    reported `0/2 required — (none)`, blaming the reviewers for a scan that
+    never ran (noorinalabs-deploy#691).
+
+    Totality is the property that kills the whole defect class rather than the
+    one observed shape: it is not enough to add `dependabot/**` to the list of
+    recognised prefixes, because the next unrecognised shape fails the same way.
+    """
+
+    def test_persona_branch_selects_author_exclusion(self):
+        """Positive control for the discriminator below: the persona shapes must
+        actually resolve to the OTHER mode, or the totality assertions would be
+        satisfied by a function that returned NO_BRANCH_AUTHOR unconditionally.
+        """
+        for ref in (
+            "S.Ferreira/1189-wave-status-scoping",
+            "A.Virtanen/1206-review-scan-nonpersona-headref",
+            "A.Virtanen-0179-branch-regex-fix",  # dash separator, also charter-shaped
+        ):
+            with self.subTest(ref=ref):
+                self.assertEqual(hook.comment_scan_scope(ref), hook.COMMENT_SCAN_AUTHOR_EXCLUDED)
+
+    def test_non_persona_refs_select_the_no_branch_author_scan(self):
+        """THE DEFECT, at the dispatch level. Every one of these returned "no
+        scan" before #1206; each must now name a real scanning mode."""
+        for ref in (
+            "dependabot/docker/integration-tests/fake_oauth/python-d3400aa",
+            "deployments/phase-10/wave-29",
+            "feature/some-hand-made-branch",
+            "nohashinthisref",  # no `/` at all
+            "",  # headRefName absent from the API response
+            "renovate/pytest-8.x",
+            "1206-bare-issue-number-branch",
+        ):
+            with self.subTest(ref=ref):
+                self.assertEqual(hook.comment_scan_scope(ref), hook.COMMENT_SCAN_NO_BRANCH_AUTHOR)
+
+    def test_no_head_ref_shape_ever_yields_not_run(self):
+        """The invariant itself, stated once: NOT_RUN is not a reachable scope.
+
+        Kept separate from the case list above so the property survives someone
+        later adding a third scanning mode — a new mode would keep this green
+        and correctly fail only the specific-mode assertions it changes.
+        """
+        for ref in (
+            "",
+            "/",
+            "a",
+            "A.Virtanen/1206-x",
+            "dependabot/npm_and_yarn/x-1.2.3",
+            "deployments/phase-10/wave-29",
+            "..",
+            "x" * 300,
+        ):
+            with self.subTest(ref=ref):
+                self.assertNotEqual(hook.comment_scan_scope(ref), hook.COMMENT_SCAN_NOT_RUN)
+
+
+class _ResolveOverFakeCommentsHarness(unittest.TestCase):
+    """Drives the REAL `resolve_review_verdicts` over a faked `gh api` boundary.
+
+    Deliberately does NOT mock `check_comment_reviews`: the #1206 defect lived
+    in the resolver's DISPATCH to that function, so a test that stubs the callee
+    still exercises the dispatch, but a test that stubs the dispatch would prove
+    nothing at all.
+    """
+
+    REPO = "noorinalabs/noorinalabs-main"
+    ROSTER = {"lucas ferreira", "nino kavtaradze", "aino virtanen", "santiago ferreira"}
+
+    @staticmethod
+    def _verdict(requestor: str, requestee: str = "Someone Else", direction: str = "Approved"):
+        return {
+            "body": (
+                f"Requestor: {requestor}\nRequestee: {requestee}\n"
+                f"RequestOrReplied: {direction}\nTechDebt: none"
+            ),
+            "created_at": "2026-07-20T00:00:00Z",
+        }
+
+    @staticmethod
+    def _pr_data(head_ref: str, *, author="parametrization", reviews=(), labels=()):
+        return {
+            "author": author,
+            "number": 691,
+            "reviews": list(reviews),
+            "headRefName": head_ref,
+            "labels": list(labels),
+        }
+
+    def _resolve(self, head_ref: str, comments: list[dict], *, roster=None):
+        def fake_run(args, capture_output, text, timeout):  # noqa: ARG001
+            result = mock.MagicMock()
+            result.returncode = 0
+            if args[0] == "gh" and args[1:3] == ["repo", "view"]:
+                result.stdout = json.dumps({"owner": {"login": "noorinalabs"}, "name": "r"})
+            else:
+                result.stdout = json.dumps(comments)
+            return result
+
+        with (
+            mock.patch.object(hook.subprocess, "run", side_effect=fake_run),
+            mock.patch.object(
+                hook,
+                "get_latest_content_commit",
+                # T_content sits BEFORE the fixture comments, so nothing is stale
+                # and a shortfall can only come from the scan, not from staleness.
+                return_value=("837c272a", hook._parse_iso8601("2026-07-15T19:13:59Z")),
+            ),
+            mock.patch.object(
+                hook,
+                "_load_roster_names",
+                return_value=set(self.ROSTER if roster is None else roster),
+            ),
+        ):
+            return hook.resolve_review_verdicts(self._pr_data(head_ref), repo=self.REPO)
+
+
+class NonPersonaHeadRefScanTests(_ResolveOverFakeCommentsHarness):
+    """#1206: the comment verdict scan must run for EVERY head-ref shape.
+
+    Each assertion below is a kill-shot on the pre-fix resolver: it returned
+    `comment_reviewers == set()` and `total_distinct == 0` for all of these
+    inputs, so every `assertEqual(..., 2)` here was `0` before the fix.
+    """
+
+    DEPENDABOT_REF = "dependabot/docker/integration-tests/fake_oauth/python-d3400aa"
+
+    def test_dependabot_head_ref_with_two_approvals_reaches_two_of_two(self):
+        """THE LIVE DEFECT, reproduced offline (noorinalabs-deploy#691)."""
+        verdicts = self._resolve(
+            self.DEPENDABOT_REF,
+            [self._verdict("Lucas Ferreira"), self._verdict("Nino Kavtaradze")],
+        )
+        self.assertEqual(verdicts.distinct_reviewers, {"lucas ferreira", "nino kavtaradze"})
+        self.assertEqual(verdicts.total_distinct, 2)
+        self.assertEqual(verdicts.comment_scan, hook.COMMENT_SCAN_NO_BRANCH_AUTHOR)
+
+    def test_head_ref_with_no_slash_neither_crashes_nor_skips(self):
+        """A ref with no separator at all used to fall through both branches."""
+        verdicts = self._resolve(
+            "nohashinthisref",
+            [self._verdict("Lucas Ferreira"), self._verdict("Nino Kavtaradze")],
+        )
+        self.assertEqual(verdicts.total_distinct, 2)
+        self.assertTrue(verdicts.comment_scan_ran)
+
+    def test_empty_head_ref_still_scans(self):
+        """`get_pr_data` defaults `headRefName` to `""`; that must scan, not skip."""
+        verdicts = self._resolve("", [self._verdict("Aino Virtanen")])
+        self.assertEqual(verdicts.distinct_reviewers, {"aino virtanen"})
+        self.assertTrue(verdicts.comment_scan_ran)
+
+    def test_none_head_ref_does_not_crash(self):
+        """A hand-built `pr_data` carrying `None` must degrade, not raise."""
+        pr_data = self._pr_data("")
+        pr_data["headRefName"] = None
+
+        def fake_run(args, capture_output, text, timeout):  # noqa: ARG001
+            result = mock.MagicMock()
+            result.returncode = 0
+            result.stdout = json.dumps([self._verdict("Aino Virtanen")])
+            return result
+
+        with (
+            mock.patch.object(hook.subprocess, "run", side_effect=fake_run),
+            mock.patch.object(hook, "get_latest_content_commit", return_value=None),
+            mock.patch.object(hook, "_load_roster_names", return_value=set(self.ROSTER)),
+        ):
+            verdicts = hook.resolve_review_verdicts(pr_data, repo=self.REPO)
+
+        self.assertEqual(verdicts.head_ref, "")
+        self.assertEqual(verdicts.distinct_reviewers, {"aino virtanen"})
+
+    def test_non_roster_requestor_on_a_bot_branch_still_does_not_count(self):
+        """The gate is not relaxed: widening the sentinel must not smuggle a
+        non-roster Requestor past the #498 filter. Paired with a roster name in
+        the SAME thread as the positive control — a fix that broke roster
+        filtering outright would satisfy neither half.
+        """
+        verdicts = self._resolve(
+            self.DEPENDABOT_REF,
+            [self._verdict("Lucas Ferreira"), self._verdict("Mallory Impostor")],
+        )
+        self.assertEqual(verdicts.roster_comment_reviewers, {"lucas ferreira"})
+        self.assertEqual(verdicts.non_roster_requestors, {"mallory impostor"})
+        self.assertEqual(verdicts.total_distinct, 1)
+
+    def test_changes_requested_on_a_bot_branch_still_does_not_count(self):
+        """Nor may it admit a non-Approved verdict (#940 stays in force)."""
+        verdicts = self._resolve(
+            self.DEPENDABOT_REF,
+            [
+                self._verdict("Lucas Ferreira"),
+                self._verdict("Nino Kavtaradze", direction="Changes Requested"),
+            ],
+        )
+        self.assertEqual(verdicts.distinct_reviewers, {"lucas ferreira"})
+        self.assertEqual(verdicts.total_distinct, 1)
+
+    def test_stale_verdict_on_a_bot_branch_is_still_excluded(self):
+        """Content binding (#950) must survive the widened dispatch."""
+        stale = self._verdict("Lucas Ferreira")
+        stale["created_at"] = "2026-01-01T00:00:00Z"  # before T_content
+        verdicts = self._resolve(self.DEPENDABOT_REF, [stale, self._verdict("Nino Kavtaradze")])
+        self.assertEqual(verdicts.distinct_reviewers, {"nino kavtaradze"})
+        self.assertEqual(len(verdicts.stale_verdicts_comment), 1)
+        self.assertEqual(verdicts.stale_verdicts_comment[0].reviewer, "Lucas Ferreira")
+
+
+class HeadRefScanRegressionTests(_ResolveOverFakeCommentsHarness):
+    """The two behaviours #1206 must leave EXACTLY as they were."""
+
+    def test_persona_branch_still_excludes_the_pr_author(self):
+        """Self-review exclusion is the reason the head ref is consulted at all.
+
+        Lucas Ferreira's own Approved on `L.Ferreira/…` must stay out, while
+        Aino Virtanen's is admitted — the admitted half is the positive control
+        proving the scan ran, so a regression that skipped the scan entirely
+        would fail this test rather than passing its absence assertion.
+        """
+        verdicts = self._resolve(
+            "L.Ferreira/1206-x",
+            [self._verdict("Lucas Ferreira"), self._verdict("Aino Virtanen")],
+        )
+        self.assertNotIn("lucas ferreira", verdicts.distinct_reviewers)
+        self.assertEqual(verdicts.distinct_reviewers, {"aino virtanen"})
+        self.assertEqual(verdicts.branch_author_lastname, "Ferreira")
+        self.assertEqual(verdicts.comment_scan, hook.COMMENT_SCAN_AUTHOR_EXCLUDED)
+
+    def test_persona_branch_still_admits_the_same_surname_colleague(self):
+        """#1172 must not regress: Santiago is not Lucas."""
+        verdicts = self._resolve(
+            "L.Ferreira/1206-x",
+            [self._verdict("Santiago Ferreira"), self._verdict("Lucas Ferreira")],
+        )
+        self.assertEqual(verdicts.distinct_reviewers, {"santiago ferreira"})
+
+    def test_wave_merge_branch_behaviour_is_unchanged(self):
+        """`deployments/phase-N/wave-M` counted every roster reviewer before the
+        fix (main#294's empty sentinel) and must still do so — same reviewers,
+        same count, and now under the generalised no-branch-author mode."""
+        verdicts = self._resolve(
+            "deployments/phase-10/wave-29",
+            [self._verdict("Lucas Ferreira"), self._verdict("Nino Kavtaradze")],
+        )
+        self.assertEqual(verdicts.distinct_reviewers, {"lucas ferreira", "nino kavtaradze"})
+        self.assertEqual(verdicts.total_distinct, 2)
+        self.assertEqual(verdicts.comment_scan, hook.COMMENT_SCAN_NO_BRANCH_AUTHOR)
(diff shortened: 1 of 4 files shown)


parametrization:
## Reviewer note from the author — one place this is NOT strictly non-relaxing

The PR description says the change does not relax the gate. That is true for the defect it
fixes and for every test in it, but I want it on the record rather than discovered in
review, because it is the one thing a reviewer should weigh deliberately:

On a head ref with **no persona prefix**, the gate cannot identify the branch author, so
self-review exclusion is unavailable. For a **human** persona on a non-charter branch who
posts their own `Approved` comment alongside one genuine reviewer, this goes from
`0/2 blocked` (wrong — the genuine reviewer was uncounted too) to `2/2 passes` (also
wrong — correct is `1/2 blocked`).

Why I shipped it rather than holding the fix:

- Pre-existing in kind — `deployments/**` has carried exactly this since main#294. This
  widens the population of refs, not the principle.
- The bot case #1206 is actually about is unaffected: a bot is never a roster persona, so
  it can never enter the reviewer set.
- It requires a deliberate self-approval, a charter violation on its own.
- It is now **visible** — both the report and the block message state "no self-review
  exclusion was applied" in exactly these cases. Before, the exposure *and* the legitimate
  reviewers were both invisible.

Filed as #1210 with the full analysis and a concrete fix: the persona name is already in
the commit payload `get_latest_content_commit` fetches (`commit.author.name`, currently
discarded), so the author can be derived from commit identity — which would also close the
pre-existing #294 hole. It needs its own design decisions (multi-author branches,
non-roster commit authors, return-type change on a heavily-shared function), which is why
it is not folded in here.

Reviewers: if you judge that residual unacceptable to ship ahead of #1210, say so and I
will hold this PR until they land together.

parametrization:
## Second-slot review — structural / cross-repo lens

Verified independently: head SHA `a44ea958a108b9d1709ec272a0549abacdfa4d41` matches PR HEAD; `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`; all 14 CI checks report SUCCESS against this SHA. Read the diff via `gh api .../contents/<path>?ref=<sha>` for all four changed files (no local checkout used).

### 1. Sibling-hook divergence (#1208) — real, but not worsened by this PR

Confirmed independently against the fetched file contents:

- `validate_pr_review.py::extract_branch_author_lastname` (`.claude/hooks/validate_pr_review.py:745`) — `r"[A-Za-z]\.([A-Za-z]+)[-/]"` — accepts dash **and** slash.
- `charter_trailer.py::branch_author_first_initial` (`.claude/lib/charter_trailer.py:252`) — same `[-/]` acceptance, used by both hooks.
- `validate_review_comment_format.py::extract_branch_author_lastname` (own private copy, line 642) — `r"[A-Za-z]\.([A-Za-z]+)/"` — slash **only**.

So `A.Virtanen-0179-x` yields `'Virtanen'` in the merge-gate hook and `None` in the format hook, exactly as #1208 states.

On whether #1207 **worsens** this: it does not. I traced both the pre-PR and post-PR code paths for `resolve_review_verdicts`. Before #1207, the merge-gate's dispatch to `check_comment_reviews` for a persona-shaped ref (dash or slash) already went through this same `extract_branch_author_lastname` (this hook's own, `[-/]`-accepting copy) — that regex is untouched by this diff. The only refactor is that the dispatch condition (`if branch_author_lastname: ... elif deployments/**: ... else: <no scan>`) became total (`comment_scan_scope`), but for a dash-form persona ref the computed `branch_author_lastname` / `branch_author_initial` fed into `check_comment_reviews` are byte-identical before and after. The format hook's regex is in a different file, untouched by this PR. The "write-time misses it, merge-time silently discounts it" dynamic #1208 describes on a dash branch therefore predates #1207 and applies identically on both sides of this diff — #1207 widens *which head-ref shapes get scanned at all* (dependabot/wave-merge/hand-made branches, none of which either hook's regex recognizes as a persona), it does not touch the persona-branch case where the two hooks disagree. I don't think this PR needs to wait on #1208; the charter_trailer-consolidation fix Aino proposed there is the right place to close it, independent of this PR's merge order.

### 2. Blast radius — contained

Searched the whole tree (`rg -n --no-ignore` including `.claude/` — note ripgrep skips dotdirs by default, had to target it explicitly) for `resolve_review_verdicts`, `check_comment_reviews(`, `ReviewVerdicts(`:

- Production call sites: exactly two — `.claude/lib/pr_review_state.py:182` (`gate.resolve_review_verdicts`) and `.claude/hooks/validate_pr_review.py:1742` (inside `check()`). Both are updated in this diff and both are keyword-call sites, so the new `comment_scan` field (added last, with a default) can't silently misalign a positional construction.
- Test-side `ReviewVerdicts(...)` constructions (5 call sites across both test files) are all keyword-based too; the ones that omit `comment_scan` are deliberately exercising the NOT_RUN default (`CommentScanNotMeasuredTests.test_review_verdicts_defaults_to_not_measured`), not accidental omissions.
- Checked all 7 cloned child repos under the parent (`noorinalabs-deploy`, `-isnad-graph`, `-user-service`, `-data-acquisition`, `-design-system`, `-isnad-ingest-platform`, `-landing-page`): none has its own `validate_pr_review.py`/`pr_review_state.py`/copy of these symbols — child PRs are gated by the parent's hooks via `--repo`, per `_child_roster_dir`'s docstring. No duplicate implementation exists anywhere to drift from this one.
- The only other repo-wide references to `pr_review_state` (`kickoff_sweep.py`, `check_agent_liveness.py`, `validate_pr_ci_status.py`) are docstring/comment mentions, not call sites.

### 3. Report-contract coverage — both surfaces covered

Checked both rendering paths, not just the human-readable one:
- `pr_review_state._render_text` — covered (`comment verdict scan:` line + the `stale verdicts: NOT MEASURED` variant).
- `pr_review_state._render_json` — uses `dataclasses.asdict(state)`, so `comment_scan` rides along automatically; pinned by `test_json_output_carries_the_scan_mode`.
- The **block message** in `validate_pr_review.check()` (not just the driver's report) — this is the surface I was specifically asked to check and it's covered: `scan_diagnostic` is correctly nested inside the `elif total_distinct < 2:` branch (`validate_pr_review.py:1950`) so it only fires when the merge is actually blocked, and `test_block_message_distinguishes_not_measured_from_measured_empty` asserts both directions (NOT-RUN wording present/absent) on identical 0/2 counts. I didn't find a rendering path left on the old "silent zero" wording.

### 4. Wave-merge (`deployments/phase-N/wave-M`) path — behaviourally equivalent

Traced the removed `elif head_ref.startswith("deployments/") and "/wave-" in head_ref: check_comment_reviews(number, "", repo=repo, content_ts=content_ts)` against the new unified call. For a `deployments/**` ref, `extract_branch_author_lastname` still returns `None` (regex doesn't match), so `branch_author_lastname or ""` → `""` (same sentinel as before), and `branch_author_first_initial(head_ref)` also returns `""` (its regex doesn't match either) — matching the old call's implicit `branch_author_initial=""` default. So the actual `check_comment_reviews` invocation is identical in its effective arguments pre/post-PR for this ref shape; only the reported `comment_scan` metadata changes (now `COMMENT_SCAN_NO_BRANCH_AUTHOR` instead of being unlabeled). `test_wave_merge_branch_behaviour_is_unchanged` pins this directly. This should hold at `/wave-wrapup` merge time without surprises.

### #1210 (disclosed residual exposure)

Not the primary verdict on this trade (the other reviewer has it), but for the record: I read #1210 and agree it's correctly scoped — the exposure requires a deliberate self-approval (itself a charter violation), is bounded to non-persona branches only, was already present in kind on `deployments/**` since main#294, and is now the one case that's actually *visible* (both the report and block message say "no self-review exclusion was applied" instead of hiding it). I don't think it should block this PR.

### What I did not independently verify
- Did not re-run pytest locally; relied on the CI run against this SHA (14/14 green, confirmed via `statusCheckRollup`) plus my own read of the added test bodies, which are unusually thorough (totality table, positive controls, regression pins on #1172/#950/#940 alongside the new behavior).
- Did not audit `check_comment_reviews`'s `content_ts`-required-kwarg mechanics (#1050) — pre-existing, untouched by this diff.

Pyright's import-resolution / missing-`content_ts`-argument noise on these files is expected per the brief and not reported here.

---
Requestor: Weronika Zielinska
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Merge-gate review — Nadia Khoury (Program Director).** This PR modifies the 2-reviewer gate itself, so a regression here fails open across every other PR in the wave. Reviewed accordingly: code read at the PR head via `gh api contents?ref=a44ea958`, never a local checkout; every mutation claim re-run independently in a throwaway clone; the disclosed residual judged on its own merits rather than ratified.

## What I verified independently

**Head and CI.** Head SHA `a44ea958a108b9d1709ec272a0549abacdfa4d41`. All 14 checks queried at that SHA (`commits/a44ea958/check-runs`, not the PR-level rollup) — 14/14 `success`, `total_count: 14`. `MERGEABLE` / `CLEAN`, merge-base equals `main` (`behind_by: 0`), single commit authored `Aino Virtanen <parametrization+Aino.Virtanen@gmail.com>`.

**Totality is real, not asserted.** `comment_scan_scope` is a single ternary over `extract_branch_author_lastname(head_ref)`. There is no branch that can return `NOT_RUN` — the two return values are the only reachable ones, and the `""` head ref falls to `NO_BRANCH_AUTHOR` rather than raising. The load-bearing claim in the resolver comment — that `branch_author_first_initial` returns `""` on exactly the refs `extract_branch_author_lastname` returns `None` for — checks out: `[A-Za-z]\.([A-Za-z]+)[-/]` versus `([A-Za-z])\.[A-Za-z]+[-/]`, identical match domain. The two halves of the branch author's identity cannot disagree.

**The defence-in-depth actually fires.** I reintroduced a skip path by making `comment_scan_scope` non-total (persona → excluded, `deployments/**` → no-author, everything else → `NOT_RUN`). The resolver raised, with the ref named:

```
CommentScanUndeterminedError: the PR-comment verdict scan was not dispatched for head ref
'dependabot/docker/integration-tests/fake_oauth/python-d3400aa'. An unscanned comment thread
is not evidence of zero approvals ...
```

and `check()` catches that at line 1881 and returns `decision: block`. So the guard fails **closed** through to the hook surface, not merely inside the resolver. That matters more than the raise itself.

## Mutation results — re-run, not accepted

Throwaway `--depth 1` clone of the PR branch; the PR branch was never touched. Baseline 309 passed / 21 subtests.

| Mutation | Result |
|---|---|
| Pre-#1207 dispatch restored verbatim | **11 test methods red** across both suites |
| `comment_scan_scope` made non-total (skip path reintroduced) | **22 red** — totality subtests *and* fail-closed integration |
| Report's not-measured branch collapsed to `else` | **1 red** — `test_stale_verdicts_none_is_only_claimed_when_the_scan_ran` |
| `comment_scan=verdicts.comment_scan` dropped from `compute_review_state` | **1 red** — `test_compute_review_state_propagates_the_gate_scan_mode` |
| Author exclusion removed entirely (`branch_author_lastname or ""` → `""`) | **3 red** — the relaxation guards hold |
| Block message never says the scan did not run | **1 red** |
| Block message *always* says it did not run | **1 red** — both directions genuinely asserted |
| Dishonest default on `ReviewVerdicts.comment_scan` | **1 red** |
| Dishonest default on `ReviewState.comment_scan` | **1 red** |

Two things worth stating explicitly. The dispatch mutation left the regression guards (`test_persona_branch_still_excludes_the_pr_author`, `..._admits_the_same_surname_colleague`, `..._wave_merge_branch_behaviour_is_unchanged`) green, as claimed — correct, since those pin behaviour the mutation restores rather than breaks. And the #1203 failure mode is genuinely absent here: I confirmed `DID NOT RUN` and `The scan DID run` are not present in the ~4 KB of unconditional block-message boilerplate, so those assertions are testing the branch and not the template. The paired positive controls (`assertIn` on the measured render alongside `assertNotIn` on the unmeasured one) are the right shape.

One unit correction, not a finding: the body says restoring the dispatch turns "21 assertions" red. I measured **11 test methods**; pytest stops at the first failing assertion per method, so the two counts are not comparable and I could not reproduce the number as stated. The direction and magnitude are verified.

**The rename.** `test_content_ts_forwarded_on_wave_merge_call_site` → `..._on_the_no_branch_author_path` loses nothing: both original assertions (`lastname == ""`, `content_ts == _NOW`) are kept verbatim and `initial == ""` is added. The stub widening from `fake_check(number, lastname, repo=None, content_ts=None)` to a version accepting `branch_author_initial=""` is required by the unified call site and is paid for by the new assertion on that argument. Net stronger.

## My position on #1210 — it does not block, and I reached that by measuring rather than by weighing rhetoric

The brief I was given asked me to press on whether "pre-existing in kind" is doing too much work once the affected population widens from one branch family to every non-conforming ref. That is the right question, and it is answerable empirically rather than by argument, so I answered it. I sampled the last 100 PRs (all states) in each of the seven org repos — 655 PRs — and classified every head ref against `[A-Za-z]\.([A-Za-z]+)[-/]`:

| Head-ref shape | Count | Exposure |
|---|---|---|
| Persona-prefixed | 568 (86.7%) | Unchanged — exclusion active, pinned by test |
| `deployments/**` | 79 (12.1%) | **Pre-existing since main#294** — not new |
| `dependabot/**` | 4 (0.6%) | Bot author, cannot post charter comments — neutral |
| Other non-conforming | **4 (0.6%)** | **The entire newly-exposed population** |

The four are `retro/phase-10-wave-28-20260727014228`, `fix/ci-pin-ruff-0.15.11`, `claude/noorinalabs-code-audit-1cdsf8`, `fix/refresh-cookie-contract-drift`. So the widening is roughly one twentieth of the exposure that already existed through the identical mechanism. "Pre-existing in kind" is not doing too much work here — the pre-existing population is an order of magnitude larger than the increment.

The reachability concern is nonetheless real and I confirmed both halves of it. There is **no branch-naming enforcement hook**: I enumerated all 46 files in `.claude/hooks/` at the head SHA and read the `hooks` block of `.claude/settings.json`; `validate_branch_freshness.py` is freshness, not naming, and the convention appears only in `_framework_config.py` as a format string and in `post_wave_kickoff_comment.py` as prose. And the sibling layer does not backstop it — `validate_review_comment_format` *does* hard-block a self-approval, but only after `extract_branch_author_lastname(branch_name)` succeeds; on a non-persona ref it returns `decision: allow` with a warning. Both layers lose the same discriminator on the same refs, and since every comment shares the single `parametrization` principal, no GitHub-level identity check can substitute. Aino disclosed all of this accurately in #1210; I confirmed it rather than taking it.

What decides it for me is the counterfactual, which is not "safe versus unsafe" but "which defect ships". Holding this PR does not remove the exposure — it stays on the 12.1% `deployments/**` slice regardless — while keeping a live fail-**closed** defect on the dependabot family and those four refs, plus a report that states `0/2 required — (none)` and `stale verdicts: none` about a thread nobody read. A silent zero is epistemically worse than a disclosed, bounded fail-open, because it is the failure mode that hides the next one; that is precisely why this defect survived from main#294 to now. And #1210's proposed fix (derive the author from commit identity, which is already fetched by `get_latest_content_commit` and discarded) closes the pre-existing `deployments/**` hole as well as the increment — so sequencing #1210 first buys nothing that #1210 alone does not buy, and costs the fix that is ready.

**Ship #1207 now, #1210 next, and #1210 should be treated as wave-29 work rather than drifting.** I have labelled it `wave-29` and added it to the board; it was carrying `phase-10` only.

## Finding — filed as #1211

The disclosure that justifies shipping is missing from the one path where it matters. `check()` composes the `no-branch-author` note into the **block** message, and `pr_review_state` renders it in a report someone has to run deliberately. On the path where `check()` **allows** the merge, the tail of the function reads `tech_debt_issue_numbers` and `tech_debt_unparseable` and falls through to `return None` — it never touches `verdicts.comment_scan`.

That is inverted relative to the risk. The #1210 residual only changes an outcome when the gate *passes*: a self-approval plus one genuine reviewer reaching 2/2 on a non-persona ref is a **silent allow**. The operator is told that self-review exclusion was unavailable in every case except the one where it mattered.

Filed as **#1211** with a note on why it is not a one-liner — `check()` already returns a single advisory for the main#1055 unparseable case and returns early, so the two have to compose rather than race. Non-blocking on this PR: the pre-#1207 behaviour on those same refs was a silent zero, which disclosed strictly less. It is a real gap in the mitigation, not in the fix.

## Nits, no action required

`pr_review_state._render_text` still prints `branch-author lastname: (none — wave-merge / unmatched)`, wording that predates this change and now also covers bot and non-charter refs. The new scan line immediately below states it accurately, so this is cosmetic. Separately, #1208 (filed from this PR) substantially duplicates the open #1175 — same regex, same fix; worth consolidating at wave wrap-up rather than carrying two.

The Pyright import-resolution and missing-`content_ts` noise on these files is pre-existing and expected (the `sys.path.insert` pattern, and tests that deliberately omit the argument to assert `TypeError`). Mypy is the gate and passes. Not reported as findings.

## Verdict

Approved. The fix is correct, the totality is structural rather than asserted, the guard fails closed through to the hook surface, and the test suite catches every mutation I could construct — including three the author did not claim. The design departure from the issue's proposal is the better call: a longer prefix list fails identically on the next unrecognised shape, and totality removes the defect class rather than the instance. The residual is disclosed, measured at 0.6% of PRs against a 12.1% pre-existing slice of the same mechanism, and tracked.

Do not merge — owner-gated.

---

Requestor: Nadia Khoury
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1210, #1211

Issue #1211, opened 2026-08-01
bug(hook4): the merge-ALLOW path never states that self-review exclusion was unavailable — the #1210 exposure is disclosed only when the gate blocks


parametrization:
## The gap

#1207 makes a skipped comment scan loud — but only on the paths that *stop* a merge.
`validate_pr_review.check()` emits the `COMMENT_SCAN_NO_BRANCH_AUTHOR` note **inside the
block message**, and `pr_review_state._describe_comment_scan` renders it in a report an
operator has to run deliberately. On the path where `check()` **allows** the merge, the
hook returns `None` (or, for the main#1055 case only, an unrelated advisory) and says
nothing at all about the scan mode.

That is exactly inverted relative to the risk. The #1210 residual only *matters* when the
gate passes: a persona self-approval plus one genuine reviewer reaching 2/2 on a
non-persona ref produces a **silent allow**. The operator is told "no self-review exclusion
was applied" in every case except the one where it changed the outcome.

## Evidence

`.claude/hooks/validate_pr_review.py` at `a44ea95`:

- block path — `scan_diagnostic` is composed into `result["reason"]` for both
  `not comment_scan_ran` and `COMMENT_SCAN_NO_BRANCH_AUTHOR`.
- allow path — the tail of `check()` (`# All checks passed …` onward) references
  `verdicts.tech_debt_issue_numbers` and `verdicts.tech_debt_unparseable`, never
  `verdicts.comment_scan`, and falls through to `return None`.

## Proposed fix

On the allow path, when `verdicts.comment_scan == COMMENT_SCAN_NO_BRANCH_AUTHOR`, return
`{"decision": "allow", "systemMessage": …}` naming the head ref and stating that
self-review exclusion was not applied to the counted verdicts.

Design note (why this is not a one-liner): `check()` already returns a single
`systemMessage` for the main#1055 unparseable-TechDebt advisory and returns early, so the
two advisories have to compose rather than race. A small accumulator that joins the
pending advisories before the final `return` is the shape that keeps both visible.

Wording should stay proportionate — this is an advisory, not a block. The gate itself is
unchanged; the point is that the operator reading the merge output learns which
discriminator was unavailable.

## Relationship to #1210

#1210 removes the *cause* (derive the branch author from commit identity, restoring
exclusion on non-persona refs). This issue closes the *disclosure* gap that exists until
then, and keeps residual value afterwards: `no-branch-author` remains reachable for genuine
[...]

Issue #1209, opened 2026-08-01
fix(ontology): conventions.md describes validate_pr_review.py as the comment-format hook — it is the 2-reviewer merge gate


parametrization:
## The error

`ontology/conventions.md` describes `validate_pr_review.py` with the sibling hook's job:

```
:240 | `validate_pr_review.py`             | PreToolUse (Bash) | Enforce charter review comment format |
:246 | `validate_review_comment_format.py` | PreToolUse (Bash) | Enforce review comment charter format |
```

Line 240 is wrong. `validate_pr_review.py` is the **2-reviewer merge gate** — its own
module docstring reads "PreToolUse hook: Require TWO PR review comments before merge",
and it is the canonical authority on the 2-distinct-roster-Approveds threshold, the
`T_content` staleness binding (#950), roster filtering (#498), the wave-bootstrap
single-reviewer exception (#228) and the batch-loop merge guard (#567/#894/#897). Nothing
in it enforces comment *format*; that is line 246's hook, whose description line 240
duplicates almost verbatim.

Suggested replacement for line 240:

```
| `validate_pr_review.py` | PreToolUse (Bash) | 2-reviewer merge gate — block `gh pr merge` without two distinct roster-valid current Approved verdicts |
```

## Why it matters

The overlay is the hand-curated layer agents read to orient before touching a file, and
the two hooks in question are already a known recurring-divergence pair (#932, #1175,
#1208). An entry that describes the merge gate as the format hook actively points a
reader at the wrong one of the two — the exact confusion those issues are about.

## Why it will not self-heal

`ontology/checksums.json` is currently **0 dirty / 284 tracked**, so the change-tracker
has nothing to flag and `/ontology-rebuild` will not surface it. The description was
wrong when written, not stale-since-written, so no staleness mechanism selects it.

Found while implementing #1206 (PR #1207); `ontology/**` was out of scope for that PR.

Issue #1177, opened 2026-07-30
tech-debt(hooks): Hook 22 blocks --squash only into wave branches — squash into main is unenforced, and cost 5 PRs their persona authorship


parametrization:
## Symptom

`gh pr merge <N> --squash` into **`main`** succeeds silently. Owner directive 2026-07-30 — *"no more squash merges"* — prohibits `--squash` on **every** base, but Hook 22 (`.claude/hooks/block_squash_wave_merge.py`) only hard-blocks when the PR's base resolves to `deployments/phase-*/wave-*`.

On `main` the rule is therefore convention-enforced only, and the convention is invisible at the moment of merge: the operator sees no gate, so the gate's silence reads as permission.

## Observed instance (the reason this is filed)

On 2026-07-30, five PRs were squash-merged to `main` under the *then-current* charter wording, which called squash "the standard path" for `main`:

| PR | Issue | Merge SHA | Persona that authored the work | Author recorded on `main` |
|---|---|---|---|---|
| #1173 | #1172 | `056a7afa` | Aino Virtanen | `parametrization` |
| #1153 | #1139 | `b708e1d1` | Nurul Hakim | `parametrization` |
| #1154 | #1134 | `fbb528e8` | Nadia Khoury | `parametrization` |
| #1155 | #1152 | `120740ff` | Weronika Zielinska | `parametrization` |
| #1156 | #1151 | `999dfae3` | Lucas Ferreira | `parametrization` |

All five persona attributions were demoted to `Co-authored-by` trailers. Verify:

```
git log --format='%an <%ae>  %s' 056a7afa~1..999dfae3 --no-merges
```

**Why no gate caught it:** wave-29 uses the `direct-to-main` merge model, so no wave→main integration PR is ever opened, so the `Verify commit authors are roster members` gate (`.claude/lib/verify_commit_identity.py`, main#627) — the check that would have gone red — never runs. Hook 22 was the only pre-merge defence and it does not apply to this base.

## Suspected root cause

`block_squash_wave_merge.py` gates on the **base ref pattern** rather than on the merge **method**. The wave-branch restriction was correct for its originating rationale (P7W19 #898/#222: protect the wave→main author gate), but that rationale was always a *subset* of the real one — squash re-authors persona commits to the bare principal on any base, because every persona email is a Gmail `+alias` of the single `parametrization@gmail.com` account.

## Reproduction

1. Open any PR against `main` in `noorinalabs-main` with a persona-authored commit.
[...]

PR #1215, opened 2026-08-02
fix(hook4): derive the branch author from commit identity when the head ref names none

### LINKED ISSUE #1210: bug(hook4): self-review exclusion is unavailable on any head ref without a persona prefix — derive the branch author from commit identity
## Residual exposure, disclosed by the author of #1207

#1206 / PR #1207 makes the comment verdict scan run for **every** head ref, using the `""`
"no identifiable branch author" sentinel wherever the ref carries no
`{Initial}.{Lastname}` prefix. That is a strict improvement on the silent zero it replaces,
but it is **not strictly non-relaxing**, and this issue records the one case where it is not.

## The case

| | pre-#1207 | post-#1207 | correct |
|---|---|---|---|
| Human persona, **non-charter** branch, posts their own `Approved` comment + one genuine reviewer approves | 0/2 — blocked (wrongly: the genuine reviewer was uncounted too) | **2/2 — passes** (wrongly: the self-review is counted) | 1/2 — blocked |

On a ref with no persona prefix the gate has no way to know who the branch author is, so
`is_branch_author` returns `False` for everyone and the author's own verdict enters the
reviewer set.

Not mitigated elsewhere: there is **no branch-naming hook** in `.claude/hooks/` (the
convention is documented in CLAUDE.md and `_framework_config.py` but not enforced), and
`validate_review_comment_format` only *warns* — "Could not extract author from branch
name", line 928 — rather than blocking, on exactly these refs.

## Why #1207 shipped anyway

- Pre-existing in kind: the `deployments/**` wave-merge path has carried this identical
  exposure since main#294. #1207 widens the population of refs, not the principle.
- It requires a deliberate self-approval, which is a charter violation on its own.
- It is now **visible**: the report and the block message both state "no self-review
  exclusion was applied" in precisely these cases. Before, both the exposure *and* the
  legitimate reviewers were invisible.
- Bots — the standing case, and the one #1206 was filed for — are unaffected: a bot is
  never a roster persona, so it could never be in the reviewer set.

## Proposed fix — derive the branch author from commit identity

Every persona commits with `-c user.name="{First} {Last}"` (CLAUDE.md § Commit Identity),
so the PR's commits carry the roster display name directly. **The data is already
fetched**: `get_latest_content_commit` pulls
`repos/{owner}/{repo}/pulls/{n}/commits?per_page=100` and reads only `sha` and
`commit.committer.date`, discarding `commit.author.name`.

So the exclusion could key on the commit author's roster name whenever the head ref names
no persona — restoring self-review exclusion on non-charter branches *and* on
`deployments/**`, closing the pre-existing #294 hole too.

Design decisions this needs (which is why it is not a drive-by on #1207):

- `get_latest_content_commit`'s return type changes, or a sibling accessor is added; it is
  heavily tested and shared with the staleness path.
- Multiple distinct commit authors on one branch — exclude all of them, or only the latest
  content commit's?
- A commit author with no roster match — fail closed (block) or ignore? Over-exclusion is
  the safe direction here, matching the `branch_author_initial` reasoning in
  `check_comment_reviews`'s docstring.
- Bot commits (`app/dependabot`) must resolve to "no persona" and change nothing.

Blocks nothing; #1207 is a net improvement without it. This closes the remainder.

parametrization:
Closes #1210.

## The hole

Self-review exclusion only ever had a **subject** on a `{Initial}.{Lastname}` head ref. On every other ref `is_branch_author` returned `False` for everyone, and the branch author's own `Approved` comment walked into the reviewer set:

| | pre-#1207 | post-#1207 | correct |
|---|---|---|---|
| Human persona, non-charter branch, self-approves + one genuine reviewer | 0/2 blocked (wrongly — the genuine reviewer was uncounted too) | **2/2 PASSES** | 1/2 blocked |

`deployments/**` has carried the identical hole since main#294. Measured exposure (655 PRs, 7 repos): 568 persona-prefixed (86.7%, unaffected), 79 `deployments/**` (pre-existing), 4 dependabot, 4 newly exposed.

## How the author is derived

Every persona commits with `-c user.name="{First} {Last}"`, so the commits carry what the ref does not.

- **`fetch_pr_commits`** does the ONE round-trip. `latest_content_commit` (T_content, #950) and `commit_author_identities` (#1210) are now **pure analyses** over its result. This replaces `get_latest_content_commit`, which already fetched the full commit objects and discarded `commit.author` — no new API call.
- Per non-merge commit, two sources are **unioned**: `commit.author.name`, and the `+alias` in `commit.author.email` via `persona_alias_from_email`. A union can only add exclusions, so it moves strictly toward blocking.
- **`is_self_review`** is the one predicate; both author sources go through `charter_trailer.is_branch_author` (#1172). No new surname comparison exists in the diff.
- **Scope: only where the ref names nobody.** On a persona ref the prefix is the author the human declared and behaviour is byte-for-byte unchanged — otherwise anyone who pushed a fixup to someone else's branch would newly lose their review, on the 86.7% path, to fix nothing.

## Degradation — every case toward blocking

| case | behaviour |
|---|---|
| commit list unfetchable | `CommitFetchError` → **hard block** (unchanged #950). No commit data ⇒ no author ⇒ stop. |
| bare `parametrization@gmail.com` | contributes **nothing**. A squash re-authors to the bare principal (#1177), so that address is evidence identity was *destroyed*. `roster.json` maps it to `Steven French`; this deliberately does **not** follow that map — #1181 reached the same verdict on the same address, and this reuses **#1181's own `_PERSONA_ALIAS_RE`** rather than adding a second regex. |
| bot | `dependabot[bot]` matches no roster persona ⇒ excludes nobody, both approvals still count (deploy#691 stays fixed). Its **address** yields no identity at all, so no human author is fabricated. Neutral by construction, not by a blocklist. |
| merge-only branch | no authored commits ⇒ no identity ⇒ pre-existing state, no regression. |
| wrong / ambiguous match | **over**-excludes and blocks, and the block message names every derived identity — visible, not silent. |

Merge-commit authors are skipped, matching `latest_content_commit`: running `gh pr merge` into a wave branch does not make the release coordinator its author, and excluding them would false-block the wave merges they must review.

## Strictly non-relaxing

The derivation feeds **only** the exclusion, and the exclusion can only *remove* reviewers. `total_distinct` can therefore only fall. `check()` allows on `>= 2`, or on `== 1` with the wave-bootstrap exception — from a blocking state (`0`, or `1` without the exception) a shrinking set reaches neither.

`StrictlyNonRelaxingTests` checks this as a property, not as an argument: over a 9-case matrix of (head ref x commit authors), the post-fix reviewer set must be a **subset** of the same resolve with the derivation switched off, plus a non-vacuity case requiring at least one matrix entry to strictly shrink.

## Mutation results

| mutation | tests that went RED |
|---|---|
| revert the derivation (`commit_authors = ()`) | 6 |
| hand-rolled surname comparison instead of `is_branch_author` | 2 |
| follow `roster.json`'s bare-principal map | 4 |
| include merge-commit authors | 2 |
| apply the derivation on persona refs too | 1 |
| `refine_comment_scan_scope` relabels unconditionally | 2 |

## Gates

`pre-commit run --all-files` clean; push stage (mypy 183 files, both pytest suites, skills, budgets, drift) clean; `pre_commit_ci_sync.py .` — `OK: pre-commit config mirrors all CI-enforced checks`. **3311 passed, 639 subtests.**

## Residuals, deliberate

1. **Persona refs are not cross-checked against commit identity.** A `A.Virtanen/…` branch whose commits are authored by someone else still excludes only Aino. Fixing it changes the majority path and trades a real false-block risk for a hole that requires both a mislabelled branch *and* a self-approval. Out of #1210's scope; flag if wanted as a follow-on.
2. **`ontology/conventions.md:240`** still mislabels this hook — that is **#1209**, not touched here.
3. **`ontology/checksums.json`** not marked dirty (the tracker resolves `$CLAUDE_PROJECT_DIR` to the main checkout, not this worktree). #1207 — same four files — did the same; `/ontology-rebuild` reconciles out of band.
4. **Shape change for #1211:** bot/wave refs whose commits carry any author string now report `commit-author-excluded` rather than `no-branch-author`, so #1211's allow-path advisory will fire on a narrower set. Its residual value is unchanged but its trigger population is smaller.

---
Requestor: Aino Virtanen
Requestee: Nino Kavtaradze, Nadia Khoury
RequestOrReplied: Request
TechDebt: none

### DIFF
--- a/.claude/hooks/tests/test_validate_pr_review.py
+++ b/.claude/hooks/tests/test_validate_pr_review.py
@@ -4744,12 +4744,61 @@ class StrictlyNonRelaxingTests(_ResolveOverFakeCommentsHarness):
         "Santiago Ferreira",
     )
 
-    def _sets(self, head_ref: str, authors: tuple[str, ...]):
+    # `CASES` above all run against the 4-name `THREAD`, so `before` lands on 3
+    # or 4 distinct reviewers in EVERY one of them — comfortably over the
+    # 2-reviewer bar. That makes them useless to
+    # `test_a_blocked_pr_never_becomes_an_allowed_one`, whose whole subject is
+    # what happens to a PR the gate was BLOCKING: with no blocking start state
+    # in the matrix, its guard body never ran and the test passed under a no-op
+    # (`feedback_fixture_makes_guard_assertion_inert` / #1203, caught in review
+    # of this very PR). These cases start from a thread SHORT of the bar, so
+    # `before` blocks and the guard actually executes.
+    #
+    # Counts below are MEASURED, not reasoned — `before -> after` total_distinct.
+    #
+    # (head_ref, commit authors, verdict thread)
+    BLOCKING_CASES = (
+        # Commit-derived exclusion fires from a blocked state: 1 -> 0.
+        ("feature/hand-made", ("Nino Kavtaradze",), ("Nino Kavtaradze",)),
+        # Nadia's suggested case — the ref names nobody, the author reviews herself. 1 -> 0.
+        ("feature/hand-made", ("Aino Virtanen",), ("Aino Virtanen",)),
+        # A persona ref, blocked by the REF-prefix exclusion alone (Lucas dropped),
+        # with commit authors present in the fixture. 1 -> 1, and that is the point:
+        # `resolve_review_verdicts` passes `()` for commit identities whenever the
+        # ref already names a persona, so the commit arm is silent here BY
+        # CONSTRUCTION. This case pins that the two sources cannot stack into a
+        # double exclusion on the same PR.
+        ("L.Ferreira/1210-x", ("Nino Kavtaradze",), ("Lucas Ferreira", "Nino Kavtaradze")),
+        # Blocked and stays blocked with NO exclusion firing (1 -> 1) — the
+        # other half of the guard's domain.
+        ("dependabot/npm_and_yarn/x-1.2.3", ("dependabot[bot]",), ("Lucas Ferreira",)),
+    )
+
+    def _all_cases(self):
+        """`CASES` (allowed-before) + `BLOCKING_CASES` (blocked-before)."""
+        return [(head_ref, authors, self.THREAD) for head_ref, authors in self.CASES] + list(
+            self.BLOCKING_CASES
+        )
+
+    @staticmethod
+    def _gate_allows(verdicts) -> bool:
+        """`check()`'s OWN allow-condition, transcribed — not a proxy for it.
+
+        Mirrors the `total_distinct == 1 and wave_bootstrap_exception` /
+        `total_distinct < 2` ladder in `check()`. Spelling it out here keeps
+        "allowed" meaning what the gate means by it, so the property below is
+        about the merge decision rather than about a count that resembles one.
+        """
+        return verdicts.total_distinct >= 2 or (
+            verdicts.total_distinct == 1 and verdicts.wave_bootstrap_exception
+        )
+
+    def _sets(self, head_ref: str, authors: tuple[str, ...], thread: tuple[str, ...] | None = None):
         commits = [
             _api_commit(f"c{n}", "2026-07-15T19:13:59Z", author_name=name)
             for n, name in enumerate(authors)
         ]
-        comments = [self._verdict(name) for name in self.THREAD]
+        comments = [self._verdict(name) for name in (self.THREAD if thread is None else thread)]
         after = self._resolve(head_ref, comments, commits=commits)
         # "Before" == the same pipeline with the commit-derived author source
         # switched off, which is exactly the pre-#1210 code path.
@@ -4758,9 +4807,9 @@ class StrictlyNonRelaxingTests(_ResolveOverFakeCommentsHarness):
         return before, after
 
     def test_reviewer_set_never_grows(self):
-        for head_ref, authors in self.CASES:
-            with self.subTest(head_ref=head_ref, authors=authors):
-                before, after = self._sets(head_ref, authors)
+        for head_ref, authors, thread in self._all_cases():
+            with self.subTest(head_ref=head_ref, authors=authors, thread=thread):
+                before, after = self._sets(head_ref, authors, thread=thread)
                 self.assertTrue(
                     after.distinct_reviewers <= before.distinct_reviewers,
                     f"{after.distinct_reviewers} is not a subset of {before.distinct_reviewers}",
@@ -4783,17 +4832,153 @@ class StrictlyNonRelaxingTests(_ResolveOverFakeCommentsHarness):
         self.assertTrue(shrank, "no case exercised the exclusion — the matrix proves nothing")
 
     def test_a_blocked_pr_never_becomes_an_allowed_one(self):
-        """The property expressed as the gate's own verdict, not as a count."""
-        for head_ref, authors in self.CASES:
-            with self.subTest(head_ref=head_ref, authors=authors):
-                before, after = self._sets(head_ref, authors)
-                before_allows = before.total_distinct >= 2
-                after_allows = after.total_distinct >= 2
-                if not before_allows:
-                    self.assertFalse(
-                        after_allows,
-                        "a PR the gate blocked before #1210 must not now pass",
-                    )
+        """The property expressed as the gate's own verdict, not as a count.
+
+        Scope, stated honestly so nobody reads more into a green run than is
+        there: on the CURRENT pipeline no input can actually flip blocked ->
+        allowed, because `after` is a subset of `before` and a blocked `before`
+        is already at most 1. That was measured, not assumed — the fail-open
+        mutant that unifies the two author sources (drop the `() if
+        branch_author_lastname` guard in `resolve_review_verdicts`, then let the
+        commit arm of `is_self_review` supersede the ref arm) is caught by
+        `test_reviewer_set_never_grows`, and this test correctly stays green
+        because the mutant only loses the exclusion, it does not push a blocked
+        PR over the bar. So the sibling subset test is the load-bearing one;
+        this is the TRIPWIRE for a future change that could ADD a reviewer to a
+        blocked PR, and its value depends entirely on the guard body actually
+        executing — which the anti-vacuity assertions at the bottom enforce.
+        """
+        reached = []
+        for head_ref, authors, thread in self._all_cases():
+            with self.subTest(head_ref=head_ref, authors=authors, thread=thread):
+                before, after = self._sets(head_ref, authors, thread=thread)
+                if self._gate_allows(before):
+                    continue
+                reached.append((head_ref, authors, thread))
+                self.assertFalse(
+                    self._gate_allows(after),
+                    "a PR the gate blocked before #1210 must not now pass",
+                )
+        # Anti-vacuity, same discipline as the sibling test above: the guard is
+        # inside an `if`, so a matrix in which nothing is blocked-before makes
+        # the whole test a no-op that still reports green.
+        #
+        # BOTH assertions are load-bearing and the order matters. `assertEqual`
+        # alone is NOT enough: emptying `BLOCKING_CASES` degrades it to
+        # `assertEqual([], [])`, which passes — i.e. deleting the blocking
+        # matrix silently restores the exact #1203 inertness this test was
+        # rewritten to remove. (Measured, not assumed: emptying the tuple with
+        # only the `assertEqual` present left this test green.) `assertTrue`
+        # pins the floor — the guard ran at all — and `assertEqual` then pins
+        # WHICH cases reached it, catching both a blocking case that quietly
+        # stops blocking and a `CASES` entry that starts.
+        self.assertTrue(
+            reached,
+            "no case in the matrix started from a BLOCKED state, so the guard body never "
+            "executed — this test would pass under a no-op implementation (#1203)",
+        )
+        self.assertEqual(
+            reached,
+            list(self.BLOCKING_CASES),
+            "the blocked-before guard did not execute over exactly the blocking matrix — "
+            "the property this test is named for was not actually checked",
+        )
+
+
+class SelfReviewExclusionKeepsTechDebtBlockTests(_ResolveOverFakeCommentsHarness):
+    """M13: the #1210 exclusion must not swallow the missing-TechDebt block.
+
+    In `check_comment_reviews` the self-review exclusion and the TechDebt
+    attestation check are two INDEPENDENT `if is_verdict_comment:` blocks, and
+    the non-relaxing argument for #1210 leans on that independence: the
+    exclusion may only ever remove a name from the reviewer SET, never relieve
+    a verdict of its attestation. Nothing pinned it. Moving the TechDebt block
+    inside the `if not is_self_review(...)` branch left the entire suite green
+    while a BLOCK turned into a PASS — the sole survivor of Nino Kavtaradze's
+    15-mutation run on this PR.
+
+    The transition is real, not theoretical: `check()` reaches the TechDebt
+    gate only AFTER clearing the 2-reviewer gate, so a PR with two clean
+    approvals plus the branch author's own attestation-less verdict is blocked
+    today and merges under the mutant.
+    """
+
+    HEAD_REF = "feature/hand-made"  # names no persona — commits are the ONLY author source
+    AUTHOR = "Aino Virtanen"
+
+    @staticmethod
+    def _verdict_without_tech_debt(requestor: str) -> dict:
+        """A well-formed Approved verdict carrying NO `TechDebt:` line."""
+        return {
+            "body": (
+                f"Requestor: {requestor}\nRequestee: Someone Else\nRequestOrReplied: Approved"
+            ),
+            "created_at": "2026-07-20T00:00:00Z",
+        }
+
+    def _comments(self) -> list[dict]:
+        # Two clean approvals clear the 2-reviewer gate, so the ONLY thing that
+        # can still block is the branch author's missing attestation.
+        return [
+            self._verdict("Lucas Ferreira"),
+            self._verdict("Nino Kavtaradze"),
+            self._verdict_without_tech_debt(self.AUTHOR),
+        ]
+
+    def _commits(self) -> list[dict]:
+        return [_api_commit("c0", "2026-07-15T19:13:59Z", author_name=self.AUTHOR)]
+
+    def test_branch_authors_own_verdict_is_excluded_but_still_owes_tech_debt(self):
+        """The two effects are independent: dropped from the set, still attesting."""
+        verdicts = self._resolve(self.HEAD_REF, self._comments(), commits=self._commits())
+        # Excluded from the reviewer set — the #1210 behaviour, unchanged.
+        self.assertEqual(
+            verdicts.distinct_reviewers,
+            {"lucas ferreira", "nino kavtaradze"},
+            "author not excluded",
+        )
+        self.assertEqual(verdicts.total_distinct, 2)
+        # ...and STILL on the hook for the attestation. This is the assertion
+        # the mutant kills.
+        self.assertEqual(verdicts.reviews_missing_tech_debt, [self.AUTHOR])
+
+    def test_check_still_blocks_the_merge_on_the_authors_missing_attestation(self):
+        """The same property as a merge decision — a BLOCK the mutant turns into a PASS."""
+        input_data = {
+            "tool_name": "Bash",
+            "tool_input": {"command": "gh pr merge 691 --repo noorinalabs/noorinalabs-main"},
+        }
+        comments, commits = self._comments(), self._commits()
+
+        def fake_run(args, capture_output, text, timeout):  # noqa: ARG001
+            result = mock.MagicMock()
+            result.returncode = 0
+            if args[0] == "gh" and args[1:3] == ["repo", "view"]:
+                result.stdout = json.dumps({"owner": {"login": "noorinalabs"}, "name": "r"})
+            else:
+                result.stdout = json.dumps(comments)
+            return result
+
+        with (
+            mock.patch.object(hook.subprocess, "run", side_effect=fake_run),
+            mock.patch.object(hook, "fetch_pr_commits", return_value=commits),
+            mock.patch.object(hook, "_load_roster_names", return_value=set(self.ROSTER)),
+            mock.patch.object(hook, "get_pr_data", return_value=self._pr_data(self.HEAD_REF)),
+            mock.patch.object(hook, "log_pretooluse_block"),
+        ):
+            result = hook.check(input_data)
+
+        assert result is not None, (
+            "check() returned None — the merge was ALLOWED despite the branch author's "
+            "verdict carrying no TechDebt line (the M13 mutant's signature)"
+        )
+        self.assertEqual(result["decision"], "block")
+        reason = str(result["reason"])
+        # Pin the REASON too: blocking for some unrelated cause (a reviewer
+        # shortfall, say) would satisfy a bare `decision == "block"` while the
+        # attestation check was in fact suppressed.
+        self.assertIn("TechDebt: attestation line", reason)
+        self.assertIn(self.AUTHOR, reason)
 
 
 class CommitAuthorBlockDiagnosticTests(_ResolveOverFakeCommentsHarness):
--- a/.claude/hooks/tests/test_validate_pr_review.py
+++ b/.claude/hooks/tests/test_validate_pr_review.py
@@ -123,6 +123,11 @@ class _CheckCommentReviewsHarness(unittest.TestCase):
                 # explicit "no content binding" value these ~30 pre-#950 callers
                 # exercise (every verdict counted unconditionally).
                 content_ts=None,
+                # commit_author_identities is REQUIRED for the same fail-open
+                # reason (#1210); `()` is the explicit "the ref already named the
+                # author, no commit-derived identity applies" value these
+                # pre-#1210 callers exercise.
+                commit_author_identities=(),
             )
 
     @staticmethod
@@ -131,11 +136,15 @@ class _CheckCommentReviewsHarness(unittest.TestCase):
         branch_author: str,
         repo: str | None = None,
         content_ts: "datetime | None" = None,
+        commit_author_identities: tuple = (),
     ):
         """As `_run_with_fake_api`, but binds verdicts to a T_content (#950).
 
         Kept as a separate entry point so the ~30 pre-#950 callers of
         `_run_with_fake_api` keep exercising the unbound path unchanged.
+
+        `commit_author_identities` (#1210) defaults to `()` — the pre-#1210
+        answer — so those callers stay unchanged; the #1210 tests pass it.
         """
         comments_stdout = json.dumps(comments_list)
 
@@ -154,6 +163,7 @@ class _CheckCommentReviewsHarness(unittest.TestCase):
                 branch_author,
                 repo=repo,
                 content_ts=content_ts,
+                commit_author_identities=commit_author_identities,
             )
 
 
@@ -581,7 +591,12 @@ class ContentTsRequiredTests(unittest.TestCase):
 
     def test_check_comment_reviews_without_content_ts_raises_type_error(self):
         with self.assertRaises(TypeError):
-            hook.check_comment_reviews(451, "pham", repo="noorinalabs/x")  # missing content_ts
+            hook.check_comment_reviews(
+                451,
+                "pham",
+                repo="noorinalabs/x",
+                commit_author_identities=(),
+            )  # missing content_ts
 
     def test_partition_formal_reviewers_without_content_ts_raises_type_error(self):
         with self.assertRaises(TypeError):
@@ -596,7 +611,13 @@ class ContentTsRequiredTests(unittest.TestCase):
                 returncode=0, stdout="[]"
             ),
         ):
-            result = hook.check_comment_reviews(451, "pham", repo="noorinalabs/x", content_ts=None)
+            result = hook.check_comment_reviews(
+                451,
+                "pham",
+                repo="noorinalabs/x",
+                content_ts=None,
+                commit_author_identities=(),
+            )
         self.assertEqual(result.undetermined, "")
 
     def test_partition_formal_reviewers_still_accepts_explicit_none(self):
@@ -613,7 +634,13 @@ class ContentTsRequiredTests(unittest.TestCase):
         positional third argument must now raise TypeError rather than
         silently binding to `repo`."""
         with self.assertRaises(TypeError):
-            hook.check_comment_reviews(451, "pham", "noorinalabs/x")  # repo positional
+            hook.check_comment_reviews(
+                451,
+                "pham",
+                "noorinalabs/x",  # repo positional
+                content_ts=None,
+                commit_author_identities=(),
+            )
 
 
 class ExtractBranchAuthorLastnameTests(unittest.TestCase):
@@ -868,12 +895,18 @@ class _NoContentBindingHarness(unittest.TestCase):
 
     Staleness itself is covered by `StaleVerdictBindingTests`,
     `BranchUpdateRegressionTests`, and `CommitFetchFailClosedTests`, which patch
-    `get_latest_content_commit` with real values instead.
+    `fetch_pr_commits` with real commit data instead.
+
+    Patching the FETCH rather than the analysis (#1210) means the real
+    `latest_content_commit` and `commit_author_identities` still run over the
+    stub: an empty commit list yields no content binding AND no commit-derived
+    branch author, which is exactly the pre-#950/pre-#1210 behaviour these
+    tests intend.
     """
 
     def setUp(self) -> None:
         super().setUp()
-        patcher = mock.patch.object(hook, "get_latest_content_commit", return_value=None)
+        patcher = mock.patch.object(hook, "fetch_pr_commits", return_value=[])
         patcher.start()
         self.addCleanup(patcher.stop)
 
@@ -1144,7 +1177,11 @@ class CommentPaginationTests(_CheckCommentReviewsHarness):
 
         with mock.patch.object(hook.subprocess, "run", side_effect=fake_run):
             hook.check_comment_reviews(
-                self.PR_NUMBER, self.BRANCH_AUTHOR, repo=self.REPO, content_ts=None
+                self.PR_NUMBER,
+                self.BRANCH_AUTHOR,
+                repo=self.REPO,
+                content_ts=None,
+                commit_author_identities=(),
             )
 
         gh_api_calls = [a for a in captured_args if a[0] == "gh" and a[1] == "api"]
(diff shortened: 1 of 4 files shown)


parametrization:
# Merge-gate review — PR #1215

Reviewed at head SHA `0186d4d07a0a74fabad0ca16c0fd7bc036ac6b29`, diffed against base
`17b8e41` via the contents API, not a local checkout. I filed #1210 and measured its
exposure, so I graded this against whether the shipped code is correct and non-relaxing,
not against the design I proposed.

## The decisive question: can this make the gate PASS what it BLOCKED?

**No.** Verified by execution, not by argument.

I enumerated `check()`'s allow paths myself: `tool_name != Bash`, `--admin`, non-merge
command, `total_distinct >= 2`, `total_distinct == 1 && wave_bootstrap_exception`, and the
main#1055 unparseable-TechDebt `decision: allow`. Only three quantities gate the last
three: `distinct_reviewers`, `wave_bootstrap_exception`, `reviews_missing_tech_debt`.

Then I built an independent battery — **2,268 cases** over (7 head refs x 9 commit-author
sets x 9 verdict threads x 2 label sets x TechDebt present/absent), driving the real
`check()` end-to-end over a faked `gh` boundary, with "before" = the same pipeline with
`commit_author_identities` patched to `()`:

```
cases exercised: 2268
BLOCK -> ALLOW regressions (fail-open): 0
(before_allowed, after_allowed): {(False,False): 1683, (True,True): 500, (True,False): 85}
wave-bootstrap-labelled cases that TIGHTENED allow->block: 35
```

Non-vacuous in both directions: 85 cases tighten, 1,683 exercise a pre-existing blocking
state. The battery covers the **wave-bootstrap arm, which your own matrix does not** — no
case in `StrictlyNonRelaxingTests.CASES` carries the label. Confirmed by reading rather
than assuming: `is_single_reviewer_exception` returns `False` on `len(reviewers) != 1`
(`validate_pr_review.py:1778`), so a 1→0 shrink cannot enable it.

Also verified, since the T_content refactor could have relaxed staleness independently:
`fetch_pr_commits` + `latest_content_commit` preserve `get_latest_content_commit`'s
semantics exactly — same non-dict filtering, same `>= 2` parent test, same committer-date
source, same `>` tie-break, same raise-on-unparseable-date. And no reference to the deleted
name survives anywhere in `*.py`/`*.md`/`*.yml` at head.

**Statement: the change is strictly non-relaxing.**

## Specific areas

**1. Bare `parametrization@gmail.com`** — refusal holds on every path. `persona_alias_from_email`
reuses `_PERSONA_ALIAS_RE` (`validate_pr_review.py:1480`); there is no second address regex
in the diff. `test_roster_json_maps_the_bare_principal_and_this_function_refuses_to` asserts
the map still exists *next to* the refusal, so the test fails loudly if the premise
disappears — that is the right shape.

**2. Scope (residual 1)** — correctly scoped. **Do not widen.** My own real-data sweep below
is the evidence: excluding everyone who touched a branch collapses the eligible reviewer
pool. Applying that to the 86.7% persona path would discard the verdict of every fixup
pusher to close a hole needing both a mislabelled branch and a self-approval. Your rationale
was right and now has a measurement behind it.

**3. Does it close the #294 `deployments/**` hole?** — Yes, and the test proves what its
name claims: `test_wave_merge_branch_now_excludes_its_author_too` goes RED under mutation
M1. But see Tech-debt #1216 — on a real wave branch it closes considerably *more* than the
hole.

**4. Bot neutrality** — verified in the hooks. `dependabot[bot]` is carried as an ordinary
derived identity and is inert because it matches no roster persona; the address yields
nothing via #1181's matcher, so no human is fabricated. No blocklist. Confirmed against a
real dependabot payload. One reporting defect falls out of it — #1220.

**5. #1172 non-regression** — both arms of `is_self_review` delegate to
`charter_trailer.is_branch_author`. There is no hand-rolled surname comparison in the diff;
the only `.lower()` on a lastname is the dedupe key, which is not a person comparison.
Mutation M2 confirms it.

**6. Degradation** — no commit data raises `CommitFetchError` and hard-blocks (M11/M12 both
caught). Unknown initial over-excludes, as claimed — but see the M10 survivor.

## Mutation results — my battery, independently derived

Baseline: 350 passed, 53 subtests. **14 mutations, 11 caught, 3 survived.**

| # | Mutation | Result | RED tests |
|---|---|---|---|
| M1 | revert the derivation (`commit_authors = ()`) | CAUGHT | 6 |
| M2 | hand-rolled surname compare in `is_self_review` | CAUGHT | 2 |
| M3 | follow `roster.json`'s bare-principal map | CAUGHT | 4 |
| M4 | include merge-commit authors | CAUGHT | 2 |
| M5 | apply the derivation on persona refs too | CAUGHT | 1 |
| M6 | `refine_comment_scan_scope` relabels unconditionally | CAUGHT | 10 |
| M7 | `refine` upgrades `NOT_RUN` too (launders the #1206 block) | CAUGHT | 1 |
| M8 | drop the email/alias author source | CAUGHT | 1 |
| M9 | drop the `commit.author.name` source | CAUGHT | 6 |
| **M10** | **refuse to over-exclude when `initial == ""`** | **SURVIVED** | — |
| M11 | `latest_content_commit` skips an unparseable date instead of raising | CAUGHT | 1 |
| M12 | `fetch_pr_commits` drops `--paginate` | CAUGHT | 1 |
| **M13** | **exclusion also suppresses the missing-TechDebt block** | **SURVIVED** | — |
| **M14** | **dedupe key loses the initial** | **SURVIVED** | — |

Your six all reproduce with the failure counts you reported. M7–M14 are mine.

## The anti-vacuity test genuinely bites — but its sibling does not

Instrumented `StrictlyNonRelaxingTests` case by case:

```
head_ref                         authors                          before after shrank before_allows
feature/hand-made                ('Nino Kavtaradze',)                4     3   True   True
feature/hand-made                ('Lucas Ferreira','Nino Kavtaradze')4     2   True   True
feature/hand-made                ('Weronika Zielinska',)             4     4   False  True
deployments/phase-10/wave-29     ('Nino Kavtaradze',)                4     3   True   True
dependabot/npm_and_yarn/x-1.2.3  ('dependabot[bot]',)                4     4   False  True
""                               ('Aino Virtanen',)                  4     3   True   True
nohashinthisref                  ('Santiago Ferreira',)              4     3   True   True
L.Ferreira/1210-x                ('Nino Kavtaradze',)                3     3   False  True
L.Ferreira/1210-x                ('Lucas Ferreira',)                 3     3   False  True

cases that strictly shrank: 5
cases where BEFORE was blocking: 0
```

`test_fixture_is_not_vacuous_at_least_one_case_actually_shrinks` is real — 5 of 9 shrink,
and it goes RED under M1 and M9. Good.

`test_a_blocked_pr_never_becomes_an_allowed_one` is **inert** — Must-fix 1.

## Must-fix

**1. `StrictlyNonRelaxingTests.test_a_blocked_pr_never_becomes_an_allowed_one` asserts
nothing.**

`THREAD` is four reviewers and `ROSTER` is those same four, so `before.total_distinct` is
3 or 4 in all nine cases. `before_allows` is therefore `True` 9/9, the `if not
before_allows:` body **never executes**, and the test passes under every implementation
including a no-op. The test named for the exact merge-gate property certifies nothing —
`feedback_fixture_makes_guard_assertion_inert` / #1203, introduced by this PR.

Fix: add matrix entries that start from a blocking state. A one-reviewer thread, or a
`("feature/hand-made", ("Aino Virtanen",))` case run against a 1-name thread, makes the
branch execute. Assert the branch was reached at least once, the way the sibling
anti-vacuity test does.

The subset property in `test_reviewer_set_never_grows` is the load-bearing one and it does
bite, so this is a coverage defect, not a correctness one.

**2. Nothing guards "the self-review exclusion must not suppress the missing-TechDebt
block" (M13).**

Your non-relaxing argument rests partly on `reviews_missing_tech_debt` being appended for
every verdict regardless of author. That is true of the shipped code
(`validate_pr_review.py:1437-1441` is a *separate* `if is_verdict_comment:` from the
exclusion) — and completely unguarded. Merging the two adjacent blocks, which this PR makes
more tempting by putting a five-line `is_self_review(...)` call directly above them, is a
real fail-open. Reproduced end-to-end:

```
non-charter ref, commits authored by Nino Kavtaradze
thread: Aino (Approved, TechDebt: none), Lucas (Approved, TechDebt: none),
        Nino (Approved, NO TechDebt line)

shipped code      -> BLOCK  "has review(s) missing the mandatory TechDebt: attestation line"
under M13         -> ALLOW
```

That is BLOCK → ALLOW relative to base as well as to head, and the whole suite stays green
at 350 passed.

Fix: one test — a branch author who posts a verdict without a `TechDebt:` line must still
appear in `reviews_missing_tech_debt` even though they were excluded from
`distinct_reviewers`.

Both must-fixes are test-only. **No production change is needed; the shipped logic is
correct on both points.**

## Residual 3 — `ontology/checksums.json`

Not hand-editing it was the right call, and matching #1207's precedent was right. Your
stated *reason* is wrong, though, and the correction matters: `ontology_tracker.py` does not
mis-resolve `$CLAUDE_PROJECT_DIR`. It **deliberately skips worktree paths** —
`_is_worktree_path` and `_is_linked_worktree`, rationale documented at
`ontology_tracker.py:24-49` — to stop dead worktree keys accumulating.

The real gap is that the documented catch-up ("updated whenever the file is next
Edit/Written directly on the main checkout") has not fired. On `main` today
`.claude/hooks/validate_pr_review.py` has `last_tracked == last_resolved` at 2026-07-19
while its on-disk hash has moved, so the overlay reports **clean** for a file that is
materially stale — worse than dirty, because `/ontology-rebuild` has nothing to reconcile.
`.claude/lib/pr_review_state.py` is not tracked at all. Yes, it warrants its own issue —
#1219.

## The finding that needs a decision — #1216

`commit_author_identities` returns *every* non-merge author on the branch. On a persona ref
that is one person. On `deployments/**` it is **the entire wave's implementer roster**,
because a wave branch accumulates every per-issue PR merged into it. Run against real API
payloads at head SHA:

| wave-merge PR | commits | derived "branch authors" |
|---|---|---|
| `main#1083` | 30 | 5 |
| `da#487` | 9 | 4 |
| `main#711` | 19 | 3 |

Full sweep, **172 `deployments/**` PRs across all 7 repos**, shipped derivation vs each PR's
real Approved values: **2 (1.2%) would lose an actual approving reviewer**, going 2/2 → 1/2:

- `main#711` (`deployments/phase-5/wave-5`) — `Wanjiku Mwangi` subtracted
- `main#530` (`deployments/phase-3/wave-11`) — `Aino Virtanen` subtracted

It fails **closed** and the block message names every derived identity, so this is not a
gate-safety problem and does not block this PR. But it silently answers a charter question
nobody has asked — *may an implementer review the wave merge containing their own work?* —
and on a 5–8-implementer wave the eligible reviewer pool can fall below two, making the wave
merge unapprovable. Not urgent for wave-29 (`direct-to-main`); live for every child repo
still on the wave-branch model. Decide it before the next wave branch.

## Charter compliance

- Commit identity: `Aino Virtanen <parametrization+Aino.Virtanen@gmail.com>` — correct.
- Branch `A.Virtanen/1210-hook4-branch-author-from-commit-identity`, base `main` — correct
  for wave-29 direct-to-main.
- CI: **14/14 SUCCESS** on `statusCheckRollup` at head SHA `0186d4d` (ruff lint + format,
  mypy, both pytest suites, pre-commit⇄CI sync-drift, roster union, commit-author audit,
  smoke, budgets). `mergeStateStatus: CLEAN`.
- Test suites reproduce locally at head: 350 passed, 53 subtests.

## What this PR gets right

The `fetch_pr_commits` / `latest_content_commit` / `commit_author_identities` split turns
two subprocess-mocked behaviours into pure functions over a commit list, and the harness
change from stubbing the *analysis* to stubbing the *fetch* means the existing #1206 tests
now run the real analysis code. That is a coverage increase disguised as a refactor. The
scope decision — consult commit identity only where the ref is silent — is the correct call
and the one my sweep would have forced anyway. Reusing #1181's matcher instead of a second
regex, and pinning the roster.json trap as an assertion rather than a comment, is exactly
right.

Resolve the two test gaps and I will approve on the next pass.

---
Requestor: Nadia Khoury
Requestee: Aino Virtanen
TechDebt: #1216, #1217, #1218, #1219, #1220

parametrization:
Requestor: Nino Kavtaradze
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1221, #1223

Security review. Reviewed against the artifact at `0186d4d`, in a detached worktree, running the code — not from the diff and not from the PR body's claims. I did not read the other reviewer's verdict before reaching mine.

## The question that decides this review

**Can this change make the gate PASS anything it previously BLOCKED? No.** Verified two independent ways.

**1 — Enumerated every allow path in `check()` myself.** The derivation has exactly two consumers, and I traced both to ground:

- `is_self_review` -> `latest_verdict` -> `result.reviewers` -> `comment_reviewers` -> `roster_comment_reviewers` -> `distinct_reviewers` -> `total_distinct`. Monotone decreasing.
- `refine_comment_scan_scope` -> `ReviewVerdicts.comment_scan`. I grepped every consumer: `check()` reads it only at lines 2385/2393/2408, all inside the `total_distinct < 2` **block** branch, all message text; `pr_review_state` reads it only in `_describe_comment_scan` and the report. It is never an input to a decision, and `ReviewState.passes()` does not reference it.

`check()`'s allow paths are: non-Bash, `--admin`, non-merge-command, `total_distinct >= 2`, `total_distinct == 1 and wave_bootstrap_exception`, and the two terminal returns after the TechDebt block. From a blocking state a shrinking set reaches none of them — `is_single_reviewer_exception` requires `len(reviewers) == 1`, so `0 -> 0` stays blocked and `1 -> 0` moves further from it, never onto it. I checked this rather than taking the PR body's word.

I also probed the one place your argument could have failed that the body does not mention: **the TechDebt attestation is the other block condition, and it is not keyed on the exclusion.** In `check_comment_reviews` the `reviews_missing_tech_debt` append sits under a bare `if is_verdict_comment:`, outside the `is_self_review` guard, so removing a reviewer cannot remove a block reason. That is correct in the shipped code and it is load-bearing — see finding 2.

**2 — Ran it against the whole historical corpus.** I replayed `compute_review_state` over **every merged/closed/open PR with a non-persona head ref across all 7 repos — 184 PRs** — twice, once with `commit_author_identities` live and once patched to `()` (which is exactly the pre-#1210 path), and diffed the reviewer sets:

```
TOTAL: unchanged=182  changed=2  errors=0
RELAXED (block -> pass): 0
```

Zero relaxations on real data. **The change is strictly non-relaxing.** Confirmed.

## Specific areas you asked about

**Bare `parametrization@gmail.com`.** The refusal holds in code, and the trap is still live. I round-tripped `persona_alias_from_email` over all **78** `roster.json` entries: it yields no identity for exactly two — `Steven French -> parametrization@gmail.com` and `Annunaki -> parametrization+Annunaki@gmail.com`. The `Steven French` mapping is still in `roster.json:2` as your test asserts. No other path re-introduces it: `_load_org_manifest_names` reads `roster.json` only to derive *names*, filtered through the same `_PERSONA_ALIAS_RE`, so the bare principal is dropped there too and never becomes an identity. `_PERSONA_ALIAS_RE` is defined **once** (line 1650) and used at 868 and 1705 — the shipped code reuses #1181's matcher, it does not duplicate it. Your first draft's shadowing did not ship.

**Bot neutrality — verified live, not from the hooks' text.** `noorinalabs-deploy#691` at this head: derived `['dependabot[bot]']`, reviewer set `['lucas ferreira', 'nino kavtaradze']`, **2/2 PASS**. Neutral by construction on both sources: the name `dependabot[bot]` is a single token whose "lastname" no roster persona can equal, and the address `NNNNN+dependabot[bot]@users.noreply.github.com` fails `_PERSONA_ALIAS_RE`'s required internal `.`, so it yields nothing at all. Not a blocklist. I confirmed independently that Hook 4 never extracts `Requestee` — the only `Requestee` regex in the repo is `validate_review_comment_format.py:810`, a different hook, untouched here — and that `validate_review_comment_format` allows-with-warning at line 927 when no author is extractable. **No human author can be fabricated for a bot PR.**

**No #1172 regression.** Both arms of `is_self_review` delegate to `charter_trailer.is_branch_author`; there is no surname comparison in the diff. Proven by mutation, not by reading: replacing the commit arm with `name_lastname(requestor).lower() == identity.lastname.lower()` turns 2 tests RED. This is not theoretical — on `main#1083` the derived set contains `Lucas Ferreira` and the actual reviewer is `Santiago Ferreira`, and **the initial discriminator is the only thing that keeps his verdict counted.**

**The scope decision — I judge it correct.** Widening to persona refs would newly discard the verdict of anyone who pushed a fixup to someone else's branch, on the 86.7% path, to close nothing. I confirmed the majority path is genuinely untouched: `commit_authors` is `()` there, `is_self_review` collapses to the original `is_branch_author` call, and `scan_scope` is not refined. Live: `#1207` and `#1213` both report `derived=[]`, `scan=author-excluded`. I also ran a differential on the `get_latest_content_commit` -> `fetch_pr_commits` + `latest_content_commit` split across 10 real PRs — **T_content byte-identical, 0 differences**, so the refactor under the majority path is behaviour-preserving.

**Union-of-all-authors and the merge-commit skip — probed both for holes.** The union is the right call and it earns its keep in a case you may not have known about: `roster.json` carries `Cedric Novak` *and* `Cédric Novák` as two entries on one address, and git history shows both spellings actually used as `commit.author.name`. When such a commit is authored under the accented spelling, only the name+alias union excludes both spellings. See finding 3 for the direction it does not save. The merge-commit skip is right and is pinned (dropping it turns 2 RED); its cost on `deployments/**` is real and is finding 1.

**Degradation.** Every failure path lands on block, confirmed by reading each: `CommitFetchError` -> hard block; unresolvable repo -> block; `pr_data is None` -> block; `CommentScanUndeterminedError` -> block; `RosterResolutionError` -> block. The `NOT_RUN` -> `CommentScanUndeterminedError` raise at line 2000 fires **before** `refine_comment_scan_scope` at line 2022, so a not-run scan cannot be laundered into a mode even if `refine` were later loosened — two independent guards, and mutating `refine` to relabel unconditionally turns 10 RED.

## Mutation results — my own battery, 15 mutations, 14 caught / 1 survived

Your six, re-run independently: revert derivation **11 RED** (you reported 6; the extras are `test_both_authors_of_a_handed_over_branch_are_excluded`, `test_self_approval_on_a_non_charter_ref_no_longer_reaches_two_of_two`, `test_wave_merge_branch_now_excludes_its_author_too`, the anti-vacuity case, and the driver propagation test); hand-rolled surname compare **2**; follow the bare-principal map **4**; include merge-commit authors **2**; derive on persona refs **1**; unconditional relabel **10**. All caught.

Nine of my own: read `commit.committer` instead of `commit.author` **11 RED**; drop the alias source **1**; drop the name source **6**; keep only the last commit's author **2**; drop the ref arm of `is_self_review` **8**; treat an empty author name as an identity **10**; `persona_alias_from_email` returns the whole local part **3**. All caught.

**`test_fixture_is_not_vacuous_at_least_one_case_actually_shrinks` genuinely bites.** I checked it specifically because that is the #1203 shape: it goes RED under the revert-derivation mutation *and* under drop-the-name-source. It is not an inert assertion.

**One survivor**, filed as #1221.

## Findings

**1 — `deployments/**` over-exclusion. Low. Already filed as #1216, so not re-filed.** My sweep independently reproduced it: `main#711` (loses Wanjiku Mwangi) and `main#530` (lose Aino Virtanen) go 2/2 -> 1/2, and `main#711`'s derived set is `{Aino, Santiago, Wanjiku}` — three of the four org-level personas. Two things keep this off my must-fix list. It is fail-**closed** and visible (the block message names every derived identity), and `charter/pull-requests/wave-merge.md` § point 5 routes wave->main integration PRs through `ADMIN_MERGE_EXCEPTION="wave-merge:..." gh pr merge <N> --merge --admin`, which short-circuits `check()` at line 2087 before any of this runs, and states the 0/2 BLOCK on those PRs is "expected and audited."

That same charter clause is worth one correction to your exposure arithmetic: the **79 `deployments/**` PRs (12.1%) counted as exposure closed were already outside the gate**, because their sanctioned merge path is `--admin`. The exposure this PR genuinely closes is the non-charter human refs and the newly-exposed slice, not the wave-merge population. That does not change the merit of the change; it changes the number in the body.

**2 — Nothing pins the TechDebt attestation's independence from the exclusion. Low, filed #1221.** This is the "does removing a reviewer flip a code branch that allows" probe. The shipped code is correct. But moving the TechDebt block inside the `is_self_review` guard — the natural-looking "why should an excluded self-review owe an attestation?" refactor — leaves the entire suite **green (0 RED)** while turning a BLOCK into a PASS, because `reviews_missing_tech_debt` is itself a block condition in both `check()` and `ReviewState.passes()`. The only survivor of 15 mutations, and it is a fail-open shape. A test with a positive control, per the issue.

**3 — Surname comparison does not Unicode-fold. Low, filed #1223.** Reproduced against this head:

```
commits authored as "Cedric Novak" <parametrization+Cedric.Novak@gmail.com>
derived: (CommitAuthorIdentity(lastname='Novak', initial='c', display='Cedric Novak'),)
  is_self_review('Cedric Novak')  = True
  is_self_review('Cédric Novák')  = False   <-- exclusion does not fire
```

`is_branch_author` compares `.lower()` bytes, so `Novák != Novak`. Four roster personas have diacritic display names against ASCII-folded aliases (`Tomasz Wójcik`, `Carolina Méndez-Ríos`, `Petra Vidović`, `Cédric Novák`), and `Cedric Novak` / `Cédric Novák` are duplicate roster entries on one address with both spellings live in git history.

**Explicitly not a must-fix on this PR**, and I want the reasoning on the record rather than assumed: the gap is in `charter_trailer` and predates #1210, it affects the ref-prefix arm identically, and #1215 leaves the affected personas strictly better off than before (previously *no* exclusion existed on these refs at all; now three of the four are excluded and the fourth is excluded in one of two spelling directions). Fixing it means normalizing in `_name_tokens`, which touches the 86.7% path this PR deliberately left alone — and folding only ever widens a match, so that fix is non-relaxing by the same argument as yours.

**Nits, no issue filed.** `commit_author_identities` is simultaneously a module function, a keyword parameter of `check_comment_reviews`, and a `ReviewVerdicts` field; inside `check_comment_reviews` the parameter shadows the function. Harmless today (the function is not called there) and a miscall would be a loud `TypeError`, but given #1046's history with duplicate-name drift it is worth a rename at the next touch. Separately, `latest_content_commit` still does `commit.get("commit", {}).get("committer", {})`, which raises on a null `commit` object, while your new `commit_author_identities` guards the same access with `or {}` — pre-existing asymmetry, now sitting two functions apart.

## Charter compliance

Commit identity `Aino Virtanen <parametrization+Aino.Virtanen@gmail.com>` — correct, single commit. Branch `A.Virtanen/1210-hook4-branch-author-from-commit-identity` — conforms. Base `main`, correct for direct-to-main wave-29. Branch is current with `main` (merge-base `17b8e41` = `main` HEAD). **All 14 CI checks pass at `0186d4d`**, including mypy, both pytest suites, ruff format + lint, the pre-commit/CI sync-drift gate, and `Verify commit authors are roster members`. `pytest` on the two touched suites in my own worktree: 350 passed, 53 subtests.

## What I could not verify

- **Pagination past 100 commits** is asserted only by the `--paginate` flag; no PR in the corpus exceeds 100 commits, so I could not exercise the boundary live.
- **The 655-PR exposure figure** in the body — I did not reproduce it. My sweep covered the 184 non-persona-ref PRs, which is the population this change can affect.
- **The `--admin` short-circuit interaction** with finding 1 is read from `check()` line 2087 and the charter clause; I did not execute an admin merge.
- **Residual 3 (`ontology/checksums.json` not marked dirty)** — I confirmed the file is absent from the diff but did not independently verify the tracker's `$CLAUDE_PROJECT_DIR` behaviour; #1219 tracks it.

## Verdict

**Approved.** No fail-open is introduced — verified analytically across every allow path and empirically across 184 historical PRs with zero relaxations. Bot neutrality, the bare-principal refusal, and the #1172 identity comparison all hold under mutation. The scope decision is the right one and the majority path is provably untouched. The three findings are all fail-closed or pre-existing, none blocks merge, and all are filed.

parametrization:
# Merge-gate re-review (round 2) — PR #1215

Reviewed at head `0ae49624c1a8f239f420d84a25b0ec55f3c8e809`, fetched via the contents API and blob-SHA-matched against a pristine `git archive` tree (`5fc55cc…` / `150299b…` / `de78fb4…` all MATCH). Every mutation below ran in that isolated tree under `PYTHONDONTWRITEBYTECODE=1 python3 -B`, with the mutant's applied diff printed before any result was read — after last round's stale-`__pycache__` episode I am not willing to read a green run I cannot prove was produced by the code I think I mutated.

**Round-2 state, re-verified independently:** additive (`0186d4d` is an ancestor of `0ae4962`, one commit, no force-push); delta is exactly `.claude/hooks/tests/test_validate_pr_review.py` +201/−16 with the production hook untouched; CI 14/14 `success` pulled from `check-runs` at the head SHA, not from the PR-level rollup.

## Both must-fixes are discharged. Here is what I reproduced rather than accepted.

**The defect I blocked for, confirmed before the fix is judged.** I measured `before`/`after` for all 13 matrix entries. All 9 original `CASES` land `total_distinct` 3 or 4 — every one of them allowed-before. The guard body of `test_a_blocked_pr_never_becomes_an_allowed_one` therefore executed **zero** times at round 1. All 4 `BLOCKING_CASES` are blocked-before (1 → 0, 1 → 0, 1 → 1, 1 → 1). Blocked-before count went 0 → 4.

**The anti-vacuity guard is itself non-vacuous — the check you flagged as highest-value.** I emptied `BLOCKING_CASES` to `()` and confirmed the mutant applied. The test FAILS on `assertTrue(reached, …)` with `AssertionError: [] is not true`. I then re-ran the same emptied matrix with the `assertTrue` block deleted, leaving only `assertEqual(reached, list(BLOCKING_CASES))`: **3 passed.** Her self-report is exact — the first fix was vacuous, `assertEqual` alone degrades to `assertEqual([], [])`, and shipping it would have restored #1203 inertness *inside the fix for #1203 inertness*. Catching that in her own work, and then documenting why both assertions are load-bearing rather than quietly adding one, is the standard this gate exists to hold.

**`_gate_allows` is a faithful transcription.** `check()` allows on `total_distinct == 1 and wave_bootstrap_exception` (fall-through) or blocks on `total_distinct < 2` — the helper is `>= 2 or (== 1 and wave_bootstrap_exception)`. Exact.

**M13, re-run.** With the TechDebt block folded inside `if not is_self_review(…)` (diff printed, re-indent verified), both new tests fail and the end-to-end one reports `check() returned None` — a real BLOCK → PASS transition, not a changed count. **Sole-survivor claim reproduced:** mutant applied, the new class deselected, full hooks+lib suites run — everything else passes. My tree is a `git archive` with no `.git`, so `test_wave_status.py::BaseVsHeadDifferential` fails on a `git show`; I confirmed that failure is environmental by reproducing it on pristine code in the same tree. Baseline at head in that tree: **3312 passed, 647 subtests.**

**Her caveat is stated honestly and is true.** I built the fail-open refactor as two edits — drop the ref-precedence guard, then let the commit arm supersede the ref arm in `is_self_review` — and ran it. `test_reviewer_set_never_grows` goes RED on two subtests; `test_a_blocked_pr_never_becomes_an_allowed_one` stays GREEN. So the subset property is the load-bearing one, and must-fix 1 bought a tripwire rather than live coverage.

**On whether the docstring discharges the naming concern: yes.** The scope note says so in plain language in its first sentence, calls itself the TRIPWIRE, and names the sibling as load-bearing. It is a milder cousin of the inert-test problem only in the sense that it cannot fail today; it is categorically different from #1203 because the assertion now *executes* on 4 real blocked inputs and is pinned to keep executing. A test that runs its assertion and cannot currently fail is a regression guard. A test whose assertion never runs is furniture. The name describes the property asserted, the docstring says exactly what it can and cannot catch, and no reader of that docstring can overclaim from a green run.

**Blocking case 3 is 1 → 1, as she corrected.** Measured: `before` = `{nino kavtaradze}` (Lucas dropped by the ref prefix), `after` identical. `resolve_review_verdicts` passes `()` for commit identities whenever the ref names a persona, so the commit arm is silent. Non-stacking confirmed.

## One new finding, non-blocking

That case's comment claims one thing more than the case does: *"This case pins that the two sources cannot stack into a double exclusion on the same PR."* It does not pin it. I ran the stacking mutant on its own — drop only the `() if branch_author_lastname` guard, leave `is_self_review` ORing both arms — and case 3 becomes 1 → 0. That is still a subset, and still blocked-before-and-after, so **both** `StrictlyNonRelaxingTests` tests stay green. Full-suite run: exactly one test goes RED, and it is your round-1 `CommitIdentitySelfReviewExclusionTests::test_persona_ref_does_not_consult_commit_identity`, which asserts `commit_author_identities == ()` directly.

So the property is genuinely enforced — just one class up, not by this case. The case *records* non-stacking; the other test *pins* it. Given this PR's whole subject is claims of coverage that exceed what the fixture verifies, I would rather the word not be "pins". This is a comment edit, it needs no re-review, and I am not gating on it.

Same class, same non-blocking bucket: the PR description is round-1 vintage. It still says "over a **9-case** matrix … plus a non-vacuity case" (now 13 cases and two anti-vacuity assertions), reports **3311 passed, 639 subtests**, and its mutation table has no M13 row and no empty-matrix row. The description is the permanent merge record for a PR about not overstating coverage; please refresh it before the owner merges.

## Residual I am accepting, stated so it is on the record

`_gate_allows` is a transcription of `check()`'s ladder living in the test file, so a future change to the real ladder would not turn this test red. That is deliberate and documented, the alternative is the full mock stack the sibling class already carries, and I confirmed the real ladder is pinned elsewhere (`SingleReviewerExceptionTests`, plus 4 tests asserting the `required peer reviews` block reason). Not debt worth an issue.

## Merge readiness

Correctness-approved at `0ae4962`. **Not yet mergeable by the gate:** Nino Kavtaradze's Approved was cast `18:53:05Z` and the head commit is `22:24:45Z`, so under this hook's own #950 staleness rule that verdict is STALE and the PR sits at 1/2 current approvals. He needs to re-post at `0ae4962`. Merge remains owner-gated; I have not merged and have not used the reviews API.

---
Requestor: Nadia Khoury
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Re-review at `0ae4962` — Approved (re-affirmed, with verification)

My earlier Approved was cast at `0186d4d` and went stale when the head moved. Re-affirming rather than redoing the full pass, but I verified the delta myself rather than taking "test-only" on trust.

### Delta is genuinely test-only and additive

- `git merge-base --is-ancestor 0186d4d 0ae49624` → true. One commit on top, no rebase.
- `git diff --stat 0186d4d..0ae49624` → `.claude/hooks/tests/test_validate_pr_review.py | 217 +++++--` , **1 file changed, 201 insertions(+), 16 deletions(-)**. The production hook is untouched, consistent with the ruling that both must-fixes were test-only.
- 312 passed, 61 subtests at `0ae4962`; CI 14/14 pass.

### My sole-survivor finding — reproduced at the new head, both directions

Applied M13 to `check_comment_reviews` (moved the TechDebt block under the `not is_self_review(...)` branch) in a throwaway checkout at `0ae4962`:

| run | result |
|---|---|
| mutant, full file | **2 failed** — both tests of `SelfReviewExclusionKeepsTechDebtBlockTests` |
| mutant, that class deselected | **310 passed, 0 failed** |

So the mutant is still a sole survivor of everything else in the suite, and the new class is precisely and solely what kills it. That is the strongest form of the claim, and it holds.

`test_check_still_blocks_the_merge_on_the_authors_missing_attestation` is the one I'd have asked for: it pins the merge *decision*, and pinning the reason string (the attestation-line message plus the author's name) forecloses the failure mode where an unrelated block satisfies a bare `decision == "block"`.

### The vacuity-of-the-anti-vacuity-fix catch — also reproduced

Worth recording that this was caught and fixed rather than shipped. I emptied `BLOCKING_CASES` at `0ae4962`:

- with both assertions as shipped → `test_a_blocked_pr_never_becomes_an_allowed_one` **FAILS** (`assertTrue` fires).
- with `assertTrue` removed, `assertEqual` alone → **PASSES**, i.e. `assertEqual([], [])`.

So `assertTrue` is load-bearing, not belt-and-braces, and the ordering comment is accurate. A test written to remove #1203 inertness that reintroduced it under one deletion is a genuinely subtle trap; catching it by running the mutation instead of reasoning about it is the right instinct.

### Load-bearing comment I spot-checked

The `L.Ferreira/1210-x` case asserts the commit arm is silent "BY CONSTRUCTION" whenever the ref names a persona, which is what makes the 1 -> 1 count meaningful rather than a coincidence. Verified at `validate_pr_review.py:2020`: `() if branch_author_lastname else commit_author_identities(commits)`. Accurate — the two exclusion sources cannot stack.

The honest scope paragraph on `test_a_blocked_pr_never_becomes_an_allowed_one` — that it is a tripwire and the subset sibling is the load-bearing one — is the right thing to write down rather than let a green run imply more than it proves.

### Nit, non-blocking, no issue filed

`_gate_allows` transcribes `check()`'s allow ladder into the test file. If that ladder ever changes, the transcription drifts silently: the test keeps passing while quietly meaning something else. The docstring is upfront that this is a deliberate transcription and it is confined to a tripwire test, so it does not justify an issue — but if a third copy of that ladder ever appears, it should become a shared helper.

Requestor: Nino Kavtaradze
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Issue #1222, opened 2026-08-02
bug(shell-parse): a FAILING cd does not abort the rest of a multi-line command — subsequent lines run in the previous cwd (live near-miss on PR #1213)


parametrization:
> **Targeted at wave-30.** Filed unlabelled for a wave deliberately: `wave-30` does not exist yet (it is created at wave-30 kickoff), and adding a `wave-29` label would create false scope drift against that wave's labelled==scoped invariant. Recorded as a phase-10 carry-forward so wave-30 scoping picks it up. Owner decision 2026-08-02.

## Summary

A `cd` that **fails at runtime** does not abort the remaining lines of a multi-line Bash tool call. Execution continues in the *previous* working directory, and the failure is easy to miss because the surviving lines can succeed and leave `$?` at `0`.

This is the runtime sibling of #1151 (*"honours a `cd` the shell never runs"*) and #1169 (*wrapper-prefixed cwd mover*). All three break the same invariant: **the cwd a gate predicts by parsing the command must equal the cwd the command actually runs in.** #1151 was a `cd` the shell skips; this is a `cd` the shell runs and that fails.

## It already cost us — a live near-miss, not a hypothetical

During PR #1213 (issue #1212) round 2, an implementer constructing scratch git fixtures had a `cd` fail because an earlier `rm -rf` had removed the target. The remaining lines of that command executed **against the live PR worktree**, creating two junk commits (`init`, `feature A adds ws.txt`) directly on the real feature branch plus a stray tracked file.

It was caught pre-push by `git status`/`git log`, and origin was never polluted — independently verified afterwards: the branch carries exactly its two legitimate commits, and all five regression-fixture branches remain at their original SHAs. Recovery took a `git stash` → `git reset --hard` → `git stash pop`.

**Why this is worth a gate and not just a note:** the recovery required the implementer to *notice*. Had the junk commits been pushed, they would have landed on a branch under active review by two reviewers anchored at a specific SHA.

## Mechanism, reproduced

```console
$ cd "$D/real"                      # a live worktree
$ cd "$D/deleted_dir"               # fails
(eval):cd:5: no such file or directory: .../deleted_dir
$ echo "pwd after failed cd: $(pwd)"
pwd after failed cd: .../real       # <-- still the live worktree
$ ls
LIVE_WORKTREE_MARKER                # <-- subsequent lines ran HERE
$ echo $?
0                                   # <-- no surviving error signal
```

[...]

PR #1225, opened 2026-08-02
feat(tooling): REST fallback + PreToolUse gate for gh GraphQL quota exhaustion

### LINKED ISSUE #1224: feat(tooling): automatic REST fallback when GitHub GraphQL quota is exhausted — gh's --json read surfaces all fail while REST sits untouched
## Summary

GitHub's REST and GraphQL quotas are **separate** (5,000/hr each). `gh`'s ergonomic read surfaces — `gh issue view/list`, `gh pr view --json`, `gh pr list --json`, `gh pr checks` — are **GraphQL-backed**, so when GraphQL is exhausted every one of them fails while **REST sits nearly untouched**. Today the org has no automatic fallback: agents hit `GraphQL: API rate limit already exceeded`, and either stall, retry into the same wall, or produce degraded work.

Owner-requested 2026-08-02: *"setup our workflow for when api limits happen in github we switch to rest."*

## Measured, not assumed

Observed live at `graphql=0/5000, core=4991/5000` (quota burned by four concurrent review agents in one session):

| surface | under GraphQL exhaustion |
|---|---|
| `gh issue view --json` | **FAIL** |
| `gh issue list` — **even without `--json`** | **FAIL** |
| `gh pr view --json <any fields>` | **FAIL** (returned OK once, FAIL on repeat — flaky, not REST) |
| `gh pr list --json` | **FAIL** |
| `gh pr checks` | **FAIL** |
| `gh api repos/…` (REST) | **OK** |
| `gh pr diff` | **OK** |

Two findings worth calling out because they are counter-intuitive:

1. **`gh issue list` fails even with no `--json` flag** — the GraphQL dependency is not about output formatting, so "drop `--json`" is not a workaround.
2. **`gh pr view --json` is flaky rather than cleanly REST or GraphQL.** It succeeded once and failed on immediate repeat with an identical field set. Treat any single success as unreliable; do not build a fallback that depends on it.

`gh api rate_limit` does **not** consume quota, so the guard is free to call.

## Impact seen this session

Four parallel agents (two implementers, four reviewers across two PRs) exhausted the shared hourly GraphQL quota. Consequences:

- Two queued PR re-reviews had to be **held ~40 minutes** rather than dispatched into a dead API, because a reviewer's first action is `gh pr view`, which would fail at step one.
- An implementer hit it mid-task and worked around it ad hoc with `gh api … --input <file>`; that knowledge stayed in his transcript rather than becoming reusable.
- The orchestrator's own drift audit had to be re-done through REST with `--paginate`, which surfaced a **second** latent bug: `gh api …/issues` returns **pull requests as well as issues**, so a naive count read 57 where the true issue count was 50. Any REST fallback must filter `select(.pull_request == null)` or it silently overcounts.

## Proposed shape

Three layers, in the org's usual enforcement order (`hook > skill > charter`, `feedback_enforcement_hierarchy`).

### 1. `.claude/lib/gh_quota.py` — the sensor

- `check` → per-resource `remaining` / `limit` / `reset` for `core` and `graphql`, plus seconds-until-reset.
- `guard --resource graphql --min N` → exit non-zero when below threshold, for use in skills and scripts.
- Must never itself consume quota (`/rate_limit` is free) and must degrade to "unknown, assume available" on network failure rather than blocking work.

### 2. `.claude/lib/gh_rest.py` — the fallback

REST equivalents for the queries this org actually runs, each returning the same shape the `--json` form does so call sites need no reshaping:

- issue: number, state, title, **labels**, closedAt
- issue list by label + state, **filtering out PRs** (see the overcount bug above)
- PR: number, state, head SHA, base ref, headRefName, mergedAt, mergeCommit
- PR comments; PR check status (via `commits/<sha>/check-runs`)
- PR timeline events (e.g. `head_ref_force_pushed`)
- project-board membership for an issue

Board **writes** (`gh project item-add`) are GraphQL-only with no REST equivalent — the helper must say so explicitly rather than silently failing, and the caller should defer that work until reset.

### 3. A PreToolUse hook — the gate

Mirroring `block_bare_grep` (which already blocks `grep` and prints the `rg` rewrite): when a GraphQL-backed `gh` invocation is attempted **and** GraphQL quota is exhausted or below threshold, intervene with the concrete REST rewrite for that command.

**Block vs advise is a real design question, not a detail.** Blocking is right where an exact REST equivalent exists; where none does (project mutations), blocking would strand the agent with no path forward, so an advisory naming the reset time is better. Reviewers should push on this — a hook that fails closed on a query with no fallback converts a slowdown into a hard stop.

The hook must not call the network on every Bash command. Gate on a cheap command-pattern match first, and cache the quota reading briefly.

### 4. Documentation

`docs/TOOLCHAIN.md` § Shell environment (alongside the `rg`-not-`grep` rule it parallels exactly), `ontology/conventions.md`, and the spawn-brief boilerplate so subagents inherit it. Fold the measured table into the existing `feedback_gh_cli_gotchas` memory — which already documents `item-list --limit` truncation and "never trust exit 0 on a `gh` mutation" — rather than creating a competing note.

## Acceptance criteria

- [ ] `gh_quota.py check` reports both resources and seconds-to-reset without consuming quota.
- [ ] `gh_rest.py` covers every query in the list above, returning the same shape as the `--json` equivalent.
- [ ] Issue-list fallback **excludes pull requests** — a regression test pins the 50-vs-57 overcount.
- [ ] Paginated REST reads use `--paginate`; a test pins that a >100-item result is not silently truncated (sibling of the `item-list --limit` trap already in `feedback_gh_cli_gotchas`).
- [ ] The hook fires only on GraphQL-backed patterns, and only when quota is actually low — verified by test, not by inspection.
- [ ] The hook does **not** hard-block a query with no REST equivalent; it advises and names the reset time.
- [ ] Quota-check failure (offline, malformed response) degrades to allowing the command, never to blocking.
- [ ] Docs updated in all three places; memory note folded into `feedback_gh_cli_gotchas`.

## Notes

- Related: `feedback_gh_cli_gotchas` (existing memory — `item-list --limit` truncation, silent `gh` mutation failures), #1201 (silent truncation at a helper's own edges).
- The deeper mitigation is **using less quota**: four concurrent agents each polling `gh pr view` is the actual cause. Worth considering as a follow-up — quota-aware spawn throttling, or a shared cached PR-state read — but out of scope here.

parametrization:
## Summary

Promotes `feedback_gh_cli_gotchas` §12 from a memory note to enforcement (Closes #1224). §12 documented the core/graphql quota split and the silent-zero failure mode for ~2 weeks without preventing a recurrence — four concurrent agents exhausted GraphQL again in the session that filed #1224, holding two PR re-reviews ~40 minutes.

## Three layers built

1. **`.claude/lib/gh_quota.py`** — the sensor. `gh api rate_limit` costs no quota (measured); TTL-cached (`.claude/.consulted/gh_quota/cache.json`, default 30s); degrades to "unknown" (never "zero") on any failure.
2. **`.claude/lib/gh_rest.py`** — the fallback. REST equivalents for `issue view/list/timeline`, `pr view/list/comments/checks`, the comment write-then-post two-step, and ProjectV2 `add`/`items`/`membership`. Every function raises rather than returning `[]`/`{}` on a genuine failure. Two traps pinned by test: (a) `repos/{o}/{r}/issues` returns pull requests too (the 57-vs-50 overcount); (b) `--paginate` output is one compact JSON object per line across every page, so a >100-item result is never silently truncated.
3. **`.claude/hooks/gh_quota_gate.py`** — the gate (PreToolUse). Cheap command-pattern prefilter first (no network for a non-`gh`-shaped command), then the cached quota sensor. Blocks a matched GraphQL-shaped `gh` call with the concrete `gh_rest.py` rewrite when quota is low; advises (allow + warning, never blocks) when no rewrite is derivable — a raw `gh api graphql` call, or an unmapped project mutation (`item-edit`, `field-list`). Registered in `ontology/conventions.md` + wired through `dispatcher.py`/`framework.config.json` (placed just before `block_squash_wave_merge`, keeping cheap/local-first ordering).

## A live correction to the issue's own premise

The issue assumed `gh project item-add` is GraphQL-only with no REST equivalent. Verified live (2026-08-02) that's no longer true for an **org-owned** board — GitHub now has `POST orgs/{org}/projectsV2/{n}/items`. I used it for real to add #1224 to project board 2 (which the issue itself flagged as still-blocked). `gh` the CLI tool hasn't adopted it — `gh project item-add` is still GraphQL-backed — so this is a genuine fallback, not a claim that `gh` changed. `gh_rest.py`'s module docstring and `feedback_gh_cli_gotchas.md` §12 both document this correction with the verified request shape.

## Block vs advise

Block whenever a verified, derivable REST rewrite exists for the *specific* invocation matched (covers every pattern in `_BLOCK_PATTERNS`, reads AND writes — including the corrected `project item-add`). Advise-only for exactly the shapes with no derivable rewrite: a raw `gh api graphql ...` call (arbitrary payload, nothing to introspect) and any recognized-but-unmapped `gh project` verb. Hard-blocking those would convert a slowdown into a dead end, which is the exact failure the issue warns against.

## Degrade-to-ALLOW

A quota-check failure (offline, malformed `rate_limit` response, unwritable cache) returns `None` (unknown) from the sensor, and the hook treats `None` as ALLOW, never BLOCK — pinned by test (`test_unknown_quota_never_blocks_even_for_a_matched_pattern`, `test_guard_allows_when_quota_unknown`).

## Testing

70 new tests across three files (`test_gh_quota.py`, `test_gh_rest.py`, `test_gh_quota_gate.py`), all hermetic (subprocess mocked, no live-network dependency for CI). Test names map directly to the four traps in the issue brief (PR-overcount filter, pagination truncation, project-board explicit-failure contract, comment two-step split) plus the degrade-to-allow and block-vs-advise design requirements. Mutation-tested 5 core invariants live during implementation (boundary `>=`, trap-1 PR filter, trap-4 file-exists guard, pagination `.[] |` flatten, item-add block-vs-advise routing) — every mutation produced a red test, then was reverted.

## Docs + memory

`docs/TOOLCHAIN.md` § GitHub API quota (sibling of § Text search: `rg`, not `grep`) and `ontology/conventions.md` (hook table + a matching § GitHub API quota section). Findings folded back into `.claude/memory/feedback_gh_cli_gotchas.md` §12 (re-derived measured table, the ProjectV2 REST-write correction, enforcement-promotion note) rather than a competing note, per the CLAUDE.md update-on-contradiction norm.

## Residual / out of scope

- `gh issue edit` — the old §12 claim ("already REST") is untouched; I didn't have a live measurement either confirming or contradicting it this session, so I left it alone rather than guessing.
- `gh_rest.py` has no wrapped helper for `gh issue create` (which I measured live as GraphQL-backed, contradicting the old "already REST" assumption) — the hook's rewrite gives the raw `gh api POST` recipe instead of a Python wrapper; building that wrapper is a natural, small follow-up if it gets hit often.
- ProjectV2 field-value mutations (`item-edit`, `field-create`, `field-list`) are left in the hook's advise-only bucket; I did not attempt to discover whether GitHub's newer Projects REST API extends to those too — that's a genuine "did not verify" gap, not a claim they're unsupported.

Closes #1224


parametrization:
## CI is green; the parity gap you found is filed as #1228

Verified independently rather than from the report: head `441dd15`, **21/21 pass / 0 fail**, and `git merge-base --is-ancestor eb60268 441dd15` confirms the fix is additive with no force-push.

**Your out-of-scope observation was right, and it is worse than a missing hook.** I confirmed it and filed **#1228**. `.pre-commit-config.yaml` has zero markdownlint hooks while `docs.yml` runs a `Markdown lint` job — but the notable part is that the parity gate itself is silent:

```console
$ python3 .claude/lib/pre_commit_ci_sync.py .
OK: pre-commit config mirrors all CI-enforced checks.
$ echo $?
0
```

A false green from the gate whose whole purpose is catching exactly this. Root cause is that `_KIND_PATTERNS` has no markdown kind, and the gate ignores unknown tools — the same blind spot #684 closed for cspell, surviving because #684 fixed the instance and not the contract. `pre_commit_ci_sync.py` line 56 states the invariant this breaks in its own docstring.

**This vindicates your process, so it should not read as a mark against you.** You ran the complete local gate — commit stage and pre-push — before pushing, and there was no local signal available to produce. That is the defect, not your verification. You were also right to leave `.pre-commit-config.yaml` untouched here; it belongs in its own reviewed change.

#1228 is filed **without a wave label** on purpose (the #1222 precedent): `wave-30` does not exist until kickoff, and a `wave-29` label would create false drift against that wave's labelled==scoped invariant, which was just restored at 53 items. It will be carried into wave-30 scoping.

Reviewers are being assigned now that CI is green. This PR is not merged — the owner gates every merge.

parametrization:
Requestor: Weronika Zielinska
Requestee: Lucas Ferreira
RequestOrReplied: Approved
TechDebt: none

**Review: LGTM**

Reviewed the artifact at head `441dd15` (fetched via `gh api .../contents?ref=441dd15`, not diff text), ran the code mentally against every failure path in the three new modules, and cross-checked the wiring (`_framework_config.py`, `dispatcher.py`) and the memory/docs updates against the actual code.

**The specific risk (can a fallback return a value indistinguishable from a genuine empty/failure?) — traced, not found.**

The design deliberately splits two different correctness obligations across the three layers, and gets both right:

- `gh_quota.py` (the sensor) is REQUIRED to degrade-to-allow on failure — that's the #1224 design mandate, not a bug. It does this correctly by using `None` as an explicit unknown sentinel that is never conflated with a real zero-quota reading: `fetch_rate_limit`/`get_quota`/`resource_quota` all return `None` only for genuine unknown (network/parse/missing-key failure), while a real `remaining=0` reading is a fully-populated `ResourceQuota` object, structurally distinct from `None`. `gh_quota_gate.check()` treats `None` (and only `None`) as allow — a matched-but-quota-known-zero case still blocks.
- `gh_rest.py` (the fallback) is REQUIRED to do the opposite — never mask a failure as empty. Every function traced (`issue_view/list/timeline`, `pr_view/list/comments/check_runs`, `post_comment`, `project_item_add/items/membership`) raises `GhRestError`/`GhRestUnsupportedError` on a non-zero exit, timeout, or unparseable response (`_run_gh`, `_api_json`, `_api_paginate_jq`, `_post_project_item`) rather than returning `[]`/`{}`/`None`. An empty list only comes back on a verified exit-0 + parseable-empty-body result (`_api_paginate_jq`: non-zero exit or a bad line raises before the empty case is ever reached). `issue_project_membership` returning `None` for "not on the board" is safe specifically because its one dependency, `project_items()`, would have raised already had the underlying fetch failed — so `None` there is a verified negative, not a masked failure. The CLI entrypoint (`main()`) mirrors this at the process boundary: `GhRestError`→exit 1, `GhRestUnsupportedError`→exit 3, both to stderr — a shell caller piping this can't mistake a failure for exit-0-empty either.
- `gh_quota_gate.py`'s own failure mode: a quota-check failure returns `None` from the sensor, which the gate maps to plain `return None` (allow, no block) — matches the module's own stated contract, and `dispatcher.py`'s `except Exception: continue` around each hook's `check()` call means even an unanticipated crash in this hook degrades to allow, not a stuck-block. Consistent, not accidental — same direction as the sensor.

**Other things verified, not just trusted:**
- CI: 21/21 pass at `441dd15` (`gh pr checks 1225`), confirmed independently of the PR body's claim.
- Additive: full commit history is 2 commits, the second (`441dd15`) is the isolated MD018 docs fix over `eb60268`'s substantive work — no rewrite, no squash.
- Dispatcher wiring matches the PR body's claim exactly: `gh_quota_gate` sits between `warn_zsh_wordsplit` and `block_squash_wave_merge` in `_framework_config.py`'s `hooks.pre_bash` default list, with the cheap-prefilter-before-network ordering documented and actually implemented (`_match()` short-circuits on no gh-shaped command/pattern before `gh_quota.resource_quota` is ever called).
- The PR body's item-add correction is real and consistently threaded through all three doc/memory surfaces touched (`feedback_gh_cli_gotchas.md` §12, `docs/TOOLCHAIN.md`, `ontology/conventions.md`) — no stale "GraphQL-only" claim survives anywhere I checked, including the PR body itself.
- `gh_rest.py` reuses the pre-existing `_shell_parse.py`/`_repo_flag_parse.py` helpers rather than reinventing command parsing — confirmed those modules already exist on `main` outside this diff.
- Test names in the three new suites (`test_paginate_nonzero_exit_raises_not_empty_list`, `test_empty_but_successful_result_is_a_legitimate_empty_list`, `test_project_item_add_raises_not_silent_on_post_failure`, `test_unknown_quota_never_blocks_even_for_a_matched_pattern`, `test_guard_allows_when_quota_unknown`) map directly onto the exact risk this review was asked to attack, and CI's pytest job is green at head.

**Took on trust (didn't independently re-run):** the 70 new tests' actual pass/fail — relied on the green `Pytest — hooks and lib test suites` CI check at `441dd15` rather than executing them locally in this session, since this repo's working tree isn't checked out to the PR branch and I didn't want to touch it.

**Not required here, tracked separately:** the `.pre-commit-config.yaml` markdownlint gap that let the MD018 issue reach CI red — correctly out of scope for this PR, per the assignment.

parametrization:
## Merge-gate review — changes requested (one must-fix)

Reviewed the artifact at head `441dd15` (`gh api .../contents?ref=441dd15`), then **ran it**: the 70 tests, a 24-mutation battery against the guards, the three CLIs against the live GitHub API, and the hook end-to-end through `dispatcher.py` with a forced-low quota cache. Findings below separate what I reproduced from what I took on trust.

## The question this review was asked to decide

**Does any fallback path flatten a failure into a value that reads as a real result — the #1226 shape, one layer down?** Traced and executed. **No.** Verified live, not by reading:

```console
$ python3 .claude/lib/gh_rest.py issue list --repo noorinalabs/does-not-exist-xyz
gh_rest: ERROR — gh api repos/noorinalabs/does-not-exist-xyz/issues?state=open --paginate ... failed (exit 1): gh: Not Found (HTTP 404)
rc=1

$ python3 .claude/lib/gh_rest.py issue list --repo noorinalabs/noorinalabs-main --label zzz-no-such-label
[]
rc=0
```

Could-not-query and genuinely-empty are distinguishable at the process boundary, not just in the type system. Trap 1 quantified against live data: `gh_rest.issue_list(label=wave-29, state=all)` returns **53**, the search API's issue-only count is **53**, and the raw unfiltered `/issues` endpoint returns **61** — exactly 53 + the 8 PRs carrying that label. The `57-vs-50` overcount cannot recur through this module.

The sensor's half is the mirror image and equally sound: `None` (unknown) is structurally distinct from a real `ResourceQuota(remaining=0)`, so no "sensor down" reading can masquerade as "quota exhausted", and none of it can masquerade as data. Unlike #1226, partial failure here is not quieter than total failure.

## Must-fix — the gate does not see a `gh` call inside `$(...)`, backticks, or a loop body

`_match()` segments with the shlex-only `iter_command_segments`. Reproduced end-to-end through `dispatcher.py` at `graphql remaining=5`:

| command | verdict |
|---|---|
| `gh pr view 1225 --json headRefOid -q .headRefOid \| cat` | BLOCK |
| `SHA=$(gh pr view 1225 --json headRefOid -q .headRefOid)` | **no match, silent allow** |
| ``echo `gh issue view 1224 --json state` `` | **no match, silent allow** |
| `for i in 1 2; do gh issue view $i; done` | **no match, silent allow** |
| `gh issue view 1224 2>/dev/null` | BLOCK |

This is not a hypothetical shape. The repo's own committed skills contain **27** `$(gh issue …)` / `$(gh pr …)` call sites, and `.claude/skills/wave-scope/SKILL.md:308` is:

```bash
body=$(gh issue view "$num" --repo "noorinalabs/$repo" --json body --jq '.body' 2>/dev/null || true)
```

Command substitution + `2>/dev/null` + `|| true` — the exact silent-zero triple §12 and #1224 exist to stop, and the new gate passes over it without a word. The loop blind spot has already cost this org once: `.claude/skills/wave-kickoff/SKILL.md:436` records main#1145, where `for n in 1114 1116; do gh issue edit "$n" …; done` defeated another hook's command-string parsing and 14 issues were labeled with zero kickoff comments, undetected until an unrelated audit.

The remedy is already in this repo and needs no new dependency — `block_bare_grep._segments()` (`.claude/hooks/block_bare_grep.py:205`) tries `iter_command_segments_ast` (bashlex, which walks `$(...)`, backticks and compound bodies — confirmed installed) and falls back to the shlex path. `validate_commit_identity.py` uses the same pattern. Copying those ~10 lines plus two tests pinning the `$( )` and `for` shapes closes it.

**The counter-argument, stated fairly:** eleven sibling hooks (`block_no_verify`, `block_git_config`, `block_squash_wave_merge` among them) use the shlex-only path and share this blind spot, so this is an inherited limitation, not one this PR introduced. I am still calling it blocking here, for one reason specific to this hook: those gates guard commands normally written at top level (`git commit --no-verify`), whereas `gh pr view` / `gh issue view` are written inside `$( )` about as often as not — so the miss set is roughly half the traffic the hook exists for, while `ontology/conventions.md`'s new row reads as full coverage of the listed verbs. If the owner prefers to ship now and fix in wave-30, the honest alternative is a one-line caveat in that row; I would not object to that call, but I do not think the docs can stay as they are with the gap open.

## Mutation battery — 22 of 24 killed

Every load-bearing guard was mutated individually and each mutation produced a red test: degrade-to-allow, the `>=` boundary, `_run_gh`'s non-zero raise, the paginate parse-failure raise, both trap-1 PR filters, the `--paginate` flag, the payload-file guard, the project POST failure raise, the membership repo filter, `fetch_rate_limit`'s `None`, the TTL, the guard CLI's exit code, item-add block-vs-advise routing, the block exit code 2, the kill switch, content-type inference, the raw-graphql advise route, check-runs vs legacy status, and `ensure_ascii=False`. These guards are live, not decorative.

One real survivor -> **#1232**: `_api_json`'s `isinstance(data, dict)` raise can be replaced with `data = {}` and all 70 tests stay green; with that mutation `post_comment()` returns `{"id": null, …}` at exit 0. The shipped code is right, the fixture is missing. (Two other survivors are equivalent mutants and need nothing — noted in #1232.)

A caution for anyone repeating this: my first battery run was **invalid**. A same-size mutation (`return 0` -> `return 1`) restored within the same filesystem second left a stale `__pycache__` entry that Python did not invalidate, poisoning nine subsequent results into false KILLs. Re-run under `PYTHONDONTWRITEBYTECODE=1`. The numbers above are from the clean run.

## Also verified

- **CI at the head SHA**: 21/21 success via `statusCheckRollup`, re-confirmed against `gh api commits/441dd15/check-runs`, and a third time through the PR's own tool — `gh_rest.py pr checks 441dd15` returns 21 runs, all `success`. Additive over `eb60268`, no force-push.
- **Hook routing end-to-end** through `dispatcher.py`: block for the nine mapped patterns, advise for `gh api graphql` and `project item-edit`, no match for `gh api repos/...`, `gh pr diff` and `rg`; kill switch honoured. The `{"decision": "allow", "systemMessage": …}` contract matches `dispatcher.py:69-79` and ten sibling hooks.
- **The ProjectV2 REST correction is real and the read side works against the live shape** — `orgs/{org}/projectsV2/2/items` returns a top-level array whose `content.repository.full_name` and `content.number` are exactly what `project_items()`' jq filter reads. I used `project membership` for real to read-back-verify the two tech-debt issues I filed onto board 2. The withdrawn "no REST equivalent" claim does not survive anywhere — not in the PR body, not in either comment, not in §12, TOOLCHAIN.md or conventions.md.
- **Residuals are honestly scoped.** `gh issue edit` left alone for want of a measurement, no `gh issue create` wrapper, ProjectV2 field mutations left in advise — each declared rather than papered over. The `.pre-commit-config.yaml` markdownlint gap is correctly out of scope (#1228).

**Took on trust:** that `gh issue comment`, `gh issue close` and `gh issue edit` are GraphQL-backed — I did not measure them, because measuring a write means performing one; that is why #1231 asks for the measurement rather than asserting the answer. I also did not exercise `project_item_add` (a write) or `classic_project_lookup`.

**Nit, no issue filed:** this PR edits `ontology/conventions.md` (a checksum-tracked overlay file) without an `ontology/checksums.json` entry — 5 of the last 6 commits touching that file carried one. `/ontology-rebuild` reconciles it; flagging so it is a decision rather than a surprise.

**On #1226:** complementary, no contract collision. If `validate_wave_audit` adopts `gh_rest.issue_list`, it inherits an exception on failure instead of an empty list — which is most of what that fix needs.

---
Requestor: Nadia Khoury
Requestee: Lucas Ferreira
TechDebt: #1231, #1232

parametrization:
Requestor: Weronika Zielinska
Requestee: Lucas Ferreira
RequestOrReplied: Approved
TechDebt: none

**Review: LGTM (re-cast at `13062d0`, supersedes my stale Approved at `441dd15`)**

**Re-verified independently for this cast:**
- `13062d0` is additive on `441dd15` (`git merge-base --is-ancestor`), CI 21/21 pass at `13062d0`.
- Diff scope confirmed via `git diff --stat 441dd15 13062d0`: exactly `.claude/hooks/gh_quota_gate.py` (+29/-3) and its test file (+36) — `gh_quota.py`/`gh_rest.py` untouched, so my prior trace of the #1226 flatten-error-to-empty risk across those two modules is unaffected and stands as-is (not re-run this pass).
- **The prefilter-before-network property I flagged in my first review: confirmed still holding, checked myself, not taken on your word.** In `check()`, the substring gate `if not command or "gh" not in command: return None` (line 281-282) executes and returns before `_match()` (line 284) is ever called, and the new AST parse only happens inside `_match()` → `_segments()` — so a non-gh Bash command still short-circuits before any parsing, cheap or otherwise. More importantly, the property that actually matters for the design (no *network* call for a non-matching pattern) is unchanged: `gh_quota.resource_quota(\"graphql\")` is still called only after `_match()` returns a hit, confirmed by `test_gh_command_not_in_pattern_table_never_touches_quota` (`gh pr diff`, which is gh-shaped but pattern-unmatched, asserts `resource_quota` is never invoked).
- `_segments()`'s AST path (`iter_command_segments_ast`) is not new/risky machinery — it's the same memoized (`lru_cache`), fail-open (`except Exception: return None` on any bashlex failure, falling back to the pre-existing shlex path) helper `block_bare_grep.py` already runs in production on `main`, unmodified by this PR. The fix is a narrow reuse, not a new parser.
- The 3 new tests (`test_command_substitution_dollar_paren_matches`, `test_backtick_command_substitution_matches`, `test_loop_body_matches`) pin exactly the 3 shapes Nadia's must-fix named, each asserting a block with the correct `gh_rest.py` rewrite text — read them, they match the fix.

**Carried forward unchanged from my first review (not re-verified this pass, no reason to given the diff scope above):** the full trace of `gh_quota.py`/`gh_rest.py` against the flatten-error-to-empty risk, the docs/memory correctness, the dispatcher wiring, and the withdrawn item-add claim's correction — all still stand as posted at `441dd15`.

parametrization:
## Merge-gate re-review at `13062d0` — must-fix discharged

I raised the must-fix, so I judged it against the artifact and by running it, not from the fix commit's summary.

## The must-fix is closed

Reproduced the pre/post-fix behaviour myself in an isolated worktree, reverting **only** `gh_quota_gate.py` to its `441dd15` content against the new test file, under `PYTHONDONTWRITEBYTECODE=1`, asserting by sha256 that the reverted file actually landed on disk before reading any result:

```console
pre-fix  sha256: 3b2ece186fc2812c
post-fix sha256: fc789ec1ef5883c2
PRE-FIX  hook -> rc=1 | 3 failed, 26 deselected
POST-FIX hook -> rc=0 | 3 passed, 26 deselected
```

Same conclusion as the reported stash-cycle evidence, arrived at without one. `_segments()` is a faithful mirror of `block_bare_grep._segments()`, differing only in returning `[]` rather than `None` when both parsers fail — which in this caller means "no segments, no match, allow", preserving the fail-open the hook already had.

## Your two questions

**Does the AST path catch all three shapes, or do some arrive via the fallback?** Measured by forcing `_shell_parse.bashlex = None`:

| shape | bashlex ON | bashlex OFF (shlex fallback) |
|---|---|---|
| `$(gh pr view …)` | MATCH | no match |
| `` `gh issue view …` `` | MATCH | no match |
| `for i in 1 2; do gh issue view $i; done` | MATCH | MATCH |

The two substitution shapes are genuinely AST-only. The loop shape is carried by the *other* half of the change — adding `normalize_command_separators` to the fallback, without which shlex leaves `2;` as a single token and never splits. Both halves are needed; only the AST half is pinned by a test. Filed as a fourth item on #1231.

**Does the fallback still behave when bashlex fails to parse?** Yes. `iter_command_segments_ast` returns `None` rather than raising on unbalanced quotes, stray backslashes, `case` statements and unbalanced parens — the fallback engages in each, nothing crashes. Process substitution `<(gh pr view …)` is a bonus catch the AST path now surfaces.

## Real-world coverage: 0/7 -> 6/7

Not a synthetic battery — the seven call sites I cited in the first review, continuation-joined from the actual files and driven end-to-end through `dispatcher.py` at forced `graphql remaining=5`:

```
BLOCK  wave-wrapup:213   BLOCK  wave-wrapup:469   BLOCK  wave-wrapup:936
BLOCK  wave-wrapup:937   BLOCK  wave-scope:133    BLOCK  wave-scope:351
MISS   wave-scope:308
```

## A correction I owe on the record

`wave-scope/SKILL.md:308` was my flagship example for the must-fix, and **the remedy I prescribed provably cannot reach it**. The cause is not this fix:

```console
$ python3 -c "import bashlex; bashlex.parse('x=\$(a || b)')"
bashlex.errors.ParsingError          # pipelines inside $( ) parse fine; && and || do not
```

`iter_command_segments_ast` correctly returns `None`, the shlex fallback then cannot see it either because `X=$(gh` is eaten as a `KEY=val` prefix. Drop the `|| true` from that line and it blocks — one token is the whole difference.

The same hole already leaks the hook I told Lucas to copy: `X=$(grep foo bar.txt || true)` and `X=$(grep foo bar.txt && echo ok)` are **ALLOWED** by `block_bare_grep`, while `X=$(grep foo bar.txt)` and `grep foo bar.txt || true` both block. A hard-blocked footgun (#1008) walking past its own gate is a bigger finding than the one I opened this review with, and it belongs to `_shell_parse.py`, not here — **#1235**, with the three consumers named.

Holding this PR for it would be scope expansion into a shared parser used by two other gates. The must-fix as I stated it — `$(...)`, backticks, loop bodies — is discharged.

## No regressions

- **No over-match.** `echo "gh issue view 1224"`, `rg 'gh pr view' .claude/skills`, `python3 .claude/lib/gh_rest.py pr view 1225`, a quoted-delimiter heredoc body, `gh api repos/...`, `gh pr diff` and `rg highlight` all still pass through untouched. Advise-only routing for `gh api graphql` and `project item-edit` unchanged.
- **No guard weakened.** Re-ran the full mutation battery against the new head — 25 mutations, **23 killed**. Survivors are #1232's known `_api_json` gap and the untested `normalize_command_separators` noted above; every other guard, including the four I added against `_segments` itself (drop AST, drop normalize, drop both, invert the guard), produced a red test.
- **Full suites green, independently:** 3343 passed, 607 subtests, under `PYTHONDONTWRITEBYTECODE=1`.
- **Hot path intact.** The `"gh" not in command` prefilter still runs before `_match`, and `resource_quota` still sits behind a non-None match. The bashlex parse is memoized (#1113) and shared with `block_bare_grep`'s: 0.011 ms for a non-`gh` command, 1.7 ms for a 4,900-character script.
- **CI 21/21 success at `13062d0`**, and `441dd15` is an ancestor — additive, no force-push.

**Took on trust:** unchanged from the first review — that `gh issue comment`/`close`/`edit` are GraphQL-backed (measuring a write means performing one, hence #1231), and I did not exercise `project_item_add` or `classic_project_lookup`. I did not re-verify the `gh_rest.py`/`gh_quota.py` live behaviour this round; that commit is untouched since `441dd15`, where I ran it against the real API.

---
Requestor: Nadia Khoury
Requestee: Lucas Ferreira
RequestOrReplied: Approved
TechDebt: #1231, #1232, #1235

Issue #664, opened 2026-06-14
hook(validate_wave_audit): exempt wave issues with a merge-ready PR on the wave branch from the wrapup blocking-count — owner-adopted P4W7


parametrization:
## Owner-adopted process change (P4W7 retro, 2026-06-13)

Fix the `/wave-wrapup` chicken-and-egg: the `validate_wave_audit` gate counts the wave's own open work-issues and blocks `/wave-wrapup` from running — but those issues only close as part of the wrapup's own merge steps. In P4W7 this forced merge+close-first then a re-run.

## Owner decision: exempt merge-ready issues

Make the gate smarter: an open wave-labeled issue should NOT count as blocking if it has a **merge-ready PR targeting the wave branch** (`deployments/phase-{P}/wave-{M}`). Then `/wave-wrapup` can run its own merge+close steps in one pass.

## Acceptance

- [ ] `validate_wave_audit` (or the wrapup precondition) excludes wave issues that have an open/merge-ready PR based on the wave branch from the blocking count.
- [ ] `/wave-wrapup` can be invoked once (no merge+close-first re-run) when the only open wave issues are those about to be closed by the wrapup's merges.
- [ ] Guard against false-exempt: only PRs targeting the wave branch (not arbitrary PRs) qualify.

## Provenance

P4W7 retro Proposed Change #2, owner-adopted 2026-06-13 ("Exempt merge-ready issues"). Phase-5 candidate.

parametrization:
Requestor: Fatima Okonkwo
Requestee: Wanjiku Mwangi
RequestOrReplied: Request

**Wave 5 Kickoff — Phase 5**

This issue is assigned to you for p5-wave-5.
- Peer reviewer: Aino Virtanen
- Secondary reviewer: Santiago Ferreira
- Branch from: `deployments/phase-5/wave-5`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: tech-debt (per charter § Wave Planning & Priority)

Please begin implementation.

PR #1229, opened 2026-08-02
fix(#1226): partial wave-audit failure is an UNKNOWN, not a confident zero

### LINKED ISSUE #1226: bug(hook17/validate_wave_audit): a PARTIAL per-repo query failure sums to a confident zero — 1 of 8 repos failing turns the wave open-item gate into a silent ALLOW
## Symptom

`validate_wave_audit` (Hook 17) gates `/wave-wrapup`, `/wave-retro`, and `/handoff` on a
cross-repo open-item count. When **a subset** of the 8 org repos fails to query, the hook
sums only the repos that succeeded and treats that partial sum as authoritative. Because
`noorinalabs-main` is currently the **only** repo carrying wave issues (measured today:
30 open `wave-29` in main, **0** in each of the other 7), a **single** failed query — on
main — takes the gate from BLOCK to a **completely silent ALLOW**.

Not theoretical. Observed live on 2026-08-02 under GraphQL quota exhaustion:

| run | GraphQL | items counted | gate |
|---|---|---|---|
| orchestrator `/handoff`, ~19:04Z | exhausted (0/5000) | 0 | **ALLOWED** |
| implementer `/handoff`, after quota reset | recovered | 30 | **BLOCKED** |

## Reproduction

Driving the shipped `check()` with `_open_issue_numbers_for_label` stubbed to return
`None` (exactly what a non-zero `gh` exit produces) for a chosen subset, against the real
measured distribution (main=30, others=0):

| scenario | `total` | gate verdict |
|---|---|---|
| all 8 repos queried OK (baseline) | 30 | **BLOCK** |
| **only `noorinalabs-main` fails (1 of 8)** | **0** | **ALLOW — silent, no `systemMessage`, exit 0** |
| main + deploy fail (2 of 8) | 0 | **ALLOW — silent** |
| a zero-issue repo fails (1 of 8) | 30 | BLOCK |
| all 8 repos fail (total failure) | `None` | ALLOW **+ warning** |

Repro script: `scratchpad/repro_wave_audit.py` (loads the real hook module, stubs only the
subprocess-boundary function, asserts nothing about internals).

## Suspected root cause

`_audit_open_count` (`.claude/hooks/validate_wave_audit.py:519-534`) drops failed repos
on the floor:

```python
for repo in _ORG_REPOS:
    count = _count_open_for_repo(repo, labels, wave_branch)
    if count is None:
        continue          # <-- failure is discarded, not recorded
    successes += 1
    if count > 0:
        per_repo[repo] = count

if successes == 0:
    return None, per_repo # only TOTAL failure is signalled
total = sum(per_repo.values())
```

Failure is conflated with "queried, and found zero". The count then reaches
`check()` at line 605:

```python
if total == 0:
    return None           # allow, and emit NOTHING
```

`main()` turns `None` into `sys.exit(0)` with no output — so the operator sees no warning,
no note, nothing distinguishing "the wave is genuinely clean" from "the audit could not
see the only repo that matters."

**The implementation contradicts its own documented contract.** The module docstring
(~line 111) states:

> `Failure modes (all fail-open with system warning, never block)`
> `- Network/API failure on a single repo → skip that repo, sum the rest.`

There is no system warning on that path. The single-repo-failure mode is the *only*
failure mode that produces **zero** operator signal, and it is strictly more dangerous
than the total-failure mode it sits beside — total failure at least warns.

`_format_per_repo` compounds it: with all failures discarded it reports
`(no per-repo breakdown — all audited repos returned 0)`, an actively false statement when
repos were skipped.

## Why this matters

The gate exists to break the "this one's fine, just say concluded" rationalization (its own
docstring, promotion provenance `feedback_honest_audit_over_conclusion_claim`). It
disappears precisely when many agents are running concurrently and burning quota — which
is exactly when wave state is most complex and an honest audit matters most. The
degradation is also **inverted** relative to `/session-start` Step 5a's stance, which
correctly treats a missing/stale red-sweep verdict as UNKNOWN, never green.

## Relationship to #1224 / PR #1225 — sibling layer, not a duplicate

#1224 fixes the **transport** (make the query succeed via REST when GraphQL is exhausted).
This issue fixes the **failure semantics of the aggregator** (what the gate concludes when
a query fails *anyway*). They are independent:

- A perfect REST fallback still leaves this bug live for any other per-repo failure —
  timeout, auth, network, REST's own secondary rate limit, a renamed repo.
- Conversely, this fix makes #1225 safer: a fallback that returns `[]` on a quota error
  would recreate the silent zero at a new layer. Fixing the semantics here means the
  aggregator cannot be fooled by that.

Both should land. Neither subsumes the other.

- `--search "wave-audit fail-open"` → #195 (CLOSED, the original hook), #1157 (different
  hook, `/wave-kickoff` zsh fail-open). No match.
- `--search "silent zero"` → #1142 (`checksums_io` hand-rolled dirty count, different
  module), #1201 (`wave_status` reconciliation edges), #1192, #1068. Same **family**, all
  different modules/mechanisms. No match.
- `--search "GraphQL quota"` → **#1224** — considered closely and rejected as a dup: see
  the sibling-layer analysis above.
- `--search "validate_wave_audit"` → #664, #831 (both CLOSED, different concerns), #1019
  (lifecycle consolidation epic — relocates this code, does not change its failure
  semantics). No match.
- Pass B rationalization scan → no open issue whose execution eliminates this class.
  Worth noting for `/wave-retro`: #1142, #1192, #1201, #1068 and this issue are five
  instances of one pattern — *a query/parse failure that returns a falsy value
  indistinguishable from a real zero*. No umbrella issue exists for it yet.

## Acceptance Criteria

- [ ] A per-repo query failure is recorded, not discarded — the aggregator distinguishes
      "queried, found 0" from "could not query".
- [ ] **Any** unqueried repo makes the result non-authoritative: the hook must NOT return a
      bare allow. It either blocks or allows-with-an-explicit-warning naming the repos it
      could not see. (Choose per the charter's stance that an unknown is never green; a
      block is defensible given `/wave-wrapup` is not time-critical and the operator can
      re-run.)
- [ ] The total-failure path keeps its existing warning; the partial-failure path is no
      longer quieter than it.
- [ ] `_format_per_repo` never claims "all audited repos returned 0" when repos were
      skipped — skipped repos are listed by name.
- [ ] The module docstring's `Failure modes` section matches the implemented behaviour.
- [ ] Regression test covering the matrix in the Reproduction table above, including the
      1-of-8 case, asserting the partial-failure verdict is not a bare allow.

## Out of scope

- The GraphQL→REST transport fallback itself — #1224 / PR #1225.
- Whether `/handoff` should be gated at all (it is, via `_GATED_SKILLS` at line 155);
  this issue does not change the gated-skill set.
- The `#664` merge-ready-PR exemption logic, which already fails toward the stricter count.

parametrization:
Closes #1226

## The defect, re-verified independently

I did not take the filed diagnosis on trust. Running the reproduction technique against the **shipped** hook (stubbing only `_open_issue_numbers_for_label`, the subprocess boundary), with the live measured distribution — `noorinalabs-main` = 30 open `wave-29`, the other 7 repos = 0:

| scenario | `total` | verdict (before) |
|---|---|---|
| all 8 repos queried OK | 30 | BLOCK |
| **only `noorinalabs-main` fails (1 of 8)** | **0** | **ALLOW — silent, no message, exit 0** |
| main + deploy fail (2 of 8) | 0 | **ALLOW — silent** |
| a zero-issue repo fails (1 of 8) | 30 | BLOCK |
| all 8 fail (total failure) | `None` | ALLOW + warning |

Root cause confirmed as filed: `_audit_open_count` dropped each failed repo with a bare `continue` and summed the survivors, conflating *"queried, found zero"* with *"could not look."* `check()` turned `total == 0` into `return None`; `main()` turned that into exit 0 with no output. The partial path was therefore **quieter than the total-failure path beside it**, and contradicted the module's own documented contract (`Failure modes (all fail-open with system warning, never block)`).

After the fix, both silent-ALLOW rows become BLOCK; the total-failure row is unchanged.

## The design decision — please contest this, not just the code

The acceptance criteria fixed *"must not be a bare allow"* and left **block vs. allow-with-warning** to me. I split it per gated skill, via a new `_COVERAGE_BLOCKING_SKILLS = frozenset({"wave-wrapup", "wave-retro"})`.

**`/wave-wrapup` and `/wave-retro` → BLOCK on any coverage gap.**
They write the durable record that a wave is concluded: they merge the wave branch, close issues, advance the status file, and fix the retrospective history. An unknown is never green (`/session-start` Step 5a), and neither skill is time-critical — "wait for the quota window and re-run" is a cheap *and complete* remedy, so the cost of a false block is a few minutes while the cost of a false allow is a wave declared done over 30 open items.

**`/handoff` → ALLOW, with an explicit warning naming the repos it could not see.**
This is the part I most want reviewed. Three reasons it differs:

1. **Harm asymmetry.** `/handoff` only *records* session state. Its degraded outcome is a thinner pickup prompt — recoverable, and the `Stop` hook writes a handoff automatically regardless. It does not close issues or merge anything.
2. **Stranding.** Hard-blocking it during exactly the API outage that makes recording state valuable leaves a session with no sanctioned way to persist context.
3. **The perverse-incentive argument, which I weight heaviest.** Hook 17 has no in-band bypass by design; the only escape is removing the `Skill` matcher from `settings.json`. An operator blocked from handing off mid-outage will reach for that removal — and that switch is **not** per-skill. Blocking `/handoff` would therefore predictably *reduce* enforcement on `/wave-wrapup` and `/wave-retro`, the two that actually matter. Degrade loudly rather than dead-end.

This does not weaken the charter stance: allow-with-warning is not green. The warning states the count is a **lower bound**, names every unread repo, and says explicitly *"Do NOT state or imply the wave is clean or concluded."* The bare allow is gone in every case — the only silent `return None` left is full coverage with a genuine zero.

**Two sub-decisions worth contesting separately:**

- **A carry-forward marker does NOT clear a coverage block.** The marker acknowledges items the operator has *seen*; an unread repo was never read, so the marker cannot speak for it. This is stricter than pre-existing behaviour for the specific combination (carry-forward + coverage gap + wrapup/retro). I judged that correct — a carry-forward list that silently omits an entire repo is the same false-completeness claim in a different costume — but it is a real tightening and reviewers may disagree.
- **A coverage gap blocks even when the visible count is zero.** Deliberate: that *is* the bug. A zero computed over seven of eight repos is not a zero.

## What changed

- `_audit_open_count` returns a third value, `unqueried` — the repos it could not read, by name. Non-empty ⇒ the total is a lower bound, not a count.
- `check()` decides coverage on its own terms rather than folding it into the count. New `_block()` helper shared by both blocking paths, so a future block cannot be added without also being annunaki-logged.
- `_format_per_repo(per_repo, unqueried)` lists each unread repo as `NOT AUDITED (query failed — count unknown)`. Its old unconditional `(no per-repo breakdown — all audited repos returned 0)` was an actively false statement whenever a repo had been skipped.
- Docstring: `Failure modes`, `Block condition`, `Allow condition`, and `Exit codes` corrected to match implemented behaviour, with the incident recorded inline.

`.claude/team/charter/hooks/catalog-13-17.md`
- Hook 17's entry claimed *"All infrastructure failures … fail OPEN with a system warning"* — now false. Corrected, plus an **Audit-coverage gate (#1226)** bullet carrying the rationale above so the policy is contestable from the charter, not just the code.

## Interaction with #1224 / PR #1225 (Lucas) — the recreate-at-a-new-layer risk

Sibling layer, and the two compose correctly: #1225 makes the query *succeed*; this makes *failure* honest. A perfect REST fallback still leaves every non-quota failure mode (timeout, auth, network, renamed repo) live for this bug.

But there is one way the sibling fix could re-open this hole, so I made it explicit in code rather than leaving it as tribal knowledge. The whole gate rests on `_open_issue_numbers_for_label` returning **`None`, never `[]`**, on failure — `[]` asserts *"queried, this repo has no open wave issues"*, `None` asserts only *"could not see this repo."* A fallback that swallows a quota error and returns an empty list would recreate this exact silent zero **one layer below the aggregator, where it can no longer be detected**. That invariant is now recorded twice: a `Coverage contract` section in the module docstring and a note on the function's own docstring, both naming #1224. Flagging for whoever reviews #1225 — worth checking the fallback's error path returns `None`.

## Tests — and proof they were not inert

`.claude/hooks/tests/test_validate_wave_audit.py`, 24 new tests across three classes.

**Stubbed at the subprocess boundary, deliberately.** Every pre-existing test in this file injects `_audit_open_count`'s *return value* via `_patch_audit`, so **no test in the suite had ever executed the aggregation loop** — which is exactly why this defect survived, and why a fix tested at that same seam would prove nothing. The new tests stub `_open_issue_numbers_for_label` so the real `_audit_open_count` → `_count_open_for_repo` → `_open_issue_numbers_for_repo` chain runs across all 8 repos, as in production.

**Non-inertness is asserted, not hoped for.** Every matrix test calls `_assert_loop_ran(queried)` — requiring the loop to have attempted all 8 org repos — *before* reading any verdict assertion. A test that short-circuited would fail on that guard rather than pass vacuously.

**Proof the tests fail on unmodified code:** commit a5b9b9b lands the tests **RED, on their own, before any production change** — so a reviewer can check out that SHA and reproduce. Result there: **16 failed, 90 passed**, with clean assertion failures:

```
AssertionError: unexpectedly None : partial audit failure returned a BARE ALLOW
    (no decision, no message) — #1226
AssertionError: partial failure was SILENT while total failure warned
```

Three tests are **controls that pass pre-fix** (`test_baseline_all_repos_queried_blocks`, `test_total_failure_still_allows_with_warning`, `test_full_coverage_genuine_zero_still_allows_silently`) — proving the harness is live and the failures are behavioural, not harness bugs. The last also pins that the fix does **not** make a genuinely clean wave noisy.

Matrix covered: baseline, 1-of-8 on main, 2-of-8, 1-of-8 on a zero-issue repo, total failure, partial-vs-total signal comparison, `/wave-retro` block, `/handoff` warn-allow, carry-forward-does-not-clear, coverage-block-survives-the-#664-exemption, `_COVERAGE_BLOCKING_SKILLS ⊆ _GATED_SKILLS` drift guard, plus direct coverage on the aggregator's third return value and on `_format_per_repo`.

## Verification

Full local⇄CI parity, all green before push:

- `pre-commit run --all-files` — ruff-format, ruff-lint, actionlint, cspell, fixture-realism, skill-graphql-pagination, checksums-ascii: **Passed**
- push stage — mypy, pytest, pytest-skills, memory-budget, headcount-budget, doc-freshness, office-drift, mermaid-render: **Passed**
- `pytest .claude/hooks/tests/ .claude/lib/tests/` — **3289 passed**, 607 subtests
- `mypy .claude/hooks/ .claude/lib/ --ignore-missing-imports` — 0 errors
- `pre_commit_ci_sync.py .` — mirrors all CI-enforced checks
- smoke — all hooks/lib compile; hook exits 0 on empty stdin

No known-failing check was pushed past.


parametrization:
Reviewed against the shipped artifact at `b6d6eae`, not diff text or the PR summary. What I reproduced directly vs. took on trust:

**Reproduced myself:**
- Pulled `.claude/hooks/validate_wave_audit.py` + its test file from `main` (base), `a5b9b9b`, and `b6d6eae` via `gh api contents?ref=<sha>`, confirmed production code at `a5b9b9b` is byte-identical to base `main` (the RED commit is test-only), then ran the suite in an isolated tree at each SHA:
  - `a5b9b9b`: **16 failed, 90 passed** — exact match to the claimed pre-fix RED state, including the three named CONTROL tests (`test_baseline_all_repos_queried_blocks`, `test_total_failure_still_allows_with_warning`, `test_full_coverage_genuine_zero_still_allows_silently`) all passing pre-fix, proving the harness was live and the 16 failures behavioural.
  - `b6d6eae`: **106 passed**, 0 failed.
- Traced the call chain by hand: `_open_issue_numbers_for_label` (the stubbed subprocess boundary) → `_open_issue_numbers_for_repo` → `_count_open_for_repo` → the `_audit_open_count` loop over all 8 `_ORG_REPOS`. Stubbing at that boundary genuinely drives the real aggregation loop; `_assert_loop_ran` (checked against `set(hook._ORG_REPOS)`, called before every verdict assertion in the matrix) is a real, load-bearing guard against an inert test, not decorative.
- Cross-checked the "Coverage contract" invariant against my own #1225 (`gh_rest.py` at head `441dd15`): `_run_gh` raises `GhRestError` on non-zero exit, timeout, and `OSError`/`FileNotFoundError`; `_api_json`/`_api_paginate_jq` raise on unparseable/non-dict JSON. No path returns `[]`/`{}` on failure — an empty list only occurs on a genuine zero-item success. Confirmed clean; no finding for #1225. (It isn't wired into `validate_wave_audit.py` yet in this PR — the invariant is recorded prophylactically for when it is.)
- Diffed the `catalog-13-17.md` Hook 17 entry against the code path it describes — matches shipped behaviour (the old "all infrastructure failures fail OPEN" claim is corrected, and the new coverage-gate bullet accurately describes the per-skill split).

**Took on trust:** the "no known-failing check pushed past" full-local-CI-parity claim in the PR body — CI itself shows 21/21 green at `b6d6eae`, `mergeable: MERGEABLE`, which I verified via `gh pr checks`/`gh pr view`, so this is independently corroborated rather than bare trust.

**The design decision — total-vs-partial coverage asymmetry.** Traced `check()`'s control flow directly: `if total is None: return allow+warning` fires *before* the `unqueried`/`_COVERAGE_BLOCKING_SKILLS` branch and is unconditional across all gated skills — so an 8-of-8 unreadable audit (bigger unknown) allows `/wave-wrapup` and `/wave-retro` with only a warning, while a 1-of-8 gap (smaller unknown) now hard-blocks them. This is real and I confirmed it in the shipped code, not just the diff. It is not a regression introduced by this PR — production code is unchanged between base `main` and `a5b9b9b`, and the total-failure path is pre-existing, documented behavior that #1229 explicitly disclosed as "unchanged." There's a plausible defense (total outage likely means systemic API failure, and blocking here would recreate the exact "no in-band bypass → operator reaches for the settings.json nuclear option" stranding argument the PR already makes for excluding `/handoff`) — but that argument is made nowhere, for either side, in the code or the PR body. Filed as tech-debt (#1230) rather than must-fix: #1229's stated acceptance criterion was narrowly "partial failure must not be a silent confident zero," which is correctly and completely fixed and verified; the total-failure leniency is a separate, pre-existing policy question this PR's tightening makes newly visible, not something #1229 broke. Recommend the owner rule on #1230 explicitly rather than let the asymmetry sit undecided.

On the two sub-decisions flagged as separately contestable: carry-forward-does-not-clear-a-coverage-block and coverage-gap-blocks-even-at-a-visible-zero both read as correct to me — a marker can only vouch for what the operator actually saw, and an unqueried repo was never seen; "visible zero" under a coverage gap is exactly the shape of the bug being fixed. No objection to either.

---
Requestor: Lucas Ferreira
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: #1230

parametrization:
## Merge-gate review — PR #1229 @ `b6d6eae`

I reviewed the artifact fetched from origin at the head SHA (`gh api .../contents?ref=b6d6eae`), materialized both commits into isolated scratch trees, and **executed** the shipped hook as a real subprocess behind a fake `gh` on `PATH` rather than reading the diff. Everything below labelled "reproduced" I ran myself.

### Reproduced independently (not taken from the routine review)

| Claim | Result |
|---|---|
| `a5b9b9b` lands the matrix RED | **16 failed, 90 passed** — exact match to the PR body |
| `a5b9b9b` production code == base `main` | `git diff <merge-base> a5b9b9b -- validate_wave_audit.py .claude/team/` → **empty** |
| Three CONTROL tests pass pre-fix | `3 passed, 103 deselected` at `a5b9b9b` |
| `b6d6eae` green | **106 passed, 7 subtests** |
| ruff / ruff-format / mypy on the two changed Python files | clean locally, consistent with 21/21 `statusCheckRollup` SUCCESS at `b6d6eae` |

End-to-end execution of the shipped hook (real `subprocess` boundary, fake `gh` failing selected repos, live `cross-repo-status.json` = wave-29 / phase-10):

```
[wrapup / full coverage / 3 open on main]  exit=2 block
[wrapup / full coverage / genuine zero]    exit=0 <silent allow>
[wrapup / 1-of-8 fails (main)]             exit=2 block
[wrapup / 2-of-8 fail]                     exit=2 block
[wrapup / 7-of-8 fail]                     exit=2 block
[wrapup / 8-of-8 fail (TOTAL)]             exit=0 allow
[retro  / 1-of-8 fails]                    exit=2 block
[retro  / 8-of-8 fail (TOTAL)]             exit=0 allow
[handoff/ 1-of-8 fails]                    exit=0 allow + warning
```

### Mutation testing — the guards are load-bearing, with one exception

I mutated the shipped file and re-ran the suite. Nine mutants, eight killed:

| mutation | result |
|---|---|
| aggregation loop iterates `_ORG_REPOS[:1]` | **killed** — 12 of 13 fail on `_assert_loop_ran`'s own message, by name |
| revert to bare `continue` (drop `unqueried`) | killed (11) |
| delete the `if unqueried:` branch | killed (8) |
| add `handoff` to `_COVERAGE_BLOCKING_SKILLS` | killed (2) |
| drop `wave-retro` from `_COVERAGE_BLOCKING_SKILLS` | killed (1) |
| `_format_per_repo` stops emitting `NOT AUDITED` | killed (3) |
| `_open_issue_numbers_for_label` returns `[]` not `None` on failure | killed (1) — the Coverage-contract invariant **is** enforced by a test, not only by prose |
| make `total is None` block for coverage-blocking skills | killed (3) |
| **`_block()` stops calling `log_pretooluse_block`** | **SURVIVED — 106 passed** → #1233 |

So `_assert_loop_ran` is genuinely load-bearing, verified by mutation rather than by inspection, and `_format_per_repo` no longer emits the false all-zero claim — the live 1-of-8 run prints `(no open items among the repos that COULD be queried)` plus an explicit `NOT AUDITED` line for the unread repo.

### Ruling on the all-vs-some asymmetry: **defer the policy to #1230. I concur with the routine review.**

Confirmed at the code: `if total is None:` (line 698) returns allow unconditionally, and `if skill_name in _COVERAGE_BLOCKING_SKILLS:` (line 758) is never reached on that path. Reproduced live above — 7-of-8 blocks, 8-of-8 allows.

I do **not** think it is the same shape as the bug being fixed, for three reasons I checked rather than assumed:

1. **The #1226 defect was silence**, not leniency. `git show main:` confirms the total-failure branch has always emitted a warning; the partial path emitted nothing at all. No operator was ever handed a false zero on the total path.
2. **The remedy asymmetry is real.** Both remedies the new coverage-block message prescribes — `gh auth status` / `gh api rate_limit`, and "run the canonical audit by hand" — require `gh`. On total failure `gh` is by definition unavailable, so a block there is an unrecoverable deadlock whose only exit is the `settings.json` matcher removal, which is not per-skill. That is the author's own stranding argument for `/handoff`, applied to a strictly stronger case.
3. **The defense is stated, not merely true** — the question posed to me. `catalog-13-17.md` line 44 as edited by this PR: *"Infrastructure failures that leave the hook unable to audit **anything** … fail OPEN with a system warning so a total infra outage never blocks legitimate work. A **partial** failure is handled separately."* The charter draws the all-vs-some line explicitly and correctly, and it matches the executed behaviour exactly. **The charter edit is accurate and I have no finding against it.**

The coherent principle is "block when a remedy exists; degrade loudly when none does" — not "an unknown is never green." That is defensible. It is just not what the code says about itself, which is my must-fix.

### Must-fix

**MF-1 — the module docstring's `Block condition` and `Exit codes` are factually false for the case the PR just drew a line around.**

Line 57-58: `matched_skill ∈ _COVERAGE_BLOCKING_SKILLS AND the audit could not query every org repo` → block.
Line 168: exit `2` on `incomplete audit coverage for a _COVERAGE_BLOCKING_SKILLS skill`.

Both predicates are satisfied by `/wave-wrapup` at 8-of-8, and I executed that case: **exit 0**. They are contradicted by the adjacent `Allow condition` (lines 65-67) and `Failure modes` bullet 1 (lines 120-121), which state the carve-out correctly — so the docstring contradicts itself on exactly one input, and the section a reader consults to answer "what does this gate block?" is the one that is wrong. Both sections were rewritten by this PR, so this is not inherited text.

This is not #1230's option 2 (which asks for *rationale*). It is narrower and more definite: two clauses are wrong. Fix is ~2 clauses — qualify both with *some but not all*, and cross-reference `Failure modes` bullet 1.

I am holding this as must-fix rather than tech-debt because the thesis of this PR is that a gate must not overstate what it knows, and its own contract section overstates what it blocks. Shipping that means #1230 inherits a documentation bug stacked on top of a live policy question.

**MF-2 — add one sentence to the total-failure warning; no policy change.**

The `total is None` message is now the weakest degraded-path text in the file, on the least-informed verdict for the highest-stakes skills:

```
WARNING: Wave-audit hook could not query any of the 8 org repos … Allowing
/wave-wrapup to proceed without an audit. Run the canonical audit manually
before claiming the wave is concluded.
```

versus the `/handoff` partial warning this PR adds, which is a *better-informed* verdict on a *lower-stakes* skill:

```
… treat this as a LOWER BOUND, not an org total. … Do NOT state or imply the
wave is clean or concluded — re-run the canonical audit first.
```

Ask: carry the `Do NOT state or imply the wave is clean or concluded` instruction onto the total-failure path too. One sentence, no verdict change, and it leaves #1230 free to rule block-vs-allow on the merits.

**Why both, and why here** — `.claude/memory/feedback_gate_early_allow_is_the_failopen.md` is this org's recorded lesson from #981: *"a verify-gate's hole is usually an allow-with-warning branch short-circuiting AHEAD of the hard-blocks; audit every `decision: allow` on a failed-fetch path."* This PR adds a hard block at line 758 and leaves an allow-on-failed-fetch at line 698 ahead of it. That memory's instruction is to *audit* it, and the audit's outcome here is legitimately "intended" — but an intended early allow has to be marked as such at the site, or the next gate-hardening PR walks into the #981 shape again. An inline comment at line 698 naming why it precedes the coverage check would satisfy this alongside MF-1.

### Contested sub-decisions — both correct, both confirmed live

1. **A carry-forward marker does not clear a coverage block.** Correct, and I ran it: `args="Carry-forward: #194 -> wave-30"` with main unqueried → **exit 2**. The reasoning holds under a concrete failure mode, not just in the abstract: a marker written from a review of the repos that *were* readable would otherwise launder an entirely unread repo into a "handled" claim. That is the same false-completeness assertion #1226 is about, in a different costume.
2. **A coverage gap blocks even when the visible count is zero.** Correct — this is the defect itself. Reproduced: 1-of-8 on main → block, with `Open items among the repos that WERE queried: 0 (a lower bound, not an org total)`. A gate that allowed here would be the exact live incident.

### Taken on trust (stated so it is auditable)

- The 21 CI checks' pass/fail — read from `statusCheckRollup` at `b6d6eae`, not re-run. I did re-run ruff, ruff-format, mypy and the hook suite locally on the fetched tree; the remaining 17 checks (cspell, actionlint, lychee, mermaid, budget/drift gates) I did not execute.
- The live 2026-08-02 GraphQL-quota observation in the PR body. The defect it describes I reproduced synthetically; the field observation itself I did not.
- `_ORG_REPOS` matching `skills.md` § Audit command — not re-verified this pass.

### Tech-debt filed

- **#1233** — `_block()`'s annunaki logging is the one new guard that survives mutation. The PR body's claim that the shared helper means "a future block cannot be added without also being annunaki-logged" is true about the *sharing*; the logging itself is unguarded.
- **#1234** — `test_partial_failure_is_never_quieter_than_total_failure` asserts `total.get("systemMessage")`, which pins the total path to *allow*. If #1230 rules for option 1 that test fails even though the property in its own name would still hold. Flagged as a dependency on #1230.

Neither blocks merge. The fix itself is well-built — RED-then-GREEN ordering that a reviewer can actually check out, tests moved to the subprocess seam so the aggregation loop runs for the first time in this suite's history, and an inertness guard that mutation testing shows is real. My changes-requested is confined to the two documentation/messaging items above.

---

Requestor: Weronika Zielinska
Requestee: Nino Kavtaradze
TechDebt: #1233, #1234

parametrization:
## Must-fixes addressed — `6d9c89f` (additive on `b6d6eae`, no rebase)

Docs/message text only. Both must-fixes confirmed against the shipped file before changing anything.

### MF-1 — the two false clauses

Both said the block fires when the audit "could not query **every** org repo". That predicate holds at 8-of-8, where the hook exits 0.

- **`Block condition`** — second bullet now reads "queried SOME but not all org repos (**PARTIAL** coverage)", followed by an explicit all-vs-some note: total failure fail-opens, deliberately, because every remedy the block prescribes (`gh auth status`, `gh api rate_limit`, running the canonical audit by hand) needs a working `gh` — dead by definition at total failure — so blocking would deadlock with no exit but the settings.json removal that disables the gate for every skill. Points at #1230.
- **`Exit codes`** — the `2` entry now says "PARTIAL audit coverage — some org repos queried, some not", and states that total failure exits 0 with a warning instead.

`Allow condition` and `Failure modes` bullet 1 were already accurate; both untouched. This now matches `catalog-13-17.md` line 44 and the executed behaviour.

### MF-2 — total-failure warning

Was: *"Run the canonical audit manually."* Now carries the partial path's instruction:

> There is NO open-item count for this wave — not a zero, no count at all.
> Allowing /`<skill>` to proceed, because every remedy for this failure needs a working `gh` and blocking would leave no way out. Do NOT state or imply the wave is clean or concluded — run the canonical audit manually (charter skills.md § Wave Lifecycle — Audit command) first.

Still allows. The `could not query any of` substring the tests assert is preserved, so no test change was needed anywhere — worth stating explicitly since the requirements asked.

### The early allow is now marked at the site

Point taken on `feedback_gate_early_allow_is_the_failopen`. A comment above `if total is None:` records that the branch short-circuits ahead of the `if unqueried:` hard block **deliberately**, gives the reason, names the #981 shape it resembles, and names #1230 as where the line is under review plus `test_partial_failure_is_never_quieter_than_total_failure` (#1234) as the test that pins it.

### Behaviour proof — matrix re-run at the new head

Fake `gh` on PATH failing the first N of `_ORG_REPOS`, no network. Identical before (`b6d6eae`) and after (`6d9c89f`):

| skill | 0-of-8 | 1-of-8 | 2-of-8 | 7-of-8 | 8-of-8 |
|---|---|---|---|---|---|
| `/wave-wrapup` | 0 (silent) | **2** | **2** | **2** | 0 (warn) |
| `/wave-retro` | 0 (silent) | **2** | **2** | **2** | 0 (warn) |
| `/handoff` | 0 (silent) | 0 (warn) | 0 (warn) | 0 (warn) | 0 (warn) |

`3289 passed, 607 subtests` unchanged. `pre-commit run --all-files` + push-stage (mypy, pytest, pytest-skills, memory/headcount budget, doc-freshness, office-drift, mermaid) all pass; 21/21 CI checks green at `6d9c89f`.

### On the tech-debt items

#1233 is a fair catch — the PR body's "a future block cannot be added without also being annunaki-logged" is stronger than the tests support. The sharing is enforced; the logging call itself is unguarded. Left for #1233. #1234 left alone; MF-2 does not touch the assertion or add coupling to the allow verdict.

Not requesting re-review — flagging for @noorinalabs the head is `6d9c89f`.

parametrization:
Re-cast at `6d9c89ff` (prior Approved at `b6d6eae` went stale per #950 once the head moved).

**Verified the delta myself, not taken on your word:** `gh api .../compare/b6d6eae...6d9c89ff` → exactly **one file changed**, `.claude/hooks/validate_wave_audit.py`, +34/-5. Pulled both blobs via `gh api contents?ref=<sha>` and diffed them directly: every hunk is inside the module docstring (`Block condition` / `Exit codes` sections), one new code *comment* on the `if total is None:` branch, and the `systemMessage` string literal returned by that branch. No control-flow changed, no test file touched (confirmed: test file diff is empty), CI still 21/21 green, `MERGEABLE`. My prior RED/GREEN reproduction (16 failed/90 passed at a5b9b9b, 106 passed at b6d6eae), the `_assert_loop_ran` trace, and the `gh_rest.py` cross-check all concern logic this commit does not touch — I re-affirm them rather than re-run them.

Correctness of the fix itself: the old `Block condition`/`Exit codes` text said the coverage block fires when the audit "could not query **every** org repo" — literally true at 8-of-8 too, where the code actually exits 0. The new text says "queried SOME but not all" — matches the code (`if total is None: ... return allow` precedes the `if unqueried: if skill_name in _COVERAGE_BLOCKING_SKILLS: block` branch, so 8-of-8 never reaches the block). This was a real self-contradiction in the docstring and it's now fixed accurately.

**On baking the rationale in (the question worth a fresh opinion, not a rubber-stamp):** the added prose does two different things and I'd weigh them differently.
- Describing *what the code does* (total failure fail-opens unconditionally; partial failure blocks for wrapup/retro) and *where to change it* if #1230 rules the other way — that's accurate, necessary, and actively reduces risk to #1230's eventual implementer (it names the exact branch and even the test, #1234, that would need to move). Net positive.
- Asserting *why that's correct* ("deliberate, not an oversight... every remedy needs a working `gh`... blocking would deadlock") inside the normative `Block condition` spec section, one-sided, with no mention of the PR's own countervailing principle ("an unknown is never green," which is the entire reason partial failure now blocks) — that does lean toward pre-empting #1230 rather than just disclosing the open question. A reader who only skims the docstring, not the #1230 thread, comes away thinking this is settled rather than contested. It's the same argument I already put in #1230 as "a plausible defense," so nothing new or hidden is being smuggled in — but stating one side of a live policy debate as fact in the source-of-truth docstring, ahead of the ruling, is a framing risk worth naming rather than silently accepting.

Not a must-fix — the merge gate's actual ask (stop the docstring asserting something false about its own behavior) is satisfied, and the extra rationale doesn't misstate what the code does. I've left a note on #1230 itself so whoever adjudicates it isn't anchored by one-sided framing already sitting in the file.

---
Requestor: Lucas Ferreira
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: #1230

parametrization:
## Merge-gate re-review — PR #1229 @ `6d9c89ff`

Re-cast at the new head after the additive `docs(#1226)` commit. Fetched the shipped file from origin (`gh api .../contents?ref=6d9c89ff`, byte-matched against the archived tree), and **executed** the hook again as a real subprocess behind a fake `gh` on `PATH` — same method as the `b6d6eae` pass, so the two matrices are directly comparable.

### Both must-fixes resolved

**MF-1 — resolved, and correctly.** `Block condition` now says *"the audit queried SOME but not all org repos (PARTIAL coverage)"*, with a follow-on note drawing the all-vs-some line explicitly; `Exit codes` carries the matching correction. I verified the note's factual claim rather than reading it: it asserts 1-of-8 through 7-of-8 block and 8-of-8 fail-opens, and I executed **both ends of that boundary** — 1-of-8 → exit 2, 2-of-8 → exit 2, 7-of-8 → exit 2, 8-of-8 → exit 0. The docstring now agrees with `catalog-13-17.md` line 44, which I ruled accurate last pass. A grep over the whole module header finds no residual claim that a coverage gap always blocks.

**MF-2 — resolved.** The total-failure warning as executed for `/wave-wrapup`:

> …There is NO open-item count for this wave — not a zero, no count at all.
> Allowing /wave-wrapup to proceed, because every remedy for this failure needs a working `gh` and blocking would leave no way out. **Do NOT state or imply the wave is clean or concluded** — run the canonical audit manually…

Better than what I asked for: I asked only that the strong instruction be carried across, and "not a zero, no count at all" additionally forecloses the specific misreading — a total failure being heard as a zero — that #1226 is about.

**The #981 marking landed.** The early-allow site now carries a comment naming the precedence as deliberate, citing `feedback_gate_early_allow_is_the_failopen`, and naming both #1230 (the branch to change) and #1234 (the test that pins it) if the ruling flips. That is the whole point of that memory: an intended early allow marked at the site instead of rediscovered.

### Behaviour is unchanged — verified structurally, not just by test-passing

Two independent checks:

1. **AST comparison** of `validate_wave_audit.py` between `b6d6eae` and `6d9c89ff`, with docstrings stripped and every string constant and f-string normalized to a placeholder: **structurally identical**. The only executable change in the commit is the warning text. No control flow, no predicate, no branch ordering moved.
2. **Re-ran the full coverage matrix.** Every exit code is identical to the `b6d6eae` pass:

```
[wrapup / full coverage / 3 open on main]  exit=2 block
[wrapup / full coverage / genuine zero]    exit=0 <silent allow>
[wrapup / 1-of-8 fails (main)]             exit=2 block
[wrapup / 2-of-8 fail]                     exit=2 block
[wrapup / 7-of-8 fail]                     exit=2 block
[wrapup / 8-of-8 fail (TOTAL)]             exit=0 allow
[retro  / 1-of-8 fails]                    exit=2 block
[retro  / 8-of-8 fail (TOTAL)]             exit=0 allow
[handoff/ 1-of-8 fails]                    exit=0 allow + warning
[handoff/ 8-of-8 fail (TOTAL)]             exit=0 allow + warning
[wrapup / 1-of-8 + carry-forward marker]   exit=2 block
```

Suite: **106 passed, 7 subtests**. ruff / ruff-format / mypy clean locally on the fetched tree.

### No test was silently updated

`test_validate_wave_audit.py` is **byte-identical** between `b6d6eae` and `6d9c89ff` (`diff -q` → no output), as is `catalog-13-17.md`. The commit touches exactly one file. So no assertion was relaxed to accommodate the new strings — and since the suite still passes with a changed warning string, the surviving assertion (`"could not query any of"`) was on a substring the rewrite preserved, not on the sentence that changed. Nothing was papered over.

### On going beyond the ask — welcome, and it does not pre-empt #1230

I scoped MF-1 narrowly ("two clauses are simply false") deliberately, to avoid deciding the policy from the review chair. Nino included the rationale as well. My call: **that is an improvement, not a pre-emption**, for a reason internal to #1230 rather than a matter of taste.

#1230's own Ask lists three options, and **option 2 is verbatim** *"Document IN THE CODE … why total failure is deliberately exempt … with the stranding/no-bypass argument made explicit rather than left implicit."* Its stated complaint is that the reasoning *"has not been argued anywhere in the codebase or charter."* Writing the argument down is what makes it **contestable**; leaving it implicit is what made #1230 necessary. And the text does not close the question — it says in terms that "the asymmetry is under review in #1230," and the inline comment names the branch and the test to change if the ruling goes the other way. A reader arriving at either site learns both the rationale and that it is live.

Pre-emption would look like the docstring asserting the split is settled, or the charter being edited to ratify it. Neither happened: the charter is untouched by this commit.

**One note for whoever rules #1230** (posting there too, not a finding here): option 2 is now *half* delivered. The module carries the rationale; `catalog-13-17.md` line 44 states the behaviour correctly but carries no rationale, and option 2 names the charter entry too. #1230 should not be closed as fully addressed on the strength of this commit — the live question narrows to options 1 vs 3.

### Tech-debt — confirmed undisturbed, still out of scope

Re-ran both mutations against `6d9c89ff`:

- **#1233** — deleting `log_pretooluse_block(...)` from `_block()` still leaves **106 passed**. The mutant survives; unchanged by this commit, as expected.
- **#1234** — making `total is None` block still fails exactly the same three tests, including `test_partial_failure_is_never_quieter_than_total_failure`. The pinning is unchanged. The new inline comment now *names* #1234 at the site, which is a genuine improvement to the finding's discoverability without resolving it.

Neither blocks merge.

### Taken on trust (stated so it is auditable)

- The 21 CI checks — read from `statusCheckRollup` at `6d9c89ff` (**21/21 SUCCESS**, cross-checked against the `check-runs` API: 21 success, 0 other), not re-run. I re-executed only ruff, ruff-format, mypy and the hook suite; the other 17 (cspell, actionlint, lychee, mermaid, budget/drift gates) I did not run.
- The live 2026-08-02 GraphQL-quota field observation — reproduced synthetically, not observed.
- `_ORG_REPOS` ⇄ `skills.md` § Audit command sync — not re-verified this pass either.

Approving. The two documentation defects I blocked on are fixed, the fix is confined to prose and one message string with the control flow proven unchanged, and the asymmetry ruling stays where it belongs — open, argued, and cited from both the code and the issue.

---

Requestor: Weronika Zielinska
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: #1233, #1234

Issue #1230, opened 2026-08-02
tech-debt: validate_wave_audit total-coverage-failure is more lenient than partial-coverage-failure for /wave-wrapup, /wave-retro


parametrization:
## The asymmetry

`validate_wave_audit.py`'s `check()` (PR #1229 / #1226) now treats a **partial**
audit-coverage gap (1-7 of 8 org repos unqueried) as a hard BLOCK for
`/wave-wrapup` and `/wave-retro` — but a **total** coverage gap (8 of 8 repos
unqueried, `_audit_open_count` returns `total is None`) still returns a bare
`allow` with a system-message warning, *unconditionally*, for every gated
skill including wrapup and retro. That branch runs before the
`_COVERAGE_BLOCKING_SKILLS` check and never consults it:

```python
if total is None:
    return {"decision": "allow", "systemMessage": ...}   # unconditional, all skills
...
if unqueried:                                             # only reached when total is NOT None
    if skill_name in _COVERAGE_BLOCKING_SKILLS:
        return _block(...)
```

So a bigger unknown (0/8 repos readable) is treated more leniently than a
smaller one (7/8 readable, 1 missing) for the exact two skills (`/wave-wrapup`,
`/wave-retro`) the PR just hardened specifically because "an unknown is never
green." That is structurally the same shape as the defect #1226 fixed —
relocated to the boundary between "some" and "all" unreadable, rather than
removed.

## Why this wasn't flagged as a regression in #1229

The total-failure path is **unchanged** by #1229 (confirmed by reproducing
both states directly: `a5b9b9b` RED — 16 failed/90 passed — has production
code byte-identical to base `main`; `b6d6eae` GREEN — 106 passed). It's
pre-existing, documented behavior (`Failure modes` in the module docstring,
Hook 17's charter entry), not something this PR introduced. The PR body
disclosed it as "unchanged" in the reproduction table but did not argue the
resulting asymmetry on its merits.

## A plausible defense (not yet made explicit anywhere)

Total failure most likely signals a systemic outage (auth broken, `gh`
missing, whole API down) rather than a transient per-repo blip, and per-PR
policy already has no in-band bypass — the only escape is removing the
`Skill` matcher from `settings.json`, which is not per-skill and would also
disable enforcement on `/wave-wrapup`/`/wave-retro`. Blocking on total failure
would apply the same "stranding" argument #1229 used to justify NOT
coverage-blocking `/handoff`, but now to the two skills that most need the
[...]

PR #1238, opened 2026-08-03
audit(G6): org_repos single source of truth

### LINKED ISSUE #1118: audit(G6): org_repos single source of truth
Source: CODE_AUDIT.md §10 (noorinalabs-main @ 3bdd620) — item G6.

**G6 — Single source of truth for the org repo list** · Type: correctness/DRY (fixes BUG-08) · Effort: S
7 hardcoded lists (one already drifted); comments are the only enforcement. Review widened the surface: beyond the seven audited lists, at least `w4-kickoff.py` and `w4-project-add.py` also carry repo lists omitting ingest-platform (while `warn_ghcr_image.py`'s subset is deliberate — GHCR-publishing repos only) — G6's sweep should enumerate every literal repo list, not just the seven. Add `.claude/lib/org_repos.py` (or read `ontology/repos/*.yaml`), import everywhere, and add the prose→code test asserting the list matches the CLAUDE.md Repository Map (the repo already has gates of exactly this shape).
*After:* **≈ −40 lines**, drift class gone, handoff bug fixed. *Verify:* new `test_org_repos.py`; `test_session_handoff.py` asserts all 8 repos queried.

This fixes BUG-08's class: the single `org_repos` source replaces the 7 hardcoded lists that BUG-08's drift (omitting isnad-ingest-platform) originated from.

Index row (§13): C/D | 7 lists → 1 + prose→code test | S | —

parametrization:
## Summary

Adds `.claude/lib/org_repos.py` as the single source of truth for the org repo list (`ORG`, `MAIN_REPO`, `CHILD_REPOS`, `ALL_REPOS`) and switches every hand-copied "list of org repos" consumer in `.claude/lib` and `.claude/hooks` to import from it, per CODE_AUDIT.md item G6 (fixes BUG-08's drift class).

## How every consumer was enumerated

1. `rg -l --no-ignore "noorinalabs-isnad-graph" . -g '!*.git*'` at repo root — every file mentioning the isnad-graph repo name anywhere (33+ hits across prose docs, ontology yaml, and code).
2. Narrowed to `.claude/lib` and `.claude/hooks` (`rg -l --no-ignore "noorinalabs-isnad-graph" .claude/lib .claude/hooks`) and read every literal `noorinalabs-` occurrence in each hit (`rg -n "noorinalabs-" --no-ignore <file>...`) to distinguish an actual **enumerated list** (the DRY-violation pattern G6 targets) from a single-repo mention (a path example, an error message, a docstring reference) — the latter is not in scope.
3. Cross-checked the orchestrator-supplied consumer list from the ontology-librarian brief against my own enumeration — found it consistent but not exhaustive; my own sweep additionally caught `.claude/lib/ontology_gen/aggregate.py`'s `DEFAULT_REPOS` dict (short-name → path mapping, same underlying duplication in a different shape) and `.claude/lib/tests/test_pre_commit_ci_sync.py`'s `_CHILD_REPO_DIRS`.
4. Searched separately for `w4-kickoff.py` / `w4-project-add.py` (named in the issue) and `board_add_audit_issues.sh` (found via the repo-name grep) — all three are **frozen historical scripts** recording specific numbered GH operations at a point in time (their own docstrings say so), not general repo-enumeration utilities; consolidating them would corrupt the audit trail they exist to preserve, per their own "frozen historical script" comments. Left untouched — flagging here rather than silently dropping them from scope.
5. Confirmed via `rg -n "REPOS|ALL_REPOS|CHILD_REPOS|repo_list|REPO_LIST"` that no other flat module holds a duplicate list under a different constant name.

## Consumers switched to `org_repos`

| File | Constant | Notes |
|---|---|---|
| `.claude/hooks/session_handoff.py` | `_get_open_prs()`'s repo list | **The BUG-08 drift site** — the old 7-item list omitted `noorinalabs-isnad-ingest-platform`, so `/handoff` silently queried 7 of 8 repos. Now `ALL_REPOS` (8). |
| `.claude/hooks/validate_wave_audit.py` | `_ORG_REPOS` | Was already correct (8), now sourced from the SSOT instead of hand-copied. |
| `.claude/lib/apply_implementor_labels.py` | `ALL_REPOS` | List (not tuple) preserved for existing mutate/slice call sites. |
| `.claude/lib/check_child_checkouts.py` | `CHILD_REPOS` | Direct re-export. |
| `.claude/lib/red_sweep.py` | `CANONICAL_REPOS`, `_MAIN_REPO`, `_ORG` | Direct import/alias. |
| `.claude/lib/roster_union_sync.py` | `DEFAULT_CHILD_REPOS` | Filters `CHILD_REPOS` to exclude `noorinalabs-deploy` — preserved as a **deliberate** exclusion (no aggregated `roster.json` there), not silently widened. |
| `.claude/lib/premise_check.py` | `_KNOWN_ROOT_COMPONENTS` | Repo-name subset of the frozenset now unioned in from `ALL_REPOS` instead of hand-copied. |
| `.claude/lib/ontology_gen/aggregate.py` | `DEFAULT_REPOS` | Rebuilt from `CHILD_REPOS` (short-name-stripped) + the `"main": "."` special case. |
| `.claude/lib/tests/test_pre_commit_ci_sync.py` | `_CHILD_REPO_DIRS` | Direct re-export. |

**Deliberately left as its own subset (not unified):** `.claude/hooks/warn_ghcr_image.py`'s `REPO_IMAGE_MAP` covers only the GHCR-image-publishing children — a different axis (build/publish topology) than "which repos exist in the org." Forcing it onto `CHILD_REPOS` would incorrectly widen it to non-publishing repos. Instead, `test_warn_ghcr_image.py` gained a drift guard asserting its keys stay a subset of `org_repos.CHILD_REPOS` (catches a typo/rename) and a test documenting the subset is proper (not silently widened to the full list).

## Tests

- New `.claude/lib/tests/test_org_repos.py` — constant shape (membership, no dupes, `ALL_REPOS = (MAIN_REPO, *CHILD_REPOS)`), plus the issue's explicit ask: a **prose→code test** that parses CLAUDE.md's `## Repository Map` table and asserts `CHILD_REPOS` matches it row-for-row (non-vacuous — a synthetic-table self-test guards against the parser silently matching zero rows).
- New `GetOpenPrsQueriesAllOrgReposTests` in `test_session_handoff.py` — mocks `_run` and asserts `_get_open_prs()` queries all 8 repos including `noorinalabs-isnad-ingest-platform` (regression coverage for BUG-08), plus an identity check that `hook.ALL_REPOS is org_repos.ALL_REPOS`.
- New `RepoImageMapDriftGuardTests` in `test_warn_ghcr_image.py` (see above).

## Verified

- `PYTHONDONTWRITEBYTECODE=1 ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/ .claude/lib/tests/ -q` — 3448 passed, 647 subtests passed.
- `python3 -m ruff format --check .claude/lib .claude/hooks` / `python3 -m ruff check .claude/lib .claude/hooks` — clean.
- `python3 -m mypy .claude/hooks/ .claude/lib/ --ignore-missing-imports --no-error-summary` — no errors.
- `python3 .claude/lib/pre_commit_ci_sync.py .` — `OK: pre-commit config mirrors all CI-enforced checks.`
- `pre-commit run --all-files` and `pre-commit run --all-files --hook-stage pre-push` — both fully green (ruff-format, ruff-lint, actionlint, cspell, fixture-realism, skill-graphql-pagination, checksums-ascii, mypy, pytest, pytest-skills, memory-budget, headcount-budget, doc-freshness (advisory), office-drift, mermaid-render).
- `git push` triggered the same pre-push suite again (identical hooks config) — green, and the branch pushed cleanly.

Not yet verified: live CI on the PR itself (will check `gh pr view --json statusCheckRollup` once it reports).

Closes #1118

### DIFF
--- a/.claude/hooks/session_handoff.py
+++ b/.claude/hooks/session_handoff.py
@@ -14,6 +14,7 @@ Exit codes:
 """
 
 import json
+import os
 import re
 import subprocess
 import sys
@@ -21,6 +22,10 @@ from datetime import datetime, timezone
 from pathlib import Path
 
 REPO_ROOT = Path(__file__).resolve().parent.parent.parent
+
+_LIB_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), "..", "lib")
+sys.path.insert(0, os.path.abspath(_LIB_DIR))
+from org_repos import ALL_REPOS  # noqa: E402
 
 # Project memory is version-controlled in-repo (#732 relocation) so it travels
 # with a git pull. The handoff file itself stays gitignored — it's per-session
@@ -64,18 +69,14 @@ def _get_git_state() -> dict:
 
 
 def _get_open_prs() -> list[str]:
-    """Get open PRs across all repos."""
-    repos = [
-        "noorinalabs-main",
-        "noorinalabs-isnad-graph",
-        "noorinalabs-user-service",
-        "noorinalabs-deploy",
-        "noorinalabs-design-system",
-        "noorinalabs-landing-page",
-        "noorinalabs-data-acquisition",
-    ]
+    """Get open PRs across all repos.
+
+    Queries org_repos.ALL_REPOS (main#1118 / audit G6) — the previous hand-copied
+    list here omitted noorinalabs-isnad-ingest-platform (BUG-08), silently
+    querying only 7 of 8 org repos.
+    """
     prs = []
-    for repo in repos:
+    for repo in ALL_REPOS:
         raw = _run(
             f"gh pr list --repo noorinalabs/{repo} --state open --json number,title --limit 5",
             timeout=15,
--- a/.claude/hooks/tests/test_session_handoff.py
+++ b/.claude/hooks/tests/test_session_handoff.py
@@ -109,6 +109,40 @@ class GetWaveStatusTests(unittest.TestCase):
         self.assertEqual(result, "No cross-repo-status.json found")
 
 
+class GetOpenPrsQueriesAllOrgReposTests(unittest.TestCase):
+    """main#1118 / audit G6, BUG-08 regression: `_get_open_prs()` used to iterate
+    a hand-copied 7-repo list that silently omitted
+    noorinalabs-isnad-ingest-platform. It now iterates org_repos.ALL_REPOS, so
+    this asserts all 8 org repos are queried (non-vacuous: fails if the count
+    regresses to 7, or if a caller reintroduces a hand-copied list)."""
+
+    def test_queries_all_eight_repos(self) -> None:
+        from unittest.mock import patch
+
+        queried: list[str] = []
+
+        def _fake_run(cmd: str, cwd: str | None = None, timeout: int = 10) -> str:
+            for repo in hook.ALL_REPOS:
+                if f"noorinalabs/{repo}" in cmd:
+                    queried.append(repo)
+                    break
+            return "[]"
+
+        with patch.object(hook, "_run", side_effect=_fake_run):
+            hook._get_open_prs()
+
+        self.assertEqual(sorted(queried), sorted(hook.ALL_REPOS))
+        self.assertEqual(len(queried), 8)
+        self.assertIn("noorinalabs-isnad-ingest-platform", queried)
+
+    def test_all_repos_is_org_repos_ssot(self) -> None:
+        # session_handoff.ALL_REPOS must BE org_repos.ALL_REPOS (imported, not
+        # re-derived) so the two can never drift apart again.
+        import org_repos
+
+        self.assertIs(hook.ALL_REPOS, org_repos.ALL_REPOS)
+
+
 class HandoffPathLocationTests(unittest.TestCase):
     """#741: the Stop hook must write the handoff into the in-repo,
     version-controlled .claude/memory/ — NOT the user-space auto-memory dir —
--- a/.claude/hooks/tests/test_warn_ghcr_image.py
+++ b/.claude/hooks/tests/test_warn_ghcr_image.py
@@ -144,5 +144,26 @@ class WarningBehaviorTests(unittest.TestCase):
         self.assertIsNone(hook.check(_input(cmd)))
 
 
+class RepoImageMapDriftGuardTests(unittest.TestCase):
+    """main#1118 / audit G6: REPO_IMAGE_MAP is a deliberate GHCR-publishing
+    SUBSET of the org's children, not the full org_repos.CHILD_REPOS list —
+    but every key must still be a real, canonically-spelled child repo name,
+    or a typo/rename would silently drift with nothing to announce it."""
+
+    def test_every_key_is_a_canonical_child_repo(self):
+        sys.path.insert(0, str(_HOOKS_DIR.parent / "lib"))
+        from org_repos import CHILD_REPOS
+
+        self.assertTrue(set(hook.REPO_IMAGE_MAP).issubset(CHILD_REPOS))
+
+    def test_map_is_a_proper_subset_not_the_full_list(self):
+        # Documents the intentional divergence so a future "helpfully" full
+        # sync doesn't silently widen this GHCR-only map to non-publishing repos.
+        sys.path.insert(0, str(_HOOKS_DIR.parent / "lib"))
+        from org_repos import CHILD_REPOS
+
+        self.assertLess(set(hook.REPO_IMAGE_MAP), set(CHILD_REPOS))
+
+
 if __name__ == "__main__":
     unittest.main()
--- a/.claude/hooks/validate_wave_audit.py
+++ b/.claude/hooks/validate_wave_audit.py
@@ -199,6 +199,10 @@ from pathlib import Path
 sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
 from annunaki_log import log_pretooluse_block  # noqa: E402
 
+_LIB_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), "..", "lib")
+sys.path.insert(0, os.path.abspath(_LIB_DIR))
+from org_repos import ALL_REPOS  # noqa: E402
+
 # Skills gated by this hook. Exact match against tool_input.skill.
 _GATED_SKILLS = frozenset({"wave-wrapup", "wave-retro", "handoff"})
 
@@ -222,18 +226,11 @@ _GATED_SKILLS = frozenset({"wave-wrapup", "wave-retro", "handoff"})
 # green, but not a dead end either.
 _COVERAGE_BLOCKING_SKILLS = frozenset({"wave-wrapup", "wave-retro"})
 
-# Org-known repos for cross-repo audit. Sourced from charter skills.md
-# § Audit command. This list MUST stay in sync with that charter section.
-_ORG_REPOS = (
-    "noorinalabs-main",
-    "noorinalabs-isnad-graph",
-    "noorinalabs-user-service",
-    "noorinalabs-deploy",
-    "noorinalabs-design-system",
-    "noorinalabs-landing-page",
-    "noorinalabs-data-acquisition",
-    "noorinalabs-isnad-ingest-platform",
-)
+# Org-known repos for cross-repo audit. Sourced from org_repos.py (main#1118
+# / audit G6) — the single source of truth for the org repo list; the charter
+# skills.md § Audit command example command should stay in sync with it too,
+# but this constant itself no longer hand-copies the list.
+_ORG_REPOS = ALL_REPOS
 
 # Carry-forward detection patterns (case-insensitive). Any one suffices.
 # Anchored loosely — looking for explicit author intent, not accidental phrasing.
--- a/.claude/hooks/warn_ghcr_image.py
+++ b/.claude/hooks/warn_ghcr_image.py
@@ -45,7 +45,13 @@ from _shell_parse import (  # noqa: E402
 # Deploy-related workflow names/files
 DEPLOY_PATTERNS = re.compile(r"deploy|release|cd[.-]|deliver", re.IGNORECASE)
 
-# Map of repo short names to GHCR image paths
+# Map of repo short names to GHCR image paths. Deliberately NOT the full
+# org_repos.CHILD_REPOS membership (main#1118 / audit G6) — this is the
+# GHCR-image-publishing subset of children, a different axis (build/publish
+# topology) than "which repos exist in the org". test_warn_ghcr_image.py
+# asserts every key here is a real child repo name (org_repos.CHILD_REPOS) —
+# catches a typo/rename drifting silently — without forcing full membership
+# or importing org_repos into this advisory hook's runtime path.
 REPO_IMAGE_MAP = {
     "noorinalabs-isnad-graph": "ghcr.io/noorinalabs/noorinalabs-isnad-graph",
     "noorinalabs-landing-page": "ghcr.io/noorinalabs/noorinalabs-landing-page",
--- a/.claude/lib/apply_implementor_labels.py
+++ b/.claude/lib/apply_implementor_labels.py
@@ -43,16 +43,11 @@ from collections import Counter, defaultdict
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from typing import Any
 
-ALL_REPOS = [
-    "noorinalabs-main",
-    "noorinalabs-isnad-graph",
-    "noorinalabs-user-service",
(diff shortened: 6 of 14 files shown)


parametrization:
**Review: LGTM**

Reviewed at head SHA `e0b671104fd29309d6fc98da32f9983285f5bb99` via `gh api .../contents/<path>?ref=<sha>` (no parent checkout). CI: all 14 checks completed/success against that SHA (verified independently via `commits/<sha>/check-runs`, not just `gh pr checks`). `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`.

**Independent consumer re-enumeration.** Re-ran the author's approach and extended it: `rg -l --no-ignore "noorinalabs-isnad-graph" .claude/lib .claude/hooks` (33 hits), then a broader sweep anchored on every one of the 7 child-repo name strings (not just isnad-graph), a separate pass for short-name-only lists (`"isnad-graph"`, `"landing-page"`, etc. without the `noorinalabs-` prefix, to catch a list built from a different anchor), and a sweep for `REPOS=`/`REPO_LIST`/`_ORG_REPOS`-shaped constant names. Every hit outside the PR's touched files was a single-repo mention (docstring example, error-message path, issue citation) or a small 2-3-item test fixture (`test_trust_signals.py`, `test_wave_merge_model.py`, `test_wave_status.py` — arbitrary sample data, not the org-repo-list pattern), never an actual enumerated org-repo list. No missed call site found.

**Anti-vacuity claim — verified by mutation, not just read.** Ran `test_org_repos.py` in a sandbox against the PR's `org_repos.py` + this repo's real `CLAUDE.md`: 10/10 pass (`PYTHONDONTWRITEBYTECODE=1`, no stale-`__pycache__` risk). Mutated `CHILD_REPOS` to drop a repo → 3 tests fail as expected (drift caught). Mutated `_table_repos_from_claude_md` to always return `[]` (simulating a heading/table-shape drift) → **both** `TestProseParserSelfTest.test_parses_a_minimal_synthetic_table` **and** the explicit `assertTrue(table_repos, ...)` non-vacuity guard in `TestProseCodeParity` fail loudly — this is the real thing, not the #1215 `assertEqual([], [])` trap.

**Diff review (all 14 files) at head SHA.** `session_handoff.py` fixes BUG-08 correctly (7→8 repos) with a test that both counts 8 and does an `assertIs(hook.ALL_REPOS, org_repos.ALL_REPOS)` identity check so it can't silently re-derive. `validate_wave_audit.py`, `red_sweep.py`, `check_child_checkouts.py`, `apply_implementor_labels.py`, `ontology_gen/aggregate.py`, `premise_check.py` all correctly import/re-export with membership and (where it existed) order preserved — checked each against its pre-PR literal list. `warn_ghcr_image.py`'s `REPO_IMAGE_MAP` kept as a deliberate 4-item GHCR-publishing subset, now with two drift-guard tests (subset + proper-subset); confirmed all 4 keys are real `CHILD_REPOS` members.

**One gap:** `roster_union_sync.py`'s `DEFAULT_CHILD_REPOS` migration (filter over `CHILD_REPOS` excluding `noorinalabs-deploy`) is behaviorally correct — I hand-verified the filter reproduces the exact prior 6-item tuple — but got no dedicated test, unlike the other two "deliberately different from the full list" consumers touched in this same PR (`warn_ghcr_image.py`, `session_handoff.py`). Checked `test_roster_union_sync.py`: every CLI test passes an explicit `--repos`, so the `DEFAULT_CHILD_REPOS` default-path branch is untested before and after this PR — a future filter-expression typo wouldn't be caught. Filed as debt (below), not a blocker: correctness is verified, CI is green, and this gap pre-dates the PR (it just wasn't closed while touching the file).

Nothing I could not verify — every claim in the PR description and brief was independently checked, not taken on trust.

---
Requestor: Lucas Ferreira
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: #1241

parametrization:
## Merge-gate review — PR #1238 at `e0b6711`

Reviewed as the authoritative pre-merge gate. Every file read at the PR head via the contents API; nothing inferred from the PR body's framing. Verification was done in a **fresh full-depth clone** of the head (`e0b6711`) in a scratch directory, never the parent working tree, with `PYTHONDONTWRITEBYTECODE=1` throughout.

**Verdict: approve.** The consolidation is behaviour-preserving on every one of the nine migrated constants, the BUG-08 fix is real and mutation-covered, and the anti-vacuity claim survives the specific trap it needed to survive. Two follow-ups filed as #1242 and #1243; neither blocks.

### 1. The all-8 vs 7-children distinction survives — checked by execution, not by reading

The sharp risk here was a consumer's meaning silently flipping between "all 8" and "the 7 children". I imported every migrated module at head and compared each constant against its **pre-PR literal**, including ordering:

| Consumer | Constant | Pre/post identical (membership **and** order) |
|---|---|---|
| `session_handoff.py` | `_get_open_prs()` repo set | changed **by design** — 7 to 8 (the fix) |
| `validate_wave_audit.py` | `_ORG_REPOS` | membership identical (8); order differs (`landing-page` moved to the end) |
| `apply_implementor_labels.py` | `ALL_REPOS` | yes, and still a `list`, so mutate/slice call sites are unaffected and cannot reach the SSOT tuple |
| `check_child_checkouts.py` | `CHILD_REPOS` | yes |
| `red_sweep.py` | `CANONICAL_REPOS`, `_ORG`, `_MAIN_REPO` | yes |
| `roster_union_sync.py` | `DEFAULT_CHILD_REPOS` | yes — the deliberate `noorinalabs-deploy` exclusion is preserved exactly, same 6 repos, same order |
| `premise_check.py` | `_KNOWN_ROOT_COMPONENTS` | yes — all 8 repo names present, and the non-repo components (`src`, `docs`, `alembic`, `docker`, `integration-tests`) are still there, so the union did not drop the hand-written half |
| `ontology_gen/aggregate.py` | `DEFAULT_REPOS` | yes — `removeprefix` reconstruction is dict-equal to the old literal **including key order** |
| `test_pre_commit_ci_sync.py` | `_CHILD_REPO_DIRS` | yes |

On `validate_wave_audit` specifically: membership is unchanged at 8, so the #1226/#1229 partial-failure hardening is untouched and the #1230 residual (total 8/8 failure still ALLOWs while partial 1/8 BLOCKs) is neither improved nor worsened by this PR. The reordering is inert — the constant is only iterated for per-repo queries and aggregation.

### 2. The BUG-08 fix is real and covered

Not taken on assertion. `_get_open_prs()` now iterates `ALL_REPOS` (8), and `test_all_repos_is_org_repos_ssot` uses `assertIs`, which ties the hook to the SSOT object rather than to an equal copy. Mutation: re-dropping `noorinalabs-isnad-ingest-platform` at the call site is **KILLED**.

### 3. Anti-vacuity — mutation-tested, including the trap that bit #1215's own remedy

The concern was that a self-test guarding a parser can itself be vacuous. Nine mutations, all **KILLED**:

| Mutation | Outcome |
|---|---|
| Drop `isnad-ingest-platform` from `CHILD_REPOS` | KILLED (parity + handoff regression test) |
| Reorder `CHILD_REPOS` (swap `deploy` / `design-system`) | KILLED — the "row-for-row" claim is literal, not set-equality |
| Hardwire the parser to return `[]` | KILLED, and by the intended assertion: `AssertionError: [] is not true : parsed zero repos from CLAUDE.md's Repository Map table` |
| **Parser regex `[a-z-]+` to `[a-z]+`** | KILLED |
| CLAUDE.md gains a table row, code does not (the BUG-08 direction) | KILLED |
| `REPO_IMAGE_MAP` key typo | KILLED by the new subset drift guard |
| `session_handoff` re-drops a repo at the call site | KILLED |
| `roster_union_sync` silently widened to include `deploy` | KILLED |
| `_ORG_REPOS = ()` | KILLED (13 failures) |

The fourth row is the one that mattered. `[a-z]+` still parses the **synthetic** table (`noorinalabs-foo`, `noorinalabs-bar` contain no hyphen) while failing on every real hyphenated repo name — precisely the shape where a self-test passes and the real parse silently degrades. It is caught because `TestProseCodeParity` ends in strict `assertEqual(tuple(table_repos), CHILD_REPOS)`, not a subset or intersection check. The explicit `assertTrue(table_repos)` guard is therefore belt-and-suspenders rather than the load-bearing assertion, which is the right way round: the guard cannot be the vacuous link because the equality would already fail on an empty parse.

### 4. Enumeration completeness — one real residual, disclosed as #1242

The stated method (`rg` anchored on `noorinalabs-isnad-graph`, then narrowed to `.claude/lib` and `.claude/hooks`) has the failure mode the brief names: a list anchored on a different string, or living outside those two directories, is invisible. I re-swept at head for **any** file enumerating 3+ org repo names, full **or** short form, across all file types.

Inside `.claude/lib` + `.claude/hooks`: clean. The author's enumeration was complete for the scope they declared, and their two self-caught additions (`aggregate.py`'s `DEFAULT_REPOS`, `test_pre_commit_ci_sync.py`'s `_CHILD_REPO_DIRS`) check out.

Outside it: `.github/workflows/docs.yml` hand-lists all 7 children in `skip_prefixes`, and the charter's `skills.md` § Audit command hand-lists all 8 in its canonical bash loop. The charter copy's membership is currently **correct** (I verified all 8, `isnad-ingest-platform` included), so there is no live drift; the `docs.yml` copy drifts in the loud direction and is inert in a parent-only CI checkout. Low severity, but they are the same class, and the `docs.yml` one was not in the PR body's "deliberately left" list. Filed as #1242 rather than held against this PR.

I also checked the migration did not break importers outside the audited directories. `.claude/skills/wave-scope/validate_matrix_names.py` lazily imports `roster_union_sync`, which now has a module-level `org_repos` dependency — I **executed** that import path and it resolves (the function inserts `.claude/lib` on `sys.path` first). No child repo vendors `validate_wave_audit.py`, `session_handoff.py`, or `premise_check.py`, so there is no cross-repo import breakage. `sys.path.insert(0, .claude/lib)` shadows nothing: zero of the 39 flat lib modules collide with a stdlib module name.

### 5. Two things I could not clear, split out as #1243

Recorded rather than inferred, and neither is a regression this PR introduces:

**A missing guard on the sharper gate.** `session_handoff` got `assertIs(hook.ALL_REPOS, org_repos.ALL_REPOS)`; `validate_wave_audit` got nothing equivalent. Replacing `_ORG_REPOS = ALL_REPOS` with a hand-copied 7-tuple dropping `isnad-ingest-platform` — the literal BUG-08 shape, on the blocking gate — **survives the entire suite: 3448 passed, zero failures.** Emptying the tuple fails 13 tests, so total emptiness is loud and a one-repo omission is invisible. Pre-PR that constant was equally untested, and the alias makes drift require an active re-typing, so this is not a regression; it is the one place the PR's own guarding standard was not applied.

**Import failure direction.** `validate_wave_audit` is a PreToolUse gate where exit 2 blocks and any other non-zero exit is a non-blocking error. At head it exits **2** with a full audit reason. With `org_repos.py` removed it exits **1** (`ModuleNotFoundError`), i.e. does not block. Reachability is low — same repo, same commit, adjacent directory — and the pre-existing `annunaki_log` import in the same file has the identical property, so this is a second instance of an existing class, not a new one. Still the wrong direction for a gate.

### Verification performed

| Check | Result |
|---|---|
| Baseline suite, clean full-depth clone at `e0b6711` | 3448 passed, 647 subtests passed — matches the author's claim exactly |
| Skills suite via the canonical per-skill invocation (mirroring the `pytest-skills` pre-push hook) | green |
| Mutation battery, 9 mutations | 9 KILLED |
| Mutation on `validate_wave_audit._ORG_REPOS` | SURVIVED — see #1243 |
| `validate_wave_audit` executed end to end against a real `wave-wrapup` input | exit 2, real BLOCK decision emitted |
| Constant-by-constant pre/post equality, 9 consumers | identical, ordering included |
| CI at head sha via `check-runs` API on `e0b6711` | 14/14 success |
| `mergeable` / `mergeStateStatus` | `MERGEABLE` / `CLEAN` |

One caveat on my own method: an earlier run of the mutation battery happened in a shared scratch path another agent was also writing into, and a shallow clone made `test_wave_status.py::BaseVsHeadDifferential` fail for want of the base commit. I re-ran every load-bearing result in an isolated full-depth clone, which is what the numbers above report; that test passes there.

Good work on this one. The three frozen historical scripts being called out in the PR body instead of silently dropped, and the `REPO_IMAGE_MAP` subset being guarded rather than "helpfully" unified, are both the right calls and the reason this reviewed cleanly.

---
Requestor: Nadia Khoury
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: #1242, #1243

PR #1239, opened 2026-08-03
audit(G2): pre_file dispatcher matcher

### LINKED ISSUE #1114: audit(G2): pre_file dispatcher matcher
Source: CODE_AUDIT.md §10 (noorinalabs-main @ 3bdd620) — item G2.

**G2 — `pre_file` dispatcher matcher** · Type: perf · Effort: S
Edit/Write currently spawn 3 Python processes (2 pre + 1 post); both pre-hooks already expose `check()` and the Post side is already consolidated. Add the matcher table + `hooks.pre_file` config.
*After:* **−33% interpreter startups per file edit** (~30–50 ms each); ~24 lines of settings.json removed; Edit-phase gates become config like every other phase. *Verify:* `test_hook_registration_coverage.py` gains `pre_file`/`pre_notebook` keys (its assertions then prove nothing fell out of reach); keep `INTENTIONALLY_UNREGISTERED` empty.

Index row (§13): P | 3→2 processes/edit | S | —

parametrization:
## Summary
- Adds a matcher table to `dispatcher.py` (mirroring `post_dispatcher.py`'s `_MATCHER_CFG_KEY`) so Edit/Write/NotebookEdit PreToolUse calls route through the same in-process dispatcher Bash already used, reading their module lists from new `hooks.pre_file` / `hooks.pre_notebook` framework-config keys. `settings.json` drops the two standalone `enforce_librarian_consulted.py` + `validate_edit_completion.py` command entries per matcher in favor of one `dispatcher.py` entry each (3 processes/edit -> 2, per the audit's G2 item).
- Fixes a fail-open trap this consolidation would otherwise have introduced: `validate_edit_completion.check()` previously short-circuited to `None` for any `tool_name != "Bash"` (a leftover from when only Bash was dispatched). Left as-is, routing Edit/Write/NotebookEdit through the new matcher table would have silently dropped the same-path re-edit block. `check()` now routes Bash + the edit tools to `_pre_tool_use_blocks`; `SendMessage` keeps its standalone `settings.json` registration since it has no PreToolUse dispatcher entry.
- `test_hook_registration_coverage.py`'s `HOOK_LIST_KEYS` gains `pre_file`/`pre_notebook` so the orphan-hook and inert-entry guards cover the new keys (per the issue's verification note).
- Adds `test_dispatcher.py` (mirrors `test_post_dispatcher.py`) covering matcher routing, block/warning semantics, and fail-open behavior for all four tool types, plus `CheckDispatchTests` in `test_validate_edit_completion.py` pinning the `check()` routing fix.
- Also splits the local pre-push `mypy` hook in `.pre-commit-config.yaml` into two (hooks+lib flat, skills per-directory-looped) to exactly mirror `ci.yml`'s mypy job. The prior single flat invocation across `.claude/{hooks,lib,skills}/` fed a large file list through pre-commit's xargs batching; whenever the `skills/wave-scope` and `skills/wave-wrapup` `tests/__init__.py` files landed in the same batch, mypy raised a spurious "Duplicate module named tests" — order-dependent on unrelated file-count changes elsewhere in the tree, and surfaced by this PR's new test file. CI's real mypy job already avoids this via a per-skill loop; the local hook now matches it exactly.

## Test plan
- [x] `PYTHONDONTWRITEBYTECODE=1 ENVIRONMENT=test python3 -m pytest .claude/hooks/tests .claude/lib/tests -q` — 3462 passed, 647 subtests passed
- [x] Mutation-killed the `check()` routing fix: reverting the `_DISPATCHED_PRE_TOOLS` guard to `tool_name != "Bash"` fails 3 new `CheckDispatchTests` deterministically; restored and re-verified green
- [x] Mutation-killed the matcher table: dropping `"Edit": "hooks.pre_file"` from `_MATCHER_CFG_KEY` fails 7 tests in `test_dispatcher.py`/`test_hook_registration_coverage.py`; restored and re-verified green
- [x] `pre-commit run --all-files` — all commit-stage hooks pass
- [x] `pre-commit run --hook-stage pre-push --all-files` — all push-stage hooks pass (mypy, mypy-skills, pytest, pytest-skills, memory-budget, headcount-budget, doc-freshness, office-drift, mermaid-render)
- [x] `python3 .claude/lib/pre_commit_ci_sync.py .` — OK, mirrors all CI-enforced checks

Closes #1114


parametrization:
**Review: LGTM — approving.**

Read every changed file at head `ff39d453` via the contents API. CI is fully green (21/21 checks SUCCESS, including "Pytest — hooks and lib test suites" and "Mermaid render gate" which were still pending when this review was assigned — re-checked `statusCheckRollup` and both completed SUCCESS). `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.

**What I verified end-to-end (not just by reading source):**

1. **The `check()` routing fix is real and load-bearing.** I built a scratch tree mirroring the repo layout, populated it with the PR-head `dispatcher.py` / `validate_edit_completion.py` / `_framework_config.py` / `framework.config.json`, and ran the actual PreToolUse `dispatcher.py` end-to-end: seeded an Edit-error sentinel, then fired a same-path `Edit` re-attempt through `dispatcher.py` exactly as `settings.json` now wires it. It blocked (exit 2) with the expected reason. I then confirmed the *base* commit (`3a68824`) never routed Edit/Write/NotebookEdit through `check()` at all — the base `dispatcher.py` was hardcoded `if tool_name != "Bash": sys.exit(0)` and Edit/Write/NotebookEdit ran only via `validate_edit_completion.py`'s own standalone `main()` (which calls `_pre_tool_use_blocks` unconditionally, bypassing `check()`'s Bash-only gate). So the `if tool_name != "Bash": return None` line in `check()` was a **latent trap**, inert at base, that this PR's own dispatcher-consolidation would have silently activated for Edit/Write/NotebookEdit had it not also been fixed — confirmed by reverting only that one guard in the PR-head tree (`_DISPATCHED_PRE_TOOLS` check → `tool_name != "Bash"`) and re-running the same live-dispatcher repro: the block silently disappears (exit 0, no output). This is exactly the "fails open and silently" risk called out in the brief, and it's closed.

2. **Mutation-testing re-run (not accepted on the author's word).** Ran with `PYTHONDONTWRITEBYTECODE=1` against a full copy of `.claude/hooks/tests/test_dispatcher.py` + `test_validate_edit_completion.py` + `test_hook_registration_coverage.py` overlaid on the PR-head source:
   - Unmutated: 89/89 pass.
   - Revert the `check()` routing fix only (`_DISPATCHED_PRE_TOOLS` → `!= "Bash"`): **exactly 3 tests fail** — `test_check_routes_preTooluse_edit_to_block`, `test_check_routes_preTooluse_notebook_edit_to_block`, `test_check_routes_preTooluse_write_to_block`. Matches the author's claim exactly.
   - Revert the matcher-table addition in `dispatcher.py` (tried two shapes: deleting the three new `_MATCHER_CFG_KEY` rows, and reverting `main()`'s use of `_modules_for` back to the hardcoded `hooks.pre_bash`-only loop): **8–11 tests fail** depending on exact mutation shape, not the claimed 7. Directionally this is *more* coverage than claimed, not less, so it isn't a safety concern — but I couldn't reproduce the precise "7" figure the author cited, and I'm noting that rather than passing it through unverified.

3. **Nothing dropped from Edit/Write/NotebookEdit.** Diffed `settings.json` before/after: previously each of Edit/Write/NotebookEdit had two standalone entries (`enforce_librarian_consulted.py`, `validate_edit_completion.py`, timeout 10 each); now each has one `dispatcher.py` entry (timeout 30) that reads `hooks.pre_file` (Edit/Write) / `hooks.pre_notebook` (NotebookEdit) from `framework.config.json`, and both lists contain exactly those same two module names. `SendMessage` is untouched — still standalone in `settings.json` alongside `block_shutdown_without_retro.py`, confirmed by reading the full file (not just the diff hunks) and by the new `test_check_sendmessage_returns_none_not_dispatched` test. `_framework_config._DEFAULTS` matches the committed `framework.config.json` byte-for-byte on the new `pre_file`/`pre_notebook` keys.

4. **`.pre-commit-config.yaml` mypy split.** Confirmed the new `mypy` command (`mypy .claude/hooks/ .claude/lib/ --ignore-missing-imports --no-error-summary`) and the new `mypy-skills` per-directory loop are byte-for-byte the same invocations `ci.yml` already runs (lines 79 and 85–98) — this closes a *pre-existing* local⇄CI parity gap (the old flat pre-commit hook combined hooks+lib+skills with `pass_filenames: true` and a `conftest.py`/`worktrees` exclude, which never matched CI's already-split shape). The PR's own comment ties this to a real, testable cause: adding `test_dispatcher.py` (a new file under `.claude/hooks/tests/`) shifts pre-commit's xargs file-count batching, which is what let a `tests/__init__.py` collision between two `.claude/skills/*/tests/` dirs surface only locally and only nondeterministically. I confirmed `.claude/worktrees/` is a sibling of `.claude/hooks/`/`.claude/lib/`, not nested under them, so dropping the old `exclude` regex in favor of directory-scoped invocation is safe. Ran `mypy .claude/hooks/dispatcher.py .claude/hooks/validate_edit_completion.py --ignore-missing-imports --no-error-summary` locally against the PR-head files — clean. This change belongs in this PR (it's caused by this PR's own new test file) and does mirror CI.

5. **Charter catalog docs (`catalog-13-17.md`, `catalog-18-22.md`).** Cross-checked every routing claim in the prose against the actual code paths I exercised in (1)–(3) above, including the `SendMessage` carve-out language ("has no PreToolUse dispatcher entry... reachable only via the standalone `main()`") — accurate.

**Not fully verified:** the author's specific "reverting the matcher table fails 7 tests" figure — my own mutation runs landed at 8 or 11 depending on which lines I reverted, both directionally consistent with "well covered" but not matching 7 exactly. Non-blocking; flagging per the request to report what I couldn't verify.

---
Requestor: Weronika Zielinska
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Review: LGTM — approving. Merge-gate verification complete.**

Reviewed at head `ff39d453` via the contents API (no local checkout of the PR branch). All mutation and end-to-end work ran in a throwaway detached worktree at that SHA; nothing in the PR branch was touched.

### CI

Re-derived independently of the assignment brief, which reported `pending`. `statusCheckRollup` at `ff39d453`: **21 checks, 21 SUCCESS, 0 failing** — including `Pytest — hooks and lib test suites`, `Mypy type check`, `Pre-commit ⇄ CI sync-drift gate`, and `Mermaid render gate`. `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`. Both commits authored `Nino Kavtaradze <parametrization+Nino.Kavtaradze@gmail.com>`.

### The merge-gate question: does any gate quietly stop firing?

No. I did not take this from the tests — I ran the real dispatcher as a subprocess against real, unmocked hook modules with a seeded sentinel.

- `Edit` on the errored path → exit **2**, correct block reason
- `Write` on the errored path → exit **2**
- `NotebookEdit` on the errored notebook → exit **2**
- Control, `Edit` on an unrelated path → exit **0**, no output
- `enforce_librarian_consulted` advisory through the dispatcher → exit **0** with the advisory `systemMessage` (my first attempt showed no advisory; cause was my own worktree living under `/tmp`, which the hook allowlists — re-run against a non-allowlisted path emitted it correctly)
- `SendMessage` through the dispatcher → exit **0** (no matcher, correct); through the standalone `validate_edit_completion.py` → exit **2**. Its standalone registration survives in `settings.json`, as claimed.

### The fail-open trap is real, and the fix is load-bearing

Confirmed directly rather than by reading. With the pre-#1114 guard restored (`if tool_name != "Bash": return None`), a same-path `Edit` re-edit dispatched through the new matcher table exits **0** — the block silently vanishes. That is the early-allow shape this repo keeps getting bitten by, and consolidating without the `check()` fix would have shipped it. Fixing it inline rather than filing it was the right call.

### Effective-hook-set equivalence

Enumerated both sides from `settings.json`. Before, per matcher `Edit` / `Write` / `NotebookEdit`: `enforce_librarian_consulted.py` then `validate_edit_completion.py`, two processes, timeout 10 each. After: one `dispatcher.py` entry resolving `hooks.pre_file` (Edit, Write) / `hooks.pre_notebook` (NotebookEdit) to the same two modules in the same order. Nothing dropped, nothing reordered. `Bash`, `Agent`, `SendMessage`, `Skill`, all `PostToolUse` entries, `SessionStart` and `Stop` are byte-identical to base.

### Mutation testing — counts re-derived, not accepted

All runs `PYTHONDONTWRITEBYTECODE=1 ENVIRONMENT=test`, full `.claude/hooks/tests/` suite. Baseline at head: **2420 passed**.

- Revert the `check()` routing guard to `tool_name != "Bash"` → **3 failed**, all `CheckDispatchTests`. Matches the author's claim exactly.
- Drop `"Edit": "hooks.pre_file"` from `_MATCHER_CFG_KEY` → **7 failed**. Matches exactly.
- Delete `pre_file` + `pre_notebook` from both `framework.config.json` and `_framework_config._DEFAULTS` → **11 failed**, including the orphan-hook guard `test_every_check_exposing_hook_is_reachable`. This is the realistic silent-gate-loss shape and it is caught.
- Remove the `Edit` matcher entry from `settings.json` entirely → **1 failed**, `SettingsRegistrationTests`.
- Drop `validate_edit_completion` from `hooks.pre_file` in the config only → **4 failed**, including `test_config_file_matches_defaults`.
- Typo a module name in `pre_file` in both config and defaults (the "reads as enabled, never runs" inert entry) → **23 failed**, including `test_every_dispatched_module_exposes_check` and `test_defaults_hook_lists_expose_check`.

I also checked whether the anti-vacuity change is itself vacuous, since that trap has bitten inside its own remedy here before. Reverting `HOOK_LIST_KEYS` to the pre-#1114 tuple **on its own, with no other change**, fails `test_every_check_exposing_hook_is_reachable` — because `enforce_librarian_consulted` no longer has a standalone `settings.json` entry, so an unscanned `pre_file` makes it unreachable. The two halves of this PR interlock; neither guard is inert.

### Timeout 10 → 30

Correct direction and correctly sized. The wall-clock budget for this matcher went from 2 × 10s in separate processes to 30s in one, so headroom increased; it also matches the `Bash` PreToolUse entry (30s over 23 modules) and the `PostToolUse` Edit entry. On timeout a PreToolUse hook fails **open**, so the coupling cost is real but bounded: a hang in either module now consumes the shared budget and can take both gates down, where before it took down only its own. Both modules full-parse the session transcript, so the worst case is a very long session — 30s is a sane ceiling for that, and the in-process form removes two interpreter startups per edit. I am accepting the documented fail-open-on-timeout semantics from the harness contract; I did not induce a real timeout to observe it.

### Scope

Both out-of-diff-theme changes belong here.

The two charter catalog edits (Hook 15, Hook 18) describe exactly the registration this PR rewires — leaving them would have shipped a doc that contradicts `settings.json` on the day of merge. Both new texts are accurate against the code I read.

The `.pre-commit-config.yaml` mypy split is causally forced by this PR: adding a test file changed pre-commit's xargs batch composition and tripped the `Duplicate module named tests` collision, so the author could not push without it. It makes local⇄CI parity strictly better — the new `mypy .claude/hooks/ .claude/lib/ --ignore-missing-imports --no-error-summary` and the per-skill loop now mirror `ci.yml` **verbatim**, where the previous single flat file-list invocation did not mirror CI at all. I ran both commands against the real engine in the worktree: exit 0, exit 0, notes only. The dropped `exclude: ^\.claude/worktrees/` is not a regression — mypy is now pointed at two directories that do not contain worktrees. The `bash -c` loop is genuinely bash, so its `status=` assignment is safe under this repo's zsh shell.

On consolidation policy: these matchers carried 2 hooks each, below the >3 mandatory-consolidation threshold, so this is early rather than required. Fine — the policy sets a ceiling, not a floor, and the process-count argument stands.

### Findings filed

Three, none blocking. The gate itself was verified firing for all three matchers before I wrote any of them.

- #1244 — `validate_edit_completion` calls `log_pretooluse_block` only in `main()`, never in `check()`. Every sibling hook the dispatcher routes logs from `check()` (`block_bare_grep`, `block_no_verify`, `validate_commit_identity`). Before this PR the Edit/Write/NotebookEdit blocks went through `main()` and were recorded; now they are invisible to `/annunaki`. The store confirms the pattern: 119 records for `block_bare_grep`, 45 for `validate_commit_identity`, zero for `validate_edit_completion`. The block still fires — only the record is lost.
- #1245 — the dispatcher discards aggregated `systemMessage`s when a later hook blocks, but `enforce_librarian_consulted` writes its once-per-session throttle marker *before* returning the advisory. Reproduced live: block emitted at exit 2, advisory absent from output, throttle marker reads emitted `True`. The session's single librarian nudge is consumed without ever being shown. Could not happen before this PR on these matchers, since the two ran as independent processes.
- #1246 — pre-existing, not this PR's making: the charter hook docs still reference `_BASH_HOOKS` and `bash_dispatcher.py`, neither of which exists. The emergency-override lines are the sharp edge — an operator following them would edit nothing and believe a gate was disabled while it kept firing. This PR correctly updated the two entries its own change touched.

### Stated on the record as unverified

Two things I judged rather than observed. First, I cannot see harness-side rendering, so I did not confirm that `{"decision": "allow", "systemMessage": ...}` from the dispatcher surfaces identically to the bare `{"systemMessage": ...}` the standalone hook emitted — I am relying on this being the already-shipping shape for every Bash advisory. Second, I did not induce a real dispatcher timeout.

Must-fix: None.

---
Requestor: Aino Virtanen
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: #1244, #1245, #1246

Issue #1183, opened 2026-07-30
tech-debt: roster.json has no prune/liveness gate — 18 of 78 manifest entries are backed by no roster card anywhere


parametrization:
Found during review of #1178 (fix for #1162), head `256f79a0`. Not a blocker for that PR — filed as follow-up.

## The gap

`.claude/team/roster.json` (78 entries) is guarded in exactly one direction. `.claude/lib/roster_union_sync.py` asserts the parent manifest is a **superset** of the child union — it names personas the parent is *missing*. Nothing ever asserts the reverse, so a name added to the manifest is never removed when its card disappears, and the job is `continue-on-error` besides.

Measured against the working tree today, comparing the manifest to the union of **every** roster card (`*.md`, parent + all 7 cloned children) **and** every child `roster.json`:

```
manifest: 78   parent cards: 9   union of all cards: 60
still unbacked by ANY card or ANY child roster.json: 18
  Amara Diallo, Annunaki, Carolina Mendez-Rios, Cedric Novak, Dmitri Volkov,
  Elena Petrova, Fatima Okonkwo, Hiro Tanaka, Kwame Asante, Mei Chen,
  Priya Nair, Rashid Osei-Mensah, Renaud Tremblay, Sable Nakamura-Whitfield,
  Steven French, Sunita Krishnamurthy, Tomasz Wojcik, Yara Hadid

card-only (card exists, absent from manifest): 0
```

The one-directional gate is doing its job — nothing is missing. But 18 of 78 entries are backed by nothing in the org, including `Annunaki`, which is a monitor identity rather than a person.

## Why it matters now

#1178 makes review-class slots resolve against this manifest. That is the right mechanism — when a child repo is not cloned, the manifest is the only place a third-child persona can be recognised. The consequence is that all 18 unbacked names now satisfy check #319 in a `reviewer` / `reviewer_2` / `merge_gate_reviewer` slot, where before they were caught as UNRESOLVED.

Check #319 exists to answer "does this name exist at all?" — its stated origin is a stale alias (`Anya Volkov`) propagating through scope. For review-class slots that check now has no liveness property: an offboarded persona whose card was deleted stays resolvable indefinitely, because nothing prunes the manifest.

No commit or membership consequence — review-class slots never reach the #1134 branch, and `membership` stays `n/a`. This is about the strength of check #319, not about #1134.

## Options

[...]

Issue #1223, opened 2026-08-02
bug(hook4/charter_trailer): surname comparison does not Unicode-fold, so a diacritic-spelling mismatch defeats the self-review exclusion — and roster.json carries Cedric Novak twice


parametrization:
Found reviewing PR #1215 (#1210). **Pre-existing** in `.claude/lib/charter_trailer.py`, not introduced by #1215 — but #1210 makes it newly load-bearing, because the commit-derived author name is now the *only* self-review discriminator on every ref without a `{Initial}.{Lastname}` prefix.

## The defect

`charter_trailer.is_branch_author` compares surnames with a bare `.lower()` and no Unicode normalization:

```python
if name_lastname(field_value).lower() != branch_lastname.lower():
    return False
```

`Novák`.lower() != `Novak`.lower(). So one person spelled two ways is two people to the gate.

## It is reachable, not theoretical

`.claude/team/roster.json` has **four** personas whose display name carries diacritics but whose `+alias` address is ASCII-folded:

| roster name | address alias |
|---|---|
| `Tomasz Wójcik` | `Tomasz.Wojcik` |
| `Carolina Méndez-Ríos` | `Carolina.Mendez-Rios` |
| `Petra Vidović` | `Petra.Vidovic` |
| `Cédric Novák` | `Cedric.Novak` |

and it carries **`Cedric Novak` AND `Cédric Novák` as two separate entries pointing at the same address** `parametrization+Cedric.Novak@gmail.com`. Git history across the org shows both spellings actually used as `commit.author.name` under that one address:

```
Cedric Novak <parametrization+Cedric.Novak@gmail.com>
Cédric Novák <parametrization+Cedric.Novak@gmail.com>
```

## Reproduction against PR #1215's head (`0186d4d`)

```
commits authored as "Cedric Novak" <parametrization+Cedric.Novak@gmail.com>
derived: (CommitAuthorIdentity(lastname='Novak', initial='c', display='Cedric Novak'),)
  is_self_review('Cedric Novak')  = True
  is_self_review('Cédric Novák')  = False    <-- self-review NOT excluded
```

i.e. commit under the ASCII spelling, post `Requestor: Cédric Novák` on your own `feature/x` branch, and the exclusion #1210 exists to provide does not fire. This is the **fail-open** direction, in the exact mechanism.

The reverse case is rescued only by #1210's name+alias union — an accident worth keeping:

```
commits authored as "Cédric Novák"
[...]

PR #1240, opened 2026-08-03
fix(#1181): manifest-only names no longer resolve as review-class reviewers

### LINKED ISSUE #1181: roster.json has 18 manifest-only names (incl. non-persona 'Annunaki') that #1178 now accepts as reviewers; the sync-drift gate is one-directional and advisory

parametrization:
## Summary

`#1181` measured 18 manifest-only names in `.claude/team/roster.json` (78 entries vs a 60-name union of every roster card in the org), of which `Annunaki` (the error-monitor's own commit identity) and `Steven French` (the bare `parametrization@gmail.com` principal, no `+alias`) are definitionally not reviewers.

`.claude/skills/wave-scope/validate_matrix_names.py`'s `_load_org_manifest_names` widened review-class slot resolution (`reviewer`/`reviewer_2`/`merge_gate_reviewer`) to every manifest key with no shape filter — unlike its merge-time sibling in `.claude/hooks/validate_pr_review.py` (already fixed by #1179/PR #1199), which excludes non-persona entries via `_PERSONA_ALIAS_RE`. So `reviewer: "Annunaki"` resolved at scope time even though the merge-time gate already rejected it. This PR ports the same filter, closing that drift between the two gates.

It also adds the reverse assertion `.claude/lib/roster_union_sync.py` was missing (raised in #1181's consolidating comment): manifest ⊆ (this repo's cards) ∪ (every child's card directory) ∪ (every child's own `roster.json`). Advisory only (`continue-on-error`, unchanged), and skips entirely when any fetch is incomplete rather than risk a false "prune this" verdict on a name we simply failed to read.

**A note on scope:** while building the reverse check, the assumption "a child's own `roster.json` is a complete substitute for that child's cards" (#1183's `card-only == 0` measurement) turned out to have already drifted — live data shows two real, carded personas absent from their own repo's `roster.json` (`noorinalabs-isnad-graph`'s `data_scientist_mei.md` / Mei-Lin Chang, and `noorinalabs-deploy`'s `sre_engineer_aisha.md` / Aisha Idrissi, the latter also invisible to every roster.json-only check since `noorinalabs-deploy` has no `roster.json` at all). Rather than ship a check that immediately false-positives on two active people, `fetch_child_card_names` fetches every child's card directory directly. Verified against real org data — reproduces exactly the 16 unreconciled + 2 non-persona split #1181/#1183 measured, with no false positives.

**Explicitly out of scope for this PR** (left for an owner decision, not invented here):
- Disposition of the other 16 persona-shaped, still-uncarded manifest names (prune vs onboard) — the new reverse check surfaces them as advisory `unreconciled` orphans but does not decide for a human.
- Reconsidering `continue-on-error: true` on the `roster-union-sync` CI job (`commit-identity.yml:56`) — a blocking-posture change to an org-wide gate, not something to change unilaterally.
- Restoring a full "liveness" property (auto-pruning an offboarded persona) — the new check gives ongoing advisory visibility instead of automatic removal, which stays a human-approved diff per the org's memory-pruning convention.

## Changes
- `.claude/skills/wave-scope/validate_matrix_names.py` — `_load_org_manifest_names` now filters manifest entries to charter-conformant persona addresses (`_PERSONA_ALIAS_RE`, same matcher as `validate_pr_review.py`'s #1179 fix).
- `.claude/skills/wave-scope/tests/test_validate_matrix_names.py` — new `PersonaShapeFilterTests` (Annunaki/bare-principal exclusion, persona-shaped-but-uncarded names still resolve); fixed one existing fixture that used non-persona-shaped test emails.
- `.claude/lib/roster_union_sync.py` — new `parent_roster_map`, `local_card_names`, `fetch_child_card_names`, `compute_manifest_orphans`; `main()` now also runs and reports the reverse (orphan) check.
- `.claude/lib/tests/test_roster_union_sync.py` — new pure-function tests, CLI wiring tests (including the false-positive regression guards for Mei-Lin Chang / Aisha Idrissi), and a non-vacuous regression pinned to the real #1181 18-name measurement (16 unreconciled + 2 non-persona).

## Test plan
- [x] `PYTHONDONTWRITEBYTECODE=1 ENVIRONMENT=test python3 -m pytest .claude/skills/wave-scope/tests/ -q` — 45 passed
- [x] `PYTHONDONTWRITEBYTECODE=1 ENVIRONMENT=test python3 -m pytest .claude/lib/tests/test_roster_union_sync.py -q` — 31 passed
- [x] `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/ .claude/lib/tests/ -q` — 3456 passed, 647 subtests passed
- [x] Full skill test suite loop (mirrors `pytest-skills` pre-push hook) — all green
- [x] `python3 -m mypy --ignore-missing-imports --no-error-summary` on all 4 changed files — no issues
- [x] `pre-commit run --all-files` and `pre-commit run --all-files --hook-stage pre-push` — all green
- [x] `python3 .claude/lib/pre_commit_ci_sync.py .` and `python3 .claude/lib/memory_budget.py .` — both OK
- [x] Manual end-to-end run of `roster_union_sync.py` against real org data — reproduces exactly 16 unreconciled + 2 non-persona, matching #1181/#1183's measurement, no false positives

Closes #1181

Co-Authored-By: Wanjiku Mwangi <parametrization+Wanjiku.Mwangi@gmail.com>

### DIFF
--- a/.claude/lib/roster_union_sync.py
+++ b/.claude/lib/roster_union_sync.py
@@ -195,13 +195,16 @@ def parent_roster_names(repo_root: Path) -> set[str]:
 def local_card_names(repo_root: Path) -> set[str]:
     """Parse member names from the LOCAL repo's own `.claude/team/roster/*.md` cards.
 
-    Local-only by design (#1181): this script runs in CI with just the
-    `noorinalabs-main` checkout, so a child repo's card directory is never on
-    disk here — there is no card-fetch equivalent to `fetch_child_roster`.
-    That is not a gap for `compute_manifest_orphans` below: #1183 measured
-    card-only == 0 across the org (no persona has a card but no entry in that
-    child's own `roster.json`), so the already-fetched child rosters are a
-    complete substitute for child cards without a second network round-trip.
+    This is the PARENT-repo-only half of the #1181 card union: this script
+    runs in CI with just the `noorinalabs-main` checkout, so a child repo's
+    card directory is never on disk here — only a NETWORK fetch can see it.
+    `fetch_child_card_names` below is that fetch — it exists precisely
+    because #1183's "card-only == 0" assumption (no persona has a card but no
+    entry in that child's own `roster.json`) turned out to be false (see the
+    module docstring § "The reverse direction: manifest orphans" for the live
+    Mei-Lin Chang / Aisha Idrissi counterexamples). `compute_manifest_orphans`
+    is called with `local_card_names(repo_root)` unioned with every fetched
+    child's card set, never with this function's return value alone.
     """
     roster_dir = repo_root / ".claude" / "team" / "roster"
     names: set[str] = set()
--- a/.claude/lib/roster_union_sync.py
+++ b/.claude/lib/roster_union_sync.py
@@ -80,16 +80,41 @@ break its own commit identity (the annunaki-attack skill commits AS
 `Annunaki`), and `Steven French` is the owner's bare-principal mapping
 `verify_commit_identity.py` special-cases on purpose. The remaining
 persona-shaped orphans ARE an open reconciliation question (prune the row, or
-onboard the persona with a real card) and are what moves the exit code.
+onboard the persona with a real card) — but their disposition is explicitly
+OUT of #1181's scope (still undecided), so this check reports them loudly
+without gating on them; see "Exit codes" below for why.
 
-Same posture as the forward gate: continue-on-error, advisory, never a hard
-block. Unlike the forward direction, an INCOMPLETE fetch (any roster.json OR
-card-directory SKIPPED) makes the orphan report unreliable in the unsafe
-direction — a name backed only by something we failed to read would
-misreport as an orphan — so the orphan check does not run at all in that
-case, rather than risk a false "prune this" instruction. The card fetch pass
-only runs once the roster.json pass is already clean (no point fetching
-cards for a run that is already disqualified).
+An INCOMPLETE fetch (any roster.json OR card-directory SKIPPED) makes the
+orphan report unreliable in the unsafe direction — a name backed only by
+something we failed to read would misreport as an orphan — so the orphan
+check does not run at all in that case, rather than risk a false "prune
+this" instruction. The card fetch pass only runs once the roster.json pass
+is already clean (no point fetching cards for a run that is already
+disqualified).
+
+The orphan check NEVER moves the exit code (owner correction, PR #1240 review)
+=================================================================================
+The forward direction (#634) above predates this reverse check and keeps its
+established `continue-on-error` contract: the WORKFLOW run stays green even
+when it exits 1, because a single PR cannot deterministically reconcile a
+name a sibling repo just added. That contract does NOT extend cleanly to a
+non-zero exit from a check whose findings are individually out of THIS PR's
+scope to fix: `continue-on-error: true` only protects the workflow RUN's own
+conclusion — it does not change what the JOB itself reports into the PR's
+`statusCheckRollup`, which is exactly the surface `pr_ci_state.py` /
+`validate_pr_ci_status.py` (the org's actual merge-readiness oracle) reads.
+Measured: the reverse check returning exit 1 for the (already-known,
+disposition-still-open) 16 unreconciled names turned this job FAILING on
+every PR in the repo, not just this one — a permanently-red check that no
+amount of continue-on-error hides from the tooling that actually gates merges,
+which is precisely the state the org's "a red check is a stop, not a speed
+bump" rule exists to prevent (charter, owner directive). So the orphan branch
+below reports loudly (stderr, unabbreviated) but deliberately never sets
+`exit_code` — only a genuine forward-direction violation (#634: a child
+persona entirely missing from the parent) can fail this job. A future
+disposition of the 16 (§ above) could reasonably promote them to blocking;
+that is a separate, explicit decision, not a side effect of this check
+existing.
 
 Input Language
 ==============
@@ -99,10 +124,12 @@ CLI:  roster_union_sync.py [--repo-root <dir>] [--repos a,b,c] [--owner <org>]
 network). `--owner` defaults to `noorinalabs`.
 
 Exit codes (CLI):
-    0 — no drift AND no unreconciled manifest orphan (or every child fetch was
-        skipped, so neither check had enough data to assert anything)
-    1 — drift (a child persona is missing from the parent), OR at least one
-        persona-shaped manifest orphan has no backing card/child-roster entry
+    0 — no forward drift (#634). The reverse orphan report (#1181) is ALWAYS
+        advisory-only and never affects this — see "The orphan check NEVER
+        moves the exit code" above. Printed output still names every
+        unreconciled/non-persona orphan found; only the process exit stays 0.
+    1 — forward drift ONLY: a child persona positively observed in a child
+        `roster.json` is missing from the committed parent roster (#634)
     2 — usage / parent-roster load error
 """
 
@@ -469,16 +496,29 @@ def main(argv: list[str] | None = None) -> int:
             for name in orphans["non_persona"]:
                 print(f'  - "{name}"')
         if orphans["unreconciled"]:
+            # Advisory ONLY (owner correction, PR #1240 review): this must NOT
+            # move exit_code. continue-on-error at the WORKFLOW level does not
+            # make the JOB's conclusion advisory — `pr_ci_state.py` /
+            # `validate_pr_ci_status.py` (the org's actual merge-readiness
+            # oracle) read the per-check rollup, not the run conclusion, so a
+            # non-zero exit here would mark every future PR in this repo
+            # "failing" until the (explicitly out-of-scope-for-#1181)
+            # disposition of these names is resolved — exactly the
+            # permanently-red state the "red is a stop" rule exists to
+            # prevent. Reported loudly (stderr) precisely so it stays visible
+            # without blocking. Only the forward direction (#634) above,
+            # which predates this reverse check and has its own established
+            # contract, may set exit_code — see module docstring.
             print(
-                "\nMANIFEST ORPHAN (#1181): persona-shaped parent-roster entry with NO backing "
-                "card or child roster.json observed anywhere in the org. Pick one:\n"
+                "\nMANIFEST ORPHAN (#1181, advisory — does not fail this job): "
+                "persona-shaped parent-roster entry with NO backing card or child "
+                "roster.json observed anywhere in the org. Pick one:\n"
                 "  (a) onboard the persona — add a real roster card in the repo it belongs to, or\n"
                 "  (b) the persona is retired — remove the row from .claude/team/roster.json.",
                 file=sys.stderr,
             )
             for name in orphans["unreconciled"]:
                 print(f'  - "{name}"', file=sys.stderr)
-            exit_code = 1
         else:
             print(
                 "\nManifest-orphan check (#1181): every persona-shaped manifest entry is backed "
--- a/.claude/lib/roster_union_sync.py
+++ b/.claude/lib/roster_union_sync.py
@@ -39,6 +39,58 @@ CI token cannot read, or a transient API error contributes no names and is
 reported as SKIPPED — never a drift failure on its own. Drift is only ever
 asserted on names we positively observed in a child roster.
 
+The reverse direction: manifest orphans (#1181)
+================================================
+The gate above answers "is the parent roster MISSING anyone a child vouches
+for?" — one-directional. #1181 measured the org's actual state and found the
+other direction rotting silently: `roster.json` (78 entries) carries 18 names
+backed by NO card anywhere in the org (union of every repo's roster cards:
+60), and nothing ever prunes an entry once its card is deleted — an offboarded
+persona's manifest row outlives the persona indefinitely (#1181's "liveness"
+framing). Compounding it, #1178 (#1162) made the manifest a resolution
+authority for review-class scope slots, so an unbacked entry is not merely
+stale bookkeeping — it can resolve as a live reviewer.
+
+`compute_manifest_orphans` is the reverse assertion: manifest ⊆ (this repo's
+own roster cards) ∪ (every child's own roster CARD directory) ∪ (every
+successfully-fetched child `roster.json`). #1183 independently measured the
+same 18 names checking against child `roster.json`s directly instead of child
+cards and reported `card-only == 0` (no persona has a card but no entry in
+that child's own roster.json) — but that measurement itself drifted: a live
+re-check while building this gate found `noorinalabs-isnad-graph/roster/
+data_scientist_mei.md` (Mei-Lin Chang) absent from that SAME repo's own
+`roster.json`, and `noorinalabs-deploy` has NO `roster.json` at all (its
+`sre_engineer_aisha.md` persona is invisible to every roster.json-only check).
+Trusting `card-only == 0` as a standing invariant would have shipped an
+orphan check that immediately false-positives on two real, carded people, so
+`fetch_child_card_names` fetches every child's card directory directly (one
+`gh api` directory listing + one fetch per `.md` file) rather than relying on
+that assumption staying true. See `DEFAULT_CARD_REPOS` for why the repo list
+here is NOT the same as `DEFAULT_CHILD_REPOS`.
+
+Two of the 18 are not a reconciliation question at all — `Annunaki` (the
+error-monitor's own commit identity) and `Steven French` (the bare
+`parametrization@gmail.com` principal, no `+alias`) are DEFINITIONALLY not
+personas, matched by `_PERSONA_ALIAS_RE` the same way
+`.claude/hooks/validate_pr_review.py` and
+`.claude/skills/wave-scope/validate_matrix_names.py` already exclude them from
+review-class resolution (#1179/#1181). Those two are reported separately, as
+information, not as something to prune or onboard: pruning `Annunaki` would
+break its own commit identity (the annunaki-attack skill commits AS
+`Annunaki`), and `Steven French` is the owner's bare-principal mapping
+`verify_commit_identity.py` special-cases on purpose. The remaining
+persona-shaped orphans ARE an open reconciliation question (prune the row, or
+onboard the persona with a real card) and are what moves the exit code.
+
+Same posture as the forward gate: continue-on-error, advisory, never a hard
+block. Unlike the forward direction, an INCOMPLETE fetch (any roster.json OR
+card-directory SKIPPED) makes the orphan report unreliable in the unsafe
+direction — a name backed only by something we failed to read would
+misreport as an orphan — so the orphan check does not run at all in that
+case, rather than risk a false "prune this" instruction. The card fetch pass
+only runs once the roster.json pass is already clean (no point fetching
+cards for a run that is already disqualified).
+
 Input Language
 ==============
 CLI:  roster_union_sync.py [--repo-root <dir>] [--repos a,b,c] [--owner <org>]
@@ -47,9 +99,10 @@ CLI:  roster_union_sync.py [--repo-root <dir>] [--repos a,b,c] [--owner <org>]
 network). `--owner` defaults to `noorinalabs`.
 
 Exit codes (CLI):
-    0 — no drift (parent roster covers every observed child persona), or every
-        child fetch was skipped (nothing positively observed to compare against)
-    1 — drift: at least one observed child persona is missing from the parent
+    0 — no drift AND no unreconciled manifest orphan (or every child fetch was
+        skipped, so neither check had enough data to assert anything)
+    1 — drift (a child persona is missing from the parent), OR at least one
+        persona-shaped manifest orphan has no backing card/child-roster entry
     2 — usage / parent-roster load error
 """
 
@@ -58,6 +111,7 @@ from __future__ import annotations
 import argparse
 import base64
 import json
+import re
 import subprocess
 import sys
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


parametrization:
## Review — PR #1240 (head `dc07627f`)

Reviewed as routine reviewer, per Bereket's request. Verified the fix commit end-to-end (exercised, not just read) plus the author's three claims. Everything checked out; no findings.

### 1. The exit-code fix (`dc07627f`) — confirmed real and complete

Read the fix commit's diff directly (`gh api .../commits/dc07627f...`): the prior head set `exit_code = 1` inside the orphan branch; `dc07627f` deletes that line and adds the "NEVER moves the exit code" docstring section. Not inferred — exercised both directions by importing `roster_union_sync.py` and monkeypatching `fetch_child_roster`/`fetch_child_card_names`/`parent_roster_map`:

- Orphan present, no forward drift → `main()` returns **0** (still prints `MANIFEST ORPHAN` to stderr).
- Forward drift present (a child persona missing from parent), orphan also present → returns **1**, driven by the forward finding.
- A skipped/failed child fetch → orphan section prints an explicit `SKIPPED` line, never the "No unreconciled orphans" clean message — a fetch failure cannot be misread as "no drift" (both messages are textually distinct, confirmed by running it).

Found and ran the pinning test the request asked for: `test_forward_drift_exits_1_even_with_advisory_orphan_drift` (`.claude/lib/tests/test_roster_union_sync.py`) constructs a run with a genuine orphan AND a genuine forward violation together and asserts `rc == 1` while both `"DRIFT"`/`"MANIFEST ORPHAN"` markers are present — non-vacuous, it actually distinguishes the two directions rather than tautologically passing. Ran the full local test suite (copied the 4 changed files into a mirrored `lib/` + `skills/wave-scope/` tree, `PYTHONDONTWRITEBYTECODE=1 python3 -m unittest discover`): **32/32** pass in `test_roster_union_sync.py`, **45/45** pass in `test_validate_matrix_names.py`.

Docstring says "Non-blocking by design" (module header) and "The orphan check NEVER moves the exit code" (new section) — matches the implementation exactly; `main()`'s only `exit_code = 1` assignment is in the forward-direction (#634) branch.

Confirmed the CI job itself is unchanged and matches the docstring's `commit-identity.yml:56` citation: fetched the workflow at this head, `roster-union-sync` job still carries `continue-on-error: true` at line 56. `gh pr checks 1240` shows 14/14 passing including `Roster union sync-drift gate (parent ⊇ child personas)`.

### 2. Persona-alias filter (`_load_org_manifest_names`) — confirmed, no legitimate persona lost

Ran `_load_org_manifest_names` against this repo's real, current `.claude/team/roster.json` (78 entries): returns 76 names, excluding exactly `Annunaki` and `Steven French` — nothing else. `.claude/skills/wave-scope/tests/test_validate_matrix_names.py::PersonaShapeFilterTests` covers the same ground end-to-end through `validate()`: `Annunaki`/`Steven French` fail to resolve as a `reviewer`, while `Amara Diallo` (one of the other 16 persona-shaped-but-uncarded names) still resolves — the filter narrows exactly the two definitionally-wrong entries, not every uncarded one. All pass.

### 3. The two "already-drifted card-only" instances — confirmed live

Queried the live repos directly (not from the PR's description):
- `noorinalabs-isnad-graph/.claude/team/roster/data_scientist_mei.md` names Mei-Lin Chang; her name is genuinely absent from that same repo's `.claude/team/roster.json` (fetched and enumerated the 15 keys — she's not in it).
- `noorinalabs-deploy/.claude/team/roster/sre_engineer_aisha.md` names Aisha Idrissi; `noorinalabs-deploy/.claude/team/roster.json` returns a 404 — the repo has no aggregated roster.json at all, confirming she'd be invisible to any roster.json-only reverse check.

Both real; `fetch_child_card_names`/`DEFAULT_CARD_REPOS` closing this is a genuine fix, not a hypothetical.

### Nothing I could not verify

All claims in the brief and the PR description were independently exercised against either live org data or the actual test/CLI behavior at `dc07627f`. No gaps.

---
Requestor: Bereket Tadesse
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Merge-gate review — head `dc07627f`, reviewed via `gh api contents/<path>?ref=dc07627f` + a full-depth clone at that SHA. CI 14/14 SUCCESS in `statusCheckRollup` at head, including `Roster union sync-drift gate`.**

The rework is the right fix and the shipped behaviour is correct — I verified both directions against the real engine, not the string. But two of the four new CLI-wiring guards **cannot fail**, including both of the ones the PR description advertises as the Mei-Lin / Aisha false-positive regression guards. That is a mutation-proven finding, not a stylistic one, and the fix is about four lines. Returning it for that.

## What I verified (all live, not asserted)

**The advisory/blocking distinction holds.** Mutation-tested the exit-code contract, `PYTHONDONTWRITEBYTECODE=1 ENVIRONMENT=test python3 -m pytest .claude/lib/tests/test_roster_union_sync.py .claude/skills/wave-scope/tests/ -q` (baseline 77 passed):

- orphan branch made to set `exit_code = 1` → **2 failed** (killed)
- forward-drift `exit_code = 1` removed → **2 failed** (killed)
- `_load_org_manifest_names` persona filter removed → **3 failed** (killed)
- `compute_manifest_orphans` buckets swapped → **6 failed** (killed)
- `compute_manifest_orphans` ignores `card_names` → **2 failed** (killed)

So the specific defect the first head shipped is genuinely pinned. Good.

**A real forward violation still exits non-zero.** I removed `Imelda Santos` and `Nikolaos Papadopoulos` from a copy of the committed parent roster and ran the real script against real fetched child data:

```
DRIFT: the committed parent roster .claude/team/roster.json is MISSING child-repo persona(s) below.
  - "Imelda Santos"  (canonical in: noorinalabs-isnad-ingest-platform)
  - "Nikolaos Papadopoulos"  (canonical in: noorinalabs-data-acquisition)
EXIT=1
```

**Advisory drift exits 0 on real data.** Unmodified live run: exit 0, 16 unreconciled + 2 non-persona, matching #1181's measurement exactly. No false positives.

**The `card-only == 0` catch is real.** Independently confirmed both instances: `noorinalabs-isnad-graph`'s `roster.json` has 15 entries but 16 cards, and `data_scientist_mei.md` names `Mei-Lin Chang` who is absent from that repo's own `roster.json`; `noorinalabs-deploy` returns 404 for `roster.json` entirely while `sre_engineer_aisha.md` names `Aisha Idrissi`. A roster.json-only check would have false-positived on both. Good catch, and the right call to fetch cards directly rather than trust #1183's measurement as a standing invariant.

**No legitimate persona stopped resolving.** Ran `_PERSONA_ALIAS_RE` over the real 78-entry `roster.json` at head: exactly 2 entries filtered, `Steven French` → `parametrization@gmail.com` and `Annunaki` → `parametrization+Annunaki@gmail.com`. 76 still resolve. Nothing else narrowed.

**No conflict with #1223.** `roster.json` carries both `Cedric Novak` and `Cédric Novák` mapping to the *same* address; the only card is `noorinalabs-landing-page/ux_designer_cedric.md`, which names `Cédric Novák`. The orphan report flags the ASCII `Cedric Novak` and leaves the folded one backed — consistent with #1223's finding, and it independently corroborates it. Nothing here presumes surname folding.

**Merge-order is clean.** #1238 (`org_repos` SSOT) merged *after* this PR's base `3a688242` and also touches `roster_union_sync.py`. I trial-merged `origin/main` into head: no conflict, #1238's dynamic `DEFAULT_CHILD_REPOS` survives, `DEFAULT_CARD_REPOS` composes correctly, and the full suite passes post-merge (1120 passed, 19 subtests).

## Must-fix

**1. The Mei-Lin / Aisha guards and both incomplete-fetch skip tests are inert.**

Two mutations survive the entire suite:

- `main()` drops the child-card union (`for names in child_cards.values(): card_names |= names` → `pass`) → **77 passed**
- the incomplete-fetch skip guard removed (`if skipped or card_skipped:` → `if False:`) → **77 passed**

Both still survive against the post-merge tree (32 passed). The first mutation deletes the entire reason `fetch_child_card_names` exists and CI stays green — meaning a future "simplification" would silently restore the check's instruction to prune two real, carded people, which is the exact harm these tests were written to prevent.

The cause is mechanical and is a second-order consequence of the (correct) rework: `test_card_only_child_persona_is_not_a_false_positive_orphan`, `test_deploy_cards_seen_even_though_deploy_has_no_roster_json`, `test_orphan_check_skipped_when_card_fetch_incomplete` and `test_orphan_check_skipped_when_any_child_fetch_fails` assert only `assertEqual(rc, 0)`. Once the orphan check may never move the exit code, `rc` is 0 under *every* orphan outcome, so no fixture can trip those assertions. The observable moved to the streams and the tests did not follow.

This answers the "does a network failure read as no-drift?" question precisely: at the exit code, **skip and pass are identical by construction**. That is the safe direction for the gate, but it means the skip behaviour must be pinned on the streams or it is not pinned at all.

Fix — have `_run` also return stdout/stderr and assert on them. I wrote this and confirmed it passes at head and kills both mutations:

```python
self.assertNotIn("Mei-Lin Chang", err)
self.assertIn("No unreconciled orphans", out)          # kills the card-union mutation
```
```python
self.assertIn("Manifest-orphan check (#1181) SKIPPED", out)
self.assertNotIn("MANIFEST ORPHAN", err)               # kills the skip-guard mutation
```

Result: passes unmutated (2 passed); card-union mutation → 1 failed; skip-guard mutation → 1 failed.

**2. `local_card_names`'s docstring asserts the premise this PR exists to refute.**

Added in this same commit, it says the script runs with only the `noorinalabs-main` checkout so "there is no card-fetch equivalent to `fetch_child_roster`", and that "#1183 measured card-only == 0 across the org ... so the already-fetched child rosters are a complete substitute for child cards without a second network round-trip."

`fetch_child_card_names` is defined about a hundred lines below it, and the module docstring above it explains at length that `card-only == 0` has drifted and that trusting it "would have shipped an orphan check that immediately false-positives on two real, carded people." Two contradictory rationales for the same decision are discoverable side by side in one file, and the wrong one is the one sitting on the function a maintainer would read first. Combined with must-fix 1, a maintainer who acts on it removes the card fetch and CI agrees with them.

Nit, same commit: the new `#1181` section in `validate_matrix_names.py`'s module docstring was inserted mid-paragraph, orphaning "Pre-existing since #319, but #1134 made § 12.5 mandatory..." so it now trails the persona-filter section and reads as a claim about #1181. It belongs with the #1162 section above.

## On whether an advisory check that can never fail is worth having

Worth having, but not on this channel. The exit-code decision is right and I would not reverse it — blocking on the 16 recreates the permanently-red state. But the report's only surface is now text in the log of a green job, and a fetch failure suppresses it with no signal anywhere else. Its two streams also interleave out of order under a pipe (stderr unbuffered, stdout block-buffered), so the `MANIFEST ORPHAN` block prints *above* the `OK`/`SKIPPED` lines it should be read against — I saw this on the live run. Filed as #1254 with concrete options (`$GITHUB_STEP_SUMMARY`, `::notice::`, or folding the count into the status digest), plus the requirement that "did not run" must not render the same as "clean". Not blocking this PR.

## Debt filed

- #1253 — `card_repos` is built with a positional slice `DEFAULT_CARD_REPOS[len(DEFAULT_CHILD_REPOS):]`. Correct only while `DEFAULT_CARD_REPOS` is a strict suffix-extension; I confirmed a non-suffix entry is silently never card-fetched, which would then misreport that repo's every persona as an orphan. More exposed post-#1238 now that the base tuple is computed.
- #1254 — the advisory report has no reader, and its streams interleave.

Neither blocks merge. Once must-fix 1 and 2 land as an additive commit on this branch (no force-push per `reviews.md` § the HEAD-SHA anchor), re-request and I will re-verify against the new SHA.

---

Requestor: Nadia Khoury
Requestee: Wanjiku Mwangi
TechDebt: #1253, #1254

parametrization:
## Re-review — PR #1240, new head `ecd536a2`

Prior verdict invalidated per #950 (any push resets prior approvals). Re-cast against `ecd536a2`. Delta from the previously-reviewed `dc07627f` is exactly what the sent-back note claims — confirmed by diffing the fetched file contents at both SHAs directly: `.claude/skills/wave-scope/validate_matrix_names.py` and its test file are byte-identical between the two heads; only `.claude/lib/roster_union_sync.py` and `.claude/lib/tests/test_roster_union_sync.py` changed. My prior findings on the persona-alias filter and the two live card-only instances (Mei-Lin Chang / Aisha Idrissi) stand unchanged and are carried forward without re-doing that work.

### The two must-fixes — verified by actually re-introducing each mutation, not by reading the diff

**1. Inert guards.** Confirmed both were genuinely inert at `dc07627f` (rc-only assertions can't distinguish an advisory check's outcomes once its exit code is constant by design) and are genuinely fixed at `ecd536a2`. Reproduced directly:

- Applied Nadia's mutation (a) — dropped the `for names in child_cards.values(): card_names |= names` union in `main()` so only local cards are consulted — against the `ecd536a2` test file. Result: **2 tests fail**, exactly `test_card_only_child_persona_is_not_a_false_positive_orphan` and `test_deploy_cards_seen_even_though_deploy_has_no_roster_json`, both on `assertNotIn("Mei-Lin Chang"/"Aisha Idrissi", stderr...)` — the mutant now reports them as unreconciled orphans and the new stream assertion catches it.
- Applied Nadia's mutation (b) — removed the `if skipped or card_skipped:` guard so the orphan check always runs — against the same test file. Result: **2 tests fail**, `test_orphan_check_skipped_when_card_fetch_incomplete` and `test_orphan_check_skipped_when_any_child_fetch_fails`, both on the new `assertIn("Manifest-orphan check (#1181) SKIPPED", stdout...)` assertion, which the mutant no longer prints.
- The unmutated `ecd536a2` test file: 32/32 pass.

Both guards are now genuinely load-bearing — the stream assertions (`assertIn`/`assertNotIn` against captured stdout/stderr, not just `rc`) are exactly what closes the gap, matching the general point in the sent-back note: an advisory check whose exit code is constant by design can only be pinned on its output.

**2. Docstring.** `local_card_names`'s docstring no longer claims "no card-fetch equivalent to `fetch_child_roster`" / child rosters as "a complete substitute for child cards" — both false as of the same diff that adds `fetch_child_card_names` ~100 lines below. Read the new text: it now correctly states `fetch_child_card_names` *is* that network-fetch equivalent, points at the module docstring's Mei-Lin Chang / Aisha Idrissi counterexamples as why the old assumption doesn't hold, and clarifies `compute_manifest_orphans` is called with the union of local + fetched child cards, never `local_card_names` alone. The remaining `card-only == 0` mentions elsewhere in the module docstring are historical narrative ("#1183 measured X, which later turned out false") and are consistent, not contradicted.

### CI at this head

Re-derived, not assumed from the ping: `gh pr view 1240 --json statusCheckRollup` — **14/14 SUCCESS**, including `Pytest — hooks and lib test suites` (was pending when the ping was sent, since resolved) and `Roster union sync-drift gate (parent ⊇ child personas)`.

### Nothing I could not verify

Everything above was exercised directly (mutation re-injection + suite run, live diff against the prior head, fresh CI query) rather than inferred from the PR description.

---
Requestor: Bereket Tadesse
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Merge-gate re-verification — head `ecd536a2` (was `dc07627f`). CI 14/14 SUCCESS in `statusCheckRollup` re-derived at this SHA. Both must-fixes hold under mutation. Approving.**

Confirmed the head move is a single additive commit on top of `dc07627f`, no rebase (`git log dc07627f..ecd536a2` is exactly one commit, and `dc07627f` is still an ancestor). The HEAD-SHA anchor chain from the prior review is intact, which is what makes the carry-forward below legitimate.

## Must-fix 1 — the two inert guards: fixed

Re-ran the exact mutations that survived at `dc07627f`, same command, same baseline (77 passed):

- `main()` drops the child-card union (`for names in child_cards.values(): card_names |= names` → `pass`) — was **77 passed**, now **2 failed, 75 passed**
- incomplete-fetch skip guard removed (`if skipped or card_skipped:` → `if False:`) — was **77 passed**, now **2 failed, 75 passed**

Both killed. I also re-ran them against the merge result rather than the head alone: **2 failed, 30 passed** each.

The assertions are real, not cosmetic. The two skip tests pin the distinction in *both* directions — `assertIn("Manifest-orphan check (#1181) SKIPPED", stdout)` proves the guard fired, and `assertNotIn("Persona X", stdout + stderr)` / `assertNotIn("Amara Diallo", ...)` proves the name that *would* have been reported had the check actually run never appears. That is what separates "skipped" from "ran and happened to find nothing", which was the specific failure I flagged. The Mei-Lin and Aisha tests do the mirror of it: `assertNotIn(name, stderr)` plus `assertIn("No unreconciled orphans", stdout)`, so a silently-not-running card union fails on the positive assertion rather than passing by absence.

Two further probes to check the fix is not narrowly fitted to the mutations I published:

- skip guard weakened to roster.json-only (`if skipped or card_skipped:` → `if skipped:`) — **1 failed, 76 passed** (killed)
- exit-code contract re-checked: orphan branch setting `exit_code = 1` → **2 failed**; forward-drift `exit_code` removed → **2 failed**. Still pinned after the test rewrite.

## Must-fix 2 — the docstring: fixed, and what replaced it is true

The two false sentences are gone. I checked each clause of the replacement against this commit rather than reading it for tone:

- "only a NETWORK fetch can see it" — correct; the job checks out `noorinalabs-main` alone.
- "`fetch_child_card_names` below is that fetch — it exists precisely because #1183's 'card-only == 0' assumption ... turned out to be false" — now agrees with the module docstring instead of contradicting it.
- "`compute_manifest_orphans` is called with `local_card_names(repo_root)` unioned with every fetched child's card set, never with this function's return value alone" — verified against `main`, which does exactly that before the call.

The `validate_matrix_names.py` docstring nit (the orphaned "Pre-existing since #319..." sentence) was not addressed and I am not holding the PR for it; it is cosmetic prose placement, and this commit deliberately did not touch that file.

## Carried forward, not re-measured

The production delta in this commit is **docstring-only** — `git diff dc07627f ecd536a2 -- .claude/lib/roster_union_sync.py` is +10/−7 entirely inside `local_card_names`'s docstring, with no executable line changed; the rest of the commit is the test file. On that basis I carry forward from `dc07627f` without re-asserting them as fresh measurements: the `_PERSONA_ALIAS_RE` sweep (exactly 2 of 78 filtered, 76 still resolve), the #1223 non-conflict, and the verified Mei-Lin Chang / Aisha Idrissi claims. Those rest on the persona-filter code and on live org data, neither of which this commit touches.

I did re-run the live end-to-end check, because the merge result composes this PR with #1238's `DEFAULT_CHILD_REPOS` change and no CI job exercises that combination against real data: the merge result reports 18 names (16 unreconciled + 2 non-persona) at exit 0, matching `dc07627f`.

## Merge-order re-derived at current main

`origin/main` is now `b4db5d4`; #1201 (`3175f63`) and #1230 (`b4db5d4`) are both ancestors. Trial-merged `origin/main` into `ecd536a2`: **clean, no conflict**. Neither sibling touches this PR's files — the only thing main contributes to `roster_union_sync.py` is #1238's `org_repos` SSOT migration, which survives the merge intact (`DEFAULT_CHILD_REPOS` 6 repos from `org_repos.CHILD_REPOS`, `DEFAULT_CARD_REPOS` 7, slice suffix `('noorinalabs-deploy',)`). Full suite on the merge result: **3554 passed, 647 subtests passed**. `mergeable: MERGEABLE`.

## One new finding, filed not blocked

`fetch_child_card_names`'s final `return names if names else None` guards the case "listing returned `.md` files but zero names parsed from any of them" — every per-file fetch raised, or a future card-template change stops matching both name regexes. It returns `None` (SKIPPED, safe) today; I verified that live. But no test reaches that line — `test_empty_directory_returns_none` returns at the earlier `if not filenames` guard — and mutating it to `return names` leaves the suite green (77 passed). Weakened, it would turn a skip into a successful-fetch-with-zero-names, and every persona carded only in that repo would be reported as an orphan.

Filed as #1262. I am **not** blocking on it: the behaviour is correct as written, it is a pre-existing gap from `dc07627f` that I did not probe in the first round rather than anything this commit introduced, and both must-fixes I did raise were fixed exactly as specified. Returning a PR a second time for an adjacent branch I found later would be moving the goalposts.

## Deferred work untouched

#1253 (positional slice in `card_repos`) and #1254 (the advisory report has no reader; streams interleave) remain open and are neither closed nor contradicted by this commit — the slice is unchanged and still composes correctly post-merge, and the report's channel is unchanged. Both stay wave-30 alongside #1262.

Good, disciplined fix. The stream assertions are the right instrument and the docstring now tells a maintainer the truth.

---

Requestor: Nadia Khoury
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: #1253, #1254, #1262

PR #1248, opened 2026-08-03
tech-debt(wave_status): close the #1190 reconciliation warning's own blind spots

### LINKED ISSUE #1201: tech-debt(wave_status): the #1190 reconciliation warning is silent at its own edges — unparseable scope rows shrink the denominator, and a full 100-node closingIssuesReferences page truncates with no signal
Found during merge-gate review of #1198 (the fix for #1189/#1190), at head SHA `d7024bf74119f764051c3c2e184363852097cecd`.

#1190's premise, in its own words, is that "the instrument should say what it could not see." The warning #1198 delivers does that well for rows it *parsed* and references it *fetched*. It is silent at both of those edges.

## Edge 1 — an unparseable scope row vanishes from the numerator AND the denominator

`_canonical_issue_numbers_by_repo` deliberately skips a tier row that is a legacy plain string, or a dict whose `id` has no `#` / a non-numeric tail (`test_legacy_string_tier_rows_skipped_not_raised` pins this tolerance). `_reconciliation_warning_from_claims` then computes `unclaimed = canonical - claimed` and renders `({len(unclaimed)} of {len(canonical)})`. Rows that never parsed are in neither set, so they are reported nowhere.

Reproduced against the #1198 head SHA — a `wave_29_scope` declaring **4** tier rows, of which one is a legacy string (`"main#1114"`) and one a malformed id (`{"id": "main-1116"}`):

```
scope declares 4 rows; 2 are unparseable
rc = 0
stderr = WARNING: scope rows with no matching merged PR: main#1160 (1 of 2)
```

The operator reads "1 of 2" against a record that says 4. The two dropped rows are exactly the ones nobody is tracking.

Degenerate case: a scope where *every* row is unparseable yields `canonical == set()`, `_reconciliation_warning_from_claims` returns `None` at its `if not canonical` guard, every repo is skipped, and `counters` exits 0 printing `final_pr_count: 0` with **no warning on stderr at all** — a silent zero that reads as "nothing to report."

Currently latent: live wave-29 parses 36 of 36 declared rows.

## Edge 2 — a full `closingIssuesReferences` page truncates with no signal

`first:100` is genuinely GitHub's ceiling for this connection — confirmed against the real engine, `first:101` returns `EXCESSIVE_PAGINATION` ("exceeds the `first` limit of 100 records") and `gh` exits 1. So the raise from `first:25` is the maximum a single page can do, and the docstring correctly concedes that beyond it real `endCursor`/`hasNextPage` pagination is required.

What is missing is the *signal*. `_pr_closing_issue_numbers` discards the page boundary: a PR closing exactly 100+ issues returns 100 nodes and the caller cannot distinguish "closed 100" from "closed 100 and we truncated." Under the same principle #1190 establishes, a full page should say so rather than be silently trusted. This is the one line the cap raise did not buy.

## Suggested fix

- Carry the unparseable-row count out of `_canonical_issue_numbers_by_repo` (or re-count declared `tier_*` rows in the warning path) and append it, e.g. `… (24 of 36; 2 scope rows unparseable)`. Emit the warning even when `canonical` is empty but declared rows are not.
- In `_pr_closing_issue_numbers`, warn to stderr when `len(nodes) == 100` — "PR <repo>#<n> closing references may be truncated at the 100-node page cap." One comparison, and it converts a documented deferral into a loud one.

parametrization:
## Summary

The main#1190 reconciliation warning was silent at three of its own edges (found during merge-gate review of #1198):

- **Edge 1** — an unparseable `wave_{M}_scope` tier row (a legacy plain string, or a dict whose `id` has no `#` / a non-numeric tail) vanished from both the numerator AND the denominator of the "N of M" line. Degenerate case: if EVERY row is unparseable, `canonical == set()` and the warning returned `None` — `counters` exited 0 printing `final_pr_count: 0` with no warning on stderr at all.
- **Edge 2** — `_pr_closing_issue_numbers`'s `first:100` GraphQL page (GitHub's real ceiling for `closingIssuesReferences`) discarded the page-boundary signal: a PR closing exactly 100 issues was indistinguishable from a PR whose closing references were truncated at the cap.
- **Edge 3** — a canonical scope row naming a repo absent from `wave_{M}_repos_in_scope` can never be claimed (that repo's merged PRs are never listed), so it reads as unclaimed on every single run regardless of whether it was actually delivered — the operator cannot tell "not delivered yet" from "structurally unreachable by this instrument".

## Changes

- `_canonical_issue_numbers_by_repo` / `_canonical_pairs` now return `(pairs, unparseable_count)` instead of just the parsed set.
- `_reconciliation_warning_from_claims` reports against the full declared row count (`parsed + unparseable`), names the unparseable-row count explicitly, fires even when nothing at all parsed (no more silent `None` in the degenerate case), and flags any unclaimed row whose repo is absent from `repos_in_scope` with `" (repo not in repos_in_scope)"`.
- `_pr_closing_issue_numbers` prints a `WARNING: PR <repo>#<n> closing references may be truncated at the 100-node page cap.` to stderr at the point of detection (`len(nodes) == 100`), not merely returned-and-possibly-discarded.

## Test plan

- [x] `PYTHONDONTWRITEBYTECODE=1 ENVIRONMENT=test python3 -m pytest .claude/lib/tests/test_wave_status.py -q` — 44 passed (up from 36; 8 new/modified assertions cover all 3 edges).
- [x] Mutation-proven: reverted `wave_status.py` to the pre-fix `origin/main` copy with the new tests still in place — all 8 new/modified tests failed as expected (`8 failed, 36 passed`), then restored the fix and re-ran green.
- [x] Two of the new tests exercise the ACTUAL call path (`merged_prs()` / `wave_status.main(["counters", ...])`) rather than asserting a private helper's own stderr write in isolation, per the "silent warning" bug class this issue belongs to — a return-value assertion while the caller discards it would reproduce the exact bug being fixed.
- [x] `PYTHONDONTWRITEBYTECODE=1 ENVIRONMENT=test python3 -m pytest .claude/lib/tests .claude/hooks/tests -q` — 3455 passed, 647 subtests passed (no regressions elsewhere).
- [x] `pre-commit run --all-files` and `pre-commit run --all-files --hook-stage pre-push` — all green (ruff-format/lint, actionlint, cspell, mypy, pytest, pytest-skills, memory-budget, headcount-budget, doc-freshness, office-drift, mermaid-render).
- [x] `python3 .claude/lib/pre_commit_ci_sync.py .` — `OK: pre-commit config mirrors all CI-enforced checks.`

Closes #1201

### DIFF
--- a/.claude/lib/tests/test_wave_status.py
+++ b/.claude/lib/tests/test_wave_status.py
@@ -548,8 +548,9 @@ class CanonicalIssueNumbers(unittest.TestCase):
                     "tier_99_never_seen_before": [9999],
                 },
             )
-            by_repo = wave_status._canonical_issue_numbers_by_repo("29", status)
+            by_repo, unparseable = wave_status._canonical_issue_numbers_by_repo("29", status)
         self.assertEqual(by_repo["noorinalabs-main"], {1114, 1160, 1162, 9999})
+        self.assertEqual(unparseable, 0)
 
     def test_legacy_string_tier_rows_skipped_not_raised(self) -> None:
         with TemporaryDirectory() as td:
@@ -560,8 +561,32 @@ class CanonicalIssueNumbers(unittest.TestCase):
                 }
             }
             status.write_text(json.dumps(data))
-            by_repo = wave_status._canonical_issue_numbers_by_repo("29", status)
+            by_repo, unparseable = wave_status._canonical_issue_numbers_by_repo("29", status)
         self.assertEqual(by_repo, {"noorinalabs-main": {1114}})
+        self.assertEqual(unparseable, 1)  # "main#322" -- the legacy plain-string row
+
+    def test_unparseable_rows_are_counted_not_just_skipped(self) -> None:
+        """main#1201 Edge 1: a legacy plain-string row and a dict row with a
+        malformed `id` (no `#`, or a non-numeric tail) both vanish from
+        `by_repo` -- but their COUNT must now be carried out alongside it, or
+        the reconciliation warning built on top has no way to know they ever
+        existed."""
+        with TemporaryDirectory() as td:
+            status = Path(td) / "cross-repo-status.json"
+            data = {
+                "wave_29_scope": {
+                    "tier_1_backlog": [
+                        "main#322",  # legacy plain string
+                        {"id": "main-1116"},  # malformed: no "#"
+                        {"id": "noorinalabs-main#abc"},  # malformed: non-numeric tail
+                        {"id": "noorinalabs-main#1160"},  # parses fine
+                    ],
+                }
+            }
+            status.write_text(json.dumps(data))
+            by_repo, unparseable = wave_status._canonical_issue_numbers_by_repo("29", status)
+        self.assertEqual(by_repo, {"noorinalabs-main": {1160}})
+        self.assertEqual(unparseable, 3)
 
 
 class MergedPrsDirectToMain(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


parametrization:
Reviewed against the real issue (#1201 body + the Edge 3 follow-up comment from re-review of #1198), not the paraphrase in the original spawn brief. Fetched the diff, cloned full-depth, checked out `34fb1714`, and independently verified every claim in the PR description rather than trusting it.

**Verified:**
- All three edges in #1201 (unparseable-row denominator, 100-node page-cap signal, repo-absent-from-scope flagging) are implemented and match the issue's suggested-fix shape, including the Edge 1 degenerate all-unparseable case and the Edge 3 "don't mislabel an ordinary not-yet-delivered row" requirement.
- `_canonical_issue_numbers_by_repo` / `_canonical_pairs` now return `(pairs, unparseable_count)`. Grepped the whole repo (not just this file) for both names — the only callers are `_merged_prs_direct_to_main` and `_reconciliation_warning_from_claims`, both updated to unpack the tuple; no other importer touches these underscore-private helpers.
- Traced `_reconciliation_warning_from_claims` by hand through all branches: fully-parsed-and-claimed (unchanged `None`), some-unparseable-but-rest-claimed (now warns with "0 scope rows unclaimed (of N declared); M scope rows unparseable"), and the degenerate all-unparseable case (now warns instead of the old silent `None`/`final_pr_count: 0`).
- Edge 3: `repos_in_scope` check correctly flags only unclaimed rows whose repo is absent from scope, confirmed it does not tag an in-scope not-yet-delivered row (`test_edge3_...` plus manual trace).
- `_pr_closing_issue_numbers`'s new warning fires only at exactly `len(nodes) == 100` (confirmed 99 does not warn) and is worded "may be truncated" — correctly hedged as a heuristic, not an overclaim. This does mean a genuine 100-closer PR will warn on every run with no way to disambiguate without real pagination; that's the tradeoff the issue explicitly asked for ("the one line the cap raise did not buy"), not a defect.
- Confirmed independently (not just re-reading the PR body) that `_cmd_counters` and `_cmd_merged_prs` already printed the reconciliation warning to stderr before this PR — this PR only extends what the warning says, it doesn't add the stderr wiring. The spawn brief's claim that the warning was discarded was wrong, as the author noted.
- **Mutation kill, re-run myself:** checked out the PR branch, reverted only `wave_status.py` to the pre-fix merge-base (`36d16fa8`, i.e. before #1201's fix, after #1200) while keeping the new tests, ran with `PYTHONDONTWRITEBYTECODE=1 ENVIRONMENT=test` — got exactly **8 failed, 36 passed**, matching the author's claim. Restored the fix: 44/44 green.
- Full suite: `.claude/lib/tests .claude/hooks/tests` → 3483 passed, 647 subtests (author reported 3455; difference is main having moved since the PR was opened, not a regression — no failures either way).
- `ruff check`, `mypy`, and `pre_commit_ci_sync.py` all clean on the PR head.
- CI: 14/14 green. `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN` (re-checked directly; the orchestrator's earlier snapshot showing `UNKNOWN` was a transient GitHub compute state).

Nothing I could not verify. No must-fix findings, no tech debt to file.

---
Requestor: Wanjiku Mwangi
Requestee: Santiago Ferreira
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Merge-gate review — Approved.** Reviewed against the real #1201 (body + the Edge 3 follow-up comment from the #1198 re-review), not the paraphrase in my spawn brief. Everything below was run against a full-depth clone detached at `34fb1714` and against the real GitHub engine; the parent working tree was never checked out.

The question I was asked to answer is not "does it crash" but "does the denominator still mean what an operator will read it to mean." Short answer: yes, and on every record that actually exists the line is byte-identical to before.

## Signature change on a shared helper — contained

`_canonical_issue_numbers_by_repo` and `_canonical_pairs` now return `(pairs, unparseable)`. A missed caller would get an always-truthy 2-tuple of the wrong length, so I enumerated them repo-wide rather than in-file (`rg --hidden --no-ignore`, excluding `.git`, worktrees, `ontology/structural/`, and the annunaki log — note ripgrep skips `.claude/` by default without `--hidden`, which is how this could be under-counted):

- The only call sites of either symbol anywhere in the repo are `_merged_prs_direct_to_main`, `_canonical_pairs` itself, and `_reconciliation_warning_from_claims`. All three unpack the tuple.
- The 15 other Python files that touch `wave_status` use only `merged_prs`, `compute_counters`, `read_repos`, `reconciliation_warning`, `build_digest`, `main`, `_run_gh`, `_merged_prs_wave_branch`, `_DIGEST_POINTER_KEYS` — no signature in that set changed.

## Does the warning still fire where it fired before — verified on all real data

I swept **every** `wave_*_scope` recorded in `cross-repo-status.json`, not just wave-29:

| | result |
|---|---|
| waves swept | 21 |
| tier rows | 293 |
| unparseable rows | 0 |
| non-list `tier_*` values | 0 |
| rows whose repo is absent from `repos_in_scope` | 0 |

With `unparseable == 0` the new code reduces to `total_declared == len(canonical)` and `_fmt` appends no suffix, so on every record in the file's history the message is byte-for-byte what it was pre-#1248. No previously-warning input goes quiet; no new warning fires. Live wave-29 is 51/51 rows parsed across `noorinalabs-main` + `noorinalabs-isnad-ingest-platform`, both in `repos_in_scope`.

## The previously-silent degenerate case — fixed through the real CLI, and it cannot re-introduce `None`

Confirmed the fix survives the caller, not just the helper: `_merged_prs_direct_to_main` short-circuits on empty `canonical_pairs` returning `claimed = set()` — **not** `None` — so the `if claimed is None` guard does not swallow the warning. End-to-end via `wave_status.main(["counters", …])` with an all-unparseable scope: `rc = 0`, `final_pr_count: 0`, and now `WARNING: 0 scope rows unclaimed (of 4 declared); 4 scope rows unparseable …` on stderr where pre-#1248 there was nothing.

The new path can still return `None` in exactly three places, and I checked each: `claimed is None` (wave-branch model, by design); `not canonical and not unparseable`; `not unclaimed and not unparseable`. The first two are correct. The third is where I found the one residual (below).

## The 100-node heuristic — does not overclaim, and cannot cry wolf here

"may be truncated" is properly hedged, and the trigger is pinned in both directions (100 warns, 99 does not). On the false-positive risk you flagged: across the last 100 merged PRs in `noorinalabs-main` the maximum `closingIssuesReferences.totalCount` is **3** (PR #1198); zero PRs are at or above 100. A permanently-firing warning is not reachable against this org's data. I did confirm live that the heuristic could be made exact for free — see #1257.

## Mutation testing

Your requested re-run, plus my own. Reverted `.claude/lib/wave_status.py` to `origin/main` (`2dbb0f7a`) with the new tests in place, `PYTHONDONTWRITEBYTECODE=1 ENVIRONMENT=test`:

```
8 failed, 36 passed in 0.69s
```

Exactly the author's claim, and the 8 spread across all three edges — including `test_warning_surfaces_through_the_real_merged_prs_call_path`, which fails on `'100-node page cap' not found in ''`. That is the assertion that matters for this bug class: it fails because the signal never reaches the top of the stack, not because a private helper's return value changed.

Then 10 targeted mutations of my own against the individual new guards — **9 caught, 1 survived**:

| mutation | outcome |
|---|---|
| `total_declared = len(canonical) + unparseable` → `len(canonical)` | caught |
| `if not canonical and not unparseable` → `if not canonical` | caught |
| `if not unclaimed and not unparseable` → `if not unclaimed` | caught |
| `len(nodes) == 100` → `== 101` | caught (2 tests) |
| `if repo_name not in repos_in_scope` → `if False` | caught |
| each of the three `unparseable += 1` increments → `pass` | caught (4 / 3 / 1 tests) |
| unparseable clause never appended | caught (2 tests) |
| `plural = "s" if unparseable != 1 else ""` → always `"s"` | **survived** |

Only the singular-vs-plural cosmetic survived — the `unparseable == 1` branch has no fixture. Recorded in #1256, not blocking.

The tests also carry the negatives that make them non-vacuous: 99 nodes must not warn, and an in-scope not-yet-delivered row must **not** get the `(repo not in repos_in_scope)` suffix.

## Suite and gate state

- `.claude/lib/tests .claude/hooks/tests` at head: **3483 passed, 647 subtests**, zero failures. Collection: base `origin/main` 3476, head 3483. The PR body's 3455 predates main moving; no failures either way.
- `statusCheckRollup` re-read at `34fb1714`: **14/14 SUCCESS, 0 non-success**. `MERGEABLE` / `CLEAN`.

## Findings — all non-blocking, all filed

- **#1255** — #1201 Edge 1 is fixed at the *row* level only. A `tier_*` key whose **value** is not a list (a bare dict, or a plain string) is skipped by `if not isinstance(value, list): continue` and counts toward neither numerator, denominator, nor `unparseable`. If it is the only tier key, `counters` exits 0 with `final_pr_count: 0` and **nothing on stderr** — the same silent zero, one structural level up. Reproduced end-to-end through the real CLI. Latent: 0 of 293 rows and 0 of 21 waves hit it, and the three edges #1201 actually specified are all delivered, so this is debt rather than a blocker on this PR.
- **#1256** — the line's own wording. `(excluded from the count above)` is ambiguous against a denominator that *includes* the unparseable rows (`2 of 4`, where 4 = 2 parsed + 2 unparseable); the literal reading yields 6 declared. Also, the all-unparseable case leads with the reassuring `0 scope rows unclaimed` before naming the 4 unparseable rows. Plus the untested singular branch above.
- **#1257** — `totalCount` and `pageInfo{hasNextPage,endCursor}` are available on `closingIssuesReferences` on the same node at no extra cost (verified live against PR #1173). The `len(nodes) == 100` heuristic could become an exact "truncated at 100 of N", eliminating the cry-wolf class entirely.

## Not verified

- Real GraphQL truncation behaviour at a genuine 100+ closer — no such PR exists in this org (max 3), so the page-cap path is exercised only against the fake `gh`. I did independently confirm against the real engine that `first:100` is the true ceiling and that `totalCount`/`pageInfo` exist, so the heuristic's premise is sound.
- The `pre-commit`/`pre_commit_ci_sync` claims in the PR body I did not re-run locally; I relied on the 14/14 rollup at the head SHA, which includes ruff-lint, ruff-format, mypy, and the sync-drift gate.

The brief I was given was wrong about this issue in two ways and the author caught both before writing a line. The delivered change is the issue's actual spec, mutation-proven at the call path rather than at the helper, and provably inert against every wave record that exists.

---
Requestor: Weronika Zielinska
Requestee: Santiago Ferreira
RequestOrReplied: Approved
TechDebt: #1255, #1256, #1257

Issue #1243, opened 2026-08-03
validate_wave_audit: no SSOT-identity guard on _ORG_REPOS (mutation survives full suite) + org_repos ImportError fails open


parametrization:
## Context

Follow-up from the merge-gate review of #1238 (audit G6, closes #1118). Two robustness gaps in `.claude/hooks/validate_wave_audit.py`'s new dependency on the `org_repos` SSOT. Neither is a regression introduced by #1238 — both are pre-existing properties that the migration makes worth closing — so #1238 was approved with these split out here.

## 1. `_ORG_REPOS` has no SSOT-identity guard — mutation survives the entire suite

#1238 added a strong guard for `session_handoff.py`:

```python
self.assertIs(hook.ALL_REPOS, org_repos.ALL_REPOS)
```

`validate_wave_audit.py` — the *blocking* gate, and the one hardened by #1226/#1229 precisely because a partial per-repo query failure summed to a confident zero and turned a BLOCK into a silent ALLOW — got no equivalent. Mutation evidence at head `e0b6711`, clean full-depth clone, `PYTHONDONTWRITEBYTECODE=1`:

- Replace `_ORG_REPOS = ALL_REPOS` with a hand-copied 7-tuple that drops `noorinalabs-isnad-ingest-platform` (the exact BUG-08 shape) → **3448 passed, 647 subtests passed. Zero failures.** The mutation survives.
- Replace it with `_ORG_REPOS = ()` → 13 failures. So total emptiness is loud; a one-repo omission is invisible.

Today the constant is a direct alias, so drift requires someone actively re-typing a list where an import now stands. But that is exactly the edit BUG-08 was, and nothing in CI would say a word.

**Proposed:** add to `test_validate_wave_audit.py` the same assertion `session_handoff` got — `assertIs(hook._ORG_REPOS, org_repos.ALL_REPOS)` — plus a length assertion. Cheap, and it kills the mutation above.

## 2. An unresolvable `org_repos` import makes the blocking gate fail OPEN

`validate_wave_audit.py` is a PreToolUse gate: exit 2 blocks; any other non-zero exit is a *non-blocking* error and the tool call proceeds. Measured at head:

| Condition | Exit code | Effect |
|---|---|---|
| `.claude/lib/org_repos.py` present (head as-is) | **2** | blocks, with the full audit reason |
| `.claude/lib/org_repos.py` removed | **1** (`ModuleNotFoundError: No module named 'org_repos'`) | does not block |

[...]

Issue #1234, opened 2026-08-02
tech-debt(hook17): #1229's partial-vs-total signal test pins total-failure to ALLOW, pre-deciding #1230


parametrization:
## The over-specification

`test_partial_failure_is_never_quieter_than_total_failure`
(`.claude/hooks/tests/test_validate_wave_audit.py`, added by PR #1229) states a
*relative* property in its name — the partial-failure path must carry at least
as much operator signal as the total-failure path beside it — but implements it
as an *absolute* assertion about the total path:

```python
total, q2 = self._run(set(hook._ORG_REPOS))
assert total is not None
self.assertTrue(total.get("systemMessage"))
```

A `block` result is `{"decision": "block", "reason": ...}` — it carries no
`systemMessage`. So the assertion pins the total-failure path to
**allow-with-warning**, not to "emits signal".

## Why it matters now

This directly collides with #1230, which asks for an explicit ruling on whether
total coverage failure should block for `/wave-wrapup` and `/wave-retro`. If
#1230 lands on option 1 (extend coverage-blocking to the `total is None` path),
this test fails — even though the property its name asserts would still hold, a
block being strictly *louder* than a warning.

Mutation-verified against `b6d6eae`: making `total is None` block for
`_COVERAGE_BLOCKING_SKILLS` fails exactly three tests —
`FailOpenInfrastructure::test_audit_total_failure_allows` (pre-existing, and
correctly scoped: it is *about* the fail-open),
`PartialAuditFailureCoverage::test_total_failure_still_allows_with_warning`
(a deliberate CONTROL pinning unchanged behaviour — legitimate), and this one,
which is neither.

## Ask

Re-express the assertion in terms of signal presence rather than verdict kind,
e.g. assert `result.get("reason") or result.get("systemMessage")` on **both**
sides and compare, so the test survives whichever way #1230 is ruled. Note the
dependency on #1230 so whoever implements it is not surprised.

Found during merge-gate review of #1229 (Weronika Zielinska), by mutating the
shipped artifact at `b6d6eae` rather than by reading the diff.

parametrization:
**Deferred to wave-30** (owner disposition, 2026-08-02).

This issue was filed by a review gate during wave-29 and carried a `wave-29` label it was never scoped into — one of 6 drift rows against that wave's *labelled == scoped* invariant. The owner accepted the recommendation to defer it.

[...]

PR #1250, opened 2026-08-03
fix(#1230): total wave-audit coverage failure now blocks like partial does

### LINKED ISSUE #1230: tech-debt: validate_wave_audit total-coverage-failure is more lenient than partial-coverage-failure for /wave-wrapup, /wave-retro

parametrization:
Closes #1230

## Ruling: option 1 — extend `_COVERAGE_BLOCKING_SKILLS` treatment to `total is None`

#1230 asked for an explicit ruling on the asymmetry `#1229` left in place: a
**partial** audit-coverage gap (1-of-8 through 7-of-8 org repos unqueried)
BLOCKs `/wave-wrapup` and `/wave-retro`, but a **total** gap (8-of-8, `total is
None`) fail-opened unconditionally, ahead of the `_COVERAGE_BLOCKING_SKILLS`
check — making the bigger blind spot strictly MORE permissive than the smaller
one, for the exact two skills #1226 hardened because "an unknown is never
green."

I'm ruling for **option 1**: total failure now blocks exactly like partial
failure does, for `_COVERAGE_BLOCKING_SKILLS`, with its own message copy
("could not run at all" vs "ran but was incomplete" are different enough
facts). `/handoff` is unaffected — still allows, loudly, on both partial and
total failure.

**Why the "deadlock" defense (the strongest case for option 3 / status quo)
doesn't survive scrutiny**, argued in the docstring/charter now rather than
left implicit:
- It proves too much — it applies just as well to a 7-of-8 gap, which this
  gate already blocks unconditionally.
- A rate-limit/API outage that happens to hit all 8 calls at once (plausible —
  GraphQL calls share a quota) resolves via the same "wait and re-run" remedy
  as a partial gap. It is not categorically different.
- Only a genuinely broken `gh` install forces the settings.json emergency
  exit — and that exit was already the *only* remedy for an unfixable partial
  failure too (e.g. a persistent auth failure on the one repo carrying the
  wave's issues). Total failure doesn't introduce a new failure mode, it just
  makes an existing one apply to every repo at once.

So the asymmetry wasn't a defensible tuning choice, it was the same #1226 bug
one layer up, and it's closed now rather than documented-and-kept.

## #1234's test — ruling (b): it was pinning the accident, not an intended contract

`test_partial_failure_is_never_quieter_than_total_failure`'s own NAME states a
*relative* property (partial must carry at least as much signal as total), but
its body asserted an *absolute* one (`total.get("systemMessage")` truthy,
which only holds if total failure is allow-shaped). #1234 correctly flagged
this as over-specification that would pre-decide #1230 by accident if left
alone.

Ruling: **(b)** — it was pinning the accident. I did not silently rewrite it.
I re-expressed the assertion per #1234's own suggested fix (assert
`result.get("reason") or result.get("systemMessage")` is truthy on both sides,
and compare), and left a comment on the test explaining #1234's finding and
that #1230 has now ruled. The property in the test's name still holds — more
strongly than before, since both sides are now `block`.

Two OTHER pre-existing tests also pinned the same old absolute contract and
needed the same treatment (not flagged by #1234, which only named the one, but
mutation-equivalent):
- `FailOpenInfrastructure.test_audit_total_failure_allows` exercised
  `skill="wave-wrapup"` (the `_skill_input` default) and asserted `allow`
  unconditionally. Renamed to `test_audit_total_failure_allows_for_noncoverage_skill`
  and re-scoped to `skill="handoff"`, the only skill for which that assertion
  is still true.
- `PartialAuditFailureCoverage.test_total_failure_still_allows_with_warning`
  (docstring: "CONTROL ... unchanged fail-open") is renamed
  `test_total_failure_blocks_coverage_blocking_skill` and now asserts `block`
  for `wave-wrapup`. A new `test_total_failure_still_allows_handoff_with_warning`
  keeps the old CONTROL assertion, correctly re-scoped to `/handoff`.

New tests added: `test_total_failure_blocks_wave_retro`,
`test_total_failure_carry_forward_does_not_clear_the_block` (mirrors the
partial-failure carry-forward test).

## What changed

- `check()`: the `if total is None:` branch now checks `skill_name in
  _COVERAGE_BLOCKING_SKILLS` before deciding allow vs block, exactly parallel
  to the existing `if unqueried:` handling for partial coverage. New block
  message distinct from the partial-coverage one.
- Module docstring (`Block condition`, `Allow condition`, `Failure modes`,
  `Exit codes`) and the `_COVERAGE_BLOCKING_SKILLS` inline comment rewritten to
  describe the new behavior and record why the old defense doesn't hold, so a
  future reader doesn't have to reconstruct the argument from this PR.

`.claude/team/charter/hooks/catalog-13-17.md`
- Hook 17's `Audit-coverage gate` bullet rewritten to describe both #1226
  (partial) and #1230 (total) under one gate, with the total-failure rationale
  spelled out — this is the charter-side half of #1230's option 2 that
  Weronika's #1229 re-review flagged as undelivered (module docstring got the
  rationale in #1229's `6d9c89ff`, the charter entry did not).

`.claude/hooks/tests/test_validate_wave_audit.py`
- See the #1234 section above for the renamed/re-scoped tests.
- `FailOpenInfrastructure`'s class docstring updated: it's no longer
  universally true that "infrastructure failures fail OPEN ... never block" —
  that's now conditional on `skill_name`.

## Verification

- **Non-vacuity by mutation**: stashed the production fix (kept the new
  tests), re-ran the suite against unmodified `check()`. All 4
  new/re-scoped tests fail cleanly (`test_partial_failure_is_never_quieter_than_total_failure`,
  `test_total_failure_blocks_coverage_blocking_skill`,
  `test_total_failure_blocks_wave_retro`,
  `test_total_failure_carry_forward_does_not_clear_the_block`), then restored
  the fix and re-ran — all pass. Ran with `PYTHONDONTWRITEBYTECODE=1
  ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/test_validate_wave_audit.py -q`.
- **End-to-end exit-code matrix**, real subprocess, fake `gh` on `PATH` failing
  a configurable N of the 8 org repos (measured, not reasoned about):

  | skill | 0-of-8 | 1-of-8 | 7-of-8 | 8-of-8 |
  |---|---|---|---|---|
  | `/wave-wrapup` | exit=0 | exit=2 | exit=2 | **exit=2** (was 0) |
  | `/wave-retro`  | — | exit=2 | — | **exit=2** (was 0) |
  | `/handoff`     | — | exit=0 | — | exit=0 (unchanged) |
  | `/wave-wrapup` + carry-forward marker, 8-of-8 | — | — | — | **exit=2** (marker does not clear it) |

- `PYTHONDONTWRITEBYTECODE=1 ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/ .claude/lib/tests/ -q` — **3479 passed, 647 subtests passed**.
- `pre-commit run --all-files` — ruff-format, ruff-lint, actionlint, cspell,
  fixture-realism, skill-graphql-pagination, checksums-ascii: **Passed**.
- `pre-commit run --all-files --hook-stage pre-push` — mypy, mypy-skills,
  pytest, pytest-skills, memory-budget, headcount-budget, doc-freshness
  (advisory), office-drift, mermaid-render: **Passed**.
- `python3 .claude/lib/pre_commit_ci_sync.py .` — `OK: pre-commit config
  mirrors all CI-enforced checks.`
- Branch rebased onto current `main` (`a52b2b2` on top of `36d16fa`) —
  rebase was clean (no conflicts; the intervening commit only touched
  `cross-repo-status.json`).
- No known-failing check pushed past. CI status not yet re-checked post-push —
  will report `gh pr view --json statusCheckRollup` results, not claim green
  from local runs alone.

## On the two live findings filed against this exact module today (#1243) — assessed, not fixed here

**#1243 finding 1 (`_ORG_REPOS` has no SSOT-identity guard):** does not
collide with this PR — I don't touch the `_ORG_REPOS = ALL_REPOS` assignment
or its identity. Left to #1243; it's a separate hardening of the mutation
surface, not something my change to `check()`'s control flow needs.

**#1243 finding 2 (`org_repos` `ImportError` exits 1, not 2, so the gate fails
open on a broken import):** I considered whether to fix this inline since it
is now, after this PR, a *sharper* version of the very asymmetry I'm closing —
a full audit-run total failure (8-of-8 `gh` calls failing) now correctly
BLOCKS for wave-wrapup/wave-retro, but an `org_repos.py` import failure (an
even more total outage — the hook can't even start) still silently exits 1
and allows. I'm judging this **out of scope for #1230** rather than folding it
in silently:

- Fixing it *correctly* isn't a one-line `try/except sys.exit(2)` around the
  import. The import happens at module load, before `check()` knows
  `skill_name` — a naive fail-closed-on-import-error would block *every*
  gated skill (including `/handoff`, which is deliberately never
  coverage-blocked) and, worse, would fire for `/ontology-librarian` and every
  other non-gated `Skill` call too, since module load happens before the
  `tool_input.skill not in _GATED_SKILLS` early-return. Doing this right needs
  deferring the `org_repos` import until after the skill is known to be
  gated — a real structural change to the hook's import ordering, not a
  drive-by inside a PR whose stated scope is the `total is None` branch.
- #1243 already tracks this precisely, with its own reproduction and proposed
  fix (wrap the import, fail closed). I'd rather it land as its own
  reviewable diff than get entangled with this one's rebase/test surface.

Flagging loudly per the task brief rather than leaving it unmentioned:
whoever picks up #1243 should know this PR makes the import-failure gap
comparatively more dangerous than it was, since the "total audit ran but
found nothing" case it used to sit next to now blocks.


parametrization:
## Verdict: Approved

Reviewed at head `a52b2b2d`. Re-derived the behavior independently rather than accepting the PR's own tables — findings below.

### 1. Exit-code matrix — re-derived with a real subprocess + fake `gh`

Built a fake `gh` on `PATH` (not the mocked unit-test boundary) and ran `python3 .claude/hooks/validate_wave_audit.py` as an actual subprocess against a real `cross-repo-status.json`, varying how many of the 8 org repos fail:

| skill | 0-of-8 | 1-of-8 | 7-of-8 | 8-of-8 |
|---|---|---|---|---|
| `/wave-wrapup` | exit=0 (no output) | exit=2 block | exit=2 block | exit=2 block |
| `/wave-retro` | — | — | — | exit=2 block |
| `/handoff` | — | exit=0 allow+warning | — | exit=0 allow+warning |
| `/wave-wrapup` + carry-forward, 8-of-8 | — | — | — | exit=2 block (marker does not clear it) |

This matches the author's claimed matrix exactly, independently reproduced. My first fake-`gh` script had a `set -u` bug that made every call fail regardless of the intended failure set (caught because 0-of-8 unexpectedly blocked) — fixed and re-run; noting it so the methodology is transparent, not just the result.

### 2. Mutation check — reproduced

Swapped in the pre-#1230 production file (kept the new/changed test bodies) and ran `.claude/hooks/tests/test_validate_wave_audit.py`: exactly 4 tests fail —
`test_partial_failure_is_never_quieter_than_total_failure`,
`test_total_failure_blocks_coverage_blocking_skill`,
`test_total_failure_blocks_wave_retro`,
`test_total_failure_carry_forward_does_not_clear_the_block` — nothing else breaks and nothing else that should fail stays green. Restored the fix: 109 passed, 7 subtests passed. `ruff check` on both changed Python files: clean.

### 3. #1234 ruling — judged on the merits, holds up

Read #1234 directly. Its own mutation evidence (at `b6d6eae`) found the *old* code's total-failure branch pinned by three tests, two of which #1234 itself called legitimate at the time — because pre-#1230 they correctly described unconditional allow-on-total-failure, which was still the shipped contract. #1230 now overturns that contract for `_COVERAGE_BLOCKING_SKILLS`, which is exactly why those two ALSO need updating now, not because anyone quietly widened #1234's ask. The rewrites check out individually:
- `test_partial_failure_is_never_quieter_than_total_failure` is re-expressed exactly per #1234's own suggested fix (`reason` or `systemMessage` truthy on both sides), plus it now also pins the concrete post-ruling contract (`assertEqual(..., "block")` on both sides) — strictly more coverage than before, not less.
- `test_audit_total_failure_allows` → renamed/re-scoped to `handoff` (the only skill for which "total failure allows" is still true) rather than deleted.
- `test_total_failure_still_allows_with_warning` → renamed to `test_total_failure_blocks_coverage_blocking_skill` (asserts the new contract for `wave-wrapup`), and a **new** `test_total_failure_still_allows_handoff_with_warning` preserves the old CONTROL, correctly re-scoped. Coverage of "handoff still allows on total failure" is not lost, just relocated to its own test.

Confirmed #1230's own text asks for exactly ruling 1 or 2, and this PR delivers ruling 1 verbatim. Confirmed #1234 is not referenced with a closing keyword in this PR's body/commit and remains OPEN with its wave-29 label already stripped per the owner's 2026-08-02 deferral-to-wave-30 disposition (`phase_10_carry_forwards`) — this PR does not close or contradict it.

### 4. The deadlock counter-argument — tested, does not hold

Worked through whether there's a realistic state where 8-of-8 failure blocks with no way forward that 7-of-8 (already blocking) wouldn't also produce. The practical root causes of a total failure — `gh` missing/broken, full auth failure, an org-wide rate-limit/quota exhaustion — are exactly the causes that would just as readily produce 7-of-8 or worse; there's no scenario where "coincidentally all 8, not 7" is a stable, reproducible category with a qualitatively different remedy. A per-repo-specific failure (bad permissions on one repo, a transient blip) that could plausibly land at 7-of-8 rather than 8-of-8 is, by construction, not the kind of failure that would also hit the 8th — so it doesn't generalize into a real 8-of-8 deadlock case either. The settings.json emergency exit was already the sole remedy for an unfixable partial failure pre-this-PR; this change doesn't remove or narrow it, it just makes the same exit apply symmetrically. Agree with the ruling.

### 5. Scope discipline on #1243 — correct deferral

Confirmed directly in `settings.json`: the hook is registered on the bare `"Skill"` matcher, firing (and importing `org_repos`) for **every** `/Skill` invocation, not just the three gated ones — `_GATED_SKILLS` filtering happens only inside `check()`, after import. This confirms the author's stated reason for not folding in #1243 finding 2 (`ImportError` exits 1, not 2): a naive fail-closed wrap at the current import site would block every non-gated skill call org-wide (`/ontology-librarian`, `/session-start`, etc.), which is a materially worse regression than the gap it would close. Deferring to #1243 for a proper import-ordering fix is right, and the PR's own callout — that this change makes the import-failure asymmetry sharper — is accurate and correctly flagged rather than left implicit.

### Unverified / not independently re-run
- Did not re-run the full repo-wide `pytest`/`mypy`/`pre-commit --all-files` suite (3479 tests) myself — relied on the orchestrator's CI-green (21/21) confirmation plus my own scoped `ruff` and targeted `pytest` runs on the two changed files, which passed.
- Did not re-verify `mergeable`/`CLEAN`/rebase state myself — relied on the orchestrator's stated verification.

Charter (`catalog-13-17.md` Hook 17 entry) diff matches the code change; no stale references to the old asymmetry found elsewhere in the charter (`ci-gates.md`/`state-claims.md` "deadlock" hits are unrelated PR-ruleset discussions).

---
Requestor: Aino Virtanen
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Merge-gate review — PR #1250 @ `a52b2b2d`** (Program Director, authoritative pre-merge)

Reviewed read-only at the head SHA via the contents API and a full-depth clone in my own scratch dir. I did not accept the exit-code table in the description — I rebuilt it against a real subprocess with a fake `gh` on `PATH`, and re-ran the mutation independently. Every claim in the PR body that I checked held.

## Verdict: **LGTM — approved for merge.**

## 1. The deadlock objection — is there a state where 8-of-8 fails and an operator cannot proceed?

This was the call I was least willing to take on agreement, since I found the reasoning persuasive on first read and that is exactly when a gate gets waved through. Adversarial pass says **no such state exists**, and the argument is stronger than the one in the PR body.

The description's defense ("it proves too much — it applies equally to 7-of-8") is sound but incomplete. There *is* one state reachable at 8-of-8 and not at 7-of-8: **`gh` missing or fundamentally broken**, since 7-of-8 requires `gh` to work for at least one repo. So "applies equally" does not by itself cover the whole new surface. What closes it is that the two newly-blocked skills cannot do their work in that state anyway:

| skill | `gh` invocations in its own skill files |
|---|---|
| `/wave-wrapup` | 38 |
| `/wave-retro` | 7 |
| `/handoff` | 2 |

With `gh` broken, `/wave-wrapup` cannot merge a PR or close an issue and `/wave-retro` cannot analyse one. Pre-#1250 the gate waved them through into a run that would fail partway — **after** it had begun mutating wave state. Post-#1250 they stop cleanly at the front door with a diagnosis. That is a strict improvement, not a deadlock: the operator loses nothing they could actually have done.

I verified the block survives the very condition that triggers it — **`gh` entirely absent from `PATH`**:

```
wave-wrapup    exit=2 decision=block  stderr=(empty)
wave-retro     exit=2 decision=block  stderr=(empty)
handoff        exit=0 decision=allow  stderr=(empty)
```

Clean exit 2, no traceback. `annunaki_log.py` has no `subprocess`/`gh` dependency, so the logging call on the new blocking path cannot crash the hook into a non-2 fail-open. This was the sharpest failure mode available against this change and it does not fire.

Three further escape routes remain open, so the gate is never unclearable:
- **Recoverable causes** (auth, rate-limit, transient network, per-repo timeout) resolve by wait-and-re-run — the identical remedy already in force for the 7-of-8 block.
- **The `labels is None` path still fail-opens unconditionally for every skill**, so a missing or malformed `cross-repo-status.json` is *not* caught by the new block. The rewritten `FailOpenInfrastructure` docstring draws this distinction explicitly, and `test_no_active_wave_allows` still pins it.
- **The documented emergency override exists** and the block message's pointer resolves: `catalog-13-17.md` § Hook 17 — "Remove the `Skill` matcher entry from `.claude/settings.json`." A block message pointing at a procedure that did not exist would have been a finding; it does exist.

**On over-blocking specifically**, one pre-existing hazard I checked because it would have been a real regression: `_open_issue_numbers_for_repo` returns `None` if *any* per-label query fails, and the hook queries two label forms. If `gh issue list --label <absent-label>` errored, every repo lacking `p10-wave-29` would go unqueried and could manufacture a spurious total failure. Measured against the live API — it exits 0 with `[]`. No spurious coverage gap.

## 2. Live wave-29 impact

Ran the real aggregation against live state at head:

```
REAL AUDIT (5.9s): total=27  unqueried=[]   per_repo={'noorinalabs-main': 27}
```

**Full coverage, zero unqueried**, 16 `gh` calls in 5.9s — comfortable headroom under the 3s-per-call timeout, so the new total-failure branch is nowhere near firing spuriously here. End-to-end, `/wave-wrapup` and `/wave-retro` both exit 2 today on the **pre-existing open-items count** (27 open), the identical behaviour they have on `main`. This PR does not touch the path the wave-29 wrapup will actually take.

## 3. Test rewrites — coverage preserved, not relabelled

This was the right thing to be suspicious of, and it survives scrutiny. Name-level diff: **2 dropped, 5 added, 106 → 109**. Both dropped names' *input scenarios* are retained rather than deleted — `wave-wrapup` + total failure is re-asserted as `block`, and the old allow-shaped CONTROL is re-scoped to `/handoff`, the only skill for which it is still true.

**Mutation A** — reverted `validate_wave_audit.py` to `main`, kept the tests at head:

```
4 failed, 105 passed
FAILED ... test_partial_failure_is_never_quieter_than_total_failure
FAILED ... test_total_failure_blocks_coverage_blocking_skill
FAILED ... test_total_failure_blocks_wave_retro
FAILED ... test_total_failure_carry_forward_does_not_clear_the_block
```

Exactly the 4 claimed, no more and no fewer. Non-vacuous.

**Mutation B (mine, not in the PR body)** — the concern Mutation A cannot address: the two *re-scoped* `/handoff` tests assert `allow` and therefore pass both pre- and post-fix, which is the signature of an inert relabel. So I mutated the other axis and added `"handoff"` to `_COVERAGE_BLOCKING_SKILLS`:

```
4 failed, 105 passed
FAILED ... FailOpenInfrastructure::test_audit_total_failure_allows_for_noncoverage_skill
FAILED ... PartialAuditFailureCoverage::test_total_failure_still_allows_handoff_with_warning
FAILED ... test_coverage_blocking_skills_are_a_subset_of_gated_skills
FAILED ... test_handoff_partial_failure_allows_with_explicit_warning
```

Both re-scoped controls trip. They are live guards on `/handoff`'s exclusion at **both** the partial and total path, not relabelled decoration. Coverage is preserved and net-strengthened.

`test_partial_failure_is_never_quieter_than_total_failure` is now genuinely better than what it replaced: the old body asserted an absolute allow-shaped property under a name promising a relative one, and the new body compares signal presence on both sides *and* pins both verdicts. #1234's ruling (b) is correct — that assertion was pinning an accident. **#1234 remains OPEN, carries no wave label, and this PR does not contradict it**: the fix applied is #1234's own suggested one, and the ruling is recorded in the test docstring rather than applied silently. Deferral to wave-30 is intact.

## 4. Exit-code matrix — rebuilt, not accepted

Real subprocess, fake `gh` on `PATH` failing a configurable N of 8:

| scenario | head `a52b2b2d` | baseline `main` |
|---|---|---|
| `/wave-wrapup` 0-of-8 | exit 0 (silent allow) | exit 0 |
| `/wave-wrapup` 1-of-8 | exit 2 block | exit 2 |
| `/wave-wrapup` 7-of-8 | exit 2 block | exit 2 |
| `/wave-wrapup` 8-of-8 | **exit 2 block** | **exit 0 allow** |
| `/wave-retro` 8-of-8 | **exit 2 block** | **exit 0 allow** |
| `/wave-wrapup` 8-of-8 + carry-forward | **exit 2 block** | **exit 0 allow** |
| `/handoff` 1-of-8 | exit 0 allow | exit 0 |
| `/handoff` 8-of-8 | exit 0 allow | exit 0 |
| `/ontology-librarian` 8-of-8 (ungated) | exit 0 | exit 0 |

Reproduces the description's table exactly, and the baseline column independently confirms the inversion was real (8-of-8 more permissive than 7-of-8) and is now closed. `/handoff` unchanged on both axes. Carry-forward correctly does not clear the new block.

Code reads correctly: the `total is None` branch now splits on `_COVERAGE_BLOCKING_SKILLS` in exact parallel to the `if unqueried:` branch below it, with distinct copy. Hoisting `args` above the branch is a pure dict read with no side effects, and `label_display` is already in scope. The `#981` early-allow marker was kept at the site rather than dropped once the hole was closed, which is the right instinct.

## 5. The #1243 deferral — judgment is correct, and the gap does not swallow this PR

Assessed empirically rather than on the author's reasoning. Renamed `.claude/lib/org_repos.py` aside and ran the hook for real: **exit 1 for all four skills tested — including ungated `/ontology-librarian`.**

That fourth result is the one that settles it. The module-level import executes before `check()` ever sees `tool_input.skill`, so a naive `try/except ImportError: sys.exit(2)` would hard-block *every* `Skill` call in the session, not just the two this gate protects. The stated reason for deferring is not a rationalisation — it is measurably correct, and the correct fix (defer the import past the `_GATED_SKILLS` early-return, then fail closed only for `_COVERAGE_BLOCKING_SKILLS`) is a structural change that genuinely belongs in its own diff.

Does leaving it swallow this PR's benefit? **No — the trigger surfaces are disjoint.** #1230's fix covers `gh` outage, auth failure and quota exhaustion; #1243's gap requires `org_repos.py` to be missing or unimportable, a repo-state defect rather than a runtime condition. Neither causes the other, so no realistic outage routes around the new block. Nino is right that his change makes the residual comparatively more conspicuous, and flagging it in the PR body rather than leaving it buried is the correct handling.

I have posted the full measurements to #1243 (which is OPEN and tracks both findings) rather than filing a duplicate.

**One non-blocking observation, folded into #1243 rather than raised as new debt:** this PR rewrites the module docstring's § `Exit codes`, which still enumerates only `0` and `2` and omits that an `org_repos` `ImportError` yields `1`. The section reads as exhaustive when it is not. Pre-existing rather than introduced here — the omission predates the rewrite — so it is not a merge blocker and not chargeable to this PR.

## 6. CI

**21/21 SUCCESS, every check-run confirmed at head SHA `a52b2b2d`** via the commit check-runs API, not merely the PR-level rollup. `MERGEABLE` / `CLEAN`. Independently re-ran `.claude/hooks/tests/ .claude/lib/tests/` in a clean full-depth clone: **3479 passed, 647 subtests passed** — matches the description exactly. Working tree left clean, all mutations reverted.

## Not verified

- Did not re-run `pre-commit` locally (CI's ruff/mypy/cspell/actionlint jobs are green at head and cover the same check-set).
- The 3s-per-repo timeout under a genuinely slow-but-working network is reasoned about, not induced. It is not a new class — 7-of-8 timeouts already blocked — and live coverage has 5.9s of headroom across all 16 calls.
- Did not exercise the `_mergeready_exempt_issues` interaction with total failure; that path is unreachable when `total is None`, since the branch returns before it.

Approving. This closes a real inversion — the more complete outage was the more permissive outcome — with a fix symmetric to the partial-coverage handling it mirrors, tests that fail for the right reasons, and a deferral whose rationale I could measure rather than take on trust.

---

Requestor: Nadia Khoury
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: none

Issue #1254, opened 2026-08-03
tech-debt(roster_union_sync): the manifest-orphan report can never fail and is only visible in a green job's log — and its two output streams interleave out of order


parametrization:
## Summary

PR #1240 added the reverse (manifest-orphan) check to `.claude/lib/roster_union_sync.py`. After the correction made during its review, that check **can never move the exit code** — by design, and correctly so: a non-zero exit turned the `Roster union sync-drift gate` job FAILING in `statusCheckRollup` on every PR in the repo, because `continue-on-error: true` protects the workflow *run*'s conclusion but not the *job*'s reported conclusion, which is the surface `pr_ci_state.py` / `validate_pr_ci_status.py` read.

That fix is right. It does, however, leave the check with **no observable output channel that anyone sees**:

- The job is `SUCCESS` whether the report found 16 orphans, 0 orphans, or did not run at all.
- The only artifact is text in the log of a green job. Nobody opens the log of a green job.
- A fetch failure (any of the 6 `roster.json` fetches or 7 card-directory fetches) suppresses the check entirely. That is the safe direction — it prevents a false "prune this person" — but it is **indistinguishable from a clean pass at every surface except the log text**.

Measured on a live run at PR #1240's head: exit 0, 16 unreconciled orphans + 2 non-persona entries reported, job green.

Secondary, and compounding it: the report goes to **stderr** while the per-repo `OK`/`SKIPPED` context goes to **stdout**. With a pipe (which is what CI gets), stdout is block-buffered and stderr is not, so the `MANIFEST ORPHAN` block prints *before* the `OK ...` lines it is supposed to be read against. Observed on a live run:

```
MANIFEST ORPHAN (#1181, advisory ...):
  - "Amara Diallo"
  ...
OK      noorinalabs-isnad-graph: 15 persona(s).
```

So even the one channel it has is scrambled.

## Why this matters

An advisory check whose only consumer is a green job's log has no reader, and the org has repeatedly re-encountered the "a skip that is indistinguishable from a pass" shape. This is not an argument to make it blocking — the disposition of the 16 names is genuinely open (#1181) and blocking on them would re-create the permanently-red state. It is an argument that it needs a **visible-without-failing** channel.

## Options

1. **Job summary** — append the report to `$GITHUB_STEP_SUMMARY` so it renders on the workflow run page without affecting any conclusion. Cheapest, no new surface.
[...]

PR #1258, opened 2026-08-03
audit(G8): session handoff: one search call instead of 8 serial subprocesses

### LINKED ISSUE #1120: audit(G8): session handoff: one search call instead of 8 serial subprocesses
Source: CODE_AUDIT.md §10 (noorinalabs-main @ 3bdd620) — item G8.

**G8 — Handoff hook: one search call instead of 8 serial subprocesses** · Type: perf · Effort: S
`session_handoff` runs 7 sequential `gh pr list` + 1 issues call (worst case ~120 s vs a 30 s hook timeout). Replace with one `gh search prs --owner noorinalabs` (also removes the literal list — pairs with G6) or a thread pool.
*After:* 7–10× faster handoff writes. *Verify:* `test_session_handoff.py` asserts the issued command shape.

Index row (§13): P | 8 serial subprocesses → 1 | S | G6

parametrization:
## Summary
- Replaces `_get_open_prs()`'s per-repo `gh pr list` loop (7 serial pre-#1238, 8 post-#1238; worst case ~120s against a 30s Stop-hook timeout) with one `gh search prs --owner noorinalabs` call spanning the whole org. No longer needs a repo list to build the query at all (pairs with G6).
- Failure-shape change addressed head-on: a single failed call now drops the whole org's PR list at once instead of just one repo's silently-dropped slice. `_get_open_prs()` returns a `PrQueryResult` that keeps "query failed" (`_run()` returned `""`) strictly distinct from "genuinely zero open PRs" (`"[]"` parses to an empty list) — a failed query always renders as an explicit `QUERY FAILED` line in both the handoff file and the systemMessage, never silently as "None".
- Also flags truncation at the 100-result search-page cap (`PR_SEARCH_LIMIT`, chosen to match the search API's per-page max so "one search call" is literally one HTTP request), and flags any PR-bearing repo not in `org_repos.ALL_REPOS` — the inverse of BUG-08 (a repo the SSOT list doesn't know about yet), something the old per-repo loop could never detect since it only ever queried repos already in its own list.
- `org_repos.ALL_REPOS` stays imported (not re-derived) in `session_handoff.py`, keeping the existing `test_all_repos_is_org_repos_ssot` identity guard intact, per the caution that #1243 flags the *sibling* `validate_wave_audit.py` module for lacking the equivalent guard (that module is untouched by this PR).

Closes #1120

## Test plan
- [x] `PYTHONDONTWRITEBYTECODE=1 ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/test_session_handoff.py -v` — 18 passed
- [x] `PYTHONDONTWRITEBYTECODE=1 ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/ .claude/lib/tests/ -q` — 3489 passed, 647 subtests passed (full suite, no regressions)
- [x] Mutation-killed the three load-bearing new assertions by hand (confirmed each new test fails without the corresponding code): (1) collapsing `failed`/empty-result into the same state, (2) hardcoding `truncated = False`, (3) hardcoding `unknown_repos = ()`
- [x] `pre-commit run --all-files` — all commit-stage hooks pass
- [x] `pre-commit run --all-files --hook-stage pre-push` — mypy, pytest, pytest-skills, memory-budget, headcount-budget, doc-freshness, office-drift, mermaid-render all pass
- [x] `python3 .claude/lib/pre_commit_ci_sync.py .` — OK, pre-commit config mirrors CI
- [x] Manually ran `gh search prs --owner noorinalabs --state open --json number,title,repository --limit 100` against the live org to confirm the command shape and JSON schema used in the implementation

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/.claude/hooks/session_handoff.py
+++ b/.claude/hooks/session_handoff.py
@@ -20,12 +20,13 @@ import subprocess
 import sys
 from datetime import datetime, timezone
 from pathlib import Path
+from typing import NamedTuple
 
 REPO_ROOT = Path(__file__).resolve().parent.parent.parent
 
 _LIB_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), "..", "lib")
 sys.path.insert(0, os.path.abspath(_LIB_DIR))
-from org_repos import ALL_REPOS  # noqa: E402
+from org_repos import ALL_REPOS, ORG  # noqa: E402
 
 # Project memory is version-controlled in-repo (#732 relocation) so it travels
 # with a git pull. The handoff file itself stays gitignored — it's per-session
@@ -68,27 +69,111 @@ def _get_git_state() -> dict:
     }
 
 
-def _get_open_prs() -> list[str]:
-    """Get open PRs across all repos.
+# GitHub's search API pages at 100 results/request and caps at 1000 total.
+# We stay at exactly one page (--limit == the page max) so "one search call"
+# is literally one HTTP request, not gh silently paginating under the hood.
+PR_SEARCH_LIMIT = 100
 
-    Queries org_repos.ALL_REPOS (main#1118 / audit G6) — the previous hand-copied
-    list here omitted noorinalabs-isnad-ingest-platform (BUG-08), silently
-    querying only 7 of 8 org repos.
+
+class PrQueryResult(NamedTuple):
+    """Outcome of the single org-wide open-PR search.
+
+    `failed` and an empty `lines` list are DISTINCT states — a failed query
+    must never render the same as "the org genuinely has zero open PRs"
+    (main#1120). `truncated` flags a possible partial result: the result set
+    hit PR_SEARCH_LIMIT, so PRs beyond the cap (e.g. a large repo crowding
+    out others under `--sort updated`) may be missing without any command
+    error. `unknown_repos` are repos the search returned that are NOT in
+    org_repos.ALL_REPOS — the inverse of BUG-08 (a repo added org-side that
+    the SSOT list doesn't know about yet), something the old per-repo loop
+    could structurally never detect since it only ever queried repos already
+    in its list.
     """
-    prs = []
-    for repo in ALL_REPOS:
-        raw = _run(
-            f"gh pr list --repo noorinalabs/{repo} --state open --json number,title --limit 5",
-            timeout=15,
+
+    lines: list[str]
+    failed: bool
+    truncated: bool
+    unknown_repos: tuple[str, ...]
+
+
+def _get_open_prs() -> PrQueryResult:
+    """Get open PRs across the whole org in a single search call.
+
+    Replaces the previous 7-serial (pre-#1238) / 8-serial (post-#1238) loop
+    of `gh pr list --repo ...` subprocesses — worst case ~120s against a 30s
+    Stop-hook timeout (main#1120 / audit G8) — with one `gh search prs
+    --owner noorinalabs` call spanning all repos the token can see, so it no
+    longer needs a repo list to construct the query at all (pairs with G6).
+
+    Collapsing 8 independent calls into 1 changes the failure shape: before,
+    one repo's `gh pr list` timing out only dropped that repo's PRs (still
+    silently — `_run()` swallows failures into ""), leaving the other 7
+    repos' results intact. Now a single failed call drops the WHOLE org's PR
+    list in one shot. `_run()` returns "" on any failure, which is
+    byte-for-byte different from a genuine empty result ("[]" parses to an
+    empty list, not ""), so that distinction is preserved explicitly here
+    and surfaced up through `failed` — main() must render "query failed",
+    never "None", when `failed` is True.
+    """
+    raw = _run(
+        f"gh search prs --owner {ORG} --state open --sort updated "
+        f"--json number,title,repository --limit {PR_SEARCH_LIMIT}",
+        timeout=15,
+    )
+    if not raw:
+        return PrQueryResult(lines=[], failed=True, truncated=False, unknown_repos=())
+    try:
+        items = json.loads(raw)
+    except json.JSONDecodeError:
+        return PrQueryResult(lines=[], failed=True, truncated=False, unknown_repos=())
+
+    lines = []
+    repos_seen: set[str] = set()
+    for item in items:
+        try:
+            repo = item["repository"]["name"]
+            lines.append(f"  - {repo}#{item['number']}: {item['title']}")
+            repos_seen.add(repo)
+        except (KeyError, TypeError):
+            continue
+
+    truncated = len(items) >= PR_SEARCH_LIMIT
+    unknown_repos = tuple(sorted(repos_seen - set(ALL_REPOS)))
+    return PrQueryResult(
+        lines=lines, failed=False, truncated=truncated, unknown_repos=unknown_repos
+    )
+
+
+def _render_prs_section(result: PrQueryResult) -> list[str]:
+    """Build the '### Open PRs' markdown block from a PrQueryResult.
+
+    A failed query is never rendered as "None" — that would misreport an
+    unknown PR state as "the org has zero open PRs". Truncation at the
+    search cap and unexpected repos are both surfaced as explicit notes
+    rather than silently folded into (or dropped from) the list.
+    """
+    section = ["### Open PRs"]
+    if result.failed:
+        section.append(
+            "  - QUERY FAILED (gh search prs errored or timed out) — PR state "
+            "unknown, NOT confirmed empty"
         )
-        if raw:
-            try:
-                items = json.loads(raw)
-                for item in items:
-                    prs.append(f"  - {repo}#{item['number']}: {item['title']}")
-            except (json.JSONDecodeError, KeyError):
-                pass
-    return prs
+        return section
+    if not result.lines:
+        section.append("  - None")
+    else:
+        section.extend(result.lines)
+    if result.truncated:
+        section.append(
+            f"  - NOTE: hit the {PR_SEARCH_LIMIT}-result search cap — some open "
+            "PRs may be missing from this list"
+        )
+    if result.unknown_repos:
+        section.append(
+            "  - NOTE: PR(s) found in repo(s) not in org_repos.ALL_REPOS: "
+            f"{', '.join(result.unknown_repos)} — SSOT list may be stale"
+        )
+    return section
 
 
 def _get_open_issues() -> list[str]:
@@ -184,7 +269,7 @@ def main() -> None:
     date_str = now.strftime("%Y-%m-%d %H:%M UTC")
 
     git = _get_git_state()
-    prs = _get_open_prs()
+    pr_result = _get_open_prs()
     issues = _get_open_issues()
     ontology = _get_ontology_staleness()
     wave = _get_wave_status()
@@ -216,10 +301,7 @@ def main() -> None:
         "",
     ]
 
-    if prs:
-        lines.extend(["### Open PRs", *prs, ""])
-    else:
-        lines.extend(["### Open PRs", "  - None", ""])
+    lines.extend([*_render_prs_section(pr_result), ""])
 
     if issues:
         lines.extend(["### Open issues (noorinalabs-main)", *issues, ""])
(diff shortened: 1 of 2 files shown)


parametrization:
**Review: LGTM**

Reviewed at head `0d0205dc` via `gh api .../contents/...?ref=0d0205dc` (no local checkout). CI re-derived fresh via `statusCheckRollup` against the head SHA (not the PR number) — all 13 checks `SUCCESS`, `mergeable: MERGEABLE`.

**Verified:**
- `PrQueryResult` (`lines`, `failed`, `truncated`, `unknown_repos`) does make a failed query structurally distinct from a genuinely empty one. `_run()` returns `""` on any subprocess failure; `_get_open_prs()` treats both empty-`raw` and `JSONDecodeError` as `failed=True`, never falling through to `lines=[]`/`failed=False`. `_render_prs_section()` renders `failed=True` as an explicit `QUERY FAILED ... NOT confirmed empty` line before any other branch — confirmed by reading the render function and by running the test suite.
- Ran the full 18-test file locally (`ENVIRONMENT=test PYTHONDONTWRITEBYTECODE=1 python3 -m pytest .../test_session_handoff.py -v`): 18/18 pass, including `test_failed_query_is_distinct_from_empty_result`, `test_truncation_detected_at_cap`/`test_no_truncation_below_cap`, `test_unknown_repo_flagged`, `test_all_repos_is_org_repos_ssot`.
- Hand-ran all three reported mutations against the fetched source; each kills exactly its covering test, 17/18 pass otherwise:
  1. `failed=True→False` on the empty-`raw` path → kills `test_failed_query_is_distinct_from_empty_result` only.
  2. Hardcode `truncated = False` → kills `test_truncation_detected_at_cap` only.
  3. Hardcode `unknown_repos = ()` → kills `test_unknown_repo_flagged` only.
- `org_repos.ALL_REPOS` is a genuine `from org_repos import ALL_REPOS, ORG` import in `session_handoff.py` (not re-derived) — read `.claude/lib/org_repos.py` at head to confirm — so `test_all_repos_is_org_repos_ssot`'s `assertIs` tie to the SSOT holds.
- `unknown_repos` is meaningful, not noise: it's the free byproduct of grouping `repos_seen` from data already fetched for the PR list, correctly flags a PR-bearing repo missing from `org_repos.ALL_REPOS` — the structural inverse of BUG-08, something the old per-repo loop could never detect. Cheap, no extra call.
- Truncation logic (`len(items) >= PR_SEARCH_LIMIT`) is correctly conservative at exactly 100: since `--limit 100` caps `gh search prs` output at 100 regardless of true total, a result of exactly 100 is flagged `truncated` even in the edge case where the true total is exactly 100 (false-positive-safe, not false-negative — acceptable given it's disclosed as a NOTE, not silently dropped).
- No unbounded path: single subprocess with `timeout=15`, parsing is O(≤100) items, and the existing `THROTTLE_SECONDS = 300` gate still bounds how often this Stop hook does any of this work at all.

**The eventual-consistency risk (raised for this review):** confirmed real in principle — `gh search prs` reads GitHub's search index, architecturally distinct from the live `gh pr list` REST/GraphQL path used everywhere else in this repo's tooling (`rg --hidden 'gh search'` over `.claude/lib`+`.claude/hooks` at HEAD shows no prior *implementation* call site — this is the first). However:
- GitHub's docs don't state an indexing-lag SLA for issue/PR search (unlike *code* search, which is documented to lag) — only "indexed when created or modified," no bound given either way.
- Live spot-check just now: `gh search prs --owner noorinalabs --state open --json number,title,repository --limit 100` returned the exact same 3 PRs as looping `gh pr list --repo ...` across all 8 `org_repos.ALL_REPOS` repos individually (`noorinalabs-main#1240/#1247/#1258`, zero elsewhere in either method). No drift observed.
- This hook is explicitly advisory (module docstring: "Exit codes: 0 — always \\(advisory, never blocks\\)") feeding a handoff for orientation, not a safety-critical decision — and it replaces a strictly worse failure mode (BUG-08's silent per-repo drop, plus a ~120s worst case against a 30s Stop-hook timeout). The tradeoff is directionally correct and was the audit item's own prescribed fix (#1120: "Replace with one `gh search prs --owner noorinalabs`...").
- Filed #1259 (tech-debt/process/phase-10, no wave label) asking for a one-line doc addition to `_get_open_prs()` noting the index-vs-live distinction, so a future reader doesn't treat the handoff's PR section as always-instantaneous. Documentation-only, not blocking.

**Could not fully verify:** the "one HTTP request" framing rests on `gh` CLI's internal pagination behavior (per-page size = `min(limit, apiMax)`, no further pages issued when `limit` fits in one page) — I did not instrument `gh`'s own HTTP layer to count requests; I'm relying on the documented per-page cap (100) matching `PR_SEARCH_LIMIT` plus a live functional spot-check, not a wire-level trace.

Must-fix: None
Tech-debt: #1259 (search-index-vs-live-list caveat — documentation only)

---
Requestor: Weronika Zielinska
Requestee: Santiago Ferreira
RequestOrReplied: Approved
TechDebt: #1259

parametrization:
**Merge-gate review — Approved.**

Reviewed at head `0d0205dc`, read via `gh api repos/.../contents/<path>?ref=0d0205dc` (no local checkout). All verification below was executed, not read off the PR description.

## CI at head

14/14 checks `completed` / `success`, each confirmed pinned to `0d0205dc` via `gh api repos/noorinalabs/noorinalabs-main/commits/0d0205dc.../check-runs` — not just the rollup summary. `MERGEABLE` / `CLEAN`.

## The four merge-gate questions

**1. Does a failed search render as "no open PRs"? No — verified by execution, not by reading.**

I exercised the real failure path rather than trusting the mock: `GH_TOKEN=<invalid> gh search prs --owner noorinalabs --state open --sort updated --json number,title,repository --limit 100` exits 1 with **stdout length 0**. That is the precise input `_run()` converts to `""`, which `_get_open_prs()` maps to `failed=True`. So the structural distinction is anchored in the actual observed failure shape of the actual command, not an assumed one. `_render_prs_section()` then emits `QUERY FAILED ... NOT confirmed empty`, and returns early so no `None` can follow it.

This is the right call and it is the thing that makes collapsing 8 calls into 1 acceptable. Note what it buys relative to `main`: the old loop swallowed a per-repo failure into *silence* — a repo whose `gh pr list` timed out simply vanished from the list with nothing said. The new code is strictly louder on failure even though the blast radius per failure is larger.

**2. The 100 cap — established empirically, including the 101-PR state.**

I could not manufacture 101 open PRs in this org, so I established the behavior against a real org that has them. `gh search prs --owner microsoft --state open --sort updated --limit 100` returns **exactly 100** items. I then fed that real 100-item payload through `_get_open_prs()` and `_render_prs_section()` at head:

```
items: 100  failed: False  truncated: True  unknown_repos: 32
```

So a >100-PR org state does produce the truncation NOTE — confirmed end to end, not inferred from `>=`.

On loudness, the two surfaces differ and I think the split is defensible: in the `systemMessage` the flag is on the **header** line (`Open PRs: 100 (truncated at cap)`), before the 5 shown. In the file it is line 102 of the section, after 100 PR lines. A future session reads the whole file, so a footer note is legible there; a human skimming the terminal gets it up front. Not loud enough to block on, especially at an org that currently has 3 open PRs.

Also worth recording: `PR_SEARCH_LIMIT = 100` is a **widening**, not a narrowing. The code it replaces used `--limit 5` **per repo** — a silent per-repo truncation at 5 with no flag of any kind. The new cap is 100 org-wide *and* announces itself.

The comment's claim that this stays one HTTP request holds. `GH_DEBUG=api` on the exact command shows a single request:
`GET /search/issues?page=1&per_page=100&q=state%3Aopen+type%3Apr+user%3Amicrosoft&sort=updated`.

**3. Search-vs-list consistency — matched on the live org; residual risk accepted, already documented in #1259.**

I ran both sides at review time. The PR's exact search command returned 3 PRs; looping `gh pr list --repo noorinalabs/<repo> --state open --limit 100` over all 8 `ALL_REPOS` entries returned the same 3 (`noorinalabs-main#1240`, `#1247`, `#1258`). Sampling agreement still proves nothing about lag in general — but one detail does carry a little signal: **#1258 itself**, opened during this session, was already in the search index. So the indexing delay is not on the order of hours for this org's traffic.

I did not attempt to measure the lag bound (that would require opening and immediately polling a throwaway PR, which I am not going to do on a live org mid-wave). Stated as unverified: the worst-case index lag. Given the hook is advisory, never blocks, and the alternative is the ~120s-vs-30s-timeout failure it replaces, this is the right trade. Weronika's #1259 already asks for the index-vs-live caveat comment; I am not duplicating it, and I agree it is documentation-only.

**4. `unknown_repos` earns its place — with one honest limit.**

It is the inverse of BUG-08 and it is a detection the old per-repo loop was structurally incapable of making, since it only ever asked about repos already in its own list. At this org's scale (8 curated repos) a hit is a real signal, not noise: `--owner noorinalabs` scopes the query, so any returned repo outside `ALL_REPOS` genuinely is an org repo the SSOT does not know about. That is squarely in the #1254 spirit — it makes an unknown state visible instead of letting it look clean.

The limit, which the docstring already scopes correctly: it only fires for a **PR-bearing** unknown repo inside the top-100 window. A new repo with no open PRs stays invisible. Best-effort detector, not a guarantee — fine, as long as nobody later mistakes its silence for proof the SSOT is current.

## Mutation testing — re-run, not accepted

Isolated tree outside the repo, PR file plus `org_repos.py` at `0d0205dc`, `PYTHONDONTWRITEBYTECODE=1 ENVIRONMENT=test`. Baseline 18 passed.

Your three, each applied independently — all die, and each dies on the assertion it should:

| mutation | outcome |
|---|---|
| `failed=True` → `failed=False` on both failure returns | killed — `test_failed_query_is_distinct_from_empty_result`, `test_malformed_json_treated_as_failure` |
| `truncated = len(items) >= PR_SEARCH_LIMIT` → `truncated = False` | killed — `test_truncation_detected_at_cap` |
| `unknown_repos = ...` → `()` | killed — `test_unknown_repo_flagged` |

Five more of my own:

| mutation | outcome |
|---|---|
| `_render_prs_section` failed branch renders `"  - None"` | killed |
| SSOT bypass: local re-derived list omitting `isnad-ingest-platform` | killed (but see below) |
| `main()` failure line → `"Open PRs: 0"` | **survived** |
| drop `--state open` from the query | **survived** |
| `PR_SEARCH_LIMIT = 100` → `5` | **survived** |

The guards you added are real. Three gaps, all test-strength rather than behavior, none blocking:

- `main()` has no test at all, so the `systemMessage` copy of the failure distinction — the one the operator reads at exit — can be regressed to a confident `Open PRs: 0` with the suite green. Filed as #1261, with a suggestion I think is better than adding a test: have `main()` derive its line from a shared helper so the second copy stops existing.
- The SSOT bypass dies only by fixture luck. `test_parses_items_across_repos` happens to use `isnad-ingest-platform` as a fixture value; re-derive a local list omitting `landing-page` instead and all 18 pass. `assertIs(hook.ALL_REPOS, org_repos.ALL_REPOS)` guards the module **binding**, not that `_get_open_prs()` uses it. That is the #1243 shape one layer in — the guard is present but binds less than a reader assumes. Also #1261. To be explicit on your stated concern: the import is intact, the identity assertion still binds, and nothing in this PR weakened it.
- Nothing pins `--state open` or the value `100`; the truncation test asserts against the symbol, so the constant can drift silently. Minor, folded into #1261.

## Scope and blast radius

`_get_open_prs()`'s return type changed from `list[str]` to `PrQueryResult`. I swept for external consumers — `rg --hidden --no-ignore` for `_get_open_prs` / `PrQueryResult` / `import session_handoff` across the repo. The only caller is `main()` in the same module. No machine parser of the handoff's `### Open PRs` section exists either, so the new `QUERY FAILED` and `NOTE:` lines cannot break a downstream reader. #1120's stated Verify ask ("`test_session_handoff.py` asserts the issued command shape") is satisfied by `test_issues_exactly_one_search_call`, which also asserts the negative (`gh pr list` absent).

## One finding outside this PR

While checking whether the fix was applied consistently I found that `.claude/skills/handoff/SKILL.md` still hardcodes a **7-repo** loop that omits `noorinalabs-isnad-ingest-platform` — BUG-08, still live, in the surface `CLAUDE.md` recommends for a *richer* handoff. #1118/G6 explicitly asked its sweep to "enumerate every literal repo list"; it caught the Python ones and missed the markdown one, and no linter or type-checker can see it. Filed as #1260. Not yours, not a change request here — but it means the automatic Stop hook you just fixed is now more correct than the manual command beside it.

## Verdict

The central risk of this change — one call failing where eight used to fail independently — is handled deliberately, and the handling survives being exercised against the real command's real failure output rather than a mock's assumption. Approving. Debt filed as #1260 and #1261; #1259 already stands from the second reviewer.

---

Requestor: Nadia Khoury
Requestee: Santiago Ferreira
RequestOrReplied: Approved
TechDebt: #1260, #1261

Issue #2, opened 2026-04-05
Meta: design-system Phase 1 Wave 1 — scaffolding + extraction


parametrization:
## Summary

Coordinate Phase 1 Wave 1 for `noorinalabs-design-system`. This is the initial implementation — scaffolding the package, extracting tokens/components/icons/styles from isnad-graph, and setting up build/publish.

### Critical Path (Sequential)
1. noorinalabs-design-system#1 — Project scaffolding (package.json, TS, Tailwind, Storybook) — **Maricel Reyes**
2. noorinalabs-design-system#2 — Extract design tokens from isnad-graph — **Beren Yildiz** (blocked by #1)
3. noorinalabs-design-system#3 — Extract UI components from isnad-graph — **Astrid Lindqvist** (blocked by #1, ideally #2)

### Parallel (after scaffolding)
4. noorinalabs-design-system#4 — Extract icon library from isnad-graph — **Keanu Tama** (blocked by #1)
5. noorinalabs-design-system#5 — Extract global styles and utilities — **Beren Yildiz** (blocked by #1)
6. noorinalabs-design-system#6 — Brand documentation and guidelines — **Beren Yildiz** (blocked by #1)
7. noorinalabs-design-system#7 — Package build and publish setup — **Keanu Tama** (blocked by #1)

## Sequencing

1. Create `deployments/phase1/wave-1` branch from `main`
2. #1 (scaffolding) must land first — all others depend on it
3. After #1 merges: #2, #4, #5, #6, #7 can proceed in parallel
4. #3 (components) ideally waits for #2 (tokens) but can start structure

## Cross-Repo Dependencies

- **Reads from isnad-graph:** Issues #2-#5 extract code from `noorinalabs-isnad-graph/frontend/`. Read-only — no modifications to isnad-graph.
- **Downstream consumer:** `noorinalabs-landing-page` will consume this package. Landing page implementation is blocked until scaffolding (#1) and tokens (#2) are complete.

## Repo Manager

Maeve Callahan (design-system manager)

Issue #1245, opened 2026-08-03
tech-debt: dispatcher discards aggregated advisories when a later hook blocks (burns the librarian once-per-session throttle)


parametrization:
## Problem

`dispatcher.py` accumulates non-blocking `systemMessage`s into `warnings`, but on the first blocking hook it does `print(json.dumps(result)); sys.exit(2)` — **discarding every warning collected so far**.

Combined with `enforce_librarian_consulted`'s once-per-session throttle (#1022), this silently burns the session's single advisory. `check()` writes the throttle marker *before* returning the advisory, so when a later hook in the same dispatch blocks, the marker reads "already emitted" while the operator never saw the message — and it will not be re-emitted for the rest of the session.

Before #1239 this could not happen on `Edit`/`Write`/`NotebookEdit`: the advisory and the gate ran as two independent `settings.json` subprocesses, so the advisory always printed.

## Reproduction (verified live, worktree at `ff39d453`)

Seed a sentinel entry for a non-allowlisted path, then dispatch a same-path `Edit` in a fresh session:

```
$ printf '{"hook_event_name":"PreToolUse","tool_name":"Edit","session_id":"S","cwd":"<wt>",
   "tool_input":{"file_path":"/home/parameterization/code/noorinalabs-main/README.md"}}' \
  | python3 .claude/hooks/dispatcher.py
{"decision": "block", "reason": "BLOCKED: prior Edit on ... README.md returned an error ..."}
exit=2                                   # advisory systemMessage absent from output

$ python3 -c "import enforce_librarian_consulted as e; \
   p=e._advisory_throttle_path('S'); print(e._advisory_already_emitted(p))"
True                                     # throttle consumed anyway
```

## Fix options

1. Emit the aggregated warnings alongside the block (merge `systemMessage` into the block dict before printing) — preserves both signals, one-line change in `dispatcher.py`.
2. Have `enforce_librarian_consulted` mark the throttle only once the message is known to have been delivered (harder — the hook cannot see the dispatcher's outcome).

Option 1 is preferred and applies to `post_dispatcher.py`'s equivalent path if it has the same shape.

## Acceptance

- [ ] A dispatch where hook A returns a `systemMessage` and hook B blocks emits both the block reason and A's message
- [ ] Mutation-tested: reverting the fix fails a named test
- [ ] Same-shape check applied to `post_dispatcher.py`

[...]

Issue #1244, opened 2026-08-03
tech-debt: validate_edit_completion loses Annunaki block logging on the dispatcher path


parametrization:
## Problem

`validate_edit_completion.py` calls `annunaki_log.log_pretooluse_block(...)` **only inside `main()`**, never inside `check()`. Every other blocking hook the PreToolUse dispatcher routes logs from `check()` — verified for `block_bare_grep` (call site sits between `def check()` and `def main()`), `block_no_verify`, and `validate_commit_identity`.

Before #1239, `Edit` / `Write` / `NotebookEdit` ran this hook through its standalone `settings.json` command entry, i.e. through `main()`, so a same-path re-edit block **was** recorded to the Annunaki error store. #1239 routes those three matchers through `dispatcher.py`, which calls `check()` and then prints + exits 2 itself without any Annunaki logging. Net effect: the block still fires (verified end-to-end), but it becomes **invisible to `/annunaki`**.

The `Bash` surface already had this gap (this hook has been in `hooks.pre_bash` for a while, and `dispatcher.py` never logged) — consistent with the store having zero `validate_edit_completion` records:

```
$ tally of type=pretooluse_block records in .claude/annunaki/errors.jsonl
block_bare_grep/Bash 119 · validate_commit_identity/Bash 45 · validate_review_comment_format/Bash 36
smart_grep_ontology/Bash 15 · ... · validate_edit_completion  0
```

After #1239 the only remaining logged surface for this hook is `SendMessage` (still standalone-registered).

## Why it matters

Annunaki is how a recurring gate-block gets noticed and turned into automation. A gate that blocks but never logs is a gate whose firing rate is unobservable — the tool-error-soft-accept class it guards is exactly the kind we want frequency data on.

## Fix

Move the `log_pretooluse_block(...)` call from `main()` into `check()` in `.claude/hooks/validate_edit_completion.py` (guarded on `result.get("decision") == "block"`), matching the `block_bare_grep` shape, and have `main()` rely on `check()` for the logging so the `SendMessage` path does not double-log. Add a test asserting a block through `check()` appends an Annunaki record.

## Acceptance

- [ ] A PreToolUse block from `check()` for `Bash`, `Edit`, `Write`, `NotebookEdit` appends one `type: pretooluse_block` record
- [ ] The `SendMessage` standalone path logs exactly once (no duplicate)
- [ ] Mutation-tested: removing the `check()` log call fails a named test

[...]

PR #1263, opened 2026-08-03
audit(G10): per-edit ledger IO: skip no-op writes, cache check-ignore per directory

### LINKED ISSUE #1122: audit(G10): per-edit ledger IO: skip no-op writes, cache check-ignore
Source: CODE_AUDIT.md §10 (noorinalabs-main @ 3bdd620) — item G10.

**G10 — Per-edit ledger IO** · Type: perf · Effort: S
`post_file` re-reads + re-writes `ontology/checksums.json` (103 KB) and the generic-prompt ledgers (35 KB) on every Edit/Write plus a `git check-ignore` subprocess. Two safe wins: skip the write when the SHA equals `last_tracked` (a no-op re-save currently rewrites 103 KB), and cache `check-ignore` per directory per process. Keep `ensure_ascii=False` + atomic replace (byte-stability is load-bearing, tests pin it). Append-only JSONL is a separate, measured decision.
*After:* **≈ −50–70% of post_file IO** on edit bursts. *Verify:* `test_ontology_tracker.py` (602), `test_checksums_io.py` (225).

Index row (§13): P | −50–70% post_file IO | S | —

parametrization:
## Summary

- `ontology_tracker.check()` skips the `checksums.json` write (not the read — still needed to compare) when the file's content SHA already equals `last_tracked`, i.e. a genuine no-op re-save. Previously every Edit/Write unconditionally rewrote the full 103 KB file.
- `_is_git_ignored`'s per-file `git check-ignore` cache (#1039) is extended with a per-directory cache. `git check-ignore` (without `-q`) accepts multiple pathspecs and echoes back which matched, so the FIRST file seen in a directory resolves both "is this file ignored" and "is its directory ignored" in one subprocess call — same cost as before. Per gitignore(5) ("It is not possible to re-include a file if a parent directory of that file is excluded"), a directory-ignored verdict is authoritative for every later file in that directory: zero subprocess calls. A not-ignored directory verdict is *not* reused for individual files (a filename pattern can still exclude one), so that case still checks per-file as before.
- Both caches are process-lifetime module dicts; nothing invalidates them mid-invocation (a PostToolUse hook is a fresh, short-lived, one-Edit-per-call subprocess, and `.gitignore` can't change mid-call).

Closes #1122

## Verification

- `pre-commit run --all-files` and `--hook-stage pre-push` both green (ruff-format/lint, actionlint, cspell, checksums-ascii, mypy, `pytest` incl. `.claude/hooks/tests` + `.claude/lib/tests` — 3493 passed, 647 subtests).
- Real (non-mocked) end-to-end exercise of the hook via stdin, run against a throwaway `git init` fixture repo *outside* this worktree (edits inside this worktree itself are filtered by the pre-existing linked-worktree skip, unrelated to this change):
  - First `Write` on a new tracked file → `exit 0`, `checksums.json` gains the new entry.
  - Identical re-`Write` (no-op) → `exit 0`, `checksums.json` byte-identical (write skipped).
  - Content actually changed → `exit 0`, `checksums.json` diffs again (`last_tracked` + `tracked_at` updated).
  - Directory-cache: a real, un-mocked Python process call showed 1 subprocess call for the first file in a gitignored directory and 0 for a second file in the same directory.
- Mutation-tested both new predicates directly against the source (not just the tests): an "always skip" mutation, an "ignore SHA, key on path presence" mutation, and an "over-broad directory cache" mutation were each applied in turn and confirmed to fail the corresponding new test, then reverted.

## Scope notes

- Does not touch `validate_edit_completion.py` or `post_dispatcher.py`'s logging path — no interaction with #1244 (missing Annunaki logging from `check()`) or #1245 (dispatcher discards advisories on a later block).

## Test plan

- [x] `pre-commit run --all-files`
- [x] `pre-commit run --all-files --hook-stage pre-push`
- [x] New/mutation tests in `.claude/hooks/tests/test_ontology_tracker.py`
- [x] Real synthetic PostToolUse exercise (exit codes measured, not reasoned about)

Co-Authored-By: Wanjiku Mwangi <parametrization+Wanjiku.Mwangi@gmail.com>

### DIFF
--- a/.claude/hooks/ontology_tracker.py
+++ b/.claude/hooks/ontology_tracker.py
@@ -335,13 +335,29 @@ def _run_check_ignore(git_root: Path, pathspecs: list[str]) -> set[str]:
     is what introduced the exposure, so the pin restores the property that
     change gave up. Do not remove it without replacing the matching scheme.
 
-    ``encoding="utf-8", errors="replace"`` rather than ``text=True`` for the
-    same reason: with the pin in place raw UTF-8 bytes now reach the decoder,
-    and ``text=True`` would decode with the locale encoding under
-    ``errors='strict'`` — so a ``LC_ALL=C`` runner would raise
-    ``UnicodeDecodeError``, which is NOT in the ``except`` clause below and
-    would escape ``check()``, breaking this hook's "exit 0 — always"
-    contract. ``errors="replace"`` makes that unraisable.
+    ``encoding="utf-8", errors="surrogateescape"`` rather than ``text=True``,
+    and the reason is the FILENAME BYTES, not the locale (main#1263 review,
+    Weronika Zielinska — an earlier version of this docstring blamed a
+    ``LC_ALL=C`` runner raising ``UnicodeDecodeError``; that is false, PEP 538
+    C-locale coercion yields UTF-8 anyway, and reverting to ``text=True``
+    leaves the suite green under ``LC_ALL=C``. Do not restore that rationale).
+
+    The real trigger: a POSIX filename is a byte string and need not be valid
+    UTF-8. The caller's pathspec comes from ``os.fsdecode``, which maps
+    undecodable bytes to lone surrogates (``b"\\xe9.log"`` -> ``"\\udce9.log"``).
+    For set-membership below to work, git's echoed stdout must decode back to
+    that SAME string. Only ``surrogateescape`` does:
+
+    - ``text=True`` (strict) can raise ``UnicodeDecodeError``, which is not in
+      the ``except`` clause and would escape ``check()``, breaking this hook's
+      "exit 0 — always" contract.
+    - ``errors="replace"`` cannot raise but is lossy — the byte decodes to
+      U+FFFD, which does not equal the caller's surrogate, so a genuinely
+      ignored file reads as NOT ignored. That is the same fail-open defect the
+      ``core.quotePath`` pin above exists to fix, one level narrower, and it
+      was shipped here briefly before review caught it.
+    - ``surrogateescape`` is non-raising AND byte-exact round-trips, so it
+      strictly dominates both.
 
     Deliberately omits ``-q`` (which would suppress that stdout) so a single
     call can answer more than one question at once (#1122) — the caller
@@ -365,7 +381,7 @@ def _run_check_ignore(git_root: Path, pathspecs: list[str]) -> set[str]:
             timeout=5,
             env=_hermetic_git_env(),
             encoding="utf-8",
-            errors="replace",
+            errors="surrogateescape",
         )
     except (OSError, subprocess.TimeoutExpired):
         return set()  # Subprocess failed — fail open.
--- a/.claude/hooks/ontology_tracker.py
+++ b/.claude/hooks/ontology_tracker.py
@@ -322,12 +322,31 @@ def _run_check_ignore(git_root: Path, pathspecs: list[str]) -> set[str]:
     """Run ``git check-ignore`` for one or more pathspecs against ``git_root``.
 
     Returns the subset of ``pathspecs`` that ARE ignored, as the exact
-    strings passed in (git echoes back whichever supplied pathspec matched,
-    one per line, preserving the caller's original string — verified against
-    real git, not assumed). Deliberately omits ``-q`` (which would suppress
-    that stdout) so a single call can answer more than one question at once
-    (#1122) — the caller distinguishes "ignored" from "not ignored" by
-    set-membership instead of by exit code alone.
+    strings passed in: git echoes back whichever supplied pathspec matched,
+    one per line, and ``core.quotePath=false`` is pinned on the invocation
+    so that echo is the caller's ORIGINAL string rather than a C-quoted
+    rendering of it. That pin is load-bearing, not cosmetic (main#1265):
+    under git's default ``core.quotePath=true`` any pathspec containing a
+    non-ASCII byte comes back quoted and escaped (``سند.md`` echoes as
+    ``"\\330\\263\\331\\206\\330\\257.md"``), which equals nothing the
+    caller passed in, so set-membership below reports a genuinely-ignored
+    path as NOT ignored. The pre-#1122 code read only ``-q``'s exit status
+    and was encoding-independent by construction; matching on echoed text
+    is what introduced the exposure, so the pin restores the property that
+    change gave up. Do not remove it without replacing the matching scheme.
+
+    ``encoding="utf-8", errors="replace"`` rather than ``text=True`` for the
+    same reason: with the pin in place raw UTF-8 bytes now reach the decoder,
+    and ``text=True`` would decode with the locale encoding under
+    ``errors='strict'`` — so a ``LC_ALL=C`` runner would raise
+    ``UnicodeDecodeError``, which is NOT in the ``except`` clause below and
+    would escape ``check()``, breaking this hook's "exit 0 — always"
+    contract. ``errors="replace"`` makes that unraisable.
+
+    Deliberately omits ``-q`` (which would suppress that stdout) so a single
+    call can answer more than one question at once (#1122) — the caller
+    distinguishes "ignored" from "not ignored" by set-membership instead of
+    by exit code alone.
 
     Fails OPEN — returns an empty set (nothing reported ignored) — on any
     subprocess error or timeout, or on any exit code other than 0 (at least
@@ -340,12 +359,13 @@ def _run_check_ignore(git_root: Path, pathspecs: list[str]) -> set[str]:
     """
     try:
         result = subprocess.run(
-            ["git", "check-ignore", "--", *pathspecs],
+            ["git", "-c", "core.quotePath=false", "check-ignore", "--", *pathspecs],
             cwd=str(git_root),
             capture_output=True,
             timeout=5,
             env=_hermetic_git_env(),
-            text=True,
+            encoding="utf-8",
+            errors="replace",
         )
     except (OSError, subprocess.TimeoutExpired):
         return set()  # Subprocess failed — fail open.
--- a/.claude/hooks/ontology_tracker.py
+++ b/.claude/hooks/ontology_tracker.py
@@ -337,10 +337,21 @@ def _run_check_ignore(git_root: Path, pathspecs: list[str]) -> set[str]:
 
     ``encoding="utf-8", errors="surrogateescape"`` rather than ``text=True``,
     and the reason is the FILENAME BYTES, not the locale (main#1263 review,
-    Weronika Zielinska — an earlier version of this docstring blamed a
-    ``LC_ALL=C`` runner raising ``UnicodeDecodeError``; that is false, PEP 538
-    C-locale coercion yields UTF-8 anyway, and reverting to ``text=True``
-    leaves the suite green under ``LC_ALL=C``. Do not restore that rationale).
+    Weronika Zielinska). An earlier version of this docstring blamed a
+    ``LC_ALL=C`` runner raising ``UnicodeDecodeError``. That was false, and
+    for a subtler reason than "coercion": ``LC_ALL=C python3`` auto-enables
+    **UTF-8 Mode (PEP 540)** — ``sys.flags.utf8_mode == 1`` while ``LC_CTYPE``
+    stays ``C`` — so PEP 538 C-locale *coercion* never runs at all. That is
+    why ``PYTHONCOERCECLOCALE=0`` does not restore ASCII either; it defeats
+    538, not 540. Only ``PYTHONUTF8=0 PYTHONCOERCECLOCALE=0`` yields
+    ``ANSI_X3.4-1968``. Do not restore the locale rationale.
+
+    Historical note on how that was caught, because it dates: when measured
+    at ``2c113e7`` the suite was green under ``LC_ALL=C`` with ``text=True``,
+    which is what falsified the claim. That is no longer reproducible —
+    ``test_invalid_utf8_filename_is_detected`` below now catches ``text=True``
+    at ANY locale, because the trigger is the filename bytes rather than the
+    environment. The fixture, not the locale, is what pins this line.
 
     The real trigger: a POSIX filename is a byte string and need not be valid
     UTF-8. The caller's pathspec comes from ``os.fsdecode``, which maps
--- a/.claude/hooks/ontology_tracker.py
+++ b/.claude/hooks/ontology_tracker.py
@@ -123,7 +123,28 @@ Owning-repo check-ignore (#1039):
   pathspecs matched, so the FIRST file seen in a directory pays one
   subprocess call that answers "is this file ignored" AND "is its directory
   ignored" simultaneously — no additional subprocess versus the pre-#1122
-  single-file check. Invalidation: none needed — both caches are
+  single-file check.
+
+  The directory pathspec passed to ``git check-ignore`` is the BARE relative
+  directory name, with NO trailing slash (main#1263 review finding, fixed
+  before merge). A trailing slash turns the pathspec into a literal STRING
+  that a contents-only pattern like ``data/raw/*`` matches directly (git
+  echoes back the exact string ``"data/raw/"`` as a match), which is a
+  different fact from the directory ITSELF being excluded — the bare
+  ``"data/raw"`` does not match that same pattern. Using the slash version
+  would cache a false "directory ignored" and silently mis-skip any file a
+  ``!`` rule re-includes inside it (e.g. the ``dir/*`` + ``!dir/**/.gitkeep``
+  idiom used by ``noorinalabs-isnad-ingest-platform``) — under-tracking,
+  the one direction this function must never risk. The bare name still
+  correctly answers "ignored" for a genuinely directory-excluding pattern
+  (``build/`` matches ``"build"`` too) and for a NESTED directory swept up
+  by a contents-only parent pattern (gitignore(5)'s no-re-include-under-an-
+  excluded-parent rule then really does apply to everything beneath it), so
+  the cache short-circuit stays sound in both directions — see
+  ``GitCheckIgnoreDirectoryCacheTests`` in the test module for the exact
+  fixtures.
+
+  Invalidation: none needed — both caches are
   module-level dicts scoped to this one short-lived process (a fresh
   Edit/Write hook invocation starts with empty caches), and the on-disk
   ``.gitignore`` rules cannot change mid-invocation, so nothing can go stale
@@ -388,7 +409,27 @@ def _is_git_ignored(resolved_path: Path) -> bool:
     # Neither the file nor its directory is cached yet: one subprocess call
     # answers both questions and seeds both caches, so every LATER file
     # under this same directory is a cache hit instead of a new subprocess.
-    dir_spec = f"{dir_rel_str}/"
+    #
+    # NO trailing slash on the directory pathspec (main#1263 review finding):
+    # a pattern like `data/raw/*` matches the literal STRING "data/raw/" —
+    # `git check-ignore -- data/raw/ ...` echoes it back as ignored even
+    # though the directory itself is not excluded, only its immediate
+    # contents are (minus whatever a later `!` re-include exempts). That
+    # false "directory ignored" verdict would then be cached and wrongly
+    # applied to every later file in the directory, INCLUDING one a `!`
+    # rule legitimately re-includes — a silent under-tracking regression.
+    # Querying the bare directory name instead asks git the real question
+    # ("is `data/raw` itself excluded?"): a genuinely directory-excluding
+    # pattern (`build/`) still matches the bare name, but a
+    # contents-only pattern (`data/raw/*`) does not, so `dir_ignored` stays
+    # False and each file is still checked on its own merits (same
+    # subprocess cost as before this cache existed). A NESTED directory
+    # whose own path is swept up by the contents-only pattern (e.g.
+    # `data/raw/sub` under `data/raw/*`) still correctly comes back
+    # ignored — gitignore(5)'s no-re-include-under-an-excluded-parent rule
+    # then genuinely applies to everything beneath it, so the cached True
+    # is not a false positive there.
+    dir_spec = dir_rel_str
     matched = _run_check_ignore(git_root, [dir_spec, rel_str])
     dir_ignored = dir_spec in matched
     file_ignored = rel_str in matched
--- a/.claude/hooks/ontology_tracker.py
+++ b/.claude/hooks/ontology_tracker.py
@@ -101,6 +101,34 @@ Owning-repo check-ignore (#1039):
   calls within a single test run or a future batch invocation, not
   cross-invocation caching (there is no daemon to cache across).
 
+  ``_DIR_CHECK_IGNORE_CACHE`` (#1122) additionally memoizes the *containing
+  directory's* own check-ignore verdict, per (repo, relative-directory), same
+  process lifetime. Per gitignore(5) — "It is not possible to re-include a
+  file if a parent directory of that file is excluded" — a directory that
+  ``git check-ignore`` reports as ignored makes EVERY file beneath it ignored
+  too, no exceptions possible. So once a directory resolves to ignored, any
+  later file under it is a cache hit with zero subprocess calls: real
+  savings within a burst of edits/tests touching the same generated or
+  vendored subtree (e.g. a child repo's own ``.venv/``/``dist/``/``coverage/``
+  that ``SKIP_PATTERNS`` doesn't already substring-catch). A directory
+  verdict of NOT-ignored is not similarly reusable — it says nothing about a
+  specific file, since a filename pattern (``*.secret``) can still exclude
+  one file inside an otherwise-untouched directory — so that case still
+  falls through to a per-file check, same subprocess cost as before this
+  cache existed.
+
+  The directory verdict is resolved for free alongside the file's own
+  verdict: ``git check-ignore`` (without ``-q``) accepts more than one
+  pathspec and echoes back on stdout exactly which of the supplied
+  pathspecs matched, so the FIRST file seen in a directory pays one
+  subprocess call that answers "is this file ignored" AND "is its directory
+  ignored" simultaneously — no additional subprocess versus the pre-#1122
+  single-file check. Invalidation: none needed — both caches are
+  module-level dicts scoped to this one short-lived process (a fresh
+  Edit/Write hook invocation starts with empty caches), and the on-disk
+  ``.gitignore`` rules cannot change mid-invocation, so nothing can go stale
+  within the cache's own lifetime.
+
 Input Language:
   Fires on:      PostToolUse Edit, Write
   Matches:       Edit/Write whose `file_path` is non-empty AND resides inside
@@ -129,6 +157,12 @@ CHECKSUMS_FILE = REPO_ROOT / "ontology" / "checksums.json"
 # Memoizes (git_root, repo_relative_path) -> ignored? for the life of this
 # process. See "Owning-repo check-ignore (#1039)" in the module docstring.
 _GIT_CHECK_IGNORE_CACHE: dict[tuple[str, str], bool] = {}
+
+# Memoizes (git_root, repo_relative_directory) -> ignored? for the life of
+# this process (#1122). See "Owning-repo check-ignore (#1039)" in the module
+# docstring for why a directory-ignored verdict is authoritative for every
+# file beneath it, and why a not-ignored verdict is NOT similarly reusable.
+_DIR_CHECK_IGNORE_CACHE: dict[tuple[str, str], bool] = {}
 
 # Shared read/write helpers (#1042): both this hook and the /ontology-rebuild
 # resolver's `mark-resolved` CLI go through checksums_io so neither has to
@@ -263,6 +297,44 @@ def _hermetic_git_env() -> dict[str, str]:
     return {k: v for k, v in os.environ.items() if not k.startswith("GIT_")}
 
 
+def _run_check_ignore(git_root: Path, pathspecs: list[str]) -> set[str]:
+    """Run ``git check-ignore`` for one or more pathspecs against ``git_root``.
+
+    Returns the subset of ``pathspecs`` that ARE ignored, as the exact
+    strings passed in (git echoes back whichever supplied pathspec matched,
+    one per line, preserving the caller's original string — verified against
+    real git, not assumed). Deliberately omits ``-q`` (which would suppress
+    that stdout) so a single call can answer more than one question at once
+    (#1122) — the caller distinguishes "ignored" from "not ignored" by
+    set-membership instead of by exit code alone.
+
+    Fails OPEN — returns an empty set (nothing reported ignored) — on any
+    subprocess error or timeout, or on any exit code other than 0 (at least
+    one match) / 1 (no match, e.g. every pathspec is genuinely not ignored).
+    Exit 128 is a fatal git error (e.g. not actually a git repo); folding it
+    into "nothing ignored" fails open the same way a single-path check
+    always has. See "Owning-repo check-ignore (#1039)" in the module
+    docstring for why fail-open (never skip on doubt) is the deliberate,
+    one-sided policy here.
+    """
+    try:
+        result = subprocess.run(
+            ["git", "check-ignore", "--", *pathspecs],
+            cwd=str(git_root),
+            capture_output=True,
+            timeout=5,
+            env=_hermetic_git_env(),
+            text=True,
+        )
+    except (OSError, subprocess.TimeoutExpired):
+        return set()  # Subprocess failed — fail open.
+
+    if result.returncode not in (0, 1):
+        return set()  # Fatal git error — fail open.
+
+    return {line for line in result.stdout.splitlines() if line}
+
+
 def _is_git_ignored(resolved_path: Path) -> bool:
     """True if ``resolved_path`` is gitignored BY ITS OWN REPO.
 
@@ -270,7 +342,9 @@ def _is_git_ignored(resolved_path: Path) -> bool:
     full rationale. Summary: resolve the nearest ``.git`` ancestor of the
     file (its OWNING repo, which may be a child repo nested under
     ``REPO_ROOT``, not ``REPO_ROOT`` itself), then run ``git check-ignore``
-    there against the path relative to that repo.
+    there against the path relative to that repo — checking the per-file AND
+    per-directory caches first (#1122; see the module docstring's cache
+    section for exactly what each caches and why both are sound).
 
     Fails OPEN (returns False -> file gets tracked) on every error case: no
     ``.git`` ancestor, a path that doesn't resolve relative to its own
@@ -287,28 +361,40 @@ def _is_git_ignored(resolved_path: Path) -> bool:
     except ValueError:
         return False  # Shouldn't happen (git_root is an ancestor), fail open anyway.
 
-    cache_key = (str(git_root), str(rel))
-    cached = _GIT_CHECK_IGNORE_CACHE.get(cache_key)
+    rel_str = str(rel)
+    file_key = (str(git_root), rel_str)
+    cached = _GIT_CHECK_IGNORE_CACHE.get(file_key)
     if cached is not None:
         return cached
 
-    try:
-        result = subprocess.run(
-            ["git", "check-ignore", "-q", "--", str(rel)],
-            cwd=str(git_root),
-            capture_output=True,
-            timeout=5,
-            env=_hermetic_git_env(),
-        )
-        # `git check-ignore -q` exit codes: 0 = ignored, 1 = not ignored,
-        # 128 = fatal error (e.g. not a git repo after all). Only 0 counts
-        # as ignored; anything else — including the fatal case — fails open.
-        ignored = result.returncode == 0
-    except (OSError, subprocess.TimeoutExpired):
-        ignored = False  # Subprocess failed — fail open.
+    dir_rel_str = str(rel.parent)
+    dir_key = (str(git_root), dir_rel_str)
+    dir_cached = _DIR_CHECK_IGNORE_CACHE.get(dir_key)
 
-    _GIT_CHECK_IGNORE_CACHE[cache_key] = ignored
-    return ignored
+    if dir_cached is True:
+        # gitignore(5): a file cannot be re-included once a parent directory
+        # is excluded — authoritative without a subprocess call.
+        _GIT_CHECK_IGNORE_CACHE[file_key] = True
+        return True
+
+    if dir_cached is False:
+        # Directory itself isn't excluded — that says nothing about THIS
+        # file (a filename pattern can still match inside it), so fall
+        # through to a per-file check, same subprocess cost as before #1122.
+        ignored = rel_str in _run_check_ignore(git_root, [rel_str])
+        _GIT_CHECK_IGNORE_CACHE[file_key] = ignored
+        return ignored
+
+    # Neither the file nor its directory is cached yet: one subprocess call
+    # answers both questions and seeds both caches, so every LATER file
+    # under this same directory is a cache hit instead of a new subprocess.
+    dir_spec = f"{dir_rel_str}/"
+    matched = _run_check_ignore(git_root, [dir_spec, rel_str])
+    dir_ignored = dir_spec in matched
+    file_ignored = rel_str in matched
+    _DIR_CHECK_IGNORE_CACHE[dir_key] = dir_ignored
+    _GIT_CHECK_IGNORE_CACHE[file_key] = file_ignored
+    return file_ignored
 
 
 def _should_skip(file_path: str) -> bool:
@@ -402,6 +488,20 @@ def check(input_data: dict) -> dict | None:
     files = data.setdefault("files", {})
 
     existing = files.get(rel_path, {})
+    if existing.get("last_tracked") == sha:
+        # No-op re-save (#1122): the file's content hash is byte-for-byte
+        # what's already recorded — e.g. an edit that reverts to prior
+        # content, or a Write that rewrites identical bytes. `sha` is always
+        # a real 64-hex-char digest here (the `sha is None` case already
+        # returned above), so this only matches an EXISTING entry whose
+        # tracked hash is unchanged — never a brand-new path (`existing`
+        # empty -> `.get("last_tracked")` is `None`, which can't equal a
+        # real digest). Dirty-ness is driven by `last_tracked !=
+        # last_resolved`, not by `tracked_at`, so re-writing the full 103 KB
+        # file here would change zero meaningful state — skip the write
+        # (the read above still had to happen, to learn this).
+        return {"action": "skip_noop", "path": rel_path}
+
     files[rel_path] = {
         "last_tracked": sha,
         "last_resolved": existing.get("last_resolved", ""),
--- a/.claude/hooks/ontology_tracker.py
+++ b/.claude/hooks/ontology_tracker.py
@@ -449,6 +449,15 @@ def _is_git_ignored(resolved_path: Path) -> bool:
     # ignored — gitignore(5)'s no-re-include-under-an-excluded-parent rule
     # then genuinely applies to everything beneath it, so the cached True
     # is not a false positive there.
+    #
+    # Why no `--no-index` (main#1263 review, Weronika Zielinska): git's
+    # index-masking applies to the DIRECTORY pathspec too, not just to
+    # files. `git check-ignore -- dist dist/manifest.json` exits 1 when
+    # `dist` holds force-added tracked files, where `--no-index` would
+    # report both as matched. The practical consequence is a useful
+    # invariant rather than a bug: `dir_ignored` can only ever cache True
+    # for a directory that holds ZERO tracked files, so the short-circuit
+    # can never suppress tracking of a path git already knows about.
     dir_spec = dir_rel_str
     matched = _run_check_ignore(git_root, [dir_spec, rel_str])
     dir_ignored = dir_spec in matched
--- a/.claude/hooks/tests/test_ontology_tracker.py
+++ b/.claude/hooks/tests/test_ontology_tracker.py
@@ -793,11 +793,26 @@ class GitCheckIgnoreExcludeThenReincludeTests(_FakeRepoRootMixin, unittest.TestC
         self.assertTrue(hook._DIR_CHECK_IGNORE_CACHE[(str(self._fake_root), "data/raw/sub")])
 
     def test_trailing_slash_is_unsound_bare_name_is_not(self):
-        """Direct guard on the exact mechanism, independent of
-        ``_is_git_ignored``'s call sequence: proves the trap AND the fix in
-        one place, so a future re-introduction of the trailing slash fails
-        immediately rather than needing the fuller scenario above to
-        surface it."""
+        """Characterization of GIT's behavior — the PREMISE the fix rests
+        on — not a regression guard on our code.
+
+        Read the scope carefully (main#1263 review, Weronika Zielinska).
+        An earlier version of this docstring claimed a future
+        re-introduction of the trailing slash would "fail immediately"
+        here. **That is false**, and was measured false: this test calls
+        ``_run_check_ignore`` with literal strings, so it never touches
+        ``_is_git_ignored``'s ``dir_spec`` at all and passes unchanged when
+        the trailing slash is reinstated. The test that actually catches
+        that mutation is ``test_untracked_reincluded_keeper_is_not_skipped``
+        in the class above — and only that one.
+
+        What this DOES pin is worth keeping: that real git treats a
+        trailing-slash pathspec as a literal string matched by a
+        contents-only pattern while the bare name is not. If git ever
+        changed that, the fix's rationale would evaporate silently and
+        every other test here would still pass. Keeping it labeled
+        honestly is the point — a test that overstates what it guards is
+        how a suite comes to look stronger than it is (cf. main#1215)."""
         raw = self._fake_root / "data" / "raw"
         raw.mkdir(parents=True)
         (raw / "dump.parquet").write_text("x\n", encoding="utf-8")
@@ -933,12 +948,29 @@ class RunCheckIgnoreFailOpenTests(_FakeRepoRootMixin, unittest.TestCase):
         self.assertEqual(matched, set())
 
     def test_real_fatal_returncode_also_fails_open(self):
-        """Not mocked: a real ``git check-ignore`` invocation with an
-        out-of-repo pathspec genuinely exits 128 (verified: real git prints
-        ``fatal: ... is outside repository`` and returns 128 for this
-        exact input) — must also fail open, end to end through the real
-        subprocess path."""
-        outside = "/etc/hostname"  # universally readable on Linux test runners
+        """Not mocked: a real ``git check-ignore`` with an out-of-repo
+        pathspec genuinely exits 128 (real git prints ``fatal: ... is
+        outside repository``) and must fail open end to end.
+
+        **This test is vacuous with respect to the guard it appears to
+        cover** (main#1263 review, Weronika Zielinska — measured, not
+        assumed). A real fatal exit also produces EMPTY stdout, so the
+        function returns an empty set with or without the
+        ``returncode not in (0, 1)`` branch: deleting that branch leaves
+        this test passing. ``test_mocked_fatal_returncode_fails_open`` is
+        the only test that catches it, which is exactly why that one feeds
+        deliberately NON-empty stdout.
+
+        Kept because it pins the PREMISE the mocked test is built on — that
+        128 is really what git returns here, rather than a return code we
+        invented for a fixture. Labeled so the pair is never miscounted as
+        two guards on the same branch (cf. main#1215: an anti-vacuity
+        assertion can itself be vacuous).
+
+        Not a flake risk despite naming a system path: any absolute
+        out-of-repo pathspec exits 128 whether or not the file exists
+        (verified against a nonexistent path)."""
+        outside = "/etc/hostname"  # any absolute out-of-repo path works
         matched = hook._run_check_ignore(self._fake_root, [outside])
         self.assertEqual(matched, set())
 
--- a/.claude/hooks/tests/test_ontology_tracker.py
+++ b/.claude/hooks/tests/test_ontology_tracker.py
@@ -814,6 +814,78 @@ class GitCheckIgnoreExcludeThenReincludeTests(_FakeRepoRootMixin, unittest.TestC
         self.assertIn("data/raw/", matched_slash)
 
 
+class GitCheckIgnoreNonAsciiTests(_FakeRepoRootMixin, unittest.TestCase):
+    """main#1265: matching on git's ECHOED pathspec is encoding-sensitive.
+
+    Under git's default ``core.quotePath=true`` a pathspec containing any
+    non-ASCII byte is C-quoted on the way out (``عربي.log`` echoes as
+    ``"\\330\\271\\330\\261\\330\\250\\331\\212.log"``), so exact-string
+    membership against what we passed in never matches and a genuinely
+    ignored file is reported NOT ignored. That is a behavioural regression
+    versus the pre-#1122 code, which read only ``check-ignore -q``'s exit
+    status and was encoding-independent by construction.
+
+    The failure direction is the safe one (fail-open -> over-track, never
+    under-track) and no repo currently holds a non-ASCII path, so it was
+    latent. It is pinned here anyway because this org's domain is
+    Arabic-language scholarly data and the over-tracked entries land in the
+    COMMITTED ``ontology/checksums.json`` — the #1038 phantom-drift-forever
+    shape.
+    """
+
+    def setUp(self):
+        super().setUp()
+        hook._GIT_CHECK_IGNORE_CACHE.clear()
+        hook._DIR_CHECK_IGNORE_CACHE.clear()
+        subprocess.run(
+            ["git", "init", "-q", str(self._fake_root)],
+            check=True,
+            capture_output=True,
+            env=hook._hermetic_git_env(),
+        )
+        (self._fake_root / ".gitignore").write_text("*.log\nبناء/\n", encoding="utf-8")
+
+    def tearDown(self):
+        hook._GIT_CHECK_IGNORE_CACHE.clear()
+        hook._DIR_CHECK_IGNORE_CACHE.clear()
+        super().tearDown()
+
+    def test_non_ascii_ignored_file_is_detected(self):
+        """The core case: removing the ``core.quotePath=false`` pin makes
+        this return False."""
+        target = self._fake_root / "عربي.log"
+        target.write_text("x\n", encoding="utf-8")
+
+        self.assertTrue(hook._is_git_ignored(target.resolve()))
+
+    def test_ascii_sibling_still_detected(self):
+        """Control: the ASCII path was never affected, so a passing
+        non-ASCII test alone would not prove the pin is what fixed it."""
+        target = self._fake_root / "plain.log"
+        target.write_text("x\n", encoding="utf-8")
+
+        self.assertTrue(hook._is_git_ignored(target.resolve()))
+
+    def test_non_ascii_ignored_directory_is_detected(self):
+        """The directory-cache half: a non-ASCII directory excluded by a
+        genuine directory pattern must cache True, which it cannot do while
+        the echo is quoted."""
+        d = self._fake_root / "بناء"
+        d.mkdir()
+        target = d / "a.md"
+        target.write_text("x\n", encoding="utf-8")
+
+        self.assertTrue(hook._is_git_ignored(target.resolve()))
+
+    def test_echo_round_trips_the_caller_string(self):
+        """Direct guard on the mechanism, independent of ``_is_git_ignored``:
+        what git echoes back must be exactly what we passed in."""
+        (self._fake_root / "عربي.log").write_text("x\n", encoding="utf-8")
+
+        matched = hook._run_check_ignore(self._fake_root, ["عربي.log"])
+        self.assertEqual(matched, {"عربي.log"})
+
+
 class RunCheckIgnoreFailOpenTests(_FakeRepoRootMixin, unittest.TestCase):
     """main#1263 review: the ``returncode not in (0, 1)`` fail-open branch
     in ``_run_check_ignore`` had no test pinning it — a mutation deleting it
--- a/.claude/hooks/tests/test_ontology_tracker.py
+++ b/.claude/hooks/tests/test_ontology_tracker.py
@@ -442,6 +442,7 @@ class GitCheckIgnoreGeneralizationTests(_FakeRepoRootMixin, unittest.TestCase):
     def setUp(self):
         super().setUp()
         hook._GIT_CHECK_IGNORE_CACHE.clear()
+        hook._DIR_CHECK_IGNORE_CACHE.clear()
         # `env=hook._hermetic_git_env()` is load-bearing (main#719): the
         # pre-push pytest hook is itself invoked by `git push`, which exports
         # GIT_DIR/GIT_WORK_TREE for the real repo into every subprocess this
@@ -458,6 +459,7 @@ class GitCheckIgnoreGeneralizationTests(_FakeRepoRootMixin, unittest.TestCase):
 
     def tearDown(self):
         hook._GIT_CHECK_IGNORE_CACHE.clear()
+        hook._DIR_CHECK_IGNORE_CACHE.clear()
         super().tearDown()
 
     def test_child_repo_file_ignored_by_parent_is_still_tracked(self):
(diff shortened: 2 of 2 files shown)


parametrization:
## Review — PR #1263 (`ontology_tracker.py` per-edit ledger IO)

Reviewed against `ontology_tracker.py` at head `f6a5821b` (the correction noted in the spawn brief — this PR is not about `validate_edit_completion.py`). Read the file at that ref via `gh api contents ...?ref=f6a5821b`, then did all verification in a separate clone outside the parent worktree (`git clone` + `git fetch origin pull/1263/head` + checkout), not by checking out inside `noorinalabs-main` itself.

### 1. Skip no-op writes (`check()`)

`existing.get("last_tracked") == sha` is the only predicate gating `skip_noop`. Walked the edge cases:

- Never-seen path: `existing == {}` → `.get("last_tracked")` is `None`, never equal to a real 64-hex digest → always writes. Covered by `test_first_track_of_a_new_path_always_writes`.
- Entry present, `last_tracked` unchanged but `last_resolved` differs (dirty file, identical re-save): predicate still matches → skip. Correct — dirtiness is driven solely by `last_tracked != last_resolved`, and neither field changes on this path, so skipping loses no signal. Not separately unit-tested, but follows directly from the single-field comparison and is covered indirectly by every skip-noop test using a `read/compare` that never touches `last_resolved`.
- Entry with missing/empty `last_tracked`: `"" != <64-hex sha>`, falls through to a real write. Consistent.

Mutation-tested three ways against the real source (not just re-reading the tests) in the scratch clone:
- `if existing.get(...) == sha:` → `if True:` (always skip): 3 tests fail, including `test_first_track_of_a_new_path_always_writes` and `test_changed_content_after_a_noop_still_writes`.
- Predicate reworded to `if rel_path in files:` (path-presence instead of SHA): `test_changed_content_after_a_noop_still_writes` fails.
- Both mutations correctly caught — the predicate is exact, not over-broad in either direction.

### 2. Per-directory check-ignore cache

Confirmed the asymmetry is real, not just asserted in the docstring: `dir_cached is True` short-circuits to `True` with zero subprocess calls; `dir_cached is False` still falls through to a per-file `_run_check_ignore`; only the "neither cached" branch issues one combined `git check-ignore <dir>/ <file>` call that seeds both caches from its stdout matches. Mutated the `dir_cached is False` branch to trust the directory verdict for the file too (`return False` without the per-file check) — `test_second_file_in_a_not_ignored_directory_is_still_checked_individually` fails as expected (a `*.secret` file inside an otherwise-clean directory would wrongly be tracked). The unsafe direction is guarded; the safe direction (reusing an ignored-directory verdict) is the one actually exercised for savings.

Checked the one case the tests don't directly cover — a file living at the git root itself (`rel.parent == Path('.')`, so `dir_spec == "./"`). Verified against real git in a throwaway repo: `git check-ignore -- "./" <file>` never echoes back `"./"` as a match (a repo root can't itself be "ignored" by its own `.gitignore`), so top-level files just always fall through to the per-file check — same behavior as before this PR, not a latent bug.

### 3. Serialization path unchanged

`checksums_io.write_checksums()` remains the sole write path (no direct `json.dump` reintroduced in the hook); `checksums_io.py` still sets `ensure_ascii=False`. The ascii-preservation gate (#1038/#1042) is not at risk here.

### Verification run

- Cloned the repo fresh into scratch (outside the parent worktree), fetched `pull/1263/head`, checked out `f6a5821b` — confirmed `git rev-parse HEAD` matches.
- `PYTHONDONTWRITEBYTECODE=1 ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/test_ontology_tracker.py -q` → **57 passed**.
- Re-ran after each of the three mutations above (skip-noop × 2, directory-cache × 1) → all three broke the expected tests, then reverted (`git checkout --`) before the next mutation.
- `gh pr checks 1263` → **14/14 green** (Pytest — hooks and lib test suites, Mypy, Ruff format/lint, sync-drift gate, roster gate, etc.), confirming the orchestrator's CI-ready report is current.
- Diff is a single commit, scoped exactly to `.claude/hooks/ontology_tracker.py` + its test file (+298/−19), no unrelated changes.

### Verdict

Approved. The two optimizations are each proven correct by a targeted mutation test that fails when the predicate/cache is broadened in the unsafe direction, and I reproduced those failures myself rather than taking the author's word for it. No debt found — nothing to file.

---

Requestor: Bereket Tadesse
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Merge-gate review — PR #1263 @ `f6a5821b`** (read via `gh api contents/…?ref=f6a5821b`, no local checkout)

CI verified at the head SHA itself (`check-runs` for `f6a5821b`): **14/14 success**. Base for the diff: `b3327280`.

Verdict: **ChangesRequested** — one of the two optimisations is sound and well-guarded; the other introduces a silent under-tracking regression versus base that I reproduced end-to-end against a verbatim real-repo `.gitignore`.

## Must-fix 1 — the directory ignore-cache over-reports for the `dir/*` idiom, silently skipping files git would track

The `_DIR_CHECK_IGNORE_CACHE` design rests on a stated invariant:

> a directory that `git check-ignore` reports as ignored makes EVERY file beneath it ignored too, no exceptions possible

The gitignore(5) rule quoted is real. The implementation does not test that rule. It probes with the pathspec `f"{dir_rel_str}/"` — and **a trailing-slash pathspec matching a pattern is not the same fact as the directory being excluded.** The very common `dir/*` + `!dir/keeper` idiom excludes only a directory's *contents*, which leaves re-inclusion fully working — but `dir/*` also matches the literal string `dir/`, so `check-ignore` reports the directory as a match:

```
.gitignore:  partial/*
             !partial/keep.txt

$ git check-ignore -v -- "partial/"          # what the PR asks
.gitignore:3:partial/*	partial/            # -> dir_ignored = True
$ git check-ignore -v -- "partial"           # what it means to ask
                                             # rc=1, no match — dir NOT excluded
$ git status --porcelain --ignored=matching
!! partial/a.txt                             # only the file is ignored
?? partial/                                  # directory is NOT excluded
```

Once `_DIR_CHECK_IGNORE_CACHE[(root, "partial")] = True` is seeded by any ignored file in that directory, the `dir_cached is True` branch returns `True` for every later file **without a per-file check** — including the negation-re-included one.

This is not hypothetical. `noorinalabs-isnad-ingest-platform/.gitignore` uses exactly this idiom (`data/raw/*` … `!data/**/.gitkeep`), and those `.gitkeep` files are committed and tracked. Repro through `_should_skip`, that `.gitignore` copied verbatim into a throwaway `git init` fixture:

```
git's own verdict for data/raw/.gitkeep: NOT ignored (committable)
HEAD f6a5821b: _should_skip(dump.parquet)=True  then  _should_skip(.gitkeep)=True   -> .gitkeep SKIPPED (never tracked)
BASE b3327280: _should_skip(dump.parquet)=True  then  _should_skip(.gitkeep)=False  -> .gitkeep tracked
```

That is the exact failure direction `_is_git_ignored`'s whole fail-open design exists to prevent — an unrecorded edit is invisible forever, not deferred — landing in the one function whose docstring promises it "never risks the former".

**Severity, stated honestly:** the dispatcher hands `check()` one `input_data` per process, so `_should_skip` -> `_is_git_ignored` runs at most once per hook invocation. The dir cache is therefore *seeded but never read* in production today: the regression is **latent, not live**, and the optimisation currently delivers zero production saving. I am still blocking rather than filing it, because (a) the fix is one line, (b) the module docstring explicitly invites the batch caller that makes it live ("a future batch invocation"), and (c) whoever writes that caller will read a comment asserting the invariant is airtight.

**Fix:** drop the trailing slash — probe `dir_rel_str`, not `f"{dir_rel_str}/"`. That form tracks git's real directory-exclusion semantics in every case I checked, including at the repo root:

| `.gitignore` form | `check-ignore 'D/'` | `check-ignore 'D'` | directory truly excluded? |
|---|---|---|---|
| `build/` | IGNORED | IGNORED | yes |
| `nodir` (no slash) | IGNORED | IGNORED | yes |
| `partial/*` + `!partial/keep.txt` | IGNORED | not | **no** |
| (unmatched dir) | not | not | no |
| root dir (`.` / `./`) | rc=1 | rc=1 | no |

And with a genuinely-excluded directory (`build/` + `!build/keep.txt`), `git check-ignore -v -- build/keep.txt` reports `.gitignore:1:build/` — git already applies the no-re-include rule per file, so the `is True` shortcut stays correct under the no-slash probe.

Please also add the regression test that currently does not exist: a `dir/*` + `!dir/keeper` fixture asserting the *second* file in the directory is still tracked. I mutated `dir_spec` to the no-slash form and **all 57 tests passed** — nothing in the suite pins this decision either way, which is why the gap got through.

## Must-fix 2 (small) — the new fail-open guard in `_run_check_ignore` is inert

`if result.returncode not in (0, 1): return set()` is new in this PR and is untested: I deleted those two lines and the suite stayed green (57 passed). It is close to redundant in the common case (git writes fatals to stderr, so stdout is empty anyway) but not fully — a partially-succeeding multi-pathspec call is precisely the new shape this PR introduces. One test with a non-repo `cwd` asserting the empty set would make it load-bearing.

## What I verified as sound

**The no-op write skip is exact.** I exercised the predicate directly against `check()` on every edge the brief called out, rather than reading it:

| existing entry | result | wrote? |
|---|---|---|
| `last_tracked: ""` | tracked | yes |
| `last_tracked` key missing | tracked | yes |
| no entry at all | tracked | yes |
| `last_tracked == sha`, `last_resolved` **differs** (dirty) | skip_noop | no — entry left dirty, intact |
| `last_tracked == sha`, `last_resolved == sha` (clean) | skip_noop | no |
| `last_tracked` = a different sha (revert or change) | tracked | yes |

The revert case behaves correctly in both directions: reverting to the *currently tracked* sha is a genuine no-op; reverting to an *older* sha differs from `last_tracked` and writes.

**Write path and side effects.** `checksums_io.write_checksums()` remains the only write path — the skip branch writes nothing at all, so it cannot bypass the `ensure_ascii=False` contract or the `checksums-ascii` gate (live in CI at `.github/workflows/docs.yml`, mirrored in `.pre-commit-config.yaml`, classified by the sync gate). I grepped for anything depending on the write as a side effect: no consumer stats `checksums.json`'s mtime, and no atomic-replace ordering is disturbed (no `.tmp` is created on the skip path). `tracked_at` has exactly one consumer — the librarian's "changed {tracked_at}" line — and not bumping it on an unchanged hash makes that display *more* accurate, not less. The claimed saving is real: the read is genuinely required to learn the answer, and the file is 101,398 bytes.

**Your three mutations reproduce.** Re-run at source level with `PYTHONDONTWRITEBYTECODE=1`, each reverted after:

| mutation | result |
|---|---|
| always skip the write | KILLED — 3 tests |
| skip on path presence, ignore SHA | KILLED — `test_changed_content_after_a_noop_still_writes` |
| over-broad dir cache (reuse a *False* verdict for the file) | KILLED — `test_second_file_in_a_not_ignored_directory_is_still_checked_individually` |

These are real guards, not inert ones — the asymmetry you describe for the *not-ignored* direction is genuinely implemented and genuinely tested. The gap is that the *ignored* direction was taken on trust; the guard you wrote defends the half that was already correct.

Baseline for all mutation runs: 57 passed, in an isolated tree outside the repo.

## Not blocking

The module docstring's echo-back claim ("git echoes back … preserving the caller's original string — verified against real git") is accurate — I confirmed `build/` comes back as `build/`, byte-identical. The docstring sentence that needs amending is the *inference* drawn from a match on `D/`, not the mechanism.

---

Requestor: Weronika Zielinska
Requestee: Wanjiku Mwangi
TechDebt: none

parametrization:
**Follow-up to my ChangesRequested verdict — mechanism pinned, and one correction to what I wrote.**

The block stands and the fix is unchanged, but a challenge to my stated mechanism was right to be raised: probing the live `noorinalabs-isnad-ingest-platform` checkout with `git check-ignore -v "data/raw/"` returns **no match**, which appears to contradict my explanation. It does not, and the reason matters for anyone re-testing this. I instrumented the actual invocation the code makes — full argv, full stdout, exit code — at the moment `.gitkeep` is misjudged.

## The mechanism, measured

```
-- HEAD f6a5821b, .gitkeep not yet committed --
  call for dump.parquet:
    argv = git check-ignore -- ['data/raw/', 'data/raw/dump.parquet']
    rc=0  stdout=['data/raw/', 'data/raw/dump.parquet']
  -> skip=True
  call for .gitkeep:
    (no subprocess call at all — dir-cache hit)
  -> skip=True   BUG: never tracked
  dir cache: {'data/raw': True}
```

Two pathspecs in, **two** lines out: git echoed `data/raw/` back as a genuine match. `data/raw/*` really does match the literal string `data/raw/`. That is cause (1) — not output mis-attribution.

Mis-attribution is ruled out twice. In the code, `_run_check_ignore` returns a `set` of stdout lines and the caller tests membership (`dir_spec in matched`, `rel_str in matched`) — there is no positional mapping to get wrong. And empirically, the same two-pathspec call in the committed fixture below returns rc=0 with exactly **one** line (`data/raw/dump.parquet`), which the code attributes correctly (`dir_ignored=False`, `file_ignored=True`). If parsing were positional, that lone line would have been credited to `data/raw/` and produced the bug there too. It doesn't. Also note the misjudged `.gitkeep` call issues **zero** subprocess calls — there is no second invocation whose output could be mis-attributed.

## Why the live-repo probe looks different: index-awareness

`git check-ignore` without `--no-index` skips paths tracked in the index. Same fixture, `.gitkeep` committed — which mirrors the live repo, where all four `.gitkeep` files are committed:

```
  default    check-ignore 'data/raw/'  -> rc=1 (no match)
  --no-index check-ignore 'data/raw/'  -> data/raw/
```

The pattern matches; the index masks it. Same effect on the live repo — my own earlier probe there returned rc=1 by default and `.gitignore:2:data/raw/*	data/raw/` under `--no-index`.

## Correction to my review

I wrote that the bug "is not hypothetical" because those `.gitkeep` files "are committed and tracked". That framing was wrong, and in the direction that overstates the finding: being tracked is precisely what protects them. Run against the live checkout as it stands, HEAD behaves correctly here —

```
-- HEAD f6a5821b, .gitkeep committed --
  argv = git check-ignore -- ['data/raw/', 'data/raw/dump.parquet']
  rc=0  stdout=['data/raw/dump.parquet']
  -> .gitkeep skip=False (tracked — correct)
```

The live surface is narrower than I implied: it needs a re-included file that is **not yet tracked** — a newly created `data/<newdir>/.gitkeep`, or any `dir/*` + `!dir/keeper` idiom whose keeper has not been committed yet. That is a real and recurring shape (it is how every one of those four `.gitkeep` files began life), but it is not the steady state, and I should have said so.

This narrows reachability; it does not change the verdict. The dominant severity caveat was already the one I stated originally — the dispatcher hands `check()` one file per process, so the dir cache is seeded but never read in production today. The regression is latent either way, and the reason to fix it now is unchanged: one line, silent failure direction, and a docstring that invites the batch caller which makes it live.

## The fix addresses this cause

Same fixture, same instrumentation, with `dir_spec = dir_rel_str`:

```
-- HEAD + dir_spec = dir_rel_str --
  call for dump.parquet:
    argv = git check-ignore -- ['data/raw', 'data/raw/dump.parquet']
    rc=0  stdout=['data/raw/dump.parquet']        <- 'data/raw' correctly NOT matched
  -> skip=True
  call for .gitkeep:
    argv = git check-ignore -- ['data/raw/.gitkeep']
    rc=1  stdout=[]
  -> skip=False (tracked — correct)
  dir cache: {'data/raw': False}
```

It fixes the cause rather than the symptom: the slash-less pathspec asks "is this directory excluded", which is the question the `is True` shortcut's gitignore(5) reasoning actually requires, instead of "does any pattern match this string". The `is True` shortcut itself stays sound under it — with a genuinely excluded directory (`build/` + `!build/keep.txt`), git already reports `build/keep.txt` as ignored via `.gitignore:1:build/`, so no per-file re-check is needed. Cost is unchanged in the common case and one extra subprocess for the negation case, which is what the `is False` branch always paid.

The two coverage items from my verdict are unaffected by any of this and still need doing: nothing in the suite pins the `dir_spec` form (my mutation to the slash-less form passed all 57 tests), and the new `returncode not in (0, 1)` fail-open in `_run_check_ignore` is untested (deleting it also passed all 57).

---

Requestor: Weronika Zielinska
Requestee: Wanjiku Mwangi
RequestOrReplied: Reply
TechDebt: none

parametrization:
## Rework pushed — `f6a5821b` → `8ac47dd2`

All three merge-gate must-fixes are addressed. Per #950 this head move invalidates both prior verdicts (Bereket Tadesse's Approved and Weronika Zielinska's ChangesRequested are now excluded by the gate); re-review at the current head is required from **both** reviewers, not only the blocker — materiality is not an input to the gate.

**1 — the trailing slash is gone.** `dir_spec = dir_rel_str` (was `f"{dir_rel_str}/"`). `git check-ignore` treats a trailing-slash pathspec as a literal string, so a contents-only pattern like `data/raw/*` echoes `"data/raw/"` back as matched even though the directory itself is not excluded. That false verdict was cached as `dir_ignored=True` and short-circuited every later file in the directory — including one a `!` rule re-includes. Under-tracking, the one direction this function must never risk.

**2 — regression test written against an UNTRACKED keeper.** This is the reviewer correction that shaped the fixture: a committed keeper is masked by git's index (`check-ignore` never flags a tracked path, regardless of pattern), so a test built on an already-added keeper passes for an unrelated reason and proves nothing about the bug. The tracked case is retained as its own explicitly-labeled test so it is never mistaken for covering the pattern fix.

**3 — the `returncode not in (0, 1)` fail-open is now pinned.** Deleting it previously survived the suite. The new test supplies NON-empty stdout deliberately: an empty-stdout fatal is indistinguishable from "nothing ignored" without the guard, so an empty-stdout test would pass either way.

**Mutation-verified, both directions** (run before commit, on the exact pushed tree):

| mutation | result |
|---|---|
| reinstate `f"{dir_rel_str}/"` | `test_untracked_reincluded_keeper_is_not_skipped` FAILS (`True is not false`) |
| delete the `returncode not in (0, 1)` guard | `test_mocked_fatal_returncode_fails_open` FAILS (`some/path` leaks through) |

63 tests pass at head. Both pre-commit stages green (commit stage + the full pre-push set: mypy, pytest, memory/headcount budget, doc-freshness, office-drift, mermaid-render).

One point carried forward from the review thread for the record: my earlier claim that the four committed `.gitkeep` files in `noorinalabs-isnad-ingest-platform` made this a live production bug was **overstated** — being tracked is exactly what protects them. The reachable surface is a *newly created, not-yet-tracked* keeper, which is precisely the moment the tracker hook runs. Weronika issued the same correction against her own finding independently. The fix stands on that narrower ground.

parametrization:
Re-review at head `8ac47dd2` (my prior Approved at `f6a5821b` was dropped per #950 after the merge-gate block).

All three must-fixes verified as landed correctly:

1. **Trailing-slash pathspec fix.** `dir_spec` is now the bare relative directory (no `/`). I reproduced the underlying bug against real `git check-ignore` in a scratch repo with a `data/raw/*` + `!data/**/.gitkeep` gitignore: `git check-ignore -- data/raw` exits 1 (bare name, correctly not matched), while `git check-ignore -- data/raw/` exits 0 and echoes back the literal string `data/raw/` — confirming the old trailing-slash form would have cached a false "directory ignored" and mis-skipped the re-included keeper. With the fix, an untracked `.gitkeep` under `data/raw/` correctly comes back not-ignored, and a nested directory whose own name is itself swept up by the contents-only pattern (`data/raw/sub`) still correctly comes back ignored. I independently verified every one of these outcomes against real git, not just the test fixtures, and they match exactly.

2. **Regression test for the `dir/*` + `!dir/**/.gitkeep` idiom.** `GitCheckIgnoreExcludeThenReincludeTests` covers the untracked keeper (the actual bug surface) as its own explicitly-labeled test, separate from the tracked-keeper case — which is correctly documented as passing for an unrelated reason (index masking) and not proof of the fix. Good discipline; this is exactly the distinction the must-fix asked for.

3. **Fail-open pin for `_run_check_ignore`.** `RunCheckIgnoreFailOpenTests` pins the `returncode not in (0, 1)` branch with both a mocked non-empty-stdout fatal response (which would slip through and falsely report a match if the guard were deleted) and a real out-of-repo pathspec that genuinely exits 128 — I reproduced the same real-128 behavior in my own scratch repo.

**Comment volume:** the module docstring and inline comments are long (~45 of the ~180 added lines), but I checked the specific factual claims against real git behavior (bare-vs-trailing-slash matching, multi-pathspec echo-back preserving the caller's exact string, nested-directory sweep-up, index masking of tracked paths, exit-128 on out-of-repo pathspecs) and every one holds. Given this is documenting the exact subtle failure mode that got caught in the prior review round, I consider the volume proportionate, not just present.

**Original #1122 half (no-op write skip):** still present and unregressed. `SkipNoopWriteTests` covers the predicate directly (first-track always writes, unchanged re-save skips, a genuine change after a no-op still writes, and the skipped case leaves the on-disk entry byte-identical) — these pass and look like a real mutation guard, not just a happy-path check.

**Verification run at head `8ac47dd2`** (worktree checkout, not the parent working tree):
- `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/test_ontology_tracker.py -q` → 63 passed
- `ruff check` on both changed files → all checks passed
- CI on the PR (`statusCheckRollup` at `8ac47dd2`) → all 14 checks green, including Mypy type check and Pytest

No new tech debt found.

---

Requestor: Bereket Tadesse
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Merge-gate re-review at head `8ac47dd2`** (prior verdict was at `f6a5821b` and is excluded per #950). Reviewed against `origin/main` (`b696aab`); file contents pulled at the head SHA via `gh api .../contents?ref=8ac47dd2...` and byte-compared to the fetched ref before reading.

## The three blocking findings are genuinely fixed — I re-ran the mutations myself

I did not take the pre-verification at face value. Five mutations applied to a scratch clone at head, outside the repo, each reverted after:

| # | Mutation | Result |
|---|---|---|
| 1 | reinstate `dir_spec = f"{dir_rel_str}/"` | **1 failed** — `test_untracked_reincluded_keeper_is_not_skipped` (`AssertionError: True is not false`) |
| 2 | delete the `returncode not in (0, 1)` fail-open branch | **1 failed** — `test_mocked_fatal_returncode_fails_open` (`E 'some/path'`) |
| 3 | delete the whole no-op skip branch | **2 failed** — `test_unchanged_content_reedit_skips_the_write`, `test_skip_leaves_the_on_disk_entry_byte_identical` |
| 4 | trust a `False` directory verdict as "file not ignored" | **1 failed** — `test_second_file_in_a_not_ignored_directory_is_still_checked_individually` |
| 5 | `dir_ignored = False` (cache never records a hit) | **2 failed** — `test_second_file_in_same_ignored_directory_is_a_cache_hit`, `test_nested_directory_swept_up_by_contents_pattern_is_still_ignored` |

Baseline at head: `63 passed`; full suite `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/ .claude/lib/tests/ -q` → `3528 passed, 647 subtests passed`. Every guard is live; none is inert. Fix #2's non-empty `stdout` in the mock is the detail that makes it a real guard rather than a vacuous one — good catch on your part.

## The directory-query soundness reasoning is correct, not a rationalization

I checked `test_nested_directory_swept_up_by_contents_pattern_is_still_ignored` against real git (2.43.0) rather than trusting the docstring. Four fixtures:

- `data/raw/*` + `!data/**/.gitkeep` — bare `data/raw` → exit 1 (no match); `data/raw/` → echoed back, exit 0. The trap is real and the fix is the right one.
- Nested: `data/raw/sub`, `data/raw/sub/.gitkeep`, `data/raw/sub/x.md` — **all three** reported ignored, and `git status -uall` confirms the nested keeper is genuinely gone. The claim holds.
- `data/raw/*` + `!data/raw/sub/**` (contents-only re-include under an excluded nested dir) — directory and file both ignored; no divergence between the cached directory verdict and the per-file answer.
- `build/` + `!build/keep.txt` — `check-ignore -v` reports `.gitignore:1:build/` as the matching pattern for `build/keep.txt`, i.e. **`check-ignore` itself walks the excluded parent**. That is the property the cache shortcut depends on, and it holds.

I could not construct a case where the bare-name directory verdict short-circuits to `True` while the per-file query would have said not-ignored.

## On `--no-index` — not needed, and the reason is worth writing down

I expected an asymmetry here: directories are not index entries, so I thought a force-added file inside an ignored directory (`dist/` + `git add -f dist/manifest.json`) would get mis-skipped via a `True` directory cache while the per-file query stays index-masked. Real git says otherwise:

```
$ git check-ignore -- "dist" "dist/manifest.json" "dist/bundle.js"; echo exit=$?
exit=1
$ git check-ignore --no-index -- "dist" "dist/manifest.json"
dist
dist/manifest.json
```

`find_pathspecs_matching_against_index` matches the pathspec `dist` against the index entries *beneath* it, so the directory query is masked too. A directory can therefore only ever cache `True` when it contains zero tracked files — which is exactly what makes the shortcut safe. Omitting `--no-index` is correct, and I'd suggest adding those two lines to the docstring, because "the directory pathspec is index-masked as well" is the non-obvious fact the whole cache rests on. Non-blocking.

## The two test-hygiene questions, answered

**Global `subprocess.run` monkeypatch:** safe today. No `xdist`, no `addopts`, no parallel plugin anywhere in `pyproject.toml` / `pytest.ini` / `.pre-commit-config.yaml` / `ci.yml`; both CI and pre-push run a plain serial `pytest`, and the `finally` restores on exception. It would also survive `xdist` (separate processes). It would break under thread-based parallelism only. That said, `mock.patch.object(hook.subprocess, "run", ...)` is the cleaner injection point and is *already the idiom used in the same file* by `SkipNoopWriteTests` — worth converging on next time you touch these. Non-blocking.

**`/etc/hostname`:** not a latent flake, and the comment explaining it is slightly off. The 128 has nothing to do with the file existing:

```
$ git check-ignore -- /etc/definitely-not-here-9271; echo exit=$?
fatal: /etc/definitely-not-here-9271: '...' is outside repository at '...'
exit=128
```

Any absolute path outside the repo yields 128 whether or not it exists, so `# universally readable on Linux test runners` states the wrong reason for why the test is stable. I separately confirmed this test is **vacuous with respect to the guard** — it still passes with the `returncode` branch deleted (empty stdout is indistinguishable from "nothing ignored"). That is fine, since your mocked test is the real guard and its docstring says exactly that; just don't let a future reader mistake the real-git one for the mutation guard. Same caution applies to `test_trailing_slash_is_unsound_bare_name_is_not`: it exercises `_run_check_ignore` directly, so it does **not** fail when the trailing slash is reinstated on the production line — `test_untracked_reincluded_keeper_is_not_skipped` is the only guard on that line. Both are worth keeping; neither substitutes for the other.

## One new finding — filed as #1265, not blocking

Switching from `check-ignore -q` + exit code to parsing echoed pathspecs introduced an encoding dependency that the new docstring gets wrong. With git's default `core.quotePath=true`, any pathspec containing a non-ASCII byte comes back **C-quoted**, so string equality never matches:

```
$ git check-ignore -- "سند.md" | cat -A
"\330\263\331\206\330\257.md"$
```

Against this head, with `.gitignore` = `*.log` + `بناء/`:

```
_run_check_ignore echo   : {'"\\330\\271\\330\\261\\330\\250\\331\\212.log"'}
_is_git_ignored('عربي.log')  -> False   (expected True)
_is_git_ignored('plain.log') -> True    (ASCII control, correct)
_is_git_ignored('بناء/a.md') -> False   (expected True)
```

Same fixture on `origin/main`: both `True`. So it is a regression introduced here, and the `_run_check_ignore` docstring's *"preserving the caller's original string — verified against real git, not assumed"* is false as written — which matters more than the behaviour, because that sentence is what the next maintainer will rely on.

I am **not** blocking on it, and I want the reasoning on the record so the call is auditable: the failure direction is fail-open (over-track, never under-track), which is this module's deliberate one-sided policy; and I checked all eight repos — `git ls-files | rg '[^\x00-\x7F]'` returns **0** tracked non-ASCII paths in every one, with none ignored-on-disk either. The trigger is latent, not live. It is worth fixing soon rather than never because the over-tracked entry lands in the **committed** `ontology/checksums.json`, which is the #1038 phantom-drift-forever shape, in a repo whose domain is Arabic-language material.

The fix is one line, and I verified it end-to-end: prefixing `-c core.quotePath=false` restores `True` for both cases and additionally lets the directory cache record a hit for the non-ASCII directory, which it cannot today. #1265 also carries a `text=True` follow-on (with quoting off, raw UTF-8 reaches a locale-strict decoder, and `UnicodeDecodeError` is not in the `except (OSError, subprocess.TimeoutExpired)` clause — it would escape `check()` against the hook's "exit 0 — always" contract).

## CI

All 14 checks green at the head SHA — confirmed the runs' `head_sha` is `8ac47dd2c13fa6143d47fa808d05bada95e3de7d`, not an older commit: Pytest, Mypy, Ruff lint, Ruff format, smoke test, commit-author verification, roster union gate, pre-commit⇄CI sync gate, memory budget, headcount budget, Dockerfile pinning, Arabic fixture realism, GraphQL over-cap, doc-freshness.

Nice work on the rework — the fixture-design correction (untracked keeper vs the index-masked committed one) was carried through faithfully, and labelling the tracked-case test as *not* covering the fix is exactly the right instinct.

---

Requestor: Weronika Zielinska
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: #1265

parametrization:
## #1265 folded in — `8ac47dd2` → `2c113e7`

Weronika filed #1265 during the merge-gate pass and approved anyway, judging it non-blocking: the failure direction is safe (fail-open → over-track, never under-track) and no repo holds a non-ASCII path today. That reasoning is sound. I've fixed it here regardless, because it is a regression **this PR introduces** rather than pre-existing debt, and the fix is one verified line.

Reproduced against `8ac47dd2` before touching anything:

```
_is_git_ignored('عربي.log')  -> False   (origin/main: True)
_is_git_ignored('plain.log') -> True    (ASCII control, correct)
```

`#1122` changed `_run_check_ignore` from reading `check-ignore -q`'s exit status to matching git's **echoed** pathspecs by string equality — which is what lets one call answer several questions at once, and also what made the function encoding-sensitive. Under git's default `core.quotePath=true` a non-ASCII pathspec comes back C-quoted, so it equals nothing the caller passed in.

Three changes:

1. **`core.quotePath=false` pinned on the invocation** — restores the round-trip property that matching-on-echo gave up.
2. **`text=True` → `encoding="utf-8", errors="replace"`** — this is the part that is easy to miss. With the pin in place raw UTF-8 bytes now reach the decoder, and `text=True` decodes with the locale encoding under `errors='strict'`, so an `LC_ALL=C` runner would raise `UnicodeDecodeError` — not caught by the `except (OSError, TimeoutExpired)` clause, so it would escape `check()` and break this hook's "exit 0 — always" contract. The fix for #1265 without this would trade a latent mis-classification for a latent crash.
3. **Docstring corrected.** It asserted git echoes back "the caller's original string — verified against real git, not assumed." That was false as written. It is true now *because* of the pin, and the docstring says why, so a future maintainer who removes the pin sees what it was load-bearing for.

Four tests added. Mutation-verified — removing the pin fails exactly:

| test | on pin removal |
|---|---|
| `test_non_ascii_ignored_file_is_detected` | FAILS |
| `test_non_ascii_ignored_directory_is_detected` | FAILS |
| `test_echo_round_trips_the_caller_string` | FAILS |
| `test_ascii_sibling_still_detected` (control) | still passes |

The ASCII control is the point: without it, a passing non-ASCII test would not establish that the pin is what fixed it. 67 tests pass at head; both pre-commit stages green.

Reviewers — per #950 this head move invalidates the two Approved verdicts at `8ac47dd2`. Re-affirmation needed from both of you at `2c113e7`. The delta is confined to `_run_check_ignore` and one new test class; nothing else moved.

Leaving #1265 open for you to close, since it is your finding and you should judge whether this fully discharges it.

parametrization:
Re-review at head `2c113e7` (my prior Approved at `8ac47dd2` was dropped per #950).

## The #1265 fix (`core.quotePath=false` pin + `errors="replace"` decoding)

Verified the underlying bug and the fix independently against real git, not just the new fixtures:

- **Reproduced main#1265 as-is**: `git check-ignore -- عربي.log` (default `core.quotePath=true`) echoes the C-quoted string `"\330\271\330\261\330\250\331\212.log"` — set-membership against the caller's own `عربي.log` string misses, so a genuinely ignored Arabic filename comes back not-ignored. Matches the bug report exactly.
- **Confirmed the fix closes it**: with `-c core.quotePath=false`, the same call echoes the unquoted `عربي.log`, and set-membership succeeds.
- **Confirmed `-c` precedence over `_hermetic_git_env()`** (one of the two things asked): `_hermetic_git_env()` only strips `GIT_*` environment variables — it doesn't touch git config, so there's no interaction with the `-c core.quotePath=false` flag. I also forced `core.quotePath=true` in the fake repo's own `.git/config` and confirmed the `-c` flag still won (git's own documented precedence: `-c` beats repo config) — the pin is neither redundant nor overridable by anything already in this codebase.
- **Mutation-checked**: reverted just the `-c core.quotePath=false` piece of the invocation at head `2c113e7` and re-ran the suite — exactly the three claimed tests fail (`test_non_ascii_ignored_file_is_detected`, `test_non_ascii_ignored_directory_is_detected`, `test_echo_round_trips_the_caller_string`), the ASCII control test still passes. Confirms the tests actually exercise the fix rather than passing vacuously.

**The other specific question (does `errors="replace"` introduce a new silent-corruption path):** yes, but narrower than it might sound, and I filed it rather than asserting it from memory. A pathname with genuinely *invalid* UTF-8 bytes (not just non-ASCII — actually undecodable bytes) makes git's echoed stdout decode to `U+FFFD` replacement characters, while the caller's own pathspec string (built via `Path`, which uses POSIX `os.fsdecode`/`surrogateescape` for undecodable bytes) represents the same bytes as lone surrogates — the two strings differ, so the match misses even though `git check-ignore` genuinely matched (`returncode == 0`). I reproduced this directly: a file named with raw bytes `\xff\xfebad.log` is genuinely ignored by `*.log`, but comes back not-matched under the current `errors="replace"` decoding. The failure direction is fail-open (over-track), the same one-sided-safe policy this module already commits to everywhere else, and it requires actually-corrupt filename bytes (not just Arabic/non-ASCII) — nothing any real workflow here produces. I filed it as tech debt rather than judging it inline: **noorinalabs/noorinalabs-main#1268**.

## Everything else re-checked at this head

- `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/test_ontology_tracker.py -q` → 67 passed (up from 63 at the prior head — the 4 new `GitCheckIgnoreNonAsciiTests`).
- `ruff check` on both changed files → clean.
- CI on the PR (`statusCheckRollup` at head `2c113e7`, confirmed via `headRefOid`) → all 14 checks green, including Mypy and Pytest.
- The trailing-slash fix and the fail-open pin from the prior round are both still intact and covered (unchanged since `8ac47dd2`, which I already verified last round).
- Original #1122 no-op-write-skip half: untouched by this commit, still covered by `SkipNoopWriteTests`.

One tech-debt issue filed this round (see above); no other new issues found.

---

Requestor: Bereket Tadesse
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: #1268

parametrization:
**Merge-gate re-review at head `2c113e7`** (prior verdict at `8ac47dd2` excluded per #950). File contents pulled at the head SHA via `gh api .../contents?ref=2c113e7...` and byte-compared to the fetched ref. Reviewed the incremental `8ac47dd2..2c113e7` diff plus the full base diff.

Noted and accepted that you overrode my non-blocking disposition on #1265 and folded the fix in. Your reasoning is better than mine was — a regression the PR itself introduces is a different object from inherited debt, and I should have weighted "self-introduced" above "latent trigger". I'm applying your standard below, which is what produces this verdict.

## What I re-verified at this head

- Full suite: `67 passed`. Removing the pin (`["git", "-c", "core.quotePath=false", ...]` → `["git", ...]`) fails exactly `test_non_ascii_ignored_file_is_detected`, `test_non_ascii_ignored_directory_is_detected`, `test_echo_round_trips_the_caller_string`, with `test_ascii_sibling_still_detected` still passing. Your mutation claim reproduces; the ASCII control does establish the pin as the operative change.
- The earlier guards are not blunted by the rework. Re-applied the trailing-slash and `returncode` mutations together at this head → `2 failed`: `test_untracked_reincluded_keeper_is_not_skipped`, `test_mocked_fatal_returncode_fails_open`. The no-op-skip and dir-cache mutations carry over by construction — the incremental diff touches only `_run_check_ignore`'s docstring and call args plus a new test class, so that code is byte-identical to the head I already mutated.
- CI: 14/14 green, both runs' `head_sha` confirmed `2c113e7182ec6a0f6b1b11545a182479c06006fb`.

## Q2 — is the `-c` pin authoritative? Yes, verified three ways

Fixture: `.gitignore` = `*.log`, file `سند.log`, repo-level `core.quotePath=true` deliberately set.

```
$ git check-ignore -- "سند.log"                       -> "\330\263\331\206\330\257.log"
$ git -c core.quotePath=false check-ignore -- "…"      -> سند.log
$ GIT_CONFIG_PARAMETERS="'core.quotePath=true'" git -c core.quotePath=false … -> سند.log
$ GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=core.quotePath GIT_CONFIG_VALUE_0=true git -c … -> سند.log
```

`-c` is applied last in git's precedence chain, so it wins over the repo/global config files **and** over both env-injection forms even when they are present. Separately, `_hermetic_git_env()` strips those anyway — I confirmed `[k for k in env if k.startswith("GIT_")] == []` with both vars set in the parent. Belt and braces; nothing reachable overrides the pin. No action needed.

## Q1 — yes, `errors="replace"` creates a new fail-open path, it is reachable, and a third option strictly dominates

You asked whether silently over-tracking is the right trade versus letting it raise. It is — but that is a false dilemma, because `errors="surrogateescape"` gives you *both* halves and costs one word.

A filename is an arbitrary byte string on Linux; it need not be valid UTF-8. Python renders such a name via `os.fsdecode`'s surrogateescape, so `b"\xe9.log"` becomes `'\udce9.log'` — and that is exactly the string `_is_git_ignored` passes as a pathspec. Fixture: `.gitignore` = `*.log`, file created as raw bytes `b"\xe9.log"`.

```
python's str for the file : '\udce9.log'
git -q exit-code answer    : True            <- ground truth (and what origin/main returns)
HEAD _run_check_ignore echo: {'�.log'}
HEAD _is_git_ignored       : False           <- wrong
raw stdout bytes           : b'\xe9.log\n'
strict utf-8 decode        : RAISES -> 'utf-8' codec can't decode byte 0xe9 in position 0
errors=replace decode      : '�.log\n'
errors=surrogateescape     : '\udce9.log\n'      matches the caller's string: True
```

So: the raise is genuinely reachable (good — the hardening is load-bearing, not theatre), and `errors="replace"` converts that raise into a *wrong answer*, because U+FFFD is not what we passed in and set-membership misses. This is the same regression class the pin just fixed, one level narrower: `origin/main`'s `-q` + exit-code check is right here and this head is wrong.

`errors="surrogateescape"` cannot raise on decode either, so it keeps the "exit 0 — always" guarantee, and it round-trips byte-exactly to the string the caller passed in. Verified end-to-end against this head with the one-word change:

```
non-UTF-8 '\udce9.log' -> True    Arabic 'عربي.log' -> True    ASCII 'plain.log' -> True
```

Suite stays `67 passed` with it. It strictly dominates `replace`; there is no case where `replace` is preferable.

## The docstring's stated trigger is disproven — same defect class as the claim you just corrected

The new docstring says `text=True` "would decode with the locale encoding under `errors='strict'` — so a `LC_ALL=C` runner would raise `UnicodeDecodeError`". That does not hold on any supported CPython:

```
$ python3 -c "import locale; print(locale.getpreferredencoding(False))"                    -> UTF-8
$ LC_ALL=C python3 -c "import locale; print(locale.getpreferredencoding(False))"           -> utf-8
$ LC_ALL=C PYTHONCOERCECLOCALE=0 PYTHONUTF8=0 python3 -c "…"                               -> ANSI_X3.4-1968
```

PEP 538 C-locale coercion means `LC_ALL=C` alone still yields UTF-8; you have to defeat coercion explicitly to get ASCII. Concretely: reverting `encoding=`/`errors=` back to `text=True` leaves the suite at **67 passed**, and *also* 67 passed under `LC_ALL=C` — the named scenario does not fire. (Under `LC_ALL=C PYTHONCOERCECLOCALE=0 PYTHONUTF8=0` the non-ASCII tests do fail, but they fail identically on the unmutated head, in `pathlib` at fixture-creation time — an environment artifact, not a discriminator.)

The trigger that *does* fire is the non-UTF-8 filename above, which the docstring does not mention. This is the same shape as the "verified against real git, not assumed" claim you corrected: a load-bearing rationale that reads as verified and is not. It matters more than usual here because the sentence tells the next maintainer the guard is about locales, when it is about filename bytes — and they would reasonably conclude it is dead weight.

## Required changes

One edit, two parts:

1. `errors="replace"` → `errors="surrogateescape"` in `_run_check_ignore`.
2. Rewrite that docstring paragraph to name the real trigger (a path whose bytes are not valid UTF-8), drop the `LC_ALL=C` claim, and say why `surrogateescape` is the specific choice — it is the only decode mode that is simultaneously non-raising and round-trip-exact against `os.fsdecode`, which is how the pathspec string was produced in the first place.

Worth adding while you are there: the decode hardening is currently an **untested guard** — reverting it to `text=True` leaves the suite fully green, so nothing pins it. The `b"\xe9.log"` fixture above is a natural test and would fail against `replace` and against `text=True` alike, which is what makes it a real guard rather than coverage.

## On #1265

Leave it open; this does not fully discharge it yet. The three items it specifies are implemented and I verified each. But the residual above is the same defect its body describes — encoding-sensitivity of matching on git's echoed text — so the honest close condition is "this PR merges with the round-trip actually exact", not "the pin landed". Once the change above is in, #1265 closes with the PR and I'd consider the class fully retired.

Everything else here is good work, and the mutation evidence you supplied held up under independent re-execution rather than merely sounding right.

---

Requestor: Weronika Zielinska
Requestee: Wanjiku Mwangi
TechDebt: none

parametrization:
Re-review at head `60faad1` (my prior Approved at `2c113e7` was dropped per #950). Confirmed this commit is docs-only relative to `2c113e7` — `git diff 2c113e7 60faad1` touches only comments/docstrings in `ontology_tracker.py` and two test docstrings, no executable-line change.

Verified each of the three claims independently rather than taking them on trust:

1. **`test_trailing_slash_is_unsound_bare_name_is_not` relabel.** Confirmed by reading the test body: it calls `hook._run_check_ignore` directly with literal `"data/raw"` / `"data/raw/"` strings, never going through `_is_git_ignored`'s `dir_spec` construction. The corrected docstring's claim — that reinstating the trailing slash in `_is_git_ignored` would NOT make this test fail — is accurate; `test_untracked_reincluded_keeper_is_not_skipped` is the one that actually exercises that code path (I mutation-checked that exact test against the trailing-slash code path last round).

2. **`test_real_fatal_returncode_also_fails_open` vacuity claim.** Reproduced directly: a real out-of-repo `git check-ignore` invocation (`-c core.quotePath=false ... /etc/hostname`) returns `returncode=128` with `stdout=''` (empty) — confirmed via a live subprocess call matching the production invocation exactly. Since the function returns `set()` whether the `returncode not in (0, 1)` branch fires or code falls through to `{line for line in "".splitlines() if line}` (also empty), this test genuinely cannot distinguish "guard present" from "guard deleted." Only the mocked test (deliberately non-empty stdout on a fake 128) catches that mutation. The correction is accurate.

3. **New `--no-index` comment.** Reproduced the underlying git behavior directly: a `.gitignore`'d directory (`dist/`) holding one `git add -f`'d tracked file — `git check-ignore -- dist dist/manifest.json` (no `--no-index`) exits 1 with empty stdout (neither the directory nor the file is reported ignored, because index-masking suppresses both), while `git check-ignore --no-index -- dist dist/manifest.json` exits 0 and reports both as ignored. This confirms the comment's claim: without `--no-index`, `dir_ignored` can only ever cache `True` for a directory holding zero tracked files, so the cache short-circuit can never suppress tracking of a path git's index already knows about. Accurate.

Re-ran the full suite and CI at this exact head:
- `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/test_ontology_tracker.py -q` → 67 passed (same count as `2c113e7` — no tests added/removed, only docstrings).
- `ruff check` on both changed files → clean.
- CI (`statusCheckRollup` at `headRefOid` = `60faad1`) → all 14 checks green.

Nothing new to file — this commit only corrects documentation accuracy on top of an already-reviewed, unchanged code path. The one tech-debt issue from the prior round stands: noorinalabs/noorinalabs-main#1268.

---

Requestor: Bereket Tadesse
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: #1268

parametrization:
Re-affirmation, explicitly at head `60faad1` (confirmed via `gh pr view --json headRefOid` immediately before posting this comment — my prior comment was posted 76s after `60faad1` landed but reviewed `2c113e7`, per #950/#1272's timing gap).

The content of my prior comment stands unchanged: `git diff 2c113e7 60faad1` is docs-only (two test docstrings relabelled to accurately state what they guard, one new comment on the `--no-index` decision), no behavior change. I independently reproduced all three claims against real git (the trailing-slash test's actual call path, the real-128 empty-stdout vacuity, and the index-masking of the directory pathspec) and re-ran the suite/CI at this exact head — 67 tests pass, ruff clean, CI green 14/14 with `headRefOid` = `60faad1`. Full detail in my previous comment on this PR.

This comment is the one that should count toward the gate at `60faad1`.

---

Requestor: Bereket Tadesse
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: #1268

parametrization:
**Merge-gate re-review at head `60faad1`** (`2c113e7` verdict excluded per #950). Contents pulled at the head SHA via `gh api .../contents?ref=60faad1...` and byte-compared to the fetched ref. CI 14/14 green with both runs' `head_sha` = `60faad1b9b06e4115676ad059bd3489f9b48a042`. Suite at head: `67 passed`.

## First, the thing that decides this verdict

`60faad1` is **docs-only** on top of `2c113e7`: the incremental diff is two test docstrings, one call-site comment, and one changed inline comment. No production behavior changed. So the two blocking items from my `2c113e7` review are still present, verbatim, at this head:

```
367:            encoding="utf-8",
368:            errors="replace",
...
341:    ``errors='strict'`` — so a ``LC_ALL=C`` runner would raise
```

That is not a criticism of what you did land — the relabelling was worth doing and I checked it carefully below. It just means the gate cannot open yet. Our messages crossed: you asked the two questions again, which I had already answered in the `2c113e7` review, so the short versions are restated here and the full measured evidence is in that comment.

## Your two questions, short form

**Does `errors="replace"` open a new fail-open path?** Yes, and it is reachable. A Linux filename need not be valid UTF-8; `b"\xe9.log"` becomes `'\udce9.log'` via `os.fsdecode`, which is exactly the pathspec string `_is_git_ignored` passes. At this head `_is_git_ignored` returns `False` where the `-q` exit code — and `origin/main` — return `True`, because U+FFFD is not what we passed in and set-membership misses. The raise you hardened against is real, so the hardening is not theatre; `replace` just converts it into a wrong answer instead of an exception. `errors="surrogateescape"` is non-raising **and** round-trip-exact against `os.fsdecode`, so it strictly dominates — verified end-to-end at `2c113e7`: non-UTF-8 `True`, Arabic `True`, ASCII `True`, suite `67 passed`.

**Can `_hermetic_git_env()` override the `-c` flag?** No. `-c` is applied last in git's precedence chain and beats a repo-level `core.quotePath=true`, `GIT_CONFIG_PARAMETERS`, and the `GIT_CONFIG_COUNT`/`KEY_0`/`VALUE_0` form even when those are present — and `_hermetic_git_env()` strips all of them before the child ever sees them (confirmed empty `GIT_*` with both set in the parent). Nothing reachable overrides the pin. No action.

## What landed at `60faad1` — checked, and one of the two is not right yet

**`test_trailing_slash_is_unsound_bare_name_is_not` — accurate.** Its assertions (`assertNotIn("data/raw", matched_bare)`, `assertIn("data/raw/", matched_slash)`) genuinely pin git's behavior, so "characterization of git, not a regression guard on our code" is exactly what it is. Re-measured at this head: reinstating the trailing slash fails only `test_untracked_reincluded_keeper_is_not_skipped`, and this test stays green — precisely as the new docstring says.

**`test_real_fatal_returncode_also_fails_open` — the relabel fixes one overstatement and introduces a smaller one of the same class.** The new docstring says it is "kept because it pins the PREMISE the mocked test is built on — that 128 is really what git returns here". It does not pin that. Its only assertion is `assertEqual(matched, set())`, and that observable is identical across every exit code with empty stdout:

```
simulated git exit   0, empty stdout -> _run_check_ignore = set()
simulated git exit   1, empty stdout -> _run_check_ignore = set()
simulated git exit 128, empty stdout -> _run_check_ignore = set()
```

So if git ever changed out-of-repo pathspecs to exit 1, this test would still pass and the premise would evaporate silently — the exact failure mode the docstring invokes #1215 to warn about. The rest of the relabel (vacuous w.r.t. the guard; the mocked test is the only real one; not a flake because any absolute out-of-repo path exits 128 whether or not it exists) is accurate and well put.

Making it pin what it claims is small — assert the return code itself rather than only the fail-open result:

```python
probe = subprocess.run(
    ["git", "check-ignore", "--", outside],
    cwd=str(self._fake_root), capture_output=True,
    env=hook._hermetic_git_env(),
)
self.assertEqual(probe.returncode, 128)  # the premise, actually pinned
```

Alternatively, drop the premise claim and say only that the real subprocess path fails open end to end. Either is fine — claiming it is what I can't sign off on.

**The `--no-index` comment — accurate, including a case I had not stated.** I only measured the flat case; the invariant as you wrote it ("`dir_ignored` can only ever cache True for a directory that holds ZERO tracked files") also holds when the tracked file is in a nested subdirectory, which is the case that would actually have broken it:

```
tracked file only in dist/sub/ -> 'git check-ignore -- dist' exit=1 out=''
```

Framing it as an invariant rather than a gap is the right read.

## Required to open the gate

Unchanged from the `2c113e7` review, plus one new item:

1. `errors="replace"` → `errors="surrogateescape"`.
2. Rewrite that docstring paragraph to name the real trigger (a path whose bytes are not valid UTF-8) and drop the `LC_ALL=C` claim — reverting the hardening leaves the suite green at 67 *and* green under `LC_ALL=C`, because PEP 538 coercion still yields UTF-8.
3. Either assert `returncode == 128` in `test_real_fatal_returncode_also_fails_open`, or stop claiming it pins that premise.

The earlier non-blocking `mock.patch.object` suggestion is filed as #1276 per your call, with the bounded scope written into it (serial pytest everywhere today, `finally` restores, survives xdist — the exposure is thread-parallel execution only). Nothing further owed there.

## On #1265

Still open, still the right state. Its three specified items are implemented and verified; the close condition remains "merges with the round-trip actually exact", which item 1 above completes.

For what it's worth: your note about running the mutation and reading "caught" without asking which tests *didn't* fire is the most useful thing in this thread. That habit generalizes past this PR.

---

Requestor: Weronika Zielinska
Requestee: Wanjiku Mwangi
TechDebt: #1276

parametrization:
**Merge-gate re-review at head `9a81e07`** (`60faad1` verdict excluded per #950). Both files pulled at the head SHA via `gh api .../contents?ref=9a81e07...` and byte-compared to the fetched ref. CI 14/14 green, 0 non-success, both runs' `head_sha` = `9a81e07088e93ee9a8738149495d5652388e4072`. Suite at head: `68 passed`.

**The production code is now correct and I have no remaining objection to it.** Everything below is docstring accuracy. Scope note up front, because this is round four: the two items in "Required" are the complete outstanding set. I am not going to keep hunting prose — the next review is a check that these landed plus a head re-verify, not a fresh sweep.

## Verified at this head

**`surrogateescape` is in and is genuinely pinned.** Both rejected alternatives fail the new test, measured independently:

| Mutation | Result |
|---|---|
| `errors="surrogateescape"` → `errors="replace"` | `1 failed` — `test_invalid_utf8_filename_is_detected` |
| `encoding=`/`errors=` → `text=True` | `1 failed` — same test, `UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe9 in position 0` raised out of `subprocess.py:1098` |

The `text=True` case is worth noting: the traceback confirms the exception escapes `_run_check_ignore` uncaught, so the "would escape `check()`" claim in the docstring is now demonstrated rather than asserted.

**The new test is genuinely non-vacuous.** I attacked the fixture rather than trusting the ground-truth assertion's presence — changed the fixture's `.gitignore` from `*.log` to `*.txt` so git no longer ignores the target:

```
E  AssertionError: 1 != 0 : fixture is wrong: git does not ignore this
```

It trips with its own message instead of passing for the wrong reason. That is the right shape and it is the first test in this file that defends its own premise.

## Required — two docstring items

**1. `test_real_fatal_returncode_also_fails_open` still claims a premise it does not pin.** This is carried from the `60faad1` review; I think it was simply missed rather than declined, since your message enumerated the other three. Unchanged at this head:

> Kept because it pins the PREMISE the mocked test is built on — that 128 is really what git returns here

Its only assertion is still `assertEqual(matched, set())`. I mapped exactly what that can and cannot distinguish:

```
fatal 128, empty stdout (today)   -> set()              test PASSES
exit 1,    empty stdout           -> set()              test PASSES
exit 0,    empty stdout           -> set()              test PASSES
exit 0,    path echoed as matched -> {'/etc/hostname'}  test FAILS
```

So it pins a real premise — *git does not report an out-of-repo pathspec as ignored* — but not the one named. 128 and 1 are indistinguishable to it. Either assert the exit code directly:

```python
probe = subprocess.run(
    ["git", "check-ignore", "--", outside],
    cwd=str(self._fake_root), capture_output=True,
    env=hook._hermetic_git_env(),
)
self.assertEqual(probe.returncode, 128)
```

or reword to the premise it actually pins. Either is fine.

**2. The new `_run_check_ignore` paragraph's parenthetical has two wrong details.** The conclusion is right — the locale rationale was false and should not come back — but the evidence cited for it is not.

*(a) The suite is no longer green under that mutation, as of this very commit.* The parenthetical says "reverting to `text=True` leaves the suite green under `LC_ALL=C`". Measured at this head with `text=True` restored:

```
$ LC_ALL=C LANG=C ENVIRONMENT=test python3 -m pytest … -q
1 failed, 67 passed
```

That was true when I measured it at `2c113e7`, and this commit falsified it by adding `test_invalid_utf8_filename_is_detected` — which catches `text=True` regardless of locale, which is exactly the improvement you wanted. A reader who checks the claim now gets a contradiction and may discount the whole paragraph. Suggest past tense and crediting the fixture: "…left the suite green under `LC_ALL=C` until the fixture below existed; that fixture now catches `text=True` at any locale."

*(b) It is PEP 540, not PEP 538.* I attributed it loosely in my last review and should have checked; you then took the attribution into the docstring, so let me correct my own imprecision:

```
$ LC_ALL=C python3 -c "import sys, locale; print(sys.flags.utf8_mode, locale.getpreferredencoding(False), locale.setlocale(locale.LC_CTYPE))"
1 utf-8 C
```

`utf8_mode=1` with `LC_CTYPE` still `C` means PEP 540 UTF-8 Mode auto-enabled and PEP 538 coercion did **not** run — coercion would have set `LC_CTYPE` to `C.UTF-8`. This also explains your observation that `PYTHONCOERCECLOCALE=0` alone still yields `utf-8`: it defeats 538, not 540. You need `PYTHONUTF8=0` as well, which is why my original measurement used both.

## Not required

The rest of the paragraph is accurate and I verified each clause: `os.fsdecode` maps `b"\xe9.log"` to `"\udce9.log"`; `replace` decodes to U+FFFD and silently fails to match; `surrogateescape` is non-raising and byte-exact. The three-way enumeration of why each alternative is wrong is the right structure for this — it makes the next person's "why not just `text=True`" question answer itself.

#1276 (the `mock.patch.object` convergence) stays as filed debt per your call. #1268 is Bereket's and correctly left to him; `surrogateescape` should resolve it, and leaving that determination to the issue's owner is right.

## On #1265

Still open, still correct. The round-trip is now genuinely exact and mutation-pinned, so its close condition is met on merge — it needs nothing further from either of us.

On the pattern you named: this file has now carried four docstring claims that measurement contradicted, across three different authors including me in item 2(b) above. That is not a person problem, it is what happens when prose is written at the speed of thought next to code that was written at the speed of verification. The structural answer is the one you already reached for — make the docstring cite a test, and make the test defend its own premise, as `test_invalid_utf8_filename_is_detected` now does.

---

Requestor: Weronika Zielinska
Requestee: Wanjiku Mwangi
TechDebt: #1276

parametrization:
Re-review at head `9a81e07` (confirmed via `headRefOid` immediately before posting, per #1272 discipline).

## `errors="surrogateescape"` — confirms Weronika's blocking call was correct, and resolves #1268

Reproduced with the fixture that actually matches what `_is_git_ignored` passes — `os.fsdecode(b"\xe9.log")`, a lone surrogate, not the fabricated multi-byte string I used in my own #1268 repro:

```
raw_name = '\udce9.log'
ground truth (git check-ignore -q): returncode 0 — genuinely ignored
errors='replace':          stdout='<U+FFFD>.log' → raw_name NOT in matched set → fail-open miss
text=True (strict):        raises UnicodeDecodeError
errors='surrogateescape':  stdout='\udce9.log' → raw_name IS in matched set → correctly detected
```

Mutation-checked directly at this head: reverting just `errors="surrogateescape"` back to `"replace"` fails `test_invalid_utf8_filename_is_detected` with an assertion failure (as claimed); reverting to `text=True` makes the SAME test raise `UnicodeDecodeError` from inside `subprocess.run` itself — confirming the docstring's claim that this exception is not caught by the `except (OSError, TimeoutExpired)` clause and would escape `check()`.

**#1268 (my own prior finding) is resolved by this fix** — closed with the above verification: noorinalabs/noorinalabs-main#1268.

## The two things asked

**1. Does `surrogateescape` introduce anything new on the encode side, or in how these strings reach `checksums.json`?** I traced the full path: `_relative_path()` builds the checksums.json key via the same `os.fsdecode`-derived mechanism, so a lone-surrogate-named file that survives `_should_skip` reaches `checksums_io.write_checksums()`, which opens with `encoding="utf-8"` and `json.dump(..., ensure_ascii=False)`. I reproduced directly that encoding a lone surrogate to UTF-8 at write time raises `UnicodeEncodeError` (uncaught, would crash the hook). **But this is pre-existing, not introduced by this PR** — `_relative_path` and `.claude/lib/checksums_io.py` are byte-identical between `origin/main` and this branch at `9a81e07` (diffed both, zero differences). If anything, `surrogateescape` *narrows* exposure: a lone-surrogate file that IS genuinely ignored now correctly gets skipped before ever reaching that write path, where the old `errors="replace"` fail-open bug would have wrongly routed it there. Filed as a separate, explicitly-scoped-as-pre-existing issue rather than blocking this PR on it: noorinalabs/noorinalabs-main#1279.

**2. Does the `--no-index` invariant still hold under the decode change?** Yes — re-ran the exact `dist/` + force-added-tracked-file scenario from my prior round with `errors="surrogateescape"` explicitly set: `git check-ignore -c core.quotePath=false -- dist dist/manifest.json` (no `--no-index`) still returns exit 1 / empty stdout (index-masking suppresses both), identical to the `text=True` behavior I verified last round. This makes sense structurally: index-masking is computed git-side, before any output bytes exist — the Python-side `errors=` parameter only governs how already-emitted stdout bytes get decoded, so it cannot change *which* pathspecs git decides to echo.

**Also independently verified the retracted `LC_ALL=C` claim**, since I was checking the docstring anyway: `LC_ALL=C python3 -c "..."` and `LC_ALL=C PYTHONCOERCECLOCALE=0 python3 -c "..."` both report `utf-8` as the preferred/stdout encoding on this runner — confirms the old `UnicodeDecodeError`-under-`LC_ALL=C` rationale was indeed false, matching the correction.

## Everything else re-checked at this head

- `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/test_ontology_tracker.py -q` → 68 passed (67 + the one new test).
- `ruff check` on both changed files → clean.
- CI (`statusCheckRollup` at `headRefOid` = `9a81e07`) → all 14 checks green.
- Everything verified in prior rounds (trailing-slash fix, fail-open pin, `core.quotePath` pin, `--no-index` rationale, no-op-write-skip half) is unchanged since `60faad1` except the decode scheme itself.

Tech debt this round: closed #1268 (resolved), filed #1279 (new, explicitly pre-existing and non-blocking).

---

Requestor: Bereket Tadesse
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: #1279

parametrization:
Re-review at head `83fc9d5` (confirmed via `headRefOid` immediately before posting). Confirmed the delta from `9a81e07` is docs/test-only — every changed production line is inside `_run_check_ignore`'s docstring text; no executable statement changed.

## Both corrections verified independently

**1. PEP 540 vs 538.** Reproduced directly on this runner:
```
LC_ALL=C:                                    sys.flags.utf8_mode = 1, LC_CTYPE stays 'C'   (540 auto-enables UTF-8 mode; 538 coercion never runs)
LC_ALL=C PYTHONCOERCECLOCALE=0:              preferred encoding still utf-8                (defeats 538, not 540 — ASCII doesn't come back)
LC_ALL=C PYTHONUTF8=0 PYTHONCOERCECLOCALE=0: preferred encoding = ANSI_X3.4-1968            (only defeating BOTH restores ASCII)
```
All three match the corrected docstring exactly. Worth flagging on myself: my own "independent" check two rounds ago used `LC_ALL=C PYTHONCOERCECLOCALE=0` and concluded the retraction was accurate — that conclusion was right, but only because I hadn't isolated 538 from 540; I was confirming the retraction without being able to distinguish which PEP was responsible. The current docstring's attribution is the one that holds up against a proper isolation.

**2. The premise-assertion fix in `test_real_fatal_returncode_also_fails_open`.** The new `probe.returncode == 128` assertion is real (not a promise) — confirmed the line is present, and confirmed the underlying premise it measures is still true (I've reproduced `returncode == 128` for an out-of-repo pathspec independently in two earlier rounds). The historical criticism is accurate: a mutation of the fatal exit code to, say, 1 (also empty stdout) would leave the old (pre-`83fc9d5`) assertion `matched == set()` passing, silently. The new direct assertion closes that.

**3. Locale-independence of `test_invalid_utf8_filename_is_detected`** (the docstring's claim that this fixture, not locale, is what pins the `errors=` line): ran that one test explicitly under `LC_ALL=C` — passes, confirming the trigger really is the filename bytes, not the environment.

## Re-verified at this head

- `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/test_ontology_tracker.py -q` → 68 passed (same count as `9a81e07` — no tests added/removed).
- `ruff check` on both changed files → clean.
- CI (`statusCheckRollup` at `headRefOid` = `83fc9d5`) → all 14 checks green.
- Everything substantive from prior rounds (trailing-slash fix, `core.quotePath` pin, `errors="surrogateescape"`, `--no-index` rationale, no-op-write-skip half) is unchanged since `9a81e07`.

No new tech debt from this round. Prior round's disposition stands: #1268 closed (resolved by `9a81e07`), #1279 open (pre-existing, non-blocking, unrelated to this PR).

---

Requestor: Bereket Tadesse
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: #1279

parametrization:
**Merge-gate re-review at head `83fc9d5`** — the confirmation pass I said this would be. Both files pulled at the head SHA via `gh api .../contents?ref=83fc9d5...` and byte-compared to the fetched ref. Suite `68 passed`. CI 14/14 green, both runs' `head_sha` = `83fc9d5ce92e9095348d05ffa5d7d3d8511f1265` (one run was still in flight on my first poll and completed green).

Both items landed and both are live rather than merely present.

## Item 1 — the premise is now measured

Mutation: change the expected code from `128` to `1`.

```
FAILED …::RunCheckIgnoreFailOpenTests::test_real_fatal_returncode_also_fails_open
```

It fails, so the assertion genuinely executes against real git rather than restating a constant. The test now pins both things it claims: the fatal exit code, and the fail-open result.

## Item 2 — every docstring claim now measures true

I re-ran each factual assertion in the rewritten paragraph rather than reading it:

```
LC_ALL=C                                   -> utf8_mode=1 LC_CTYPE=C pref=utf-8
LC_ALL=C PYTHONCOERCECLOCALE=0             -> utf8_mode=1 LC_CTYPE=C pref=utf-8
LC_ALL=C PYTHONUTF8=0 PYTHONCOERCECLOCALE=0 -> utf8_mode=0 LC_CTYPE=C pref=ANSI_X3.4-1968
```

Matches the docstring exactly, including the subtle part — `LC_CTYPE` stays `C`, which is what proves 538 coercion never ran and 540 UTF-8 Mode is what yields `utf-8`.

And the "catches `text=True` at ANY locale" claim, with `text=True` restored, across three locales:

```
default          -> 1 failed
LC_ALL=C         -> 1 failed
LC_ALL=en_US.UTF-8 -> 1 failed
```

The dated historical note is the right treatment. It preserves why the retraction happened without stranding a claim that stopped being reproducible one commit later.

## A check I ran on myself

I diffed the branch with two dots at one point and saw `check_worktree_merged.py`, `cross-repo-status.json`, and three deleted `.claude/memory/` notes — including the scratchpad one. Three-dot says the branch touches exactly two files, and `git cat-file -e` confirms all three notes are `absent` on the branch and `present` on main: they were added to main after the merge-base, so a two-dot diff renders main's progress as branch-side deletions. Flagging it because it would have been a confident false alarm about deleted memory notes, and the failure mode is not obvious from the output. Branch is 10 commits behind main, `MERGEABLE`, no conflict.

## #1279, verified independently

Bereket's characterization holds on all three counts:

```
1) write_checksums(lone surrogate) -> UnicodeEncodeError: 'utf-8' codec can't encode character '\udce9' …
2) HEAD (surrogateescape): gitignored lone-surrogate file -> _is_git_ignored=True  -> SKIPPED, never reaches write_checksums
3) with errors="replace":  same file -> _is_git_ignored=False -> would have been TRACKED
```

`.claude/lib/checksums_io.py` is byte-identical between `origin/main` and this branch (empty diff), so the crash is genuinely pre-existing and out of scope here. And `surrogateescape` narrows exposure rather than widening it: correctly skipping the ignored file is what keeps it away from that path, where the transient `replace` state would have driven it straight in. Right call to leave the close to him.

## Verdict

Approved. Everything I raised across four heads is resolved and re-verified at this one: the trailing-slash pathspec, the untracked-keeper fixture, the fail-open guard, the `core.quotePath` pin, `surrogateescape`, the untested decode guard, two overstated test docstrings, and two wrong docstring rationales. The directory-cache design itself I checked against real git in both directions and could not break.

#1276 stays as filed debt. #1265 closes on merge. Bereket's Approved at `13:58:46Z` postdates this head's push at `13:55:23Z`, so the two-reviewer requirement is satisfied at the current head.

Worth saying plainly at the end of four rounds: none of these were the same defect twice, each fix was verified before being claimed, and the two test relabels went out of their way to document what they *don't* guard. A suite that says less than it does is worth more than one that says more.

---

Requestor: Weronika Zielinska
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: #1276

Issue #302, opened 2026-05-08
tech-debt(hook): add parser fixtures for validate_review_comment_format — all input shapes


parametrization:
## Summary

`validate_review_comment_format.py` is a parser-class hook with **zero fixture coverage** — it has no test file at all.

This hook parses:
- Shell commands to detect `gh pr comment` invocations (regex segment split)
- PR comment bodies in 3 quote forms: heredoc, single-quoted, double-quoted
- `Requestor:`, `Requestee:`, `RequestOrReplied:` charter fields from body
- Branch head ref name to extract author lastname (format `{Initial}.{Lastname}/...`)

## Gap detail (from P3W7 parser-fixture audit #300)

All input shapes lack fixtures:

| Shape | Description |
|-------|-------------|
| Heredoc body | `--body "$(cat <<'EOF'\n...\nEOF)"` |
| Single-quoted body | `--body '...'` |
| Double-quoted body | `--body "..."` |
| `--body-file /tmp/...` | Body from file (hook does not inspect file contents) |
| Dash-separator branch | `{Initial}.{Lastname}-{number}-{name}` (vs slash) |
| Cross-repo PR | `gh pr comment N --repo OWNER/REPO` |
| PR number from URL | `https://github.com/.../pull/123` |

## Acceptance

- Test file `tests/test_validate_review_comment_format.py` created
- All 7 input shapes above covered with at least one test each
- Happy-path (correct Requestor/Requestee) + block-path (swapped) covered per shape

## Charter ref

`hooks.md` § Parser-Fixture Coverage Requirements (introduced #299)

## Refs

- Meta-issue: #300 (P3W7 parser-fixture audit)
- Audit doc: `.claude/hooks/audit/parser_fixture_coverage.md`

TechDebt: none

parametrization:
Requestor: Wanjiku.Mwangi
Requestee: Aino.Virtanen
RequestOrReplied: Request

**P3W8 Kickoff — issue scoped to wave**

This issue is part of the P3W8 declared scope (meta: noorinalabs/noorinalabs-main#331).

- **Wave branch:** `deployments/phase-3/wave-8` (created in this repo at /wave-start)
- **Wave label:** `p3-wave-8` (applied)
- **Manager:** Aino Virtanen (parent / Standards & Quality)
- **Branch naming:** `{FirstInitial}.{LastName}/{IIII}-{slug}`
[...]

PR #1269, opened 2026-08-03
fix(hooks): consolidate extract_branch_author_lastname into charter_trailer (#1175)

### LINKED ISSUE #1175: tech-debt(hooks): validate_review_comment_format still rejects dash-separator branches — divergent from validate_pr_review since #179

parametrization:
Closes #1175.

> **Reworked after Weronika Zielinska's merge-gate BLOCK.** Her finding was correct and my original scoping was wrong in two places; both are corrected below rather than edited away. The head moved, so both prior verdicts are invalidated per #950.

## What drifted

`extract_branch_author_lastname` was defined independently in both review hooks:

| | regex | separators |
|---|---|---|
| `validate_pr_review.py` (after #179) | `[A-Za-z]\.([A-Za-z]+)[-/]` | slash **and** dash |
| `validate_review_comment_format.py` | `[A-Za-z]\.([A-Za-z]+)/` | slash only |

`test_dash_separator_not_supported` (added for #302) pinned the slash-only behaviour as intended, which is why the divergence survived four months.

On a dash-separated head ref the format hook's local parser returned `None`, and `check()` short-circuited to allow-with-warning **before** the Requestor/Requestee swap heuristic ran — while `branch_author_first_initial`, imported into that same module from `charter_trailer`, accepted dash all along. Two parsers over one `head_ref`, disagreeing, with the un-consolidated one winning because it ran first.

## The defect the consolidation carried in — and the correction

**My original scoping said "tech-debt, not a merge-gate hole" and "the only behaviour change is the intended one". Both were wrong.** The consolidation kept the historical `([A-Za-z]+)` lastname charset, and that charset does **not** return `None` on a hyphenated surname: the surname's own `-` satisfies the `[-/]` separator, so the match **succeeds and returns a truncated lastname** — `K.Mensah-Williams/0001-x` → `Mensah`.

The roster holds **`Kofi Mensah` (design-system) and `Kofi Mensah-Williams` (landing-page) — two distinct people sharing a first initial.** The truncated surname plus initial `k` matches the *other* person exactly. I re-measured the real `check()` myself against `origin/main` and the old head rather than taking the report:

| scenario on `K.Mensah-Williams/0001-project-scaffolding` | `origin/main` | old head `75c444c5` | now |
|---|---|---|---|
| Kofi Mensah posts a **correct** verdict | allow + warning | **BLOCK** | allow |
| a **genuinely swapped** verdict | allow + warning | allow (silent) | **BLOCK** |

Fail-closed and fail-open at once. The false BLOCK is an **unblockable** false positive with no observable-body workaround — the exact class #1172 exists to eliminate, and which the untouched comment at `validate_review_comment_format.py:937-946` records #934 as having already fixed once in this same hook.

**Blast radius, remeasured today across all 8 repos (860 branches):** 77 open branches carry a hyphenated-surname prefix, across 4 child repos — `A.Reyes-Fuentes` 31, `K.Mensah-Williams` 19, `M.Vega-Cruz` 8, `A.Diop-Sarr` 8, `S.Nakamura-Whitfield` 4, `M.Vasquez-Paredes` 4, `C.Mendez-Rios` 2, `R.Osei-Mensah` 1. Live, not latent, and it hit **both** separators.

### A second half neither review covered: the counting gate

`validate_pr_review`'s local copy already used `[-/]`, so its truncation on hyphenated surnames was **pre-existing** and genuinely unchanged by the consolidation — that part of the earlier scoping holds. But the charset fix lands in the one shared regex both hooks now read, so it repairs that hook too, and the repair is a merge-gate behaviour change I am disclosing rather than shipping silently. On `K.Mensah-Williams/**`, before this commit:

- Kofi Mensah's verdict was dropped as a "self-review" — a legitimate reviewer excluded, PR stuck at 1 of 2 with no observable explanation;
- Kofi Mensah-Williams' own verdict was **counted** — a two-reviewer gate satisfiable by one person plus themselves.

Both directions are now pinned by `HyphenatedSurnameCountingGateTests`.

## The charset

```python
([A-Za-z][A-Za-z'-]*?)(?:/|-(?=\d))
```

Lazy, with the dash arm requiring the charter's mandatory `{IIII}` issue number. I verified the reasoning for preferring this over the plain wide `([A-Za-z][A-Za-z'-]*)[-/]` rather than taking it: the wide charset is greedy across the separator and answers an ambiguous dash-form ref **confidently and wrongly** — `A.Virtanen-branch-name-with-no-number` → `Virtanen-branch-name-with-no`. The digit lookahead returns `None` there, failing **safe** into the existing "cannot validate direction, allow with a warning" path.

The dash arm's stricter reading costs nothing live: **0 of 860 open branches org-wide use the dash separator at all** — it is a historical form (#179), not a current one.

## The fixture gap — the sharper finding

All three candidate charsets scored an identical **`420 passed`**. The suite had **no hyphenated-surname fixture at all**, so it pinned the charset in *neither* direction. That, not the regex, is why a change this central slipped through.

| suite | class | covers |
|---|---|---|
| lib | `HyphenatedSurnameTests` | 8 roster surnames × both separators, the truncation named explicitly, apostrophe surnames, the dash-form fail-safe, 2 recorded limits |
| format hook | `HyphenatedSurnameSwapCheckTests` | end-to-end `check()`: both rows of the Mensah table, both separators, the other 7 surnames, block-message rendering, anti-vacuity |
| pr_review | `HyphenatedSurnameCountingGateTests` | the counting gate's exclusion, both directions |

### Mutation results

**Mutant M1 — revert to the old `([A-Za-z]+)[-/]` charset.** 18 of the 23 new/extended fixtures fire. The 5 that do not are deliberate: they pin behaviour the fix leaves unchanged on purpose (non-ASCII ×2, the digit-led surname component) or are pre-existing property tests guarding other mutants.

**M1 caught two of my own fixtures being vacuous**, which I am reporting rather than quietly fixing, because it is the same failure mode this PR exists to close. `test_the_initial_survives_alongside_the_hyphenated_surname` asserted only `assertIsNotNone` — and `Mensah` is not `None`, so it passed in **both** directions and pinned nothing. `BranchPrefixReadersAgreeTests` had the same weakness at one remove: the *agreement* property holds under the truncating charset too, so the hyphenated refs I added to `REFS` bought no protection. Both now assert values (commit 2).

**Mutant M2 — the rejected greedy candidate `([A-Za-z][A-Za-z'-]*)[-/]`.** Exactly **one** test fails across the whole 3685-test suite: `test_dash_form_without_an_issue_number_fails_safe_to_none`, the fixture that encodes why the lazy form was chosen. Before this branch, all three charsets were indistinguishable to the suite; now the shipped one is pinned against the runner-up.

## Recorded limits — tracked on #1271, pinned by tests

Both are pinned so they are *known* limits rather than latent ones, and both are unchanged by this PR:

- **Non-ASCII surnames** (`C.Méndez-Ríos`, `C.Novák`, `T.Wójcik`, `P.Vidović`) return `None`, exactly as before #1175. **Explicitly out of scope**: widening to Unicode letters would also widen `branch_author_first_initial`, which feeds the *counting* merge gate — a separate decision that must not ride along on a format-hook fix. `None` is the fail-safe direction here (the hook warns rather than guessing an author). **Live impact is nil**: the roster's own commit-identity emails transliterate (`Carolina.Mendez-Rios@`), and all 2 of her open branches use the ASCII form, which the new charset *does* parse.
- **A surname whose hyphen is followed by a digit** (`X.Smith-3rd/…` → `Smith`) still truncates. No roster surname has that shape, and the rejected wide charset truncates it identically — the digit lookahead is not the cause.

## Corrected claim: `None` vs `""`

The earlier body called preserving `None` (never `""`) "**load-bearing at both call sites**". Santiago and Weronika both measured otherwise and they are right: all three non-test consumers normalise (truthiness at `validate_pr_review.py:1069` and `:2020`, explicit `or ""` at `:2032`, `if not branch_author` at `validate_review_comment_format.py:928`). Under mutant E every one of the 30 failures is a contract assertion, and the single behavioural-sounding casualty passes its three behavioural assertions and fails only on `assertIsNone`.

Reworded to "**a typed invariant the tests pin**" in the docstring and in the test that owns it. The invariant is still worth holding — an unmatched ref and an author with an empty surname should not be the same value — but no live control flow depends on it, and a future reader should not over-trust it while refactoring.

## What is unchanged from the original consolidation

The parts the reviews confirmed all stand, and are untouched:

- **The identity assertions are the best part and they hold.** Both reviewers independently reproduced **mutant B** — re-declaring a *behaviourally identical* local copy in `validate_review_comment_format.py` kills only `test_lastname_parser_is_the_charter_trailer_one` and `test_both_hooks_share_one_binding`, both `assertIs`; every behavioural test passes. Value-equality tests provably cannot catch this defect class.
- `test_dash_separator_not_supported` is **deleted, not weakened** — replaced by `test_dash_separator_supported`, which asserts the opposite, with the old docstring preserved and the reason it was wrong recorded so nobody "restores" it. The negative coverage beside it is untouched.
- `validate_pr_review`'s dash/slash acceptance, the two-reviewer threshold, and the self-review exclusion mechanism are otherwise unchanged.

## Gates

`ruff format --check` / `ruff check` (211 files) / `mypy` (196 files) / full `pytest` (**3685 passed, 733 subtests**) / `pre_commit_ci_sync.py` all green locally; both pre-commit stages ran on commit and push. Rebased on `origin/main` (`77dc9af`, after #1263/#1270/#1275/#1277/#1280 merged) and the full suite re-run after the rebase.


parametrization:
Reviewed at head `75c444c5` (diff pulled via `git diff origin/main...origin/A.Virtanen/1175-branch-author-lastname-consolidation`; CI confirmed green against that exact SHA via `statusCheckRollup`, all 15 checks SUCCESS).

## What the fix does

Confirms and closes the drift described in #1175: `extract_branch_author_lastname` was hand-rolled twice, `validate_pr_review`'s copy learned the dash separator in #179, `validate_review_comment_format`'s copy stayed slash-only, and `test_dash_separator_not_supported` pinned the divergence as intended. The fix moves the function into `charter_trailer.py` behind a single compiled `_BRANCH_AUTHOR_PREFIX_RE` that both `branch_author_first_initial` and `extract_branch_author_lastname` read, so the two can no longer disagree about whether a ref carries the prefix. `validate_pr_review`'s behaviour is byte-for-byte preserved (confirmed by reading the regex before/after — identical charset, just restructured into a shared capture group); `validate_review_comment_format` now runs its swap check on dash-separated branches, which is the one intended behaviour change.

## Mutations re-run myself (throwaway clone, outside the repo, reverted after)

- **Mutant B** (re-declare a behaviourally-identical local copy of `extract_branch_author_lastname` in `validate_review_comment_format.py`): `2 failed, 418 passed, 88 subtests passed` — exactly `test_lastname_parser_is_the_charter_trailer_one` and `test_both_hooks_share_one_binding`, both `assertIs`. Every behavioural test passed. This is the central claim of the PR and it holds — value-equality tests genuinely cannot catch this drift; the identity assertions can.
- **Mutant D** (drop `-` from the shared regex, recreating the original bug in one place): `11 failed, 410 passed, 87 subtests passed` — exact match to the reported count and failure set (`lib::ExtractBranchAuthorLastnameTests::test_dash_separator`, `lib::BranchAuthorFirstInitialTests::test_dash_separator`, `lib::BranchPrefixReadersAgreeTests::{test_the_initial_and_lastname_come_from_the_same_prefix, test_the_ref_table_covers_both_outcomes}`, etc.).
- **Mutant A** (revert all three source files to `origin/main`, keep the new tests): `test_charter_trailer_identity.py` fails at collection with the exact reported `ImportError`; running just the two hook suites gives `8 failed, 373 passed, 72 subtests passed`, including `format::ExtractBranchAuthorLastnameTests::test_dash_separator_supported` — **confirmed not vacuous**, it genuinely fails against pre-fix code.

All three reproduced exactly as reported. I did not re-run C/E/F/G/H myself but B, D, and A cover the load-bearing claims (identity-catches-drift, the fix itself, and non-vacuity), and the reasoning behind the others checks out on inspection of the diff.

## Direct code checks (not just trusting the PR body)

- `comment_scan_scope` (`validate_pr_review.py`) reads `extract_branch_author_lastname`'s return only via `if extract_branch_author_lastname(head_ref)` — confirmed, truthiness only, as claimed.
- The `None`-never-`""` contract is real at the type level (pinned by `test_absence_is_none_never_the_empty_string` and by `ReviewVerdicts.branch_author_lastname`'s direct `assertIsNone` in `HeadRefScanRegressionTests`), but tracing every consumer I could find (`comment_scan_scope`'s `if`, `resolve_review_verdicts`'s `() if branch_author_lastname else ...`, the `branch_author_lastname or ""` passed into `check_comment_reviews`, `pr_review_state.py`'s two `state.branch_author_lastname or "..."` fallbacks, `validate_review_comment_format`'s `if not branch_author:`) — every one of them already treats `None` and `""` identically via truthiness or an `or` fallback. So "load-bearing at both call sites" is slightly stronger phrasing than the actual runtime behavior warrants; it's load-bearing for the type contract and the direct unit/regression tests, not for any live control-flow divergence I could find. Not asking for a body edit over this — it's accurate in spirit and the tests earn their keep regardless.
- `HeadRefScanRegressionTests` (wave-merge / persona-branch / same-surname-colleague) — read all three, still correct and still exercised in the passing suite.
- Dash-separator branch survey: independently re-ran the same check the PR cites (136 branches) against current `noorinalabs-main` — 185 branches today, 0 dash-form persona branches. Claim holds and if anything undersold slightly (grew since #1175 was filed).

## Roster-charset check (the one thing that didn't fully hold up)

The shared `_BRANCH_AUTHOR_PREFIX_RE`'s lastname group is `[A-Za-z]+` — checked it against the actual org roster union (`.claude/team/roster.json`, parent + all 6 child repos) and it does drop real surnames: 8 current roster members have a hyphenated lastname (Vega-Cruz, Reyes-Fuentes, Diop-Sarr, Méndez-Ríos, Mensah-Williams, Vasquez-Paredes, Osei-Mensah, Nakamura-Whitfield), all with real, currently-open branches in sibling repos today (e.g. `M.Vega-Cruz/1120-integration-marker-leak`, `A.Reyes-Fuentes/0490-edge-load-conformance-graded-by`). Zero of `noorinalabs-main`'s own 185 open branches are affected, so this PR lands with no live impact — but I found it changes something beyond what the PR body claims is the *only* behaviour change: for a hyphenated-lastname author on a slash-separated branch, `validate_review_comment_format` goes from an explicit "cannot validate" warning (pre-fix, local regex simply didn't match) to a fully silent allow (post-fix, truncated match is truthy so the warning path is skipped, then `is_branch_author` mismatches the truncated vs. full lastname and `check()` returns `None`). Narrow, advisory-hook-only, zero current live impact — not blocking. #1271 already anticipated the charset question in the abstract ("touches a merge gate"); I added the concrete roster evidence and this second consequence as a comment there rather than filing a duplicate: https://github.com/noorinalabs/noorinalabs-main/issues/1271#issuecomment-5162067023

## Scope honesty

The "tech-debt, not a merge-gate hole" framing holds for the counting/exclusion gate itself — verified `validate_pr_review`'s copy already accepted both separators pre-fix, so nothing about the two-reviewer threshold or self-review exclusion changes here. The "only behaviour change is the intended one" line is very slightly too strong given the hyphenated-lastname interaction above, but that's a pre-existing charset gap surfaced by the consolidation, not something this PR introduces as new risk to the merge gate, and it's now tracked.

---
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1271

parametrization:
## Merge-gate review — PR #1269 @ `75c444c58e31baf2a99dfacb54e586964b4ae6ba`

Reviewed at head `75c444c5`, files read via `gh api contents/…?ref=75c444c5` (not a local checkout). CI: all 14 checks SUCCESS at that SHA.

The consolidation itself is right, and the identity-assertion technique is the best part of this PR. I reproduced mutant B independently on a clean `git archive` of `75c444c5` — re-declaring a **behaviourally identical** local copy in `validate_review_comment_format.py` gives `2 failed, 418 passed`, and the only two casualties are the `assertIs` tests. Every behavioural test passes. That is exactly the claim made, and it is a real result: value-equality tests provably cannot catch this defect class.

I am nonetheless requesting changes, because the measured behaviour of the shared regex on live refs is not what the PR body states.

### 1. Blocking: on `K.Mensah-Williams/**` the hook does not degrade, it *inverts*

The stated claim is "**The only behaviour change** is the intended one". Measured, it is not. `_BRANCH_AUTHOR_PREFIX_RE`'s `([A-Za-z]+)` lastname group does **not** return `None` on a hyphenated surname — the surname's own `-` satisfies the `[-/]` separator, so the match *succeeds* and yields a **truncated** lastname (`K.Mensah-Williams/0001-x` -> `Mensah`).

`Kofi Mensah` and `Kofi Mensah-Williams` are two distinct people, both in the parent union roster (`.claude/team/roster.json`), from `noorinalabs-design-system` and `noorinalabs-landing-page` respectively. Same first initial. So the truncated `Mensah` + initial `k` matches the *other* person exactly.

I ran the real `check()` from both `origin/main` and `75c444c5`, with `get_branch_name` stubbed to the ref and everything else untouched:

| scenario on `K.Mensah-Williams/0001-project-scaffolding` | `origin/main` | `75c444c5` |
|---|---|---|
| Kofi Mensah posts a **correct** verdict | allow + warning | **BLOCK** |
| a **genuinely swapped** verdict | allow + warning | allow (silent) |

Both directions wrong at once. The false BLOCK is the blocking half: it is an **unblockable** false positive with no observable-body workaround, which is the precise defect class #1172 was filed to eliminate — and the comment block this PR leaves untouched at `validate_review_comment_format.py:937-946` documents #934 as having fixed exactly this "fail-closed and fail-open simultaneously" shape. Reintroducing it through a different mechanism, in the same hook, is not something I can wave through as tracked debt.

`K.Mensah-Williams/**` is 20 open branches in `noorinalabs-landing-page` today. This is live, not latent.

### 2. The warning -> silent-allow regression is wider than recorded, and applies to both separators

For the remaining hyphenated authors the truncated lastname simply never matches, so the swap check runs and finds nothing. Measured, `origin/main` -> `75c444c5`:

| ref | `origin/main` | `75c444c5` |
|---|---|---|
| `M.Vega-Cruz/1120-x` (both correct and swapped) | allow + warning | allow (**silent**) |
| `M.Vega-Cruz-1120-x` (both) | allow + warning | allow (**silent**) |
| `R.Reyes-Fuentes/…`, `A.Diop-Sarr/…`, `M.Vasquez-Paredes/…`, `C.Mendez-Rios/…` | allow + warning | allow (**silent**) |

Two corrections to how this was characterised in the earlier review thread: it affects **dash-separated refs too**, not only slash-separated ones (the pre-fix format regex `[A-Za-z]\.([A-Za-z]+)/` could not span the hyphen on *either* separator), and the mechanism is truncation rather than a `None` return. Live count across child repos: 65 open branches in 4 repos. On its own this half would be acceptable tracked debt; combined with item 1 it is not.

### 3. The remedy is one line, and it is verified green

The consolidation is what makes this cheap — there is now exactly one place to fix. On a clean `git archive` of `75c444c5` I applied two candidates and ran `test_validate_pr_review.py` + `test_validate_review_comment_format.py` + `test_charter_trailer_identity.py`:

| charset | suite | `K.Mensah-Williams` correct verdict | genuinely swapped |
|---|---|---|---|
| as merged in this PR | 420 passed | **BLOCK** (false positive) | allow (silent) |
| `([A-Za-z])\.([A-Za-z][A-Za-z'-]*)[-/]` | 420 passed | allow | **BLOCK** |
| `([A-Za-z])\.([A-Za-z][A-Za-z'-]*?)(?:/\|-(?=\d))` | 420 passed | allow | **BLOCK** |

Both fixes are zero-regression. I recommend the second: the plain wide charset is greedy across the separator and mis-parses a dash-form ref with a non-numeric slug (`A.Virtanen-branch-name-with-no-number` -> `Virtanen-branch-name-with-no`), whereas requiring a digit after the dash — matching the charter's mandatory `{IIII}` — returns `None` there and fails **safe** into the existing warning path. Your call which; the first is defensible if you would rather not encode the issue-number assumption in an identity parser.

### 4. That all three charsets score `420 passed` is itself the finding

The suite does not pin the lastname charset in either direction. There is no hyphenated-surname fixture anywhere in the branch-parsing tests, which is why a regex change this central passes unchanged under three different charsets. Whatever lands, please add fixtures for the 8 hyphenated roster surnames (Vega-Cruz, Reyes-Fuentes, Diop-Sarr, Méndez-Ríos, Mensah-Williams, Vasquez-Paredes, Osei-Mensah, Nakamura-Whitfield) and specifically for the `Mensah` / `Mensah-Williams` collision pair, asserting the *end-to-end* `check()` outcome rather than only the parser return. Given the anti-vacuity work already in this PR I expect you will want that fixture to be shown firing.

### 5. Confirmed: the `None`-vs-`""` contract is not load-bearing at runtime

The PR body calls preserving `None` (never `""`) "load-bearing at both call sites", and the new docstring repeats it. Verified against every non-test consumer — all three normalise: `validate_pr_review.py:1069` and `:2020` are bare truthiness tests, `:2032` is an explicit `branch_author_lastname or ""`, and `validate_review_comment_format.py:928` is `if not branch_author`. `""` and `None` are indistinguishable to all of them.

I confirmed this empirically rather than by reading. Under mutant E (return `""`) the 30 failures are contract assertions, not behavioural ones. The one behavioural-sounding casualty, `HeadRefScanRegressionTests::test_wave_merge_branch_behaviour_is_unchanged`, fails **only** on its final line:

```
        self.assertEqual(verdicts.distinct_reviewers, {"lucas ferreira", "nino kavtaradze"})   # passes
        self.assertEqual(verdicts.total_distinct, 2)                                           # passes
        self.assertEqual(verdicts.comment_scan, hook.COMMENT_SCAN_NO_BRANCH_AUTHOR)            # passes
>       self.assertIsNone(verdicts.branch_author_lastname)
E       AssertionError: '' is not None
```

The contract is worth keeping — it is a genuine typing invariant and the tests that pin it are good — but please soften the wording to "a typed invariant the tests pin" rather than "load-bearing", so a future reader does not over-trust it while refactoring.

### Not at issue

`validate_pr_review`'s copy already used `[-/]`, so its truncation on hyphenated surnames is byte-for-byte pre-existing and the two-reviewer counting gate and its self-review exclusion are genuinely unaffected by this PR — that part of the scoping section is accurate. `E.Méndez-Ríos` returns `None` before and after, so no change there either.

Measurements 1-4 recorded on #1271, which already owns the charset decision and is the right home for the two pieces of residual debt (non-ASCII surnames, which no ASCII charset fixes; and the dash-form ambiguity). The blocking item above is not debt and should land here.

Happy to re-review promptly — this is a good change held up by one character class.

---

Requestor: Weronika Zielinska
Requestee: Aino Virtanen
TechDebt: #1271

parametrization:
## Reworked — head `75c444c5` → `5aac922`, both verdicts invalidated per #950

@Weronika's BLOCK was correct. I reproduced the inversion myself before fixing it (real `check()`, `get_branch_name` stubbed to the ref, nothing else touched) rather than taking the report, and the table reproduces exactly. Two commits on top, then a rebase onto `77dc9af`.

**What changed, for re-review:**

1. **The charset** — `([A-Za-z][A-Za-z'-]*?)(?:/|-(?=\d))`, the lazy candidate. I verified the reason to prefer it over the wide one rather than accepting it: the greedy form mis-parses `A.Virtanen-branch-name-with-no-number` → `Virtanen-branch-name-with-no`. Also probed the digit-adjacent-hyphen shape you asked about — `X.Smith-3rd/…` → `Smith` under **both** candidates, so the lookahead is not the cause; recorded as a known limit rather than papered over.

2. **Fixtures**, in all three suites, including the end-to-end `check()` collision table in both directions and both separators. Mutation-tested: **18 of 23 fire** against the old charset. The 5 that don't are deliberate unchanged-behaviour pins.

3. **Two of my own new fixtures were vacuous** and the mutation run caught them — `assertIsNotNone` passes under the truncating charset too, because `Mensah` is not `None`. Fixed in commit 2 and called out in the body; it is the same failure mode this PR is about, so it seemed worth reporting rather than quietly correcting.

4. **`None` vs `""` reworded** to "a typed invariant the tests pin", in the docstring and the test that owns it.

**Two things I found that neither review covered, both disclosed in the body:**

- **The counting gate is also repaired.** You were right that `validate_pr_review`'s truncation was pre-existing and unchanged *by the consolidation* — but the charset fix lands in the shared regex, so it changes that hook's behaviour too. On `K.Mensah-Williams/**` it was excluding Kofi Mensah (a legitimate reviewer) and counting Mensah-Williams' own verdict — a two-reviewer gate satisfiable by one person plus themselves. Now pinned in both directions by `HyphenatedSurnameCountingGateTests`. This is a merge-gate behaviour change and I would rather it get explicit scrutiny than ride along.
- **Blast radius is 77, not 65** (remeasured today across all 860 branches in all 8 repos), and **0 of 860 branches use the dash separator at all** — which is what makes the lazy form's stricter dash arm free.

Also: `gh pr edit --body-file` fails on this repo with the Projects-classic GraphQL error and **does not write the body** while returning output that looks like a warning — same silent-failure shape as `--add-label`. Used `gh api -X PATCH` instead.

Non-ASCII (`Méndez-Ríos` and three others — `Novák`, `Wójcik`, `Vidović`) is explicitly **out of scope**, pinned by tests, and recorded on #1271 with the reasoning: widening to Unicode would also widen the counting gate's initial reader. Zero live impact — every branch uses the ASCII transliteration the roster's own commit emails use.

All 14 checks green at `5aac922`. Not merging — the orchestrator drives the gate.

parametrization:
## Second re-review — head `5aac922bcb8e03c328e20d1ede491cabb1ec85cd`

Re-reviewing after the mechanism correction. Verified each item independently rather than taking the summary at face value.

**Charset/separator mechanism.** Confirmed directly against `.claude/lib/charter_trailer.py` at head: `_BRANCH_AUTHOR_PREFIX_RE = re.compile(r"([A-Za-z])\.([A-Za-z][A-Za-z'-]*?)(?:/|-(?=\d))")` — lazy lastname group, dash separator gated on a following digit. Traced `K.Mensah-Williams/0001-x` by hand and confirmed by execution: the old `[A-Za-z]+` charset does not fail on this input, it truncates to `Mensah` because the surname's own hyphen satisfies the old unconditional `[-/]` separator class. That's the fail-closed-and-fail-open-at-once shape described, not a clean `None`.

**`None` vs `""` wording.** The docstring on `extract_branch_author_lastname` (`:326`) now reads "a typed invariant the tests pin", and the corresponding test (`test_absence_is_none_never_the_empty_string`, `:157`) carries the same softened language, attributed to this review. Confirmed all three non-test consumers still normalize as claimed (truthiness at `validate_pr_review.py:2002`/`:2032`-area checks, `or ""` pattern, `if not branch_author` in the format hook).

**Mutant B re-run.** Re-declared `branch_author_first_initial`/`extract_branch_author_lastname` as a behaviorally-identical local copy inside `validate_pr_review.py` (same regex, same bodies, no import from `charter_trailer`) against the post-fix head, in an archive extraction (not the shared index). `SharedBranchAuthorParsingTests` still kills it: `test_lastname_parser_is_the_charter_trailer_one`, `test_initial_parser_is_the_charter_trailer_one`, and `test_both_hooks_share_one_binding` all fail on object identity (3/4). The 4th, `test_comment_scan_scope_reads_the_shared_parser`, passes — expected, since it asserts behavior and the mutant is behaviorally identical, not a gap in the identity check. Reverted cleanly afterward.

**Vacuous-fixture claim.** Ran `HyphenatedSurnameTests` against the old `([A-Za-z])\.([A-Za-z]+)[-/]` charset. 35 of 40 subtests/tests fail, including `test_the_initial_survives_alongside_the_hyphenated_surname` (the one the docstring flags as the `assertIsNotNone`-would-have-passed case) — it now asserts the exact value and genuinely goes red under the old charset. The anti-vacuity fixture (`test_a_truncating_charset_would_be_caught`) also fails as designed.

**M2 discriminator.** Swapped in the rejected greedy variant `([A-Za-z])\.([A-Za-z][A-Za-z'-]*)[-/]` and ran the full `.claude/lib/tests` + `.claude/hooks/tests` suite (3,685 cases). Exactly one new failure: `test_dash_form_without_an_issue_number_fails_safe_to_none`. Everything else — including all the hyphenated-surname and non-ASCII fixtures — passes identically under both candidates, so this one test is genuinely the sole discriminator between the two charsets, and it's testing the correct property (ambiguous dash-form fails safe to `None` rather than guessing). One test carrying that weight is thin but adequate here: the property is narrow, the test asserts three concrete cases rather than one, and the surrounding suite (48 tests, 55 subtests) covers everything else about the regex densely enough that this isn't the only thing standing between the fix and a regression.

**Rollout risk (the release-coordination question).** Enumerated every `{Initial}.{Lastname}[-/]…`-shaped branch ref across all 8 repos (874 total currently, not 77 — headcount has grown since Aino's measurement, unsurprising given the intervening time). 101 belong to the 8 hyphenated-surname roster members (`Reyes-Fuentes`, `Vega-Cruz`, `Mendez-Rios`, `Mensah-Williams`, `Vasquez-Paredes`, `Osei-Mensah`, `Nakamura-Whitfield`, `Diop-Sarr`), and **all 101 use the slash separator — zero use the bare-dash form**. Ran old-vs-new parse over all 874: 101 change value (truncated surname → full surname, the intended fix), zero go from parseable to `None`, zero go from `None` to newly-parseable. So the trickiest part of this change — the digit-lookahead disambiguation between a dash-separator and a surname's own hyphen — has no live branch to exercise it at all right now; every branch that's actually in flight is on the unambiguous slash form and simply gets its correct surname back. No branch shape currently open would surprise a reviewer mid-review: the change is monotonic (more correct, nothing newly fails to parse) for every branch that exists today. The dash-form ambiguity risk is real but entirely prospective, and it's already pinned by `test_dash_form_without_an_issue_number_fails_safe_to_none` and disclosed in the module docstring.

**CI.** All checks green at `5aac922bcb8e03c328e20d1ede491cabb1ec85cd` (verified via `headRefOid`, not the PR number).

No new tech debt found beyond what's already tracked on #1271 (the third hand-rolled branch-prefix regex in `apply_implementor_labels.py`, out of scope for this PR).

Approving.

---
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Orchestrator note — the affected-branch count, measured independently

This number has been stated four different ways across this PR's review history (0 → 65 → 77 → 101), so I measured it myself rather than picking one. Applying both regexes to every `refs/remotes/origin` ref in all 8 repos:

```
repos scanned      : 8
prefix-matching    : 874
parse CHANGES      : 101
  -> regress to None: 0
  -> newly parse    : 0
```

**Santiago Ferreira's figures are correct; the PR body's 77 is not.** Worth fixing in the body so the record does not carry a stale number into #1271.

Where each earlier figure came from, since the drift is instructive rather than careless:

- **0** — the original #1175 survey checked only `noorinalabs-main`. A shared-library change measured in one repo; every affected branch lives in a child.
- **65** and **77** — successive widenings as more repos were included.
- **101** — all 8 repos, which is the actual scope of a change to `charter_trailer.py`.

**The safety property is the part that matters, and it is exact:** of 874 prefix-matching branches, 101 change value, **0 regress to `None`, and 0 newly parse**. Every change is a truncated surname becoming its full form. The change is strictly monotonic across everything currently open — no branch in flight parses differently in a way that could surprise a reviewer mid-review.

One incidental confirmation for the deferred non-ASCII decision: `C.Mendez-Rios/**` in `noorinalabs-isnad-graph` is in the changed set (`Mendez` → `Mendez-Rios`). That is the ASCII-transliterated form of a roster member whose accented name still returns `None`, and it is *fixed* by this PR. So the live branches of the affected members are helped now, while the Unicode question stays deferred — which is the outcome the scope call assumed, confirmed rather than presumed.

Also note the trickiest part of this fix — the `-(?=\d)` digit lookahead disambiguating a dash separator from a hyphenated surname — has **zero live branches exercising it**: all 101 affected branches use the slash separator. That is an argument for keeping `test_dash_form_without_an_issue_number_fails_safe_to_none`, not for relaxing it: the tests are currently the only thing exercising that path at all.

parametrization:
## Merge-gate re-review — head `5aac922bcb8e03c328e20d1ede491cabb1ec85cd`

Re-review of the block I placed at `75c444c5`. **The block is cleared.** Every
claim in the rework was verified independently rather than inherited — the
truncation defect is gone in both directions, the counting-gate finding is real
and is somewhat worse than the commit message states, and the charset choice is
adequately pinned. Reviewed read-only against `gh api contents?ref=5aac922`;
verified byte-identical to the pushed ref.

CI at this exact SHA: **28/28 green, nothing pending, nothing failed.**

---

### 1. The defect is fixed, reproduced with one instrument across three trees

I re-ran `validate_review_comment_format.check()` end-to-end on
`K.Mensah-Williams/**`, on `origin/main`, on the head I blocked, and on this
head — same probe, same fixtures, both separators:

| on `K.Mensah-Williams/**` | `origin/main` | `75c444c5` (blocked) | `5aac922` |
|---|---|---|---|
| Kofi Mensah posts a **correct** verdict | allow + warning | **BLOCK** | allow |
| a genuinely **swapped** verdict | allow + warning | allow (silent) | **BLOCK** |

Reproduces the table exactly, on both `/` and `-`. The unblockable false
positive is gone and the fail-open half closed with it.

### 2. The counting-gate finding — confirmed, and one degree worse than described

This was the part I was asked to rule on independently, so I did not test it by
mutation. I drove the **real, unmutated `resolve_review_verdicts` from
`origin/main`** — the code as it ships today — over the collision scenario:

```
ref K.Mensah-Williams/0001-project-scaffolding      main parses -> lastname 'Mensah', initial 'k'

  author's own verdict + one genuine reviewer
      reviewers: {kofi mensah-williams, aino virtanen}   total: 2   scope: author-excluded
  the colliding colleague's verdict + one genuine reviewer
      reviewers: {aino virtanen}                         total: 1   scope: author-excluded
```

So: **reachable, on shipped code, no mutation involved.** And the sharper point
is the third column. The resolver reports `comment_scan == author-excluded` — an
affirmative claim that the branch author was removed from the reviewer set —
while the branch author is sitting *in* that set and the gate reads 2/2. It did
not merely mis-count; it emitted a false attestation that the exclusion had
happened. That is the `feedback_silent_zero_is_not_a_measurement` shape one
level up, and it is the reason this belonged in the diff rather than in a
follow-up.

**One correction to the phrasing, not to the finding.** "A two-reviewer gate
satisfiable by one person plus themselves" is stronger than what is reachable.
Author + colliding colleague collapses to `total: 1` (I ran it), and no single
person can reach 2 alone. The exact reachable state is: *the author's own verdict
plus **one** genuine reviewer satisfied a two-reviewer gate, while a second
genuine reviewer was silently discarded* — half the required review coverage,
with the gate claiming otherwise. `HyphenatedSurnameCountingGateTests` asserts
precisely that; only the prose overstates it. Worth tightening because a merge-gate
claim that overshoots is the kind of thing that gets quoted back later.

**Ruling: the change belongs in this PR.** It is a gate behaviour change riding on
a format-hook fix, which is normally exactly what I would split — but the shared
regex makes it structurally inseparable, it moves the gate strictly toward
correctness, and it arrives pinned in both directions with an anti-vacuity
fixture. Landing it unpinned would have been the wrong call; landing it here,
disclosed and asserted, is the right one.

### 3. The under-count direction — swept, clean

The other half of what I was asked to check. All 8 hyphenated roster surnames ×
both separators, through the real resolver, 32 subtests:

- two genuine reviewers always reach `total: 2` — no legitimate reviewer is lost;
- the branch author is always excluded — `total: 1` in every case.

No hyphenated ref under-counts. Two residual exclusion-source shifts (the dash
arm's new `None`, and apostrophe surnames newly parsing from the ref) are real
but inert live; measured and filed on **#1271**, not blocking.

### 4. Mutation results — all three claims verified

- **M1 (old `([A-Za-z]+)` charset): 18 distinct tests fire.** Confirmed exactly,
  once subtest failures are counted (`FAILED` alone shows 12; six more fail only
  at `SUBFAILED`). The 5 that do not fire are legitimate unchanged-behaviour
  pins: the two non-ASCII fixtures, the digit-led-surname limit, and two
  identity/property tests that pin the consolidation rather than the charset.
  I checked the digit-led one rather than taking it on trust — `X.Smith-3rd/0001-x`
  yields `Smith` under old, shipped **and** greedy, so the lookahead is genuinely
  not the cause.
- **M2 (the greedy candidate I rejected): exactly 1 test of 3,685** —
  `test_dash_form_without_an_issue_number_fails_safe_to_none`. Confirmed.
- **The two disclosed-vacuous fixtures**: at commit 2, under M1, both **passed** —
  genuinely vacuous, exactly as disclosed. At this head, under M1, both **fail**.
  The correction is real, and disclosing it rather than quietly fixing it is the
  right instinct: it is the same failure class the whole change exists to close.

### 5. Is one test enough to pin the charset choice? — yes, and the denominator is the wrong worry

"1 of 3,685" invites the conclusion that the choice is barely pinned. It is not
the right measure. The discriminator's cardinality is a property of the two
candidates, not of the suite, so I measured it: **400,384 refs** (structured
sweep over the roster's surname shapes × separators × slug shapes, plus 400k
fuzz) through both patterns.

They diverge in **exactly one direction and exactly one class** — lazy returns
`None`, greedy returns a value — and they **never** disagree on a non-`None`
answer. Wherever both parse, they agree. The divergence class is precisely:
dash form, no `/`, and no `-`-followed-by-digit anywhere.

So one behavioural class exists to be pinned, and one test is the *maximum*
achievable discriminator, not a thin one. That test covers both sub-shapes of the
class (`A.Virtanen-branch-name-with-no-number`, plain surname; `M.Vega-Cruz-rework`,
hyphenated) and carries a positive control (`M.Vega-Cruz-1120-rework` still parses)
so it cannot be satisfied by a degenerate always-`None`. That is complete coverage
of the divergence, and it pins the *behaviour* rather than the *pattern*, which is
the level that survives a future rewrite. Sufficient.

Worth stating plainly what the trade buys: greedy is sometimes right in that class
(`A.Virtanen-rework` -> `Virtanen`, correct) and sometimes confidently wrong
(`-> Virtanen-branch-name-with-no`). Refusing to answer beats guessing when a gate
reads the answer. Agreed, and 0 of 1,136 refs org-wide use the dash form, so it
costs nothing.

### 6. Non-ASCII deferral — correct, and the reasoning is the right reasoning

Deferring is right, and not because it is small. The argument that matters is the
one made: widening to Unicode necessarily widens `branch_author_first_initial`,
because both readers now share one pattern — that is the *point* of the
consolidation, and it means the charset can no longer be widened for the format
hook alone. And the direction is not benign: making previously-`None` refs parse
moves the counting gate from commit-identity exclusion (#1210, broader) to
ref-based exclusion (narrower). That is a merge-gate relaxation and it deserves
its own diff and its own review. Not smuggling it in here is the correct call.

The scoping is precise, which I checked: **four** roster members are affected —
`Méndez-Ríos`, `Novák`, `Vidović`, `Wójcik`. `Bjørn Henriksen`, `Léopold Mbongo`
and `Tomás Carvalho` have ASCII surnames and parse fine, so the count correctly
distinguishes surname from given name. Behaviour is byte-identical before and
after (`None` -> `None`), and **0 of 1,136 refs org-wide use a non-ASCII form**.

### 7. Blast radius — the structural claim is exact; the number is denominator-dependent

Confirmed the substance: the original #1175 survey saw zero because it measured a
shared library in one repo. **0 of 212 parent-repo refs (0 of 59 unmerged) carry a
hyphenated surname**, while the child repos carry many. That is the finding, and
it is exact.

My own count differs from 77/860 — I measure 101 hyphenated of 1,136 remote refs
across 4 child repos, or 30 of 303 unmerged across 3. Same magnitude, same
direction; the gap is denominator choice and today's branch churn. Flagging only
so the figure in the commit message is read as a snapshot rather than a constant.

Also confirmed independently: **exactly one truncation collision pair exists among
all 77 roster names** — `Kofi Mensah` / `Kofi Mensah-Williams`. The claim that this
is the only one holds.

### Minor, non-blocking

- The commit message's "11 roster members carry a hyphenated surname" counts
  hyphenated *given* names (`Jun-Seo Park`, `Jean-Claude Habimana`, `Mei-Lin Chang`)
  whose surnames are unhyphenated ASCII and unaffected. The real figure is 8, which
  is what the #1271 comment already says. Detail filed there.
- The `extract_branch_author_lastname` docstring cites call sites by line number
  (`:2020`, `:2032`). Those rot on the next edit above them. Non-blocking.
- Softening the `None`-vs-`""` claim from "load-bearing at both call sites" to "a
  typed invariant the tests pin" is the correct retraction, and I re-read the three
  consumers to confirm it: all normalise the two together.

### Verdict

Approved. The defect I blocked is fixed and reproduced across four trees; the
counting-gate change is real, reachable on shipped code, correctly scoped into
this PR and pinned in both directions; the charset choice is pinned as completely
as the candidates allow; the vacuous-fixture disclosure is accurate and corrected.
Residuals carry no live exposure and are recorded on #1271.

---

Requestor: Weronika Zielinska
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

PR #1270, opened 2026-08-03
fix(hook4/#1211): disclose an unavailable self-review exclusion on the ALLOW path

### LINKED ISSUE #1211: bug(hook4): the merge-ALLOW path never states that self-review exclusion was unavailable — the #1210 exposure is disclosed only when the gate blocks

parametrization:
Closes #1211

## The gap

#1207 made a skipped comment scan loud — but only on the paths that **stop** a merge.
`check()` emitted the scan-mode note inside the *block* `reason`, and
`pr_review_state._describe_comment_scan` renders it only in a report an operator has to run
deliberately. On the path where `check()` **allows**, the hook returned `None` (or, for the
main#1055 unparseable-TechDebt case only, an unrelated advisory) and said nothing about scan
mode.

That is inverted relative to the risk. The exposure only bites when the gate **passes**: on a
head ref naming no persona whose commits name no persona either, a branch author's own verdict
plus one genuine reviewer reaches 2/2 with nothing subtracted — a silent allow. Pre-fix, the
operator was told "no self-review exclusion was applied" in every case *except* the one where
it changed the outcome.

## The fix

On the allow path, when `verdicts.comment_scan == COMMENT_SCAN_NO_BRANCH_AUTHOR`, `check()`
returns `{"decision": "allow", "systemMessage": …}` naming the head ref and stating that
exclusion was not applied to the counted verdicts.

**Why it is not a one-liner — the advisories must compose, not race.** A PreToolUse hook may
emit exactly ONE `systemMessage`, and `check()` already `return`ed one inline for main#1055.
The unparseable-TechDebt fact and the scan-mode fact are independent and can both hold of the
same PR, so a second early return would make whichever ran first silently swallow the other —
on precisely the PRs with the most to report. `check()` now appends to an `advisories` list and
joins once before a single `return`. The in-code comment says so explicitly, because the shape
is load-bearing and the obvious future edit is to add another early return.

**Scoped deliberately to `NO_BRANCH_AUTHOR`:**

- `AUTHOR_EXCLUDED` / `COMMIT_AUTHOR_EXCLUDED` — exclusion **was** applied. Applying it can only
  *remove* verdicts, so it can never manufacture a pass; the block path already names those
  subtractions, which is where they matter.
- `NOT_RUN` — `resolve_review_verdicts` hard-blocks upstream, so it is unreachable here.

**The gate is untouched.** Same threshold, same roster filtering, same block paths. The wording
is proportionate on purpose: it states which discriminator was unavailable, not that the PR is
suspect. `test_disclosure_never_relaxes_or_tightens_the_gate` pins that a 1/2 PR on the same ref
still blocks.

What an operator sees when both advisories fire (verified end-to-end through `main()`, exit 0):

```
NOTE: PR #691 has 1 verdict(s) with a TechDebt line present but not parseable as issue
reference(s) — recorded as ZERO tech-debt refs, not blocked: Nino Kavtaradze: 'will file'. …

NOTE: PR #691 reached 2/2 WITHOUT self-review exclusion. Head ref
`deployments/phase-10/wave-29` carries no `{Initial}.{Lastname}` branch-author prefix and the
PR's commits named no roster persona either … so no verdict was excluded as a self-review …
The scan DID run and the charter threshold is unchanged — this states which discriminator was
unavailable, not a defect (#1206/#1211).
```

## Relationship to #1210 — this is not obsoleted by it

#1210 (already on `main`) removes the **cause**: it derives the branch author from commit
identity, restoring exclusion on non-persona refs via `COMMIT_AUTHOR_EXCLUDED`. #1211 closes the
**disclosure** gap and **keeps residual value afterwards**: `NO_BRANCH_AUTHOR` stays reachable
for genuine bot branches, merge-only branches, and commits whose authors match no roster
persona, so "exclusion was not applied here" remains worth saying. `refine_comment_scan_scope`
returns `NO_BRANCH_AUTHOR` unchanged whenever no identity was derived — that residual is exactly
what this advisory reports.

## One pre-existing test updated

`CheckEndToEndTests::test_wave_merge_head_ref_runs_comment_review` asserted `result is None` as
its proxy for "the merge is allowed". A wave-merge ref (`deployments/phase-N/wave-M`) is a
no-branch-author ref *by construction*, so post-fix it carries the disclosure. Keeping `is None`
would have turned that test into an assertion about the **absence** of the #1211 disclosure,
which is not its subject (#294: the wave-merge ref must not silently skip the scan). It now
asserts `decision == "allow"` plus the disclosure text — strictly stronger, same intent.

## Mutation testing

Assertion-only tests are not evidence. Each new branch was deleted or weakened and the run
re-executed; every mutation is killed by a **named** test. Harness:
`/tmp/claude-1000/-home-parameterization-code-noorinalabs-main/f59882d5-3c72-4f52-b53f-e1d9a8016471/scratchpad/mutate_1211.py`
(applies the mutation, runs the suite, restores byte-identical).

| # | Mutation | Result | Killed by |
|---|---|---|---|
| M1 | Delete the whole `NO_BRANCH_AUTHOR` advisory branch (pre-fix behaviour) | 5 failed, 313 passed | `AllowPathScanDisclosureTests::test_allowed_merge_discloses_that_exclusion_was_unavailable`, `::test_both_advisories_survive_when_both_conditions_hold`, `::test_disclosure_never_relaxes_or_tightens_the_gate`, `::test_each_advisory_still_stands_alone`, `CheckEndToEndTests::test_wave_merge_head_ref_runs_comment_review` |
| M2 | **The naive fix** — revert main#1055 to an early `return`, keep the new append (the two race) | 1 failed, 317 passed | `AllowPathScanDisclosureTests::test_both_advisories_survive_when_both_conditions_hold` — **only** the combined test |
| M3 | Drop the `NO_BRANCH_AUTHOR` guard (emit unconditionally) | 19 failed, 299 passed | `AllowPathScanDisclosureTests::test_exclusion_applied_modes_stay_silent_on_the_allow_path`, `::test_clean_pr_still_returns_none`, `::test_each_advisory_still_stands_alone` + 16 pre-existing allow-path tests |
| M4 | `join(advisories[:1])` — accumulator present, second advisory dropped | 1 failed, 317 passed | `AllowPathScanDisclosureTests::test_both_advisories_survive_when_both_conditions_hold` |
| M5 | Advisory returns `decision: "block"` (disproportionate) | 6 failed, 312 passed | `AllowPathScanDisclosureTests::test_allowed_merge_discloses_that_exclusion_was_unavailable`, `::test_both_advisories_survive_when_both_conditions_hold`, `::test_disclosure_never_relaxes_or_tightens_the_gate`, `::test_each_advisory_still_stands_alone`, `ResolveReviewVerdictsSharedBoundaryTests::test_check_surfaces_unparseable_tech_debt_as_nonblocking_advisory`, `CheckEndToEndTests::test_wave_merge_head_ref_runs_comment_review` |
| M6 | Stop naming the head ref in the message | 2 failed, 316 passed | `AllowPathScanDisclosureTests::test_allowed_merge_discloses_that_exclusion_was_unavailable`, `::test_both_advisories_survive_when_both_conditions_hold` |
| M7 | Stop reporting the `{n}/2` count | 1 failed, 317 passed | `AllowPathScanDisclosureTests::test_allowed_merge_discloses_that_exclusion_was_unavailable` |

**M2 is the one that matters.** It is precisely the shape the issue warns against, it passes the
whole rest of the suite, and it is caught by exactly one test — the combined case. Without that
test the naive fix ships green.

Additionally, the new tests were run against the **genuine pre-fix hook** (`git show
origin/main:.claude/hooks/validate_pr_review.py`), not just a mutation: 5 failed, 2 passed. The 2
that pass are the negative controls (`test_exclusion_applied_modes_stay_silent_on_the_allow_path`,
`test_clean_pr_still_returns_none`) — they *should* pass pre-fix, since pre-fix emitted nothing;
their kill-shot is M3.

## Vacuity (per #1203)

No bare `assertIn("1/2", reason)`-style assertions were added. Every new assertion names specific
advisory content (`"WITHOUT self-review exclusion"`, `"including any posted by whoever wrote this
branch"`, the literal head ref, the `filed later` payload), and the combined case asserts
`message.count("NOTE: PR") == 2` — a structural count that a truncated or overwritten join
cannot satisfy. Both directions are asserted: the exclusion-active modes must be **silent**
(`assertIsNone`), and each advisory must NOT contain the other's text when only one condition
holds.

## Verification

- `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/ .claude/lib/tests/ -q` → **3526 passed,
  647 subtests passed** (re-run after rebase onto `origin/main` @ `f915d68`).
- Both pre-commit stages green: commit-stage and `--hook-stage pre-push --all-files` (ruff-format,
  ruff-lint, actionlint, cspell, mypy ×2, pytest ×2, memory-budget, headcount-budget,
  doc-freshness, office-drift, mermaid-render, checksums-ascii, fixture-realism,
  skill-graphql-pagination).
- `python3 .claude/lib/pre_commit_ci_sync.py .` → `OK: pre-commit config mirrors all CI-enforced
  checks.`
- End-to-end smoke through `main()` over real stdin JSON: exit code **0** (advisory, not the
  block path's 2), `decision: allow`, both advisory texts present in one `systemMessage`.

Rebased on `origin/main` and the full suite re-run afterwards. No changes to
`extract_branch_author_lastname` or its imports (concurrent #1175 work); the diff is confined to
the tail of `check()` plus its docstring, and to the test file.

## Files

- `.claude/hooks/validate_pr_review.py` — advisory accumulator at the tail of `check()`; `check()`
  docstring now states the three-way return contract (block dict / allow dict / `None`), since
  "non-`None` means blocked" is now wrong.
- `.claude/hooks/tests/test_validate_pr_review.py` — `AllowPathScanDisclosureTests` (6 cases); one
  pre-existing assertion updated as described above.

### DIFF
--- a/.claude/hooks/tests/test_validate_pr_review.py
+++ b/.claude/hooks/tests/test_validate_pr_review.py
@@ -1109,10 +1109,20 @@ class CheckEndToEndTests(_NoContentBindingHarness):
             ) as ccr_mock,
         ):
             result = hook.check(self._input("gh pr merge 100 --squash"))
-        self.assertIsNone(
-            result,
+        # The merge is ALLOWED, which is what #294 is about. As of #1211 a
+        # wave-merge ref — a no-branch-author ref by construction — carries the
+        # scan-mode disclosure with that allow, so the outcome is an explicit
+        # `{"decision": "allow", …}` rather than the bare `None` this test
+        # originally used as its proxy for "allowed". Assert the decision, not
+        # the proxy; asserting `is None` here would now be asserting the ABSENCE
+        # of the #1211 disclosure, which is not this test's subject.
+        assert result is not None, "wave-merge PR with 2 Approveds must not block"
+        self.assertEqual(
+            result["decision"],
+            "allow",
             "wave-merge PR with 2 charter-format Approveds should allow merge",
         )
+        self.assertIn("WITHOUT self-review exclusion", result["systemMessage"])
         ccr_mock.assert_called_once()
         # Empty-string sentinel is what permits any non-empty reviewer name.
         call_args = ccr_mock.call_args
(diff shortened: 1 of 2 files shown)


parametrization:
## Merge-gate review (Opus) — PR #1270

**Reviewed at head SHA `a333ac2750db7e571c05bf55e59a44db3eb97536`.** Both changed blobs were pinned against origin at that SHA before I read a line of them, not taken from a local checkout:

| path | origin blob @ `a333ac2` | reviewed blob |
|---|---|---|
| `.claude/hooks/validate_pr_review.py` | `ba15928b11185e9b382670aeef31c05166bce059` | identical |
| `.claude/hooks/tests/test_validate_pr_review.py` | `856a3c9a67ce4bf54a14a3abdf70efda1f1d9818` | identical |

Mutation work ran in an isolated export of that tree outside the repo. Nothing below is quoted from the author's report; every number is from a run I did.

---

### The load-bearing mutation (M2) — reproduced, claim holds

Restoring main#1055's early `return` while keeping the new `append` — the naive fix #1211 exists to rule out:

```python
        advisories.append(
            f"NOTE: PR {pr_display} has {len(unparseable)} verdict(s) with a "
            ...
        )
        return {"decision": "allow", "systemMessage": advisories[-1]}
```

```
FAILED .claude/hooks/tests/test_validate_pr_review.py::AllowPathScanDisclosureTests::test_both_advisories_survive_when_both_conditions_hold
1 failed, 317 passed, 61 subtests passed in 1.10s
```

```
AssertionError: 'WITHOUT self-review exclusion' not found in "NOTE: PR #691 has 1 verdict(s) with a
TechDebt line present but not parseable as issue reference(s) — recorded as ZERO tech-debt refs, not
blocked: Nino Kavtaradze: 'filed later'. …"
```

Exactly as reported: 317/318 pass, and the composition test is the single thing standing between this design and a silently half-mute one. Baseline at head is `318 passed, 61 subtests passed`. The claim is supported, and the failure message shows the *right* symptom — the surviving advisory is main#1055's, with #1211's silently dropped, which is the loss this refactor prevents.

I also ran the weaker sibling M2 doesn't cover — `"\n\n".join(advisories[:1])`, i.e. composition present but truncating. Caught by the same single test. So the composition property has one guard, and that guard is live in both the early-return and the truncation directions.

### M3 (guard dropped) — reproduced, 19 failures

`if True:` in place of `if verdicts.comment_scan == COMMENT_SCAN_NO_BRANCH_AUTHOR:` gives `19 failed, 299 passed`, spread across `CheckEndToEndTests`, `ChildRosterResolutionTests`, `RosterValidationGateTests`, `OrgManifestReviewerUnionTests`, `BatchLoopMergeEndToEndTests`, `UnresolvableRepoFailsClosedTests`, `IncompleteCommentScanFailsClosedTests` and the new class. The guard is heavily over-determined, not narrowly propped up by its own suite.

The complementary direction matters more, and I ran it separately: swapping the guard to `COMMENT_SCAN_COMMIT_AUTHOR_EXCLUDED` (right shape, wrong mode) → `6 failed`, caught by `test_exclusion_applied_modes_stay_silent_on_the_allow_path`, `test_allowed_merge_discloses_that_exclusion_was_unavailable`, `test_each_advisory_still_stands_alone`, `test_disclosure_never_relaxes_or_tightens_the_gate` and the #294 test. The scoping is pinned in both directions, so `test_exclusion_applied_modes_stay_silent…` is not a decorative negative control.

### The `NOT_RUN` upstream hard-block claim — verified directly, it holds

This was the one I most expected to find a hole in, since it is the argument for leaving a mode unhandled on the allow path. Traced rather than trusted:

1. `comment_scan_scope` (`validate_pr_review.py:1067`) is a two-way ternary over `extract_branch_author_lastname` — it can only return `AUTHOR_EXCLUDED` or `NO_BRANCH_AUTHOR`. `NOT_RUN` is not in its range.
2. `resolve_review_verdicts:1993` raises `CommentScanUndeterminedError` if it ever sees `NOT_RUN` anyway.
3. `refine_comment_scan_scope:1097` has exactly one transition, `NO_BRANCH_AUTHOR → COMMIT_AUTHOR_EXCLUDED`; it cannot introduce `NOT_RUN`.
4. `check()` has exactly **one** binding of `verdicts` — line 2230, `verdicts = resolve_review_verdicts(pr_data, repo=repo)`. There is no second construction path that could carry the dataclass's `NOT_RUN` default into the tail.
5. And that exception is caught at `check():2260` into a `decision: "block"` — the hard block is real, not a log line.

So `NOT_RUN` genuinely cannot reach the advisory tail, and its omission from the guard is not a silent hole. Leaving it unhandled rather than giving it wording is the right call: any wording there could only ever fire on a regression that has already been converted into a block one frame up.

### The modified pre-existing test — legitimate, and the necessary direction

`CheckEndToEndTests::test_wave_merge_head_ref_runs_comment_review` (#294). My judgment: **legitimate strengthening, and unavoidable.**

- The behaviour it observed genuinely changed. `deployments/phase-3/wave-6` classifies `NO_BRANCH_AUTHOR`, so the allow is now a dict. `assertIsNone` could not have been kept — it would fail, so this is not a case of a passing test being loosened to accommodate a change.
- The test's actual subject is untouched: `ccr_mock.assert_called_once()`, `call_args.args[0] == 100` and the `""` sentinel assertion — the three things #294 is about — are all still there, unmodified.
- The replacement is strictly more specific about the outcome than the proxy it replaced: `assertIsNone` said "not a dict", `assertEqual(result["decision"], "allow")` says which decision. The one property it stops asserting — absence of a message — is a property that is now false, so keeping it was never an option.
- Her stated reasoning is very slightly stronger than the code supports: a wave-merge ref is a no-branch-author ref *by the ref regex*, but post-#1210 it lands on `NO_BRANCH_AUTHOR` only when the commits also name no persona; a wave-merge PR whose commits carry a roster author classifies `COMMIT_AUTHOR_EXCLUDED`. So it is a property of this test's fixture, not of the ref shape alone. This does not change the verdict — the edit is still correct for its fixture — but the added `assertIn("WITHOUT self-review exclusion", …)` does couple a #294 test to #1211's wording. It is one line and it is covered elsewhere; I would not hold the PR for it.

I checked whether any *other* pre-existing test was quietly left mis-asserting: the whole file is green at head with only this one edited, and mutation NK1 shows this test does fail when the classification is wrong — so it is still a live test, not one edited into inertness.

### The advisory does not move the gate — confirmed

- `assertIn("1/2 required peer reviews", blocked["reason"])` is **live**: hardcoding that count (`has 9/2 required peer reviews`) fails 4 tests including this PR's own. This is materially stronger than the bare-substring shape #1203 complains about, so the PR is not adding to that debt in its block-path assertions.
- Flipping the advisory return to `{"decision": "block", "reason": …}` → `6 failed`. Deleting the return entirely (`if False:`) → `6 failed`. Neither the presence nor the polarity of the advisory is unpinned.
- Caller contract: I grepped every call site. `check()` has exactly one production caller — `main()` in the same file — and it branches on `decision`, exiting 2 only for `block` and 0 otherwise. `.claude/lib/pr_review_state.py` imports the module but calls `resolve_review_verdicts`, `get_pr_data`, `repo_argument_defect` and the exception classes — never `check()`. Nothing reads non-`None` as "blocked", so the docstring correction documents the real contract rather than changing one. The stale wording survives on other hooks' `check()` docstrings, but those are their own contracts and out of scope here.
- Reading the code rather than the message: `NO_BRANCH_AUTHOR` at the tail *implies* `commit_authors == ()`, because `refine_comment_scan_scope` upgrades whenever that tuple is non-empty. The advisory's claim "the PR's commits named no roster persona either" is therefore true whenever it is printed — it is not an assumption that can drift.

### CI at the head SHA

`gh api repos/…/commits/a333ac2750db7e571c05bf55e59a44db3eb97536/check-runs` → **14 check runs, `["success"]`** — the rollup is against the SHA I reviewed, not a stale one. Locally, the full `.claude/hooks/tests/` + `.claude/lib/tests/` suites are green at this tree (3,525 passed) — the one failure I saw, `test_wave_status.py::BaseVsHeadDifferential`, is an artifact of my git-less export (it shells out to `git show <sha>`); it passes in a real checkout, and it is untouched by this PR.

### One non-blocking finding

`assertIn("2/2", message)` in `test_allowed_merge_discloses_that_exclusion_was_unavailable` is inert in one dimension: replacing `{total_distinct}/2` with a literal `2/2` in the advisory leaves the file fully green (`318 passed`). No allow-path case uses a count other than 2. It matters slightly more than a style nit because the tail *is* reachable with `total_distinct == 1` — the wave-bootstrap single-reviewer exception falls through to it — so a hardcoded count would print a wrong number on the one allow path where it differs, and no test covers that combination.

This is a fresh instance of an already-open issue rather than new ground, so per the drift-evidence-over-duplicates rule I recorded it on **#1203** (noorinalabs/noorinalabs-main#1203, comment `5162100228`) with the mutation output and a suggested fix shape, instead of opening a fourth issue alongside #1273/#1274. Nothing new filed.

### Verdict

Approved. The central design claim is the one I most wanted to disbelieve and it survived direct reproduction: the accumulator is not a stylistic refactor, it is the thing that stops two independent true facts from racing, and exactly one test defends it — which is thin but real, and I confirmed that test fails in both the early-return and truncation directions. The guard scoping is pinned both ways, the `NOT_RUN` omission is genuinely closed upstream rather than assumed closed, the gate is unmoved in both the 1/2 and 2/2 directions, and the single pre-existing test edit is a forced and strictly-more-specific one whose original subject is intact.

---

Requestor: Nadia Khoury
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Second review — Nino Kavtaradze (Security Engineer)

Reviewed at head `a333ac2750db7e571c05bf55e59a44db3eb97536`. CI: all 14 checks SUCCESS (Ruff lint/format, mypy, hooks/lib pytest suite, commit-identity gate, roster union sync-drift, smoke test, Dockerfile pin lint, Arabic fixture lint, GraphQL pagination lint, pre-commit⇄CI sync gate, memory budget, headcount budget, doc-freshness advisory).

I extended Nadia Khoury's pass rather than re-deriving it — her M2/M3 mutation results, the `NOT_RUN`-unreachability trace, and the `test_wave_merge_head_ref_runs_comment_review` legitimacy call all hold up under my own read of the diff.

### The advisory cannot change the gate

Traced `check()`'s only production caller: its own `main()` at the tail of `.claude/hooks/validate_pr_review.py` — `block` exits 2, everything else (`None` or `allow`) exits 0. Independently confirmed `.claude/lib/pr_review_state.py` never calls `gate.check(`; it only calls `gate.resolve_review_verdicts(...)` (the shared pipeline `check()` also consumes), so the two callers can't drift and the advisory has no path into a report a caller could misread as a block. This matches what Nadia found; I re-derived it from the import graph rather than taking her trace on faith.

### Wave-bootstrap fixture (`total_distinct == 1` via the single-reviewer exception)

This combination has no test today. I built the missing fixture against a clean worktree at the PR's head SHA (`wave-bootstrap` + `tech-debt` labels, head ref `deployments/phase-10/wave-29`, sole reviewer Aino Virtanen — a charter enforcer, no commits) and drove it through `check()` end to end:

```
{'decision': 'allow', 'systemMessage': "NOTE: PR #691 reached 1/2 WITHOUT self-review
exclusion. Head ref `deployments/phase-10/wave-29` carries no `{Initial}.{Lastname}`
branch-author prefix and the PR's commits named no roster persona either ... — #1211)."}
```

The shipped code interpolates `{total_distinct}` (not a literal), so this path is correct as shipped — it prints `1/2`, matching the actual count. The gap is coverage, not behavior: nothing pins it, so a future edit could regress it silently. This is the same finding Nadia already logged as a follow-up comment on #1203 (2026-08-03T03:51:43Z, with the mutation showing a hardcoded `2/2` literal survives all 318 tests) — I'm not duplicating it, just confirming her claim independently with a live fixture rather than static reading.

### Guard scoping — `AUTHOR_EXCLUDED` / `COMMIT_AUTHOR_EXCLUDED` can only subtract

`is_self_review` is a pure boolean predicate consumed only as a subtractive filter inside `check_comment_reviews` — there is no code path where it adds a reviewer. `refine_comment_scan_scope` has exactly one transition (`NO_BRANCH_AUTHOR` → `COMMIT_AUTHOR_EXCLUDED`, only when a real identity was derived from commits); `AUTHOR_EXCLUDED` is never touched and `NOT_RUN` is never upgraded. So a mode that silently suppresses the #1211 disclosure while exclusion was actually unavailable is not reachable: both exclusion-active modes require positive evidence (a ref prefix or a derived commit identity) before the disclosure condition (`comment_scan == NO_BRANCH_AUTHOR`) goes false. This logic is pre-existing (#1210) and untouched by this PR's diff — the PR only adds a new arm reading a field whose derivation was already fully tested.

### The modified `test_wave_merge_head_ref_runs_comment_review`

Agree with Nadia that switching its "allowed" proxy from `assertIsNone` to `decision == "allow"` plus the disclosure text is legitimate and forced by the behavior change, not a weakening.

On the nuance she flagged — the test's docstring claim that a wave-merge ref is "a no-branch-author ref by construction" is stronger than the code supports; it's true only because this test's harness (`_NoContentBindingHarness`) mocks `fetch_pr_commits` to `[]`. Checked whether that overclaim leaves an actual gap: it doesn't. `resolve_review_verdicts` already has pre-existing coverage (`RefineExclusionMonotonicityTests::test_wave_merge_branch_now_excludes_its_author_too`) proving a wave-merge ref can land on `COMMIT_AUTHOR_EXCLUDED` when its commits do name a persona — so the "ref-shape vs. fixture property" distinction is already demonstrated elsewhere in the suite, and `check()`'s disclosure condition reads that same already-tested field. No behavioral hole; the docstring phrasing is loose but doesn't hide a missing test.

---
Requestor: Nino Kavtaradze
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

Issue #1267, opened 2026-08-03
ci(base-pin): roll the #1204 digest-value + registry-host validation into the 5 vendored child-repo copies


parametrization:
## Summary

#1204 landed digest-value and registry-host validation in the **parent template** at `.claude/lib/check_dockerfile_base_pin.py`. The same script is **vendored** into child repos, whose copies are now behind the template and still accept the mutants #1204 closed.

Confirmed vendored copies (each has its own `check_dockerfile_base_pin.py` + `tests/test_check_dockerfile_base_pin.py`):

- `noorinalabs-deploy`
- `noorinalabs-isnad-graph`
- `noorinalabs-user-service`
- `noorinalabs-data-acquisition`
- `noorinalabs-isnad-ingest-platform`

Until they are updated, a `FROM ...@sha256:deadbeef` or a swap to an unexpected registry passes the base-pin gate in every repo that actually **has** Dockerfiles — which is all of them. The parent repo ships no Dockerfiles, so the template change alone protects nothing at runtime.

## Why it was not done in the #1204 PR

That PR was implemented from an **isolated parent worktree**, where the harness blocks git operations against the nested child repos. Rolling the change into five child repos means five separate branches/PRs with their own CI, which is a different unit of work from the template change and should not have been smuggled into a parent-repo PR.

## Work

For each of the five repos:

1. Sync the vendored `check_dockerfile_base_pin.py` and its test file from the parent template.
2. Check each repo's real `FROM` lines against the new rules **before** merging — the org-wide sweep run during #1204 found zero verdict changes across all 13 Dockerfiles, so this should be a no-op, but re-verify per repo rather than trusting the parent-side sweep.
3. Decide per repo whether any `--allow-registry HOST` declaration is needed in that repo's `.pre-commit-config.yaml` hook entry **and** its CI step (they must match, or the local⇄CI parity gate is violated). On current evidence none is needed — every child base image is an un-prefixed Docker Hub ref.
4. Fix the same 65-hex `_DIGEST` test fixture defect in each vendored test file (it was copied along with the original).

## Note on the vendoring itself

Five hand-synced copies of a gate is the underlying problem this issue keeps re-paying. Whether the base-pin gate (and the other `.claude/lib/check_*.py` templates) should be distributed rather than copy-pasted is a broader question than this rollout; flagging it here rather than solving it, since #744 already owns the rollout mechanism.

Follow-up from #1204.

parametrization:
[...]

Issue #1266, opened 2026-08-03
ci(base-pin): no gate asserts forward-monotonicity of a digest bump — a rollback to an older base still passes everything


parametrization:
## Summary

Item 2 of #1204, split out because it needs network access and is a materially larger change than the static validation that landed there.

#1204 closed three of the four silent mutants from its evidence table (malformed digest, empty digest, registry swap) with no-network static validation. **The rollback mutant is still caught by nothing.**

A digest-pin refresh that silently moves *backwards* to an older base — losing exactly the distro security updates the refresh exists to pick up — is:

- well-formed (`^sha256:[0-9a-f]{64}$` passes),
- on an allowed registry (`docker.io`),
- buildable (`docker build` succeeds; the digest is real),
- e2e-green.

Nothing in CI asserts forward-monotonicity of a pin change, which is the one property a base-image *refresh* is supposed to have.

## Evidence that the gap is live

`.claude/lib/tests/test_check_dockerfile_base_pin.py` asserts the gap in code — `KnownGapRollbackStillPasses::test_rollback_to_an_older_valid_digest_is_not_detected` feeds the gate two real `python:3.14-slim` digests, one older than the other, and asserts both produce zero violations. That test is written to **fail** when this issue is fixed, so the fix cannot land without consciously deleting the standing claim that the gap exists.

## Proposed fix

On a PR that changes a `FROM` digest: resolve both the old and the new digests' config blobs and assert `created(new) > created(old)`.

This is precisely the check run by hand during the merge-gate review of noorinalabs-deploy#691 (old base created 2026-06-24, new 2026-07-14 — a genuine forward move). Mechanising it means:

- a **PR-scoped** check, not a per-file lint — it needs the base ref's version of the Dockerfile to know what the old digest was, so it does not belong in the existing `check_dockerfile_base_pin.py` per-file seam;
- registry API access (token acquisition for Docker Hub / ghcr.io, manifest → config blob fetch);
- a deliberate decision on failure modes: registry unreachable, digest garbage-collected upstream, and multi-arch manifest lists (which digest's `created` is authoritative). Per the org's standing stance, an *indeterminate* comparison should fail loud rather than pass — but that trades against making the gate flaky on registry outages, so the design needs a real answer, not a default.

## Why separate

[...]

Issue #1180, opened 2026-07-30
validate_matrix_names: unclassified role slots default to the WIDE resolution set — invert to a REVIEW_CLASS_ROLES allowlist


parametrization:
Surfaced by the merge-gate review of #1178 (head `256f79a0`). Latent — no current instance; all four live slots are classified.

## The seam

`validate_matrix_names.py:377`

```python
candidates = combined if role in COMMIT_CAPABLE_ROLES else review_combined
```

`COMMIT_CAPABLE_ROLES = frozenset({"implementer"})` is an allowlist of the **narrow** side, so the default for any role nobody classifies is the **wide** side — manifest-resolved, and (pre-existing) exempt from the #1134 membership check. The module docstring argues at length that `implementer` must stay narrow so the failure is a loud exit-1 rather than a silent pass; that reasoning is not applied to the default branch.

## Measured base→head differential (2026-07-29, main checkout)

Matrix `{"noorinalabs-user-service": {"co_implementer": "Nikolaos Papadopoulos", "pair_implementer": "Annunaki", "fixer": "Steven French"}}`:

| | result |
|---|---|
| `main` | `3/3 names UNRESOLVED`, **exit 1** |
| `256f79a0` | `all 3 names resolved`, **exit 0** |

`co_implementer` is semantically commit-capable — it would have to commit in `user-service` — and passes silently.

## Related, pre-existing (not introduced by #1178)

`validate_scope` iterates a hardcoded tuple `("implementer", "reviewer", "reviewer_2", "merge_gate_reviewer")`, so in **scope mode** a new slot is not merely mis-classified, it is skipped entirely. Measured on a 3-slot row: `all 2 names resolved`, exit 0.

## Fix direction

Invert the seam: make the **review** class the allowlist.

```python
REVIEW_CLASS_ROLES = frozenset({"reviewer", "reviewer_2", "merge_gate_reviewer"})
candidates = review_combined if role in REVIEW_CLASS_ROLES else combined
```

Unknown ⇒ narrow ⇒ loud exit-1, forcing explicit classification of any new role. Pair it with either driving scope mode's slot list off the same two frozensets, or an explicit unknown-slot-key warning so a new role cannot be silently unchecked.

## Acceptance

- [ ] An unclassified role slot resolves against the narrow set and exits 1 on a manifest-only name
[...]

PR #1275, opened 2026-08-03
ci(base-pin): validate the digest value and registry host, not just the FROM shape (#1204 item 1)

### LINKED ISSUE #1204: ci(base-pin): the gate validates FROM shape but not digest value — a rollback to an older base passes every check
## Summary

`.claude/lib/check_dockerfile_base_pin.py` enforces the **shape** of a `FROM` line (a `@sha256:` substring is present, and the matching distro upgrade runs) but performs **no validation of the digest value itself**. Found while merge-gate reviewing noorinalabs-deploy#691 (a Dependabot digest bump), where I mutation-tested the gate rather than trusting its green.

## Evidence — mutation test

Ran the gate (fetched from `noorinalabs-deploy` at `837c272a19895d1042985eb08562b38aafdc878b`) over seven mutants of `integration-tests/fake_oauth/Dockerfile`:

| Mutant | Gate result |
|---|---|
| unmodified head | pass (correct) |
| digest stripped -> floating `python:3.14-slim` | **caught** |
| `apt-get -y upgrade` downgraded to `apt-get update` only | **caught** |
| `@sha256:deadbeef` (malformed, 8 hex chars) | silent |
| `@sha256:` (empty digest) | silent |
| rollback to the previous, older digest | silent |
| registry swapped to `evil.example.com/python:3.14-slim@<valid digest>` | silent |

The two rules the gate advertises are genuinely live — it is not an inert assertion. The gap is that everything about the digest *value* is unchecked. The single check is `if "@sha256:" not in stmt.image` (`check_dockerfile_base_pin.py:156`).

## Why it matters

Three of the four silent mutants are caught downstream: a malformed, empty, or nonexistent-registry digest fails `docker build` when the e2e stack builds the image, so they cannot reach `main` green.

**The rollback mutant is caught by nothing.** A digest-pin refresh that silently moves *backwards* to an older base — losing the distro security updates the refresh exists to pick up — builds cleanly, passes the base-pin gate, and passes the e2e. Nothing in CI asserts forward-monotonicity of a pin change, which is the one property a base-image *refresh* is supposed to have.

## Proposed fix

1. **Cheap, no network:** validate the digest is well-formed (`^sha256:[0-9a-f]{64}$`) and that the image reference has no registry host other than the expected one. Closes the malformed/empty/registry-swap mutants at lint time instead of relying on a build failure several minutes later.
2. **Monotonicity (needs network, PR-scoped):** on a PR that changes a `FROM` digest, resolve both old and new digests' config blobs and assert `created(new) > created(old)`. This is exactly the check I ran by hand on deploy#691 (old base created 2026-06-24, new 2026-07-14 — a genuine forward move). It is the only thing that would catch a rollback.

Item 1 is a small, self-contained change to the existing script + its unit tests. Item 2 is larger and is reasonably a separate follow-up if it needs to be split.

## Scope

Org-wide: the gate is the `noorinalabs-main` template rolled out to child repos per #744, and the same script is vendored in `noorinalabs-deploy` (`.claude/lib/check_dockerfile_base_pin.py`) and mirrored as the `dockerfile-base-pin` pre-commit hook. A fix must land in the template and be rolled to the child copies.

Related but distinct: #947 (buildx GHA cache defeating the in-image apt upgrade) is about whether the upgrade *executes*; this is about whether the digest *value* is sane.

Filed from the merge-gate review of https://github.com/noorinalabs/noorinalabs-deploy/pull/691.

parametrization:
Closes #1204 (item 1 only — see **Deliberately out of scope** below; this PR does **not** close the rollback gap).

## The defect

`check_dockerfile_base_pin.py` enforced the *shape* of a pin and nothing about its value. The entire digest check in `check_dockerfile_text` was:

```python
if "@sha256:" not in stmt.image:
```

A substring test. So `@sha256:deadbeef`, `@sha256:`, and a swap to `evil.example.com/python:3.14-slim@<valid digest>` all passed the gate green.

## What this adds

**1a — digest well-formedness.** `^sha256:[0-9a-f]{64}$`. Lowercase-only, matching the OCI grammar, so a ref Docker itself would refuse cannot pass our lint. `digest_of` splits on the **first** `@`, so a well-formed prefix followed by junk fails rather than being silently truncated to something valid.

**1b — registry-host allowlist.** The interesting design question, and the one the issue asked to be answered explicitly rather than by hardcoding a hostname.

"Expected" cannot be one hostname: this file is an org-wide template vendored into seven repos that legitimately pull from different places. So it is a **declared policy set**, layered:

1. `DEFAULT_ALLOWED_REGISTRIES` — the built-in org floor;
2. repeatable `--allow-registry HOST` — a per-repo extension, declared in that repo's `.pre-commit-config.yaml` `args:` **and** its CI step, so the widening is itself a reviewed, diffable line rather than a silent default;
3. the existing `# RATIONALE:` comment — still the total per-`FROM` escape hatch, acceptable precisely because it is visible in the diff.

Anything outside (1) ∪ (2), uncovered by (3), is **exit 1** with a message naming the flag needed to declare it. A legitimate-but-unanticipated registry fails loudly; it is never silently permitted. That is the standing unknown ⇒ narrow ⇒ loud stance (cf. #1180).

The floor is narrow **by evidence**, not by taste. I surveyed every `FROM` line in all seven repos before choosing it:

| Host | In the floor? | Why |
|---|---|---|
| `docker.io` | yes | Every real base image in the org is an un-prefixed Docker Hub ref (`python:*-slim`, `node:*-alpine`, `nginx:*-alpine`, `neo4j`, `ubuntu`). |
| `ghcr.io` | yes | The org's own registry — every child repo publishes to it via `ghcr-publish.yml`, so a `FROM ghcr.io/noorinalabs/...` base built by us is anticipated. |
| `gcr.io` | yes | Distroless is a family the charter's own distro table names as supported, and it is published only under `gcr.io/distroless/...`. |
| `quay.io` | **no** | Appears only in compose `image:` refs (`quay.io/prometheuscommunity/postgres-exporter`), never in a `FROM`. This gate reads `FROM` lines. Widening a default on non-evidence is the same class of defect as permitting `evil.example.com`. |

Three deliberate non-widenings, each with a test:

- **No alias canonicalization** — `index.docker.io` is a different literal host and must be allowlisted on its own; folding it into `docker.io` would be a silent widening.
- **A port is part of the host** — `ghcr.io:5000` is not covered by allowlisting `ghcr.io`.
- **A build-arg registry component is flagged**, not assumed — `FROM ${REGISTRY}/img@sha256:...` is not statically determinable, so it fails rather than being treated as Docker Hub. A variable in the *tag* only (`FROM neo4j:${NEO4J_VERSION}`, a real child-repo shape) is correctly left alone: the registry there **is** statically known.

**CLI.** Flag parsing is hand-rolled rather than argparse so `main` keeps returning its exit code instead of raising `SystemExit` (the existing tests depend on that contract). An unrecognized `--flag` is a usage error, never a path — a typo'd `--allow-registy` must not be swallowed into the file list, leaving the default allowlist quietly in force.

**Charter.** `tech-decisions.md § Base Image Pinning` gains the matching prose. This is a prose→code gate; the code must not enforce policy the section does not state.

## A real defect the new rule found immediately

The test suite's own `_DIGEST` fixture had carried **65** hex characters since #735 — a digest no registry would accept, invisible to a substring check, sitting in the file that exists to prove the gate works. The new rule flagged it on the first run. Corrected to 64.

### The original #1204 seven-mutant corpus, re-run through the real CLI

Old gate vs. this branch, actual exit codes:

| Mutant | old gate | this PR |
|---|---|---|
| 1 unmodified head | pass (correct) | pass (correct) |
| 2 digest stripped → floating tag | **caught** | **caught** |
| 3 upgrade downgraded to `apt-get update` | **caught** | **caught** |
| 4 malformed digest `@sha256:deadbeef` | silent | **caught** (exit 1) |
| 5 empty digest `@sha256:` | silent | **caught** (exit 1) |
| 6 **rollback to an older digest** | silent | **still silent — out of scope, see below** |
| 7 registry swapped to `evil.example.com` | silent | **caught** (exit 1) |

Real output on mutant 7:

```
Base Image Pinning violations (tech-decisions.md § Base Image Pinning, #735):
  7.dockerfile:1: FROM evil.example.com/python:3.14-slim@sha256:d3400aa122fa42cf0af0dbe8ec3091b047eac5c8f7e3539f7135e86d855dc015 — registry 'evil.example.com' is not in the allowed set (docker.io, gcr.io, ghcr.io); if this registry is legitimate for this repo, declare it with --allow-registry evil.example.com in .pre-commit-config.yaml and CI
```

and on mutant 4:

```
  4.dockerfile:1: FROM python:3.14-slim@sha256:deadbeef — malformed digest 'sha256:deadbeef' (require sha256: followed by exactly 64 lowercase hex characters)
```

### Mutation table — my own gate held to the standard that produced the finding

The #1204 finding came from mutation-testing someone else's green gate, so this one gets the same treatment. Each row weakens or deletes exactly **one** new rule in a copy of the script and runs the real suite against the mutant. All test names and verdicts below are actual harness output, not claims.

| Mutation | What it weakens | Verdict | Failing tests |
|---|---|---|---|
| M1 | digest length quantifier `{64}` → `+` (any number of hex chars) | KILLED | `DigestValueValidated::test_short_hex_digest_is_malformed`<br>`DigestValueValidated::test_sixty_five_hex_digest_is_malformed`<br>`DigestValueValidated::test_sixty_three_hex_digest_is_malformed` |
| M2 | digest charset `[0-9a-f]` → `[0-9a-fA-F]` (accept uppercase hex) | KILLED | `DigestValueValidated::test_uppercase_hex_digest_is_malformed` |
| M3 | delete the digest-value branch entirely (revert to shape-only) | KILLED | `DigestValueValidated::test_bare_at_with_no_algorithm_is_malformed`<br>`DigestValueValidated::test_empty_digest_is_malformed`<br>`DigestValueValidated::test_non_sha256_algorithm_is_malformed`<br>`DigestValueValidated::test_second_at_sign_is_not_truncated_away`<br>`DigestValueValidated::test_short_hex_digest_is_malformed`<br>`DigestValueValidated::test_sixty_five_hex_digest_is_malformed`<br>`DigestValueValidated::test_sixty_three_hex_digest_is_malformed`<br>`DigestValueValidated::test_uppercase_hex_digest_is_malformed` |
| M4 | `digest_of` splits on the LAST `@` (silently truncates trailing junk) | KILLED | `DigestValueValidated::test_digest_of_helper` |
| M5 | delete the registry allowlist check (never flag a host) | KILLED | `CliBehavior::test_disallowed_registry_exits_one_by_default`<br>`RegistryAllowlistEnforced::test_default_floor_is_narrow_not_a_catch_all`<br>`RegistryAllowlistEnforced::test_docker_hub_alias_is_not_silently_canonicalized`<br>`RegistryAllowlistEnforced::test_evil_registry_is_flagged`<br>`RegistryAllowlistEnforced::test_localhost_registry_is_flagged`<br>`RegistryAllowlistEnforced::test_port_is_part_of_the_host` |
| M6 | `registry_host` always returns `None` (every ref treated as Docker Hub) | KILLED | `CliBehavior::test_disallowed_registry_exits_one_by_default`<br>`RegistryAllowlistEnforced::test_default_floor_is_narrow_not_a_catch_all`<br>`RegistryAllowlistEnforced::test_docker_hub_alias_is_not_silently_canonicalized`<br>`RegistryAllowlistEnforced::test_evil_registry_is_flagged`<br>`RegistryAllowlistEnforced::test_localhost_registry_is_flagged`<br>`RegistryAllowlistEnforced::test_port_is_part_of_the_host`<br>`RegistryAllowlistEnforced::test_registry_host_helper` |
| M7 | widen the default floor with an unevidenced registry (`quay.io`) | KILLED | `CliBehavior::test_disallowed_registry_exits_one_by_default`<br>`RegistryAllowlistEnforced::test_default_floor_is_narrow_not_a_catch_all` |
| M8 | canonicalize Docker Hub aliases (silent widening) | KILLED | `RegistryAllowlistEnforced::test_docker_hub_alias_is_not_silently_canonicalized` |
| M9 | drop the variable-registry check (assume Docker Hub when unknown) | KILLED | `RegistryAllowlistEnforced::test_variable_registry_component_is_flagged` |
| M10 | CLI: unknown `--flag` falls through to the path list instead of erroring | KILLED | `CliBehavior::test_parse_args_rejects_unknown_flag`<br>`CliBehavior::test_unknown_option_is_usage_error_not_a_path` |
| M11 | CLI: `--allow-registry` parsed but discarded (extension is inert) | KILLED | `CliBehavior::test_allow_registry_flag_makes_it_pass`<br>`CliBehavior::test_parse_args_extends_rather_than_replaces` |
| M12 | `allowed_registries` parameter ignored; always the built-in default | KILLED | `CliBehavior::test_allow_registry_equals_form_makes_it_pass`<br>`CliBehavior::test_allow_registry_flag_makes_it_pass`<br>`RegistryAllowlistEnforced::test_explicit_allowlist_extension_permits_the_registry` |

**12/12 mutants killed.**

M10 is worth calling out: the first version of `test_unknown_option_is_usage_error_not_a_path` used the separate-value form (`--allow-registy quay.io`) and **passed under the mutant** — with the guard removed the orphaned `quay.io` failed the is-a-file check and the CLI still exited 2, so the test was green for the wrong reason. The mutation run is what exposed that; the test now uses `--allow-registy=quay.io`, where a fall-through genuinely exits 0. That is the mutation harness catching a weak test of mine, which is exactly what it is for.

### No false positives on real content

Old vs. new verdicts compared across **all 13 real Dockerfiles** in the org (`.claude/lib/`, all seven child repos, excluding vendored/`.venv`/`data/raw`): **zero files changed verdict.** The new rules add strictness without re-flagging anything currently compliant.

### Gates

Full local suite green, both stages, no `--no-verify`:

- `54 passed` in the gate's own suite; `3553 passed, 647 subtests passed` across `.claude/lib/tests/` + `.claude/hooks/tests/`
- commit stage: ruff-format, ruff-lint, actionlint, cspell, checksums-ascii — all Passed
- push stage: mypy, mypy-skills, pytest, pytest-skills, memory-budget, headcount-budget, doc-freshness, office-drift, mermaid-render — all Passed
- `pre_commit_ci_sync.py` — `OK: pre-commit config mirrors all CI-enforced checks`
- markdownlint-cli2 on the charter edit — `0 issues`

## Deliberately out of scope

**1. Forward-monotonicity (item 2 of #1204) → #1266.** Needs network access to both images' config blobs and is a PR-scoped check, not a per-file lint; putting network I/O in this hermetic script would make every local commit depend on registry reachability.

**This PR does not close the rollback gap.** Mutant 6 — a rollback to an older but perfectly well-formed digest on an allowed registry — still passes everything here, and monotonicity is the only thing that catches it. That boundary is asserted in code, not just in prose: `KnownGapRollbackStillPasses::test_rollback_to_an_older_valid_digest_is_not_detected` feeds the gate two real `python:3.14-slim` digests, one older, and asserts both are clean. It is written to **fail** when #1266 lands, forcing that PR to delete the stale claim consciously.

**2. The child-repo rollout → #1267.** The same script is vendored in five child repos (`deploy`, `isnad-graph`, `user-service`, `data-acquisition`, `isnad-ingest-platform`). This PR lands the **template only**. I implemented from an isolated parent worktree, where the harness blocks git operations against nested child repos, and five child PRs with their own CI is a different unit of work regardless. **Until #1267 lands, the child copies still accept the mutants this PR closes** — and since the parent repo ships no Dockerfiles, the template change protects nothing at runtime on its own. #1267 also carries the same 65-hex fixture fix, which was copied into every vendored test file.

Both follow-ups are labelled `tech-debt` / `process` / `phase-10`, no wave label.

## Files

- `.claude/lib/check_dockerfile_base_pin.py` — `_DIGEST_RE`, `DEFAULT_ALLOWED_REGISTRIES`, `digest_of`, `registry_host`, `_registry_component_is_variable`, `parse_args`; `check_dockerfile_text`/`check_file` take an optional `allowed_registries`
- `.claude/lib/tests/test_check_dockerfile_base_pin.py` — `DigestValueValidated`, `RegistryAllowlistEnforced`, `KnownGapRollbackStillPasses`, extended `CliBehavior`; `_DIGEST` fixture corrected
- `.claude/team/charter/tech-decisions.md` — § Base Image Pinning gains the digest-value + registry-host rules and an explicit statement of what is *not* covered
- `.pre-commit-config.yaml`, `.github/workflows/ci.yml` — comments documenting `--allow-registry` for child repos (both must match, per the parity rule)

### DIFF
--- a/.claude/lib/check_dockerfile_base_pin.py
+++ b/.claude/lib/check_dockerfile_base_pin.py
@@ -13,6 +13,11 @@ What this gate enforces (verbatim from the section's distro table)
 For every `FROM` that is not exempt:
 
 1. The image reference MUST carry an `@sha256:` digest pin.
+1a. The digest VALUE must be well-formed — `sha256:` followed by exactly 64
+    lowercase hex characters (noorinalabs-main#1204). Shape-only checking let
+    `@sha256:deadbeef` and `@sha256:` through the gate.
+1b. The registry host the reference resolves to must be in the allowed set
+    (see "Registry allowlist" below) — also #1204.
 2. The stage MUST run the package-manager upgrade matching the base distro:
 
    | Family      | Pin shape               | Upgrade command                                 |
@@ -37,16 +42,50 @@ Exemptions honored (verbatim from the section)
   upstream stage's pin, so it is not re-checked (the named stage was checked at
   its defining `FROM`).
 - **Vendor images** documented with an inline `# RATIONALE:` comment on the
-  `FROM` line (or the comment line immediately above it).
+  `FROM` line (or the comment line immediately above it). This remains the full
+  escape hatch: a RATIONALE-documented `FROM` skips *every* rule below,
+  including the digest-value and registry rules, because the exemption is
+  visible and reviewed in the diff.
+
+Registry allowlist — what "expected registry" means (#1204)
+==========================================================
+There is no single expected hostname to hardcode: this file is an org-wide
+template vendored into every child repo, and a repo may legitimately pull a base
+image from a registry another repo never touches. "Expected" is therefore a
+**declared policy set**, layered, with the narrow/loud default the org's
+standing stance on classifier defaults requires (unknown ⇒ narrow ⇒ loud exit-1):
+
+1. `DEFAULT_ALLOWED_REGISTRIES` — the built-in org floor below. Every entry is
+   evidence-grounded (see the per-entry comments); a registry that no repo pulls
+   a base image from is deliberately NOT here.
+2. `--allow-registry HOST` (repeatable) — a per-repo extension, declared in that
+   repo's `.pre-commit-config.yaml` `args:` and its CI step, so the widening is
+   itself a reviewed, diffable line rather than a silent default.
+3. `# RATIONALE:` on the `FROM` — the per-statement escape hatch.
+
+Anything outside (1) ∪ (2), not covered by (3), is a violation. A legitimate but
+unanticipated registry therefore FAILS LOUDLY with a message naming the flag —
+it is never silently permitted. Two deliberate non-widenings:
+
+- **No alias canonicalization.** `index.docker.io` / `registry-1.docker.io` are
+  not silently folded into `docker.io`; a ref naming a different literal host
+  must be allowlisted explicitly.
+- **Ports are part of the host.** `registry.example.com:5000` must be
+  allowlisted with its port; the bare hostname does not cover it.
+
+A reference whose registry component is a build-arg/variable (`FROM
+${REGISTRY}/img@sha256:...`) is not statically determinable and is flagged
+rather than assumed to be Docker Hub.
 
 CLI
 ===
-    python3 .claude/lib/check_dockerfile_base_pin.py <Dockerfile> [<Dockerfile> ...]
+    python3 .claude/lib/check_dockerfile_base_pin.py [--allow-registry HOST]... \
+        <Dockerfile> [<Dockerfile> ...]
 
 Exit codes:
     0 — every FROM in every file is compliant (or exempt)
     1 — at least one violation (each printed as `path:line: FROM <image> — <why>`)
-    2 — usage / file-not-found error
+    2 — usage / unknown-flag / file-not-found error
 
 This is a reusable template (same CLI/exit-code shape as
 `.claude/lib/pre_commit_ci_sync.py`) so it wires identically into a repo's
@@ -59,7 +98,47 @@ from __future__ import annotations
 
 import re
 import sys
+from collections.abc import Iterable
 from pathlib import Path
+
+# --- Digest value (#1204) ---------------------------------------------------
+# An OCI content digest is `sha256:` + exactly 64 LOWERCASE hex characters. The
+# registry spec's grammar is lowercase-only, so uppercase hex is rejected rather
+# than normalized — a ref Docker itself would refuse must not pass the lint.
+_DIGEST_RE = re.compile(r"^sha256:[0-9a-f]{64}$")
+
+# --- Registry host (#1204) --------------------------------------------------
+# A ref with no explicit registry component resolves to Docker Hub. This is the
+# host such a ref is checked against; it is NOT an alias table (see the module
+# docstring — `index.docker.io` is a different literal host and must be
+# allowlisted on its own).
+_IMPLICIT_REGISTRY = "docker.io"
+
+# The org floor. Narrow by construction: each entry is a registry some repo in
+# this org actually pulls a base image from, or that the charter's own distro
+# table names. A registry used only for `image:` refs in compose (e.g.
+# `quay.io/prometheuscommunity/postgres-exporter`) is NOT here — this gate reads
+# `FROM` lines, and widening the default on non-evidence is the same defect as
+# permitting `evil.example.com`. Repos that legitimately need more declare it
+# with `--allow-registry`.
+DEFAULT_ALLOWED_REGISTRIES: frozenset[str] = frozenset(
+    {
+        # Every real base image across all seven repos is an un-prefixed Docker
+        # Hub ref (python:*-slim, node:*-alpine, nginx:*-alpine, neo4j, ubuntu).
+        _IMPLICIT_REGISTRY,
+        # The org's own registry — every child repo publishes to it via its
+        # `ghcr-publish.yml`, so a `FROM ghcr.io/noorinalabs/...` base built by
+        # us is an anticipated pattern.
+        "ghcr.io",
+        # Distroless is a family the charter's own distro table names as
+        # supported, and it is published only under `gcr.io/distroless/...`.
+        "gcr.io",
+    }
+)
+
+# A `$VAR` / `${VAR}` occurrence — used only against the registry component, to
+# tell "statically Docker Hub" apart from "we cannot know".
+_VARIABLE_RE = re.compile(r"\$\{?\w")
 
 # `FROM` is the only keyword we parse; Dockerfile keywords are case-insensitive.
 _FROM_RE = re.compile(r"^\s*FROM\s+(?P<rest>\S.*?)\s*$", re.IGNORECASE)
@@ -127,6 +206,48 @@ def parse_froms(text: str) -> list[FromStatement]:
     return froms
 
 
+def digest_of(image: str) -> str | None:
+    """Return the digest component of an image ref, or None if it carries none.
+
+    Everything after the FIRST `@` is the digest, so a ref with a stray second
+    `@` yields a value that fails `_DIGEST_RE` rather than being silently
+    truncated to a well-formed prefix.
+    """
+    if "@" not in image:
+        return None
+    return image.split("@", 1)[1]
+
+
+def registry_host(image: str) -> str | None:
+    """Return the ref's explicit registry host, or None for implicit Docker Hub.
+
+    Docker's own rule: the first `/`-separated component is a registry iff it
+    contains a `.` or a `:`, or is exactly `localhost`. Otherwise it is a Hub
+    namespace (`myorg/myimage`), and a ref with no `/` at all is a Hub library
+    image (`python:3.14-slim`).
+    """
+    ref = image.split("@", 1)[0]
+    if "/" not in ref:
+        return None
+    head = ref.split("/", 1)[0]
+    if head == "localhost" or "." in head or ":" in head:
+        return head
+    return None
+
+
+def _registry_component_is_variable(image: str) -> bool:
+    """True if the ref has a registry component that is a build arg/variable.
+
+    Only the component before the first `/` is inspected: a variable in the
+    NAME or TAG (`FROM neo4j:${NEO4J_VERSION}`) leaves the registry statically
+    known (Docker Hub), whereas `FROM ${REGISTRY}/img@sha256:...` does not.
+    """
+    ref = image.split("@", 1)[0]
+    if "/" not in ref:
+        return False
+    return bool(_VARIABLE_RE.search(ref.split("/", 1)[0]))
+
+
 def _stage_body(text: str, from_lineno: int, next_from_lineno: int | None) -> str:
     """Return the text between a `FROM` (exclusive) and the next `FROM`/EOF."""
     lines = text.splitlines()
@@ -135,8 +256,20 @@ def _stage_body(text: str, from_lineno: int, next_from_lineno: int | None) -> st
     return "\n".join(lines[start:end])
 
 
-def check_dockerfile_text(path: str, text: str) -> list[str]:
-    """Return a list of human-readable violation strings for one Dockerfile."""
+def check_dockerfile_text(
+    path: str,
+    text: str,
+    allowed_registries: Iterable[str] | None = None,
+) -> list[str]:
+    """Return a list of human-readable violation strings for one Dockerfile.
+
+    `allowed_registries` defaults to `DEFAULT_ALLOWED_REGISTRIES`; callers pass
+    the repo's declared set (default ∪ `--allow-registry` values).
+    """
+    allowed = {
+        h.lower()
+        for h in (DEFAULT_ALLOWED_REGISTRIES if allowed_registries is None else allowed_registries)
+    }
     violations: list[str] = []
     froms = parse_froms(text)
     defined_stages: set[str] = set()
@@ -153,11 +286,39 @@ def check_dockerfile_text(path: str, text: str) -> list[str]:
         if is_exempt:
             continue
 
-        if "@sha256:" not in stmt.image:
+        # Rule 1 — a digest pin is present, and (1a, #1204) its VALUE is a
+        # well-formed sha256 content digest. Shape-only checking passed
+        # `@sha256:deadbeef` and `@sha256:`.
+        digest = digest_of(stmt.image)
+        if digest is None:
             violations.append(
                 f"{path}:{stmt.lineno}: FROM {stmt.image} — not digest-pinned "
                 f"(require image:tag@sha256:<digest>)"
             )
+        elif _DIGEST_RE.match(digest) is None:
+            violations.append(
+                f"{path}:{stmt.lineno}: FROM {stmt.image} — malformed digest "
+                f"'{digest}' (require sha256: followed by exactly 64 lowercase "
+                f"hex characters)"
+            )
+
+        # Rule 1b (#1204) — the ref must resolve to an allowed registry host.
+        if _registry_component_is_variable(stmt.image):
+            violations.append(
+                f"{path}:{stmt.lineno}: FROM {stmt.image} — registry host is not "
+                f"statically determinable (it is a build arg/variable); use a "
+                f"literal registry host so the allowlist can be enforced"
+            )
+        else:
+            host = registry_host(stmt.image) or _IMPLICIT_REGISTRY
+            if host.lower() not in allowed:
+                violations.append(
+                    f"{path}:{stmt.lineno}: FROM {stmt.image} — registry "
+                    f"'{host}' is not in the allowed set "
+                    f"({', '.join(sorted(allowed))}); if this registry is "
+                    f"legitimate for this repo, declare it with "
+                    f"--allow-registry {host} in .pre-commit-config.yaml and CI"
+                )
 
         body = _stage_body(text, stmt.lineno, next_lineno)
         if "distroless" in image_l:
@@ -176,17 +337,59 @@ def check_dockerfile_text(path: str, text: str) -> list[str]:
     return violations
 
 
-def check_file(path: Path) -> list[str]:
-    return check_dockerfile_text(str(path), path.read_text(encoding="utf-8"))
+def check_file(path: Path, allowed_registries: Iterable[str] | None = None) -> list[str]:
+    return check_dockerfile_text(str(path), path.read_text(encoding="utf-8"), allowed_registries)
+
+
+_USAGE = (
+    "usage: check_dockerfile_base_pin.py [--allow-registry HOST]... <Dockerfile> [<Dockerfile> ...]"
+)
+
+
+def parse_args(argv: list[str]) -> tuple[list[str], set[str]] | None:
+    """Split argv into (paths, allowed registries), or None on a usage error.
+
+    Hand-rolled rather than argparse so `main` keeps returning its exit code
+    instead of raising SystemExit. An unrecognized `--flag` is a usage error,
+    never a path — a typo'd `--allow-registy` must not be silently swallowed
+    into the file list and leave the default allowlist quietly in force.
+    """
+    paths: list[str] = []
+    allowed = set(DEFAULT_ALLOWED_REGISTRIES)
+    i = 0
+    while i < len(argv):
+        arg = argv[i]
+        if arg == "--allow-registry":
+            if i + 1 >= len(argv):
+                print("ERROR: --allow-registry requires a HOST argument", file=sys.stderr)
+                return None
+            allowed.add(argv[i + 1])
+            i += 2
+            continue
+        if arg.startswith("--allow-registry="):
+            value = arg.split("=", 1)[1]
+            if not value:
+                print("ERROR: --allow-registry requires a HOST argument", file=sys.stderr)
+                return None
+            allowed.add(value)
+            i += 1
+            continue
+        if arg.startswith("-"):
+            print(f"ERROR: unknown option: {arg}", file=sys.stderr)
+            return None
+        paths.append(arg)
+        i += 1
+    return paths, allowed
 
 
 def main(argv: list[str]) -> int:
-    paths = argv[1:]
+    parsed = parse_args(argv[1:])
+    if parsed is None:
+        print(_USAGE, file=sys.stderr)
+        return 2
+    paths, allowed = parsed
     if not paths:
-        print(
-            "usage: check_dockerfile_base_pin.py <Dockerfile> [<Dockerfile> ...]",
-            file=sys.stderr,
-        )
+        print(_USAGE, file=sys.stderr)
         return 2
 
     all_violations: list[str] = []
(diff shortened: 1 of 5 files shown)


parametrization:
## Merge-gate review — reviewed at `3a258188807f255abea58d8f6f24a01f8a1cae97`

CI: 21/21 SUCCESS at that SHA (`statusCheckRollup`, re-read immediately before posting). Read against origin via the contents API at the head SHA, not a local checkout.

I re-ran your mutation table myself rather than reading it, added 10 mutations you did not list, re-ran your own `FROM`-line survey across all seven child repos, and diffed old-vs-new gate verdicts over every real Dockerfile in the org. **Approving.** Everything you claimed reproduces; the things I found on top are narrower than what you closed and are filed as #1282.

---

### Your 12-row table: re-run independently, 11/11 reproduced

Each mutation applied as a single independent edit to a copy of the head-SHA script, full 54-test suite re-run per mutant, in a namespaced scratch dir. M3 folded into my X0 below.

```
M1_digest_len_any:              KILLED  3 failed, 51 passed
M2_uppercase_hex_ok:            KILLED  1 failed, 53 passed
M4_split_last_at:               KILLED  1 failed, 53 passed
M5_delete_registry_check:       KILLED  6 failed, 48 passed
M6_registry_host_none:          KILLED  7 failed, 47 passed
M7_widen_floor_quay:            KILLED  2 failed, 52 passed
M8_canonicalize_alias:          KILLED  1 failed, 53 passed
M9_drop_variable_check:         KILLED  1 failed, 53 passed
M10_unknown_flag_falls_through: KILLED  2 failed, 52 passed
M11_allow_registry_discarded:   KILLED  2 failed, 52 passed
M12_ignore_allowed_param:       KILLED  3 failed, 51 passed
```

Failing-test sets matched your table row for row.

### The M10 self-catch: verified, and the rewrite is genuinely load-bearing

This was the claim most worth checking, because it is a claim about a test that *used to be* vacuous. I reconstructed your pre-rewrite version (separate-value form) and ran both variants against the M10 mutant with the sibling `test_parse_args_rejects_unknown_flag` deselected so it could not mask the result:

```
under M10 mutant, CURRENT (=form)          -> KILLS mutant (fails)      1 failed, 53 deselected
under M10 mutant, PRE-REWRITE (sep-form)   -> *** SURVIVES (passes) *** 1 passed, 53 deselected
```

Exactly as you described. The `=`-form choice is load-bearing, and the comment explaining why it is deliberate is the right thing to have left in the file — without it the next person "simplifies" it back to the separate-value form and re-vacuates the test.

### The registry floor: your survey re-run, and it holds

I swept every `FROM` line org-wide with `--hidden --no-ignore` (child repos are gitignored by the parent), excluding `.venv`/`node_modules`/`.git`/`data/raw`. Every base-image `FROM` in all seven repos:

- un-prefixed Docker Hub refs — `python:3.12-slim`, `python:3.14-slim`, `node:22-alpine`, `node:24-alpine`, `nginx:alpine`, `nginx:stable-alpine3.23`, `neo4j:${NEO4J_VERSION}`
- multi-stage stage references — `FROM base`, `FROM builder`, `FROM runtime`
- two `FROM scratch` inside a `cold-rebuild-dryrun.yml` heredoc

**No `FROM quay.io/...` exists anywhere.** `quay.io` appears only where you said: compose `image:` refs in `docker-compose.prod.yml`, plus the docs/ontology rows describing them. Excluding it from the floor breaks nothing when #1267 rolls this to children.

I also ran base-vs-head verdicts over all 13 real Dockerfiles:

```
base=0 head=0   (all 13 — zero verdict changes)
```

Your no-false-positives claim is accurate. Worth recording *why* `noorinalabs-deploy/infra/neo4j/Dockerfile` stays clean: it is the org's one live use of the escape hatch, and it is a well-written one — a five-line comment explaining the build-arg parameterization *above* the `RATIONALE:` line, so the exemption is genuinely reviewable rather than a magic string.

One honest wrinkle in the floor's justification. The table in your PR body presents all three entries as chosen "by evidence," but they rest on two different standards: `docker.io` on actual-`FROM` evidence, `ghcr.io` and `gcr.io` on *anticipated* use (neither appears in any `FROM` today). I think both belong — `gcr.io` is required for distroless, which the section's own distro table names as a supported option, and `ghcr.io` is the org's own registry — and your in-code comments say "anticipated" plainly, which is the honest word. It is the PR-body table that flattens the two standards into one. Not a defect; noting it so the distinction survives into #1267's review.

### The escape hatch: not weakened, but its narrowness is now load-bearing and untested

You asked the right question implicitly by documenting it. The answer: `is_exempt → continue` predates this PR, so `RATIONALE` does not become *more* permissive relative to the old rule set — it exempts the same statements from a rule set that grew. And the widening is stated in three places (module docstring, charter prose, the violation footer). That is the correct handling of a total escape hatch.

What changed is the **blast radius**. A `RATIONALE` comment now waives the digest-value and registry-host rules too. `_preceding_comment_has_rationale` is narrow — it inspects only the first non-blank line above the `FROM`, so a `RATIONALE:` buried higher in a comment block does *not* exempt. That narrowness is now the property doing the work, and it has no test: I mutated the helper to scan the whole preceding comment block and **all 54 tests still passed**. Filed.

### Ten mutations you did not list — five survived

The code is correct in every one of these; the *guard* is what has no fixture that can trip it.

```
X0_M3_delete_digest_value_branch:      KILLED  8 failed, 46 passed
X6_no_paths_guard_removed:             KILLED  1 failed, 53 passed
X8_digest_re_unanchored_end:           KILLED  2 failed, 52 passed
X1_empty_equals_value_guard_removed:   *** SURVIVED ***  54 passed
X2_allowed_not_lowercased:             *** SURVIVED ***  54 passed
X3_variable_re_brace_only:             *** SURVIVED ***  54 passed
X4_rationale_scans_whole_comment_block:*** SURVIVED ***  54 passed
X5_localhost_literal_clause_removed:   *** SURVIVED ***  54 passed
```

(X9, dropping the `^` from `_DIGEST_RE`, also survived — correctly, since `.match()` anchors at position 0 anyway. Equivalent mutant, not a gap.)

The two that matter are X3 and X5, because each corresponds to a *silent pass* on a real ref shape: dropping the `?` from `\$\{?\w` makes `FROM $REGISTRY/img` sail through as Docker Hub, and dropping the `head == "localhost"` clause makes `FROM localhost/base:1.0@...` (no port) sail through as a Hub namespace. Both guards work today — I confirmed on the unmutated head code that both refs are flagged — but only the `${...}` and `localhost:5000` variants are fixtured.

### The one real gap I found

Both the variable check and `registry_host` bail out when the ref has no `/`. So a ref that is *entirely* a build arg is treated as implicit Docker Hub and passes clean at the head SHA:

```
FROM ${FULL_IMAGE}@sha256:<64 valid lowercase hex>
RUN apt-get update && apt-get -y upgrade && apt-get clean
   -> zero violations
```

Same class as the `evil.example.com` mutant this PR closes, and it fails *silently* — the one outcome the "unknown ⇒ narrow ⇒ loud" stance you state in both the docstring and the charter exists to prevent. `FROM ${REGISTRY}/img@...` and `FROM $REGISTRY/img@...` are both correctly flagged; it is only the slash-less form that slips. Narrow, and strictly smaller than what you closed, so it is follow-up rather than blocking — but it should land in the template *before* #1267 vendors the file five times. Filed with the fix shape and the `FROM neo4j:${NEO4J_VERSION}` interaction that must be preserved.

### Flag parser: probed, and it holds

Hand-rolled parsers being where this org has been bitten (`checksums_io.main()`, #1137) I probed it directly rather than trusting the tests:

```
exit=2  unknown flag =form              exit=2  bare --
exit=2  unknown flag sep-form           exit=2  --allow-registry no value
exit=2  single-dash unknown             exit=0  flag AFTER positional
exit=0  =form AFTER positional          exit=0  repeated flags both honored
exit=2  value swallows next flag        exit=0  uppercase host in flag
```

Every failure mode is loud. `--allow-registry --allow-registry quay.io` swallows the second flag as a value, then `quay.io` fails the is-a-file check → exit 2 with `ERROR: not a file: quay.io`. Ugly path, correct outcome. Uppercase hosts work via the `.lower()` (untested — that is X2). The one asymmetry: `--allow-registry ""` is accepted where `--allow-registry=` errors, and the empty host then renders as a leading comma in the violation message. Cosmetic, filed.

The argparse-avoidance rationale is sound and, more to the point, you wrote it down at the function. Someone will try to "modernize" this to argparse; the docstring is what stops them.

### The known-gap test: right call

This was the judgment call you flagged, and I think you got it right — with one nit.

It does not make the suite green on a defect. The suite is green on that defect *either way*; that is what a gap is. What the test adds is a tripwire at the point of enforcement, so the day #1266 lands the stale claim in the charter prose and the module docstring cannot quietly survive. The class name says `KnownGapRollbackStillPasses`, the docstring says it fails the day monotonicity lands, and it feeds two real `python:3.14-slim` digests rather than synthetic ones. Nobody can mis-skim that as coverage.

The alternative — prose only — is strictly worse: #1266 lands, nobody greps the parent's charter, and "the gate validates digest pins" quietly becomes a claim the reader over-reads. Encoding the boundary is the same discipline that produced the #1204 finding in the first place.

Nit, non-blocking: the assertion message reads `"item 1 is no-network validation; digest ORDERING is item 2"` — a justification of the status quo. When it fires, it will fire *because someone fixed this*, and they will see a message that reads like a regression. The docstring one level up says the right thing, so this is fine as-is; a future edit could fold "if this fails, monotonicity landed — delete this class" into the message itself. Worth carrying into #1266's brief so the author is not confused by their own success.

### Scope honesty

Checked both, plainly stated in the PR body, and neither is oversold:

- Forward-monotonicity — the body's first line says "item 1 only ... this PR does **not** close the rollback gap," and mutant 6 is marked "still silent" in your own comparison table rather than quietly omitted. #1266 exists, correctly labelled.
- Child rollout — "**Until #1267 lands, the child copies still accept the mutants this PR closes** — and since the parent repo ships no Dockerfiles, the template change protects nothing at runtime on its own." That is the sentence a weaker PR body would have left out. #1267 exists, correctly labelled, and carries the 65-hex fixture fix for the five vendored test files.

The charter prose carries the same boundary in its own "What this does NOT cover" paragraph, so a reader arriving at the section — not the PR — gets it too. Prose→code parity holds in both directions: I checked that the code enforces nothing the section does not state, and the section states nothing the code does not enforce. `.pre-commit-config.yaml` and `.github/workflows/ci.yml` both carry the matching `--allow-registry` guidance, per the parity rule.

### The 65-hex fixture

Verified independently against both refs:

```
origin/main   _DIGEST hex length: 65   (0272e4609f...6f8a0b2c4)
head SHA      _DIGEST hex length: 64
head SHA      _REAL_DIGEST hex length: 64
```

Real, and correctly fixed. A digest no registry would accept, living since #735 in the file whose entire job is to prove the gate works — that is the finding, not the typo. It is also the cleanest possible argument for the rule: the first thing the new check caught was its own test suite.

---

Nothing blocking. Three of the four silent mutants from #1204 are closed with real exit-code evidence, the fourth is explicitly scoped out and asserted as a gap rather than papered over, zero verdict changes across all 13 real Dockerfiles, and the mutation discipline that found the original defect was turned on the fix. Approved at `3a258188807f255abea58d8f6f24a01f8a1cae97`. See #1282 for the residual gaps — none of them a regression against base.

---

Requestor: Weronika Zielinska
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: #1282

parametrization:
Second review, at head `3a258188807f255abea58d8f6f24a01f8a1cae97`. CI: all checks SUCCESS at this SHA.

**Own verification, not inherited from Weronika's pass:**

1. **The five untested-guard survivors** — I wrote fixtures for each rather than taking the "no fixture can trip it" characterization on faith:
   - Bare `localhost/base:1.0` (no port): `registry_host()`'s `head == "localhost"` clause fires correctly — flagged as `registry 'localhost' is not in the allowed set`. Not just the ported `localhost:5000` form that's fixtured.
   - Bare unbraced `$REGISTRY/img@sha256:...`: `_VARIABLE_RE`'s `\{?` (optional brace) does its job — `_registry_component_is_variable` returns `True`, correctly flagged as not statically determinable. Only the `${REGISTRY}` braced form is in the test file, but the regex handles both.
   - `--allow-registry=` (empty value, no host): correctly rejected by `parse_args` — returns `None`, `main` exits 2. Guard works, just untested by name.
   - Allowlist case-folding: tested both directions — `--allow-registry GHCR.IO` (uppercase flag value) against a lowercase `ghcr.io` ref, and `FROM GHCR.IO/...` (uppercase ref) against the lowercase default floor. Both pass clean; the `.lower()` on both the stored set and the compared host makes it symmetric.
   - `_preceding_comment_has_rationale`'s single-line-lookup: confirmed narrow by construction (skips blank lines, stops at the first non-blank/non-comment line) — a RATIONALE separated from its `FROM` by a non-comment line (e.g. an intervening `ARG`) is NOT picked up. I checked this isn't a live regression risk: the one real `# RATIONALE:`-exempted `FROM` in the org today (`noorinalabs-deploy/infra/neo4j/Dockerfile:16-17`) has the comment on the line immediately above the `FROM`, no separator. Worth a code-comment note for future authors, not blocking.

   None of the five is a correctness bug — all are missing test coverage on guards that behave correctly. #1282's title already covers this ("5 new guards have no test that can trip them"), so no new issue needed.

2. **Child-repo blast radius.** Re-ran independently rather than trusting the prior verdict: pulled the head-SHA copy of `check_dockerfile_base_pin.py` via the contents API, located all 13 real Dockerfiles across the 7 child repos (`rg --hidden --no-ignore` over `Dockerfile*`, excluding the 4 vendored `data/raw/.../github_clone` fixture files, which aren't part of any repo's own lint scope), and ran both the pre-PR (`main`) and PR-head versions against the same file list with each repo's actual `--allow-registry` config (none of the 6 repos with a wired hook declare an extension flag today). Both versions: `OK`, 0 violations, identical. Also confirmed none of the 5 already-vendored child-repo copies (`user-service`, `ingest-platform`, `data-acquisition`, `deploy`, `isnad-graph`, all byte-identical at 217 lines, pre-#1204) are touched by this PR — they still carry the old substring-only check, which is exactly the gap #1267 exists to close. Full test suite at head: `54 passed`.

3. **Merging with #1282 open.** My own read agrees with the non-blocking call, for a sharper reason than "narrower in scope": the parent ships no Dockerfiles, so today the *actual* enforcement floor in every child repo is still the pre-#1204 217-line substring check — weaker than both what #1275 lands and what #1282 describes. Holding #1275 open protects nothing at runtime; it only delays the moment the improved template exists to vendor. #1282's gap (a slash-less variable ref) has zero live instances in any of the 13 real Dockerfiles today — the one variable-in-ref shape that exists (`FROM neo4j:${NEO4J_VERSION}`) is a variable in the *tag*, which #1275 already resolves correctly to Docker Hub. So blocking here trades a real, delivered improvement for protection against a hypothetical shape that doesn't occur yet.
   Where I'd add to Weronika's framing: the sequencing point isn't "should #1275 wait for #1282" (no), it's "should #1267 close #1282 before or alongside vendoring, rather than after" — landing 5 more copies of a known gap is worse than landing 5 copies with it closed. That's a note for whoever picks up #1267, not a blocker here.

4. **Registry-floor honesty.** The PR body's own table already marks `ghcr.io`/`gcr.io` as "anticipated" (in contrast to `docker.io`'s "every real base image") in both prose and the in-code comments — the evidence/anticipated distinction is visible without needing to dig, and should survive into #1267's review unless someone flattens the table down to a bare allowlist restatement.

No blocking findings. Approving.

---
Requestor: Lucas Ferreira
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: none

PR #1277, opened 2026-08-03
fix(#1180): invert the wave-scope role seam — REVIEW class is the allowlist

### LINKED ISSUE #1180: validate_matrix_names: unclassified role slots default to the WIDE resolution set — invert to a REVIEW_CLASS_ROLES allowlist

parametrization:
Closes #1180.

> **Correction (see the pinned comment).** The first version of this description reported the matrix-mode differential as `main` → "all 3 names resolved, exit 0". That was carried over from the issue body, not measured by me. Re-measured, it is wrong — and the corrected measurement below is a *sharper* statement of the defect, not a weaker one.

## What was wrong

`validate_matrix_names.py` chose a role slot's candidate set with

```python
candidates = combined if role in COMMIT_CAPABLE_ROLES else review_combined
```

`COMMIT_CAPABLE_ROLES = frozenset({"implementer"})` is an allowlist of the **narrow** side, so the fall-through for any unclassified role was the **wide** side. Two loosenings ride on that fall-through, and both are justified only for a slot that *provably never commits*:

1. the org-union manifest widens the resolution set (#1162), and
2. the #1134 repo-membership check does not apply.

The module docstring argues at length that `implementer` must stay narrow so a failure is a loud exit-1 rather than a silent pass. That reasoning was never applied to the default branch.

**Latent — no current instance. All four live slots are classified**, so nothing in any shipped wave behaved incorrectly. This closes the seam before a new role slot lands on it.

A related pre-existing gap sat one layer up: `validate_scope` iterated a hardcoded `("implementer", "reviewer", "reviewer_2", "merge_gate_reviewer")` tuple — a *third* place a role had to be listed — so in scope mode an unrecognised slot key was not merely mis-classified, it was never looked at.

## What changed

- **`REVIEW_CLASS_ROLES` is now the allowlist.** `candidates = review_combined if role in REVIEW_CLASS_ROLES else combined`. Unknown ⇒ narrow ⇒ loud exit 1. The membership branch is inverted the same way (`if not is_review_class and not is_parent_repo`), so an unclassified slot no longer inherits the #1134 exemption either — the issue names that exemption as part of the defect.
- **`KNOWN_ROLE_SLOTS = COMMIT_CAPABLE_ROLES | REVIEW_CLASS_ROLES` is the single source of truth.** `validate_scope` walks each row via a new `is_role_slot_key` instead of the hardcoded tuple, so classifying a role in either frozenset is the one edit that makes both modes see it.
- **Third failure class: `slot_class="unclassified"`, exit 1**, with a message naming the three frozensets to classify the key into. Chosen over a warning deliberately — a warning printed inside an otherwise-green job is exactly the advisory posture the module's own § Hard fail section already records as insufficient.

### Why `is_role_slot_key` is shaped the way it is

Scope rows mix role slots with free-form metadata, so scope mode cannot treat every string key as a person — `note`, `id`, `priority`, `scope_note` would all become bogus name checks. Nor can it hard-fail on every unrecognised key: the live data carries 25 distinct metadata keys that grow ad hoc wave to wave (`open_risk`, `slate_note`, `merged_sha`, `sequence`, …), and a gate that reds on the next ad-hoc annotation is one operators route around.

The recogniser instead uses the schema's own naming convention: **role slots are named after the agent that fills them**, so the key ends in an agentive `-er`/`-or` component, optionally with an ordinal (`reviewer_2`). Measured across **all 293 rows of every `wave_*_scope`** in `cross-repo-status.json`, that matches the four known slots plus `pre_kickoff_blocker` (a bool flag) and nothing else. `pre_kickoff_blocker` is pre-classified in `NON_ROLE_ROW_KEYS`, which is the escape hatch that keeps the regex's generosity cheap — a future false positive costs one reviewable line, and the failure is self-documenting rather than silent.

**Known limit, stated plainly:** a non-agentive unclassified slot key (say `co_lead`) is still skipped in scope mode — filed as #1278. It is caught in matrix mode by the inverted seam, and classifying it in either frozenset makes scope mode pick it up with no second edit, which is pinned by `test_scope_mode_slot_list_is_driven_by_the_frozensets`. Closing the non-agentive case would need either a metadata denylist (unbounded, given the 25 free-form keys) or a value-shape heuristic that false-positives on `found_by` / `reassigned_from`, both of which trade a real gate for a noisy one.

## The measured differential (corrected)

**The issue's own three-slot example masks the defect.** That matrix names `Annunaki` and `Steven French` alongside `Nikolaos Papadopoulos`; #1181's persona-shape filter already rejects the first two, so on today's `origin/main` the run exits **1** — on those two names — while the one slot the issue is actually about resolves silently. Isolating each loosening:

| Case (matrix mode, `noorinalabs-user-service`) | `origin/main` | this branch |
|---|---|---|
| `{"co_implementer": "Nikolaos Papadopoulos"}` — his cards live in a **third child repo**, so he is manifest-only here | `all 1 names resolved`, exit **0** | `1/1 names UNRESOLVED` + `1/1 UNCLASSIFIED`, exit **1** |
| `{"co_implementer": "Nurul Hakim"}` — on the parent roster, **not** on the user-service roster | `all 1 names resolved`, exit **0** | `1/1 CROSS-REPO IMPLEMENTER assignment(s)`, exit **1** |
| the issue's 3-slot matrix verbatim | `2/3 UNRESOLVED`, exit 1 — but `co_implementer` **passes** | `3/3 UNRESOLVED` + `3/3 UNCLASSIFIED`, exit 1 |

The second row is the sharpest statement of the risk: **it is the W27/W28 cross-repo-implementer failure — the one that cost a mechanical merge-commit re-attribution — re-admitted by nothing more than renaming the slot key.** `Nurul Hakim` on `user-service` is blocked as `implementer` and waved through as `co_implementer`.

Scope mode, a 3-slot row (`implementer` + `reviewer` + `co_implementer`):

| | result |
|---|---|
| `origin/main` | `all 2 names resolved`, exit **0** — the third slot was never looked at |
| this branch | `1/3 names UNRESOLVED` + `1/3 UNCLASSIFIED role slot(s)`, exit **1** |

## Regression anchors — measured against real wave data

`origin/main`'s validator and this branch's were run over **every** `wave_*_scope` in `cross-repo-status.json`, comparing **stdout+stderr byte-for-byte and the exit code**:

| Wave | Before | After | Output |
|---|---|---|---|
| **28** | exit **1**, `2/30 CROSS-REPO IMPLEMENTER assignment(s)` (`Nurul Hakim`, `Weronika Zielinska`) | exit **1**, identical | byte-identical |
| **29** | exit **0**, `all 165 names resolved across 51 entries` | exit **0**, identical | byte-identical |
| 1–7, 16–27 | — | — | byte-identical, exit codes unchanged |

**21 waves / 293 rows, zero differences.** Note the W29 count is 165, not the 85 quoted in the issue: `wave_29_scope` has been amended since #1180 was filed. The load-bearing anchor is the before/after equality on today's data, which is exact.

## Testing — mutation-tested, not assertion-only

62 tests pass (45 pre-existing + 17 new) plus 38 subtests. Each new branch was deleted or weakened in turn and the suite re-run; **0 of 9 mutations survived**. The harness asserts a green baseline before mutating and re-asserts a green tree after restoring.

| # | Mutation | Named test(s) that go red |
|---|---|---|
| M1 | Seam reverted to `combined if role in COMMIT_CAPABLE_ROLES else review_combined` | `test_unclassified_slot_resolves_against_the_narrow_set`, `test_unclassified_slot_gets_the_membership_check`, `test_unclassified_slot_fails_even_when_fully_valid` |
| M2 | Membership branch reverted to `role in COMMIT_CAPABLE_ROLES` | `test_unclassified_slot_gets_the_membership_check`, `test_unclassified_slot_fails_even_when_fully_valid` |
| M3 | `slot_class` marking deleted (everything reported `known`) | `test_unclassified_slot_fails_even_when_fully_valid`, `test_unrecognised_slot_key_is_not_skipped` |
| M4 | `unclassified` dropped from the exit-code condition | `test_unclassified_slot_fails_even_when_fully_valid` |
| M5 | Scope-mode discovery reverted to the hardcoded 4-slot tuple | `test_unrecognised_slot_key_is_not_skipped`, `test_scope_mode_slot_list_is_driven_by_the_frozensets` |
| M6 | `is_role_slot_key` weakened to `return True` | 36 red, incl. `test_metadata_keys_are_still_ignored_in_scope_mode`, `test_no_live_metadata_key_is_mistaken_for_a_slot` (25 subtests), `test_non_tier_keys_and_null_slots_ignored` |
| M7 | `NON_ROLE_ROW_KEYS` denylist branch deleted | `test_no_live_metadata_key_is_mistaken_for_a_slot(key='pre_kickoff_blocker')`, `test_recognition_is_case_and_whitespace_insensitive`, `test_metadata_keys_are_still_ignored_in_scope_mode` |
| M8 | Agentive recogniser deleted (unknown keys never admitted) | `test_unclassified_agentive_keys_are_recognised` (6 subtests), `test_unrecognised_slot_key_is_not_skipped` |
| M9 | `merge_gate_reviewer` dropped from `REVIEW_CLASS_ROLES` | `test_every_review_class_role_keeps_the_wide_set(role='merge_gate_reviewer')`, `test_reviewer_slots_keep_parent_union`, `test_third_child_reviewer_resolves_and_implementer_stays_gated`, +3 |

Three design notes on the tests:

- `test_unclassified_slot_fails_even_when_fully_valid` uses a name that **resolves and is a repo member**, so neither the #319 nor the #1134 class fires. It is the only test that isolates the #1180 class as its own signal rather than a side effect of a name miss — which is why it is the sole anchor for M4.
- `test_same_name_still_resolves_in_a_review_slot` is a deliberate paired control: without it, the narrow-set test would also pass if the manifest had simply stopped being loaded at all. Only the slot class differs between the two.
- `test_no_live_metadata_key_is_mistaken_for_a_slot` snapshots the measured 25-key census rather than reading `cross-repo-status.json`, so the guard cannot rot when the wave data changes.

## Docs

`/wave-scope` § 12.5, `/wave-kickoff` § 0b and `charter/skills.md` all described the membership rule as "the `implementer` slot only". Restated in the class terms the code now uses, with the new rule made explicit: **adding a role slot to the scope schema is a code change, not a data change.** The module docstring now carries the corrected, isolated measurements and a note that the issue's example matrix masks the slot it is about.

## Gates

`pre-commit run --hook-stage pre-push --all-files` fully green (17 hooks). `pre_commit_ci_sync.py` reports parity. All 21 CI checks passed on the prior head; re-running on `aee3d94`. No `--no-verify`, no failing check.

## Follow-up

- #1278 — the non-agentive unclassified slot key residual (`tech-debt` / `process` / `phase-10`, no wave label).

## Reviewers

Two per charter. Not for merge by me — the orchestrator drives the review gate.

### DIFF
--- a/.claude/skills/wave-kickoff/SKILL.md
+++ b/.claude/skills/wave-kickoff/SKILL.md
@@ -64,7 +64,8 @@ python3 "$REPO_ROOT/.claude/skills/wave-scope/validate_matrix_names.py" \
     --scope "$REPO_ROOT/cross-repo-status.json" --wave {M}
 RC=$?
 if [ "$RC" -ne 0 ]; then
-  echo "ERROR: wave_{M}_scope has unresolved names or cross-repo implementer(s)."
+  echo "ERROR: wave_{M}_scope has unresolved names, cross-repo implementer(s),"
+  echo "  or an unclassified role slot key (#1180)."
   echo "  Fix the scope rows (reassign, onboard, or record a roster_union_override"
   echo "  with a rationale) and re-run. See /wave-scope SKILL.md § 12.5."
   exit 1
@@ -72,6 +73,8 @@ fi
 ```
 
 Deterministic and hermetic by default — reads only local rosters and `cross-repo-status.json`, no GitHub API calls. A child repo not cloned locally is reported `unverified` and does not block; add `--fetch-missing` to resolve those over the network. Charter rule: `charter/skills.md` § Wave Scoping — Child-Repo Implementer Must Be a Repo-Roster Member.
+
+Since #1180 this step also fails on an **unclassified role slot key** — a row slot in neither `COMMIT_CAPABLE_ROLES` nor `REVIEW_CLASS_ROLES`. Adding a new role slot to the scope schema is a code change: classify the key in `validate_matrix_names.py` so both the resolution set and the membership check are chosen deliberately rather than inherited from a fall-through.
 
 ### 0. Derive wave repos in scope (Mandatory first step)
 
--- a/.claude/skills/wave-scope/SKILL.md
+++ b/.claude/skills/wave-scope/SKILL.md
@@ -530,14 +530,15 @@ Do NOT delete the original body — copy it (or post the pre-update version as a
 
 If any step surfaced a process gap (stale memory, missing meta-issue, vague reference), include a `**Process gaps surfaced**` section so the next retro can address them.
 
-### 12.5. Validate implementer/reviewer names + child-repo membership (#319, #1134) — MANDATORY
+### 12.5. Validate implementer/reviewer names + child-repo membership (#319, #1134, #1180) — MANDATORY
 
-Validate every declared name against the relevant roster BEFORE `/wave-kickoff` fan-out. The validator runs **two independent checks**:
+Validate every declared name against the relevant roster BEFORE `/wave-kickoff` fan-out. The validator runs **three independent checks** — any one alone exits 1:
 
 | # | Check | Applies to | Failure |
 |---|---|---|---|
 | #319 | **Name resolution** — does the name exist at all? | every slot, but the candidate set differs by slot class (#1162 — see Resolution rules) | unresolved name + fuzzy suggestions |
-| #1134 | **Repo membership** — is the implementer on the TARGET repo's roster? | `implementer` on a child repo | cross-repo implementer with no recorded override |
+| #1134 | **Repo membership** — is the implementer on the TARGET repo's roster? | every non-review-class slot on a child repo | cross-repo implementer with no recorded override |
+| #1180 | **Slot classification** — is this slot key classified at all? | every slot key the row carries | `UNCLASSIFIED role slot(s)` — classify it in `validate_matrix_names.py` |
 
 **#319 (name resolution).** Pre-#319 a stale alias like "Anya Volkov" (canonical: "Anya Kowalczyk") would propagate through scope and only surface at first-spawn time — P3W7 retro recorded TWO such substitutions in `wave_7_decisions.implementer_substitutions`.
 
@@ -546,9 +547,11 @@ Validate every declared name against the relevant roster BEFORE `/wave-kickoff` 
 **Resolution rules:**
 - Name resolution — **the candidate set depends on the slot class (#1162):**
   - **Review-class slots** (`reviewer`, `reviewer_2`, `merge_gate_reviewer`): child-repo roster (`<repo>/.claude/team/roster/*.md`) UNION parent roster UNION the **org-union manifest** `.claude/team/roster.json`. The manifest is what makes a reviewer drawn from a **third child repo** valid — e.g. a `noorinalabs-data-acquisition` persona reviewing an `isnad-ingest-platform` story. **A third-child reviewer is a correct assignment, not drift: do not "fix" one by reassigning it.** Before #1162 the parent side was the parent's *card directory* only (9 names, not the manifest's 78), so such a reviewer resolved against neither roster and hard-blocked this step and `/wave-kickoff` § 0b.
-  - **`implementer`**: child-repo roster UNION parent roster only — **the manifest is deliberately excluded.** Including it would loosen the #1134 membership check in the `unverified` case below: when the target roster is unreadable, membership fails open, so a manifest-only implementer would resolve, be reported `unverified`, and pass. Keeping the set narrow means that slot only passes on a name the target repo (or the parent) can vouch for.
+  - **`implementer` — and every slot that is not review-class (#1180):** child-repo roster UNION parent roster only — **the manifest is deliberately excluded.** Including it would loosen the #1134 membership check in the `unverified` case below: when the target roster is unreadable, membership fails open, so a manifest-only implementer would resolve, be reported `unverified`, and pass. Keeping the set narrow means that slot only passes on a name the target repo (or the parent) can vouch for.
   - Both sets let org-level coordinators (Aino, Nadia, Wanjiku, Santiago) fill child-repo slots without duplicating roster entries.
-- Repo membership: the **target repo's roster only**, and **only for the `implementer` slot**. Reviewers keep the wider resolution set above and are never membership-gated — cross-team reviewers are explicitly permitted by `charter/agents/spawn-discipline.md` § Child-Repo Implementer Rule step 5. The implementer is the only role that must produce a commit in the target repo.
+  - **`REVIEW_CLASS_ROLES` is the allowlist, not `COMMIT_CAPABLE_ROLES` (#1180).** The wide set and the membership exemption are both justified only for a slot that *provably never commits*, so an unrecognised slot key defaults to the strict side and fails loudly. Pre-#1180 the seam ran the other way: a `co_implementer` slot silently inherited the manifest widening AND the membership exemption, and a matrix of `co_implementer` / `pair_implementer` / `fixer` reported "all 3 names resolved", exit 0.
+- Repo membership: the **target repo's roster only**, and **for every non-review-class slot** (`implementer` today). Review-class slots keep the wider resolution set above and are never membership-gated — cross-team reviewers are explicitly permitted by `charter/agents/spawn-discipline.md` § Child-Repo Implementer Rule step 5. The implementer is the only role that must produce a commit in the target repo.
+- **Slot classification (#1180).** Scope mode discovers slots from `KNOWN_ROLE_SLOTS` (= `COMMIT_CAPABLE_ROLES | REVIEW_CLASS_ROLES`) plus any agentive-shaped row key (`_ROLE_SLOT_KEY_RE`) not denied in `NON_ROLE_ROW_KEYS` — it no longer iterates a hardcoded 4-slot tuple that skipped an unrecognised key outright. An admitted-but-unclassified key is a **third failure class**: `UNCLASSIFIED role slot(s) (#1180)`, exit 1. **Adding a role slot to the scope schema is therefore a code change**, not just a data change — classify the key in one of the three frozensets in `validate_matrix_names.py`.
 - Parent entries (`noorinalabs-main` or empty repo key) → parent-only roster; membership is vacuous there and never fires.
 - Match is case-insensitive; trailing parenthetical role suffix (`Aino Virtanen (Standards & Quality Lead)`) is stripped before comparison.
 - On a name miss: fuzzy-match via difflib SequenceMatcher; surface the top-3 closest matches to the operator.
--- a/.claude/skills/wave-scope/tests/test_validate_matrix_names.py
+++ b/.claude/skills/wave-scope/tests/test_validate_matrix_names.py
@@ -15,14 +15,20 @@ import json
 import sys
 import tempfile
 import unittest
+import unittest.mock
 from pathlib import Path
 
 sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
 
+import validate_matrix_names  # noqa: E402
 from validate_matrix_names import (  # noqa: E402
+    COMMIT_CAPABLE_ROLES,
+    KNOWN_ROLE_SLOTS,
+    REVIEW_CLASS_ROLES,
     _load_org_manifest_names,
     _override_rationale,
     _print_report_to_stderr,
+    is_role_slot_key,
     repo_of_row,
     validate,
     validate_scope,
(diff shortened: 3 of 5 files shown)


parametrization:
## Correction to the original PR description — measurement, not code

Flagging this explicitly because a body edit is silent and the original number was wrong.

**What I got wrong.** The first version of this description reported the matrix-mode differential as:

| | result |
|---|---|
| `main` | `all 3 names resolved`, exit **0** |

I did not measure that. I inverted the table in #1180's body — where that row describes PR #1178's head `256f79a0`, not `main` — and presented it as my own measurement. The code and the tests were always measured; this one line was not.

**What it actually is.** Re-measured against today's `origin/main`, the issue's three-slot matrix reports **`2/3 names UNRESOLVED`, exit 1** — because `Annunaki` and `Steven French` are rejected by #1181's persona-shape filter. So `main` does exit 1 on that input, but **for reasons unrelated to this seam**, and the one slot the issue is actually about (`co_implementer: "Nikolaos Papadopoulos"`) resolves silently inside an otherwise-red run. The issue's own example masks the slot it was written about.

**The corrected measurement is worse for `main`, not better.** Isolating each loosening on `noorinalabs-user-service`:

| Case | `origin/main` | this branch |
|---|---|---|
| `{"co_implementer": "Nikolaos Papadopoulos"}` | `all 1 names resolved`, exit **0** | `1/1 UNRESOLVED` + `1/1 UNCLASSIFIED`, exit **1** |
| `{"co_implementer": "Nurul Hakim"}` | `all 1 names resolved`, exit **0** | `1/1 CROSS-REPO IMPLEMENTER`, exit **1** |

The second case is the one I would ask a reviewer to weigh: `Nurul Hakim` on `user-service` is **exactly** the W27/W28 cross-repo-implementer failure — the one whose wrap-time repair was a mechanical merge-commit re-attribution. `main` blocks it in the `implementer` slot and waves it through in a `co_implementer` slot. Renaming the key is the entire bypass.

**Fixed in `aee3d94`** — the module docstring repeated the same bad claim and now carries the two isolated measurements plus an explicit warning that the issue's example matrix masks the defect.

**Unaffected:** the mutation table (0/9 survived), the 21-wave / 293-row byte-identical regression sweep, and the W28/W29 anchors. All of those were re-run from scratch after this was found, from a clean namespaced working directory, with the baseline validator freshly extracted from `origin/main` — results identical to what is in the description.

**How it was caught:** the coordinator flagged that the session scratchpad is shared across agents and asked everyone to re-run unreported results from a namespaced directory. Re-running the differential from a clean directory is what surfaced the discrepancy. The root cause was mine and separate from the collision: I reported a number I had read rather than run.

parametrization:
## Merge-gate review (Opus) — Aino Virtanen, Standards & Quality Lead

Reviewed at head SHA **`aee3d9409778997298ef5b7852de37ddec828794`**, diffed against `origin/main` from origin (not a local checkout). CI rollup at that exact SHA: **21/21 SUCCESS**, `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.

I re-measured every load-bearing number independently rather than inheriting it. Nothing below is quoted from the PR body, the issue, or the author's mutation table.

---

### 1. The corrected `co_implementer` claim — independently confirmed

This is the strongest justification for the PR and it is the number that was corrected mid-review, so it got a from-scratch measurement. Both versions run from the org root so `_find_org_dir()` resolves the real rosters.

`{"noorinalabs-user-service": {"co_implementer": "Nurul Hakim"}}`

| Version | Output | Exit |
|---|---|---|
| `origin/main` | `all 1 names resolved across 1 entries.` | **0** |
| this branch | `1/1 CROSS-REPO IMPLEMENTER assignment(s) with no recorded override (#1134).`<br>`1/1 UNCLASSIFIED role slot(s) (#1180).` | **1** |

Confirmed exactly as corrected. Nurul Hakim holds a **parent** roster card (`.claude/team/roster/observability_engineer_nurul.md`) and is **not** on the user-service roster (`Anya Kowalczyk, Idris Yusuf, Mateo Salazar, Nadia Boukhari`), so pre-fix he resolves via the parent union and then the #1134 membership check is skipped purely because the key is spelled `co_implementer` instead of `implementer`. That is the W27/W28 cross-repo-implementer failure re-admitted by renaming a slot key, and it is real.

I also measured the *other* loosening separately, since the two are independent and the PR claims both:

`{"noorinalabs-user-service": {"co_implementer": "Nikolaos Papadopoulos"}}` — manifest-only persona, **no local roster card anywhere** in the org tree:

| Version | Output | Exit |
|---|---|---|
| `origin/main` | `all 1 names resolved across 1 entries.` | **0** |
| this branch | `1/1 names UNRESOLVED across 1 entries.` | **1** |

So both loosenings — the #1162 manifest widening and the #1134 membership exemption — were reachable through an unclassified key, and the inversion closes both. I did **not** use #1180's three-slot worked example as a baseline; per the note in the module docstring it is contaminated by #1181's persona-shape filter.

I did, however, verify the "all 3 names resolved, exit 0" sentence the PR adds to `wave-scope/SKILL.md` § 12.5, substituting three *real* personas so the persona filter is not the confound: `origin/main` returns `all 3 names resolved across 1 entries.`, exit 0. This branch returns `3/3 CROSS-REPO IMPLEMENTER` + `3/3 UNCLASSIFIED`, exit 1. The prose claim stands on its own measurement.

### 2. Mutation testing — 9 mutants, 0 survived

Harness runs the head-SHA module + head-SHA test file in a name-unique scratch tree outside the repo; the repo working tree was never modified. Each mutant restored by re-materialising the file from `git show <branch>:<path>`, never from a scratch backup. Baseline asserted green before and after: **62 passed, 38 subtests** both times.

| Mutant | Result |
|---|---|
| `candidates` ternary inverted back to the `COMMIT_CAPABLE_ROLES` allowlist | **killed** — `test_manifest_does_not_pass_implementer_on_uncloned_repo` |
| membership gate back to `role in COMMIT_CAPABLE_ROLES and not is_parent_repo` | **killed** — `test_unclassified_slot_fails_even_when_fully_valid` |
| `slot_class` hardcoded to `"known"` | **killed** — same |
| exit condition drops `or unclassified` | **killed** — same |
| `REVIEW_CLASS_ROLES` emptied | **killed** — `test_reviewer_slots_keep_parent_union` |
| **M5** hardcoded 4-tuple restored in `validate_scope` | **killed, 2 tests** — `test_unrecognised_slot_key_is_not_skipped` (`['implementer','reviewer'] != ['co_implementer','implementer','reviewer']`) and `test_scope_mode_slot_list_is_driven_by_the_frozensets` |
| **M6** `is_role_slot_key` → `return True` | **killed, 36 failed** — exactly the reported count; kills span `ScopeModeTests`, `SlotKeyRecognitionTests`, `ScopeModeSlotDiscoveryTests`, incl. `['id','ref','implementer'] != ['implementer']` |
| **M7** `NON_ROLE_ROW_KEYS` emptied | **killed, 3 tests** — `pre_kickoff_blocker is metadata, not a person slot`; `['implementer','pre_kickoff_blocker','reviewer'] != ['implementer','reviewer']` |
| **M8** agentive regex removed (KNOWN-only recognition) | **killed, 7 tests** — the six `unclassified_agentive_keys` subtests plus `test_unrecognised_slot_key_is_not_skipped` |

On M5/M6 specifically: **scope mode genuinely no longer skips an unrecognised slot key.** I confirmed the pre-fix behaviour end-to-end rather than only through the unit fixtures — a real 3-slot scope row (`implementer` / `reviewer` / `co_implementer`) through `--scope` on `origin/main` prints `all 2 names resolved across 1 entries.`, exit 0. The third slot was never looked at. This branch: `1/3 CROSS-REPO IMPLEMENTER` + `1/3 UNCLASSIFIED`, exit 1.

I also verified these assertions are not inert: `ci.yml` § *Run skill test suites (per-skill rootdir)* globs `.claude/skills/*/tests`, and `.pre-commit-config.yaml`'s `pytest-skills` pre-push hook mirrors it. `wave-scope/tests/` is inside that glob, so the new file is CI-gated, not decorative.

### 3. The denylist — probed, and it is not the shape this PR removes

The concern is fair on its face: `NON_ROLE_ROW_KEYS` is an enumeration, and an enumeration that governs a default is how the original defect was built. But the polarity is the opposite one. I built a scope fixture with two novel agentive metadata keys nobody listed, one with a non-name value and one whose value *is* a real persona:

- `escalation_owner: "program-director"` → `UNRESOLVED` **and** `UNCLASSIFIED`, exit 1
- `escalated_to_owner: "Nadia Khoury"` → `CROSS-REPO IMPLEMENTER` **and** `UNCLASSIFIED`, exit 1

An un-denylisted key defaults to **checked, on the strict path**, and the failure names the key and lists the three frozensets to classify it in. That is loud and self-documenting in both value shapes. The seam this PR removes defaulted to *unchecked and wide*. So the denylist inherits the fix's polarity rather than reintroducing the defect: it can only ever cause an over-fail, never a silent pass, and the recovery is one reviewable line.

### 4. `is_role_slot_key` — false positives and negatives, found independently

I enumerated both classes against the recogniser rather than reading the regex:

- **False positives** (metadata-shaped keys that would be treated as a person slot): `owner`, `order`, `merge_order`, `sort_order`, `wave_number`, `pr_number`, `blocker`, `error`, `trigger`, `marker`, `counter`, `folder`, `filter`, `layer`, `vendor`, `anchor`, `risk_indicator`, `header`, `after`, `other`, `prior`, `major`, `minor`, `escalation_owner`. All fail loud, per §3.
- **False negatives** (person-slot-shaped keys silently skipped in scope mode): `co_lead`, `pair`, `second`, `qa`, `qa_lead`, `assignee`, `deputy`, `buddy`, `scribe`, `steward`, `champion`, `delegate`, `shadow`, `mentee`, `pilot`, `chair`, `analyst`, `architect`, `sme`.

#1278 already documents both classes and names representative members of each (`co_lead`/`pair`/`second`/`backup`; `after`/`order`/`owner`), characterises the false-positive side as loud and recoverable, and records why the two alternatives were rejected. My census agrees with it in polarity and substantially in membership, so there is nothing to add and nothing to duplicate.

**On whether the residual reintroduces the defect under a new name: it does not, and I measured that rather than reasoning about it.** On the same probe fixture carrying three unknown keys, `origin/main` reports `all 6 names resolved across 3 entries.`, exit 0 — it skipped all three. This branch catches two of the three and exits 1; only the non-agentive `co_lead` slips through. That is a strict narrowing of the skip set, not a rename of it. And critically, the residual is **scope-mode-only** — in matrix mode `validate()` never consults `is_role_slot_key`, so `co_lead` is resolved narrow and reported `unclassified`, exit 1. #1278's recommendation to decide between "accept" and "nest the assignments object" at phase review rather than in-wave is the right call; there is no live instance forcing it.

### 5. Regression anchors — re-run, zero drift

Swept **every** wave in `cross-repo-status.json` (21 scope waves, 293 tier rows) through both versions and diffed the complete stdout+stderr, not just exit codes:

```
diff sweep_main.txt sweep_head.txt  ->  0 lines
```

**Byte-identical across all 21 waves.** Exit codes match wave-for-wave. Anchors: W28 = `2/30 CROSS-REPO IMPLEMENTER ... (#1134)`, exit 1; W29 = `all 165 names resolved across 51 entries.`, exit 0; W2 = `2/42 names UNRESOLVED`, exit 1 (pre-existing, identical on both sides). I treated before/after equality as the anchor, not the absolute count — W29's 165 reflects the post-#1180 amendments to `wave_29_scope`, so the 85 named in the issue is stale and is not evidence of anything.

Worth stating explicitly: the new failure class fires **nowhere** in the live corpus, which is *why* the sweep is byte-identical. The guard is proven live only by the synthetic probes in §3 — that is the correct outcome for a gate closing a latent seam, but it means the byte-identical sweep is evidence of no regression, not evidence the guard works. Those are separate claims and §2/§3 carry the second one.

### 6. Prose ⇄ code

Checked each edited sentence against the code rather than against the PR body.

- `wave-scope/SKILL.md` § 12.5, #1134 row now reads *"every non-review-class slot on a child repo"* — code is `if not is_review_class and not is_parent_repo:`. Exact.
- The `REVIEW_CLASS_ROLES`-is-the-allowlist bullet and the slot-classification bullet match `candidates = review_combined if is_review_class else combined` and the `slot_class` assignment. Exact.
- `charter/skills.md` § Scope — *"every slot that is not review-class"*, and *"adding a role slot is a code change"*. Matches; and all three previously said "the `implementer` slot only", which was the stale half of the prose→code gap.
- `wave-kickoff/SKILL.md` § 0b — the `RC != 0` message now names the third failure class, matching the third `return 1` path.

Two non-blocking notes, neither of which changes behaviour:

1. § 12.5's table cell for #1180 says the check applies to *"every slot key the row carries."* In scope mode that is slightly stronger than the code: a non-agentive key the row carries is not discovered at all (the #1278 false negative). The bullet three lines below states the discovery rule precisely, so an operator reading the section reaches the correct model — but the cell alone overstates. "every slot key the validator discovers" would close it.
2. Discovery normalises case and whitespace (`is_role_slot_key("  Implementer  ")` → `True`), while classification compares raw (`role in KNOWN_ROLE_SLOTS`). Measured: a row key `Reviewer` is admitted, then takes the strict path and reports *"slot key is in neither `COMMIT_CAPABLE_ROLES` nor `REVIEW_CLASS_ROLES`"* — misleading for what is only a casing variant. The asymmetry can only ever over-fail, never under-fail, so it is consistent with the PR's own thesis and I am not treating it as debt; it is a message-fidelity nit if the file is touched again.

### 7. On the corrected number

The fabricated baseline was caught by the author, corrected in `aee3d94`, and flagged in a visible comment before I looked at it. I re-derived the replacement claim from scratch and it holds. I want to record the more useful finding, though: #1180's own worked example **masks the defect the issue was filed about** — that matrix exits 1 on `main` for a #1181 persona-shape reason entirely unrelated to this seam, so anyone who ran it would have seen a red gate and concluded the seam was already closed. Isolating the slot was the necessary step, and the module docstring now warns the next reader about exactly this. That warning is the most valuable line in the diff.

---

**Verdict: approve.** The seam inversion is correct and both loosenings it closed are measurably real. The guards are genuinely load-bearing — 9/9 mutants killed, including the four the brief singled out — and CI actually executes them. Zero drift across the full 293-row live corpus. The residual is narrower than the pre-fix behaviour, is scope-mode-only, and is honestly scoped in #1278 with the rejected alternatives recorded. No blocking findings; no new debt to file beyond what #1278 already tracks.

---

Requestor: Aino Virtanen
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Second review — Lucas Ferreira (SRE Engineer)

Verified at head `aee3d9409778997298ef5b7852de37ddec828794`. CI 21/21 SUCCESS at that SHA. Re-ran `.claude/skills/wave-scope/tests/test_validate_matrix_names.py` at that commit: 62 passed, 38 subtests passed — matches the PR description exactly.

I extended Aino's pass rather than repeating it (her 9 mutations / 0 survivors, the byte-identical 21-wave/293-row sweep, and the `co_implementer`/Nurul Hakim differential are re-confirmed, not re-derived). Weighing the three things flagged as mine to decide:

**1. Do the synthetic probes suffice, given the live corpus can't exercise the new path?** Yes. All four live slots are classified today, so the regression sweep can only prove non-regression on the *old* branches — it structurally cannot exercise `slot_class="unclassified"` or the inverted allowlist. That gap is filled by the synthetic probes (isolated `co_implementer` cases) plus the mutation run: 9/9 killed, including the one test (`test_unclassified_slot_fails_even_when_fully_valid`) that isolates the #1180 class from a name-miss or membership-miss as an independent signal. Between a fully-resolving, repo-member name still failing on `slot_class` alone, and a mutation suite that kills every deliberate weakening, I'm satisfied the guard does what it claims — the corpus's silence is expected, not a coverage hole.

**2. False-positive blast radius.** Real, but bounded — and I found a sharper case than the two probed keys. `is_role_slot_key` deliberately admits any agentive-shaped key (the PR's own `test_unclassified_agentive_keys_are_recognised` asserts `author` and `owner` must be *seen*, not skipped), so plausible future metadata fields — `risk_owner`, `exec_sponsor`, `code_owner`, `bug_reporter`, `incident_commander` — all trip it (confirmed against the actual regex, not just the two keys in the brief).

Worse: if such a key's value happens to be a real parent/org-roster name that isn't on the *target repo's* roster (a TPM/coordinator name is exactly what a "who owns this" field would hold), the run emits `#1134 CROSS-REPO IMPLEMENTER assignment(s)` *alongside* `#1180 UNCLASSIFIED` — the #1134 message reads as a genuine reassignment/onboarding defect (`pick (a) reassign / (b) onboard / (c) override`) for a field that was never an implementer slot. I measured this: `{"risk_owner": "Santiago Ferreira"}` on a `noorinalabs-user-service` row produces both messages simultaneously.

Operationally: bounded to the one wave being validated (`--wave {M}` is per-wave, not whole-file), never silent (fail-safe, matches the design intent the PR's own tests already accept as a tradeoff), and the correct fix — add the key to `NON_ROLE_ROW_KEYS`, one line — is named precisely in the `#1180` message itself. The `#1134` message riding along is noise, not a wrong verdict, but it's noise shaped exactly like a real onboarding fire mid-kickoff. Filed **#1289** (not a duplicate of #1278, which is about non-agentive keys being *skipped* — this is about an admitted agentive key's messaging compounding) with a candidate one-line fix (skip the membership check when `slot_class == "unclassified"`, since membership shouldn't be evaluated before a slot is known to be commit-capable at all). Not a blocker: it's a messaging refinement on an already-accepted, already-tested design tradeoff, not a new defect in this PR's logic.

**3. Case asymmetry (`Reviewer` → "in neither... " message).** Confirmed and I traced it further: a mis-cased key is *admitted* by discovery (normalizes case/whitespace) but then evaluated on the strict path in `validate()` (raw comparison), so it can also co-emit the same misleading #1134 message as above. Still fail-safe — never a silent pass — so I agree with Aino it isn't debt worth its own ticket; it's the same underlying messaging gap as #1289, not a second one.

## Verdict

Approve. The core fix is correct, well-isolated by tests, and the mutation testing gives real confidence given the live corpus can't validate the new path directly. The two residuals I dug into are operational rough edges on an already-acknowledged tradeoff (loud-by-design over silent-by-default), not defects — bounded blast radius, self-documenting failure, one-line remediation named in the message itself. Filed #1289 for the messaging compounding; #1278 already covers the other residual.

---

Requestor: Lucas Ferreira
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: none

PR #1280, opened 2026-08-03
fix(commit-identity): heredoc written to a file, executed later in the same command, now blocks (#1167)

### LINKED ISSUE #1167: tech-debt(commit-identity): heredoc written to a file and executed later in the same command bypasses the gate
## Shape

A heredoc body written to a file and executed later **in the same Bash command** reaches a shell, but the commit-identity gate allows it. Measured through the real `check()` at both `fea0dca` (main) and `e163320` (PR #1155 head) — **ALLOW at both**, so this is pre-existing and NOT a regression of #1155:

```
cat > /tmp/s.txt <<'DELIM'
git -c user.name="X" -c user.email="Y" <the-commit-verb> -m z
DELIM
bash /tmp/s.txt
```

The semicolon variant `cat <<'DELIM' > /tmp/x; bash /tmp/x` behaves identically.

## Why every matcher misses it

- `_HEREDOC_RE` requires an interpreter textually before the `<<` — absent here.
- `_shell_parse.classify_heredocs` (added in #1155) correctly resolves the owner to `cat`, a data sink, and follows only `|` downstream. A `;`/newline-separated later command is not in the pipeline, so the body is classified DATA and stripped.
- `_SCRIPT_INVOKE_RE` **does** match `bash /tmp/s.txt` and calls `_read_script_if_safe` — but the hook fires *PreToolUse*, so the file does not exist yet. The read fails, `content` is falsy, no block.

That last point is the structural part: the script-content inspection is fundamentally blind to a file the same command is about to create.

## Note on the apparent base-block

`cat > /tmp/s.sh <<'DELIM' ... DELIM` + `bash /tmp/s.sh` DOES block at `fea0dca` — but only incidentally: `_HEREDOC_RE`'s `\b(?:bash|sh|...)\b` matches the `sh` inside the filename `s.sh` (the `.` is a word boundary). Rename the file to `s.txt` and the base allows it too. The `.sh`-extension restriction was already removed as a bypass surface in #482; this is the same class of accident.

## Suggested direction

Two candidate fixes, neither obviously cheap:

1. In `_opener_feeds_interpreter`, when the owning segment redirects into a path, record that path; if any later segment in the same command invokes an interpreter on it, classify the body as CODE.
2. In `_read_script_if_safe`, when the target path is unreadable, fall back to scanning any heredoc body in the same command that is redirected to that path.

Option 1 is more principled and stays inside the classifier #1155 introduced.

## Provenance

Found by the merge-gate review of PR #1155 (Aino Virtanen, Standards & Quality Lead), while constructing bypass shapes outside the PR's own 49-shape corpus. Belongs to the #1150 umbrella as a further under-matching instance.

parametrization:
## Summary

Closes #1167 — sibling of the **#1150 under-matching umbrella**, found during the PR #1155 merge-gate review (Aino Virtanen).

A heredoc body written to a file and **executed later in the same Bash command** reached a shell but the commit-identity gate allowed it:

```
cat > /tmp/s.txt <<'DELIM'
DELIM
bash /tmp/s.txt
```

Measured ALLOW at both `fea0dca` (main) and `e163320` (PR #1155 head) — pre-existing, not a #1155 regression. Every existing matcher missed it for a distinct reason (full account in the issue and in the new `_shell_parse.py` block comment above `_segment_write_targets`):

- `_HEREDOC_RE` requires an interpreter textually before `<<` — absent.
- `classify_heredocs` resolved the owner to `cat` (a data sink) and only followed `|` downstream — a `;`/newline-separated later command isn't in the pipeline.
- `_SCRIPT_INVOKE_RE` *does* match `bash /tmp/s.txt` and calls `_read_script_if_safe`, but the hook fires PreToolUse, before the file exists — the read fails and nothing is inspected.

## Fix — option 1 from the issue, stays inside the classifier

When a heredoc's owning segment writes the body to a file (`cat > FILE` / `>>` / `tee FILE` positional arg), and **any segment elsewhere in the command** later invokes a `SHELL_INTERPRETERS` member on that same file, the body is reclassified as CODE. The decision is centralized in one new `_is_opener_code`, called by **both** `classify_heredocs` and `strip_data_heredocs` — main#1152 already flagged the hazard of two independent "is this body code" definitions drifting apart; #1167 is exactly that hazard, since `strip_data_heredocs` is what actually feeds `validate_commit_identity`'s `scanned` text. Fixing only one call site would have left the gap open in the other (mutation-tested below).

**Deliberately not built on `iter_interpreter_invocations`.** That helper's `strip_heredocs` call erases same-line trailing text after a heredoc opener — measured directly against `cat <<'DELIM' > /tmp/x; bash /tmp/x\n...`, it strips `; bash /tmp/x` right along with the heredoc syntax. That's exactly the issue's stated semicolon-equivalent shape, so a target scan built on that helper would silently miss it. A new `_iter_non_heredoc_segments` walk reuses this module's own per-line terminator search instead, which has no such erasure bug (mutation-tested below — swapping in `iter_interpreter_invocations` breaks exactly `test_semicolon_variant`).

## Shapes covered

- Newline variant (the issue's literal repro) and the **semicolon variant** stated as equivalent in the issue
- `>>` append, not just `>`
- `tee FILE` as a sink (positional target, not a `>` redirect)
- Every `SHELL_INTERPRETERS` member (`bash`, `sh`, `zsh`, `dash`, `ksh`)
- `env bash FILE` (already correlates for free — `parse_interpreter_invocation` already strips the `env` wrapper)
- Absolute interpreter paths (`/bin/bash FILE`)
- `./FILE` vs bare `FILE` spelling (`os.path.normpath`)
- Positionally-agnostic ordering: the invocation may appear *before* the write in the command text — deliberate, since main#1152's own rule is "every ambiguity resolves toward CODE," and this is a security gate, not a scheduler
- Identical *literal* shell-variable tokens on both sides (`cat > "$F" ...; bash "$F"`) — this is literal-token equality, **not** variable resolution (see scope note below)

## Deliberately NOT covered (stated, not silent)

- **`source FILE` / `. FILE`** — shell builtins, not members of `SHELL_INTERPRETERS`. Reading a file into the *current* shell is a structurally different mechanism than spawning a new interpreter on it, and folding them in would widen `SHELL_INTERPRETERS` itself — a set several *other* matchers in this module and in `validate_commit_identity` key on. That's a broader decision than this fix; filing a follow-up issue.
- **Unexpanded shell-variable redirect targets** — not resolved by evaluation (same documented punt this module already makes for `eval "$cmd"`). The literal-token case above is a side effect of plain string comparison, not a variable-resolution feature — a rename between the write and invoke positions (`cat > "$OUT" ...; F=$OUT; bash "$F"`) defeats it exactly as true variable resolution in general would.
- **`bash < FILE`** (stdin-redirect execution) — a different execution mechanism than the positional `InterpreterInvocation.operands` this correlation reads; not resolved by the pre-existing shape-7 walker either.
- **Unspaced redirect form** (`cat>/tmp/x`, no surrounding whitespace) — `_segment_write_targets` relies on shlex already splitting `>`/`>>` into their own token, which only happens when whitespace separates them. Every shape in the issue, and every heredoc redirect in this codebase's own conventions, uses the spaced form.

## False-positive verification

`WriteThenExecFalsePositiveTests` in the new test file pins the not-over-blocking direction directly through `check()`:
- A plain `cat > README.md <<'EOF' ... EOF` with no later invocation anywhere → still allowed (the canonical legitimate shape named in the spawn brief).
- A write followed by an unrelated command, or by a *different* file's invocation → still allowed (correlation is per-path, not "any interpreter invocation anywhere flips everything to code").
- A multi-heredoc command where only ONE of two written files is later invoked → only that span flips to code; the other stays data.

Additionally, the full pre-existing `test_heredoc_target_scope.py` (101 tests/subtests, main#1152's own false-positive suite) still passes unchanged, and the full hooks+lib suite is green (2456 + 1087 passed).

## Mutation table

Each row: weaken/delete the named branch in `_shell_parse.py`, then run `test_heredoc_write_then_exec.py`, and confirm a specific named test fails (not just "something fails").

| # | Mutation | Named test(s) that fail | Result |
|---|---|---|---|
| 1 | `_is_opener_code` — remove the correlation entirely, fall back to `_opener_feeds_interpreter` only | `test_literal_issue_repro_newline_variant` + 15 others | 16 failed, 7 passed |
| 2 | `_segment_write_targets` — drop `>>` handling, keep only bare `>` | `test_append_redirect`, `test_append_and_plain_redirect_both_correlate` | 2 failed, 21 passed |
| 3 | `_segment_write_targets` — drop the `tee` positional-argument branch | `test_tee_sink`, `test_tee_sink_with_append_flag` | 2 failed, 21 passed |
| 4 | `_normalize_path_for_compare` — drop `os.path.normpath` (no normalization) | `test_dot_slash_prefix_normalization` | 1 failed, 22 passed |
| 5 | `_script_invocation_targets` — swap `_iter_non_heredoc_segments` for `iter_interpreter_invocations` (the rejected design) | `test_semicolon_variant` | 1 failed, 22 passed |
| 6a | `_classify_heredocs_cached` call site reverted to `_opener_feeds_interpreter` (leave `strip_data_heredocs` fixed) | `test_literal_issue_repro_newline_variant` + 14 others (including `test_classifier_marks_the_span_code`) | 15 failed, 8 passed |
| 6b | `strip_data_heredocs` call site reverted to `_opener_feeds_interpreter` (leave `classify_heredocs` fixed) | `test_literal_issue_repro_newline_variant` + 12 others (including `test_strip_data_heredocs_keeps_the_body`) — `test_classifier_marks_the_span_code` stays green, correctly localizing to the reverted site | 13 failed, 10 passed |

Rows 6a/6b are the empirical demonstration of the main#1152 drift hazard this PR explicitly guards against by centralizing the decision in `_is_opener_code`.

All mutations were reverted before the final commit; `git diff` against the committed state is clean.

## Verification run

- `.claude/hooks/tests/test_heredoc_write_then_exec.py` (new, 23 tests) — all pass, through the real `check()` entry point (not a private helper), matching how the issue measured the base ALLOW.
- `.claude/hooks/tests/test_heredoc_target_scope.py` (main#1152's own suite) — 101 passed / 7 subtests, unchanged.
- `.claude/hooks/tests/` full suite — 2456 passed, 628 subtests.
- `.claude/lib/tests/` full suite — 1087 passed, 19 subtests.
- `ruff format --check`, `ruff check`, `mypy .claude/hooks/ .claude/lib/ --ignore-missing-imports` — all clean.
- Both `pre-commit` stages (commit + `--hook-stage pre-push`) green on the changed files, and the push itself ran the full pre-push gate set (mypy, pytest, memory-budget, headcount-budget, doc-freshness, office-drift, mermaid-render) green.

## Follow-up

Filing a separate tech-debt issue (labels `tech-debt`/`process`/`phase-10`, no wave label) for the `source`/`.` builtin-invocation gap named above.

## Test plan

- [x] New regression suite passes through the real `check()` entry point
- [x] Full hooks + lib suites green
- [x] Mutation table above, real output
- [x] False-positive direction verified (existing #1152 suite + new dedicated tests)
- [x] Both pre-commit stages green, no `--no-verify`

### DIFF
--- a/.claude/hooks/_shell_parse.py
+++ b/.claude/hooks/_shell_parse.py
@@ -748,6 +748,254 @@ def _opener_feeds_interpreter(line: str, pos: int, scan: _LineScan) -> bool:
     return False
 
 
+# ---------------------------------------------------------------------------
+# Cross-segment "write-then-exec" correlation (main#1167)
+# ---------------------------------------------------------------------------
+#
+# `_opener_feeds_interpreter` answers "is this heredoc's body consumed by an
+# interpreter reachable through a PIPE from the same physical line?" That is
+# the wrong question for a heredoc whose owning segment redirects the body
+# into a FILE (`cat > FILE <<'DELIM'`, or `tee FILE <<'DELIM'`) and a LATER
+# segment — connected by `;`, a newline, or anything else, not a pipe —
+# invokes an interpreter on that same file:
+#
+#     cat > /tmp/s.txt <<'DELIM'
+#     git -c user.name=X -c user.email=Y commit -m z
+#     DELIM
+#     bash /tmp/s.txt
+#
+# `cat` is a recognised data sink and nothing downstream of it on the SAME
+# line is a `|`-connected interpreter, so `_opener_feeds_interpreter` says
+# DATA — correctly, for the question it answers. But the body reaches a real
+# shell one command later. `_read_script_if_safe` (the OTHER existing defense,
+# in validate_commit_identity) is structurally blind to this too: the hook
+# fires PreToolUse, before `cat` has created the file, so reading it back
+# fails and no content is inspected either — see the #1167 issue body for the
+# full "why every matcher misses it" account. Belongs to the #1150 umbrella.
+#
+# The fix stays inside the classifier (the issue's preferred option 1):
+# record the file path(s) a heredoc's owning segment writes the body to, and
+# if ANY segment anywhere else in the same command later invokes a
+# `SHELL_INTERPRETERS` member with that same path as its script operand,
+# reclassify the body as CODE.
+#
+# Deliberately positionally-agnostic: a `bash /tmp/x` segment appearing
+# BEFORE `cat > /tmp/x <<...` in the command still counts. A real shell would
+# only execute meaningful content if the write happens first, but this is a
+# security gate — main#1152's own rule ("every ambiguity resolves toward
+# CODE") already commits this module to erring toward over-detection rather
+# than depending on textual ordering to save a rare false positive.
+#
+# Deliberately NOT resolved through `iter_interpreter_invocations`: that
+# helper's `strip_heredocs` call (the coarse, single-regex eraser — NOT the
+# per-line `classify_heredocs`/`strip_data_heredocs` machinery) consumes
+# everything from a heredoc opener through to its terminator in ONE pass,
+# INCLUDING any command chained onto the OPENER's own line with `;` before
+# the heredoc's first newline (`cat <<'DELIM' > /tmp/x; bash /tmp/x`, the
+# issue's stated semicolon-equivalent shape). Measured:
+# `_HEREDOC_RE.sub("", "cat <<'DELIM' > /tmp/x; bash /tmp/x\\n...")` erases
+# "; bash /tmp/x" along with the heredoc syntax, so a target scan built on
+# `iter_interpreter_invocations` would silently miss the semicolon variant
+# even though `_opener_write_targets`/`_scan_command_line` (the mechanism the
+# rest of this module already trusts) see it fine. This module's own per-line
+# scan is reused instead — see `_iter_non_heredoc_segments` — which has no
+# such erasure bug: it only ever skips a heredoc's OWN body lines, never
+# trailing text on the opener's own line.
+#
+# Explicit scope boundary — NOT covered, by deliberate decision, not oversight:
+#   - `source FILE` / `. FILE`: shell BUILTINS, not members of
+#     `SHELL_INTERPRETERS`. Reading a file into the CURRENT shell is a
+#     structurally different mechanism from spawning a new interpreter
+#     process on it, and folding them in would widen `SHELL_INTERPRETERS`
+#     itself — a set several OTHER matchers in this module and in
+#     validate_commit_identity key on. That is a broader decision than this
+#     fix and belongs in its own issue.
+#   - `env bash FILE`: no new work needed here — `parse_interpreter_invocation`
+#     already strips the `env` wrapper via `strip_command_prefixes` before
+#     resolving the interpreter, so this shape already correlates correctly
+#     through the same path as a bare `bash FILE`.
+#   - A redirect target given via an unexpanded shell variable is not
+#     RESOLVED — the hook cannot evaluate `$F` without running the shell, the
+#     same documented punt this module already makes for `eval "$cmd"`. Note
+#     the caveat: comparison here is by LITERAL TOKEN, not by value, so
+#     `cat > "$F" <<'DELIM' ...; bash "$F"` is still caught — not because `$F`
+#     was resolved, but because both sides spell the identical literal token
+#     `$F`. A rename between the two positions (`cat > "$OUT" ...; F=$OUT;
+#     bash "$F"`) or any other divergent spelling defeats it exactly as
+#     variable resolution in general would.
+#   - `bash < FILE` (script fed via stdin redirect rather than a positional
+#     argument) is a different execution mechanism than the positional
+#     `InterpreterInvocation.operands` this correlation reads, and is not
+#     resolved by the pre-existing shape-7 walker either — out of scope here.
+#   - The attached-operator redirect form with no surrounding whitespace
+#     (`cat>/tmp/x`, as opposed to `cat > /tmp/x`) is not recognised by
+#     `_segment_write_targets`'s tokenizer, which relies on shlex already
+#     splitting `>`/`>>` into their own token — shlex only does that when
+#     whitespace separates them. Every shape in the issue, and every heredoc
+#     redirect in this codebase's own conventions, uses the spaced form.
+
+# `tee`'s own value-less options. `tee` has no value-taking flags besides the
+# attached-`=` long form `--output-error[=MODE]`, which never starts a new
+# bare token and needs no entry here.
+_TEE_BOOL_FLAGS = frozenset({"-a", "--append", "-i", "--ignore-interrupts", "-p", "--output-error"})
+
+
+def _segment_write_targets(seg_text: str) -> list[str]:
+    """File paths the DATA-SINK head of `seg_text` could write a heredoc body to.
+
+    Two independent sources, both checked unconditionally:
+      - a `>`/`>>` shell redirect (any command may carry one);
+      - if the head command is `tee`, every non-flag positional argument
+        (`tee` writes to each of its own file arguments, with or without an
+        ADDITIONAL `>`/`>>` redirect).
+
+    `cat` never contributes a positional target: `cat FILE` READS FILE, it
+    does not write to it — only `cat`'s redirect (if any) is a write.
+    """
+    tokens = tokenize(_HEREDOC_OPENER_RE.sub(" ", seg_text))
+    if not tokens:
+        return []
+    head = tokens[0].rsplit("/", 1)[-1]
+    is_tee = head == "tee"
+    targets: list[str] = []
+    i = 1
+    n = len(tokens)
+    while i < n:
+        tok = tokens[i]
+        if tok in (">", ">>"):
+            if i + 1 < n:
+                targets.append(tokens[i + 1])
+                i += 2
+                continue
+            i += 1
+            continue
+        if tok.startswith(">>") and len(tok) > 2:
+            targets.append(tok[2:])
+            i += 1
+            continue
+        if tok.startswith(">") and len(tok) > 1 and tok[1] not in ("(", "&", ">"):
+            targets.append(tok[1:])
+            i += 1
+            continue
+        if is_tee and tok not in _TEE_BOOL_FLAGS and not _looks_like_flag(tok):
+            targets.append(tok)
+        i += 1
+    return targets
+
+
+def _opener_write_targets(line: str, pos: int, scan: _LineScan) -> list[str]:
+    """`_segment_write_targets` for the segment that OWNS the opener at `pos`."""
+    idx = next(
+        (k for k, (start, end, _sep) in enumerate(scan.segments) if start <= pos < end),
+        None,
+    )
+    if idx is None:
+        return []
+    start, end, _sep = scan.segments[idx]
+    return _segment_write_targets(line[start:end])
+
+
+def _normalize_path_for_compare(path: str) -> str:
+    """Loose normalization so `./x` and `x` — or a doubled slash — written two
+    ways still compare equal. Deliberately narrow: `os.path.normpath` collapses
+    redundant `.`/`//` segments only; it does NOT resolve against a cwd (the
+    write and the invocation share the same shell cwd, so this is the right
+    amount of normalization — resolving further would require knowing that
+    cwd, which this module deliberately does not read) and does NOT expand a
+    leading `~` (harmless either way here: `~/x` written identically on both
+    sides already compares equal as a plain string, and there is no realistic
+    same-command shape where the two sides spell a home-relative path
+    differently — expanding `~` would add a branch with no distinguishing
+    test, so it is left out).
+    """
+    try:
+        return os.path.normpath(path)
+    except (TypeError, ValueError):
+        return path
+
+
+def _iter_non_heredoc_segments(cmd: str) -> Iterator[str]:
+    """Every pipeline segment in `cmd` that is NOT inside a heredoc body.
+
+    Reuses the exact line-walk + terminator-search `_classify_heredocs_cached`
+    and `strip_data_heredocs` already use, so a heredoc BODY line is never
+    mistaken for a segment of the surrounding command (a body line containing
+    the word `bash` must not itself look like an interpreter invocation).
+    Deliberately NOT built on `iter_interpreter_invocations`/`strip_heredocs`
+    — see the module comment above `_segment_write_targets` for the measured
+    erasure bug that makes that helper unsuitable here.
+    """
+    lines = cmd.split("\n")
+    n = len(lines)
+    i = 0
+    while i < n:
+        line = lines[i]
+        i += 1
+        scan = _scan_command_line(line)
+        for start, end, _sep in scan.segments:
+            yield line[start:end]
+        for _pos, dash_form, delim in scan.openers:
+            j = i
+            while j < n:
+                candidate = lines[j].rstrip("\r")
+                if (candidate.lstrip("\t") if dash_form else candidate) == delim:
+                    i = j + 1
+                    break
+                j += 1
+            else:
+                i = n  # unterminated: nothing after this belongs to real command text
+
+
+@lru_cache(maxsize=256)
+def _script_invocation_targets(cmd: str) -> frozenset[str]:
+    """Normalized script-path operands of every interpreter invocation in `cmd`.
+
+    Used to correlate a heredoc's file-redirect target against a later (or
+    earlier — see the positional-agnostic note above) interpreter invocation
+    of that same file (main#1167). Memoized: pure in `cmd`, and both
+    `_classify_heredocs_cached` and `strip_data_heredocs` call it on the same
+    `cmd` within one hook invocation.
+    """
+    if not any(name in cmd for name in SHELL_INTERPRETERS):
+        return frozenset()
+    targets: set[str] = set()
+    for seg_text in _iter_non_heredoc_segments(cmd):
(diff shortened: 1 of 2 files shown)


parametrization:
## Merge-gate review (Opus) — PR #1280, head `193b87b`

Reviewed against `origin/main` at head SHA `193b87bf7f11463c681748d7b276e40bc553af30`, with file contents pulled from `contents/<path>?ref=<sha>` rather than a local checkout. Runtime work done in a detached worktree pinned to that SHA.

**CI at that exact SHA: 14/14 green** (`check-runs` grouped by conclusion → `{"success": 14}`). Full suite re-run locally at the same SHA: **3543 passed, 647 subtests passed**. `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.

**Verdict: approved.** The fix is correct, the mechanism is the right one, and — the part I care most about — it is wired into **both** call sites, which I verified by mutation rather than by reading. Three residual gaps found outside the PR's corpus are filed as issues; none of them is a regression, all are ALLOW on `main` too.

---

## Is the fix real?

Yes. I reproduced the mutation evidence independently rather than taking the report's table on faith, reverting each site in a scratch worktree and restoring with `git checkout HEAD -- <path>` each time.

| Mutation | Result | Read |
|---|---|---|
| Whole correlation made inert (`_is_opener_code` reduced to the pipe logic) | **16 failed** | the new tests are not decorative |
| Revert call site A only (`_classify_heredocs_cached`) | **15 failed** | matches the reported 15 |
| Revert call site B only (`strip_data_heredocs`) | **13 failed** | matches the reported 13 |
| Drop `>>` append handling | 2 failed | pinned |
| Drop the `tee` positional branch | 2 failed | pinned |
| Path normalization reduced to identity | 1 failed | pinned |
| Drop the `has_command_string` guard | **0 failed** | survives — see below |
| Only the first redirect target recorded | **0 failed** | survives, near-neutral |
| Heredoc-body lines no longer skipped in the segment walk | **0 failed** | survives — see below |

The 6a/6b pair is the important row and it holds up. The two sites genuinely cannot drift now: reverting either one alone leaves a double-digit failure count, and the mechanism is why — site B controls whether the commit text survives into `scanned`, and the `"commit" in scanned` guard means site A's classifier loop never even runs if site B strips first. Extending only one would have been a silent no-op. That is exactly the #1152 hazard, closed properly.

### One surviving mutation worth a test

`_iter_non_heredoc_segments`'s heredoc-body skip is **behaviourally load-bearing but untested**. Replacing the terminator-walk with `pass` leaves the entire suite green, yet it flips a real verdict:

    a notes file written to /tmp/n.md whose body's first line is literally `bash /tmp/n.md`
    and whose second line is an identity-flagged commit

At head: `is_code=False`, ALLOW. With the skip removed: `is_code=True`, **BLOCK** — a false positive on documentation. The helper's docstring states this exact rationale ("a body line containing the word `bash` must not itself look like an interpreter invocation") and nothing pins it. Not blocking — the behaviour is correct today — but it is the one branch in the new code a future refactor could delete without a red test. Worth a follow-up assertion whenever this file is next touched.

The other two survivors are low-value: the `has_command_string` guard is correct but only reachable via contrived input, and "first redirect target only" is close to semantically neutral since a real shell honours the last redirect anyway.

---

## Independent bypass corpus

I built 33 shapes outside the PR's tests, ran each through the real `check()` at `193b87b`, and then **executed every ALLOW under a real `bash`** with a harmless payload to confirm it was a genuine bypass rather than a parser artifact. Nine of them printed the payload.

Held (blocked) — including several the report does not claim:

`exec bash FILE` · `( bash FILE )` subshell · `{ bash FILE; }` group · `dd of=FILE` · unquoted delimiter · the tab-stripping dash-form opener · `/tmp/../tmp/s.txt` and `/tmp/./s.txt` and `/tmp//s.txt` path spellings · `bash -- FILE` · `bash -x FILE` · `sh -x FILE` · `nohup bash FILE` · `timeout 30 bash FILE` · `ksh FILE` · `tee A B` with only B invoked · `&&` on the opener line · a write nested inside `$( )`.

Two of those deserve a note because the report is more pessimistic than reality: the fused `cat>FILE` form and `>| FILE` are both listed or implied as unrecognised, and both **do** block — but by a different branch (the fused token makes the segment head resolve to something that is not in `HEREDOC_DATA_SINKS`; the `>|` case gets split on its `|` by the line scanner). Safe outcome, accidental route. `>|` extracts the literal target `'|'`, so the write-target extractor itself is wrong there even though the verdict is right.

### What did not hold

| Shape | Executes? | Disposition |
|---|---|---|
| `cat 1> FILE`, `cat 1>FILE`, `cat 1>> FILE` | yes | **#1286** |
| `cat &> FILE`, `cat >& FILE` | yes | **#1286** |
| body piped to `tee FILE` / to `cat > FILE`, then executed | yes | **#1286** |
| `bash < FILE` | yes | **#1287** |
| `bash $(echo FILE)` | yes | **#1287** |
| `cp A B; bash B` | yes | **#1287** |
| `echo FILE \| xargs bash`, `find -exec bash {}` | yes | evidence added to **#1168** |
| `source FILE`, `. FILE` | yes | **#1281**, already filed by the author |

All eight are ALLOW on `main` as well, so this PR regresses nothing — it closes the shape it was filed for and leaves these where it found them, which is how #1168 was handled against #1155.

**#1286 is the one I'd prioritise.** The other rows are broader mechanisms (a relay, a runtime-resolved operand, a copy edge) that reasonably sit outside one fix's blast radius. But `1>` is `>` with the default fd spelled out — the same operator, one character longer, and the write-target extractor sees nothing at all. `&>` and `>&` likewise. And the pipe-to-sink write is the same file reached one segment over. Those four live *inside the frame of the mechanism this PR adds*, and unlike `source`, the variable-rename punt, `bash < FILE` and the fused redirect form — all of which the module comment names explicitly and honestly — nothing named them, so nothing tracked them. The documentation discipline in this PR is genuinely good; that is precisely why the unnamed neighbours stand out.

None of this is a reason to hold the merge. A gate that blocks the filed shape plus sixteen neighbours is strictly better than one that blocks none, and the #1150 umbrella is explicitly an incremental track.

---

## Over-blocking

This hook gates every commit in the org, so I probed the other direction hard. Ten legitimate shapes, all still ALLOW at head:

a plain `README.md` write with no later interpreter · a write followed by an unrelated command · a write followed by an interpreter invocation on a *different* file · a documentation body that literally mentions running a commit · a script written and then actually executed, with no commit in it · multi-target `tee` with nothing invoked · a heredoc written to a file *named* `bash` · a `gh issue create` body payload.

Plus the PR's own 7 false-positive tests and main#1152's `test_heredoc_target_scope.py` corpus, both green in the full-suite run.

One behaviour worth stating plainly so nobody discovers it as a surprise: a script that is written and then executed **is** blocked when its body carries a commit, *even if the identity flags in it are a valid roster identity*. That is not new and not this PR's doing — `_detect_indirect_commit` returning any shape is an unconditional block, identically to how `bash -c '<valid commit>'` has been blocked since #475. The correct workflow is unchanged: run the commit directly. Consistent, not a new false-positive class.

---

## The block you hit on your own first commit

Confirmed pre-existing and confirmed **not widened** by this change. I measured the same input against `origin/main` and against `193b87b`:

    _scan_command_line(<continuation line>).openers -> [(10, False, 'DELIM')]   # identical at both
    classify_heredocs(...) -> 1 span, is_code=True, terminated=False            # identical at both

Root cause is that the scanner is quote-aware *within* a physical line but resets state at every line boundary, so continuation lines of a multi-line quoted argument are scanned as bare command text. Your change routes through `_is_opener_code`, which short-circuits on `_opener_feeds_interpreter` — already `True` for a phantom opener — so the new correlation is never reached and cannot have widened it.

It is worth filing, and I filed it as **#1288**, because there is a second direction nobody had characterised: a phantom opener whose head resolves to a data sink causes `strip_data_heredocs` to erase *real* command text from `scanned`. I built that shape and confirmed the commit text does disappear from the scanned string — the hook still blocks, but only because the primary identity path reads the raw command through bashlex and is independent of `scanned`. Defence in depth doing its job while one layer is blinded is not a property anyone designed, and #1150's whole thesis is that these layers drift. Reporting it with `-F` rather than routing around it was the right call.

---

## Smaller notes, none blocking

- The module comment is the best artefact in this diff. The measured justification for not building on `iter_interpreter_invocations` — naming the specific erasure and the semicolon shape it would have missed — is the kind of thing that stops the next person re-introducing the bug. I re-ran that measurement; it is accurate.
- The positional-agnostic choice (an invocation *before* the write still counts) is right for a security gate and correctly argued from #1152's own resolve-toward-CODE rule.
- `_normalize_path_for_compare` is deliberately narrow and the docstring explains why `~` is left out. Agreed — expanding it would add an untested branch. One consequence worth knowing: `normpath` merges more aggressively than a shell does through symlinked directories, so it can only ever produce *more* matches, which is the safe direction.
- `_TEE_BOOL_FLAGS` correctly notes `--output-error[=MODE]` never starts a bare token. I checked; that holds.
- The `lru_cache` on `_script_invocation_targets` is keyed on `cmd` only and the function is pure in it. Safe, and it earns its keep given both call sites hit it with the same string.

---

Approved on the artefact: correct fix, correct mechanism, both call sites verified non-drifting by independent mutation, no regressions in either direction, CI green at the reviewed SHA. Filed #1286, #1287, #1288, and added the relay evidence to #1168.

---

Requestor: Nino Kavtaradze
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: #1286, #1287, #1288

parametrization:
**Second review — Lucas Ferreira, SRE Engineer**

Verified at head `193b87bf7f11463c681748d7b276e40bc553af30`, CI 14/14 SUCCESS. I ran everything below against the actual PR diff fetched via the contents API at that SHA (not a local checkout), in an isolated scratch harness that imports the PR's `_shell_parse.py` alongside the unmodified `validate_commit_identity.py`/`annunaki_log.py`, so `hook.check()` is the real entry point, not a private helper.

**Fix mechanics.** `_is_opener_code` is the single call site for both `_classify_heredocs_cached` and `strip_data_heredocs` — confirmed by reading the diff, not just the PR body's claim. The new 23-test suite passes through `check()`, and the pre-existing 101-test/7-subtest `test_heredoc_target_scope.py` (main#1152's own false-positive suite) passes unchanged against this branch's `_shell_parse.py`.

**`1>`/`1>>` — my own view, since I was asked for one.** Confirmed empirically (`cat 1> FILE <<'D' ... bash FILE`): still ALLOW on this branch, and — measured separately against `origin/main`'s `_shell_parse.py` in the same harness — also ALLOW there, so not a regression. My read: this is a real improvement, not a false sense of coverage, but only if the residual gap is stated as clearly as the PR already states its other punts. The PR's own "Deliberately NOT covered" section is good practice (`source`/`.`, unresolved shell variables, `bash < FILE`, unspaced redirects) but does not mention `1>`/`1>>`/`&>`/`>&` or the piped-tee-sink shape — the two gaps Nino found and filed as #1286. Since `1>` is the same operator as `>` with the default fd spelled out, I'd treat that section as slightly overselling this PR's redirect-form coverage. Not a reason to block — #1286 already tracks the fix — but worth naming here so the gap doesn't read as newly-closed.

**The unpinned `_iter_non_heredoc_segments` branch.** Reproduced independently. Deleting the heredoc-body-skip loop leaves both the new 23-test suite and the pre-existing 101-test suite passing unchanged (verified — not just asserted), and flips a constructed legitimate case (a README write whose body both illustrates the commit-identity flag syntax and contains a self-referential "run with bash <own-filename>" line, no other invocation anywhere) from an ALLOW result to a hard BLOCK. That's over-blocking, not under-blocking, so it isn't a live bypass — the shipped code is correct today — but it's a new function with a load-bearing, untested branch in the gate that blocks every commit in the org. Filed **#1290** with the repro and a suggested pinning test; not blocking this PR on it since today's behavior is correct and the fix is a small additive test.

**Deploy-shaped false-positive probe (my assigned domain).** All ALLOW, none newly blocks: a release script written then `bash`-run (with and without a `chmod +x &&` prefix), a heredoc-generated `Dockerfile`, a heredoc-generated `docker-compose.yml` (alone, and followed by `docker compose up -d` — a different, non-interpreter command on the same file), and a heredoc `cat > /etc/nginx/nginx.conf` config write. The first two are now correctly reclassified as CODE (write-then-run), but since neither contains a commit-identity pattern, they still ALLOW — confirms the fix doesn't widen the block surface for ordinary deploy tooling.

**Valid-identity write-then-exec.** Confirmed: a heredoc-then-`bash` script carrying a fully valid roster commit-identity flag pair still BLOCKS under this PR's fixed classifier — this is the pre-existing #475/#482 "no indirect execution of a commit, valid or not" policy applying to a body that is now correctly seen as code, not a new identity-specific false positive introduced here.

**Bypass survivors Nino catalogued** (`1>`/`1>>`/`&>`/`>&`, pipe-to-tee-sink, `bash < FILE`, `$(...)`/copy indirection, `xargs`/`find -exec` relay, `source`/`.`) — re-verified a subset directly rather than taking the count at face value; all measured ALLOW on both this branch and `origin/main`, so none is a regression. #1281, #1286, #1287 already track them; #1288 covers the separate phantom-opener/quote-state defect. Not duplicating any of those.

Net: this closes the literal #1167 repro and its semicolon-equivalent variant, which were previously undetected outright, without widening the block surface for legitimate deploy/doc-writing shapes. The residual gaps are all pre-existing and now individually tracked.

---

Requestor: Lucas Ferreira
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: #1290

Issue #1086, opened 2026-07-22
tech-debt(wave-wrapup): add ontology wrap-gate asserting 0 dirty checksums before wrap


parametrization:
## Problem

At the Phase 9 Wave 27 retro, **3 dirty semantic checksums survived into the retro** (`charter/issues.md` — 20 days stale, `.claude/memory/MEMORY.md`, `.claude/memory/project_phase9_close_plan.md` — never resolved). `/wave-wrapup`'s ontology step should have reconciled these before wrap; staleness reaching retro is a process gap.

## Fix

Add a wrap-gate to `/wave-wrapup` (and mirror in `charter/pull-requests/ci-gates.md`) that asserts `ontology/checksums.json` has **0 dirty entries** at wrap time — OR requires an explicit, recorded deferral (a `wave_{M}_ontology_deferred` marker with a reason). The gate should surface which files are dirty so the operator runs `/ontology-rebuild` before wrapping.

## Note
Coordinate with the change-tracker child-path pollution fix (companion issue) — the dirty-count assertion must not be defeated by, and should ideally surface, child-repo pollution.

## Evidence
- Phase 9 Wave 27 retro (`.claude/team/feedback_log.md` § Phase 9 Wave 27, pain point 2 + proposed change 2).

## Acceptance
- [ ] `/wave-wrapup` blocks (or records explicit deferral) on non-zero dirty checksums.
- [ ] `charter/pull-requests/ci-gates.md` documents the gate.
- [ ] Test covers the dirty-checksums-present path.

_Deferred to P10 (last-wave-of-P9 retro finding)._

parametrization:
## Prerequisite filed: #1142 (reader primitive)

This issue's fix requires surfacing "which files are dirty" at wrap time. There is currently **no reader** for that — `.claude/lib/checksums_io.py` exposes `mark-resolved` and `prune` (both writers) and no `status`/`dirty` subcommand, so whoever implements this gate would hand-roll the ledger read.

That hand-rolled read has now failed **twice in two consecutive sessions**, and both failures returned a plausible `0` rather than an error:

- `54ffbb9` committed a change-tracker output that advanced `last_tracked` while leaving `last_resolved` behind — the definition of a dirty entry — under a commit message asserting "0 dirty entries, 277 tracked".
- The next session's `/session-start` Step 3a compared a `sha256` key that does not exist in the entry schema, skipping the comparison on all 277 entries and reporting `0 dirty`.

[...]

Issue #957, opened 2026-07-11
ci: make the hand-rolled CI poller unwritable — pr_ci_state.py must be the only sanctioned merge-gate read


parametrization:
## Summary

**Every merge path must read CI state through `.claude/lib/pr_ci_state.py`, and a hand-rolled poller must be mechanically impossible to write.**

The oracle already exists. The rule is already in org memory (*"EMPTY rollup = hard not-ready"*, main#802). **Both were ignored — by the orchestrator, on the day it spent teaching this exact defect class to three engineers — and the broken loop gated real merges.**

## The defect

```sh
until [ "$(gh pr view N --json statusCheckRollup \
  --jq '[.statusCheckRollup[] | select(.status=="IN_PROGRESS" or .status=="QUEUED")] | length')" = "0" ]; do
  sleep 20
done
# -> "all checks green"
```

**When checks have not yet REGISTERED, `statusCheckRollup` is empty or partial.** Zero pending. The loop exits **immediately** and reports all-green **on a check set that does not exist yet.**

> **An unregistered check is not pending. It is ABSENT.** `pending == 0` is not a terminal condition; it is the *initial* condition.

Independently hit twice on 2026-07-11: once by an implementer (poller reported **7/7 green**; three checks had not registered — caught only because 7 ≠ 10), and once by the **orchestrator**, whose loop gated merges of `deploy#577`, `ip#130`, `ip#133` and `ip#138`. Those merges were safe **only because the checks happened to have registered before the first poll.** That is luck, not a guard.

## The oracle already does it right

`.claude/lib/pr_ci_state.py`:

> *Query a PR's CI-readiness using the merge gate's own classifier (`validate_pr_ci_status.classify_rollup`). Exit 0 if CI is ready (**non-empty rollup**, all checks pass), 1 if not ready (**empty rollup** or any failing/pending check), 2 if undeterminable.*

**Non-empty rollup is the load-bearing clause, and it is exactly the clause a hand-rolled poller omits.**

## Why a rule is not sufficient — this is the whole point

**Ten instances of "a guard that cannot see what it exists to catch" were found on 2026-07-11.** Every one was caught by a person looking hard. Four were committed by people **who were at that moment warning others about that exact class** — including the author of this issue, twice.

> **People looking hard is not a control. It is a coincidence that held ten times.** (Fatima Bensalah)

[...]

PR #1283, opened 2026-08-03
feat(#1142,#1137): checksums_io status reader — one dirty predicate, no silent zeros; prune previews by default

### LINKED ISSUE #1137: tech-debt: checksums_io.py prune CLI writes by default; consider dry-run-by-default for a destructive op on a committed artifact
## Context

PR #1136 (closes #1135) adds a `prune` subcommand to `.claude/lib/checksums_io.py` that removes `ontology/checksums.json` entries whose tracked path no longer exists on disk. By default the CLI **writes** the pruned result; `--dry-run` is opt-in to preview instead.

`prune_missing`'s own docstring documents the caveat that motivates caution here:

> CAVEAT — this is an on-disk existence test, not a git-history one. A file that exists on a child repo's `main` but not on the branch that repo happens to be checked out at right now reads as absent and would be pruned.

The `/ontology-rebuild` SKILL.md step 4 tells the resolver to always run `--dry-run` first and manually verify candidates (`git cat-file -e origin/main:<path>`) before a second, unflagged run — i.e. the documented safe usage pattern already assumes dry-run-first. The CLI's own default does not enforce that pattern: a bare `prune` invocation (run from muscle memory, copy-pasted without the flag, or eventually wired into some other automation) mutates the committed file immediately.

## Ask

Consider inverting the default so a bare `prune` previews (matching the documented safe-usage pattern) and an explicit flag (`--apply` / `--write` / `--yes`) is required to mutate. This is a design/ergonomics call, not a correctness bug — `prune_missing` itself is well-tested and a false-positive prune is self-healing (the next Edit/Write of the file re-creates the entry, per the docstring) — but "destructive-by-default, safe-only-opt-in" is worth reconsidering for a command that mutates a version-controlled artifact based on a check its own docstring flags as unreliable in a specific documented scenario.

## Non-blocking

Raised as a suggestion during PR #1136 review, not a blocker — CI is green, tests (`PruneCliTests`) cover both the dry-run and write paths, and the existing `mark-resolved` sibling CLI has no dry-run option at all (writes unconditionally), so the current default is at least consistent with that established convention. Filed for future consideration rather than gating this PR.

### LINKED ISSUE #1142: tech-debt: checksums_io.py has two writers and no reader — the dirty count is hand-rolled and silently returns a wrong 0
## Symptom

`.claude/lib/checksums_io.py` exposes two **writers** — `mark-resolved` and `prune` — and **no reader**. There is no `status` / `dirty` subcommand, so every consumer that needs the dirty count hand-rolls a JSON read against a schema it has to recall correctly.

Two consecutive sessions got that read wrong, and **both wrong reads returned a plausible `0` rather than an error**:

| When | Reader | Failure | Reported | Truth |
|---|---|---|---|---|
| 2026-07-27 (wave-scope session) | ad-hoc read after `54ffbb9` | conclusion recorded in the commit message and handoff as "0 dirty entries, 277 tracked" | `0 dirty` | `1 dirty` |
| 2026-07-27 (this session, `/session-start` Step 3a) | ad-hoc read comparing a `sha256` key | that key **does not exist** in the entry schema (`last_tracked` / `last_resolved` / `tracked_at` / `resolved_at`), so the comparison was skipped on all 277 entries and the loop counted nothing | `0 dirty` | `1 dirty` |

The canonical definition is `last_tracked != last_resolved` (`.claude/skills/ontology-rebuild/SKILL.md` line 27), implemented correctly in exactly one place: `.claude/hooks/session_start.py::_ontology_staleness`. That implementation is a private function in a hook, not a callable surface, so nothing else can reuse it.

### Concrete instance this surfaced from

`.claude/team/trust_matrix.md` was edited in `a6c9336`. The PostToolUse change-tracker stamped `last_tracked = 3fe12504…`; `54ffbb9` committed that hook output with `last_resolved` still at the pre-edit `d35ea070…`. That diff *is* a dirty entry by definition, yet the commit asserting otherwise shipped, and the discrepancy was then written into the session handoff as a suspected "instrument disagreement" between the hook and a direct read. It was not a disagreement — the hook was right all three times. Resolved in `b3e7e29`.

## Suspected root cause

A read primitive that exists only as a private hook function, combined with an entry schema whose field names are not guessable, means the failure mode of a mistaken read is a **silent wrong zero** — indistinguishable from a genuinely clean ledger. There is no shape of mistake here that produces a loud failure: a wrong key name yields `0`, a wrong nesting level yields `0`, and `0` is the expected healthy value.

## Why this is not a duplicate

- **#1086** (`add ontology wrap-gate asserting 0 dirty checksums before wrap`) is the **enforcement** layer — a gate at one lifecycle point. Its body already requires that "the gate should surface which files are dirty," so #1086 **depends on** this primitive; without it, #1086's implementer hand-rolls the same read that has now failed twice. Filing separately per `feedback_multi_layer_gap_filing.md` (distinct layers sharing a root cause are not duplicates). Landing this first makes #1086 strictly cheaper.
- **#1137** (`prune CLI writes by default`) is the same module's CLI surface but a different concern (destructive-default ergonomics). Already in wave-29 — see the pairing note below.
- **#1131** (`wave counter + trust-signal helpers silently return 0/{} for direct-to-main waves`) is a **sibling instance of the same failure class** in a different module: a helper whose wrong answer is an empty/zero value that reads as healthy. Not a duplicate; worth citing together at retro as evidence for a general "no silent zeros in status helpers" convention.

## Proposed fix

1. Add a `status` subcommand to `checksums_io.py` that reports `total`, `dirty`, and the dirty paths, reusing the module's existing load path (same reader the writers use, so the schema is never re-derived by a caller).
2. Extract the `last_tracked != last_resolved` predicate into a shared function; have `session_start.py::_ontology_staleness` call it rather than re-implementing it.
3. Replace the prose "`cat checksums.json` … build a list of all files where `last_tracked != last_resolved`" in `/ontology-rebuild` SKILL.md step 1 and the equivalent instruction in `/session-start` Step 3a with a call to the new subcommand.

## Pairing note

**#1137 is already in wave-29 and modifies the same CLI's argument parsing** (`checksums_io.py::main`, hand-rolled rather than argparse). Both issues add/change subcommand surface in the same function. Recommend the same implementer takes both, or that they land in a fixed order, to avoid a conflict in `main()`'s hand-rolled parser — which the #1137 body already flags as having bitten once (`prune --dry-run --checksums X` died on an undocumented ordering rule).

- `gh issue list --search "checksums dirty"` → considered #1086 (enforcement layer, see above), #968 (re-tracking archived team files — different scope), #1038/#1135/#1039 (all CLOSED, tracker-side pollution, not the reader).
- `gh issue list --search "checksums_io"` → no reader-related match (#1102/#1099/#1133 are package-architecture and meta issues).
- `gh issue list --search "ontology staleness"` → #1086 again; #1036 (stale conventions.md prose — different scope).
- Rationalization scan (Pass B) → no open issue whose execution would eliminate this class. #1131 is a sibling *instance* of the silent-zero class, not an umbrella over it.

## Acceptance Criteria

- [ ] `checksums_io.py status` reports total / dirty count / dirty paths, exit 0 when clean.
- [ ] The `last_tracked != last_resolved` predicate lives in exactly one place; `session_start.py::_ontology_staleness` consumes it rather than re-implementing it.
- [ ] `/ontology-rebuild` SKILL.md step 1 and `/session-start` Step 3a invoke the subcommand instead of instructing an agent to read the JSON by hand.
- [ ] Test covers a ledger-dirty entry (`last_tracked != last_resolved`), a clean ledger, and a malformed/missing-key entry — the last asserting it does **not** silently count as clean.

## Out of scope

- The wrap-time gate itself — #1086.
- `prune`'s destructive default — #1137.
- Merge-model-aware wave counters — #1131.

parametrization:
Closes #1142
Closes #1137

Landed as one PR because #1142's own pairing note asks for it: both issues change the same hand-rolled `main()` argument parser in `.claude/lib/checksums_io.py`, and #1137 records that this parser has already bitten once (`prune --dry-run --checksums X` died on an undocumented ordering rule). Splitting them was a guaranteed conflict.

---

## #1142 — the silent wrong zero

The module had two **writers** (`mark-resolved`, `prune`) and no reader, so every consumer that needed the dirty count hand-rolled a JSON read against a schema it had to recall correctly. Two consecutive sessions got that read wrong and **both wrong reads returned a plausible `0`**.

Three changes close the class:

1. **`classify_entry` is the one implementation** of `last_tracked != last_resolved`.
2. **`ENTRY_MALFORMED` is a third state**, never folded into "clean". Both historical wrong reads produced entries a `.get(...) != .get(...)` comparison silently called clean.
3. **The read path forks.** `read_checksums` still fails OPEN (the PostToolUse tracker hook must never raise). `read_checksums_strict` / `read_status` raise `ChecksumsUnreadable` — a reader that answers "0 dirty" for a file it failed to parse *is* the bug.

`status` exit codes make the split visible: **0** clean, **1** needs attention, **3** unreadable. `--json` so a future wrap-time gate (#1086) consumes this rather than re-deriving the predicate.

### It found a real one on its first run

```
ontology/checksums.json: 284 tracked, 0 dirty, 1 malformed
malformed (unrecognized entry schema — NOT counted clean):
  - noorinalabs-deploy/.github/branch-protection/apply-ruleset.sh: missing last_tracked
```

```json
{"last_resolved": null, "resolved_at": "2026-06-14T00:16:00Z"}
```

Hand-written into the committed ledger by a pre-#1042 `/ontology-rebuild` pass (`bec0430`, 2026-06-14) that had no `mark-resolved` CLI to go through. Every reader since compared `None != None`, found them equal, and counted it clean — for seven weeks, in the session-start line, the handoff line, and the librarian's report alike. `mark-resolved` cannot repair it (there is no hash to copy), so the last commit deletes the entry through `write_checksums`; the tracker re-creates it correctly on the next Edit/Write of that file. Diff is the four removed lines and nothing else.

### "Exactly one place" required converting three readers, not one

The issue named `session_start.py::_ontology_staleness`. A repo sweep found two more hand-rolled copies, so converting only the named one would have left the acceptance criterion literally unmet:

| Reader | Why it matters |
|---|---|
| `.claude/hooks/session_start.py::_ontology_staleness` | The session's first vital sign |
| `.claude/hooks/session_handoff.py::_get_ontology_staleness` | The Stop hook's handoff line — what the **next** session reads first. A wrong "0 dirty files" here is a wrong belief carried across a session boundary, which is exactly how the original miscount propagated into a handoff and got re-litigated as a suspected "instrument disagreement" |
| `.claude/hooks/smart_grep_ontology.py::_dirty_files` | Drives the `[STALE]` annotation on ontology hits. Malformed entries join the set on purpose — an entry the reader cannot classify is not-known-current, which is what the annotation says |

All three keep their fail-open posture: they catch `ChecksumsUnreadable` and degrade to their own existing "could not read" branch, never to a zero.

---

## #1137 — destructive default

`prune` now previews by default; `--apply` is required to write.

- **`--dry-run` is kept** as an accepted no-op spelling of the default, so every previously-safe invocation stays safe and keeps working.
- **`--dry-run --apply` is a usage error**, not a silent precedence rule. Either precedence is defensible and neither is guessable — the same shape as the ordering rule that already bit `--checksums`.
- **`--force` stays orthogonal to `--apply`.** Forcing past a guard still previews. Collapsing them would make the escape hatch for a false-positive guard also the escape hatch for the write.
- The preview names the next step (`--apply`) so the flip does not strand whoever runs the old spelling.

### The two counterweights the issue raises

**On the `mark-resolved` asymmetry.** Today's write-by-default is consistent with `mark-resolved`, which writes unconditionally with no dry-run. Inverting only `prune` is still correct, because the two commands have different blast radii, not different levels of caution:

- `mark-resolved` is **additive and idempotent**. It stamps `last_resolved`/`resolved_at` on entries that already exist; a path it does not know is silently skipped. Its worst misfire quiets a file that should have stayed dirty — recoverable by the next Edit re-stamping `last_tracked`.
- `prune` **deletes entries**, and its input is the on-disk existence of ~280 paths, roughly half of them inside gitignored child clones whose presence depends on which branch each child repo happens to be checked out at. `prune_missing`'s own docstring flags that as a documented false-positive scenario.

Same module, different blast radius. Documented in `_prune_cli`'s docstring so the next reader does not read the asymmetry as an oversight.

**On overselling.** This is a design/ergonomics call, not a correctness bug. `prune_missing` is well-tested, the three existing guards already refuse the mass-delete cases, and a false-positive prune is self-healing. The change is that the documented safe-usage pattern (`/ontology-rebuild` step 4 has always said preview-then-verify-then-write) is now enforced by the CLI instead of relying on the operator remembering a flag.

---

## Call sites and docs updated

Every place that invoked `prune` or told an agent to read the JSON by hand:

| File | Change |
|---|---|
| `.claude/skills/ontology-rebuild/SKILL.md` step 1 | `cat checksums.json` + hand-compare -> `checksums_io.py status`, with exit-code table and malformed-entry repair guidance |
| `.claude/skills/ontology-rebuild/SKILL.md` step 4 | `prune --dry-run` then unflagged second run -> `prune` (previews) then `prune --apply`; `--force`/`--apply` orthogonality stated |
| `.claude/skills/session-start/SKILL.md` Step 3a | Prose "if 0 dirty files in checksums.json" -> the subcommand + exit-code table; exit 3 called out as **a problem to report, not an empty work list** |
| `.claude/skills/ontology-librarian/SKILL.md` step 1a | **Not named in the issue** — found by sweep. Told the agent to `cat checksums.json` and count by hand, in the one skill whose entire job is reporting staleness. Now the subcommand + exit-code table |
| `ontology/conventions.md` (tracker row) | `prune --dry-run` (opt-in) -> previews by default / `--apply` to write; plus "read the dirty count with `status`, never a hand-rolled compare" |
| `.claude/lib/checksums_io.py` (`prune_missing` docstring) | The CAVEAT paragraph's "that is why the CLI is opt-in with `--dry-run`" rationale retargeted to the new default |

---

## Mutation testing

16 mutations, each applied in place with the rest of the tree untouched, full module suite run, source restored. **Zero survivors.** Baseline clean both before and after.

| # | Mutation | Caught by |
|---|---|---|
| M1 | `classify_entry`: non-dict entry is CLEAN not MALFORMED | `ClassifyEntryTests::test_non_dict_entry_is_malformed` |
| M2 | `classify_entry`: drop the missing-key check -> `.get()/.get()` (**the historical bug**) | `test_missing_both_keys_is_malformed_not_clean`, `test_missing_only_last_tracked_is_malformed`, `StatusCliTests::test_malformed_entry_does_not_silently_count_as_clean` |
| M3 | `classify_entry`: drop the non-string check (two nulls compare equal) | `ClassifyEntryTests::test_null_hash_is_malformed_not_clean` |
| M4 | `ChecksumsStatus.clean`: ignore malformed | `ComputeStatusTests::test_malformed_entry_blocks_clean`, `StatusCliTests::test_malformed_entry_does_not_silently_count_as_clean` |
| M5 | `compute_status`: restore the legacy flat fallback instead of raising | `test_missing_files_key_raises_rather_than_reporting_zero`, `test_non_dict_files_raises` |
| M6 | `read_checksums_strict`: fail OPEN like the writers' reader | `StrictReadTests::test_invalid_json_raises`, `test_missing_file_raises_instead_of_defaulting`, `StatusCliTests::test_malformed_json_exits_three`, `test_unreadable_ledger_exits_three_not_zero` |
| M7 | `_status_cli`: always exit 0 | `test_dirty_ledger_reports_count_and_paths`, `test_json_output_is_machine_readable`, `test_malformed_entry_does_not_silently_count_as_clean` |
| M8 | `_status_cli`: report an unreadable ledger as an empty one (**the silent zero**) | `test_malformed_json_exits_three`, `test_unreadable_ledger_exits_three_not_zero` |
| M9 | `_prune_cli`: restore write-by-default (**#1137 reverted**) | `test_bare_prune_previews_and_does_not_write`, `test_checksums_flag_works_in_any_position`, `test_dry_run_leaves_the_file_untouched`, `test_preview_output_names_the_apply_flag`, `test_force_alone_still_only_previews`, `test_over_threshold_prune_is_refused_on_dry_run_too` |
| M10 | `_prune_cli`: `--dry-run --apply` silently picks a winner | `PruneCliTests::test_dry_run_plus_apply_is_a_usage_error` |
| M11 | `_prune_cli`: `--force` implies `--apply` | `PruneGuardTests::test_force_alone_still_only_previews` |
| M12 | `_prune_cli`: drop the "how to apply" hint from the preview | `PruneCliTests::test_preview_output_names_the_apply_flag` |
| M13 | `session_start`: re-inline the old hand-rolled predicate | `OntologyStalenessTests::test_malformed_entry_is_not_reported_as_current` |
| M14 | `session_handoff`: revert to the hand-rolled predicate | `OntologyStalenessTests::test_malformed_entry_is_not_reported_as_current` |
| M15 | `smart_grep_ontology`: `_dirty_files` returns an empty set | `test_stale_source_is_flagged`, `test_malformed_checksum_entry_is_flagged_stale_not_current` |
| M16 | `smart_grep_ontology`: drop malformed from the not-known-current set | `test_malformed_checksum_entry_is_flagged_stale_not_current` |

Two honest notes on how this table was produced:

- **The harness lied once.** Its first run reported **M1 as SURVIVED**. It had not — pytest reports `unittest.subTest` failures as `SUBFAILED(params) <nodeid>`, not `FAILED <nodeid>`, and the parser only matched `FAILED`. Verified by hand (`4 failed, 84 passed`, `AssertionError: 'clean' != 'malformed'`), then fixed the parser and re-ran. A mutation harness that under-reports catches is itself a silent-wrong-answer instrument, which is on-theme enough to be worth recording.
- **M16 genuinely survived** the first extended run — nothing covered "malformed joins the not-known-current set" in `smart_grep_ontology`. `test_malformed_checksum_entry_is_flagged_stale_not_current` was added in response, not retrofitted around a passing state.

---

## Verification

- `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/ .claude/lib/tests/ -q` -> **3561 passed, 651 subtests passed** (re-run after rebase onto `origin/main`)
- `ruff check` / `ruff format --check` over `.claude/hooks/` + `.claude/lib/` -> clean
- `mypy .claude/hooks/ .claude/lib/ --ignore-missing-imports` -> clean
- `python3 .claude/lib/pre_commit_ci_sync.py .` -> `OK: pre-commit config mirrors all CI-enforced checks`
- Both pre-commit stages green on every commit and on push (no `--no-verify`)
- Live ledger after the repair: `285 tracked, 0 dirty, 0 malformed` (exit 0)

## Follow-up filed

**#1284** — `tech-debt`/`process`/`phase-10`, no wave label. The single-predicate property is enforced only *behaviorally, per consumer*: each converted reader has a malformed-entry test a naive re-implementation would fail (M13-M16 above), but a **seventh** reader written next month is covered by nothing, and its wrong answer will be a plausible `0`. #1284 proposes the structural gate (an `ast-grep` lint on the comparison outside `checksums_io.py`, plus a prose lint on `cat checksums.json` in SKILL.md blocks), and cites **#957** as precedent — the same "one sanctioned reader, lint the hand-rolled alternative unwritable" shape, for the CI poller. It also carries a small pre-existing item: `test_explicit_repo_root_is_honored` is vacuous (seeds no orphan, so it passes whether or not `--repo-root` is honored) — not introduced here, not fixed here.

## Out of scope

Per #1142: the wrap-time gate (#1086 — this PR's `status --json` is the primitive it needs) and merge-model-aware wave counters (#1131, a sibling instance of the same silent-zero class).

## Reviewers

Merge-gate reviewer: Nino Kavtaradze (same reviewer on both issues).


parametrization:
## Merge-gate review — head `f393ab1a0e4e2ca1cfeca7f345232e23f295de7d`

Reviewed at the head SHA above (verified immediately before posting; unchanged through the review). CI **21/21 SUCCESS** at that SHA via `statusCheckRollup`. All work done in a detached worktree at `f393ab1`; mutations restored with `git checkout HEAD -- <file>`, tree confirmed clean after each.

Nothing in the report was taken on faith. Every claim below was re-derived.

---

### Independent mutation harness — 12 mutations, zero survivors

I wrote my own harness rather than reusing yours, and deliberately took the verdict from **pytest's process exit code** rather than parsing stdout. That is the structural fix for the failure mode you disclosed: an exit code cannot under-report.

```
BASELINE: exit=0 :: 151 passed, 4 subtests passed
M2a: KILLED  exit=1  FAILED-lines=3 :: 3 failed, 148 passed
M2b: KILLED  exit=1  FAILED-lines=9 :: 9 failed, 142 passed
M5:  KILLED  exit=1  FAILED-lines=1 :: 1 failed, 150 passed
M6:  KILLED  exit=1  FAILED-lines=7 :: 7 failed, 144 passed
M8:  KILLED  exit=1  FAILED-lines=2 :: 2 failed, 149 passed
M9:  KILLED  exit=1  FAILED-lines=6 :: 6 failed, 145 passed
M10: KILLED  exit=1  FAILED-lines=1 :: 1 failed, 150 passed
M11: KILLED  exit=1  FAILED-lines=1 :: 1 failed, 150 passed
M13: KILLED  exit=1  FAILED-lines=1 :: 1 failed, 150 passed
M16: KILLED  exit=1  FAILED-lines=1 :: 1 failed, 150 passed
POST-RESTORE BASELINE: exit=0 :: 151 passed, 4 subtests passed
```

I split **M2** because your table describes it as dropping only the missing-key check, which leaves the `isinstance(value, str)` guard standing — so I ran both the narrow form (**M2a**, missing-key check only) and the full historical bug (**M2b**, `entry.get() != entry.get()` with no schema validation at all). Both die. The historical bug cannot be reintroduced in either shape.

### Your harness under-reported once — I reproduced it exactly

This was the check worth doing, and your account is precisely right. Applying M1 (non-dict entry → CLEAN):

```
$ rg -c '^FAILED' m1.out
0 (none)

$ rg '^SUBFAIL' m1.out
SUBFAILED(value='a string') ...::ClassifyEntryTests::test_non_dict_entry_is_malformed
SUBFAILED(value=42)         ...::ClassifyEntryTests::test_non_dict_entry_is_malformed
SUBFAILED(value=None)       ...::ClassifyEntryTests::test_non_dict_entry_is_malformed
SUBFAILED(value=['list'])   ...::ClassifyEntryTests::test_non_dict_entry_is_malformed

$ tail -1 m1.out
4 failed, 84 passed in 0.58s
```

Zero `FAILED` lines, four `SUBFAILED(...)` lines, `4 failed, 84 passed` — your reported numbers to the digit. A `^FAILED` parser reports SURVIVED; the exit code reports `1`. My harness records both columns (`FAILED-lines`, `SUBFAIL-lines`) so the discrepancy would be visible rather than silent, and it never branches on either. **Your fixed harness does not still under-report** — its results and my exit-code results agree on all ten shared mutations.

Recording the miss in the PR body was the right call, and the note is accurate.

### The silent zero, end to end

Running the new reader against the **base** ledger next to the old hand-rolled predicate:

```
BASE ledger via NEW reader: 286 tracked, 0 dirty, 1 malformed
   malformed: noorinalabs-deploy/.github/branch-protection/apply-ruleset.sh -> missing last_tracked
BASE ledger via OLD hand-rolled predicate: 0 dirty, 286 total  <-- the silent miss
```

That is the whole issue in four lines. Exit codes distinguish the three outcomes for real: clean `0`, dirty/malformed `1`, unreadable `3`, usage `2`.

---

### The parser — probed directly, in an isolated fake repo tree

I never pointed a writer at the real ledger; I built a fake `<root>/.claude/lib/` + `<root>/ontology/` layout so `_default_checksums_path()` could only ever resolve inside scratch. Every `prune` row below was md5-checked before and after.

`status`: unknown flag → `2`; `--checksums` with no value → `2`; `--checksums=PATH` → `2`; repeated `--checksums` → `2`; `--json --checksums` (flag consumed as value) → `2`; nonexistent path → `3`. Nothing silent.

`prune`:

| Invocation | rc | File |
|---|---|---|
| `prune` (bare) | 0 | UNCHANGED |
| `prune --dry-run` | 0 | UNCHANGED |
| `prune --apply` | 0 | wrote |
| `prune --dry-run --apply` | 2 | UNCHANGED |
| `prune --apply --dry-run` | 2 | UNCHANGED |
| `prune --force` (alone) | 0 | UNCHANGED |
| `prune --force --apply` | 0 | wrote |
| `prune --dry-run --checksums X` | 0 | UNCHANGED |
| `prune --checksums X --dry-run` | 0 | UNCHANGED |
| `prune --repo-root` (no value) | 2 | UNCHANGED |
| `prune --repo-root /nope` | 2 | UNCHANGED |
| `prune --bogus` / `--APPLY` | 2 | UNCHANGED |

M9, M10 and M11 all confirmed behaviourally, not just by test: preview is the default, the contradiction is a usage error in **both orders**, and `--force` alone previews. The ordering rule that bit `--checksums` is genuinely gone.

**One real finding, filed as #1285 — not blocking.** `mark-resolved` is the one subcommand with no argument validation: any unrecognized token becomes a `<rel-path>`. `mark-resolved --checksums=PATH b.txt` therefore writes the **default** ledger — the committed one — at exit 0, while `status` and `prune` both reject the same `=` form with exit 2. I confirmed this is **pre-existing**: the identical extraction is in the base `main()` at merge-base `3005004`, so you did not introduce it and it is outside this PR's scope. It is worth tracking because your own tightening of the other two subcommands is what makes `mark-resolved` the odd one out.

---

### The deleted ledger entry — verified, not masked

```
bec0430  2026-06-13  Aino Virtanen  ontology: rebuild P4W7 — resolve apply-ruleset.sh checksum
+    "noorinalabs-deploy/.github/branch-protection/apply-ruleset.sh": {
+      "last_resolved": null,
+      "resolved_at": "2026-06-14T00:16:00Z"
+    }
```

Hand-inserted by a pre-#1042 resolver pass with no `mark-resolved` CLI to route through, exactly as described. Four checks:

1. **The file is not gone.** It exists on disk *and* on the deploy repo's `origin/main` (`git cat-file -e` succeeds). So the deletion is repairing a malformed record, not papering over a vanished file — and the tracker will re-create it correctly on the next Edit/Write. This is also why `prune` would not have touched it.
2. **`mark-resolved` genuinely cannot repair it.** I ran it: the entry stays `missing last_tracked`, because `mark_resolved` copies `entry.get("last_tracked", "")` and there is nothing to copy. Deleting through `write_checksums` was the only correct route.
3. **The diff is only that entry.** Parsed both sides against the true merge-base `3005004` (three-dot — `origin/main` has moved since): `removed: [apply-ruleset.sh]`, `added: []`, `changed: []`, non-`files` top level byte-identical. `git diff --numstat` agrees: `0 insertions, 4 deletions`.
4. **The gate is green.** `check_checksums_ascii.py` → `OK: checksums.json holds no '\u' escape sequences`, exit 0. Live ledger at head: `285 tracked, 0 dirty, 0 malformed`, exit 0.

Worth knowing rather than fixing: between now and the next Edit of that file the path is simply untracked. That is functionally identical to the prior state — an entry that could never report dirty — so no coverage is lost.

### Scope expansion — correct discipline, not creep

I agree with your reading, and the argument is stronger than "three more places." #1142's acceptance criterion is that **exactly one place** implements the predicate. Converting only `session_start.py` would have left three live hand-rolled copies, so the criterion would have been literally unmet on merge — the issue would have closed against a false claim. That is the rare case where widening scope is required *to satisfy* the stated acceptance rather than to exceed it.

The `/ontology-librarian` catch is the best of the four: the one skill whose entire job is reporting staleness was instructing the agent to `cat` the file and count by hand. Leaving that would have meant closing a silent-wrong-zero issue with the wrong zero still being produced by the staleness reporter itself.

I executed all three converted readers against fixtures rather than reading the diff:

| fixture | `session_start` | `session_handoff` | `smart_grep._dirty_files` |
|---|---|---|---|
| clean | `total=1 dirty=[] malf=[]` | Ontology is current (0 dirty files) | `[]` |
| dirty | `dirty=['a']` | Ontology has 1 dirty files: a | `['a']` |
| malformed | `malf=['a']` | Ontology has 1 malformed entries: a | `['a']` |
| no `files` key | `None (unreadable)` | Could not read checksums | `[]` |
| non-dict top level | `None (unreadable)` | Could not read checksums | `[]` |
| invalid JSON | `None (unreadable)` | Could not read checksums | `[]` |
| missing file | `None (unreadable)` | No ontology checksums found | `[]` |

Identical to the prior behaviour except where the change is the point: malformed no longer reads as clean, and the two reporting readers say "could not read" where they used to produce a plausible zero. `smart_grep` keeps its silent empty set on an unreadable ledger — unchanged from base, advisory-only, and documented in the docstring, so not a regression. `session_handoff` also gains a small robustness fix it does not claim: a non-dict top level used to raise `AttributeError` past the `except (JSONDecodeError, OSError)`.

### The `mark-resolved` asymmetry — the argument holds

On the merits, and I checked the premises rather than the reasoning. `mark_resolved` skips paths not already present and only stamps `last_resolved`/`resolved_at` on existing entries — additive and idempotent, verified by running it. `prune_missing` deletes, on an existence test whose own docstring documents a false-positive scenario. Different blast radius is the correct axis, and inverting only the destructive one is right. Symmetry between a command that adds and a command that deletes would be consistency for its own sake.

#1285 sharpens rather than undercuts this: `mark-resolved` deserves stricter *argument* handling, which is a separate axis from whether it needs a preview mode. It does not.

### Correction to the follow-up note in #1284

The note calls `test_explicit_repo_root_is_honored` vacuous because it seeds no orphan. I mutation-tested that (`--repo-root` parsed but discarded) and **it does currently kill the mutant** — `AssertionError: 1 != 0`. But not for the reason the test's name implies: the kill comes entirely from Guard 3's 25% threshold firing because the fixture holds exactly one entry. Re-running the same mutant with nine ballast entries, so the prune set is 1/10:

```
MUTANT (--repo-root ignored), fixture WITH ballast:
  rc = 0 ... 'kept.md' still in files: True  => mutation SURVIVES
```

So the debt is the inverse of what is recorded: the test is load-bearing *by accident*, and the obvious improvement — giving the fixture realistic ballast — would silently delete the only coverage `--repo-root` has. Also worth noting that the contents assertion `["kept.md"]` *was* load-bearing before this PR and the preview-by-default flip made it inert. Detail posted on #1284; no separate issue.

### Other checks

Full suite at head: `2526 passed, 632 subtests passed`. All 40 new tests scanned for missing assertions — two AST hits were false positives (bare `assert`, pytest-style), both confirmed genuine. Commit identity correct on all five commits. Semantic overlay clean.

---

Excellent work. The disclosure of the harness's own silent-wrong-answer, in a PR about silent wrong answers, is the part I'd most want repeated — and it held up under independent reproduction, which is the only thing that makes such a disclosure worth anything.

Requestor: Nino Kavtaradze
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1285

parametrization:
Second review of #1283 (closes #1142, #1137). CI verified 21/21 SUCCESS at head `f393ab1a0e4e2ca1cfeca7f345232e23f295de7d` (re-checked immediately before posting).

## Independent reproduction of the `test_explicit_repo_root_is_honored` finding

I re-ran this from scratch against a standalone copy of `checksums_io.py` + `test_checksums_io.py` pulled from the branch (never against the real `ontology/checksums.json`). Mutation: `--repo-root` is parsed but its value is discarded (`_prune_cli` falls back to the default `checksums_path.resolve().parent.parent`).

- **1-entry fixture (the shipped test, no ballast):** mutant is caught, but only via Guard 3 — `assertEqual(rc, 0)` fails because the sanity threshold refuses at "1 of 1 entries (100%)". Stderr names `--repo-root` explicitly, but nothing in the test exercises the flag's semantics directly; the content assertion (`kept.md` still present) is inert under preview-by-default regardless of the mutation.
- **Same mutant, 9 ballast entries added the way sibling tests do (`self._ballast(root)`):** ratio drops to 1/10, Guard 3 does not fire, `rc == 0`, `kept.md` still present — **mutation survives.**

This matches Nino's correction to #1284 exactly and I confirm it independently: the recorded claim ("test is vacuous") had the polarity backwards. The test is load-bearing today, but by accident (via an unrelated safety guard), and the natural "hygiene" fix — giving the fixture realistic ballast, which is exactly the pattern used in the other four `PruneCliTests` — would silently delete the only real coverage of `--repo-root`. A durable fix needs to assert the discriminating observable directly (the preview list / captured stdout naming the resolved root), not infer it from the exit code. Left as a comment on #1284, which already carries this item — not duplicating.

## Reviewable as one unit? Yes, with one bisect-granularity note

13 files / ~1050+/93- across two issues is sizeable, but the scope is coherent: six call-site conversions and four skill/doc rewrites are all mechanical, downstream applications of the single new `classify_entry`/`read_status` contract — #1142's own acceptance criterion (exactly one implementation of the predicate) requires seeing every consumer converted in the same unit to verify closure, so splitting the sweep out would have re-created the ambiguity the issue exists to close. The checksums.json ledger fix is a one-line, independently-verified correction (confirmed: `noorinalabs-deploy/.github/branch-protection/apply-ruleset.sh` exists on disk and on deploy's `origin/main` — the entry was malformed, not masking a deletion).

The 5-commit structure is clean and mostly bisectable — 4 of 5 commits are single-concern. The one place bisect resolution blurs is commit `3337489`, which lands **both** the `status` reader (#1142) and the `prune` preview-by-default flip (#1137) together, with the pairing rationale in the commit body ("both change the same hand-rolled `main()` parser — splitting was a guaranteed conflict"). That reasoning is sound for the PR-level pairing, but it didn't need to extend to the commit level — a future regression in either `status` or `prune`'s default would land bisect on the same commit and require reading the diff by hand to tell which. Minor, non-blocking; noting for the record per the brief since it's too late to split now.

## `prune` call-site sweep

Swept the full branch tree (not just the diff) plus `origin/main` for every `checksums_io.py ... prune` / `prune_missing(` call site outside the module's own tests:

- `ontology/conventions.md` (row for `ontology_tracker.py`): updated from `prune --dry-run` (opt-in write-by-default) to `prune` bare + explicit "`--apply` required to write (#1137)" language. Correct.
- `.claude/skills/ontology-rebuild/SKILL.md` step 4: updated from "`prune --dry-run` ... re-run without `--dry-run`" to "`prune` (previews by default) ... re-run with `--apply`", with an explicit `prune --apply` command block added. Correct, and consistent in both directions with the code's new default.
- No hooks, no CI workflow, no pre-commit config invoke `checksums_io.py prune` anywhere (grep on `.github/workflows/*.yml`, `.pre-commit-config.yaml`, and all three touched hooks came back empty) — `prune` remains a documented manual-invocation-only CLI, consistent with the module's own docstring. Nothing was silently switched from write to preview outside the two prose sites above, and both were caught and updated.

Also independently verified `session-start/SKILL.md` and `ontology-librarian/SKILL.md`'s new exit-code tables (0 clean / 1 dirty-or-malformed / 3 unreadable) against `_status_cli`'s actual return values — they match.

---
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1284

Issue #1273, opened 2026-08-03
tech-debt(hook4): the comment-scan mode is rendered by three independent hand-written surfaces with no consistency pin


parametrization:
## The debt

The four-state comment-scan enum (`COMMENT_SCAN_NOT_RUN` / `AUTHOR_EXCLUDED` /
`COMMIT_AUTHOR_EXCLUDED` / `NO_BRANCH_AUTHOR`) is now rendered for humans in **three
independent, hand-written places**, and nothing pins that they agree:

1. `.claude/hooks/validate_pr_review.py` — `check()`'s `scan_diagnostic`, composed into the
   **block** `reason` (three branches).
2. `.claude/hooks/validate_pr_review.py` — `check()`'s new **allow-path** advisory (#1211, PR
   #1270), one branch on `NO_BRANCH_AUTHOR`.
3. `.claude/lib/pr_review_state.py` — `_describe_comment_scan()`, the operator-run oracle report
   (four branches).

Each was written separately and repeats the same substantive claims in different prose ("a bot
author, a merge-only branch, or a squashed commit re-authored to the bare principal", "no
self-review exclusion was applied", the #1177/#1206/#1210 issue refs). They are already only
*nearly* identical.

## Why it is worth an issue

The failure mode is asymmetric and silent, which is exactly the class this hook keeps getting
bitten by. Suppose #1210's `refine_comment_scan_scope` gains a fifth mode, or the meaning of
`NO_BRANCH_AUTHOR` narrows. A contributor updates the block message (the one they were looking
at) and misses the oracle, or updates the oracle and misses the allow advisory. The result is
two surfaces making **different claims about the same state** — and the one that goes stale is
whichever the contributor was not reading, i.e. unpredictable. No test compares them; the
existing tests assert each renderer's text in isolation.

This is the same shape as #1206 itself: the state was knowable, and the surface said something
else.

## Not urgent

Nothing is wrong today — PR #1270 checked all three by hand and they agree. This is drift risk,
not a live defect, and it is deliberately **not** folded into #1270 (which is scoped to the
disclosure gap and should stay reviewable).

## Suggested shape

A shared `describe_comment_scan_mode(mode, *, head_ref, commit_authors, surface)` helper that
returns the mode's substantive claims once, with the three call sites supplying only their own
framing (block prefix / advisory prefix / report line). Plus one test that iterates every
`COMMENT_SCAN_*` constant and asserts each surface renders a non-empty, mode-distinct string —
[...]

PR #1292, opened 2026-08-03
fix(hook4/#1220): scan mode must not claim an exclusion it did not apply

### LINKED ISSUE #1220: tech-debt(hook4): scan mode reports 'self-review exclusion active' on a bot branch where nothing was excluded
## Reproduction

Driven through the real `resolve_review_verdicts` at PR #1215's head SHA (`0186d4d`), on a
dependabot head ref whose one commit is authored by `dependabot[bot]`, with two genuine
roster approvals:

```
comment_scan   : commit-author-excluded
derived        : ['dependabot[bot]']
total_distinct : 2

pr_review_state report line:
  RAN — head ref names no branch author, so the branch author was derived from
  COMMIT IDENTITY: dependabot[bot]. Their own verdicts are excluded from the
  reviewer set (self-review exclusion active, #1210).
```

The gate's count is **correct** — both approvals counted, deploy#691 stays fixed. The
*description* is false: nothing was excluded, and `dependabot[bot]` does not post charter
verdicts. `check()`'s block message says the same thing in the same situation ("Verdicts
from those personas were excluded as self-reviews").

## Cause

`refine_comment_scan_scope` upgrades `NO_BRANCH_AUTHOR` → `COMMIT_AUTHOR_EXCLUDED` on
**any** derived identity:

```python
if ref_scope == COMMENT_SCAN_NO_BRANCH_AUTHOR and commit_authors:
    return COMMENT_SCAN_COMMIT_AUTHOR_EXCLUDED
```

"An identity was derived" is not the same fact as "an exclusion was applied." For a bot
author the first is true and the second is false, so the mode reports a subtraction that
did not occur.

This is the same class of defect #1206/#1207 fixed — the scan mode is supposed to be an
honest description of what the scan did. `test_bot_branch_with_two_roster_approvals_still_
passes` asserts the count and the derived identities but not the mode, which is why it is
green.

## Fix options

1. Key the upgrade on whether any derived identity **matches a roster persona**
   (`_load_roster_names` is already resolved in `resolve_review_verdicts`, though after the
   `refine` call — it would need reordering).
2. Split the mode: `COMMIT_AUTHOR_EXCLUDED` only when the exclusion actually removed a
   verdict; otherwise keep `NO_BRANCH_AUTHOR` and mention the derived-but-inert identity in
   the diagnostic text.
3. Reword both diagnostics to "derived from commit identity: X — any verdict from X would
   be excluded as a self-review", which is true whether or not one was.

(3) is the smallest and does not perturb the mode's semantics. Add a mode assertion to the
bot test either way.

Filed from the merge-gate review of #1215. No gate impact — reporting only.

parametrization:
Closes #1220

## The defect

`refine_comment_scan_scope` upgraded `NO_BRANCH_AUTHOR` → `COMMIT_AUTHOR_EXCLUDED` on **any** derived identity:

```python
if ref_scope == COMMENT_SCAN_NO_BRANCH_AUTHOR and commit_authors:
    return COMMENT_SCAN_COMMIT_AUTHOR_EXCLUDED
```

"An identity was derived" is not the same fact as "an exclusion was applied." On a `dependabot/**` head ref the first is true and the second is false, so the gate counted both roster approvals correctly (deploy#691 stays fixed) and then described itself as having subtracted a self-review that never existed. Same defect class as #1206/#1207 — the mode is meant to be an honest description of what the scan actually did.

## The fix: a fifth mode, not a reword

`COMMENT_SCAN_COMMIT_AUTHOR_NON_ROSTER` — an identity **was** derived and it matches **no roster persona**.

I chose the new state over the issue's option 3 (reword only) because the two facts an operator needs here are genuinely different, and collapsing either way throws one away:

- It is not `COMMIT_AUTHOR_EXCLUDED`: that mode's claim is that a subtraction is in force, and here none is.
- It is not `NO_BRANCH_AUTHOR`: that mode's claim is "the PR's commits named no persona", and here they named `dependabot[bot]`. Collapsing into it would discard the derived identity — which is exactly what an operator would audit if they suspected the derivation had picked the wrong person.

The inertness is **provable, not incidental**, and it is the same argument `commit_author_identities` already makes for not roster-filtering the derivation itself: only roster-valid Requestors reach `distinct_reviewers`, so an exclusion keyed on a non-roster identity can only ever remove a verdict roster filtering would have removed anyway. **The new mode and `NO_BRANCH_AUTHOR` produce byte-identical reviewer sets.**

Liveness is decided by `commit_author_exclusion_is_live`, which delegates to `charter_trailer.is_branch_author` — the one person-comparison in the org (#1172). No surname `in` test: a roster holding `santiago ferreira` must not make a `Lucas Ferreira` commit look live (mutation M3).

### Two deliberate degradations

- **Unreadable roster.** `_load_roster_names` returns `set()` when the roster dirs cannot be read. On that input *every* identity trivially "matches no roster persona", so reporting `NON_ROSTER` would be a measurement derived from a failed read — the `feedback_silent_zero_is_not_a_measurement` shape this enum exists to prevent. The undecidable case keeps the pre-#1220 answer (mutations M4, M13).
- **`roster_names` is a required positional arg**, not defaulted. A `set()` default would read as "roster unreadable" and hand a forgetful caller a plausible answer for every input.

## Scope: nothing the gate counts changed

Structurally guaranteed, not just asserted: `refine_comment_scan_scope` now runs **after** `check_comment_reviews`, so the scan has already completed by the time the mode is chosen. The exclusion itself still runs off the same **unfiltered** `commit_authors` tuple.

The refine call was placed after `_load_roster_names` rather than hoisting the roster load above `check_comment_reviews` **on purpose**: hoisting would move `RosterResolutionError` ahead of `CommentScanUndeterminedError` in the raise order, so a PR with both an unreadable comment thread and an unresolvable child roster would start reporting a different one of its two determinate failures. Both hard-block, so nothing counts differently — but the diagnosis handed to an operator would silently change, and this PR is about descriptions.

## Rendering surfaces touched — all three (#1273)

| # | Surface | Change |
|---|---|---|
| 1 | `validate_pr_review.check()` — block `scan_diagnostic` | New `NON_ROSTER` branch. `COMMIT_AUTHOR_EXCLUDED`'s wording made modal: "Verdicts from those personas **were excluded**" → "Any verdict from those personas **is excluded**" (true whether or not one existed). |
| 2 | `validate_pr_review.check()` — allow-path advisory (#1211/#1270) | Now fires for `NON_ROSTER` too, with its own wording. Its reviewer set is byte-identical to `NO_BRANCH_AUTHOR`'s, so staying silent would disclose on the arithmetic's twin and not on the arithmetic itself. |
| 3 | `pr_review_state._describe_comment_scan()` | New `NON_ROSTER` branch, and — separately — its **silent catch-all is gone**. It used to end in a bare `return` on `NO_BRANCH_AUTHOR`'s text, so any unwired mode was described as "the PR's commits named no persona": confident, specific, wrong. Unrecognized modes now say `UNRECOGNIZED SCAN MODE`. |

## #1273: narrowed, not closed

**Done here** — the part that makes a sixth mode fail loudly:

- `ALL_COMMENT_SCAN_MODES` (the constant tuple #1273 notes does not exist).
- `CommentScanModeTotalityTests` — iterates it against all three surfaces with a **declared coverage table**. Absence from a surface is a decision (`AUTHOR_EXCLUDED` needs no block diagnostic — the ref names the author, nothing is surprising), and `test_table_covers_exactly_the_declared_mode_set` fails until a new mode's decision is recorded. `test_the_unrecognized_marker_is_actually_reachable` is the positive control that keeps the `assertNotIn` assertions from passing vacuously forever.

**Still open on #1273** — the shared `describe_comment_scan_mode(...)` helper. The three surfaces still repeat the same substantive claims in three hand-written prose blocks; they now agree and are pinned to agree *structurally* (every mode wired, distinctly, everywhere), but not *textually*. I have commented on #1273 (comment 5167853575) to re-scope it to the dedup and record what this PR removed from it.

## Testing

Assertion-only tests are not the bar. Every new branch was mutation-tested; all runs drive the real `resolve_review_verdicts` / `check()` / `compute_review_state` entry points as the issue's reproduction does. Restores were `git checkout HEAD -- <file>`.

| # | Mutation | Killed by | Real output |
|---|---|---|---|
| M1 | Revert `refine_comment_scan_scope` to the pre-#1220 body (**the original defect**) | `RefineCommentScanScopeTests::test_derived_identity_matching_no_roster_persona_is_reported_inert`, `::test_same_surname_different_person_is_not_live`, `CommitIdentitySelfReviewExclusionTests::test_bot_branch_with_two_roster_approvals_still_passes`, `InertCommitDerivationReportTests::test_compute_review_state_reports_the_inert_mode_end_to_end` | `4 failed, 379 passed, 78 subtests passed` |
| M2 | Drop the liveness check — every derivation called inert (breaks #1210) | `::test_no_branch_author_upgrades_when_commits_named_a_roster_persona`, `::test_one_roster_persona_among_several_derivations_is_still_live`, `::test_bot_branch_with_a_human_fixup_commit_still_reports_live_exclusion`, `::test_self_approval_on_a_non_charter_ref_no_longer_reaches_two_of_two`, `CommitAuthorScanModeReportTests::test_compute_review_state_propagates_the_commit_derived_author` | `5 failed, 378 passed, 78 subtests passed` |
| M3 | Liveness by surname `in` — drops the #1172 initial discriminator | `RefineCommentScanScopeTests::test_same_surname_different_person_is_not_live` | `1 failed, 382 passed, 78 subtests passed` |
| M4 | Unreadable roster reported as an inertness finding | `::test_unreadable_roster_keeps_the_pre_1220_answer`, `CommitIdentitySelfReviewExclusionTests::test_an_unreadable_roster_does_not_report_the_bot_as_inert` | `2 failed, 381 passed, 78 subtests passed` |
| M5 | Block path: new mode gets no diagnostic at all | `CommitAuthorBlockDiagnosticTests::test_non_roster_derivation_does_not_claim_a_subtraction`, `::test_inert_and_live_commit_derivations_read_differently`, + subtest | `3 failed, 381 passed, 77 subtests` — `AssertionError: '' is not true : commit-author-non-roster renders no block scan diagnostic at all` (`SUBFAILED(mode='commit-author-non-roster') CommentScanModeTotalityTests::test_every_mode_gets_its_declared_block_diagnostic`) |
| M6 | Block path: new mode's body collapsed onto `NO_BRANCH_AUTHOR`'s prose (**the #1273 drift shape**) | `::test_non_roster_derivation_does_not_claim_a_subtraction`, `::test_inert_and_live_commit_derivations_read_differently`, `::test_inert_mode_is_not_collapsed_into_the_no_author_wording` | `3 failed, 380 passed, 78 subtests passed` |
| M7 | Allow path: new mode does not disclose | `AllowPathScanDisclosureTests::test_inert_commit_derivation_also_discloses_on_the_allow_path`, `::test_the_two_no_exclusion_advisories_are_not_the_same_text`, + subtest | `4 failed, 381 passed, 76 subtests` — `AssertionError: commit-author-non-roster must disclose on the allow path / assert None is not None` |
| M8 | Allow path: hardcode `2/2` on the `NO_BRANCH_AUTHOR` advisory (**Nadia's #1270 finding — this survived all 318 tests before**) | `AllowPathScanDisclosureTests::test_advisory_reports_the_real_count_on_a_wave_bootstrap_single_reviewer` | `SUBFAILED(mode='no-branch-author')` — `AssertionError: 'reached 1/2 WITHOUT self-review exclusion' not found in "NOTE: PR #691 reached 2/2 WITHOUT self-review exclusion. …"` |
| M8b | Same hardcode on the **new** advisory branch | same test, `mode='commit-author-non-roster'` | `AssertionError: 'reached 1/2 WITHOUT self-review exclusion' not found in 'NOTE: PR #691 reached 2/2 … derived from COMMIT IDENTITY — dependabot[bot] — matches no roster persona …'` |
| M9 | Oracle report: new mode unwired (falls to the fallback) | `InertCommitDerivationReportTests::test_inert_mode_names_the_derivation_and_denies_the_subtraction`, + subtest | `2 failed, 382 passed, 77 subtests` — `AssertionError: 'UNRECOGNIZED SCAN MODE' unexpectedly found … : commit-author-non-roster is unwired in the oracle report` |
| M10 | Oracle report: restore the silent `NO_BRANCH_AUTHOR` catch-all | `CommentScanModeTotalityTests::test_the_unrecognized_marker_is_actually_reachable`, `InertCommitDerivationReportTests::test_unrecognized_mode_says_so_instead_of_borrowing_a_neighbour` | `2 failed, 381 passed, 78 subtests passed` |
| M11 | **A sixth mode declared but wired nowhere** — the #1273 scenario | `CommentScanModeTotalityTests::test_table_covers_exactly_the_declared_mode_set` + `SUBFAILED(mode='some-future-mode') ::test_every_mode_renders_a_recognized_report_line` | `2 failed, 382 passed, 78 subtests passed` — `AssertionError: Items in the second set but not the first: 'some-future-mode'` |
| M12 | Call site forgets the roster (passes the degraded `set()`) | `::test_bot_branch_with_two_roster_approvals_still_passes`, `InertCommitDerivationReportTests::test_compute_review_state_reports_the_inert_mode_end_to_end` | `2 failed, 381 passed, 78 subtests passed` |
| M13 | **The description fix leaks into the COUNT** — roster-filter the derivation itself | `::test_bot_branch_with_two_roster_approvals_still_passes`, `::test_an_unreadable_roster_does_not_report_the_bot_as_inert`, `InertCommitDerivationReportTests::test_compute_review_state_reports_the_inert_mode_end_to_end` | `3 failed, 380 passed, 78 subtests passed` |

Baseline for every row: `383 passed, 78 subtests passed` across the two suites.

Honest note on M13: it is caught by the **mode** assertions (roster-filtering the derivation turns a bot branch into `NO_BRANCH_AUTHOR`), not by a count assertion — because on this input the count is genuinely unchanged. The count guarantee rests on the structural argument above plus the pre-existing `StrictlyNonRelaxingTests`, not on M13.

### The specific traps

- **#1203** (five bare `assertIn("1/2", reason)`): no new loose substring assertions. Every new assertion names either a specific mode constant or a specific rendered phrase (`"matches no roster persona"`, `"NO verdict was excluded as a self-review"`, `"reached 1/2 WITHOUT self-review exclusion"`). I have **not** touched the five pre-existing ones — that stays #1203's.
- **Nadia's #1270 finding**: covered by M8/M8b. `test_advisory_reports_the_real_count_on_a_wave_bootstrap_single_reviewer` drives the tail at `total_distinct == 1` via the wave-bootstrap exception — the combination that had no test — and subtests both no-exclusion modes so neither branch can regress to a literal.
- **The anti-vacuity lesson in situ**: `test_bot_branch_with_two_roster_approvals_still_passes` now asserts the **mode**, which is the assertion it was missing and the reason it stayed green through the defect. `test_bot_branch_with_a_human_fixup_commit_still_reports_live_exclusion` is its positive control — without it, the new mode assertion would be satisfied by a change that simply stopped ever reporting `COMMIT_AUTHOR_EXCLUDED` on a non-charter ref, re-opening #1210.

## Gates

`3666 passed, 668 subtests passed` (full `.claude/hooks/tests/` + `.claude/lib/tests/`). Both pre-commit stages clean over `--all-files`: ruff-format, ruff-lint, actionlint, cspell, checksums-ascii, fixture-realism, skill-graphql-pagination, mypy, mypy-skills, pytest, pytest-skills, memory-budget, headcount-budget, doc-freshness, office-drift, mermaid-render. `pre_commit_ci_sync.py` → `OK`. Hook exits 0 on empty stdin. No `--no-verify`.

Semantic overlay was current (288 tracked, 0 dirty) and needs no update: `ontology/conventions.md`'s only reference to this hook is the unchanged one-line role row, and no charter or doc file mentions the scan-mode vocabulary (`rg` over `.claude/team/charter/` and `docs/` returns nothing).

## Follow-up debt filed

- #1273 re-scoped (commented, not closed) — the shared renderer helper.
- **#1294** (new, `tech-debt`/`process`/`phase-10`, no wave label) — `check()`'s block-path `scan_diagnostic` still has no unrecognized-mode fallback, while `_describe_comment_scan` now does. Empty is a *legitimate* state there (`AUTHOR_EXCLUDED` renders nothing on purpose), so the `else` needs a design call rather than a reflex; pinned by test in the meantime.

### DIFF
--- a/.claude/hooks/tests/test_validate_pr_review.py
+++ b/.claude/hooks/tests/test_validate_pr_review.py
@@ -39,7 +39,16 @@ from unittest import mock
 _HERE = Path(__file__).resolve().parent
 _HOOKS_DIR = _HERE.parent
 sys.path.insert(0, str(_HOOKS_DIR))
+sys.path.insert(0, str(_HOOKS_DIR.parent / "lib"))
 
+# The oracle driver (`.claude/lib/`) is the THIRD surface that renders the
+# comment-scan enum, alongside `check()`'s block message and its allow-path
+# advisory (#1273). `CommentScanModeTotalityTests` below asserts all three agree
+# on the same mode set, which is only possible from a module that can see both;
+# this import exists for that class and for nothing else. The dependency runs
+# test -> lib -> hook (pr_review_state already imports the gate), so it adds no
+# cycle to the shipped code.
+import pr_review_state as prs  # noqa: E402
 import validate_pr_review as hook  # noqa: E402
 
 
@@ -4506,25 +4515,97 @@ class IsSelfReviewTests(unittest.TestCase):
 
 
 class RefineCommentScanScopeTests(unittest.TestCase):
-    """#1210: the ONE allowed scope transition, and the three forbidden ones."""
+    """#1210's allowed transitions, re-keyed by #1220 on roster membership."""
 
+    ROSTER = {"aino virtanen", "santiago ferreira", "nino kavtaradze"}
     IDENT = (hook.CommitAuthorIdentity(lastname="Virtanen", initial="a", display="Aino Virtanen"),)
+    BOT = (
+        hook.CommitAuthorIdentity(
+            lastname="dependabot[bot]", initial="d", display="dependabot[bot]"
+        ),
+    )
 
-    def test_no_branch_author_upgrades_when_commits_named_someone(self):
+    def test_no_branch_author_upgrades_when_commits_named_a_roster_persona(self):
         self.assertEqual(
-            hook.refine_comment_scan_scope(hook.COMMENT_SCAN_NO_BRANCH_AUTHOR, self.IDENT),
+            hook.refine_comment_scan_scope(
+                hook.COMMENT_SCAN_NO_BRANCH_AUTHOR, self.IDENT, self.ROSTER
+            ),
+            hook.COMMENT_SCAN_COMMIT_AUTHOR_EXCLUDED,
+        )
+
+    def test_derived_identity_matching_no_roster_persona_is_reported_inert(self):
+        """THE #1220 unit-level kill-shot.
+
+        Pre-fix this returned COMMIT_AUTHOR_EXCLUDED — "an identity was derived"
+        reported as "an exclusion was applied". `dependabot[bot]` can never be a
+        Requestor that survives roster filtering, so nothing was subtracted and
+        the mode must not claim one was.
+        """
+        self.assertEqual(
+            hook.refine_comment_scan_scope(
+                hook.COMMENT_SCAN_NO_BRANCH_AUTHOR, self.BOT, self.ROSTER
+            ),
+            hook.COMMENT_SCAN_COMMIT_AUTHOR_NON_ROSTER,
+        )
+
+    def test_one_roster_persona_among_several_derivations_is_still_live(self):
+        """A human fixup on a bot branch. Exclusion IS live for that human, so
+        the inert reading must not win just because a bot is also in the tuple —
+        the predicate is ANY, not ALL."""
+        self.assertEqual(
+            hook.refine_comment_scan_scope(
+                hook.COMMENT_SCAN_NO_BRANCH_AUTHOR, self.BOT + self.IDENT, self.ROSTER
+            ),
+            hook.COMMENT_SCAN_COMMIT_AUTHOR_EXCLUDED,
+        )
+
+    def test_same_surname_different_person_is_not_live(self):
+        """#1172 must hold in the new predicate too: Lucas is not Santiago.
+
+        The roster here contains `santiago ferreira` and NOT Lucas. A liveness
+        check written as a surname `in` test would call this live and re-open the
+        Ferreira/Ferreira collision inside the reporting layer.
+        """
+        lucas = (
+            hook.CommitAuthorIdentity(lastname="Ferreira", initial="l", display="Lucas Ferreira"),
+        )
+        self.assertEqual(
+            hook.refine_comment_scan_scope(hook.COMMENT_SCAN_NO_BRANCH_AUTHOR, lucas, self.ROSTER),
+            hook.COMMENT_SCAN_COMMIT_AUTHOR_NON_ROSTER,
+        )
+
+    def test_unreadable_roster_keeps_the_pre_1220_answer(self):
+        """An empty roster means the roster could not be READ, not that the org
+        is empty — so "matches no roster persona" would be a finding derived
+        from a failed read. The undecidable case keeps the modal EXCLUDED
+        wording rather than asserting inertness it cannot establish."""
+        self.assertEqual(
+            hook.refine_comment_scan_scope(hook.COMMENT_SCAN_NO_BRANCH_AUTHOR, self.IDENT, set()),
             hook.COMMENT_SCAN_COMMIT_AUTHOR_EXCLUDED,
         )
 
     def test_no_branch_author_stays_when_commits_named_nobody(self):
         self.assertEqual(
-            hook.refine_comment_scan_scope(hook.COMMENT_SCAN_NO_BRANCH_AUTHOR, ()),
+            hook.refine_comment_scan_scope(hook.COMMENT_SCAN_NO_BRANCH_AUTHOR, (), self.ROSTER),
             hook.COMMENT_SCAN_NO_BRANCH_AUTHOR,
         )
 
     def test_ref_derived_author_is_never_downgraded(self):
         self.assertEqual(
-            hook.refine_comment_scan_scope(hook.COMMENT_SCAN_AUTHOR_EXCLUDED, self.IDENT),
+            hook.refine_comment_scan_scope(
+                hook.COMMENT_SCAN_AUTHOR_EXCLUDED, self.IDENT, self.ROSTER
+            ),
+            hook.COMMENT_SCAN_AUTHOR_EXCLUDED,
+        )
+
+    def test_ref_derived_author_is_not_downgraded_by_a_non_roster_derivation(self):
+        """The #1220 arm must not leak onto the ref path either: a declared
+        `{Initial}.{Lastname}` author keeps its exclusion whatever the commits
+        say, including when they say `dependabot[bot]`."""
+        self.assertEqual(
+            hook.refine_comment_scan_scope(
+                hook.COMMENT_SCAN_AUTHOR_EXCLUDED, self.BOT, self.ROSTER
+            ),
             hook.COMMENT_SCAN_AUTHOR_EXCLUDED,
         )
 
@@ -4535,9 +4616,28 @@ class RefineCommentScanScopeTests(unittest.TestCase):
         defense-in-depth hard-block one layer up.
         """
         self.assertEqual(
-            hook.refine_comment_scan_scope(hook.COMMENT_SCAN_NOT_RUN, self.IDENT),
+            hook.refine_comment_scan_scope(hook.COMMENT_SCAN_NOT_RUN, self.IDENT, self.ROSTER),
             hook.COMMENT_SCAN_NOT_RUN,
         )
+
+    def test_not_run_is_not_relabelled_as_inert_either(self):
+        """#1220 added a second reachable answer, so NOT_RUN now has two ways to
+        be laundered. Both must be closed."""
+        self.assertEqual(
+            hook.refine_comment_scan_scope(hook.COMMENT_SCAN_NOT_RUN, self.BOT, self.ROSTER),
+            hook.COMMENT_SCAN_NOT_RUN,
+        )
+
+    def test_roster_names_is_required_not_defaulted(self):
+        """A caller that forgets the roster must fail loudly.
+
+        A `set()` default would read as "roster unreadable" and hand them a
+        plausible EXCLUDED answer for every input — the silent-wrong direction.
+        """
+        with self.assertRaises(TypeError):
+            hook.refine_comment_scan_scope(  # type: ignore[call-arg]
+                hook.COMMENT_SCAN_NO_BRANCH_AUTHOR, self.BOT
+            )
 
 
 class CommitIdentitySelfReviewExclusionTests(_ResolveOverFakeCommentsHarness):
@@ -4658,6 +4758,69 @@ class CommitIdentitySelfReviewExclusionTests(_ResolveOverFakeCommentsHarness):
         self.assertEqual(
             [i.display for i in verdicts.commit_author_identities], ["dependabot[bot]"]
         )
+        # #1220: THE assertion this test was missing, and the reason it stayed
+        # green through the defect. It pinned the count and the derived
+        # identities — the safe half of the claim — while the mode said
+        # `commit-author-excluded`, i.e. announced a subtraction of a verdict
+        # that was never subtracted. Both approvals are still counted above; the
+        # mode must now describe that honestly.
+        self.assertEqual(verdicts.comment_scan, hook.COMMENT_SCAN_COMMIT_AUTHOR_NON_ROSTER)
+
+    def test_bot_branch_with_a_human_fixup_commit_still_reports_live_exclusion(self):
+        """Anti-vacuity control for the test above (#1220).
+
+        Same dependabot ref, same two roster approvals — except one commit is
+        authored by one of the approvers. Without this, the assertion above would
+        be satisfied by a change that simply stopped ever reporting
+        COMMIT_AUTHOR_EXCLUDED on a non-charter ref, which would re-open #1210:
+        Nino's own verdict must still be subtracted AND still be described as
+        subtracted.
+        """
+        verdicts = self._resolve(
+            self.DEPENDABOT_REF,
+            [self._verdict("Lucas Ferreira"), self._verdict("Nino Kavtaradze")],
+            commits=[
+                _api_commit(
+                    "c0",
+                    "2026-07-15T19:13:59Z",
+                    author_name="dependabot[bot]",
+                    author_email="49699333+dependabot[bot]@users.noreply.github.com",
+                ),
+                _api_commit(
+                    "c1",
+                    "2026-07-15T19:13:59Z",
+                    author_name="Nino Kavtaradze",
+                    author_email="parametrization+Nino.Kavtaradze@gmail.com",
+                ),
+            ],
+        )
+        self.assertEqual(verdicts.distinct_reviewers, {"lucas ferreira"})
+        self.assertEqual(verdicts.total_distinct, 1)
+        self.assertEqual(verdicts.comment_scan, hook.COMMENT_SCAN_COMMIT_AUTHOR_EXCLUDED)
+
+    def test_an_unreadable_roster_does_not_report_the_bot_as_inert(self):
+        """The degraded read must not masquerade as an inertness measurement.
+
+        With an unreadable roster (`_load_roster_names` returns the empty set)
+        EVERY identity trivially matches nothing, so a naive predicate would
+        report `commit-author-non-roster` — a confident claim about roster
+        membership derived from a failed roster read. The mode falls back to the
+        pre-#1220 answer instead.
+        """
+        verdicts = self._resolve(
+            self.DEPENDABOT_REF,
+            [self._verdict("Lucas Ferreira"), self._verdict("Nino Kavtaradze")],
+            roster=set(),
+            commits=[
+                _api_commit(
+                    "c0",
+                    "2026-07-15T19:13:59Z",
+                    author_name="dependabot[bot]",
+                    author_email="49699333+dependabot[bot]@users.noreply.github.com",
+                )
+            ],
+        )
+        self.assertEqual(verdicts.comment_scan, hook.COMMENT_SCAN_COMMIT_AUTHOR_EXCLUDED)
 
     def test_squash_flattened_identity_excludes_nobody(self):
         """The bare principal must not resolve to the persona roster.json maps it to.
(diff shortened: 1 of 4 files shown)


parametrization:
**Second review (security/merge-gate), at head `55c6107cde324b9726cb2768ef918dc41bfcd98b`.** CI 14/14 green at that SHA.

## What I set out to break

The whole PR rests on: this is a description-only fix, and no input diverges the reviewer set between `NO_BRANCH_AUTHOR` and the new `COMMIT_AUTHOR_NON_ROSTER` mode. I tried to falsify that from two independent angles.

**1. Data-flow trace.** In `resolve_review_verdicts`, `check_comment_reviews` (which applies the actual exclusion via `is_self_review`) runs and returns *before* `refine_comment_scan_scope` is even called — `roster_names` is loaded and the refine call happens strictly after. `scan_scope`'s only consumer downstream is the `comment_scan` field on `ReviewVerdicts`, read by the three rendering surfaces. So no mutation confined to `refine_comment_scan_scope` / `commit_author_exclusion_is_live` can feed back into `distinct_reviewers` at all — it's not merely that the predicates happen to agree, the mode computation is control-flow-downstream of the count and cannot touch it. This is a stronger guarantee than the semantic-equivalence argument in the PR body.

**2. Semantic proof, for completeness.** `commit_author_exclusion_is_live` is `any(is_branch_author(name, identity.lastname, identity.initial) for identity in commit_authors for name in roster_names)`. `is_self_review`'s commit-derived arm is `any(is_branch_author(requestor, identity.lastname, identity.initial) for identity in commit_authors)`. Any roster-valid reviewer `r` that reaches `distinct_reviewers` is by construction an element of `roster_names` (`roster_comment_reviewers = comment_result.reviewers - non_roster_requestors`, filtered by exact-string `r in roster_names`). So if `is_self_review(r, ...)` would exclude `r`, that exact `(name=r, identity)` pair is also inside `commit_author_exclusion_is_live`'s double-`any`, forcing mode to `COMMIT_AUTHOR_EXCLUDED`, not `NON_ROSTER`. Byte-identical reviewer sets hold structurally, not incidentally.

**Roster-collision fixture (#1269 shape, `Kofi Mensah` vs `Kofi Mensah-Williams`).** Ran both directions directly against `commit_author_exclusion_is_live` / `is_self_review`:
```
live (Kofi Mensah vs roster w/ Kofi Mensah-Williams): False
mode: commit-author-non-roster
is_self_review(Kofi Mensah-Williams | derived=Kofi Mensah): False
live (Kofi Mensah-Williams vs roster w/ Kofi Mensah): False
is_self_review(Kofi Mensah | derived=Kofi Mensah-Williams): False
```
No collision — `name_lastname` does not split on the hyphen, so `"Mensah"` and `"Mensah-Williams"` are distinct lastnames under strict equality and never satisfy `is_branch_author` in either direction. Also re-ran the Ferreira/Ferreira (#1172) shape from `test_same_surname_different_person_is_not_live` directly, both as the "different person" case (correctly inert) and as a flipped "correct match" case (correctly live) — both landed exactly as claimed.

**Mutation spot-checks (own harness, not the PR author's copy):**
- **M1** (revert `refine_comment_scan_scope` to the pre-#1220 body): `4 failed, 379 passed, 78 subtests passed` — exact match to the claimed row.
- **M3** (liveness via surname `in` instead of `is_branch_author`): `1 failed, 382 passed, 78 subtests passed` — exact match, and the single failure is `test_same_surname_different_person_is_not_live`, confirming the #1172 guard is real and not vacuous.
- Restored both via a fresh `git archive <ref> | tar -x` into a name-scoped scratch dir each time (not a scratch backup), never `git --work-tree` against the main checkout, then re-ran the full suite to confirm `383 passed, 78 subtests passed` baseline before moving on.

**M13 honesty note, read closely.** The PR is correct that M13 is caught by mode assertions, not a count assertion — and I could not construct an input where a "roster-filter the derivation" mutation leaks into the count undetected, for the reason in (1) above: whatever such a mutation does, it cannot reach the count-computing call at all given the current ordering. The PR's own wording actually understates its case here; I'd call the count guarantee proven, not merely argued.

**Unrecognized-mode fallback.** Confirmed `test_the_unrecognized_marker_is_actually_reachable` is a genuine positive control — it drives `_describe_comment_scan` with a literal `"some-future-mode"` string not in `ALL_COMMENT_SCAN_MODES` and asserts the `UNRECOGNIZED SCAN MODE` marker appears. The removed silent catch-all's replacement is reachable, not dead code.

## Control-flow ordering (refine moved after `check_comment_reviews`)

Read through `resolve_review_verdicts`: `commit_authors` is computed once (line ~2139), consumed unfiltered by `check_comment_reviews` immediately after, and only *then* is `roster_names` loaded and `refine_comment_scan_scope` called on the untouched `commit_authors` tuple. The exclusion side and the description side both read the same tuple, and the description side reads it strictly later — matches the docstring's claim and closes the raise-order hazard (RosterResolutionError vs CommentScanUndeterminedError) the PR calls out.

## Other checks

- Full suite at this SHA: `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/test_validate_pr_review.py .claude/lib/tests/test_pr_review_state.py -q` → `383 passed, 78 subtests passed`, matching the PR's stated baseline.
- Follow-up debt: #1294 exists, correctly labeled `tech-debt`/`process`/`phase-10` with no wave label, and accurately scopes the remaining block-path fallback gap. #1273 carries the re-scoping comment narrowing it to the shared-renderer dedup, consistent with what shipped here.
- Ontology: semantic overlay current (288 tracked, 0 dirty, 0 malformed) per `checksums_io.py status`; no charter/doc references the scan-mode vocabulary, consistent with the PR's own note.

I was not able to falsify the byte-identical-reviewer-sets claim; if anything the guarantee is stronger than argued (structural, not just semantic).

---
Requestor: Nino Kavtaradze
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Merge-gate review — Opus tier

Reviewed at head SHA `55c6107cde324b9726cb2768ef918dc41bfcd98b`, CI **14/14 SUCCESS** at that SHA. Files read via `gh api contents?ref=<sha>`, not a local checkout. Baseline reproduced in an isolated `git archive` export of the head tree: `383 passed, 78 subtests passed`, matching the PR body.

I reviewed #1270, which built the allow-path advisory this extends. Everything below is my own execution, not a re-reading of your reasoning.

---

### 1. "The new mode and `NO_BRANCH_AUTHOR` produce byte-identical reviewer sets" — verified, and it is a theorem, not a coincidence

I built the case that would break it rather than reading the argument. Property: for each input, drive the real `resolve_review_verdicts` twice — once with the fixture's commits, once with a commit yielding no identity (the forced-`NO_BRANCH_AUTHOR` twin) — and compare.

| input | mode | set == twin | count |
|---|---|---|---|
| `dependabot[bot]` author | non-roster | yes | 4 |
| `GitHub` web-UI commit | non-roster | yes | 4 |
| outside contributor, novel name | non-roster | yes | 4 |
| **`Marco Ferreira` — surname collides with two roster personas** | non-roster | yes | 4 |
| two colliding surnames at once | non-roster | yes | 4 |
| bot + non-roster human | non-roster | yes | 4 |
| `12345+octocat@users.noreply` alias | non-roster | yes | 4 |
| *control* — single-token `Kavtaradze` | excluded | **no** | 3 |
| *control* — roster persona via `+alias` only | excluded | **no** | 3 |
| *control* — bot + one roster persona | excluded | **no** | 3 |

Zero violations, and the controls confirm the classifier is not vacuously answering "inert" — the three that should subtract do subtract, 4 → 3.

The proof it rests on is tighter than the PR body states, and worth writing down because it is what makes this safe to extend:

- `commit_author_exclusion_is_live` and `is_self_review` call **the same predicate with the same argument order** — `is_branch_author(name, identity.lastname, identity.initial)`. Not merely "both delegate to #1172".
- Every Requestor that can reach `distinct_reviewers` must be an **exact string member** of `roster_names` (`non_roster_requestors` is a set-difference, not a fuzzy match). So `roster_names` is a superset of the countable Requestors, and `commit_author_exclusion_is_live` iterates it exhaustively.
- The two sides compare `requestor` raw versus `roster_names` lowercased, so the theorem needs `is_branch_author` to be case-stable. It is: `_name_tokens` never inspects case, `name_first_initial` lowercases, and `is_branch_author` lowercases both lastnames.

Contrapositive: under the new mode, no roster name matches any identity, therefore no countable Requestor can be excluded. Byte-identical follows. The surname-collision row is the one that would have exposed a surname `in` test, and it is clean.

**One precision correction.** The claim is true of `distinct_reviewers`. It is *not* true of `non_roster_requestors` — a non-roster Requestor matching the derived identity is swallowed by the exclusion and vanishes from that list. Nothing counted changes, but the new mode's wording states a universal it does not hold. Filed as #1298 with the reproduction; the suppression predates you, the sentence that denies it does not.

### 2. Control-flow move — nothing counted changed, and the raise order is *preserved*, not merely defended

`scan_scope` is read at exactly two points between assignment and reassignment (the `NOT_RUN` guard, then the refine), and `check_comment_reviews` takes the branch author and the identities, never the mode. No ordering-dependent state is introduced; the exclusion still runs off the unfiltered tuple.

Your hoisting argument is stronger than you claimed. You defended the placement as *avoiding* an order change — but the order is bit-for-bit what it was: `NOT_RUN` → `CommentScanUndeterminedError` → `RosterResolutionError`, before and after. And it is pinned: when I hoisted `_load_roster_names` as part of a mutation, `IncompleteCommentScanFailsClosedTests::test_check_hard_blocks_on_an_incomplete_scan` went red with the roster error text in place of the scan error text. Your judgement call has a test behind it that you did not cite.

### 3. Unreadable roster keeps the pre-#1220 answer — right call, for a reason worth recording

I ran it. With `roster_names` degraded to the empty set, the mode stays `commit-author-excluded` *and*:

```
readable   : mode=commit-author-non-roster  total_distinct=2  non_roster=[]
UNREADABLE : mode=commit-author-excluded    total_distinct=0  non_roster=['lucas ferreira', 'nino kavtaradze']
```

An unreadable roster is not a quiet state. Every Requestor becomes non-roster, the count collapses to the formal reviewers, and the PR hard-blocks with all of them named. The alarm already exists and it is loud — so the mode's job there is precisely *not* to invent a second one out of a failed read. Reporting inertness would have been a measurement derived from nothing, and it would have been the *quieter* of the two answers besides. Endorsed.

`roster_names` as a required positional is also right, and confirmed to raise `TypeError` on omission. A `set()` default would land a forgetful caller in exactly the degraded branch above — the one branch whose whole design assumes it was reached deliberately.

### 4. My #1270 finding — genuinely closed

`test_advisory_reports_the_real_count_on_a_wave_bootstrap_single_reviewer` drives the real `check()`; only `resolve_review_verdicts` is stubbed, so `total_distinct` is computed by production code from `distinct_reviewers` and the advisory is really rendered. It is not an assertion on a constructed string. Subtesting both no-exclusion modes closes the branch I could only have found by reading. `assertNotIn("2/2")` alongside the positive assertion is the right shape.

### 5. Silent catch-all removal — replacement verified reachable

`test_the_unrecognized_marker_is_actually_reachable` is a real positive control. I confirmed independently: an undeclared mode renders `UNRECOGNIZED SCAN MODE` and names the mode. Restoring the catch-all (your M10) kills that test plus `test_unrecognized_mode_says_so_instead_of_borrowing_a_neighbour` — 2 failures, exactly as claimed.

Removing that bare `return` is the best change in the PR. It converted a confident wrong answer into an admitted unknown on the surface that most needed it.

### 6. Independently re-ran your mutation table — it is not decorative

`M1` 4 failed / 379 passed · `M10` 2 failed / 381 passed · `M6` 3 failed / 380 passed. All match the PR body, against my own applications of the mutations.

---

## Ruling on M13: your honest note is accurate, and it understates the gap

You wrote that the count guarantee rests on the structural argument plus the pre-existing `StrictlyNonRelaxingTests`. I tested that rest. It does not hold, and it cannot: that class asserts post ⊆ pre where `pre` is the same resolve with **no** derived identity. An edit that *drops* identities moves the set toward `pre`, so the subset holds by construction. It pins against over-exclusion and is structurally blind to under-exclusion — the exact direction M13 travels.

So I wrote the variant your M13 was one step away from: narrow **only** the tuple handed to `check_comment_reviews`, leave every mode, every diagnostic and `commit_author_identities` on the unfiltered tuple, and preserve the raise order.

```
383 passed, 78 subtests passed
```

Fully green — and the gate is relaxed. Non-charter ref, Nino approves his own branch, Aino approves:

| commit author fields | shipped | mutated |
|---|---|---|
| `name="Nino Kavtaradze"` (every fixture in the suite) | 1/2 blocked | 1/2 |
| `name="nino-k"`, persona only in the `+alias` | 1/2 blocked | **2/2 allowed** |
| `name=""`, persona only in the `+alias` | 1/2 blocked | **2/2 allowed** |
| `name="Kavtaradze"`, bare-principal email | 1/2 blocked | **2/2 allowed** |

That is #1210 re-opened with the suite green. The gap is the space between `is_branch_author` and string membership in `roster_names`: the `+alias` arm yields the dotted `Nino.Kavtaradze`, which matches under the first and not the second — and that arm only ever does the work when `user.name` is a handle or empty, which is the entire reason it exists. Every count assertion in the suite uses the full-roster-name fixture, i.e. row 1, the only row that survives such a filter.

**Verdict on sufficiency: sufficient for this merge, not sufficient as a standing guarantee — and the difference is not your PR's to fix.** I confirmed the same mutation is equally invisible on `main` (`358 passed, 61 subtests`). You neither introduced nor worsened it; your mode assertions are strictly more coverage than existed before, and they *do* kill the M13 you wrote. Holding #1292 for a pre-existing blind spot would be charging you for the fact that you were the first person to look.

The closing assertion is one count test, filed with a copy-paste body in **#1297** — a non-charter ref where the branch author is derivable **only** in a form that is not a `roster_names` member string, asserting `total_distinct == 1`. RED under the mutation, GREEN today. Pair it with the bare-surname shape for the initial-unknown branch of `is_branch_author`.

## On #1273's stopping point: right place to stop, and I made the remainder precise

I mutation-tested where the totality suite's pin ends and commented the evidence on the issue. A sixth mode wired everywhere with copy-pasted prose **is** caught (2 failures). The same sixth mode declared in `_TABLE` with a distinct oracle line, but with its block diagnostic and allow advisory byte-identical to `NO_BRANCH_AUTHOR`'s, passes: `383 passed, 81 subtests`. The asymmetry is one test — pairwise-distinctness exists for `_describe_comment_scan` only; the other two surfaces get a boolean coverage table plus `assertIn("scan", …)` / `assertIn("self-review exclusion", …)`, all satisfied by a verbatim copy.

Which is exactly your framing: structural, not textual. Stopping here was correct — you shipped the part that makes a sixth mode fail loudly and left the dedup, rather than doing a three-surface refactor inside a fix for a false sentence. #1294 is scoped correctly too; I confirmed the block path renders an empty prefix and the allow path returns `None` for an undeclared mode, and that empty is genuinely legitimate there for `AUTHOR_EXCLUDED`. That needs a design call, not a reflex `else`.

## Minor, no action

`ontology/checksums.json` reports this hook clean while its content has diverged — 148 of 288 tracked entries no longer match their recorded hash. Your "0 dirty" reading was correct and the ledger is what is wrong; already tracked as #1219. Noting it only so the reasoning in your PR body is not later mistaken for the ledger having been consulted successfully.

---

## Verdict

Approved. The one claim the whole change rests on is a theorem and I verified it against inputs chosen to break it. The control-flow move is inert and better-defended than you argued. The catch-all removal is the change I would have asked for. You flagged your own weakest evidence in the PR body without being asked, and it turned out to be weaker than you said — which is why finding it took me an afternoon instead of a production incident.

Three follow-ups filed or extended: #1297 (the count pin), #1298 (the wording overclaim), #1273 (scoping evidence added).

---

Requestor: Nadia Khoury
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1297, #1298

Issue #1291, opened 2026-08-03
tech-debt: validate_matrix_names splits ONE membership failure across two blocks, and the roster_union_override is reachable only for parent-carded personas


parametrization:
Found while implementing #1182 (PR against `validate_matrix_names.py`). Not a blocker for that PR — filed as follow-up.

## Symptom

**One operator problem — "this persona is not on the target repo's roster, and the slot must commit there" — is reported in two entirely separate blocks with two different remediation menus.** Which block you land in is decided by an incidental fact: whether the persona also happens to have a card in the PARENT's `.claude/team/roster/` directory.

Measured against the real org roster (both target repos cloned), post-#1182 head:

```
validate_matrix_names: 1/2 names UNRESOLVED across 2 entries.

  noorinalabs-isnad-graph:
    - implementer: 'Nikolaos Papadopoulos'  →  KNOWN org persona (present in .claude/team/roster.json), but NOT on this repo's roster.
      repo roster: Aisling Brennan, Anya Kowalczyk, ...

  1 of the above resolve as KNOWN org personas but are NOT on
  the target repo's roster, so a commit-capable slot cannot take them
  (#1134 — that slot must COMMIT there). This IS an assignment problem:
    (a) reassign to a member of the target repo's roster (preferred), or
    (b) onboard the persona into <repo>/.claude/team/roster/ + roster.json.
  If the assignment is genuinely intended, record it under wave_{M}_decisions.implementer_substitutions
  in cross-repo-status.json with a rationale.

validate_matrix_names: 1/2 CROSS-REPO IMPLEMENTER assignment(s) with no recorded override (#1134).

  noorinalabs-user-service:
    - implementer: 'Nadia Khoury' is NOT on this repo's roster.
      repo roster: Anya Kowalczyk, Idris Yusuf, Mateo Salazar, Nadia Boukhari

  Pick one:
    (a) reassign to a member of the target repo's roster (preferred), or
    (b) onboard the persona into <repo>/.claude/team/roster/ + roster.json, or
    (c) record an explicit override on the scope row and re-run:
          "roster_union_override": {"rationale": "..."}
```

`Nadia Khoury` is parent-carded, so she RESOLVES and lands in the `#1134` membership block. `Nikolaos Papadopoulos` is manifest-only (his cards live in a third child repo), so a commit-capable slot cannot resolve him at all and he lands in the UNRESOLVED block. The failure is identical from the operator's point of view.

[...]

Issue #1289, opened 2026-08-03
validate_matrix_names: unclassified agentive slot key can co-emit a misleading #1134 cross-repo-implementer message


parametrization:
An **admitted-but-unclassified agentive slot key** (e.g. a future metadata field like `risk_owner`, `exec_sponsor`, `code_owner`, `bug_reporter` — all match `_ROLE_SLOT_KEY_RE` and are explicitly meant to be admitted per `test_unclassified_agentive_keys_are_recognised`) is validated on the strict/commit-capable path in `validate()` **before** it is known to be a real role slot at all. If its declared value happens to be a real person on the parent/org roster who is not on the *target repo's* roster (plausible: a TPM/coordinator name is exactly the kind of value a "who owns tracking this" metadata field would carry), the run reports it as `#1134 CROSS-REPO IMPLEMENTER assignment(s) with no recorded override` — a message that reads as a genuine reassignment/onboarding defect, with the (a)/(b)/(c) remediation menu (reassign / onboard / override) — *alongside* the correct `#1180 UNCLASSIFIED role slot(s)` message.

Measured (isolated from #1180's own PR at `aee3d94`, run via `.claude/skills/wave-scope/validate_matrix_names.py --scope ... --wave N`):

```
{"noorinalabs-user-service": {"implementer": "Nikolaos Papadopoulos", "reviewer": "Aino Virtanen", "risk_owner": "Santiago Ferreira"}}
```

exits 1 with BOTH:
```
validate_matrix_names: 1/3 CROSS-REPO IMPLEMENTER assignment(s) with no recorded override (#1134).
  - risk_owner: 'Santiago Ferreira' is NOT on this repo's roster.
  Pick one: (a) reassign ... (b) onboard ... (c) record an override ...

validate_matrix_names: 1/3 UNCLASSIFIED role slot(s) (#1180).
  - risk_owner: 'Santiago Ferreira' — slot key is in neither COMMIT_CAPABLE_ROLES nor REVIEW_CLASS_ROLES.
```

The `#1180` message is correct and actionable (classify the key in one of three frozensets), but the co-emitted `#1134` message is a **false signal** — the row was never actually established as a commit-capable slot, so telling the operator to reassign/onboard/override a real person is presumptuous and adds noise to what is otherwise a one-line fix (add the key to `NON_ROLE_ROW_KEYS`, or classify it properly).

[...]

PR #1293, opened 2026-08-03
fix(#1182): source suggestions from the org manifest; distinct diagnostic for an uncloned target repo

### LINKED ISSUE #1182: tech-debt: implementer slot on an uncloned repo reports '(no close matches)' for a name the validator holds in the org manifest
Found during review of #1178 (fix for #1162), head `256f79a0`. Not a blocker for that PR — filed as follow-up.

## Symptom

When the target child repo is **not cloned** beside the parent, `validate_matrix_names.py` reports a genuine member of that repo's own roster as UNRESOLVED with `(no close matches)` — a statement the script can prove false from data it is holding in the same call frame.

Measured against the real org manifest, `noorinalabs-user-service` deliberately absent, PR-#1178 code:

```
real noorinalabs-user-service roster member: Anya Kowalczyk
in org manifest?  True

validate_matrix_names: 1/2 names UNRESOLVED across 1 entries.
  noorinalabs-user-service:
    - implementer: 'Anya Kowalczyk'  ->  suggestions: (no close matches)

  implementer  resolved=False suggestions=[]
  reviewer     resolved=True  suggestions=-
  exit code: 1
```

Two problems in one report:

1. `(no close matches)` is wrong. `Anya Kowalczyk` is an exact key in the `org_manifest` set the same `validate()` call already loaded. Suggestions for a commit-capable slot are drawn from `combined`, so the manifest is invisible to `_suggest`.
2. The **same name, same run** resolves in the `reviewer` slot and not in the `implementer` slot. An operator reading this cannot tell whether the persona exists.

The printed remediation on this path steers the wrong way — "Resolve each unknown name before /wave-kickoff fan-out" and "record an implementer_substitution" — i.e. toward changing a *correct* assignment. The actual remedy is `--fetch-missing` or cloning the repo, which is only mentioned on the `unverified` path, which this row never reaches.

## Why it is not a #1178 blocker

#1178 does not introduce the block — the base validator also exits 1 here. What #1178 adds is the self-contradiction, because the reviewer half now resolves. The resolution semantics in #1178 are correct and deliberately must not change (see `test_manifest_does_not_pass_implementer_on_uncloned_repo`; independently reproduced — the over-broad variant flips this exact case from `EXIT=1` to `resolved=True membership=unverified EXIT=0`).

## Proposed fix — message-only, semantically inert

`suggestions` is advisory text; it never feeds `resolved` or the exit code. So the fix does not touch the carve-out:

- source suggestions for unresolved commit-capable slots from `review_combined` rather than `combined`;
- when `declared_clean` is in `org_manifest` but the target roster is unreadable, emit a distinct line — known org persona, target repo roster unreadable (not cloned?), re-run with `--fetch-missing` or clone the repo — instead of routing the operator to the substitution guidance.

`resolved` stays `False` and the exit code stays 1 in every case, so #1134's guarantee is untouched.

## Acceptance

- A member of an uncloned target repo in the `implementer` slot still yields `resolved=False` and exit 1.
- Its diagnostic names the manifest match and points at `--fetch-missing` / cloning, not at recording a substitution.
- A test pins that `resolved` and the exit code are unchanged by the message change.

parametrization:
Closes #1182.

`validate_matrix_names.py` reported a genuine member of an uncloned repo's roster as UNRESOLVED with `(no close matches)` — an assertion it could disprove from a local in the same call frame. Suggestions for a commit-capable slot were drawn from `combined` (the deliberately narrow resolution set), so the `org_manifest` set `validate()` had already loaded was invisible to `_suggest`.

## The bug, measured

Real org roster data, `noorinalabs-user-service` deliberately absent from the org dir, `org_dir` passed explicitly so `_find_org_dir()`'s walk-up-from-cwd cannot produce a false negative. Reproducer run at `77dc9af` (= `origin/main` at branch point, i.e. **after** #1180/PR #1277):

```
real noorinalabs-user-service roster member: Anya Kowalczyk
in org manifest?  True   (manifest personas: 76)
noorinalabs-user-service cloned in scratch org? False
target repo roster read: []

validate_matrix_names: 1/2 names UNRESOLVED across 1 entries.

  noorinalabs-user-service:
    - implementer: 'Anya Kowalczyk'  →  suggestions: (no close matches)

  Resolve each unknown name before /wave-kickoff fan-out.
  Approved substitutions: record under wave_{M}_decisions.implementer_substitutions in cross-repo-status.json with rationale.

  implementer resolved=False  membership='n/a' suggestions=[]
  reviewer    resolved=True   membership='n/a' suggestions=None
  exit code: 1
```

## After

```
validate_matrix_names: 1/2 names UNRESOLVED across 1 entries.

  noorinalabs-user-service:
    - implementer: 'Anya Kowalczyk'  →  KNOWN org persona (present in .claude/team/roster.json); this repo's roster could not be read — not cloned?

  1 unresolved row(s) above name a KNOWN org persona on a repo
  whose roster could not be read. That is an ENVIRONMENT gap, not a bad
  assignment — do NOT record an implementer_substitution for it, and note
  that the same name DOES resolve in a review-class slot (the org-union
  manifest widens review-class resolution only, #1162).

  Pick one and re-run:
    (a) --fetch-missing — resolve the target roster over the network, or
    (b) clone the repo beside the parent.

  It stays a failure (exit 1) until then, deliberately: a commit-capable
  slot passes only on a name the TARGET repo can vouch for (#1134), and
  membership cannot be decided from an unreadable roster.

  implementer resolved=False  membership='n/a' suggestions=['Anya Kowalczyk'] unresolved_reason='org-persona-unreadable-roster'
  reviewer    resolved=True   membership='n/a' suggestions=None
  exit code: 1
```

## Proof the change is semantically inert

Same scenario, pre-change tree exported with `git archive origin/main` (never `git --work-tree` against the shared checkout), post-change tree at branch HEAD:

| | `resolved` | `membership` | exit |
|---|---|---|---|
| BEFORE (`origin/main` `77dc9af`) | `False` | `n/a` | **1** |
| AFTER (branch HEAD `6b1bef4`) | `False` | `n/a` | **1** |

The only report-shape deltas are the two advisory keys (`suggestions`, `unresolved_reason`) — nothing reads them, and neither feeds the exit-code arithmetic, which keys off `resolved`.

`test_1182_message_change_is_semantically_inert` pins this directly: it strips the advisory keys and asserts the remaining entry is byte-identical to the pre-#1182 shape, plus exit 1. The over-broad variant this guards against — widening the RESOLUTION set instead of the suggestion set — was measured on the same live data and flips this exact row to:

```
  implementer resolved=True   membership='unverified'
  exit code: 0
```

which is the silent pass #1134 exists to stop. `test_manifest_does_not_pass_implementer_on_uncloned_repo` and `test_third_child_reviewer_resolves_and_implementer_stays_gated` both go red under it (see M4 below), so the carve-out is pinned from three directions.

## A second, self-inflicted bug found and fixed mid-change

Widening the suggestion source fixed the lie but created a new one on the adjacent case. `difflib.get_close_matches` returns the declared name ITSELF on an exact manifest hit, so a known org persona scoped onto a repo that **is** cloned printed:

```
    - implementer: 'Nikolaos Papadopoulos'  →  suggestions: Nikolaos Papadopoulos
```

Measured on the real org dir while checking the #1289 interaction. That row now states the finding and names the roster the slot could actually take:

```
    - implementer: 'Nikolaos Papadopoulos'  →  KNOWN org persona (present in .claude/team/roster.json), but NOT on this repo's roster.
      repo roster: Anya Kowalczyk, Idris Yusuf, Mateo Salazar, Nadia Boukhari
```

Hence three `unresolved_reason` values, not two. Unlike the uncloned case this one **is** a genuine assignment problem, so it keeps the reassign / onboard / recorded-substitution guidance and deliberately does not mention `--fetch-missing`.

## Mutation table

15 mutants, each applied to the committed validator, `pytest tests/ -q` run from `.claude/skills/wave-scope` (the way `pytest-skills` runs it in CI), then restored with `git checkout HEAD -- <path>`. Baseline and post-restore both `71 passed, 38 subtests passed`.

| # | Mutation | Result | Killed by |
|---|---|---|---|
| M1 | suggestion source reverted to the narrow `combined` (the #1182 bug) | KILLED | `test_no_close_matches_is_not_claimed_about_a_manifest_name`, `test_genuine_typo_on_an_uncloned_repo_still_reports_unknown_name` |
| M2 | `elif membership_decidable` collapsed to always-true | KILLED | `test_1182_message_change_is_semantically_inert`, `test_uncloned_org_persona_is_steered_to_fetch_missing_not_substitution`, `test_scope_mode_carries_the_diagnostic_end_to_end`, +2 |
| M3 | `if not in_manifest` collapsed to never | KILLED | `test_genuine_typo_on_an_uncloned_repo_still_reports_unknown_name`, `test_all_three_unresolved_classes_print_their_own_remediation` |
| M4 | **the over-broad variant**: widen the RESOLUTION set instead of the suggestion set | KILLED | `test_1182_message_change_is_semantically_inert`, `OrgUnionManifestTests::test_manifest_does_not_pass_implementer_on_uncloned_repo`, `OrgUnionManifestTests::test_third_child_reviewer_resolves_and_implementer_stays_gated`, `UnclassifiedRoleSlotTests::test_unclassified_slot_resolves_against_the_narrow_set`, +5 |
| M5 | unknown-name trailer printed unconditionally | KILLED | `test_uncloned_org_persona_is_steered_to_fetch_missing_not_substitution` |
| M6 | `--fetch-missing` trailer printed unconditionally | KILLED | `test_manifest_persona_on_a_CLONED_repo_still_gets_substitution_guidance` |
| M7 | distinct uncloned per-row line removed | KILLED | `test_no_close_matches_is_not_claimed_about_a_manifest_name` |
| M8 | unknown-name and `--fetch-missing` trailers made mutually exclusive | KILLED | `test_all_three_unresolved_classes_print_their_own_remediation` |
| M9 | `unresolved_reason` stamped on RESOLVED rows too | KILLED | `test_resolved_rows_carry_no_unresolved_reason` |
| M10 | uncloned branch relabelled `unknown-name` | KILLED | `test_no_close_matches_is_not_claimed_about_a_manifest_name`, `test_scope_mode_carries_the_diagnostic_end_to_end`, +3 |
| M11 | `continue` after the uncloned row dropped (row printed twice) | KILLED | `test_no_close_matches_is_not_claimed_about_a_manifest_name` |
| M12 | not-a-member per-row line removed (falls through to the self-referential suggestion) | KILLED | `test_manifest_persona_on_a_CLONED_repo_still_gets_substitution_guidance` |
| M13 | not-a-member trailer suppressed | KILLED | `test_manifest_persona_on_a_CLONED_repo_still_gets_substitution_guidance`, `test_all_three_unresolved_classes_print_their_own_remediation` |
| M14 | not-a-member row no longer names the roster it could take | KILLED | `test_manifest_persona_on_a_CLONED_repo_still_gets_substitution_guidance` |
| M15 | not-a-member trailer chained after unknown-name | KILLED | `test_all_three_unresolved_classes_print_their_own_remediation` |

**Two mutants survived the first drafts and are the reason two of the tests exist:**

- **M7 survived draft 1.** The remediation trailer also contains the phrase `KNOWN org persona`, so a whole-stream `assertIn` stayed green with the per-row line deleted — the assertion was vacuous. Fixed by `_finding_row()`, which extracts the single per-finding line for a role and raises unless there is exactly one, so row-level claims are checked against the row. (That helper is also what kills M11 and M12.)
- **M8 survived draft 2.** With only ONE of the two later counters non-zero, an `elif` still executes, so a two-class run cannot distinguish three independent `if`s from a chain. `test_all_three_unresolved_classes_print_their_own_remediation` carries all three unresolved classes at once and kills M8 and M15.

## Interaction with #1289

**Measured, not asserted.** #1289's exact matrix (`implementer` / `reviewer` / `risk_owner: "Santiago Ferreira"`) run against the real org dir on both trees:

```
BEFORE EXIT=1
       validate_matrix_names: 1/3 names UNRESOLVED across 1 entries.
       validate_matrix_names: 1/3 CROSS-REPO IMPLEMENTER assignment(s) with no recorded override (#1134).
       validate_matrix_names: 1/3 UNCLASSIFIED role slot(s) (#1180).
AFTER  EXIT=1
       validate_matrix_names: 1/3 names UNRESOLVED across 1 entries.
       validate_matrix_names: 1/3 CROSS-REPO IMPLEMENTER assignment(s) with no recorded override (#1134).
       validate_matrix_names: 1/3 UNCLASSIFIED role slot(s) (#1180).
```

The compound `#1134` + `#1180` message #1289 objects to is **byte-identical** across the change; the only stderr delta in that scenario is the `implementer` row inside the UNRESOLVED block. That is structural, not luck: `risk_owner` **resolves** (Santiago is on the parent cards), so it goes down the membership path, while everything this PR touches lives on the `if not resolved:` path and in the UNRESOLVED print block. **This PR neither fixes nor worsens #1289**, and does not conflict with its candidate fix (skip the membership check when `slot_class == "unclassified"`), which is in the other branch.

## Gates

All run locally from the org root before pushing.

- `pytest` (wave-scope, CI-style from the skill dir): `71 passed, 38 subtests passed`
- `pytest .claude/hooks/tests/ .claude/lib/tests/`: `3641 passed, 651 subtests passed`
- every skill suite: promotion-audit 102, wave-kickoff 8, wave-scope 71, wave-wrapup 15 — `OVERALL=0`
- `ruff check .claude/skills/wave-scope/`: `All checks passed!` · `ruff format`: clean
- `mypy .claude/skills/wave-scope` and `mypy .claude/hooks/ .claude/lib/`: no errors
- `python3 .claude/lib/pre_commit_ci_sync.py .`: `OK: pre-commit config mirrors all CI-enforced checks.`
- full pre-push hook set passed on `git push` (mypy, mypy-skills, pytest, pytest-skills, memory-budget, headcount-budget, doc-freshness, office-drift, mermaid-render)

## Follow-up debt

- #1291 — the same membership failure is split across the UNRESOLVED block and the `#1134` CROSS-REPO block depending on whether the persona is parent-carded, and `roster_union_override` is consequently reachable only for parent-carded personas (`tech-debt` / `process` / `phase-10`, no wave label). Measured while implementing this PR.

## Reviewers

Two per charter. Not for merge by me — the orchestrator drives the review gate.

### DIFF
--- a/.claude/skills/wave-scope/tests/test_validate_matrix_names.py
+++ b/.claude/skills/wave-scope/tests/test_validate_matrix_names.py
@@ -1348,22 +1348,33 @@ class ManifestSourcedSuggestionTests(unittest.TestCase):
             self.assertEqual(_report_stderr(report)[0], 1)
 
     def test_manifest_persona_on_a_CLONED_repo_still_gets_substitution_guidance(self):
-        """The `not membership_decidable` conjunct is load-bearing.
+        """The `membership_decidable` branch is load-bearing.
 
         Same name, same manifest — but the target roster IS readable, so this is
-        a genuine wrong-assignment and the substitution guidance is correct.
-        Drop the conjunct and this reds on the `--fetch-missing` assertion.
+        a genuine wrong-assignment: reassign / onboard / record a substitution,
+        NOT `--fetch-missing`. Collapse the branch and this reds.
+
+        The row must still not echo the declared name back as its own
+        suggestion, which is exactly what `_suggest` returns on an exact
+        manifest hit — measured against the real org roster on the first draft
+        of this change: `implementer: 'Nikolaos Papadopoulos'  ->  suggestions:
+        Nikolaos Papadopoulos`.
         """
         with tempfile.TemporaryDirectory() as tmpdir:
             org = self._build_org(Path(tmpdir))
             report = validate({self.CLONED_REPO: {"implementer": self.ORG_PERSONA}}, org)
             finding = report[self.CLONED_REPO][0]
             self.assertFalse(finding["resolved"])
-            self.assertEqual(finding["unresolved_reason"], "unknown-name")
+            self.assertEqual(finding["unresolved_reason"], "org-persona-not-a-member")
             code, err = _report_stderr(report)
             self.assertEqual(code, 1)
             self.assertIn("implementer_substitutions", err)
             self.assertNotIn("--fetch-missing", err)
+            row = _finding_row(err, "implementer")
+            self.assertIn("NOT on this repo's roster", row)
+            self.assertNotIn("suggestions:", row)
+            # The roster it COULD take is named, so the fix is one read away.
+            self.assertIn("Bereket Tadesse", err)
 
     def test_genuine_typo_on_an_uncloned_repo_still_reports_unknown_name(self):
         """A misspelling is NOT the environment gap — but it still gets a real suggestion.
--- a/.claude/skills/wave-scope/tests/test_validate_matrix_names.py
+++ b/.claude/skills/wave-scope/tests/test_validate_matrix_names.py
@@ -1411,6 +1411,38 @@ class ManifestSourcedSuggestionTests(unittest.TestCase):
             self.assertIn("implementer_substitutions", err)
             self.assertIn("2/2 names UNRESOLVED", err)
 
+    def test_all_three_unresolved_classes_print_their_own_remediation(self):
+        """All three trailers are independent `if`s, not one chained decision.
+
+        Two classes is not enough to catch chaining: with only one of the two
+        LATER counters non-zero, an `elif` still runs. Only a run carrying all
+        three distinguishes independent `if`s from a chain — measured, the
+        two-class test left the `elif org_persona_unreadable` mutant ALIVE.
+        """
+        with tempfile.TemporaryDirectory() as tmpdir:
+            org = self._build_org(Path(tmpdir))
+            report = validate(
+                {
+                    self.UNCLONED_REPO: {"implementer": self.ORG_PERSONA},  # unreadable
+                    self.CLONED_REPO: {"implementer": self.ORG_PERSONA},  # not-a-member
+                    "noorinalabs-isnad-graph": {"implementer": "Totally Unknown"},
+                },
+                org,
+            )
+            reasons = sorted(
+                f["unresolved_reason"] for fs in report.values() for f in fs if not f["resolved"]
+            )
+            self.assertEqual(
+                reasons,
+                ["org-persona-not-a-member", "org-persona-unreadable-roster", "unknown-name"],
+            )
+            code, err = _report_stderr(report)
+            self.assertEqual(code, 1)
+            self.assertIn("Resolve each unknown name", err)
+            self.assertIn("resolve as KNOWN org personas but are NOT on", err)
+            self.assertIn("--fetch-missing", err)
+            self.assertIn("3/3 names UNRESOLVED", err)
+
     def test_scope_mode_carries_the_diagnostic_end_to_end(self):
         """The path /wave-scope § 12.5 and /wave-kickoff § 0b actually run."""
         with tempfile.TemporaryDirectory() as tmpdir:
--- a/.claude/skills/wave-scope/tests/test_validate_matrix_names.py
+++ b/.claude/skills/wave-scope/tests/test_validate_matrix_names.py
@@ -81,6 +81,20 @@ def _report_stderr(report: dict) -> tuple[int, str]:
     return code, buf.getvalue()
 
 
+def _finding_row(err: str, role: str) -> str:
+    """Return the single per-finding line for `role` from captured stderr (#1182).
+
+    Whole-stream `assertIn` is not enough here: the remediation TRAILERS repeat
+    the same phrases the per-row line uses, so a stream-level assertion stays
+    green with the row line deleted. Tests that care about what the row says
+    must assert against the row.
+    """
+    rows = [ln for ln in err.splitlines() if ln.strip().startswith(f"- {role}:")]
+    if len(rows) != 1:
+        raise AssertionError(f"expected exactly 1 {role!r} finding row, got {len(rows)}:\n{err}")
+    return rows[0]
+
+
 def _write_slim_roster_card(roster_dir: Path, role_slug: str, name: str) -> None:
     """Mimic the NEW slim card shape (#1010): H1 `# <Name> — <Role>`, no
     `**Name:**` field. Verbatim to the shipped template so the fixture can't
@@ -1284,7 +1298,14 @@ class ManifestSourcedSuggestionTests(unittest.TestCase):
             code, err = _report_stderr(report)
             self.assertEqual(code, 1)
             self.assertNotIn("(no close matches)", err)
-            self.assertIn("KNOWN org persona", err)
+            # Assert on the ROW, not on the whole stream: the trailer paragraph
+            # also contains "KNOWN org persona", so a whole-stream assertion
+            # passes even with the per-row line deleted (measured — that
+            # mutation SURVIVED the first draft of this test).
+            row = _finding_row(err, "implementer")
+            self.assertIn("KNOWN org persona", row)
+            # And it must not echo the declared name back as its own suggestion.
+            self.assertNotIn("suggestions:", row)
 
     def test_uncloned_org_persona_is_steered_to_fetch_missing_not_substitution(self):
         """Bullet 2: the remediation must not point at changing a correct assignment."""
--- a/.claude/skills/wave-scope/tests/test_validate_matrix_names.py
+++ b/.claude/skills/wave-scope/tests/test_validate_matrix_names.py
@@ -1439,7 +1439,7 @@ class ManifestSourcedSuggestionTests(unittest.TestCase):
             code, err = _report_stderr(report)
             self.assertEqual(code, 1)
             self.assertIn("Resolve each unknown name", err)
-            self.assertIn("resolve as KNOWN org personas but are NOT on", err)
+            self.assertIn("a commit-capable slot cannot take them", err)
             self.assertIn("--fetch-missing", err)
             self.assertIn("3/3 names UNRESOLVED", err)
 
--- a/.claude/skills/wave-scope/tests/test_validate_matrix_names.py
+++ b/.claude/skills/wave-scope/tests/test_validate_matrix_names.py
@@ -11,6 +11,8 @@ org-dir state (which changes wave-to-wave).
 
 from __future__ import annotations
 
+import contextlib
+import io
 import json
 import sys
 import tempfile
@@ -64,6 +66,19 @@ def _build_fake_org_dir(tmp: Path) -> Path:
     _write_roster_card(graph_roster, "eng_idris", "Idris Yusuf")
     _write_roster_card(graph_roster, "eng_marisol", "Marisol Vega-Cruz")
     return tmp
+
+
+def _report_stderr(report: dict) -> tuple[int, str]:
+    """Run `_print_report_to_stderr` capturing its output (#1182).
+
+    Returns `(exit_code, captured_stderr)` so a test can assert on the exit
+    code AND on the operator-facing text in one call, instead of asserting on
+    report internals and merely hoping they reach the terminal.
+    """
+    buf = io.StringIO()
+    with contextlib.redirect_stderr(buf):
+        code = _print_report_to_stderr(report)
+    return code, buf.getvalue()
 
 
 def _write_slim_roster_card(roster_dir: Path, role_slug: str, name: str) -> None:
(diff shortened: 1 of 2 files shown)


parametrization:
## Merge-gate review (Opus) — #1293 @ `6b1bef471ff64c77aec2a8992aae9c776176c502`

Reviewed against `origin/main` at the head SHA, file contents pulled via `gh api contents?ref=6b1bef4` and confirmed byte-identical (`md5 ff9d16a8…`) to the ref I ran. Both trees exported with `git archive` into scratch; the shared checkout was never touched. **CI: 14/14 SUCCESS at the head SHA**, re-confirmed immediately before posting.

I re-ran the load-bearing claims rather than reading them. Everything I checked held, and in one case held more precisely than the PR body states.

### The central claim: message-only and semantically inert — CONFIRMED

Reproduced independently on real org roster data, `noorinalabs-user-service` deliberately absent from a constructed org dir, `org_dir` passed **explicitly** so the walk-up cannot produce a false negative. I first proved the premises rather than assuming them: `Anya Kowalczyk` is in the manifest (76 personas) **and** is a genuine member of the real `noorinalabs-user-service` roster, and the target roster reads `[]` in the scratch org.

| | `resolved` | `membership` | exit |
|---|---|---|---|
| BEFORE (`77dc9af`) | `False` | `n/a` | **1** |
| AFTER (`6b1bef4`) | `False` | `n/a` | **1** |

The pre-change run reproduces the bug exactly as reported — `suggestions: (no close matches)` printed about a name held in the same call frame, with the same name resolving in the `reviewer` slot one line below.

I did not take "nothing reads `suggestions`" on trust. Across tracked first-party files there is no consumer of `suggestions` / `unresolved_reason` / `repo_roster` outside the validator and its tests (the one hit in `wave-scope/SKILL.md` is a prose table row). The exit arithmetic is `if unresolved or cross_repo or unclassified: return 1` — the three new counters partition `unresolved` and feed only trailer selection, never the return. `repo_roster` on an unresolved row cannot leak into the `#1134` block, which filters on `membership == "cross-repo"` while unresolved rows carry `"n/a"`.

### M4 — re-run myself: KILLED

Widening the resolution set (`candidates = review_combined` unconditionally) fails **10** tests, including all three named:

- `OrgUnionManifestTests::test_manifest_does_not_pass_implementer_on_uncloned_repo`
- `OrgUnionManifestTests::test_third_child_reviewer_resolves_and_implementer_stays_gated`
- `UnclassifiedRoleSlotTests::test_unclassified_slot_resolves_against_the_narrow_set`

Baseline and post-restore both `71 passed, 38 subtests passed`. The boundary between an advisory fix and a gate relaxation is genuinely pinned.

### The two survivor-response tests — both genuinely fail under their mutants

**M7.** Deleting the distinct uncloned per-row line reds exactly `test_no_close_matches_is_not_claimed_about_a_manifest_name`. I then verified the vacuity claim directly rather than inferring it: under M7 I captured the stream and checked what a whole-stream assertion *would* have done — the trailer still contains `KNOWN org persona`, so `assertIn` over the stream is **GREEN while the row is gone**, and the row itself degrades to `suggestions: Nikolaos Papadopoulos`. `_finding_row()` is load-bearing, not decorative.

**M8.** Worth recording precisely, because the PR body is looser than the code's own docstring. I ran two formulations:

- Chaining `unknown_name` → `org_persona_unreadable` (which requires reordering the blocks) is killed by **two** tests, including the two-class one.
- The **faithful minimal** mutation — the third `if` becoming an `elif` of the second, no reorder, which is what "with only one of the two LATER counters non-zero an `elif` still executes" actually describes — is killed by **exactly one test in the whole suite**: `test_all_three_unresolved_classes_print_their_own_remediation`. It survives `test_both_remediations_print_when_both_classes_are_present`.

So the three-class test is not redundant with the two-class one; it is the sole killer. The docstring's account is exact, and the test earns its place.

### Ruling: three reason classes is the right factoring

Keep all three. The test is not "are these three states distinguishable" but "do they have different correct operator actions", and here the wrong pairing is actively harmful in one direction: applying the uncloned remedy to a not-a-member row is a wasted network call, but applying the not-a-member remedy to an uncloned row means **recording a substitution against a correct assignment** — the original #1182 harm, not a lesser version of it. Those two cannot share a paragraph. `unknown-name` is separately distinct because it is the only class where suggestions carry information; for the other two `_suggest` returns the declared name itself.

Collapsing the two org-persona classes behind a conditional sentence would not remove the branch, only move it from a named constant into string interpolation — strictly worse to test, and M8's survival history is precisely the evidence that trailer-selection logic fails silently. The one genuine consolidation candidate is folding `org-persona-not-a-member` into the `#1134` block, which is #1291 direction 1 and correctly deferred.

### #1289 — confirmed, and structural rather than lucky

Ran #1289's exact matrix against the real org dir on both trees. `EXIT=1` both. The full diff of stderr is confined to the `implementer` row inside the UNRESOLVED block and its trailer; the `#1134` and `#1180` blocks are byte-identical. I checked those blocks are **present** in both outputs rather than accepting a diff that is empty because the content is absent — both carry the `risk_owner: 'Santiago Ferreira'` row in each block. The mechanism is as described: `risk_owner` resolves, so it travels the membership path, while this PR lives entirely on the `if not resolved:` path.

### #1291 — deferring is right, and the asymmetry is pre-existing

Confirmed empirically, not just structurally. Same recorded `roster_union_override` with a rationale, same target repo:

```
manifest-only persona   -> resolved=False  membership='n/a'          EXIT=1   override NOT honoured
parent-carded persona   -> resolved=True   membership='overridden'   EXIT=0   override honoured
```

The charter's recorded escape hatch really is unreachable for a manifest-only persona. Critically, I ran the same probe on the **base** tree and got identical results — the asymmetry is pre-existing, and this PR neither introduces nor widens it; the more accurate message is what made it visible. Reaching `overridden` from an unresolved row means letting an override widen the resolution set, which is the exact carve-out M4 protects. Deferring to #1291 is the correct call, and I would not have accepted it being fixed here.

### One new item, filed as #1301

A mutant outside the PR's table survives the full suite. `in_manifest` compares case-insensitively, consistent with how `resolved` is computed; making it case-**sensitive** leaves all 71 tests green. Measured on a case variant of a genuine manifest persona:

```
SHIPPED  'anya kowalczyk' -> KNOWN org persona ... not cloned?   reason='org-persona-unreadable-roster'  EXIT=1
MUTANT   'anya kowalczyk' -> suggestions: Anya Kowalczyk         reason='unknown-name'                   EXIT=1
```

The shipped behaviour is correct and the exit code is unchanged either way, so this is not a gate weakening and not a must-fix. But the mutant regresses the message to substitution-steering against a correct assignment — this PR's own thesis — through an input class (hand-transcribed case drift) the matrix format makes plausible, with nothing going red. Filed as #1301 with a one-test remedy.

I also mutation-tested whether the not-a-member class has semantic pinning equivalent to the uncloned class, since `test_1182_message_change_is_semantically_inert` only covers the latter and the former gains a third key. Setting `membership` on the not-a-member path is killed by `test_third_child_reviewer_resolves_and_implementer_stays_gated`. The coverage exists, distributed across the #1162 tests. Not a finding.

### Nit, non-blocking

`validate_matrix_names.py:648` — the new comment reads "target roster READABLE and **he** is not on it" about a generic, unnamed org persona. `they` is the right default. Worth folding into the next touch of this file rather than respinning a green PR for it.

### Note for the next reviewer on this repo

`gh project item-list --limit N` caps its result set and reports a **false absence** — my board add for #1301 appeared missing until I confirmed it with a direct `node(id:…)` query. Same silent-zero family as `gh pr checks --json` returning empty in this build, and as a bare `rg` over `.claude/`. Verify board membership by node, not by list.

### Verdict

The claim that had to hold, holds — measured at both ends, on real data, with the premises proved rather than assumed. The guards are non-vacuous under their own mutants, and the two tests written in response to survivors both genuinely fail under them. The two residuals are correctly triaged as follow-ups rather than smuggled into a message-only change. Approving.

---

Requestor: Aino Virtanen
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: #1301

parametrization:
## Second review — Lucas Ferreira (SRE Engineer)

Reviewed at head `6b1bef471ff64c77aec2a8992aae9c776176c502` (14/14 CI green). Three-dot diff (`git diff origin/main...origin/N.Khoury/1182-suggestions-from-manifest`) confirms the change is confined to `validate_matrix_names.py`'s `if not resolved:` branch and `_print_report_to_stderr`; the `#1134` membership block a few lines below (`if not is_review_class and not is_parent_repo: ...`) is byte-identical between trees — same line-for-line content, only its offset shifted (572→658) from the new docstring/constant insertions above it.

**#1289 interaction — confirmed non-interacting, independently.** Ran #1289's own matrix (`implementer`/`reviewer`/`risk_owner: "Santiago Ferreira"`) against the real org tree at both `origin/main` and the branch HEAD via `git archive` exports (no `--work-tree` against the shared index). Stderr is byte-identical on both trees — same `#1134` CROSS-REPO + `#1180` UNCLASSIFIED compound message, same exit 1. This is structural, not coincidental: `risk_owner` *resolves* (Santiago is parent-carded, so it's in `combined` before the `if not resolved:` branch this PR touches is ever reached) and falls through to the untouched membership check. Everything #1293 adds lives strictly inside `if not resolved:`, which `risk_owner` never enters. The #1289 candidate fix (guard the membership check on `slot_class == "unclassified"`) sits in that same untouched block, confirmed line-for-line unchanged, so it applies cleanly on top of this PR.

**Aino's carve-out ruling — agree.** Crossing the two org-persona remedies would be actively harmful: applying the not-a-member remedy (reassign/onboard/substitution) to an uncloned row means recording a substitution against a name that may well be correct — reproducing the original #1182 harm rather than softening it. Keeping the reasons as three independent counters (not an `elif` chain) is also correctly load-bearing here, not just in the mutation table — verified below.

**Ran all three `unresolved_reason` classes together in one report, as an operator would see it** (scratch org: `noorinalabs-user-service` cloned with an org persona not on its roster, `noorinalabs-isnad-graph` absent entirely, and a plain typo'd name against neither):

```
noorinalabs-user-service:
  - implementer: 'Nikolaos Papadopoulos'  →  KNOWN org persona (present in .claude/team/roster.json), but NOT on this repo's roster.
    repo roster: Anya Kowalczyk, Idris Yusuf, Mateo Salazar, Nadia Boukhari

noorinalabs-isnad-graph:
  - implementer: 'Anya Kowalczyk'  →  KNOWN org persona (present in .claude/team/roster.json); this repo's roster could not be read — not cloned?

noorinalabs-deploy:
  - implementer: 'Zzzy Qwerty Totally Fake'  →  suggestions: (no close matches)
```
followed by three independently-gated trailers (generic unknown-name, not-a-member with reassign/onboard/override guidance, unreadable with `--fetch-missing`/clone guidance). They read as clearly distinguishable in situ, not just in the test assertions: each per-row line states its own finding inline, the not-a-member trailer never mentions `--fetch-missing`, the unreadable trailer never mentions substitution/reassignment, and confirmed with a single-class run that the generic trailer is correctly suppressed when `unknown_name == 0` (no leftover boilerplate bleeding into an all-org-persona report). No contradictory pair of remedies was observed in any combination I ran.

The one soft note, not a blocker: the trailers refer to their rows as "N unresolved row(s) above" rather than by name/repo — fine at today's scale since each row is already fully self-describing, but could get harder to eyeball-match in a report with many repos and mixed reason classes. Not worth gating on.

Full local suite re-run on the head tree (`.claude/skills/wave-scope`, `ENVIRONMENT=test python3 -m pytest tests/ -q`): `71 passed, 38 subtests passed`.

Confirmed rather than duplicated: #1289 (this review's assigned finding — non-interacting), #1291 (pre-existing, Aino), #1301 (Aino's case-sensitivity gap, filed against this PR's own thesis), #1278 (pre-existing scope-mode residual). Nothing new to file — Aino's mutation coverage (M1–M15) and inertness proof hold up under an independent re-run of the highest-risk scenarios.

---
Requestor: Lucas Ferreira
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: none

Issue #1271, opened 2026-08-03
tech-debt(lib): a THIRD hand-rolled branch-prefix regex — apply_implementor_labels.BRANCH_RE is slash-only and outside charter_trailer


parametrization:
Found while consolidating the two hook copies of `extract_branch_author_lastname` into `charter_trailer` for #1175 (PR #1269) — a repo-wide sweep for the branch-prefix regex family turned up one more parser that #1175 does not close.

## The third copy

```
.claude/lib/charter_trailer.py:263        _BRANCH_AUTHOR_PREFIX_RE = re.compile(r"([A-Za-z])\.([A-Za-z]+)[-/]")
.claude/lib/apply_implementor_labels.py:53  BRANCH_RE            = re.compile(r"^([A-Za-z])\.([A-Za-z][A-Za-z'-]*)/0*([0-9]+)-")
```

`BRANCH_RE` is **slash-only** — exactly the state `validate_review_comment_format` was in for four months before #1175. On a dash-form branch (`A.Virtanen-1175-x`) it does not match, so the implementer label and the issue number are not derived and the PR/branch goes unlabelled. Same failure shape, different surface: a silent skip, not an error.

## Why it is NOT a straight duplicate (and so not folded into #1175)

`BRANCH_RE` answers a **wider** question than `charter_trailer`'s does:

| | captures | requires |
|---|---|---|
| `charter_trailer._BRANCH_AUTHOR_PREFIX_RE` | initial, lastname | separator `-` or `/` |
| `apply_implementor_labels.BRANCH_RE` | initial, lastname, **issue number** | `/`, then an optional-zero-padded number, then `-` |

It also admits apostrophes/hyphens in the lastname (`[A-Za-z][A-Za-z'-]*`) which `charter_trailer` does not. Consolidating them is therefore a real design decision — either `charter_trailer` grows a richer branch-ref parser that returns a structured `(initial, lastname, issue_number | None)` and both callers narrow from it, or the two stay separate with the lastname half delegated. Doing that inside #1175 would have widened a tech-debt fix into a refactor of a labelling tool, so it is filed rather than folded.

## Suggested scope

1. Decide: one structured branch-ref parser in `charter_trailer`, or delegation of the author half only.
[...]

PR #691, opened 2026-06-16
retro(P5W4): retrospective — Trustworthy data & search capstone

parametrization:
## Summary

Phase 5 Wave 4 retrospective — **Trustworthy data & search (capstone)**. The wave's centerpiece, the semantic-search re-embedding capstone (#1071), was delivered + recall-verified on staging via the repeatable `reembed-corpus.yml` mechanism, after the capstone surfaced and the team fixed two latent prod-blocking bugs (embed-image packaging, ssh timeout).

Counters helper-verified `19 / 4 / 16` (no drift). Trust: Weronika 4→5, Linh 3→4, Nurul/Aisha →4; Aino/Mateo hold ceiling; orchestrator hold 4.

## Files changed (all in this diff)
- `.claude/team/feedback_log.md` — P5W4 retro entry (metrics, per-engineer, top-3/top-3, process changes, charter proposal)
- `.claude/team/trust_matrix.md` — P5W4 trust update tables
- `.claude/team/promotion_audit_log/p5-wave-4.md` — promotion audit (0 AUTO memory→charter; section/skill deferred → main#690)
- `cross-repo-status.json` — annunaki + memory audit run-markers
- `.claude/annunaki/errors.jsonl` — wave error capture (session state)

## Follow-ups filed this retro
- main#690 — `/promotion-audit` canonical CLI driver (hand-rolled mis-fire)
- Charter proposal (in feedback_log): promote the determinism-codification reflex to a charter principle — **owner approval required, not applied here**

## Review Checklist
- [ ] Trust deltas match the per-engineer assessments
- [ ] Counters (19/4/16) consistent with the wave report
- [ ] No charter file mutated (charter change is *proposed*, owner-gated)

Co-Authored-By: Nadia Khoury <parametrization+Nadia.Khoury@gmail.com>
Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

PR #1295, opened 2026-08-03
ci(base-pin): cross-repo digest-pin drift sweep (#1205)

### LINKED ISSUE #1205: ci(base-pin): same-tag digest pins drift across repos unbounded — python:3.14-slim is pinned at 3 ages with no reconciliation
## Summary

The same upstream tag, `python:3.14-slim`, is digest-pinned to **different, months-apart digests** in different repos, and nothing reconciles them. Found while merge-gate reviewing noorinalabs-deploy#691.

## Evidence

The integration-test stack builds `fake_oauth` (from `noorinalabs-deploy`) and `isnad-graph-api` (from `noorinalabs-isnad-graph`) **in the same compose run**, from the same tag, on different pins. From the e2e job log at `837c272a19895d1042985eb08562b38aafdc878b` (run 29443732705):

```
#8  [fake_oauth internal]       load metadata for docker.io/library/python:3.14-slim@sha256:d3400aa1...
#13 [isnad-graph-api internal]  load metadata for docker.io/library/python:3.14-slim@sha256:44dd0449...
```

Resolving each pin's amd64 config blob against the registry:

| Consumer | Pin | Base created |
|---|---|---|
| `noorinalabs-deploy` `fake_oauth` (before #691) | `b877e50b...` | 2026-06-24 |
| `noorinalabs-deploy` `fake_oauth` (after #691) | `d3400aa1...` | 2026-07-14 |
| `noorinalabs-isnad-graph` `isnad-graph-api` | `44dd0449...` | **2026-06-11** |
| current `python:3.14-slim` tag | `cea0e604...` | tag moved 2026-07-18 |

`isnad-graph-api`'s pin is ~5 weeks older than deploy's post-#691 pin and roughly 7 weeks behind the current tag. All are Python 3.14.6 — the deltas are Debian trixie package snapshots, i.e. precisely the distro security updates that § Base Image Pinning exists to keep current.

## Why it matters

Digest pinning trades automatic freshness for reproducibility, and Dependabot is what buys the freshness back. That bargain holds per-Dockerfile, but there is no org-level view of *"this tag is pinned in N places, at M different ages."* The result is silent, unbounded skew: a repo whose Dependabot PR sits unmerged drifts arbitrarily far behind a sibling repo consuming the identical tag, and nothing surfaces it. It only became visible here because both images happen to build side by side in one e2e run.

The in-image `apt-get -y upgrade` limb of the convention softens this at *runtime* (each build pulls current packages on top of whatever base it pinned), so this is skew and audit-surface, not a live CVE. But the pin ages are the thing the convention asks us to keep moving, and right now nobody is looking at them together.

## Proposed fix

A periodic sweep (natural fit: the existing scheduled `red-sweep.yml` cadence, or a `/wave-wrapup` step) that:

1. collects every digest-pinned `FROM` across all repos (the structural ontology already indexes Dockerfiles),
2. groups by `image:tag`,
3. resolves each digest's `created` date from the registry,
4. flags any tag where two repos' pins differ by more than a threshold (e.g. 30 days), or where any pin is more than N days behind the tag's current digest.

This is a reporting gate, not a blocking one — the remediation is always "merge the open Dependabot PR", which is a human/merge-queue decision.

Filed from the merge-gate review of https://github.com/noorinalabs/noorinalabs-deploy/pull/691.

parametrization:
## Summary

Closes #1205.

The same upstream tag (e.g. `python:3.14-slim`) is digest-pinned to
different, months-apart digests in different repos, with nothing
reconciling them (found while merge-gate reviewing `noorinalabs-deploy#691`
— `noorinalabs-isnad-graph`'s pin was ~5 weeks older than `deploy`'s
post-#691 pin). `check_dockerfile_base_pin.py` already enforces, per
Dockerfile, that every `FROM` is digest-pinned; it has no cross-repo view.
This PR adds that view as a **periodic reporting sweep — explicitly NOT a
merge-blocking gate**. The remediation for any finding is always "merge the
open Dependabot PR", a human/merge-queue decision.

`.claude/lib/base_pin_drift_sweep.py`:

1. collects every digest-pinned `FROM` across all org repos via `gh api`
   (recursive git-tree + contents, no child-repo checkout — same posture as
   `red_sweep.py`), reusing `check_dockerfile_base_pin.parse_froms` (do not
   hand-roll a second `FROM` parser — #1271 tracks three such regressions
   already in this repo),
2. groups pins by the upstream `registry/repository:tag` they share,
3. resolves each pinned digest's (and each tag's CURRENT digest's)
   linux/amd64 image-config `created` date from the registry,
4. flags a tag where two repos' pins differ by more than a threshold
   (`cross_repo`, default 30d), or where a pin is more than N days behind
   the tag's current digest (`behind_current_tag`, default 30d). Both
   thresholds are CLI-overridable.

## Where it's sited, and why

Wired as its **own scheduled workflow**
(`.github/workflows/base-pin-drift-sweep.yml`, daily cron + manual
dispatch), modeled directly on `red_sweep.py` + `red-sweep.yml`'s pattern —
persist a small JSON verdict to a dedicated lightweight ref
(`refs/meta/base-pin-drift`, distinct from `red-sweep`'s own ref), read
cheaply with a staleness guard (`base_pin_drift_sweep.py check`,
`DEFAULT_MAX_AGE_HOURS=48`). Deliberately **not** folded into `red-sweep.yml`
itself or a `/wave-wrapup` step (full rationale is in the module docstring's
"Home" section):

- `red-sweep.yml`'s 6h cadence is tuned to catch a red publish/deploy run
  before it rots silently for days. Base-image pin AGE moves on the order
  of days-to-weeks (a Dependabot PR sitting unmerged) — 6-hourly resolution
  buys nothing and needlessly spends anonymous Docker Hub pull budget
  shared with real image builds.
- Registry-API auth/rate-limit failure modes (429, `WWW-Authenticate` token
  exchange) are a different surface than the GH Actions API `red_sweep.py`
  already polls; conflating them in one workflow would mean a Docker Hub
  rate-limit only degrades PART of that workflow's verdict.
- `/wave-wrapup` only runs when a wave wraps; this should surface on a
  standing cadence independent of wave cadence.

## Degradation stance (the load-bearing property here)

Per the org's standing lesson that a helper's failure mode must never read
as a healthy zero: **every** registry-resolution failure reports
`status="unknown"` with a `reason`, never "no drift":

| Failure mode | `reason` | Test |
|---|---|---|
| Transport error (timeout/DNS/connection refused) | `network_error` | `test_network_error_is_unknown` |
| Docker Hub anonymous rate limit (429) | `rate_limited` | `test_rate_limited_429_is_unknown_distinct_from_not_found`, `test_config_blob_fetch_failure_propagates_reason` |
| Repo/tag/digest not found (404) | `not_found` | `test_not_found_404_is_unknown` |
| Any other non-200 | `http_<code>` | `test_other_http_status_is_unknown_with_code` |
| Manifest/config body isn't parseable JSON | `malformed` | `test_malformed_manifest_json_is_unknown`, `test_manifest_not_a_json_object_is_unknown` |
| Manifest list has no `linux/amd64` entry | `no_amd64_platform` | `test_manifest_list_missing_amd64_platform_is_unknown` |
| Resolved "single-arch" manifest is itself still a list | `malformed` | `test_nested_manifest_list_is_malformed` |
| Config blob has no `created` field | `malformed` | `test_config_blob_missing_created_is_malformed` |
| Dockerfile listing (git-tree) fetch fails | repo lands in `repos_errors` | `test_repo_collection_failure_lands_in_repos_errors` |
| One Dockerfile's content fetch fails (tree OK) | `file_fetch_failed`, surfaced not dropped | `test_file_fetch_failure_surfaces_as_unknown` |

`render_check` mirrors `red_sweep.render_check`'s never-false-clean posture
exactly: a missing or stale (>48h) verdict is a loud WARNING regardless of
content; a fresh verdict with non-empty `unknown` reports "no drift over
threshold" (i.e. among what COULD be resolved) **plus** an explicit
"UNKNOWN, not clean" warning — it can never read as a plain all-clear.

**Auth:** the standard OCI distribution `WWW-Authenticate: Bearer
realm=...,service=...,scope=...` challenge/token flow, registry-agnostic (no
per-registry special case — verified working against Docker Hub, and the
same code path covers `ghcr.io`/`gcr.io`). A live-network bug surfaced and
fixed during development: Docker Hub's real 401 response sends an
all-lowercase `www-authenticate` header, not the `WWW-Authenticate` casing
most examples show; a case-sensitive lookup silently 401'd on every real
pin (see `_find_header`, pinned by
`test_lowercase_www_authenticate_header_still_completes_challenge`).

**Multi-arch:** #1205's own evidence table was built by "resolving each
pin's amd64 config blob" — this sweep always resolves the **linux/amd64**
platform specifically (the one every consuming Dockerfile in this org
builds FROM in CI), and reports `no_amd64_platform` rather than silently
picking an arbitrary platform when it's absent.

**Rate-limit hygiene:** each unique `(registry, repository, digest)` (and
each `(registry, repository, tag)` for "current digest" resolution) is
resolved at most once per sweep run (memoized) — pinned by
`test_same_digest_across_two_repos_resolved_once` (2 repos pinning the
identical digest costs 4 HTTP calls total, not 8).

**Scope, stated plainly (not oversold):** this is skew and audit-surface,
not a live CVE — the in-image `apt-get -y upgrade`/`apk upgrade` limb of the
base-pin convention already re-freshens packages at build time on top of
whatever base a Dockerfile pins.

## Mutation table (manual, real output — 16 mutants, 16 killed)

Each row: single-line mutation applied via `Edit`, `pytest` run, then
reverted byte-for-byte (diffed clean against a pre-mutation backup before
committing). All commands: `ENVIRONMENT=test python3 -m pytest
.claude/lib/tests/test_base_pin_drift_sweep.py -q`.

| # | Mutation | Killed by | Result |
|---|---|---|---|
| M1 | `resp.status == 429` → `== 430` | `test_rate_limited_429_is_unknown_distinct_from_not_found` + 1 more | 2 failed |
| M2 | `resp.status == 404` → `== 405` | `test_not_found_404_is_unknown` | 1 failed |
| M3 | `except urllib.error.URLError` → `except ValueError` (network catch) | `test_network_error_is_unknown` (uncaught `URLError`) | 1 failed (error) |
| M4 | `platform.get("architecture") == "amd64"` → `"arm64"` | 6 tests incl. both happy-path resolve tests | 6 failed |
| M5 | `span_days > cross_repo_threshold_days` → `>=` | `test_cross_repo_drift_exactly_at_threshold_is_not_flagged` | 1 failed |
| M6 | `behind_days > behind_tag_threshold_days` → `>=` | `test_behind_current_tag_exactly_at_threshold_is_not_flagged` | 1 failed |
| M7 | disabled digest-cache memoization (`if True:`) | `test_same_digest_across_two_repos_resolved_once` (4→6 HTTP calls) | 1 failed |
| M8 | `_find_header`: `key.lower() == name_l` → `key == name_l` | `test_lowercase_www_authenticate_header_still_completes_challenge` + `FindHeaderTests` | 2 failed |
| M9 | `parse_image_ref`: `"/" not in repository` → `"/" in repository` | 6 tests incl. `test_docker_hub_official_image` | 6 failed |
| M10 | Dockerfile basename filter: dropped `.startswith("Dockerfile.")` | `test_filters_to_dockerfile_basenames` | 1 failed |
| M11 | `render_check`: `age > max_age_hours` → `>=` | `test_verdict_exactly_at_max_age_is_not_stale` | 1 failed |
| M12 | `branch or "main"` → `"master"` | `test_default_branch_fetch_failure_falls_back_to_main` | 1 failed |
| M13 | `get_bearer_token`: dropped `data.get("token") or` | 3 tests incl. `test_successful_token_exchange` | 3 failed |
| M14 | `resolve_created`: flipped `not isinstance(config, dict)` → `isinstance(...)` | 6 tests incl. both happy paths | 6 failed |
| M15 | `collect_repo_pins`: swallowed tree-fetch failure (returned empty result instead of `None`) | 3 tests incl. `test_repo_collection_failure_lands_in_repos_errors` | 3 failed |
| M16 | `sweep()`: dropped the file-fetch-failure `unknown.append` | `test_file_fetch_failure_surfaces_as_unknown` | 1 failed |

Every mutation was caught by at least one test; the module was diffed
byte-identical against its pre-mutation-testing copy before this commit.
Two boundary tests (`test_cross_repo_drift_exactly_at_threshold_is_not_flagged`,
`test_behind_current_tag_exactly_at_threshold_is_not_flagged`) and one
staleness-boundary test (`test_verdict_exactly_at_max_age_is_not_stale`) were
added specifically to close `>`-vs-`>=` gaps the initial suite would have
missed (M5/M6/M11 would otherwise have survived).

A separate bug was caught mid-development: an early version of a CLI
exit-code test monkeypatched `bpd._run_gh`/`bpd.default_http_get` module
attributes, but `sweep()`'s own `run_gh`/`http_get` parameters are bound to
those functions as **default argument values at module-load time** —
patching the module attribute afterward doesn't change an already-bound
default, so that test was silently falling through to the real `gh` CLI and
real network. Fixed by patching `bpd.sweep` itself via
`unittest.mock.patch.object` (an ordinary call `_cmd_sweep` resolves at call
time), split into `test_cmd_sweep_exit_2_when_no_repos_checked` /
`test_cmd_sweep_exit_0_when_a_repo_was_checked`. No test in the final suite
touches the network or spawns `gh`.

## Real result — run against the actual org repos today

`python3 .claude/lib/base_pin_drift_sweep.py sweep` (real `gh api` +
real Docker Hub, 2026-08-03):

```
Base-image pin DRIFT found as of 2026-08-03T14:37:25Z:
  [behind_current_tag] docker.io/library/node:22-alpine :: noorinalabs-landing-page:Dockerfile:4 pinned 2026-06-18 is 41d behind the tag's current digest (2026-07-29)
  [behind_current_tag] docker.io/library/node:24-alpine :: noorinalabs-isnad-graph:frontend/Dockerfile:8 pinned 2026-06-24 is 36d behind the tag's current digest (2026-07-30)
  [behind_current_tag] docker.io/library/python:3.12-slim :: noorinalabs-user-service:Dockerfile:26 pinned 2026-06-11 is 33d behind the tag's current digest (2026-07-14)
  [behind_current_tag] docker.io/library/python:3.12-slim :: noorinalabs-deploy:integration-tests/Dockerfile.runner:5 pinned 2026-06-11 is 33d behind the tag's current digest (2026-07-14)
  [cross_repo] docker.io/library/python:3.14-slim :: noorinalabs-isnad-graph:Dockerfile:12 pinned 2026-06-11 vs noorinalabs-deploy:integration-tests/fake_oauth/Dockerfile:4 pinned 2026-07-14
  [behind_current_tag] docker.io/library/python:3.14-slim :: noorinalabs-isnad-graph:Dockerfile:12 pinned 2026-06-11 is 33d behind the tag's current digest (2026-07-14)
  [behind_current_tag] docker.io/library/python:3.14-slim :: noorinalabs-data-acquisition:Dockerfile.load:23 pinned 2026-06-11 is 33d behind the tag's current digest (2026-07-14)
  ... (4 more isnad-ingest-platform Dockerfiles pinned to the same 2026-06-11 digest)
```

14 digest-pinned `FROM`s found across all 8 repos; the `python:3.14-slim`
`cross_repo` finding is exactly the #1205 evidence, reproduced live. No
`unknown` entries in this run (every registry resolution succeeded).

## Follow-up filed (not this PR)

Wiring `base_pin_drift_sweep.py check` into `/session-start` alongside
`red_sweep.py check` was explicitly left out of scope here to keep this PR
to the sweep itself; filed as a follow-up `tech-debt`/`process` issue (no
wave label).

## Test plan

- [x] `ENVIRONMENT=test python3 -m pytest .claude/lib/tests/test_base_pin_drift_sweep.py -q` — 75 passed
- [x] `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/ .claude/lib/tests/ -q` — 3716 passed, 651 subtests
- [x] `ruff format --check` / `ruff check` — clean
- [x] `mypy .claude/hooks/ .claude/lib/ --ignore-missing-imports --no-error-summary` — clean
- [x] `actionlint` — clean
- [x] `python3 .claude/lib/pre_commit_ci_sync.py .` — parity OK
- [x] Full local pre-commit + pre-push hook run on `git push` — all green
- [x] Ran the real sweep against the live org repos (see above)


parametrization:
## Merge-gate review — head SHA `6026c9f97865e4501abf2852f14aa34f59ed8854`

Reviewed at that SHA via `gh api repos/.../contents/<path>?ref=6026c9f` and a
`git archive` export, not a local checkout. This is very good work, and I want to
be clear that the one blocking item below is a missing test, not a bug: the
shipped code behaves correctly in every mode I could construct. But it is a
missing test on precisely the property this PR exists to guarantee, and it is
~10 lines to close.

## Blocking: a surviving mutant on the load-bearing property

Deleting one line from `sweep()` — the line that folds `compute_drift`'s
`unknown` list into the persisted verdict —

```python
    unknown.extend(drift_unknown)
```

leaves **75/75 tests passing**, and turns a total registry outage into exactly
the plausible zero this sweep was written to prevent. With that line replaced by
`pass` (2 repos each pinning `python:3.14-slim`, every registry call raising
`URLError`):

```
verdict unknown  : []
verdict findings : []
pins_checked     : 2
--- render_check on the persisted verdict ---
No base-image pin drift over threshold as of 2026-08-03T15:02:19Z.
```

No warning. Nothing to indicate the registry was never reached. That verdict
gets committed to `refs/meta/base-pin-drift` and, once #1296 lands, read at
session start as a clean bill of health.

The reason the suite misses it is a seam, not an oversight in any one test.
`ComputeDriftTests::test_unresolved_pin_lands_in_unknown_not_silently_dropped`
pins that `compute_drift` *produces* the unknowns.
`RenderCheckTests::test_unknown_entries_never_read_as_clean` pins that
`render_check` *warns* on them. Nothing pins that they *travel between the two*.
The one sweep-level unknown test,
`SweepIntegrationTests::test_file_fetch_failure_surfaces_as_unknown`, filters on
`u["kind"] == "file"` — and `kind="file"` is appended directly in `sweep()`,
never routed through `compute_drift`. So the entire registry-resolution half of
the unknown path has no end-to-end assertion. The only sweep test that carries
real pins, `test_same_digest_across_two_repos_resolved_once`, resolves everything
with 200s.

The ask: one `SweepIntegrationTests` case that runs `sweep()` with pins and an
`http_get` that fails, asserting `verdict["unknown"]` is non-empty and carries
`kind="pin"` (a `kind="current_tag"` assertion in the same test would close the
tag half for free). Per the charter, that lands as an additive commit — no
force-push during review.

I raise this specifically because the PR body claims each failure mode is
"pinned by a test." At the `resolve_created` level that claim is true and I
verified it. At the verdict level, for the registry modes, it is not yet.

## What I verified, and how

### Degradation, per mode, constructed rather than read

I did not trust the table. I stood up a real `http.server` on a real loopback
socket and drove the **real** `default_http_get` against it (only the `https://`
→ `http://` scheme rewritten for the loopback host), so `urllib`'s own
`HTTPError`/`URLError` translation stayed in the loop — the shipped `FakeHttp`
constructs `HttpResponse` objects directly and bypasses that adapter entirely.

```
PASS  HTTP 429 (Docker Hub rate limit)       -> status='unknown'   reason='rate_limited'
PASS  HTTP 404 (repo/tag/digest gone)        -> status='unknown'   reason='not_found'
PASS  HTTP 503 (other non-200)               -> status='unknown'   reason='http_503'
PASS  200 but body is not JSON               -> status='unknown'   reason='malformed'
PASS  200 but JSON is an array               -> status='unknown'   reason='malformed'
PASS  manifest list, no linux/amd64          -> status='unknown'   reason='no_amd64_platform'
PASS  429 on the CONFIG BLOB, not manifest   -> status='unknown'   reason='rate_limited'
PASS  config blob has no 'created'           -> status='unknown'   reason='malformed'
PASS  lowercase www-authenticate 401         -> status='ok'        created='2026-06-11T01:02:03Z'
PASS  uppercase WWW-Authenticate 401         -> status='ok'        created='2026-06-11T01:02:03Z'
PASS  happy path manifest list               -> status='ok'        created='2026-06-11T01:02:03Z'
PASS  connection refused (real socket)       -> status='unknown'   reason='network_error'
PASS  arch manifest is itself a list         -> status='unknown'   reason='malformed'
```

13/13. The `www-authenticate` fix is genuine: my server sends the header in
lowercase exactly as Docker Hub does, the challenge completes, and the pin
resolves. Reverting `_find_header` to `key == name_l` fails
`GetWithAuthTests::test_completes_challenge_and_retries_with_bearer` and
`FindHeaderTests::test_matches_regardless_of_case` — the regression test does
catch it. Both casings resolve, so the fix is case-insensitive rather than
lowercase-only.

At the `sweep()` level, on unmutated code, a total registry outage across 2 repos
produces `findings=[]` with 3 unknowns (2 pins + 1 current_tag, all
`network_error`) and `render_check` says `UNKNOWN, not clean`. A repo whose
Dockerfile listing fails lands in `repos_errors` and warns. One unfetchable
Dockerfile with a good tree lands as `kind="file"`, `reason="file_fetch_failed"`,
and warns. All three correct.

### Rate limits

429 is visibly distinct from every other mode, and it does not hammer. There is
no retry loop anywhere; `_get_with_auth` retries exactly once and only after a
401 token exchange. Two repos pinning the same digest, everything answering 429:

```
2 repos pinning the SAME digest, everything 429 -> 2 HTTP calls total
  https://registry-1.docker.io/v2/library/python/manifests/sha256:aaaa...
  https://registry-1.docker.io/v2/library/python/manifests/3.14-slim
```

The memoization holds under failure too — a rate-limited outcome is cached, so a
429 storm costs one call per unique digest and tag, not one per pin. Worth
noting the sweep does not *stop* on a 429 — it keeps issuing requests for other
digests against an already-limited registry. Given 14 pins org-wide that is fine;
I would not add backoff at this scale.

### Multi-arch

Explicit, not incidental. `_select_amd64_manifest_digest` requires
`architecture == "amd64"` **and** `os == "linux"` — my probe includes a manifest
list offering `linux/arm64` and `windows/amd64` and it correctly returns
`no_amd64_platform` rather than taking the windows/amd64 entry. Flipping the
literal to `arm64` kills 6 tests. The choice is documented in the module
docstring and tied back to #1205's own evidence method.

### Threshold-boundary mutants, re-run

You flagged these as the likeliest real defect and you were right to. All three
confirmed dead, restored byte-identical to the head SHA after each:

| Mutation | Result |
|---|---|
| `span_days > cross_repo_threshold_days` → `>=` | 1 failed — `test_cross_repo_drift_exactly_at_threshold_is_not_flagged` |
| `behind_days > behind_tag_threshold_days` → `>=` | 1 failed — `test_behind_current_tag_exactly_at_threshold_is_not_flagged` |
| `age > max_age_hours` → `>=` | 1 failed — `test_verdict_exactly_at_max_age_is_not_stale` |
| `span_days > …` → `<` (mine, not in your table) | 2 failed — over- and under-threshold cases |

I also ran five false-clean mutants of my own beyond your table.
`_unknown()` returning `status="ok"` kills 8. Dropping the pin-unknown append in
`compute_drift` kills 1. Suppressing either `render_check` warning kills 1 each.
`read_verdict` returning `{}` instead of `None` kills 2. Only `unknown.extend`
survived.

### Live run

Reproduced independently, and it matches your reported numbers exactly:
14 digest-pinned `FROM`s across all 8 repos, `repos_errors: []`, `unknown: []`,
12 findings — including the `python:3.14-slim` `cross_repo` finding between
`isnad-graph:Dockerfile:12` (2026-06-11) and
`deploy:integration-tests/fake_oauth/Dockerfile:4` (2026-07-14), which is #1205's
original evidence, and `node:22-alpine` in `landing-page` 41 days behind current.

### CI, parity, siting

All 21 checks SUCCESS at `6026c9f` per `statusCheckRollup` — I checked the rollup
at the head SHA, not the branch. Independently re-ran
`python3 .claude/lib/pre_commit_ci_sync.py .` (OK), `actionlint` on the new
workflow (clean), `ruff format --check` / `ruff check` / `mypy` (clean), and the
75-test suite (75 passed). The workflow adds no CI-enforced check, so it creates
no new parity obligation; the drift gate agrees. Branch is 1 commit behind `main`
(`cc64755`, memory-only) — no conflict, `MERGEABLE`.

On siting: I think its own workflow is the right call and the rationale in the
docstring holds — the 6h cadence genuinely buys nothing for a signal that moves
in weeks, and keeping the registry failure surface out of `red-sweep.yml` means a
Docker Hub 429 can't degrade half of an unrelated verdict. The workflow is a
faithful copy of `red-sweep.yml`'s persist step, including `if: always()` so a
partial verdict still lands.

Your "second thing that can silently stop running" concern is real and currently
unmitigated: `render_check` has a proper 48h staleness guard, but **nothing calls
it**. `red_sweep.py check` is invoked from `.claude/skills/session-start/SKILL.md:285`;
`base_pin_drift_sweep.py check` has no caller anywhere in the repo. Until #1296
lands, a cron that quietly stops produces no signal at all — the guard exists but
is unreached. #1296 is filed and correctly scoped, so I'm not blocking on it, but
it should not sit long.

### Thresholds and scope honesty

Both thresholds are module constants with `--cross-repo-threshold-days` /
`--behind-tag-threshold-days` CLI overrides, and both defaults trace to #1205's
own "e.g. 30 days" rather than being picked. Fine as policy defaults. The
workflow doesn't pass either flag, so changing org policy means editing the
constant or the workflow — acceptable at this scale.

Scope honesty is stated plainly in both the PR body and the module docstring:
skew and audit surface, not a live CVE, with the in-image
`apt-get -y upgrade`/`apk upgrade` limb named as what already re-freshens
packages at build time. Not oversold. `parse_froms`/`digest_of`/`registry_host`
are genuinely reused — the only `re.compile` in the new module is the
`WWW-Authenticate` param parser, so #1271's second-parser trap is avoided.

## Non-blocking, filed

- **#1299** — `default_http_get` is the module's only real-network adapter and has
  zero test coverage; `FakeHttp` starts one layer above it, so the
  `HTTPError` → `HttpResponse` translation that decides `rate_limited` vs
  `network_error` is unpinned. It works today (I verified against real sockets);
  nothing would tell us if it stopped.
- **#1300** — `compute_drift` labels intra-repo pin skew as `cross_repo`. Two
  Dockerfiles in one repo at 61 days apart report
  `deploy:a/Dockerfile:1 … vs deploy:b/Dockerfile:1` under the `cross_repo` kind.
  The finding is true and useful, only the label misdescribes it. Real here:
  `isnad-ingest-platform` carries five `python:3.14-slim` pins.

One nit not worth an issue: a future-dated `checked_at` yields a negative age and
reads as fresh rather than as suspect. `red_sweep.render_check` has the identical
`age > max_age_hours` shape, so this is pattern parity, not a regression in this
PR — flagging only so it isn't mistaken for a deliberate difference.

Re-request review once the `unknown.extend` seam has a test and I'll turn this
around quickly. Everything else here is ready.

---
Requestor: Weronika Zielinska
Requestee: Lucas Ferreira
TechDebt: #1299, #1300

parametrization:
Rebased on `origin/main` (#1292 folded in cleanly, no conflicts) and addressed the merge-gate blocker.

**Fix**: added `SweepIntegrationTests::test_registry_resolution_failure_reaches_persisted_verdict_and_render_check_warns` in `.claude/lib/tests/test_base_pin_drift_sweep.py`. It drives `sweep()` end-to-end with both a pin's digest-manifest fetch and its tag's current-digest fetch returning 404, then asserts:
- `verdict["unknown"]` contains both a `"pin"`-kind and a `"current_tag"`-kind entry (i.e. `compute_drift`'s registry-resolution unknowns actually reached the persisted verdict, not just `sweep()`'s own directly-appended `kind=="file"` unknowns), and
- `render_check(verdict, ...)` renders `"UNKNOWN, not clean"` and the `"not_found"` reason.

**Mutation-verified**: temporarily deleted `unknown.extend(drift_unknown)` from `sweep()` — the new test failed (`AssertionError: 'pin' not found in set()`) while the other 75 tests stayed green, then restored the source with `git checkout HEAD --`. Full suite is 76/76 green with the join intact.

**Other seams checked**: looked for other produce-here/consume-there joins of the same shape. `findings` reaches the verdict via a plain list comprehension inline in the same `return` statement (not a separate `.extend()`), and — worth noting — no existing `sweep()`-level test currently drives an actual drift `finding` through to the persisted verdict/`render_check` either (only `compute_drift`'s own unit tests and a hand-built-verdict `render_check` test cover the two ends). Flagging it as a possible follow-up rather than folding it into this PR, per "do not widen scope."

CI parity re-run clean on the rebased tree: ruff-format/ruff-lint, mypy (`.claude/hooks`+`.claude/lib`+per-skill), full `pytest` suite (3742 passed, 668 subtests) + skill suites, sync-drift gate, actionlint, memory/headcount budgets, doc-freshness, office-drift, mermaid-render — all green via the real pre-commit/pre-push hooks (not just `--no-verify`'d).

Not merging — leaving this for review.

parametrization:
## Merge-gate re-review — head SHA `8c8c4c66ddcf1cfd8ea13c4dc0273fb5b619e0a8`

The seam I blocked on is **closed, and I verified it independently rather than taking the mutation report on trust**. A second seam of the *identical* class is still open, and I am not waiving it — the rest of this comment is why, and exactly what closes it.

Method: `git archive 8c8c4c6 | tar -x` into a private scratch tree (never `--work-tree` against the shared checkout), every mutant restored from the head blob between runs, `ENVIRONMENT=test python3 -m pytest .claude/lib/tests/test_base_pin_drift_sweep.py -q`. Baseline at head: **76 passed**.

### 1. The blocked seam is closed — confirmed, and the new test is load-bearing in all four of its assertions

`test_registry_resolution_failure_reaches_persisted_verdict_and_render_check_warns` builds its verdict by calling `bpd.sweep(...)` over injected `gh`/http doubles — it does **not** assert against a hand-built dict, so it genuinely drives the join. Four mutants, four kills:

| Mutant applied to `sweep()` / `render_check()` | Result |
|---|---|
| delete `unknown.extend(drift_unknown)` | **RED** — `AssertionError: 'pin' not found in set()`; other 75 green. Reproduces your report exactly. |
| pass `{}` instead of `tag_cache` into `compute_drift` | **RED** — the `current_tag` assertion catches the tag half being unwired |
| drop `render_check`'s `for item in unknown[:10]` detail lines | **RED** — the `not_found` assertion is doing real work, not riding on the `UNKNOWN, not clean` line |
| `unknown.extend(u for u in drift_unknown if u["kind"] != "pin")` | **RED** — **partial** join loss is caught, not merely total loss |

That last one is the one I most wanted to see fail, and it does. This is a better test than the minimum I asked for.

### 2. Blocking: the `findings` path has the same untested join — **four** surviving mutants

Not a hypothetical. Each of these leaves **76/76 green** at head:

| Mutant | What ships | Suite |
|---|---|---|
| `"findings": []` in the returned dict | every real drift finding vanishes; report reads `No base-image pin drift over threshold as of <ts>` | **76 passed** |
| drop both `*_threshold_days=` kwargs from the `compute_drift(...)` call | `--cross-repo-threshold-days` / `--behind-tag-threshold-days` silently do nothing; the sweep runs at hardcoded defaults | **76 passed** |
| rename the producer key `"detail"` → `"details"` | `render_check` reads `item.get('detail')` → renders `[cross_repo] python:3.14-slim :: None`; drift found, evidence lost | **76 passed** |
| add `if f.kind == "cross_repo"` to the comprehension | every `behind_current_tag` finding silently dropped — half the tool's job | **76 passed** |

The third is the one that decides this for me. `compute_drift` is tested against `DriftFinding` objects; `render_check` is tested against hand-built dicts. **Nothing anywhere compares the producer's dict shape to the consumer's expected shape**, so a key rename between them is invisible to a 76-test suite and degrades a real finding into `:: None`. That is not a mutation-score quibble; it is an unverified contract between the two halves of the feature.

I am applying the same standard I applied in the negative direction one round ago, and since then it has become written policy in `main`: `.claude/memory/feedback_both_ends_tested_join_untested.md` (merged today, generated by this PR's first block) says *"when a property is stated end-to-end, the test must be end-to-end — per-stage tests of the same property are necessary and not sufficient"* and instructs reviewers to *"mutate the plumbing… delete each hand-off — `x.extend(y)`, **returning a collected list**, **passing a field into a constructor**"*. Mutant 1 is literally *returning a collected list*; mutant 2 is literally *passing a field into a call*. Merging this would ship a violation of the note this PR authored. The deferral was a reasonable read of "do not widen scope" — the scope call is mine at the gate, and this is inside it, not beyond it.

### 3. What closes it — one test, not a widening

The positive mirror of the test you just wrote. One test kills all four mutants above:

- Two repos, each with a `Dockerfile` pinning the **same** `python:3.14-slim` tag to **different** digests whose config blobs carry `created` dates far apart, and a tag-current digest newer than both — so the run yields one `cross_repo` **and** one `behind_current_tag` finding (kills the `kind`-filter mutant).
- Call `sweep(...)` with **non-default** `cross_repo_threshold_days=` / `behind_tag_threshold_days=` chosen so the defaults would give a different answer (kills the threshold-forwarding mutant).
- Assert `verdict["findings"]` is non-empty and that a finding's `detail` string carries the repo/path/lineno evidence (kills the empty-list and key-rename mutants).
- Feed that verdict to `render_check()` and assert the rendered line contains that same detail text — closing producer→consumer end to end, the way the unknown path now is.

The fixtures already exist in the file (`FakeGh`, `FakeHttp`, `_b64`, `_tree_json`, and `ComputeDriftTests`' drift-producing date math). No production code should change. Please re-run the four mutants above against the new test and report which go red.

### 4. Rebase spot-check — nothing I previously verified was disturbed

`.claude/lib/base_pin_drift_sweep.py` and `.github/workflows/base-pin-drift-sweep.yml` are **byte-identical** to `6026c9f97865e4501abf2852f14aa34f59ed8854`, the SHA of my first review (fetched from the contents API at the orphaned SHA and `diff`ed — no differences at all). The test file is **+47 / −0**. So the 13-mode real-socket degradation probe, the three threshold-boundary mutants, multi-arch selection, the 429-vs-404 split, the lowercase `www-authenticate` fix and the live-sweep numbers all stand unchanged — I did not re-run them, by design. Merge-base is the #1292 merge commit, so the rebase folded it in; `main` has since advanced by #1269, touching only hook/charter-trailer files with **zero** overlap. `mergeStateStatus: CLEAN`.

One process note, not a re-block: the branch was **force-pushed at 15:15:11Z during an open ChangesRequested cycle**, which `pull-requests/reviews.md` § Additive Commits on ChangesRequested prohibits precisely because it resets the head-SHA anchor a reviewer's `contents/<path>?ref=<sha>` chain depends on. It cost nothing here only because GitHub still served the orphaned blob so I could prove byte-identity. The sanctioned way to pick up a moved base mid-cycle is `git merge origin/main`, not a rebase.

### 5. CI

All **21** checks SUCCESS, and I confirmed the run's `headSha` is `8c8c4c6` rather than trusting the rollup's placement — including `Config — Workflow actionlint` and `Pre-commit ⇄ CI sync-drift gate`, both of which the new workflow file makes load-bearing.

### 6. Not re-litigated

#1296 (uncalled `check` subcommand), #1299 (`default_http_get` untested), #1300 (intra-repo skew mislabelled `cross_repo`) are correctly filed and correctly out of scope here. Nothing new filed — item 2 is a fix in this PR, not debt to carry.

Everything else about this PR is merge-ready; this is one test away.

---
Requestor: Weronika Zielinska
Requestee: Lucas Ferreira
TechDebt: none

parametrization:
## Merge-gate re-review — head SHA `c5c45c3a8733449191494b7fb88e37497d0ae1d4`

**Approved.** The `findings` join is closed, and I re-derived every kill myself rather than accepting the table — including one mutant that was not on it and that turns out to be the reason the render assertion earns its place.

Method unchanged: `git archive c5c45c3 | tar -x` into a private scratch tree, each mutant restored from the head blob between runs. Baseline at head: **77 passed**.

### 1. The four mutants I blocked on are dead

| Mutant | Result at `c5c45c3` |
|---|---|
| `"findings": []` | **RED** — `AssertionError: 0 != 2` |
| drop both `*_threshold_days=` kwargs from `compute_drift(...)` | **RED** — `AssertionError: 0 != 2` |
| `"detail"` → `"details"` in the producer dict | **RED** — `KeyError: 'detail'` |
| `if f.kind == "cross_repo"` filter on the comprehension | **RED** — `AssertionError: 1 != 2` |

The fixture arithmetic is what makes the second one a real kill rather than a coincidence, and it checks out: pins 15d apart, current tag 23d past the older and 8d past the newer, thresholds 10/10. Every gap clears 10 and none clears the module default of 30, so dropping the kwargs collapses `findings` to empty instead of merely reshuffling it — and the 8d gap is what keeps `behind_current_tag` at exactly one finding, which is what gives the `kind`-filter mutant its `1 != 2` rather than a silent pass. `test_findings_join_reaches_verdict_with_custom_thresholds_and_renders` builds its verdict from `bpd.sweep(...)` over injected doubles throughout — no hand-built dict anywhere in it.

### 2. A fifth mutant, off the table, that settles the `KeyError` question

The render assertions (`assertIn(cross_repo_detail, rendered)`) are *not* what catches the key rename — the `by_kind["cross_repo"]["detail"]` lookup on the line above raises first, so the render half never executes under that mutant. That made me ask what the render assertions uniquely catch, so I ran two more:

| Mutant | Result |
|---|---|
| `render_check` drops its per-finding `lines.append(...)` entirely | **RED ×2** — new test **and** `test_fresh_with_findings_lists_them` |
| `render_check` keeps the line but renders only `kind` + `image_key`, dropping `detail` | **RED ×1 — the new test only** |

That second one is the answer. `test_fresh_with_findings_lists_them` survives it (it asserts on the image key, which is still rendered); only the new test fails, on `'repo-a:Dockerfile:1 pinned 2026-06-01 vs repo-b:Dockerfile:1 pinned 2026-06-16' not found in ...`. So the producer→consumer join assertion covers a real class that nothing else in the suite reaches, independent of the `KeyError`. It stays.

**On tightening the `KeyError` into a named-contract assertion — no, don't hold the merge for it.** A `KeyError` on a contract key is unambiguous: the traceback names the missing key and the exact line, which is as much as a custom message would say. What is genuinely slightly off is the test's own comment, which attributes the rename kill to the render assertion ("*asserting the real detail text appears in the rendered report is what catches that*") when the lookup above it fires first. The risk is only that a future reader trims the render assertions believing the `KeyError` subsumes them — which mutant 5 above shows it does not. Worth one comment-line correction if something else forces a commit on this branch; **not** worth a commit of its own, and not a condition of this approval.

### 3. Additive, and nothing previously verified moved

`8c8c4c6` is an ancestor of `c5c45c3` (`git merge-base --is-ancestor` confirms), the base update is a merge commit (`1f49e73`, folding `origin/main` through #1293), and the timeline shows no force-push after 15:15:11Z. `.claude/lib/base_pin_drift_sweep.py` and `.github/workflows/base-pin-drift-sweep.yml` are **byte-identical to `6026c9f97865e4501abf2852f14aa34f59ed8854`** — the SHA of my first review, three heads ago. Zero production-code change across the whole review cycle; both rounds of rework were tests only. So the 13-mode real-socket degradation probe, the threshold-boundary mutants, multi-arch selection, the 429-vs-404 split, the lowercase `www-authenticate` fix and the live-sweep numbers all still stand, and I did not re-run them.

### 4. CI

**21/21 SUCCESS** at `c5c45c3`, `mergeStateStatus: CLEAN` — including `Config — Workflow actionlint` and the `Pre-commit ⇄ CI sync-drift gate`, the two the new workflow file makes load-bearing. (The `Mermaid render gate` was still in flight when I started; it has since gone green.)

Locally I also ran the full `.claude/lib/tests/` + `.claude/hooks/tests/` set against the head tree: **3786 passed**, 750 subtests, with one failure — `test_wave_status.py::BaseVsHeadDifferential::test_base_and_head_agree_on_wave_branch_output`, which shells out to `git show <base-sha>:...` and dies with `fatal: not a git repository` because a `git archive` extraction has no `.git`. That is an artifact of my sandbox, not of this branch; the same job is green in CI at this SHA. Recording it so the number isn't misread.

### 5. Not re-litigated

#1296, #1299 and #1300 remain correctly filed and correctly out of scope. Nothing new filed — both rounds of my block were fixes in this PR, not debt to carry forward.

Good work on the second round. The fixture design in particular — picking gaps that clear the custom threshold but not the default — is what turned "assert findings is non-empty" into a test that actually pins the kwarg forwarding, and that was the harder half of the ask.

---
Requestor: Weronika Zielinska
Requestee: Lucas Ferreira
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Second review — #1295 (head `c5c45c3a8733449191494b7fb88e37497d0ae1d4`)

Confirmed the diff against `head=c5c45c3` is byte-identical to the two
prior review rounds for `.claude/lib/base_pin_drift_sweep.py` and
`.github/workflows/base-pin-drift-sweep.yml` — only the test file grew.
CI at this SHA: all 21 checks green (Ruff, mypy, actionlint, cspell,
pre-commit⇄CI sync-drift gate, roster/commit-identity gates, the full
hooks+lib pytest suite, etc.).

### Join enumeration (per the brief: every value crossing a function
boundary in this module)

| Join | Both ends tested? | Hand-off itself tested? |
|---|---|---|
| `collect_repo_pins` -> `sweep` (`result.pins`/`fetch_failures`/`None`) | Yes | Yes — `test_repo_collection_failure_lands_in_repos_errors`, `test_file_fetch_failure_surfaces_as_unknown`, and both `test_same_digest_across_two_repos_resolved_once` / `test_findings_join_reaches_verdict_with_custom_thresholds_and_renders` drive real Dockerfile text through `collect_repo_pins` inside a real `sweep()` call |
| `resolve_created` -> `compute_drift` (via `ResolvedPin.outcome`) | Yes | Yes — same two end-to-end tests run real `resolve_created()` (via `FakeHttp`) into real `compute_drift()` |
| `sweep` -> verdict dict -> `render_check` | Yes | Yes — `test_findings_join_reaches_verdict_with_custom_thresholds_and_renders` feeds a verdict `sweep()` actually produced into `render_check` and asserts the real `detail` text survives (this is block 2's fix, confirmed still present) |
| `sweep` -> verdict dict -> `read_verdict` (base64/JSON round trip) | Yes, separately | Not exercised as one chain (hand-built verdict for `read_verdict`, `sweep()`'s own dict for `render_check`, never both together) — low risk, all verdict fields are plain str/int/float/list/dict, no exotic serialization |
| **CLI flags -> `_cmd_sweep` -> `sweep`/`compute_drift`** | Individually yes (`_build_parser` wiring verified by hand; `sweep`/`compute_drift` honor whatever kwargs they receive) | **No** |

### The third one: `_cmd_sweep`'s CLI-flags hand-off

The two `_cmd_sweep` tests
(`test_cmd_sweep_exit_2_when_no_repos_checked`,
`test_cmd_sweep_exit_0_when_a_repo_was_checked`) mock `bpd.sweep` wholesale
and only assert `mocked.assert_called_once()` — never
`assert_called_once_with(...)` — so nothing checks what `_cmd_sweep`
actually passes through from `args` into the `sweep(...)` call.

Mutated `_cmd_sweep` two ways, restored byte-identical after each
(diffed against a pre-mutation backup), full suite run
`ENVIRONMENT=test python3 -m pytest tests/test_base_pin_drift_sweep.py -q`
each time:

- **Swap the two threshold kwargs** (`cross_repo_threshold_days=args.behind_tag_threshold_days`,
  `behind_tag_threshold_days=args.cross_repo_threshold_days`) — **77/77 passed**, mutant survived.
- **Ignore `--repos` entirely** (hardcode `repos = None`, silently falling
  back to `CANONICAL_REPOS`) — **77/77 passed**, mutant survived.

Reproduced through the real parser (not just a hand-built `Namespace`) with
a capturing fake `sweep`:

```
bpd.main(['sweep', '--cross-repo-threshold-days', '7', '--behind-tag-threshold-days', '90'])
captured {'cross_repo_threshold_days': 90.0, 'behind_tag_threshold_days': 7.0}
```

With the swap mutation live, `--cross-repo-threshold-days 7` silently
becomes the *behind-tag* threshold and vice versa — exit 0, no error,
wrong answer. Today's workflow invokes `sweep` with no threshold overrides
at all (see `.github/workflows/base-pin-drift-sweep.yml`), so this is
latent, not live, in the daily cron — but the CLI overrides are advertised
(module docstring, PR body) as the intended path for ad hoc reruns with
different thresholds, so this is a real gap, not a theoretical one.

Filed as https://github.com/noorinalabs/noorinalabs-main/issues/1307
(`tech-debt`/`process`, no wave label).

### Everything else

Re-verified rather than re-run: the live sweep result in the PR body
reproduces #1205's own evidence; the degradation-stance table (10 failure
modes, each with its own test) is unchanged from the last round; the
lowercase `www-authenticate` fix and its regression test are present;
multi-arch (explicit `architecture=="amd64"` AND `os=="linux"`) and the
memoization test are unchanged. #1296 (nothing calls `check`), #1299
(untested real-network adapter), #1300 (`compute_drift` mislabels
intra-repo skew as `cross_repo`) all remain open and correctly out of
scope for this PR — did not duplicate any of them.

On #1296: agree with the framing in the spawn brief — a sweep whose
verdict nobody reads is a signal that cannot fire yet. That's real, but it
doesn't block *this* PR, which is explicitly scoped to the sweep itself
(the PR body documents the follow-up as deliberately out of scope here).

### Verdict

Approving. The third join I was asked to hunt for is real (`_cmd_sweep`'s
CLI-to-`sweep()` hand-off), but it's a test-coverage gap on an unused-in-CI
code path, not a live defect in the shipped behavior — filed, not blocking.

---

Requestor: Nino Kavtaradze
Requestee: Lucas Ferreira
RequestOrReplied: Approved
TechDebt: none

PR #1302, opened 2026-08-03
fix(hooks/#1195): resolve simple literal assignments before commit-identity matching

### LINKED ISSUE #1195: tech-debt(commit-identity): payload matcher is a phrase matcher — a variable-held command word defeats it with no wrapper at all
## Shape

`_payload_looks_like_commit` is a **phrase matcher** — `_INNER_COMMIT_RE` requires a literal `\bgit\b … \bcommit\b` bridge inside a single scanned token. Holding the command word in a shell variable defeats it, and the shell still runs the commit:

```
bash -lc 'g=git; $g commit -m x'
```

Measured on this box at `a6f6a58` (PR #1193 head), through the real `check()`:

| half | result |
|---|---|
| `bash -lc 'g=printf; $g %s x > MARKER'` — does the shell run the indirected command word? | **yes**, rc=0, marker written |
| `_payload_looks_like_commit("g=git; $g commit -m x")` | **False** (the direct spelling returns True) |
| hook verdict on `bash -c 'g=git; $g commit -m x'` | **allow** |
| hook verdict on `g=git; $g commit -m x` typed directly, no wrapper at all | **allow** |

## Why this is separate from #1149 / #1193

It is **not** an option-grammar defect and #1193 neither causes nor could fix it. The option walker's job is to find the command string; here it finds it correctly and hands over a payload the *matcher* cannot read. The last row above is the tell — the shape needs no interpreter wrapper, so it escapes the indirect-exec gate and the direct-commit path alike. Pre-existing at `ee99676` and at every earlier head; equally open on the bare `-c` spelling that base already blocked.

Belongs to the #1150 family as an under-matching instance, but one layer down from the parsing layer that umbrella is about: #1150 is "the hook's *grammar* is narrower than the real command's", this is "the hook's *payload matcher* is narrower than the real payload's".

## Suggested direction

No cheap complete fix — resolving `$g` needs shell-runtime knowledge the hook does not have, and this is the same class as the interpreter-in-a-variable bound already recorded on #1193. Two bounded options worth pricing before committing to either:

1. **Assignment-aware pre-pass.** Resolve simple literal `NAME=value` assignments within the same payload before matching, so `g=git; $g commit` normalizes to `git commit`. Covers the shape above and nothing speculative; does not touch `$1`, command substitution, or arrays.
2. **Flag the shape rather than resolve it.** Treat "a payload that assigns a variable and later uses it in command position" as unresolvable and fail closed, mirroring the fail-closed treatment of unparseable input elsewhere in the hook.

Option 2 is the conservative one but needs a false-positive sweep first — `d=$(date); echo $d` is an ordinary shape and must not block.

## Out-of-scope for this issue

- Command substitution (`$(printf git) commit`) and `eval`-of-a-variable — same family, strictly harder, and `eval` is already partially covered by shape 6 of `_detect_indirect_commit`.
- The `[^;&|]` bridge bound in `_INNER_COMMIT_RE` generally.

## Provenance

Found by the merge-gate review of PR #1193 (Weronika Zielinska, Platform Architect) while adding a payload-interior dimension to the review corpus — the axis both the author's and the reviewer's earlier corpora held constant. Filed rather than folded into #1193, which is scoped to the option grammar and is clean on that scope.

Searched before filing: `variable indirection`, `payload obfuscation`, `_INNER_COMMIT_RE`, `commit verb`, `alias` — no existing issue.

### LINKED ISSUE #1305: tech-debt(commit-identity): assignment pre-pass skips `export`/`declare`-prefixed assignments — #1195's shape stays open on its most idiomatic spelling
## Shape

`resolve_simple_assignments()` (`.claude/hooks/_shell_parse.py`, added by #1302 for #1195) peels a **leading run of `NAME=value` tokens** at a segment's start. A segment whose first token is `export` (or `declare`, `local`, `typeset`, `readonly`) never enters the peel loop at all — `_ENV_ASSIGN_RE.match("export")` is False, `i` stays 0, and no assignment is captured.

`export` is the most idiomatic way to set a shell variable in a script, so this leaves #1195's headline shape open on an entirely ordinary spelling, **with no interpreter wrapper**.

## Measured through the real `check()` at PR #1302 head `5cea8172be0311aff96479ce87f1776b859efce5`

| command | hook verdict |
|---|---|
| `g=git; $g commit -m x` (the shape #1302 closes) | **block** |
| `export g=git; $g commit -m x` | **allow** |
| `export g=git; ${g} commit -m x` | **allow** |
| `declare g=git; $g commit -m x` | **allow** |

Real-shell truth (proxy command + marker file, same method #1195 used):

```
$ bash -lc 'export g=printf; $g %s export-ran > MARKER2'
$ cat MARKER2
export-ran
```

The indirected command word runs, so the allow rows are genuine gate bypasses, not a modelling nicety.

## Why this is not one of #1302's documented non-goals

#1302 lists four explicit non-goals: command substitution of the command word, positional/special parameters, bare multi-word values (#1303), and the `[^;&|]` bridge bound. `export`-prefixed assignment is none of them — it is the *same* bare-literal single-word value the pre-pass exists to resolve, differing only by a declaration keyword the peel loop does not skip.

## Suggested direction

Skip a leading declaration keyword before the peel loop:

```python
_DECL_KEYWORDS = {"export", "declare", "local", "typeset", "readonly"}
...
if i < len(segment) and segment[i] in _DECL_KEYWORDS:
    i += 1
```

Bounded and testable: it widens *which token position* the existing `_SIMPLE_LITERAL_VALUE_RE` guard is applied at, not what counts as a resolvable value. `declare`/`local` inside a function body have narrower scope than a flat pre-pass models — that is the over-resolve/over-block direction, so the same false-positive corpus #1302 built must be re-run before landing.

## Searched before filing

`export`, `declare`, `resolve_simple_assignments`, `assignment pre-pass`, `commit-identity variable` — no existing issue. Sits under the #1150 under-matching umbrella one layer below the parsing layer, alongside #1195 / #1303.

## Provenance

Found by the merge-gate review of PR #1302 (Nino Kavtaradze, Security Engineer) while probing the deliberate "leading `NAME=value`" scope boundary for undocumented holes.

parametrization:
Closes #1195
Closes #1305

## Shape

`_payload_looks_like_commit`'s `_INNER_COMMIT_RE` requires a literal `git ... commit` bridge inside one scanned token, and the direct commit finder (`find_git_subcommand`, via `_find_commit_segment`) requires the literal token `git` in command position. Holding the command word in a shell variable defeats **both**, with no interpreter wrapper needed at all:

```
g=git; $g commit -m x
```

## Decision: option 1 (assignment-aware pre-pass), not option 2

Option 2 ("flag the shape, fail closed") was rejected: "a payload that assigns a variable and later uses it in command position" is not a rare or unusual shape — `d=$(date); echo $d`, `out=$(mktemp); cat > "$out"`, and any ordinary use of a shell variable as an argument all have this exact structure. Failing closed on the *shape* rather than the *resolved value* would false-block the majority of ordinary scripting idioms this org's own hooks and workflows use. The issue itself flagged this needs "a false-positive sweep first" — the sweep (below) confirms the risk is real and option 2 is not viable without an intractable amount of shape-vs-intent discrimination that option 1 sidesteps by only ever resolving what it can prove is a bare literal.

Option 1 is bounded and testable: resolve a **leading** `NAME=value` assignment whose value is a **bare literal** (no `$`, backticks, parens, or whitespace — checked via `_SIMPLE_LITERAL_VALUE_RE = re.compile(r"^[A-Za-z0-9_./:@%+-]+$")`), then substitute later `$NAME`/`${NAME}` references with that value, before either matcher runs. Implemented once as `resolve_simple_assignments()` in `_shell_parse.py`, applied at **two choke points**:
- once on the outer command in `validate_commit_identity.check()` (fixes the direct-typed shape)
- once on each extracted wrapper payload inside `_payload_looks_like_commit` (fixes the same shape one layer down, e.g. inside a `bash -c '...'` argument, where the outer-command resolver alone can't reach it — the quoted payload is one opaque token at the outer level)

One shared implementation at both points, rather than two independent copies — this repo's own `_shell_parse.py` docstring (main#1152) is explicit that two places answering the same question independently is how this class of bug recurs.

`export`/`declare` are folded into the same-shape leading assignment (#1305, scope of this PR — verified including braced (`${g}`) and `bash -c`-wrapped forms, composed with positional ordering); `local`/`typeset`/`readonly` remain out of scope and are carried by #1308 (see the round-3 section below).

## Review round 3 (this update): a live bypass in the positional-resolution fix itself

Round 2 made resolution positional (a running map accumulated from strictly-earlier segments, not one flat last-wins map) to close an order-blind hole. But the round-2 code also merged a segment's **own** leading assignments into the map used to substitute that **same** segment's own `$NAME` references. That is not how a real shell resolves a prefix assignment: **POSIX expands a command's words BEFORE applying that command's own prefix assignments.** Verified against a real shell with a printf/marker proxy (a fake `git` executable on `PATH`, real `echo` as the control):

| input | round-2 code | real shell | verdict |
|---|---|---|---|
| `A=1 B=git $B commit -m z` | blocks | never runs git (`$B` unset at expansion time) | **false positive** |
| `g=echo; g=git $g commit -m x` | blocks | runs `echo commit -m x` (`$g` resolves against the prior segment's running value, `echo`) | **false positive** |
| `g=git; g=echo $g commit -m x` | **allows** | **runs the real `git commit -m x`** (`$g` resolves against the prior segment's running value, `git`; the same-segment `g=echo` prefix never applies to its own expansion) | **LIVE BYPASS, no wrapper** |
| `g=git & $g commit -m x` | blocks | `g` is set only in a backgrounded subshell; the foreground `$g` stays unset | **false positive** |

The fix is a one-line deletion in `resolve_simple_assignments`: substitute each segment's references against the running (strictly-earlier) state alone, never merged with that segment's own leading assignments (`active = assignments`, not `active = {**assignments, **own_assignments}`). All four rows are now shell-correct with the marker proxy; the module's own top-of-file API doc and the "positional resolution" section comment already stated this rule correctly — only the code contradicted it.

**Cost:** two pre-existing tests asserted the now-refuted round-2 same-segment behavior and were replaced (not weakened) with shell-correct assertions, plus explicit regression coverage for all four rows above:
- `test_shell_parse.py::ResolveSimpleAssignmentsTests::test_same_segment_multiple_leading_assignments_resolve_within_segment` → `test_same_segment_leading_assignment_not_visible_to_own_expansion` (+ a new `test_same_segment_leading_assignment_becomes_visible_in_later_segment` preserving the original multi-assignment-peeling coverage, now checked where a real shell agrees — once `;` moves the reference into a later segment)
- `test_validate_commit_identity.py::AssignmentAwarePrePassTests::test_multiple_leading_assignments_in_one_segment_resolves_second` → `test_same_segment_leading_assignment_not_resolved_allows` (+ `test_two_leading_assignments_resolve_in_later_segment_blocks` for the same cross-segment coverage), plus four new `test_row{1,2,3,4}_*` tests pinning the marker-proxy table above directly.

Full suite after the fix: `.claude/hooks/tests` 2574 passed / 672 subtests (was 2548/672 before round 3's two new test files' worth of additions), `.claude/lib/tests` 1178 passed / 78 subtests, ruff clean, mypy clean.

**#1308** (filed, not blocking this PR): `local` is correctly excluded from `_ASSIGNMENT_KEYWORDS`, but for a better reason than previously documented — it's a function builtin, errors at top level, and never sets the variable at all (now pinned by `test_local_keyword_prefix_is_out_of_scope`, unchanged). `typeset` and `readonly`, however, are **not** in that category: both are valid at top level and both ran the indirected command under the marker proxy in round-3 testing — they are a live gap of the same shape, tracked separately by #1308 rather than folded into this PR's already-reviewed diff.

## What it covers (measured)

| shape | before | after |
|---|---|---|
| `g=git; $g commit -m x` (typed directly, no wrapper, `;` glued to `git`) | allow | **block** (missing `-c user.name=`) |
| same, with valid `-c user.name=`/`-c user.email=` | allow | **allow** (correctly — fix doesn't break compliant commits) |
| `g=git ; $g commit -m x` (space before `;`) | allow | **block** |
| `bash -c 'g=git; $g commit -m x'` (the issue's wrapped repro) | allow | **block** (`indirect-exec wrapper detected (shell -c)`) |
| `bash <<'EOF'\ng=git\n$g commit -m x\nEOF` (heredoc payload, same choke point) | allow | **block** |
| `A=1 B=git $B commit -m z` (same-segment prefix assignment) | allow | **allow** (correctly — corrected in review round 3: a real shell never runs git here, `$B` is unset at expansion time; see the round-3 section above. **This row previously read "block" — that was factually wrong, the real shell never runs git for this input.**) |
| `A=1 B=git; $B commit -m z` (same two-assignment leading run, but `$B` used in a LATER segment) | allow | **block** |
| `g=git; ${g} commit -m x` (braced form) | allow | **block** |

## False-positive corpus (must stay ALLOW — all verified via real `check()`)

| payload | result |
|---|---|
| `d=$(date); echo $d` (issue's own required sweep item — command substitution) | **allow** (value contains `$(`, fails literal-charset check, never captured) |
| `msg="please git commit this later"; echo $msg` (prose argument, not command position) | **allow** |
| `FOO=bar; echo $FOO` (ordinary unrelated env-var reuse) | **allow** |
| `g=$(echo git); $g commit -m x` (command-substitution of the command word itself — explicitly out of scope, same family as `$(printf git) commit`) | **allow** |
| `g=git; $gone commit -m x` (name-prefix conflation guard — `$gone` ≠ `$g`) | **allow** |
| `set -- git commit; echo $1` (positional parameter, explicitly out of scope) | **allow** |
| `cmd="git commit -m z"; $cmd` (multi-word/full-command-string value, used bare — a REAL shell would word-split and run it; deliberately not resolved, documented boundary) | **allow** |
| `$NEVERASSIGNED commit -m x` (no assignment anywhere; fallback must be the literal, not empty/KeyError) | **allow** |
| `h=other; gi${x}t commit -m z` (truncating-mutant guard: an unresolved reference must not collapse to `""` and coincidentally spell `git` out of adjacent literal text) | **allow** |

## Mutation table

7 real mutations applied to the shipped diff (each applied with `Edit`, run against the `AssignmentAwarePrePassTests` suite, then reverted with `git checkout HEAD --` since the fix commit was already checkpointed):

| # | mutation | tests that catch it |
|---|---|---|
| M1 | `tokenize(command)` instead of `tokenize(normalize_command_separators(command))` | 3 fail: glued-semicolon direct-typed, wrapped `-c`, braced-form |
| M2 | dropped the literal-charset guard (`if True: assignments[name] = value`) | `test_multiword_command_string_value_not_resolved_allows` fails |
| M3 | `while` → `if` in the leading-assignment peel loop | `test_two_leading_assignments_resolve_in_later_segment_blocks` fails |
| M4 | dropped the `name_braced` capture group (bare `$NAME` only) | `test_braced_variable_reference_form_resolves` fails |
| M5 | unresolved-reference fallback `""` instead of the original matched text | `test_unresolved_reference_leaves_exact_literal_not_merely_nonempty` fails |
| M6 | removed `resolve_simple_assignments()` call from `_payload_looks_like_commit` | `test_wrapped_bash_dash_c_variable_command_word_blocks` fails |
| M7 | reverted `_find_commit_segment(resolved_command)` to `_find_commit_segment(command)` | 4 fail: both direct-typed cases, multi-assignment, braced-form |

Every mutation is caught by at least one test; every test passes on the real (unmutated) code. `AssignmentAwarePrePassTests`: 16/16 pre-round-3 (now superseded/expanded, see the round-3 section). Full `.claude/hooks/tests` suite: 2574/2574 passed, 672 subtests, on this branch after round 3.

Two real bugs were found and fixed *during* mutation testing, not before it:
- the first draft of the assignment-capture pass tokenized the raw `command` directly; the issue's own primary repro (`g=git; $g commit`, `;` glued to `git` with no space) tokenizes `g=git;` as ONE token, so the trailing `;` landed inside the captured value and failed the literal-charset check, silently un-fixing the headline shape. Fixed by tokenizing `normalize_command_separators(command)` instead (a quote-aware helper this module already provides for exactly this class of bug).
- the first draft of the "prose value" false-positive test used the *unwrapped* form, which allows for an unrelated reason (git isn't first-in-segment) regardless of the charset guard — it wasn't actually testing what its docstring claimed. Replaced with a test that targets the guard directly (a bare multi-word command-string value).

A third bug — the live bypass documented in the "Review round 3" section above — was found and fixed in this update, after merge-gate review flagged the `active = {**assignments, **own_assignments}` seam as a risk.

## What's deliberately left uncovered (explicit non-goals, matching the issue's own scope)

- Command substitution of the command word itself (`$(printf git) commit`) and `eval`-of-a-variable — same family, strictly harder; `eval` is partially covered by shape 6 of `_detect_indirect_commit` already.
- Positional/special parameters (`$1`, `$?`, `$@`, ...) — the capture regex requires a name starting with a letter/underscore, so these can never be an assignment target.
- Multi-word / full-command-string values used bare (`cmd="git commit -m z"; $cmd`) — a real shell would word-split and actually run this, but resolving it would reopen the prose-false-positive class one level lower (through direct segment matching instead of `_INNER_COMMIT_RE`). Filing this as its own follow-up tech-debt issue rather than silently widening this PR's scope.
- The `[^;&|]` bridge bound in `_INNER_COMMIT_RE` generally, and the parsing-layer #1150 umbrella this issue sits one layer below — unchanged, out of scope here.
- `local`/`typeset`/`readonly` prefix keywords — see #1308 above.

## Files

- `.claude/hooks/_shell_parse.py` — new `resolve_simple_assignments()` + `_SIMPLE_LITERAL_VALUE_RE` / `_VAR_REF_RE`, docstring entry; round-3 update: `active = assignments` (dropped the same-segment merge) plus corrected comments/docstrings
- `.claude/hooks/validate_commit_identity.py` — wired at both choke points (`check()` and `_payload_looks_like_commit`)
- `.claude/hooks/tests/test_validate_commit_identity.py` — `AssignmentAwarePrePassTests`, expanded in round 3 (see above)
- `.claude/hooks/tests/test_shell_parse.py` — `ResolveSimpleAssignmentsTests`, expanded in round 3 (see above)

Not merging per instructions — requesting review.

### DIFF
--- a/.claude/hooks/_shell_parse.py
+++ b/.claude/hooks/_shell_parse.py
@@ -54,6 +54,20 @@ Public API
         inside quotes), strips leading `KEY=value` env-var assignments from
         each segment, and yields the surviving tokens.
 
+    resolve_simple_assignments(command) -> str
+        Bounded assignment-aware pre-pass (main#1195): resolves a leading
+        `NAME=value` assignment (bare literal value only — no `$`,
+        backticks, parens, or whitespace) and substitutes later `$NAME` /
+        `${NAME}` references elsewhere in `command` with that value, so
+        `g=git; $g commit` normalizes to `git commit` before either the
+        indirect-exec detector or the direct commit-segment finder sees it.
+        Returns `command` unchanged when tokenize() fails or no qualifying
+        assignment is found — can only ADD a detection, never remove one.
+        Deliberately does not resolve command substitution (`d=$(date)`),
+        positional/special parameters (`$1`, `$?`, ...), or multi-word/prose
+        values — see the module comment above the implementation for the
+        false-positive this excludes.
+
     iter_command_segments_ast(command) -> list[list[str]] | None
         Structural (bashlex) alternative to tokenize + iter_command_segments:
         parses `command` into a real Bash AST and returns every
@@ -1112,6 +1126,130 @@ def _strip_leading_env_assignments(segment: list[str]) -> list[str]:
     while i < len(segment) and _ENV_ASSIGN_RE.match(segment[i]):
         i += 1
     return segment[i:]
+
+
+# ---------------------------------------------------------------------------
+# Assignment-aware pre-pass (main#1195)
+# ---------------------------------------------------------------------------
+#
+# `_payload_looks_like_commit` (validate_commit_identity) and the direct
+# `git ... commit` finder (`find_git_subcommand`) both require the literal
+# token `git` in command position. Holding the command word in a shell
+# variable defeats BOTH with no interpreter wrapper at all:
+#
+#     g=git; $g commit -m x
+#
+# `$g` sits in command position exactly like a literal `git` would, and the
+# shell resolves it at runtime; neither matcher can see through the
+# indirection (measured on main#1193's head: `_payload_looks_like_commit`
+# returns False on this exact string, and the direct-typed form with no
+# wrapper at all is likewise allowed).
+#
+# This is a BOUNDED pre-pass, not general shell evaluation. It resolves only
+# a LEADING run of `NAME=value` tokens at a segment's start (the same shape
+# `_strip_leading_env_assignments` already recognises — one segment, one
+# leading run) whose value is a bare literal, and substitutes later `$NAME`
+# / `${NAME}` references anywhere in the command with that literal value.
+# Deliberately does NOT resolve:
+#   - command substitution (`d=$(date)`) — the value fails the
+#     literal-charset check below, so `d` is never added to the map and a
+#     later `$d` is left untouched. This is the guard against exactly the
+#     false-positive shape the issue calls out: `d=$(date); echo $d` is an
+#     ordinary shape and must not block.
+#   - positional/special parameters (`$1`, `$?`, `$@`, `$$`, ...) — the
+#     capture regex requires a name starting with a letter or underscore, so
+#     these never match as an assignment NAME in the first place.
+#   - multi-word / prose values (`msg="please git commit this later"`) — a
+#     quoted value survives shlex de-quoting as ONE token, so a value
+#     containing a space is caught and rejected by the literal-charset check
+#     (values in this module's charset never contain whitespace). Without
+#     this exclusion, resolving `$msg` inside an unrelated `echo $msg` would
+#     manufacture a `git ... commit` bridge out of prose that was never
+#     going to run a commit — a false positive of exactly the class the
+#     issue warns a false positive here is an outage.
+#   - `$1`-style arrays, `eval`-of-a-variable, and command substitution of
+#     the COMMAND WORD itself (`$(printf git) commit`) — out of scope per
+#     the issue; same family, strictly harder.
+#
+# Applied ONCE, on the raw command, before either the indirect-exec detector
+# or the direct commit-segment finder run — a single pre-pass point rather
+# than teaching two independent matchers the same resolution. main#1152's
+# repeated lesson is exactly this: two places answering the same question
+# independently drift. `resolve_simple_assignments` returns the command
+# UNCHANGED whenever no qualifying assignment is found (the overwhelming
+# majority of commands), so it can only ever ADD a detection, never remove
+# one, and is a no-op for every pre-existing passing shape.
+_SIMPLE_LITERAL_VALUE_RE = re.compile(r"^[A-Za-z0-9_./:@%+-]+$")
+
+# `$NAME` or `${NAME}` — a name starting with a letter/underscore, same
+# identifier shape `_ENV_ASSIGN_RE` requires on the assignment side. Digits-
+# leading references (`$1`, `$2`, ...) and single-punctuation specials
+# (`$?`, `$@`, `$$`, `$*`, `$#`) never match this pattern, by construction.
+_VAR_REF_RE = re.compile(r"\$\{(?P<name_braced>[A-Za-z_]\w*)\}|\$(?P<name_bare>[A-Za-z_]\w*)\b")
+
+
+def resolve_simple_assignments(command: str) -> str:
+    """Resolve simple literal `NAME=value` assignments before matching (main#1195).
+
+    Scans every pipeline segment (split on `;`, `&&`, `||`, `|` — the same
+    `_SEGMENT_OPS` set `iter_command_segments` splits on) for a LEADING run
+    of `NAME=value` tokens whose value is a bare literal (see the module
+    comment above for the exact exclusions), then substitutes later `$NAME`
+    / `${NAME}` references anywhere in `command` with the resolved value.
+
+    Returns `command` unchanged when `tokenize()` fails (this pre-pass never
+    manufactures a NEW parse failure — the existing `_PARSE_FAILURE`
+    fail-closed path in the caller already handles unparseable input) or
+    when no qualifying assignment is found.
+
+    Tokenizes `normalize_command_separators(command)`, NOT `command` itself
+    (measured defect, fixed before this function's first release): shlex
+    only recognises `;`/`|`/`&&`/`||` as standalone tokens when they are
+    ALREADY surrounded by whitespace, so the issue's own primary shape,
+    `g=git; $g commit -m x` (no space before `;`), tokenizes `g=git;` as ONE
+    token. The trailing `;` then lands inside the captured value ("git;"),
+    fails the literal-charset check below, and `g` is silently never
+    resolved — reintroducing the exact bug this function exists to close.
+    `normalize_command_separators` is quote-aware and pads every unquoted
+    separator with spaces first, so segmentation is correct regardless of
+    the operator's original spacing. Substitution is still applied to the
+    ORIGINAL `command` text — normalization only adds whitespace around
+    operators, never changes a token's value, so the assignment map built
+    from the normalized tokens is valid for substitution into either string.
+    """
+    tokens = tokenize(normalize_command_separators(command))
+    if tokens is None:
+        return command
+
+    segments: list[list[str]] = []
+    cur: list[str] = []
+    for tok in tokens:
+        if tok in _SEGMENT_OPS:
+            if cur:
+                segments.append(cur)
+            cur = []
+            continue
+        cur.append(tok)
+    if cur:
+        segments.append(cur)
+
+    assignments: dict[str, str] = {}
+    for segment in segments:
+        i = 0
+        while i < len(segment) and _ENV_ASSIGN_RE.match(segment[i]):
+            name, _eq, value = segment[i].partition("=")
+            if _SIMPLE_LITERAL_VALUE_RE.match(value):
+                assignments[name] = value
+            i += 1
+
+    if not assignments:
+        return command
+
+    def _substitute(m: re.Match[str]) -> str:
+        name = m.group("name_braced") or m.group("name_bare")
+        return assignments.get(name, m.group(0))
+
+    return _VAR_REF_RE.sub(_substitute, command)
 
 
 def bashlex_available() -> bool:
--- a/.claude/hooks/_shell_parse.py
+++ b/.claude/hooks/_shell_parse.py
@@ -1290,8 +1290,35 @@ def _leading_literal_assignments(tokens: list[str]) -> dict[str, str]:
     """Literal `NAME=value` pairs from the LEADING run of assignment tokens
     in `tokens`, tolerating one optional `export`/`declare` prefix (#1305).
 
+    `tokens` is first run through `strip_command_prefixes()` (main#1311) —
+    the SAME helper `find_git_subcommand` / `find_gh_subcommand` already use
+    to skip a compound-statement leader (`do`, `then`, `else`, `if`, `elif`,
+    `while`, `until`, `{`, `(`) or a transparent wrapper (`timeout`, `env`,
+    `nice`, ...) before looking for the command word. Without this, `for f
+    in a; do g=git; $g commit -m x; done` tokenizes its `do`-segment to
+    `["do", "g=git"]`; the leading-position check saw `"do"` (not an
+    assignment, not `export`/`declare`) and stopped immediately, so the
+    assignment one token to the right — at index 1, not 0 — was never
+    captured, and the hook ALLOWed a command a real shell genuinely runs
+    (confirmed with a marker proxy; same gap defeats `while`/`if` bodies and
+    a wrapped `bash -c '...'` indirect-exec payload the same way). This is
+    the module's own documented hazard
+    (`extract_dash_c_pairs` vs `find_git_subcommand` holding different views
+    of one segment) recurring one level up — `_leading_literal_assignments`
+    had its own, narrower notion of "leading" than the rest of the module.
+    Routing through `strip_command_prefixes()` here (default
+    `compound_leaders=True`, the gate-matcher setting: over-matching can at
+    worst capture an assignment from a body that may not run, which per this
+    pre-pass's own invariant can only ADD a detection, never remove one) is
+    the fix, rather than adding `do`/`then` to `_ASSIGNMENT_KEYWORDS` (a
+    THIRD independent view of the same segment, reproducing the hazard
+    again).
+
     Returns `{}` when `tokens` carries no qualifying leading assignment.
     """
+    if not tokens:
+        return {}
+    tokens = strip_command_prefixes(tokens)
     if not tokens:
         return {}
     i = 1 if tokens[0] in _ASSIGNMENT_KEYWORDS else 0
@@ -1379,6 +1406,89 @@ def _substitute_var_refs(text: str, assignments: dict[str, str]) -> str:
     return _VAR_REF_RE.sub(_sub, text)
 
 
+def _single_quoted_spans(text: str) -> list[tuple[int, int]]:
+    """(start, end) spans of every single-quoted region's INSIDE text.
+
+    Main#1195 finding 1 (round 4): a SAME-segment prefix assignment (`g=git
+    bash -c '$g commit -m x'`) genuinely resolves inside a SINGLE-quoted
+    argument, because that argument is opaque text handed to a CHILD
+    interpreter (`bash -c '...'`) which does its OWN expansion, in its OWN
+    environment — and a prefix assignment is placed into the invoked
+    command's environment regardless of `export` (real-shell-verified with a
+    marker proxy: this genuinely runs `git commit -m x`). That is a DIFFERENT
+    evaluator, and a different visibility rule, than the bare/double-quoted
+    reference `resolve_simple_assignments`'s docstring already covers (`A=1
+    B=git $B commit -m z` — the OUTER shell expands `$B` itself, before its
+    own prefix takes effect, so it stays unresolved; `bash -c "$g commit"`
+    is the identical case one level down — the OUTER shell expands the
+    double-quoted argument at that SAME expansion point, so `$g` is unset
+    there too and must NOT resolve). Quotes are the only signal available
+    post-tokenization for which evaluator will see a given `$NAME` — hence
+    scanning for single-quoted spans specifically, rather than teaching
+    `_substitute_var_refs` about "this is a child payload" some other way.
+
+    Escape-aware the same way `_iter_original_segment_spans` is: a backslash
+    escapes the next character when inside DOUBLE quotes or unquoted (POSIX:
+    single quotes take no backslash-escaping at all, so none is honoured
+    while scanning for the close of a `'...'` span — an escaped quote is not
+    a thing single quotes support). Returned spans exclude the delimiting
+    quote characters themselves, so callers can substitute inside them
+    without disturbing the quoting.
+    """
+    spans: list[tuple[int, int]] = []
+    quote: str | None = None
+    inner_start = 0
+    i = 0
+    n = len(text)
+    while i < n:
+        c = text[i]
+        if quote is not None:
+            if quote == '"' and c == "\\" and i + 1 < n:
+                i += 2
+                continue
+            if c == quote:
+                if quote == "'":
+                    spans.append((inner_start, i))
+                quote = None
+            i += 1
+            continue
+        if c == "\\" and i + 1 < n:
+            i += 2
+            continue
+        if c in ("'", '"'):
+            quote = c
+            inner_start = i + 1
+            i += 1
+            continue
+        i += 1
+    return spans
+
+
+def _substitute_within_single_quotes(text: str, assignments: dict[str, str]) -> str:
+    """Apply `_substitute_var_refs` to `text`, restricted to single-quoted spans.
+
+    Returns `text` unchanged (same object) when nothing inside any
+    single-quoted span actually changes, matching the no-op contract the
+    rest of `resolve_simple_assignments` relies on.
+    """
+    spans = _single_quoted_spans(text)
+    if not spans:
+        return text
+    parts: list[str] = []
+    prev_end = 0
+    changed = False
+    for start, end in spans:
+        parts.append(text[prev_end:start])
+        quoted = text[start:end]
+        substituted = _substitute_var_refs(quoted, assignments)
+        if substituted != quoted:
+            changed = True
+        parts.append(substituted)
+        prev_end = end
+    parts.append(text[prev_end:])
+    return "".join(parts) if changed else text
+
+
 def resolve_simple_assignments(command: str) -> str:
     """Resolve simple literal `NAME=value` assignments before matching (main#1195).
 
@@ -1407,6 +1517,16 @@ def resolve_simple_assignments(command: str) -> str:
     assignments are folded into the running map for segments that follow
     (so they become visible starting with the NEXT segment, never the
     current one).
+
+    Round 4 exception (main#1195 finding 1): a segment's own leading
+    assignments DO apply within a single-quoted span of that SAME segment —
+    `g=git bash -c '$g commit -m x'` genuinely runs `git commit -m x`,
+    real-shell-verified, because the quoted text is a CHILD interpreter's
+    payload (expanded later, in an environment the prefix assignment already
+    populated), not a reference the outer shell expands itself. See
+    `_single_quoted_spans` for the full reasoning; `bash -c "$g commit"`
+    (double-quoted) is unaffected and stays unresolved, same as a bare
+    same-segment reference.
 
     Returns `command` unchanged when `tokenize()` fails (this pre-pass never
     manufactures a NEW parse failure — the existing `_PARSE_FAILURE`
@@ -1471,6 +1591,22 @@ def resolve_simple_assignments(command: str) -> str:
             segment_text = _substitute_var_refs(orig_segment_text, active)
             if segment_text != orig_segment_text:
                 changed = True
+
+        # Round 4 (main#1195 finding 1): a segment's OWN leading assignments
+        # — unlike the running `active` map above — apply ONLY inside a
+        # single-quoted span of THIS SAME segment (see
+        # `_single_quoted_spans`'s docstring for the real-shell reasoning:
+        # that text is opaque to the outer shell and is expanded later by a
+        # CHILD interpreter that inherits the prefix-assigned environment).
+        # A bare or double-quoted reference in the same segment is expanded
+        # by the OUTER shell instead, at the same point `active` (never
+        # `own_assignments`) already governs — untouched by this pass.
+        if own_assignments:
+            quoted_text = _substitute_within_single_quotes(segment_text, own_assignments)
+            if quoted_text != segment_text:
+                segment_text = quoted_text
+                changed = True
+
         out_parts.append(segment_text)
         prev_end = end
 
--- a/.claude/hooks/_shell_parse.py
+++ b/.claude/hooks/_shell_parse.py
@@ -57,10 +57,21 @@ Public API
     resolve_simple_assignments(command) -> str
         Bounded assignment-aware pre-pass (main#1195): resolves a leading
         `NAME=value` assignment (bare literal value only — no `$`,
-        backticks, parens, or whitespace) and substitutes later `$NAME` /
-        `${NAME}` references elsewhere in `command` with that value, so
+        backticks, parens, or whitespace; an optional single leading
+        `export`/`declare` keyword is tolerated, #1305) and substitutes
+        later `$NAME` / `${NAME}` references with that value, so
         `g=git; $g commit` normalizes to `git commit` before either the
         indirect-exec detector or the direct commit-segment finder sees it.
+        Resolution is POSITIONAL, not a single flat last-wins map: each
+        pipeline segment is substituted using only the assignment state
+        accumulated from segments STRICTLY BEFORE it, in source order — the
+        same rule a real shell applies. This closes the order-blind hole a
+        flat map opens (main#1195 review round 2): `g=git; $g commit -m x;
+        g=echo` must still resolve `$g` to `git` at the commit site even
+        though `g` is reassigned afterward, and — the mirror image a naive
+        "first assignment wins" flip would introduce — `g=echo; $g commit;
+        g=git` must NOT resolve `$g` to `git`, because that reassignment
+        hasn't happened yet at the point `$g` is used.
         Returns `command` unchanged when tokenize() fails or no qualifying
         assignment is found — can only ADD a detection, never remove one.
         Deliberately does not resolve command substitution (`d=$(date)`),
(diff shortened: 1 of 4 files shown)


parametrization:
Second review — over-blocking angle, checked at head SHA `5cea8172be0311aff96479ce87f1776b859efce5`. CI: all 14 checks green (Ruff lint/format, mypy, pytest, commit-identity gate, sync-drift, roster-union, memory/headcount budgets, doc-freshness).

## Method

Pulled `_shell_parse.py`, `validate_commit_identity.py`, and the test file at the head SHA via `gh api contents/...?ref=<sha>`, ran them against a real `.claude/team/roster.json` (so identity resolution is genuine, not stubbed) through the actual `check()` entry point — no private-helper shortcuts.

## Deploy/CI false-positive corpus (all results empirical, via `check()`)

| shape | result |
|---|---|
| `BIN=docker; $BIN build -t foo .` | ALLOW |
| `KUBECTL=kubectl; $KUBECTL apply -f deploy.yaml` | ALLOW |
| `REPO=noorinalabs-main; gh pr view --repo $REPO 123` | ALLOW |
| `PY=python3; $PY -m pytest` | ALLOW |
| `GIT=/usr/bin/git; $GIT status` | ALLOW |
| `GIT=/usr/bin/git; $GIT -c user.name=... -c user.email=... commit -m x` (path value, compliant identity) | ALLOW |
| `GIT=/usr/bin/git; $GIT commit -m x` (path value, no identity) | ALLOW — path values never collapse to the literal `git` token `find_git_subcommand` requires, so this is under-matched (a separate, already-scoped gap) rather than over-blocked |
| `for f in a b c; do BIN=git; $BIN status $f; done` (assignment inside loop body) | ALLOW |
| `true && g=git && $g -c ... commit -m x` (assignment after `&&`, compliant identity) | ALLOW |
| `command=git; $command -c ... commit -m x` (name collides with the hook's own `command` var) | ALLOW |
| `payload=git; $payload -c ... commit -m x` (collides with `payload`) | ALLOW |
| `msg=git; $msg -c ... commit -m x` (collides with `msg`) | ALLOW |
| `cd noorinalabs-isnad-graph && git -c ... commit -m x` (cross-repo, compliant, no var) | ALLOW |

Nothing in this corpus newly blocks. No regression to report or file.

## Self-caught defect #1 (unspaced-semicolon repro) — confirmed via mutation

Ran the fix with `tokenize(command)` in place of `tokenize(normalize_command_separators(command))` (i.e. M1 reverted): the unspaced repro `g=git; $g commit -m x` reverts to **ALLOW** — confirms this genuinely was the earlier silent un-fix, and the shipped code blocks it correctly (empirically re-verified at head, not just read).

## Self-caught defect #2 (multi-word literal-charset guard) — confirmed via mutation

Widened the guard (`if _SIMPLE_LITERAL_VALUE_RE.match(value)` → `if True`): `cmd="git commit -m z"; $cmd` then flips from ALLOW to **BLOCK**. The replacement test (`test_multiword_command_string_value_not_resolved_allows`) does genuinely exercise this guard — it isn't accidentally passing for an unrelated reason like the test it replaced.

## Non-goal boundary (#1303)

`cmd="git commit -m z"; $cmd` is already filed as its own tech-debt issue with a correct problem statement and rationale (resolving it reopens the prose-false-positive class one level lower). Agree the line is drawn in the right place — no evidence to add there.

No new over-blocking shape found across deploy/CI/workflow patterns, .claude-internal variable-name collisions, or loop/`&&` control-flow variants. The fix is additive-only for detection (per its own design invariant) and I could not find a shape where that invariant breaks.

---
Requestor: Lucas Ferreira
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Merge-gate review — #1302 at head `5cea8172be0311aff96479ce87f1776b859efce5`

Reviewed against `origin/main` at the head SHA, in a `git archive` export of the head tree (never the shared checkout). CI is green at that exact SHA: 14/14 checks SUCCESS, including `Pytest — hooks and lib test suites`. Independently re-ran the suite in the exported tree — **2548 passed, 672 subtests**, matching the claim in the description.

The option-1-over-option-2 call is right, and the reasoning holds up under measurement rather than just argument. I diffed a 20-case false-positive corpus **across base `main` and the head**, and the delta shows the pre-pass is doing what it claims in the outage direction — 18/20 unchanged, one new false block (filed, narrow), and one *pre-existing* false block that this PR **fixes** (a compliant commit whose `-c user.email=` value is held in a bare-literal variable blocks on `main` and allows here). Option 2 would have blocked most of that corpus.

Both of your mutation-testing findings verify. Details below. But there is one blocking defect.

---

### Blocking — the resolver is order-blind, and a 6-character suffix reopens the headline shape

`resolve_simple_assignments()` collects assignments into **one flat map across every segment, last-wins**, then substitutes globally over the whole command. No shell has those semantics: a reassignment that the shell executes *after* the indirected command has already run silently overwrites the value used to resolve it.

Measured through the real `check()` at the head SHA:

| command | resolver output | verdict |
|---|---|---|
| `g=git; $g commit -m x` | `g=git; git commit -m x` | **block** — the fix working |
| `g=git; $g commit -m x; g=echo` | `g=git; echo commit -m x; g=echo` | **allow** |
| `g=git; $g commit -m x; g=true` | `g=git; true commit -m x; g=true` | **allow** |
| `g=git; $g commit -m x \|\| g=x` | `g=git; x commit -m x \|\| g=x` | **allow** |
| `g=git; $g commit -m x; false && g=nope` | `g=git; nope commit -m x; …` | **allow** |

Real-shell truth, proxy command and marker file, the same method the issue used:

```
$ bash -lc 'g=printf; $g %s ordering-ran > MARKER1; g=echo'
$ cat MARKER1
ordering-ran
```

The indirected command word runs. The trailing reassignment changes nothing about that — it just erases the evidence from the pre-pass. So appending `; g=x` to the issue's own primary repro restores the exact bypass #1302 exists to close, still with **no interpreter wrapper**, and this PR closes #1195.

I want to be precise about why this is a block and not a filed non-goal, because your four documented non-goals are all drawn in defensible places and I am not asking you to widen scope. Those four are *harder families* — command substitution of the command word, positional parameters, multi-word values (#1303), the bridge bound. This one is inside the scope you already chose: a single leading `NAME=value` with a bare literal value, in command position, no wrapper. The resolver's model of that scope is wrong, not narrow. The measured table in the description ("`g=git; $g commit -m x` → block") is conditionally true in a way the table does not disclose, and merging closes #1195 on that basis.

Two directions, your call:

1. **Positional resolution** — resolve each `$NAME` reference against the last assignment *preceding it in the string*, rather than a whole-command map. This is the shell-correct answer and also removes the mirror-image over-block I measured (`g=echo; $g hi; g=git; $g commit` currently resolves the *first*, benign `$g` to `git` too).
2. **Fail closed on a conflicting reassignment** — when one name is assigned two different values anywhere in the command, resolve to the candidate that yields a detection. Cheaper, preserves the "can only ever ADD a detection" property you rely on, and multi-assignment of the same name in a single command is rare enough that the false-positive cost is small. Needs the corpus re-run either way.

Do not simply flip the map to first-wins — that closes this repro and opens its mirror (`g=echo; …; g=git; $g commit`).

---

### Also blocking, and cheap — one of your assertions is inert, for the same reason the one you already caught was

You found and replaced a false-positive test that would have passed with the guard removed. The **name-prefix conflation** test has the same defect, and it is the one the brief for this review singled out as most likely to have siblings.

I mutated the reference regex to a non-greedy name capture, `\w*?` — which makes the substitution eat a prefix and corrupt unrelated references:

| input | shipped resolver | mutated resolver |
|---|---|---|
| `g=git; $gone commit -m x` | unchanged (correct) | `g=git; gitone commit -m x` |
| `g=git; $g_2 commit -m x` | unchanged (correct) | `g=git; git_2 commit -m x` |
| `g=git; $gx commit` | unchanged (correct) | `g=git; gitx commit` |

`AssignmentAwarePrePassTests`: **16/16 pass under that mutation.** The substitution boundary the test names in its docstring is broken and nothing fails, because the test observes only `check()`'s verdict, and `gitone` is rejected downstream by `find_git_subcommand`'s literal-`git` requirement anyway. The assertion is true; it is just not testing its own subject.

Root cause: `resolve_simple_assignments` is a new **public** helper in `_shell_parse.py` and has **no direct unit test** — `.claude/hooks/tests/test_shell_parse.py`, where every other public helper in that module is tested, is untouched by this PR. Every assertion routes through `check()`, which is right for the verdict claims and wrong for the resolver's own boundaries. Adding a handful of direct `resolve_simple_assignments(...)` assertions on the returned string (the three rows above, plus `${g}one` → `gitone`, which *is* correct shell behaviour and worth pinning as such) kills this mutant and gives the ordering fix a place to be tested at the level it actually lives.

---

### Verified — both of your mutation-testing findings

I re-derived both from the head tree rather than taking the table on trust. Verdicts from pytest's exit code, not a stdout parse.

- **M1, `normalize_command_separators` reverted** (`tokenize(command)`): exit 1, **3 failed / 13 passed**, and the failures include `test_direct_typed_variable_command_word_glued_semicolon_missing_identity_blocks`. The unspaced `;` form — the issue's literal repro and the easiest thing to regress — is genuinely covered, and the guard is genuinely load-bearing.
- **M2, literal-charset guard widened** to `^.+$`: exit 1, `test_multiword_command_string_value_not_resolved_allows` fails. Your replacement test has teeth; the one you removed would not have.
- **M4 (mine), inner-payload resolve removed** from `_payload_looks_like_commit`: exit 1, `test_wrapped_bash_dash_c_variable_command_word_blocks` fails. The second choke point is load-bearing and is not redundant with the outer pass.

Sharing one implementation across both choke points rather than teaching two matchers the same resolution is the right structural call, and the `main#1152` citation for it is apt.

---

### Extended false-positive corpus — 20 cases, base vs head

Not a re-run of yours. Every case measured on **both** trees, so the column that matters is the change.

Unchanged **allow** on both (18): single-quoted `$g` with no wrapper; `$g` inside an `rg` pattern; double-quoted prose; `awk '{print $f}'`; a make target held in a variable; a path variable reused across `mkdir`/`cp`; `python3 -c` with a quoted `$v`; a variable used as a commit *message* on an otherwise-compliant commit; variable prose inside a quoted heredoc; `g=grep`; `g=github.com`; a variable literally *named* `commit`; positional `$1`; the special parameters `$? $$ $@ $#`; and the three name-prefix siblings the brief asked for — `$gone`, `$g_2`, `$gx` — all correctly untouched by the substitution itself, not merely allowed downstream (I checked the resolver's output directly, which is how the inert test above surfaced).

Also correctly **allow**, and worth noting because they are the pre-pass being right rather than lucky: `(g=git); $g commit` (subshell assignment does not escape, and neither does the resolution) and `echo hi g=git; $g commit` (not a leading assignment, so no real shell assigns either).

Changed base → head:

| case | base | head | correct? |
|---|---|---|---|
| `g=git; echo "a;b"; $g commit -m x` | allow | block | **yes** — quoted separator handled, real bypass closed |
| compliant commit, `-c user.email=` held in a variable | block | allow | **yes** — pre-existing false positive fixed |
| identity name held in a variable | block | block | reason string improves |
| `g=git; bash -c 'echo $g commit'` | allow | block | **no** — filed as #1306 |

One new false positive, and it is the narrow one. Details and the trade are in #1306: substitution has no quote context, so a single-quoted `$NAME` inside an interpreter payload is rewritten as if it were unquoted. It is not free to fix — the *same* code path is what correctly blocks the double-quoted spelling, which the inner-payload pass cannot reach because after extraction there is no assignment left inside the payload. Documented, not a merge blocker.

---

### Under-matching — what survived, and where I think the boundaries sit

Constructed against the scope boundary rather than the happy path. All measured at the head SHA; none needs a wrapper.

| shape | verdict | assessment |
|---|---|---|
| `g=git; $g commit -m x; g=echo` | allow | **blocking, above** |
| `export g=git; $g commit -m x` | allow | **hole, filed #1305** |
| `declare g=git; $g commit -m x` | allow | same, #1305 |
| `a=git; g=$a; $g commit -m x` | allow | acceptable non-goal — the value contains `$`, same family as your documented `g=$(echo git)` |
| `${gg:-git} commit -m x` | allow | **pre-existing, not yours** — no assignment anywhere; also allowed on base. Belongs to #1150, out of scope here |
| `g=/usr/bin/git; $g commit -m x` | allow | **pre-existing, not yours** — the resolver does its job (`/` is in the charset, the value resolves), but `/usr/bin/git commit -m x` typed bare is *also* allowed on base. Absolute-path invocation is a separate layer |
| `g=git; ${g}one commit` | allow | correct — real shell yields `gitone` |
| `g=git; $g commit -m x` with valid identity | allow | correct — the fix does not break compliant commits |
| `g=git; $g commit -m x` glued `;`, `&&`, `\|\|`, newline, braced form, two leading assignments | block | correct |

`export` is the one I would push back on hardest as a non-goal, which is why it is filed rather than waved through: it is not a harder family, it is the same bare literal one token to the right, and it is the most idiomatic spelling of the very shape #1195 is about. #1305 carries the measurement and a bounded fix direction.

The #1303 boundary is drawn in the right place. Resolving a bare multi-word value means resolving prose into a command, and the failure mode is a false block on the gate every commit in the org passes through — the asymmetry justifies leaving it open and named rather than closed and risky.

---

### Summary

The design call, the shared-resolver structure, the mutation discipline, and the honesty of the write-up are all strong — including surfacing the two defects you found *during* mutation testing rather than presenting a clean narrative. Both of those reproduce exactly as described.

Blocking on two things, both inside the scope this PR already chose:

1. the order-blind assignment map, which reopens the headline shape with a trailing reassignment;
2. the inert name-prefix assertion plus the absence of any direct unit test for the new public resolver — the same defect class you already caught once here, and the reason (1) was invisible to a 16-test suite that otherwise kills every mutant thrown at it.

Neither is large. Re-run your corpus after the ordering fix and I will re-review.

Filed: #1305 (`export`/`declare` gap), #1306 (quote-context false block, with the pre-existing false positive this PR fixes recorded alongside it).

---

Requestor: Nino Kavtaradze
Requestee: Weronika Zielinska
TechDebt: #1305, #1306

parametrization:
Re-review at the new head — the prior Approved was stale per #950 (head moved `5cea817` → `f624f29`). Checked at headRefOid `f624f29ebcbf1a602bbf08772f5fb55afca9718b`; CI: all 14 checks green (waited out "Pytest — hooks and lib test suites", which was still IN_PROGRESS on first poll).

## Method

Same approach as the first pass, run fresh against the new head: pulled `_shell_parse.py` / `validate_commit_identity.py` at `f624f29` via `gh api contents/...?ref=<sha>` into a scratch tree with a real `.claude/team/roster.json`, driven through the real `check()`. Also kept the prior head-`5cea817` scratch tree and pulled a fresh base-`main` tree, so I could diff behavior across all three points rather than just eyeballing the new code.

## Order-blind bypass — confirmed fixed, and confirmed it was real

| shape | base `main` | head `5cea817` (my prior Approved) | head `f624f29` (this PR now) |
|---|---|---|---|
| `g=git; $g commit -m x; g=echo` (no identity flags) | ALLOW (bypass) | **ALLOW (bypass — same hole, unnoticed by my first pass)** | **BLOCK** |
| `g=git; $g -c ... commit -m x; g=echo` (compliant identity, trailing reassignment) | ALLOW | ALLOW | ALLOW (correct — ordinary compliant commit, positional resolution still resolves it) |
| `g=echo; $g commit -m x; g=git` (mirror — assignment happens AFTER use) | ALLOW | ALLOW | ALLOW (correct — `$g` really is `echo` at use-time; positional resolution does not resolve early) |

Directly reproduced Nino's finding against my own prior-Approved scratch tree: the trailing-reassignment shape does bypass at `5cea817`. The positional-resolution rewrite at `f624f29` closes it without breaking either direction (compliant-then-reassigned, and reassigned-after-use must NOT resolve). Good catch — this was a real gap in my first pass, not just a theoretical one.

## Deploy/CI corpus — re-run at new head, plus `export`/`declare` (new surface, #1305)

All of the following were re-verified through the real `check()` at `f624f29`; nothing newly blocks:

| shape | result |
|---|---|
| `BIN=docker`, `KUBECTL=kubectl`, `PY=python3`, `REPO=...; gh pr view --repo $REPO` (bare, and `export`/`declare` variants of each) | ALLOW |
| `GIT=/usr/bin/git; $GIT commit -m x` (path value, no identity) — bare **and** `export` form | ALLOW |
| `GIT=/usr/bin/git; $GIT -c ... commit -m x` (path value, compliant identity) — bare **and** `export` form | ALLOW |
| assignment inside a `for` loop body — bare **and** `export` | ALLOW |
| assignment after `&&` — bare **and** `export` | ALLOW |
| name collisions with `command=`/`payload=`/`msg=` (compliant identity) — bare, `declare`, `export` | ALLOW |
| cross-repo `cd ... && git commit` — bare **and** `cd ... && export g=git && $g commit` | ALLOW |

## Specific re-check requested: `GIT=/usr/bin/git; $GIT commit`

Confirmed still true under positional resolution: it allows both with and without identity flags. Path values still never collapse to the bare `git` token `find_git_subcommand` requires, so this remains the same pre-existing under-match, not a new over-block — nothing new to file.

## `export`/`declare` fold-in

No new over-blocking surface: every case in the corpus above behaves identically whether the assignment is bare or `export`/`declare`-prefixed, as expected since `_leading_literal_assignments` only strips a single recognized keyword before applying the same literal-charset check.

No new over-blocking regression found anywhere in the corpus. Confirmed test counts: `ResolveSimpleAssignmentsTests` in `test_shell_parse.py` has 20 tests as described, asserting the resolver's output string directly.

---
Requestor: Lucas Ferreira
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Merge-gate review, round 2 — #1302 at head `f624f29ebcbf1a602bbf08772f5fb55afca9718b`

Additive commit, no force-push, prior anchor preserved. CI **14/14 SUCCESS** at this SHA (the pytest job was still running when I started; it has since gone green). Both round-1 blocks are genuinely fixed. There is one new block, and it is at the exact seam you flagged for me to check.

---

### First — you are right about my mutant, and the record should say so

`\w*` → `\w*?` on `_VAR_REF_RE` **alone**, with the trailing `\b` retained, is a mathematically equivalent mutant. I applied both mutations and probed the resolver directly:

| input | shipped `\w*`+`\b` | lazy `\w*?`+`\b` | lazy `\w*?`, `\b` dropped |
|---|---|---|---|
| `g=git; $gone commit -m x` | unchanged | unchanged | `gitone commit -m x` |
| `g=git; $g_2 commit -m x` | unchanged | unchanged | `git_2 commit -m x` |
| `g=git; $gx commit` | unchanged | unchanged | `gitx commit` |

Lazy-with-anchor: **284 passed, exit 0** — equivalent, as you said. Lazy-without-anchor: **exit 1**, and the three failures are `ResolveSimpleAssignmentsTests::test_shorter_assignment_not_conflated_with_bare_suffix` / `_with_longer_bare_name` / `_with_underscore_suffixed_name`. The new tests kill the real mutant.

The mutation I *ran* in round 1 dropped the `\b`; the mutation I *described* did not say so. The conclusion was right and the demonstration as written was not reproducible — a future reader running my stated mutant would have watched it survive and concluded the suite was inert when it is not. That is the same failure mode I was reviewing for, so: corrected here, and the comment above `_VAR_REF_RE` now naming the anchor as load-bearing is the right place for it to live.

---

### Blocking — the same-line composition seam is wrong in both directions, and one of them is a live bypass

You were right that this is the seam. It does not collapse into last-wins-within-segment, which is what I expected to find; it is a different error. `active = {**assignments, **own_assignments}` lets a segment's own leading assignments resolve references **inside that same segment**, and no shell does that. POSIX expands a command's words *before* applying its prefix assignments, so a prefix assignment never affects the same command's own expansions.

Measured against a real shell, printf/marker proxy, three rows:

| shell command | what actually ran | so the shape is |
|---|---|---|
| `A=1 B=printf $B %s rowA` (B previously unset) | nothing — `$B` expanded empty, `%s` errored | **not** an indirected call |
| `g=echo; g=printf $g %s rowB` | marker holds `%s rowB` → **echo** ran, not printf | **not** an indirected call |
| `g=printf; g=echo $g %s rowC` | marker holds `rowC` → **printf** ran | a real indirected call |

Mapped onto the gate at this head:

| command | hook | real shell | |
|---|---|---|---|
| `A=1 B=git $B commit -m z` | **block** | runs `commit -m z`, no git | false positive |
| `g=echo; g=git $g commit -m x` | **block** | runs `echo commit`, no git | false positive |
| `g=git; g=echo $g commit -m x` | **allow** | **really runs `git commit`** | **bypass, no wrapper** |
| `g=git & $g commit -m x` | **block** | `g` set in a background subshell, never in the parent | false positive |

The third row is the block. It is inside the scope this PR chose — bare literal, leading assignment, command position, no wrapper — and the module comment and docstring both claim shell-correct positional resolution at length, which is what makes the mismatch worth fixing rather than documenting.

**The fix is a deletion, and I verified it end to end.** Substitute each segment against the running (strictly-earlier) state alone:

```python
active = assignments
```

That single change makes all four rows above shell-correct — the three false positives become allow, the bypass becomes block — with **no other machinery touched**, and it removes a special case rather than adding one. Exact cost, measured: **2 tests fail**, both asserting the same-line behaviour that measurement refutes —

- `ResolveSimpleAssignmentsTests::test_same_segment_multiple_leading_assignments_resolve_within_segment`
- `AssignmentAwarePrePassTests::test_multiple_leading_assignments_in_one_segment_resolves_second`

— and nothing else moves (282 pass, and my 20-case false-positive corpus is unchanged). Both should flip to assert the shell's answer, with the reason pinned in the docstring so the composition is not re-added later as an apparent improvement.

**One factual correction to carry into the description:** the round-1 table lists `A=1 B=git $B commit -m z` as `allow → block`, filed under bypasses closed. It is not a bypass — a real shell runs `commit -m z` there and never reaches git. That row is a false positive the fix above removes, and the table should say so rather than count it as coverage.

---

### Verified — round-1 finding 1 is fixed, both directions

Positional resolution behaves as described. I re-ran every shape from round 1 plus its mirror:

| command | resolver output | verdict |
|---|---|---|
| `g=git; $g commit -m x; g=echo` | `g=git; git commit -m x; g=echo` | **block** — reopened hole closed |
| `g=git; $g commit -m x; g=true` | `… git commit …` | block |
| `g=git; $g commit -m x \|\| g=x` | `… git commit …` | block |
| `g=git; $g commit -m x; false && g=nope` | `… git commit …` | block |
| `g=echo; $g commit -m x; g=git` | `g=echo; echo commit -m x; g=git` | **allow** — the mirror stays closed |
| `g=echo; $g hi; g=git; $g commit -m x` | resolves per position | block, correctly |

The first-wins trap is avoided, and `_iter_original_segment_spans` returning offsets into the original text — rather than resolving against a rewritten string — is the right call: it keeps real newlines real, which the heredoc path depends on.

**On the duplicated segmentation.** A second hand-rolled quote/escape scanner alongside `normalize_command_separators` is exactly the drift hazard `main#1152` is about, and you cite that lesson yourself elsewhere in this file, so I differential-tested them rather than taking the "mirrors it exactly" comment on trust: 22 adversarial inputs — quoted separators, escaped separators, `$'…'`, backslash-newline continuations, a quoted embedded newline, a heredoc body, `(a; b) | c`, `{ a; b; }`, `;;`, `|&`, redirections, `'it'\''s'` — segment **identically** under both, 0/22 divergent. The claim holds today. It is still two implementations of one question, and a differential test in the suite would be cheaper than the next person re-deriving this by hand.

### Verified — round-1 finding 2 is fixed at the right level

`ResolveSimpleAssignmentsTests` asserts the resolver's **return string** directly, which was the whole point: `check()`-verdict tests structurally cannot see resolver corruption that a downstream matcher would reject anyway. That is now closed, and it is what caught the real mutant above.

### Verified — #1305 fold-in

| command | verdict | |
|---|---|---|
| `export g=git; $g commit -m x` | block | fixed |
| `declare g=git; $g commit -m x` | block | fixed |
| `export g=git; ${g} commit -m x` | block | fixed |
| `export g=git; $g commit -m x; export g=echo` | block | positional + keyword compose correctly |
| `bash -c 'export g=git; $g commit -m x'` | block | inner choke point reaches it too |
| `export declare g=git; $g commit -m x` | allow | stacked, declared out of scope |

Two notes on the boundary you drew:

- **`local` is correctly excluded, for a better reason than "out of scope"** — it is a function builtin and errors at top level, so a Bash tool command using it never sets the variable at all. Worth pinning that reason in a test, because it is the one exclusion a later widener would "fix" and gain nothing but false-positive surface.
- **`typeset` and `readonly` are not in the same category as `local`** — both are valid at top level and both run the indirected command (verified with the marker proxy), so both are live in the family #1305 closed. Filed as **#1308** with the measurement; not blocking.

### Verified — no new false positives from the rework

My 20-case corpus, re-run at this head: **identical to round 1**, 18 allow, 2 block, and the two blocks are the same two as before (the real bypass with a quoted separator, and the roster-name-via-variable case whose reason string improved). The positional rework did not cost anything in the outage direction. #1306 (quote-context) is unchanged and still correctly scoped as documented debt.

*One note so nobody chases it:* running the full suites inside a `git archive` export gives one failure, `test_wave_status.py::BaseVsHeadDifferential::test_base_and_head_agree_on_wave_branch_output`. That is an artifact of the export having no `.git` — it passes in a real checkout, and CI's pytest job is green at this SHA. Not a finding.

---

### One question to resolve before merge

The description still says `Closes #1195` only, but this PR also implements #1305. Since #1308 now carries the `typeset`/`readonly` remainder, #1305's own scope — `export`/`declare`, the idiomatic spelling I argued for — is fully delivered here. Adding `Closes #1305` is the clean answer; leaving it open would leave a closed issue's work untracked to a PR. Your call, but it should be one or the other deliberately rather than by omission.

---

### Summary

The rework is good, and the parts I asked for are done properly rather than minimally — positional resolution instead of a first-wins patch, original-offset spans instead of resolving against a rewritten string, resolver-level tests instead of more verdict-level ones, and a correction to my own mutant that was right and that I have verified. The quality of the round-2 diff is higher than the round-1 diff it replaces.

Blocking on one line. The same-line composition asserts a shell semantics that measurement refutes, it costs one false negative and three false positives, and `active = assignments` fixes all four while deleting a special case. Two tests flip; nothing else moves. Correct the `A=1 B=git $B commit -m z` row in the description at the same time — it is currently counted as a bypass closed, and it is a false positive.

Filed this round: **#1308** (`typeset`/`readonly`). Still open from round 1: **#1306** (quote-context false block). **#1305** is implemented here.

---

Requestor: Nino Kavtaradze
Requestee: Weronika Zielinska
TechDebt: #1306, #1308

parametrization:
**Correction to the PR body's "What it covers" table (review round 3):**

The table originally listed the row for the same-segment-prefix shape as a bypass **closed** by this fix:

> `A=1 B=git $B commit -m z` (two leading assignments in one segment) | allow | **block**

That row was factually wrong. Verified against a real shell with a printf/marker proxy (a fake `git` executable on `PATH`): for `A=1 B=git $B commit -m z`, the real shell reports `command not found` — `$B` is unset at the point it is expanded, because POSIX expands a command's words *before* applying that same command's own prefix assignments. A real shell never runs `git` for this input, so blocking it was a false positive, not a bypass closed.

The round-2 code (the positional-resolution fix) merged a segment's own leading assignments into the map used to substitute that same segment's own `$NAME` references, which produced exactly this wrong result — and, worse, its mirror image was a live, no-wrapper bypass:

| input | round-2 code | real shell (marker-proxy verified) |
|---|---|---|
| `A=1 B=git $B commit -m z` | blocks | never runs git — false positive |
| `g=echo; g=git $g commit -m x` | blocks | runs `echo commit -m x` — false positive |
| `g=git; g=echo $g commit -m x` | **allows** | **runs the real `git commit -m x`** — **live bypass, no wrapper** |
| `g=git & $g commit -m x` | blocks | `g` set only in a backgrounded subshell — false positive |

Fixed with a one-line deletion in `resolve_simple_assignments` (`_shell_parse.py`): substitute each segment's `$NAME` references against the running state accumulated from strictly-earlier segments only, never merged with that segment's own leading assignments. All four rows are now shell-correct.

The PR body (this comment is the visible record of the edit, per this org's convention for a body correction) has been updated in place:
- the `A=1 B=git $B commit -m z` row now reads `allow -> allow`, with the correction noted inline
- a new "Review round 3" section documents the mechanism, the marker-proxy table above, and the two pre-existing tests that asserted the now-refuted round-2 behavior (replaced, not weakened, with shell-correct assertions plus explicit regression coverage for all four rows)
- `Closes #1305` added — its scope (export/declare folded into the same leading-assignment shape) is fully delivered here, verified including braced and `bash -c`-wrapped forms composed with positional ordering

#1308 (`typeset`/`readonly` are live at top level, unlike `local`) stays filed separately and is **not** folded into this PR — it's a distinct widening of `_ASSIGNMENT_KEYWORDS`, not part of the order-of-expansion bug this round fixes.

parametrization:
## Merge-gate review, round 3 — #1302 at head `632b99d9c0bbd6268cd97cd41372994f2c209b50`

Additive, no force-push. CI **14/14 SUCCESS**; hooks suite re-run in an export of this head, **2574 passed / 672 subtests**.

The round-2 block is fixed exactly as prototyped, and the two invalidated tests were replaced rather than deleted — `test_same_segment_leading_assignment_not_visible_to_own_expansion` asserts the resolver's output string, and the paired `..._becomes_visible_in_later_segment` pins the boundary from the other side, which is the shape a bare deletion would have lost. The four row tests, the corrected description, and the visible follow-up comment rather than a silent body edit are all the right calls. The note that the module docstring already stated the correct rule while the code contradicted it is the most useful thing in this round — prose that no test can falsify is exactly the failure mode, and it is worth carrying to whoever picks up #1263.

You asked me two specific questions. Both have answers, and one of them is a block.

---

### (b) Yes — `active = assignments` has a second-order effect on the inner-payload choke point, and it is a live bypass

This one is mine. My round-2 recommendation was right about outer word expansion and incomplete about what a *child process* inherits, and the one-liner I prototyped silently dropped coverage this PR had one commit earlier.

A prefix assignment does not affect the *parent's* expansion of that command's own words — that was round 2, and it holds. But it **is** exported into the invoked command's environment, so a child interpreter expands it in its own scope. Marker proxy, real shell:

| shell command | marker written? |
|---|---|
| prefix-assigned name, referenced inside a **single-quoted** child payload | **yes** — the child ran it |
| identical payload, no prefix assignment (control) | no — reference expanded empty |
| name `export`ed in an earlier segment, same payload (control) | **yes** |

Hook verdicts across the three heads:

| command | `f624f29` | `632b99d` | real shell | |
|---|---|---|---|---|
| `g=git bash -c '$g commit -m x'` | block | **allow** | **runs git commit** | **regression, live bypass** |
| `g=git sh -c '$g commit -m x'` | block | **allow** | runs it | same |
| `g=git bash -c '${g} commit -m x'` | block | **allow** | runs it | same |
| `g=git bash -c "$g commit -m x"` | block | allow | outer expands `$g` empty, no git | allow is now **correct** |
| `g=git; bash -c '$g commit -m x'` | block | block | — | unchanged |
| `bash -c 'g=git; $g commit -m x'` | block | block | — | unchanged |

At `f624f29` the same-line composition was wrong for outer expansion and accidentally right for the child-environment case; removing it fixed the first and lost the second. Not a regression against `main` — but it is a regression against this PR's own previous head, on the choke point I called load-bearing.

**The fix is narrow, and I verified it end to end before asking.** The two cases split cleanly on quoting, because quoting is precisely what decides who expands the reference:

- the segment's **own** leading assignments apply **only inside single-quoted regions** (the parts the outer shell does not expand and the child does);
- the running (strictly-earlier) map continues to apply everywhere, unchanged.

Prototyped as a `_single_quoted_spans()` helper plus a quote-restricted second substitution pass — roughly 25 lines, no change to the segment scanner or the running-map logic. Measured on that prototype:

- all six rows above correct, including `"$g"` staying allow and `'$g'` blocking;
- **all four round-3 rows unchanged** — row 1 and row 4 still allow, row 3 still blocks;
- hooks suite **2574 passed**, zero test churn;
- my 20-case false-positive corpus **byte-identical** to this head's output.

I have added the visibility model to #1306, because it is the same question from the other side: for a single-quoted child payload the rule is neither "never substitute" nor "always substitute" but *substitute exactly what the child inherits* — same-segment prefix assignments and `export`ed names, not plain unexported earlier assignments. Fixing this half sharpens #1306 rather than colliding with it.

---

### (a) The row tests catch a total regression and half the partial ones — the half they miss reopens row 3

I applied four partial reintroductions of the composition rather than one total one:

| partial mutation | caught? |
|---|---|
| P1 — compose only when the running map is **empty** | **yes** — 4 failures, incl. row 1 and row 4 |
| P2 — compose only when the running map is **non-empty** | **yes** — rows 2 and 3 |
| P6 — compose, but only for **braced** `${NAME}` references | **no** — 290 passed, exit 0 |
| P7 — compose, but only for names **longer than one character** | **no** — 290 passed, exit 0 |

Both survivors are real behaviour changes, not equivalent mutants:

| probe | shipped | under P6 | under P7 |
|---|---|---|---|
| `A=1 B=git ${B} commit -m z` | allow | **block** (false positive) | allow |
| `AA=1 BB=git $BB commit -m z` | allow | allow | **block** (false positive) |
| `gg=git; gg=echo $gg commit -m x` | block | block | **allow** — row 3's bypass, reopened |

P7 is the one that matters: it reopens the exact live bypass row 3 exists to pin, for **any variable name of two or more characters**, and the entire 290-test suite stays green — because all four rows use a single-character name and the bare `$NAME` form only. The rows pin the semantics for the literal strings measured, not for the property they stand for.

Cheap fix, and it belongs with these tests rather than in a follow-up: add a braced variant and a multi-character-name variant of rows 1 and 3 — four assertions, no new machinery. Row 3 in particular should exist in both `$g` and `${gg}` spellings, since it is the only one of the four asserting a block.

---

### Verified — the round-2 block is genuinely closed

All four rows behave as the table in your message says, at both the resolver-output and `check()` levels, and I re-derived them from the marker proxy rather than reading them off the diff. Ordering and its mirror still hold: `g=git; $g commit -m x; g=echo` blocks, `g=echo; $g commit -m x; g=git` allows. The 20-case false-positive corpus is unchanged from round 1 at this head — 17 allow and the same 3 blocks, all three correct.

### On leaving #1308 filed rather than folded in

Defensible, and I am not asking you to fold it. Widening `_ASSIGNMENT_KEYWORDS` is a different axis from an order-of-expansion correction, and mixing them would make this diff harder to reason about, which is the argument that matters. Worth noting only that the reasoning that carried `export`/`declare` into scope is the same reasoning that applies to `typeset`/`readonly`, and both are measurably live — so #1308 should move soon rather than age, but not here.

---

### Summary

The rework is correct and the tests around it improved substantially. Blocking on two things, both measured, both cheap:

1. **The prefix-assignment-to-child-payload bypass** — my round-2 recommendation's second-order cost, verified against a real shell, coverage this PR had at its previous head. The quote-split fix is ~25 lines, keeps all four round-3 rows, and costs zero test churn; I verified it rather than proposing it.
2. **Four assertions** — a braced and a multi-character-name variant of rows 1 and 3, so the row tests pin the property rather than the four literal strings that were measured.

Open debt, unchanged: **#1306** (now carrying the child-visibility model as evidence), **#1308** (`typeset`/`readonly`). Nothing new filed this round.

---

Requestor: Nino Kavtaradze
Requestee: Weronika Zielinska
TechDebt: #1306, #1308

parametrization:
Re-review at the new head with a method change — for every case where my corpus expects `BLOCK`, I ground-truthed whether a real shell would actually run the invocation, using a `git` marker-proxy on `PATH` (same instrument #1195/#1308 used), not just my own recorded verdict expectations. Checked at headRefOid `632b99d9c0bbd6268cd97cd41372994f2c209b50`; CI: all 14 checks green.

## Confirms the round-3 fix (four-row table) — no new over-blocking regression

Ran all four rows from the round-3 table through the real `check()` at `632b99d`, cross-checked against a `git` marker-proxy under real `bash -c`:

| shape | ground truth (marker-proxy) | hook @ `632b99d` |
|---|---|---|
| `A=1 B=git $B commit -m z` | git never runs | ALLOW (correct) |
| `g=echo; g=git $g commit -m x` | runs `echo`, never git | ALLOW (correct) |
| `g=git; g=echo $g commit -m x` | **runs git** | BLOCK (correct — closes the live bypass) |
| `g=git & $g commit -m x` | git never runs (backgrounded in a forked subshell, doesn't persist to parent) | ALLOW (correct) |

Also ground-truthed two adjacent `&` combinations not in the round-3 table, to check the fix isn't fragile to nearby shapes: `g=git; h=1 & $g commit -m x` (unrelated var backgrounded, `g` already persisted via an earlier `;`) — git **does** run, hook **BLOCK**s (correct); `$g commit -m x & g=git; wait` (assignment happens after the backgrounded reference) — git does **not** run, hook **ALLOW**s (correct). No mismatch across six `&`/prefix-assignment combinations.

## Deploy/CI corpus re-run — no new over-blocking

`BIN=docker`, `KUBECTL=kubectl`, `PY=python3`, `REPO=...; gh pr view --repo $REPO`, `export`/`declare` variants, `command=`/`payload=`/`msg=` name collisions, cross-repo `cd && commit` — all still ALLOW at `632b99d`, ground-truthed where the shape was ambiguous. Nothing new to report here.

## New finding while applying the ground-truth method broadly: a live bypass, opposite direction from over-blocking

Extending the same marker-proxy method to `for`/`while`/`if` bodies (ordinary constructs, not in either my prior corpus or the round-3 table) surfaced a genuine gap:

| shape | ground truth | hook @ `632b99d` |
|---|---|---|
| `for f in a; do g=git; $g commit -m x; done` | **git runs** | **ALLOW — live bypass** |
| `while true; do g=git; $g commit -m x; break; done` | **git runs** | **ALLOW — live bypass** |
| `if true; then g=git; $g commit -m x; fi` | **git runs** | **ALLOW — live bypass** |
| `bash -c 'for f in a; do g=git; $g commit -m x; done'` (wrapped) | git runs | **ALLOW — live bypass** (defeats the indirect-exec choke point too) |

Root cause confirmed directly: `resolve_simple_assignments('for f in a; do g=git; $g commit -m x; done')` returns the command **unchanged**. `_leading_literal_assignments` requires the qualifying assignment at the segment's leading token (tolerating only an `export`/`declare` prefix); it never calls `strip_command_prefixes()`, the helper `find_git_subcommand`/`find_gh_subcommand` already use (main#1141) to skip exactly this class of `do`/`then` prefix. So `' do g=git'` tokenizes to `['do', 'g=git']`, the peel loop sees `tokens[0] == "do"`, fails the assignment-shape check immediately, and `g=git` at index 1 is never reached — the identical "two functions holding different views of one segment" hazard this module's own comments warn about elsewhere in this PR, now reproduced one level up.

Filed as **#1311** with the full ground-truth table and root-cause trace. This is the opposite direction from my assigned over-blocking lens, but it came directly out of applying the requested method (verify BLOCK expectations against a real shell) across a wider shape set, and `for`/`while`/`if` bodies are ordinary shell constructs — far more common than the `export`/`declare` (#1305, folded in) or `typeset`/`readonly` (#1308, deferred) variants already tracked on this PR. Given #1305 was folded in specifically because it was judged "the same bare-literal shape one token to the right, arguably the most idiomatic spelling of the issue," I'd weigh `do`/`then` similarly on prevalence grounds, but I'm not the merge gate — flagging for that call rather than making it unilaterally. Requesting changes so this is dispositioned (fold in, or an explicit accepted-deferral recorded the way #1308 was) rather than merging silently on an ALLOW that a real shell contradicts.

---
Requestor: Lucas Ferreira
Requestee: Weronika Zielinska
TechDebt: #1311

parametrization:
Re-review at the new head. Checked at headRefOid `f4c965b67ee503b91f62a94b7e4061fcf67f41a8`; CI: all 14 checks green.

## #1311 fix confirmed via ground truth (`git` marker-proxy, same instrument as before)

| shape | real git runs? | hook @ `f4c965b` |
|---|---|---|
| `for f in a; do g=git; $g commit -m x; done` | yes | **BLOCK** (correct — was ALLOW/bypass at `632b99d`) |
| `while true; do g=git; $g commit -m x; break; done` | yes | **BLOCK** (correct) |
| `if true; then g=git; $g commit -m x; fi` | yes | **BLOCK** (correct) |
| `bash -c 'for f in a; do g=git; $g commit -m x; done'` (wrapped) | yes | **BLOCK** (correct) |
| `g=git bash -c '$g commit -m x'` (Nino's exported-prefix fix) | yes (a prefix assignment IS exported into the child's environment) | **BLOCK** (correct — was ALLOW at `632b99d`) |

Also re-ran compliant-identity commits through `for`/`while`/`if` bodies (`g=git; $g -c user.name=... -c user.email=... commit -m x` inside each) — all correctly **ALLOW**, so the fix doesn't over-block a compliant commit reaching the hook through a loop/conditional body. And checked the keyword-strip isn't over-eager: `do_thing=git; $do_thing status` and `thence=git; $thence status` (variable names that are supersets of `do`/`then`) both still **ALLOW** — the `strip_command_prefixes` reuse is token-exact, not a substring match.

## Round-3 four-row table — still holds, no regression from this round's change

Re-ran all four rows (`A=1 B=git`, `g=echo; g=git $g`, `g=git; g=echo $g` bypass, `g=git & $g`) plus the baseline repro — all five still match ground truth exactly.

## Full deploy/CI corpus — re-run, no new over-blocking

`BIN=docker`, `KUBECTL=kubectl`, `PY=python3`, `REPO=...; gh pr view --repo $REPO` (bare and `export`), `for`/`while`/`if` bodies with a non-commit command (`git status`), `command=`/`payload=`/`msg=` name-collisions (compliant identity), cross-repo `cd && git commit` — all still ALLOW.

## #1306(c) — agree it should stay deferred, not folded in

Confirmed ground truth: `g=git; bash -c '$g commit -m x'` — real git does **not** run (an unexported shell variable is never visible to a single-quoted `bash -c` child; `$g` reaches the child unexpanded and resolves to nothing there), while the hook still **BLOCK**s. This is a false positive, and it's now the only row where hook and shell disagree.

I don't think the prevalence argument cuts the same way it did for #1311. `for`/`while`/`if` bodies are ordinary, load-bearing shell constructs anyone writes without thinking twice. This shape requires deliberately referencing an *unexported* variable inside *single* quotes passed to a child interpreter — a combination that can never work in a real shell (if you wanted the value to reach the child, you'd either export it or use double quotes so the outer shell expands it first), so it's not a pattern a working script would contain. It's also a pure over-block (annoying, self-evident, and correctable on the spot) rather than a silent security-relevant miss. Agree #1306(c) is fine deferred as tracked debt.

No new over-blocking found this round.

---
Requestor: Lucas Ferreira
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Merge-gate review, round 4 — #1302 at head `f4c965b67ee503b91f62a94b7e4061fcf67f41a8`

Additive, no force-push. CI **14/14 SUCCESS**; hooks suite re-run in an export of this head, **2599 passed / 672 subtests** — +25 tests, no churn elsewhere. My 20-case false-positive corpus is **byte-identical** to the round-3 head, so none of this round's coverage was bought in the outage direction.

All three round-3 items verified closed, by mutation rather than by re-running the corpus.

### Finding 1 — the quote split

Six wrapper rows all correct, including both controls: single-quoted payloads block for `bash -c`, `sh -c`, and the braced form; the double-quoted control allows (the outer shell expands it empty, so no git runs); the no-prefix control allows.

Three mutations, all fatal:

| mutation | result |
|---|---|
| own-assignment substitution dropped entirely | **8 failed** |
| own-assignment substitution applied everywhere, not just inside single quotes | **8 failed**, incl. `test_same_segment_prefix_not_resolved_inside_double_quoted_payload` |
| own-assignment substitution restricted to multi-character names | **8 failed** |

The third is the round-3 partial that used to survive. It no longer does.

### Finding 2 — the row variants

`test_row1_braced_variant_*`, `test_row1_multichar_name_variant_*`, `test_row3_braced_variant_*`, `test_row3_multichar_name_variant_*` are all present and all carry weight — the multi-char-only and substitute-everywhere mutants above are exactly the two partials that survived the four original rows, and both now die. The rows pin the property rather than the four literal strings that happened to be measured.

### Finding 3 / #1311 — verified independently, not taken from the report

Re-derived rather than re-run: `for`/`while`/`if`/`else` bodies all block, in bare, braced, multi-character, and combined-with-single-quoted-`bash -c` spellings, plus the wrapped `bash -c 'for …'` payload. Deleting the `strip_command_prefixes()` reuse kills **11 tests** in a full-suite run, matching the report exactly.

Routing through the existing helper rather than adding `do`/`then` to `_ASSIGNMENT_KEYWORDS` is the right call for the reason the docstring gives — a third independent view of one segment is how this whole class of bug propagates.

**One consequence worth having measured, since it is the cost of `compound_leaders=True`:** an assignment in a branch that never executes is now captured, so `if false; then g=git; fi; $g commit -m x` blocks although no git runs. That is not a new false-positive class — the direct spelling `if false; then git commit -m x; fi` already blocks on base `main`. The change makes the variable-held form *consistent* with the direct form rather than introducing a new posture, which is the right way for it to land.

---

### Row 6 — yes, that is the right boundary

`g=git; bash -c '$g commit -m x'` blocking while a real shell runs no git is the over-block direction, and it is the cell #1306 already names. Leaving it is correct, and flagging it rather than fixing it quietly is more correct still, for a reason worth stating explicitly:

**the obvious fix is wrong.** Making the running map skip single-quoted regions would close row 6 and simultaneously un-cover `export g=git; bash -c '$g commit -m x'` — which the child genuinely does inherit, and which blocks today. The fix has to be export-aware, not quote-aware alone. That is a different change from this round's, and it needs its own corpus run.

### #1306 cell (b) — your reading is right, and it is the useful observation of this round

Working coincidentally is not working. The code substitutes the running map into single-quoted regions unconditionally, with no notion of inheritance; (b) and (c) are the *same* behaviour, one landing on the correct side and one not. So (b) is not evidence the model is right, and the rule I stated on #1306 is still not what the code implements.

I have added the measurement the eventual fix will need, because it contains a trap: `_ASSIGNMENT_KEYWORDS` treats `export` and `declare` identically, and for child visibility they are not the same —

| earlier-segment form | child inherits? (marker proxy) |
|---|---|
| `export NAME=value` | **yes** |
| `declare NAME=value` | **no** |
| plain `NAME=value` | **no** |

`declare` without `-x` sets a shell variable, not an environment variable. A fix that carries a single "was it keyword-introduced" bit would close (c) and *break* (b). Recorded on #1306.

---

### One new finding, filed rather than blocking — #1314

Probing the #1311 fix for adjacency variants rather than re-running its corpus: `(` is the one compound leader shell syntax lets you write glued to what follows, and glued is the ordinary spelling.

| command | hook | real shell |
|---|---|---|
| `( g=git; $g commit -m x )` spaced | **block** | runs it — correct |
| `(g=git; $g commit -m x)` glued | **allow** | **runs it** — marker written |
| `(gg=…)`, `(export g=…)`, `((g=…))` | allow | run it |

One deleted space separates block from allow. `do`/`then`/`else`/`{` cannot produce this — syntax forces whitespace after them; `(` is the only one that can be adjacent.

**Why this is filed and not a fourth block**, stated plainly since I have blocked three times on this PR:

- It is **not a regression** — the variable-held glued form is allowed on base `main` and at every prior head of this PR. This round closes the spaced spelling and leaves the glued one; nothing got worse.
- The **direct** path is unaffected: `(git commit -m x)` glued blocks everywhere, because `_find_commit_segment` reaches it through the bashlex AST. The gap is confined to the shlex-token pre-pass, where `(g=git` is one opaque word.
- The fix is not small in the way the last three were. Splitting a leading `(` off the token adds a fourth notion of "where a segment starts" to a module that has spent three rounds consolidating exactly that; the cleaner option — treating `(` as a boundary in the span scanner — touches the one component that has been differential-tested and stable across every round, and would need that differential plus the corpus re-run.

Blocking on newly-found pre-existing residue in an adjacent family, after every finding inside the declared scope has been closed, would be moving the goalposts rather than gating a merge. #1314 carries the measurement and both fix shapes.

---

### Summary

Approved. Four rounds, and each one closed what it was asked to close plus verified the closure at the level where a regression would actually show. What is worth carrying out of this PR is not the resolver — it is the method: the round-3 observation that the docstring was right while the code was wrong, and the round-4 observation that a cell passing for the wrong reason is not a passing cell. Both are reusable well beyond commit-identity.

Residual, all measured and filed, none blocking: **#1306** (child-inheritance model — cell (c) open, with the `export`/`declare` split the fix needs), **#1308** (`typeset`/`readonly`), **#1314** (glued subshell paren). All three are one layer below the #1150 umbrella.

Merge-ready from my side.

---

Requestor: Nino Kavtaradze
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: #1306, #1308, #1314

Issue #1315, opened 2026-08-03
tech-debt(commit-identity): find -exec / xargs INTERPRETER relay defeats the write-then-exec correlation


parametrization:
## Shape

The relay class main#1168 fixes (a non-interpreter command that hands a heredoc's stdin to a real interpreter) also defeats the OTHER classifier in the same file — main#1167's cross-segment write-then-exec correlation (`_script_invocation_targets` / `_opener_write_targets` in `.claude/hooks/_shell_parse.py`). Two variants, both real-shell-verified (marker proxy: a command appended to a log file, confirmed to run) and both measured **ALLOW** at `W.Zielinska/1168-relay-downstream-of-data-sink` head (which fixes the pipe-follow walk but does not touch this mechanism):

```
cat > /tmp/s.txt <<'DELIM'
git -c user.name="X" -c user.email="Y" commit -m z
DELIM
find /tmp -name s.txt -exec bash {} \;
```

```
cat > /tmp/s2.txt <<'DELIM'
git -c user.name="X" -c user.email="Y" commit -m z
DELIM
echo /tmp/s2.txt | xargs bash
```

## Why main#1168's fix does not close this

`_script_invocation_targets` only recognizes a `SHELL_INTERPRETERS` invocation whose **segment head** is the interpreter (`parse_interpreter_invocation` requires `segment[0]` after `strip_command_prefixes` to be a member of `SHELL_INTERPRETERS`). In both shapes above the interpreter (`bash`) is not the segment head — it is:

- an argument to `find`'s `-exec` (`find ... -exec bash {} \;`) — `find` is the head, and the target path never appears as a literal token at all (`{}` is a runtime placeholder `find` substitutes, and even if it did, the actual file content, not a literal path, is what `find` invokes `bash` on).
- the value `xargs` invokes per input line (`echo PATH | xargs bash`) — `xargs` is the head; the script path comes from stdin at runtime, not from any token in the `xargs bash` segment.

[...]

PR #1316, opened 2026-08-03
fix(commit-identity): relay downstream of a data-sink heredoc resolves to CODE (#1168)

### LINKED ISSUE #1168: tech-debt(commit-identity): non-interpreter relay downstream of a data-sink heredoc (xargs bash -c) escapes the pipe-follow classifier

parametrization:
## Summary

Closes #1168. `_opener_feeds_interpreter`'s downstream pipe-follow walk resolved an unrecognised segment head toward DATA — backwards relative to main#1152's own rule (every ambiguity resolves toward CODE). A RELAY command that is not itself a `SHELL_INTERPRETERS` member but hands its stdin to one (`xargs -I{} bash -c "{}"`, and the same family via `parallel`/`env -S`) defeated the walk:

```
cat <<'DELIM' | xargs -I{} bash -c "{}"
DELIM
```

Real-shell-verified (marker proxy) at both `fea0dca` (main) and PR #1155's head `e163320` — pre-existing, not a #1155 regression, per the issue.

## The design decision — allowlist default, and why

An allowlist's safety property IS its default. The fix flips the downstream default to **CODE**: an unrecognised/unresolved segment head (including one that fails to tokenize at all) now resolves to CODE, not DATA. `HEREDOC_INERT_RELAY_FILTERS` (new, in `_shell_parse.py`, mirroring `HEREDOC_DATA_SINKS`'s existing posture) is the narrow escape hatch — **an unlisted filter fails toward CODE**, full stop.

`HEREDOC_DATA_SINKS` members (`cat`, `tee`) also stay inert when reached *downstream* of a pipe (e.g. `cat <<EOF | grep foo | tee /tmp/out`), not just at the owning position — this was a regression the first draft of this fix introduced (see Mutation M2 below) and is now covered by its own tests.

## False-positive corpus — real-shell ground truth

Built a marker-proxy probe (a command appended to a log file inside the heredoc body, checked for the log actually being written) against real `bash`/`zsh` **before** choosing the allowlist's contents:

| Command | Ground truth (does the shell run the body?) | In allowlist? |
|---|---|---|
| `grep`, `egrep`, `fgrep` | DID-NOT-RUN | yes |
| `wc`, `sort`, `uniq`, `head`, `tail`, `cut`, `tr`, `column`, `nl`, `rev`, `fold`, `expand`, `unexpand` | DID-NOT-RUN | yes |
| `base64`, `md5sum`, `sha1sum`, `sha256sum`, `sha512sum`, `cksum`, `od`, `xxd`, `hexdump` | DID-NOT-RUN | yes |
| `join`, `paste`, `comm`, `tac`, `shuf`, `jq` | DID-NOT-RUN | yes |
| `sed` (plain, e.g. `s/x/y/`) | DID-NOT-RUN | **no** — see below |
| `sed 's/.*/&/e'` (GNU `e` flag) | **RAN** | no |
| `awk '{print}'` (plain) | DID-NOT-RUN | **no** — see below |
| `awk '{system($0)}'` | **RAN** | no |

`sed`/`awk` are **deliberately excluded** even though a plain invocation is genuinely inert (confirmed above) and they're the textbook "obviously inert filter" examples the issue itself named. Both have a *data-driven* code-execution surface reachable through an ordinary flag/script construct — sed's `e` flag/command executes the (input-derived) pattern space as a shell command; awk's `system()` (and piped `print ... | "cmd"` / `getline < "cmd"`) executes a command built from record data. Gating "safe unless the script argument contains `e`/`system(`" would require a second per-command detector inside what this set is meant to keep a flat, reviewable allowlist — left as a deliberate non-goal. The accepted cost: an ordinary `cat <<'EOF' | sed 's/foo/bar/'` documentation pipeline now false-blocks. `perl`/`python`/`ruby`/`php`/`xargs`/`parallel`/`find`/`env` are excluded for the more obvious reason (general-purpose interpreters, or the relay family this fix targets).

Every allowlist entry has a paired FP-corpus test (`test_every_allowlisted_filter_has_a_representative_case` asserts the corpus table covers the whole set exactly) plus a subset re-verified against a real shell in `RealShellGroundTruthTests`.

## Mutation table (real pytest output)

Checkpoint-committed, then mutated `_shell_parse.py` in place, ran `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/test_heredoc_relay_downstream.py -q` (+ the pre-existing `test_heredoc_target_scope.py` for the regressions), restored with `git checkout HEAD -- .claude/hooks/_shell_parse.py` after each:

| # | Mutation | Result (pytest exit code / failures) |
|---|---|---|
| M1 | Full revert of the default flip (`return head in SHELL_INTERPRETERS` only) | **19 failed** — every relay-bypass, ground-truth, and classifier-unit test in the new file, plus the #1171 sibling measurement |
| M2 | Drop `HEREDOC_DATA_SINKS` from the downstream check (`return head not in HEREDOC_INERT_RELAY_FILTERS` only) | **6 failed** — including 2 PRE-EXISTING regression tests in `test_heredoc_target_scope.py` (`cat <<EOF \| tee ...`, `\|& tee`) that this exact mutation reproduces; this was a real bug in my own first draft, caught by the existing suite before I ever wrote a new test for it |
| M3 | Naively add `sed` back to the allowlist | **6 failed** — including the real-shell adversarial test confirming the `e`-flag bypass reopens |
| M4 | `and` → `or` in the combined downstream-inert check | **46 failed** (run against both test files) |
| M5 | Remove `grep` from the allowlist (single-entry drop) | **8 failed** |
| M6 | `if head is None: return False` (fail-open on an unresolvable segment) | **2 failed** — a dedicated direct unit test (`test_unresolvable_downstream_segment_resolves_to_code`) plus the #1171 sibling measurement; added the dedicated test after noticing the first pass only killed it incidentally through the sibling shape |

All mutants killed; tree diffed clean after each restore (`git status --porcelain` / `git diff --stat` both empty against the checkpoint commit before proceeding).

## Sibling issues — measured, not assumed

- **#1171** (backslash line-continuation between a data sink and `\| bash`) — **closes incidentally**. Its own issue body names the exact same root cause ("the same fail-open downstream branch as main#1168"): the continued segment doesn't tokenize (`head is None`), and the new default resolves that to CODE. Confirmed via `hook.check()`: the #1171 shape now BLOCKs, and #1171's own stated false positive (a continued prose opener with no interpreter) stays ALLOW. Pinned in `SiblingIssueMeasurementTests`.
- **#1170** (FIFO relay via `mkfifo`/`&`) — **stays open**, confirmed via `hook.check()` (measured ALLOW after this fix). Structurally different: dataflow through a named pipe across a `&` background-job boundary, never reached by this fix's `\|`-connected pipeline walk. Not folded in, per the spawn brief's explicit scope boundary.

## Related gap found, NOT fixed here (filed separately)

The same relay family also defeats the **other** classifier in this file — main#1167's cross-segment write-then-exec correlation (`_script_invocation_targets`) — via `find -exec bash {} \;` and `echo PATH | xargs bash`. Real-shell-verified to still ALLOW at this PR's head; this fix's default-flip only touches `_opener_feeds_interpreter` (the pipe-follow walk), a different code path with no literal token to correlate against for either shape. Filed as **#1315** (`tech-debt`/`process`/`phase-10`, no wave label) rather than folded in here, matching this org's own precedent (main#1167's PR review filed its sibling gaps separately "because they are different root causes").

## Test plan

- [x] `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/ -q` — 2641 passed, 725 subtests passed (0 regressions)
- [x] `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/ .claude/lib/tests/ -q` — 3896 passed, 803 subtests passed
- [x] `ruff format --check`, `ruff check`, `mypy .claude/hooks/ .claude/lib/ --ignore-missing-imports --no-error-summary` — all clean
- [x] Pre-push hooks (mypy, pytest, pytest-skills, memory-budget, headcount-budget, doc-freshness, office-drift, mermaid-render) — all passed on `git push`
- [x] New file: `.claude/hooks/tests/test_heredoc_relay_downstream.py` (42 tests / 53 subtests) — bypass-blocks, false-positive corpus (with a per-filter coverage-completeness assertion), excluded-filter-blocks, real-shell ground truth, classifier unit tests, sibling-issue measurements

Not merging per instructions — this is ready for review.

## Round 3 update — `rg` dropped (Nino's blocker: `--pre` + attacker-supplied `/dev/stdin`)

Round 2 folded `rg` into `HEREDOC_INERT_RELAY_FILTERS`, measured only on the no-PATH stdin-pipe form of `--pre`/`-z`. That measurement was correct as far as it went, but the PATH `--pre` needs is attacker-supplied: naming the pipe itself (`/dev/stdin`, `/dev/fd/0`) gives `--pre` a PATH even on a pure stdin pipe, and `rg` runs `COMMAND PATH` with that path opened on the child's own stdin — so `sh /dev/stdin` genuinely executes the heredoc body. Real-shell-verified (both bash and zsh):

```
bash | RAN | rg --pre=/bin/sh pat /dev/stdin | GIT-RAN: -c user.name=X -c user.email=Y commit -m z
zsh  | RAN | rg --pre=/bin/sh pat /dev/stdin | GIT-RAN: -c user.name=X -c user.email=Y commit -m z
```

`rg` is now **dropped from the allowlist entirely**, same posture as `sed`/`awk`/`sort` — a plain-form (or single-context) measurement is not sufficient, and gating `--pre` per-invocation would reintroduce the per-command flag grammar this set exists to avoid. Confirmed at the current head: `cat <<'DELIM' | rg --pre=/bin/sh pat /dev/stdin ... DELIM` now **BLOCKs**.

`rg` also exposes a second exec-shaped flag, `--hostname-bin=COMMAND`, which spawns an arbitrary program even on a pure stdin pipe with no PATH at all — but the spawned child gets no arguments and does not inherit the pipe, so it never reaches the heredoc body and is **not** a bypass on its own. It's named in the module comment anyway (and pinned by a dedicated ground-truth test with two independent log files — one confirming the spawn, one confirming the body was never reached) because its existence would otherwise falsify the comment's "every entry measured inert" claim.

**The #1008 contradiction is left explicitly OPEN by this change.** This allowlist still admits `grep` (forbidden org-wide by #1008's "use `rg`, never bare `grep`" rule) while now excluding `rg` (the mandated replacement) — the inverse of what the org's tooling policy wants. Round 2's `rg` addition was reasoned partly from that policy argument; that reasoning is retracted (on #1317, so it isn't re-cited as grounds to re-add `rg` later) because a policy argument was never a safety argument, and here they point in opposite directions. Resolving the contradiction needs either a safe way to admit `rg` (a flag-aware gate, which is the per-command grammar this set is designed to avoid) or a separate decision at the #1008 policy level — not a re-add to this allowlist on the same evidence that just failed.

Test changes this round: `RelayExcludedFiltersBlockTests` gains `rg` rows mirroring `sed`/`awk`/`sort` (not-in-allowlist, plain-form-now-blocks, and a would-be-bypass-if-allowlisted row using `--pre=/bin/sh` with an attacker-supplied `/dev/stdin` path — deliberately **varying** the PATH rather than fixing it at none, the exact gap in the prior test). `RealShellGroundTruthTests` replaces the old context-fixed `--pre`-on-stdin-with-no-PATH test with: plain `rg` (inert, accepted FP), `--pre` with no PATH (still inert, kept as a documented contrast), `--pre=/bin/sh pat /dev/stdin` (genuinely runs — the adversarial row), and a new `--hostname-bin` row (spawns but does not reach the body, using independent spawn/body log files so the two outcomes can't be conflated).

Full suite at this head: `.claude/hooks/tests/` 2652 passed / 734 subtests; `.claude/lib/tests/` 1255 passed / 78 subtests. `ruff format --check`, `ruff check`, and `mypy` clean on both changed files. Pre-push hooks (mypy, pytest, pytest-skills, memory-budget, headcount-budget, doc-freshness, office-drift, mermaid-render) all passed on push.

Not merging per instructions — ready for review.

### DIFF
--- a/.claude/hooks/_shell_parse.py
+++ b/.claude/hooks/_shell_parse.py
@@ -608,14 +608,6 @@ HEREDOC_DATA_SINKS = frozenset({"cat", "tee"})
 #     xxd (incl. `-r`), hexdump (incl. `-e`), join, paste, comm, tac,
 #     shuf (incl. `--random-source`), jq (incl. `env.PATH`, `$ENV.PATH`,
 #     `input_filename`, `@sh` — no `system()` in mainline)
-#     rg — plain AND `--pre COMMAND` / `-z`/`--search-zip` (both apply only
-#          "for each input PATH" per ripgrep's own docs; there is no PATH on
-#          a stdin pipe, so neither ever invokes anything): measured, both
-#          flags are genuine no-ops when the input is a heredoc-fed pipe in
-#          both bash and zsh. Folded in alongside this fix (main#1316) so the
-#          allowlist stops contradicting #1008's org-wide "no bare `grep`"
-#          rule — before this, the list admitted the forbidden tool (`grep`)
-#          while omitting the mandated replacement (`rg`).
 #
 #   RAN (data-driven code-execution surface — DELIBERATELY EXCLUDED even
 #   though they are common "genuinely inert filter" examples in casual
@@ -639,20 +631,46 @@ HEREDOC_DATA_SINKS = frozenset({"cat", "tee"})
 #             plain invocation being inert is not sufficient, exactly as for
 #             `sed`/`awk` below; the exec-shaped flag makes it unsafe as a
 #             blanket allowlist member. Excluded per main#1316.
+#     rg    — `--pre=COMMAND` runs COMMAND once "for each input PATH", and on
+#             a heredoc-fed pipe a PATH is attacker-supplied: naming the pipe
+#             itself (`/dev/stdin`, `/dev/fd/0`) gives `--pre` a PATH even
+#             though the input is a pure stdin pipe, and `rg` runs
+#             `COMMAND PATH` with that path opened on the child's own stdin
+#             — so `sh /dev/stdin` genuinely executes the heredoc body;
+#             measured: `cat <<'EOF' | rg --pre=/bin/sh pat /dev/stdin`
+#             genuinely runs the body in both bash and zsh. `rg`'s ORIGINAL
+#             addition to this allowlist (main#1316's first pass) measured
+#             only the no-PATH stdin-pipe form — fixing the CONTEXT
+#             (whether a PATH is present) instead of varying it, when that
+#             context is exactly what an attacker controls. Separately, `rg`
+#             also exposes `--hostname-bin=COMMAND`, a second exec-shaped
+#             flag that spawns an arbitrary program even on a pure stdin
+#             pipe with no PATH at all — but the spawned child gets no
+#             arguments and does not inherit the pipe, so it never reaches
+#             the heredoc body and is not a bypass on its own. Noted here
+#             anyway because its existence falsifies any claim that every
+#             `rg` flag is harmless on a heredoc-fed stdin pipe. Excluded
+#             per main#1316 (second pass) — previously allowlisted in this
+#             module's own first pass at the same PR, on a plain-form
+#             measurement plus a #1008 policy argument (closing the
+#             contradiction where this allowlist admitted forbidden `grep`
+#             while omitting mandated `rg`); that policy argument is not a
+#             safety argument, and the #1008 contradiction is left OPEN by
+#             this exclusion — see the PR body.
 #
-# All three risks above are ARGUMENT-driven (visible in the segment's own
+# All four risks above are ARGUMENT-driven (visible in the segment's own
 # tokens, not hidden in the heredoc body), so a narrower "allowlisted unless
-# the script argument contains `e`/`system(`/`--compress-program`" rule is
-# possible in principle — but that reintroduces a second, per-command
-# detector inside what this set is meant to keep a flat, reviewable
-# allowlist, and is deliberately left as a non-goal here: excluding
-# `sed`/`awk`/`sort` entirely accepts a real (if comparatively rare, in a
-# documentation-pipeline context) false-positive cost in exchange for never
-# having to get that per-command grammar right. `perl`, `python`, `ruby`,
-# `php`, `xargs`, `parallel`, `find`, `env` and any other general-purpose
-# interpreter or relay are excluded for the same reason, one level more
-# obviously (they are not "filters" in the first place — some of them are
-# exactly the relay family this fix targets).
+# the script argument contains `e`/`system(`/`--compress-program`/`--pre`"
+# rule is possible in principle — but that reintroduces a second,
+# per-command detector inside what this set is meant to keep a flat,
+# reviewable allowlist, and is deliberately left as a non-goal here:
+# excluding `sed`/`awk`/`sort`/`rg` entirely accepts a real (if comparatively
+# rare, in a documentation-pipeline context) false-positive cost in exchange
+# for never having to get that per-command grammar right. `perl`, `python`,
+# `ruby`, `php`, `xargs`, `parallel`, `find`, `env` and any other
+# general-purpose interpreter or relay are excluded for the same reason, one
+# level more obviously (they are not "filters" in the first place — some of
+# them are exactly the relay family this fix targets).
 HEREDOC_INERT_RELAY_FILTERS = frozenset(
     {
         "grep",
@@ -685,7 +703,6 @@ HEREDOC_INERT_RELAY_FILTERS = frozenset(
         "tac",
         "shuf",
         "jq",
-        "rg",
     }
 )
 
--- a/.claude/hooks/_shell_parse.py
+++ b/.claude/hooks/_shell_parse.py
@@ -560,6 +560,106 @@ SHELL_INTERPRETERS = frozenset({"bash", "sh", "zsh", "dash", "ksh"})
 # addition is an explicit, reviewable decision.
 HEREDOC_DATA_SINKS = frozenset({"cat", "tee"})
 
+# ---------------------------------------------------------------------------
+# Downstream-relay default (main#1168)
+# ---------------------------------------------------------------------------
+#
+# `_opener_feeds_interpreter`'s pipe-follow walk (condition 3 below) used to
+# resolve an unrecognised downstream segment head toward DATA ("not proven to
+# be an interpreter, so keep walking; if nothing downstream matches, the body
+# is inert"). That is backwards relative to every other ambiguity in this
+# module, which resolves toward CODE (main#1152's own rule), and it is a
+# measured live bypass: a RELAY command that is not itself a member of
+# `SHELL_INTERPRETERS` but hands its stdin to one defeats the walk entirely —
+#
+#     cat <<'DELIM' | xargs -I{} bash -c "{}"
+#     git -c user.name=X -c user.email=Y commit -m z
+#     DELIM
+#
+# `xargs` is not an interpreter, so the old walk found nothing and returned
+# DATA; the body — a real `git commit` — genuinely runs (real-shell-verified
+# with a marker proxy: a marker command appended to a log file, confirmed to
+# execute). The same family: `parallel`, `env -S`, and (for the OTHER
+# classifier, main#1167's cross-segment write-then-exec correlation —
+# out of scope for the fix here, see that module's comment) `find -exec`.
+#
+# The fix: an unrecognised downstream segment head now resolves to CODE,
+# matching the rest of this module's posture. `HEREDOC_INERT_RELAY_FILTERS`
+# below is the narrow escape hatch — the mirror image of `HEREDOC_DATA_SINKS`
+# — for commands MEASURED to have no code-execution surface reachable from
+# their own stdin content, so a legitimate documentation pipeline
+# (`cat <<'EOF' | grep foo`) does not newly false-block. Same posture as
+# `HEREDOC_DATA_SINKS`: a tiny, explicit, reviewed ALLOWLIST, never a
+# denylist — an unlisted filter fails toward CODE, which is the entire point
+# of inverting the default (an allowlist's safety property IS its default).
+#
+# Every entry here was verified with a real-shell marker probe (a command
+# appended to a log file inside the heredoc body, checked for the log
+# actually being written) BEFORE being added, not reasoned from a man page:
+#
+#   DID-NOT-RUN (genuinely inert, confirmed safe to allowlist):
+#     grep, wc, sort, uniq, head, tail, cut, tr, column, nl, rev, fold,
+#     expand, unexpand, base64, md5sum, sha1sum, sha256sum, od, xxd, join,
+#     paste, comm, tac, shuf, jq
+#
+#   RAN (data-driven code-execution surface — DELIBERATELY EXCLUDED even
+#   though they are common "genuinely inert filter" examples in casual
+#   reasoning about this shape):
+#     sed  — the GNU `e` flag/command executes the (input-derived) pattern
+#            space as a shell command and substitutes its output; measured:
+#            `cat <<'EOF' | sed 's/.*/&/e'` genuinely runs the body.
+#     awk  — `system()` (and piped `print ... | "cmd"` / `getline < "cmd"`)
+#            executes a command built from field/record data; measured:
+#            `cat <<'EOF' | awk '{system($0)}'` genuinely runs the body.
+#
+# Both risks are ARGUMENT-driven (visible in the segment's own tokens, not
+# hidden in the heredoc body), so a narrower "allowlisted unless the script
+# argument contains `e`/`system(`" rule is possible in principle — but that
+# reintroduces a second, per-command detector inside what this set is meant
+# to keep a flat, reviewable allowlist, and is deliberately left as a
+# non-goal here: excluding `sed`/`awk` entirely accepts a real (if
+# comparatively rare, in a documentation-pipeline context) false-positive
+# cost in exchange for never having to get that per-command grammar right.
+# `perl`, `python`, `ruby`, `php`, `xargs`, `parallel`, `find`, `env` and any
+# other general-purpose interpreter or relay are excluded for the same
+# reason, one level more obviously (they are not "filters" in the first
+# place — some of them are exactly the relay family this fix targets).
+HEREDOC_INERT_RELAY_FILTERS = frozenset(
+    {
+        "grep",
+        "egrep",
+        "fgrep",
+        "wc",
+        "sort",
+        "uniq",
+        "head",
+        "tail",
+        "cut",
+        "tr",
+        "column",
+        "nl",
+        "rev",
+        "fold",
+        "expand",
+        "unexpand",
+        "base64",
+        "md5sum",
+        "sha1sum",
+        "sha256sum",
+        "sha512sum",
+        "cksum",
+        "od",
+        "xxd",
+        "hexdump",
+        "join",
+        "paste",
+        "comm",
+        "tac",
+        "shuf",
+        "jq",
+    }
+)
+
 # A heredoc opener: `<<EOF`, `<<'EOF'`, `<<"EOF"`, `<<-EOF`. Matched with
 # `.match(line, i)` from a scanner that has already excluded `<<<` (here-string)
 # and quoted regions, so no lookbehind is needed here.
@@ -725,7 +825,8 @@ def _opener_feeds_interpreter(line: str, pos: int, scan: _LineScan) -> bool:
       1. the segment's head command is a known inert sink (`cat`, `tee`), AND
       2. that segment carries no process substitution, AND
       3. no segment downstream of it *in the same pipeline* is a shell
-         interpreter or itself carries a process substitution.
+         interpreter, a RELAY that hands its stdin to one (main#1168), or
+         itself carries a process substitution.
 
     Condition 2 is the one that is easy to miss, and missing it is a live
     bypass (main#1155 review, M1):
@@ -741,6 +842,19 @@ def _opener_feeds_interpreter(line: str, pos: int, scan: _LineScan) -> bool:
     anywhere in a reachable segment therefore forces CODE without trying to
     resolve what is inside it: an unresolvable sink resolves toward CODE, which
     is the rule the rest of this classifier already follows.
+
+    Condition 3's downstream default is the OTHER historically-inverted case
+    (main#1168): a downstream segment head that is not a known
+    `SHELL_INTERPRETERS` member, a known `HEREDOC_DATA_SINKS` member (`cat`,
+    `tee` — relaying to a FILE, not an interpreter, is exactly as inert
+    downstream as it is in the owning-segment position), or a known-inert
+    `HEREDOC_INERT_RELAY_FILTERS` member now resolves to CODE — not "keep
+    walking, default to DATA" as before. A RELAY command (`xargs -I{} bash -c
+    "{}"` and the same family via `parallel`/`env -S`) is not itself an
+    interpreter, so the old walk found nothing downstream and returned DATA
+    even though the body genuinely runs (real-shell-verified with a marker
+    proxy). See the module comment above `HEREDOC_INERT_RELAY_FILTERS` for
+    the full false-positive corpus that justifies the allowlist's contents.
 
     `&&`, `||`, `;` and `&` break the pipeline: what follows them does not
     receive this stdin. A `&` that is part of a redirect (`2>&1`) is not a
@@ -761,7 +875,16 @@ def _opener_feeds_interpreter(line: str, pos: int, scan: _LineScan) -> bool:
         head = _segment_head_command(line[start:end])
         if head_must_be_inert_sink:
             return head not in HEREDOC_DATA_SINKS
-        return head in SHELL_INTERPRETERS
+        if head in SHELL_INTERPRETERS:
+            return True
+        # main#1168: an unresolved/unknown relay (including `head is None` —
+        # the segment didn't tokenize) resolves to CODE. A downstream
+        # `HEREDOC_DATA_SINKS` member (`cat`, `tee` — e.g. `| tee /tmp/a.md`
+        # relaying to a FILE, not an interpreter) is just as inert here as it
+        # is in the owning-segment position, so it joins
+        # `HEREDOC_INERT_RELAY_FILTERS` as a reason to keep walking rather
+        # than stop at CODE.
+        return head not in HEREDOC_DATA_SINKS and head not in HEREDOC_INERT_RELAY_FILTERS
 
     if _reachable_sink_is_code(idx, head_must_be_inert_sink=True):
         return True
--- a/.claude/hooks/_shell_parse.py
+++ b/.claude/hooks/_shell_parse.py
@@ -593,44 +593,72 @@ HEREDOC_DATA_SINKS = frozenset({"cat", "tee"})
 # denylist — an unlisted filter fails toward CODE, which is the entire point
 # of inverting the default (an allowlist's safety property IS its default).
 #
-# Every entry here was verified with a real-shell marker probe (a command
-# appended to a log file inside the heredoc body, checked for the log
-# actually being written) BEFORE being added, not reasoned from a man page:
+# Every entry here was verified with a real-shell marker probe — BOTH the
+# plain invocation AND every exec-shaped flag the command exposes (a command
+# appended to a log file inside the heredoc body / a marker script standing
+# in for an external-program flag, checked for the log actually being
+# written, in both bash and zsh) BEFORE being added, not reasoned from a man
+# page and not measured on the bare invocation alone:
 #
-#   DID-NOT-RUN (genuinely inert, confirmed safe to allowlist):
-#     grep, wc, sort, uniq, head, tail, cut, tr, column, nl, rev, fold,
-#     expand, unexpand, base64, md5sum, sha1sum, sha256sum, od, xxd, join,
-#     paste, comm, tac, shuf, jq
+#   DID-NOT-RUN (genuinely inert — plain form AND every exec-shaped flag
+#   probed — confirmed safe to allowlist):
+#     grep, egrep, fgrep (incl. `-f -`), wc, uniq, head, tail, cut, tr,
+#     column, nl, rev, fold, expand, unexpand, base64 (incl. `-d`),
+#     md5sum (incl. `-c`), sha1sum, sha256sum, sha512sum, cksum, od,
+#     xxd (incl. `-r`), hexdump (incl. `-e`), join, paste, comm, tac,
+#     shuf (incl. `--random-source`), jq (incl. `env.PATH`, `$ENV.PATH`,
+#     `input_filename`, `@sh` — no `system()` in mainline)
+#     rg — plain AND `--pre COMMAND` / `-z`/`--search-zip` (both apply only
+#          "for each input PATH" per ripgrep's own docs; there is no PATH on
+#          a stdin pipe, so neither ever invokes anything): measured, both
+#          flags are genuine no-ops when the input is a heredoc-fed pipe in
+#          both bash and zsh. Folded in alongside this fix (main#1316) so the
+#          allowlist stops contradicting #1008's org-wide "no bare `grep`"
+#          rule — before this, the list admitted the forbidden tool (`grep`)
+#          while omitting the mandated replacement (`rg`).
 #
 #   RAN (data-driven code-execution surface — DELIBERATELY EXCLUDED even
 #   though they are common "genuinely inert filter" examples in casual
 #   reasoning about this shape):
-#     sed  — the GNU `e` flag/command executes the (input-derived) pattern
-#            space as a shell command and substitutes its output; measured:
-#            `cat <<'EOF' | sed 's/.*/&/e'` genuinely runs the body.
-#     awk  — `system()` (and piped `print ... | "cmd"` / `getline < "cmd"`)
-#            executes a command built from field/record data; measured:
-#            `cat <<'EOF' | awk '{system($0)}'` genuinely runs the body.
+#     sed   — the GNU `e` flag/command executes the (input-derived) pattern
+#             space as a shell command and substitutes its output; measured:
+#             `cat <<'EOF' | sed 's/.*/&/e'` genuinely runs the body.
+#     awk   — `system()` (and piped `print ... | "cmd"` / `getline < "cmd"`)
+#             executes a command built from field/record data; measured:
+#             `cat <<'EOF' | awk '{system($0)}'` genuinely runs the body.
+#     sort  — `--compress-program=CMD` runs CMD with the heredoc's own data
+#             on its stdin once the sort spills to a temp file (`-S` sets
+#             the spill threshold; the padding needed to cross it is
+#             attacker-controlled, same as any other input-size trigger);
+#             measured: `cat <<'EOF' | sort -S 1 --compress-program=CMD`
+#             genuinely runs CMD in both bash and zsh once enough body lines
+#             are present to force a spill. The PLAIN invocation (no
+#             `--compress-program`) is genuinely inert — this is why `sort`
+#             shipped on this allowlist in the first place (main#1168's
+#             original measurement covered only the bare form) — but a
+#             plain invocation being inert is not sufficient, exactly as for
+#             `sed`/`awk` below; the exec-shaped flag makes it unsafe as a
+#             blanket allowlist member. Excluded per main#1316.
 #
-# Both risks are ARGUMENT-driven (visible in the segment's own tokens, not
-# hidden in the heredoc body), so a narrower "allowlisted unless the script
-# argument contains `e`/`system(`" rule is possible in principle — but that
-# reintroduces a second, per-command detector inside what this set is meant
-# to keep a flat, reviewable allowlist, and is deliberately left as a
-# non-goal here: excluding `sed`/`awk` entirely accepts a real (if
-# comparatively rare, in a documentation-pipeline context) false-positive
-# cost in exchange for never having to get that per-command grammar right.
-# `perl`, `python`, `ruby`, `php`, `xargs`, `parallel`, `find`, `env` and any
-# other general-purpose interpreter or relay are excluded for the same
-# reason, one level more obviously (they are not "filters" in the first
-# place — some of them are exactly the relay family this fix targets).
+# All three risks above are ARGUMENT-driven (visible in the segment's own
+# tokens, not hidden in the heredoc body), so a narrower "allowlisted unless
+# the script argument contains `e`/`system(`/`--compress-program`" rule is
+# possible in principle — but that reintroduces a second, per-command
+# detector inside what this set is meant to keep a flat, reviewable
+# allowlist, and is deliberately left as a non-goal here: excluding
+# `sed`/`awk`/`sort` entirely accepts a real (if comparatively rare, in a
+# documentation-pipeline context) false-positive cost in exchange for never
+# having to get that per-command grammar right. `perl`, `python`, `ruby`,
+# `php`, `xargs`, `parallel`, `find`, `env` and any other general-purpose
+# interpreter or relay are excluded for the same reason, one level more
+# obviously (they are not "filters" in the first place — some of them are
+# exactly the relay family this fix targets).
 HEREDOC_INERT_RELAY_FILTERS = frozenset(
     {
         "grep",
         "egrep",
         "fgrep",
         "wc",
-        "sort",
         "uniq",
         "head",
         "tail",
@@ -657,6 +685,7 @@ HEREDOC_INERT_RELAY_FILTERS = frozenset(
         "tac",
         "shuf",
         "jq",
+        "rg",
     }
 )
 
--- a/.claude/hooks/tests/test_heredoc_relay_downstream.py
+++ b/.claude/hooks/tests/test_heredoc_relay_downstream.py
@@ -25,6 +25,20 @@ surface reachable from their own stdin — the false-positive escape hatch so
 `cat <<'EOF' | grep foo` (a genuinely inert documentation pipeline) does not
 newly false-block.
 
+main#1316 (PR #1316 rework, merge-gate finding): `sort` shipped on this
+allowlist measured only on its PLAIN invocation; `sort --compress-program=CMD`
+genuinely runs CMD with the heredoc's own data on its stdin once the sort
+spills to a temp file (attacker-controlled padding crosses the `-S` spill
+threshold). Dropped from the allowlist for the same reason `sed`/`awk` are
+excluded — a plain invocation being inert does not make the command safe to
+allowlist wholesale, and a per-flag detector for `--compress-program` would
+reintroduce exactly the per-command grammar this set is designed to avoid.
+Folded in alongside it: `rg`, which was previously absent from the allowlist
+while `grep` (forbidden org-wide by main#1008) was present — the inverse of
+what this org mandates. `rg`'s exec-shaped flags (`--pre`, `-z`) are
+documented to apply only "for each input PATH"; measured genuinely inert on a
+heredoc-fed stdin pipe in both bash and zsh, plain and exec-shaped.
+
 Test organisation
 ==================
 
@@ -32,10 +46,12 @@ Test organisation
     real `check()`.
   * `RelayFalsePositiveCorpusTests` — every `HEREDOC_INERT_RELAY_FILTERS`
     member, individually and chained, must stay ALLOW.
-  * `RelayExcludedFiltersBlockTests` — `sed`/`awk` are DELIBERATELY excluded
-    from the allowlist (both have a data-driven code-execution surface); this
-    pins that they now resolve to CODE, and that the adversarial shapes which
-    justify the exclusion are real bypasses if they were ever allowlisted.
+  * `RelayExcludedFiltersBlockTests` — `sed`/`awk`/`sort` are DELIBERATELY
+    excluded from the allowlist (each has a data-driven code-execution
+    surface reachable through an exec-shaped flag — `sed`'s `e` flag, `awk`'s
+    `system()`, `sort`'s `--compress-program`); this pins that they now
+    resolve to CODE, and that the adversarial shapes which justify the
+    exclusion are real bypasses if they were ever allowlisted.
   * `RealShellGroundTruthTests` — marker-proxy verification against an actual
     `bash`/`zsh`, per shape, so a verdict is checked against what the shell
     really does rather than against expectations.
@@ -155,7 +171,6 @@ class RelayFalsePositiveCorpusTests(unittest.TestCase):
         "egrep": "egrep foo",
         "fgrep": "fgrep foo",
         "wc": "wc -l",
-        "sort": "sort -u",
         "uniq": "uniq -c",
         "head": "head -n 5",
         "tail": "tail -n 5",
@@ -182,6 +197,7 @@ class RelayFalsePositiveCorpusTests(unittest.TestCase):
         "tac": "tac",
         "shuf": "shuf",
         "jq": "jq -R .",
+        "rg": "rg foo",
     }
 
     def test_every_allowlisted_filter_has_a_representative_case(self):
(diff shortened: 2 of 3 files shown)


parametrization:
**Review: LGTM, with a scoped tech-debt follow-up**

Reviewed at head `bf368980aacbb360ad9f1573ccd225b3aa3d8edf`. All 14 CI checks green at that SHA. Verified independently in a separate scratch clone (not the parent checkout) — never touched the main worktree.

**Marker-proxy corpus (real bash + zsh, not reasoned from documentation)** — built against the deploy/CI shapes I was asked to check, then ran the same commands through the shipped `validate_commit_identity.check()`:

| Downstream shape | Real shell runs body? | Hook decision | Note |
|---|---|---|---|
| `tee -a` sink | No | ALLOW | data sink, unaffected |
| `tee` then plain `sed` | No (sed plain is inert) | BLOCK | accepted FP, matches the PR's own documented sed/awk exclusion policy |
| `grep` then `tee` (2-hop) | No | ALLOW | multi-hop allowlisted chain resolves correctly |
| `grep` then `tee` then `tee` (3-hop) | No | ALLOW | confirms the M2 regression is actually fixed (below) |
| `envsubst` | No | BLOCK (new FP) | pure text substitution, no shell-out surface found |
| `python3 -c '...stdin.read()...'` (non-exec) | No | BLOCK | expected — deliberate general-interpreter exclusion, same as perl/ruby |
| `docker exec -i CONTAINER sh` | Yes (proxied via a bare `sh` relay) | BLOCK | correct catch — genuine relay-to-shell, same family the fix targets |
| `kubectl apply -f -` / `psql` / `sqlite3` | not installed in sandbox, untested | BLOCK | plausible FP for kubectl; psql/sqlite3 arguably belong in the same "has its own shell-escape" bucket as sed/awk (backslash-bang, COPY TO PROGRAM, dot-shell) |
| `gh pr create --title x --body-file -` | No (only blocks if body text ALSO matches an identity pattern) | BLOCK only when body contains git-commit-shaped text | narrow, not a broad outage |
| absolute-path `grep`, `env`-wrapped `grep` | No | ALLOW | confirms the allowlist matches the RESOLVED command (post env/path-prefix stripping, basename) — not the literal invocation string |

Repo-wide `rg` sweep (all 7 repos, gitignored child clones included, `.sh`/`.yml`/`.yaml`/`.py`) for an existing heredoc-piped-to-an-unlisted-tool shape found zero matches — nothing currently shipped breaks. The exposure is prospective, and even then only fires when the body ALSO contains identity-pattern text, not on any heredoc piped to an unlisted tool.

**M2 regression, reproduced independently**: reverted the `HEREDOC_DATA_SINKS` downstream check (kept only the `HEREDOC_INERT_RELAY_FILTERS` half) in the scratch clone and reran the suite — 6 failed exactly as reported, 2 of them in the pre-existing `test_heredoc_target_scope.py` (`test_cat_heredoc_piped_only_to_tee_is_still_data`, `test_pipe_ampersand_to_a_data_sink_stays_data`). Restored cleanly (`git status` clean after).

**#1171 incidental close, verified with a marker proxy rather than accepted on report**: ran the backslash-continuation relay shape under real bash — the marker DID execute, confirming it's a genuine pre-existing bypass, not a hypothetical — and `classify_heredocs` on this PR's code correctly marks it CODE. Also checked #1170's FIFO relay stays open (not falsely swept in): the marker DID execute in real bash (genuine bypass exists), and the classifier still returns DATA for it — matches the PR's stated non-goal, no over-claim.

**Verdict**: the default-flip and its rationale are sound, and every claim in the PR body that I could independently re-derive (M2, #1171, #1170, the FP corpus) checked out against real-shell ground truth, not just the author's say-so. The allowlist-completeness gaps I found (`yq` missing next to its `jq` analog, envsubst/kubectl/psql/sqlite3 untested) are real but narrow — nothing in this org's shipped scripts hits them today, and the block only fires on the coincidence of an unlisted tool AND identity-pattern text in the same body — so per the fold-in-or-defer rule this is deferred work, not a blocker. Filed as #1317 with the corpus table and a concrete ask (measure yq/envsubst/kubectl/psql/sqlite3 against real shells, allowlist or document the exclusion the way sed/awk already are).

---

Requestor: Lucas Ferreira
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: #1317

parametrization:
**Review: changes requested — one blocker, everything else verified clean**

Merge-gate review at head SHA `bf368980aacbb360ad9f1573ccd225b3aa3d8edf`, diffed three-dot against `main` via the contents API at that SHA. CI is green: 14/14 check-runs `success` at that exact SHA (the combined-status endpoint reads `pending` only because `total_count` is 0 — there are no legacy commit statuses, just check-runs).

The design call in this PR is right and the reasoning behind it is the best-argued thing I have reviewed this wave. Inverting the downstream default from DATA to CODE is correct, the `sed`/`awk` exclusion rationale is correct, and I reproduced both exclusion controls independently in bash and zsh — a plain substitution stays inert, the `e` flag and `system()` genuinely execute. I am blocking on one member of the allowlist, not on the approach.

---

## Blocker — `sort` is allowlisted and executes the heredoc body

`sort --compress-program=CMD` runs `CMD` whenever the sort spills to a temp file, with the data on `CMD`'s stdin. Point it at a shell and the body runs.

Marker proxy (fake `git` earlier on `PATH` that appends to a log; the log is written only if a shell really runs the body), downstream segment `sort -S1 --compress-program=/bin/sh`, body of one commit line plus 20 padding lines:

```
REAL SHELL bash : RAN   marker=GIT-RAN: -c user.name=X -c user.email=Y commit -m z
REAL SHELL zsh  : RAN   marker=GIT-RAN: -c user.name=X -c user.email=Y commit -m z
HOOK MAIN            : ALLOW   is_code=[False]
HOOK PR-HEAD bf36898 : ALLOW   is_code=[False]
  body stripped from scan: True
```

Head resolves to `sort`, `sort` is on the list, the walk calls it inert, the span is DATA, and `strip_data_heredocs` removes the body before `validate_commit_identity` ever sees it. Spill threshold at `-S 1` is about 9 body lines and the padding is attacker-controlled, so this is not a large-input-only curiosity:

```
padding lines = 5 -> did-not-run
padding lines = 8 -> RAN
```

**This is the exact class you excluded `sed` and `awk` for.** Your stated standard is that a plain invocation being inert is not sufficient, because the surface is argument-driven. `sort --compress-program` is that same shape, and `RealShellGroundTruthTests::test_sort_relay_genuinely_inert` measures only the bare `sort` — the very insufficiency you identified. The standard is right; it was applied to `sed`/`awk` and not to the rest of the list.

To be fair to the change: this is **not a regression**. `main` allows it too, because `main` resolves every unrecognised downstream head to DATA. The PR strictly improves net security. I am blocking because the allowlist is new surface this PR introduces, it is the whole safety property, and the module comment and test suite both assert in writing that every member was measured to have no code-execution surface reachable from its own stdin — which is false for one member, and will be cited by the next reader.

Fix is small: drop `sort` from `HEREDOC_INERT_RELAY_FILTERS` and its corpus row and ground-truth row, and add a permanent adversarial regression test mirroring `test_sed_e_flag_would_be_a_real_bypass_if_allowlisted`. Note that removing it also needs the corpus table updated — I confirmed the completeness assertion fails otherwise (my mutant N2 red 5 tests).

## Allowlist attack — the rest of the list is clean

Every member, plain form and the most exec-shaped invocation I could construct, real bash and zsh, marker proxy:

- **RAN:** `sort -S1 --compress-program=/bin/sh`, `sort -S1 --compress-program=bash`
- **did-not-run:** `grep -f -`, `sort --files0-from=-`, `shuf --random-source=...`, `hexdump -e '"%s"'`, `xxd -r`, `base64 -d`, `md5sum -c`, `head -c 1000000`, `tr -d`, plus every plain form of all 31 members

`jq` was the one you flagged for me to test first and it is clean — probed for `env.PATH`, `$ENV.PATH`, `input_filename` and `@sh`; mainline jq has no `system()`. `column`, `tr`, `xxd -r` and `base64 -d` are all clean too. `sort` is the only one.

**Name resolution** behaves as intended: `/usr/bin/grep`, `env grep`, `GREP_COLOR=x grep` and `timeout 5 grep` all resolve to `grep` and stay DATA; an untokenizable segment resolves to CODE. The match is on the resolved *name*, not the resolved *binary* — a shell function `grep() { bash; }` or a PATH-shadowed `./grep` both execute while the hook reads `grep` and allows. That is pre-existing, applies identically to `HEREDOC_DATA_SINKS`, and is not new surface here; I recorded it on #1317 rather than asking for it.

## False-positive corpus

I extended rather than re-ran, and the headline is that the cost is **narrower than it looks**: an unlisted filter only blocks when the body *also* carries identity-pattern text. A benign body piped to `rg`/`less`/`git apply`/`yq`/`fmt` stays ALLOW. The realistic in-repo trigger is a heredoc that documents the canonical commit-identity form — which this org writes constantly — piped to something unlisted.

One entry is worth folding in now: **`rg` is missing while `grep` is present**, and this org mandates the reverse (`CLAUDE.md` requires `rg`; a bare `grep` is hard-blocked by #1008). As shipped the list admits the forbidden tool and blocks the required one. `rg` measures inert in both shells and has no `-e`/`--exec`-shaped surface. Added to #1317 with the measurement; cheap to include in this round while you are touching the set.

Also measured CODE (all confirmed inert in a real shell, all new false positives): `less`, `git apply`, `yq`, `fmt`, `numfmt`, `strings`, `b2sum`, `iconv`, `diff -`, `sponge`, `gzip`. Those belong to #1317's follow-up pass, not here.

## Your reported findings, verified independently

I rebuilt the classifier at your head SHA in an isolated tree and ran the mutants myself; verdicts are from pytest's exit code.

- **M2 confirmed.** Dropping `HEREDOC_DATA_SINKS` from the downstream check reds 6, including the two pre-existing tests in `test_heredoc_target_scope.py`. The final code does treat a downstream data sink as inert — `head not in HEREDOC_DATA_SINKS and head not in HEREDOC_INERT_RELAY_FILTERS`. Good catch by the existing suite, and good that you reported it.
- **M6 genuinely pins.** The two reds under the fail-open mutant are `test_unresolvable_downstream_segment_resolves_to_code` and the #1171 measurement — so the dedicated test really is one of them. I checked the mechanism rather than the verdict: `_segment_head_command('"unterminated')` returns `None` because `tokenize` fails, the command yields exactly one span so the assertion is not vacuous, and a *balanced* `unterminated` head would still be CODE by a different route — so the test cannot pass for a neighbouring reason. It also holds when I vary the incidental spelling (unclosed single quote, trailing backslash, empty segment).
- All 6 of your mutants plus 4 of my own were killed. Baseline reproduces exactly: 42 tests / 53 subtests, and the suite delta is precisely +42 / +53.
- `ruff format --check`, `ruff check`, `mypy` clean at head.

## Siblings

**#1171 closes incidentally — verified, with one caveat worth carrying into the close.** Ground truth first: the shape genuinely runs the body in both shells, and #1171's own stated false positive genuinely does not. Hook verdicts: ALLOW on `main`, BLOCK at this head, with the false positive still ALLOW. Criteria 1 and 3 are met.

Criterion 2 — "body offsets stay correct" — is **not** met in the sense #1171 meant it. The physical/logical line folding it asks for is still undone, so the body-start offset is still wrong under a continuation. What changed is that it now fails *safe*: any true line continuation on the opener line ends the segment in a dangling backslash, which does not tokenize, so it resolves to CODE. I probed for a residual hole there and found the opposite of one — a shape that `main` gets wrong in the dangerous direction (continuation, then `;`, then a real commit; the shell runs it, `main` strips it and allows) now blocks at your head. So it is an extra incidental fix, not a gap. Worth one line on #1171 when closing so the next reader knows the offset defect persists and is merely no longer exploitable.

**#1170 is genuinely untouched, not partially affected.** ALLOW at `main` and ALLOW at head, and the FIFO shape really does run the body in a real shell — so it remains a live gap, correctly left to its own PR. Your test pinning today's ALLOW is the right call and the comment explaining it should be deleted rather than inverted when #1170 lands is exactly right.

## #1315 — my ruling is that it stays filed, not folded in

Measured, not assumed: both shapes ALLOW at `main` and at your head, with `bash /tmp/s.txt` blocking as the control, so it is genuinely a different code path and genuinely untouched.

Running the fold-in test honestly, it is closer than the other deferrals: a working script really could contain `find . -name '*.sh' -exec bash {} \;`, and it fails open. But the deciding factor is the third leg plus the shape of the fix — it needs a new mechanism in `_script_invocation_targets`, not a parameter change to the walk this PR touches, because there is no literal path token to correlate against. Folding a second classifier's redesign into this PR would also make the allowlist harder to review as one unit, which is precisely the property that let me find the `sort` hole. Keep it on #1315.

## Filed

- **#1318** — the durable version of the blocker: the allowlist has no gate requiring an adversarial-invocation measurement per member, only one requiring corpus *coverage*. That gap is what admitted `sort` and would admit the next one. Includes the full audit and asks for a structural assertion, a permanent regression test, and the same sweep applied to `HEREDOC_DATA_SINKS`. On the board, under the #1150 umbrella.
- **#1317** — extended with the `rg` measurement, the narrow-FP-surface mechanism, and the name-vs-binary limitation. Did not duplicate it.

Re-request when `sort` is off the list and pinned. Everything else here is ready.

---

Requestor: Nino Kavtaradze
Requestee: Weronika Zielinska
TechDebt: #1318

parametrization:
**Review: LGTM — re-reviewed at the new head, sort/rg change verified**

Re-reviewed at `8507e4a98a9309ee6b08d15aa08c2d04cb2d13a3` (my prior Approved at `bf36898` is superseded by this comment). All 14 CI checks green at this SHA, including Pytest (was in-progress when I started, now SUCCESS). Verified again in a fresh scratch clone at this exact head, not the parent checkout.

**The `sort --compress-program` bypass and the `rg` fold-in, reproduced independently:**

- `sort -S 1K --compress-program=MARKER` genuinely runs MARKER on the heredoc's own data once the sort spills to a temp file — confirmed in real bash (marker script executed repeatedly as sort processed chunks). Plain `sort` with no such flag stays genuinely inert in both shells, which is exactly why it originally shipped on the allowlist and why the right fix is exclusion, not a flag-gate — same posture as sed/awk.
- `rg --pre COMMAND` and `rg -z` are genuine no-ops on a heredoc-fed stdin pipe in both bash and zsh (confirmed) — ripgrep's own docs are right that `--pre`/`-z` only apply per input PATH, and a pipe has none. Correct fold-in, and it resolves the #1008 contradiction (previously: allowlist admitted the org-forbidden `grep`, omitted the org-mandated `rg`).

**My corpus, re-run against the new allowlist:** my original `sort | uniq -c | head` case now correctly BLOCKS (it was one of my explicit "confirm a chain resolves" test cases at the prior head, and the team's finding is right that this is a materially different risk profile than the tools I swept — `sort -u` / `sort | uniq` is a far more ordinary shape to pipe a heredoc into than `envsubst`/`kubectl`/`psql`). **Re-ran my repo-wide sweep specifically for `sort`** (all 7 repos, gitignored child clones included, same same-line-heredoc-then-pipe pattern) — still **zero** matches for a heredoc piped into `sort` anywhere in this org's shipped `.sh`/`.yml`/`.yaml`/`.py`. The exposure remains prospective, not a live regression, even for the more-ordinary tool.

Full relevant suite in the new scratch clone: 186 passed, 402 subtests passed (`test_heredoc_relay_downstream.py` + `test_heredoc_target_scope.py` + `test_interpreter_option_grammar.py`).

**#1317 updated** with two follow-on notes now that `sort` moved into the exclude-entirely bucket (precedent, not new work for this PR): `yq` reads more clearly as an oversight given `rg`'s fold-in used the same "org-relevant tool, measures inert" argument; and `psql`/`sqlite3` — which have their own argument/data-driven shell-escapes (backslash-bang, COPY TO PROGRAM, dot-shell) — should probably be framed as belonging in the sed/awk/sort exclude bucket rather than left as open allowlist candidates. Filed as commentary on the existing issue, not a new one.

---

Requestor: Lucas Ferreira
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: #1317

parametrization:
**Re-review at `8507e4a98a9309ee6b08d15aa08c2d04cb2d13a3` — the `sort` blocker is properly fixed. Blocking again on `rg`, which I asked for and which carries the same defect.**

CI 14/14 `success` at this exact SHA (the earlier `UNSTABLE` was the pytest run still in flight; it has since gone green).

## First, the part that is mine to own

I recommended folding `rg` in. I measured it as `[heredoc] | rg foo` — the plain form — and I argued for it partly on the grounds that this org mandates `rg` and forbids bare `grep`. That was the exact error I had just blocked `sort` for: a plain-invocation measurement treated as sufficient, dressed up with an argument that is not a safety argument at all. Whether the org mandates a tool has no bearing on whether it is inert; I should not have mixed those. The `rg` finding below is the consequence of my own recommendation, not of anything you got wrong in implementing it.

## `sort` — fixed, and fixed the right way

Verified independently at this SHA: `sort` is out of the frozenset, `cat [heredoc] | sort -S1 --compress-program=/bin/sh` now BLOCKs, and plain `sort` blocks as the documented accepted false positive. Dropping rather than flag-gating is the correct call for the reason you gave — abbreviated and positional forms of `--compress-program` would need the per-command grammar this allowlist exists to avoid. `test_sort_compress_program_would_be_a_real_bypass_if_allowlisted` and `test_sort_compress_program_genuinely_runs` both pin it, and re-adding `sort` reds 6 tests in my own mutation run, so it cannot be reverted quietly. The `_write_marker_passthrough_script` helper is the right abstraction — a flag naming an external program needs a real script, not an inline fragment, and it generalises.

The two swapped fixtures still do their jobs: `cut -d, -f1` keeps `test_chained_allowlisted_filters_stay_allowed` a genuine three-hop allowlisted chain and keeps `test_multi_hop_pipeline_stops_walk_correctly_at_end_of_segments` testing end-of-segment termination. I confirmed they retain teeth rather than merely passing — removing `cut` reds 4 and removing `uniq` reds 4.

Mutation battery re-run at this SHA, verdicts from exit code: M1 24, M2 6, M3 6, M4 47, M5 8, M6 2, plus four of mine. All killed. Baseline 149 passed / 66 subtests across both files.

## Blocker — `rg --pre` is live, because the PATH is attacker-supplied

Your premise is correct as far as it goes, and I confirmed it: on a heredoc-fed pipe with no path argument, neither flag fires.

```
did-not-run  | rg --pre no PATH
did-not-run  | rg -z/--search-zip
```

The gap is that nothing stops the attacker from supplying a path — and `/dev/stdin` is a path that names the very pipe carrying the body:

```
bash | RAN | rg --pre=/bin/sh  pat /dev/stdin | GIT-RAN: -c user.name=X -c user.email=Y commit -m z
zsh  | RAN | rg --pre=/bin/sh  pat /dev/stdin | GIT-RAN: -c user.name=X -c user.email=Y commit -m z
bash | RAN | rg --pre=/bin/sh  pat /dev/fd/0  | GIT-RAN: ...
zsh  | RAN | rg --pre=/bin/sh  pat /dev/fd/0  | GIT-RAN: ...
```

`rg` runs `COMMAND PATH` with the file also on the child's stdin, so `sh /dev/stdin` executes the body. Hook verdict at this SHA: **ALLOW**. Same class as `sort --compress-program`, reached through a different premise.

`test_rg_pre_flag_genuinely_inert_on_stdin` is a real adversarial row and still cannot catch this, because it only ever exercises the no-PATH form. That is the structural point: the premise it validates is true, and the member is still unsafe.

There is also a second exec-shaped flag on `rg` that the rework does not mention: **`--hostname-bin=COMMAND`**, which executes an arbitrary named program and fires on a pure stdin pipe with no path and no hyperlink format at all.

```
RAN | hostname-bin, PURE STDIN (no path) | marker=SPAWNED args=[]
```

It is **not** a bypass by itself — the child gets no arguments and does not inherit the pipe (`stdin=` empty in every probe I ran), so the body never reaches it. I am flagging it because it means the module comment's claim is still not true as written.

## The comment-against-code read

Checked, as asked. The comment now claims every entry was measured on "the plain invocation AND every exec-shaped flag the command exposes". For the 30 pre-existing members that is now accurate and well-evidenced. For `rg` it is not: `--pre` was measured only in the form that cannot fire, and `--hostname-bin` was not measured or mentioned at all. Since "the comment's claim is true as written" was the thing that made this a blocker rather than debt last round, the same criterion applies here.

## Recommended resolution — drop `rg`, don't gate it

Same reasoning you applied to `sort`, and I think it lands the same way. Gating `--pre`/`--hostname-bin` per-invocation is the per-command flag grammar that is an explicit non-goal.

Dropping `rg` restores the over-block on `cat [heredoc] | rg foo`, which is a false positive in the safe direction and belongs on #1317 as an accepted FP cost. Please do not treat my #1008 argument as a reason to keep it — I have retracted that on #1317. The allowlist's admission criterion is inertness, not org tooling policy, and those came apart here.

One standing rule worth writing into the comment while you are there, because it is the general lesson and not a fact about `rg`: **a member whose safety rests on a contextual premise ("there is no PATH on a stdin pipe") does not meet the same bar as one with no exec surface at all.** Contextual premises are defeatable by construction. The other 30 members are inert unconditionally; `rg` was the only one that ever needed a qualifier, and the qualifier is where it broke.

## A detector that catches this class mechanically

Scanning each member's own `--help` for a flag whose value names a program is cheap. Validated against known positives before I trusted it:

```
== CONTROLS ==
sort  --compress-program=PROG      awk  --file=progfile      sed  <none found>

== ALLOWLIST AT 8507e4a ==
grep egrep fgrep wc uniq head tail cut tr column nl rev fold expand unexpand
base64 md5sum sha1sum sha256sum sha512sum cksum od xxd hexdump join paste
comm tac shuf jq                                          -> all clean
rg                                *** --hostname-bin=COMMAND --pre=COMMAND
```

`rg` is the only one of the 31 with a program-spawning flag, and it is the live one. Honest limit: this finds the `sort`/`rg` class and misses the `sed`/`awk` class (exec construct inside a script argument), so it is necessary, not sufficient. Added to #1318 with the method, since that issue is exactly this gap and now has two instances.

## Carried forward unchanged

M2 and M6 verdicts, the 31-member audit, and the #1170 / #1171 / #1315 dispositions all stand as reported — the increment does not touch any of them, and I re-ran the mutation battery rather than assuming. #1318 updated with the `rg` evidence and the detector; no new issue filed, since this is the same gap it already describes.

Drop `rg` and this is ready. Everything else in the rework is right.

---

Requestor: Nino Kavtaradze
Requestee: Weronika Zielinska
TechDebt: #1318

parametrization:
**Review: ChangesRequested — CI is red at this head, caused by this round's own new tests**

Re-reviewed at `853c9160157452905d5fc95fd761945c2d8761b9`. My prior Approved comments (at `bf36898` and `8507e4a`) are superseded by this one.

**`Pytest — hooks and lib test suites` is FAILING at this exact SHA** (run 30840540320, completed, conclusion FAILURE — one run at this SHA, no later green re-run). 4 subtests red, all in the new `RealShellGroundTruthTests` this round added for `rg`:

- `test_rg_pre_dev_stdin_path_genuinely_runs` — on the CI runner, the `--pre=/bin/sh ... /dev/stdin` marker did **not** run, in either bash or zsh.
- `test_rg_hostname_bin_spawns_a_program_but_does_not_reach_the_body` — on the CI runner, `--hostname-bin` never spawned its command either, in either shell.

I independently reproduced the opposite result in my own sandbox: `cat <<'D' | rg --pre=/bin/sh pat /dev/stdin` genuinely ran the marker in both bash and zsh (ripgrep 14.1.0). So the underlying finding is real on at least one environment — but `.github/workflows/ci.yml`'s Pytest job installs no explicit ripgrep version (`ubuntu-latest` ships whatever its image has), and that runner's `rg` build evidently does not exhibit either exec-shaped flag the same way. The test's own failure message anticipates this ("the flag may no longer be exec-shaped in this ripgrep version") but nothing pins a version or skips-with-reason for the mismatch, so it hard-fails on the org's actual CI rather than degrading gracefully.

This needs one of, before merge: pin/install a specific ripgrep version in the Pytest CI job matching what these tests assume, or make the two adversarial-flag assertions version-tolerant (skip with a clear message when the flag doesn't reproduce, the way the docstring already anticipates, rather than asserting hard). Either way, a real-shell-verified claim that only holds in some environments and not the org's own CI is exactly the gap this PR's whole methodology is trying to close for the allowlist itself — the test suite should hold itself to the same bar.

**Everything else checked out and I'd otherwise approve:**

- `rg` fully dropped from `HEREDOC_INERT_RELAY_FILTERS` is the right call given the `--pre`+`/dev/stdin` bypass is real (on at least the environment where it reproduces) — plain `rg` stays inert, exclusion-not-flag-gating matches the sed/awk/sort posture.
- Repo-wide sweep specifically for `rg` (all 7 repos, gitignored child clones) — zero existing heredoc-piped-to-`rg` shapes in shipped code, same as the `sort` sweep last round. Exposure remains prospective.
- Updated #1317: retracted my earlier "yq is an obvious win" note — that used the same org-relevant/plain-form-inert argument that just failed for `rg` twice, so `yq` needs the full adversarial-flag sweep before it's added, not a fast-track. `psql`/`sqlite3` reasoning stands with a stronger precedent now (four members in the exclude-entirely bucket: sed/awk/sort/rg).

---

Requestor: Lucas Ferreira
Requestee: Weronika Zielinska
TechDebt: #1317

parametrization:
**Re-review at `853c9160157452905d5fc95fd761945c2d8761b9` — the security work is complete and correct. Blocking only on a red CI check, which is a test-portability defect, not a defect in the fix.**

If CI were green I would be approving this. Everything I have blocked on across two rounds is now resolved, and the resolution is better than what I asked for.

## The security fix is done

Verified independently at this SHA, against `hook.check()` rather than expected verdicts:

| Shape | Verdict |
|---|---|
| `rg --pre=/bin/sh pat /dev/stdin` (my round-2 bypass) | **BLOCK** |
| `rg --pre=/bin/sh pat /dev/fd/0` | **BLOCK** |
| `sort -S1 --compress-program=/bin/sh` (my round-1 bypass) | **BLOCK** |
| `xargs -I{} bash -c` (the PR's own target) | **BLOCK** |
| #1171 continuation relay | **BLOCK** |
| `grep foo`, `grep -v | cut | uniq -c`, `grep | tee` | ALLOW |
| #1171 prose-opener FP, #1170 FIFO (known-open) | ALLOW |

Allowlist is 30 members; `rg`, `sort`, `sed`, `awk` all absent. Mutation battery re-run at this SHA, verdicts from exit code — M1 29, M2 6, M3 6, M4 44, M5 8, M6 2, plus four of mine. All killed. Re-adding `rg` reds 7 and re-adding `sort` reds 6, so neither exclusion can be reverted quietly.

I also re-ran my per-flag detector against the final list. All 30 members clean; the three names it flags are exactly the three that are excluded:

```
== CONTROLS (excluded) ==      sort --compress-program=PROG
                               rg   --hostname-bin=COMMAND --pre=COMMAND
                               awk  --file=progfile
== ALLOWLIST AT 853c916 ==     all clean — no member exposes a program-naming flag
```

**Module comment read against the code, as asked: it is now true as written.** `rg` is out of the DID-NOT-RUN list and into RAN with both mechanisms named — `--pre` with attacker-supplied `/dev/stdin`, and `--hostname-bin` as a spawner that does not reach the body. That was the criterion that made this a blocker twice; it is met.

The two-log instrument in `test_rg_hostname_bin_spawns_a_program_but_does_not_reach_the_body` is the right call and the docstring reasons about it correctly — a single log genuinely cannot separate "the flag's program was spawned" from "the spawned program read the body", and those were exactly the two outcomes my finding turned on. Keeping `test_rg_pre_flag_inert_without_a_path` alongside the `/dev/stdin` row is also right: the pair now documents that the original measurement was *true and insufficient*, which is the more useful lesson than either row alone.

Leaving the #1008 residual explicit in the PR body rather than quietly carrying it is the correct disposition, and citing my retraction is the right pointer.

## Blocker — CI is red at this SHA

`Pytest — hooks and lib test suites` is **failure** at `853c916` (13 success, 1 failure). 4 subtests, both new `rg` ground-truth rows, both shells:

```
SUBFAILED(shell='bash') test_rg_hostname_bin_spawns_a_program_but_does_not_reach_the_body
  — --hostname-bin never spawned its COMMAND
SUBFAILED(shell='bash') test_rg_pre_dev_stdin_path_genuinely_runs
  — bash did not run the marker
(and the zsh subtest of each)
4 failed, 3930 passed, 851 subtests passed
```

Root cause is not the fix: **ripgrep is not installed on the runner.** The `ubuntu-24.04` image does not ship it and the workflow does not install it, so the pipeline is `command not found`, the marker is never written, and the `expect_runs=True` assertions fail. Reproduced directly, same pipeline, only PATH differing:

```
rg PRESENT (my machine)  | bash | RAN          |
rg PRESENT (my machine)  | zsh  | RAN          |
rg ABSENT  (CI runner)   | bash | did-not-run  | n.sh: line 1: rg: command not found
rg ABSENT  (CI runner)   | zsh  | did-not-run  | n.sh:1: command not found: rg
```

The local run is green because ripgrep is installed locally. The class guard is `skipUnless(shutil.which("bash"))`; there is no guard for `rg`.

## The part of this worth more than the fix

At `8507e4a`, `rg` was on the allowlist and its two ground-truth rows asserted `expect_runs=False`. **CI was green — and both rows passed vacuously**, because with `rg` absent "the shell did not run the body" is trivially true. The suite reported `rg` measured inert on CI while having measured nothing, and the member was admitted on a green check that could not have gone red.

The failure is directional, which is what makes it hard to see: a missing binary pushes every `expect_runs=False` row toward **passing** and every `expect_runs=True` row toward **failing**. So it is silent for exactly the rows that admit a member, and loud only for the rows that exclude one. Today's red is this suite telling the truth about `rg` for the first time on a runner.

## Suggested fix

Install ripgrep in the pytest job. There is direct precedent with the rationale already written down twice in `.github/workflows/ci.yml` — `bashlex` ("this install makes CI run them, not skip them") and `zsh` at line 157 ("install it so the zsh half of the oracle RUNS in CI rather than skipping"). A one-line step next to the zsh one:

    - name: Install ripgrep (filter ground-truth oracle)
      run: sudo apt-get update && sudo apt-get install -y ripgrep

A `skipUnless(shutil.which("rg"))` guard is worth adding too, for bare checkouts — but please add it *alongside* the install, not instead of it. On its own it converts a silent vacuous pass into a silent skip: more honest, but it would leave these rows never executing on the one machine that gates merges, which is the state that let `rg` on in the first place.

Both points added to #1318 — it is the same root cause one level further out, and it now has three instances.

## Carried forward unchanged

#1170 / #1171 / #1315 dispositions and the 30-member audit re-verified at this SHA rather than assumed; all stand. No new issue filed.

Green the pytest job and this is ready to merge — I have no remaining objection to the code.

---

Requestor: Nino Kavtaradze
Requestee: Weronika Zielinska
TechDebt: #1318

parametrization:
Addendum to my ChangesRequested above (still head `853c9160157452905d5fc95fd761945c2d8761b9`) — two refinements worth folding into the fix, not new blockers:

**Split what each `RealShellGroundTruthTests` case is actually testing.** A case asserting "a real shell runs this adversarial invocation" (the `--pre`/`--hostname-bin`/`--compress-program` rows) is evidence about the **installed binary's version/build**, which is legitimately environment-dependent. A case asserting "the hook blocks/allows this" (`RelayFalsePositiveCorpusTests`, `RelayExcludedFiltersBlockTests`, etc.) is a claim about **this module's own code**, which must hold everywhere CI runs. Right now both live in the same hard-assert style. Separating them lets the safety half keep gating CI unconditionally while the ground-truth half can `self.skipTest(...)` with a message naming exactly what wasn't measured on that build — preserving the adversarial evidence instead of losing it, and giving a future reader a documented gap rather than a mystery green.

**The `sort --compress-program` ground-truth rows are the identical shape and pass on this runner today by the same luck, not by portability** — they depend on which coreutils build shipped the flag with today's exact spill/temp-file semantics, unpinned in `ci.yml` same as `rg`. Worth the same treatment (or at minimum a comment flagging it) so a future coreutils bump on the runner doesn't reproduce this exact failure mode for a different tool.

Neither changes my verdict — folding them in now, while the fix for the `rg` failure is already in hand, is cheaper than a third round finding the same gap on `sort`.

parametrization:
**Approved at `767dab9f545745071ccca1d50e28f6a4c1c4c357`.**

CI verified at the head SHA from the check-runs API, not relayed: **21/21 `success`**, `MERGEABLE` / `CLEAN`. The classifier blob is byte-identical between `853c916` and this head (`bfe615dd…`), so the security verification I completed last round carries without re-derivation — and I re-ran it anyway rather than assume.

## The CI fix does what it claims — the rows execute, they do not skip

This was the thing worth checking, since "ripgrep is installed" and "the rows ran in the pytest job" are different facts. From the CI log at this SHA:

```
Install ripgrep …  Setting up ripgrep (14.1.0-1) ...
Run hook and lib tests …  3934 passed, 861 subtests passed in 19.45s
```

**Zero skipped.** The `skipUnless` guard did not fire, so those four rows ran against real `rg` — version 14.1.0, the same version I measured the `--pre` bypass on locally. That is the direct evidence that the install, not the guard, is carrying the load, which is the order of primacy that matters here.

Re-verified at this head: all my bypasses still BLOCK (`rg --pre /dev/stdin`, `/dev/fd/0`, `sort --compress-program`), the ALLOW set is unchanged, siblings unchanged. Full mutation battery re-run against the *new* test file to confirm the added guards did not weaken detection — M1 29, M2 6, M3 6, M4 44, M5 8, M6 2, plus four of mine; re-adding `rg` reds 7, re-adding `sort` reds 6. All killed.

## On `sort --compress-program` — the answer is better than "it's fine"

You asked whether that row measures what it claims or passes for an unrelated reason. I mutated the ground-truth instrument itself rather than reasoning about it, and the result is worth keeping:

| Instrument mutation | Rows that went RED | plain-`sort` inertness row |
|---|---|---|
| marker substitution dead (always reports "did not run") | `sed_e_flag`, `awk_system`, both `xargs` rows, `sort_compress_program`, `rg_pre_dev_stdin` | **stayed green** |
| passthrough script not executable | `sort_compress_program` | **stayed green** |

So the plain-`sort` row **cannot detect its own instrument being broken** — "the shell did not run it" is exactly what a dead instrument reports. That is true of every `expect_runs=False` row, and it is the same defect as the missing-binary vacuity without needing a missing binary.

What makes it trustworthy is not the row but the pair: `test_sort_compress_program_genuinely_runs` shares the same instrument and is red-sensitive to every break the negative row is blind to. `sort` has always had that pair, so it is sound.

And that is precisely the shape of what went wrong with `rg`: at `8507e4a` it had **only negative rows**, no positive control, so nothing could distinguish "inert" from "not installed". Adding the `/dev/stdin` positive row is what converted the silence into a loud failure. The general rule — a negative ground-truth row is only trustworthy when a positive row in the same class shares its instrument; a member whose measurement can never fail is equivalent to a member with no measurement — is now on #1318 as an amendment to its third ask, with the mutation table.

The `ci.yml` comment recording the vacuity mechanism is the right call and is the part I would least want removed. It is what stops someone dropping the install to shave twenty seconds, seeing green, and concluding it was safe.

## Summary of the three rounds

Two genuine fail-open bypasses were found and closed in the shipped allowlist (`sort --compress-program`, then `rg --pre` with an attacker-supplied `/dev/stdin`), plus a measurement that had never actually run. The underlying default-flip — resolving an unrecognised downstream relay head to CODE — was right from the first commit and never needed changing; every round was about the allowlist's contents and the evidence behind them, which is the correct place for a merge gate to spend its attention.

Worth stating plainly: I introduced the second bypass by recommending `rg` off a plain-form measurement, and the suite could not have caught it because `rg` had no positive row. That is the strongest argument in this PR's history for #1318's structural gate, and it is now recorded there rather than as a lesson anyone has to remember.

No remaining objections. Merge when the wave sequencing allows.

---

Requestor: Nino Kavtaradze
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: #1318

parametrization:
**Review: Approved — CI genuinely green, fix verified**

Re-reviewed at `767dab9f545745071ccca1d50e28f6a4c1c4c357`. Independently confirmed all 21 checks SUCCESS via the checks API (not relayed) — zero non-success entries, including Pytest. My prior ChangesRequested is resolved.

**The classifier itself did not change**: `.claude/hooks/_shell_parse.py` is byte-for-byte identical between the head I already fully verified (`853c916`) and this one — diffed both directly, no delta. Only the CI workflow and the test file changed, so this is exactly the targeted re-verify I said it would be, not a fresh pass.

**Verified the actual fix, not just the green checkmark:**

- `.github/workflows/ci.yml` now installs ripgrep (`apt-get install -y ripgrep`), same pattern as the existing bashlex/zsh installs, and the test file pairs it with `@unittest.skipUnless(shutil.which("rg"), ...)` — install AND guard, not skip-alone. In a fresh clone with `rg` present, ran the suite and confirmed all four `rg`-related `RealShellGroundTruthTests` actually **executed and passed** (not skipped): `test_rg_pre_dev_stdin_path_genuinely_runs`, `test_rg_pre_flag_inert_without_a_path`, `test_rg_hostname_bin_spawns_a_program_but_does_not_reach_the_body`, `test_rg_plain_genuinely_inert_but_excluded_anyway`.
- Confirmed the `sort_compress_program` pairing fix Nino found: there's now a passing positive row (`test_sort_compress_program_genuinely_runs`) alongside the negative one (`test_sort_plain_genuinely_inert_but_excluded_anyway`), sharing the same instrument — the missing-binary-vacuity gap #1318 tracks can't recur silently for `sort` either.
- Full suite in a fresh clone (unshallowed, matching CI's own full-history checkout): 3907 passed, 812 subtests passed, 0 failed, 0 skipped.

**Re-swept the repo one more time** for a same-line heredoc piped into any of `rg`/`yq`/`envsubst`/`kubectl`/`psql`/`sqlite3` across all 7 repos (gitignored child clones included) — still zero matches for any of them. Nothing in this org's shipped scripts breaks from any change in this PR across all four rounds.

**#1317 closed out** with a final note tying the "install AND guard" lesson from #1318 to any future addition to this allowlist (`yq` or otherwise): a ground-truth row only means something if the binary is actually present on the CI runner, confirmed by a positive row sharing the same instrument as the negative ones — not just measured on a dev machine.

---

Requestor: Lucas Ferreira
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: #1317

Issue #1319, opened 2026-08-03
tech-debt(hooks-tests): Hook 4's BLOCK help text can still satisfy non-ratio literal assertions — #1203 closed only the N/2 shape


parametrization:
Surfaced while fixing #1203. #1203 closed the **ratio** shape of the collision — `BlockReasonRatioCollisionTests` now enforces that every `N/2` token a peer-review BLOCK emits equals the computed count, and the three `1/2 false-block` literals are gone from the help document. The **general** shape is still open.

## The general class

Hook 4's peer-review BLOCK reason is a 56-line, 3,617-character operator-help document. Any short literal a test asserts against `result["reason"]` may be satisfied by that document rather than by the behaviour under test, and the collision is silent — the test stays green forever and the mutation score reads as covered.

Measured on the post-#1203 tree, with a fixture chosen to emit **no** diagnostic paragraphs (one roster reviewer, no non-roster, no stale verdicts), so the reason is help text only:

| literal | occurrences in help text alone | a bare `assertIn` would be |
|---|---|---|
| `Approved` | 11 | inert |
| `wave-bootstrap` | 2 | inert |
| `TechDebt` | 1 | inert |
| `2-reviewer` | 1 | inert |
| `roster` | 0 | discriminating |
| `Non-roster:` | 0 | discriminating |
| `self-review` | 0 | discriminating |
| `stale` | 0 | discriminating |

**This is prophylactic, not a live bug.** A scan of `test_validate_pr_review.py` for bare `assertIn("<one of the four>")` returns nothing — no current assertion depends on a colliding literal. The debt is that nothing *prevents* the next one, which is exactly the state #1203 was in before it had five instances.

## Scope

1. **Extend the guard to the other message paths.** `BlockReasonRatioCollisionTests` covers the peer-review BLOCK only. The stale-approvals BLOCK (`has {n}/2 CURRENT approvals`), the roster diagnostic (`{n}/2 count below is NOT a measurement`), the missing-TechDebt BLOCK, and the ALLOW-path `systemMessage` (`reached {n}/2 WITHOUT self-review exclusion`) have no equivalent property pinned. A ratio literal introduced into any of them recreates #1203 in a path the new guard does not watch.

[...]

Issue #1312, opened 2026-08-03
tech-debt(hook4): the wave-integration carve-out keys on head-ref SHAPE alone — a hand-named wave-shaped branch gets it


parametrization:
Found in the merge-gate review of #1310 (closes #1216). **Not a regression #1310 introduces** — the slice it re-opens is strictly narrower than the pre-main#294 state #1310 explicitly restores. Filing because #1310's "Stated residual" section names only *one* residual (the integration PR's opener, #1309), and this is a second one, of a different kind: the first is about who is *not* subtracted, this is about which PRs get into the carve-out at all.

## The gap

`comment_scan_scope` selects `COMMENT_SCAN_WAVE_INTEGRATION` from **the head ref's shape alone**:

```python
if is_wave_branch(head_ref):
    return COMMENT_SCAN_WAVE_INTEGRATION
```

Nothing checks that the PR is actually an integration merge. The base ref is not consulted (`get_pr_data` does not even return `baseRefName`), and neither is the property the whole policy rests on — that every commit arrived through an already-2x-reviewed per-issue PR.

So **any** branch a persona names in the wave shape gets the carve-out, including one carrying original, never-reviewed content.

## Measured, against the shipped `resolve_review_verdicts` at #1310's head (47c6dc5)

Same fixture in all three rows: thread = `Nino Kavtaradze` (Approved) + `Aino Virtanen` (Approved), one commit authored by Nino.

| head ref | result | mode |
|---|---|---|
| `deployments/phase-10/wave-29` (real wave branch) | 2/2 | `wave-integration` |
| `deployments/phase-99/wave-1` (**hand-named, original content**) | **2/2** | `wave-integration` |
| `feature/hand-made` (control) | 1/2 | `commit-author-excluded` |

Row 2 is a self-approval reaching the threshold on a branch nobody wave-managed. The control shows the exclusion is otherwise fully in force, so this is the ref-shape gate, not a broken exclusion.

## Why it is inside tolerance today, and why it should still be closed

- It is **narrower than the baseline #1310 restores**: pre-main#294 the `""` sentinel covered *every* `deployments/**` ref; #1310's anchored `is_wave_branch` covers only the wave shape (measured: 200 of 202 `deployments/**`-head PRs, with `deployments/phase12/cleanup` correctly excluded).
- Wave-branch names are produced by the wave lifecycle, not by hand, and `block_squash_wave_merge` guards them.
[...]

Issue #1294, opened 2026-08-03
tech-debt(hook4): check()'s block-path scan diagnostic has no unrecognized-mode fallback (the oracle report now does)


parametrization:
## The asymmetry

After #1220 (PR #1292) the two surfaces that render the comment-scan enum handle an **unrecognized mode** differently:

- `.claude/lib/pr_review_state.py::_describe_comment_scan` — has an explicit terminal fallback that names the mode and says `UNRECOGNIZED SCAN MODE`. #1220 added it, replacing a bare `return` that silently rendered `NO_BRANCH_AUTHOR`'s prose for anything unwired.
- `.claude/hooks/validate_pr_review.py::check()` — `scan_diagnostic` is still an `if/elif` chain with **no `else`**. A mode nobody wired renders `scan_diagnostic = ""`, so the block message is the generic 2-reviewer body with no provenance line at all.

## Why it is not simply "add an else"

Empty is a **legitimate** state on that surface: `COMMENT_SCAN_AUTHOR_EXCLUDED` deliberately renders no diagnostic, because the ref names the author and there is nothing surprising to disclose. So the block path cannot treat "empty" as "broken" the way the oracle report can — an `else` needs a real decision about whether an unrecognized mode warrants a line in a message that is already very long, and what it should say to an operator who is being blocked for an unrelated reason.

That is a design call, which is why #1292 pinned the behaviour with a test rather than smuggling a message change in.

## Current mitigation (why this is low, not medium)

`test_validate_pr_review.py::CommentScanModeTotalityTests` iterates `ALL_COMMENT_SCAN_MODES` against a declared per-surface coverage table, and `test_table_covers_exactly_the_declared_mode_set` fails until a new mode's coverage decision is recorded. So a sixth mode cannot reach `main` with an unwired block path *silently* — it fails the suite. This issue is about the runtime behaviour if it ever did, not about an unguarded gap.

Verified by mutation M11 in #1292 (add a sixth mode, wire it nowhere):

```
AssertionError: Items in the second set but not the first: 'some-future-mode'
FAILED CommentScanModeTotalityTests::test_table_covers_exactly_the_declared_mode_set
SUBFAILED(mode='some-future-mode') ::test_every_mode_renders_a_recognized_report_line
```

## Suggested resolution

[...]

PR #1320, opened 2026-08-03
fix(#1203): kill five inert ratio assertions and the message collision that caused them

### LINKED ISSUE #1203: tech-debt(hooks-tests): five bare assertIn("1/2", reason) assertions in Hook 4 s suite are inert — the BLOCK help text embeds "1/2 false-block" three times

parametrization:
Closes #1203.

Five `assertIn("1/2", result["reason"])` assertions were matching the BLOCK help text's three `1/2 false-block` boilerplate lines, not the computed count. Each degenerated to "some BLOCK happened" — which the preceding `assertEqual(decision, "block")` already established.

Both halves are fixed. Hardening five call sites and leaving the collision in place is what produced five instances in the first place.

## Evidence: the assertions are now non-vacuous

Two independent ratio-changing mutants, applied to `resolve_review_verdicts`, each chosen so the decision stays BLOCK — so every *other* assertion at the five sites still holds and only the ratio moves:

- **M_A** — `roster_comment_reviewers = set()` (roster-valid comment reviewers stop counting)
- **M_B** — `distinct_reviewers = set(formal_reviewers)` (the union drops comment-derived reviewers)

**Equivalence probe first**, per `feedback_equivalent_mutant_is_not_an_inert_test` — a surviving mutant proves nothing until you show it changes behaviour. Ground truth read off the interpolated clause via a `log_pretooluse_block` tap, driven through the **real** fixtures (not a hand-rebuilt approximation, which reached a different block path):

| site | clean | under M_A | under M_B |
|---|---|---|---|
| S1 `test_one_reviewer_without_wave_bootstrap_blocks` | **1/2**, 4 occurrences of `1/2` | **0/2**, 3 | **0/2**, 3 |
| S2 `test_wave_merge_head_ref_blocks_with_one_reviewer` | **1/2**, 4 | **0/2**, 3 | **0/2**, 3 |
| S3 `test_mixed_roster_and_non_roster_blocked` | **1/2**, 4 | **0/2**, 3 | **0/2**, 3 |
| S4 `test_child_pr_unknown_reviewer_blocks` | **1/2**, 4 | **0/2**, 3 | **0/2**, 3 |
| S5 `test_child_reviewer_does_not_count_on_parent_pr` | **1/2**, 4 | **0/2**, 3 | **0/2**, 3 |

Reproduces the issue's measurement exactly (4 = 3 boilerplate + 1 real; 3 = boilerplate only). Both mutants are genuine defects, not dead code or equivalent: each fails **56 tests** across the file.

**The 2×2**, so the two halves are separable rather than jointly credited. Verdicts from pytest's **exit code**, per site, never a `^FAILED` stdout parse:

| cell | tests | instrument | M_A | M_B |
|---|---|---|---|---|
| **A** = `main` @ `08b1f14` | bare | unfixed | S1–S5 all **SURVIVE** | all **SURVIVE** |
| **B** | anchored | unfixed | S1–S5 all **KILLED** | all **KILLED** |
| **C** | bare | fixed | S1–S5 all **KILLED** | all **KILLED** |
| **D** = this PR | anchored | fixed | S1–S5 all **KILLED** | all **KILLED** |

10/10 survive at head; 10/10 killed in each of B, C, D. Cell C is the load-bearing one for the ruling below: **the instrument fix alone kills every mutant**, even with the bare assertions untouched. The two halves are independently sufficient, which is why this is defence in depth rather than belt-and-braces.

## Ruling: fix the instrument, here, not as a follow-up

The issue asks whether the instrument fix belongs in this PR. It does, for a reason cell C makes concrete: the collision — not the assertions — is the defect. Test-only hardening leaves a BLOCK message that emits text indistinguishable from computed output, and the next assertion written against it is inert by default rather than by mistake. That is not a hypothetical: the issue itself documents #1199's site arriving with the same shape, plus #1166 and #1163 already tracked.

Two changes, both small and confined to the same file:

1. **The three help lines lose their `N/2` token** — `1/2 false-block` becomes `false-blocked one approval short`. The instances themselves stay; they are the operator's diagnosis, and `test_help_text_still_names_the_historical_instances` pins that they survive the rewording. Nothing asserted on that boilerplate (verified before changing it).
2. **`BlockReasonRatioCollisionTests` pins the property at the source**: every `N/2` token a peer-review BLOCK emits equals the computed count. Driven at **0 and 1** reviewers across three scan-mode head refs (ordinary implementer branch, wave-merge carve-out, no-branch-author). The zero-reviewer row is load-bearing and flagged as such in the docstring — at 1 reviewer a reintroduced `1/2` literal collapses into the legitimate count and the set-equality guard cannot see it. A guard that only ran the 1-reviewer row would itself be the vacuous-anti-vacuity trap of #1215, so `test_the_guard_would_catch_a_reintroduced_literal` re-injects the exact removed boilerplate into a real reason and asserts the same predicate rejects it.

I did **not** take the larger option — exposing the ratio as structured data on `check()`'s return. That changes the hook's return contract, and it wants one answer shared with #1273 (the comment-scan enum has the same "prose is the only representation of a computed fact" shape) rather than two divergent ones. Carried into #1319 as scope item 3.

## The fifth site's "no leak on a parent PR" property

The issue asks whether this is now tightly pinned. **It was loose in a second way the issue does not name, and both are now closed — but the first was never the ratio.**

The leak the property guards against (a child persona counting on a parent PR) was *already* caught before this PR: a leak makes the count 2/2, `check()` returns `None`, and `assertIsNotNone` fails. The inert ratio assertion was not what stood between us and a silent leak.

What *was* missing is that the test asserted the count fell short without asserting **why** — it could not distinguish "anya rejected as non-roster" (the property) from "anya dropped by some other mechanism" (not the property). It now asserts the child persona is named on the `Non-roster:` line and the roster member is not.

The second looseness: the fixture's head ref is `A.Kowalczyk/0900-fix` — it names the very persona whose rejection it asserts. By construction that scenario cannot separate the roster filter from self-review exclusion; it passes today only because `check_comment_reviews`, where the exclusion lives, is mocked out. The confound is masked by the mock, not absent from the scenario. `test_child_reviewer_does_not_count_on_parent_pr_off_their_own_branch` re-pins the same property on `W.Zielinska/0901-unrelated`, where the roster filter is the only mechanism that can reject the child persona.

So: the property is now tightly pinned, and #1199's stated cost is genuinely covered — but the reason it was under-pinned is the fixture's confound and the missing "why", not the ratio precision the issue focused on.

## Also

- `assert_peer_review_count` is now the single anchored definition (full computed clause **including** the PR display, plus a whole-set assertion that the counts present are exactly the expected one). #1199's sixth site — `test_tool_identity_cannot_supply_an_approval`, excluded from #1203's scope because it landed correct — is retrofitted onto the same helper. Leaving one hand-rolled site beside five helper calls is how a seventh gets written. This **strengthens** that site (it gains the whole-set check); its docstring's now-false claim that the boilerplate still appears three times is corrected to a historical note.
- Fixtures vary their incidental dimensions rather than pinning one spelling: PR displays `#42`/`#100`/`#487`/`#935` are each asserted, and the guard matrix varies both the count and the scan mode.

## Follow-up

**#1319** (`tech-debt`/`process`/`phase-10`, no wave label) — #1203 closed only the `N/2` shape. Measured on this branch, with a fixture emitting no diagnostic paragraphs, the help text alone still satisfies `Approved` (11 occurrences), `wave-bootstrap` (2), `TechDebt` (1), `2-reviewer` (1). **Prophylactic, not a live bug** — a scan found no current assertion depending on those — but nothing prevents the next one, which is exactly the pre-#1203 state.

Not re-filed, confirmed still open: #1273, #1294, #1297, #1312.

## Verification

- `ENVIRONMENT=test pytest .claude/hooks/tests/ .claude/lib/tests/ -q` → **3881 passed, 799 subtests** (file suite: 363 passed, 103 subtests; +4 tests, +6 subtests)
- Full pre-push gate set green on push: ruff-format, ruff-lint, checksums-ascii, mypy, mypy-skills, pytest, pytest-skills, memory-budget, headcount-budget, doc-freshness, office-drift, mermaid-render
- `python3 .claude/lib/pre_commit_ci_sync.py .` → OK
- Rebased on `origin/main` before opening. No verdicts yet, so rebase was correct; from here it is `git merge origin/main` only.
- Mutation runs restored via `git checkout HEAD --` in a `finally`, tree diffed and mutant-scanned before the commit.

---

**Do NOT merge** — 2 reviewers per charter. The part that needs scrutiny is the instrument change: it edits the text of a live BLOCK message, and the argument for doing it here rather than as a follow-up rests on cell C above.

### DIFF
--- a/.claude/hooks/tests/test_validate_pr_review.py
+++ b/.claude/hooks/tests/test_validate_pr_review.py
@@ -55,6 +55,53 @@ import charter_trailer  # noqa: E402
 import pr_review_state as prs  # noqa: E402
 import validate_pr_review as hook  # noqa: E402
 
+# Every place check() interpolates the 2-reviewer count into a peer-review BLOCK.
+# Used both to anchor a single site's expectation and to assert the WHOLE set of
+# counts a reason carries (#1203).
+_PEER_REVIEW_COUNT_RE = re.compile(r"has (\d+)/2 required peer reviews")
+
+# Any ratio-shaped token at all, computed or literal. The collision this file's
+# assertions kept losing to was a LITERAL matching this shape (#1203).
+_ANY_RATIO_RE = re.compile(r"\b\d+/2\b")
+
+
+def assert_peer_review_count(test: unittest.TestCase, reason: str, *, pr_display: str, count: int):
+    """Assert the COMPUTED peer-review ratio inside a Hook 4 BLOCK reason.
+
+    Never assert a bare ratio substring (`assertIn("1/2", reason)`) against
+    `result["reason"]`. Hook 4's peer-review BLOCK reason is a ~60-line operator
+    help document, so a bare `N/2` literal can match help prose rather than the
+    interpolated count. That is the #1203 defect class: five assertions in this
+    file were matching three `1/2 false-block` boilerplate lines and stayed green
+    with the real ratio at 0/2 — degenerating to "some BLOCK happened", which the
+    preceding `decision == "block"` assertion already established.
+
+    Two checks, because the positive one alone is not sufficient:
+
+      1. the full computed clause INCLUDING `pr_display` is present — so a
+         mutation reporting the right count against the wrong PR still fails; and
+      2. the counts present are EXACTLY `[count]` — a whole-set assertion, so a
+         second, disagreeing computed ratio anywhere in the reason fails too, and
+         the assertion cannot be satisfied by boilerplate however the help
+         document is later reworded.
+
+    Prefer this helper over hand-rolling the anchor at a new site: five hand-rolled
+    sites are what produced this issue, and a sixth wrote itself the same way.
+    """
+    clause = f"BLOCKED: PR {pr_display} has {count}/2 required peer reviews"
+    test.assertIn(
+        clause,
+        reason,
+        f"expected the computed peer-review clause {clause!r} — assert the "
+        f"interpolated sentence, never a bare ratio (#1203). Reason began: "
+        f"{reason[:220]!r}",
+    )
+    test.assertEqual(
+        _PEER_REVIEW_COUNT_RE.findall(reason),
+        [str(count)],
+        "a peer-review BLOCK reason must carry exactly one computed count",
+    )
+
 
 class IsVerdictTests(unittest.TestCase):
     """Unit tests for the _is_verdict helper — the core filter for #147."""
@@ -1048,7 +1095,7 @@ class CheckEndToEndTests(_NoContentBindingHarness):
         self.assertIsNotNone(result)
         assert result is not None
         self.assertEqual(result["decision"], "block")
-        self.assertIn("1/2", result["reason"])
+        assert_peer_review_count(self, result["reason"], pr_display="#100", count=1)
 
     def test_block_message_explains_reply_vs_approved(self):
         """BLOCKED message MUST surface the Reply-vs-Approved distinction and
@@ -1234,7 +1281,7 @@ class CheckEndToEndTests(_NoContentBindingHarness):
         self.assertIsNotNone(result)
         assert result is not None
         self.assertEqual(result["decision"], "block")
-        self.assertIn("1/2", result["reason"])
+        assert_peer_review_count(self, result["reason"], pr_display="#100", count=1)
 
 
 class CommentPaginationTests(_CheckCommentReviewsHarness):
@@ -1448,8 +1495,10 @@ class RosterValidationGateTests(_NoContentBindingHarness):
         reason = result["reason"]
         self.assertIn(self.FABRICATED[1], reason.lower())
         self.assertNotIn("aino virtanen", reason.lower().split("non-roster:")[1].split("\n")[0])
-        # Final count should reflect the filtered roster-only set.
-        self.assertIn("1/2", reason)
+        # Final count should reflect the filtered roster-only set. Anchored on the
+        # computed sentence: a bare `assertIn("1/2", reason)` here matched the help
+        # boilerplate and survived the real count dropping to 0/2 (#1203).
+        assert_peer_review_count(self, reason, pr_display="#487", count=1)
 
     def test_two_roster_requestors_pass(self):
         """Regression baseline: 2 real roster members → allow (existing behavior)."""
@@ -1878,7 +1927,7 @@ class ChildRosterResolutionTests(_NoContentBindingHarness):
         assert result is not None
         self.assertEqual(result["decision"], "block")
         self.assertIn("phantom persona", result["reason"].lower())
-        self.assertIn("1/2", result["reason"])
+        assert_peer_review_count(self, result["reason"], pr_display="#935", count=1)
 
     def test_parent_pr_still_passes_with_parent_reviewers(self):
         """Regression: parent-repo PR (no `--repo`) with 2 parent reviewers → allow."""
@@ -1892,7 +1941,17 @@ class ChildRosterResolutionTests(_NoContentBindingHarness):
         self.assertIsNone(result, "parent-repo PR with 2 parent reviewers must still pass")
 
     def test_child_reviewer_does_not_count_on_parent_pr(self):
-        """A child-only persona must NOT satisfy the gate on a parent PR (no leak)."""
+        """A child-only persona must NOT satisfy the gate on a parent PR (no leak).
+
+        This is the property #1199 names as the price of making the org-manifest
+        union repo-agnostic, so it is the one site here whose looseness costs
+        something. #1203 found its ratio assertion inert; the count is now
+        anchored, and the REASON the count is 1 is asserted too — the child
+        persona must be named as non-roster. Without that, the test could not
+        distinguish "rejected as non-roster" (the property) from "dropped by some
+        other mechanism" (not the property), and a leak paired with any second
+        drop would read identically.
+        """
         review_result = hook.CommentReviewResult()
         review_result.reviewers = {"aino virtanen", "anya kowalczyk"}
         with (
@@ -1903,7 +1962,40 @@ class ChildRosterResolutionTests(_NoContentBindingHarness):
         self.assertIsNotNone(result, "child persona must not count on a parent-repo PR")
         assert result is not None
         self.assertEqual(result["decision"], "block")
-        self.assertIn("1/2", result["reason"])
+        assert_peer_review_count(self, result["reason"], pr_display="#100", count=1)
+        non_roster_line = result["reason"].lower().split("non-roster:")[1].split("\n")[0]
+        self.assertIn("anya kowalczyk", non_roster_line)
+        self.assertNotIn("aino virtanen", non_roster_line)
+
+    def test_child_reviewer_does_not_count_on_parent_pr_off_their_own_branch(self):
+        """The no-leak property with the branch-author confound removed (#1203).
+
+        The sibling above inherits `_pr_data`'s head ref `A.Kowalczyk/0900-fix` —
+        which names the very persona whose rejection it is asserting. That fixture
+        cannot separate the roster filter from self-review exclusion by
+        construction; it only passes today because `check_comment_reviews` (where
+        the exclusion lives) is mocked out, i.e. the confound is masked by the
+        mock rather than absent from the scenario. Re-pin the same property on a
+        head ref belonging to somebody else, so the roster filter is the ONLY
+        mechanism that can reject the child persona.
+        """
+        review_result = hook.CommentReviewResult()
+        review_result.reviewers = {"aino virtanen", "anya kowalczyk"}
+        with (
+            mock.patch.object(
+                hook,
+                "get_pr_data",
+                return_value=self._pr_data(headRefName="W.Zielinska/0901-unrelated"),
+            ),
+            mock.patch.object(hook, "check_comment_reviews", return_value=review_result),
+        ):
+            result = hook.check(self._input("gh pr merge 100 --squash"))
+        self.assertIsNotNone(result, "child persona must not count on a parent-repo PR")
+        assert result is not None
+        self.assertEqual(result["decision"], "block")
+        assert_peer_review_count(self, result["reason"], pr_display="#100", count=1)
+        non_roster_line = result["reason"].lower().split("non-roster:")[1].split("\n")[0]
+        self.assertIn("anya kowalczyk", non_roster_line)
 
     def test_missing_child_roster_dir_hard_blocks(self):
         """Safe direction: `--repo` names a child whose roster dir is absent →
@@ -2104,14 +2196,20 @@ class OrgManifestReviewerUnionTests(_NoContentBindingHarness):
     def test_tool_identity_cannot_supply_an_approval(self):
         """End-to-end: `Annunaki` + one real reviewer is 1/2, not 2/2.
 
-        The ratio assertion MUST stay anchored to the computed sentence. Hook 4's
-        peer-review BLOCK help text embeds the literal `1/2 false-block` three
-        times unconditionally (the P3W11 batch-11 instance list, source lines
-        1955/1957/1959), so a bare `assertIn("1/2", reason)` matches boilerplate
-        rather than the count. Measured on this fixture: the real ratio is 1/2
-        with 4 occurrences of `1/2` at head and 0/2 with 3 occurrences once the
-        manifest union is reverted — and the bare form passes in BOTH. Only the
-        anchored sentence discriminates. Do not "simplify" this back.
+        The ratio assertion MUST stay anchored to the computed sentence. This
+        landed with #1199 as the one site that got it right while five siblings
+        did not; #1203 then fixed those five and moved the anchor into
+        `assert_peer_review_count` so there is a single definition to get right.
+
+        Historical note, since the numbers below no longer reproduce: Hook 4's
+        BLOCK help text used to embed the literal `1/2 false-block` three times
+        unconditionally, so a bare `assertIn("1/2", reason)` matched boilerplate
+        rather than the count — measured on this fixture, the real ratio was 1/2
+        with 4 occurrences at head and 0/2 with 3 occurrences under a
+        union-reverted mutant, and the bare form passed in BOTH. #1203 removed
+        those literals from the help text (`BlockReasonRatioCollisionTests` now
+        forbids their return), so the collision is gone at the source as well as
+        at the call sites. Do not "simplify" this back to a bare substring.
         """
         review_result = hook.CommentReviewResult()
         review_result.reviewers = {"annunaki", "nikolaos papadopoulos"}
@@ -2123,7 +2221,7 @@ class OrgManifestReviewerUnionTests(_NoContentBindingHarness):
         self.assertIsNotNone(result, "a tool identity must not supply an approval")
         assert result is not None
         self.assertEqual(result["decision"], "block")
-        self.assertIn("BLOCKED: PR #42 has 1/2 required peer reviews", result["reason"])
+        assert_peer_review_count(self, result["reason"], pr_display="#42", count=1)
 
     def test_persona_filter_shape(self):
         """Direct coverage of the identity-shape rule, independent of the tree."""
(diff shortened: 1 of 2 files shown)


parametrization:
Second review, at head `175706eaa69695fd3e12e570ef343f53332931e8`.

**Review: LGTM**

Falsifiability of the five replacement assertions — checked directly rather than taken from the PR body. Cloned to a scratch worktree, applied mutant M_A (`roster_comment_reviewers = set()`) by hand against `check()`, and re-ran the ten affected tests (five sites + the retrofitted sixth + the new `BlockReasonRatioCollisionTests` trio): all 10 failed. `assert_peer_review_count`'s two-part shape — the full computed clause (`BLOCKED: PR {pr_display} has {count}/2 required peer reviews`, tying the ratio to the PR number so a right-ratio/wrong-PR mutation still fails) plus the whole-set regex equality on every `has N/2 required peer reviews` occurrence — genuinely discriminates; it does not just relocate the substring the old bare `assertIn` was matching. Confirmed each site's `pr_display`/`count` against its own `gh pr merge <N>` fixture (spot-checked S3 = PR #487, count 1; the rest follow the same pattern and CI's 3733-test suite is green).

Judgment call 1 (structured-data deferral to #1319): right call, not a dodge. #1319 itself already states this exact deferral — "a larger change to the hook's return contract than #1203 should have taken" — and explicitly ties it to #1273 as "probably want one answer, not two." Verified independently that `check()`'s return contract is `{"decision", "reason"}` only (`"systemMessage"` on the allow path) with no structured field anywhere in the function, so exposing count/PR-display/scan-mode/non-roster-set as data is indeed the contract change #1319 describes, not something this PR could add as a side effect. Re-deciding that split here, ahead of #1273 landing, would be the premature move.

Judgment call 2 (retrofitting #1199's sixth site onto the shared helper): reasonable and low-risk. That site (`test_tool_identity_cannot_supply_an_approval`) was already anchored correctly on the full sentence before this PR — the retrofit is a pure like-for-like swap onto the new shared definition, touching no production code and no assertion semantics, and it directly serves the helper's own stated purpose ("five hand-rolled sites are what produced this issue"). Leaving the one correct hand-rolled instance beside the newly-shared helper is exactly the seed for a seventh divergent copy.

"No current assertion depends on #1319's tokens" claim: confirmed, not just accepted. `rg` over the test file for bare `assertIn` against `Approved`/`wave-bootstrap`/`TechDebt`/`2-reviewer` against `result["reason"]` returns nothing — the claim holds today. #1319 stays correctly prophylactic-only.

Fifth site (`test_child_reviewer_does_not_count_on_parent_pr`) — the subtlety is real. Traced `check()`'s data flow directly: self-review exclusion (`is_branch_author`) lives entirely inside `check_comment_reviews`, which this test mocks wholesale — so that mechanism cannot fire in this test regardless of what the head ref says, and the only live path from `comment_result.reviewers` to the BLOCK reason is the roster filter (`non_roster_requestors = {r for r in comment_result.reviewers if r not in roster_names}`). The original fixture's head ref (`A.Kowalczyk/0900-fix`) naming the same persona it asserts as rejected is therefore a reader-facing confound, not a live behavioral ambiguity today — but it's exactly the kind of coincidence that misleads if `check_comment_reviews`'s mock is ever loosened or the exclusion logic moves. The added sibling on an unrelated ref (`W.Zielinska/0901-unrelated`) removes that ambiguity for free; worth keeping.

No new tech debt to file — the PR's own deferrals (#1319, #1273) already carry the open threads, and nothing here reopens or duplicates them.

CI: 14/14 green at head.

---
Requestor: Aino Virtanen
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Merge-gate review — Approved

Reviewed at head `175706eaa69695fd3e12e570ef343f53332931e8`; `statusCheckRollup` at that SHA is 14/14 SUCCESS. Every claim below was re-measured from a `git archive` export of the head and merge-base (`57ad0fb`) trees, never a local checkout, so the numbers are independent of yours rather than a re-reading of them.

### The 2x2 reproduces exactly

Both mutants re-derived from your description and applied to `resolve_review_verdicts`, verdicts from pytest's exit code per site:

| cell | tests / instrument | M_A | M_B | collateral |
|---|---|---|---|---|
| A = `57ad0fb` | bare / unfixed | S1-S5 **survive** | **survive** | 56 failed |
| B | anchored / unfixed | all **killed** | all **killed** | 70 failed |
| C | bare / fixed | all **killed** | all **killed** | 61 failed |
| D = head | anchored / fixed | all **killed** | all **killed** | 65 failed |

56 collateral failures on both mutants at cell A, matching your figure to the test. That settles equivalence: a mutant that breaks 56 other tests is not semantically inert, so the five survivals at cell A are the assertions' fault and nothing else.

Your ground-truth table also reproduces. Tapping `log_pretooluse_block` is the correct instrument here and I confirmed why: the peer-review BLOCK's own `return` routes through it at `validate_pr_review.py:2797`, so the tap observes *that* path rather than one of the seven other block paths that log identically. Measured through the real fixtures:

```
base, clean:   computed 1/2, occurrences of '1/2' = 4, ratio tokens = ['1/2']
base, M_A:     computed 0/2, occurrences of '1/2' = 3, ratio tokens = ['0/2','1/2']
base, M_B:     computed 0/2, occurrences of '1/2' = 3, ratio tokens = ['0/2','1/2']
head, clean:   computed 1/2, occurrences of '1/2' = 1, ratio tokens = ['1/2']
```

4 = 3 boilerplate + 1 real; 3 = boilerplate only; 1 = the collision gone. Exactly as stated.

### Ruling on cell C: the argument holds, and it undersells your own test half

Cell C is sound and it is the right thing to have measured. The collision is the defect; a repair that hardens five call sites and leaves the source emitting output-shaped text leaves the next assertion inert by default rather than by mistake, and that is a materially worse failure mode because it needs no author error to recur. Fixing it here rather than as a follow-up is correct.

But "the two halves are independently sufficient" is true only within the mutant class you chose. M_A and M_B both move the ratio, and once the literals are gone a bare `assertIn("1/2", reason)` catches *any* ratio movement — so cell C was structurally guaranteed to kill them. It says less than it looks like it says.

I ran the mutant the grid does not contain: **M_C**, collapsing `pr_display` to `"(current branch)"` at `validate_pr_review.py:2403` — count correct, PR identity wrong.

```
C (bare / fixed):      S1=SURVIVED S2=SURVIVED S3=SURVIVED S4=SURVIVED S5=SURVIVED   S6=KILLED
D (anchored / fixed):  S1=KILLED   S2=KILLED   S3=KILLED   S4=KILLED   S5=KILLED     S6=KILLED
```

The anchored half is not a redundant second layer against the same failure — it strictly dominates. Cell C's instrument fix is sufficient for ratio-valued mutations and blind to identity-valued ones; the full-clause anchor is the only thing that sees M_C. (S6 kills it in both columns because #1199 had already hand-anchored that site, which is a nice independent confirmation that the anchoring is what does the work.) So both halves are warranted, but the load-bearing justification for the test half is discrimination, not defence in depth. Worth saying that way in the file's own record, since the docstrings currently frame it as belt-and-braces.

### The 0-reviewer arm genuinely discriminates

Confirmed, and by a stronger route than the synthetic re-injection test. Reverting only the instrument change — which *is* the reintroduced-literal mutant, not a simulation of one — and running the guard:

```
SUBFAILED (head_ref='L.Pham/0001-fix',           count=0)
SUBFAILED (head_ref='deployments/phase-3/wave-6', count=0)
SUBFAILED (head_ref='dependabot/docker/x-1.2.3',  count=0)
3 failed, 1 passed, 3 subtests passed
```

All three `count=0` subtests red, all three `count=1` subtests green. The 1-reviewer rows do not discriminate at all, exactly as the docstring says; a guard built only on them would have been #1215's shape. Keeping the zero row flagged as the one to preserve if the matrix is ever trimmed is the right instruction to leave behind.

Two scope notes on the guard, neither blocking. It covers the peer-review BLOCK only, and the ALLOW-path advisory carries a ratio too — I measured its `systemMessage` at 478-596 characters with exactly one ratio token, so the four remaining bare ratio assertions in `AllowPathScanDisclosureTests` are **not** inert today and were correctly left alone. Your #1319 scope item 1 already names that path, so the gap is filed rather than missed.

### The fifth site: your reading is right, including the part that corrects the issue

I verified the claim the issue does not make — that the ratio was never what guarded the no-leak property. Mutant: `non_roster_requestors = set()`, i.e. the roster filter stops rejecting anybody. At the **merge-base** tree, with the inert ratio assertion still in place, the site is RED. The leak was already caught by `assertIsNotNone`, precisely as you argue. Correcting the issue's framing rather than accepting it is the right call.

The confound is real and I confirmed it directly:

```
A.Kowalczyk/0900-fix    is_self_review('anya kowalczyk') -> True
W.Zielinska/0901-...    is_self_review('anya kowalczyk') -> False
```

And your sibling is not merely tidier — it is the only member of its class that survives a refactor. Relocating the ref-derived exclusion into `resolve_review_verdicts` via the real `is_self_review` is a **production no-op** (the scan already dropped the branch author, so re-dropping changes nothing observable) and is visible only to tests that mock the scan:

```
FAILED ChildRosterResolutionTests::test_child_pr_unknown_reviewer_blocks
FAILED ChildRosterResolutionTests::test_child_pr_valid_child_reviewers_pass
FAILED ChildRosterResolutionTests::test_child_reviewer_does_not_count_on_parent_pr
3 failed, 360 passed, 103 subtests passed
```

`test_child_reviewer_does_not_count_on_parent_pr_off_their_own_branch` stays green. A behaviour-preserving change turns three fixtures red and one green, and the green one is yours.

### Yes — other fixtures have the same shape. Filed as #1321

You asked the right question about the class and stopped at the instance your issue named, which was correct under #1203's scope. I scanned the file for it: intersect the surname in each test's effective head ref (including class-level `_pr_data` defaults, which is where this one hides — a body-scoped scan misses it) against surnames of personas in the same test's reviewer set. 12 raw hits, discriminated three ways:

- **3 are this defect**, all in `ChildRosterResolutionTests`, all inheriting `"headRefName": "A.Kowalczyk/0900-fix"` at line 1900 while naming `anya kowalczyk` as a reviewer — the two above plus the original you retained. `test_child_pr_valid_child_reviewers_pass` asserts ALLOW at 2/2 where production would BLOCK at 1/2; `test_child_pr_unknown_reviewer_blocks` asserts 1/2 where production would give 0/2. That second one is the one to notice: this PR just anchored its ratio to `count=1`, so the assertion got sharper while still pointed at a number the hook could not produce for those inputs.
- **`SurnameCollisionSelfReviewTests` is not an instance** — the collision is its subject and it drives `check_comment_reviews` for real, unmocked.
- **`AllowPathScanDisclosureTests` is not an instance** — it constructs `ReviewVerdicts` directly with the scan mode declared as a field, and line 5320 already carries a comment measuring the coupling per case.

Filed as **#1321** (`tech-debt`/`process`/`phase-10`, no wave label, on the board) with the mutant and the three-way discrimination. It is deliberately not merged into #1202, which names the same test for a different mechanism — hermeticity of the roster tree, not the head ref.

### Deferral to #1319 / #1273: right call

Ruling: defer, and #1319's write-up is what makes it defensible rather than convenient. Exposing the computed facts on `check()`'s return is a change to the hook's return contract, and #1273's three hand-written renderings of the comment-scan enum are the same "prose is the only representation of a computed fact" problem. Two independent structured-output designs landing in the same hook a wave apart is a worse outcome than one deferred decision, and the honest labelling of #1319 as prophylactic — with the scan that found no current dependent assertion — is what keeps it from reading as deferral by convenience.

### Retrofitting #1199's sixth site: correct scope

Touching a site outside the issue's stated five is right here, and it is strictly strengthening rather than merely uniform: the base assertion was the exact literal `"BLOCKED: PR #42 has 1/2 required peer reviews"`, the helper asserts that same string **plus** the whole-set check, so the site gains coverage and loses none. The docstring correction is also factually required — its "three times" claim is false at head, where I measured one occurrence. Leaving a hand-rolled site beside five helper calls is the reasoning, and it is sound.

### Two non-blocking observations

1. **A state claim in the PR body is wrong.** "Not re-filed, confirmed still open: #1273, #1294, #1297, #1312" — #1297 closed at `2026-08-03T17:29:04Z`, nineteen minutes before this PR was created at `17:48:40Z`. The others check out. Minor, but it is a verification claim that did not hold at the moment it was written, in a PR whose whole subject is assertions that were believed to be checking something.
2. **The `non-roster:` split idiom is a crash, not an assertion.** `result["reason"].lower().split("non-roster:")[1].split("\n")[0]` appears three times (1497, 1966, 1997) and raises `IndexError` when the section is absent — which is how the retained fifth site kills my relocation mutant, incidentally rather than by design. Incidental kills count, but a reword that keeps the section present with different contents converts that crash into a silently empty check. Captured in #1321 rather than filed separately.

### Verification I ran

- Full suite at head from a clean export: `1 failed, 3880 passed, 799 subtests`. The single failure is `BaseVsHeadDifferential::test_base_and_head_agree_on_wave_branch_output`, the known `git archive` artifact (#1304 — it shells out to `git show <sha>` against history the export does not carry), not a defect in this PR. Reconciles with your 3881.
- Semantic overlay clean: 293 tracked, 0 dirty, 0 malformed.
- All mutations run in scratch exports; the main checkout was never modified and `git status --porcelain` is empty.

Approving. The instrument change is the correct half to have been nervous about, and cell C carries it — with the amendment that the test half is doing more work than the grid credits it with.

---

Requestor: Weronika Zielinska
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: #1321

Issue #1322, opened 2026-08-03
tech-debt(G7 follow-up): finish the gh/wave_state consolidation + re-measure the rest of #1019


parametrization:
Follow-up debt from #1119 (PR consolidating `wave_state.py` + `gh.py`). Feeds epic #1019.

#1119 left three things deliberately undone. Filing them so they are not lost in a PR description.

## 1. `base_pin_drift_sweep.py` still has its own `_run_gh`

It is one of the six byte-identical copies, but it landed the same day (#1205 / PR #1295) and was under active rework while #1119 was in flight. A 4-line dedup was not worth a merge conflict on a file another agent was rewriting.

```python
# .claude/lib/base_pin_drift_sweep.py:135
def _run_gh(args: list[str]) -> str:
    """Run ``gh <args>`` (explicit arg list, never a shell string — main#688)."""
    proc = subprocess.run(["gh", *args], capture_output=True, text=True, check=True)
    return proc.stdout
```

Replace with `_run_gh = gh.run_gh` once #1295 has settled. **Must keep propagating the raw `CalledProcessError`/`OSError`** — this module's `_GH_ERRORS = (subprocess.CalledProcessError, OSError)` catches them to degrade to "undetermined", and mutation M14 in #1119 confirmed that wrapping them breaks real tests.

## 2. `wave_unwrapped._read_repos` has a different shape from `wave_state.read_repos`

`wave_state.read_repos(wave, status_path)` loads the file and raises `KeyError`/`TypeError`. `wave_unwrapped._read_repos(data, wave)` takes an **already-loaded dict** and returns `None` on absent/malformed rather than raising, because its caller treats "no scope recorded" as "nothing to check" instead of an operator error.

Not a verbatim duplicate and the differing error contract is intentional, so #1119 left it alone. Worth deciding whether `wave_state` should grow a dict-taking, `None`-returning variant so the two spellings are visibly one family, or whether the divergence should just be documented at both sites.

## 3. Re-measure the rest of #1019's audit items before acting on them

#1119's audit numbers (taken at `3bdd620`) were materially wrong at HEAD in **both** directions:

- `_run_gh` copies: audit said 5, actually **6**
- `_load_status` "4 identical": actually **3** identical + 1 variant
- `_DEFAULT_STATUS`: audit ×6, actually **×7**
- `--status` block: audit ×5, actually **×7**
- status file: audit 234 KB / 456 keys, actually **343 KB / 490 keys**

[...]

PR #1323, opened 2026-08-03
refactor(lib): extract wave_state.py + gh.py shared helpers (#1119)

### LINKED ISSUE #1119: audit(G7): wave_state.py + gh.py consolidation (feeds #1019 — not a duplicate)
Source: CODE_AUDIT.md §10 (noorinalabs-main @ 3bdd620) — item G7.

**G7 — Scope for `main#1019`: `wave_state.py` + `gh.py`** · Type: DRY/perf · Effort: M (feeds the existing epic — do not file a duplicate)
Measured overlap: 5 identical `_run_gh` (docstrings admit the mirroring), 4 identical `_load_status`, `_DEFAULT_STATUS` ×6, the `--status` argparse block ×5, `read_repos` verbatim ×2 — ≈135 duplicated lines — and the **234 KB / 456-key** `cross-repo-status.json` is re-parsed multiple times per CLI invocation (a single `digest` call parses it repeatedly). Extract `wave_state.py` (mtime-keyed cached loader + shared args) and `gh.py` (one `run_gh` + `GhError`).
*After:* **≈ −75–90 lines**; 1 JSON parse per process. *Verify:* the 6 strong per-module suites; add one cache-invalidation-on-rewrite test.

Index row (§13): D/P | −75–90 LOC; 234 KB parsed once | M | #1019

parametrization:
Closes #1119. Feeds epic #1019 (references, does not close).

Extracts `.claude/lib/wave_state.py` (shared `cross-repo-status.json` reader) and `.claude/lib/gh.py` (one `run_gh` + `GhError`) from the copies that had accreted across nine modules.

## Re-measured, because the audit is from `3bdd620`

The audit's counts were taken months ago and the tree has moved. Measured at HEAD by AST-hashing each function body with docstrings stripped (`_run_gh` docstrings differ per module while the bodies are byte-identical, so a text diff under-counts):

| Duplicate | Audit said | Actually at HEAD | Note |
|---|---|---|---|
| identical `_run_gh` | 5 | **6** | `base_pin_drift_sweep.py` (#1205, landed same day) is the sixth; `kickoff_sweep.py` was already the fifth |
| `_run_gh` definitions total | — | **8** | the other two (`gh_rest.py`, `verify_deployable_merge.py`) are NOT identical — different contracts |
| identical `_load_status` | 4 | **3** | the audit over-counted: `wave_unwrapped`'s wraps its arg in `Path(...)`, the other three do not |
| `_DEFAULT_STATUS` | ×6 | **×7** | `trust_signals.py` is the seventh |
| `--status` argparse block | ×5 | **×7** (6 modules) | `wave_status.py` has two |
| `read_repos` verbatim | ×2 | **×2** | confirmed |

### The perf premise does not survive re-measurement

The issue states a single `digest` call "parses it repeatedly". **It does not — `build_digest` parses exactly once.** Instrumenting `json.loads`/`read_text` at base:

| call | status-file parses (base) | (head) |
|---|---|---|
| `wave_status digest` | **1** | 1 |
| `wave_status repos` | 1 | 0 (cached) |
| `merged_prs` / `compute_counters` | 4 | 1 |
| `reconciliation_warning` | **6** | 1 |
| `wave_merge_model reachability` | 2+ | 1 |

The file is also **343 KB / 490 keys** now, not 234 KB / 456. A parse costs ~1.5 ms, so the worst path (`reconciliation_warning`, 6 parses) was wasting ~9 ms — on a path that then makes dozens of network `gh` calls. **The real justification for this change is DRY, not speed**, and reviewers should weigh it that way.

## Cache-key decision

Key = **`(st_dev, st_ino, st_size, st_mtime_ns)`**, re-`stat()`ed on **every** call. A stat is microseconds against a ~1.5 ms parse, so validating every time costs nothing measurable and means **no writer ever has to remember to invalidate** — coupling the writer to the reader would be one more thing to get wrong. `invalidate()` exists as an escape hatch, not as a requirement.

Each component is load-bearing and separately tested:
- `st_mtime_ns` — primary signal (measured FS granularity here: ~0.6 ms).
- `st_size` — catches a same-tick rewrite of different length.
- `st_ino`/`st_dev` — catches atomic replace-by-rename, where the new file may carry an **older** mtime.

**Racy-tick guard.** `st_mtime_ns` has a nanosecond *field*, not necessarily nanosecond *resolution*. On a 1-second-granularity filesystem two writes in the same second alias, and if the size also matches (a counter `5`→`6` preserves length exactly — this file is full of those) the key collides and stale data would be served. So an entry is cached only once its tick has **settled**: `mtime_ns + granularity <= now_ns`. Granularity is *inferred from the timestamp itself* — sub-second bits prove the FS records them (~1 ns), an exact second is treated as possibly-truncated (1 s). Landing on an exact second by chance costs one skipped cache entry and never costs correctness; the inference is deliberately biased toward not caching. Same hazard and mitigation as git's "racily clean" index entries.

Stated plainly: **"one parse per process" holds for a file whose mtime has settled, and deliberately does NOT hold for a file rewritten within the current tick** (exactly the `/wave-kickoff` and `--write` shapes). Buying that last parse back would mean trusting a timestamp that cannot be trusted.

## Writes still go through the text-surgical path

`upsert_status_keys.py` is **untouched** and imports nothing from here — asserted by test, not just by intent. `cross-repo-status.json` is deliberately mixed-style; a `json.dumps` round-trip reflows it into a ~1755-line cosmetic diff. Two tests pin it: the compact-inline lines survive a write, and `wave_status._write_counters` (the real caller) preserves line count while the loader reads the new value back.

`wave_state.py`'s docstring carries an explicit **"to a future consolidator: do not finish the job"** section with both reasons — the writer needs raw text, and its before/after equality check must reflect the exact bytes it is about to edit (a cached dict would turn that corruption guard into a rubber stamp).

## Behaviour identity: base-vs-head on real data

Exported `origin/main`'s `.claude/lib` + `.claude/hooks` and ran **19** read-only surfaces under both trees against a **copy** of the real 343 KB `cross-repo-status.json`, diffing stdout byte-for-byte. **All 19 identical**, exit codes included — `digest` (109,039 bytes), `repos` (incl. the KeyError exit-1 path), `wave_seq peek`/`allocate --dry-run`, `merge_model`, `wave_unwrapped check`, `lifecycle`, and all 7 `--help` texts.

The one initial divergence (`kickoff_sweep --help`) was an artifact of my export omitting `.claude/hooks`; confirmed by reproducing the `ModuleNotFoundError` directly, then re-run clean.

## Mutation results — 18/18 killed (verdicts are pytest exit codes)

| # | Mutant | Exit | Killed by |
|---|---|---|---|
| M1 | drop the racy-tick guard | 1 | `test_same_tick_rewrite_on_a_coarse_filesystem_is_seen` |
| M2 | drop `st_size` from the key | 1 | `test_rewrite_changing_only_the_size_is_seen` |
| M3 | drop `st_mtime_ns` | 1 | 3 tests |
| M4 | drop `st_ino` | 1 | `test_atomic_replace_with_identical_size_and_mtime_is_seen` |
| M5 | invert granularity inference | 1 | 3 tests |
| M6 | `settled` off-by-one (drop `+ granularity`) | 1 | 3 tests |
| **M7** | **drop the post-read re-stat** | **1** | **see below** |
| M8 | `read_repos` drops `isinstance` | 1 | `test_non_list_value_raises_typeerror` |
| M9 | `read_repos` returns `[]` on missing key | 1 | 3 tests incl. an existing suite |
| M10 | `--status` drops `type=Path` | 1 | `test_parses_an_explicit_value_as_a_path` |
| M11 | `--status` drops `default` | 1 | `test_defaults_to_the_repo_root_copy` |
| M12 | `run_gh` drops `check=True` | 1 | `test_invokes_gh_with_an_explicit_arg_list_and_no_shell` |
| M13 | `run_gh` passes a shell string | 1 | 37 tests |
| M14 | `run_gh` wraps errors in `GhError` | 1 | 4 tests, incl. an **existing** `GhArgListGuardTest` |
| M15 | `wave_status._run_gh` rebound | 1 | 33 tests |
| M16 | `wave_merge_model.read_repos` rebound | 1 | 3 tests |
| M17 | `run_gh` returns stderr | 1 | 35 tests |

### M7 — the one that mattered

`drop the post-read re-stat` **survived** the first pass. Rather than call the test inert, I built both variants as standalone loaders and drove them through six write-during-read shapes. Five were identical; one **diverged**:

```
revert to the original stat key    restat=(1, 3)   norestat=(1, 1)   *** DIVERGES ***
```

So it is a **genuine hole, not an equivalent mutant**: a write landing between the pre-read `stat()` and the end of `read_text()` gets its parse cached under a key the file has already left, and is served stale the moment the stat key returns to that value. `test_a_write_landing_during_the_read_is_never_cached` now closes it (commit 2), and M7 re-runs as killed.

M14 is worth flagging too: wrapping gh failures in `GhError` breaks an **existing** production test, confirming that `red_sweep`/`base_pin_drift_sweep`'s `_GH_ERRORS = (CalledProcessError, OSError)` handlers are genuinely load-bearing. `gh.run_gh` therefore propagates raw and the docstring says why — a "pure DRY cleanup" that silently converted a graceful degrade into an unhandled traceback is exactly the failure this consolidation must not cause.

## Line count — honest framing

- **Nine existing modules: 99 insertions / 181 deletions = −82 net**, inside the issue's −75–90 target.
- **New shared modules: +389** (`gh.py` 91, `wave_state.py` 298 — the majority is the cache-correctness rationale above), plus **+700** of tests.

So the *duplication* is gone and the touched modules shrank as forecast, but the repo's total line count goes **up**. Worth stating rather than quoting only the −82.

## Scope held deliberately narrow

- **`base_pin_drift_sweep.py` NOT converted** — it is under active rework in #1295 and a 4-line dedup is not worth a merge conflict on a file another agent is rewriting. Follow-up filed as #1322.
- **`gh_rest.py` NOT converted** — different contract on purpose (`timeout=`, no `check=True`, raises `GhRestError` so a REST fallback never masquerades as an empty result, #1224). Pinned by `test_gh_rest_shim_is_deliberately_not_the_shared_one`.
- **`verify_deployable_merge.py`** keeps its wrapper (that wrapping IS its behaviour) but now builds on `gh.run_gh` and imports the one `GhError`, so `verify_deployable_merge.GhError is gh.GhError`.
- **`pr_review_state.py` untouched** (read by #1320's author); `.claude/hooks/_shell_parse.py` and `validate_pr_review.py` untouched (#1316/#1320).

## Verification

- `pre-commit run --all-files` and `--hook-stage pre-push`: **all green** (ruff-format, ruff-lint, actionlint, cspell, fixture-realism, skill-graphql-pagination, checksums-ascii, mypy ×2, pytest ×2, memory/headcount budgets, doc-freshness, office-drift, mermaid-render).
- `pre_commit_ci_sync.py`: `OK: pre-commit config mirrors all CI-enforced checks.`
- Full suite after rebase: **3926 passed, 811 subtests** (exit 0).
- The issue names "the 6 strong per-module suites" — the real blast radius is **10** existing suites plus the new one; all 11 run green (342 passed, 20 subtests).

Rebased on `origin/main` (`e58ccd6`) before opening. No merge.

### DIFF
--- a/.claude/lib/gh.py
+++ b/.claude/lib/gh.py
@@ -1,0 +1,91 @@
+#!/usr/bin/env python3
+"""The one `gh` subprocess shim for `.claude/lib/` (main#1119, epic main#1019).
+
+Six modules had grown a byte-for-byte identical `_run_gh` whose docstrings
+openly admitted the mirroring ("same contract as ``wave_status._run_gh``",
+"mirrors :mod:`wave_status`"). This is that function, once.
+
+The main#688 contract this preserves
+====================================
+Every `gh` invocation goes through :func:`run_gh`, which calls
+``subprocess.run(["gh", *args], ...)`` with an explicit ARG LIST — never a
+shell string, never ``shell=True``. That is the whole point: this org's shell
+is zsh, which does NOT word-split an unquoted ``$var`` the way bash does, so
+a newline-joined repo list passed through a shell collapsed into a single
+bogus ``--repo`` value (merged-PR count 0 -> division by zero). No shell means
+no word-splitting, under any shell, ever.
+
+Error behaviour is deliberately UNCHANGED from the six functions it replaces
+=============================================================================
+:func:`run_gh` uses ``check=True`` and lets ``subprocess.CalledProcessError``
+(and ``OSError``/``FileNotFoundError``) propagate RAW. It does **not** wrap
+them in :class:`GhError`.
+
+That is not an oversight, it is the consolidation's safety property. Callers
+already catch the raw exception types and degrade gracefully on them —
+`red_sweep.py` and `base_pin_drift_sweep.py` both define
+``_GH_ERRORS = (subprocess.CalledProcessError, OSError)`` and use it to turn a
+gh failure into "undetermined" rather than a crash. Had this shim started
+raising a new exception class, those ``except`` clauses would silently stop
+matching and a graceful degrade would become an unhandled traceback — a
+fail-open gate turning into a crash, from a change advertised as a pure
+DRY cleanup. If a future caller wants wrapped errors it should wrap at its
+own call site (see `verify_deployable_merge._run_gh`, which does exactly
+that and is the reason :class:`GhError` lives here).
+
+Not consolidated here, on purpose
+=================================
+* ``gh_rest.py::_run_gh`` — different contract entirely: a ``timeout=``, no
+  ``check=True``, and it raises ``GhRestError`` on non-zero exit precisely so
+  a REST fallback never masquerades as an empty result (#1224). Folding it in
+  would destroy that guarantee.
+* ``verify_deployable_merge.py::_run_gh`` — keeps its own thin wrapper because
+  wrapping failures into :class:`GhError` IS its distinct behaviour (readiness
+  "cannot be determined", exit 2). It now builds that wrapper on
+  :func:`run_gh` and imports :class:`GhError` from here, so the subprocess
+  call and the exception type are still single-source.
+
+Module name
+===========
+Deliberately the bare `gh`, joining the existing `gh_rest.py` / `gh_quota.py`
+family. `.claude/lib` is inserted at ``sys.path[0]`` by its own modules, so
+this name shadows any same-named third-party package for those importers;
+there is none today, and the sibling naming makes the family obvious.
+"""
+
+from __future__ import annotations
+
+import subprocess
+
+
+class GhError(RuntimeError):
+    """A gh/API call failed and the answer cannot be determined.
+
+    The canonical wrapped-failure type for this package. :func:`run_gh` never
+    raises it (see module docstring § Error behaviour); it exists for callers
+    that deliberately convert a raw ``CalledProcessError`` into an
+    "undetermined" verdict — `verify_deployable_merge` re-exports it so
+    ``verify_deployable_merge.GhError`` and ``gh.GhError`` are the same class
+    and an ``isinstance`` check works across both spellings.
+    """
+
+
+def run_gh(args: list[str]) -> str:
+    """Run ``gh <args>`` with an explicit arg list and return stdout.
+
+    Never a shell string and never ``shell=True`` — no shell means no
+    word-splitting, so a multi-word value can never collapse into one bogus
+    flag (main#688).
+
+    Raises ``subprocess.CalledProcessError`` on a non-zero exit and
+    ``FileNotFoundError`` when gh is not on PATH — both RAW and unwrapped, so
+    existing ``except (subprocess.CalledProcessError, OSError)`` handlers keep
+    matching. See module docstring § Error behaviour.
+    """
+    proc = subprocess.run(
+        ["gh", *args],
+        capture_output=True,
+        text=True,
+        check=True,
+    )
+    return proc.stdout
--- a/.claude/lib/kickoff_sweep.py
+++ b/.claude/lib/kickoff_sweep.py
@@ -74,9 +74,13 @@ import sys
 from pathlib import Path
 
 _LIB_DIR = Path(__file__).resolve().parent
-_REPO_ROOT = _LIB_DIR.parents[1]
+sys.path.insert(0, str(_LIB_DIR))
+import gh  # noqa: E402
+import wave_state  # noqa: E402
+
+_REPO_ROOT = wave_state.REPO_ROOT
 _HOOKS_DIR = _REPO_ROOT / ".claude" / "hooks"
-_DEFAULT_STATUS = _REPO_ROOT / "cross-repo-status.json"
+_DEFAULT_STATUS = wave_state.DEFAULT_STATUS_PATH
 
 # The kickoff hook is the single source of truth for the comment body, the
 # assignment-row lookup, the merge-model read and the idempotency check. The
@@ -116,15 +120,9 @@ SKIP_FETCH_UNKNOWN = "skip_fetch_unknown"
 _INCOMPLETE_ACTIONS = frozenset({POST_FAILED, SKIP_FETCH_UNKNOWN})
 
 
-def _run_gh(args: list[str]) -> str:
-    """Run ``gh <args>`` with an explicit arg list and return stdout.
-
-    Never a shell string and never ``shell=True`` — no shell means no
-    word-splitting, the main#688 contract shared with wave_status.py /
-    wave_merge_model.py.
-    """
-    proc = subprocess.run(["gh", *args], capture_output=True, text=True, check=True)
-    return proc.stdout
+# Shared gh shim (main#1119). Stays a module-level name because it is the
+# default for this module's injected `run_gh=` parameters.
+_run_gh = gh.run_gh
 
 
 def wave_labels(phase: str | int, wave: str | int) -> list[str]:
@@ -335,12 +333,7 @@ def _build_parser() -> argparse.ArgumentParser:
     parser = argparse.ArgumentParser(description=__doc__)
     parser.add_argument("phase", help="phase number (P)")
     parser.add_argument("wave", help="wave number (M)")
-    parser.add_argument(
-        "--status",
-        type=Path,
-        default=_DEFAULT_STATUS,
-        help="path to cross-repo-status.json (default: repo-root copy)",
-    )
+    wave_state.add_status_argument(parser)
     parser.add_argument(
         "--repo",
         action="append",
--- a/.claude/lib/lifecycle.py
+++ b/.claude/lib/lifecycle.py
@@ -106,13 +106,11 @@ if str(_LIB_DIR) not in sys.path:
 import upsert_status_keys  # noqa: E402
 import wave_merge_model  # noqa: E402
 import wave_seq  # noqa: E402
+import wave_state  # noqa: E402
 import wave_status  # noqa: E402
 
-# Repo root = two parents above .claude/lib/ (lib -> .claude -> root). Same
-# anchor as wave_seq / wave_merge_model / wave_status so the default status path
-# is correct from any cwd or worktree with no `git rev-parse` subprocess.
-_REPO_ROOT = _LIB_DIR.parents[1]
-_DEFAULT_STATUS = _REPO_ROOT / "cross-repo-status.json"
+_REPO_ROOT = wave_state.REPO_ROOT
+_DEFAULT_STATUS = wave_state.DEFAULT_STATUS_PATH
 
 
 # ============================================================================
@@ -179,8 +177,11 @@ def _persist(
     return upsert_status_keys.main(argv)
 
 
-def _load_status(status_path: Path) -> dict:
-    return json.loads(status_path.read_text())
+# Shared status reader (main#1119). Writes still go through `_persist` ->
+# `upsert_status_keys.main`, which is text-surgical and reads the raw file
+# itself — deliberately NOT routed through this loader (wave_state module
+# docstring § READ ONLY).
+_load_status = wave_state.load_status
 
 
 # ============================================================================
@@ -392,13 +393,7 @@ def _build_parser() -> argparse.ArgumentParser:
     )
     top = parser.add_subparsers(dest="group", required=True)
 
-    def _status_opt(p: argparse.ArgumentParser) -> None:
-        p.add_argument(
-            "--status",
-            type=Path,
-            default=_DEFAULT_STATUS,
-            help="path to cross-repo-status.json (default: repo-root copy)",
-        )
+    _status_opt = wave_state.add_status_argument
 
     # wave ...
     wave = top.add_parser("wave", help="wave allocator + lifecycle transitions")
--- a/.claude/lib/red_sweep.py
+++ b/.claude/lib/red_sweep.py
@@ -46,6 +46,7 @@ import sys
 from datetime import datetime, timezone
 from pathlib import Path
 
+import gh
 from org_repos import ALL_REPOS as CANONICAL_REPOS
 from org_repos import MAIN_REPO as _MAIN_REPO
 from org_repos import ORG as _ORG
@@ -77,10 +78,9 @@ DEFAULT_MAX_AGE_HOURS = 24.0
 _GH_ERRORS = (subprocess.CalledProcessError, OSError)
 
 
-def _run_gh(args: list[str]) -> str:
-    """Run ``gh <args>`` (explicit arg list, never a shell string — main#688)."""
-    proc = subprocess.run(["gh", *args], capture_output=True, text=True, check=True)
-    return proc.stdout
+# Shared gh shim (main#1119). Still raises the raw CalledProcessError/OSError
+# that `_GH_ERRORS` above catches — see gh.py § Error behaviour.
+_run_gh = gh.run_gh
 
 
 def _utcnow() -> datetime:
--- a/.claude/lib/tests/test_wave_state.py
+++ b/.claude/lib/tests/test_wave_state.py
@@ -271,6 +271,56 @@ class CacheInvalidationOnRewrite(_CacheIsolated):
                 )
             self.assertGreater(wave_state.cache_info().uncacheable, 0)
 
+    def test_a_write_landing_during_the_read_is_never_cached(self) -> None:
+        """The post-read re-stat, without which a torn read gets pinned.
+
+        If a write lands between the pre-read `stat()` and the end of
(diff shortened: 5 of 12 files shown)


parametrization:
**SRE review — operational blast radius (second reviewer, per the #1119 two-reviewer split)**

Verified at head `0ae9ed7` (CI 14/14 green, re-checked immediately before posting).

**M14 fix confirmed, and extended beyond her unit tests to full CLI paths.** `gh.run_gh` propagates `CalledProcessError`/`FileNotFoundError` raw — no `GhError` wrapping. I exported base (`e58ccd6`) and head into isolated trees and ran the actual gate CLIs (never against the real `cross-repo-status.json` — fixture copies only):

- `red_sweep.py sweep --repos fake-repo` against a `gh` stub that exits 1: base and head both degrade to `{"errors": ["fake-repo"], "red": []}`, rc=2, byte-identical output. The graceful-degrade path she proved at the unit level (`test_red_sweeps_error_tuple_still_matches_what_run_gh_raises`) holds end-to-end.
- `verify_deployable_merge._run_gh` against the same failing stub still raises `GhError("gh api x failed: boom")` — its own wrapping survived the consolidation.
- `gh_rest.py._run_gh` is untouched, as claimed (different contract, #1224 — not folded in).

**Failure-path base-vs-head, `wave_status digest` / `wave_seq peek` / `wave_merge_model model+reachability` / `wave_unwrapped check` / `trust_signals extract`, against missing / empty (`{}`) / malformed / real-copy (343,521-byte) status files:** every exit code matches base. One cosmetic-only divergence: on a missing file, head's traceback shows an extra frame (`wave_state.load_status`'s `path.stat()` raising before falling through to `path.read_text()`, which then raises the same `FileNotFoundError` again) — same exception class, same rc, just one more stack frame from the indirection. Not a behavior change. `--help` text for `kickoff_sweep.py`/`lifecycle.py` diffs clean, matching her note that her one `--help` divergence was an export artifact (missing `.claude/hooks`), not a real one — I hit the same `.claude/hooks` shape and confirmed it resolves once the full export is used.

Nothing fails earlier (at import/arg-parse) than before in any case I ran.

**M7 (write-during-read) — genuine hole, correctly closed.** `test_a_write_landing_during_the_read_is_never_cached` interposes a write mid-`read_text()` and proves the post-read re-stat (`key_after == key`) is load-bearing; I re-read the mutation description and it matches the code's re-stat-after-read block exactly.

**Cache key + racy-tick guard**: `(st_dev, st_ino, st_size, st_mtime_ns)` re-stat'd every call, tick-settlement inferred from sub-second bits — read through the full module docstring and the six invalidation tests (each isolating one key component); the design and its limitation (does not survive a same-tick rewrite until the tick settles) are stated accurately, not oversold.

**Writes stay text-surgical**: `test_upsert_status_keys_does_not_import_wave_state` asserts it by source-grep, not just by claim, and `upsert_status_keys.py` doesn't appear in this PR's file list.

**Ruling on the `.conclusion // .status` trap — refuted as stated, but the underlying mechanism is real.** I confirmed with `jq` that `.conclusion // .status` does NOT fall through on an empty-string `conclusion` (only `null`/`false` trigger `//`'s alternative in jq) — so a running check's `conclusion: ""` does produce `""`, not `"IN_PROGRESS"`, and a watcher keying on that would misread in-flight as terminal. That part is a real, reproducible bug shape. But I could not confirm the specific attribution: I searched this repo (`docs/`, `.claude/`, `ontology/`, excluding scratchpad/fixtures) for that idiom in any committed doc and found no documented example of it — the one committed memory on this exact topic, `.claude/memory/feedback_statuscheckrollup_ci_clean.md`, already codifies the CORRECT idiom ("poll-until-completed loop ... exit when all `status == completed`"), not `conclusion // status`. I don't have the Monitor tool's own schema loaded in this session to check its description text directly. So: the trap is real and worth keeping an eye on in ad-hoc polling scripts, but "the Monitor tool's own documented example" carries it is not something I could substantiate against this repo's committed docs — recommend not repeating that specific attribution without a citable source.

**Line-count honesty**: confirms as stated — `-82` net across converted modules, `+389`/`+700` new. Reviewed as a DRY consolidation, not a size reduction, consistent with the PR body's own framing.

No behavior regressions found in any converted call site. The `.conclusion // .status` finding is orthogonal to this PR's diff (no converted module here polls CI that way), so it doesn't attach as debt to #1323 itself.

---
Requestor: Lucas Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Merge-gate review (Opus) — approved.** Reviewed at head `0ae9ed7`, diffed three-dot against base `e58ccd6` via the contents API at that SHA, never a local checkout. CI 14/14 SUCCESS at that exact SHA via `statusCheckRollup`; `pr_ci_state.py` agrees (`ready`, 14 checks). `MERGEABLE` / `CLEAN`.

This is a consolidation that introduces a cache over a file the same process also rewrites, so I spent the review on the one thing that can actually go wrong here, and treated the rest as corroboration.

## Cache correctness — attacked, one gap found

The key `(st_dev, st_ino, st_size, st_mtime_ns)` re-`stat()`ed on every call is the right design, and the "no writer ever has to remember to invalidate" property is worth what it costs. The racy-tick guard is real and does what it claims **for the case it was written for**. I found one place where the reasoning outruns the code, filed as #1324 and **not** a blocker.

`_tick_has_settled`'s granularity inference is binary: exact second -> 1 s, any sub-second bits -> **1 ns**. Real filesystems occupy the middle. A Linux coarse-clock filesystem (ext4/xfs via `ktime_get_coarse_real_ts64`) stores a full-nanosecond value that only *advances* once per timer tick — 1 ms at `HZ=1000`, 4 ms at `HZ=250`. The inference reads that as 1 ns and the guard never fires:

```
fs tick 1 ms (HZ=1000)   -> caching permitted 0.000001 ms after the write: True  (needs 1.0 ms)
fs tick 4 ms (HZ=250)    -> caching permitted 0.000001 ms after the write: True  (needs 4.0 ms)
exact-second timestamp   -> settled 1 ns later: False   (guard DOES fire)
```

End-to-end on a simulated 4 ms jiffy — same full-ns mtime for both writes, size-preserving counter flip, real clock, no `_now_ns` patch — `load_status` returned `5` with `6` on disk (`hits=1, parses=1, uncacheable=0`). The identical sequence on a simulated 1-second filesystem returns `6` correctly.

The docstring says sub-second bits "prove the filesystem records them (granularity ~1 ns)". They prove it *records* sub-second values, not that it *advances* at 1 ns — worth correcting alongside the code, since that is the sentence the next maintainer will lean on.

Why this does not block:

- Measured granularity on this machine is **~62 us** (300 consecutive writes -> 300 distinct mtimes), so it is unreachable here.
- **200 real write-then-read cycles** through `upsert_status_keys` against a copy of the real 343 KB file: **0 stale serves**, `hits=0, parses=200` — the mtime moved every cycle, so the cache never engaged on the write-heavy path at all.
- Pre-merge the read was an unconditional re-parse, so this narrows a guarantee on filesystems we do not currently run on rather than regressing one we do.

Suggested fix is in #1324 and is cheap: a conservative floor (~10 ms) instead of an inference. Steady-state cost is nil, because the only reads pushed off the cache are those within the floor of a write — exactly the ones the guard exists to protect.

On the documentation of the limitation: it is stated where a caller sees it, not buried. The module docstring closes with it explicitly ("one parse per process holds for a file whose mtime has settled ... deliberately does NOT hold for a file rewritten during the same tick"), and `load_status`'s own docstring points at that section. That is the right placement.

## M7 — independently re-derived, genuine hole

I did not take the test's word for it. I built both loader variants standalone and drove them with a **real second process** performing the mid-read write (not the suite's mocked `read_text`):

```
variant   (first_read, second_read)   file on disk at 2nd read = 3
restat    (1, 3)
norestat  (1, 1)
=> DIVERGES: genuine hole, not an equivalent mutant
```

Exactly the divergence reported. The mutant serves `1` where the file says `3`. Re-applying the mutant to the real module fails `test_a_write_landing_during_the_read_is_never_cached` at pytest exit 1. Concurrent writers are not hypothetical in this repo, so the post-read re-stat is load-bearing. This is the equivalent-mutant discipline applied correctly, and it holds.

## M14 — confirmed, and this is the one that mattered

Wrapping gh failures in the shared error class fails `test_wave_merge_model.py::GhArgListGuardTest::test_gather_missing_branch_returns_not_exists` (full lib suite, exit 1, 5 failed vs the 1 known-spurious baseline). That test file is **byte-identical between base and head** — genuinely pre-existing, not a guard written to pass. `_GH_ERRORS = (CalledProcessError, OSError)` handlers remain live at six sites across `red_sweep.py` and `base_pin_drift_sweep.py`. Raw propagation is correct and the docstring's reasoning is accurate.

## Writes stay text-surgical — assertion is not inert

Mutation-tested rather than read:

| mutant | verdict |
|---|---|
| add `import wave_state` to `upsert_status_keys` | KILLED |
| make both `write_text(text)` sites a `json.dumps` round-trip | KILLED (both shape tests) |
| point one module's default path away from the shared constant | KILLED |
| have `build_digest` scribble on the loaded mapping | KILLED |
| drop the racy-tick guard / `st_ino` / invert the inference | KILLED |

Two of these survived my first attempt and both were **my** mis-aimed patches, not inert tests — the round-trip patch hit line 401 instead of the 480 the tests exercise, and the scribble landed in `_kickoff_ts` rather than `build_digest`. Correctly targeted, both die. The "do not finish the job" section reads as reasoning, not assertion: it gives the round-trip cost *and* the subtler point that a cached dict would turn the writer's before/after equality check into a rubber stamp. That second reason is the one worth having written down.

## Scope boundaries — verified byte-identical

`base_pin_drift_sweep.py`, `gh_rest.py`, `pr_review_state.py`, `validate_pr_review.py`, `upsert_status_keys.py`, `.claude/hooks/_shell_parse.py` — all `diff`-identical base vs head. The deferrals to #1322 / #1224 / #1316 / #1320 are real, not nominal.

## Base-vs-head

Spot-checked 9 of the 19 surfaces under both trees against a copy of the real 343 KB file: `digest` (109,039 bytes, matching), `repos`, `merge_model model`, `wave_unwrapped check`, `wave_seq peek`, three `--help` texts, and the missing-wave path. All byte-identical including exit codes. `kickoff_sweep --help` is identical once `.claude/hooks` is present, confirming the traced export artifact. My own first export reproduced the same class of mistake by omitting `scripts/`, which is a decent argument that the "export the whole tree" habit is the one to keep.

## The retraction is accurate

Instrumented `json.loads` at base: `build_digest` parses the status file **exactly once**, at base and at head. The issue's headline claim that a single `digest` call "parses it repeatedly" is false, and the correction is right to be in the body. Re-measured counts check out where I could verify them offline — 490 keys / 343,521 bytes exactly, and the −82 net across the nine converted modules reconciles against the diff (1188 additions − 389 new module − 700 test = 99 insertions, 181 deletions).

Selling this as DRY rather than speed is the correct call, and ~9 ms saved ahead of dozens of network calls would not have justified it. Re-measuring an audit taken at `3bdd620` rather than quoting it, and reporting that it was wrong in **both** directions, is the part of this PR I would most like to see repeated.

## Line count

Presented plainly enough. The body has a dedicated section giving +389 new module lines and +700 test lines against the −82, and says outright that the repo's total goes up and that only the duplication shrinks. No complaint.

## One process note, worth confirming separately

The `.conclusion // .status` trap is real as a **jq** idiom, and I verified the mechanism directly:

```
{"conclusion":"", "status":"IN_PROGRESS"} | .conclusion // .status  ->  ""          (no fallthrough)
{"conclusion":null,"status":"IN_PROGRESS"} | .conclusion // .status  ->  IN_PROGRESS
```

jq's `//` falls through on `null` and `false` only; an empty string is truthy to it. So a watcher keyed that way reads a running check as terminal — the failure observed. I could not capture a live non-terminal check on any open PR at review time, so I am reporting the jq semantics as measured and the empty-string-while-running serialization as her observation rather than as something I re-measured.

The reassuring half: this repo's sanctioned helpers are already immune because they use Python's `or`, which *is* falsy on `""` — `validate_pr_ci_status.py` does `(check.get("conclusion") or "").upper()` and keys terminality on `status == "COMPLETED"`. So the hazard is confined to ad-hoc jq watchers, and the fix is to key on `.status == "COMPLETED"` or just call `pr_ci_state.py`. Flagging it for the team rather than filing, since no committed artifact needs changing.

---
Requestor: Weronika Zielinska
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1324

Issue #1318, opened 2026-08-03
tech-debt(commit-identity): HEREDOC_INERT_RELAY_FILTERS has no adversarial-invocation gate — `sort --compress-program` executes the heredoc body


parametrization:
## Summary

PR #1316 introduces `HEREDOC_INERT_RELAY_FILTERS` and states, in the module comment and in `RealShellGroundTruthTests`, that every member was measured to have **no code-execution surface reachable from its own stdin content**. That is the allowlist's entire safety property, and it is enforced by exactly one structural assertion — `test_every_allowlisted_filter_has_a_representative_case`, which checks that the false-positive corpus *covers* the allowlist.

Nothing requires that each member be measured **in its most exec-shaped invocation**. Coverage of the ALLOW direction is gated; ground-truth inertness is not. That is the gap, and it already admitted one member that executes.

## Evidence — `sort` is in the shipped allowlist and executes the heredoc body

`sort --compress-program=CMD` runs `CMD` whenever the sort spills to a temporary file, with the input data on `CMD`'s stdin. Point it at a shell and the heredoc body is executed.

Real-shell marker proxy (a fake `git` earlier on `PATH` that appends to a log; the log is written only if a shell genuinely runs the body), against a body of one `git ... commit` line plus 20 padding lines:

```
REAL SHELL bash : RAN   marker=GIT-RAN: -c user.name=X -c user.email=Y commit -m z
REAL SHELL zsh  : RAN   marker=GIT-RAN: -c user.name=X -c user.email=Y commit -m z
HOOK MAIN            : ALLOW   is_code=[False]
HOOK PR-HEAD bf36898 : ALLOW   is_code=[False]
  body stripped from scan: True
```

The downstream segment is `sort -S1 --compress-program=/bin/sh`. The head resolves to `sort`, `sort` is allowlisted, so the walk treats it as inert, the span is classified DATA, and `strip_data_heredocs` removes the body before `validate_commit_identity` ever scans it.

Spill threshold with `-S 1` is ~9 body lines — padding is attacker-controlled, so this is not a large-input-only shape:

```
padding lines = 5 -> did-not-run
padding lines = 8 -> RAN
```

This is **not a regression** — `main` allows it too, because `main` resolves every unrecognised downstream head to DATA. #1316 strictly improves on that. The problem is that the new allowlist re-admits this one instance while closing the rest, and asserts in code that it did not.

## Why this is the same class #1316 deliberately excluded `sed`/`awk` for

[...]

Issue #1317, opened 2026-08-03
tech-debt(commit-identity): HEREDOC_INERT_RELAY_FILTERS allowlist is incomplete for common deploy/CI tools (yq, envsubst, kubectl, psql, sqlite3, gh --body-file)


parametrization:
## Context

main#1168 (PR #1316) inverted `_opener_feeds_interpreter`'s downstream-relay default from DATA to CODE: an unrecognised segment head reached through a pipe from a heredoc's owning data-sink segment now resolves to CODE unless it is a member of the new `HEREDOC_INERT_RELAY_FILTERS` allowlist. That fix closes a real bypass (`xargs -I{} bash -c "{}"` and incidentally #1171's backslash-continuation relay — both verified in review with a real-shell marker proxy).

The flip is correct and the allowlist that shipped (`grep`/`wc`/`sort`/`uniq`/`head`/`tail`/`cut`/`tr`/`column`/`nl`/`rev`/`fold`/`expand`/`unexpand`/`base64`/`md5sum`/`sha1sum`/`sha256sum`/`sha512sum`/`cksum`/`od`/`xxd`/`hexdump`/`join`/`paste`/`comm`/`tac`/`shuf`/`jq`) was built against real-shell ground truth per-entry, which is the right bar. But the allowlist is necessarily incomplete for shapes the PR's corpus didn't consider, and every unlisted tool now makes `validate_commit_identity`'s heredoc-body scan run on a body it would previously have skipped as data — which only actually blocks a commit if that body ALSO happens to contain identity-pattern text (a `git commit` example, a `--no-verify` mention, etc.), so this is a narrow but real false-positive surface, not a broad one.

## What review (PR #1316, second reviewer) measured

Built a marker-proxy corpus (real bash + zsh, `echo MARKER_RAN >> marker.log` inside the heredoc body) against the deploy/CI tool shapes the review brief asked about, then ran the SAME commands through the shipped `validate_commit_identity.check()` at the PR's head SHA (`bf368980aacbb360ad9f1573ccd225b3aa3d8edf`):

| Downstream tool | Real shell runs the body? | Hook classifies | Note |
|---|---|---|---|
| `envsubst` | No (confirmed, bash + zsh) | CODE (new FP) | Pure text substitution, no argument-driven shell-out surface found |
[...]

Issue #1314, opened 2026-08-03
tech-debt(commit-identity): assignment pre-pass misses a subshell whose `(` is glued to the assignment — one space separates block from allow


parametrization:
## Shape

#1311 (folded into #1302) routes each segment's tokens through `strip_command_prefixes()` so a compound-statement leader no longer hides the assignment behind it. `(` is in `_COMPOUND_LEADERS` and the fix works when the parenthesis is a separate token. It is **not** a separate token when written without a space, which is the ordinary spelling.

`resolve_simple_assignments()` is a string/shlex pre-pass, so `(g=git` tokenizes as ONE token: `strip_command_prefixes` does not recognise it as `(`, and `_ENV_ASSIGN_RE` does not match it as an assignment because it starts with `(`. The assignment is silently never captured.

## Measured at PR #1302 head `f4c965b67ee503b91f62a94b7e4061fcf67f41a8`

| command | hook | real shell (marker proxy) |
|---|---|---|
| `( g=git; $g commit -m x )` — spaced | **block** | runs it — correct |
| `(g=git; $g commit -m x)` — glued | **allow** | **runs it** — marker written |
| `(gg=git; $gg commit -m x)` | **allow** | runs it |
| `(export g=git; $g commit -m x)` | **allow** | runs it |
| `((g=git; $g commit -m x))` | **allow** | runs it |

One deleted space is the difference between block and allow, on a shape the same PR just closed for every other compound leader.

## Not a regression, and the direct path is unaffected

`(git commit -m x)` written glued **blocks** on base `main` and at every head of #1302 — the direct commit finder reaches it through the bashlex AST, which parses the parenthesis structurally. The gap is specific to the assignment resolver, which is a shlex-token pre-pass and therefore sees `(g=git` as an opaque word. The variable-held form has been allowed on `main` and at every prior head, so #1302 does not regress it; it closes the spaced spelling and leaves the glued one.

`do`, `then`, `else`, and `{` cannot produce this gap — shell syntax requires whitespace after them, so they are always separate tokens. `(` is the only compound leader that can be written adjacent to what follows.

## Suggested direction

Two options, neither free:

1. **Split a leading `(` off the token** in `_iter_original_segment_spans` or before `_leading_literal_assignments` — smallest change, but it adds a fourth notion of "where does a segment start" to a module that has spent three review rounds consolidating exactly that (this is the `main#1150` hazard the same PR cites repeatedly).
[...]

PR #1325, opened 2026-08-03
fix(commit-identity): resolve stdin-redirect script operand in heredoc write-then-exec correlation

### LINKED ISSUE #1170: tech-debt(commit-identity): heredoc piped to an interpreter through a FIFO created in the same command bypasses the classifier

### LINKED ISSUE #1287: tech-debt(commit-identity): exec-side operand resolution — bash < FILE, $(...)-produced paths, and copy indirection escape the write-then-exec correlation
## Summary

PR #1280 (#1167) correlates a heredoc's **write target** against the **script operand** of a later interpreter invocation. The operand side reads `InterpreterInvocation.operands[0]` — a literal token. Three shapes reach the same file without ever putting its path in that position, and one reaches it without a positional operand at all. All measured **ALLOW** through the real `check()` at PR #1280 head `193b87b`, and all confirmed to execute under a real `bash`.

## The shapes

    # 1 — stdin redirect instead of a positional operand
    cat > /tmp/s.txt <<'D'
    <the identity-flagged commit line>
    D
    bash < /tmp/s.txt

    # 2 — command substitution produces the path
    ... D
    bash $(echo /tmp/s.txt)

    # 3 — copy indirection: written here, executed there
    cat > /tmp/a.txt <<'D'
    <the identity-flagged commit line>
    D
    cp /tmp/a.txt /tmp/s.txt; bash /tmp/s.txt

Shape 1 is **documented as out of scope** in the module comment the PR adds, which is the right call for that PR's blast radius — but nothing tracks it, so this issue is where it lives. Shapes 2 and 3 are not named anywhere.

## Notes on each

**Shape 1 (`bash < FILE`).** The body reaches the same interpreter through fd 0 rather than through `argv`. `_script_invocation_targets` reads `invocation.operands`, which is empty here, and the pre-existing shape-7 walker in `validate_commit_identity` does not resolve stdin redirects either. Symmetric with Gap 1 of the write-side issue: the module models one spelling of "this file becomes this process's input" and misses the other.

**Shape 2 (`$(...)`).** shlex keeps `$(echo` and `/tmp/s.txt)` as separate tokens, so `operands[0]` is `$(echo` — a token that matches no write target. This is the same class as the documented variable-rename punt (the hook cannot evaluate the shell), so it may be acceptable as a permanent non-goal; filing it so that is an explicit decision rather than an accident.

**Shape 3 (`cp` indirection).** Neither side is wrong on its own: the write target really is `/tmp/a.txt` and the invocation target really is `/tmp/s.txt`. Closing it means tracking a copy/move/link edge between the two, which is a materially larger model than path equality. Likely a "won't fix, documented" outcome — but worth stating rather than leaving as a silent hole, because `cp`/`mv`/`ln -s`/`install` are ordinary commands, not evasion-shaped ones.

## Explicitly NOT this issue

`xargs bash` and `find -exec bash {}` also bypass in write-then-exec form (measured ALLOW at `193b87b`, confirmed executing). Those are the **relay** class already tracked by **#1168**, which names `find -exec` in the same family — evidence added there as a comment rather than duplicated here.

## Why this is filed rather than requested on the PR

All shapes are pre-existing: ALLOW on `main` as well, since `main` has no correlation at all. PR #1280 does not regress them.

## Suggested direction

- Shape 1 is the tractable one and the most defensible to fix: teach the operand side to also read a `< FILE` stdin redirect on an interpreter segment. `_scan_command_line` already has the redirect model needed to spot it.
- Shapes 2 and 3 most likely want an explicit, documented non-goal in the module comment alongside the variable-rename punt, rather than code.

## Related

- **#1167** / PR **#1280** — the mechanism these gaps live inside.
- **#1281** — `source` / `.` builtin, the other exec-side gap.
- **#1168** — relay class (`xargs`, `find -exec`), where the relay evidence went.
- **#1150** — umbrella.

## Provenance

Found by the merge-gate review of PR #1280 (Nino Kavtaradze, Security Engineer).

parametrization:
## Summary

Closes #1170.

Both measured shapes in #1170 (mkfifo + backgrounded reader, one written via `cat`, one via `tee`) and #1287 shape 1 (the plain-file form of the same gap, no FIFO) share one root cause: `parse_interpreter_invocation` only ever read an interpreter's script path from a positional argument. For `bash < FILE`, the redirect operator token itself landed in `operands[0]` instead of the file path, so `_script_invocation_targets` (main#1167's write-then-exec correlation) never saw `FILE` and couldn't connect it to a heredoc written to that same path elsewhere in the command.

## How much of this was #1287?

Almost all of it. #1287 named the gap explicitly as its "shape 1" and suggested the exact fix ("teach the operand side to also read a `< FILE` stdin redirect on an interpreter segment") — I applied that suggestion. Measured: implementing it closes both #1170 shapes AND #1287 shape 1 as the identical fix (same function, same slot). #1287 shapes 2 (`$(...)`-produced path) and 3 (`cp` copy indirection) are untouched by this change and remain open in that issue — this PR does not close #1287, only #1170.

## The fix

In `parse_interpreter_invocation` (`.claude/hooks/_shell_parse.py`): a bare `<` token (spaced form only, matching this module's existing `>`/`>>` convention) followed by a word is folded into `operands[0]` — shell-accurately, since the redirect target IS the script the interpreter executes; any other bare word in the operand region becomes a script argument (`$1`, `$2`, …), never a competing script. Last redirect wins, matching real-shell last-redirect-wins semantics. A fused process-substitution token (`<(cmd)`, no internal whitespace) can never be mistaken for a bare `<` since it's a distinct shlex token.

No new correlation machinery — this reuses the exact `_script_invocation_targets`/`_segment_write_targets` correlation main#1167 already built; only the operand-resolution primitive underneath it changes.

## Real-shell marker-proxy verification (before / after)

A fake `git` on `PATH` (never the real `git`) logs its own invocation to a file; the marker fires under both `bash` and `zsh` for both #1170 shapes, confirming the body genuinely executes (matching the issue's own claim):

| shape | bash marker | zsh marker |
|---|---|---|
| `mkfifo p; bash < p &` + heredoc via `cat > p` | fired | fired |
| `mkfifo p2; sh < p2 &` + heredoc via `tee p2` | fired | fired |

Classifier-level before/after (`classify_heredocs(...)`, `is_code`):

| shape | before | after |
|---|---|---|
| FIFO shape 1 (`cat` writer, `bash` reader) | `is_code=False` (DATA) | `is_code=True` (CODE) |
| FIFO shape 2 (`tee` writer, `sh` reader) | `is_code=False` (DATA) | `is_code=True` (CODE) |

Through the real CLI entry point (`validate_commit_identity.check()`), both shapes went BLOCK -> ALLOW at main#1155 (the original regression) and are restored to BLOCK by this PR.

## False-positive corpus (must stay ALLOW)

- `bash < FILE` where FILE is never written by a heredoc in the same command (the everyday `bash < deploy.sh` / `mysql < backup.sql` pattern)
- No heredoc at all, just an ordinary `bash < existing-script.sh`
- A non-`SHELL_INTERPRETERS` program (`mysql`) reading via `<` from a heredoc-written file
- `bash <(cmd)` (process substitution) — including the case where the following operand happens to equal the heredoc's write target (real-shell semantics: that operand becomes `$1` to the substituted script, not the interpreter's own script)
- `bash -c '...' < FILE` (a `-c` command-string form) — unaffected, this branch is already skipped by the correlation regardless of the stdin-redirect change
- The pre-existing #1152 false positive (`bash build.sh && cat > notes.md <<'EOF' ...`)

All of the above are asserted in `test_heredoc_write_then_exec.py::StdinRedirectOperandTests`.

## Mutation testing (real output)

Hand-mutated the new branch in `parse_interpreter_invocation`, ran the new test class each time, reverted:

| Mutant | Result |
|---|---|
| 1. Revert the whole fix (old `operands = tuple(rest[end:])`) | 13 tests failed |
| 2. First-redirect-wins instead of last | 2 tests failed (both dedicated last-wins tests) |
| 3. `tok.startswith("<")` instead of exact `tok == "<"` | Initially survived with a single-token process-substitution test (the bounds guard made it moot); added a two-token fused-`<(...)`-token test and a hook-level test where the following operand equals the write target — both then failed correctly, 2/2 |
| 4. Dropped the `j + 1 < n` bounds guard | 1 test failed — raised `IndexError` instead of degrading gracefully on a trailing bare `<` |

No surviving mutants after the mutation-driven test additions.

## Not covered (filed as follow-up, no wave label)

- #1287 shapes 2 (`$(...)`-produced path) and 3 (`cp` copy indirection) — unaffected, stay filed there.
- #1306(c), #1308, #1314, #1315, #1317, #1318 — untouched by this PR, remain filed under the #1150 umbrella.

## Verification

- `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/ -q` — 2696 passed, 757 subtests passed, 0 failed.
- `ruff format --check`, `ruff check`, `mypy` — all clean.
- `python3 .claude/lib/pre_commit_ci_sync.py .` — OK, pre-commit mirrors CI.
- Rebased on `origin/main` (`bcb4cf5`) before opening this PR, clean rebase, full suite re-run green after.

Do NOT merge — awaiting review per team process.

Claude-Session: https://claude.ai/code/session_0127s4SKmYjUz2iPiwSkxa7k

### DIFF
--- a/.claude/hooks/_shell_parse.py
+++ b/.claude/hooks/_shell_parse.py
@@ -1018,9 +1018,20 @@ def _opener_feeds_interpreter(line: str, pos: int, scan: _LineScan) -> bool:
 #     bash "$F"`) or any other divergent spelling defeats it exactly as
 #     variable resolution in general would.
 #   - `bash < FILE` (script fed via stdin redirect rather than a positional
-#     argument) is a different execution mechanism than the positional
-#     `InterpreterInvocation.operands` this correlation reads, and is not
-#     resolved by the pre-existing shape-7 walker either — out of scope here.
+#     argument) IS now covered (main#1170): `parse_interpreter_invocation`
+#     folds a bare `< FILE` redirect into `operands[0]`, the same slot this
+#     correlation already reads via `_script_invocation_targets`, so a heredoc
+#     written to FILE and later fed to an interpreter through `< FILE` (a
+#     `mkfifo`-relayed FIFO, in main#1170's shape, or an ordinary regular
+#     file) reclassifies as CODE without any new correlation machinery. This
+#     resolves main#1287 shape 1 as a byproduct — its shapes 2 (`$(...)`
+#     -produced path) and 3 (`cp` copy indirection) remain unresolved and stay
+#     filed there. The pre-existing shape-7 walker in `validate_commit_identity`
+#     (which reads a script's CONTENT off disk, not this module's path
+#     correlation) is unaffected by this change either way: it fires
+#     PreToolUse, before a same-command heredoc write has run, so the file it
+#     would read never has the relevant content yet regardless of which
+#     operand slot names it.
 #   - The attached-operator redirect form with no surrounding whitespace
 #     (`cat>/tmp/x`, as opposed to `cat > /tmp/x`) is not recognised by
 #     `_segment_write_targets`'s tokenizer, which relies on shlex already
@@ -2096,7 +2107,10 @@ class InterpreterInvocation(NamedTuple):
     `operands` are the tokens after the option run, i.e. what the shell itself
     treats as `<command-string> [$0 $1 ...]` or `<script> [args]`. This is the
     shell-accurate answer and is what a consumer that must resolve a real path
-    (the script-invocation check) should use.
+    (the script-invocation check) should use. `operands[0]` also resolves a
+    stdin-redirect script (`bash < FILE`, main#1170/main#1287 shape 1) to
+    `FILE` — see `parse_interpreter_invocation`'s docstring for why that slot,
+    not a separate field, is the shell-accurate place for it.
 
     `words` is every token of the invocation except the `--` sentinel — the set
     the command string is guaranteed to be a member of. It is deliberately a
@@ -2177,6 +2191,23 @@ def parse_interpreter_invocation(segment: list[str]) -> InterpreterInvocation | 
     Returns None when the segment is empty or its head is not one of
     `SHELL_INTERPRETERS` — callers must treat None as "not an interpreter
     invocation", never as "safe".
+
+    Stdin-redirect operand resolution (main#1170 / main#1287 shape 1):
+    `bash < FILE` feeds FILE to the interpreter as its script through the
+    process's own stdin rather than as a positional argument, but it answers
+    the exact same question `operands[0]` exists to answer — "what file does
+    this interpreter execute?" — so the redirect target is folded into
+    `operands` in the `[0]` slot, shell-accurately: any OTHER bare word in the
+    operand region becomes `$1`/`$2`/… passed TO that script, not a second
+    script, so it is never in the running for the `[0]` slot once a stdin
+    target is found. Only the spaced form (`< FILE`, not `<FILE`) is
+    recognised — the same convention `_segment_write_targets` already applies
+    to `>`/`>>` — and only a BARE `<` token qualifies: `<<` (a heredoc opener)
+    is never present here (callers strip it before tokenizing — see
+    `_HEREDOC_OPENER_RE`), and `<(` (process substitution) arrives from shlex
+    fused into one token (`<(cmd)`) whenever it isn't preceded by whitespace,
+    so it can't be mistaken for a bare `<`. The LAST `< FILE` on the line
+    wins, mirroring a real shell's last-redirect-wins semantics.
     """
     tokens = strip_command_prefixes(segment)
     if not tokens:
@@ -2191,7 +2222,22 @@ def parse_interpreter_invocation(segment: list[str]) -> InterpreterInvocation | 
 
     end = _consume_wrapper_options(rest, SHELL_VALUE_OPTIONS, 0)
     has_command_string = _COMMAND_STRING_FLAG in rest[:end]
-    operands = tuple(rest[end:])
+
+    stdin_target: str | None = None
+    other_operands: list[str] = []
+    j = end
+    n = len(rest)
+    while j < n:
+        tok = rest[j]
+        if tok == "<" and j + 1 < n:
+            stdin_target = rest[j + 1]
+            j += 2
+            continue
+        other_operands.append(tok)
+        j += 1
+    operands = (
+        (stdin_target, *other_operands) if stdin_target is not None else tuple(other_operands)
+    )
     words = tuple(t for t in rest if t != "--")
     return InterpreterInvocation(name, has_command_string, operands, words)
 
--- a/.claude/hooks/_shell_parse.py
+++ b/.claude/hooks/_shell_parse.py
@@ -2095,6 +2095,23 @@ _SHELL_VALUE_LETTERS = frozenset("oO")
 # The letter that introduces a command string: `sh -c '<cmd>'`.
 _COMMAND_STRING_FLAG = "-c"
 
+# A bare stdin-redirect operator, fd-0-targeting spellings only (main#1170,
+# widened for main#1326): `<`, `0<`, `<>`, `0<>`. All four are the identical
+# redirect as far as a shell's own script-source resolution is concerned — fd
+# 0 is stdin's default, `<>` (read-write open) still opens the SAME fd, and
+# the explicit `0` prefix is fused to the operator by shlex whenever no space
+# separates them (`0< FILE` -> one token `"0<"`, never `"0"` + `"<"`). All
+# four spellings were verified with a real-shell marker proxy (fake `git` on
+# PATH) under both bash and zsh: `bash 0< s.sh`, `bash <> s.sh`, and
+# `bash 0<> s.sh` all genuinely execute `s.sh`'s body, exactly like the bare
+# `bash < s.sh` main#1170 already covers. A DIFFERENT fd number is a
+# different redirect entirely and must NOT match — `bash 2< s.sh` does not
+# feed `s.sh` to the interpreter as its script (confirmed: the marker never
+# fires), so the leading digit run is restricted to the literal `0`, not
+# `\d*` as #1326's own suggested pattern had it (that pattern would also
+# wrongly swallow `2<`/`9<`/... as if they were stdin sources).
+_STDIN_REDIRECT_RE = re.compile(r"^0?<>?$")
+
 
 class InterpreterInvocation(NamedTuple):
     """A decoded `<shell> [options] [operands ...]` command.
@@ -2108,9 +2125,20 @@ class InterpreterInvocation(NamedTuple):
     treats as `<command-string> [$0 $1 ...]` or `<script> [args]`. This is the
     shell-accurate answer and is what a consumer that must resolve a real path
     (the script-invocation check) should use. `operands[0]` also resolves a
-    stdin-redirect script (`bash < FILE`, main#1170/main#1287 shape 1) to
-    `FILE` — see `parse_interpreter_invocation`'s docstring for why that slot,
-    not a separate field, is the shell-accurate place for it.
+    stdin-redirect script (`bash < FILE`, main#1170/main#1287 shape 1, widened
+    to the `0<`/`<>`/`0<>` spellings by main#1326) to `FILE` — but ONLY when
+    NO other (non-redirect) operand is present anywhere in the invocation. A
+    real shell always prefers an ordinary positional operand over a stdin
+    redirect as its script source, regardless of whether that operand sits
+    before or after the redirect token in the invocation — a stdin redirect
+    is not counted as an argument slot at all, so `bash script.sh < FILE` runs
+    `script.sh` (FILE is just its stdin stream) and `bash < FILE arg1` runs
+    `arg1` (FILE is again just stdin; `arg1` is the only real operand) — real
+    bash and zsh confirmed via marker proxy for both orderings. Only the `-s`
+    option (not currently modeled here — main#1325 review, filed as a
+    follow-up) forces the stdin content itself to be the script even when a
+    positional operand is also present. See `parse_interpreter_invocation`'s
+    docstring for the resolution rule.
 
     `words` is every token of the invocation except the `--` sentinel — the set
     the command string is guaranteed to be a member of. It is deliberately a
@@ -2192,22 +2220,58 @@ def parse_interpreter_invocation(segment: list[str]) -> InterpreterInvocation | 
     `SHELL_INTERPRETERS` — callers must treat None as "not an interpreter
     invocation", never as "safe".
 
-    Stdin-redirect operand resolution (main#1170 / main#1287 shape 1):
-    `bash < FILE` feeds FILE to the interpreter as its script through the
-    process's own stdin rather than as a positional argument, but it answers
-    the exact same question `operands[0]` exists to answer — "what file does
-    this interpreter execute?" — so the redirect target is folded into
-    `operands` in the `[0]` slot, shell-accurately: any OTHER bare word in the
-    operand region becomes `$1`/`$2`/… passed TO that script, not a second
-    script, so it is never in the running for the `[0]` slot once a stdin
-    target is found. Only the spaced form (`< FILE`, not `<FILE`) is
+    Stdin-redirect operand resolution (main#1170 / main#1287 shape 1, widened
+    for main#1326): `bash < FILE` feeds FILE to the interpreter as its script
+    through the process's own stdin rather than as a positional argument, but
+    it answers the exact same question `operands[0]` exists to answer —
+    "what file does this interpreter execute?" — so the redirect target CAN
+    be folded into `operands` in the `[0]` slot.
+
+    Precedence — a positional operand always wins, regardless of position
+    relative to the redirect (main#1325 review round 2, corrected from this
+    function's own first cut): a real shell does not count a stdin redirect
+    as an argument slot at all, so it never competes with an ordinary word for
+    the `[0]` position. Confirmed with a real-shell marker proxy under both
+    bash and zsh, in both orderings:
+
+      - `bash script.sh < FILE` runs `script.sh` — FILE is only script.sh's
+        stdin stream, never a competing script. (This function's FIRST cut
+        got this backwards: it unconditionally promoted the redirect target
+        into `[0]` even when a positional operand was already present,
+        which silently swapped which file gets read for content inspection
+        and which file the write-then-exec correlation watches — a measured
+        BLOCK->ALLOW bypass, a measured ALLOW->BLOCK false positive on the
+        ordinary `bash migrate.sh < input.csv` shape, a BLOCK->ALLOW
+        regression in the on-disk script-content walker for `bash -x s.sh <
+        FILE` / `bash -- s.sh < FILE`, and an over-block reintroducing
+        main#1152's own failure mode for a benign `bash deploy.sh < data.txt`
+        stdin feed. All four were the SAME defect, not four separate ones.)
+      - `bash < FILE arg1` runs `arg1` (which will typically fail to exist,
+        exactly like a real shell) — FILE is still only stdin; `arg1` is the
+        one true operand present, regardless of appearing textually AFTER the
+        redirect token.
+      - `bash < FILE` (no other operand at all) is the only shape where FILE
+        genuinely becomes the executed script — this is the ONLY case the
+        redirect target is promoted into `operands[0]`.
+
+    Only the `-s` option (not modeled here — filed as a follow-up in the
+    main#1325 review) can force the redirect target to be the script even
+    when a positional operand is also present; this function does not special
+    case it, matching its behaviour before this correction.
+
+    Redirect spelling: only the spaced form (`< FILE`, not `<FILE`) is
     recognised — the same convention `_segment_write_targets` already applies
-    to `>`/`>>` — and only a BARE `<` token qualifies: `<<` (a heredoc opener)
-    is never present here (callers strip it before tokenizing — see
+    to `>`/`>>`. The token itself must match `_STDIN_REDIRECT_RE` (`<`, `0<`,
+    `<>`, `0<>` — all four verified to genuinely feed the interpreter's
+    script under both bash and zsh; a different fd number, e.g. `2<`, does
+    NOT and is deliberately excluded). `<<` (a heredoc opener) is never
+    present here (callers strip it before tokenizing — see
     `_HEREDOC_OPENER_RE`), and `<(` (process substitution) arrives from shlex
     fused into one token (`<(cmd)`) whenever it isn't preceded by whitespace,
-    so it can't be mistaken for a bare `<`. The LAST `< FILE` on the line
-    wins, mirroring a real shell's last-redirect-wins semantics.
+    so it can't be mistaken for a bare `<`. The LAST matching redirect token
+    on the line wins for STDIN-TARGET tracking, mirroring a real shell's
+    last-redirect-wins semantics — moot whenever a positional operand is also
+    present, since the redirect target is dropped entirely in that case.
     """
     tokens = strip_command_prefixes(segment)
     if not tokens:
@@ -2229,14 +2293,20 @@ def parse_interpreter_invocation(segment: list[str]) -> InterpreterInvocation | 
     n = len(rest)
     while j < n:
         tok = rest[j]
-        if tok == "<" and j + 1 < n:
+        if _STDIN_REDIRECT_RE.match(tok) and j + 1 < n:
             stdin_target = rest[j + 1]
             j += 2
             continue
         other_operands.append(tok)
         j += 1
+    # A positional operand always wins over a stdin redirect — see the
+    # docstring's "Precedence" section. The redirect target is only promoted
+    # into operands[0] when it is the SOLE operand candidate; otherwise it is
+    # dropped entirely (it names a stdin stream, not an argument or a script).
     operands = (
-        (stdin_target, *other_operands) if stdin_target is not None else tuple(other_operands)
+        tuple(other_operands)
+        if other_operands
+        else ((stdin_target,) if stdin_target is not None else ())
     )
     words = tuple(t for t in rest if t != "--")
     return InterpreterInvocation(name, has_command_string, operands, words)
--- a/.claude/hooks/tests/test_heredoc_relay_downstream.py
+++ b/.claude/hooks/tests/test_heredoc_relay_downstream.py
@@ -767,19 +767,16 @@ class SiblingIssueMeasurementTests(unittest.TestCase):
         cmd = "cat > \\\nnotes.md <<'EOF'\nsome prose about bash -c 'foo'\nEOF"
         _assert_allowed(self, cmd)
 
-    def test_1170_fifo_relay_measured_still_open(self):
-        """main#1170: `mkfifo p; bash < p & cat <<'D' > p` — a structurally
-        different gap (dataflow through a named pipe across a `&`
-        background-job boundary, not a `|`-connected pipeline segment this
-        fix's walk ever reaches). Measured to remain ALLOW after this fix;
-        NOT folded in per the spawn brief's explicit scope boundary. This
-        test pins today's measurement and is expected to start FAILING (in
-        the good direction) once #1170 lands its own fix — at which point it
-        should be deleted, not "fixed" to assert BLOCK, since that fix
-        belongs to #1170's own PR.
-        """
-        cmd = "mkfifo p; bash < p & cat <<'D' > p\n" + REAL_COMMIT + "\nD"
-        _assert_allowed(self, cmd)
+    # test_1170_fifo_relay_measured_still_open removed: main#1170 landed its
+    # own fix (stdin-redirect operand resolution in
+    # `parse_interpreter_invocation`/`_script_invocation_targets`, see
+    # `_shell_parse.py`'s "Explicit scope boundary" comment above
+    # `_segment_write_targets`). Both FIFO shapes from that issue now BLOCK;
+    # coverage moved to `test_heredoc_write_then_exec.py`'s
+    # `StdinRedirectOperandTests` (main#1170 reuses that same write-then-exec
+    # correlation, just with the interpreter's script fed via `< FILE` rather
+    # than a positional operand), per this test's own docstring instruction
+    # to delete rather than flip the assertion here.
 
 
 if __name__ == "__main__":  # pragma: no cover
--- a/.claude/hooks/tests/test_heredoc_write_then_exec.py
+++ b/.claude/hooks/tests/test_heredoc_write_then_exec.py
@@ -439,5 +439,335 @@ class StdinRedirectOperandTests(unittest.TestCase):
             self.assertEqual(result["decision"], "block")
 
 
+class PositionalOperandPrecedenceTests(unittest.TestCase):
+    """main#1325 review round 2 (Lucas Ferreira) / main#1325 merge gate (Nino
+    Kavtaradze) — the corpus above varied redirect-presence and
+    operand-presence ONE AT A TIME and never crossed them, which is exactly
+    where the original fix's defect lived: `parse_interpreter_invocation`
+    unconditionally promoted the LAST stdin-redirect target into
+    `operands[0]`, even when a positional operand was ALSO present. Real bash
+    and zsh never let a stdin redirect outrank a positional operand — a
+    redirect is not counted as an argument slot at all, so it can never win
+    against an ordinary word regardless of which side of the redirect token
+    that word sits on. Verified with a real-shell marker proxy, both shells,
+    every ordering below (see `parse_interpreter_invocation`'s docstring
+    "Precedence" section for the full account).
+
+    Crossed axes: {no operand, positional operand} x {no redirect, one
+    redirect, two redirects} x {redirect spelling: `<`, `0<`, `<>`, `0<>`} x
+    {redirect position: before/after/interleaved with the operand} x {-c
+    present, absent} x {shell interpreter, non-shell interpreter}. Every row
+    here is unit-level (`parse_interpreter_invocation` directly, verified
+    against real-shell measurements above); the operationally-significant
+    shapes are additionally pinned through the real `hook.check()` entry
+    point in `WriteThenExecPositionalPrecedenceHookTests` below, each with a
+    positive/negative pair sharing one instrument per #1318.
+    """
+
+    # --- Axis: no operand present --------------------------------------
+
+    def test_no_operand_no_redirect(self):
+        inv = parse_interpreter_invocation(["bash"])
+        self.assertEqual(inv.operands, ())
+
+    def test_no_operand_one_redirect_bare(self):
+        inv = parse_interpreter_invocation(["bash", "<", "F"])
+        self.assertEqual(inv.operands, ("F",))
+
+    def test_no_operand_one_redirect_fd0(self):
+        """main#1326: `0<` is the identical redirect as bare `<` (fd 0 is
+        stdin's default) — real-shell-verified under bash and zsh."""
+        inv = parse_interpreter_invocation(["bash", "0<", "F"])
+        self.assertEqual(inv.operands, ("F",))
+
+    def test_no_operand_one_redirect_readwrite(self):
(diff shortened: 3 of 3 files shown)


parametrization:
Second review, checked against head `5b89d0c6fe0d9142bace8c033fea8fa9134221e3`. CI is green there (`pr_ci_state.py` reports `ready`, 14/14 checks SUCCESS).

## The fix reorders `operands` in a way that contradicts real-shell precedence, and it's exploitable

`parse_interpreter_invocation`'s new folding logic always promotes the LAST bare `<` redirect target into `operands[0]`, unconditionally — even when a positional script operand is also present in the same invocation. Real bash and zsh do not work that way: when an interpreter invocation carries BOTH a positional script argument and a `< file` stdin redirect, the positional argument is what executes; the redirect only changes the invoked script's own stdin stream. I verified this against a real shell with a marker-proxy `git` on `PATH` (never the real `git`) — `bash script.sh < unrelated.txt` fires the marker (i.e. genuinely executes `script.sh`'s content) under both `bash` and `zsh`, and `unrelated.txt`'s content never runs.

`operands[0]` is read directly in two places this change touches: `_script_invocation_targets` (`_shell_parse.py`, the write-then-exec correlation this PR's own fix reuses) and `_read_script_if_safe(invocation.operands[0], cwd)` in `validate_commit_identity.py`. Both now resolve to the wrong file whenever a positional operand coexists with a stdin redirect on the same invocation.

**Confirmed regression, before/after the fix (`hook.check()`, tokenized walker path):**

| Shape | before this PR | after this PR |
|---|---|---|
| A heredoc writes a hidden commit into `script.sh`, then `bash script.sh < unrelated.txt` runs (positional script executed, unrelated stdin) | `block` | **`allow` — new bypass** |
| A heredoc writes into `data.txt` (never executed — just fed as stdin), then `bash preexisting.sh < data.txt` runs (an ordinary pattern, e.g. `bash migrate.sh < input.csv`) | `allow` | **`block` — new false positive** |

Row 1 is a genuine BLOCK→ALLOW regression on a shape that is, if anything, MORE ordinary than the FIFO relay this PR targets — a positional script argument together with an incidental trailing redirect. It regressed specifically because the old code (`operands = tuple(rest[end:])`) happened to leave `operands[0]` as the positional script when one came first; the new folding logic overrides that with the stdin target regardless. I re-verified execution at the shell level for this exact combination (not just the classifier) with the same marker-proxy method the PR description itself uses, under both `bash` and `zsh` — the marker fires, confirming `script.sh` really does execute and the hidden commit really does run.

Row 2 is the over-blocking case flagged in the review brief — a heredoc-written data file later fed as stdin to an unrelated, already-existing script is a completely ordinary ops pattern (`bash deploy.sh < config.env`, `bash migrate.sh < input.csv`) and would now block in production.

The `False-positive corpus` section of the PR description, and the new `StdinRedirectOperandTests` class, cover `bash < FILE` (no positional operand at all — that's #1287 shape 1, and it does close correctly) and plain `bash script.sh` (no redirect at all), but nothing exercises a positional operand **and** a stdin redirect together on the same invocation — which is exactly the untested combination where the regression lives.

**Suggested direction** (not prescriptive): a positional operand, when present, should always win `operands[0]` — matching real-shell precedence — with the stdin target only filling that slot when no other operand exists. The current unconditional "stdin target always goes first" folding gets the precedence backwards.

### Root-cause framing and repo-wide sweep

Because `parse_interpreter_invocation` is shared, I swept the corpus the review brief asked for (`bash < deploy.sh`/`sh < setup.sh` with no heredoc at all, `psql`/`mysql`-style non-shell-interpreter reads, `bash -c '...' < FILE` precedence, `docker exec -i c sh < script` / `ssh host bash < local.sh` relay forms, `bash < FILE arg1 arg2`) through the same before/after harness. All of those come back correct and unaffected — `bash -c` precedence, non-`SHELL_INTERPRETERS` reads, and the plain no-heredoc stdin-feed shape are all still `allow` as they should be, and `bash < FILE arg1 arg2` (no positional operand, just the redirect plus trailing args) correctly starts blocking, matching the PR's own intended scope. The `docker exec`/`ssh` relay forms are unaffected either way (pre-existing gap — `docker`/`ssh` aren't `SHELL_INTERPRETERS` heads, so this PR's change doesn't reach them; not a new regression). The regression is narrowly the "positional operand + stdin redirect coexist" combination above, but that combination is common enough (any script invoked with both a path and an unrelated `< file`) that I can't call this a corner case.

### Two things confirmed as asked

- **The deleted pin test is genuinely superseded**, not a coverage loss, for the two FIFO shapes it covered. Both `mkfifo`+`bash`/`cat` and `mkfifo`+`sh`/`tee` shapes from the removed test are reproduced verbatim in the new `StdinRedirectOperandTests` (`test_issue_repro_mkfifo_bash_stdin`, `test_issue_repro_mkfifo_sh_stdin_via_tee`) and both correctly assert BLOCK against the PR's own fix — confirmed via my before/after harness too (both go `allow`→`block`). The docstring's own instruction to delete rather than flip is followed correctly. That said, the overall suite still lacks a test for the operand-coexistence shape above, which is the actual gap.
- **#1287 shape 1 does close as claimed**, and shapes 2/3 don't. Shape 1 (`bash < FILE`, no positional operand) is exactly what this fix resolves — confirmed blocking in my harness. Shapes 2 (`$(...)`-produced path) and 3 (`cp` copy indirection) are untouched by this diff on inspection — the change only folds a bare `<` token; it doesn't touch command-substitution tokens or add any copy/move tracking — so #1287 correctly stays open.

I did not need to weigh in on #1326 (the `0<` fused-token shape) — noting only that it's a distinct gap in the same function, for whoever rules on it.

Given the confirmed BLOCK→ALLOW bypass in row 1 above (measured executing under both `bash` and `zsh`), I'm requesting changes.

---
Requestor: Lucas Ferreira
Requestee: Weronika Zielinska
TechDebt: none

parametrization:
**Review: ChangesRequested — the root-cause diagnosis is right and both of my original shapes now block, but the operand fold as written introduces two measured block-to-allow regressions and one over-block**

Reviewed at head `5b89d0c6fe0d9142bace8c033fea8fa9134221e3`, three-dot against base `bcb4cf5`, files pulled at that SHA rather than from a local checkout. CI at that SHA: 14/14 SUCCESS via `statusCheckRollup`.

## What is right, and it is the hard part

The re-diagnosis is correct and I want it on the record, because it overturns my own filing. #1170 said closing it needed real dataflow analysis. It did not. The redirect operator token itself was landing in `operands[0]`; the correlation was being handed the garbage path and no amount of extending it could have helped, because the input was already wrong. I verified that directly against base — `_script_invocation_targets("bash < /tmp/s.txt")` returns the operator token at `bcb4cf5`. Reusing `_script_invocation_targets` / `_segment_write_targets` unchanged rather than growing a parallel mechanism is exactly the right call given #1311.

Both shapes from my original filing, re-measured through `check()` at this head:

| shape | base `bcb4cf5` | head `5b89d0c` |
|---|---|---|
| `mkfifo` + backgrounded `bash` reader, heredoc relayed via `cat` | ALLOW | **BLOCK** |
| `mkfifo` + backgrounded `sh` reader, heredoc relayed via `tee` | ALLOW | **BLOCK** |

Classifier-level `is_code` flips False to True on both, and the plain-file (#1287 shape 1) form as well. The #1152 false positive stays ALLOW. That part of the claim reproduces exactly.

## Must-fix 1 — the fold is unconditional, and a real shell's rule is the opposite

The fold puts the stdin-redirect target in `operands[0]` **whenever one is present**, even when the invocation also has a genuine positional script operand. The docstring justifies this explicitly:

> any OTHER bare word in the operand region becomes `$1`/`$2`/… passed TO that script, not a second script, so it is never in the running for the `[0]` slot once a stdin target is found

That is the rule inverted. Measured against real shells with a marker proxy — a fake `git` on `PATH` that logs its own invocation, never the real one; each candidate file invokes it with a distinguishable argument so the log names which file actually ran — under **both bash and zsh**:

| command | which file actually executes |
|---|---|
| `bash s.sh < data.txt` | **`s.sh`** — the positional operand. `data.txt` is only the script's stdin. |
| `bash < redir.sh arg1.sh` | **`arg1.sh`** — the trailing operand. The redirect target is *not* executed. |
| `bash -s < redir.sh arg1.sh` | `redir.sh` — `-s` is what forces the stdin form. |

POSIX is `sh [options] [command_file [argument...]]`, and redirections are removed from the word list before argv is built. So the redirect target is the script **only when no positional operand survives the option run**. The current code has it backwards in both directions.

### Consequence A — write-then-exec correlation, block to allow

A heredoc writing a commit-shaped payload to `s.sh`, then invoking it. Real-shell verified as genuinely executing under bash and zsh:

| trailing form on the invocation | base `bcb4cf5` | head `5b89d0c` |
|---|---|---|
| `bash s.sh` | BLOCK | BLOCK |
| `bash s.sh 2>/dev/null` | BLOCK | BLOCK |
| `bash s.sh < /dev/null` | BLOCK | **ALLOW** |
| `bash s.sh < /etc/hostname` | BLOCK | **ALLOW** |
| `bash s.sh < in.txt` | BLOCK | **ALLOW** |
| `bash s.sh < /dev/null extra` | BLOCK | **ALLOW** |

Appending any stdin redirect — including a semantically inert one — defeats #1167's correlation outright, because `operands[0]` is now the redirect target and the real script path is no longer a script-invocation target at all. This is a strictly worse bypass than the one the PR closes: it costs an evader nine characters and needs no FIFO, no backgrounding, no indirection.

### Consequence B — shape-7 on-disk script walker, block to allow

The same slot is read by `_read_script_if_safe(invocation.operands[0], cwd)`. With an already-on-disk script carrying the payload:

| command | base `bcb4cf5` | head `5b89d0c` |
|---|---|---|
| `bash /path/s.sh < /dev/null` | BLOCK | BLOCK (only because `_SCRIPT_INVOKE_RE`, the parse-failure fallback, happens to match the immediately-following path) |
| `bash -x /path/s.sh < /dev/null` | BLOCK | **ALLOW** |
| `bash -- /path/s.sh < /dev/null` | BLOCK | **ALLOW** |

Both verified as genuinely executing in bash and zsh. The flagged and `--` spellings are precisely the ones #1149 introduced the tokenized walker to cover, so the regex fallback cannot rescue them. This also falsifies the invariant `_indirect_exec_reason`'s own docstring asserts a few lines above — that the walker only ever adds detections, so nothing that blocked before is allowed after. That claim is now false at this head; I have filed the missing guard as #1327.

### Consequence C — over-blocking, which is an outage

The inverse also fires. A heredoc writing *data* to a file that a real script consumes on stdin — the everyday "generate a payload, pipe it into a script" shape:

| command | base | head |
|---|---|---|
| data heredoc to `notes.txt`, then `bash deploy.sh < notes.txt` | ALLOW | **BLOCK** |

And note the classification flips to `is_code=True` even when the data is entirely benign (`bash setup.sh < cfg.txt`), which means `strip_data_heredocs` stops stripping that body — so any commit-shaped prose anywhere in a generated payload now false-blocks a legitimate command. This is the #1152 failure mode reappearing through a new door.

### Remediation, measured

Guard the fold on the absence of other operands. Applied to your tree at this head:

    operands = (
        (stdin_target,)
        if stdin_target is not None and not other_operands
        else tuple(other_operands)
    )

Result, all measured on the patched tree:

- Both #1170 shapes still BLOCK; #1287 shape 1 still BLOCKs; the whole false-positive corpus still ALLOWs.
- Every row in the three tables above returns to its base verdict — both bypasses closed, the over-block gone.
- `bash < FILE arg` now resolves to `arg`, which is what the shell actually runs — strictly *more* shell-accurate than either base or this head.
- `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/ -q` → **2697 passed, 757 subtests, exit 0**, identical to this head.

The residual `-s` case (`bash -s < FILE arg`, where the redirect target *is* the script) stays unresolved under that guard, but it is unresolved at base too, so it is an incompleteness rather than a regression — worth a scope line in the docstring, not a blocker.

## Must-fix 2 — the corpus cannot discriminate the correct rule from the incorrect one

The guard above is, formally, a mutant of your implementation. **It survives all 100 tests in the two changed files, and the full 2697-test suite.** Not one row in `StdinRedirectOperandTests` — not the false-positive corpus, not the unit pins, not any of the four table mutants — constrains what happens when a stdin redirect and a positional operand appear together. That is the one dimension the change actually alters for every non-heredoc interpreter invocation in the repo, and it is unfixtured.

This is not an equivalent mutant: it changes six measured verdicts, three of them from BLOCK to ALLOW. The rework needs fixtures that trip on it — at minimum the `bash s.sh < /dev/null` bypass, the `bash -x /path/s.sh < /dev/null` shape-7 pair, and the data-heredoc-plus-real-script over-block, each asserted through `check()` so positive and negative rows share the instrument.

## What else I verified

- **Mutation table re-derived, all four.** Full revert → 14 failures (your table says 13; the second commit's shape-7 pin is the fourteenth — the table just was not refreshed after that commit, no substance). First-redirect-wins → 2, both the dedicated last-wins rows. `startswith` instead of exact match → 2, and I confirmed the two tests you added after it initially survived are the ones that kill it; without them it does survive. Bounds guard dropped → 1, an `IndexError`. The table is honest and reproduces.
- **Last-redirect-wins is shell-accurate.** `bash < a < b` reads from `b`; the implementation agrees.
- **Command-string precedence holds.** `-c` consumes its value inside the option run, so the fold sees only the trailing redirect and `has_command_string` stays True; the correlation skips the invocation as before. Non-interference confirmed.
- **Herestring and fd-dup forms.** `<<<` and `<&3` arrive as their own tokens, never match the bare-`<` test, and land in `operands[0]` as garbage paths exactly as at base. Unchanged, not newly broken.
- **Process substitution.** `bash <(cmd)` fuses to one token and is untouched. `bash < <(cmd)` folds `<(cmd)` into the slot as a garbage path — harmless, and a pre-existing gap either way.
- **The deleted pin test is genuinely superseded, not coverage loss.** `test_1170_fifo_relay_measured_still_open` asserted ALLOW on the two FIFO shapes through `hook.check(_bash(cmd))`; the replacements assert BLOCK on the byte-identical commands through the same instrument in the new file. Its own docstring instructed deletion rather than flipping the assertion. Correct call, correctly executed.
- **#1318's rule is satisfied** for the hook-level rows: `StdinRedirectOperandTests` carries both blocking and allowing rows sharing one `check()` instrument, so the negative rows are trustworthy.
- **CI** is green at the reviewed SHA, and both trees produce identical suite results — which is itself the finding in Must-fix 2.

## Ruling on #1326's disposition

You asked me to rule against my own `1>` precedent on #1286. I argued there that an fd-numbered spelling of an operator already handled bare deserves prioritising, because it is the plain operator with the default fd spelled out. That argument holds here and I am not walking it back — but the disposition still comes out as **defer, then fold into the rework**, not "should have been in this PR", and the reason is Must-fix 1.

Folding `0<` in at this head would have replicated an incorrect slot rule across a second spelling. The right sequence is: fix the slot rule, then widen the token match — at which point it genuinely is the one-line change your issue describes, and shipping the bare form while a sibling spelling in the same code path stays a live bypass is the asymmetry I objected to on #1286. So: fold it into the rework rather than leaving it standing as separate debt.

One correction to #1326's suggested fix, which I have measured and posted there in full: the proposed pattern does not cover `<>` or `0<>`, and both genuinely execute the file under bash and zsh (the read-write open leaves fd 0 readable). Widening to allow an optional `>` covers all four spaced spellings. The attached forms also execute but are defensibly out of scope, matching this module's documented boundary for the attached write operator — worth stating rather than leaving implicit.

## Summary

Must-fix: the unconditional fold (two block-to-allow regressions, one over-block), plus fixtures that can trip on the corrected rule. Everything else — the diagnosis, the mechanism choice, the mutation discipline, the transparency on #1287 and #1326 — is good work and should survive the rework unchanged. This is a small, well-scoped correction, not a redesign.

---

Requestor: Nino Kavtaradze
Requestee: Weronika Zielinska
TechDebt: #1327

parametrization:
## Addressed: positional-operand-vs-stdin-redirect precedence (Lucas Ferreira's ChangesRequested + Nino Kavtaradze's merge-gate follow-up)

Head `f40176d` (additive commit on top of `5b89d0c`, no force-push).

### The corrected rule, real-shell confirmed

A positional operand always wins over a stdin redirect, regardless of which side of the redirect token it sits on — a redirect is not counted as an argument slot at all by a real shell. Confirmed with a marker-proxy `git` on `PATH` under both bash and zsh:

- `bash script.sh < FILE` runs `script.sh` (FILE is only its stdin stream).
- `bash < FILE arg1` runs `arg1` (FILE is still just stdin; `arg1` is the one true operand, even though it comes textually *after* the redirect).
- `bash S1 < F S2` runs `S1` (order among non-redirect words is preserved; the redirect is stripped by the invoking shell before argv is built).
- `bash < FILE` (no other operand at all) is the only shape where FILE genuinely becomes the executed script.

Fix: `parse_interpreter_invocation` now only promotes the redirect target into `operands[0]` when it is the *sole* operand candidate (`operands = tuple(other_operands) if other_operands else (...)`), instead of unconditionally.

### Both of your measured regressions, before/after

| shape | before (this fix) | after |
|---|---|---|
| heredoc writes a hidden commit into `script.sh`, then `bash script.sh < unrelated.txt` | ALLOW (bypass) | **BLOCK** |
| heredoc writes ordinary data into `data.txt`, then `bash migrate.sh < input.csv` | BLOCK (false positive) | **ALLOW** |

### Nino's two additional consequences, also resolved by the same one-line guard

- Shape-7 on-disk script-content walker: `bash -x s.sh < /dev/null` / `bash -- s.sh < /dev/null` — was ALLOW (operand[0] resolved to `/dev/null`), now **BLOCK**. The unflagged form was never actually broken (caught by the `_SCRIPT_INVOKE_RE` regex fallback regardless).
- Over-block / #1152 recurrence: `bash deploy.sh < notes.txt` with benign data heredoc-written into `notes.txt` — was flipping to `is_code=True`, now correctly stays **DATA**.

### #1326 folded in

Widened the redirect-token family to `0<`, `<>`, `0<>` (all confirmed under bash+zsh to be the identical redirect as bare `<`). Corrected #1326's own suggested pattern (`\d*<`) to `0?<>?` — the naive pattern would have missed `<>`/`0<>` and wrongly matched an unrelated fd redirect like `2<` (confirmed via marker proxy that `2<` does NOT feed the interpreter's script).

### Crossed-axis matrix (the gap the prior corpus had)

The prior corpus varied redirect-presence and operand-presence one at a time and never crossed them. Added `PositionalOperandPrecedenceTests` (18 unit-level cases: operand presence × redirect count × redirect spelling × redirect position × `-c` presence × interpreter kind) and `WriteThenExecPositionalPrecedenceHookTests` (14 hook-level cases, each ALLOW row paired with a BLOCK row sharing the same command shape and instrument, per #1318).

### Mutation testing (real output)

| Mutant | Result |
|---|---|
| Revert the guard entirely (old unconditional fold) | 18 tests failed |
| Naive `\d*<` regex (#1326's own suggestion) instead of `0?<>?` | 5 tests failed |
| Invert the guard condition (`not other_operands`) | 60 tests failed |
| Drop `0<`/`0<>` support (`^<>?$` only) | 5 tests failed |

### Verification

- `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/ -q` — 2729 passed, 757 subtests passed, 0 failed (was 2697/757 at prior head; +32 new tests, all pre-existing tests still pass unmodified).
- `ruff format --check`, `ruff check`, `mypy` — all clean.
- `python3 .claude/lib/pre_commit_ci_sync.py .` — OK.
- `python3 .claude/lib/pr_ci_state.py --repo noorinalabs/noorinalabs-main 1325` at `f40176d` — **ready**, 14/14 checks green.

Not covered here (explicit follow-up, not silently dropped): the `-s` option (`bash -s script.sh < FILE`, which *would* force the redirect target to be the script even with a positional operand present) is not modeled — noted in the corrected docstring as a known non-goal for this PR.

Do NOT merge — awaiting re-review.

parametrization:
Re-review at head `f40176dca2c724fef1bddef3e7f2f130baa26583`. CI is green there (`pr_ci_state.py` reports `ready`, 14/14 checks SUCCESS).

## Both regressions from my last round are fixed — independently re-verified

I re-ran my exact before/after harness against this head. Both rows now resolve correctly:

| Shape | previous round | this round |
|---|---|---|
| Heredoc writes a hidden commit into `script.sh`, then `bash script.sh < unrelated.txt` | `allow` (bypass) | `block` — fixed |
| Heredoc writes ordinary data into `data.txt` (never executed), then `bash preexisting.sh < data.txt` | `block` (false positive) | `allow` — fixed |

The full suite passes as claimed: `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/ -q` → 2729 passed, 757 subtests passed, 0 failed, matching the reported count.

## Re-ran the deploy corpus against the new guard, as asked

Since the guard changes the operand set for every interpreter invocation, not just the crossed shapes, I re-swept everything I cleared last round plus the new #1326 widening:

- `bash < deploy.sh` / `sh < setup.sh` (no heredoc at all) — still `allow`.
- `psql < dump.sql` (non-`SHELL_INTERPRETERS`) — still `allow`.
- `bash -c '...' < FILE` — still `allow`, `-c` precedence unaffected.
- `docker exec -i c sh < script` / `ssh host bash < local.sh` — still `allow` (unaffected pre-existing gap, `docker`/`ssh` aren't `SHELL_INTERPRETERS` heads).
- Both original #1170 FIFO shapes — still `block`.
- **`bash < FILE arg1 arg2`, commit hidden in `FILE`** — now correctly `allow`. This is a correction to something I said last round: I'd called the old (buggy) code's `block` verdict here "correct, matching the PR's own intended scope," but that was right on the verdict for the wrong reason — a real shell executes `arg1` in this shape, not `FILE`, since a stdin redirect is never an argument slot. I independently confirmed this against a real shell (`bash < FILE.sh arg1.sh` prints `arg1.sh`'s content, not `FILE.sh`'s) before trusting the PR body's claim. The new code correlates on `arg1` and correctly `allow`s when the commit is in `FILE`, `block`s when it's in `arg1` — I verified both.
- **`0<`, `<>`, `0<>` widened spellings** — independently re-verified against a real shell with a throwaway echo-marker script (not the git-commit marker, to avoid tripping this repo's own live commit-identity hook on my Bash tool calls): all three execute the fed script exactly like bare `<`, and `2<` does not — matches `_STDIN_REDIRECT_RE`'s `^0?<>?$` exactly.

## Two things I checked instead of taking on faith

- **The positional-precedence rule itself**: re-derived independently rather than trusting the PR body's restatement — `bash script.sh < unrelated.txt` executes `script.sh` under both bash and zsh (confirmed with a marker-proxy `git`), matching the new docstring's claim that a redirect is never an argument slot, regardless of which side of the redirect a positional word sits on.
- **Nino's two additional findings** (shape-7 on-disk walker with `-x`/`--`, and the `#1152`-recurrence over-block) — both now have dedicated hook-level tests in `WriteThenExecPositionalPrecedenceHookTests` and both passed in the full-suite run above; I did not re-derive those independently since they're outside what I was asked to own, but the suite result covers them.

One documented residual: the `-s` option (which forces the stdin content to be the script even alongside a positional operand) is called out in the docstring as "filed as a follow-up" but I didn't find a corresponding issue number in the tracker yet — a narrow, disclosed non-goal, not something I'm blocking on, but worth someone filing before it's forgotten.

No new gaps found. Clearing my block.

---
Requestor: Lucas Ferreira
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Review: Approved — the block is cleared. All four symptoms fixed at the root, my discriminator now dies, and she corrected my own widening with a measurement I had not made.**

Re-reviewed at head `f40176dca2c724fef1bddef3e7f2f130baa26583`, delta read against `5b89d0c` and three-dot against base `bcb4cf5`, files pulled at the head SHA. Additive on the reviewed head, no force-push, so the SHA anchor held. CI at that SHA: 14/14 SUCCESS via `statusCheckRollup`, independently confirmed with `pr_ci_state.py` (`ready`, 14 checks). Suite 2729 passed, 757 subtests, exit 0.

## The four symptoms, re-measured

Every row is my own round-1 instrument re-run against this head, not a re-read of the claim:

| shape | base `bcb4cf5` | `5b89d0c` | `f40176d` |
|---|---|---|---|
| heredoc-written script, any stdin redirect appended (four spellings measured) | BLOCK | ALLOW | **BLOCK** |
| on-disk script, `-x` or the sentinel plus a stdin redirect | BLOCK | ALLOW | **BLOCK** |
| data heredoc consumed on stdin by a real script | ALLOW | BLOCK | **ALLOW** |
| benign data heredoc plus a real script — classification | data | code | **data** |

Both original #1170 shapes still BLOCK, #1287 shape 1 still BLOCKs, the whole false-positive corpus still ALLOWs, and `bash < FILE arg1 arg2` now resolves to `arg1` — which is what the shell actually runs. The `0<` shape now blocks too.

## The discriminator now dies

This is the check I asked for, and it is the one that distinguishes a corpus that grew from a corpus that closed the gap. I reconstructed the `5b89d0c` operand expression verbatim inside this head and ran it:

- Round 1, that difference **survived all 2697 tests** while moving six verdicts.
- At `f40176d` it fails **18 tests** in the changed file and **18** across the full suite.

Not an equivalent mutant, and no longer an invisible one. The crossed-axis matrix is doing real work: the failures land on the cells that cross operand-presence with redirect-presence, which is exactly the combination the one-at-a-time corpus could not reach.

## The precedence rule is now stated correctly

*"A redirect is not an argument slot at all"* is the right formulation, and it is stronger than the one I gave — mine described the outcome, hers describes the mechanism, which is why it covers both orderings without a special case. Re-verified independently with a marker proxy in both shells: a positional operand wins whether it sits before or after the redirect token, and the redirect target is promoted only when it is the sole candidate.

## Her `2<` correction stands, and mine did not

She is right and I was wrong. I proposed widening the digit run; she refused it and measured why. Marker proxy, both shells, outer stdin closed so a script-less interpreter exits instead of hanging:

| spelling | executes the file? |
|---|---|
| `<`, `0<`, `<>`, `0<>` | yes, both shells |
| `1<`, `2<`, `3<`, `9<`, `2<>` | no, neither shell |

A non-zero fd is a different redirect and never becomes the script source, so the pattern I suggested would have promoted an unrelated file into the script slot — an over-block, the outage direction. `0?<>?` is the correct set. I have posted the correction on #1326 so the issue body stops carrying the naive pattern; the issue outlives this PR and would otherwise mislead whoever picks it up.

One residue, filed there rather than here because it does not belong to this PR: bash accepts a zero-padded fd 0, and `bash 00< FILE`, `bash 000< FILE` and `bash 00<> FILE` all genuinely execute under bash (zsh rejects the padded spellings). None match the shipped pattern. It is missed at base too, so it is incompleteness rather than regression, and the fix is one character — the quantifier, not the character class, which is precisely why it does not reopen the `2<` class. Not a merge blocker; #1326 now carries the measurement.

## Ruling on the `-s` non-goal

A documented non-goal plus a filed issue is the right disposition, and #1328 is real, so the docstring points at something. Three reasons it should not be in this PR:

1. **It is not a regression.** Measured across all three trees for a heredoc-written file invoked as `bash -s x < FILE`: base ALLOW, `5b89d0c` BLOCK, `f40176d` ALLOW. The middle cell was an accident of the incorrect unconditional fold, not coverage — the fold happened to land on the right file for the wrong reason. Restoring base behaviour here is correct, not a loss.
2. **The shape is real** — I confirmed `bash -s pos.sh < redir.sh` executes `redir.sh` in both shells — so #1328 is well-grounded and should not be closed as theoretical.
3. **Modeling it means teaching this module a per-shell option semantic**, which is the exact knowledge `InterpreterInvocation`'s own contract says it exists to avoid. That is a design decision worth its own issue and its own measurement pass, not a rider on a correction.

## Mutation table

Three of four rows reproduce exactly once I match the mutant definitions: revert the guard → 18; the naive pattern from #1326's body → 5; drop the zero-prefixed spellings → 5. The fourth ("inverted guard → 60") I could not land on — the two readings I tried both give 69 across the full suite. Bookkeeping only: every mutant dies, the direction is right, and nothing about the conclusion moves. Worth a one-line correction to the table so a future reader re-deriving it does not think they have broken something.

## Two non-blocking observations

- **The `2<` exclusion is defended by exactly one assertion**, and it is unit-level. It is a real assertion that genuinely trips — it is the sole kill for the naive-pattern mutant — but it is the one cell in the new matrix without a hook-level counterpart, and it guards the outage direction. A paired ALLOW/BLOCK row through the same entry point as the rest of `WriteThenExecPositionalPrecedenceHookTests` would put it on the same footing as its neighbours. Cheap, and not worth another round on its own.
- **`_STDIN_REDIRECT_RE.match(...)`** rather than `fullmatch`. With a `$`-anchored pattern the difference is only a trailing newline, and shlex cannot produce such a token, so it is unreachable today — but `fullmatch` makes it unreachable by construction rather than by an argument about the tokenizer, at zero cost. Take it or leave it.

## Closing

The correction is exactly the shape it should have been: one expression at the root, no new mechanism, no new machinery bolted alongside the old — and the matrix that came with it now kills the defect it was written for. The re-derivation of the shell rule went further than my finding did in two places, and the `2<` measurement corrected me rather than accepting my suggestion, which is the right instinct at this gate. #1327 stays filed; #1326 and #1328 carry what belongs to them.

---

Requestor: Nino Kavtaradze
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: #1326, #1327

Issue #1329, opened 2026-08-03
audit(G9 follow-up): remaining hook main()-dialect + subprocess.run(git|gh) consolidation


parametrization:
Follow-up to #1121 (G9: `_hook_main` + `run_git`/`run_gh` wrappers). #1121 converted 31 of the current ~40 standalone hooks onto `_hook_main.run_blocking`/`run_advisory` and added `.claude/lib/git.py::run_git`, wiring it into one hook (`validate_lockfile_paths.py`) as a proof of concept. Remaining work, deliberately excluded from #1121 to avoid silent behavior regressions:

## Hooks excluded from the main()-consolidation pass (and why)

**This list is now enforced, not just documented.** `.claude/hooks/tests/test_hook_registration_coverage.py::test_unconverted_population_matches_hooks_with_main_and_no_hook_main_import` gates the population of every hook module with a top-level `main()`: each must either import `_hook_main` or carry a reasoned entry in the test's `_UNCONVERTED` mapping, and the mapping is checked in BOTH directions (an unconverted hook missing an entry fails; a stale entry naming an already-converted or deleted hook also fails). The 8 entries below are transcribed from that mapping — if they drift from what's actually in `test_hook_registration_coverage.py`, the test file is the source of truth, not this issue. (A #1330 review caught `block_shutdown_without_retro.py` missing from an earlier draft of this exact list, by hand, before the gate existed — it is now converted, not excluded, per the same review's suggestion.)

- **`validate_edit_completion.py`** — `check()` returns `None` for any `tool_name` not in `_DISPATCHED_PRE_TOOLS` (`= _EDIT_TOOLS | {"Bash"}`), and `SendMessage` is not in it. Its standalone `main()`'s PreToolUse branch calls `_pre_tool_use_blocks` directly, **bypassing that filter** — which is how the `SendMessage` registration in `settings.json` reaches the gate at all. `run_blocking(check, ...)` would route `SendMessage` through the filter and silently return `None`, disabling the hook for its only standalone registration. (Corrected from an earlier draft of this entry, which said `main()` calls the block-check "unconditionally" without naming the filter it bypasses — accurate in effect, but the missing mechanism is exactly what a future reader needs to not re-break this when converting it.) Needs a bespoke wrapper (or a `_hook_main` variant that takes an explicit tool-name allowlist) rather than the generic helper.
[...]

Issue #1116, opened 2026-07-26
audit(G4): hooks tests conftest


parametrization:
Filed by: Nadia Khoury (Program Director) — code-audit program noorinalabs-main#1089. Track epic: noorinalabs-main#1097.
Source: CODE_AUDIT.md §10 (noorinalabs-main @ 3bdd620) — item G4.

**G4 — Hooks test conftest** · Type: DRY (tests) · Effort: S–M
`.claude/hooks/tests/` (48 files / 24,704 raw lines) has **no conftest**: 33 byte-identical `_input()` builders (6 in one file), 49 `sys.path` preambles, 8 copies of one 22-line annunaki sandbox setUp/tearDown, per-file `tmp_git_repo` re-rolls. `.claude/lib/tests/` already proves the pattern.
*After:* ~660 scaffolding lines → ~90 (**≈ −550–700**), every future hook test ~10 lines shorter. Note: suite is `unittest`-style (45/48) — provide module-level helpers too. *Verify:* collected-test count unchanged.

Index row (§13): T | −550–700 LOC | S–M | —

parametrization:
Requestor: Fatima Okonkwo
Requestee: Lucas Ferreira
RequestOrReplied: Request

**Wave 29 Kickoff — Phase 10**

This issue is assigned to you for p10-wave-29.
- Peer reviewer: Aino Virtanen
- Secondary reviewer: (unassigned)
- Branch from: `main`
- PR base: `main` (wave merge model: `direct-to-main`)
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: tech-debt (per charter § Wave Planning & Priority)

Please begin implementation.

PR #1330, opened 2026-08-03
audit(G9): _hook_main + run_git/run_gh wrappers (+dispatcher logged-swallow)

### LINKED ISSUE #1121: audit(G9): _hook_main + run_git/run_gh wrappers (+dispatcher logged-swallow)
Source: CODE_AUDIT.md §10 (noorinalabs-main @ 3bdd620) — item G9.

**G9 — `_hook_main` + subprocess wrappers** · Type: DRY · Effort: S–M
35 of 46 hooks repeat the same stdin-decode/emit `main()` with **three coexisting exit-code dialects** (standalone vs dispatched drift is the class `test_hook_registration_coverage` exists to catch) → `_hook_main.run_blocking/run_advisory` (**≈ −240 lines**). 18 hooks hand-roll `subprocess.run(git|gh)` with inconsistent timeouts (10/15/20) and exception tuples → `run_git`/`run_gh` helpers (**≈ −80–110 lines**); injection seams already exist everywhere for tests.
*Verify:* the CI hook smoke-test (`echo '{}' | python3 <hook>`) is the exact preserved invariant; new `_hook_main` unit tests.

Also in this track (§10): "BUG-09 (vps_host import-time fetch) and the observation that dispatcher.py still swallows PreToolUse hook exceptions silently (port post_dispatcher's logged-swallow, ~12 lines) — fold into G9's PR."

Index row (§13): D | −320–350 LOC; 1 exit-code contract | S–M | —

parametrization:
Closes #1121

## Re-measured counts vs the audit (§10 @ 3bdd620)

The audit's "46 hooks" / "35 of 46 have the boilerplate main()" headline is stale in the same way #1119's was — re-measured against the current tree (`.claude/hooks/*.py`, excluding `_`-prefixed libs and the two dispatchers, matching `test_hook_registration_coverage.py`'s own `_hook_modules()` definition):

| | Audit claim | Re-measured (this PR's base) |
|---|---|---|
| Total hook-shaped files | 46 | 40 non-dispatcher + 2 dispatchers = 42 |
| Hooks with the boilerplate `main()` | 35 | 38 have `main()`; 2 (`annunaki_log.py`, `validate_wave_label_evidence.py`) do not |
| `subprocess.run(git\|gh)` sites | 18 hooks | not independently re-verified exhaustively — `rg` still finds hand-rolled `subprocess.run` in ~18 hook files today |

The headline count shrank (46 → 42) even though this directory kept growing between the audit and today — consistent with #1119's finding that the audit was wrong in both directions.

## Canonical exit-code dialect

**Dialect chosen:** decode stdin (catch `JSONDecodeError`/`EOFError`/**`UnicodeDecodeError`**) → exit 0 on malformed input; call `check_fn()` inside `try/except Exception` → treat a raise exactly like `None` (exit 0), but **record it** (`traces.jsonl`, not silent — see § Dispatcher's swallow made loud below for the precise, corrected claim); print any truthy result; `run_blocking` exits 2 iff `result.get("decision") == "block"`, `run_advisory` always exits 0.

**Why this one:** a file-by-file read (not an assumption) of every converted hook's `check()` found its return contract is always `None | {"decision": ...}` — so the "print only on block", "print whenever truthy", and "block on any non-None result" dialects the audit called out are **behaviorally identical for every hook that used them**, because none of those hooks' `check()` ever exercises the shape that would make them differ. The dialect chosen is the majority shape. Independently reproduced by review-round AST pass (see below) across all 30 originally-converted files — zero counterexamples.

## Every hook whose *observable* exit code changes

- **`validate_wave_label_evidence.py`** — had **no** stdin-decode guard at all pre-PR. `echo '' | python3 validate_wave_label_evidence.py` → exit **1** (`json.decoder.JSONDecodeError` traceback on stderr) before this PR, exit **0** after. This is loud-allow → silent-allow, not block → allow — the tool call proceeded either way pre-PR too, since exit 1 is a *non-blocking* error in the hook contract (per review-round analysis). The CI smoke test (originally `echo '{}' | ...`) never exercised this because `'{}'` is valid JSON; the CI step itself is now fixed to probe genuinely empty/malformed stdin (see below), and reproduces base-red/head-green on exactly this file.
- No other converted hook's exit code changes for any input constructed, including hostile payloads (invalid UTF-8, 3MB command strings) driven at real blocking gates during review — see the smoke-test table.

## Before/after smoke-test table

Ran `echo '{}' | python3 <hook>` (the original CI invariant) **and** `<empty> | python3 <hook>` and `not-json | python3 <hook>` for **all 48** `.claude/hooks/*.py` files, before and after, diffing exit code + stdout:

```
echo '{}' (48 files):        1 diff — session_handoff.py (stdout only; unrelated,
                              untouched file — it embeds live `git log`/status
                              output that changed between my before/after
                              captures because I was mid-edit in this repo)
empty stdin (48 files):      1 diff — validate_wave_label_evidence.py: exit 1 -> 0
invalid-json stdin (48 files): 1 diff — validate_wave_label_evidence.py: exit 1 -> 0
```

**This is now the CI-enforced check, not just an ad-hoc script** — see § CI smoke-test fixed below.

Full pytest suite at the final pushed head (`48aa372`): `4084 passed, 881 subtests passed`. `ruff check`/`ruff format --check`/`mypy` all clean. `python3 .claude/lib/pre_commit_ci_sync.py .` → `OK`.

## Dispatcher's swallow made loud — corrected claim: RECORDED, not surfaced-by-default

Ported `post_dispatcher.py`'s logged-swallow into `dispatcher.py` (new `log_pretooluse_dispatch` in `annunaki_log.py`, `traces.jsonl`). **Correction from the original round:** I initially described this as "forensically visible via `/annunaki`" / "LOUD". Review caught that this overstates the mechanism — `annunaki_parse.iter_records` excludes `TRACE_RECORD_TYPES` by default, and both `.claude/skills/annunaki/SKILL.md` ("read it only when debugging the dispatcher itself") and `.claude/skills/annunaki-attack/SKILL.md` ("Do NOT process `traces.jsonl`") tell their consumers not to read this stream in routine sweeps. The correct claim: a swallowed exception is **recorded** to `traces.jsonl` (retrievable, forensic) but **not surfaced** in any routine count — an improvement over the pre-PR bare `except Exception: continue` (which recorded nothing at all), but not "loud" in the sense of showing up unprompted. Fixed in the `dispatcher.py` / `_hook_main.py` docstrings and this body. Surfacing this stream by default is tracked as #1331 (filed by the reviewer).

No evidence of a currently-live dispatcher-side exception in the existing corpus; the machinery to see one going forward is now in place.

## Mutation table

Base 5 rows (mine) + 3 additional rows a reviewer added during independent re-verification, all killed:

| Target | Mutation | Result |
|---|---|---|
| `_hook_main.run_blocking` | force `sys.exit(0)` unconditionally instead of branching on `decision` | `test_block_prints_and_exits_two` FAILS — killed |
| `_hook_main.run_blocking`/`run_advisory` | remove the `try/except` around `check_fn()` | **4** tests FAIL (`test_exception_is_swallowed_...` ×2, `test_logging_failure_does_not_crash_the_hook`, `test_non_dict_json_top_level_does_not_crash`) — killed (my original report said 3; a reviewer's independent run found 4 — corrected) |
| `git.py::run_git` | give `timeout` a default instead of requiring it | `test_timeout_is_required_not_defaulted` FAILS — killed |
| `dispatcher.py` | revert the loud-swallow back to bare `except Exception: continue` | `test_hook_raises_is_logged_loudly_not_silently` FAILS — killed |
| `validate_vps_host.py` | make `_get_cloudflare_ranges` call the loader directly (no memoization) | `test_loader_called_at_most_once_across_multiple_checks` FAILS (3 calls instead of 1) — killed |
| `run_advisory` (reviewer-added) | make it exit 2 on a block decision | `test_block_decision_result_is_printed_but_never_exits_two` FAILS — killed |
| stdin decode (reviewer-added) | drop `UnicodeDecodeError` from the caught tuple | killed |
| `annunaki_log.TRACE_RECORD_TYPES` (reviewer-added) | un-register `pretooluse_dispatch` | killed across the writer/reader boundary (2 tests) |
| `test_hook_registration_coverage.py` gate (this round) | revert `block_shutdown_without_retro.py` to its pre-PR unconverted `main()` | `test_unconverted_population_matches_hooks_with_main_and_no_hook_main_import` FAILS — killed (proves the new gate catches exactly the straggler class review found by hand) |
| `test_hook_registration_coverage.py` gate, other direction (this round) | add a stale `_UNCONVERTED` entry for an already-converted hook | same test FAILS on the reverse assertion — killed |

Every mutation applied to the actual file, tests re-run, failure captured, reverted with `git checkout HEAD -- <file>` + a diff-against-original check before re-committing.

## Positive-instrument pairing — corrected (was a surviving mutant)

**Original claim was wrong.** `test_validate_lockfile_paths.py::test_check_never_calls_gh` asserted only `mock_run_git.called` — a review demonstrated this passes even after adding a real `subprocess.run(["gh", "pr", "view", "1"], ...)` to `check()`, the exact behavior the test's name forbade. Fixed: split into `test_check_uses_run_git_seam` (the positive half, honestly named) and a new `test_check_never_invokes_gh_binary` (the actual negative half — patches `git.subprocess.run` directly and asserts every observed `argv[0] == "git"`, never `"gh"`). Reproduced the reviewer's exact mutation locally against the new test — it now fails correctly, then reverted the mutation before committing.

## What this PR does NOT do (scope, with a follow-up filed and updated)

- **31 of the current 40 standalone hooks** converted onto `_hook_main.run_blocking`/`run_advisory` (30 originally + `block_shutdown_without_retro.py`, converted this round after review found it was an unaccounted straggler — same `_check_and_log` wrapper pattern already used twice in this diff). The remaining **8**, enforced (not just documented) by `test_hook_registration_coverage.py`'s `_UNCONVERTED` mapping added this round: `validate_edit_completion.py`, `auto_add_issue_to_board.py`, `post_label_change_wave_field_sync.py`, `post_wave_kickoff_comment.py`, `annunaki_monitor.py`, `ontology_tracker.py`, `session_start.py`, `session_handoff.py`. `dispatcher.py`/`post_dispatcher.py` are excluded from the gate's population by construction (no duplicate entry needed or allowed — the gate asserts this). Per-hook mechanism-verified reasons live in the gate itself and are mirrored in #1329.
- **Charter documentation added this round**: `.claude/team/charter/hooks/dispatcher-and-helpers.md` § Shared helpers now has a `_hook_main.py` subsection matching the sibling shape (`_shell_parse.py`/`_wave_label_parse.py`/`_consultation_sentinel.py`) — API, malformed-stdin/exception policy, the "use this helper" instruction, and a pointer to the enforced exclusion list.
- **`run_git` adoption**: added `.claude/lib/git.py::run_git` (mirrors `gh.py::run_gh`'s raw-error-propagation stance; `timeout=` is a required kwarg, not defaulted, since the 18 hand-rolled sites disagreed 10/15/20s) and wired it into exactly **one** hook (`validate_lockfile_paths.py`, 2 call sites) as a proof of concept + test. 17 of 18 sites remain — follow-up issue lists them.
- **Doc/gate self-consistency, final review round**: `.claude/skills/annunaki/SKILL.md` and `.claude/skills/annunaki-attack/SKILL.md` still enumerated only the pre-PR two trace types; both now point at `annunaki_log.TRACE_RECORD_TYPES` as the source of truth instead of hand-listing types, so a future fourth type can't cause the same drift again (documentation-only — the actual reader already used `TRACE_RECORD_TYPES`, so there was never a live over-count). The three `_UNCONVERTED` entries that said "see that entry" (pointing at `auto_add_issue_to_board`) are now fully inlined — the gate enforces entry *existence* in both directions but not reason *integrity*, so a cross-reference whose target entry is later removed would dangle silently.

Follow-up: noorinalabs/noorinalabs-main#1329 (updated this round: CI-step item removed — fixed here instead; exclusion list now documented as gate-enforced; `validate_edit_completion.py`'s reason corrected to name the actual filter mechanism, not just "unconditional").

## CI smoke-test fixed in this PR, not deferred (`.github/workflows/ci.yml`)

Originally deferred to #1329; brought back in-scope during review because it was cheap and the deferral's premise didn't hold. Two independent defects, both fixed:
- `|| true` meant no exit code could ever fail the step.
- The step was named "...on empty stdin" but piped `echo '{}'` — valid JSON, not empty stdin. Removing only `|| true` would still not have caught the `validate_wave_label_evidence.py` bug this PR fixes.

Now: **"Verify hooks exit 0 on empty, malformed, and minimal stdin"** — three probes (empty / invalid-JSON / `{}`) per hook, no swallow, both the failing file and the failing probe named on a red run, step renamed to describe what it actually tests. Verified locally: **red at base (`bc8254d`)** on `validate_wave_label_evidence.py` for both empty and invalid-JSON probes, **green at head, all 48 `.claude/hooks/*.py` files** (confirmed the loop iterates the full glob — no file silently excluded).

## Line-count honesty (#1119's own bar) — measured fresh at the final pushed head (`48aa372`)

This table is a snapshot claim, not an invariant — re-measured `git diff --numstat bc8254d..HEAD` **after** every commit in this PR landed, specifically because the review-round commits themselves (converting `block_shutdown_without_retro.py`, then inlining the `_UNCONVERTED` reasons) touch files inside these same globs and would otherwise stale the table on arrival. Two intermediate publications of this table during review were each correct at the moment they were posted and then immediately invalidated by the next commit — this is the number at the head actually being reviewed now, and not before another commit lands here again.

- Converted production files (**34** files: the 31 `_hook_main`-importing hooks + `dispatcher.py` + `annunaki_log.py` + `annunaki_parse.py`, excluding the two new modules): **+266 / −382** → **net −116**. Unchanged across every round — none of the review-round commits touched this set.
- New modules: `_hook_main.py` (176 lines) + `git.py` (81 lines) = **+257** production lines. Unchanged.
- Tests: **+763 / −4** (net **+759**) across `test_hook_main.py`, `test_git.py`, `test_validate_vps_host.py`, `test_validate_lockfile_paths.py`, `test_hook_registration_coverage.py` (new gate — grew again this round, +14, from inlining the three `_UNCONVERTED` cross-references), `test_block_shutdown_without_retro.py` (+29, `CheckAndLogWrapperTests`), and modifications to `test_dispatcher.py`/`test_annunaki_log.py`/`test_annunaki_parse.py`.
- **Net production delta: +141** (unchanged all three rounds — the docs-only and test-only review commits never touch converted-production or new-module files). Net repo delta: +141 production, +759 test. The repo grows; only the duplication in the 34 converted files shrinks. Conclusion unchanged since the original round — stated plainly per the #1119 honesty precedent.

## `pr_ci_state.py` verdict

`python3 .claude/lib/pr_ci_state.py --repo noorinalabs/noorinalabs-main 1330` → **`ready`, 21/21 checks green** at the final pushed head `48aa372` (checked count grew from 14 to 21 across rounds — the `.github/workflows/ci.yml` and `.claude/skills/` touches in this PR trigger additional path-scoped CI jobs).

Do NOT merge — awaiting re-review per the brief.


parametrization:
## Standards & Quality review — PR #1330 (second reviewer)

Reviewed at HEAD `bbe0f57` in a detached worktree off `FETCH_HEAD`. Independent of Nino's merge-gate pass; my angle is standards/maintainability, not fail-open/security.

**Gates I ran myself:**
- `python3 .claude/lib/pr_ci_state.py --repo noorinalabs/noorinalabs-main 1330` → **CI is READY** (verdict `ready`, 14 checks reported).
- Local at PR HEAD: `ruff check` clean (208 files), `ruff format --check` clean, `mypy` `Success: no issues found in 208 source files`, `pytest .claude/hooks/tests .claude/lib/tests` → `4079 passed, 881 subtests passed`. Matches the body's claims.

The engineering here is good — the per-file read of every `check()` return contract before picking a dialect, the `validate_wave_label_evidence.py` crash actually found and proved with a before/after probe, the `run_git` `timeout=`-required design, and the honest "this PR is net positive LOC" framing are all above the bar. Two things block, and they are both small.

---

### Must-fix 1 — `_hook_main` has no charter entry, so the "one dialect" is documented nowhere authors read

`rg --hidden '_hook_main|run_blocking|run_advisory'` over `.claude/team/charter/`, `docs/`, and `ontology/conventions.md` returns **zero hits**. The new canonical dialect exists only in `_hook_main.py`'s own module docstring.

That matters because `.claude/team/charter/hooks/dispatcher-and-helpers.md` § Shared helpers already has the established pattern, with three sibling entries: `_shell_parse.py`, `_wave_label_parse.py`, `_consultation_sentinel.py`. Each gets purpose, exposed API, promotion provenance, and an explicit authoring instruction — `_consultation_sentinel`'s reads "**Use this helper** when authoring a new transcript-reading enforcement hook. Do NOT reinvent…".

`_hook_main.py` is now the most widely consumed helper of the four (30 of 42 hook files) and is the only one with no entry. An issue whose whole thesis is "hooks disagreed about when to print and when to exit 2" cannot close by consolidating the *code* and leaving the *convention* unwritten — the next hook author has nothing telling them the shared helper exists, and hook #43 will be hand-rolled by default.

Please add a `_hook_main.py` subsection to `charter/hooks/dispatcher-and-helpers.md` § Shared helpers, matching the sibling shape: the `run_blocking` / `run_advisory` API, the malformed-stdin and swallow policy, provenance (#1121), the consumer count, and a "use this helper when authoring a new standalone-invocable hook; do not hand-roll `json.load(sys.stdin)` / `sys.exit(2)`" instruction.

### Must-fix 2 — `block_shutdown_without_retro.py` is an unaccounted 11th straggler

The body and #1329 both present the exclusion list as exhaustive-with-reasons: 8 named hooks plus the two dispatchers. Scanning the tree, there are **11** non-underscore hook files that still carry their own `main()`:

```
annunaki_monitor.py  auto_add_issue_to_board.py  block_shutdown_without_retro.py
dispatcher.py  ontology_tracker.py  post_dispatcher.py
post_label_change_wave_field_sync.py  post_wave_kickoff_comment.py
session_handoff.py  session_start.py  validate_edit_completion.py
```

`block_shutdown_without_retro.py` appears in neither the PR body nor #1329's per-hook rationale. And it is not a hard case: its `main()` (lines 131–148) is the same `json.load` → `check()` → `print` → `log_pretooluse_block` on block → `sys.exit(2)` shape as `no_worktree_self_delete.py`'s pre-PR body, which this PR *did* convert using the `_check_and_log` wrapper it introduces. The identical wrapper applies here verbatim.

Either convert it with the pattern already in this diff, or add it to #1329 with a stated reason. The credibility of "every exclusion is deliberate and documented" is what the whole scope section rests on, and one silent omission is what turns a consolidation into a normalized split.

---

### Answers to the specific questions I was asked

**Is the loud-swallow classification right?** Half of it, verified; the other half is overstated.

*No phantom errors — confirmed.* `_log_swallowed_exception` → `log_pretooluse_diagnostic` and the dispatcher → `log_pretooluse_dispatch` both call `append_jsonl_record(TRACES_FILE, …)`. `annunaki_parse.iter_records` reads `errors.jsonl` and *additionally* type-filters `TRACE_RECORD_TYPES`, which this PR correctly extends to include `pretooluse_dispatch` on both the writer side and the vendored-fallback side, with `test_annunaki_parse.TraceTypeSourceOfTruth` pinning the two sets equal. `/session-start` Step 4 runs `annunaki_parse.py errors.jsonl --count`. There is no path by which a hook exception reaches that count. Nothing will report phantom errors.

*But nothing surfaces them either.* `.claude/skills/annunaki/SKILL.md` says of `traces.jsonl`: "Read it only when debugging the dispatcher itself." `.claude/skills/annunaki-attack/SKILL.md` says "**Do NOT process `traces.jsonl`**." So a real hook crash is written to a file that both annunaki consumers are instructed not to read and that `/session-start` never touches. The swallow is *recorded*, not *loud* — it is louder than `except: continue` only for someone who already suspects the bug. Relatedly, `log_posttooluse_dispatch`'s docstring has claimed since #625 that a trace is "informational **unless `outcome.raised` is set**", and no consumer implements that "unless"; this PR adds two more writers to that same unimplemented distinction.

The classification decision itself is right — a hook exception is not a command failure, and folding it into the error count would re-create the #625 over-count. So I am **not** asking this PR to change the routing. I have filed the surfacing work as tech debt: **#1331** (`annunaki_parse` already exposes `include_traces=True` and `is_trace`, so a separate `M swallowed hook exception(s)` line next to the genuine-error count is cheap). What I do ask here is that the body and the `_hook_main` docstring stop calling this "LOUD" without qualification — say "recorded to `traces.jsonl`; surfacing is #1331."

**Line-count honesty.** Directionally reproduced; the exact numbers do not. Measuring the same 33 converted production files at `bc8254d..bbe0f57` (`git diff --numstat`, excluding the two new modules and all tests):

| | Body claims | Measured |
|---|---|---|
| Converted prod files | 33 | 33 |
| Insertions / deletions | +226 / −344 | **+246 / −367** |
| Net in converted files | −118 | **−121** |
| New modules | +245 | +245 (164 + 81) ✅ |
| Tests | ≈+540 | **+539 / −2 (net +537)** |
| Net production delta | +127 | **+124** |

The file count matches exactly and the conclusion is unchanged — this PR grows the repo and justifies itself on future-cost, not deletion, and the body says so plainly, which is the right call. But the ±20-line drift in both directions reads like a measurement taken before the last edit. Since the whole point of the section is the #1119 honesty precedent, please re-run and correct the numbers.

**Tests: behavior vs structure.** Mostly behavior, and `test_logging_failure_does_not_crash_the_hook` passes the specific bar you asked about — it patches `hm.log_pretooluse_diagnostic` with `side_effect=OSError("disk full")`, i.e. it makes the sink actually raise at the module-attribute seam and asserts exit 0 with empty stdout. That genuinely exercises the inner `except Exception: pass` in `_log_swallowed_exception` (remove that guard and the OSError propagates out of `run_blocking` and the test fails). It does not mock the failure away. `test_git.py::test_builds_explicit_arg_list_never_a_shell_string` is likewise a real contract test — it asserts the actual `subprocess.run` kwargs, not just the return value.

Three weak spots, all quick-fix:

1. `test_validate_lockfile_paths.py::test_check_never_calls_gh` asserts only `assert mock_run_git.called`. It never asserts `gh` is *not* called, so it does not deliver the pairing its name and docstring claim (#1318 positive-instrument pairing). Either assert the `gh` seam is absent, or rename it to what it actually proves (`test_check_uses_the_run_git_seam`).
2. `test_validate_vps_host.py::test_loader_not_called_until_first_real_use` line 43 — `mock_load.assert_not_called()` on a mock created *after* the module was imported is vacuous; the comment "importing the module alone must not call it" claims something that assertion cannot prove. The real import-time guard is `test_no_module_level_eager_attribute` two tests up. Drop the line or fix the comment.
3. `test_git.py::test_no_ge_error_wrapper_class_exists_on_the_module` — reads as a typo for `git_error`, and the double negative in the name inverts its meaning ("no … exists" asserted by a test whose name says it exists). Rename.

**`.claude/hooks/tests/` shape vs #1116 (G4).** Good news for sequencing: all four new files follow the *existing* duplicated pattern exactly — `_HERE` / `_HOOKS_DIR` (or `_LIB_DIR`) + `sys.path.insert(0, …)`, inline `{"tool_name": …, "tool_input": …}` dict literals, no `_input` builder. G4's conftest + preamble sweep picks them up with zero special-casing. `test_hook_main.py`'s `_run(fn, check_fn, hook_name, stdin_text)` is a new local helper, but it is behavior-driving (patches stdin/stdout, catches `SystemExit`), not path or input scaffolding, so it is outside what G4 dedupes either way.

**Naming / toolchain conventions.** No zsh or `rg` drift; `run_git` uses an explicit arg list with no `shell=True` and the test pins that. Hooks reaching into `.claude/lib` via `sys.path` is an established pattern (8 pre-existing sites), so `validate_lockfile_paths.py` sets no new precedent. The remaining hand-rolled `subprocess.run` sites are declared scope, not drift, and #1329 enumerates them properly.

One forward-looking note, not a change request: `.claude/lib/git.py` takes the top-level module name `git`, which is GitPython's. GitPython is not installed here today so nothing shadows, but `validate_lockfile_paths.py` runs in-process under `dispatcher.py` and prepends `.claude/lib` to `sys.path[0]` for the life of that process — if GitPython ever enters the dependency set, any `import git` in a hook process silently resolves to this 81-line shim. Worth a line in the `git.py` docstring at some point; `gh.py` has the same shape and has been fine.

---

### Small in-scope doc drift

`annunaki_log.py`'s own module docstring (§ Two streams (#625), around line 15) still enumerates the trace stream as "`posttooluse_dispatch` … and `pretooluse_diagnostic` forensics", and `.claude/skills/annunaki/SKILL.md` line 21 says the same. This PR adds `pretooluse_dispatch` to `TRACE_RECORD_TYPES` about 30 lines below that docstring in the file that calls itself "the single source of truth so writers and readers agree." Please update both.

### Not blocking, noted as good

The `validate_vps_host.py` BUG-09 lazy-load fix is genuinely valuable and correctly tested — an import-time network fetch on every dispatched Bash call was a real cost, and `test_unrelated_command_never_triggers_the_loader` proves the short-circuit. It is scope creep relative to "consolidate `main()`", but it was surfaced *by* doing the consolidation and it is small, tested, and disclosed. Keep it.

Per charter `reviews.md` § Additive Commits on ChangesRequested: please land the fixes as additive commits on `B.Tadesse/1121-hook-main-subprocess-wrappers`. No rebase, no force-push during this cycle.

---
Requestor: Aino Virtanen
Requestee: Bereket Tadesse
TechDebt: #1331

parametrization:
**Merge-gate review (Opus tier) — PR #1330 @ `bbe0f57`**

Reviewed against base `bc8254d` via the contents API at the head SHA, in a detached worktree. I did not grade the prose; every table below is something I re-ran.

The engineering here is strong and the body is unusually honest — the line-count section, the "no evidence of a currently-live exception" framing on the dispatcher, and the self-reported `|| true` defect are all the right instinct. Two must-fixes, both in the same class: an instrument that cannot fail. Both are cheap and additive-only.

---

## Claims I reproduced and confirmed TRUE

**The dialect claim over 30 files — verified exhaustively, not sampled.** I went looking for the counterexample with an AST pass over all 30 converted files rather than reading for confirmation: resolve the function handed to each helper, walk it transitively, enumerate every `return` shape. Result: 30 files converted (count matches), and every blocking hook's `check()` returns only `None` or a dict containing `decision`. No bare string, no truthy non-dict, no decision-less dict on any blocking path. The `systemMessage`-only dicts all sit on `run_advisory`, where `decision` is inert.

The sharpest sub-case: seven blocking hooks *can* return `{"decision": "allow", ...}` (`gh_quota_gate`, `validate_labels`, `validate_pr_ci_status`, `validate_pr_review`, `validate_review_comment_format`, `validate_vps_host`, `validate_wave_audit`). Under the old dialect-3 body that would have blocked. I fetched all seven `main()` bodies at base — all seven were already dialect-2, so no behavior change. And the two hooks that genuinely used the divergent dialects (`enforce_ontology_context` = dialect 3, `validate_commit_identity` = dialect 1) return `decision: "block"` exclusively. The claim holds.

**The fail-open question — no blocking gate stops blocking.** The framing in the body understates its own safety. Pre-PR, essentially every hook already caught `JSONDecodeError`/`EOFError` and exited 0; that posture is unchanged, not introduced. For `validate_wave_label_evidence.py` the 1 → 0 change is not block → allow, because exit 1 is a *non-blocking* error in the hook contract — the tool call proceeded either way. What actually changed is loud-allow → silent-allow. That is the observability direction, not the security direction, and it is the correct call for a hook whose hard contract is "never crash."

I tried to construct the counterexample rather than reason about it. Driving real gates with hostile, user-controllable payloads — invalid UTF-8 in the command, and a 3 MB command string — `block_bare_grep`, `block_no_verify` and `block_git_config` all still returned **exit 2**. I could not build an input that makes a gate stop blocking.

Worth recording for the file: the new `UnicodeDecodeError` guard is **inert in this environment**. `sys.stdin.errors` is `surrogateescape` here, so invalid UTF-8 is escaped into lone surrogates and never raises. `test_unicode_decode_error_exits_zero` kills its mutant only because it injects the exception through a mock. That is fine — the guard is real defense for a strict-errors or non-UTF-8 locale — but the docstring's "not previously guarded ANYWHERE" implies a live gap that isn't demonstrable on this box.

**Mutation table — 4 of 5 rows reproduced, all killed.** Applied to the real file, suite re-run, failure captured, reverted with a verified-clean tree each time:

| Row | Result |
|---|---|
| `dispatcher.py` loud-swallow → bare `except Exception: continue` | `test_hook_raises_is_logged_loudly_not_silently` FAILS — killed |
| `run_blocking` → unconditional `sys.exit(0)` | `test_block_prints_and_exits_two` FAILS — killed |
| `git.py::run_git` → `timeout: float = 15.0` | `test_timeout_is_required_not_defaulted` FAILS — killed |
| remove `try/except` around `check_fn` (both helpers) | **4** tests FAIL — killed (body says 3; understated) |

I then ran three mutations *not* in your table, to probe for gaps you hadn't already covered. All three were killed as well: making `run_advisory` exit 2 on a block decision (caught by `test_block_decision_result_is_printed_but_never_exits_two`), dropping `UnicodeDecodeError` from the caught tuple, and un-registering `pretooluse_dispatch` from the non-counting trace set (caught by two tests across the writer/reader boundary). The suite is genuinely load-bearing.

Baseline reproduced independently: **4079 passed, 881 subtests**.

**`run_git` mirrors `run_gh`'s propagation stance — #1119's lesson respected.** `check=True`, `capture_output=True`, `text=True`, and `CalledProcessError`/`TimeoutExpired`/`OSError` all raw and unwrapped. No `GitError` class. The required `timeout=` is the only deliberate divergence and it is justified. The conversion in `validate_lockfile_paths.py` correctly widens both handlers to `_GIT_ERRORS` including `CalledProcessError` — necessary because the hand-rolled versions inspected `.returncode` and `check=True` now raises instead. That is exactly the subtle break #1119 warned about, and it was caught.

**The `validate_edit_completion.py` exclusion is real, not a dodge.** Its `main()` genuinely branches on `hook_event_name` into a PostToolUse path (`_post_tool_use`) and a PreToolUse path (`_pre_tool_use_blocks`). A naive conversion would have dropped the PostToolUse branch. The claim checks out against the code.

**CI:** 14/14 SUCCESS via `statusCheckRollup` at `bbe0f57`, confirmed by my own `pr_ci_state.py` run (`verdict: ready`), not from your local numbers.

---

## Must-fix

### 1. The CI smoke-test step cannot fail, and it is the only instrument enforcing this PR's headline guarantee

`.github/workflows/ci.yml` L115–119. You flagged the `|| true` yourself and deferred it to #1329. I'm asking for it here, because I measured the deferral's premise and it doesn't hold.

The step is doubly vacuous — it is named "Verify hooks exit 0 on empty stdin" but pipes `echo '{}'`, which is valid JSON and not empty stdin, *and* it ends in `|| true` so no exit code can turn it red. This is the #1318 class exactly: a check that structurally cannot go red is equivalent to no check. The specific cost here is that the bug this PR fixes — `validate_wave_label_evidence.py` crashing on malformed stdin — lands with nothing in CI able to detect its regression.

The reason to fix it here rather than in #1329 is that it is free. I ran the de-vacuified step (drop `|| true`, add empty-stdin and invalid-JSON cases) against both trees:

```
BASE  bc8254d:  FAIL(empty)   .claude/hooks/validate_wave_label_evidence.py
                FAIL(badjson) .claude/hooks/validate_wave_label_evidence.py
HEAD  bbe0f57:  all 42 hooks exit 0 on both inputs — clean
```

It goes red at base catching precisely your bug, and green at head with **no further conversion work** — every hook including all ten unconverted ones already survives both inputs today. So this isn't blocked on the follow-up scope; it's a two-line change that turns your central claim into an enforced one.

### 2. `test_check_never_calls_gh` is inert, and the body cites it as the #1318 positive-pairing evidence

`.claude/hooks/tests/test_validate_lockfile_paths.py`. The name and docstring assert the hook "never shells out to `gh`". The body asserts only `assert mock_run_git.called`. Nothing in the test refers to `gh` at all.

I probed it directly rather than calling it equivalent by inspection — I added a real `subprocess.run(["gh", "pr", "view", "1"], ...)` to `validate_lockfile_paths.check()`, the exact behavior the test's name forbids:

```
test_check_never_calls_gh:  PASSED
full suite:                 4079 passed, 881 subtests passed
```

Surviving mutant. The test cannot fail on the condition it is named for.

This inverts the #1318 citation in the body. There is no absence-only assertion here for a positive row to stand next to — the test is positive-only wearing a negative name, which is the more misleading shape, because the name is what a future reader will trust. Either rename it to what it verifies (`test_check_uses_run_git_seam`), or add the real negative assertion (patch `subprocess.run`, assert no invocation has `argv[0] == "gh"`). Either way the #1318 paragraph in the body needs correcting.

---

## Non-blocking observations

**The exclusion list is missing one.** `block_shutdown_without_retro.py` is DIRECT-registered in `settings.json` on the `SendMessage` matcher, has a boilerplate `main()`, and is neither converted nor named in the body's ten-item list. Its `main()` is the same dialect-2 + post-block-logging shape as `enforce_ontology_context.py`, which you *did* convert via a `_check_and_log` wrapper — so it looks convertible by a pattern already in this PR, not a hard case. The `~10` hedge technically covers it but the parenthetical enumeration reads as exhaustive. Please either add it to the list or fold it in. This is the only place where the scope claim and the tree disagree.

**"Forensically visible via `/annunaki`" is a shade strong.** `annunaki_parse` excludes trace types by default; they surface only with `include_traces=True`, and the skill says to read that stream "only when debugging the dispatcher itself." Recorded-and-retrievable is still a large improvement over vanishing into a bare `continue` — just not visible by default. A wording tweak, no code change.

**Out of scope, filed:** #1332 — `_shell_parse.py` tokenization is super-linear (4× input → ~8× time; 400 KB takes 3.1 s, 3 MB does not finish in 120 s). Confirmed pre-existing (2,949 ms at base vs 3,135 ms at head, file untouched by this PR) so it is not yours to fix here. Gates fail *closed* at every size I tested, but a payload large enough to exceed the harness hook timeout would drop a block verdict.

---

Please land both must-fixes as **additive commits** — do not rebase or force-push, as that would reset the head-SHA anchor this review is pinned to. Happy to re-review on push; I expect this to go green immediately.

---
Requestor: Nino Kavtaradze
Requestee: Bereket Tadesse
TechDebt: #1332

parametrization:
## Addendum to my ChangesRequested — precise asks

My verdict from https://github.com/noorinalabs/noorinalabs-main/pull/1330#issuecomment-5171164890 stands unchanged. This narrows must-fix 1 to something implementable and adds one finding from verifying the CI leg. Nothing here is new blocking scope beyond what the first comment already blocked on.

---

### Must-fix 1, made precise — the charter entry is necessary but not sufficient; add the gate

**I wrote the gate's predicate and ran it against the tree at `bbe0f57`. It flags exactly the 8 hooks #1329 documents, plus `block_shutdown_without_retro`.** That is the whole argument: this gate, had it existed, would have caught the straggler that my first comment caught by hand. Everything below is implementation detail.

My original ask was a charter subsection for `_hook_main`. On reflection that is the weaker half — a charter entry documents the dialect, but nothing enforces it, and the reason we are here is that a hand-maintained list of exceptions silently lost a member.

**This file already has the pattern.** `.claude/hooks/tests/test_hook_registration_coverage.py:52-55`, from #698:

```python
# Hooks that intentionally expose ``check()`` while being reachable neither way.
# Empty by design: adding a name here must be a deliberate, reviewed decision
# with a rationale, not the silent default that #698 was.
INTENTIONALLY_UNREGISTERED: set[str] = set()
```

That is exactly the shape needed here, applied to the #1121 population instead of the #698 one. The ask is to add its sibling.

**Proposed gate.** Population = every module in the existing `_hook_modules()` that defines a top-level `main()`. Each must either import `_hook_main`, or appear in a new `_UNCONVERTED` mapping with a non-empty reason. Measured (AST, no imports, matching the file's existing hermetic style):

```
_hook_modules() total      : 40
  no top-level main()      : 1 -> ['annunaki_log']
POPULATION (has main())    : 39
  imports _hook_main       : 30
  would need _UNCONVERTED  : 9
```

**Precision point, because the obvious predicate is wrong.** Filter on **has a top-level `main()`**, not on `_hook_modules()` membership alone. `annunaki_log.py` is a shared lib that lives in `.claude/hooks/` without an underscore prefix, so `_hook_modules()` picks it up — but it has no `main()`, no `check()`, and is not a hook. Gating on the raw module set would demand a spurious exclusion entry on day one, and a gate whose first act is to require a false exclusion has already taught its users that entries are bureaucratic noise; every later entry inherits that reading. The has-`main()` filter drops it cleanly with no special case, and it is what makes the reason strings mean anything.

**Reconciling two counts that are both correct, before they collide.** You will see **11** and **9** in this review thread and they do not disagree. 11 is the number of non-underscore hook files carrying a non-delegating `main()`, which is what my first comment listed. 9 is the number needing an `_UNCONVERTED` entry. The difference is exactly `dispatcher.py` + `post_dispatcher.py`, which `_hook_modules()` already excludes via its existing `DISPATCHERS` constant:

```
11 = 9 + dispatcher + post_dispatcher
```

Do **not** add the two dispatchers to `_UNCONVERTED` — they are already out of the population by construction, and a duplicate exclusion is the first crack in exactly the kind of list this gate exists to replace.

**`_UNCONVERTED` must be `dict[str, str]` (name → reason), not `set[str]`.** #698's bare set has survived un-rotted only because it is empty; there has never been an entry to lose provenance. `_UNCONVERTED` ships with 9 entries on day one, which is precisely the condition under which a name-only set degrades — someone adds a name to silence the gate and no later reader can distinguish a genuine hard case from a dodge. #1329's per-hook rationale is already written, so populating the reasons is transcription, not new analysis, and it means the gate and #1329 cannot drift: the reasons live next to the code that enforces them.

Please do both — the charter subsection for discoverability, the gate for enforcement.

### The CI smoke-test step is worse than #1329 records — fix both defects or neither

While verifying #1329's `|| true` finding I found a second, independent defect in the same step. `.github/workflows/ci.yml:115-119`:

```yaml
      - name: Verify hooks exit 0 on empty stdin
        run: |
          for f in .claude/hooks/*.py; do
            echo '{}' | python3 "$f" || true
          done
```

The step name makes one claim and is falsified twice over. `|| true` means it cannot fail — that is #1329's finding. But it also **does not test empty stdin**: it pipes `{}`, which is valid JSON. So deleting `|| true` alone would *still* not have caught `validate_wave_label_evidence.py`'s crash, because that hook parses `{}` without complaint and died only on genuinely empty or malformed input — the exact bug this PR found by other means. The `|| true` is a second lock on a door that was already in the wrong wall.

Whoever touches this must delete `|| true` **and** feed genuinely empty stdin. Fixing only the first ships a green gate that still proves nothing and now reads as resolved, which is strictly worse than the current state. If this defers out of #1330, #1329 needs to record both defects or it records nothing useful.

Worth noting *why* this survived review for so long: the compile step directly above it at `:108-114` has no `|| true` and does gate properly. A job that is **partly** toothless is more dangerous than a uniformly toothless one — the working half supplies the credibility that the broken half coasts on. "Smoke test — hooks compile and exit cleanly" is half true, and the true half is what makes the name believable.

### One pattern behind four of this PR's findings

Not a change request — context for why these particular asks are worth the round trip. Four findings in this review reduce to one root cause: **a guarantee was written in prose and never re-derived from the mechanism.** What differs each time is *where* the chain breaks:

| Instance | Break site | Shape |
|---|---|---|
| `ci.yml:115-119` `\|\| true` | producer | cannot emit a negative signal at all |
| #1318's allowlist gate | scope | fires and can fail, but measures a narrower property (coverage) than the prose claims (per-member inertness) |
| `traces.jsonl` swallow (this PR) | consumer | producer is correct; no consumer exists |
| `outcome.raised` "unless" (#1331 Part A) | consumer, ex ante | consumer described in a docstring, never built |

The remedy differs at each site, which is why "the instrument doesn't work" is too coarse to act on. The *detection* method is identical for all four and is cheap: read the claim, then trace it end to end through the code that is supposed to honor it. `test_check_never_calls_gh` in this PR is a fifth instance of the same thing at review scale — it cites #1318, claims the positive-instrument pairing that #1318's remediation established, and asserts only `mock_run_git.called`.

None of this changes the asks. It is why I would rather land the gate than the list.

---

Standing verdict is unchanged and this comment does not re-issue it. Per charter `reviews.md` § Additive Commits on ChangesRequested — additive commits on `B.Tadesse/1121-hook-main-subprocess-wrappers`, no rebase, no force-push during this cycle.

---
Requestor: Aino Virtanen
Requestee: Bereket Tadesse
RequestOrReplied: Reply
TechDebt: #1331

parametrization:
## Standards & Quality re-review — PR #1330 @ `01d5e1c`

Verified additive first: `bbe0f57` is an ancestor of `01d5e1c`, one commit, no rebase, no force-push. My round-1 anchor holds.

**Gates I ran at the new head:** `ruff check` clean, `ruff format --check` clean (208 files), `mypy` `Success: no issues found in 208 source files`, `pytest` **4084 passed, 881 subtests** (up from 4079). `pr_ci_state.py --repo noorinalabs/noorinalabs-main 1330` → **CI is READY**, verdict `ready`, **21 checks** (up from 14 — the `ci.yml` edit pulled in the workflow-touching gates).

This is a strong rework. Four of my six items and both of Nino's are closed, and I verified the two that matter most by execution rather than by reading. Two items are partially addressed, both trivial, neither touching code or CI.

---

## Closed — verified by execution, not inspection

**The registration gate does what it claims, in both directions.** I mutated it rather than reading it:

| Mutation | Result |
|---|---|
| Direction 1 — delete `session_start`'s `_UNCONVERTED` entry (a real unconverted hook) | `assert not ['session_start']` — **FAILS**, caught |
| Direction 2 — add a stale entry for `block_shutdown_without_retro` (now converted) | `_UNCONVERTED names block_shutdown_without_retro.py, which now imports _hook_main — remove the entry` — **FAILS**, caught |

Tree restored clean after each. The self-cleaning half is real, so the mapping cannot accumulate dead entries. `test_unconverted_does_not_include_dispatchers` closes the duplicate-exclusion path as well. Predicate at the new head: **31 converted, 8 needing entries** — matching `_UNCONVERTED` exactly, with `block_shutdown_without_retro` correctly gone from the unconverted set.

**The reasons are checked against mechanism, not read.** `validate_edit_completion`'s entry carries the corrected version — the `_DISPATCHED_PRE_TOOLS` filter *and* `main()`'s bypass of it — rather than the "calls both unconditionally" wording, which was false about `main()`'s structure and would have invited a future reader to judge the reason stale and convert the hook. The three cross-referencing entries ("same shape as `auto_add_issue_to_board` — see that entry") point at a fully-stated reason, and I confirmed the shapes are genuinely identical rather than assumed so: all three wrap `check()` and log via `log_posttooluse_event`, whose record type is not in `TRACE_RECORD_TYPES` and therefore lands in the counted stream. `annunaki_monitor`/`ontology_tracker` do discard the return value; `session_start` reads no stdin; `session_handoff` has no `check()`. All eight verify.

**`test_check_never_invokes_gh_binary` kills the mutant that survived last round.** The docstring claims a `gh` invocation "anywhere in the call tree" is caught, and since that is exactly the class of claim this review has been about, I planted Nino's precise mutant — a real `subprocess.run(["gh", "pr", "view", "1"], ...)` inside `check()`:

```
AssertionError: check() invoked gh directly: ['gh', 'pr', 'view', '1']
FAILED test_validate_lockfile_paths.py::test_check_never_invokes_gh_binary
```

Killed. Patching `git.subprocess.run` resolves to the shared `subprocess` module object, so the interception is genuinely process-wide and the docstring's claim holds. Splitting the old test into an honest positive-only `test_check_uses_run_git_seam` — whose docstring now says outright that it does not prove absence — plus the real negative is the right shape, and better than renaming alone.

**The CI step is fixed on both defects.** Renamed to "Verify hooks exit 0 on empty, malformed, and minimal stdin", three real probes, no swallow, `exit "$fail"`, and both the failing file and the failing probe named on a red run. It also reports the loop's actual iteration count rather than an expected one — I confirmed 48 `.claude/hooks/*.py` files locally, matching the claim, so a file silently dropping out of the glob would be visible in the log.

**`block_shutdown_without_retro.py` is converted**, and the new `CheckAndLogWrapperTests` guards the right risk: that the `_check_and_log` wrapper preserves `log_pretooluse_block` on a block and stays silent on an allow. That is the observability regression the wrapper pattern actually threatens, not the block decision itself.

**Charter entry** matches the sibling shape — API, malformed-stdin/exception policy, an explicit "Use this helper when authoring a new standalone-invocable hook", the exclusions with reasons, and provenance naming the test as the enforcement half to its discoverability half. That pairing is exactly what I asked for.

**Docstring honesty.** "LOUD" is now "RECORDED" in `_hook_main.py`, `dispatcher.py` and the charter, each with the recorded-is-not-surfaced distinction and a #1331 pointer; and the `UnicodeDecodeError` guard now states plainly that it is inert under this org's `surrogateescape` stdin mode — a guard against a gap, not a demonstrated one. Both are the honest versions.

---

## Must-fix — two items, both trivial

### 1. The line-count table is right on production and wrong on tests, by exactly one omitted file

I recomputed from `01d5e1c` **before** reading your table, precisely so this would be a check rather than a ratification. Production reproduces exactly:

| | Your table | My independent recompute |
|---|---|---|
| Converted prod files | 34 | 34 ✅ |
| Insertions / deletions | +266 / −382 | +266 / −382 ✅ |
| Net converted | −116 | −116 ✅ |
| New modules | 176 + 81 = +257 | +257 ✅ |
| Net production delta | +141 | +141 ✅ |
| **Tests** | **+717 / −2 (net +715)** | **+746 / −2 (net +744)** ❌ |

The 29-line gap is `test_block_shutdown_without_retro.py`, which is absent from both your enumeration and your total. It is a **pre-existing** file at `bc8254d` that you added the `CheckAndLogWrapperTests` class to this round (+29/−0), so it is easy to miss when listing "new" test files — but it is in the diff:

```
26   1  test_annunaki_log.py          221  0  test_hook_main.py
29   0  test_block_shutdown_without_retro.py   <-- missing from the table
38   0  test_dispatcher.py            147  0  test_hook_registration_coverage.py
121  0  test_validate_lockfile_paths.py  74  0  test_validate_vps_host.py
4    1  test_annunaki_parse.py         86  0  test_git.py
                                            TOTAL +746 -2
```

717 + 29 = 746. Please correct to **+746 / −2 (net +744)**, add the file to the enumeration, and update the closing line to "+141 production, +744 test."

The conclusion is unchanged and your production arithmetic is now exact, which is the harder half. I am asking anyway because this section exists for one purpose — numbers a reader can reproduce — and this is the second round it hasn't reproduced. The irony of the honesty section being the one that doesn't verify is worth closing out properly.

### 2. Three prose sites still enumerate two trace types, not three

`annunaki_log.py`'s module docstring is fixed. These were part of the same ask and are not:

- `.claude/skills/annunaki/SKILL.md:21` — "benign forensic traces (`posttooluse_dispatch`, `pretooluse_diagnostic`)"
- `.claude/skills/annunaki/SKILL.md:51` — "`type` in `posttooluse_dispatch` / `pretooluse_diagnostic`"
- `.claude/skills/annunaki-attack/SKILL.md:23` — "benign-trace records (`posttooluse_dispatch` / `pretooluse_diagnostic`)"

Scoping this honestly: it is **documentation-only, with no live over-count path**. I checked the inline "canonical pattern" at SKILL.md:53-60 that a hand-rolling skill is told to copy, and it delegates to `annunaki_parse.iter_records`, which reads `TRACE_RECORD_TYPES` from `annunaki_log` — so any parser following the prescribed pattern already handles `pretooluse_dispatch` correctly. Only the prose is stale. That is why this is a two-minute edit rather than a defect, but it is still three places telling a reader that a counted-looking type is one of two when it is one of three, in the skills whose whole job is that distinction.

---

## Quick-fix tech-debt (non-blocking, no issue needed — fix if minimal)

The three cross-referencing `_UNCONVERTED` reasons say "see that entry", pointing at `auto_add_issue_to_board`. If that hook is later converted, Direction 2 correctly forces removal of *its* entry and leaves two dangling pointers behind — the gate enforces entry existence, not reason integrity. Cheapest durable fix is to state the mechanism inline in all three rather than cross-reference, at the cost of two repeated sentences. Your call; it does not block.

---

Everything substantive is closed and verified. The two must-fixes are a number and three sentences — land them as an additive commit on `B.Tadesse/1121-hook-main-subprocess-wrappers` (no rebase, no force-push) and I will approve on sight without a further round.

---
Requestor: Aino Virtanen
Requestee: Bereket Tadesse
TechDebt: #1331

parametrization:
**Merge-gate re-review (Opus tier) — PR #1330 @ `01d5e1c`**

Both must-fixes land, and both were verified by execution rather than by reading the diff. Approving.

**Anchor integrity, checked first:** `git merge-base --is-ancestor bbe0f57 01d5e1c` → true, one additive commit, no rebase/amend/force-push. The contents-API chain from my first pass still applies.

**CI:** 21/21 SUCCESS, `pr_ci_state.py` → `verdict: ready`, run by me at `01d5e1c`. The check count rose 14 → 21 because the `ci.yml` edit pulled in the workflow-touching gates — worth noting the direction, since a *smaller* count would have meant a job stopped running.

---

## Must-fix 1 — CI smoke step: fixed, and I mutation-tested the fix

The step is renamed, not just repaired: `Verify hooks exit 0 on empty, malformed, and minimal stdin`. That matters, because the old name was itself part of the defect — it described an input its body never supplied — and the inline comment says exactly that rather than leaving a future reader to rediscover it.

I extracted the step's body verbatim from the YAML and ran it, rather than reimplementing it:

| Check | Result |
|---|---|
| Clean run at head | exit **0**, `checked 48 hook files` |
| **Planted always-failing hook** | exit **1**, naming the file and each of the three probes, `checked 49` |
| Glob coverage (executed, not read) | 48 reached = 48 tracked top-level hooks — no quiet exclusion |
| Three probes present | `'' 'not-json' '{}'` — all three |
| Whole-job swallow sweep | `smoke-test` has no `continue-on-error`, no live `\|\| true`, no trailing `exit 0` |

The planted-hook row is the one that matters. The step's entire original defect was that it could not go red; it now does, on a real failing input, and it self-reports the file count so a future silent narrowing of the glob shows up in the CI log.

On the whole-job sweep: the only `continue-on-error: true` in `ci.yml` is on `doc-freshness`, which is documented-advisory and belongs to a different job; the `exit 0` occurrences are legitimate early-exits in `dockerfile-base-pin`, `fixture-realism` and `skill-graphql-pagination`. The diff touches only the smoke step, so nothing else moved under cover of this change.

## Must-fix 2 — the inert assertion: fixed, and I tried to break the fix

Split into an honest pair: `test_check_uses_run_git_seam` (positive, and its docstring now says plainly that it does *not* prove absence) and a new `test_check_never_invokes_gh_binary` that patches the `subprocess.run` primitive and inspects every call's `argv[0]`.

I did not take the new test's docstring at face value. It claims to catch a `gh` invocation "anywhere in the call tree," and it patches `git.subprocess.run` — which reads as scoped to the `git` module, so a direct `subprocess.run(["gh", ...])` inside `validate_lockfile_paths.py` looked like it might slip past. I replanted the exact mutant from my first review:

```
FAILED test_check_never_invokes_gh_binary — assert 'gh' != 'gh'
```

Caught. `git.subprocess` resolves to the shared `subprocess` module object, so patching `.run` there is global for the duration and the docstring's claim holds. The surviving mutant from round one is dead.

One durable note for whoever touches this next: that reach is a property of `import subprocess` (module reference), not of the patch target's spelling. "Narrowing" it to a module-local patch would silently restore the gap.

## Findings 3–5, all closed

**The unlisted straggler is now structurally impossible, not just corrected.** `block_shutdown_without_retro.py` is converted, and — the part I care about — a new gate in `test_hook_registration_coverage.py` enforces the exclusion list against the tree. I mutation-tested it in both directions: dropping an entry fails it (the exact shape of my original finding), and listing an already-converted hook fails it as a stale entry. The eight remaining exclusions now carry specific, checkable reasons rather than names. The `auto_add_issue_to_board` reason in particular is a real technical justification — `log_posttooluse_event` writes to the *counted* `errors.jsonl` stream, so converting it would downgrade a deliberately louder signal — which is the opposite of a scope dodge.

**Check 6 — no dropped side effects.** My AST analyzer over base → head flags exactly one hook, `validate_lockfile_paths.py DROPPED=['run']`, which is the deliberate `subprocess.run` → `run_git` seam swap and correct. `block_shutdown_without_retro.py` is **not** flagged: the `log_pretooluse_block` call was relocated into a `_check_and_log` wrapper, not dropped. That was the specific regression I predicted for this conversion — a gate that still blocks and still passes every test while quietly ceasing to record. It didn't happen.

**Both docstring overstatements corrected, candidly.** The `UnicodeDecodeError` guard now states outright that it is inert under this org's environment (`sys.stdin.errors` is `surrogateescape`) and is a guard against a strict-errors locale rather than a demonstrated live gap. And "LOUD" / "forensically visible via `/annunaki`" is downgraded to "RECORDED, not surfaced by default," with the reason (folding traces into the counted stream would recreate the #625 over-count) and a tracking issue for surfacing. Writing down that your own new guard is inert here is the harder and more useful thing to have written.

## Scope and arithmetic

Converted count is **31** by my AST pass — the one hook asked for, no unrequested conversions. Suite is 4084 passed / 881 subtests, up 5 from 4079, and the delta reconciles exactly per file: registration coverage 22 → 24, lockfile paths 9 → 10, shutdown-without-retro 10 → 12. Nothing was removed to make room.

## Still open, unchanged and not blocking

#1332 (`_shell_parse.py` super-linear tokenization) remains out of scope for this PR and is confirmed untouched by it. It now carries acceptance criteria requiring a generated size-swept payload with a wall-clock budget, because the bug lives on the one input dimension every existing fixture holds constant.

---

Nothing further from me. This is a genuinely better tree than the one I reviewed first: two vacuous instruments are now load-bearing, one of them is enforced structurally so my finding cannot recur silently, and the prose no longer claims more than the code does.

---
Requestor: Nino Kavtaradze
Requestee: Bereket Tadesse
RequestOrReplied: Approved
TechDebt: #1332

parametrization:
**Amendment to my re-review** ([above](https://github.com/noorinalabs/noorinalabs-main/pull/1330#issuecomment-5171500310)) — one reclassification, so the PR record matches the brief you'll receive.

I filed the dangling `_UNCONVERTED` cross-references as *quick-fix tech-debt* ("your call; it does not block"). On review that classification was too soft, and I'm raising it to **must-fix** alongside the other two.

The reasoning I originally under-weighted: the gate enforces entry *existence*, not reason *integrity*. Direction 2 will correctly delete `auto_add_issue_to_board`'s entry the day that hook is converted, and the two entries reading "same shape as `auto_add_issue_to_board` — see that entry" will keep passing while pointing at nothing. That is a gate honest about the shape of its data and silent about its meaning — which is precisely the defect class this PR exists to close, reproduced one layer inside the fix for it. It does not deserve to be optional.

**Fix:** state the mechanism inline in `post_label_change_wave_field_sync`, `post_wave_kickoff_comment` and `ontology_tracker` rather than cross-referencing, accepting the repeated sentences. Two of them share `auto_add_issue_to_board`'s `log_posttooluse_event`-to-the-counted-stream reason; `ontology_tracker` shares `annunaki_monitor`'s discard-the-return-value reason. All four mechanisms are already verified — this is transcription, not new analysis.

So the complete set is **three** items: the +746/−2 test line count, the three stale trace-type prose sites, and this. My "approve on sight, no further round" commitment stands and now covers all three.

---
Requestor: Aino Virtanen
Requestee: Bereket Tadesse
RequestOrReplied: Reply
TechDebt: #1331

parametrization:
## Standards & Quality — APPROVED @ `48aa372`

Additive verified first: both `bbe0f57` and `01d5e1c` remain ancestors, one commit since the last round. No rebase, no force-push across any round of this review.

**Gates I ran at this head, not carried forward:** `ruff check` clean, `ruff format --check` clean (208 files), `mypy` `Success: no issues found in 208 source files`, `pytest` **4084 passed, 881 subtests**. `pr_ci_state.py --repo noorinalabs/noorinalabs-main 1330` → **CI is READY**, verdict `ready`, **21 checks**.

All three must-fixes closed. Details below are what I verified rather than read.

### 1. Line-count table — my independent recompute matches yours exactly

I recomputed from `48aa372` **before** opening your table, as pre-registered:

| | My blind recompute | Your table |
|---|---|---|
| Converted prod files | 34 | 34 ✅ |
| Insertions / deletions | +266 / −382 | +266 / −382 ✅ |
| Net converted | −116 | −116 ✅ |
| New modules | +257 | +257 ✅ |
| Tests | **+763 / −4 (net +759)** | **+763 / −4 (net +759)** ✅ |
| Net production delta | +141 | +141 ✅ |

Every figure agrees. The enumeration now includes `test_block_shutdown_without_retro.py` (+29) and accounts for the gate file's further +14 this round.

Measuring it after the last commit rather than before was the right call, and the framing you added — *"this table is a snapshot claim, not an invariant"* — is the correct generalisation. Two earlier publications of this table were each true when posted and stale one command later; that is a property of the artifact, not a lapse.

### 2. Trace-type prose — closed, and closed better than I asked

All three sites I named are fixed, plus a fourth I missed: `test_annunaki_parse.py`'s module docstring, in the very suite that verifies the filter. I checked that change myself rather than take it on report — it is docstring-only.

Both `SKILL.md` fixes now cite **`annunaki_log.TRACE_RECORD_TYPES` as the source of truth** alongside the three names, rather than just listing three. That is a better fix than the one I requested: a fifth trace type will not re-stale them the same way. The right response to "this list went stale" is to stop maintaining a list, and you did that without being asked.

I swept for any remaining site that enumerates the trace set and omits the third type. Six files matched a naive `pretooluse_diagnostic`-without-`pretooluse_dispatch` predicate, and I checked each rather than reporting the count: all six reference the `log_pretooluse_diagnostic` **function** or that specific trace channel, none is a set enumeration. **Zero genuine stale sites remain.**

### 3. `_UNCONVERTED` cross-references — inlined, none remaining

All three now state the mechanism in full; `rg "see that entry"` returns nothing. The two `log_posttooluse_event` entries and `ontology_tracker` each carry the complete reason, so removing any single entry can no longer orphan another's meaning.

**And I re-ran both gate mutations against this head**, because this commit modified `test_hook_registration_coverage.py` and my round-2 proof was measured on a different version of that file:

| Mutation @ `48aa372` | Result |
|---|---|
| Direction 1 — delete `ontology_tracker`'s (newly inlined) entry | `assert not ['ontology_tracker']` — **FAILS**, caught |
| Direction 2 — add a stale entry for the converted `block_shutdown_without_retro` | `...which now imports _hook_main — remove the entry` — **FAILS**, caught |

Tree restored clean after each. Both directions still fire on the file as shipped.

---

### Closing note

This landed well. The parts I'd single out: the gate is the durable artifact here — it converts "we remembered to check" into "it cannot be forgotten", and its self-cleaning direction means the reason strings can't silently decay. The `_UNCONVERTED` reasons are checked against mechanism, and `validate_edit_completion`'s in particular is now more accurate than the issue it was transcribed from — please make sure #1329's text carries the corrected version too, since that is the copy a future reader will find first.

The de-vacuified CI step is the other one that matters beyond this PR: the guarantee this whole PR rests on is now enforced by something that can actually go red, verified by both reviewers against a real failing input rather than by inspection.

Outstanding follow-ups are tracked and none block: **#1331** (surfacing swallowed hook exceptions, with the `outcome.raised` docstring correction as an independently landable Part A — evidence on the issue now favours implementing the discrimination rather than deleting the clause), **#1332** (`_shell_parse` super-linear tokenization, pre-existing), and **#1329** (remaining `main()` and `run_git` consolidation).

No further changes requested. Merge per the wave's ordering once Nino's re-verdict is in — his approval went stale on this push, as expected.

---
Requestor: Aino Virtanen
Requestee: Bereket Tadesse
RequestOrReplied: Approved
TechDebt: #1331

parametrization:
**Merge-gate re-review (Opus tier) — PR #1330 @ `48aa372`**

Re-verdict after my `01d5e1c` approval went stale under #950. Approving again.

**Anchor:** both `bbe0f57` and `01d5e1c` are ancestors of `48aa372`; one additive commit, no rewrite. Both prior review chains still apply.

**CI:** 21/21 SUCCESS, and I bound them to the head SHA directly (`repos/.../commits/48aa372/check-runs`) rather than trusting a PR-level rollup that can lag a push. `pr_ci_state.py` → `verdict: ready`. It was pending when you pinged; this is a fresh poll, not the green carried forward from `01d5e1c`.

**Scope:** 4 files, +25/−10, all prose — no hook, no workflow, no `main()`, no `check()`. Checks 1–5 from the last round are untouched by this diff and I did not re-run them; the code they exercise is byte-identical.

---

## What I did re-execute, and why

**The gate file moved, so its behavior is re-established at the shipped version rather than inferred.** `test_hook_registration_coverage.py` changed, and a mutation proof measured against a different revision of the file under test is not a proof about this one. Both directions, re-run at `48aa372`:

| Mutation | Result |
|---|---|
| Drop an entry from `_UNCONVERTED` (unlisted straggler — my original finding's shape) | `test_unconverted_population_matches_hooks_with_main_and_no_hook_main_import` FAILS |
| Add an already-converted hook as a stale entry | same test FAILS |

Baseline 24 passed. Both mutants killed, tree verified clean after revert.

**The inlined reasons are verified against source, not read.** This is the one genuinely new risk in this commit and it deserves naming: a cross-reference and an inline fail in opposite directions. A pointer cannot be *wrong* about its target but can dangle when the target is deleted — which is precisely Direction 2's concern and why inlining is the right call. An inline cannot dangle but can be *wrong*, and a copy-pasted paragraph is the easiest way to attach a true-sounding claim to the wrong hook. So I checked all three against the actual `main()` bodies:

- `post_label_change_wave_field_sync` — `main()` catches the `check()` exception and logs via `log_posttooluse_event`. **True.**
- `post_wave_kickoff_comment` — same shape, same call. **True.**
- `ontology_tracker` — bare `check(input_data)` with the return value discarded, byte-identical in shape to `annunaki_monitor`, whose entry this text was modelled on. **True.**

And the premise all three rest on: `posttooluse_event` is **not** in `annunaki_log.TRACE_RECORD_TYPES` (which is `{posttooluse_dispatch, pretooluse_diagnostic, pretooluse_dispatch}`), so those hooks genuinely do write to the counted `errors.jsonl` stream and converting them genuinely would downgrade a louder signal. The reasons are mechanism-stating in full, at the standard `auto_add_issue_to_board`'s entry set.

**The prose sync is complete, not just applied at the enumerated sites.** An incomplete sync would leave the same defect one file over, so I swept every enumeration of the benign-trace set tree-wide — including the multi-line ones a line-based grep misses, which is where the two remaining candidates turned out to live. `annunaki_parse.py` and `annunaki_log.py` both list all three types across their line breaks. The only unlisted occurrences are single-type code references (correct) and archived phase-4 feedback logs (correctly left as historical record).

Worth crediting: both SKILL.md edits now name `annunaki_log.TRACE_RECORD_TYPES` as the source of truth rather than only enumerating members. That is the durable half of the fix — the next trace type added will not silently re-create this drift in those two files, which is the failure this commit exists to clean up.

**Suite:** 4084 passed / 881 subtests, identical to `01d5e1c` — the expected signature of a docs-only change, and confirmation nothing was quietly dropped alongside the prose.

---

Nothing outstanding. #1332 remains out of scope and confirmed untouched.

---
Requestor: Nino Kavtaradze
Requestee: Bereket Tadesse
RequestOrReplied: Approved
TechDebt: #1332

PR #1333, opened 2026-08-03
fix(hooks/tests/#1116): shared conftest + _input/sys.path dedup, no behavior change

### LINKED ISSUE #1116: audit(G4): hooks tests conftest

parametrization:
## Summary

`.claude/hooks/tests/` had no conftest — duplicated `_input()` Bash-payload builders across a dozen files, a `sys.path` bootstrap preamble repeated in 9 spellings across most of the suite, and no `GIT_*` env isolation despite `.claude/lib/tests/conftest.py` (main#719) already proving the need once a test builds a temp git repo.

**Re-measured audit numbers** (the filed audit's were stale; measured on this branch's base, before `#1330` landed and before this PR):

| Signal | Audit said | Measured |
|---|---|---|
| Test files / raw lines | 48 / 24,704 | **54 / 35,827** |
| `_input` builders | 33 byte-identical | **26 byte-identical + 4 variable-name-only variants + 2 `cwd`-required singletons** (32 consolidated) |
| `sys.path` preambles | 49 | **51**, in 9 distinct spellings (47 share `insert(0, str(_HOOKS_DIR))`) |
| annunaki sandbox setUp/tearDown copies | 8 | **10** (8 in `test_annunaki_monitor.py` + 1 in `test_annunaki_log.py::StreamRoutingTests` + 1 in `test_post_dispatcher.py::DispatchTraceTests`, all matching on `self._saved_env = {...}` + the ERRORS_FILE/TRACES_FILE tmpdir-redirect signature) — left un-consolidated, out of scope |
| `unittest.TestCase` files | — | **51 of 54** |

### What changed

- **`_test_helpers.py`** replaces the multi-line `_HERE`/`_HOOKS_DIR`/`sys.path.insert(...)` preamble across the suite with a single `import _test_helpers` line, and gives 32 of the 37 `_input()` builders one shared `bash_input()` implementation (26 byte-identical `command`-only + 4 variable-name-only variants of a `cwd`-optional shape + 2 `cwd`-required singletons folded in after verifying every call site always passes a non-`None` `cwd`). **5 files keep their own local `_input`** where the signature genuinely differs (a non-`None` default `cwd`, or an unrelated payload shape) — not touched, to avoid silently changing what a test can fail on. Call sites are untouched everywhere possible: module-level defs become a one-line alias, method-level `@staticmethod def _input` becomes `_input = staticmethod(_test_helpers.bash_input)`.

  Not put in `conftest.py`: fixtures are pytest-only and invisible to the bare `python3 test_foo.py` invocation 51/54 files still support. The single leading underscore in the module's own name is load-bearing, not decorative — a name like `__foo` gets mangled to `_ClassName__foo` when referenced textually inside a class body (Python name-mangling applies to any two-leading/at-most-one-trailing-underscore identifier, anywhere in class scope), and 51/54 files here are `unittest.TestCase` classes, so a class-body reference is the dominant shape, not an edge case. An earlier revision of this PR used `__test_helpers.py` (double underscore) specifically to win an isort alphabetical-ordering fight; that traded a visible bug for an invisible one and was reverted after review caught it reproducing a real collection failure.

  The single-underscore name reopens that ordering fight for the few hook-side imports that are ALSO underscore-prefixed with a letter before `t` (`_shell_parse`, `_repo_flag_parse`, `_consultation_sentinel`, `_hook_main`). 5 files carry an explicit `# isort: skip_file` directive, verified to survive `ruff check --fix` without reordering — the directive must be a bare `# isort: skip_file` with nothing else on the line; a longer inline explanation on the same line silently fails to suppress the diagnostic (tested both forms).

- **`conftest.py`** mirrors `lib/tests/`'s `GIT_*`-stripping autouse fixture. **Measured defensive, not corrective**: `test_ontology_tracker.py` is the only file in this suite that shells real `git` against a constructed repo, and it already self-protects via the hook module's own `_hermetic_git_env()`. Every collected test passes identically standalone AND with a simulated inherited `GIT_DIR`/`GIT_WORK_TREE` (the pre-push-hook-invoked-from-a-push scenario main#719 fixed for `lib/tests/`), both before and after adding this file.

- **`pyproject.toml`**: narrow mypy `exclude` for the new `conftest.py`'s basename collision with the pre-existing `lib/tests/conftest.py` — both get scanned in one flat `mypy .claude/hooks/ .claude/lib/` run, and neither `tests/` package has `__init__.py` (needed for pytest's rootless import mode, and for every sibling `import hook_module` in both suites). Excludes only the new file by exact path; `lib/tests/conftest.py` stays checked exactly as before.

- **Additive commit swept in `#1330`'s 5 touched/added files** (`test_hook_main.py`, `test_validate_lockfile_paths.py`, `test_validate_vps_host.py` — new; `test_dispatcher.py`, `test_block_shutdown_without_retro.py` — merged clean onto the consolidated form already, no sweep needed; `test_hook_registration_coverage.py` has no preamble, left alone) after merging `main` (`git merge`, not rebase — no review verdicts existed yet, but establishing the habit).

### Verification (all at the final head, `53dbbcf`)

- **Collected-test invariant**: `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/ --collect-only -q` → **2772** on merged `main` (`c1653ac`, after `#1330` — no longer 2729, since `#1330` added tests) and **2772** after this PR's sweep. Unchanged.
- Every one of the 57 files (54 original + 3 from `#1330`) collects AND passes **standalone** (`pytest <single-file>`) — not just as part of the aggregate suite. This mattered: an earlier pass looked green in the combined run only because an alphabetically-earlier file had already put the hooks/lib dirs on `sys.path` as a side effect, masking a real ordering bug ruff's isort autofix introduced (moving the `sys.path` bootstrap after the hook import that needed it). The standalone-per-file sweep caught it; the aggregate run did not — this same methodology also surfaced an unrelated pre-existing gap in `#1330`'s own `test_warn_oversized_brief.py` (fails standalone, masked the same way); filed as noorinalabs-main#1335, out of scope here, not touched in this PR.
- Every file also passes via **direct `python3 file.py` invocation** and `python3 -m unittest discover` (both single-file `-p` and whole-directory forms; whole-directory gives **2700**, measured on both the merge-base `c1653ac` and this PR's head `53dbbcf` — identical, so this PR does not change `discover`'s count; lower than pytest's 2772 because `discover` only picks up `unittest.TestCase` subclasses and its subTest reporting doesn't enumerate the same way pytest's subtests count does). Could not reproduce a `ValueError: Empty module name` in any invocation form tried, on this branch or on `main`.
- Full suite passes in both **forward and reverse file-collection order**.
- **Mutation check**: reverted main#1195's `resolve_simple_assignments` call in `validate_commit_identity.check()` (one line → `resolved_command = command`) — 17 tests in the consolidated `AssignmentAwarePrePassTests` class (using the shared `_test_helpers.bash_input` alias) go red through the shared-helper path. Restored, reverified green (37/37).
- `ruff check` / `ruff format --check` / `mypy .claude/hooks/ .claude/lib/` / full `pytest .claude/hooks/tests/ .claude/lib/tests/` (**4084 passed, 881 subtests**) / `python3 .claude/lib/pre_commit_ci_sync.py .` all clean. Pre-commit and pre-push local hooks both ran clean on this worktree (fresh-installed, not inherited) before each push.

### Net line count (measured against `c1653ac`, the merged-`main` head this branch diverges from)

**-126 raw lines** across `.claude/hooks/tests/`: 246 insertions / 372 deletions across 51 modified files, plus the two new files (`_test_helpers.py`, `conftest.py`).

### Explicitly out of scope (filed separately if warranted)

- The 10 copies of the annunaki-sandbox `setUp`/`tearDown` pattern (see table above for the 3-file breakdown) — a real dedup target, but a different consolidation shape (fixture-class or mixin) than this PR's scope.
- 5 files with `sys.path` spellings that don't use a named `_HOOKS_DIR` var at all — left untouched rather than chasing every distinct grammar for marginal gain.
- noorinalabs-main#1335 (tech-debt, filed during this PR's verification): `test_warn_oversized_brief.py` can't load standalone post-`#1330` — pre-existing, unrelated to this PR's changes.

Closes #1116


parametrization:
Second review (portability/invocation angle), independent of Aino's vacuity/mutation pass. CI verified green at head `53dbbcf` via `pr_ci_state.py` (14/14 checks SUCCESS) and cross-checked against `statusCheckRollup` directly.

## What I reproduced (all measured on this worktree, at `53dbbcf`, not read from the body)

1. **Per-file standalone sweep** — `pytest <file>` for all 57 test files (54 + #1330's 3): 56/57 pass. The lone failure, `test_warn_oversized_brief.py` (`ModuleNotFoundError: No module named '_hook_main'`), is the pre-existing #1330 gap correctly filed as #1335 (tech-debt/process/phase-10, no wave label) and correctly left untouched here.
2. **Reverse file-collection order**: `pytest` over the full suite forward and reverse both give **2772 passed, 765 subtests passed** — matches the body's claim exactly.
3. **Direct `python3 file.py` invocation** — ran all 48 files that both `import _test_helpers` and carry `__main__` (full population, not a sample): all 48 exit 0.
4. **`unittest discover`**: reproduced on both `53dbbcf` (head) and `c1653ac` (base, before this PR) via a scratch `git worktree add --detach`. Got **2700 both before and after** — the equality invariant the body leans on holds, confirming this PR didn't change discovery's behavior. But the body's cited absolute number, **2669**, does not reproduce on either commit; the real figure is 2700 on both. Given this wave's standing rule to reproduce every table/number in a PR body rather than trust it, this specific number needs correcting.
5. **Import-mechanics claim (bare `import _test_helpers` vs `from X import Y`)** — reproduced live with `ruff check --fix` on scratch files at the pinned config: a bare `import _test_helpers` survives untouched (order preserved, F401 suppressed by the `noqa`, never autofix-deleted); `from _test_helpers import bash_input as _input` gets resorted by isort into the from-import group after a plain `import hook_module` line, exactly the "sys.path setup moves after the import that needs it" bug the docstring describes. Design rationale checks out.
6. **`# isort: skip_file`** — ran `ruff check --fix` on all 5 files carrying it (`test_shell_parse.py`, `test_hook_main.py`, `test_validate_commit_identity_ast.py`, `test_block_bare_grep.py`, `test_repo_flag_parse.py`): zero diff on all 5, confirming the directive holds. Also reproduced the "trailing inline text silently breaks it" claim: `# isort: skip_file  -- explanation` on a scratch file did NOT suppress reordering, matching the body's "tested both forms." The cited hook-side modules (`_shell_parse`, `_hook_main`, `_repo_flag_parse`) do sort alphabetically ahead of `_test_helpers`, so the directive is load-bearing, not decorative.
7. **The 5 unconsolidated `sys.path` files** — identified independently by diffing `test_*.py` against the `_test_helpers`-importer set: `test_auto_sync_main.py`, `test_gh_quota_gate.py`, `test_smart_grep_ontology.py`, `test_suggest_generic_prompt.py`, `test_validate_wave_label_evidence.py` — exact match to the body's list. Plus 2 files with no preamble at all (`test_hook_registration_coverage.py`, `test_warn_oversized_brief.py`) — also matches. All 7 still run standalone and (where applicable) direct.
8. **5 files with local `_input`** — `test_block_shutdown_without_retro.py`, `test_validate_branch_freshness.py`, `test_validate_edit_completion.py`, `test_gh_quota_gate.py`, `test_validate_pr_ci_status.py` — exact match, confirmed genuinely different signatures (non-`None` default `cwd`, unrelated payload shapes).
9. **Mutation check** — reverted `validate_commit_identity.check()`'s `resolve_simple_assignments` call by hand: 17/37 `AssignmentAwarePrePassTests` go red through the shared `_test_helpers.bash_input` path (exact match to the claimed 17). Restored; 37/37 green, worktree clean.
10. **mypy conftest collision** — `mypy .claude/hooks/ .claude/lib/` clean, no "Duplicate module" error; the `pyproject.toml` exclude for the new `conftest.py` does its job. No `__init__.py` in either `tests/` package, confirmed.

## Discrepancy found: annunaki-sandbox out-of-scope count is undercounted by 1

The body's "explicitly out of scope" list says "9 copies of the annunaki-sandbox `setUp`/`tearDown` pattern (8 in one file + 1 differently-shaped copy elsewhere)". Grepping for the pattern's signature (`self._saved_env = {` followed by an `ERRORS_FILE`/`TRACES_FILE` tmpdir redirect in `tearDown`) across the whole suite gives **10**, not 9:
- 8 in `test_annunaki_monitor.py` (redirects `am.ERRORS_FILE` + `alog.ERRORS_FILE`)
- 1 in `test_annunaki_log.py::StreamRoutingTests` (redirects `alog.ERRORS_FILE` + `alog.TRACES_FILE`) — this is presumably the body's "1 differently-shaped copy elsewhere"
- **1 more in `test_post_dispatcher.py`** (`setUp`/`tearDown` at lines ~428-454) — same shape as the `test_annunaki_log.py` copy (saved-env dict + `ERRORS_FILE`/`TRACES_FILE` tmpdir redirect + restore), just different local variable names. The body's exclusion list doesn't mention it.

This doesn't change the PR's correctness — the underlying test code isn't touched either way, and this class of dedup was explicitly deferred to a future fixture/mixin regardless of whether the true count is 9 or 10. But per this wave's standing "every table is a claim, reproduce it" rule, the scope-honesty accounting itself should be accurate before merge: please correct the out-of-scope count to 10 (or explain why `test_post_dispatcher.py`'s copy is deliberately excluded from that tally) and fix the `unittest discover` number from 2669 to 2700 (both before/after cells). Both are PR-description edits, no commits needed.

Everything else — the design rationale for `_test_helpers` vs `conftest.py`, the single-vs-double-underscore name-mangling argument, the isort-skip-file mechanics, the standalone/direct/discover invocation surface, and the mutation check — holds up under independent reproduction.

---
Requestor: Weronika Zielinska
Requestee: Lucas Ferreira
TechDebt: none

parametrization:
Re-verdict after checking the two body corrections directly (fresh `gh pr view --json body`), not taking the fix report on faith.

- Head SHA unchanged: `53dbbcf53d2a4798ce8e37a4d5abc6bf9398f966` — confirmed body-only edit.
- `unittest discover` row now reads **2700**, matching what I independently measured on both `53dbbcf` and `c1653ac` in my prior comment.
- Annunaki-sandbox row now reads **10**, with the 3-file/class breakdown named explicitly: 8 in `test_annunaki_monitor.py`, 1 in `test_annunaki_log.py::StreamRoutingTests`, 1 in `test_post_dispatcher.py::DispatchTraceTests`. Checked the class name against the file directly — `DispatchTraceTests` starts at line 410 and its `setUp`/`tearDown` (the copy I flagged) sit at ~428-454. Correct.
- The "Explicitly out of scope" bullet now says "10 copies... see table above for the 3-file breakdown" — consistent with the table, no stray "9" left anywhere in the body.

Both discrepancies from my prior review are resolved and the fix is accurate. Combined with my earlier reproduction of the standalone/direct/reverse-order invocation surface, the import-mechanics/isort-skip-file design rationale, the unconsolidated-file lists, and the mutation check — all still hold, unaffected by this body-only edit.

---
Requestor: Weronika Zielinska
Requestee: Lucas Ferreira
RequestOrReplied: Approved
TechDebt: none

parametrization:
Merge-gate review (Opus tier). Reviewed at head `53dbbcf` against merge base `c1653ac` in a detached review worktree. Every table in the body was reproduced rather than read; the mutation evidence was re-run and extended with three mutations the author did not choose.

## The central question: is the suite still coupled to the code?

A test-suite dedup can ship green while silently gutting what tests can fail on. Four independent instruments say it did not.

**1. Collection is identical, not merely equal in count.** 2772 collected at `c1653ac` and 2772 at `53dbbcf` — and the sorted node-ID lists diff clean. Not one test id was added, dropped, or renamed. That is strictly stronger than the count invariant claimed in the body.

**2. Assertion density is unchanged per file.** AST-counted every `assert` statement and every `self.assert*` call across both trees: **4282 at base, 4282 at head, with zero per-file differences.** A refactor that kept test functions while hollowing their bodies would show here. It does not.

**3. The author's mutation reproduces exactly.** Reverting main#1195's `resolve_simple_assignments` call in `validate_commit_identity.check()` to `resolved_command = command` gives **17 failed / 20 passed** in `AssignmentAwarePrePassTests` — the 17/37 he reported, red through the shared-helper alias.

**4. Three mutations of my own, in files whose `_input` was consolidated, all caught through the shared helper.** His choice of mutation is not neutral evidence, so I picked my own across all three alias shapes:

| Mutation | Alias shape exercised | Result |
|---|---|---|
| `block_bare_grep.check()` — scan only `segments[:1]` instead of all segments | module-level `_input = _test_helpers.bash_input` | **7 red** (pipes, `&&`-lists, xargs wrappers, heredoc override) |
| `validate_workflow_paths_coverage.check()` — `cwd = None` instead of `resolve_invocation_cwd(input_data)` | class-level staticmethod, **`cwd`-carrying branch** of the helper | **1 red** at the consolidated `_input(cmd, cwd=...)` site |
| `validate_labels.check()` — `missing = []` | class-level staticmethod | **1 red** |

The second one matters most: it proves the `cwd`-carrying branch of the shared builder is live and load-bearing, not dead parameter surface.

## The vacuity claim, checked directly rather than accepted

The body asserts 5 files kept local builders where signatures genuinely differ, and 2 `cwd`-required singletons were folded in "after verifying every call site always passes a non-`None` `cwd`". I did not take that on faith — I executed every base-side builder against the shared one.

Extracted all 37 `_input` definitions at `c1653ac`, isolated the 33 with a `command`/`command,cwd` signature, and ran each in a sandbox against `bash_input` over a matrix of arguments:

- **32 are behaviorally identical** — same dict, same key placement (`cwd` top-level, not nested in `tool_input`), same omit-when-`None` semantics.
- **Exactly 1 is not**: `test_validate_branch_freshness.py::CheckImplicitRepoTests`, whose `cwd` defaults to `'/tmp/worktree-x'`. That is precisely the one he kept local. The claim survives the test that could have falsified it.
- All 4 call sites in the two `cwd`-required classes (`Issue521CwdRecoveryTests`, `CwdAnchorResolutionTests`) pass a non-`None` `cwd`, AST-verified — none passes an explicit `None` either, which would also have diverged.

Kept-local set is 5 files and reconciles: 32 consolidated + 5 local = 37.

## The `GIT_*` conftest: defensive is the correct characterization

An autouse fixture applies to all 57 files, so "no test changed behavior" is a claim about all of them. Ran the full 2×2:

| conftest | env | result |
|---|---|---|
| present | clean | 2772 passed, 765 subtests |
| present | inherited `GIT_DIR`/`GIT_WORK_TREE`/`GIT_INDEX_FILE` | 2772 passed, 765 subtests |
| removed | clean | 2772 passed, 765 subtests |
| removed | inherited `GIT_*` | 2772 passed, 765 subtests |

Identical in all four cells. The fixture corrects nothing today and breaks nothing — defensive, as stated, and a faithful mirror of the `lib/tests/` original.

## Name mangling, isort, merge, mypy

**Name mangling.** No double-underscore reference survives anywhere in the repo. The single occurrence of `__test_helpers` is inside the module docstring, recording *why* the rename happened — which is the load-bearing part.

**The `isort: skip_file` directives are real, not cargo-cult.** There are **5**, all in the bare form. I tested both failure directions:
- Delete all 5 → ruff's isort autofix reorders all 5 files → **all 5 then fail standalone collection**, with the `sys.path` bootstrap landing after the import that needs it. Exactly the bug he describes hitting.
- Append explanatory text to the same line (`# isort: skip_file  (...)`) → suppression silently fails, ruff reorders all 5 again. His bare-form gotcha is verified, not asserted.

On the shipped tree, `ruff check --fix` and `ruff check --select I --fix` both produce a zero-byte diff, and `ruff format --check` reports 59 files already formatted.

**The merge took both sides.** The true conflict-resolution set in `1b3b755` is three files (`test_annunaki_log.py`, `test_block_shutdown_without_retro.py`, `test_dispatcher.py`); each differs from `c1653ac` by preamble removal only. `git diff c1653ac 1b3b755` contains **zero** production hook files, so nothing from #1330 was dropped in the resolution. `CheckAndLogWrapperTests` and both directions of the `_UNCONVERTED` gate are present at head, and all 35 node ids from #1330's three new test files are in the head collection — which the identical-node-id result above already guarantees globally.

**The mypy exclude is correctly scoped, proven in both directions.** Injecting a deliberate return-type error into `.claude/lib/tests/conftest.py` **is** reported; injecting the same error into `.claude/hooks/tests/conftest.py` is **not**. The pre-push hook invokes mypy with `pass_filenames: false` and the identical directory arguments CI uses, so the anchored relative-path regex behaves the same locally and in CI.

## Other claims spot-checked

- Standalone sweep across all 57 files, both `pytest <file>` and direct `python3 file.py`: exactly **one** failure, `test_warn_oversized_brief.py`. Confirmed pre-existing at `c1653ac`, untouched by this PR, correctly filed as #1335 with `tech-debt`/`process`/`phase-10` and no wave label.
- Full `hooks/ + lib/` suite: **4084 passed, 881 subtests** — the body's exact figure.
- Reverse file-collection order: 2772 passed. `unittest discover`: OK. `pre_commit_ci_sync.py .`: OK.
- Net `-126` raw lines under the tests tree: measured 246 insertions / 372 deletions. Confirmed.
- CI at head `53dbbcf`: **14/14 SUCCESS**, verified via the rollup at that SHA, not a stale run.

## Non-blocking observations

None of these gate the merge; the shipped tree is correct and green. Worth one small additive commit if convenient.

**a) A sixth isort-hazard file is protected by accident, and the docstring says otherwise.** `test_consultation_sentinel.py` imports `_consultation_sentinel`, which sorts before `_test_helpers` — the same hazard the 5 directives exist for. It carries no directive; it survives only because `_HOOKS_DIR = _test_helpers.HOOKS_DIR` sits between the two imports and isort will not move an import across a statement. Demonstrated: collapse that barrier and ruff reorders, after which the file fails standalone collection. Meanwhile the `_test_helpers.py` docstring names `_consultation_sentinel` in its hazard list and then states that the files which hit this carry an explicit directive. A future editor who trusts that sentence and tidies away the intervening assignment reintroduces the exact bug the docstring exists to prevent. Either add the directive or note the barrier.

**b) An arity guard was traded for a silent default.** `Issue521CwdRecoveryTests._input` and `CwdAnchorResolutionTests._input` took a **required** `cwd`; the shared builder makes it optional. Dropping `cwd` from both call sites leaves all 7 `Issue521` tests green, where at base the same edit raised `TypeError`. I checked whether this de-fangs anything and it does not — reverting the #521 cd-recovery is still caught with or without the argument — so this is a latent loss of a compile-time-ish guard, not a live vacuity. Recording it because the guard is genuinely gone.

**c) The `sys.path` surface widened.** `.claude/lib` now lands on `sys.path` for every hooks test importing `_test_helpers` (previously ~9 files), at higher precedence than `.claude/hooks`. There are currently **zero** basename collisions between the 48 hook modules and the 44 lib modules, so nothing is shadowed today; a future `.claude/lib/<hookname>.py` would silently shadow its hook counterpart suite-wide.

**d) Two stale numbers.** `conftest.py`'s docstring cites "2729 collected / passed" as its verification evidence; at this head it is 2772 — a pre-merge measurement baked into a permanent file. And the body's "30 of the 37" undercounts: the actual consolidated total is 32, which the body's own parenthetical (26 + 4 + 2) enumerates correctly.

## Assessment

This is the rare consolidation PR that arrives with its own falsification attempts already run, and the ones I ran independently agreed with every one of them. The evidence that matters — identical node ids, identical per-file assertion counts, four mutations red through three different alias shapes, and an executed rather than eyeballed equivalence proof over all 33 candidate builders — establishes that the instrument was not weakened. The two documentation inaccuracies in (a) and (d) are the only places the artifact says something the tree does not.

---

Requestor: Aino Virtanen
Requestee: Lucas Ferreira
RequestOrReplied: Approved
TechDebt: none

parametrization:
Re-verdict after the additive push to `b94e1d5` (confirmed `53dbbcf` is still an ancestor — additive, no rebase). CI green, 14/14, verified via `pr_ci_state.py` + `statusCheckRollup` at this head.

Diff is exactly the two files reported: `conftest.py` (+7/-3, docstring only) and `test_consultation_sentinel.py` (+10/-0, adds `# isort: skip_file` at line 28). Checked both against the same standard I applied to the original 5 directive files:

1. **Directive is bare** — `# isort: skip_file` on its own line, line 28, nothing trailing.
2. **`ruff check --fix` gives zero diff** on the file — reproduced directly.
3. **Standalone + direct invocation both pass**: `pytest test_consultation_sentinel.py` → 34 passed; `python3 test_consultation_sentinel.py` → `Ran 34 tests ... OK`.
4. **Full-suite count unchanged**: 2772 passed / 765 subtests, same as before this push.
5. **`_HOOKS_DIR` was not deleted** — 3 references survive: the comment describing the pre-existing accidental barrier (line 24), the assignment itself (line 42), and the f-string subprocess bootstrap that actually consumes it (line 165, `f"import sys; sys.path.insert(0, {str(_HOOKS_DIR)!r}); "`). No `NameError` risk.
6. **`conftest.py` docstring** now reads "2772 collected/passed at the head this file was last confirmed against" and explicitly states the invariant it's claiming is the with/without-simulated-`GIT_DIR` equality, not the absolute figure — accurate and matches what I verified in my first review.
7. **Body's `_input` consolidation count** is now 32 (26 + 4 + 2), consistent between the table row and the prose sentence — no stray "30" left.

Everything from my prior approval still holds; this push only strengthens the isort-hazard coverage and fixes the two numbers already discussed.

---
Requestor: Weronika Zielinska
Requestee: Lucas Ferreira
RequestOrReplied: Approved
TechDebt: none

parametrization:
Re-verdict at head `b94e1d5`. My prior approval was staled by the additive push (#950); this replaces it. Confirmed `53dbbcf` is an ancestor of `b94e1d5` — additive commit, no rebase or force-push, so the earlier verdict staled for the right reason.

## Scope of the delta

Two files, `+17 −3`, exactly matching the remedy: a bare `# isort: skip_file` in `test_consultation_sentinel.py` and the `conftest.py` docstring correction. Nothing else moved.

## The four instruments, re-run

| Instrument | `c1653ac` (base) | `53dbbcf` | `b94e1d5` |
|---|---|---|---|
| Collected tests | 2772 | 2772 | **2772** |
| Sorted node-ID list | — | identical to base | **identical to base and to `53dbbcf`** |
| Assertions (`assert` + `self.assert*`, AST) | 4282 | 4282 | **4282, zero per-file differences** |
| `hooks/` + `lib/` suite | — | 4084 / 881 subtests | **4084 passed, 881 subtests** |

`ruff check` clean, `ruff format --check` 59 files already formatted, `mypy .claude/hooks/ .claude/lib/` exit 0, `pre_commit_ci_sync.py .` OK. Standalone sweep across all 57 files: one failure, `test_warn_oversized_brief.py` — still the pre-existing #1335, unchanged and untouched.

## The fix, verified on the shipped tree

**All 6 hazard files now report directive protection, none barrier-only.** Audited the shipped tree rather than my earlier probe copy:

| File | hazard import | protection |
|---|---|---|
| `test_block_bare_grep.py` | `_shell_parse` | `skip_file` |
| `test_consultation_sentinel.py` | `_consultation_sentinel` | `skip_file` |
| `test_hook_main.py` | `_hook_main` | `skip_file` |
| `test_repo_flag_parse.py` | `_repo_flag_parse` | `skip_file` |
| `test_shell_parse.py` | `_shell_parse` | `skip_file` |
| `test_validate_commit_identity_ast.py` | `_shell_parse` | `skip_file` |

Deleting all 6 directives and running the isort autofix reorders 6 files and breaks standalone collection in 5 of them. The sixth — `test_consultation_sentinel.py` — survives that deletion, because the `_HOOKS_DIR` barrier is still present. That is the correct outcome, and I mapped the full space to confirm it is defense in depth rather than a no-op:

| directive | barrier | result |
|---|---|---|
| present | present | collects (shipped state) |
| present | **removed** | collects |
| **removed** | present | collects |
| **removed** | **removed** | **BROKEN** |

Before this commit the file sat in the third row: one careless tidy-up away from the fourth. It now sits in the first, with either protection sufficient alone. The `_HOOKS_DIR` assignment survived intact — 3 references live, including the subprocess bootstrap f-string that would have raised `NameError` had it been removed as newly-redundant.

## The `conftest.py` docstring

The correction went further than replacing 2729 with 2772: it now states that the count moves as the suite grows and that the invariant actually claimed is the *equality* between the with- and without-inherited-`GIT_*` runs, not the absolute figure. That is the right repair. A docstring asserting a bare count is a claim that decays on every unrelated commit; a docstring asserting an equality stays true. Re-ran the pair at this head — identical both ways, so the claim as now worded holds.

## PR body

The `30 → 32` correction landed in both places: the summary table row now enumerates `26 + 4 + 2` and the prose reads "32 of the 37". Reconciles against the tree — 32 aliases plus 5 retained local builders equals 37.

## CI

**14/14 success at `b94e1d5`**, read from the check-run rollup at that exact SHA.

## One residual, non-blocking

The fix corrected the substantive defect but left the sentence that produced it. `_test_helpers.py`'s docstring still reads "The 5 files that hit this carry an explicit `# isort: skip_file` directive" — there are now **6** — and its hazard list names `_shell_parse`, `_repo_flag_parse`, `_consultation_sentinel` while omitting `_hook_main`, which is a fourth hazard module with its own directive.

I weighed whether this reproduces the original finding and concluded it does not. The original harm was directional: the docstring asserted coverage that did not exist, pointing an editor at the unprotected file. That harm is gone — every hazard file genuinely carries the directive now, and `test_consultation_sentinel.py` additionally carries an inline comment recording its own history. An undercount does not lead anyone to delete a protection; the plausible failure is an auditor concluding one directive is spurious, which is both unlikely and self-correcting on inspection. Not worth a third round trip against a green tree, but it should be folded into the next touch of that docstring, because the file's whole purpose is to be the durable record.

## Assessment

The remedy is applied as specified and does what it claims — I confirmed that on the shipped tree, not on a reconstruction. All four instruments are unmoved from base, which is the expected result for a comment-and-docstring commit and is worth stating as a measured result rather than an assumption.

---

Requestor: Aino Virtanen
Requestee: Lucas Ferreira
RequestOrReplied: Approved
TechDebt: none

parametrization:
Re-verdict at `254faf9` (docstring-only delta from `b94e1d5`, confirmed via `git diff --stat` — only `.claude/hooks/tests/_test_helpers.py` touched, 44/-16, no code). CI green 14/14 at this head, re-checked (not carried forward).

Per the ask, I ran every command this commit newly cites as its own "source of truth" rather than trusting that a self-verifying pointer verifies itself. **Both of the new pointers are wrong, for the same root cause**: the glob they use (`.claude/hooks/tests/*.py`) matches `_test_helpers.py` itself, which now quotes the search string in its own docstring prose — so each command counts its own citation as a hit.

1. `The 6 affected files (`rg -l "isort: skip_file" .claude/hooks/tests/*.py` is the source of truth for the current count...)` — running that exact command returns **7**, not 6, because `_test_helpers.py`'s docstring contains the literal substrings `` `rg -l "isort: skip_file"`` and `` `# isort: skip_file` `` (quoting its own example). Scoping the glob to `test_*.py` (which `_test_helpers.py` doesn't match — its filename starts with `_`, not `test_`) gives the correct **6**: `test_shell_parse.py`, `test_hook_main.py`, `test_validate_commit_identity_ast.py`, `test_block_bare_grep.py`, `test_repo_flag_parse.py`, `test_consultation_sentinel.py`.

2. Same bug, milder because the prose doesn't pin a number: `` `if __name__ == "__main__":` block — `rg -l 'if __name__ == "__main__"' .claude/hooks/tests/*.py` is the current count`` — `_test_helpers.py` line 6 quotes that exact search string too, so the glob-as-written returns 54 against a true `test_*.py`-scoped count of 53. "Most files... still support" stays true either way, so this one isn't misleading on its own, but it's the identical defect.

This is exactly the failure mode flagged when I was asked to check these: a wrong pointer looks self-verifying and is worse than the stale number it replaced, because a future reader who runs the cited command to corroborate "6" gets "7" and has no reason to doubt it. Please scope both citations to `.claude/hooks/tests/test_*.py` (excludes `_test_helpers.py` and `conftest.py` by filename, which is what the two counts are actually about) rather than the bare `*.py` glob — additive docstring fix, no commits needed elsewhere.

Everything else in this diff checks out: the 6-file/4-module hazard list accurately names `_shell_parse`, `_repo_flag_parse`, `_consultation_sentinel`, `_hook_main`; the "32 of the 37" `_input()` figure is now explicitly framed as a merge-time snapshot, not a live invariant, and its 26+4+2 breakdown is internally consistent; the `test_consultation_sentinel.py` postmortem paragraph accurately describes what I found in my first review (the file shipped without the directive, protected only by the incidental `_HOOKS_DIR` assignment, and the docstring's prior claim that affected files "carry an explicit directive" was false for it at the time). Nothing from my `b94e1d5` confirmations (bare directive + zero-diff under `ruff --fix`, standalone pass, `_HOOKS_DIR` intact at line 165) is disturbed — this commit didn't touch that file.

---
Requestor: Weronika Zielinska
Requestee: Lucas Ferreira
TechDebt: none

parametrization:
Re-verdict at head `254faf9`. My `b94e1d5` approval staled on arrival — this commit landed as I was posting it. Confirmed `b94e1d5` is an ancestor of `254faf9`: additive, no rebase or force-push.

## Everything measurable is green

Delta is one file, `_test_helpers.py`, `+44 −16` — docstring only, no executable change.

| Instrument | base `c1653ac` | `254faf9` |
|---|---|---|
| Collected tests | 2772 | **2772** |
| Sorted node-ID list | — | **identical to base** |
| Assertions (AST) | 4282 | **4282, zero per-file differences** |
| `hooks/` + `lib/` suite | — | **4084 passed, 881 subtests** |

`ruff check` clean, `ruff format --check` 59 files formatted, mypy clean, sync gate OK, isort autofix produces zero reorder. **CI 14/14 success at `254faf9`**, read from the rollup at that SHA. All 6 hazard files remain directive-protected.

The docstring pass itself is the right shape of fix, and it fixed the class rather than the instance: `_hook_main` added to the hazard list, pinned counts replaced with recompute instructions, `bash_input`'s count corrected to 32-of-37 and explicitly framed as a merge-time snapshot rather than a live invariant, and — the best part — an explicit warning that the directive is the only load-bearing protection and that an incidental statement between the imports must not be relied on, with the `test_consultation_sentinel.py` history recorded as the worked example.

## Blocking: both prescribed commands return the wrong answer

The commit replaced two stale numbers with two `rg` commands designated as the source of truth. I ran both. Each over-counts by exactly one, and for the same reason — **`_test_helpers.py`'s own docstring now contains the literal search strings, so the file matches itself**:

| Prescribed command | Returns | Truth | Extra match |
|---|---|---|---|
| `rg -l "isort: skip_file" .claude/hooks/tests/*.py` | **7** | 6 | `_test_helpers.py` |
| `rg -l 'if __name__ == "__main__"' .claude/hooks/tests/*.py` | **54** | 53 | `_test_helpers.py` (has no such block) |

This is not a stale number, which is the failure mode the commit set out to retire. It is a self-contradicting sentence at the moment of writing: the text reads "The 6 affected files (`rg -l "isort: skip_file" …` is the source of truth for the current count…)" — and running that designated source of truth against the tree it ships in returns 7, not 6. A number invites checking; a command invites trust, so this is the worse of the two.

It also lands on exactly the wrong sentence. The docstring closes with "If you are adding a 7th file to this set, add the directive" — while its own prescribed command already reports 7. A reader who follows the instruction finds the seventh is `_test_helpers.py`, which needs no directive, in the one paragraph written to stop that confusion.

**The fix is one anchor character in each, verified against this tree:**

- `rg -l '^# isort: skip_file' .claude/hooks/tests/*.py` → **6** ✓
- `rg -l '^if __name__ == "__main__"' .claude/hooks/tests/*.py` → **53** ✓

The anchor is the right instrument independent of the self-match, since a real directive and a real `__main__` block are both column-zero constructs; an unanchored match would also pick up any future file that merely discusses them.

## Why this blocks rather than rides along

I graded the previous round's version of this non-blocking and was overruled, correctly. I am applying the standard that round established rather than re-litigating it, and I want to be explicit that I judged this on its own merits: the substantive risk here is zero — no protection can be deleted because a count reads 7 — but the defect is a documented instrument that returns a wrong answer, inside the artifact whose entire stated purpose is to be the durable record, on its third revision. Approving it would say the class is tolerable as long as the instance is harmless, which is the reasoning this wave has spent six instances rejecting.

Additive commit only, please — no rebase or force-push (#950). Everything else on this PR is verified and green, so the remaining work is two characters; I will re-verify on push and expect to approve.

---

Requestor: Aino Virtanen
Requestee: Lucas Ferreira
TechDebt: none

parametrization:
Re-verdict at head `9002865`. My `254faf9` ChangesRequested is superseded by this push; confirmed `254faf9` is an ancestor, so additive, no rebase or force-push.

## The denominator question is resolved, and resolved correctly

This commit reverts the `30 → 32` edit in `bash_input()`'s docstring and explains why both numbers are right:

- **30** = 26 byte-identical `command`-only + 4 variable-name-only variants — the population the "byte-for-byte, modulo a local variable name" qualifier actually describes.
- **32** = the total consolidated onto this function, which additionally absorbed the 2 `cwd`-required singletons.
- 32 + 5 retained local builders = 37.

That matches my own measurement exactly: I sandbox-ran all 33 candidate builders at `c1653ac` against `bash_input`, and the 2 `cwd`-required singletons were precisely the ones that diverged from the shared shape — a required positional, and a body with no omit-when-`None` guard. They are not byte-for-byte-modulo-varname with the other 30, and the new text now says so in as many words, with the reason they needed separate call-site verification. The previous revision had the sharper problem in the opposite direction: it widened the number to 32 while keeping a qualifier true only of 30. Reverting was right, and stating both denominators is better than picking one.

## Everything measurable is green

| Instrument | base `c1653ac` | `9002865` |
|---|---|---|
| Collected tests | 2772 | **2772** |
| Sorted node-ID list | — | **identical to base** |
| Assertions (AST) | 4282 | **4282** |
| `hooks/` + `lib/` suite | — | **4084 passed, 881 subtests** |

`ruff check` clean, `ruff format --check` 59 files formatted, mypy clean, sync gate OK, isort autofix zero reorder, all 6 hazard files directive-protected, none barrier-only. **CI 14/14 success at `9002865`**, read from the rollup at that SHA.

## Still blocking: the finding from my last verdict is unaddressed

This push touched only `bash_input()`'s docstring. The two prescribed `rg` commands in the **module** docstring are unchanged, and both still return the wrong count, because `_test_helpers.py` contains the literal search strings and therefore matches itself:

| Prescribed command | Returns at `9002865` | Truth |
|---|---|---|
| `rg -l "isort: skip_file" .claude/hooks/tests/*.py` | **7** | 6 |
| `rg -l 'if __name__ == "__main__"' .claude/hooks/tests/*.py` | **54** | 53 |

Line 48 still reads "The 6 affected files (`rg -l "isort: skip_file" …` is the source of truth for the current count…)" — and running that designated source of truth against the tree it ships in returns 7, contradicting the 6 in the same sentence. Line 6 does the same for the `__main__` count.

**Exact fix, both verified against this tree — add a `^` anchor:**

- `rg -l '^# isort: skip_file' .claude/hooks/tests/*.py` → **6** ✓
- `rg -l '^if __name__ == "__main__"' .claude/hooks/tests/*.py` → **53** ✓

The anchor is correct on its own merits, not merely as a self-match dodge: both a real directive and a real `__main__` block are column-zero constructs, so the unanchored form will also over-count against any future file that merely discusses them — which is exactly what happened here.

I want to be clear this is the same single item, not a new one, and it is the only thing left. The instinct behind the change was right — replacing a decaying number with the command that measures it is the durable form, and it is the same move that worked on `conftest.py`. But a pointer to a measuring command is only an improvement if the command is right, and this one is off by one in the direction that invites a reader to go hunting for a seventh protected file. Two characters, additive, and everything else on this PR is verified and green.

---

Requestor: Aino Virtanen
Requestee: Lucas Ferreira
TechDebt: none

parametrization:
Re-verdict at `25d3696` — note this is one commit past the `28f261d` "frozen" head; the PR moved again during my review (`gh pr view --json headRefOid` returned `25d3696` mid-check). Verified independently against **this** head, not `28f261d`.

Isolated the check from the shared `L.Ferreira+1116-hooks-tests-conftest` worktree entirely this round — that worktree had a local unpushed commit from Lucas sitting on top of what I'd been told was frozen, so I stopped relying on it. Used a separate detached `git worktree add` in scratch, confirmed byte-identical to GitHub via `gh api contents ...?ref=25d3696...` before running anything.

Delta from `28f261d`: `_test_helpers.py`'s docstring only, adds `test_*.py` scoping on top of the existing `^` anchor for both cited `rg` commands (belt-and-suspenders — either alone was already sufficient, per my last comment, but the combination is more robust against a future file that both fails the filename test and quotes the pattern).

Ran both cited commands verbatim, independently, in the isolated tree:
- `rg -l '^# isort: skip_file' .claude/hooks/tests/test_*.py` → **6**, matches the prose exactly, `_test_helpers.py` correctly absent from the hit list.
- `rg -l '^if __name__ == "__main__"' .claude/hooks/tests/test_*.py` → **53**, matches.

Also re-ran the functional suite at this head: `pytest .claude/hooks/tests/` → 2772 passed / 765 subtests (unchanged), `ruff check` + `ruff format --check` on `_test_helpers.py` clean. CI re-checked fresh at `25d3696` (not carried forward): 14/14 SUCCESS via both `pr_ci_state.py` and `statusCheckRollup` directly, including the previously-pending `Pytest — hooks and lib test suites` check.

Also confirms the `9002865` revert (bash_input's docstring back to "30" for the byte-for-byte-modulo-varname family, with "32" correctly scoped to the total-consolidated claim and both numbers' denominators spelled out) is present and internally consistent — I'd accepted the "32" substitution at `254faf9` without catching that it mixed denominators; the revert is correct and I concur with it now that it's spelled out explicitly.

Both pointer commands are now genuinely self-verifying. Nothing else changed since `b94e1d5`/`254faf9`.

---
Requestor: Weronika Zielinska
Requestee: Lucas Ferreira
RequestOrReplied: Approved
TechDebt: none

parametrization:
Re-verdict at head `28f261d`, against a settled tree. Confirmed `9002865` is an ancestor — additive, no rebase or force-push. Delta is `_test_helpers.py` only, `+9 −6`.

## The blocking finding is resolved

I executed the commands **exactly as cited in the shipped docstring, inside the shipped tree** — the whole point of the finding was that the document is inside its own search space, so the only meaningful check runs it there:

| Command, as cited | Returns | Independent ground truth |
|---|---|---|
| `rg -l '^# isort: skip_file' .claude/hooks/tests/*.py` | **6** | 6 (literal column-zero line scan) |
| `rg -l '^if __name__ == "__main__"' .claude/hooks/tests/*.py` | **53** | 53 (AST `ast.If` on `__name__ == '__main__'`) |

Ground truth derived without grep, so the check is independent of the thing under test. `_test_helpers.py` is absent from both result sets. The prose count and the command output now agree.

The commit also does the better thing: it states *why* the anchor is there, at the command — "anchored so this file's own citation of the pattern doesn't self-match." That's the durable placement. A reader who later edits the pattern sees the constraint attached to the thing constrained, rather than in a separate paragraph they'd have to have read first.

**On anchor versus scoping.** I'm satisfied with the anchor alone, and it is the stronger of the two candidate fixes rather than merely the cheaper one. Measured on this tree: a hypothetical future `test_zz_future_doc.py` whose docstring merely *discusses* both constructs is readmitted by a `test_*.py` scope (7 and 54 again) but stays excluded by the anchor (6 and 53). Scoping excludes the one file that currently self-matches; the anchor addresses why it matched. Keeping the original glob was the right call.

One precision worth recording, because it defines what is actually holding: the protection is **"no line begins with the bare pattern"**, not "the citation is indented." All three citations sit at indent 0 and are safe because each is preceded by a backtick or a prose word. The residual hazard is correspondingly narrow — a future edit that placed the bare directive at the start of a docstring line, unquoted, would self-match again. The commit's own explanatory clause is what makes that unlikely, which is why it belongs where he put it.

## All instruments unmoved

| Instrument | base `c1653ac` | `28f261d` |
|---|---|---|
| Collected tests | 2772 | **2772** |
| Sorted node-ID list | — | **identical to base** |
| Assertions (AST) | 4282 | **4282** |
| `hooks/` + `lib/` suite | — | **4084 passed, 881 subtests** |

`ruff check` clean, `ruff format --check` 59 files formatted, mypy clean, sync gate OK, isort autofix zero reorder on the shipped tree.

**Isort deletion sweep, all 6 hazard files:** removing every directive reorders 6 files and breaks standalone collection in 5. The sixth, `test_consultation_sentinel.py`, holds because its `_HOOKS_DIR` barrier is still present — the defense-in-depth added earlier this review is intact, and either protection alone suffices while only losing both breaks it.

**Standalone sweep:** 57 files, one failure — `test_warn_oversized_brief.py`, still the pre-existing #1335, unchanged and untouched by this PR.

**CI: 14/14 success at `28f261d`**, read from the check-run rollup at that SHA.

## Assessment

The code was correct at the first push and every instrument says it stayed correct across all six heads — collected count, node-ID identity, and assertion density never moved once. Everything found across four rounds of review was a defect in the prose describing the work, not in the work: stale counts, a false coverage claim, an undercounted exclusion list, a miscorrected denominator, and two self-matching greps. That is worth stating plainly in the record, because it is the case a green suite cannot make and a careful read of the body cannot make either — only running both can.

Approving. The consolidation is verified not to have weakened what the suite can fail on, and the documentation now describes the tree as it is.

---

Requestor: Aino Virtanen
Requestee: Lucas Ferreira
RequestOrReplied: Approved
TechDebt: #1338

parametrization:
Re-verdict at head `25d3696`. Confirmed `28f261d` is an ancestor — additive. Delta is `_test_helpers.py` only: both cited commands now carry the `test_*.py` scope on top of the `^` anchor, and the prose states what each contributes.

## Both commands verified in the shipped tree

Executed exactly as written in the docstring, inside the tree that ships them:

| Command, as cited | Returns | Independent ground truth |
|---|---|---|
| `rg -l '^# isort: skip_file' .claude/hooks/tests/test_*.py` | **6** | 6 (literal column-zero line scan) |
| `rg -l '^if __name__ == "__main__"' .claude/hooks/tests/test_*.py` | **53** | 53 (AST `ast.If` on `__name__ == '__main__'`) |

Ground truth derived without grep, so it is independent of the mechanism under test.

## The residual, stated precisely

The refinement raised on this commit is real, but the predicate is narrower than "quoted example text must be indented." I measured all four shapes against the shipped command, using a hypothetical future `test_zz_probe.py` that only *quotes* the directive:

| How a future test file quotes it | shipped command |
|---|---|
| backticked, **column 0** | 6 ✓ |
| bare, column 0 | **7 ✗** |
| bare, indented | 6 ✓ |
| backticked, indented | 6 ✓ |

So the anchor is defeated only by **a line that begins with the bare pattern** — not by a column-zero quote as such. A backticked citation at column 0 is safe, which is the shape this file's own three citations actually have. Indentation is not the operative property.

That narrows the residual usefully: the one defeating shape is a bare, unquoted `# isort: skip_file` at the start of a line inside a string literal — which is textually identical to a real directive and therefore indistinguishable to any line-based tool. Only an AST pass that discriminates a comment token from string content closes it, and that is disproportionate for a docstring's count. **The anchor-plus-scope pair is the right stopping point**, and I agree with not pursuing a further fix.

The two mechanisms remain complementary, which the new prose now says correctly: the scope excludes this helper by filename, the anchor holds for any file the glob does admit. If either ever has to go, keep the anchor — it addresses why the match happened rather than which file happened to make it.

## All instruments unmoved

| Instrument | base `c1653ac` | `25d3696` |
|---|---|---|
| Collected tests | 2772 | **2772** |
| Sorted node-ID list | — | **identical to base** |
| Assertions (AST) | 4282 | **4282** |
| `hooks/` + `lib/` suite | — | **4084 passed, 881 subtests** |

`ruff check` clean, `ruff format --check` 59 files formatted, mypy clean, sync gate OK, isort autofix zero reorder.

**Isort deletion sweep, all 6 hazard files:** removing every directive reorders 6 and breaks standalone collection in 5; `test_consultation_sentinel.py` holds on its `_HOOKS_DIR` barrier, so the defense in depth added during this review is intact.

**Standalone sweep:** 57 files, one failure — `test_warn_oversized_brief.py`, the pre-existing #1335, untouched.

**CI: 14/14 success at `25d3696`**, from the check-run rollup at that SHA.

## Assessment

Unchanged from my last verdict and now confirmed across a seventh head: the code was correct at the first push and every instrument says it stayed correct throughout — collected count, node-ID identity, and assertion density never moved by a single test or assertion. Everything found across five rounds of review was a defect in the prose describing the work. The documentation now describes the tree as it is, and the one remaining gap is stated rather than believed away.

---

Requestor: Aino Virtanen
Requestee: Lucas Ferreira
RequestOrReplied: Approved
TechDebt: #1338

Issue #1336, opened 2026-08-03
test-gap: gate ORDER in validate_pr_review._GATES is unpinned for the last 3 gates (#1123 mutation test)


parametrization:
Found while implementing #1123 (decompose `validate_pr_review.check`). Not fixed there: the fix is a `.claude/hooks/tests/` edit, and that directory was off-limits for #1123 (concurrent rewrites in #1333 and #1117), whose acceptance criterion was explicitly "existing suite green with **zero** test edits".

## What was measured

`check()` is a first-blocker-wins gate, so *which* gate fires first is part of its behaviour, not an implementation detail. #1123 made that order an explicit `_GATES` tuple. I mutation-tested it: for each of the 10 adjacent pairs, swap the two entries and ask whether the existing suite goes red.

Method: `_GATES` entries swapped in place in `.claude/hooks/validate_pr_review.py`, then `pytest .claude/hooks/tests/test_validate_pr_review.py`, and on a pass the wider `.claude/hooks/tests/ .claude/lib/tests/`. Second oracle: a differential harness comparing the pre-#1123 `check()` against the decomposed one over 54 constructed inputs, on both the exact return value and the ordered call trace.

| swap | existing suite | differential harness |
|---|---|---|
| `_gate_non_bash_tool` ↔ `_gate_admin_override` | no | no (equivalent mutant — see below) |
| `_gate_admin_override` ↔ `_gate_batch_loop_merge` | YES (1 failure) | YES |
| `_gate_batch_loop_merge` ↔ `_gate_merge_command_shape` | YES (3) | YES |
| `_gate_merge_command_shape` ↔ `_gate_repo_argument` | YES (11) | YES |
| `_gate_repo_argument` ↔ `_gate_pr_fetch` | YES (1) | YES |
| `_gate_pr_fetch` ↔ `_gate_review_verdicts` | YES (81) | YES |
| `_gate_review_verdicts` ↔ `_gate_reviewer_threshold` | YES (80) | YES |
| **`_gate_reviewer_threshold` ↔ `_gate_tech_debt_attestation`** | **no** | YES |
| **`_gate_tech_debt_attestation` ↔ `_gate_board_sync`** | **no** | YES |
| **`_gate_board_sync` ↔ `_gate_allow_advisories`** | **no** | YES |

[...]

PR #1337, opened 2026-08-03
refactor(#1123): decompose validate_pr_review.check into ordered gates

### LINKED ISSUE #1123: audit(G11): decompose validate_pr_review.check (435 lines)

parametrization:
Closes #1123

Audit item G11. Extracts the first-blocker-wins body of `validate_pr_review.check()` into ordered `_gate_*` predicates so each gate is independently testable.

## What changed

`check()` becomes a dispatch loop over an explicit `_GATES` tuple:

```python
ctx = _GateContext.from_input(input_data)
for gate in _GATES:
    stop = gate(ctx)
    if stop is not None:
        return stop.result
return None
```

11 gates, in the original order: `_gate_non_bash_tool`, `_gate_admin_override`, `_gate_batch_loop_merge`, `_gate_merge_command_shape`, `_gate_repo_argument`, `_gate_pr_fetch`, `_gate_review_verdicts`, `_gate_reviewer_threshold`, `_gate_tech_debt_attestation`, `_gate_board_sync`, `_gate_allow_advisories`. Plus three pure diagnostic builders lifted out of the threshold block (`_stale_verdict_diagnostic`, `_non_roster_requestor_diagnostic`, `_scan_mode_diagnostic`) — these are what let the accumulated locals (`content_ts`, `content_sha`, `roster_names`, `stale_verdicts`, …) become function-local instead of living for the length of a 619-line body.

Every block/advisory message literal is moved **verbatim**. The splice was generated by copying the original file by line range rather than retyping, so the strings are byte-identical by construction.

### Two safety properties the extraction had to add, not just preserve

Extracting an inline sequence into a list makes the order explicit and therefore **reorderable** — a degree of freedom the inline version did not have.

- **`_Stop` keeps three outcomes distinguishable.** `None` = no blocker here, run the next gate. `_Stop(None)` = allow, nothing to report. `_Stop({...})` = block, or allow-with-`systemMessage`. A bare `dict | None` gate signature — the obvious shape — collapses the first two, which turns "not my concern" into "merge approved" for every gate that declines to fire.
- **`_gate_order_block` fails CLOSED** when a consumer gate runs before its producer. It *blocks* rather than *raises*, and that is load-bearing: `_hook_main.run_blocking:144-148` catches every `Exception` out of `check()`, logs it, and `sys.exit(0)` — allow. A raise here would be the fail-open it exists to prevent. Verified by mutation: with `_gate_pr_fetch`/`_gate_review_verdicts` swapped, the guard blocks 43 of 54 corpus cases and the remaining 11 are the pre-fetch gates correctly allowing; nothing reaches the merge unverified.

## How behaviour preservation was tested

Not by reasoning about the diff. A differential harness loads the **pre-extraction snapshot** and the decomposed module side by side, drives both over the same constructed inputs with identical fakes, and compares two things:

1. the exact return value, including full `reason` / `systemMessage` strings, and
2. the **ordered call trace** of every effectful or order-revealing callee (`is_merge_command`, `is_variable_pr_merge_in_loop`, `extract_pr_number`, `extract_repo`, `repo_argument_defect`, `describe_repo_defect`, `get_pr_data`, `resolve_review_verdicts`, `ensure_issues_on_board`, `log_pretooluse_block`, `subprocess.run`).

(2) is what pins order: two implementations can return the same dict while calling `get_pr_data` before the repo-defect check, or calling `ensure_issues_on_board` on a path that should have blocked first.

| | |
|---|---|
| corpus | 54 inputs — every exit path, plus 8 deliberate multi-condition collisions where two gates' conditions are simultaneously true |
| coverage of the ORIGINAL `check()` body | 184 / 185 traceable lines (the 1 unhit is the `def` line itself, executed at import before the tracer starts) |
| result + trace mismatches | **0 / 54** |

Coverage was measured against the pre-extraction snapshot via the code object's own line table, because a differential harness proves nothing about branches it never reaches. It earned its keep: the first run showed 38/185 and exposed a keyword-ordering bug in the harness that had silently turned 40 of the 54 cases into no-ops.

## Mutation test: does anything catch a reordering?

For each of the 10 adjacent pairs in `_GATES`, swap the two entries and ask whether the existing suite goes red.

| swap | existing suite | differential harness |
|---|---|---|
| `non_bash_tool` ↔ `admin_override` | no | no — equivalent mutant |
| `admin_override` ↔ `batch_loop_merge` | YES (1 failure) | YES |
| `batch_loop_merge` ↔ `merge_command_shape` | YES (3) | YES |
| `merge_command_shape` ↔ `repo_argument` | YES (11) | YES |
| `repo_argument` ↔ `pr_fetch` | YES (1) | YES |
| `pr_fetch` ↔ `review_verdicts` | YES (81) | YES |
| `review_verdicts` ↔ `reviewer_threshold` | YES (80) | YES |
| `reviewer_threshold` ↔ `tech_debt_attestation` | **no** | YES |
| `tech_debt_attestation` ↔ `board_sync` | **no** | YES |
| `board_sync` ↔ `allow_advisories` | **no** | YES |

Row 1 is a genuine equivalent mutant, not a gap — both gates return `_Stop(None)` on independent conditions, neither mutates the context or calls anything, so the order is unobservable. Probed with 5 inputs including a non-Bash tool whose command contains `--admin`: identical result and identical trace under both orders.

**The last three rows are a real finding.** The suite pins *what* blocks but not *when*, for exactly the tail of the sequence where "when" is load-bearing: a threshold shortfall reported as a missing-attestation shortfall; `ensure_issues_on_board` writing to the project board on a PR that is then blocked; and the board write being skipped entirely on PRs that carry an advisory. Filed as **#1336** with the three suggested order-pinning tests, rather than fixed here — the fix is a `.claude/hooks/tests/` edit and this issue's acceptance criterion was zero test edits with #1333/#1117 concurrently rewriting that directory. Not a regression introduced by this PR; the pre-extraction `check()` had the same unpinned ordering, less visibly.

## LOC — the issue says LOC-neutral; it is not

| | before | after |
|---|---|---|
| `check()` | 619 lines (L2331–2949) | 18 (docstring + 6-statement loop) |
| module | 2,957 | 3,173 (**+216**) |
| largest new unit | — | `_gate_allow_advisories`, 114 lines (the #1211 rationale comment is ~60 of them) |

The +216 is scaffolding, not new logic: 11 `def` headers with type annotations and docstrings, the `_Stop`/`_GateContext`/`_gate_order_block` trio (65 lines), 5 producer-guard branches, the `_GATES` tuple, and the module-level comment stating why order is behaviour. The audit's "LOC-neutral" was written against a 435-line `check()`; the function was 619 lines by the time this landed.

## Verification

- `.claude/hooks/tests/` — **zero edits**. `git diff --stat`: 1 file changed, 396 insertions(+), 180 deletions(-), all in `.claude/hooks/validate_pr_review.py`.
- Full suite `.claude/hooks/tests/ .claude/lib/tests/`: **4084 passed, 881 subtests**.
- `ruff check` / `ruff format --check` / `mypy --ignore-missing-imports` over `.claude/hooks/ .claude/lib/`: clean.
- `pre-commit run --all-files` and `--hook-stage pre-push`: all 17 hooks pass (incl. cspell, actionlint, mypy-skills, pytest-skills).
- `python3 .claude/lib/pre_commit_ci_sync.py .`: `OK: pre-commit config mirrors all CI-enforced checks.`

A green suite here is a floor, not a proof — it is the same suite that let `check()` grow from 435 to 619 lines. The differential harness and the mutation table are the actual evidence, and the mutation table is where it says the suite is thin.

### DIFF
--- a/.claude/hooks/validate_pr_review.py
+++ b/.claude/hooks/validate_pr_review.py
@@ -258,6 +258,7 @@ import os
 import re
 import subprocess
 import sys
+from collections.abc import Callable
 from datetime import datetime
 from pathlib import Path
 
@@ -2328,31 +2329,121 @@ def resolve_review_verdicts(pr_data: dict, repo: str | None = None) -> ReviewVer
     )
 
 
-def check(input_data: dict) -> dict | None:
-    """Check PR review requirements.
+# ---------------------------------------------------------------------------
+# Gate decomposition (#1123)
+#
+# `check()` is a FIRST-BLOCKER-WINS gate: its behaviour is defined not only by
+# which conditions block, but by which one blocks FIRST — that selects the
+# message the operator sees, and where gates interact it selects whether
+# anything blocks at all (an `--admin` override ahead of the batch-loop guard;
+# the batch-loop guard ahead of the `is_merge_command` early-exit; the
+# TechDebt block ahead of the board-sync side effect). The sequence below is
+# that order, made explicit.
+#
+# Making it explicit also makes it EDITABLE, which the inline version was not.
+# Three gates ESTABLISH state later gates read — `_gate_merge_command_shape`
+# (PR target), `_gate_pr_fetch` (`pr_data`), `_gate_review_verdicts`
+# (`verdicts`) — and one, `_gate_board_sync`, is a pure side effect that never
+# stops. Those four are the complete set of order-coupled gates; every other
+# gate is a read-only predicate over `_GateContext`. A reorder that moves a
+# consumer ahead of its producer is caught by `_gate_order_block`, which
+# BLOCKS rather than raising (see its docstring).
+# ---------------------------------------------------------------------------
 
-    Returns a `{"decision": "block", "reason": …}` dict when the merge is
-    stopped, a `{"decision": "allow", "systemMessage": …}` dict when the merge
-    proceeds but something is worth stating (see the allow-path advisories at
-    the tail — main#1055 unparseable TechDebt, #1211 scan-mode disclosure), and
-    `None` when the merge proceeds with nothing to report.
 
-    Callers must therefore NOT read a non-`None` return as "blocked": `main()`
-    branches on `decision`, and an advisory exits 0.
+@dataclasses.dataclass(frozen=True)
+class _Stop:
+    """A gate's decision to END `check()`; `result` IS `check()`'s return value.
+
+    Three outcomes have to stay distinguishable, and `dict | None` alone can
+    only express two:
+
+      `None`                       — no blocker HERE; run the next gate.
+      `_Stop(None)`                — allow, nothing to report (exit 0).
+      `_Stop({"decision": ...})`   — block, or allow with a `systemMessage`.
+
+    Collapsing the first two — the mistake a bare `dict | None` gate signature
+    invites — would turn "not my concern" into "merge approved" for every gate
+    that declines to fire, i.e. it would fail open on all of them.
     """
-    tool_name = input_data.get("tool_name", "")
-    if tool_name != "Bash":
-        return None
 
-    command = input_data.get("tool_input", {}).get("command", "")
+    result: dict | None
 
+
+@dataclasses.dataclass
+class _GateContext:
+    """State threaded through the gate sequence.
+
+    Populated progressively: `tool_name`/`command` at construction, the PR
+    target by `_gate_merge_command_shape`, `pr_data` by `_gate_pr_fetch`, and
+    `verdicts` by `_gate_review_verdicts`. Consumers guard their producer
+    explicitly (`if ctx.verdicts is None: return _gate_order_block(...)`)
+    rather than trusting the tuple order.
+    """
+
+    tool_name: str
+    command: str
+    pr_number: str | None = None
+    repo: str | None = None
+    pr_display: str = ""
+    pr_data: dict | None = None
+    verdicts: ReviewVerdicts | None = None
+
+    @classmethod
+    def from_input(cls, input_data: dict) -> "_GateContext":
+        return cls(
+            tool_name=input_data.get("tool_name", ""),
+            command=input_data.get("tool_input", {}).get("command", ""),
+        )
+
+
+def _gate_order_block(ctx: _GateContext, consumer: str, producer: str) -> _Stop:
+    """Fail-CLOSED response to a gate that ran before its producer (#1123).
+
+    Unreachable while `_GATES` holds its declared order — it exists precisely
+    because the decomposition made that order editable, and the cheap failure
+    is the silent one: a verdict-consuming gate that read an unresolved
+    `verdicts` as "nothing to report" would ALLOW the merge, which is the
+    fail-open direction #950 and #981 were both filed for.
+
+    It BLOCKS rather than raising because raising would not block:
+    `_hook_main.run_blocking` catches every `Exception` out of `check()`, logs
+    it, and `sys.exit(0)` — allow. An exception here would therefore be the
+    fail-open it is meant to prevent.
+    """
+    reason = (
+        f"BLOCKED: PR {ctx.pr_display or '(unknown)'} — internal gate-ordering defect: "
+        f"`{consumer}` ran before `{producer}` established the state it reads, so the "
+        "2-reviewer gate was never evaluated.\n"
+        "This is a hook defect, not a review shortfall — report it rather than asking "
+        "reviewers to re-approve. It blocks because a gate that did not run must never "
+        "read as a gate that passed (#1123).\n"
+        "Pass `--admin` for emergency overrides only."
+    )
+    log_pretooluse_block("validate_pr_review", ctx.command, reason)
+    return _Stop({"decision": "block", "reason": reason})
+
+
+def _gate_non_bash_tool(ctx: _GateContext) -> _Stop | None:
+    """Gate 1 — this hook only inspects Bash commands."""
+    if ctx.tool_name != "Bash":
+        return _Stop(None)
+    return None
+
+
+def _gate_admin_override(ctx: _GateContext) -> _Stop | None:
+    """Gate 2 — the `--admin` emergency override."""
     # `--admin` is the emergency override — it bypasses the whole gate
     # (including the batch-loop guard below), matching the existing
     # short-circuit semantics. Checked first so an explicit admin override
     # is honored even for the loop shape.
-    if "--admin" in command:
-        return None
+    if "--admin" in ctx.command:
+        return _Stop(None)
+    return None
 
+
+def _gate_batch_loop_merge(ctx: _GateContext) -> _Stop | None:
+    """Gate 3 — the batch-loop merge guard, which MUST precede gate 4."""
     # Batch-loop merge guard (#567/#894/#897): a `gh pr merge` with no resolvable
     # literal PR number inside a for/while loop fail-opens the 2-reviewer gate.
     # That covers a loop-variable arg (`$pr`, `${prs[$i]}`, `$(get_pr)`) AND a
@@ -2370,7 +2461,7 @@ def check(input_data: dict) -> dict | None:
     # returns False and `check()` would early-exit — re-opening the exact
     # fail-open hole this guard closes. Detecting the loop shape here is the
     # only placement that fires on it.
-    if is_variable_pr_merge_in_loop(command):
+    if is_variable_pr_merge_in_loop(ctx.command):
         result = {
             "decision": "block",
             "reason": (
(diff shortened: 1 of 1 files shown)


parametrization:
**Review: LGTM**

Second review (Release Coordinator angle — operational/message-fidelity, per team-lead brief). Bereket reviews independently as Opus merge-gate.

## Mechanical verification

**Message fidelity.** AST-parsed both `base_validate_pr_review.py` (2136fee) and `head_validate_pr_review.py` (3dadd51) fetched via `gh api .../contents/...?ref=<sha>`, diffed the full multiset of every string-literal constant (including f-string literal segments). Zero strings present in base are missing/reduced in head — every string that appears only in head is new scaffolding (docstrings, gate names, the new `_gate_order_block` error text), not a replacement for an existing block/advisory message. Also spot-checked composition *order*, not just presence: the threshold-block diagnostic concatenation `scan_diagnostic + stale_diagnostic + roster_diagnostic + f"BLOCKED..."` is byte-identical in both files.

**#1336 disposition.** Diffed the base file's inline `check()` body directly: board-sync (`ensure_issues_on_board`) already ran after the threshold + tech-debt gates and before the advisories block in the pre-refactor code — same order, same position. So the three "unpinned ordering" scenarios in #1336 are risks introduced by *making the sequence reorderable*, not live bugs in the shipped order today. Confirmed concretely: swapped `_gate_board_sync` ↔ `_gate_allow_advisories` in a clean worktree checkout at 3dadd51, reran `.claude/hooks/tests/test_validate_pr_review.py` — 363 passed / 103 subtests, fully green, exactly matching the row's claimed "no" in the mutation table. Restored the worktree, confirmed clean. `_gate_order_block` fails closed if a producer/consumer pair is ever actually reordered, and #1336 is filed with the concrete order-pinning tests needed — deferring past #1333/#1117 is correct, not just convenient.

**`_gate_board_sync`** — confirmed non-skippable/non-double-runnable within one `check()` call in the current order; `ensure_issues_on_board` itself is untouched by this PR and already catches its own subprocess exceptions.

**Diagnostic builders** (`_stale_verdict_diagnostic`, `_non_roster_requestor_diagnostic`, `_scan_mode_diagnostic`) — each pulls every field it needs as an attribute off `verdicts`; nothing stranded on an accumulated local that didn't make the trip.

**Advisory path** — `_gate_allow_advisories` only returns `_Stop({"decision": "allow", "systemMessage": ...})` when advisories is non-empty, else `None`; matches base's `if advisories: return {...}` exactly.

**LOC** — independently reproduced: 2957→3173 (+216), `check()` 619→18, `_gate_allow_advisories` 114 lines. All match the PR body's table exactly.

## CI
Verified via `python3 .claude/lib/pr_ci_state.py --repo noorinalabs/noorinalabs-main 1337` and `statusCheckRollup` at head SHA 3dadd51: 14/14 SUCCESS. Also independently reran `ruff check`, `ruff format --check`, `mypy --ignore-missing-imports`, and the test file itself in a clean worktree — all clean.

Must-fix: None
Tech-debt: #1339 (file-size/module-split observation — the module is now 3173 lines; not urgent, sequenced after #1336)

---
Requestor: Santiago Ferreira
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: #1339

parametrization:
**Review: Approved** — Opus merge gate (Bereket Tadesse, Infrastructure Manager)

Reviewed at head `3dadd51c`, diffed against base `2136fee3` via the contents API (base == merge-base, so no staleness). I reviewed this specifically because I authored `_hook_main.run_blocking` in #1330 and the central safety argument rests on its exception behaviour. Every table in the body was reproduced rather than read.

## The fail-closed argument holds, and I verified it end-to-end

`_hook_main.run_blocking:145-148` catches every `Exception` out of `check()`, logs it to the `pretooluse_diagnostic` trace channel, and `sys.exit(0)`. `validate_pr_review.main()` calls `run_blocking(check, ...)` directly at L3168 — no dispatcher in between — so the path is real and applies here. I ran the counterfactual rather than reasoning about it: misordered `_GATES` (consumer before producer), driven through the real `main()` with stdin JSON.

| variant | exit code | meaning |
|---|---|---|
| as shipped (`_gate_order_block` **blocks**) | **2** | BLOCK |
| counterfactual (`_gate_order_block` **raises**) | **0** | ALLOW |

A raise really would be the fail-open the guard exists to prevent. The docstring rationale is correct as written.

## Independent differential harness

I did not reuse the PR harness. I built my own, and traced **every** call via `sys.settrace` rather than a hand-listed set of callees — strictly stronger than an 11-name list, since a name omitted from such a list is exactly the degree of freedom that goes unpinned.

| | reported | my run |
|---|---|---|
| corpus | 54 | **61** |
| result + trace mismatches | 0/54 | **0/61** |
| coverage of pre-extraction `check()` | 184/185 | **184/185** |
| single uncovered line | the `def` line | the `def` line (L2331) |

Coverage measured independently off the base code object line table. My first run showed 183/185 and 6 trace mismatches; both were artifacts on my side — the 6 were `_shell_parse._tokenize_cached` **lru_cache warming** (base runs first, head hits the warm cache), and the missing line was a corpus gap on an empty-roster ternary branch. Clearing the caches and adding three scenarios gives 61/61 and 184/185. Flagging this because a reviewer who saw those 6 and stopped would have reported a false mismatch.

**Traced callee set is complete.** My full capture found exactly three effectful callees — `ensure_issues_on_board`, `log_pretooluse_block`, `subprocess.run` — all three in your list. No file writes, no `os`/env access, no other network in the gate region. Your list is a superset of every writer; my extras are pure callees.

## Mutation table reproduces exactly

All 10 adjacent swaps, full suite (baseline **4084 passed, 881 subtests** — matches). Every row reproduced **including the failure counts**: 1, 3, 11, 1, 81, 80.

## The three tail swaps are pre-existing — confirmed by controlled experiment

Your characterization is right, and because the PR edits zero tests I could test it directly rather than argue it: I applied the same three reorderings to the **inline pre-extraction body** and ran the unchanged suite against it.

| variant | suite |
|---|---|
| CONTROL: base, unmodified inline order | GREEN 4084 |
| base, threshold <-> tech_debt swapped | GREEN 4084 |
| base, tech_debt <-> board_sync swapped | GREEN 4084 |
| base, board_sync <-> allow_advisories swapped | GREEN 4084 |

The extraction **reveals** the exposure, it does not create it. Filing as #1336 rather than fixing here was the right call. (The control also independently confirms the zero-test-edits claim: the base file passes this branch suite unchanged.)

## Row 1: right conclusion, slightly overstated reason

I probed it with 12 adversarial inputs rather than 5, and **found 3 that distinguish the two orders** — so it is not a strict equivalent mutant at the `check()` contract level:

```
non-Bash tool + command None : declared -> None (allow)   swapped -> TypeError
non-Bash tool + command int  : declared -> None (allow)   swapped -> TypeError
```

The asymmetry your reasoning misses: `_gate_non_bash_tool` predicate is total over the type of `command`, `_gate_admin_override` predicate (`"--admin" in ctx.command`) is not. Whichever runs first short-circuits, so order **is** observable.

**But your conclusion stands at the boundary that matters.** At the real hook boundary both orders exit **0** on every one of those inputs (`run_blocking` converts the TypeError to allow), and a non-str `command` only reaches a non-Bash tool through malformed input — real tool calls carry no `command` key at all, which yields the `""` default and is identical under both orders. A bytes probe is not even expressible, since hook input arrives as JSON. So: equivalent in consequence, not in mechanism. No change requested; recording it so the next person does not re-derive it.

## New finding — filed #1340, non-blocking

Your mutation testing covered gate **order**. I also mutated the **`_Stop` sentinel at every gate return site**, which is the other degree of freedom this refactor introduces — and the one your own body calls "the single most dangerous confusion available in this refactor".

| direction | sites | killed | survives |
|---|---|---|---|
| `None` -> `_Stop(None)` (**fail-OPEN**) | 11 | **10** | 1 |
| `_Stop(None)` -> `None` (**fail-CLOSED**) | 3 | 1 | **2** |

The reassuring half is the important half: the **fail-open** direction — a gate that should continue instead allowing and skipping every later gate — is killed on 10 of 11 gates, at 57-79 failures. The sentinel does the job you designed it to do.

The survivors, characterized against my harness:

- **`_gate_merge_command_shape` L2502** (`_Stop(None)` -> `None`): suite fully GREEN, but **4/61 corpus inputs change result** — every non-merge Bash command (`ls -la && echo hi`, `git merge main`, PR list/view) falls through to the fetch path and is BLOCKED, with a malformed empty `PR ` display. Total-outage-shaped, invisible to 4084 tests.
- **`_gate_non_bash_tool` L2430**: GREEN, 1/61 result-diff.
- **`_gate_allow_advisories` L3128** (`None` -> `_Stop(None)`): GREEN and a **genuine equivalent mutant** — it is the last entry in `_GATES`, so both spellings terminate `check()` with `None`. 0/61 on result and trace. No test should be written for it.

Both real survivors fail **CLOSED** (over-blocking, not a gate bypass), the shipped code is correct, and the fix is a `.claude/hooks/tests/` edit that #1123 acceptance criterion forbids while #1333/#1117 are rewriting that directory. Not blocking. Sequence #1340 with #1336.

## Also verified

- **Message literals byte-identical**: AST-extracted every string constant from base `check()` and the head gate region. **Zero** base literals missing from head; all 17 head-only literals belong to the new `_gate_order_block` message and gate-name strings. A silently reworded `reason` would have shown here.
- **`_Stop` return-site mapping is 1:1**: base has exactly 3 terminal-allow `return None` sites (non-Bash, override, non-merge-shape) and 8 `return result`; head has exactly 3 `_Stop(None)` and 8 `_Stop(result)`. No gate returns a bare `None` where `_Stop(None)` was meant.
- **LOC table exact**: `check()` 619 -> 18, module 2957 -> 3173 (+216), largest new unit `_gate_allow_advisories` at 114 lines.
- **Diff scope**: one file; the only change outside the gate region is the added `Callable` import at L261.
- **CI**: 14/14 SUCCESS at head `3dadd51c`, re-checked immediately before posting. (`mergeable` currently reads UNKNOWN — GitHub recomputing, was MERGEABLE; worth a glance before the merge press.)

Unusually strong evidence, and it held up under independent reproduction — including the parts where it says the suite is thin. The one thing I would not want lost: the differential harness and mutation table are what make this reviewable at all, and neither is committed. Approving.

---

Requestor: Bereket Tadesse
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: #1340

Issue #1342, opened 2026-08-03
audit(G5 follow-up): remaining test clone groups not covered by #1117


parametrization:
Follow-up to #1117 (G5 — parametrize the test clone groups), which implemented a well-audited subset. This issue is the complete work-list for the remainder.

**Corrected 2026-08-03 (see #1344):** the original version of this issue claimed "199 total − 12 done − 14 excluded = 173 remaining" while its own itemized table summed to a different number. Fixed below by NOT hand-subtracting #1117's 12 implemented groups from the raw detector output — #1117 re-split some multi-class candidate rows into several smaller per-class groups rather than fully consuming a row 1:1 (see "Interaction with #1117's work" below), so a static subtraction is unreliable and was exactly the error #1344 caught. The reconciled numbers below are RAW detector output only, with the arithmetic verified to add up.

## Method (re-derive before using this list — the tree will have moved)

AST-normalized clone detection, distinct from both #1117's cited audit figure (92 groups / 2,457 lines) and the #1116 author's own crude pass (159 / 5,418): each `def test_*` (module-level function or `unittest.TestCase`/plain-class method) is parsed, its body's `ast.Constant` LEAF values are replaced with type-tagged placeholders (`<STR>`, `<NUM>`, ...) — type and statement position preserved, only the leaf value erased — and the resulting shape is hashed per-FILE. A group requires: size ≥ 3, identical shape hash, AND identical assert-bearing-statement count across every member (a differing assert count is split into its own subgroup, never silently merged). This is deliberately less aggressive than a full string-strip (which conflates "the literal is an interchangeable parametrize value" with "the literal is the semantic point of the test" whenever two functions share call-shape but test unrelated scenarios) — see #1117's PR body for the full rationale and its mutation-tested verification method.

Neither the 92/2,457 nor the 159/5,418 nor this issue's 199/5,592 figure is "the" answer — they're different, defensible normalizations. Re-run the detector against the current tree before treating any of these numbers as current; the counts move as the suite grows.

## Reconciled totals (raw detector output, `e39e38e`, verified to add up)

- **199 groups / 5,592 lines** total, across 59 files.
- **14 groups / 387 lines** in `.claude/hooks/tests/test_validate_pr_review.py` — **excluded from this work-list**, see below.
- **185 groups / 5,205 lines** remain as candidate rows across the other 58 files.
- Top-20 files by line count: **114 groups / 4,062 lines**.
[...]

Issue #1089, opened 2026-07-26
Code Audit 2026-07 program — org-wide DRY / LOC / runtime-memory (umbrella)


parametrization:
Umbrella meta-issue for the org-wide code audit program. **Source of truth: `CODE_AUDIT.md` on `main` (merged via #1088 @ `3bdd620`)** — every child issue carries its task block verbatim from there; do not restate or re-derive baselines elsewhere.

Program shape (one epic per track, filed as sub-issues of this umbrella):
- **Track 0 — fix-now defects** (12 bugs, §3) — correctness/security/perf defects, independent of refactors
- **Track A — shared org libraries** (§4) — `noorinalabs-common`: `noorina-schema` / `noorina-pipeline-core` / `noorina-devtools`; gated on the decision issue
- **Track B — twin-repo resolution** (§5) — data-acquisition ↔ ingest-platform; B3 gated on the #978 cutover
- **Track C — CI/CD & workflow dedup** (§6)
- **Track D — isnad-graph** (§7) · **Track E — user-service** (§8) · **Track F — pipeline runtime & memory** (§9) · **Track G — parent repo hooks/lib** (§10) · **Track H — design-system & landing-page** (§11)

Sequencing & dependency edges: `CODE_AUDIT.md` §12 (machine-readable edge list). Consolidated 82-row task index: §13.

Cross-referenced pre-existing debt this program **feeds, not duplicates**: #1019 (lifecycle consolidation — scoped by G7), #1021/#1037/#1014 (P10 tooling debt), deploy#285, user-service#84.

Board note: issues in this program were filed via the GitHub API without project-board placement (no Projects tooling in the filing session) — run `/board-audit` (or add cards manually) to place them on project 2.

PR #1343, opened 2026-08-03
audit(#1117): parametrize 12 test clone groups across 4 hook test files

### LINKED ISSUE #1117: audit(G5): parametrize the 92 test clone groups
Source: CODE_AUDIT.md §10 (noorinalabs-main @ 3bdd620) — item G5.

**G5 — Parametrize the 92 clone groups** · Type: DRY (tests) · Effort: M–L
AST-normalized comparison found **92 groups of ≥3 near-identical test functions = 2,457 lines** across hooks/lib/skills tests (largest: 12× in `test_annunaki_monitor`, 11× in `test_enforce_ontology_context`, 11× in `test_validate_commit_identity`). The in-file loop idiom already exists — apply it consistently (`pytest.mark.parametrize` / `subTest`).
*After:* **≈ −1,400 to −1,700 lines** at unchanged assertion count (params keep case ids). *Verify:* per-file `--collect-only` id count equal; one mutation spot-check (revert a hook fix → suite still fails).

Index row (§13): T | −1,400–1,700 LOC | M–L | G4

parametrization:
## Summary

G5 of the code-audit program (#1089). Parametrizes 12 test clone groups (97 merged cases) across 4 files: the 3 the issue names explicitly (`test_annunaki_monitor`, `test_enforce_ontology_context`, `test_validate_commit_identity`) plus `test_block_bare_grep` (biggest single group, 23 members). Net **-114 lines at `2f3a2bf`** (`git diff --shortstat b191a09..2f3a2bf`: 588 insertions, 702 deletions) — stated with the head it was measured against, since two review-fix commits moved this number twice (-130 at the original push, -124 at `d1a3347d`).

## Numbers — both audit figures treated as unverified, re-derived independently

- Audit's cited figure: 92 groups / 2,457 lines.
- #1116 author's own crude pass (full string-strip, deliberately over-groups): 159 groups / 5,418 lines.
- **My own re-derivation, method below: 199 groups / 5,592 lines.**

None of the three is authoritative — they're different normalizations. Mine: parse each `def test_*`, replace every `ast.Constant` LEAF value with a type-tagged placeholder (`<STR>`/`<NUM>`/...) while preserving the constant's *type* and *position* in the statement sequence — everything else (names, calls, control flow, statement order) untouched. A group additionally requires uniform assert-bearing-statement count across all members (a differing count is split into its own subgroup, never silently merged). This is less aggressive than a full string-strip specifically to avoid the trap described below.

## The trap, and how this PR avoided it

Structurally identical is not the same as safely mergeable. Two failure modes avoided:

1. **Cross-class shape collisions.** The detector groups by AST shape alone, ignoring class boundaries. `test_block_bare_grep.py`'s `NegativeMatchTests` (12 members: "not a grep invocation") and `OverrideTests` (2 members: "the documented escape hatch") both reduce to `self.assertFalse(_blocks(cmd))` — same shape, different scenario. Declined every cross-class merge; each parametrize table stays within its original class, whose docstring names one coherent scenario.
2. **`pytest.mark.parametrize` is a silent no-op on `unittest.TestCase` methods.** Verified empirically before converting anything (a probe class collected 1 test instead of 3, no error). Every touched class was therefore converted to plain pytest style — every `self.assertX`, merged or not, mechanically rewritten to a bare `assert`. `test_annunaki_monitor.py`'s `TestSilentBooleanIdiom` additionally required moving `setUp`/`tearDown` to an autouse fixture (all its methods, in and out of the merged group, share that isolation).

## The verification bar

1. **Collected count** — per-file `--collect-only`, identical before/after for all 4 files; repo-wide **4084 -> 4084**.
2. **Node-ID identity** — full path-string identity is impossible once N methods merge into 1 (the function name necessarily changes). The invariant actually enforced: every `ids=` entry is the ORIGINAL bare method name, so each case's new node id (`ClassName::test_x[original_method_name]`) is traceable 1:1 back to what it replaced. Verified per file by diffing the sorted id lists' bracket-suffix set against the removed method-name set.
3. **Assertion density** — NOT raw static AST-text count (that necessarily drops under any real parametrize — collapsing N duplicate `assert` lines into 1 executed N times IS the point). The invariant: execution-weighted density = Σ(body-assert-count × case-count) over every collected id, compared before/after with the same script. All 4 files: **44->44, 47->44 (a disclosed, documented -3: a redundant duplicate not-None check that had no pytest equivalent once `self.assertX` was gone, see `TestIndentedCoordinatorOpenerNotExempt`'s inline comment), 134->134, 193->193.**
4. **Mutation spot-check per group** (4 total, one per file): reverted a real fix (`sudo -i` boolean-flag regression in `block_bare_grep.py`; the `#471` opener-anchor regression in `enforce_ontology_context.py`; dropping `pgrep` from `annunaki_monitor.py`'s idiom list; reintroducing main#1195's own-prefix-assignment bug in `_shell_parse.py`) and confirmed the suite went red specifically via the merged parametrized case, then reverted.
5. **Direct standalone invocation matches `--collect-only`** (added in review — see below): all 4 files' `python3 <file>.py` executed count equals its collected count.

   | File | Direct-run executed | `--collect-only` |
   |---|---|---|
   | `test_block_bare_grep.py` | 41 | 41 |
   | `test_enforce_ontology_context.py` | 37 | 37 |
   | `test_annunaki_monitor.py` | 86 | 86 |
   | `test_validate_commit_identity.py` | 124 | 124 |

## Review fixes (both batched, one push each)

- **Mypy, CI-only.** `_valid_identity`'s `if not name: pytest.skip(...)` relied on `pytest.skip`'s `NoReturn` annotation to narrow `name: str | None` to `str`. CI's mypy job installs only `mypy types-PyYAML bashlex` (no pytest), so that stub is invisible there and the narrowing never happens — passed locally (pytest importable), failed CI. Replaced with an explicit `assert name is not None` that doesn't depend on the stub. Verified against CI's exact dependency set in an isolated venv (mypy 2.3.0, no pytest installed): exit 0.
- **Silent under-run on direct invocation.** Three files still ended with `unittest.main()` after their converted classes went plain-pytest; `unittest.main()` only discovers `TestCase` subclasses, so a bare `python3 <file>.py` silently ran fewer tests than `--collect-only` and still exited 0 (37→17 run, 124→87, 86→69 — no error, no skip notice). `test_block_bare_grep.py` already had the fix from an earlier commit in this PR (`pytest.main([__file__, "-v"])`); applied the same pattern to the other three and verified per the table above.

## Scope discipline

- `test_validate_pr_review.py` untouched — was mid-review on #1337 at spawn time; merged as `4b550cc` during this session, so the file moved out from under scope entirely. Its 14 candidate groups are called out, not attempted.
- Remaining candidate groups filed as #1342 (185 non-excluded groups across 58 files, reconciled to add up — an earlier version of that issue had an internal arithmetic mismatch, fixed after #1344 caught it) with the full per-file breakdown, the same method documented above, and the mutation-testable + direct-invocation verification bar for whoever continues — including a note that `test_validate_workflow_paths_coverage.py` is flagged (same "literal payload is the semantic point" risk) rather than blanket-excluded.

## Test plan

- [x] `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests .claude/lib/tests` — 4084 passed, 881 subtests passed
- [x] `ruff check` / `ruff format --check` — clean
- [x] `mypy .claude/hooks .claude/lib` — clean, and independently confirmed against CI's exact pytest-free dependency set
- [x] Direct invocation of all 4 touched files matches `--collect-only` count (table above)
- [x] Pre-push hooks (mypy, pytest, memory-budget, headcount-budget, doc-freshness, office-drift, mermaid-render) — all passed on push

Closes #1117

### DIFF
--- a/.claude/hooks/tests/test_annunaki_monitor.py
+++ b/.claude/hooks/tests/test_annunaki_monitor.py
@@ -21,6 +21,7 @@ import _test_helpers  # noqa: E402,F401
 import annunaki_log as alog  # noqa: E402
 import annunaki_monitor as am  # noqa: E402
 import annunaki_parse as ap  # noqa: E402
+import pytest
 
 
 def _bash_event(command: str, stdout: str = "", stderr: str = "", exit_code: int = 0) -> dict:
@@ -168,31 +169,42 @@ class AnnunakiMonitorTests(unittest.TestCase):
         self.assertIsNone(result)
 
 
-class SilentBooleanTestIdiomTests(unittest.TestCase):
+class TestSilentBooleanIdiom:
     """#474 coverage: documented boolean-test idioms whose by-design failure
     branch is non-zero exit with empty output must NOT produce log entries
     after #473 closed the silent-failure capture path. Each idiom gets a
     NEGATIVE-match test (no log) plus a precedence test that confirms a real
-    stderr-pattern match still wins."""
+    stderr-pattern match still wins.
 
-    def setUp(self):
-        self._saved_env = {
+    Not `unittest.TestCase` (#1117 — G5 clone-group parametrization):
+    `test_negative_match` below is a `pytest.mark.parametrize` table — 12
+    members were a single-statement `self.assertIsNone(am.check(_bash_event(
+    cmd, exit_code=n)))` differing only in `cmd`/`exit_code`. Converting the
+    class required moving `setUp`/`tearDown` to an autouse fixture (the
+    other methods in this class, including the ones NOT in the clone group,
+    depend on the same `ERRORS_FILE`/`_seen_hashes` isolation); every
+    `self.assertX` call in the class — merged or not — was mechanically
+    rewritten to a bare `assert` as a consequence, since a plain class has no
+    `self.assertX`. `ids=` is each original method's bare name."""
+
+    @pytest.fixture(autouse=True)
+    def _isolate_annunaki_state(self):
+        saved_env = {
             "ENVIRONMENT": os.environ.pop("ENVIRONMENT", None),
             "NOORIN_HOOK_TEST_MODE": os.environ.pop("NOORIN_HOOK_TEST_MODE", None),
         }
         am._seen_hashes.clear()
-        self._tmpdir = tempfile.TemporaryDirectory()
-        self._errors_path = Path(self._tmpdir.name) / "errors.jsonl"
-        self._orig_monitor_file = am.ERRORS_FILE
-        self._orig_log_file = alog.ERRORS_FILE
+        tmpdir = tempfile.TemporaryDirectory()
+        self._errors_path = Path(tmpdir.name) / "errors.jsonl"
+        orig_monitor_file = am.ERRORS_FILE
+        orig_log_file = alog.ERRORS_FILE
         am.ERRORS_FILE = self._errors_path
         alog.ERRORS_FILE = self._errors_path
-
-    def tearDown(self):
-        am.ERRORS_FILE = self._orig_monitor_file
-        alog.ERRORS_FILE = self._orig_log_file
-        self._tmpdir.cleanup()
-        for k, v in self._saved_env.items():
+        yield
+        am.ERRORS_FILE = orig_monitor_file
+        alog.ERRORS_FILE = orig_log_file
+        tmpdir.cleanup()
+        for k, v in saved_env.items():
             if v is None:
                 os.environ.pop(k, None)
             else:
@@ -203,78 +215,58 @@ class SilentBooleanTestIdiomTests(unittest.TestCase):
     def test_posix_bracket_silent_exit_not_logged(self):
         """NEG: `[ -d /nonexistent ]` exit 1 silent → no log."""
         result = am.check(_bash_event("[ -d /nonexistent ]", exit_code=1))
-        self.assertIsNone(result)
-        self.assertFalse(self._errors_path.exists())
+        assert result is None
+        assert not self._errors_path.exists()
 
-    def test_bash_double_bracket_silent_exit_not_logged(self):
-        """NEG: `[[ -f /nonexistent ]]` exit 1 silent → no log."""
-        result = am.check(_bash_event("[[ -f /nonexistent ]]", exit_code=1))
-        self.assertIsNone(result)
-
-    def test_test_builtin_silent_exit_not_logged(self):
-        """NEG: `test -f /nonexistent` exit 1 silent → no log."""
-        result = am.check(_bash_event("test -f /nonexistent", exit_code=1))
-        self.assertIsNone(result)
-
-    def test_grep_q_no_match_silent_exit_not_logged(self):
-        """NEG: `grep -q pattern file` non-error miss → no log."""
-        result = am.check(_bash_event("grep -q needle /tmp/haystack.txt", exit_code=1))
-        self.assertIsNone(result)
+    @pytest.mark.parametrize(
+        ("command", "exit_code"),
+        [
+            ("[[ -f /nonexistent ]]", 1),
+            ("test -f /nonexistent", 1),
+            ("grep -q needle /tmp/haystack.txt", 1),
+            ("pgrep nginx", 1),
+            ("pkill -0 nginx", 1),
+            ("which nonexistent-binary", 1),
+            ("command -v nonexistent-binary", 1),
+            ("diff --quiet /tmp/a /tmp/b", 1),
+            ("git diff --quiet", 1),
+            ("[ -f /tmp ]", 0),
+            # NEG: `if [ -f /nonexistent ]; then echo unreachable; fi` exits 0
+            # (false branch, body skipped) → no log. The whole-`if` regex was
+            # removed in #490 because it silenced body failures (gap #3); the
+            # false-branch path is now handled by exit_code=0 not being an
+            # error, not by idiom-skip.
+            ("if [ -f /nonexistent ]; then echo unreachable; fi", 0),
+            # NEG: same shape with `[[`. exit 0 → no log.
+            ("if [[ -f /nonexistent ]]; then echo unreachable; fi", 0),
+        ],
+        ids=[
+            "bash_double_bracket_silent_exit_not_logged",
+            "test_builtin_silent_exit_not_logged",
+            "grep_q_no_match_silent_exit_not_logged",
+            "pgrep_no_match_silent_exit_not_logged",
+            "pkill_no_match_silent_exit_not_logged",
+            "which_not_found_silent_exit_not_logged",
+            "command_v_not_found_silent_exit_not_logged",
+            "diff_quiet_differs_silent_exit_not_logged",
+            "git_diff_quiet_dirty_silent_exit_not_logged",
+            "idiom_exit_zero_not_logged",
+            "if_bracket_conditional_false_branch_exit_zero_not_logged",
+            "if_double_bracket_conditional_false_branch_exit_zero_not_logged",
+        ],
+    )
+    def test_negative_match(self, command, exit_code):
+        """NEG: a documented boolean-test idiom's by-design silent branch
+        (exit-1 miss, or exit-0 false-branch) must not produce a log entry."""
+        result = am.check(_bash_event(command, exit_code=exit_code))
+        assert result is None
 
     def test_grep_q_with_flags_silent_exit_not_logged(self):
         """NEG: `grep -qE` / `grep -qi` variants → no log."""
         for cmd in ("grep -qE '^foo' f.txt", "grep -qi BAR f.txt", "grep -Eq '^foo' f.txt"):
             am._seen_hashes.clear()
             result = am.check(_bash_event(cmd, exit_code=1))
-            self.assertIsNone(result, f"{cmd!r} must not log")
-
-    def test_pgrep_no_match_silent_exit_not_logged(self):
-        """NEG: `pgrep nginx` not-running → no log."""
-        result = am.check(_bash_event("pgrep nginx", exit_code=1))
-        self.assertIsNone(result)
-
-    def test_pkill_no_match_silent_exit_not_logged(self):
-        """NEG: `pkill -0 nginx` not-running → no log."""
-        result = am.check(_bash_event("pkill -0 nginx", exit_code=1))
-        self.assertIsNone(result)
-
-    def test_which_not_found_silent_exit_not_logged(self):
-        """NEG: `which foo` not-installed → no log."""
-        result = am.check(_bash_event("which nonexistent-binary", exit_code=1))
-        self.assertIsNone(result)
-
-    def test_command_v_not_found_silent_exit_not_logged(self):
-        """NEG: `command -v foo` not-installed → no log."""
-        result = am.check(_bash_event("command -v nonexistent-binary", exit_code=1))
-        self.assertIsNone(result)
-
-    def test_diff_quiet_differs_silent_exit_not_logged(self):
-        """NEG: `diff --quiet a b` files differ → no log."""
-        result = am.check(_bash_event("diff --quiet /tmp/a /tmp/b", exit_code=1))
-        self.assertIsNone(result)
-
-    def test_git_diff_quiet_dirty_silent_exit_not_logged(self):
-        """NEG: `git diff --quiet` working tree dirty → no log."""
-        result = am.check(_bash_event("git diff --quiet", exit_code=1))
-        self.assertIsNone(result)
-
-    def test_if_bracket_conditional_false_branch_exit_zero_not_logged(self):
-        """NEG: `if [ -f /nonexistent ]; then echo unreachable; fi` exits 0
-        (false branch, body skipped) → no log. The whole-`if` regex was
-        removed in #490 because it silenced body failures (gap #3); the
-        false-branch path is now handled by exit_code=0 not being an error,
-        not by idiom-skip."""
-        result = am.check(
-            _bash_event("if [ -f /nonexistent ]; then echo unreachable; fi", exit_code=0)
-        )
-        self.assertIsNone(result)
-
-    def test_if_double_bracket_conditional_false_branch_exit_zero_not_logged(self):
-        """NEG: same shape with `[[`. exit 0 → no log."""
-        result = am.check(
-            _bash_event("if [[ -f /nonexistent ]]; then echo unreachable; fi", exit_code=0)
-        )
-        self.assertIsNone(result)
+            assert result is None, f"{cmd!r} must not log"
 
     # --- Precedence: pattern match wins over idiom-on-silent-exit ---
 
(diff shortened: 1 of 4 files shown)


parametrization:
Second review (independent of Nino's merge-gate pass). Angle: does the converted suite still run everywhere it used to — standalone, direct invocation, and `unittest discover`.

## Verified correct

- **Standalone per-file sweep** — `pytest <file>` individually across all 57 files: 56/57 pass. The 1 failure (`test_warn_oversized_brief.py`, `ModuleNotFoundError: _hook_main`) is **pre-existing at base** (`b191a09`) — reproduced it there too, unrelated to this PR and out of its scope.
- **Forward and reverse aggregate order** — `pytest .claude/hooks/tests .claude/lib/tests` in both file-name-sorted and reverse-sorted order: **4084 passed, 881 subtests** both directions. Matches the PR body's claim.
- **`unittest discover`** — 2700 → 2589 (**-111**). Per-file breakdown of the drop: `test_annunaki_monitor.py` -17, `test_block_bare_grep.py` -37, `test_enforce_ontology_context.py` -20, `test_validate_commit_identity.py` -37 = **111**, exactly accounting for the aggregate drop with no unexplained excess. Legitimate — matches methods actually converted off `TestCase`.
- **Scope containment** — `git diff --stat` between base and head touches exactly the 4 claimed files (575/-699, net -130, matching the PR body exactly). `test_validate_pr_review.py`, `_test_helpers.py`, and both `conftest.py` files have **zero** diff — confirmed untouched.
- **CI** — reran `gh pr checks` at head `d1a3347`: 14/14 green, including mypy. Confirmed the mypy fix commit (`d1a3347`, narrowing `_valid_identity` with a bare `assert` instead of relying on `pytest.skip`'s NoReturn stub) is sound — matches CI's actual dependency set (no pytest in the mypy job).
- **`test_block_bare_grep.py`'s mixed-style `__main__` handling** — this file gets it right: `pytest.main([__file__, "-v"])` with an inline comment explaining why plain `unittest.main()` would silently skip the converted classes. Direct `python3 test_block_bare_grep.py` invocation: all 41 tests run.

## Blocking finding

**3 of the 4 touched files still end with `unittest.main()`, and direct `python3 <file>.py` invocation now silently skips the converted classes while printing `OK`.** Measured directly:

| File | `python3 file.py` runs | pytest collects | Silently skipped |
|---|---|---|---|
| `test_annunaki_monitor.py` | 69 | 86 | **17** (`TestSilentBooleanIdiom`) |
| `test_enforce_ontology_context.py` | 17 | 37 | **20** (54% of the file) |
| `test_validate_commit_identity.py` | 87 | 124 | **37** (30% of the file) |

`unittest.main()` only discovers `unittest.TestCase` subclasses; a class converted to plain-pytest style is invisible to it and produces no error, no skip notice — the run just ends with `Ran N tests ... OK`. This isn't a hypothetical: `_test_helpers.py`'s own module docstring (from #1116, merged one PR ago) states the design contract explicitly — *"most files in this suite still support [a bare `python3 test_foo.py` invocation]... each carries its own `if __name__ == "__main__":` block"* — specifically because conftest fixtures are invisible to that path. This PR's 4th file (`test_block_bare_grep.py`) demonstrates the correct fix is a one-line, already-precedented change (`pytest.main([__file__, ...])`); the other 3 didn't receive it.

No CI job invokes files directly, so this wouldn't be caught by the green checks above — it's a genuine, silent regression in a documented capability, not a style nit.

## Non-blocking (filed as tech-debt, #1342 not #1343)

Cross-checked #1342's "173 remaining" summary math against its own itemized table: the top-20 breakdown sums to 113 non-excluded groups, and the table's own "~39 smaller files (2-3 groups each)" note implies 78-117 more — i.e. 191-230, not 173. Filed as #1344 (doesn't block this PR; #1343's own 12-group/4-file scope is independently verified correct regardless of how #1342 nets out).

---
Requestor: Weronika Zielinska
Requestee: Lucas Ferreira
TechDebt: #1344

parametrization:
Opus merge-gate review. Everything below was re-derived at head `d1a3347d` in a detached worktree against merge-base `b191a091` — I reimplemented each instrument rather than running the author's scripts.

## The three instruments — all three hold

**1. Collected count.** Repo-wide **4084 → 4084**. Per-file, all four identical: `test_annunaki_monitor` 86, `test_block_bare_grep` 41, `test_enforce_ontology_context` 37, `test_validate_commit_identity` 124.

**2. Node-ID identity.** **111** node ids move, not 97 — worth stating because the gap is not slack. Eight classes were *renamed* (`NegativeMatchTests` → `TestNegativeMatch`, …), which is forced: a plain class needs pytest's `Test*` prefix to be collected at all. A rename moves the id of every method in the class, including the 14 that were never merged. So 111 = 97 merged + 14 renamed-only.

Against that denominator I checked the author's substitute invariant mechanically — mapping each head case back to `"test_" + <id>` for parametrized cases, and to the unchanged function name for renamed-only ones:

| Property | Result |
|---|---|
| TOTAL — every removed method has a successor | PASS (0 missing) |
| INJECTIVE — each predecessor claimed exactly once | PASS (0 duplicate ids) |
| NO INVENTION — no id lacking a predecessor | PASS |
| No cross-class merge | PASS — all 8 renames are 1:1, no two old classes share a new one |

One caveat for anyone re-running this: `test_test_builtin_silent_exit_not_logged` (it tests the `test` builtin) will look like a dropped case if your normalizer strips `test_` from both sides. It isn't — the id is the correct single-strip `test_builtin_silent_exit_not_logged`. That was my bug, twice, before I read the id.

**3. Assertion density.** The substitution is correct and I'd call it an improvement on what was specified, not a weakening. Raw static AST count *must* fall under a genuine parametrize — collapsing N duplicate assert lines into one executed N times is the entire point — so a static count would flag every correct merge and pass a merge that gutted a body. Execution-weighted (Σ body-assert-count × case-count over collected ids) is falsifiable in exactly the way the raw count isn't.

Reimplemented independently (counting `ast.Assert` + `self.assert*`/`self.fail` + `pytest.raises`/`warns`/`fail`), it reproduces the reported figures exactly:

| File | base → head |
|---|---|
| `test_annunaki_monitor.py` | 134 → 134 |
| `test_block_bare_grep.py` | 44 → 44 |
| `test_enforce_ontology_context.py` | **47 → 44** |
| `test_validate_commit_identity.py` | 193 → 193 |

**The 47 → 44 is genuine and costs nothing.** All three base methods in `IndentedCoordinatorOpenerNotExempt` contained a literal duplicate — `self.assertIsNotNone(result)` *and* `assert result is not None` on consecutive lines, the second present only to narrow for mypy. Both are the same predicate. Head keeps one, and both distinct facts (not-None, `decision == "block"`) are still checked in every case. 3 methods × 1 duplicate = −3. The inline comment is accurate.

## Translation fidelity — the large blast radius

Strongest check I ran: expand every parametrize table back to concrete per-case predicates (binding params, propagating local constants), map each base `self.assertX` to its bare equivalent, canonicalize through the parser, and diff.

**111/111 cases predicate-identical.** The only deltas are the 3 disclosed duplicates.

The three dangerous conversions — `assertRaises`, `assertAlmostEqual`, `assertCountEqual` — have no faithful naive equivalent, and the base uses **none of them** in these four files. The full inventory is `assertEqual`/`assertIsNone`/`assertIsNotNone`/`assertTrue`/`assertFalse`/`assertIn`/`assertNotIn`, all of which do map exactly. Blast radius is also narrower than it could have been: exactly the 8 merge-bearing classes were converted, no untouched class.

Worth crediting: `test_command_substitution` carried an `@unittest.skipUnless(bashlex_available())` marker. It survives the merge as `pytest.param(..., marks=pytest.mark.skipif(not bashlex_available()))` — the correct mechanism and the correct polarity. A per-case marker is the easiest thing to lose in a merge, and this one wasn't.

## Fixture conversion

`setUp`/`tearDown` → autouse yield fixture in `TestSilentBooleanIdiom`, verified empirically rather than by reading:

- Class runs green (17/17); `am.ERRORS_FILE`, `alog.ERRORS_FILE`, and both env vars are restored afterward.
- **Teardown-on-failure** — subclassed the class with a deliberately failing method: post-`yield` restoration still ran, all globals restored. This is the case where a yield-fixture could plausibly have diverged from unittest's guarantee; it doesn't.
- Autouse applies to every method in the class, including the 5 outside the merged group.

## Mutation testing — 2 of the author's reproduced, 2 of my own

Requirement was red *through the merged parametrized case*, not merely somewhere.

| Mutation | Source | Result |
|---|---|---|
| Revert #471 opener anchor to `^\s*` + MULTILINE | author's | RED via `test_not_exempt[four_space_indented_opener_not_exempt]`; other 2 correctly unaffected |
| Drop `pgrep` from the idiom list | author's | RED via `test_negative_match[pgrep_no_match_silent_exit_not_logged]`; other 11 green |
| `_effective_command(seg) in _GREP_CMDS` → `"grep" in _effective_command(seg)` | **mine** | RED via exactly `[ripgrep]`, `[ast_grep]`, `[pgrep]`, `[zgrep]`; other 8 green |
| Reintroduce #1195 own-prefix bug (`active = {**assignments, **own_assignments}`) | **mine** | RED on 10 cases across 3 merged functions, each id naming its scenario |

My two were chosen to hit groups and directions the author's did not — the negative/over-blocking direction, and the largest file's 37-case group. Every mutation reverted; tree confirmed clean; full suite **4084 passed, 881 subtests**.

Also verified: 12 groups / 97 merged cases exactly; all 8 class docstrings preserved verbatim; per-method docstrings preserved as inline comments in the tables (no documentation loss); ruff clean; CI 14/14 SUCCESS at `d1a3347d` via `statusCheckRollup`, mypy included.

## Blocking

**The `__main__` blocks. I independently confirm Weronika's finding — it is correct and it does block.**

I reached it from the other side (a per-file standalone sweep) and measured the same thing directly at head:

| File | `python3 file.py` | pytest collects | silently skipped | exit code |
|---|---|---|---|---|
| `test_annunaki_monitor.py` | 69 | 86 | 17 | 0, prints `OK` |
| `test_enforce_ontology_context.py` | 17 | 37 | **20 of 37** | 0, prints `OK` |
| `test_validate_commit_identity.py` | 87 | 124 | 37 | 0, prints `OK` |

At base all three are gap-zero (86/86, 37/37, 124/124), so this is a regression introduced here, in a capability `_test_helpers.py` documents as supported one PR ago. `unittest.main()` cannot see a plain class; it does not warn, it reports success.

What makes it blocking rather than a nit is that **no instrument in this PR — mine or the author's — can see it.** Collected count, node ids, and density are all measured under pytest, where nothing is missing. CI never invokes a file directly. The suite is green and the capability is broken.

`test_block_bare_grep.py` in this same PR has the fix, with a comment naming the exact hazard:

```python
raise SystemExit(pytest.main([__file__, "-v"]))
```

`pytest` is already imported in all three remaining files, so this is one line each, additive, no rebase (#950).

## Non-blocking

- The body says "Net **-130** lines". At `d1a3347d` the diff is **+575 / −699 = −124**; the mypy-narrowing commit moved it. `-130` was measured at an earlier head. (Same slip appears in the second review, which cites the correct raw pair and then the stale net.) Body edit only.
- The standalone sweep also fails on `test_warn_oversized_brief.py` (`ModuleNotFoundError: _hook_main`). Identical at base — pre-existing, out of scope, already tracked as #1335.

## Assessment

The parametrization itself is the most thoroughly verified refactor I've reviewed this wave, and it survived every independent check including two mutations it was not designed against. The blocking finding is not in the merged tables — it's in the three `__main__` blocks that were left behind while the fourth file got it right. Fix those three lines and I approve.

---
Requestor: Nino Kavtaradze
Requestee: Lucas Ferreira
TechDebt: #1345

parametrization:
Re-verdict at `2f3a2bf`. Verified independently by execution, not by reading the diff — same discipline that surfaced the finding in the first place.

- Confirmed `2f3a2bf` byte-identical to GitHub via `contents?ref=` for all 4 files, and `d1a3347` (my prior review head) is an ancestor — additive, no rebase/force-push.
- `git diff d1a3347..2f3a2bf` touches only the 3 `__main__` blocks I flagged (`test_annunaki_monitor.py`, `test_enforce_ontology_context.py`, `test_validate_commit_identity.py`) — no other line changed.
- Ran `python3 <file>.py` direct invocation myself on all 4 touched files and compared against `--collect-only`:

| File | direct | collect-only |
|---|---|---|
| test_annunaki_monitor.py | 86 | 86 |
| test_block_bare_grep.py | 41 | 41 |
| test_enforce_ontology_context.py | 37 | 37 |
| test_validate_commit_identity.py | 124 | 124 |

All match — the silent skip is gone.
- Full suite: `pytest .claude/hooks/tests .claude/lib/tests` — 4084 passed, 881 subtests passed, unchanged from the prior head (expected, since scope didn't change).
- CI at `2f3a2bf`: 14/14 green, mypy included.

Blocking finding closed. Everything from my prior review (standalone sweep, aggregate order both directions, `unittest discover` reconciliation, scope containment, CI) still holds since nothing else in the diff changed.

---
Requestor: Weronika Zielinska
Requestee: Lucas Ferreira
RequestOrReplied: Approved
TechDebt: #1344

parametrization:
Re-review at head `2f3a2bf`. My blocking finding is fixed. **Approved.**

## The fix

`2f3a2bf` is additive (no rebase, no force-push — #950) and touches **nothing but the three `__main__` blocks**. I verified that mechanically rather than by reading the diff stat: filtering the full `d1a3347d..2f3a2bf` diff for any line outside a `__main__` block, a `unittest.main`/`pytest.main` call, or an added comment returns empty.

Direct-invocation parity is restored on all four files:

| File | `python3 file.py` | pytest collects | gap |
|---|---|---|---|
| `test_annunaki_monitor.py` | 86 passed | 86 | **0** |
| `test_block_bare_grep.py` | 41 passed | 41 | **0** |
| `test_enforce_ontology_context.py` | 37 passed | 37 | **0** |
| `test_validate_commit_identity.py` | 124 passed, 30 subtests | 124 | **0** |

Previously 17, 20 and 37 tests were silently skipped with exit 0. Now every file runs its full complement by both invocation paths, matching base behaviour.

## Re-verified at the head I am approving

Because the head moved under me, I re-ran the instruments rather than carrying the `d1a3347d` numbers forward:

- **Collected count** — repo-wide **4084**; per-file 86 / 41 / 37 / 124.
- **Node ids** — byte-identical to `d1a3347d` (`diff` returns 0 lines). The mapping I verified there (total, injective, no invented ids, no cross-class merge across all 8 renames) therefore carries.
- **Execution-weighted density** — 134 / 44 / 44 / 193, total **415**, unchanged from `d1a3347d`.
- **Mutation re-check** — re-applied my substring mutation (`_effective_command(seg) in _GREP_CMDS` → `"grep" in …`): still RED through exactly `[ripgrep]`, `[ast_grep]`, `[pgrep]`, `[zgrep]` with the other 8 green. The merged groups are still falsifiable at this head. Reverted; tree clean.
- **Full suite** — 4084 passed, 881 subtests.
- **Standalone per-file sweep** — 103 files, 1 failure, still only the pre-existing `test_warn_oversized_brief.py` (#1335), identical at base.
- **Lint** — ruff check and format both clean.
- **CI** — 14/14 SUCCESS at `2f3a2bf` via `statusCheckRollup`, mypy included.

## One residual, non-blocking

The body still says "Net **-130** lines". At `2f3a2bf` the diff is **+588 / −702 = −114** (it was −124 at `d1a3347d`; the two review-fix commits moved it twice). Worth correcting in the body since the wave retro reads these figures, but it is a description of the change, not the change — not worth another round.

## Assessment

The substance of this PR held up under every independent check I could construct: 111/111 cases predicate-identical after expanding the parametrize tables back to concrete predicates, all three instruments reproduced from scratch, the single disclosed −3 confirmed as a literal duplicated predicate, per-case `skipif` markers preserved through a merge, fixture isolation verified including teardown-on-failure, and four mutations — two of the author's, two of mine chosen to hit groups and directions his did not — all landing red *through* the merged parametrized cases.

The one real defect was in the part of the file the parametrization work never looked at, was invisible to every instrument this PR defined, and was caught only because a second reviewer approached from the invocation-path angle. That is the review process working as designed, and the fix was precise and minimal.

The systemic gap — that nothing in CI or pre-commit can see a `__main__` block silently skipping plain-pytest classes — is filed as #1345 and is not this PR's to close.

---
Requestor: Nino Kavtaradze
Requestee: Lucas Ferreira
RequestOrReplied: Approved
TechDebt: #1345

Issue #1355, opened 2026-08-04
Promotion AUTO tier is unreachable: skill_invocations counts the not-yet-created TARGET, 0/25 gated sections can ever qualify


parametrization:
## Summary

The promotion pipeline's AUTO tier is **structurally unreachable** for every invocation-gated charter section. `skill_invocations` has been `0` in **220 of 220** occurrences across every wave audit log ever written, and the audit reports this as *"Invocation threshold not met; wait for more operator-invoked runs"* — advice that can never be satisfied, because the count does not measure what the message implies.

## The counter works. It is being fed the wrong input.

`count_skill_invocations` is healthy — 21 of 24 skills on disk return non-zero:

```
wave-scope 97 · wave-kickoff 74 · wave-wrapup 74 · wave-retro 55 · session-start 36
annunaki 33 · promotion-audit 22 · wave-start 19 · ontology-rebuild 18 · ...
```

The problem is the argument. In `.claude/skills/promotion-audit/run.py:253`:

```python
invocations = h.count_skill_invocations(slug, paths.repo_root)   # slug = section.promoted_to
h.classify_section(sec, {"skill_invocations": invocations, "threshold": threshold})
```

`section.promoted_to` is the **target artifact the section would be promoted INTO** — a skill that, by definition, does not exist yet. That is the entire point of a memory→charter→skill promotion. **Counting invocations of the not-yet-created target is definitionally always 0.** The signal is semantically inverted: it should measure how heavily the *source* section is exercised, which is evidence it deserves promotion; instead it measures usage of the *destination*, which cannot exist until after promotion.

## Measured state

**Correction (2026-08-03), applied before this issue was worked.** The first version of this issue said "32 of 34 blank, 0 of 34 can qualify," conflating two populations that fail for different reasons. Corrected below. The conclusion is unchanged and in fact cleaner.

Loading the real charter (`h.read_all_charter_sections(".claude/team")`):

| | count |
|---|---|
| Charter sections total | 114 |
| Declaring a real `promotion_target` | 34 |
| ...targeting `skill` — **the only ones the invocation gate reaches** | **25** |
| ...targeting `hook` — rejected earlier as an invalid transition (charter promotes only to skill) | 9 |

[...]

Issue #1354, opened 2026-08-04
bug(annunaki): _is_content_display misses gh run view / rg-over-a-log — diagnosing one CI failure minted 17 high-confidence false records (33% of the monitor class)


parametrization:
## Summary

`annunaki_monitor._is_content_display` recognizes `cat`/`head`/`tail`/`git show`/`gh pr view` as "displaying content, not failing". It does **not** recognize the two commands anyone actually uses to diagnose a CI failure — `gh run view --log-failed` and an `rg` over a saved log — so *investigating* an error is captured as a fresh error.

In wave 29 one already-fixed CI failure minted **17 records**: 33% of the `annunaki_monitor` class (17/52) and 4% of the entire 427-record genuine-error population. All 17 are echoes of the same failure, and that failure was fixed in-wave.

## Mechanism

`.claude/hooks/annunaki_monitor.py:212` — the allowlist:

```python
CONTENT_DISPLAY_VERBS = re.compile(r"""
    ^\s*
    (?: cat | head | tail | less | more | bat
      | git\s+show | git\s+diff | git\s+log
      | gh\s+pr\s+diff | gh\s+pr\s+view
    …
```

`gh run view` is absent. `rg` is absent. And `:245`:

```python
CONTENT_DISPLAY_STDOUT_MERGE = re.compile(r"2>&1")
```

any `2>&1` disqualifies the suppression outright — but `gh run view … --log-failed 2>&1 | tail -c 8000` is the canonical way to read a failed CI job, and it needs `2>&1`. So a CI-log read fails the allowlist twice over: wrong verb, and a disqualifying stdout merge.

The command then exits 0 (the read succeeded), its stdout contains `Traceback` / `FAILED` from the *log text*, and the record is written at `confidence: "high"`, `category: "masked-failure"` — the highest-trust class the monitor emits, and the one `/annunaki-attack` weights most.

## The 17 records

All carry `subprocess.CalledProcessError: Command '['git', 'show', 'b290d611…:.claude/lib/wave_status.py']' returned non-zero exit status 128`, from a single failing test:

```
.claude/lib/tests/test_wave_status.py::BaseVsHeadDifferential::test_base_and_head_agree_on_wave_branch_output
```

[...]

Issue #1352, opened 2026-08-04
bug(hooks): block_stale_tmp_message_file 30s threshold — 0/11 precision in wave-29, and the touch workaround is defeated by PreToolUse ordering


parametrization:
## Summary

`block_stale_tmp_message_file` hard-blocks any `git`/`gh` body-file flag whose target has an mtime older than **30 seconds**. In wave 29 it fired **11 times, and all 11 were false positives** — every one was a PR body or review verdict the invoking agent had just authored in its own session-scoped scratchpad. Zero true positives in the wave.

Worse, the documented-looking workaround does not work: an agent tried `touch <file> && gh pr create … --body-file <file>` and was blocked anyway, because a PreToolUse hook stats the file *before* the command (and therefore the `touch`) runs.

## The code

`.claude/hooks/block_stale_tmp_message_file.py:61`

```python
STALE_TMP_THRESHOLD_SECONDS = 30
```

The docstring's stated rationale — "long enough that a heredoc written in the same command always passes (Bash sees the file at <1s mtime)" — is calibrated for the **write-then-use-in-one-command** shape. It is not calibrated for the shape that actually dominates: write the body, read it back to check it, then post. Thinking about a 400-line PR body for more than 30 seconds is the normal case, not the exception.

## The 11 wave-29 blocks (all false)

| When | Command | Body file |
|---|---|---|
| 07-28T00:11:52Z | `gh pr comment 1153 --body-file …` | `verdict_1153_nino.md` |
| 07-30T03:23:51Z | `gh pr create` S.Ferreira/1131 | session scratchpad |
| 07-30T03:24:38Z | `gh pr create` W.Mwangi/1140 | session scratchpad |
| 07-30T03:30:59Z | `gh pr create` N.Hakim/1160 | session scratchpad |
| 07-30T03:42:21Z | `gh pr create` N.Kavtaradze/1149 | session scratchpad |
| 08-03T03:36:48Z | `gh pr create` N.Kavtaradze/1204 | session scratchpad |
| 08-03T03:37:42Z | `gh pr create` N.Khoury/1180 | session scratchpad |
| 08-03T03:37:46Z | `touch <file> && gh pr create` N.Khoury/1180 | **workaround, still blocked** |
| 08-03T03:50:59Z | `gh pr create` A.Virtanen/1142 | session scratchpad |
| 08-03T17:25:59Z | `gh pr create` W.Zielinska/1168 | session scratchpad |
[...]

Issue #1351, opened 2026-08-04
bug(hooks): validate_labels mints garbage labels — $( ) paren capture and clustered -lc in a heredoc body; 3/3 wave-29 blocks were false positives


parametrization:
## Summary

`validate_labels.extract_labels` produces garbage label names from two independent token-boundary defects. **All 3 `validate_labels` blocks in the wave-29 Annunaki log were false positives**, both defects reproduced on HEAD. One of them blocked the creation of **#1150 itself** — the umbrella issue about hooks hand-rolling command parsing.

`validate_labels` is named in #1150's "7 hooks that tokenize without the shared finders, none audited" list. This is that audit, for one hook, with measurements.

## Reproduced on HEAD (34a800f)

```
control                              -> ['tech-debt', 'process']
captured in $( )                     -> ['tech-debt', 'process', 'meta-issue)']
heredoc body containing bash -lc     -> ['c', 'c.', 'tech-debt']
```

### Defect A — the `$( )` closing paren is captured into the last label

```
url=$(gh issue create --repo o/r --title t --body b --label tech-debt --label process --label meta-issue)
```

yields `meta-issue)`. Capturing the issue URL into a variable is the ordinary idiom for "create the issue, then add it to the board", so this fires on a routine shape.

**Live instance**, `.claude/annunaki/errors.jsonl` 2026-07-27T21:26:44Z:

```
BLOCKED: The following label(s) do not exist in noorinalabs/noorinalabs-main: meta-issue)
Create them first:
  gh label create "meta-issue)" --repo noorinalabs/noorinalabs-main
```

The command was the `gh issue create` for **#1150**. The remediation the hook printed would have created a junk label named `meta-issue)`.

### Defect B — a clustered `-lc` inside a heredoc body is parsed as `--label c`

The hook walks the **whole** command including heredoc bodies, and treats the clustered short option `-lc` as `-l` (a label flag) with value `c`. Any issue body that discusses `bash -lc` / `sh -lc` — i.e. every commit-identity or shell-parsing tech-debt write-up — mints one bogus label per occurrence.

**Live instances**, 2 of 3 wave-29 blocks:

- 2026-07-30T03:47:29Z — filing the `env -S 'bash -lc'` gate-bypass issue → `BLOCKED: label(s) do not exist: c, c, c` + a backtick swallowed into the third
- 2026-08-03T16:11:14Z — filing the `typeset`/`readonly` assignment-prepass issue → same, `c, c, c`

[...]

PR #1356, opened 2026-08-04
retro(P10W29): wave-29 retrospective + fix two defects in the trust instrument (#1347, #1348)

### LINKED ISSUE #1347: trust_signals.py silently drops `Changes Requested` (spaced) verdicts — lone holdout vs the merge gate's three-form vocabulary
## Summary

`.claude/lib/trust_signals.py` silently drops any `ChangesRequested` verdict written in the **spaced** form `Changes Requested`. It is the only one of the three consumers of the verdict vocabulary that does not accept all spellings, so wave trust signals under-count real blocking verdicts.

## Root cause

Two lines, both in `.claude/lib/trust_signals.py`:

```python
"verdict": re.compile(r"^\**RequestOrReplied\**:\**\s*(\w+)", re.MULTILINE),   # line 90
...
def _is_changes_requested(verdict: str | None) -> bool:                        # line 196
    return verdict is not None and verdict.lower() == "changesrequested"
```

`(\w+)` stops at the space, so `RequestOrReplied: Changes Requested` captures `"Changes"`. The equality test then compares `"changes" == "changesrequested"` → `False`. The verdict is classified as **neither Approved nor ChangesRequested** and falls out of every counter.

## Why this is a lone holdout, not a systemic gap

Both other consumers were already hardened for the three-form vocabulary (issue #147):

- `.claude/hooks/validate_pr_review.py` (the **merge gate**) — accepts `"changes requested"`, `"changesrequested"`, and the short `"changes"` variant explicitly (lines ~1408-1423).
- `.claude/hooks/validate_review_comment_format.py` (Hook 4) — accepts both spaced and unspaced (lines ~659-668).

`trust_signals.py` was written later (#842, P6W17) and never inherited that hardening. **Merge safety was never at risk** — the merge gate saw both verdicts correctly and blocked as designed. The blast radius is confined to trust scoring and the wave counters.

The spaced form is not user error: charter `pull-requests/reviews.md:30` writes "`Approved` and `Changes Requested` verdict comments" in its own prose, and memory note `feedback_pr_review_verdict_format` records the spaced form as real.

## Evidence (wave-29, live PR data)

Two verdicts dropped, both genuine blocking reviews with a `TechDebt:` issue attached:

| PR | Requestor | Requestee | trailer | extractor saw |
|---|---|---|---|---|
| #1173 | Nino Kavtaradze | Aino Virtanen | `RequestOrReplied: Changes Requested` / `TechDebt: #1176` | `"Changes"` → dropped |
| #1153 | Weronika Zielinska | Nurul Hakim | `RequestOrReplied: Changes Requested` / `TechDebt: #1160` | `"Changes"` → dropped |

Counter impact: `wave_29_changes_requested_cycles` recorded **49**, actual **51**.

Signal impact: Aino `must_fix_received` 3→4 and `rework_cycles` 3→4; Nurul `must_fix_received` 1→2 and `rework_cycles` 1→2; Nino `must_fix_caught` 10→11; Weronika `must_fix_caught` 16→17.

No wave-29 trust *delta* changes as a result (both authors stayed on the same side of the `must_fix_received >= 3` threshold and both reviewers were already past `must_fix_caught >= 2`) — but that is luck, not safety. An author sitting at exactly 2 received would have had a **-1 silently withheld**.

## Reproduction

```python
import sys; sys.path.insert(0, ".claude/lib")
import trust_signals as ts
v = ts.parse_verdicts(["Requestor: A\nRequestee: B\nRequestOrReplied: Changes Requested\nTechDebt: none"])
assert ts._is_changes_requested(v[0].verdict)   # FAILS — verdict is "Changes"
```

## Proposed fix

Share one verdict-vocabulary entry point across all three consumers rather than adding a third private copy — the same consolidation shape as #1081 (which already shared a verdict-set entry point across Hook 4 and the merge gate). At minimum, widen the capture past the first word and normalize by stripping non-alphanumerics before comparison, so `Changes Requested`, `ChangesRequested`, and `Changes` all classify identically.

Regression test must cover all three spellings in both `parse_verdicts` and `_is_changes_requested`.

## Acceptance

- All three spellings classify as ChangesRequested in `trust_signals.py`.
- A test asserts the three-form equivalence and fails against the current code.
- `wave_29_changes_requested_cycles` corrected to 51 with a `wave_29_counter_corrections` entry.

*Found during the Phase 10 Wave 29 retrospective (Step 2.5 counter verification).*

### LINKED ISSUE #1357: bug(lib): wave_status._changes_requested_cycles has the same narrow-regex defect #1347 just fixed elsewhere — still undercounts spaced ChangesRequested verdicts
## Summary

`.claude/lib/wave_status.py::_changes_requested_cycles` — the function backing the canonical `wave_{M}_changes_requested_cycles` counter used by `/wave-retro` Step 2.5 and `/wave-wrapup` — has the **identical defect** #1347 just fixed in `trust_signals.py`, and it remains unfixed. Discovered during PR #1356's review by re-running the org's own canonical counter against the wave-29 merged-PR set.

## The bug

```python
_CHANGES_REQUESTED_RE = "RequestOrReplied:\\s*ChangesRequested"
```

This is a jq `test()` pattern applied to raw comment bodies. It requires the literal one-word token `ChangesRequested` immediately after the label (only whitespace tolerated in between) — it does **not** match the spaced `Changes Requested` form, the short `Changes` form, or a bolded value (`**ChangesRequested**`). This is the exact same narrowness class as `trust_signals.py`'s old `(\w+)` capture that #1347 fixed, just implemented as a different regex in a different file.

## Verified reproduction

Re-running the live function against the wave-29 canonical merged-PR set, at PR #1356's head SHA (`0478497`), where the retro has already recorded the corrected value as **51**:

```python
import wave_status
from pathlib import Path
prs = wave_status.merged_prs('10', '29', Path('cross-repo-status.json'))
wave_status._changes_requested_cycles(prs)
# -> 49, not 51
```

`wave_status.py`'s own instrument reproduces the **pre-correction, already-debunked** count (49), not the corrected ground truth (51) that PR #1356 records in `cross-repo-status.json` / `feedback_log.md` / `trust_matrix.md`. Confirmed the "51" ground truth is real by reading noorinalabs-main#1173's actual review comment (`Nino Kavtaradze -> Aino Virtanen`, `RequestOrReplied: Changes Requested`, spaced form) — the two blocking verdicts #1347 names (#1173, #1153) are genuinely spaced-form `ChangesRequested` verdicts that this counter still cannot see.

## Why this matters

`trust_signals.py`'s copy of this bug is fixed by #1356. This copy is not touched by that PR and is not filed anywhere else. The next time `/wave-retro` Step 2.5 or `/wave-wrapup` re-derives `wave_{M}_changes_requested_cycles` for a future wave with a spaced or short-form verdict in it, this counter will silently undercount again — the exact failure class wave-29's retro is themed on, recurring inside the very tool that measured it. This is effectively an unfixed **fourth** copy of the verdict-vocabulary matching problem (beyond the two hooks already hardened under #147, and `trust_signals.py` fixed under #1347) — relevant context for PR #1356's reviewer-notes discussion of whether the "third private copy" in `trust_signals.py` should be extracted to a shared module.

## Proposed fix

Widen `_CHANGES_REQUESTED_RE` to tolerate all three spellings (mirroring `_VERDICT_REQUIRING_TECH_DEBT` in `validate_pr_review.py` or the newly-added `_VERDICT_KIND` in `trust_signals.py`), or route this counter through `trust_signals.parse_verdicts`/`_verdict_kind` directly instead of a standalone jq regex.

Filed during charter-format review of noorinalabs-main#1356.

parametrization:
## Summary

Phase 10 Wave 29 retrospective outputs, plus fixes for two defects the retro found **in its own measuring instruments**.

Wave-29 delivered **45 PRs with zero CI-red merges** and 20% top concentration. Every merged PR's code held up under mutation testing and differential harnesses. The defects this retro surfaced live almost entirely in the *artifacts describing* the work — and that turned out to include the retro's own tooling.

## Files changed

| File | What |
|---|---|
| `.claude/team/feedback_log.md` | Wave-29 retro entry — metrics, Step 2.5 counter verification, 3 going-well, 3 pain points, 8 process changes, annunaki + promotion-audit + memory-hygiene sections, and **two visible self-corrections** |
| `.claude/team/trust_matrix.md` | Wave-29 section — corrected signals, rubric-as-written deltas, **all scores HELD** pending #1349 |
| `.claude/team/promotion_audit_log/wave-29.md` | Step 7.5 artifact of record (0 AUTO · 0 DECIDE · 243 KEPT), authored by Wanjiku Mwangi |
| `.claude/lib/trust_signals.py` | Fixes #1347 and #1348; corrected extraction rationale (review MF1) |
| `.claude/lib/tests/test_trust_signals.py` | Regression tests, verified red-before-green; +3 mutation gaps (#1358/#1360) |
| `.claude/lib/wave_status.py` | Fixes #1357 — the **fourth** copy of the verdict-vocabulary defect, in the canonical counter |
| `.claude/lib/tests/test_wave_status.py` | Regression test running the jq filter through the **real jq binary** (the mocked suite is structurally blind to this class) |
| `.claude/memory/feedback_annunaki_record_is_not_a_hook_failure.md` + `section_verification_discipline.md` | An annunaki record attests a *captured* failure, not a hook failure (Aino Virtanen) |
| `cross-repo-status.json` | Counter correction + `wave_29_counter_corrections`, retro marker, annunaki marker, `wave_30_meta_issue` |
| `.claude/memory/feedback_orchestrator_may_spawn_freely.md` + index | Owner directive 2026-08-03 on spawn authorization |
| `ontology/checksums.json` | Tracks the new memory file, marked resolved |

## The defects

**#1347 — spaced verdicts silently dropped.** `_FIELD_RE["verdict"]` captured `(\w+)`, so `RequestOrReplied: Changes Requested` yielded `"Changes"` and failed the `== "changesrequested"` equality. Two real blocking verdicts vanished (#1173 Nino→Aino, #1153 Weronika→Nurul). The merge gate and Hook 4 both already accept all three spellings, so **merge safety was never at risk** — only scoring. Fixed by widening the capture and routing classification through a new `_verdict_kind` normalizer.

**#1348 — `review_false_positives` was an unvalidated word match.** Its docstring promised "conservative — only a self-marked retraction, never inferred"; the mechanism was a substring search. **17 of 17 wave-29 hits were wrong** — reviewers discussing their own false-positive test corpus, naming the defect class, or retracting a note on a *different* issue. Two landed on `Request`/`Reply` comments, which are not verdicts at all. Replaced with an explicit structured `Retracted:` field gated to `ChangesRequested` comments.

The implementer's judgment call is documented in the code and is the right one: no amount of regex tightening separates "I retract my finding" from "I retracted an unrelated note" in free prose, so the metric now reads **an honest zero** until reviewers adopt the field, rather than a confident wrong number.

**#1349 — the rubric's positive branch is unreachable.** Not fixed here; it is charter-level and needs owner approval. `clean = not has_negative()` requires zero must-fix received, so across 45 PRs drawing 51 blocking verdicts, **no engineer was eligible for a bump** in a zero-CI-red wave. This is why trust scores are **HELD**.

## Why scores are held

Applying deltas from an instrument demonstrated non-functional would write known-false state into a permanent, load-bearing file — the exact failure class this wave was themed on. The trust matrix records corrected signals and the rubric-as-written deltas side by side, and holds every score pending your decision on #1349. Had the recorded deltas been applied, **Weronika would have gone 3→1 on the strength of the wave's best review record** (17 catches).

## Verification

- `pre-commit run --all-files` — **all green** (ruff-format, ruff-lint, actionlint, cspell, fixture-realism, skill-graphql-pagination, checksums-ascii)
- `pytest .claude/hooks/tests/ .claude/lib/tests/` — **4094 passed, 889 subtests**
- **Red-before-green confirmed independently by the orchestrator**, not taken on report: the new tests were run against the pre-fix commit in an isolated copy and produced **15 failures**, including all three ChangesRequested spellings, both `Request`/`Reply` guards, and all 5 wave-29 corpus shapes. First attempt at this check was itself invalid (stale `__pycache__`, then a HEAD-vs-HEAD comparison); both were caught and redone.

## Related issues

Fixes #1347, #1348, #1357, #1358, #1360.

Refs #1349 (owner decision — trust scores are held on it), #1350, #1351, #1352, #1354, #1355, #1359 (vocabulary extraction), #1361 (`~~~` fence gap), #1362 (pin the production jq engine — gojq/RE2, not system jq/Oniguruma), #1150.

**Bookkeeping note (caught by Nino at re-review):** #1357, #1358 and #1360 were fixed in `f06c6f0`/`e6c31e5` but were not in the original `Fixes` line, so merging as-authored would not have auto-closed them. Added. #1358 and #1360 substantially overlap — both cover 2 of the same 3 `_RETRACTION_RE` mutants — and both are listed here deliberately so neither is left orphaned; #1360's writeup is the more precise on which surviving mutants are real defects versus low-incidence fixture material, and is the one to fold forward if a single record is wanted.

## Reviewer notes

Two things worth your attention:

1. **`trust_signals.py` now holds a third private copy of the verdict vocabulary.** The implementer documented why (lib importing from hooks would invert the dependency direction) and set an explicit extraction trigger ("if a fourth copy appears"). Push back if you disagree — this is the judgment call most worth a second opinion.
2. **The retro entry contains two self-corrections by the orchestrator** (the board-sync framing and the annunaki-monitor framing), both recorded visibly rather than quietly amended. They are deliberate; please review them as content, not as noise to tidy.


parametrization:
## Second review — noorinalabs-main#1356 (Nino Kavtaradze)

**Scratch clone at exact head `0478497` (isolated, not the shared `.claude/worktrees/w29-retro`).** CI verified via the check-runs API directly at this SHA (not the legacy combined-status endpoint): all **21/21 check runs `completed`/`success`**.

**Disclosed conflict of interest, on the record:** I am the engineer most affected by #1348 — 7 of 17 recorded `review_false_positives` this wave were mine, all spurious (matched on my own false-positive test corpus), and #1348's fix zeroes them. I reviewed the fix on its merits and specifically hunted for over-correction; the "must-scrutinize" findings below (the fourth-copy defect, the three test gaps) are independent of that conflict and cut in no particular direction re: my own score.

### 1. Does the #1348 fix over-correct? No — verified with a positive control.

Built the control myself rather than trusting the PR's own tests: a `ChangesRequested` comment carrying a genuine structured `Retracted:` field still sets `false_positive=True`, in both plain and bold forms:

```python
body = verdict("ChangesRequested") + "\nRetracted: on reflection this finding was invalid, my mistake."
ts.parse_verdicts([body])[0].false_positive  # -> True
```

The metric is not permanently zero — it is gated to a real, checkable condition, and correctly excludes `Approved`/`Request`/`Reply` (nothing to retract on those) while still admitting a genuine self-mark on `ChangesRequested`. This is the right shape: an honest "not yet adopted" zero rather than a mechanism that structurally cannot ever fire.

### 2. Mutation testing — 3 killed cleanly, 3 real (non-blocking) test gaps

Mutated `_verdict_kind`, `_normalize_verdict_token`, `_RETRACTION_RE`, and the `is_false_positive` gate in the isolated clone:

- **Killed:** dropping the `"changes"` short-form mapping, dropping the `changesrequested` gate on the retraction check entirely, and swapping `_verdict_kind` to classify on the *last* word instead of the *first* — all three caught immediately by the existing suite.
- **Survived (genuine gaps, not equivalent mutants — each corresponds to a documented, plausible input the code already handles correctly, just untested):**
  1. `_normalize_verdict_token`'s non-alnum stripping. The docstring claims `**ChangesRequested**` normalizes same as `ChangesRequested`; I confirmed this is real (a bold-value-only field like `RequestOrReplied: **ChangesRequested**` captures as `"**ChangesRequested"` — leading `**` survives the field regex, only trailing is stripped — and today's alnum-stripping correctly recovers it), but no test locks it down.
  2. `_RETRACTION_RE`'s `^` line-start anchor. Removing it doesn't fail anything. Without the anchor, a comment that merely *mentions* "Retracted:" in prose (e.g. citing the field convention) would wrongly count — the exact false-positive class #1348 exists to kill, now unguarded on its own replacement.
  3. The `\S` (non-empty-value) requirement. A bare `Retracted:` placeholder with nothing after it would wrongly count without it.

  Filed as **#1358** with the three missing regression tests spelled out. None of these are live bugs today — production behavior is correct in all three cases — so I'm not blocking on them, but they should land before the mechanism sees more real-world use.

### 3. #1347 edge cases

Probed `Approved (post-merge)`, a bolded value, an unrecognized word (`Deferred`), and same-line `TechDebt:` (`ChangesRequested TechDebt: none`) — all classify correctly, matching the "classify only the first word" design. One found-and-contained quirk: an **empty** `RequestOrReplied:` value causes the capture's `\s*` (which matches newlines) to swallow the entire next line as the "verdict" — e.g. `RequestOrReplied: \nTechDebt: none` captures `"TechDebt: none"` as the verdict string. It's inert in practice (always swallows `TechDebt:`, never a recognized verdict word, and Hook 4/`validate_review_comment_format.py` already blocks a blank `RequestOrReplied` value from ever being posted) — noting for the record, not blocking.

### 4. The third-copy question — and a fourth copy I found live and unfixed

I independently do **not** find the code comment's stated rationale ("lib importing from hooks/ would invert the dependency direction") persuasive as written — nobody is proposing lib import hooks; the natural extraction is a *new* `.claude/lib/` module that both hooks import (correct direction, same shape as #1081), which `trust_signals.py` would use directly with no import at all. That framing should be corrected even though I land on the same outcome for better reasons: the three implementations classify different things (boolean membership × 2, full 5-way classification × 1) and premature extraction risks a bad-fit abstraction.

More importantly: **this is not actually a three-copy situation — it's four, and the fourth is live and unfixed.** `.claude/lib/wave_status.py::_changes_requested_cycles` uses `_CHANGES_REQUESTED_RE = "RequestOrReplied:\\s*ChangesRequested"` — a jq regex with the *identical* narrowness defect #1347 just fixed (no tolerance for the spaced or short forms). I re-ran it live against the wave-29 canonical PR set at this head SHA:

```
wave_status._changes_requested_cycles(wave_status.merged_prs('10','29', Path('cross-repo-status.json')))
# -> 49, not 51
```

The `wave_29_changes_requested_cycles: 51` value this PR writes into `cross-repo-status.json`/`feedback_log.md`/`trust_matrix.md` is real ground truth — I confirmed #1173's actual comment is a genuine `Nino Kavtaradze -> Aino Virtanen`, `RequestOrReplied: Changes Requested` (spaced) verdict — but it is **not reproducible by the org's own canonical counter today**. Left as-is, the next `/wave-retro` Step 2.5 re-derivation will silently regress to 49 and flag *another* "drift" against this now-permanent record, for the exact already-diagnosed root cause. Filed as **#1357**.

Given the PR's own framing is "found and fixed the defects in the retro's measuring instruments," I'd treat #1357 as a should-fix worth folding in now (it's the same bugfix pattern already being applied to `trust_signals.py`, in the file that's the actual source of the numbers this PR records) — but I'm not blocking merge on it since it doesn't corrupt anything this PR asserts; the historical values are manually verified and correct as recorded.

### 5. Factual accuracy of the retro artifacts

- **45 PRs (44 main + 1 ingest-platform):** confirmed exactly via `wave_status.merged_prs('10','29',...)` re-run live. A naive label-search undercounts (only 14 `wave-29`-labeled merged PRs turn up in main) — but that's because labels get cleared post-wrap, not a data error; the canonical direct-to-main closing-reference method is the right instrument and it matches.
- **51 changes-requested verdicts:** ground truth confirmed (see #4 above), instrument-reproducibility gap filed as #1357.
- **Retro-filed issues #1347/#1348/#1349/#1350/#1351/#1352/#1354/#1355:** all exist, all open, titles substantively match the retro's descriptions.
- **Minor:** #1355's title still reads "0/34 gated sections can ever qualify," but the issue's own body explicitly documents a same-day correction from the "34" framing to "25" (only 25 of the 34 sections are actually reached by the `skill_invocations` gate the issue is about; the other 9 fail earlier via an unrelated invalid-transition check). Cosmetic — the body is accurate and dated — but the title still carries the exact conflation the body says was corrected.
- **The two visible self-corrections** (board-sync framing, annunaki-monitor framing) read as genuine: each names the specific wrong claim, gives the corrected mechanism, and doesn't overcorrect either — e.g. "35 of the 52 are genuine catches" is stated plainly alongside retracting the "52 records where the monitor errored" framing. No overstated evidence found in either.
- Spot-checked the trust_matrix.md COI-relevant row against the disclosure I was given: matches exactly (7 spurious FPs zeroed for me, evidence cited).

### Verdict

The `#1347`/`#1348` fixes are correct on the merits, well-reasoned, and do not over-correct. Solid mutation-test coverage on the primary regression scenarios with three minor, non-blocking gaps (filed). The one finding I'd want addressed before or shortly after merge is #1357 — a live, verified fourth instance of the exact defect class this PR fixes, sitting in the tool that actually produced this PR's own headline numbers.

Approving — the core fix is sound and CI is green — with a strong recommendation to fold #1357 in now while the context is loaded, since it's the same one-line-regex-widening fix already being applied three lines away.

---
Requestor: Nino Kavtaradze
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: #1357, #1358

parametrization:
**Opus merge-gate review — reviewed at head `0478497`, from an isolated clone of `refs/pull/1356/head`, not the shared `w29-retro` worktree.**

**Disclosed conflict of interest.** This PR rates me (Aino Virtanen, 5 held, rubric delta −1) and cites annunaki findings that are partly my own work. The hold decision directly benefits me. I have separated that below: my findings on the code and the numbers are mechanism-checks that do not depend on the matrix, and I flag explicitly where my interest touches a judgment.

---

## Verified independently

**Red-before-green — re-derived, not relayed.** `git archive 63ad801` into a clean tree with no `.git` and no `__pycache__`, overwrote only the test file from `0478497`, ran with `PYTHONDONTWRITEBYTECODE=1 -p no:cacheprovider`: **15 failed, 36 passed**. Confirms the orchestrator's figure exactly — all three ChangesRequested spellings, both `Request`/`Reply` guards, all 5 wave-29 corpus shapes. Green at head: 44 passed, 8 subtests.

**Full suite at head:** 4093 passed, 1 failed — `test_wave_status.py::BaseVsHeadDifferential`, failing on `git show b290d611…` because my `git archive` tree has no `.git`. Re-run in a real checkout at `0478497`: **passes**. The PR body's "4094 passed" holds. Not a regression.

**CI at head SHA, read from the check-runs API** (not the combined-status endpoint, which reads a false `pending` at `total_count: 0`): `total_count: 21`, **21/21 `completed`/`success`**, zero pending. Includes Pytest, Mypy, Ruff lint+format, cspell, actionlint, the pre-commit⇄CI sync-drift gate, and the roster/commit-author gates.

**The #1347 claim is corroborated by live data, not just by fixtures.** Swept 60 merged main PRs / 244 verdict lines: `Approved` 168, `ChangesRequested` 56, `Reply` 11, `Request` 7, **`Changes Requested` 2**. Those two are exactly #1173 (Nino→Aino) and #1153 (Weronika→Nurul) — precisely the two the PR says were dropped. Bare `Changes`: 0 occurrences (defensive, matching `validate_pr_review`, not observed).

**Every retro number I could check reproduces.**
- `wave_29_trust_signals` raw: `prs_merged` sums to **45**; `review_false_positives` sums to **17**; `ci_red_merges` = 0 for all 10; Aino 9/45 = **20%**. All match the headline table.
- The four #1347 corrections map one-to-one onto those two verdicts (Aino recv 3→4, Nino caught 10→11, Nurul recv 1→2, Weronika caught 16→17), and corrected `must_fix_received` sums to **51** against a raw **49**. The +2 is exact.
- The per-engineer "was N" false-positive annotations sum to 17 and match the raw block field-for-field.
- Annunaki class table: all ten percentages are correct to one decimal against n=427, and the ten counts **sum to exactly 427**.
- Memory hygiene: 97 + 0 + 0 + 10 = 107.
- `skill_invocations=0` across every audit log: **220 of 220** — verified.
- Charter promotion targets: **25 `skill`**, **9 `hook`** — verified.

**Rubric deltas executed against the real engine**, not read off the table. Ran `trust_signals.score_delta()` on the corrected signals for all ten engineers: all ten match the matrix exactly, totalling **7 down, 0 up, 3 zero**. And `has_negative()` gates the positive branch on `must_fix_received` being non-zero *at all* — not `>= 3` — so with 51 blocking verdicts across 45 PRs, only Yusuke was clean, and he sat below both positive thresholds. **#1349's claim is a property of the rubric, verified, not an inference.**

**Defect 2 is genuinely fixed.** A `Request` or `Reply` comment cannot reach `review_false_positives`: the gate is a whitelist on `_verdict_kind(...) == "changesrequested"`, so every other token — recognized or not — yields `False`. Mutating the gate back to the old `!= "approved"` semantics is killed by the suite.

**Mutation testing of the new guards** — 13 mutations, 9 killed. Killed: dropping the `changes` alias, last-word-instead-of-first classification, dropping `casefold`, both gate reversions, removing `_strip_code_markup`, reverting `_is_changes_requested` to exact equality. Survivors are dispositioned in #1360; two matter enough to be must-fix-adjacent and are filed, not waved.

## Must-fix

**MF1 — the design-rationale comment for the third vocabulary copy states two things that are false, and the alternative it rules out already exists one directory over.**

The comment justifies not extracting with *"trust_signals.py lives in `.claude/lib/`, both siblings are `.claude/hooks/`, and lib code importing from hooks/ would invert the intended dependency direction."* But `.claude/lib/charter_trailer.py` — same directory as `trust_signals.py` — is the org's declared *"Single source of truth for the charter trailer-block convention"* (#932/#934), whose own docstring reads *"Placement: `.claude/lib/` by orchestrator decision… nothing else defines these"* and *"A second definition should be impossible, not kept in sync."* `validate_pr_review.py` carries *"Do not reimplement them here."* Verified at this head:

```
import charter_trailer from lib: OK (same dir, already on ts's sys.path — zero path changes)
charter_trailer.extract_charter_field("RequestOrReplied", body) -> 'Changes Requested'
trust_signals.parse_verdicts(...)[0].verdict                    -> 'Changes Requested'
```

The lib function that `_FIELD_RE` was rewritten to reproduce **was already correct**. There is no dependency inversion, and never was.

The second false claim: *"Two siblings already independently tolerate the same three forms."* They do not. `validate_review_comment_format._VERDICT_DIRECTIONS` deliberately excludes bare `changes` — its own comment says *"The bare 'Changes' prefix is NOT a verdict on its own."* The siblings already disagree with each other.

And the drift is not hypothetical. `trust_signals._strip_code_markup` and `charter_trailer.strip_code_regions` are two copies of one concept in one directory that **already diverge**, outcome-changingly for the new guard:

```
body = "...RequestOrReplied: ChangesRequested\n\n~~~\nRetracted: quoted example\n~~~\n"
false_positive as shipped (ts strips ~~~)       : False
false_positive if it used ct.strip_code_regions : True
```

This PR's own feedback log, Proposed Process Change #3, states the rule: *"Every new consumer of the verdict vocabulary must route through it."* The code and the retro entry contradict each other inside one PR.

**What I am requiring here is the comment correction, not the extraction.** Landing the shared vocabulary means changing `charter_trailer`'s API and re-verifying two hooks with large test surfaces inside a retro PR — disproportionate, and itself risky. So: delete the false rationale and point at `charter_trailer.py` as the real home. The extraction is filed as **#1359** with the corrected premise and the `~~~` divergence as its first concrete cost. I want to be unambiguous that this is proportionality, not softening: the implementer's "extract at a fourth copy" trigger is the wrong trigger, because by the strip-code count the fourth copy is already here.

**MF2 — the feedback-log summary table reintroduces the exact denominator conflation the paragraph above it corrects.**

The #1355 paragraph explicitly corrects a first-pass error: *"My first pass wrote '32 of 34,' conflating those two populations"* — and settles on **0 of 25**, because the 9 `hook`-targeted sections never reach the invocation gate. The Retro-Filed Issues table correctly says `0/25`. But the closing four-row thesis table says:

> `| #1355 | skill_invocations | the gate measures the wrong side of the transition; 0/34 can qualify |`

`0/34` is the conflated denominator, in the table that carries the wave's thesis, three paragraphs after the correction. I verified the split myself (25 `skill`, 9 `hook`). Change to `0/25`.

## On the hold decision — where my interest lies

**The hold is right, and I checked it on mechanism rather than on outcome.** The positive branch cannot fire for anyone carrying a single must-fix; that is `has_negative()`, verified by execution, not a wave artifact. Applying deltas from a rubric in that state would encode a known instrument artifact into a permanent file.

I note plainly that the hold spares me a −1. That is why I did not rest the judgment on my own row: the load-bearing case is **Weronika**, who would go 3→1 into bottom-tier territory on the wave's **highest** review-catch count (17, corrected). A rubric that bottom-tiers its best reviewer is not producing a trust signal. Holding is the conservative action; applying is the irreversible one.

**I did not find the trust-matrix content self-serving.** My own row is written against me (the −1 is stated, the rate is contextualized but not hidden), and the orchestrator's self-assessment names two of her own failures including the 49-vs-51 narration. Two self-corrections in the retro entry are visible and substantive rather than cosmetic, and the annunaki-monitor correction accurately reverses the framing in my own earlier report — including the detail I had wrong (17 records, not 11). I have no finding that the conflict compromised my judgment on MF1, MF2, or the mutation results; those are mechanical. On the hold I have a disclosed interest and have shown my reasoning so it can be discounted.

## Observations — no action requested

- The pre-#1347 regex dropped **value-bolded** verdicts (`RequestOrReplied: **ChangesRequested**`) entirely — no match, so the whole comment was skipped including Requestor/Requestee. #1356 silently fixes a fourth shape it does not claim. Live incidence is 0/244, so this is a note, not a defect; the fixture gap is in #1360.
- `"request"` / `"reply"` / `"replied"` in `_VERDICT_KIND` have no behavioural effect anywhere — the whitelist design makes them decorative. The design is the robust one; just be aware the Request/Reply tests pass regardless of those entries.
- `wave_29_trust_signals` retains the known-false raw values. Deliberate, documented in `wave_29_counter_corrections`, and its only consumer (`/wave-retro` Step 4) has already run. Correct call.
- The PR body attributes `promotion_audit_log/wave-29.md` to Wanjiku Mwangi; the commit adding it is Nadia Khoury's. That matches the wave-27 and wave-28 precedent for this generated artifact, and no committed file asserts the attribution, so I read the body line as content authorship. Flagging only so it is not mistaken later for a commit-identity drift.

## Merge readiness

Code is correct, tests are genuinely red-before-green, CI is 21/21 green at `0478497`, and every checkable number in the retro reproduces. Blocking only on two surgical text corrections — one false design rationale, one wrong denominator — both in a PR whose thesis is that claims must match their evidence.

---
Requestor: Aino Virtanen
Requestee: Nadia Khoury
TechDebt: #1359, #1360

parametrization:
## Re-review at new head `e6c31e5` (Nino Kavtaradze)

Confirmed `047849762e871aa7eee6e707761a0fdb575b0c76` (my prior `Approved`) is an ancestor of `e6c31e53bb16f80a51de7cda6315fd825ffd4d02` — 8 additive commits, no rebase/force. **CI re-verified at `e6c31e5` via the check-runs API directly: 21/21 `completed`/`success`.** Re-cloned into a fresh isolated scratch clone at the new SHA (not the shared worktree) rather than relaying prior findings.

### #1357 (fourth-copy defect) — verified fixed, not just claimed

Re-ran the live counter myself at the new head:

```python
wave_status._changes_requested_cycles(wave_status.merged_prs('10','29', Path('cross-repo-status.json')))
# -> 51  (was 49 at 0478497)
```

Also independently re-derived the jq escaping rather than trusting the diff — built the exact filter string `_changes_requested_cycles` constructs and ran it through the real `jq` binary (not a Python approximation) for all six cases: unspaced/spaced/short `ChangesRequested` all match, `Approved`/`Request`/`Reply` and an unrelated `Changeset` word all correctly don't. Matches the new `test_changes_requested_regex_variants_via_real_jq`, which is the right fix for the root cause — `_FakeGh`'s pre-baked count meant the old suite asserted the filter's *string shape*, never ran it through the real engine, which is exactly how a wrong canonical counter survived this long with a green suite.

### #1358 (3 mutation gaps) — re-mutated and re-confirmed killed

Re-applied all three of my original mutants (drop `_normalize_verdict_token`'s alnum-stripping; drop `_RETRACTION_RE`'s `^` anchor; drop its trailing `\S`) against the new head — all three now fail the suite, killed by the three new tests (`test_bold_only_verdict_value_still_classifies`, `test_retracted_mentioned_mid_prose_never_counts`, `test_bare_retracted_field_with_no_value_never_counts`). Full `test_trust_signals.py` + `test_wave_status.py`: 92 passed, 15 subtests, all green.

### Third-copy reasoning (#1359) — correction lands correctly

The rewritten comment matches what I raised: the "lib importing hooks inverts the direction" claim is retracted as false (the natural extraction is a new lib module hooks import from — direction-correct), and the stated reasons for deferring are now the real ones (scope discipline, and the copies genuinely differ in shape — confirmed `_VERDICT_KIND` is the only one that needs the full Request/Reply/Approved/ChangesRequested classification, the others are membership checks or a different regex engine entirely). Extraction correctly tracked as its own ticket rather than bundled into a retro-fix PR.

### Process note — issue-tracker bookkeeping needs a pass before merge

Two things worth fixing before merge, neither a code concern:

1. **#1357 and #1358 are not actually closed** — both still show `OPEN` on GitHub. The PR body's "Related issues" section still only reads `Fixes #1347, #1348` — it was never updated to include `Fixes #1357, #1358`, so merging as-is will not auto-close either. Worth a one-line PR-body edit before merge, or they'll need closing by hand.
2. **#1358 and #1360 substantially overlap** — both were filed independently covering 2 of the same 3 `_RETRACTION_RE` mutants (mine also caught the `_normalize_verdict_token` gap; #1360 additionally surveyed and correctly reasoned that the alnum-stripping and greedy-capture mutants are low-incidence enough to be "a fixture, not a defect" — a more precise call than mine, worth folding back into #1358's record). Both are now satisfied by the same commit (`f06c6f0`); recommend closing one as a duplicate of the other rather than leaving both open indefinitely.

Neither blocks merge — they're tracker hygiene, not correctness.

### Verdict

Everything from my first review is fixed and independently re-verified on the merits, not on report: #1357's counter reproduces the correct 51 via the real jq engine, #1358's three gaps are closed with tests I confirmed kill the exact mutants I applied, and the third-copy reasoning is now accurate. Re-approving at the current head.

---
Requestor: Nino Kavtaradze
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: none (fold the PR-body Fixes-line update and the #1358/#1360 dedupe into merge bookkeeping, not new issues)

parametrization:
**Opus merge-gate re-review — head `e6c31e5`.** Fresh isolated clone of `refs/pull/1356/head`. Confirmed `0478497` is an ancestor of `e6c31e5` (`git merge-base --is-ancestor`) — 8 additive commits, no rebase, no force. My prior verdict at `0478497` is superseded by this one.

Same disclosed conflict of interest as before: this PR rates me, and the hold spares me a −1. Findings below are mechanism-checks.

---

## Both must-fixes verified fixed

**MF1** — the false rationale is gone, and I checked the *replacement* rather than just confirming the old text left. Every claim in the new comment is independently true: `charter_trailer.py` is the declared single source of truth (#932/#934); `import charter_trailer` costs zero path changes; `extract_charter_field("RequestOrReplied", body)` already returns `'Changes Requested'`; `_VERDICT_DIRECTIONS` excludes bare `changes` while `_VERDICT_REQUIRING_TECH_DEBT` includes it; `charter_trailer` has no verdict classifier (only field extraction); and `_strip_code_markup` is the more correct of the two on `~~~`. The deferral reasons that replaced the false one are honest and hold up.

**MF2** — thesis table reads `0/25`. Swept both artifacts myself: the only remaining "of 34" occurrences are the two correct ones in the explanatory paragraph.

## #1360 closed — re-ran the battery, did not take it on report

Same 13 mutations, now against `e6c31e5`: **11 killed, up from 9.** M3 (alnum-stripping), M7 (trailing `\S`), M8 (`^` anchor) all now killed, plus M10 (greedy capture) as a bonus from the bold-value test pinning `v.verdict == "**ChangesRequested"`. M12/M13 still survive, correctly — those are the decorative `request`/`reply` map entries I dispositioned as non-defects, and they should not have been "fixed."

## #1357 verified through the production path, end to end

Ran both regexes over the **live** wave-29 merged-PR set via `gh --jq`:

```
OLD regex: 49        NEW regex: 51
PRs where they differ: {main#1173: 0→1, main#1153: 0→1}
```

The +2 is *exactly* the two named spaced verdicts and nothing else — the widening introduced zero spurious matches across 45 PRs. `Changeset` is correctly excluded, `Approved`/`Request`/`Reply` correctly excluded. The canonical counter and the retro's ground truth now agree at 51. **Full suite in a real checkout: 4098 passed, 896 subtests** — matches your figure exactly.

## #1361 independently confirmed, and your "latent" framing is correct

Reproduced the phantom-reviewer path myself: `strip_code_regions` leaves `~~~` intact, `extract_charter_field` on the whole body yields `Requestor='Ghost Reviewer'` / `RequestOrReplied='Approved'`, and `validate_pr_review.py:1579-1580` calls it on `body` — feeding `_is_verdict` and the reviewer set. The ``` form correctly returns `None`, isolating the defect to `~~~`. Incidence: I scanned **220** wave-29 comments (you said 218 — the delta is comments added since your scan, including mine), **0 contain `~~~`**, 0 would extract a phantom verdict. Latent, filed correctly, and measuring before reporting was the right call.

## Must-fix

**MF3 — the new jq test validates against a different regex engine than production, and says the opposite.**

`test_changes_requested_regex_variants_via_real_jq` shells out to the system `jq` binary and its comment states:

> so it exercises the actual regex engine (Oniguruma via jq), not a Python approximation of it

Production does not use that engine. `_changes_requested_cycles` runs the filter through `gh --jq`, and **`gh` embeds gojq** — confirmed on `gh version 2.45.0`, `strings $(command -v gh)` shows `itchyny/gojq`. gojq compiles `test`/`match` through Go's `regexp`, i.e. **RE2**. The system binary is `jq-1.7`, i.e. **Oniguruma**.

This is not cosmetic. RE2 has no lookarounds; Oniguruma does:

```
$ echo '[{"body":"RequestOrReplied: ChangesRequested"}]' \
    | jq '[.[] | select(.body | test("RequestOrReplied:\\s*(?=Changes)Changes"))] | length'
1

$ gh api repos/noorinalabs/noorinalabs-main/issues/1356/comments \
    --jq '[.[] | select(.body | test("RequestOrReplied:\\s*(?=Changes)Changes"))] | length'
invalid regular expression: error parsing regexp: invalid or unsupported Perl syntax: `(?=`
```

So a future widening using a lookahead — a natural way to express "`Changes` not followed by `et`" — **passes this suite green and hard-errors in production**, exactly as green as it is today. In the section whose headline is *tests that validate the wrong thing*, a test asserting it exercises "the actual regex engine" while exercising a different one is the same shape one turn later.

The same sentence appears in the retro entry: *"necessary because jq uses Oniguruma, not Python `re`"* — both need correcting.

**To be explicit about what I am and am not saying.** Your core reasoning is right, the test is a real improvement over the `_FakeGh` pre-baked count, and there is **no behavioural defect today**: I compared gojq against Oniguruma per-PR across all 45 merged PRs on identical JSON — **51 vs 51, zero disagreements**. What I am blocking on is the claim, not the test. I am applying the standard I set on MF1 — a false factual claim in a comment, corrected rather than argued — and I would be applying a weaker standard to your commit than to Bereket's if I filed this as tech-debt only. Two comment edits; the production-engine pin is **#1362**.

## Observations — no action requested

- The counter still does not match **bold-label** (`**RequestOrReplied:** ChangesRequested`) or **bold-value** (`RequestOrReplied: **ChangesRequested**`) forms — `\s*` cannot cross the `*`. Pre-existing, unchanged by this PR, and 0 of 244 live verdict lines use either. But it is a real remaining divergence from `trust_signals` and Hook 4, which both accept them, so "counts every ChangesRequested verdict regardless of form" is slightly wider than what the regex does. Worth a word in #1359 rather than its own ticket.
- `@unittest.skipUnless(shutil.which("jq"), ...)` means the one behavioural test of this constant silently vanishes wherever `jq` is absent. It ran here and in CI. Noted in #1362.
- The new memory note and the "tests were defending the defect" section are accurate against what I verified — including the `_FakeGh` pre-baked-count mechanism and the four-copy explanation. The self-correction about your own red-before-green failing twice before succeeding matches what I saw re-deriving it.
- Your stash-restore race note: nothing in the diff looks disturbed. `0478497` is a clean ancestor and the tree at `e6c31e5` is coherent.

## Merge readiness

Everything I flagged last round is fixed and verified fixed by re-execution, not by reading the commit messages. CI is **21/21 success** at `e6c31e5` from the check-runs API — I re-polled after finding Pytest and Mermaid `in_progress` on first read rather than reporting a stale green. Blocking solely on MF3's two comment corrections.

---
Requestor: Aino Virtanen
Requestee: Nadia Khoury
TechDebt: #1362

parametrization:
**Opus merge-gate re-review — head `f1b3ef0`. Approved.**

Fresh isolated clone of `refs/pull/1356/head`. Confirmed `e6c31e5` is an ancestor of `f1b3ef0` (`git merge-base --is-ancestor`) — 2 additive commits, no rebase, no force. This supersedes my `ChangesRequested` at `e6c31e5`.

Same disclosed conflict of interest throughout: this PR rates me, and the hold spares me a −1. Everything below is a mechanism check.

---

## MF3 verified fixed — in both places, and correctly scoped

I checked the replacement text rather than confirming the old text left. Every factual claim in the new comment is one I independently verified: gh 2.45.0 embeds `itchyny/gojq`; gojq compiles through Go's `regexp`/RE2; system `jq-1.7` is Oniguruma; the lookaround reproduction is exact; and the 51-vs-51 zero-disagreement figure is my own measurement, accurately reproduced.

Two things I want to credit specifically, because they are better than what I asked for:

1. The test comment now states the limit **as a limit** — "treat a green here as evidence about the pattern's semantics, not about its portability to gojq." That is the correct epistemic framing, not just a factual correction.
2. The retro sentence was fixed at the root, not patched: "through the real engine" became "through a real jq", and the *reason* the test is necessary is now stated engine-neutrally (jq's escape-collapsing before any regex engine). The original reasoning survives; only the false attribution is gone.

The `skipUnless` limit I raised as a secondary is recorded too, and both point at #1362.

## Everything else re-verified at this head, not carried forward

- **Full suite:** 4098 passed, 896 subtests, in a real checkout. Matches your figure.
- **Mutation battery:** same 13 mutations against `f1b3ef0` — **11 killed, 2 survived**, identical to `e6c31e5`. No regression. The two survivors remain the decorative `request`/`reply` map entries I dispositioned as non-defects; correctly still untouched.
- **The jq test itself:** passes, 7 subtests.
- **Production code is untouched since `e6c31e5`** — `git diff` over `wave_status.py` and `trust_signals.py` is empty. The `_CHANGES_REQUESTED_RE` I verified against the live corpus (49→51, difference exactly main#1173 and main#1153, zero spurious matches across 45 PRs) is byte-identical at this head, so that result carries without re-running.
- **CI:** first read showed 18/21 with Mermaid, Pytest and the roster gate `in_progress`. Re-polled to completion: **21 total, 21 success, 0 pending, 0 non-success.**

## The new memory note

`feedback_precommit_stash_reverts_concurrent_edits` is accurate on the mechanism — pre-commit stashes *unstaged* working-tree changes, so an explicit pathspec scopes what is committed, not what is stashed. The observation that this fails "with a well-formed wrong answer and a clean `git status`" is the right generalization, and "re-read the file after editing, not just at write time" is the only detection that actually works. It correctly explains the push anomaly you flagged in-thread. One cosmetic thing, not worth a round: the trailing `[[feedback_verify_after_edit_not_at_write]] if that note exists` reads oddly inside the note — an unresolved wikilink is fine on its own and needs no hedge.

## On the sixth-instance framing

I agree with recording it as an instance rather than a footnote, and I want to be precise about what it does and does not demonstrate, since it is now in a permanent artifact.

It **is** the class: a claim about a mechanism that did not match the mechanism, written into a durable file, one turn after documenting four instances of exactly that.

It is **not** evidence that the test was wrong. The test was and is a genuine improvement over the `_FakeGh` pre-baked count, there was no behavioural defect, and the two engines agree on everything the constant currently uses. The retro says this plainly ("There was no behavioural defect — the block was on the **claim**, correctly"), which is the part that keeps the instance honest rather than self-flagellating. A future reader should take from it that the claim was wrong, not that shelling out to jq was.

## Merge readiness

Approved. Both must-fixes from round one and the one from round two are fixed and verified fixed by re-execution. CI green at `f1b3ef0`, re-polled to completion.

Open deferrals tracked, none blocking: **#1359** (extract the verdict vocabulary into `charter_trailer.py`; the bold-label/bold-value gap is noted there), **#1360** (still open — correct, it closes on merge with the fix, not before), **#1362** (pin the production regex engine).

Merge gating now rests where you placed it: the owner's ruling on **#1349**. That is the right gate — the trust matrix is in this diff and all ten scores are held on it.

---
Requestor: Aino Virtanen
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: #1359, #1360, #1362

parametrization:
## Re-confirm at `f1b3ef0` (Nino Kavtaradze)

Confirmed `e6c31e53bb16f80a51de7cda6315fd825ffd4d02` (my prior `Approved`) is an ancestor of `f1b3ef08e18eab61e50b71b308d9776e62d5d5c0` — 2 additive commits. **CI re-verified at `f1b3ef0` via the check-runs API directly: 21/21 `completed`/`success`, 0 pending.** Fresh isolated scratch clone, not the shared worktree.

**Production code is byte-identical to `e6c31e5`** — verified myself (`git diff e6c31e5 f1b3ef0 -- .claude/lib/wave_status.py .claude/lib/trust_signals.py` is empty), not taken on Aino's word. Only `test_wave_status.py` (comments), two memory files, and `feedback_log.md` changed. My prior 49→51 live re-derivation stands unchanged.

**MF3's RE2-vs-Oniguruma claim, independently reproduced, not relayed:**

```
$ strings $(command -v gh) | rg -i gojq   # confirms gh 2.45.0 embeds itchyny/gojq
$ gh api -X GET /rate_limit --jq '"x" | test("(?=x)x")'
invalid regular expression "(?=x)x": error parsing regexp: invalid or unsupported Perl syntax: `(?=`
```

and confirmed the current (unmutated) `_CHANGES_REQUESTED_RE` matches correctly through the real `gh --jq` path (gojq/RE2) for all three ChangesRequested spellings and correctly excludes `Changeset`/`Approved` — this is a scope-limit finding about a *future* widening using lookarounds, not a live defect in what's shipped. Correctly filed as `#1362` and correctly left in scope-limit territory rather than blocking. One small note: `#1362` isn't listed in the PR body's `Refs` line alongside `#1359`/`#1361` — worth adding for discoverability, not a blocker.

`test_trust_signals.py` + `test_wave_status.py`: 92 passed, 15 subtests, green in my own run.

### Verdict

Re-confirming `Approved` at `f1b3ef0`. Everything since my last verdict is documentation/test-comment-only, independently verified rather than taken on report, and CI is green at this exact head.

---
Requestor: Nino Kavtaradze
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: none new (add #1362 to the PR body's Refs line as a small housekeeping item)

PR #1376, opened 2026-08-10
fix(#1375): retire the deprecated Agent team_name parameter (10 files / 18 references) + hook-block it

### LINKED ISSUE #1375: charter mandates the deprecated Agent team_name parameter (10 files / 18 references); botfarm_inc hard-blocks it on the same harness
## Problem

The charter mandates that every `Agent` spawn pass `team_name: "noorinalabs"`. The harness has since deprecated that parameter, and the live `Agent` tool schema in a current session (Claude Code **v2.1.226**) documents it as:

> `team_name`: **Deprecated; ignored.** The session has a single implicit team.

So the org's most-repeated spawn instruction points at a parameter the tool no longer honours. Every spawn brief written from the charter carries a dead argument.

## Evidence — a sibling repo already reversed this

`~/code/botfarm_inc` (same machine, same harness version, same `CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS=1` user setting) **hard-blocks** the parameter with a PreToolUse hook on the `Agent` matcher, `.claude/hooks/validate_no_team_name.py`, exit 2. Its docstring states the reasoning directly:

> The `team_name` argument is deprecated and MUST NOT be set: passing it (e.g. `team_name: "reddit-bot"`) is a hard error that this hook turns into a clean block, so the spawn fails fast with an actionable message instead of **silently registering a ghost agent against a non-existent team**.
>
> Previously this hook validated that a `~/.claude/teams/<team>/config.json` existed before allowing a `team_name` spawn. That model is gone [...] the correct number of `team_name` values is zero. The check is therefore **inverted** — any spawn carrying a `team_name` is blocked.

So the two repos on this machine hold **opposite** doctrines for the same harness: botfarm_inc blocks what noorinalabs-main mandates.

## Observed symptom

Agents spawned from `noorinalabs-main` are grouped under a named team and must be navigated into (left-arrow) rather than being selectable from the main session view, which is the botfarm_inc behaviour the owner wants.

**Stated honestly: the causal link between `team_name` and that UI grouping is inferred, not measured.** Confirming it requires spawning a throwaway agent with and without the parameter and comparing. What *is* measured is the deprecation and the doctrine conflict above — and that alone justifies the change, independent of the UI question.

## Scope — two distinct uses of `team_name`, only one of which changes

**Category A — the `Agent`-tool parameter (REMOVE).** 14 occurrences:

| file | lines |
|---|---|
| `CLAUDE.md` | 40 |
| `.claude/team/charter/agents/orchestration-model.md` | 64, 65, 70 |
| `.claude/team/charter/agents/naming-and-teams.md` | 25, 27, 45, 47, 49 |
| `.claude/team/charter/agents.md` | 27 |
| `.claude/team/lifecycle.md` | 130 |
| `.claude/skills/session-start/SKILL.md` | 176 |
| `.claude/skills/wave-kickoff/SKILL.md` | 499 |
| `.claude/context/project-architecture.md` | 52, 54, 56 |
| `.claude/memory/feedback_orchestrator_may_spawn_freely.md` | 30 |
| `.claude/memory/section_spawn_delegation.md` | 30 |

**Category B — the skill `args:` frontmatter (DO NOT TOUCH in this issue).** `args: team_name, …` in the frontmatter of `retro`, `team-reset`, `wave-kickoff`, `wave-retro`, `plan-phase`, `wave-start`, `wave-wrapup`. This is a *skill invocation argument*, a different mechanism from the Agent-tool parameter that merely shares a name. It is arguably also stale — nobody invokes `/wave-retro noorinalabs 10 29`, they invoke `/wave-retro 10 29` — but changing skill arg parsing is a separate, riskier change and is deliberately out of scope here. **Filed as a follow-up rather than bundled**, so this PR stays a pure doctrine correction.

## Fix

1. Remove the Category-A mandate from all 11 files; state instead that the session has a single implicit team and `team_name` must not be passed.
2. Port botfarm_inc's `validate_no_team_name` PreToolUse hook so a brief written from a stale cached copy of the charter cannot silently reintroduce it. Enforcement-tier reasoning per `feedback_enforcement_hierarchy` (hook > skill > charter > memory): the charter text is the tier that just failed, so the remedy belongs one tier up.
3. Correct the two memory notes that prescribe the parameter.

## Acceptance

Per the wave-30 bar, the hook lands with a test that **fails against the pre-fix implementation** — trivially satisfied here, since there is no hook today, so any test asserting a `team_name`-carrying spawn is blocked fails before the change. The doctrine edits are verified by a repo-wide `rg` returning zero Category-A matches.

---
Owner-approved 2026-08-09. Found while diagnosing why agents spawned from this repo are not selectable from the main session view.

parametrization:
Closes #1375.

## What

The charter mandated `team_name: "noorinalabs"` on every `Agent` spawn, across 14 sites. The live tool schema on this harness (**v2.1.226**) documents that parameter as:

> `team_name`: **Deprecated; ignored.** The session has a single implicit team.

So the org's most-repeated spawn instruction pointed at a dead argument, and every brief written from the charter carried it.

## Why now — a sibling repo already reversed this

`~/code/botfarm_inc` — same machine, same harness version, same `CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS=1` user setting — **hard-blocks** the parameter via `.claude/hooks/validate_no_team_name.py`. Its docstring records that the hook originally *validated* a team config existed and was later **inverted** rather than deleted, because "the correct number of `team_name` values is zero."

The two repos on this machine therefore held **opposite** doctrines for the same harness. This PR aligns `noorinalabs-main` with the one that matches the tool.

## Scope — two things share the name; only one changes

**Category A — the Agent-tool parameter (removed, 14 sites):** `CLAUDE.md`; charter `agents.md`, `agents/orchestration-model.md` (×3), `agents/naming-and-teams.md` (×5); `team/lifecycle.md`; skills `session-start` + `wave-kickoff`; `context/project-architecture.md`; and the two memory notes that prescribed it.

The per-repo `team_name` **tables** are deleted outright rather than annotated — with the parameter ignored they have no consumer. Per-repo **rosters** under `<repo>/.claude/team/roster/` are untouched and are what "which repo's team" now means.

**Category B — the `args: team_name` frontmatter in 7 skills (deliberately NOT touched):** that is a *skill-invocation argument*, a different mechanism that merely shares the name. It is arguably also stale — nobody invokes `/wave-retro noorinalabs 10 29` — but changing skill arg parsing is a separate, riskier change. Filed as a follow-up rather than bundled, so this PR stays a pure doctrine correction.

## Enforcement

New `validate_no_team_name` PreToolUse hook on the `Agent` matcher, registered **first** so a spawn carrying a dead parameter fails before the more expensive brief/ontology checks run.

Rationale is the enforcement hierarchy (`feedback_enforcement_hierarchy`: hook > skill > charter > memory). Charter text is precisely the tier that just failed — a sentence cannot stop a brief written from a stale cached copy of the charter; a hook can.

Two deliberate design calls, both mirroring `enforce_ontology_context`:

- **Empty/whitespace `team_name` is treated as absent** — it names no team, so there is nothing to correct and blocking it would be pure friction.
- **Malformed input fails OPEN** — a PreToolUse hook that raises surfaces to the user as block-with-error, which is worse than allowing through a shape the tool will reject itself.

## Verification

Beyond the unit tests, the hook was driven **end-to-end through its real entry point** and shown to *separate the two classes* — not merely to fire:

| input | exit | result |
|---|---|---|
| `Agent` + `team_name: "noorinalabs"` | **2** | block, with an actionable reason |
| `Agent`, no `team_name` | 0 | allow |
| `Bash` carrying a `team_name` key | 0 | allow (guards against a widened matcher) |

That third row matters: a test asserting only "it blocks" would also pass against a hook that blocks everything.

- **17 new tests**; full suite **4177 passed / 947 subtests**.
- `ruff format --check` + `ruff check` clean; commit-stage and push-stage hooks green (cspell, mermaid-render, memory-budget, office-drift, doc-freshness included).
- `settings.json` diff is **+10 lines with no reformat churn**.
- Repo-wide `rg` for Category-A mandates returns zero, with one intentional exception: a historical note in `section_spawn_delegation.md` that *explains* the retirement.

## On the acceptance bar

Wave-30's bar is "a fix lands only with a test that fails against the pre-fix implementation." Here that is satisfied **structurally rather than interestingly** — with no hook before this commit, every case fails at import. Recording that plainly rather than presenting it as strong evidence: the bar's spirit is aimed at *modified* gates, and a brand-new gate meets it trivially.

## What is NOT proven

The owner's original symptom was that agents spawned from this repo are grouped under a named team and must be navigated into (left-arrow) rather than being selectable from the main session view. **The causal link between `team_name` and that UI grouping is inferred, not measured** — confirming it would require spawning a throwaway agent with and without the parameter and comparing.

What *is* measured is the deprecation and the cross-repo doctrine conflict, and that alone justifies the change. Reviewers should not treat the UI improvement as established.

---

**Reviewers:** please pay particular attention to (1) whether any Category-A site was missed or any Category-B site was touched by mistake, and (2) the hook's fail-open stance on malformed input — that is a deliberate choice inherited from `enforce_ontology_context`, and it is the kind of choice that deserves a second opinion rather than inheritance-by-default.

https://claude.ai/code/session_019rKDyqj6F6MhDtjX75RBm5


parametrization:
Reviewed at head `4d4d9c8d4bdb8c45f4da93d46441ea3829e85e8d` via `gh pr diff`/`gh api` (no local checkout). Ontology: semantic overlay 301 tracked / 0 dirty; structural index current (0 source files changed since last build).

**1. Coverage — independently re-derived, matches.** Ran my own `rg --hidden --no-ignore "team_name"` sweep of the pre-fix tree (root-level tracked files only — excluded `.git/`, `.claude/worktrees/`, and two stale `scratchpad/wz-1248` / `scratchpad/wm-1248` full-repo copies that are noise, not doctrine). Found Category-A `team_name` mandates in exactly the same 10 files the diff touches: `CLAUDE.md`, `agents.md`, `orchestration-model.md` (×3), `naming-and-teams.md` (×5), `lifecycle.md`, `session-start/SKILL.md`, `wave-kickoff/SKILL.md`, `project-architecture.md` (×3), and the two memory notes. Post-diff, none of those strings remain as live mandates. The 7 Category-B `args: team_name` skill-frontmatter sites (`retro`, `team-reset`, `wave-kickoff`, `wave-retro`, `plan-phase`, `wave-start`, `wave-wrapup`) are confirmed untouched — including `wave-kickoff/SKILL.md`, which the diff edits at line ~499 (prose) while leaving its own `args: team_name` frontmatter at line 4 alone. No Category-A site missed, no Category-B site touched.

Nit, non-blocking: the issue/PR title says "14 sites," but the issue body's own Category-A table lists 18 distinct line references when summed row-by-row (1+3+5+1+1+1+1+3+1+1). Coverage is correct against the table either way — this is a headline-vs-table arithmetic slip, not a gap.

**2. Information loss — checked, none.** The deleted per-repo `team_name` tables mapped session-context → a now-ignored parameter string; nothing else referenced them. Per-repo rosters (`<repo>/.claude/team/roster/`) are a separate, untouched artifact class and remain populated and canonical for commit identity/domain ownership/reviewer pairing, confirming the PR's claim.

**Real gap found (filed as tech debt, not a blocker for this PR):** 3 of the 7 child repos still carry their own live `team_name` mandate, now inconsistent with the parent's blanket retirement — `noorinalabs-user-service` charter:19, `noorinalabs-deploy` charter:602-610 (a full table), `noorinalabs-data-acquisition` charter:504-506. This PR's own hook is repo-scoped (lives in this repo's `.claude/settings.json`) and does not reach an isolated session opened directly in those 3 repos. Filed as #1377 — the same "opposite doctrines" problem #1375 diagnosed between noorinalabs-main and botfarm_inc now exists one level down, between noorinalabs-main and 3 children. Out of stated scope for #1376 (which is `noorinalabs-main`-only, correctly), but should not go undiscovered.

**3. Memory-note edits — coherent.** `feedback_orchestrator_may_spawn_freely.md` correctly replaces the old mandate with "never pass `team_name`" + hook citation. `section_spawn_delegation.md`'s one-liner correctly frames the retired clause in the past tense ("used to carry... was retired by #1375") — reads as history, not as a live instruction, matching the PR's claim of being the one intentional survivor.

**4. Hook registration order — correct, low risk.** `validate_no_team_name` is inserted as its own `Agent`-matcher block ahead of the existing one (`validate_wave_context`, `enforce_ontology_context`, `warn_oversized_brief`). Read all three existing hooks: `warn_oversized_brief` is advisory-only (never blocks), and the two blocking hooks are pure functions of `tool_input` with no side effects on the allow path — so registration order changes only *which* block reason surfaces first on a spawn that violates multiple gates, not any hook's individual behavior. That's the PR's stated intent (cheap check first) and is a reasonable trade-off, not an oversight.

**5. Numbers re-verified myself, not re-read:** CI 21/21 green, recounted from `statusCheckRollup` at the head SHA. 17 new tests — recounted test methods + parametrize cases in the diff (2 + 4 + 4 + 3 + 4 = 17), exact match. `settings.json` diff — recounted, exactly +10 lines, no reformat churn. Could **not** independently verify "4177 passed / 947 subtests" (full suite) without running it, which is out of scope for a read-only diff review — relying on CI's own green "Pytest — hooks and lib test suites" for that signal.

**6. Design choices** (empty/whitespace `team_name` treated as absent; fail-open on malformed input) — reviewed, consistent with `enforce_ontology_context` precedent, and covered by the `MalformedInputFailsOpen` / `test_spawn_without_team_name_is_allowed` test classes.

No must-fix findings. Approving.

---
Requestor: Weronika Zielinska
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1377

parametrization:
## Correction: the headline count was wrong (found by Weronika Zielinska)

The review caught a real counting error in my own issue and PR titles, and I've verified it independently rather than taking it on faith. **It does not change the diff, the coverage, or the verdict** — every Category-A site is still correctly handled — but the headline number was stated under the wrong denominator.

Three numbers, all correct, none interchangeable:

| number | denominator |
|---|---|
| **10** | files changed (Category A) |
| **14** | lines matching the specific `rg` regex I used when first scoping |
| **18** | line-references listed in #1375's own body table |

I wrote **"14 sites"** in both titles. "Sites" naturally reads as *places in the charter*, which is the 18 — so the headline under-counted against the table sitting directly beneath it. The 14 was really "lines my regex matched", a denominator that appears nowhere in the issue and that I never stated.

Titles on #1375 and #1376 are now corrected to **"10 files / 18 references"**, which carries its denominators inside the number.

**The commit message still says "14 sites" and is deliberately not being amended.** Amending would change the head SHA, and the Opus merge-gate review is in flight against `4d4d9c8` right now — per this org's head-SHA anchoring rule, an approval predating the head is not an approval. Trading a live review for a headline-count fix is the wrong way round. This comment is the durable record instead.

### Why this is worth more than a footnote

This is a textbook instance of [`feedback_state_the_denominator_with_the_number`](https://github.com/noorinalabs/noorinalabs-main/blob/main/.claude/memory/feedback_state_the_denominator_with_the_number.md) — *"a wrong number gets challenged; a right-under-a-different-denominator number gets ABSORBED, because each party verifies their own and attributes the mismatch to the other."* That note was in context at session start and I reproduced the exact failure it describes anyway, which is the note's own stated point: knowing the rule does not prevent it.

It is also the second counting error of this session. The first was a `comm` sort-order mismatch during `/wave-scope 10 30` that reported "85 carry-forwards closed / 0 open" when the truth was 2 closed / 83 open — that one announced itself with a `comm: file 2 is not in sorted order` warning and was caught immediately. **This one was silent**, which is exactly why it survived to the title of two artifacts: nothing errored, and 14 is a perfectly plausible count.

Credit to the reviewer for checking the number against the table rather than reading past it — the brief asked her to verify the body's claims independently, and this is what that produced.

parametrization:
**Merge-gate review** at head `4d4d9c8d4bdb8c45f4da93d46441ea3829e85e8d`. Files read via `gh api contents?ref=<sha>`; execution done in a detached worktree pinned to that SHA, byte-verified against the API content before running anything. Ontology: semantic overlay 301 tracked / 0 dirty (exit 0 from `checksums_io.py status`).

I re-ran every number I rely on. **CI 21/21 SUCCESS** at the head SHA, recounted from `check-runs` and `statusCheckRollup` independently. **Full suite reproduces exactly: 4177 passed / 947 subtests**, so that figure in the PR body is honest. The scope claims hold — my own repo-wide sweep at head finds every surviving `team_name` mention in prose is a *prohibition*, and all 7 Category-B `args: team_name` frontmatter sites are untouched.

The question I was asked is not "does the hook work" but "can this suite tell a working hook from a broken one." Mutation testing answers that, and the answer is **yes for the decision function, no for the block delivery**.

## Mutation results

15 mutants, each applied singly against the **full** suite (`.claude/hooks/tests/` + `.claude/lib/tests/`), then restored. A **null mutant was run first as a control** and correctly survived — worth stating, because my first harness pass silently passed an unrecognised `--timeout` to pytest, which returned rc=4 for *every* run and reported all 15 mutants "killed." That run was a gate incapable of the outcome it described, which is this wave's theme arriving uninvited. The results below are from the corrected harness.

**All five guards you named are covered.** Killed: the `tool_name != "Agent"` early return; the `isinstance(tool_input, dict)` check; the `isinstance(team_name, str)` check; the `.strip()`; and every element of the block dict I could weaken independently — `"decision": "block"` -> `"allow"`, dropping the `{team_name}` interpolation, dropping the remedy sentence, dropping `SendMessage`, dropping `#1375`. Also killed: inverting the empty-check to always-allow, and renaming the lookup key to `teamName`. That is a genuinely well-pinned `check()`.

**Four survivors, all in the `main()` / `_check_and_log` layer:**

| # | exact mutation | result |
|---|---|---|
| **M11** | `from _hook_main import run_blocking` -> `run_advisory`, and `run_blocking(_check_and_log, ...)` -> `run_advisory(_check_and_log, ...)` | **4177 passed — SURVIVED** |
| M12 | delete `log_pretooluse_block("validate_no_team_name", prompt, result["reason"])` | 4177 passed — survived |
| M13 | `run_blocking(_check_and_log, ...)` -> `run_blocking(check, ...)` | 4177 passed — survived |
| M14 | `prompt = raw[:200]` -> `prompt = raw` | 4177 passed — survived |

**M11 is the finding.** Under that mutation I drove the hook through its registered entry point with the exact payload it exists to stop:

```
$ printf '%s' '{"tool_name":"Agent","tool_input":{"prompt":"spawn","team_name":"noorinalabs"}}' \
    | python3 .claude/hooks/validate_no_team_name.py
exit=0
```

The gate silently allows, and **4177 tests stay green**. That is #1243 exactly — a BLOCK gate that fails open because a PreToolUse non-2 exit does not block — and this suite cannot see it. `test_hook_registration_coverage.py` does not close it either: it asserts only that a hook *imports* `_hook_main`, and the mutant still does.

## Exit-code correctness — verified, and clean

Driven end-to-end, not read. The **shipped** hook is correct:

| payload | exit |
|---|---|
| `Agent` + `team_name: "noorinalabs"` | **2** (block, reason on stdout) |
| `Agent` + `team_name: "  noorinalabs  "` | **2** |
| `Agent`, no `team_name` | 0 |
| `Bash` carrying a `team_name` key | 0 |
| `Agent` + `team_name: ["noorinalabs"]` / `{...}` / `""` | 0 |
| non-`dict` `tool_input` / non-JSON stdin / empty stdin | 0 |

Your end-to-end table reproduces. The hook does **not** have the #1243 defect; what it lacks is a test that would notice if it ever acquired one.

## Can the gate be bypassed by ordering? No — and I have evidence, not an assumption

This PR creates the first duplicate-matcher situation in this `settings.json`: two separate `Agent` groups. The risk worth checking was **shadowing** — if the harness took only the first matching group, the new group would silently disable `validate_wave_context`, `enforce_ontology_context` and `warn_oversized_brief`. It does not: the sibling repo `~/code/botfarm_inc`, same machine and same harness, registers **seven** separate `Agent`-matcher groups that all run in production. Every matching group runs.

#1245 does not apply. That is a `dispatcher.py` aggregation defect, and `dispatcher.py` is registered only for `Bash`/`Edit`/`Write`/`NotebookEdit` — `Agent` hooks are invoked directly as independent processes, so there is no aggregation to discard. The bare command with no `|| …` tail is also right: a missing script exits non-zero and fails **closed**, and adding a tail would trip `_FAIL_OPEN_GUARD`.

One unverified claim in the PR body, non-blocking: "registered first so a spawn … fails before the more expensive brief/ontology checks run" assumes the harness short-circuits on the first blocking group. I could not confirm that, and it is cosmetic either way.

## The fail-open stance — you asked for a call, so here it is

**Split the question; the two halves have different answers.**

The fail-open on a non-`dict` `tool_input` is **right**. That is a whole-payload shape the tool validates itself, it is not `team_name`-specific, and raising there would surface as block-with-error.

The fail-open on a non-`str` `team_name` is the **wrong predicate**, though currently costless. The gate's own docstring says "the correct number of `team_name` values in any spawn is **zero**" — the signal it cares about is *key presence*. It then decides by inspecting the value's *type*, which is a different question. So `null`, `[]`, `{}` and `""` all pass a gate whose stated principle says they should not.

Is it an exploitable bypass? **No, on measured evidence.** Across **1266 real `Agent` tool_use calls** in the local transcript corpus, `team_name` appears in **165** and its value is *always* the string `'noorinalabs'` — zero non-strings, zero empties, and the harness never injects the key itself. The only writer is an LLM copying the charter's literal `team_name: "noorinalabs"`. So the adversarial framing does not bite here; the sloppy-caller framing is the real one, and a sloppy caller writing `team_name: null` currently gets no correction.

That same measurement is why the tightening is free: a key-presence predicate blocks exactly the same 165 calls and newly blocks **0**. Identical on real traffic, strictly stronger on shapes that have not occurred. I am filing it as tech debt rather than a must-fix (#1378, with the patch) because the shipped behaviour is safe and this is precision, not a defect. But my answer to "is inheritance-by-default right here" is **no** — inherit the `tool_input` guard, drop the `team_name` one.

Same reasoning covers the empty-string call. `""`-is-absent is defensible on its own terms, and I would not block on it; note only that it is an exception carved by a *side effect of the type check* rather than stated once. The suggested predicate makes it the single explicit exception.

## On the acceptance bar

Your framing is correct and I am not going to pretend otherwise: with no prior module, "every case fails at import" is the bar met trivially. Recording that plainly was the right call. But it means the bar did no work here, and mutation testing is what has to do it instead — which is how M11 surfaced. For a **new blocking gate** I want the entry point pinned, and that is not a novel demand: `test_validate_pr_review.py`, `test_validate_wave_label_evidence.py` and `test_validate_commit_identity.py` all drive their hook through `subprocess` and assert the exit code. This file is the departure from local practice, not the fix.

## Must-fix

**MF1 — pin the exit code at the entry point.** Add a test that runs `.claude/hooks/validate_no_team_name.py` as a subprocess and asserts `returncode == 2` with a `decision: block` body on stdout for a `team_name`-carrying `Agent` payload, and `returncode == 0` for a clean spawn, a `Bash` call, and malformed stdin. That kills M11 and closes the #1243 class permanently. Roughly ten lines, matching the three sibling test files above.

**MF2 — one dangling anchor this PR introduced.** `.claude/team/charter/agents.md:27` links to `agents/naming-and-teams.md#team-names`, but this PR renames that heading to `## Team Names — RETIRED: never pass \`team_name\``, whose anchor is `#team-names--retired-never-pass-team_name`. `orchestration-model.md:64` — also touched by this PR — uses the new anchor correctly, so one of the two call sites was updated and the other was not. CI cannot catch it: lychee runs without `--include-fragments`, so it verifies the file exists and never the fragment (filed as #1380). One-word fix.

Neither is a defect in the hook's live behaviour. Both are the difference between a gate that happens to work and a gate that is held in place.

Nothing else blocks. The doctrine correction itself is clean, the scope split between Category A and B is exactly as described, and the "what is NOT proven" section about the UI grouping is the right kind of honesty — I did not attempt to verify that causal link either, and agree it is not needed to justify the change.

---
Requestor: Nino Kavtaradze
Requestee: Aino Virtanen
TechDebt: #1378, #1379, #1380

parametrization:
## Both must-fixes addressed — `2101124`

@Nino — thank you for the mutation run. Discarding your first harness pass over the `--timeout` rc=4 artifact, and running a null-mutant control before trusting the results, is the difference between a mutation report and a mutation *measurement*. The finding it produced is the best one this PR could have received.

### MF1 — fixed, and it is worse than reported

I reproduced M11 before fixing it. Under `run_blocking` → `run_advisory`, the hook does not merely fail to block:

```
$ echo '{"tool_name":"Agent","tool_input":{"prompt":"spawn","team_name":"noorinalabs"}}' \
    | python3 .claude/hooks/validate_no_team_name.py
{"decision": "block", "reason": "BLOCKED: Agent spawn passes team_name=\"noorinalabs\", ..."}
exit=0
```

It prints the **byte-identical** block payload and exits 0 — the gate announces `BLOCKED` while silently allowing. The passing and broken states are observationally identical on stdout; the exit code is the *only* discriminator. That is why 17 `check()`-level tests could not see it, and it is the wave-30 theme reproduced inside the fix for the wave-30 theme.

Added `MainEntrypointExitCode` — 8 subprocess tests through the registered entry point, following `test_warn_pipe_mask_rc.py`'s local idiom rather than inventing one, exactly as you noted this file was the departure from practice.

**Verified the fix kills the mutant** rather than assuming it:

| state | result |
|---|---|
| M11 applied | **2 failed** / 23 passed |
| restored | 25 passed |

Worth recording: `test_team_name_spawn_emits_block_decision` **still passes under the mutant**, because the mutant does print the block JSON. So of the 8 new cases exactly one class of assertion is load-bearing, and the class docstring says so — a reader must not conclude that "8 entry-point tests" means 8 independent guards.

### MF2 — fixed, after my first verification of it was wrong

`agents.md:27` now points at `#team-names--retired-never-pass-team_name`, matching `orchestration-model.md:64`.

My first check reported this link **DEAD after fixing it**, which was a false alarm from a broken instrument: the extraction regex `[a-z0-9-]+` stopped at the underscore, and my slug function stripped `_` — which GitHub preserves. Re-derived with GitHub's actual rule (drop all but alphanumeric/space/hyphen/underscore, spaces → hyphens); all **5** `naming-and-teams.md` fragments across both files resolve. Flagging it because a fragment-checker that is wrong about underscores would also be wrong about #1380's fix.

### Deliberately NOT in this round

**#1378, #1379, #1380** stay as you filed them. You scoped all three as non-blocking on measured evidence, and a must-fix round should address must-fixes rather than quietly widen scope under cover of a re-review.

On #1378 specifically — your reasoning is right and I want it on the record rather than buried in an issue: the docstring says the correct number of values is *zero* (key-presence), while the code branches on value *type*, so `null`/`[]`/`{}` pass a gate whose stated principle excludes them. That is a prose-guarantee-vs-mechanism gap in something I wrote two hours ago. Your 1266-call / 165-occurrence measurement is what makes it safely deferrable — and the same measurement is what makes the fix free, so it should not sit long.

### Also corrected

"14 charter sites" → **"10 files / 18 references"** in a test docstring. Weronika found the headline count error; 14 was the match count of one ad-hoc regex, a denominator stated nowhere. Titles on #1375/#1376 corrected earlier; the first commit message retains "14 sites" and was deliberately not amended, because your review was in flight against `4d4d9c8` and a new SHA would have invalidated it.

### Verification

- Full suite **4185 passed** / 947 subtests (was 4177; +8).
- `ruff format --check` + `ruff check` clean across all 109 hook files.
- Anchors: 5/5 resolve.

Re-review requested on `2101124`.

parametrization:
**Merge-gate re-review** at head `2101124f9062dfcb05034bc716398a41570f61b9`. Delta from my prior review confirmed against the compare API: exactly two files, `test_validate_no_team_name.py` (+71/−2) and `agents.md` (+1/−1). `validate_no_team_name.py` is **byte-identical** to `4d4d9c8`, so the end-to-end behaviour I verified last round still stands unchanged. Both must-fixes verified by instrument, not by reading.

## MF1 — closed, and it holds against sharper mutants than the one I reported

Re-ran the battery against the new suite in a worktree pinned to `2101124`, null-mutant control first (survived, 4185 passed / 947 subtests — your `+8` reproduces exactly).

| mutation | before MF1 | now |
|---|---|---|
| `run_blocking` → `run_advisory` (+ call site) | survived 4177 | **KILLED** (2 failed) |
| **`run_blocking` → hand-rolled main that prints the block JSON and `sys.exit(1)`** | — | **KILLED** (2 failed) |
| **predicate removed: blocks every `Agent` call, reason intact** | — | **KILLED** (5 failed) |
| **block JSON emitted to stderr instead of stdout, exit 2 correct** | — | **KILLED** (1 failed) |
| **`settings.json` registration group deleted** | — | **KILLED** (1 failed) |

The second row is the one I most wanted and did not ask for explicitly: it is #1243's *literal* shape — the gate delivers a correct, complete block payload and exits **1**. Your assertions are `assertEqual(returncode, 2)` rather than a truthiness or `!= 0` check, so they catch it. An `assertNotEqual(rc, 0)` would have passed that mutant and looked identical in a diff.

Row three matters for the other half of the wave's theme: the class can distinguish **can-fail** from **can-pass**. A gate that blocks everything is killed by the four allow-path cases, so the suite pins a discriminator, not just a tripwire.

Row five is a bonus I checked rather than assumed — the gate cannot be silently unhooked, because `test_hook_registration_coverage.py` catches removal of the `settings.json` entry.

**Your correction to my report is right and it is the more precise statement.** I described M11 as "exits 0 and silently allows"; you reproduced that it prints the *byte-identical* `{"decision": "block", ...}` to stdout while exiting 0. That is strictly worse than what I wrote — the failure is not silent, it is **mislabelled**, which is harder to catch by eye than silence. Noting `test_team_name_spawn_emits_block_decision` still passes under the mutant, in the docstring, is the right way to record which assertion is load-bearing; that test is not wasted (it kills the stderr mutant above), but it is not what closes #1243.

## MF2 — closed, and your underscore correction is correct

Verified independently with GitHub's real slug rule rather than by eye. `## Team Names — RETIRED: never pass \`team_name\`` slugs to `team-names--retired-never-pass-team_name` — **the underscore is preserved**, because `github-slugger` filters on `[\w\- ]` and `\w` includes `_`. Your first extraction regex `[a-z0-9-]+` stopped at the underscore and your slug function stripped it; both errors point the same way and produced a confident false DEAD. Good catch on your own work, and you are right that it matters for #1380 — a fragment checker wrong about `_` would report false positives across every `team_name`/`args:`-style anchor in the charter.

All four distinct fragments targeting `naming-and-teams.md` (5 link instances — two now share the retargeted anchor) resolve to real headings. Repo-wide: **136 fragment links across 162 markdown files, 0 dangling.**

## Your question 1 — are the 8 cases the right set?

Yes. I went looking for a shape they cannot distinguish and found nothing outside the layer already tracked in #1379. The block-delivery contract — exit code, stdout channel, stdout being parseable JSON, the allow paths, and both malformed-stdin paths — is pinned. `M12`/`M13`/`M14` (annunaki logging deleted, wrapper bypassed, truncation removed) still survive at 4185, exactly as you said and exactly as I scoped them; that is #1379's job, not this PR's.

One measured observation, and I am **not** asking you to change it: the subprocess helper inherits the parent environment rather than setting it, so with `ENVIRONMENT` unset a block-path case writes a real record to `.claude/annunaki/errors.jsonl` — the #452 pollution class. I checked before raising it and this is **not** a departure: `test_validate_commit_identity.py`, `test_validate_pr_review.py` and `test_validate_wave_label_evidence.py` all drive annunaki-writing hooks through `subprocess.run` with no `env=` too. You followed the idiom; the idiom is the gap. Filed repo-wide as #1381. Leave it out of this PR.

## Your question 2 — should MF2 carry a test?

**No, and I would decline it if you offered.** A test pinning this one anchor guards 1 of 136 fragment links while 135 stay unguarded — that is a gate that reads as coverage and delivers a sample, which is the shape this wave exists to stop. The honest options are "all fragments" or "none," and "all fragments" is `.lychee.toml`, not pytest.

What I can now tell you is that the systemic fix is cheaper than I claimed when I filed it. I hedged in #1380 that enabling `include_fragments` would need a triage backlog. Measured at your head SHA: **0 dangling repo-wide.** So it should be a one-line config change that lands green, not a cleanup project — I have posted that correction to #1380.

**Provenance limit:** lychee is not installed on this machine, so that 0 comes from my own reimplementation of the slugger, not from lychee itself. Mine is a line-regex over ATX headings; it does not handle Setext headings or explicit `<a name>` anchors, and lychee parses markdown properly. That is precisely why the change belongs in its own PR where the CI run is the proof — and why putting an unexecuted lychee config change into this one would be trading a verified fix for an unverified gate.

## Not addressed, and correctly so

You left #1378, #1379, #1380 open and said you would not widen a must-fix round under cover of re-review. That is the right call and I want it on the record as such: I scoped all three non-blocking on measured evidence, and a reviewer who then accepts them being folded into a re-review round has made the original scoping meaningless. #1378 in particular is a design change to a live gate and deserves its own diff and its own two reviewers, not a hurried commit on an approved branch.

## Re-verified this round

CI **21/21 SUCCESS** at `2101124` from `statusCheckRollup` and `check-runs` independently. Full suite **4185 passed / 947 subtests**, reproduced locally. Hook file byte-identical to the previously-reviewed SHA. Worktree removed.

Both must-fixes are closed by instrument. Approving.

---
Requestor: Nino Kavtaradze
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1378, #1379, #1380, #1381

parametrization:
Re-confirming at new head `2101124f9062dfcb05034bc716398a41570f61b9` (prior verdict at `4d4d9c8` went stale per #950). Re-derived the delta myself via the compare API rather than trusting either report — `gh api .../compare/4d4d9c8...2101124` returns exactly **2 files**, matching the claim: `validate_no_team_name.py` itself does not appear in that file list at all (byte-identical, confirmed independently, not just asserted).

**1. `test_validate_no_team_name.py` (+71/−2) — read the full patch.** Adds `MainEntrypointExitCode`, 8 subprocess tests that invoke the real `python3 validate_no_team_name.py` entry point (not `check()` directly) and assert on process exit code + stdout JSON. This is a genuine gap close, not padding: every prior test called `hook.check()` in-process, so a `run_blocking` → `run_advisory` mutation in `main()` would leave all 17 original tests green while the hook prints the identical `{"decision": "block", ...}` payload and exits 0 — announcing a block while allowing the spawn through. Reproduced-live claim is consistent with the #1243 fail-open shape cited. The 8 cases cover: blocked exits 2, block payload shape, whitespace-padded `team_name` still exits 2, clean spawn exits 0, non-`Agent` tool exits 0, non-dict `tool_input` exits 0, empty stdin exits 0, non-JSON stdin exits 0 — a reasonable spread across the same allow/block boundary already unit-tested, now pinned at the process boundary. Uses `_test_helpers.HOOKS_DIR` (checked it resolves to `.claude/hooks/`, so `_HOOKS_DIR / "validate_no_team_name.py"` is a real path) — no test infra issue.

Also note: this same patch's docstring quietly drops the "14 charter sites" framing for "10 files / 18 references" — matches the arithmetic I flagged in my first pass. Good.

**2. `agents.md` (+1/−1) — the doctrine-link fix, independently re-derived.** Did not reuse either of your slug computations. Fetched `naming-and-teams.md` at `2101124` directly and read the literal heading: `## Team Names — RETIRED: never pass `team_name`` (line 43). Ran GitHub's actual slugger rule by hand: strip markdown (backticks around `team_name`), lowercase, delete any character outside `[\w\-\s]` (the em-dash and the colon both get **deleted**, not replaced with a space — that's what produces the double-hyphen, since the em-dash sits between two spaces that both survive and then convert), replace remaining spaces with hyphens. That gives `team-names--retired-never-pass-team_name`, matching the target exactly. Confirmed both consumers now point at it: `orchestration-model.md:64` and `agents.md:27` both resolve to `#team-names--retired-never-pass-team_name` at `2101124`. No collision risk — only one `## Team Names…` heading exists in the target file.

**3. CI — reconfirmed at the new head, not re-read:** `gh pr view --json headRefOid,statusCheckRollup` at `2101124` shows 21/21 `SUCCESS`. Could not re-verify "4185 passed / 947 subtests" without running the full suite — same limitation as my first pass, noted rather than assumed.

Nothing in the delta changes my original findings (coverage, memory-note framing, hook ordering) — those all still hold since none of the files they touch are in this delta. #1377 (child-repo doctrine gap) stands as filed, unaffected by this delta.

No must-fix findings. Approving at `2101124`.

---
Requestor: Weronika Zielinska
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1377

nino:
@parametrization fyi I'm still not the nino you're looking for

Issue #1379, opened 2026-08-10
validate_no_team_name: _check_and_log side effects are untested — 3 mutants survive the full suite


parametrization:
Follow-up from the merge-gate review of #1376. Mutation testing at head `4d4d9c8` against the full suite (`.claude/hooks/tests/` + `.claude/lib/tests/`, 4177 passed / 947 subtests) found three surviving mutants, all inside the untested `_check_and_log` side-effect layer of `.claude/hooks/validate_no_team_name.py`.

Method: apply one mutation, run the full suite, restore. A null mutant (no-op replacement) was run first as a control and correctly left the suite green, so a green result here means the mutant genuinely survived rather than the harness silently failing.

| # | exact mutation | suite result |
|---|---|---|
| M12 | delete the line `log_pretooluse_block("validate_no_team_name", prompt, result["reason"])` from `_check_and_log` | **4177 passed** — survived |
| M13 | `run_blocking(_check_and_log, ...)` -> `run_blocking(check, ...)` in `main()` (bypasses the wrapper entirely) | **4177 passed** — survived |
| M14 | `prompt = raw[:200]` -> `prompt = raw` (truncation removed) | **4177 passed** — survived |

M12 and M13 have the same live effect: **the gate blocks but records nothing**. The Annunaki error log would then show zero `validate_no_team_name` block events, and a reader would reasonably conclude from that zero that nobody is passing `team_name` any more — a silent instrument reporting as a clean result. That is the failure shape `feedback_silent_zero_is_not_a_measurement` describes.

M14 is cosmetic (unbounded prompt text into the log record) but is likewise unpinned.

## Suggested fix

Add a test that monkeypatches / captures `validate_no_team_name.log_pretooluse_block` and asserts, for a `team_name`-carrying spawn driven through `_check_and_log`:

1. it is called exactly once (kills M12 and M13),
2. with the hook name and the block reason, and
3. with a prompt argument of length <= 200 for an over-long prompt (kills M14),

and that it is **not** called for an allowed spawn.

Related: #1376 merge-gate review, and the entry-point coverage gap tracked separately.

Issue #1217, opened 2026-08-02
tech-debt(test): two #1210 mutations survive the hook4 suite — unknown-initial over-exclusion and the same-surname dedupe key are unguarded


parametrization:
## Two mutations survive the #1215 suite

From the independent mutation battery run during the merge-gate review of #1215
(14 mutations; 11 caught, 3 survived — the third is tracked separately as a must-fix on
the PR itself). Both survivors weaken the #1210 guard **relative to the shipped code**;
neither is a fail-open relative to pre-#1210, which is why they are tech-debt rather than
must-fix.

### M10 — the "unknown initial ⇒ over-exclude" claim is unguarded

`CommitAuthorIdentity`'s docstring states:

> `initial` is `""` when none is derivable (a single-token author name), which
> `is_branch_author` reads as "unknown" and falls back to lastname-only — the
> OVER-excluding, fail-closed direction.

Shipped code does exactly that. But this mutation of `is_self_review` in
`.claude/hooks/validate_pr_review.py` keeps the whole suite **green**:

```python
    return any(
        identity.initial != ""
        and is_branch_author(requestor, identity.lastname, identity.initial)
        for identity in commit_authors
    )
```

That flips the stated degradation direction — a commit authored with a mononym
(`git -c user.name="Virtanen"`) would stop excluding anyone, so a self-approval from that
branch would count. 350 passed, 53 subtests passed.

**Guard:** assert `is_self_review("Aino Virtanen", "", "", (CommitAuthorIdentity(lastname="Virtanen", initial="", display="Virtanen"),))` is `True`.

### M14 — the dedupe key's initial component is unguarded

In `commit_author_identities`:

```python
identities.setdefault((identity.lastname.lower(), identity.initial), identity)
```

Dropping `identity.initial` from that key keeps the suite green. Under the mutation, a
branch authored by **both** Lucas Ferreira and Santiago Ferreira collapses to whichever
appeared first, so the second one's verdict is no longer excluded — the main#1172
collision, re-created in the dedupe rather than in the comparison.

`test_every_non_merge_author_is_returned_not_just_the_latest` does not catch it because
its two authors are `Ferreira` and `Virtanen` — different surnames.

**Guard:** the same test with two same-surname, different-initial authors, asserting both
`("ferreira", "l")` and `("ferreira", "s")` are derived.

Both fixes are test-only; no production change is needed.

parametrization:
[...]

Issue #677, opened 2026-06-14
tech-debt(promotion-audit): empty-slug over-count produces false 24-item charter→skill AUTO cascade


parametrization:
## Source
Found during a standalone `/promotion-audit` run for p5-wave-3 (2026-06-14).

## Bug
A naive audit run flagged **24 charter sections as AUTO charter→skill promotions** — a false cascade that, if acted on, would auto-open 24 bogus skill-scaffold PRs (charter rules wrongly turned into skills).

Two compounding defects:
1. **`count_skill_invocations("")` returns 635** (empty/blank slug matches everything). A real missing slug returns 0; only the empty string over-counts. So any code path that passes `""` trips the AUTO threshold (5) for every marked section, with an identical bogus `signal=635`.
2. **Attribute-name mismatch in the documented usage.** SKILL.md Step 3's worked example reads `sections[0].promoted_to_slug`, but the `Section` dataclass exposes `promoted_to` / `promotion_target` — there is **no** `promoted_to_slug`. `getattr(s,'promoted_to_slug',None) or ""` silently degrades to `""`, feeding defect #1.

Net: an unpromoted charter section (no skill yet → no invocations) should score **signal 0 → KEPT**, but instead scores 635 → AUTO.

## Fix
- `count_skill_invocations`: return **0** for empty/None/blank slug (never match-all).
- `classify_section`: treat a section with no resolved target slug as signal 0 (KEPT), not AUTO.
- SKILL.md Step 3: correct `promoted_to_slug` → `promoted_to` (and update `helpers` if a slug accessor is intended).
- Add a regression test: a marked-but-unpromoted section classifies KEPT, not AUTO.

## Acceptance
- [ ] `count_skill_invocations("")` == 0 (test).
- [ ] Standalone `/promotion-audit` on current repo state yields 0 AUTO / 0 DECIDE (smoke test already asserts this — make it pass on the real charter, not just fixtures).
- [ ] SKILL.md worked example uses the real attribute name.

## Workaround used this run
The audit was re-run passing `section.promoted_to` (None when unpromoted) and substituting signal 0 for empty slugs → correct result: 0 AUTO · 0 DECIDE · 240 KEPT · 4 SUPERSEDED · 21 ALREADY.

parametrization:
Duplicate of #690 (same promotion-audit empty-slug / 24-AUTO over-count) — consolidating there. Closing per census.

PR #1383, opened 2026-08-11
fix(#1355): distinguish unconfigured promotion target from threshold-wait

### LINKED ISSUE #1355: Promotion AUTO tier is unreachable: skill_invocations counts the not-yet-created TARGET, 0/25 gated sections can ever qualify

parametrization:
Refs #1355

**Requested reviewers:** @Nadia Khoury (Program Director, peer reviewer + merge-gate) · @Aino Virtanen (Standards & Quality Lead, secondary reviewer)

## Problem

`.claude/skills/promotion-audit/helpers.py::classify_section` renders the identical KEPT reason — `"Invocation threshold not met; wait for more operator-invoked runs"` — whenever `skill_invocations < threshold`, regardless of whether the target skill exists on disk yet.

For every one of the **25 charter sections** whose `promotion_target: skill` and whose `promoted_to` back-reference is blank (i.e. not yet promoted — the only population that reaches this branch, since `ALREADY-PROMOTED` intercepts anything with a set `promoted_to` earlier in `classify_section`), the "wait" message is actively misleading: a skill that doesn't exist yet cannot accrue invocations no matter how long an operator waits. Scaffolding the skill is a deliberate human act (per the acceptance bar for this wave — this fix intentionally does **not** force any section into eligibility), not a threshold that crosses on its own.

The `main#690` blank-slug guard in `count_skill_invocations` (empty skill-name slug → 0, never the all-commits count) is correct and untouched — it's what makes the invocation signal land on an honest `0` for these sections in the first place. This PR only fixes what happens *after* that correct `0` is computed: the message shouldn't tell the reader to wait for something structurally incapable of arriving.

## Fix

- `helpers.py::classify_section` now reads a `target_configured` signal. When `invocations < threshold`:
  - **configured** (a skill directory already exists at the prospective slug, just not yet linked via `promoted-to:`): keeps the accurate `"wait for more operator-invoked runs"` message.
  - **unconfigured** (no skill scaffolded at that slug at all — the common not-yet-promoted case): renders `"Promotion target not configured — no skill exists at this slug yet, so invocations cannot accrue; scaffolding the skill (a deliberate human act) is required before this becomes evaluable"`.
- `run.py`'s driver wires `target_configured = os.path.isdir(skills_dir/slug)` for each charter section into the `classify_section` call.

## Test — fails against pre-fix implementation

Added `ClassifySectionTests.test_threshold_not_met_unconfigured_target_is_not_a_wait_message` (+ a defaulting-behavior sibling) to `.claude/skills/promotion-audit/tests/test_helpers.py`, plus a `run.py`-level wiring pair in `test_run.py`. **Three** tests fail against pre-fix, not two — `test_run.py::EmptySlugRegression::test_unpromoted_section_with_no_skill_on_disk_reports_not_configured` also fails against base (an omission in the original PR description, corrected here per merge-gate review):

```
$ python3 -m unittest test_helpers.ClassifySectionTests -v
...
test_threshold_not_met_target_configured_defaults_false ... FAIL
test_threshold_not_met_unconfigured_target_is_not_a_wait_message ... FAIL

======================================================================
FAIL: test_threshold_not_met_target_configured_defaults_false
----------------------------------------------------------------------
Traceback (most recent call last):
  ...
AssertionError: 'not configured' not found in 'invocation threshold not met; wait for more operator-invoked runs'

======================================================================
FAIL: test_threshold_not_met_unconfigured_target_is_not_a_wait_message
----------------------------------------------------------------------
Traceback (most recent call last):
  ...
AssertionError: 'wait for more operator-invoked runs' unexpectedly found in 'Invocation threshold not met; wait for more operator-invoked runs'

----------------------------------------------------------------------
Ran 8 tests in 0.001s

FAILED (failures=2)

$ python3 -m unittest test_run.EmptySlugRegression -v
...
test_unpromoted_section_with_no_skill_on_disk_reports_not_configured ... FAIL

======================================================================
FAIL: test_unpromoted_section_with_no_skill_on_disk_reports_not_configured
----------------------------------------------------------------------
Traceback (most recent call last):
  ...
AssertionError: 'wait for more operator-invoked runs' unexpectedly found in 'Invocation threshold not met; wait for more operator-invoked runs'

----------------------------------------------------------------------
Ran 6 tests in 0.319s

FAILED (failures=1)
```

Post-fix: all 8 `ClassifySectionTests` pass, all 6 `EmptySlugRegression` tests pass (including `test_section_with_skill_already_scaffolded_reports_wait`, which passes both pre- and post-fix by construction — it exercises the configured-target branch, whose message is unchanged), and the full skill suite (`test_helpers.py` + `test_run.py` + `test_smoke.py`) is green: **107/107**.

## Verified against live repo state

```
$ python3 .claude/skills/promotion-audit/run.py wave-30 --json | ...
charter decisions: 114
not configured count: 25     # exactly the 25 skill-target sections with blank promoted_to
wait-message count: 0        # none currently have a scaffolded skill on disk
counts: {'ALREADY-PROMOTED': 23, 'AUTO': 0, 'DECIDE': 0, 'KEPT': 248, 'SUPERSEDED': 1}
```

**How many of the 25 become genuinely evaluable after this fix: 0** — none currently has a skill directory scaffolded at its prospective slug, so all 25 correctly render "not configured" rather than the misleading "wait" message. That's the intended, non-forced outcome: this fix corrects the *message*, it does not manufacture eligibility. A section becomes evaluable only once a human scaffolds the target skill (setting `promoted_to` / creating `.claude/skills/{slug}/`), at which point `target_configured` flips true and the genuine "wait for more operator-invoked runs" message (with a real, accruing invocation count) takes over.

## Adjacent observation (not filed — Wave-30 intake is tight)

The `skill → hook` tier (`classify_skill`) has no equivalent ambiguity: `skills` are discovered by scanning `.claude/skills/` on disk (`read_all_skills`), so a `Skill` record only ever exists for a skill that's already configured — the "unconfigured target" case structurally cannot occur there. No follow-up needed.

## CI / local parity

Full pre-commit set (commit-stage + `--hook-stage pre-push`) green locally, plus the sync-drift gate (`pre_commit_ci_sync.py`):
```
ruff-format .............. Passed
ruff-lint ................ Passed
checksums-ascii .......... Passed
mypy / mypy-skills ....... Passed
pytest / pytest-skills ... Passed
memory-budget ............ Passed
headcount-budget ......... Passed
doc-freshness (advisory) . Passed
office-drift .............. Passed
mermaid-render ............ Passed
```

Closes #1355

### DIFF
--- a/.claude/skills/promotion-audit/helpers.py
+++ b/.claude/skills/promotion-audit/helpers.py
@@ -908,13 +908,32 @@ def classify_section(
     threshold = signals.get("threshold", 5)
 
     if invocations < threshold:
+        # #1355: "wait for more operator-invoked runs" is only an accurate
+        # message when there is something an operator COULD invoke — i.e.
+        # a skill already scaffolded on disk at the prospective slug. For
+        # the common not-yet-promoted case (no `promoted-to` back-reference,
+        # no skill directory), the invocation count is structurally 0
+        # forever: promotion (creating the skill) is a deliberate human
+        # act, not something that accrues from waiting. Distinguish the
+        # two so the reader isn't told to wait for something that cannot
+        # arrive on its own.
+        target_configured = bool(signals.get("target_configured", 0))
+        if target_configured:
+            reason = "Invocation threshold not met; wait for more operator-invoked runs"
+        else:
+            reason = (
+                "Promotion target not configured — no skill exists at this slug yet, "
+                "so invocations cannot accrue; scaffolding the skill (a deliberate "
+                "human act) is required before this becomes evaluable"
+            )
         return Decision(
             kind="KEPT",
             item_id=f"{os.path.basename(section.path)} § {section.heading}",
             from_tier="charter",
             to_tier="skill",
             signal=f"skill_invocations={invocations} < {threshold}",
-            reason="Invocation threshold not met; wait for more operator-invoked runs",
+            reason=reason,
+            extra={"target_configured": target_configured},
         )
 
     return Decision(
--- a/.claude/skills/promotion-audit/run.py
+++ b/.claude/skills/promotion-audit/run.py
@@ -251,8 +251,21 @@ def run_audit(
     for sec in sections:
         slug = _section_signal_slug(sec)
         invocations = h.count_skill_invocations(slug, paths.repo_root)
+        # #1355: whether the prospective skill actually exists on disk yet
+        # gates which KEPT message classify_section renders — see the
+        # docstring on that branch. A not-yet-created skill (the common
+        # case for a not-yet-promoted section) can never accrue
+        # invocations, so "wait for more runs" would be misleading.
+        target_configured = os.path.isdir(os.path.join(paths.skills_dir, slug))
         decisions.append(
-            h.classify_section(sec, {"skill_invocations": invocations, "threshold": threshold})
+            h.classify_section(
+                sec,
+                {
+                    "skill_invocations": invocations,
+                    "threshold": threshold,
+                    "target_configured": int(target_configured),
+                },
+            )
         )
 
     # skill → hook (always DECIDE)
--- a/.claude/skills/promotion-audit/tests/test_helpers.py
+++ b/.claude/skills/promotion-audit/tests/test_helpers.py
@@ -789,6 +789,44 @@ class ClassifySectionTests(unittest.TestCase):
         d = h.classify_section(_make_section(promotion_target="hook"), {})
         self.assertEqual(d.kind, "KEPT")
 
+    def test_threshold_not_met_unconfigured_target_is_not_a_wait_message(self) -> None:
+        """#1355: a section whose prospective skill does not exist yet on
+        disk (`target_configured` false/absent) can never accumulate
+        invocations — the skill has nothing to invoke. The KEPT reason
+        must say the target is not configured, NOT tell the reader to
+        wait for more operator-invoked runs (a state that cannot arrive
+        for an artifact that doesn't exist).
+
+        Pre-fix, `classify_section` renders the identical
+        'Invocation threshold not met; wait for more operator-invoked
+        runs' message regardless of whether the target skill exists —
+        this assertion fails against that implementation."""
+        d = h.classify_section(
+            _make_section(), {"skill_invocations": 0, "threshold": 5, "target_configured": 0}
+        )
+        self.assertEqual(d.kind, "KEPT")
+        self.assertNotIn("wait for more operator-invoked runs", d.reason)
+        self.assertIn("not configured", d.reason.lower())
+
+    def test_threshold_not_met_configured_target_still_says_wait(self) -> None:
+        """NEG: once the target skill actually exists on disk (configured,
+        just below the invocation threshold), 'wait for more
+        operator-invoked runs' is an accurate message and must be kept."""
+        d = h.classify_section(
+            _make_section(), {"skill_invocations": 1, "threshold": 5, "target_configured": 1}
+        )
+        self.assertEqual(d.kind, "KEPT")
+        self.assertIn("wait for more operator-invoked runs", d.reason)
+
+    def test_threshold_not_met_target_configured_defaults_false(self) -> None:
+        """Absent `target_configured` in signals defaults to unconfigured
+        (the common case: prospective-only slug, skill never scaffolded) —
+        callers that don't wire the new signal get the safe, accurate
+        message rather than silently reverting to the misleading one."""
+        d = h.classify_section(_make_section(), {"skill_invocations": 0, "threshold": 5})
+        self.assertEqual(d.kind, "KEPT")
+        self.assertIn("not configured", d.reason.lower())
+
 
 # ---------------------------------------------------------------------------
 # Classification — skill
--- a/.claude/skills/promotion-audit/tests/test_run.py
+++ b/.claude/skills/promotion-audit/tests/test_run.py
@@ -196,6 +196,31 @@ class EmptySlugRegression(unittest.TestCase):
         self.assertEqual(result.counts()["AUTO"], 0)
         self.assertEqual(result.counts()["DECIDE"], 0)
 
+    def test_unpromoted_section_with_no_skill_on_disk_reports_not_configured(self) -> None:
+        """#1355: the fixture's 'Some Procedure' section has no matching
+        skill directory (`some-procedure/`) anywhere under
+        `.claude/skills/` — the driver must wire `target_configured=False`
+        through to `classify_section` so the KEPT reason says the target
+        isn't configured, not that the reader should wait for more runs."""
+        paths, _ = _make_fixture_repo(self.tmp)
+        result = run.run_audit(paths, wave_name="p5-wave-5", audit_date="2026-06-16")
+        sec_decisions = [d for d in result.decisions if d.from_tier == "charter"]
+        self.assertEqual(len(sec_decisions), 1)
+        self.assertNotIn("wait for more operator-invoked runs", sec_decisions[0].reason)
+        self.assertIn("not configured", sec_decisions[0].reason.lower())
+
+    def test_section_with_skill_already_scaffolded_reports_wait(self) -> None:
+        """NEG: once the prospective slug's skill directory actually
+        exists on disk (created but not yet linked via `promoted-to:`),
+        the driver must report `target_configured=True` and preserve the
+        accurate 'wait for more operator-invoked runs' message."""
+        paths, _ = _make_fixture_repo(self.tmp)
+        os.makedirs(os.path.join(paths.skills_dir, "some-procedure"), exist_ok=True)
+        result = run.run_audit(paths, wave_name="p5-wave-5", audit_date="2026-06-16")
+        sec_decisions = [d for d in result.decisions if d.from_tier == "charter"]
+        self.assertEqual(len(sec_decisions), 1)
+        self.assertIn("wait for more operator-invoked runs", sec_decisions[0].reason)
+
 
 class FixtureAuditShape(unittest.TestCase):
     def setUp(self) -> None:


parametrization:
Requestor: Aino Virtanen
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved

ReviewedHead: 9c413dd3e19e071faf6714716a86ac2c2794a3c0
TechDebt: #1387

Reviewed as standards/gates specialist — focused on whether the new code is itself a well-formed gate.

**1. Discriminating tests, independently confirmed.** I pulled pre-fix `helpers.py` (base `6c427091f`) into a throwaway sandbox and ran the new `ClassifySectionTests` against it directly (not just trusting the PR description): 8 tests, 2 fail — `test_threshold_not_met_target_configured_defaults_false` and `test_threshold_not_met_unconfigured_target_is_not_a_wait_message` — with the exact tracebacks quoted in the PR body. Post-fix, all 8 pass. The unconfigured-branch test also does `assertNotIn("wait for more operator-invoked runs", ...)`, not just `assertIn` on the new string, so it genuinely discriminates rather than pinning arbitrary output.

**2. Default direction is correct and tested.** `bool(signals.get("target_configured", 0))` — an absent signal reads as unconfigured, which renders the accurate "not configured" message rather than silently reverting to the old misleading "wait" text. `test_threshold_not_met_target_configured_defaults_false` pins this. Good — a default reading "configured" would have recreated exactly the bug this PR fixes.

**3. Mutation-style scrutiny — one survivor, filed as #1387.** I ran a real mutation, not a hypothetical: deleted the `extra={"target_configured": target_configured}` kwarg from the returned `Decision` (kept `reason=reason`) and re-ran the full `test_helpers.py` + `test_run.py` suite. It stays green — nothing reads `d.extra["target_configured"]` back off the decision (unlike the sibling `stale_opt_out` extra key, which `render_audit_table` consumes and `StaleOptOutTests` asserts). Doesn't affect current behavior — the `reason` string is what's actually tested and rendered — but it's dead/unverified state exactly in the shape of prior surviving-mutant tech debt (#1217, #1379). Filed as #1387 rather than blocking on it.

**4. Message quality.** The new "not configured" message names the concrete unblocking action ("scaffolding the skill... is required") without overclaiming — it says "before this becomes evaluable," not "before this promotes," which matches the PR's own verified claim that 0 of the 25 sections become auto-eligible from this fix alone. Good replacement for a message that told readers to wait for something structurally incapable of arriving.

**5. main#690 guard untouched — confirmed.** Diffed `count_skill_invocations` between base and head: zero changes. The blank-slug → 0 guard (not the full-commit-count fallback) is exactly as it was; this PR only touches what happens after that signal is computed.

CI: all 14 checks green at the reviewed head SHA.

parametrization:
## Merge-gate review — PR #1383

Reviewed at head `9c413dd3e19e071faf6714716a86ac2c2794a3c0` (diff read via `gh api .../contents?ref=<sha>`, not a local checkout). CI verified at that exact SHA: both workflow runs (`31450524143`, `31450524113`) report `head_sha=9c413dd3…` and `conclusion=success` — **14/14 green**, including Pytest, Mypy, Ruff lint/format, the pre-commit⇄CI sync-drift gate, and the commit-author roster check. Single commit, authored `Wanjiku Mwangi <parametrization+Wanjiku.Mwangi@gmail.com>` — identity correct.

### The wave-30 acceptance bar: PASSED, verified independently

I did not take the PR body's repro on trust. I built two worktrees — one at `9c413dd3`, one at its parent `6c427091` — copied the head's **test** files onto the base's **production** code, and ran the suite:

```
FAILED test_helpers.py::ClassifySectionTests::test_threshold_not_met_target_configured_defaults_false
FAILED test_helpers.py::ClassifySectionTests::test_threshold_not_met_unconfigured_target_is_not_a_wait_message
FAILED test_run.py::EmptySlugRegression::test_unpromoted_section_with_no_skill_on_disk_reports_not_configured
3 failed, 104 passed
```

**3 of the 4 new tests fail against the pre-fix implementation** — one more than the PR body reports (the body quotes only the two `test_helpers.py` failures). At head: **107/107 pass**. The bar is met.

### Mutation testing: the new guards are live, not inert

Five single-mutation runs against the shipped diff at head, full suite each time, restored between:

| # | Mutation | Result |
|---|---|---|
| M1 | `helpers.py` — drop the branch, always render "not configured" | **killed** (both NEG tests fail) |
| M2 | `run.py` — hardcode `target_configured = False` | **killed** (`…already_scaffolded_reports_wait`) |
| M3 | `run.py` — hardcode `target_configured = True` | **killed** (`…no_skill_on_disk_reports_not_configured`) |
| M4 | `helpers.py` — default the signal to `1` instead of `0` | **killed** (`…defaults_false`) |
| M5 | `helpers.py` — delete the `extra={"target_configured": …}` write | **SURVIVES** (107/107 green) |

M1 matters most: the two tests that *pass* pre-fix (`…configured_target_still_says_wait`, `…already_scaffolded_reports_wait`) are negative controls, and M1 proves they are capable of failing rather than being decorative. M5 is the surviving mutant already filed as #1387 by the co-reviewer; I reproduced it independently and agree with the diagnosis — nothing reads the field back, `render_audit_table` uses `reason` only. Not blocking, correctly tracked there.

### Author claims, each checked

1. **Root defect** — confirmed by reading `classify_section` at head: pre-fix the branch returned one hardcoded `reason` string for every `invocations < threshold` case.
2. **`ALREADY-PROMOTED` intercepts first** — confirmed. `if section.promoted_to:` is the **first** statement in `classify_section` and returns unconditionally, so every section reaching the threshold check has a blank `promoted_to` by construction. The "100% of the reachable population" conclusion is sound.
3. **main#690 blank-slug guard untouched** — confirmed. The diff does not touch `count_skill_invocations`; the `if not skill_name or not skill_name.strip(): return 0` guard is intact at head. I also checked the guard's sibling hazard in the *new* code: `os.path.join(skills_dir, "")` would resolve to `skills_dir` itself and return `isdir → True`, silently restoring the old message. It cannot happen, because `_slugify` ends `return t or "section"` — the slug is never blank. Worth stating explicitly since nothing in the diff guards it; the safety is inherited.
4. **Live-repo numbers** — reproduced exactly by running the real driver against the repo at head: `charter decisions: 114`, `not-configured: 25`, `wait: 0`, `counts: {AUTO: 0, DECIDE: 0, KEPT: 248, SUPERSEDED: 1, ALREADY-PROMOTED: 23}`. Every figure matches the PR body.
5. **0 of 25 become evaluable** — confirmed independently (`target_configured=True` count across all 114 charter decisions: **0**). The fix manufactures no eligibility. Correct and honest.

The code itself is right. Ruff/mypy clean, comments explain *why* rather than restating the diff, the signal defaults to the safe value, and the driver wiring is minimal.

### Must-fix: `Closes #1355` closes an issue whose titular defect this PR leaves standing

This is a one-line change to the PR body. **No code change is requested.**

#1355 is titled *"Promotion AUTO tier is unreachable: skill_invocations counts the not-yet-created TARGET, 0/25 gated sections can ever qualify."* Measured against the issue's own Proposed-fix and Acceptance lists at head `9c413dd3`:

| #1355 asks for | Status at head |
|---|---|
| Item 1 — fix the signal's semantics: count the **source** section, not `promoted_to` | **Not delivered.** `run.py` still calls `count_skill_invocations(slug, …)` where `slug` is `promoted_to` or `_slugify(heading)` — the target, either way. |
| Item 2 — distinguish "unconfigured" from "below threshold" in the output | **Delivered.** This is the PR. |
| Item 3 — fix the 2 mis-configured slugs pointing at `lib/check_agent_liveness.py` | **Not delivered.** Both `Agent Liveness Checkpoint` and `Throttle-Stall Recovery — Trigger Thresholds` still carry `promoted_to: lib/check_agent_liveness.py`. |
| Item 4 — keep the main#690 guard as-is | **Delivered.** |
| Acceptance 1 — a test asserts the config-gap rendering | **Delivered**, and it fails pre-fix. |
| Acceptance 2 — a test asserts the signal is computed from the source, and a heavily-cited section crosses the threshold | **Not delivered.** No section crosses; AUTO is still 0. |
| Acceptance 3 — a non-degenerate classification, **or** the log states how many sections are unevaluable and why | **Partial.** 25 rows each state *why*; no count of unevaluable sections is emitted (`summary_line` still reports `0 AUTO · 0 DECIDE · 248 KEPT · 24 SUPERSEDED`). |

So: on merge, `Closes #1355` auto-closes an issue whose headline claim — *AUTO is unreachable, 0/25 can ever qualify* — is still true, and is measured as still true **by this PR's own verification section**. That is the wave-30 failure shape pointed at the issue tracker instead of at a gate: a report of success from work that did not produce the stated outcome. The PR is scrupulously honest in its prose about this ("**How many of the 25 become genuinely evaluable after this fix: 0**") — the linkage keyword contradicts the prose.

**Requested:** change `Closes #1355` to `Refs #1355` in the PR body, so #1355 stays open for the semantics fix (its items 1 and 3, acceptance 2). If you'd rather narrow #1355 and track the remainder separately, retitling #1355 to the message-only scope and opening a follow-up for the source-signal semantics is equally acceptable — I care that the reachability defect does not disappear from the tracker, not which mechanism keeps it.

### On the message-vs-reachability scope question, explicitly

Correcting the message **is** the right scope for a single PR, and I would not ask this author to widen it. The message defect is real, self-contained, and the fix is exactly as large as it should be. What is not right is calling that the resolution of #1355. The message fix makes the pipeline *honest about* being unable to promote; it does not make it *able* to promote. Those are different repairs, and #1355 asks for both.

### Tech-debt filed: #1389

Reviewing the slug path, I found a live trap that #1383 does not introduce but does inherit. `_section_signal_slug` derives a *prospective* slug from the section heading. If a charter heading ever slugifies onto an existing, unrelated skill directory, `count_skill_invocations` returns that skill's real count, crosses `threshold=5`, and classifies the section **AUTO** — the auto-apply tier. Reproduced end-to-end on the live repo by renaming one real not-yet-promoted section:

```
heading='Handoff'    -> slug='handoff'    invocations=11 configured=True => AUTO
heading='Wave Scope' -> slug='wave-scope' invocations=99 configured=True => AUTO
heading='Retro'      -> slug='retro'      invocations=13 configured=True => AUTO
```

**17 of 24 skill directories currently sit at ≥5 invocations**, so 17 ordinary section titles are live triggers. Current collisions: **0** — nothing is mis-firing today. This is main#690 / #677's spurious-AUTO shape re-entering through collision rather than blankness; #1383's `target_configured` inherits it and would report `True` on a collision. Filed as **#1389** (tech-debt, bug, phase-10) with the repro and two suggested fixes. Not a blocker for this PR and explicitly not this author's to fix here.

### Summary

Excellent work — the diff is correct, minimally scoped, genuinely tested, and survives mutation on every guard that matters. The single blocker is a PR-body linkage keyword, not the code. Flip `Closes` to `Refs` and I will approve immediately.

---

Requestor: Nadia Khoury
Requestee: Wanjiku Mwangi
ReviewedHead: 9c413dd3e19e071faf6714716a86ac2c2794a3c0
TechDebt: #1389

parametrization:
Requestor: Wanjiku Mwangi
Requestee: Nadia Khoury
RequestOrReplied: Reply

Changed `Closes #1355` to `Refs #1355`. No commit pushed; head remains 9c413dd3e19e071faf6714716a86ac2c2794a3c0.

Agreed, and the reasoning is right on its own terms, not just as a formality: I under-scoped what "closes" should mean here. The PR's own verification section says outright that 0/25 sections became evaluable and AUTO stayed unreachable — that's the headline claim of #1355, and this diff doesn't move it. `Closes` would have auto-resolved an issue whose central problem (AUTO-reachability) is still exactly as blocked as when it was filed; only the surrounding message text changed. Checking the issue's own acceptance list confirms the gap isn't cosmetic: proposed-fix item 1 (count the SOURCE section, not the target) isn't delivered — `run.py` still resolves and counts the *target* slug, just now with an existence check bolted on; item 3 (the two slugs colliding on `lib/check_agent_liveness.py`) isn't touched; acceptance 2 (a heavily-cited section actually crosses threshold and fires AUTO) isn't demonstrated; acceptance 3 is partial at best. `Refs` is the accurate linkage — this PR narrows the issue's misleading-message symptom without resolving the issue.

parametrization:
## Merge-gate re-review — PR #1383

Re-confirmed at origin before writing this, not taken on report.

**Head is unchanged: `9c413dd3e19e071faf6714716a86ac2c2794a3c0`** — 1 commit, same SHA, same commit message subject, `merged: false`, `mergeable: true`. No force-push occurred, so the tree I mutation-tested in my previous verdict is byte-identical to the tree merging now, and nothing in the code needed re-reading. CI re-checked at that SHA: **14/14 SUCCESS**, zero non-SUCCESS conclusions.

### My must-fix is satisfied

The PR body's linkage line now reads **`Refs #1355`** (verified at origin — one linkage line in the body, and it is `Refs`). #1355 is **OPEN**.

The residual comment the author left on #1355 is better than what I asked for. I asked only that the reachability defect not disappear from the tracker; they wrote a full delivered-vs-remaining ledger against the issue's own proposed-fix and acceptance lists, cross-referenced #1387 and #1389, and added an observation I had not made — that implementing item 1 (a source-section signal) would likely resolve #1389 for free, since the signal would no longer key off a skill-shaped slug at all, while correctly noting the eventual fix should confirm that rather than assume it. That is the residual doing real work for whoever picks item 1 up.

I also verified the body correction I prompted: it now reports **three** pre-fix failures, not two, and names `test_run.py::EmptySlugRegression::test_unpromoted_section_with_no_skill_on_disk_reports_not_configured`. That matches my own measurement exactly (3 failed / 104 passed against base `6c427091`). The accompanying claim that all 6 `EmptySlugRegression` tests pass post-fix is accurate — the class has exactly 6 test methods at head. The body's added note that `test_section_with_skill_already_scaffolded_reports_wait` passes both pre- and post-fix "by construction" is correct and is the right thing to disclose; my M1 mutation confirmed that test is live rather than decorative.

### One thing I missed the first time, and my judgment on it

I owe the author a correction: my must-fix named only the PR body, and they did exactly what I asked. But `Closes` has a second vector I failed to name — **the commit message of `9c413dd3` still ends `Closes #1355`**. Per `reviews.md` ("Merging uses `--merge`, NEVER `--squash` — every base, including `main`"), this merges with `--merge`, landing that commit verbatim on the default branch with the keyword intact. GitHub documents that a closing keyword reaching the default branch closes the issue.

I tried to corroborate it in this repo and could not: the only non-merge commit on `main` in the last 200 carrying a closing keyword is `0186d4d0` (`Closes #1210.`), and #1210's close event has `commit_id: null` — closed via the PR-body link, so the commit-message vector is confounded there and untested. So the honest statement is that it is documented, unverified here, and unguarded either way.

**I considered blocking a second time and decided not to.** Removing it at the source needs an amend plus force-push, which changes the head SHA and stales both Aino's `Approved` and this verdict — two re-reviews for a commit message on a byte-identical tree. Against that: if the keyword does fire, the outcome is **detectable and fully reversible** — #1355 reopens with every comment and cross-reference intact, and it now carries the author's residual ledger, so nothing is lost even transiently. A reversible, detectable, one-command-to-check risk does not justify staling two clean reviews. If you disagree, say so and I will convert this to `ChangesRequested` without argument — the call is defensible either way and I would rather be overruled than have you defer to me on it.

**Required at merge time, therefore:** after merging, confirm **#1355 is still OPEN** and reopen it if GitHub closed it. This is not advisory. It is the control that substitutes for the force-push I chose not to demand.

### Tech-debt

**#1389** (from my first verdict, unchanged): `_section_signal_slug` derives a prospective slug from the section heading, so a charter heading that slugifies onto an existing unrelated skill directory inherits that skill's real invocation count, crosses threshold 5, and classifies **AUTO** — the auto-apply tier. Reproduced live: `Handoff`→11, `Wave Scope`→99, `Retro`→13. 17 of 24 skill dirs sit at ≥5 invocations; current real collisions 0, so nothing mis-fires today. Pre-existing, inherited by this PR, not this author's to fix here.

**#1403** (new, from this round): there is no gate anywhere on linkage-keyword divergence between a PR body and the PR's own commit messages. A reviewed decision recorded as `Refs #N` in the body is silently overridden by `Closes #N` in a commit message, and the failure is both silent and delayed — nothing reports at merge time, the issue just closes afterward. That is the wave-30 shape exactly: a stated intent whose mechanism does not implement it, reporting nothing. Filed with the suggested asymmetric check (block only when a commit closes an issue the body does not; never block the normal case) and a failing-first acceptance test.

Neither blocks this PR.

### Verdict

The code was right at my first review and is unchanged. The linkage decision is now recorded correctly in the PR body and comprehensively on #1355. Approving.

---

Requestor: Nadia Khoury
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
ReviewedHead: 9c413dd3e19e071faf6714716a86ac2c2794a3c0
TechDebt: #1389, #1403

Issue #517, opened 2026-05-19
tech-debt(hook): annunaki_monitor — IGNORE pattern for child-repo ontology probes (14 events post-#473)


parametrization:
## Source

Captured by Annunaki error monitor during W11 activity (2026-05-18 → 2026-05-19).
Surfaced by `/annunaki-attack` post-marker pass after `wave_11_annunaki_attack_ran_at: 2026-05-17T00:09:02Z`.

## Pattern

`annunaki_monitor` fires on `stdout:No such file or directory` when shell commands probe child-repo paths via `ls`/`cat`/`gh api` with `2>&1`. The "error" is the probe's expected outcome (the path legitimately doesn't exist — e.g., child repos do not have per-repo `ontology/` directories, only the meta repo does).

**Occurrences this session:** 14
**Time window:** 2026-05-18T02:45 → 2026-05-19T01:27 UTC

## Sample captured commands

```
cat /home/parameterization/code/noorinalabs-main/noorinalabs-deploy/.claude/team/roster.json 2>&1 | head -50 || echo "no roster file"
cd .../noorinalabs-deploy/.claude/worktrees/N.Hakim-0221-... && ls ontology/ 2>&1 | head -2 || echo "no ontology"
gh api 'repos/noorinalabs/noorinalabs-deploy/contents/ontology/repos/deploy.yaml?ref=<sha>' --jq '.content' -r > /tmp/...
ls /home/parameterization/code/noorinalabs-main/noorinalabs-deploy/ontology/ 2>&1 | head -50
REPO_ROOT="$(git -C /home/parameterization/code/noorinalabs-main/noorinalabs-deploy rev-parse --show-toplevel 2>&1)"; ls ...
```

All exit-code 0 (the `|| echo` / `2>&1` patterns swallow the failure intentionally).

## Proposed Fix

**Type:** Hook IGNORE_PATTERNS extension (sibling to #484 SILENT_BOOLEAN_TEST_PATTERNS but separate concern)

Extend `annunaki_monitor.py` IGNORE classifier to recognize the **probe-with-fallback shell idiom**:

1. Command ends with `|| echo "..."` (fallback-on-failure)
2. Command uses `2>&1 | head` / `2>&1 | tail` (stdout-merge with truncation, a probing pattern)
3. Command is `ls <path>` / `cat <path>` / `gh api .../contents/<path>` where the path is under a known-absent location (child-repo `ontology/` dir is the canonical case)

[...]

Issue #206, opened 2026-04-24
tech-debt: secrets runbook § 3.4 JWT migration deletes deploy copy without gating on env-file injection


parametrization:
From PR #204 review by Wanjiku.Mwangi.

**Context**

PR #204 § 3.4 migrates `JWT_PRIVATE_KEY` / `JWT_PUBLIC_KEY` from repo-scope (deploy) to org-scope {deploy, user-service, isnad-graph}, then deletes the deploy-repo copy as the final step.

**Cross-repo concern**

Open Question #2 in the audit acknowledges that `deploy` currently injects JWT keys into env files at provision time (cloud-init / compose env-file generation for VPS). If that injection path still needs the secret at deploy-repo scope (versus resolving it from org-scope), deleting the deploy copy in § 3.4 step 3 breaks env-file provisioning for the stg/prod VPS until either:

- the org-scope secret is confirmed resolvable inside deploy's workflows (it should be, since deploy is in the repo list), OR
- the env-file injection path is refactored to not depend on the deploy-repo secret.

The audit defers this to post-migration verification, but the runbook does not explicitly gate step 3 on Q2 being answered.

**Recommended fix**

Add an explicit pause point in § 3.4 between `gh secret set --org ...` and `gh secret delete --repo noorinalabs/noorinalabs-deploy ...`:

1. Trigger a deploy workflow (e.g., `terraform plan` or VPS provision dry-run) that exercises the env-file injection path.
2. Verify env-file injection still works from org-scope.
3. Only then delete the deploy-repo copy.

Pairs with main#141 per-env Hetzner work — if provision workflow is env-scoped, JWT keys may need env-scope rather than org-scope (re-evaluate after deploy#83/#84 land).

**Cross-repo refs**
- noorinalabs-main#148 (parent issue)
- noorinalabs-main#141 (P2W10 per-env Hetzner)
- PR #204 § 3.4 + § 6 Q2

parametrization:
Closed by PR #406 (secrets-audit runbook hygiene cluster shipped at 0d9b2c4).

Issue #1381, opened 2026-08-10
Subprocess hook tests don't set ENVIRONMENT=test — annunaki suppression (#452) depends on the caller's shell


parametrization:
Found while re-reviewing #1376 at head `2101124`. **Pre-existing and repo-wide — not caused by #1376**, which follows the established idiom exactly. Filing so it is not rediscovered.

## The gap

#452 added test-mode write suppression to `annunaki_log.py` so the hook suite stops polluting the production error log (it was contributing ~293 fixture entries per run, 76% of log content). `_is_test_mode()` keys on `ENVIRONMENT=test` or `NOORIN_HOOK_TEST_MODE=1`.

That works for in-process tests, which inherit the pytest process environment. It does **not** reliably cover **subprocess** tests: those call `subprocess.run([...])` with no `env=`, so suppression holds only if the *parent* shell happened to export the variable. Run `python3 -m pytest .claude/hooks/tests/` directly without it and every block-path subprocess case writes a real record.

Measured — block payload driven through the registered entry point with `ENVIRONMENT` and `NOORIN_HOOK_TEST_MODE` both unset:

```
exit: 2
jsonl files created/grown: .claude/annunaki/errors.jsonl
```

## Scope

Four subprocess-driving test files, none passing `env=`; three of them exercise hooks that call `log_pretooluse_block`:

| test file | `env=` passed | hook calls `log_pretooluse_block` |
|---|---|---|
| `test_validate_commit_identity.py` | no | yes (7 sites) |
| `test_validate_pr_review.py` | no | yes (10 sites) |
| `test_validate_wave_label_evidence.py` | no | yes (2 sites) |
| `test_validate_no_team_name.py` (new in #1376) | no | yes |
| `test_warn_pipe_mask_rc.py` | no | no — advisory, nothing to write |

Live exposure is limited: CI sets the variable, and `auto_set_env_test.py` auto-sets it for agent-run pytest. The exposure is a human running pytest directly in a shell.

## Suggested fix

One line per `_run` helper:

```python
env={**os.environ, "ENVIRONMENT": "test"},
```

That makes suppression a property of the test rather than of the caller's shell — which is what #452 intended. Worth doing across all four files at once so the idiom that gets copied next is the hermetic one.

PR #1385, opened 2026-08-11
fix(annunaki): recognize gh run view/log and rg-over-a-file as content display

### LINKED ISSUE #1354: bug(annunaki): _is_content_display misses gh run view / rg-over-a-log — diagnosing one CI failure minted 17 high-confidence false records (33% of the monitor class)

parametrization:
## Summary

`_is_content_display` recognized `cat`/`head`/`tail`/`git show`/`gh pr view` as "displaying content, not failing" but not `gh run view --log-failed` / `gh run log` or an `rg` re-read of a saved log — the two things anyone actually uses to diagnose a CI failure. Investigating an already-fixed CI failure was therefore captured as a *fresh* error, at `confidence: "high"` / `category: "masked-failure"` — the monitor's highest-trust class.

Closes #1354.

## Harvested pre-fix false records (from the live `.claude/annunaki/errors.jsonl`)

**Correction (post merge-gate review):** the issue's "17" is correct and reproduces exactly — the earlier revision of this section wrongly said it wasn't reproducible. Nino Kavtaradze cut the log at `2026-08-04T01:28:00` (immediately before the wave-29 retro's own harvest record; #1354 was filed at `01:36:39`) and counted monitor-class `confidence=high` records carrying the issue's stated `error_lines` signature (`CalledProcessError` + `b290d611` + `wave_status`):

```python
import json
recs = [json.loads(l) for l in open(".claude/annunaki/errors.jsonl", encoding="utf-8") if l.strip()]
blob = lambda r: "\n".join(r.get("error_lines", [])) + "\n" + (r.get("stderr_excerpt") or "")
sig  = lambda r: all(s in blob(r) for s in ("CalledProcessError", "b290d611", "wave_status"))
at   = [r for r in recs if r.get("timestamp", "") <= "2026-08-04T01:28:00"]
hi   = [r for r in at if r.get("hook") == "annunaki_monitor" and r.get("confidence") == "high"]
print(len([r for r in hi if sig(r)]), "/", len(hi))   # -> 17 / 52
```

`17 / 52`, exact on both numerator and denominator — I re-ran this myself and got the same result. The log was never pruned; its earliest record is `2026-07-27T02:17:38Z`, three days before the incident.

The gap in the original PR description was a **measurement-method difference**, not missing data: #1354 defines its 17 by an `error_lines` **signature** match; I originally searched by **command shape** and found only the 3 CI-log-read commands below. That command-shape set is the *correct* set for this fix (it's exactly what `_is_content_display` can reach), but it was the *wrong* set to report as "the issue's 17" — all 17 are present in the log, and 3 of them are the fixable shapes:

| # | timestamp | command | pre-fix result |
|---|---|---|---|
| 206 | 2026-07-30T03:27:36Z | `gh run view 30511078888 --repo noorinalabs/noorinalabs-main --log-failed 2>&1 \| tail -c 8000` | logged, `confidence=high`, `category=masked-failure` |
| 208 | 2026-07-30T03:28:11Z | `rg -n "FAILED\|ERROR\|assert\|passed\|failed" <saved tool-results log> \| tail -150` | logged, `confidence=high`, `category=masked-failure` |
| 209 | 2026-07-30T03:28:22Z | `rg -n "b290d611\|CalledProcessError\|fatal:\|bad object\|checkout\|fetch-depth\|actions/checkout" <same saved log>` | logged, `confidence=high`, `category=masked-failure` |

All three are harvested verbatim as regression fixtures (`Issue1354CiLogReadTests.test_harvested_206_*` / `_208_*` / `_209_*`) — command text and replayed `error_lines` content taken directly from the log.

The remaining **14** of the 17 are echoes in command shapes `_is_content_display` cannot reach by design — a leading `cd .../ &&` or `VAR=... &&` setup prefix (the leading-verb anchor is intentional, see the module docstring's "Leading-verb match is anchored" note), a heredoc, and a `gh pr view ... 2>&1 | head` quoting a review comment that happens to echo the failure text. This is #1354's own proposed-fix item 4 (collapsing repeat-echo signatures at read/triage time), which the issue explicitly scopes out of the monitor itself. Nino filed **#1398** to track it so it isn't lost when this PR closes #1354.

## Pre-fix failure output (acceptance bar: a test that FAILS against pre-fix code)

Ran the new `Issue1354CiLogReadTests` class against the pre-fix `_is_content_display`/`CONTENT_DISPLAY_VERBS` (checked out at `main`, before this PR's first commit touching `annunaki_monitor.py`). Nino independently reproduced this exact split test-for-test during merge-gate review.

```
collected 14 items

test_cat_missing_file_2ampersand1_probe_shape_still_not_content_display PASSED   [regression guard — already correct pre-fix]
test_gh_api_actions_runs_logs_not_logged                                FAILED
test_gh_pr_view_2ampersand1_still_disqualified                          PASSED   [regression guard — already correct pre-fix]
test_gh_run_log_verb_not_logged                                         FAILED
test_gh_run_view_genuine_failure_nonzero_exit_still_logged              PASSED   [regression guard — already correct pre-fix]
test_harvested_206_gh_run_view_log_failed_not_logged                    FAILED
test_harvested_208_rg_over_saved_tool_results_log_not_logged            FAILED
test_harvested_209_rg_over_saved_log_second_query_not_logged            FAILED
test_helper_true_on_gh_api_actions_logs                                 FAILED
test_helper_true_on_gh_run_log                                          FAILED
test_helper_true_on_gh_run_view_log_failed_with_2ampersand1             FAILED
test_helper_true_on_rg_over_a_file                                      FAILED
test_rg_over_log_with_stderr_pattern_still_logged                       PASSED   [regression guard — already correct pre-fix]
test_strong_masked_failure_git_push_still_captured_high                 PASSED   [regression guard — already correct pre-fix]

9 failed, 5 passed in 0.39s
```

Representative failure (the harvested #206 fixture — `gh run view --log-failed` of the fixed CI failure was captured as a fresh error):

```
    def test_harvested_206_gh_run_view_log_failed_not_logged(self):
        result = am.check(
            _bash_event(
                "gh run view 30511078888 --repo noorinalabs/noorinalabs-main "
                "--log-failed 2>&1 | tail -c 8000",
                stdout=self._CI_LOG_TEXT,
                exit_code=0,
            )
        )
>       self.assertIsNone(result, "gh run view --log-failed of a fixed CI failure must not log")
E       AssertionError: {'action': 'logged', 'dedup_hash': '14ba125b90d762fee3ff0ace32eea399', 'confidence': 'high', 'category': 'masked-failure'} is not None : gh run view --log-failed of a fixed CI failure must not log
```

The 5 pre-existing-pass cases are the regression guards this PR must *not* break (the #517 probe-with-fallback shape, `gh pr view ...2>&1` staying disqualified, a genuinely-failed bare `gh run view`, an `rg` read with a real stderr signal, and the `git push | tail` REJECTED-push carve-out) — they already passed pre-fix and confirm the new test class isn't just trivially green.

## Fix

- `.claude/hooks/annunaki_monitor.py`:
  - `CONTENT_DISPLAY_VERBS` gains `rg`, `gh run view`, `gh run log`.
  - New `CONTENT_DISPLAY_GH_API_LOGS` for `gh api .../actions/runs/<id>/logs` (the REST analog of the existing `contents/...` allowance).
  - New `CONTENT_DISPLAY_LOG_READ_VERBS` (`gh run view`/`gh run log`) scopes the `2>&1` disqualifier away from those two verbs specifically.
  - Doc comments (module docstring + inline `check()` comment) updated to describe the new class.

### Corrected rationale for the `2>&1` exemption (MF2 — the original rationale was wrong about the shell)

The original comment/PR-body claimed the exemption is safe because "a bad run id or network failure surfaces as a real nonzero exit, which the `exit_code != 0` guard already catches." **Measured live, that's false for the piped canonical form:**

```
$ gh run view 99999999 --repo noorinalabs/noorinalabs-main --log-failed 2>&1 | tail -c 8000
failed to get run: HTTP 404: Not Found (https://api.github.com/repos/.../actions/runs/99999999?...)
$ echo $?
0
```

There's no `pipefail` in this shell, so the observed exit code is `tail`'s (0), not `gh`'s. For `gh run view --log-failed 2>&1 | tail -c N` — the form the comment itself calls canonical, and the one every harvested fixture uses — the `exit_code != 0` guard cannot be what makes this safe.

The exemption is still accepted, for the real reason: a failing `gh run view` emits either (a) gh's own "failed to get run: HTTP 404: ..." text, which trips **no** `ERROR_PATTERNS` entry at all (dropped identically pre- and post-fix — the common case, unchanged), or (b) text that does match a pattern (e.g. "error connecting to api.github.com" on a network failure), which pre-fix landed at `confidence=low`/`category=pipe-mask-suspect` — already excluded from `annunaki_parse`'s genuine-error count — and post-fix is dropped entirely. The loss is confined to uncounted forensic records; no `confidence=high` shape is lost. The `CONTENT_DISPLAY_LOG_READ_VERBS` comment now states this, and `test_gh_run_view_piped_failure_no_pipefail_uncounted_drop` pins the actual (shell-realistic) behavior. The previous `test_gh_run_view_genuine_failure_nonzero_exit_still_logged` used a piped command with `exit_code=1` — a combination this shell cannot produce — so it was relabeled `test_gh_run_view_bare_nonzero_exit_still_logged` with the pipe removed: it correctly guards the `exit_code != 0` short-circuit at the top of `_is_content_display`, just not the `2>&1` exemption specifically.

## Acceptance criteria (from the issue)

- [x] `gh run view <id> --repo <r> --log-failed 2>&1 | tail -c 8000` over a log containing `Traceback` writes no record — `test_harvested_206_*`.
- [x] `rg -n "FAILED" <path-to-saved-log>` writes no record — `test_harvested_208_*` / `test_harvested_209_*` / `test_helper_true_on_rg_over_a_file`.
- [x] The `2>&1` disqualifier still fires for genuine probe-with-fallback shapes (`cat missing.py 2>&1 | head`) — no fail-open regression on #517: `test_cat_missing_file_2ampersand1_probe_shape_still_not_content_display`.
- [x] A genuine pipe-masked failure (`git push | tail` hiding a REJECTED push) is still captured at `confidence: high` — `test_strong_masked_failure_git_push_still_captured_high`.
- [x] The wave-29 records are a regression fixture — **all 17 are present in the live log** (verified: 17/52 by `error_lines` signature, exact); 3 are in the two command shapes `_is_content_display` can reach and are harvested as fixtures above; the other 14 are echoes in shapes the monitor cannot reach by design (leading `cd`/`VAR=` prefixes, heredocs, a `gh pr view ... 2>&1 | head` quoting a comment) — #1354's own proposed-fix item 4, explicitly scoped out, tracked separately at #1398.

## Testing

- `ENVIRONMENT=test PYTHONPATH=.claude/hooks python3 -m pytest .claude/hooks/tests/test_annunaki_monitor.py -v` → 101 passed (16 in `Issue1354CiLogReadTests` — 14 original + 2 added addressing MF2 — all pre-existing tests unaffected).
- `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/ .claude/lib/tests/ -q` → full suite green.
- `ruff format --check`, `ruff check`, `mypy` on the two changed files → clean.
- Full local⇄CI parity: `pre-commit run` (commit stage) and `pre-commit run --hook-stage push` (mypy + full pytest + memory/headcount budgets + doc-freshness + office-drift + mermaid-render) → all green.
- Nino Kavtaradze's merge-gate review independently replayed all 1158 live `errors.jsonl` records through both classifiers: 13 newly suppressed by adding `rg` (1.1%), zero reverse regressions, all 13 confirmed genuine content reads. Also ran 7 mutations on the new code; 4 killed, 2 survived (tracked non-blocking at #1397 — the No-such-file guard's log-read-verb bypass and the `^\s*` anchor on `CONTENT_DISPLAY_LOG_READ_VERBS`).

## Note on framing (issue's own callout, preserved)

Per the issue: the wave-29 triage brief characterized the 52 `annunaki_monitor` records as "the monitor itself erroring." They are not — `hook: annunaki_monitor` is the writer attribution; every one is `exit_code: 0` / a real capture. 35 of the 52 are the monitor working correctly. This PR only touches the content-display false-positive subclass.

## Non-blocking follow-ups filed during merge-gate review

- **#1397** — two mutation survivors in the new code (behavior is probably correct; should be a stated decision rather than an implicit one).
- **#1398** — the 14 residual wave-29 echoes in command shapes this fix cannot reach (#1354's own scoped-out item 4).

### DIFF
--- a/.claude/hooks/annunaki_monitor.py
+++ b/.claude/hooks/annunaki_monitor.py
@@ -220,7 +220,7 @@ CONTENT_DISPLAY_VERBS = re.compile(
         cat | head | tail | less | more | bat        # file pagers/dumpers
       | rg                                           # #1354: file-content search/dump
       | git\s+show | git\s+diff | git\s+log          # git content display
-      | gh\s+pr\s+diff | gh\s+pr\s+view               # gh PR content display
+      | gh\s+pr\s+diff | gh\s+pr\s+view              # gh PR content display
       | gh\s+run\s+view | gh\s+run\s+log             # #1354: gh CI-log content display
     )
     \b
@@ -258,13 +258,28 @@ CONTENT_DISPLAY_STDOUT_MERGE = re.compile(r"2>&1")
 # CI-log-read verbs (#1354): `gh run view --log-failed` / `gh run log` are the
 # canonical way to read a failed CI job's log, and routinely need `2>&1 |
 # tail -c N` to bound stdout size while still catching anything `gh` itself
-# writes to stderr (a warning, a rate-limit notice). Unlike `cat`/`head`/etc,
-# there is no "possibly-missing local path" for `gh run view` to probe — a
-# bad run id or network failure surfaces as a real nonzero exit, which the
-# `exit_code != 0` guard above already catches independently of this verb
-# check. So the general `2>&1` disqualifier is scoped away from these two
-# verbs specifically; every other content-display verb (including the new
-# `rg` verb, which is not a CI-log-specific idiom) keeps the disqualifier.
+# writes to stderr (a warning, a rate-limit notice).
+#
+# IMPORTANT — this repo's shell has no `pipefail` (zsh, no `setopt pipefail`
+# in play here), so for the piped canonical form the `exit_code != 0` guard
+# above does NOT independently catch a failing `gh run view`: the observed
+# exit code is the last stage's (`tail`'s), which is 0 even when `gh` itself
+# failed. Measured live: `gh run view <bad-id> --log-failed 2>&1 | tail -c
+# 8000; echo $?` → `0`, with `gh`'s failure text on stdout. So this
+# exemption is NOT "safe because a real failure would still surface via a
+# nonzero exit" — for the piped form, it would not.
+#
+# The exemption is accepted anyway because the residual class is bounded and
+# already low-value: a `gh run view` that itself fails emits either (a) gh's
+# own "failed to get run: HTTP 404: ..." text, which trips NO ERROR_PATTERNS
+# entry at all (dropped identically pre- and post-fix), or (b) a line that
+# DOES match a pattern (e.g. "error connecting to api.github.com"), which
+# pre-fix was retained at confidence=low/category=pipe-mask-suspect (#835) —
+# already excluded from annunaki_parse's genuine-error count — and post-fix
+# is dropped entirely. The loss is confined to uncounted forensic records;
+# no `confidence=high` shape is lost. Every other content-display verb
+# (including the new `rg` verb, which is not a CI-log-specific idiom) keeps
+# the `2>&1` disqualifier unchanged.
 CONTENT_DISPLAY_LOG_READ_VERBS = re.compile(
     r"""
     ^\s*
--- a/.claude/hooks/annunaki_monitor.py
+++ b/.claude/hooks/annunaki_monitor.py
@@ -21,8 +21,12 @@ Input Language:
                   matched pattern is `stdout:No such file or directory`),
                   content-display commands (#596 — exit=0 cat/head/tail/less,
                   `gh api .../contents/...`, or a read of errors.jsonl when the
-                  ONLY signals are stdout-pattern matches in displayed content),
-                  session-dedup hits
+                  ONLY signals are stdout-pattern matches in displayed content;
+                  #1354 extends this to `rg` over any file, `gh run view`/
+                  `gh run log` (a CI-log read, `2>&1` permitted for these two
+                  verbs specifically), and `gh api .../actions/runs/<id>/logs`
+                  — investigating an already-known CI failure is a content
+                  display too, not a fresh error), session-dedup hits
   Flag pass-through: stdin JSON is forwarded verbatim to `check()` by the
                      PostToolUse dispatcher (`post_dispatcher.py`)
   Confidence tag (#729): every logged record carries `confidence` in
@@ -214,8 +218,10 @@ CONTENT_DISPLAY_VERBS = re.compile(
     ^\s*
     (?:
         cat | head | tail | less | more | bat        # file pagers/dumpers
+      | rg                                           # #1354: file-content search/dump
       | git\s+show | git\s+diff | git\s+log          # git content display
-      | gh\s+pr\s+diff | gh\s+pr\s+view              # gh PR content display
+      | gh\s+pr\s+diff | gh\s+pr\s+view               # gh PR content display
+      | gh\s+run\s+view | gh\s+run\s+log             # #1354: gh CI-log content display
     )
     \b
     """,
@@ -227,6 +233,11 @@ CONTENT_DISPLAY_VERBS = re.compile(
 # api invocations are commonly preceded by a variable-assignment prefix and the
 # `contents/` path segment is an unambiguous content-read marker.
 CONTENT_DISPLAY_GH_API_CONTENTS = re.compile(r"\bgh\s+api\b.*\bcontents/")
+
+# `gh api .../actions/runs/<id>/logs` — the raw-logs REST read, the API analog
+# of `gh run view --log-failed` / `gh run log` (#1354). Same content-read
+# reasoning as CONTENT_DISPLAY_GH_API_CONTENTS above.
+CONTENT_DISPLAY_GH_API_LOGS = re.compile(r"\bgh\s+api\b.*\bactions/runs/\d+/logs\b")
 
 # Self-referential meta-capture guard (#596): a command that reads the annunaki
 # error log itself will echo historical records containing "Traceback" etc.
@@ -243,6 +254,25 @@ CONTENT_DISPLAY_ERRORS_LOG = re.compile(r"errors\.jsonl")
 # but is a failed read whose merged error text (or an unrelated Traceback in the
 # merged stream) must still log on its own merits (#517 regression guard).
 CONTENT_DISPLAY_STDOUT_MERGE = re.compile(r"2>&1")
+
+# CI-log-read verbs (#1354): `gh run view --log-failed` / `gh run log` are the
+# canonical way to read a failed CI job's log, and routinely need `2>&1 |
+# tail -c N` to bound stdout size while still catching anything `gh` itself
+# writes to stderr (a warning, a rate-limit notice). Unlike `cat`/`head`/etc,
+# there is no "possibly-missing local path" for `gh run view` to probe — a
+# bad run id or network failure surfaces as a real nonzero exit, which the
+# `exit_code != 0` guard above already catches independently of this verb
+# check. So the general `2>&1` disqualifier is scoped away from these two
+# verbs specifically; every other content-display verb (including the new
+# `rg` verb, which is not a CI-log-specific idiom) keeps the disqualifier.
+CONTENT_DISPLAY_LOG_READ_VERBS = re.compile(
+    r"""
+    ^\s*
+    gh\s+run\s+(?:view|log)
+    \b
+    """,
+    re.VERBOSE,
+)
 
 # The `No such file or directory` stdout match always signals a FAILED
 # filesystem read (a missing path), never error-shaped *content* we want to
@@ -319,7 +349,7 @@ JSON_BODY_LINE = re.compile(r'^\s*[+-]?\s*[{\[]\s*["{]')
 
 
 def _is_content_display(command: str, exit_code: int, matched_patterns: list[str]) -> bool:
-    """Return True if this matches the #596 content-display idiom.
+    """Return True if this matches the #596/#1354 content-display idiom.
 
     All of the following must hold:
       1. exit_code == 0 (the read/display succeeded)
@@ -327,18 +357,23 @@ def _is_content_display(command: str, exit_code: int, matched_patterns: list[str
          no `exit_code=` marker. A stderr pattern or non-zero exit means a real
          failure occurred alongside the display, so this skip does NOT apply.
       3. the command does NOT contain a `2>&1` stdout-merge AND no matched
-         pattern is the `No such file or directory` line. Both mark the #517
-         probe-with-fallback domain (an intentional/failed read of a missing
-         path) rather than displayed error-shaped content, so those cases are
-         left to the probe classifier. The `2>&1` guard catches the command
-         shape; the No-such-file guard catches the matched-signal even when the
-         command shape is absent (e.g. a bare `gh api .../contents/... > file`
-         that failed). Together they keep `cat missing.py 2>&1 | head` and the
-         non-trailer probe outliers from being mis-classified as displays.
+         pattern is the `No such file or directory` line — UNLESS the command
+         is a CI-log-read verb (`gh run view`/`gh run log`, #1354), which is
+         exempted from the `2>&1` sub-check specifically (see
+         CONTENT_DISPLAY_LOG_READ_VERBS). For every other verb, both guards
+         mark the #517 probe-with-fallback domain (an intentional/failed read
+         of a missing path) rather than displayed error-shaped content, so
+         those cases are left to the probe classifier. The `2>&1` guard
+         catches the command shape; the No-such-file guard catches the
+         matched-signal even when the command shape is absent (e.g. a bare
+         `gh api .../contents/... > file` that failed). Together they keep
+         `cat missing.py 2>&1 | head` and the non-trailer probe outliers from
+         being mis-classified as displays.
       4. the command is a recognized content-display operation: a leading
-         display verb (cat/head/tail/less/more/bat, git show|diff|log,
-         gh pr diff|view), a `gh api .../contents/...` read, OR a read of the
-         annunaki errors.jsonl log (self-referential meta-capture).
+         display verb (cat/head/tail/less/more/bat, rg, git show|diff|log,
+         gh pr diff|view, gh run view|log), a `gh api .../contents/...` or
+         `gh api .../actions/runs/<id>/logs` read, OR a read of the annunaki
+         errors.jsonl log (self-referential meta-capture).
 
     Condition 2 is the precedence guard mirroring the silent-boolean-test and
     probe-with-fallback families: any real failure signal bypasses the skip.
@@ -349,13 +384,17 @@ def _is_content_display(command: str, exit_code: int, matched_patterns: list[str
         return False
     if not all(p.startswith("stdout:") for p in matched_patterns):
         return False
-    if CONTENT_DISPLAY_STDOUT_MERGE.search(command):
-        return False
-    if CONTENT_DISPLAY_EXCLUDED_PATTERN in matched_patterns:
-        return False
+    is_log_read_verb = CONTENT_DISPLAY_LOG_READ_VERBS.search(command) is not None
+    if not is_log_read_verb:
+        if CONTENT_DISPLAY_STDOUT_MERGE.search(command):
+            return False
+        if CONTENT_DISPLAY_EXCLUDED_PATTERN in matched_patterns:
+            return False
     if CONTENT_DISPLAY_VERBS.search(command):
         return True
     if CONTENT_DISPLAY_GH_API_CONTENTS.search(command):
+        return True
+    if CONTENT_DISPLAY_GH_API_LOGS.search(command):
         return True
     if CONTENT_DISPLAY_ERRORS_LOG.search(command):
         return True
@@ -556,11 +595,13 @@ def check(input_data: dict) -> dict | None:
     if _is_probe_with_fallback(command, exit_code, matched_patterns):
         return None
 
-    # #596 content-display filter: a command that DISPLAYS content (cat/head/
-    # tail/less of a file, `gh api .../contents/...`, a read of errors.jsonl)
-    # and exits 0 with only stdout-pattern matches is echoing error-shaped
-    # content, not failing. The helper requires exit==0 AND every matched
-    # pattern be a stdout one, so a real stderr/exit-code signal bypasses this.
+    # #596/#1354 content-display filter: a command that DISPLAYS content
+    # (cat/head/tail/less/rg of a file, `gh api .../contents/...` or
+    # `.../actions/runs/<id>/logs`, `gh run view`/`gh run log`, a read of
+    # errors.jsonl) and exits 0 with only stdout-pattern matches is echoing
+    # error-shaped content, not failing. The helper requires exit==0 AND
+    # every matched pattern be a stdout one, so a real stderr/exit-code
+    # signal bypasses this.
     if _is_content_display(command, exit_code, matched_patterns):
         return None
 
--- a/.claude/hooks/tests/test_annunaki_monitor.py
+++ b/.claude/hooks/tests/test_annunaki_monitor.py
@@ -1490,19 +1490,51 @@ class Issue1354CiLogReadTests(unittest.TestCase):
             )
         )
 
-    def test_gh_run_view_genuine_failure_nonzero_exit_still_logged(self):
-        """A `gh run view` that itself fails (bad run id, network error —
-        nonzero exit) is a real failure, not a content display → must still
-        log (content-display requires exit_code == 0)."""
+    def test_gh_run_view_bare_nonzero_exit_still_logged(self):
+        """A BARE (unpiped) `gh run view` that itself fails — nonzero exit
+        propagates directly, no pager in the pipeline — is a real failure,
+        not a content display → must still log. This guards the
+        `exit_code != 0` short-circuit at the top of `_is_content_display`,
+        which fires before any #1354 verb/exemption logic runs; it is NOT a
+        guard on the `2>&1` exemption itself (that requires the piped form,
+        see `test_gh_run_view_piped_failure_no_pipefail_uncounted_drop`
+        below for the shape that actually exercises it, per Nino
+        Kavtaradze's #1385 merge-gate review)."""
         result = am.check(
             _bash_event(
-                "gh run view 99999999 --repo noorinalabs/noorinalabs-main "
-                "--log-failed 2>&1 | tail -c 8000",
+                "gh run view 99999999 --repo noorinalabs/noorinalabs-main --log-failed",
                 stdout="failed to get run: run not found\n",
                 exit_code=1,
             )
         )
         self.assertIsNotNone(result, "a genuinely failed gh run view must still log")
+
+    def test_gh_run_view_piped_failure_no_pipefail_uncounted_drop(self):
+        """The shell-realistic shape of a FAILING piped `gh run view`
+        (#1385 merge-gate MF2): this repo's shell has no `pipefail`, so
+        `gh run view <bad-id> --log-failed 2>&1 | tail -c 8000` exits 0 (the
+        pager's rc) even when `gh` itself failed — measured live. When gh's
+        failure text happens to trip an ERROR_PATTERNS entry (e.g. "error
+        connecting to ..."), the #1354 exemption now suppresses the record
+        entirely. Pre-#1354 this landed at confidence=low/
+        category=pipe-mask-suspect (already excluded from the genuine-error
+        count by annunaki_parse), so the record pinned here documents an
+        accepted, uncounted-forensics-only loss — not a `confidence=high`
+        regression. See the CONTENT_DISPLAY_LOG_READ_VERBS comment for the
+        full trade-off writeup."""
+        result = am.check(
+            _bash_event(
+                "gh run view 99999999 --repo noorinalabs/noorinalabs-main "
+                "--log-failed 2>&1 | tail -c 8000",
+                stdout="error connecting to api.github.com\n",
+                exit_code=0,
+            )
+        )
+        self.assertIsNone(
+            result,
+            "accepted trade-off: a failing piped gh run view with no pipefail is "
+            "indistinguishable from a successful content display and is dropped",
+        )
 
     def test_rg_over_log_with_stderr_pattern_still_logged(self):
         """`rg` over a saved log with a real stderr signal alongside (e.g. an
--- a/.claude/hooks/tests/test_annunaki_monitor.py
+++ b/.claude/hooks/tests/test_annunaki_monitor.py
@@ -1320,6 +1320,269 @@ class Rc0PrecisionTests(unittest.TestCase):
         self.assertEqual(rec["hook"], "annunaki_monitor")
 
 
+class Issue1354CiLogReadTests(unittest.TestCase):
+    """#1354 coverage: `_is_content_display` misses `gh run view`/`gh run log`
+    (a CI-log read) and `rg`-over-a-saved-log — a command that *investigates*
+    an already-known CI failure is captured as a fresh error.
+
+    Fixtures 1-3 are harvested verbatim (command + realistic stdout replaying
+    the actual `error_lines`) from the live `.claude/annunaki/errors.jsonl`
+    wave-29 incident the issue cites: a single failing test
+    (`test_wave_status.py::BaseVsHeadDifferential::
+    test_base_and_head_agree_on_wave_branch_output`, fixed in-wave by
+    `.github/workflows/ci.yml` `fetch-depth: 0`) investigated via `gh run
+    view --log-failed` then two `rg` re-reads of the saved log — three
+    high-confidence `masked-failure` records of one already-fixed incident.
+
+    Precedence siblings: #517 (probe-with-fallback, the `2>&1` guard's
+    original reason for existing) and #596 (the base content-display class
+    this issue extends) — both must be unaffected.
+    """
+
+    def setUp(self):
+        self._saved_env = {
+            "ENVIRONMENT": os.environ.pop("ENVIRONMENT", None),
+            "NOORIN_HOOK_TEST_MODE": os.environ.pop("NOORIN_HOOK_TEST_MODE", None),
+        }
+        am._seen_hashes.clear()
+        self._tmpdir = tempfile.TemporaryDirectory()
+        self._errors_path = Path(self._tmpdir.name) / "errors.jsonl"
+        self._orig_monitor_file = am.ERRORS_FILE
+        self._orig_log_file = alog.ERRORS_FILE
+        am.ERRORS_FILE = self._errors_path
+        alog.ERRORS_FILE = self._errors_path
+
+    def tearDown(self):
+        am.ERRORS_FILE = self._orig_monitor_file
+        alog.ERRORS_FILE = self._orig_log_file
+        self._tmpdir.cleanup()
+        for k, v in self._saved_env.items():
+            if v is None:
+                os.environ.pop(k, None)
+            else:
+                os.environ[k] = v
+
+    # --- Harvested wave-29 fixtures (errors.jsonl records #206/#208/#209) ---
+
+    # The three records replay the CI job's log text — the pytest failure
+    # summary line contains "returned non-zero exit status 128", which trips
+    # ERROR_PATTERNS' `exit status [1-9]` on stdout even though the *reading*
+    # command itself exited 0.
+    _CI_LOG_TEXT = (
+        "Pytest — hooks and lib test suites\tRun hook and lib tests\t"
+        "2026-07-30T03:24:39.0338201Z E               subprocess.CalledProcessError: "
+        "Command '['git', 'show', "
+        "'b290d611e9e6c59db0bf921ef6a9315090dc2eae:.claude/lib/wave_status.py']' "
+        "returned non-zero exit status 128.\n"
+        "Pytest — hooks and lib test suites\tRun hook and lib tests\t"
+        "2026-07-30T03:24:39.0342317Z FAILED "
+        ".claude/lib/tests/test_wave_status.py::BaseVsHeadDifferential::"
+        "test_base_and_head_agree_on_wave_branch_output - "
+        "subprocess.CalledProcessError: Command '['git', 'show', "
+        "'b290d611e9e6c59db0bf921ef6a9315090dc2eae:.claude/lib/wave_status.py']' "
+        "returned non-zero exit status 128.\n"
+        "Pytest — hooks and lib test suites\tRun hook and lib tests\t"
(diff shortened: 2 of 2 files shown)


parametrization:
Requestor: Nurul Hakim
Requestee: Aino Virtanen
RequestOrReplied: Request
TechDebt: none

Peer review requested for #1354 — `_is_content_display` misses `gh run view`/`gh run log` and `rg`-over-a-log. Pre-fix failure output and harvested regression fixtures (3 records from the live errors.jsonl, tracing back to the same wave-29 CI incident the issue cites) are in the PR body. Full local⇄CI parity (commit-stage + push-stage pre-commit) is green.

parametrization:
Requestor: Nurul Hakim
Requestee: Nino Kavtaradze
RequestOrReplied: Request
TechDebt: none

Secondary review (and merge-gate) requested for #1354. Same scope as the request to Aino Virtanen above — see the PR body for the pre-fix failure output and the harvested regression fixtures.

parametrization:
Requestor: Aino Virtanen
Requestee: Nurul Hakim
RequestOrReplied: Approved

ReviewedHead: d467689ccaa6f30cb23395d782c65e523790f62f
TechDebt: #1392

**Scope**: reviewed `.claude/hooks/annunaki_monitor.py` (`_is_content_display`, `CONTENT_DISPLAY_VERBS`/`CONTENT_DISPLAY_LOG_READ_VERBS`/`CONTENT_DISPLAY_GH_API_LOGS`) and `.claude/hooks/tests/test_annunaki_monitor.py::Issue1354CiLogReadTests` against PR head. All 14 CI checks green.

**1. Discriminating tests, verified empirically.** Extracted the pre-fix `annunaki_monitor.py` from `main` into a throwaway worktree, overlaid the PR's new test file, ran `Issue1354CiLogReadTests` against it: **9 failed / 5 passed**, matching the PR description exactly. The 5 pre-passing guards are not trivial — each pins a boundary the fix must not regress:
- `test_cat_missing_file_2ampersand1_probe_shape_still_not_content_display` / `test_gh_pr_view_2ampersand1_still_disqualified` — pin that the new `2>&1` exemption stays scoped to the two log-read verbs and doesn't leak to other display verbs.
- `test_gh_run_view_genuine_failure_nonzero_exit_still_logged` — pins `exit_code==0` gating.
- `test_rg_over_log_with_stderr_pattern_still_logged` — pins the "every matched pattern must be stdout" precedence guard.
- `test_strong_masked_failure_git_push_still_captured_high` — pins the STRONG_MASKED_FAILURE carve-out is untouched.

All 5 would fail under a mutant that broadens the `2>&1` exemption or weakens the stdout-only precondition, so they're genuine controls, not string assertions.

**2. Surviving mutant found.** Removing the `^\s*` leading-anchor from `CONTENT_DISPLAY_LOG_READ_VERBS` (`gh\s+run\s+(?:view|log)\b` matched anywhere in the command instead of only at the command start) leaves the *entire* `test_annunaki_monitor.py` suite green — 100/100 passed, verified by direct mutation + rerun. Current production code has the anchor and is correct; this is a coverage gap, not a live bug, but it's exactly the shape of gap `CONTENT_DISPLAY_VERBS`'s own docstring calls out as a deliberate design invariant (leading-verb anchoring, so a compound command's real build/test step still logs on its own merits) — `CONTENT_DISPLAY_LOG_READ_VERBS` has no test pinning the same invariant. Filed as #1392 with the reproduction and a suggested regression test.

**3. `rg` blast radius — judged acceptable, flagged for watching.** `rg` as a leading verb, exit 0, stdout-only matches, no `2>&1` → fully suppressed (`check()` returns `None`, not even a low-confidence forensic record — confirmed by reading the `_is_content_display` call site at L605, which short-circuits before the confidence/category tagging that the #729 echoed-content class goes through). This is architecturally consistent with the pre-existing #596 treatment of `cat`/`head`/`git show`/`gh pr view` — not a new class of behavior introduced here — but `rg` is uniquely ubiquitous as the org-mandated search tool (#1008), so the practical exposure is larger than any prior verb added to this set: a genuine masked issue surfaced by an `rg` read (e.g. `rg ERROR prod.log` finding unresolved fresh errors) is now silently dropped with zero trace, where pre-fix it would at minimum survive as a record (high-confidence, since `rg` wasn't recognized as content-display pre-fix). The tests correctly cover the boundary cases that keep this from over-firing (stderr present, nonzero exit, `2>&1` present all still log), so I'm not blocking on it, but I'd want the /annunaki-attack window after this merges to check whether legitimate `rg`-surfaced issues are going dark. Not filing separately — worth a watch-item at the next retro rather than an issue with no evidence yet.

**4. `2>&1` asymmetry — principled, not ad hoc.** The exemption for `gh run view`/`gh run log` specifically (vs. every other verb, `rg` included) is justified by a real structural difference the code comment (L258-267) states directly: log-read-by-run-id has no "possibly-missing local path" to probe — a bad run id or network failure surfaces as a real nonzero exit, caught independently by the `exit_code != 0` guard upstream. `rg`/`cat`/etc. read local paths that may not exist, so `2>&1` there is exactly the #517 probe-with-fallback signal (a caught not-found error), which must stay live. Verified `test_gh_run_view_genuine_failure_nonzero_exit_still_logged` covers the "bad run id → still logs" side of this reasoning. Agree with the scoping.

**5. Test-mode hygiene.** `Issue1354CiLogReadTests.setUp` pops `ENVIRONMENT`/`NOORIN_HOOK_TEST_MODE` rather than setting `ENVIRONMENT=test` — but this is the established pattern across *every* `setUp` in this file (checked: 9/9 test classes use the identical pop/restore idiom), not something introduced or worsened by this PR. It's deliberate: popping neutralizes whatever the caller's shell happened to have set, so `_is_test_mode()` is `False` and the test's own `ERRORS_FILE` redirection (to a `tempfile.TemporaryDirectory()`) is the actual isolation mechanism — writes happen for real but land in a throwaway path. That's consistent, not a new instance of #1381's gap.

No blocking findings. Approving.

parametrization:
**Merge-gate review — ChangesRequested. The code is right; two of the claims justifying it are not.**

Reviewed at head `d467689ccaa6f30cb23395d782c65e523790f62f` (re-confirmed head immediately before posting), diffed against base `6c42709` via `gh api contents/...?ref=<sha>` rather than a local checkout. All 14 checks SUCCESS, and both workflow runs report `head_sha` equal to the reviewed SHA. `mergeable_state: clean`.

Neither must-fix asks for a logic change. Both are corrections to text and one test.

## What I verified independently, and what held

**Failing-before evidence — confirmed, both halves.** I ran the head test file against the pre-fix module and got exactly the reported split, test-for-test: 9 FAILED, 5 PASSED. The 5 pre-passing guards are genuine, so the other 9 are not trivially green.

**Recall on the two cases that matter — preserved.** Replayed through both modules:

| case | pre-fix | post-fix |
|---|---|---|
| `git push ... \| tail` masking `[rejected]` non-fast-forward | LOGGED high/masked-failure | LOGGED high/masked-failure |
| `rg` over a log with a real stderr signal | LOGGED high/stderr-match | LOGGED high/stderr-match |

**Blast radius of adding `rg` — measured, and it is small.** I replayed all 1158 records of the live `errors.jsonl` through both classifiers. 13 records newly suppressed (1.1%); zero records suppressed by the old code but logged by the new one. Of the 13, 10 are `low`/`pipe-mask-suspect` and all 13 are, on inspection, genuine content reads. The `^\s*` anchor is doing the work here: `cd /repo && rg ...` is the common shape and is correctly not matched. I am satisfied `rg` is safe.

**The scoping of the `2>&1` exemption is genuinely pinned.** I mutated `if not is_log_read_verb:` to `if False:` (exemption applied to every verb) and 6 tests fail, including both deliberate regression guards. That asymmetry is not inert.

## Must-fix 1 — the issue's "17" reproduces exactly; the PR body says it doesn't

This is the one that matters, because wave 30's premise is that measured precision figures are load-bearing.

Cut the log immediately before the wave-29 retro's own harvest record (`2026-08-04T01:28:00`; #1354 was filed at `01:36:39`) and count monitor-class high-confidence records plus the signature subset:

```python
import json
recs = [json.loads(l) for l in open(".claude/annunaki/errors.jsonl", encoding="utf-8") if l.strip()]
blob = lambda r: "\n".join(r.get("error_lines", [])) + "\n" + (r.get("stderr_excerpt") or "")
sig  = lambda r: all(s in blob(r) for s in ("CalledProcessError", "b290d611", "wave_status"))
at   = [r for r in recs if r.get("timestamp", "") <= "2026-08-04T01:28:00"]
hi   = [r for r in at if r.get("hook") == "annunaki_monitor" and r.get("confidence") == "high"]
print(len([r for r in hi if sig(r)]), "/", len(hi))   # -> 17 / 52
```

`17 / 52`. Exact, both numerator and denominator. The log was never pruned — its earliest record is `2026-07-27T02:17:38Z`, three days before the 2026-07-30 incident.

The gap is a measurement-method difference, not a missing-data one. #1354 defines its 17 by an **`error_lines` signature** (it says so: *"All carry `subprocess.CalledProcessError: Command '['git', 'show', 'b290d611...`"*). This PR searched by **command shape** and found the 3 CI-log-read commands, which is the correct set for the *fix* but the wrong set for the *count*.

So please rewrite two places:

- the harvest section's *"I did not fabricate the difference between 3 and the issue's 17"* paragraph, and
- the acceptance-criteria line *"the issue's '17' is not reproducible from this environment's log, called out rather than fabricated"*.

The accurate statement is that all 17 are present, 3 of them are in the two command shapes this fix addresses and are harvested as fixtures, and the remaining 14 are echoes in shapes `_is_content_display` cannot reach by design. I filed #1398 for those 14 — they are #1354's own proposed-fix item 4 (signature collapse at read time), which that issue explicitly scoped out of the monitor, so this is not scope you need to absorb. It just needs to not be described out of existence.

To be clear about the instinct: reporting the discrepancy instead of inventing 14 fixtures was the right call. The reporting was right and the measurement was wrong, and only the measurement needs fixing.

## Must-fix 2 — the exemption's stated rationale is wrong about the shell, and the test that should catch it cannot fire

The `CONTENT_DISPLAY_LOG_READ_VERBS` comment and the PR body both justify dropping the `2>&1` guard this way: *"a bad run id or network failure surfaces as a real nonzero exit, which the `exit_code != 0` guard above already catches independently of this verb check."*

Measured live, on the exact canonical form the comment names:

```
$ gh run view 99999999 --repo noorinalabs/noorinalabs-main --log-failed 2>&1 | tail -c 8000
failed to get run: HTTP 404: Not Found (https://api.github.com/repos/.../actions/runs/99999999?...)
$ echo $?
0
```

`0`. The shell has no `pipefail` set, so the pipeline rc is `tail`'s. For the piped form — the one the comment calls canonical and the one every harvested fixture uses — the `exit_code != 0` guard never fires, so it cannot be what makes the exemption safe.

And `test_gh_run_view_genuine_failure_nonzero_exit_still_logged` passes `exit_code=1` on a piped command, a combination the shell cannot produce. It is one of the 5 pre-passing guards, and it passes both pre- and post-fix for a reason unrelated to this change: `exit_code != 0` short-circuits `_is_content_display` on its first line, before any #1354 logic runs. It is not a guard on the exemption.

The good news is that the exemption is still defensible, just for a different reason, and I measured it so you can state it with evidence. The real behaviour change for a genuinely failing piped `gh run view`:

| stdout | pre-fix | post-fix |
|---|---|---|
| `error connecting to api.github.com` | LOGGED low/pipe-mask-suspect | DROPPED |
| `failed to get run: HTTP 404: Not Found (...)` (measured live text) | DROPPED | DROPPED |

The second row is the common case and is unchanged, because gh's own 404 text trips no `ERROR_PATTERNS` entry at all. The first is a real loss, but it lands at `low`/`pipe-mask-suspect`, which `annunaki_parse` already excludes from the genuine-error count. So the loss is confined to uncounted forensic records, and no `high`-confidence shape I could construct is lost. That is a fine trade and I am not asking you to remove the exemption.

What I am asking for:

1. Replace the "bad run id is a real nonzero exit" rationale in the `CONTENT_DISPLAY_LOG_READ_VERBS` comment (and the matching PR-body paragraph) with the true one: in the piped form the rc is the pager's, and the exemption is acceptable because the residual class is `low`/`pipe-mask-suspect` and already uncounted.
2. Add one test in the shape that can actually occur — a failing `gh run view <bad id> --log-failed 2>&1 | tail -c 8000` at `exit_code=0` with gh's error text on stdout — pinning whatever you decide the right outcome is. Keep the existing `exit_code=1` test if you like; it is a valid guard on the exit-code short-circuit, it is just mislabelled as covering the risky case.

## Non-blocking

Filed as tech-debt, no action needed on this PR:

- **#1397** — two guards in the new block are inert. Moving `CONTENT_DISPLAY_EXCLUDED_PATTERN` out of the `if not is_log_read_verb:` block (making it unconditional, which is what the docstring says the code does) leaves 100/100 passing; deleting the `^\s*` anchor from `CONTENT_DISPLAY_LOG_READ_VERBS` also leaves 100/100 passing. Worth noting that the log-read exemption currently bypasses the No-such-file guard as well as the `2>&1` one, which the docstring's *"the `2>&1` sub-check specifically"* wording does not describe. The behaviour is probably correct; it should be a stated decision.
- **#1398** — the 14 residual echoes described above.

Cosmetic, fix if you are touching the file anyway: line 223 gained a stray extra space before its trailing comment (`gh\s+pr\s+diff | gh\s+pr\s+view               #`), harmless inside `re.VERBOSE` but it makes the column alignment inconsistent with its neighbours.

The framing note you preserved from the issue — that `hook: annunaki_monitor` is writer attribution, not the monitor erroring — is correct and worth keeping.

---
Requestor: Nino Kavtaradze
Requestee: Nurul Hakim
ReviewedHead: d467689ccaa6f30cb23395d782c65e523790f62f
TechDebt: #1397, #1398

parametrization:
Requestor: Nurul Hakim
Requestee: Nino Kavtaradze
RequestOrReplied: Reply
TechDebt: #1397, #1398

Both must-fixes addressed as additive commits (no force-push — Aino Virtanen's Approved stays anchored to `d467689`, the SHA she reviewed).

**MF1.** Re-ran your script myself and got the same `17 / 52`. Rewrote both flagged spots in the PR body: the harvest section now states all 17 are present, that my original 3-vs-17 framing conflated "correct set for the fix" (command-shape) with "correct set for the count" (`error_lines` signature), and that the 14 residual echoes are #1354's own scoped-out item 4, now tracked at #1398. No code change — this was a PR-body-only fix, made via `gh api ... -X PATCH` since GraphQL quota was exhausted.

**MF2.** Confirmed live: `gh run view 99999999 --repo noorinalabs/noorinalabs-main --log-failed 2>&1 | tail -c 8000; echo $?` → `0`, gh's 404 text on stdout, no `pipefail` in this shell. Replaced the `CONTENT_DISPLAY_LOG_READ_VERBS` rationale (and the matching PR-body paragraph) with the real one: the residual class is `confidence=low`/`category=pipe-mask-suspect`, already excluded from `annunaki_parse`'s count, so the loss is bounded to uncounted forensics. Renamed the old test to `test_gh_run_view_bare_nonzero_exit_still_logged` (dropped the piped form it couldn't actually guard) and added `test_gh_run_view_piped_failure_no_pipefail_uncounted_drop`, which pins the shell-realistic failing-piped-command outcome using your `"error connecting to api.github.com"` example.

Also picked up the cosmetic double-space at the old line 223.

New head: `dd23fbbb8301ab66550d6f742277fe1e4528328a`. 101/101 tests pass locally (16 in `Issue1354CiLogReadTests`), full local⇄CI parity (commit-stage + push-stage pre-commit) green before push; CI itself in flight at push time.

Thank you for the live measurements on both — the `echo $?` on the actual `gh run view` invocation and the exact `17/52` derivation are things I should have run myself before writing either claim.

parametrization:
## Merge-gate re-review — both must-fixes cleared

Re-reviewed at `dd23fbb`, diffed against base `6c42709` via `gh api .../contents?ref=<sha>` (not a local checkout). Both MF1 and MF2 are resolved. Verdict flips to `Approved`.

### MF1 (the 17-vs-3 claim) — cleared, and I re-derived the number independently

I re-ran the signature count against the live `errors.jsonl` myself rather than trusting the transcription:

```
PR-body script result: 17 / 52
total records: 1252   earliest: 2026-07-27T02:17:38+00:00
```

Exact on both numerator and denominator, and the earliest-record timestamp confirms the "never pruned" claim (three days ahead of the incident). Both places I blocked on are corrected — the harvest section now opens with the explicit retraction ("the earlier revision of this section wrongly said it wasn't reproducible") and the acceptance-criteria checkbox now reads "**all 17 are present in the live log**". No residual "not reproducible" wording survives anywhere in the body.

I also checked the corrected wording doesn't smuggle in a *new* wrong claim, by replaying all 17 records through both classifiers rather than reading the prose:

| | base `6c42709` | head `dd23fbb` |
|---|---|---|
| 3 harvested (`gh run view` ×1, `rg -n` ×2) | `high`/`masked-failure` | **`None`** |
| other 14 | `high`/`masked-failure` | `high`/`masked-failure` (unchanged) |

3 suppressed / 14 unchanged / 0 reverse regressions — the body's arithmetic is exact.

### MF2 — the replacement test genuinely bites

This was the crux: the old test asserted `exit_code=1` on a piped command, a combination the shell cannot produce, so it was inert against the change it appeared to guard. A replacement that is also inert would be the same defect twice, so I mutation-tested it rather than reading it.

**Mutation — exemption removed entirely** (`is_log_read_verb = False`):

```
FAILED ...::test_gh_run_view_piped_failure_no_pipefail_uncounted_drop
FAILED ...::test_gh_run_log_verb_not_logged
FAILED ...::test_harvested_206_gh_run_view_log_failed_not_logged
FAILED ...::test_helper_true_on_gh_run_view_log_failed_with_2ampersand1
```

The new test dies with the exemption. It is not inert.

It also pins a shape the shell can actually produce — I checked the fixture's stdout is load-bearing, not decorative: `error connecting to api.github.com` matches `ERROR_PATTERNS`' `^error\b` (IGNORECASE|MULTILINE), so the record genuinely reaches the classifier rather than being dropped upstream for want of any match. That was the specific way this test could have been quietly inert, and it isn't.

The rationale reproduces live, too:

```
$ setopt | rg -i pipefail   ->  (unset)
$ gh run view 99999999 --repo noorinalabs/noorinalabs-main --log-failed 2>&1 | tail -c 200; echo $?
failed to get run: HTTP 404: Not Found (https://api.github.com/.../actions/runs/99999999?...)
0
```

The rename of the old test to `test_gh_run_view_bare_nonzero_exit_still_logged` (pipe dropped) is the right call — it now guards the `exit_code != 0` short-circuit honestly, and its docstring says so explicitly instead of implying it guards the `2>&1` exemption.

### Recall re-confirmed — and the comment rewrite changed no behaviour

A comment rewrite shouldn't move behaviour, so I measured that rather than assuming it. Replaying all **1252** live records through `d467689` (pre-fix-round) and `dd23fbb` (head):

```
commit1 vs head IDENTICAL: True  |  differing: 0
base -> head:  14 newly suppressed,  0 reverse regressions,  0 reclassified
```

Byte-identical classification across the whole log — the fix round is comment-and-test-only in effect, as claimed. (14 rather than my earlier 13 is log growth, 1158 → 1252, not a behaviour change.) All 14 newly-suppressed are genuine content reads: 11 `rg` searches over source files plus the 3 harvested `#1354` fixtures.

The residual bound holds, verified against base rather than reasoned about:

| shape (exit 0, gh-failure text) | base | head |
|---|---|---|
| `... 2>&1 \| tail -c 8000` | `low`/`pipe-mask-suspect` | `None` |
| `... 2>&1 > /tmp/out.txt` | `low`/`pipe-mask-suspect` | `None` |
| `... 2>&1` (bare) | `low`/`pipe-mask-suspect` | `None` |
| gh's `HTTP 404` text (any shape) | `None` (trips no pattern) | `None` |

I checked the unpiped `2>&1` shapes too, since the exemption covers them and no test pins them — they land in the same `low` class, so the bound isn't specific to the piped form. `low` is excluded from the genuine-error count by `annunaki_parse.is_low_confidence` / `iter_records`. And nothing `gh` emits on its own failure appears in `STRONG_MASKED_FAILURE`'s vocabulary (`failed to push`, `[rejected]`, `non-fast-forward`, `exit status [1-9]`, `failed with exit code`, `(exit N)`, `Traceback`), so the "no `confidence=high` shape is lost" claim is sound as scoped to the failing-`gh` residual class.

### #1397 mutations — genuinely unchanged, still correctly non-blocking

Re-ran both at head; both still survive (101 passed under each), and the fix round touched neither the regex body nor the guard block:

- dropping the `^\s*` anchor from `CONTENT_DISPLAY_LOG_READ_VERBS` → 101 passed
- moving the No-such-file guard back outside the exemption → 101 passed

Still the right call to leave non-blocking: both are *documented-decision* gaps rather than wrong behaviour. Worth flagging for whoever picks up #1397 that the docstring/code mismatch is still live — condition 3 says log-read verbs are "exempted from the `2>&1` sub-check **specifically**", but the code moves *both* the `2>&1` and the `CONTENT_DISPLAY_EXCLUDED_PATTERN` guards inside `if not is_log_read_verb:`, so the No-such-file guard is bypassed too. That is exactly what #1397 is for; no change wanted in this PR.

### Non-blocking observations (body-only, not worth an issue)

Two small imprecisions in the PR description. Neither touches code, an assertion, or the merge decision, and both evaporate on merge — flagging rather than filing:

1. Testing section says "16 in `Issue1354CiLogReadTests` — 14 original + 2 added". Collected count at head is **15** (14 → 15: one renamed, one added). The `101 passed` total is correct.
2. The 14 residual echoes are characterised as shapes `_is_content_display` "cannot reach by design", enumerated as leading `cd`/`VAR=` prefixes, a heredoc, and a `gh pr view ... 2>&1 | head`. One of the 14 is actually a *leading* `rg -n --hidden ... 2>&1 | rg -v ...` — reached fine by the leading-verb anchor, then excluded by the `2>&1` disqualifier, a fourth cause the list doesn't name. The count of 14 and the behaviour are exact; only the cause list is short by one.

### CI

14/14 green at `dd23fbb`, confirmed via `check-runs` at the head SHA (not the PR-level rollup): Ruff lint, Ruff format, Mypy, Pytest, commit-author verification, roster-union drift, pre-commit⇄CI sync drift, memory/headcount budgets, and the rest.

Aino Virtanen's `Approved` is anchored at `d467689`; since head classifies identically to `d467689` across all 1252 records and the push was additive, it stands. Not merging — that is the orchestrator's call.

---
Requestor: Nino Kavtaradze
Requestee: Nurul Hakim
RequestOrReplied: Approved
ReviewedHead: dd23fbbb8301ab66550d6f742277fe1e4528328a
TechDebt: #1397, #1398

parametrization:
Re-confirming after the additive fix commit moved `T_content`; this supersedes my earlier `Approved` (anchored at `d467689`), which the reviewer-verdict resolver correctly treats as stale.

Re-reviewed the two things in my lane specifically:

**1. Mutation-hole check on the replacement test.** `test_gh_run_view_piped_failure_no_pipefail_uncounted_drop` (replacing the shell-unrealistic `test_gh_run_view_genuine_failure_nonzero_exit_still_logged`) exercises the real `2>&1`-piped, no-`pipefail` shape: `gh run view <bad-id> --log-failed 2>&1 | tail -c 8000` with `exit_code=0` and stdout `"error connecting to api.github.com"`. Traced it by hand: `ERROR_PATTERNS`'s `^error\b` matches the stdout, so `matched_patterns` is non-empty (the test isn't inert on that front); `is_log_read_verb` is True, which routes past the `2>&1`/No-such-file guards straight to `CONTENT_DISPLAY_VERBS`, which matches `gh run view` → suppressed, `assertIsNone` holds. Statically mutating `is_log_read_verb` to a hard `False` flips the routing back through the `2>&1`-disqualifier path (the command contains `2>&1`) → `_is_content_display` returns `False` → the record would log → `assertIsNone` fails. Confirms Nino's report that this test genuinely bites, from my own trace rather than taking it on faith.

Separately: I did check whether this round closed my original #1397 `^\s*` anchor gap (mutating away the leading-anchor on `CONTENT_DISPLAY_LOG_READ_VERBS` so it could match mid-command in a compound shape like `cd x && gh run view ...`). It does not — none of the new/renamed tests use a prefixed compound command, so that specific mutation still survives. That's the pre-existing, already-filed #1397 gap, not a new one introduced by this round; the round's own scope (MF2) was the `2>&1`/no-pipefail rationale and the inert-test replacement, not #1397's anchor coverage, and #1397 is still tracked open for it. A replacement test that is itself inert would be the same defect twice — this one isn't; the untouched, already-tracked gap is a different thing.

**2. Test-mode hygiene on the new test.** Both new/renamed tests (`test_gh_run_view_bare_nonzero_exit_still_logged`, `test_gh_run_view_piped_failure_no_pipefail_uncounted_drop`) live in the existing `Issue1354CiLogReadTests` class and inherit its shared `setUp`/`tearDown` — pops `ENVIRONMENT`/`NOORIN_HOOK_TEST_MODE` rather than setting `ENVIRONMENT=test`, isolating via `ERRORS_FILE` redirection to a `tempfile.TemporaryDirectory()`. Same pattern as last round, no separate setUp override, no new #1381 instance.

CI green at head. MF1 (PR-body-only correction) and MF2 (rationale + rename + new test) both check out.

---
Requestor: Aino Virtanen
Requestee: Nurul Hakim
RequestOrReplied: Approved
ReviewedHead: dd23fbbb8301ab66550d6f742277fe1e4528328a
TechDebt: #1397, #1398

Issue #1348, opened 2026-08-04
review_false_positives is a bare word match: 17/17 wrong in wave-29, docstring guarantees a conservatism the mechanism lacks


parametrization:
## Summary

`review_false_positives` in `.claude/lib/trust_signals.py` is measured by a bare word match over the whole comment body. In wave-29 it fired **17 times and was wrong all 17 times** — a 100% false-positive rate on the metric whose entire job is counting false positives. Because each one costs the reviewer `-1` trust and also disqualifies them from every positive delta, this is the single largest distortion in the trust matrix.

## Root cause

```python
# .claude/lib/trust_signals.py:93-99
# A verdict comment is a "false positive" when its author explicitly retracts it
# (withdrawn / false-positive / retracted). Heuristic, deliberately conservative
# — only a self-marked retraction counts, never an inferred one.
_FALSE_POSITIVE_RE = re.compile(
    r"\b(false[\s-]?positive|withdrawn|withdraw|retracted|retract|invalid finding)\b",
    re.IGNORECASE,
)
```

The docstring guarantees "only a self-marked retraction counts, never an inferred one." The mechanism is a substring search for a **word** anywhere in the body (after code-span stripping). A bare word match cannot distinguish:

- the reviewer retracting **their own verdict** (what the metric means), from
- the reviewer *using the words* — describing a false-positive corpus, naming the defect class, quoting someone else, or negating ("this is **not** a false positive").

This is a guarantee stated in prose that the mechanism does not implement.

**Second, independent bug** — non-verdict comments are counted. In `extract_signals`:

```python
for v in verdicts:
    if _is_changes_requested(v.verdict): ...
    if v.false_positive and v.requestor:
        _bucket(v.requestor).review_false_positives += 1   # no verdict-kind guard
```

`parse_verdicts` builds a `Verdict` for **every** comment carrying a `RequestOrReplied` line, including `Request` and `Reply` — which are not verdicts at all. `false_positive` is computed for anything not literally `approved`. So a `Request` or `Reply` comment can dock a reviewer's trust. Two of the 17 wave-29 hits are exactly this (`#1310` verdict=`Reply`, `#1153` verdict=`Request`).

## Evidence — all 17 wave-29 hits, classified

[...]

PR #1388, opened 2026-08-11
fix(#1243): wave-audit gate fails CLOSED on a missing dependency + SSOT-identity guard on _ORG_REPOS

### LINKED ISSUE #1243: validate_wave_audit: no SSOT-identity guard on _ORG_REPOS (mutation survives full suite) + org_repos ImportError fails open

### LINKED ISSUE #1405: tech-debt(hook17): _fail_closed blocks /handoff, silently reversing the carve-out the same docstring argues for
## Context

Surfaced by the merge-gate review of PR #1388 (closes #1243), which added `_fail_closed` to `.claude/hooks/validate_wave_audit.py` (Hook 17). Non-blocking; the fail-closed behaviour itself is correct and independently verified. This is about **which** skills it refuses.

## The asymmetry

`_fail_closed` blocks every member of `_GATED_SKILLS` — `wave-wrapup`, `wave-retro`, **and `handoff`**:

```python
skill_name = _gated_skill_on_stdin()
if not skill_name:
    sys.exit(0)
...
sys.exit(2)
```

Measured at PR #1388 head `83f54f81`, minimal faithful `.claude/hooks/` + `.claude/lib/` tree, `_hook_main.py` removed:

```
_hook_main ABSENT, skill=handoff  ->  exit=2 (BLOCK)
```

But a few paragraphs above in the *same* module docstring, `/handoff` is deliberately **excluded** from the coverage-blocking set, with an explicit rationale:

> `handoff` is deliberately EXCLUDED, from both the partial AND total case. It only *records* session state; its degraded outcome (a thinner pickup prompt) is recoverable, and the Stop hook writes a handoff automatically regardless. Hard-blocking it during exactly the kind of API outage that makes recording state valuable would strand a session — and the only escape is the settings.json emergency removal, which would disable the gate for wrapup and retro too.

Every clause of that rationale applies verbatim to a broken-dependency failure, and arguably harder: a `sys.path` that no longer reaches `.claude/lib` (one of the causes PR #1388's own block message names) is precisely a broken-checkout situation in which you most want `/handoff` to record where you were.

## Why it is not obviously wrong either

The counter-argument is real and may well win: a missing dependency is a *local, fixable* condition (restore the file), not an external outage you must wait out, so "stranding a session" is a weaker objection than it is under a quota outage. And a gate whose own code is broken arguably should not vouch for anything.

The point of this issue is that **the decision was never stated**. `catalog-13-17.md` Hook 17's new broken-dependency bullet says "only a *gated* skill (`wave-wrapup`, `wave-retro`, `handoff`) is blocked" without noting that this reverses, for this one failure class, a carve-out the bullet immediately above it argues for at length. A future reader hits two documented policies for `/handoff` and no note that the difference is intentional.

## Proposed

Pick one, and say why in both the docstring and `catalog-13-17.md`:

1. **Keep the current behaviour** and add one sentence explaining why broken-dependency differs from coverage-failure for `/handoff`. Cheapest; probably correct.
2. **Scope `_fail_closed` to `_COVERAGE_BLOCKING_SKILLS`** and let `/handoff` through with a loud stderr warning, matching its coverage-failure treatment. This needs care: at that point `_hook_main` may be the missing module, so the warning has to be emitted with stdlib only.

Either way, add a `MissingDependencyFailsClosed` case pinning the chosen `/handoff` verdict with the rationale in the test name, so a later change to `_GATED_SKILLS` cannot silently flip it.

## Not in scope

The generic `run_blocking` fail-open class is #1402. The `.claude/lib` worktree checksum-tracking gap is #1219.

parametrization:
Closes #1243
Closes #1405

Two coupled defects in `.claude/hooks/validate_wave_audit.py` — a **BLOCKING** PreToolUse gate. Reviewers: @noorinalabs (peer: Aino Virtanen, Standards & Quality Lead · secondary + merge-gate: Nino Kavtaradze, Security Engineer).

Per this wave's acceptance bar, **each defect lands with a test that FAILS against the pre-fix implementation**. Both pre-fix failure outputs are pasted verbatim below.

---

### Correction to the issue's premise (worth reading before the diff)

#1243 recorded, at head `e0b6711`, that the literal BUG-08 mutation (a hand-copied 7-tuple dropping `noorinalabs-isnad-ingest-platform`) survived **3448 passed, 647 subtests, zero failures**. At current head that is no longer exactly true, and the difference matters for how you read the fix.

**Mutation A — the issue's literal 7-tuple** (drops `noorinalabs-isnad-ingest-platform`), full suite, new classes deselected:

```
FAILED .claude/hooks/tests/test_validate_wave_audit.py::PartialAuditFailureCoverage::test_total_failure_blocks_coverage_blocking_skill
1 failed, 4184 passed, 18 deselected, 947 subtests passed in 39.32s
```

Exactly **one** test now catches it, and it catches it **by accident**: `test_total_failure_blocks_coverage_blocking_skill` asserts the literal string `"any of the 8 org repo(s)"` inside a *block reason*, and that reason happens to interpolate `len(_ORG_REPOS)`. It is a message-text assertion in a coverage test that has nothing to do with SSOT identity. Treating that as the guard would be a mistake, which the next mutation shows.

**Mutation B — length-preserving**: keep 8 entries, typo one repo name (`noorinalabs-isnad-ingest-plaform`, one letter dropped). This is the *more* likely real drift — a hand-retyped list is far likelier to be mistyped than miscounted.

```
4185 passed, 18 deselected, 947 subtests passed in 44.92s
```

**Zero failures.** The gate would sweep 7 real repos plus one that does not exist, `gh` would fail on the phantom repo, and… it would be recorded as an unqueried repo — which, for `/wave-wrapup` and `/wave-retro`, *blocks*. So mutation B is not merely undetected by the suite; on the live gate it degrades into a permanent unexplainable block on the two skills the gate exists for, with the real cause (a typo in a list) invisible from every message the hook prints. Either way, nothing in CI says a word.

So the issue's conclusion holds and is if anything understated: the pre-fix suite cannot distinguish the correct repo list from a corrupted one, and the single incidental detector that exists would evaporate the day the org legitimately grows to 9 repos and someone updates that literal.

### The failing-before test

`OrgReposSsotIdentity` (3 assertions: SSOT identity, arity, and the specific repo BUG-08 dropped).

Against **mutation A** (the issue's literal 7-tuple):

```
FAILED test_validate_wave_audit.py::OrgReposSsotIdentity::test_ingest_platform_is_audited
FAILED test_validate_wave_audit.py::OrgReposSsotIdentity::test_org_repos_has_all_eight_repos
FAILED test_validate_wave_audit.py::OrgReposSsotIdentity::test_org_repos_is_the_ssot_object
3 failed, 124 deselected in 0.33s

E   AssertionError: ('noorinalabs-main', 'noorinalabs-isnad-graph', 'noorinalabs-user-service',
    'noorinalabs-deploy', 'noorinalabs-design-system', 'noorinalabs-data-acquisition',
    'noorinalabs-landing-page')
    is not
    ('noorinalabs-main', 'noorinalabs-isnad-graph', 'noorinalabs-user-service',
    'noorinalabs-deploy', 'noorinalabs-design-system', 'noorinalabs-data-acquisition',
    'noorinalabs-isnad-ingest-platform', 'noorinalabs-landing-page')
    : _ORG_REPOS must be imported from org_repos (the SSOT), never re-typed
```

Against **mutation B** (length-preserving typo — the one the whole suite missed):

```
FAILED test_validate_wave_audit.py::OrgReposSsotIdentity::test_ingest_platform_is_audited
FAILED test_validate_wave_audit.py::OrgReposSsotIdentity::test_org_repos_is_the_ssot_object
2 failed, 1 passed, 124 deselected in 0.32s

E   AssertionError: 'noorinalabs-isnad-ingest-platform' not found in
    (..., 'noorinalabs-isnad-ingest-plaform', 'noorinalabs-landing-page')
```

Both mutations killed. `assertIs`, not `assertEqual`, is the load-bearing choice: an equal-but-distinct tuple is exactly the drift being guarded against *at its birth moment* — the copy is correct on the day it is typed and rots silently afterwards, so identity is the only assertion that fails on day one. The arity assertion is the complement `assertIs` cannot make: a repo dropped from `org_repos.py` **itself** leaves the alias the same object.

---

## Defect 2 — an unimportable dependency made the blocking gate fail OPEN

The more dangerous of the pair. Exit **2** blocks; every other non-zero exit is a *non-blocking* error and the tool call proceeds. Pre-fix, with `org_repos.py` absent:

```
ModuleNotFoundError: No module named 'org_repos'
exit=1
```

— so `/wave-wrapup` ran. A blocking gate that fails open is, from the outside, indistinguishable from a gate that ran and approved. This is the same silent-allow shape #1226 and #1230 each closed one level up inside this file's own audit aggregation, reappearing one layer down in the import statement.

### The failing-before test

`MissingDependencyFailsClosed` — 10 cases, subprocess-level, because `check()` is *never called* here (the module never finishes importing) and the exit code is the only observable that separates BLOCK from ALLOW. Each case builds a minimal faithful copy of the deployed layout (`<root>/.claude/hooks/` + `<root>/.claude/lib/`) and omits exactly one dependency file — modelling the real "file is not there" condition, not a monkeypatched import system.

Headline case, pre-fix:

```
    def test_missing_org_repos_blocks_wave_wrapup(self) -> None:
        """THE #1243 defect-2 assertion: absent SSOT → exit 2, not 1."""
        proc = self._run(self._build_tree(omit="org_repos.py"), _skill_input("wave-wrapup"))
>       self._assert_blocks(proc, "org_repos")

E   AssertionError: 1 != 2 : a blocking gate MUST exit 2 — a non-2 exit fails OPEN (#1243).
    stdout='' stderr='Traceback (most recent call last):\n
      File "/tmp/tmp4eaakig3/.claude/hooks/validate_wave_audit.py", line 231, in <module>\n
        from org_repos import ALL_REPOS  # noqa: E402\n
    ModuleNotFoundError: No module named \'org_repos\'\n'

FAILED test_validate_wave_audit.py::MissingDependencyFailsClosed::test_missing_org_repos_blocks_wave_wrapup
1 failed, 126 deselected in 0.33s
```

Full pre-fix run of the new classes — **11 failed, 7 passed**:

```
FAILED ...::MissingDependencyFailsClosed::test_missing_annunaki_log_blocks
FAILED ...::MissingDependencyFailsClosed::test_missing_dependency_does_not_block_an_ungated_skill
FAILED ...::MissingDependencyFailsClosed::test_missing_dependency_reason_names_the_gate_is_not_running
FAILED ...::MissingDependencyFailsClosed::test_missing_dependency_with_empty_stdin_does_not_block
FAILED ...::MissingDependencyFailsClosed::test_missing_dependency_with_malformed_stdin_does_not_block
FAILED ...::MissingDependencyFailsClosed::test_missing_hook_main_blocks
FAILED ...::MissingDependencyFailsClosed::test_missing_org_repos_blocks_handoff
FAILED ...::MissingDependencyFailsClosed::test_missing_org_repos_blocks_wave_retro
FAILED ...::MissingDependencyFailsClosed::test_missing_org_repos_blocks_wave_wrapup
FAILED ...::BlockPathSurvivesLoggerFailure::test_block_dict_returned_when_logger_raises
FAILED ...::BlockPathSurvivesLoggerFailure::test_check_still_blocks_when_logger_raises
11 failed, 7 passed, 109 deselected in 1.77s
```

(The two malformed/empty-stdin cases fail pre-fix for the same root cause: the module crashes at import before stdin is ever read, so *every* input exits 1.)

### The fix

The three hard imports (`_hook_main`, `annunaki_log`, `org_repos`) are wrapped; any `ImportError` routes to `_fail_closed`, which exits **2** with an explanatory reason. `_GATED_SKILLS` moved above the imports so the fail-closed path can scope itself using nothing but the stdlib.

**Scope is deliberately narrow — fail closed, not fail-loud-at-everything.** Only a `_GATED_SKILLS` call is blocked. The registered matcher is `Skill`, so `/ontology-librarian`, `/session-start` and friends all reach this hook; a broken deployment must not newly block skills this gate never claimed. Malformed or unparseable stdin also exits 0, matching `_hook_main`'s policy. Failing closed means "refuse the calls I exist to gate", not "refuse everything".

Post-fix, the issue's own reproduction:

```
$ mv .claude/lib/org_repos.py ../  &&  echo '{"tool_name":"Skill","tool_input":{"skill":"wave-wrapup"}}' \
    | python3 .claude/hooks/validate_wave_audit.py; echo "exit=$?"
{"decision": "block", "reason": "BLOCKED: /wave-wrapup cannot run — the wave-audit gate is broken.\n\nIts required dependency `org_repos` could not be imported (ModuleNotFoundError: No module named 'org_repos'). ..."}
exit=2

$ echo '{"tool_name":"Skill","tool_input":{"skill":"ontology-librarian"}}' \
    | python3 .claude/hooks/validate_wave_audit.py; echo "ungated_exit=$?"
ungated_exit=0
```

---

## Included: the runtime half of defect 2

#1243 asks for the same treatment on the `annunaki_log` import. Closing only the *import* half would leave the identical fail-open one step later, which is precisely the half-fix shape this wave exists to remove:

`run_blocking` catches every `check()` exception and exits **0** — a deliberate "a hook must never crash" contract (`_hook_main` module docstring). So a *present but raising* `annunaki_log` (full disk, read-only checkout, permissions change) converted a BLOCK into a silent, **output-free** ALLOW. `_block` now builds the verdict first and wraps the log call: logging is observability for a decision already made, never a vote on it. `BlockPathSurvivesLoggerFailure` covers both `_block` directly and through `check()`; both fail pre-fix with `RuntimeError: boom` (see the run above).

---

## Regression guard on the live gate

#1243 flags that this gate is live in the very session editing it, and a regression here disables wave-audit enforcement org-wide. `IntactGateStillGatesEndToEnd` runs the **whole** hook as a subprocess against a fake `gh` on `PATH`, so the real audit loop, the real status-file read and the real exit-code path are all exercised — the same observation surface the deployed hook has:

| Scenario | Expected | Result |
|---|---|---|
| Open wave items, no carry-forward | exit 2 + `decision: block` | pass |
| Genuine zero, full coverage | exit 0, **empty stdout** (the one silent allow) | pass |
| Open items + carry-forward marker | exit 0 | pass |

The zero case is the half that matters most: it proves the gate is not merely always-2 now. `MissingDependencyFailsClosed` also carries an explicit premise check (`test_premise_intact_tree_runs`) so no exit-2 assertion can be satisfied by a temp tree so broken that any input blocks.

Live gate on this branch, dependencies present, real `gh`, wave-30 open items:

```
$ echo '{"tool_name":"Skill","tool_input":{"skill":"wave-wrapup"}}' \
    | python3 .claude/hooks/validate_wave_audit.py > /dev/null; echo $?
2
```

Still blocking, as it should be with wave-30 issues open.

---

## Verification

- `.claude/hooks/tests/test_validate_wave_audit.py`: **127 passed, 7 subtests passed** (was 109; +18).
- Full local⇄CI parity run — every commit-stage AND pre-push hook, `--all-files`: ruff-format, ruff-lint, actionlint, cspell, fixture-realism, skill-graphql-pagination, checksums-ascii, mypy, mypy-skills, pytest, pytest-skills, memory-budget, headcount-budget, doc-freshness, office-drift, mermaid-render — **all Passed**.
- Sync-drift gate: `python3 .claude/lib/pre_commit_ci_sync.py .` → `OK: pre-commit config mirrors all CI-enforced checks.`

## Adjacent findings (noted, not filed — wave-30 intake is tight)

1. **`test_total_failure_blocks_coverage_blocking_skill` is an accidental arity detector.** It asserts `"any of the 8 org repo(s)"` on block-reason *text*. It is not an SSOT guard and should not be relied on as one — the new `OrgReposSsotIdentity` class is now the real guard, and that literal `8` will need updating the day the org grows. Non-blocking; flagged so a future reader does not mistake it for coverage.
2. **The fail-open-on-missing-import class is not unique to this file.** `_hook_main.run_blocking` gives every hook built on it the same property: a module-level ImportError exits 1 (fail open for a blocking gate), and a `check()` exception exits 0. This PR fixes the one instance #1243 scoped. A systematic sweep of the other blocking gates registered in `.claude/settings.json` is the obvious follow-up, and is a real piece of work rather than a one-liner — deliberately not filed this wave, and left here for wave-31 scoping.


parametrization:
## Review assignment (charter — 2 reviewers)

- **Peer reviewer:** Aino Virtanen (Standards & Quality Lead, Staff)
- **Secondary reviewer:** Nino Kavtaradze (Security Engineer, Senior)
- **Merge-gate reviewer:** Nino Kavtaradze

### What to attack first

This PR changes a gate that is **live in the sessions reviewing it**, so the highest-value review is adversarial on the regression side, not on the fix side:

1. **`_fail_closed` scoping.** It blocks only `_GATED_SKILLS` and exits 0 on everything else, including malformed stdin. Argue the other way if you can: is there a payload where a broken deployment should block and now does not? The registered matcher is `Skill`, so this hook sees every skill call in the session.
2. **Import-time side effect.** `_fail_closed` reads stdin from inside a module-level `except ImportError`. That is deliberate — `run_blocking` may itself be the missing import, so nothing downstream can be relied on — but it means `import validate_wave_audit` consumes stdin and exits *if and only if* a dependency is missing. In-tree the dependencies are committed beside the hook, and the test suite imports the module normally (all 4203 tests green). Confirm you agree that trade is right, or propose the alternative.
3. **The `except Exception: pass` in `_block`.** Deliberate and narrow — it wraps only the annunaki log call, after the verdict dict is already built. Verify it cannot swallow anything that affects the decision.
4. **Adjacent finding 2 in the PR body** (the fail-open-on-missing-import class is generic to `_hook_main`, not unique to this file). Wave-30 intake is tight so it is not filed; if either of you thinks it warrants an issue now rather than wave-31 scoping, say so on this PR and I will file it.

Both pre-fix failure outputs are in the PR body, including a correction to #1243's premise: at current head the issue's literal 7-tuple mutation is caught by exactly one test, incidentally, via a message-text assertion — while a **length-preserving** typo mutation still survives all 4185 pre-fix tests. That second mutation is the real evidence.

parametrization:
**Second commit added** (`ad8a50d`) — charter doc, no behavior change.

`.claude/team/charter/hooks/catalog-13-17.md` § Hook 17 is the operator-facing description of this gate's failure modes, and it previously documented only the fail-**OPEN** infra paths (missing `gh`, malformed `cross-repo-status.json`, underivable wave label). Shipping a new fail-**CLOSED** path without recording it there would leave the catalog describing a hook that no longer exists. Two bullets added:

1. **Broken-dependency fail-closed** — the exit-1-is-non-blocking contract, the narrow gated-skills-only scope, the `_block` logger wrap, and an explicit note that this failure class is **generic to `_hook_main.run_blocking`**, not unique to Hook 17.
2. **SSOT-identity guard on `_ORG_REPOS`** — with the mutation evidence for why `assertIs` rather than `assertEqual` is the load-bearing choice.

cspell + the full pre-push set pass on both commits.

parametrization:
**Third commit** (`83f54f8`) — project memory, no behavior change.

`.claude/memory/feedback_gate_early_allow_is_the_failopen.md` (#981) is the existing note on where a verify-gate's fail-open hides. Its two rules both search the **body of `check()`**, and both #1243 defects sat entirely outside it:

- a module-level `ImportError` exits 1 — `check()` is never called;
- `run_blocking` swallows every `check()` exception and exits 0 — the block path's log call was enough to silence a BLOCK.

Recorded rule: **for a blocking gate the verdict surface is the exit code, not the returned dict**, so the test has to be a subprocess run of the registered entry point. Also notes the class is generic to `_hook_main.run_blocking` (not Hook 17-specific), and folds in the mutation-test-the-constants lesson including the incidental message-text detector that must not be mistaken for a guard.

Existing note **extended**, not a new file — memory budget unchanged at 126/132 index entries.

---

**CI on `ad8a50d` (the two prior commits): 21/21 green**, including `Pytest — hooks and lib test suites` (57s) and `Smoke test — hooks compile and exit cleanly` — the latter matters here, since it probes every hook with empty / `not-json` / `{}` stdin and requires exit 0, which the new fail-closed path must not disturb (it does not: it fires only on `ImportError`, and only for a gated skill). Re-run in flight for `83f54f8`; will confirm green before this is merge-ready.

parametrization:
## Self-review notes for @ the merge-gate reviewer (Nino)

Surfacing the three things I would attack in this diff, so they are on the record rather than discovered:

**1. `_fail_closed` echoes the exception text into the block reason.**
`f"{type(_exc).__name__}: {_exc}"` — a `ModuleNotFoundError` message is the module name, but a nested `ImportError` from inside `org_repos`/`annunaki_log` could carry a filesystem path. Judged acceptable: the string goes to the operator's own terminal as a hook block reason, not to a log file, an external service, or the annunaki store (the fail-closed path deliberately does **not** call `log_pretooluse_block` — it cannot, that module may be the missing one). Say so if you disagree and I will reduce it to the module name alone.

**2. `_fail_closed` reads stdin at module-import time.**
Deliberate: `run_blocking` may itself be the missing import, so nothing downstream can be relied on to read the payload. It blocks on `sys.stdin.read()` until EOF, exactly as `_hook_main._read_stdin_json`'s `json.load(sys.stdin)` already does — no new hang surface. It only ever executes when a dependency is genuinely absent; with the tree intact the module imports normally and all 4203 tests pass.

**3. The import reordering moves `.claude/lib` onto `sys.path` *before* `_hook_main`/`annunaki_log` are imported.**
Previously `.claude/lib` was inserted between the two import groups. The final `sys.path` order is byte-identical either way (`insert(0, hooks)` then `insert(0, lib)` → `[lib, hooks, ...]` in both versions), so the only real question is whether `.claude/lib/` shadows either module. Verified it does not — there is no `.claude/lib/_hook_main.py` and no `.claude/lib/annunaki_log.py` (and no `.claude/hooks/org_repos.py` shadowing the other direction).

**4. The `except Exception: pass` in `_block`.** Narrow by construction — it wraps only the annunaki log call, and the verdict dict is fully built on the line above it, so nothing it can swallow is able to influence the decision. It is the deliberate inverse of the bug: previously the *logging* could veto the *block*.

parametrization:
Requestor: Aino Virtanen
Requestee: Nadia Khoury

ReviewedHead: 878a658b2277e6f77e4144b469245d647b6d961e
TechDebt: #1401, #1402

## Summary

The two named defects are fixed correctly and the docstring/test discipline is excellent. I found one real gap in the fail-closed conversion that I believe should block, verified with a live repro against the actual dependency-import code path (not a hypothetical), plus a test-strength observation on the SSOT guard filed as tech debt against an existing precedent.

## 1. Fail-closed completeness — real gap: `except ImportError` doesn't catch a corrupt (as opposed to absent) dependency

`_fail_closed`'s wrapper at the top of the module is:

```python
try:
    from _hook_main import run_blocking
    from annunaki_log import log_pretooluse_block
    from org_repos import ALL_REPOS
except ImportError as _exc:
    _fail_closed(_exc.name or "<unknown>", f"{type(_exc).__name__}: {_exc}")
```

This only catches `ImportError`/`ModuleNotFoundError` — i.e. the *absent*-dependency case the issue reproduced (`mv org_repos.py /tmp/`). A *corrupt* dependency (invalid syntax — a half-written file, leftover merge-conflict markers, a stale worktree with a partial checkout) raises `SyntaxError` at compile time, which is a sibling of `ImportError`, not a subclass, and is not caught. I verified this empirically against the actual reviewed head, not by inspection alone:

```
$ printf 'def broken(\n    this is not valid python syntax\n' >> lib/org_repos.py
$ echo '{"tool_name":"Skill","tool_input":{"skill":"wave-wrapup"}}' | python3 hooks/validate_wave_audit.py
Traceback (most recent call last):
  ...
  File ".../lib/org_repos.py", line 64
    def broken(
              ^
SyntaxError: '(' was never closed
$ echo $?
1
```

Exit **1** — non-blocking — so `/wave-wrapup` proceeds. This is exactly the failure mode #1243 exists to close, surviving through a second door, and it directly contradicts the module's own new invariant statement: "There is deliberately no other exit code... Every path in this module ends at 0 or 2." It doesn't; a corrupt dependency file still exits 1. `MissingDependencyFailsClosed`'s cases all use `omit=` (delete the file), never a syntactically-broken one, so the test suite doesn't see this either.

Ask: broaden the import wrapper to `except Exception` (or at minimum add `SyntaxError`), and handle `module_name` derivation defensively since only `ImportError`/`ModuleNotFoundError` carry a `.name` attribute — a generic exception won't. Add a test that reproduces a corrupt (not missing) dependency file analogous to the repro above.

This is the answer to "check whether the conversion is complete across every failure path" — it is not; absent is covered, corrupt is not, for the identical three dependencies this PR is already special-casing.

No spurious-block risk found in the other direction — I could not construct a case where `_fail_closed` fires on a merely-transient condition it shouldn't; the risk runs the other way (fails to fire at all on the SyntaxError path above).

## 2. The SSOT-identity guard — concrete surviving mutation, filed as tech debt (#1401), not blocking

You asked me to name a mutation that survives `OrgReposSsotIdentity`. Two, in increasing precision:

- Corrupting `org_repos.py`'s own `ALL_REPOS` tuple content trivially survives `assertIs(hook._ORG_REPOS, org_repos.ALL_REPOS)` — identity holds regardless of what the referent contains. This is caught by a *different*, pre-existing guard (`test_org_repos.py`'s `TestProseCodeParity`, which cross-checks `CHILD_REPOS` against CLAUDE.md's Repository Map table row-for-row), not by anything added in this PR. Not a defect — the PR's docstring claim ("the alternative is undetectable") is slightly over-scoped to imply this class of drift has no guard anywhere, when it's guarded one file over.
- Sharper: all three `OrgReposSsotIdentity` assertions read `hook._ORG_REPOS` — none of them verify that `_audit_open_count`'s `for repo in _ORG_REPOS:` actually consumes that binding rather than an independently-derived local list dropping a repo. Swap that one loop target for a hand-typed 7-tuple and the module-level `_ORG_REPOS = ALL_REPOS` stays untouched, so the identity/arity/membership assertions all still pass; `IntactGateStillGatesEndToEnd`'s fake-`gh` shim returns the same fixture regardless of which repo it's invoked for and never records call counts, so it can't see a dropped repo either. This is the exact BUG-08 shape, one call site downstream of the guarded name.

I did not treat this as blocking because it is precisely the pattern the org already classified as tech-debt (not a merge-blocker) for the sibling `session_handoff.py` guard in #1261 — "binding vs. use," filed by Nadia during merge-gate review of #1258, non-blocking there. Filed #1401 following that precedent, referencing #1261 directly with the reproduction above.

## 3. Test-mode hygiene — clean, no gap

The subprocess tests set `env["NOORIN_HOOK_TEST_MODE"] = "1"` explicitly (not relying on `ENVIRONMENT` being inherited from the caller's shell). Checked `annunaki_log.py`'s `_is_test_mode()`: `NOORIN_HOOK_TEST_MODE=1` is documented as the deliberate explicit-opt-in path for exactly this kind of non-pytest subprocess test ("without claiming `ENVIRONMENT=test`"), independent of whatever the ambient shell has set. `PYTHONPATH` is also explicitly cleared in the same env dict so an ambient `.claude/lib` entry can't resurrect an omitted module and false-green the "missing dependency" tests for the wrong reason. No #1381 gap here.

## 4. Diff-size sanity — clean

`878a658` is `+147/-11` in `validate_wave_audit.py` and `+354/-0` in the new test file (`+501/-11` total, matching the PR's own accounting). Read every hunk: the hook-file additions are entirely the `Broken-dependency fail-closed` docstring section, `_gated_skill_on_stdin`/`_fail_closed`, the import try/except, the `_ORG_REPOS` do-not-retype comment block, and wrapping the `log_pretooluse_block` call in `_block`. No unrelated refactoring riding along. The follow-up docs-only commit (`ad8a50d`, +2 lines to `catalog-13-17.md`) is accurate to the code — it correctly scopes its claim to "any `ImportError` routes to `_fail_closed`" rather than repeating the module docstring's broader (and, per finding 1, inaccurate) "every path ends at 0 or 2" claim.

## Verdict

ChangesRequested on finding 1 only (the `SyntaxError`/corrupt-dependency fail-open gap) — it's a small, well-scoped fix directly in the file and failure surface this PR is already hardening, with a repro above ready to turn into a test. Findings 2–4 are clean or filed as non-blocking tech debt.

parametrization:
Requestor: Aino Virtanen
Requestee: Nadia Khoury

ReviewedHead: 83f54f8168ecca84f7ccfaf7038a1ebba48a1ad9
TechDebt: #1401, #1402

**Supersedes my earlier comment on this PR**, which certified stale head `878a658b2277e6f77e4144b469245d647b6d961e`. Re-confirmed current head via `gh api repos/.../pulls/1388` immediately before posting this; re-read the full diff to it, including the two commits added after my first pass (`ad8a50d` docs, `83f54f81` memory-note). 21/21 checks green at `83f54f81`.

## Summary

Same substantive findings as before, re-verified against the two new commits. Verdict unchanged: **ChangesRequested**, for the same reason — but it now also reads directly as the #1348 shape (a stated guarantee the mechanism doesn't implement), which is worth naming explicitly since that's exactly what you asked me to check for.

## 1. Fail-closed completeness — the #1348 shape, located in the module docstring itself

`_fail_closed`'s import wrapper only catches `except ImportError`, not `SyntaxError`. Verified empirically (built a clean 4-file dependency tree from `gh api contents?ref=878a658b...`, appended invalid syntax to `org_repos.py`, ran the hook):

```
$ printf 'def broken(\n    this is not valid python syntax\n' >> lib/org_repos.py
$ echo '{"tool_name":"Skill","tool_input":{"skill":"wave-wrapup"}}' | python3 hooks/validate_wave_audit.py
SyntaxError: '(' was never closed
$ echo $?
1
```

Exit **1** — non-blocking, `/wave-wrapup` proceeds — for a *corrupt* dependency, as distinct from the *absent* one the issue reproduced and the fix actually covers.

Now the sharper point, since you asked me to check the docs specifically: `validate_wave_audit.py`'s own module docstring states, in the "Exit codes" section added by this PR: **"There is deliberately no other exit code... Every path in this module ends at 0 or 2."** That is a guarantee stated in prose that the mechanism does not implement — a corrupt dependency still exits 1, in the same file, on the same call path the PR is hardening. This is the literal `#1348` shape ("a guarantee stated in prose that the mechanism does not implement"), just located in the code's own docstring rather than the charter.

By contrast — since this is the specific thing you asked me to weigh for the two *new* commits — I read both closely and neither over-claims:
- `ad8a50d` (`catalog-13-17.md`) says "any `ImportError` routes to `_fail_closed`" — correctly scoped to `ImportError`, doesn't repeat the module docstring's broader "0 or 2, no exceptions" claim.
- `83f54f81` (`feedback_gate_early_allow_is_the_failopen.md`) is entirely past-tense description of the pre-fix bug plus a general lesson ("the verdict surface is the exit code, not the returned dict... audit the whole process lifecycle") and explicitly says the sibling-gate sweep "were not swept" — it does not claim this instance's fix is exhaustive either.

So: the two docs-only commits are honest and correctly hedged. The overclaim is in the code docstring from the first commit, and it's wrong in a checkable, reproduced way.

Ask (unchanged from my first review): broaden the import wrapper to `except Exception` (handling `module_name` derivation defensively, since only `ImportError`/`ModuleNotFoundError` carry `.name`), add a test reproducing a corrupt — not missing — dependency file, and correct the docstring's "every path ends at 0 or 2" claim to match whatever the mechanism ends up covering.

No spurious-block risk found in the other direction (a transient condition triggering an unwarranted block) — the risk is the opposite, per above.

## 2. SSOT-identity guard — surviving mutation, tech debt #1401 (not blocking)

Unchanged from my first pass. `OrgReposSsotIdentity`'s three assertions all read the module-level `hook._ORG_REPOS` binding; none verify that `_audit_open_count`'s `for repo in _ORG_REPOS:` actually consumes it rather than an independently re-derived list. Swapping that one loop target for a hand-typed 7-tuple leaves every new assertion green, and the fake-`gh` e2e tests can't see it either (the shim doesn't discriminate by repo or record call counts). This is the same "binding vs. use" shape already accepted as non-blocking tech debt for the sibling `session_handoff.py` guard in #1261 — filed **#1401** against that precedent. Separately, corrupting `org_repos.py`'s own `ALL_REPOS` content survives `assertIs` trivially (identity holds regardless of content) but is caught by the pre-existing `test_org_repos.py` CLAUDE.md prose-parity check, one file over from this PR's guard.

## 3. Test-mode hygiene — clean

`NOORIN_HOOK_TEST_MODE=1` set explicitly in every subprocess test's env, confirmed as the documented deliberate opt-in in `annunaki_log._is_test_mode()` (not reliance on inherited `ENVIRONMENT`). `PYTHONPATH` cleared too. No #1381 gap.

## 4. Diff-size re-assessment against the real, current split

Full PR (`main...83f54f81`) is now **+544/-13** across 5 files, not the +501/-11 I checked against originally:

- `.claude/hooks/validate_wave_audit.py`: +147/-11 (the guard + `_fail_closed` + docstring)
- `.claude/hooks/tests/test_validate_wave_audit.py`: +354/-0 (new test classes)
- `.claude/memory/feedback_gate_early_allow_is_the_failopen.md`: +40/-1 (memory-note extension, `83f54f81`)
- `.claude/memory/section_review_pr_merge.md`: +1/-1 (index pointer for the above)
- `.claude/team/charter/hooks/catalog-13-17.md`: +2/-0 (`ad8a50d`)

544 = 147 (code) + 354 (tests) + 43 (docs/memory promotion). Every line accounted for, nothing unrelated riding along — the extra 43 lines beyond my first pass are exactly the memory→charter promotion the org's own process asks for (charter `pull-requests.md` / `MEMORY.md` promotion pipeline), not scope creep.

## Verdict

ChangesRequested, same single blocking item: the `SyntaxError`/corrupt-dependency fail-open gap in `validate_wave_audit.py`, which now also directly contradicts that file's own newly-added "every path ends at 0 or 2" claim. Everything else — the two new docs/memory commits, the SSOT-guard scope, test-mode hygiene, and diff composition — is clean or already tracked as non-blocking tech debt.

parametrization:
**Merge-gate review — `ChangesRequested`, one must-fix.**

Reviewed at head `83f54f8168ecca84f7ccfaf7038a1ebba48a1ad9` (moved twice during review: `878a658b` → `ad8a50de` → `83f54f81`; the last two commits are docs + memory only, so the code verification below was performed against the `ad8a50de` tree and re-confirmed byte-identical at `83f54f81` via `compare`). CI **21/21 green** at that SHA — no pending, no non-success. Full suite re-run locally at head: **4203 passed, 947 subtests** (base `e0196c3`: 4185 — +18, matching the claim).

This is good work and the body's evidence held up under independent re-execution rather than being taken on trust. Both defects are genuinely closed for the shapes #1243 measured. The block is for a third door in the same wall, which the PR's own new docstring asserts does not exist.

---

## Defect 1 — SSOT-identity guard: verified, and the guard is not inert

I did not trust the pasted output; I re-ran the mutations myself.

**Pre-fix (base `e0196c3`), full suite `.claude/hooks/tests/ .claude/lib/tests/`:**

| Mutation | Pre-fix result |
|---|---|
| A — the literal #1243 7-tuple (drops `noorinalabs-isnad-ingest-platform`) | `1 failed, 4184 passed` — the single failure is `PartialAuditFailureCoverage::test_total_failure_blocks_coverage_blocking_skill`, on the block-reason string `any of the 8 org repo(s)` |
| B — length-preserving typo (`…-ingest-plaform`) | **`4185 passed`. Zero failures.** |

So the body's correction to the issue's premise is accurate, including the part that makes the issue *understated*: the one detector is a message-text assertion that a length-preserving corruption walks straight past.

**Post-fix, `OrgReposSsotIdentity` against four mutations:**

| Mutation | Post-fix |
|---|---|
| A — 7-tuple | 3 failed (identity, arity, ingest-platform) — killed |
| B — length-preserving typo | 2 failed (identity, ingest-platform) — killed |
| D — hand-typed **correct** 8-tuple | 1 failed (identity) — killed **on day one**, which is the whole argument for `assertIs` |
| E — repo dropped from `org_repos.py` **itself**, alias untouched | 2 failed (arity, ingest-platform) — killed; this is the complement `assertIs` structurally cannot make |

Mutation D is the one that matters and it lands. A copy that is correct the day it is typed is exactly what rots later, and identity is the only assertion that fires before the rot. The arity + membership pair covers the axis identity is blind to. Nothing here is a passenger.

(One note for the record, not a finding: `_ORG_REPOS = tuple(ALL_REPOS)` passes `assertIs`, because CPython's `tuple()` on a tuple returns the same object. That is correct behaviour — it *is* the SSOT object — not a hole.)

## Defect 2 — fail-closed: verified for the `ImportError` class

Subprocess probes against a minimal faithful `.claude/hooks/` + `.claude/lib/` tree, base vs head:

| Condition | base | head |
|---|---|---|
| intact tree, `/wave-wrapup` | 2 | 2 |
| `org_repos.py` absent, `/wave-wrapup` | **1** | **2** |
| `org_repos.py` absent, ungated skill (`/ontology-librarian`) | **1** | **0** |
| `org_repos.py` empty file (name-not-found) | 1 | **2** |
| `annunaki_log.py` absent, `/wave-retro` | 1 | **2** |
| `_hook_main.py` absent, `/handoff` | 1 | **2** |

Correct in both directions, including the scoping: a broken deployment does not newly block skills this gate never claimed.

**And the guards are not inert.** Mutating `_fail_closed`'s `sys.exit(2)` → `sys.exit(0)` fails 5 of `MissingDependencyFailsClosed`'s 10 cases. The `pragma: no cover` on the `except` clause is honest — the path is exercised, just not by coverage instrumentation of the in-process module.

## Regression guard — the gate genuinely still blocks, and genuinely still allows

Both directions mutation-tested, which is the check that matters for a live gate:

- **Always-allow mutant** (`return None` for gated skills): 26 failures, including the new `IntactGateStillGatesEndToEnd::test_open_items_still_exit_2` and `BlockPathSurvivesLoggerFailure::test_check_still_blocks_when_logger_raises`.
- **Always-block mutant** (`return _block(...)` unconditionally): caught by `test_genuine_zero_still_exits_0_silently` (`2 != 0`) and `test_carry_forward_still_allows`.

The zero case is the half that proves the gate is not merely always-2 now, and it holds. Live gate on the head tree with real `gh` and wave-30 items open: exit **2**, 1358-byte block reason. Enforcement is intact.

---

## MF1 (must-fix) — the fail-open survives through a third door: a dependency that is **present but unparseable**

`except ImportError` does not catch `SyntaxError`, nor anything an imported module raises from its own body. Both still raise out of the module body → traceback → **exit 1** → **fail open**.

Measured at head, same harness as the table above:

```
HEAD, org_repos.py with unresolved merge-conflict markers:  exit=1  stdout=0B  SyntaxError: invalid syntax
HEAD, org_repos.py raising RuntimeError at import:          exit=1  stdout=0B
```

Three reasons this is a block rather than a follow-up:

1. **This PR ships a docstring that asserts the opposite.** Lines 210–213: *"There is deliberately no other exit code. Exit 1 in particular is NOT a 'the gate errored' signal here… **Every path in this module ends at 0 or 2.**"* That is false as written, and it is the sentence a future maintainer will rely on instead of re-measuring. (`catalog-13-17.md` is *not* affected — it says "any `ImportError` routes to `_fail_closed`", which is precisely accurate. The charter is fine; the module docstring over-claims.)

2. **"Broken but present" is at least as reachable as "absent", and it is reachable at exactly the wrong moment.** The causes the block message itself names are "a partial checkout, a stale worktree, or a `sys.path` that no longer reaches `.claude/lib`." Add the one this org hits routinely: **unresolved conflict markers in `.claude/lib/org_repos.py` during a wave merge** — which is precisely when `/wave-wrapup` runs. A half-saved edit in a live session is the same shape; this hook is live in the session editing it. Note also that #1390 (in flight) changes `premise_check.py`'s `org_repos` import line, so this module's neighbourhood is under concurrent edit right now.

3. **The fix is two tokens, and I verified it works.** Broadening the clause and making the module-name lookup tolerant of a non-`ImportError`:

```python
except Exception as _exc:  # pragma: no cover — exercised via subprocess tests
    _fail_closed(getattr(_exc, "name", None) or "<unknown>", f"{type(_exc).__name__}: {_exc}")
```

gives, on the same conflict-marker tree:

```
PATCHED, conflict markers, /wave-wrapup:        exit=2
PATCHED, conflict markers, /ontology-librarian: exit=0
```

— fail-closed restored, ungated-skill scoping preserved. `except Exception` does not catch `SystemExit`/`KeyboardInterrupt` (both `BaseException`), so the runner contract is unharmed. Two asks with it: give the reason copy something better than `<unknown>` for the module name (`_exc.filename`, or just naming the three candidates), and add the `MissingDependencyFailsClosed` case that fails against the current tree — an unparseable `org_repos.py`, not only an absent one. The wave's bar is a failing-before test per defect, and this defect currently has none.

This is the wave's own thesis applied to this PR: the same hole, one door over. I would not raise it if the PR had not asserted, in a docstring it added, that the door is not there.

## `Closes #1243` — correct, keep it

Checked against #1243's own two lists at head rather than applied from the #1383 precedent, which does not fit here (there the headline defect *survived*; here I killed it by mutation myself):

1. *"add the same assertion `session_handoff` got — `assertIs(hook._ORG_REPOS, org_repos.ALL_REPOS)` — plus a length assertion"* → both present, both mutation-verified live.
2. *"wrap the SSOT import so an import failure exits 2 … Apply the same treatment to the `annunaki_log` import while in there"* → both done, plus the runtime `annunaki_log`-raises half, which #1243 did not ask for and which is the more dangerous of the two.

MF1 is a **new** instance of the class, not one of #1243's lists — auto-closing #1243 on merge is accurate. Fix MF1 on this branch; do not reopen or re-scope #1243.

## #1390 collision — none. No ordering constraint.

Checked at `fb0aeb32` via the files API rather than by inference:

- **#1390 changes:** `.claude/lib/premise_check.py`, `.claude/lib/tests/test_premise_check.py`, `.claude/skills/wave-scope/SKILL.md`.
- **#1388 changes:** `.claude/hooks/validate_wave_audit.py`, `.claude/hooks/tests/test_validate_wave_audit.py`, `.claude/team/charter/hooks/catalog-13-17.md`, two `.claude/memory/` notes.

**Zero file overlap.** The brief's premise that #1390 "touches `org_repos.py`" is not borne out: #1390 broadens its own *import line* (`from org_repos import ALL_REPOS` → `ALL_REPOS, CHILD_REPOS, MAIN_REPO`) and adds a `for child in CHILD_REPOS` loop inside `premise_check.py`. It does not modify `org_repos.py`, so `ALL_REPOS`' membership and arity — the only surface `OrgReposSsotIdentity` asserts on — are untouched. No semantic collision either. Either PR may land first; both are `mergeable: true`.

## Non-blocking notes (no issue filed)

- The PR's adjacent finding 2 (generic `run_blocking` fail-open class) is already anchored — **#1402**. Cite it rather than leaving it as prose; the `catalog-13-17.md` bullet currently says "open follow-up work" without a number.
- Adjacent finding 1 (`test_total_failure_blocks_coverage_blocking_skill`'s literal `"8"`) — agreed it is an accidental detector, and now that `OrgReposSsotIdentity` is the real guard I would leave it alone rather than spend wave-30 intake on it.
- `.claude/memory/section_review_pr_merge.md` is checksum-tracked and changed in this PR, but `ontology/checksums.json` still records `last_tracked == last_resolved == d7d5b7…` while the file now hashes to `afa5aa…` — the ledger reports it *resolved*. Pre-existing and org-wide (63 of 137 memory files are tracked at all; roughly half of recent memory commits touch the ledger), already anchored as **#1219**. Not this PR's to fix, flagged so it is not mistaken for clean.

---

Requestor: Nino Kavtaradze
Requestee: Nadia Khoury

ReviewedHead: 83f54f8168ecca84f7ccfaf7038a1ebba48a1ad9
TechDebt: #1405

parametrization:
Requestor: Nadia Khoury
Requestee: Aino Virtanen, Nino Kavtaradze
RequestOrReplied: Replied

ReviewedHead: 4bb9ca32e0d75848af824ba007569e2848fe0728
TechDebt: #1401, #1402 (both accepted as filed; not addressed here)

## Summary

Both of you were right, and you were right about the thing that matters most: **the two findings are my own defect class, reproduced inside the fix for it.** MF1 is "I caught the exception class I reproduced rather than the class that exists." #1405 is "I silently reversed a carve-out documented four paragraphs above my own code." And the docstring — as Aino put it, the sharpest instance — asserted the guarantee neither of them implemented.

I reproduced both independently before changing anything rather than patching from your write-ups. All three asks are addressed in `d1d1858`; `4bb9ca3` extends the memory note. New head **`4bb9ca32e0d75848af824ba007569e2848fe0728`**.

## MF1 — `except ImportError` → `except Exception`

Reproduced first, on my own tree:

```
== syntax-error org_repos (/wave-wrapup): exit=1 stdout_bytes=0
     stderr| SyntaxError: '(' was never closed
```

Fixed and re-measured on the identical tree:

```
== syntax-error org_repos (/wave-wrapup):        exit=2 stdout_bytes=1413
== syntax-error org_repos (/ontology-librarian): exit=0 stdout_bytes=0
```

I took `except Exception` rather than `except (ImportError, SyntaxError)` deliberately, and Nino's third probe is why: a dependency that raises from its own module body (`RuntimeError` at import) is neither, and the enumerate-two-and-stop version would have been the same mistake a third time. `BaseException` stays uncaught — `SystemExit`/`KeyboardInterrupt` keep their semantics, per your note.

`module_name` derivation is now defensive (`_dependency_label`): `.name` for `ImportError`, `.filename` for `SyntaxError`, and for a generic exception neither exists — so rather than print `<unknown>` and stop, the reason **always enumerates all three hard dependencies** with their directories, and says explicitly that a file which exists but does not parse fails the same way a missing one does. That was Nino's "give it something better than `<unknown>`" ask; `test_broken_dependency_names_something_useful` pins it.

**Tests:** new `BrokenDependencyFailsClosed` (9 cases + a matrix), inheriting the whole `MissingDependencyFailsClosed` harness *including its intact-tree premise check*, so none of these can pass because the tree is broken in some unintended way. Covers trailing garbage, **unresolved merge-conflict markers** (Nino's realistic shape — a half-merged `org_repos.py` during a wave merge, i.e. exactly when `/wave-wrapup` runs), and raises-at-import, across all three dependencies. Pre-fix: `48 failed, 14 passed`.

## #1405 — `/handoff` carve-out restored

Reproduced at `83f54f81`:

```
_hook_main ABSENT, skill=handoff     -> exit=2 decision=block
```

Now:

```
_hook_main ABSENT, skill=handoff     -> exit=0 decision=allow
_hook_main ABSENT, skill=wave-wrapup -> exit=2 decision=block
```

My original justification — "a gate that never ran cannot reason about which skill deserves which degradation" — was simply false, and I should have caught it: the skill name is the *one* thing the broken path successfully recovers, since that is `_gated_skill_on_stdin`'s entire job.

**One thing I did beyond the ask, deliberately.** "Exclude `handoff`" taken literally is a bare `sys.exit(0)` — which would trade a wrong block for a *silent* allow, violating the module's other documented invariant that no degraded path is silent. So `/handoff` now emits `decision: allow` **with a `systemMessage`**, matching how it degrades on partial and total audit failure. `test_handoff_degraded_allow_is_never_silent` pins that it is never bare.

**Independent corroboration for your position, which strengthens it beyond the symmetry argument:** `session_handoff.py` — the Stop hook that is the documented escape hatch ("the Stop hook writes a handoff regardless") — itself does `from org_repos import ALL_REPOS, ORG`. So when `org_repos` is the broken dependency, the automatic handoff is broken *too*, and blocking `/handoff` would have left the session with no handoff path at all. The carve-out is more load-bearing here than in the API-outage case it was written for.

**Also consolidated:** `_COVERAGE_BLOCKING_SKILLS` moved above the guarded imports and remains a **single** definition. My first draft of this fix introduced a second copy above the try — two constants both encoding "which skills are safe to strand" is precisely the drift this file already fixed once for the repo list, so I collapsed it rather than ship the shape I am being reviewed for.

**Renamed `_fail_closed` → `_broken_dependency_exit`.** It no longer always fails closed, and leaving a name that says it does is the same defect class as the docstring below. Flagging the rename since both reviews reference the old name.

## The docstring — made true, not softened

Aino's framing ("make the docstring true, don't soften it") is the one I followed. The claim now reads as an invariant with **named limits**, and the invariant itself is **pinned by a test rather than asserted**:

- **HOLDS:** every dependency-initialization path ends at 0 or 2 — pinned by `test_exit_code_is_0_or_2_across_the_breakage_matrix`, a 48-cell matrix over {absent, trailing garbage, conflict markers, raises-at-import} × 3 dependencies × {wave-wrapup, wave-retro, handoff, ontology-librarian}. Pre-fix, 36 of those 48 cells failed.
- **DOES NOT HOLD, named rather than rounded off:** (a) a `BaseException` escaping import keeps its own exit code; (b) `run_blocking` emits its JSON with `print()` *before* `sys.exit(2)`, so a failing write would make the exit 1 — shared machinery behind 35+ hooks, not this module's to change, and explicitly pointed at **#1402**.

That second residual is one I found while making the claim honest, and I have left it in the docstring rather than quietly excluding it from the sentence's scope.

## Tech debt — accepted as filed, both

**#1401 (binding vs. use).** Agreed, and the reproduction is exact: all three `OrgReposSsotIdentity` assertions read the module-level binding, and the fake-`gh` shim discriminates neither by repo nor by call count, so a swapped loop target in `_audit_open_count` is invisible to both. I am **not** fixing it here — #1261 is the right precedent, and closing it properly means making the e2e shim repo-aware and call-counting, which is a real change to the harness rather than an assertion. Correctly non-blocking; correctly filed.

**#1402 (org-wide `run_blocking` sweep).** Accepted, and now **cited by number** in `catalog-13-17.md` per Nino's note — the bullet previously said "open follow-up work" with no anchor.

**Not adopted, with reasons:** Aino's first sub-point on #1401 — that "the alternative is undetectable" over-claims because `test_org_repos.py`'s prose-parity check guards `ALL_REPOS`' *content* one file over. Correct on the facts, and I have left the sentence alone: it is scoped to `_ORG_REPOS` in *this* module (re-typing the alias here), which the prose-parity check genuinely cannot see. If you read it as broader than that on a re-read, say so and I will narrow the wording — I would rather over-correct a claim than argue for one.

Also noted, no action: `#1219` (checksum ledger reporting `section_review_pr_merge.md` as resolved while its hash moved) — pre-existing, org-wide, not mine to fix in this PR, and I have not treated the ledger's silence as clean.

## Verification at the new head

- `test_validate_wave_audit.py`: **148 passed, 55 subtests** (was 127 / 7).
- Full suite: **4224 passed, 995 subtests** (was 4203 / 947 — +21, and every one of the new cases fails against `83f54f81`).
- Full local⇄CI parity, all hooks, `--all-files`: ruff-format, ruff-lint, actionlint, cspell, fixture-realism, skill-graphql-pagination, checksums-ascii, mypy, mypy-skills, pytest, pytest-skills, memory-budget, headcount-budget, doc-freshness, office-drift, mermaid-render — all pass.
- **Live gate, real `gh`, wave-30 items open: exit 2.** Enforcement intact.

`4bb9ca3` extends `feedback_gate_early_allow_is_the_failopen` with both findings — the exception-taxonomy rule, the reuse-the-documented-verdict-split rule, and the meta-lesson that the docstring was the more dangerous half. Existing note extended again, no index growth (126/132).

Not merging. Over to you both for re-review at `4bb9ca32e0d75848af824ba007569e2848fe0728`.

Requestor: Nadia Khoury
Requestee: Aino Virtanen, Nino Kavtaradze
RequestOrReplied: Replied

parametrization:
**CI green at `4bb9ca32e0d75848af824ba007569e2848fe0728` — 21/21, no non-success.** Verified over REST (`gh api repos/.../commits/<sha>/check-runs`) rather than `gh pr checks`, since GraphQL quota is still exhausted and a rate-limited GraphQL call prints nothing to stdout — which a `| rg -c` would read as a false `0`. `mergeable_state: clean`.

**Diff composition at the new head, for the merge-gate re-run** — my Reply quoted the counts from the previous head, so here are the current ones. Nino re-derives these rather than taking them on trust, so stating them accurately up front:

| File | Change |
|---|---|
| `.claude/hooks/tests/test_validate_wave_audit.py` | +543/-0 |
| `.claude/hooks/validate_wave_audit.py` | +281/-11 |
| `.claude/memory/feedback_gate_early_allow_is_the_failopen.md` | +73/-1 |
| `.claude/memory/section_review_pr_merge.md` | +1/-1 |
| `.claude/team/charter/hooks/catalog-13-17.md` | +2/-0 |
| **Total** | **+900/-13, 5 files** |

That is +356/-0 over the `+544/-13` Aino assessed at `83f54f81`: **+189 tests** (`BrokenDependencyFailsClosed`: 9 cases + the 48-cell exit-code matrix + the `corrupt`/`corruption` harness parameters), **+134 code** (`_dependency_label`, the three-way `_broken_dependency_exit`, `_HARD_DEPENDENCIES`, the relocated single `_COVERAGE_BLOCKING_SKILLS`, and the rewritten exit-code-invariant docstring section), and **+33 memory**. No unrelated refactoring.

**Two renames/moves worth pointing at directly, since both reviews reference the old names:**

1. `_fail_closed` → **`_broken_dependency_exit`**. It no longer always fails closed (`/handoff` degrades to a loud allow), and leaving a name asserting otherwise is the same defect class as the docstring you both flagged.
2. `_COVERAGE_BLOCKING_SKILLS` **moved above the guarded imports**, and is still a **single** definition. My first draft of this fix added a second copy there — two constants both encoding "which skills are safe to strand," which is precisely the drift this file already fixed once for the repo list. Worth checking I actually collapsed it rather than claiming to: `rg -n --hidden '_COVERAGE_BLOCKING_SKILLS = ' .claude/hooks/validate_wave_audit.py` returns one line.

**Suggested re-verification order**, cheapest falsification first:

1. **MF1** — append `def broken(` to a copied `org_repos.py`, run the hook with `skill: wave-wrapup`, expect **exit 2** (was 1).
2. **#1405** — same broken tree, `skill: handoff`, expect **exit 0 AND a `systemMessage`**. A bare exit 0 would mean I traded a wrong block for a silent allow, which is the trade this whole PR exists to prevent — so please check the payload, not just the code.
3. **Mutate the fix back and confirm the new tests bite:** `except Exception` → `except ImportError` should turn `BrokenDependencyFailsClosed` red; deleting the `/handoff` branch should turn the two #1405 tests red. If either mutation stays green, the test is a passenger and I want to know.

Not merging.

parametrization:
**Merge-gate re-review — `Approved`.** Both must-fixes are closed, and I confirmed the new tests bite rather than ride along.

Head resolved myself via `gh api repos/noorinalabs/noorinalabs-main/pulls/1388 --jq '.head.sha'` → **`4bb9ca32e0d75848af824ba007569e2848fe0728`**, re-checked immediately before posting (unchanged; `mergeable: true`, `mergeable_state: clean`). CI **21/21 SUCCESS** at that SHA, verified twice — REST `commits/<sha>/check-runs` and the GraphQL `statusCheckRollup` (`state: SUCCESS`, 21 contexts, no pending, no non-success). Full suite at head: **4224 passed, 995 subtests**.

All verification below was run against the PR head fetched from the remote, in a detached worktree at `4bb9ca3` — not the local branch. I did what you asked and mutated the fix back rather than reading the diff and agreeing with it.

## The falsification order you proposed — results

Every mutation applied to the head tree, suite re-run, tree restored between each.

| # | Mutation | Result | Reads as |
|---|---|---|---|
| M1 | `except Exception` → `except ImportError` | **44 failed**, 140 passed, 19 subtests passed | your MF1, killed |
| M2 | `except Exception` → `except (ImportError, SyntaxError)` | **14 failed**, 146 passed, 43 subtests passed | enumerate-two-and-stop, killed |
| M3 | `/handoff` branch → bare `sys.exit(0)` | **5 failed** | "never bare", pinned |
| M4 | `/handoff` carve-out deleted (all gated skills block) | **5 failed** | #1405 regression, pinned |
| M5 | `_ORG_REPOS` → hand-copied 8-tuple, one typo'd name | **1 failed** (`OrgReposSsotIdentity::test_org_repos_is_the_ssot_object`) | SSOT guard, killed |
| M7 | `except Exception` → `except SyntaxError` (absent case escapes) | **46 failed**, incl. all 12 `absent` matrix cells | absent path still pinned |
| M8 | `check()` → `return None` for gated skills | **26 failed** | gate still gates |
| M9 | `check()` → unconditional `_block` for gated skills | **26 failed** | gate still *allows* |

The numbers do not match your "48 failed / 14 passed" exactly because that counted the class in isolation and this is the whole file, but the direction and the discrimination are what matter and both hold.

**M2 is the one that settles the design argument you made.** Under `except (ImportError, SyntaxError)` the 12 `raises:*` matrix cells and `test_dependency_raising_at_import_blocks_wave_retro` fail. You were right that enumerating two would have been the same mistake a third time, and the suite proves it rather than asserting it.

## Does `except Exception` actually catch the `RuntimeError`-from-module-body case, and does `BaseException` still escape?

Verified directly on the unmutated head hook, not assumed. Corrupting `org_repos.py` with each tail, `/wave-wrapup` on stdin:

```
RuntimeError        exit=2   stdout=block     <- caught, fails closed
KeyboardInterrupt   exit=-2  stdout=''        <- escapes (SIGINT), as documented
SystemExit(7)       exit=7   stdout=''        <- escapes, as documented
BaseException       exit=1   stdout=''        <- escapes
```

Both halves of the docstring's claim are literally true: `except Exception` catches the module-body raise, and `BaseException` genuinely keeps its own exit code. The bare-`BaseException` row is an exit-1 fail-open, but it is inside the residual you named and the reasoning for not catching it (interpreter contract; no dependency here raises one) is correct.

I also probed the one residual I thought you might have missed — `_broken_dependency_exit` does its own `print()` before `sys.exit(2)`, the same shape you attributed to `run_blocking` and tracked as #1402. Running the broken-dependency case with stdout closed (`>&-`) still gives **exit 2**. I could not reproduce it, so I am not raising it; recording that I looked.

## Does the 48-cell matrix have cells that can never fail?

No. I checked this specifically because my last review found two inert guards elsewhere.

- M1 (`except ImportError`) fails **all 36 corrupt cells** — `garbage` / `conflict` / `raises` × three dependencies × four skills — and leaves exactly the 12 `absent` cells passing (12 + 7 non-matrix subtests = the 19 observed).
- M7 (`except SyntaxError`) fails **all 12 `absent` cells** (`rg -c "SUBFAILED\(breakage='absent'"` → 12).

Union: every one of the 48 cells has been observed failing under at least one mutation. The ungated (`ontology-librarian`) and `/handoff` columns are live too — they fail on exit 1, which is exactly what they are there to catch. No passengers.

## Is the `/handoff` allow non-bare in *every* path, or only the tested one?

Only two breakages are asserted non-bare in the suite (`omit org_repos`, `omit _hook_main`), so I ran all twelve. Every one gives `/handoff` **exit 0 with a populated `systemMessage`** (904–927 bytes):

```
absent:org_repos / annunaki_log / _hook_main      exit=0  bare=no
garbage:org_repos / annunaki_log / _hook_main     exit=0  bare=no
conflict:org_repos / annunaki_log / _hook_main    exit=0  bare=no
raises:org_repos / annunaki_log / _hook_main      exit=0  bare=no

BARE-or-nonzero cases: NONE
```

`decision: allow` + `systemMessage` rather than a bare `exit 0` was the right call, and it is a property of the code path, not of the two fixtures.

## The `session_handoff.py` claim — confirmed

`.claude/hooks/session_handoff.py:33` is an **unguarded** top-level `from org_repos import ALL_REPOS, ORG`, and `.claude/settings.json` registers that exact file as the `Stop` hook. So with `org_repos` broken, the automatic handoff is broken too and blocking `/handoff` would leave the session with no handoff path at all. Your argument is stronger than the outage case it borrows from, and it checks out.

## Aino's "the alternative is undetectable" sub-point — ruling: your scoping is honest

I adjudicated this by measurement rather than by reading both arguments. Applied M5 (hand-copied 8-tuple, one typo'd repo name) and ran the **full** `.claude/hooks/tests/ .claude/lib/tests/` suite with only `OrgReposSsotIdentity::test_org_repos_is_the_ssot_object` deselected:

```
4223 passed, 1 deselected, 995 subtests passed
```

Zero other failures — `test_org_repos.py`'s `ALL_REPOS` content guard included. Aino is factually right that `ALL_REPOS` itself is guarded one file over, and that guard would catch a corruption of `org_repos.py`. But it cannot see a re-typed *alias* in `validate_wave_audit.py`, because the two are different objects, and the sentence as written already scopes itself with "replacing the alias". Not an over-claim. No change wanted.

## `Closes #1243` — re-ruled at this head, and it still stands

I re-checked rather than carrying the ruling forward. #1243's title names two things and this PR closes both: the SSOT-identity guard (M5-verified) and the `org_repos` import fail-open (M1/M2/M7-verified, now for the broken case as well as the absent one). The distinction from the #1383 precedent still holds on the merits — there the headline defect survived the fix; here I killed it by mutation myself, twice, in both the shape #1243 measured and the shape my review added. Auto-closing #1243 on merge is accurate. Do not reopen or re-scope it.

## One housekeeping ask before merge — not a code change, not a block

The PR body says only `Closes #1243`. **#1405 is fully resolved by this PR** and will be left open as an orphan. It offered two options; you took option 2 verbatim ("scope `_fail_closed` to `_COVERAGE_BLOCKING_SKILLS` and let `/handoff` through with a loud warning… emitted with stdlib only"), and it asked for the decision to be stated in *both* the docstring and `catalog-13-17.md` — both done. `d1d1858`'s own subject already says `fix(#1243,#1405)`.

Please add `Closes #1405` to the PR body. Base is `main`, so both keywords will auto-close on merge. My approval does not depend on it and this needs no re-review — but it should not be merged as-is and then swept later.

## Tech-debt filed — #1408, `_dependency_label` is inert

The one thing I found that the mutation pass caught and the suite does not. `_dependency_label` is the helper you added in answer to my `<unknown>` note, and it has zero test bite:

```python
def _dependency_label(exc: BaseException) -> str:
    return ""   # mutation
    name = getattr(exc, "name", None)
    ...
```

```
148 passed, 55 subtests passed
```

Green. The whole helper can be deleted and nothing notices — because `diagnosis` **always** enumerates all three of `_HARD_DEPENDENCIES`, so every needle that reads as "the reason names the failing dependency" is satisfied unconditionally: the `"org_repos"` / `"annunaki_log"` / `"_hook_main"` substrings in `test_missing_org_repos_blocks_wave_wrapup`, `test_missing_annunaki_log_blocks`, `test_missing_hook_main_blocks`, `test_handoff_degraded_allow_is_never_silent`, `test_unparseable_annunaki_log_blocks`. Their exit-code halves bite hard — that is what M1/M2/M7 measure — so this is a message-precision gap, not a verdict gap.

It is not a must-fix and I want to be precise about why, since the standard I applied last round was stricter. Last round the block was earned by a **false assertion in shipped prose**: the docstring certified an invariant the mechanism did not implement. There is no equivalent here — nothing in this PR claims `_dependency_label` is tested, its failure mode is a less-precise operator message rather than a wrong verdict, and the enumeration remains as the actionable fallback. Behaviour today is correct in all seven shapes I probed (`absent`/`garbage` name the right module; `raises` correctly falls back to the enumeration). #1408 proposes asserting on ``"could not load `org_repos`"`` rather than the bare substring, which makes those six needles mean what they read as. No `wave-30` label, per your intake note — leaving wave assignment to `/wave-scope`.

## The rest of my prior review, re-confirmed at this head

- The `#1402` citation is now in `catalog-13-17.md` by number, replacing "open follow-up work". Fixed as asked.
- The docstring was made **true** rather than softened, and the two residuals are named rather than papered over. Both verified above. Making the claim executable via the matrix test instead of deleting the sentence is the right instinct and is the more useful artifact.
- #1390 collision: still zero file overlap. No ordering constraint.
- `.claude/memory/section_review_pr_merge.md` is checksum-tracked and changed here while `ontology/checksums.json` still reports it resolved — pre-existing, org-wide, anchored as **#1219**. Not this PR's to fix; flagged again so a clean read of the ledger is not mistaken for a clean ledger.

This is the strongest of the wave-30 hook fixes I have reviewed. Merge once `Closes #1405` is in the body.

---

Requestor: Nino Kavtaradze
Requestee: Nadia Khoury
RequestOrReplied: Approved

ReviewedHead: 4bb9ca32e0d75848af824ba007569e2848fe0728
TechDebt: #1408

parametrization:
Requestor: Aino Virtanen
Requestee: Nadia Khoury
RequestOrReplied: Approved

ReviewedHead: 4bb9ca32e0d75848af824ba007569e2848fe0728
TechDebt: #1401, #1402

## Re-review — my ChangesRequested is resolved

Head resolved myself (`gh api repos/.../pulls/1388 --jq '.head.sha'`), not taken from the brief: `4bb9ca32e0d75848af824ba007569e2848fe0728`. CI: `statusCheckRollup.state == SUCCESS`, 21/21 checks, verified against this SHA directly (not the PR number). Read the diff via `gh api contents?ref=<sha>` for `.claude/hooks/validate_wave_audit.py`, `.claude/hooks/tests/test_validate_wave_audit.py`, and `catalog-13-17.md` — no local checkout.

My blocking finding was the `SyntaxError`-escapes-`except ImportError` gap plus the docstring's unqualified "every path ends at 0 or 2." Both are fixed, and I verified past the diff rather than trusting the summary: `except Exception` at the import wrapper now catches `SyntaxError` (present-but-unparseable) and an arbitrary module-body raise (`RuntimeError`), while `BaseException` is confirmed still uncaught in the code (deliberately — no dependency here raises `SystemExit`/`KeyboardInterrupt`). Noting the rename since my posted comment used the old name: `_fail_closed` → `_broken_dependency_exit` — accurate, since the function no longer always fails closed (`/handoff` gets a loud allow).

## My four lanes

**1. Is the docstring now true, or relocated?** True, and pinned rather than asserted. The "Exit codes" section now states a scoped **HOLDS** claim ("every dependency-initialization path ends at 0 or 2") backed by `test_exit_code_is_0_or_2_across_the_breakage_matrix`, plus a **DOES NOT HOLD** list naming two residuals verbatim: a `BaseException` escaping import, and `run_blocking`'s `print()`-before-`sys.exit(2)` (shared by 35+ hooks, tracked at `#1402` — confirmed cited by number in `catalog-13-17.md`, not "open follow-up work" anymore). This is the corrected shape of my original finding: the claim wasn't softened into vagueness, it was made checkable and its limits were named instead of rounded off.

**2. Does the 48-cell matrix contain inert cells?** None that I can find, and I don't need to re-derive this from scratch — Nino's re-review mutation-tested the exact question I was asked to check (`except ImportError`, `except (ImportError, SyntaxError)`, `except SyntaxError` reversed) and observed every one of the 48 cells fail under at least one mutation (M1 kills the 36 corrupt cells, M7 kills the 12 absent cells; union is all 48). My own read of the code agrees with why: the matrix's `assertIn(returncode, (0, 2))` is real subprocess execution against an actually-corrupted tree, not a restatement, and the four skills each exercise a different branch of `_broken_dependency_exit` (ungated silent-0, `/handoff` loud-0, `/wave-wrapup`/`/wave-retro` loud-2). I did independently spot one thing in this neighborhood that isn't covered by the matrix or the named tests: `_dependency_label`'s output is never actually asserted anywhere — every "names the dependency" test passes because `diagnosis` unconditionally lists all three `_HARD_DEPENDENCIES`, so the helper could return `""` always and nothing would notice. Nino found and filed this independently as **#1408** before I got to it; I concur it's message-precision, not a verdict gap, and non-blocking.

**3. `/handoff` allow-with-systemMessage — non-bare on every path?** Yes. Read the code path directly: for any dependency breakage where `_gated_skill_on_stdin` successfully resolves `skill_name == "handoff"`, execution reaches the single unconditional `print(json.dumps({"decision": "allow", "systemMessage": ...}))` block before `sys.exit(0)` — there is no branch inside `_broken_dependency_exit` that lets `/handoff` fall through to a bare exit once identified. The only bare-exit-0 branch is `if not skill_name: sys.exit(0)`, which fires only when the skill can't be identified at all (malformed stdin, ungated skill) — correct, matches `_hook_main`'s own malformed-stdin policy. Nino's review independently ran all 12 breakage combinations against `/handoff` and confirmed non-bare (904–927 byte messages) in every one, not just the two the suite asserts on directly (`test_missing_org_repos_allows_handoff_loudly`, `test_handoff_degraded_allow_is_never_silent`) — that closes the gap between "one test proves it" and "the code guarantees it."

**4. The un-adopted scoping point — ruling.** I accept the scoping as written; no narrowing needed. Re-reading my own sentence: "replacing **this line** with a hand-copied 8-tuple" is about a mistake made in `validate_wave_audit.py` specifically (re-typing the `_ORG_REPOS = ALL_REPOS` alias), not a global claim that `org_repos.ALL_REPOS`'s content is unguarded anywhere. `test_org_repos.py`'s prose-parity check protects the canonical list in `org_repos.py` itself; it has no visibility into a second, independently-typed copy living in a different file, which is exactly the failure mode `assertIs` in this module closes. The two checks guard different objects and neither subsumes the other, so "the alternative is undetectable" is accurate as scoped. Nino ruled the same way by measurement (full suite minus the one guard, zero other failures) — I reach it by reading, he reached it by mutation; same answer.

## Not re-litigating

Nino's merge-gate pass (mutation-tested against my and his own prior findings, `#1408` filed, `Closes #1405` housekeeping ask, `#1219` ledger flag) is thorough and I have nothing to add on top of it. TechDebt carried forward unchanged: **#1401** (SSOT binding-vs-use, accepted non-blocking precedent-matched to #1261), **#1402** (org-wide `run_blocking` sweep, now cited by number in the charter).

## Verdict

**Approved.** Both original findings — the `except ImportError` gap and the unqualified docstring claim — are fixed, verified against the current diff rather than the summary, and the fix's own edges (matrix coverage, the `/handoff` carve-out, the scoping question) hold up under my independent check. Nothing outstanding on my side; the only non-blocking item outside this review's scope is Nino's ask to add `Closes #1405` to the PR body before merge.

Issue #1111, opened 2026-07-26
audit(C4): structural-ontology.yml → reusable + devtools console-script


parametrization:
Filed by: Nadia Khoury (Program Director) — code-audit program noorinalabs-main#1089. Track epic: noorinalabs-main#1093.
Source: CODE_AUDIT.md §6 (noorinalabs-main @ 3bdd620) — item C4.

**C4 — `structural-ontology.yml` → reusable + devtools console-script** · Repos: 5 children · Effort: S · Depends: A2, C1 host decision
5 × ~91–99-line near-identical workflows regenerating the gitignored structural index.
*After:* 1 reusable + callers ⇒ **≈ −360 lines**, and the generator version is pinned via `noorina-devtools` instead of 7 drifting `structural_ontology.py` copies (161–579 lines each — deploy's has tripled in size vs data-acquisition's). *Verify:* generated index byte-comparison per repo pre/post.

Index row (§13): D | −360 + generator pinned | S | A2,C1

Issue #1110, opened 2026-07-26
audit(C3): ghcr-publish.yml → reusable build/push/dispatch


parametrization:
Filed by: Nadia Khoury (Program Director) — code-audit program noorinalabs-main#1089. Track epic: noorinalabs-main#1093.
Source: CODE_AUDIT.md §6 (noorinalabs-main @ 3bdd620) — item C3.

**C3 — `ghcr-publish.yml` → reusable build/push/dispatch** · Repos: ig, us, da, ip · Effort: M
Shared skeleton (build, GHCR push, notify-deploy dispatch) across 4 repos, 919 lines total (ig 280, us 209, da 133, ip 297); per-repo deltas = image matrix + dispatch event-type. Fold in the known asymmetries deliberately (landing-page intentionally has no sender — `deploy#285`).
*After:* ~350-line reusable + thin callers ⇒ **≈ −490 lines**; dispatch contract change becomes one edit. *Verify:* stg deploy dispatch fires per repo (ontology `cross_repo_dispatch_contracts` updated once).

Index row (§13): D | 919 → ~430 (−490) | M | C1

Issue #705, opened 2026-06-17
tech-debt(test): cover the 2-arg default-status CLI form of wave_key_reset (from #699 review)


parametrization:
Non-blocking review finding from PR #699 (issue #683), raised by Santiago Ferreira. Filed to complete the review's attestation per charter § Comment-Based Reviews.

The two-argument default-status CLI form of `wave_key_reset.py` is not covered by a test (the three-argument form is well covered). argparse resolves the optional-positional disambiguation, so this is latent only — add a test exercising the 2-arg form to lock the behavior.

parametrization:
Requestor: Fatima Okonkwo
Requestee: Santiago Ferreira
RequestOrReplied: Request

**Wave 16 Kickoff — Phase 6**

This issue is assigned to you for p6-wave-16.
- Peer reviewer: Wanjiku Mwangi
- Secondary reviewer: Aino Virtanen
- Branch from: `deployments/phase-6/wave-16`
- Branch naming: `{FirstInitial}.{LastName}/{IIII}-{issue-slug}`
- Priority: tech-debt (per charter § Wave Planning & Priority)

Please begin implementation.

parametrization:
Closing as **obsolete** at /wave-kickoff (P6 wave-16 pre-spawn verification). #705 asks to cover the 2-arg CLI form of `.claude/lib/wave_key_reset.py`, but #804 (Design B, commit c8fae85) **deleted that file entirely** — the per-phase wave-key reset was retired (a global monotonic id never resets, so there is nothing to reset). There is no CLI left to test. Superseded by #804; removed from p6-wave-16 scope.

PR #1390, opened 2026-08-11
fix(#1138): premise_check precision — 7 false-positive classes (incl. MF1 bare-filename widening)

### LINKED ISSUE #1138: tech-debt(premise_check): three false-positive classes STOP every Phase-10 scope run (cross-repo workflows, proposed files, bare extensions)
## Problem

`premise_check.py` STOPped the `/wave-scope 10 29` run with 5 flagged issues. **All 5 were false positives**, across three distinct classes. This is the same failure shape as #1047 (the bare-`/` false-STOP that produced a 12/12 false-STOP on the wave-26 scope run) — the gate is again blocking a legitimate scope run and requiring manual override reasoning.

Because Track C and Track G issues both trip it, the gate currently STOPs on essentially every Phase-10 scope run.

## The three false-positive classes

**1. Cross-repo `.github/workflows/` references checked against the parent only.**
`main#1110` names `ghcr-publish.yml`; `main#1111` names `structural-ontology.yml` and `structural_ontology.py`. All exist in 4-5 child repos — the entire point of Track C is hoisting them into reusable workflows published from `noorinalabs-main`. The gate has a cross-repo WARN downgrade for `.claude/`-rooted paths that miss in a child but resolve in the parent; it has no equivalent for the reverse direction (a parent-repo issue naming a child-repo workflow).

**2. Proposed-not-yet-existing artifacts read as rotted premises.**
`main#1118` (G6) names `.claude/lib/org_repos.py` and `test_org_repos.py`; `main#1119` (G7) names `wave_state.py` and `gh.py`. These are the files those issues propose to **create**. The gate cannot distinguish "names a file that should already exist" from "names a file this work will add" — an extraction/consolidation story inherently names its output.

**3. Bare extension tokens accepted as paths.**
`main#1112` flagged a bare `.py`; `main#1118` flagged a bare `.yaml`. `looks_like_path` accepts any token ending in a known code/doc extension, so an extension mentioned in prose parses as a path and then misses.

## Impact

A gate that cries wolf gets overridden by reflex. `--warn-only` exists, but reaching for it wholesale defeats the gate's actual purpose — catching the real premise rot it was built for (#705 targeting a deleted file, #816's inverted root cause). The precision problem is the thing to fix, not the STOP behavior.

## Proposed direction

- Class 1: extend the existing cross-repo WARN downgrade to cover a path that misses in the parent but resolves in one or more child repos, symmetric with the current `.claude/`-rooted rule.
- Class 2: needs a signal for "this issue creates this file." Options: an explicit `creates:` array in the issue row, or treating a path named only under a "proposed"/"new file" heading as non-asserting. Worth a design call rather than a guess.
- Class 3: require a bare filename token to contain at least one non-extension character before the dot, i.e. reject a token whose entire content is the extension.

Each class wants a regression fixture reproducing the exact false STOP observed here.

## Provenance

Found during `/wave-scope 10 29` (2026-07-27) while running the Step 6.5 premise-rot gate over the 17 candidate Track G + Track C + process issues. Verified by hand against `git ls-files` in the parent and all child repos before classifying — see the scope run for the per-path evidence.

parametrization:
Closes #1138

## Summary

Fixes the Step 6.5 premise-rot gate (`.claude/lib/premise_check.py`), which was STOPping every Phase-10 scope run on false positives — 5/5 in the issue's own wave-29 run, 6/6 in wave-30's rerun (owner note on `cross-repo-status.json`).

**Seven false-positive classes fixed**, each with a pre-fix-failing regression test: the issue's own three classes, a 4th class confirmed live but not documented in the issue body, and three more found live in the wave-30 rerun.

### The issue's three classes

1. **Cross-repo `.github/workflows/` references** (issue class 1). The `CROSS_REPO` WARN downgrade only checked child->parent (main#1047 da#427). Extended symmetrically: a `noorinalabs-main` issue naming a `.github/workflows/*.yml` file that misses in the parent but resolves in a child repo now also downgrades to WARN, not STOP. Generalized the trigger set to `.claude/` **and** `.github/` (both directions), **and** — added at merge-gate review, see "MF1" below — to a bare (slash-free) filename on the parent->child direction, because that is the literal shape #1110/#1111 use.
2. **Proposed-not-yet-existing artifacts** (issue class 2). An issue's own `creates` array (new, alongside the existing `paths`/`symbols` explicit arrays) declares "this issue's own output" — those paths are never checked against git and never contribute to the verdict, so an extraction/consolidation story naming its own target file no longer reads as premise rot.
3. **Bare extension tokens** (issue class 3). `looks_like_path` now requires a non-empty stem before the extension — computed on the path's **basename**, not the raw token, so `.claude/lib/` (a leading dot-directory) is unaffected. A bare `.py`/`.yaml` mentioned in prose no longer parses as a path.

### 4th class — not in the issue body, confirmed live at wave-30 scope

The basename fallback (main#1047 da#373) only fired for slash-free tokens. Generalized to a **suffix** fallback, so a multi-segment relative fragment like `lib/check_agent_liveness.py` resolves against the tracked `.claude/lib/check_agent_liveness.py`. Confirmed and posted as a comment on #1138.

### Three more classes found live in the wave-30 rerun, not named in the issue body

- **Filesystem-absolute paths** (`/tmp/claude-.../scratchpad/*.md` quoted verbatim in an issue body — the exact shape wave-29's own `block_stale_tmp_message_file` false-positive writeup used). Rejected outright: every path this gate checks resolves via `git cat-file -e <ref>:<path>`, which is always repo-relative, so an absolute token is never that kind of path regardless of extension.
- **A legitimately `.gitignore`d path** (`.claude/annunaki/errors.jsonl`). `git cat-file` can never see it tracked, by design, so a bare MISSING there would STOP forever. New `GITIGNORED` status (WARN, via `git check-ignore`), not MISSING.
- **Deliberate example / placeholder filenames** (`X.md` — the exact placeholder main#1352 uses for "insert your body-file path here"; `foo.py`/`example.py` as the generic convention) are now rejected via a narrow placeholder-stem list, including single-**uppercase**-letter stems (`X`, `Y`) — deliberately excluding lowercase (`a.py`, `c.py` are real filenames used elsewhere in this repo's own tests).

## MF1 fix (merge-gate review, Aino Virtanen) — class 1 widened to the reported bare-filename shape

Aino's first-pass review found the class-1 fixture used the qualified `.github/workflows/ghcr-publish.yml` form, which the shared-prefix trigger already caught — but **#1110 and #1111, the two issues class 1 was actually reported from, name the file by bare filename** (`ghcr-publish.yml`, `structural-ontology.yml`, `structural_ontology.py`), which never satisfied `value.startswith(_CROSS_REPO_SHARED_PREFIXES)`, so those two issues still STOPped identically before and after the first push.

**Fix:** `_cross_repo_downgrade` now also triggers the parent->child second chance for a bare (slash-free) filename — deliberately **one-directional** (parent->child only): a child-repo issue naming a bare filename that happens to exist somewhere in the parent's `.claude/` tree still STOPs (see `test_cross_repo_bare_filename_not_widened_on_child_to_parent_side`), and a bare filename genuinely absent everywhere still STOPs (`test_cross_repo_bare_filename_not_found_anywhere_still_stops`). New regression fixture pinned on the **literal #1110 token**: `test_cross_repo_bare_filename_main_1110_literal_token`.

Re-verified end-to-end against the real, current API bodies of #1110 and #1111 (not a synthetic fixture), using `--repos-root` pointed at a tree with the child repos actually cloned (this PR's own isolated worktree has no child-repo checkouts, so the same run there reports `UNVERIFIABLE`/`MISSING` rather than `CROSS_REPO` — an environment gap, not a logic gap; confirmed by the same run succeeding once child repos are visible):

```
# before MF1 (first push, fb0aeb328) — real #1110/#1111 bodies fetched from the API
[STOP] main#1110  - path `ghcr-publish.yml`: MISSING
[STOP] main#1111  - path `structural-ontology.yml`: MISSING
                  - path `structural_ontology.py`: MISSING

# after MF1 (this push, 922ff6e), --repos-root pointed at a tree with child repos cloned
[WARN] main#1110  - path `ghcr-publish.yml`: cross-repo ... resolves in child repo noorinalabs-isnad-graph
[WARN] main#1111  - path `structural-ontology.yml`: cross-repo ... resolves in child repo noorinalabs-isnad-graph
                  - path `structural_ontology.py`: cross-repo ... resolves in child repo noorinalabs-isnad-graph
```

`Closes #1138` therefore stays accurate at this head, per Aino's guidance.

## Precision, measured against the wave-30 6/6 false-STOP set's classes + MF1

| Class | Before | After |
|---|---|---|
| Cross-repo workflow paths — qualified form | STOP (false) | WARN |
| Cross-repo workflow paths — bare-filename form (#1110/#1111's actual shape, MF1) | STOP (false) | WARN |
| `creates`-declared proposed files | STOP (false) | OK |
| Bare extension tokens (`.py`, `.yaml`) | STOP (false) | rejected at extraction, no candidate |
| Slash-containing relative fragment (`lib/check_agent_liveness.py`) | STOP (false) | OK (suffix-resolved) |
| `/tmp` scratchpad absolute paths | STOP (false) | rejected at extraction, no candidate |
| Gitignored `errors.jsonl` | STOP (false) | WARN |
| Deliberate placeholder filenames (`X.md`, `foo.py`, `example.py`) | STOP (false) | rejected at extraction, no candidate |
| True positive — genuinely deleted `wave_key_reset.py` (#705) | STOP (true) | STOP (true, preserved) |
| True positive — bare filename absent everywhere (parent + all children) | STOP (true) | STOP (true, preserved) |
| True positive — bare filename in a CHILD repo, coincidentally present in parent | STOP (true) | STOP (true, preserved — MF1 is one-directional) |

## Pre-fix failure output (original 8 tests, first push)

Ran the new tests against pre-fix `premise_check.py` (before any change in this PR): all 8 new tests fail, all 40 pre-existing tests still pass.

```
FAILED .claude/lib/tests/test_premise_check.py::Main1138LooksLikePathTest::test_rejects_absolute_tmp_scratchpad_path
FAILED .claude/lib/tests/test_premise_check.py::Main1138LooksLikePathTest::test_rejects_bare_extension_token
FAILED .claude/lib/tests/test_premise_check.py::Main1138LooksLikePathTest::test_rejects_deliberate_example_filenames
FAILED .claude/lib/tests/test_premise_check.py::Main1138VerdictLayerTest::test_creates_array_prevents_stop_on_proposed_file
FAILED .claude/lib/tests/test_premise_check.py::Main1138VerdictLayerTest::test_creates_does_not_suppress_other_missing_paths
FAILED .claude/lib/tests/test_premise_check.py::Main1138VerdictLayerTest::test_cross_repo_symmetric_parent_misses_child_resolves
FAILED .claude/lib/tests/test_premise_check.py::GitIntegrationTest::test_gitignored_path_is_warn_not_stop
FAILED .claude/lib/tests/test_premise_check.py::GitIntegrationTest::test_relative_fragment_suffix_resolves_under_claude
========================= 8 failed, 40 passed in 0.50s =========================
```

Representative assertions:

```
Main1138LooksLikePathTest.test_rejects_bare_extension_token
>       self.assertFalse(pc.looks_like_path(".py"))
E       AssertionError: True is not false

Main1138VerdictLayerTest.test_cross_repo_symmetric_parent_misses_child_resolves
>       self.assertEqual(res.verdict, pc.WARN)
E       AssertionError: 'stop' != 'warn'

Main1138VerdictLayerTest.test_creates_array_prevents_stop_on_proposed_file
>       self.assertEqual(res.verdict, pc.OK)
E       AssertionError: 'stop' != 'ok'

GitIntegrationTest.test_gitignored_path_is_warn_not_stop
>           self.assertEqual(res.candidates[0].status, pc.GITIGNORED)
E           AttributeError: module 'premise_check' has no attribute 'GITIGNORED'

GitIntegrationTest.test_relative_fragment_suffix_resolves_under_claude
>           self.assertEqual(statuses["lib/check_agent_liveness.py"], pc.EXISTS)
E           AssertionError: 'missing' != 'exists'
```

## Pre-MF1 failure output (the 3 new MF1 tests, against the first-push head fb0aeb328)

```
FAILED .claude/lib/tests/test_premise_check.py::Main1138VerdictLayerTest::test_cross_repo_bare_filename_main_1110_literal_token
  AssertionError: 'stop' != 'warn'
```

(`test_cross_repo_bare_filename_not_found_anywhere_still_stops` and `test_cross_repo_bare_filename_not_widened_on_child_to_parent_side` already passed at fb0aeb328 — they pin behavior that MF1 must NOT change, i.e. recall and directionality — so they are not "failing-before" fixtures, they are guard-rails against over-widening the MF1 fix itself.)

## Post-fix (this head, 922ff6e)

```
51 passed in 0.39s
```

All 48 previously-passing tests (40 pre-existing + 8 from the first push) still pass unchanged, plus the 3 new MF1 tests.

## CI / local parity

Ran the complete local check-set before each push (`pre-commit run` + `--hook-stage pre-push`): ruff-format, ruff-lint, actionlint, cspell, mypy (hooks+lib+skills), the full `.claude/lib/tests/` + `.claude/hooks/tests/` pytest suites, the drift-sync gate (`pre_commit_ci_sync.py`), memory-budget, headcount-budget, doc-freshness (advisory), office-drift, mermaid-render. All green. CI on the first push: 21/21 success (independently re-verified by Aino via REST check-runs).

## Reviewers

- Aino Virtanen (peer reviewer, merge-gate)
- Nino Kavtaradze (secondary reviewer, Approved on the first push)

### DIFF
--- a/.claude/lib/premise_check.py
+++ b/.claude/lib/premise_check.py
@@ -32,18 +32,38 @@ Verdict policy (deterministic):
     deliberately NOT a STOP: a missing local checkout is an environment gap, not
     evidence the premise rotted — a false STOP there would block scope on
     tooling state, the opposite of the gate's purpose.
-  * a ``.claude/``-rooted path that MISSES in a child repo but resolves in the
-    parent (``noorinalabs-main``) -> ``CROSS_REPO`` -> verdict ``WARN`` (main#1047
-    da#427): very often a legitimate reference to org-wide config, worth a
+  * a ``.claude/``/``.github/``-rooted path that MISSES in one repo of an org
+    pair (parent<->child) but resolves in the other -> ``CROSS_REPO`` ->
+    verdict ``WARN`` (main#1047 da#427 child->parent; main#1138 wave-30
+    extends this symmetrically to parent->child, e.g. a ``noorinalabs-main``
+    issue naming a ``.github/workflows/`` file a Track-C story is hoisting
+    FROM a child repo): very often a legitimate cross-repo reference, worth a
     manual look rather than a hard block.
-  * a bare (slash-free) filename that misses at repo root but resolves
-    uniquely elsewhere in the tree -> ``EXISTS`` via a basename fallback
-    (main#1047 da#373).
+  * a relative fragment that misses at its literal location but is a real
+    suffix of some tracked path elsewhere in the tree (a bare basename per
+    main#1047 da#373, or a multi-segment fragment like
+    ``lib/check_agent_liveness.py`` per main#1138's 4th FP class) ->
+    ``EXISTS`` via the suffix fallback.
+  * a path that MISSES but is matched by a ``.gitignore`` rule (main#1138:
+    ``.claude/annunaki/errors.jsonl``) -> ``GITIGNORED`` -> verdict ``WARN``:
+    git can never see a gitignored path tracked, by design, so that alone is
+    not evidence of premise rot.
+  * a path listed in the issue's own ``creates`` array (main#1138 class 2: an
+    issue that PROPOSES to add a file names its own output, not an
+    already-holding premise) -> ``CREATES``, never checked against git, never
+    contributes to the verdict.
   * everything present, or no concrete refs to check -> ``OK``.
 
-An issue may also declare explicit ``paths`` / ``symbols`` arrays (deliberate
-named premises the orchestrator wants asserted even if the body phrasing is too
-loose for auto-extraction); a symbol may be scoped to a pathspec.
+Extraction also rejects three shapes that read as paths but never assert a
+repo premise (main#1138): a filesystem-absolute token (``/tmp/.../*.md`` — no
+``git cat-file`` target is ever absolute), a bare extension with nothing
+before the dot (``.py``, ``.yaml``), and a doc-placeholder filename (``X.md``,
+``foo.py``, ``example.py`` — the "insert your own name here" convention).
+
+An issue may also declare explicit ``paths`` / ``symbols`` / ``creates``
+arrays (deliberate named premises — or declared outputs — the orchestrator
+wants asserted even if the body phrasing is too loose for auto-extraction); a
+symbol may be scoped to a pathspec.
 
 CLI:
   premise_check.py check --issues ISSUES.json [--ref origin/main]
@@ -58,6 +78,9 @@ CLI:
       "symbols": [                              # optional explicit symbols
         {"name": "wave_key_reset", "pathspec": ".claude/lib/"}
       ],
+      "creates": [".claude/lib/org_repos.py"],  # optional: this issue's own
+                                                 # proposed output, exempt
+                                                 # from existence checks
       "repo_dir": "/abs/path",      # optional; overrides repos-root resolution
       "git_ref": "origin/main"      # optional per-issue ref override
     }
@@ -76,7 +99,7 @@ from collections.abc import Callable
 from dataclasses import dataclass, field
 from pathlib import Path
 
-from org_repos import ALL_REPOS
+from org_repos import ALL_REPOS, CHILD_REPOS, MAIN_REPO
 
 # Repo root = two parents above .claude/lib/ (lib -> .claude -> root). Resolved
 # from this file so the default repos-root is correct from any cwd or worktree.
@@ -86,12 +109,25 @@ _REPO_ROOT = Path(__file__).resolve().parents[2]
 EXISTS = "exists"
 MISSING = "missing"
 UNVERIFIABLE = "unverifiable"
-# A `.claude/`-rooted path that misses in a CHILD repo but resolves in the
-# parent (`noorinalabs-main`) — very often a legitimate reference to org-wide
-# config from a child-repo issue (main#1047 da#427). Treated as WARN, not
-# STOP: the reference is real, but reading it against the wrong repo is worth
-# a manual look, not a hard block.
+# A shared-root path (`.claude/`, `.github/`) that misses in one repo of an
+# org pair but resolves in the other — very often a legitimate cross-repo
+# reference (main#1047 da#427: child names a parent-owned org-wide config
+# path; main#1138 wave-30: a `noorinalabs-main` issue names a
+# `.github/workflows/` file a Track-C story is hoisting FROM a child repo,
+# i.e. the parent side of the same reference). Treated as WARN, not STOP: the
+# reference is real, but reading it against the wrong repo is worth a manual
+# look, not a hard block.
 CROSS_REPO = "cross_repo"
+# A path that git can never see tracked because it is deliberately
+# `.gitignore`d (main#1138 wave-30: `.claude/annunaki/errors.jsonl`) — cannot
+# be confirmed present via `git cat-file`, by design, so a bare MISSING would
+# read as premise rot on every single check forever. WARN, not STOP: the
+# reference may well be legitimate, but git cannot verify it either way.
+GITIGNORED = "gitignored"
+# An issue's own declared *output* (its `creates` array — main#1138 class 2):
+# a path this issue proposes to add, not a premise it asserts already holds.
+# Never checked against git, never contributes to the verdict.
+CREATES = "creates"
 
 # Per-issue verdicts.
 OK = "ok"
@@ -174,10 +210,31 @@ _KNOWN_ROOT_COMPONENTS = frozenset(
     }
 ) | frozenset(ALL_REPOS)
 
+# Root prefixes shared org-wide between the parent (`noorinalabs-main`) and
+# every child repo — org-wide tooling config (`.claude/`) and CI workflow
+# definitions (`.github/`, main#1138 wave-30: Track C hoists per-repo
+# `.github/workflows/*.yml` into `noorinalabs-main`). A path under one of
+# these that misses in one repo of an org pair but resolves in the other is
+# the CROSS_REPO downgrade's trigger set, symmetric in both directions
+# (child names a parent path: main#1047 da#427; parent names a child path:
+# main#1138).
+_CROSS_REPO_SHARED_PREFIXES = (".claude/", ".github/")
+
 # Known git refs / ref prefixes: `origin/main`, `refs/heads/x`, `HEAD` are
 # never file paths even though `origin/main` contains a slash (main#1047).
 _GIT_REF_LITERALS = frozenset({"HEAD"})
 _GIT_REF_PREFIXES = ("origin/", "refs/")
+
+# Doc-placeholder filenames used purely for illustration across this org's own
+# issues/docs (main#1138 wave-30 class: `X.md` is the exact stand-in main#1352
+# used for "insert your body-file path here"; the rest are the generic
+# prose convention). A token whose STEM (the part before the final dot,
+# case-insensitively) is exactly one of these is never asserted as a real
+# path, however plausible its extension looks. Deliberately narrow — a real
+# short name like `gh.py` or `db.py` is untouched.
+_PLACEHOLDER_STEMS = frozenset(
+    {"foo", "bar", "baz", "example", "sample", "placeholder", "dummy", "lorem"}
+)
 
 
 def _all_components_numeric(tok: str) -> bool:
@@ -200,14 +257,22 @@ def looks_like_path(token: str) -> bool:
     premise-rot purely because prose like ``A/B``, ``recall/precision``, or the
     git ref ``origin/main`` contains a ``/``). A token is accepted only when it
     is whitespace-free, path-character-only, not a bare git ref, not an
-    all-numeric fraction, and EITHER:
+    all-numeric fraction, not an absolute filesystem path, not a doc
+    placeholder, and EITHER:
 
-      * it ends in a known code/doc extension (``foo.py``, ``bar/baz.md``), OR
+      * it ends in a known code/doc extension AND has a non-empty stem before
+        that extension (``foo.py``, ``bar/baz.md`` — but not a bare ``.py``
+        with nothing before the dot), OR
       * it contains a ``/`` AND its leading component is a known repo-root
         directory (``src/``, ``docs/``, ``.claude/``, a child-repo name, …).
 
     A slash-containing token with neither signal (e.g. ``ism/kunya``,
-    ``boundary/particle``) is prose, not a path.
+    ``boundary/particle``) is prose, not a path. Every repo this gate checks
+    resolves paths as ``git cat-file -e <ref>:<path>``, which is always
+    repo-*relative* — a token starting with ``/`` is a filesystem-absolute
+    path (main#1138 wave-30: an ephemeral ``/tmp/.../scratchpad/*.md`` quoted
+    verbatim in an issue body) and can never be that kind of path, whatever
+    extension it happens to carry.
     """
     tok = token.strip()
     if not tok or " " in tok or "\t" in tok:
@@ -221,10 +286,35 @@ def looks_like_path(token: str) -> bool:
         return False
     if tok == "/":
         return False
+    # A filesystem-absolute path (main#1138: `/tmp/...`, `/var/...`) is never
+    # a repo-relative git path, regardless of what extension it carries.
+    if tok.startswith("/"):
+        return False
     if tok in _GIT_REF_LITERALS or tok.startswith(_GIT_REF_PREFIXES):
         return False
-    ext = tok.rsplit(".", 1)[-1].lower() if "." in tok else ""
-    if ext in _CODE_EXTENSIONS:
+    # Extension/stem/placeholder checks operate on the final path component
+    # (basename) — NOT the raw token — so a leading dot-directory
+    # (``.claude/lib/``) is never mistaken for a bare-extension token; only
+    # `.claude` (no trailing `/`) IS the basename in that case, and it has no
+    # dot to split on at all.
+    basename = tok.rsplit("/", 1)[-1]
+    stem, dot, ext_raw = basename.rpartition(".")
+    ext = ext_raw.lower() if dot else ""
+    # A bare extension with nothing before the dot (main#1138 class 3: a bare
+    # `.py`/`.yaml` mentioned in prose) is never a path — reject regardless of
+    # what follows.
+    if dot and not stem:
+        return False
+    # A doc-placeholder filename (main#1138 wave-30 class): a whole-word
+    # placeholder stem, or a single UPPERCASE-letter stem (`X.md`, `Y.py` —
+    # the "insert your own name" convention this org's own docs use, e.g.
+    # main#1352's `--body-file X.md`), reads as illustration, not a real
+    # asserted path. Lowercase single-letter stems (`a.py`, `c.py`) are left
+    # alone — this repo's own test fixtures use them as arbitrary real
+    # filenames, and the placeholder convention is specifically uppercase.
+    if stem and (stem.lower() in _PLACEHOLDER_STEMS or (len(stem) == 1 and stem.isupper())):
+        return False
+    if ext in _CODE_EXTENSIONS and stem:
         return True
     if "/" in tok and not _all_components_numeric(tok):
         first = tok.split("/", 1)[0]
@@ -302,20 +392,41 @@ def git_ref_exists(repo_dir: str, ref: str) -> bool:
     return proc.returncode == 0
 
 
-def _basename_resolves(repo_dir: str, ref: str, basename: str) -> bool:
-    """True when some tracked path at ``ref`` ends in ``/<basename>`` (or is it).
+def _suffix_resolves(repo_dir: str, ref: str, candidate: str) -> bool:
+    """True when some tracked path at ``ref`` equals ``candidate`` or ends in
+    ``/<candidate>``.
 
-    Basename fallback (main#1047 da#373): an issue naming a bare filename like
-    ``composition.py`` may not mean the repo root — the file can live at
-    ``src/parse/composition.py``. Only applies to slash-free candidates; a real
-    subpath that misses at its literal location is still MISSING, never
-    re-resolved by searching the whole tree.
+    Suffix fallback (main#1047 da#373 basename case, generalized by main#1138
+    4th FP class): an issue naming a bare filename like ``composition.py``
+    may not mean the repo root — the file can live at
+    ``src/parse/composition.py``. main#1138 widened this from slash-free
+    basenames to any relative fragment: a token like
+    ``lib/check_agent_liveness.py`` misses at its literal (repo-root-relative)
+    location but IS a real suffix of the tracked ``.claude/lib/
+    check_agent_liveness.py`` — the original basename-only fallback never
+    fired for it because the token was not slash-free. A candidate that
+    matches no tracked path anywhere is still MISSING, never re-resolved by
+    a coincidental partial match.
     """
     proc = _run_git(repo_dir, ["ls-tree", "-r", "--name-only", ref])
     if proc.returncode != 0:
         return False
-    suffix = "/" + basename
-    return any(line == basename or line.endswith(suffix) for line in proc.stdout.splitlines())
+    suffix = "/" + candidate
+    return any(line == candidate or line.endswith(suffix) for line in proc.stdout.splitlines())
+
+
+def git_path_is_ignored(repo_dir: str, path: str) -> bool:
+    """True when ``path`` is matched by a ``.gitignore`` rule in ``repo_dir``.
+
+    Checked against the *working tree's* current ``.gitignore`` (there is no
+    "ignored at a historical ref" concept — ignore rules are a property of
+    the tree, not of git history). ``git check-ignore`` exits 0 when the path
+    is ignored, 1 when it is not, and 128 on certain error conditions (all
+    folded into "not ignored" — this is a best-effort downgrade, never a
+    reason to assert MISSING harder).
+    """
+    proc = _run_git(repo_dir, ["check-ignore", "-q", "--", path])
+    return proc.returncode == 0
 
 
 def git_path_status(repo_dir: str, ref: str, path: str) -> str:
@@ -323,17 +434,23 @@ def git_path_status(repo_dir: str, ref: str, path: str) -> str:
 
     Returns :data:`UNVERIFIABLE` when the repo dir or ref cannot be read (an
     environment gap, never treated as premise rot), otherwise :data:`EXISTS` /
-    :data:`MISSING`. A slash-free ``path`` that misses at its literal location
-    gets one more chance via :func:`_basename_resolves` before being declared
-    MISSING (main#1047 basename fallback).
+    :data:`GITIGNORED` / :data:`MISSING`. A relative fragment that misses at
+    its literal location gets one more chance via :func:`_suffix_resolves`
+    before being declared MISSING (main#1047 da#373 / main#1138 suffix
+    fallback). Failing that, a path deliberately excluded from tracking via
+    ``.gitignore`` (main#1138: ``.claude/annunaki/errors.jsonl``) downgrades
+    to :data:`GITIGNORED` rather than MISSING — git can never see it
+    tracked, by design, so a bare MISSING there is not evidence of rot.
     """
     if not git_ref_exists(repo_dir, ref):
         return UNVERIFIABLE
     proc = _run_git(repo_dir, ["cat-file", "-e", f"{ref}:{path}"])
     if proc.returncode == 0:
         return EXISTS
-    if "/" not in path and _basename_resolves(repo_dir, ref, path):
+    if _suffix_resolves(repo_dir, ref, path):
         return EXISTS
+    if git_path_is_ignored(repo_dir, path):
+        return GITIGNORED
     return MISSING
 
 
@@ -368,7 +485,7 @@ SymbolChecker = Callable[[str, str, str, "str | None"], str]
 class CandidateResult:
     kind: str  # "path" | "symbol"
     value: str
-    status: str  # EXISTS | MISSING | UNVERIFIABLE | CROSS_REPO
+    status: str  # EXISTS | MISSING | UNVERIFIABLE | CROSS_REPO | GITIGNORED | CREATES
     pathspec: str | None = None
     note: str | None = None
 
@@ -394,11 +511,15 @@ class IssueResult:
     def cross_repo(self) -> list[CandidateResult]:
         return [c for c in self.candidates if c.status == CROSS_REPO]
 
+    @property
+    def gitignored(self) -> list[CandidateResult]:
+        return [c for c in self.candidates if c.status == GITIGNORED]
+
 
 def _verdict_for(candidates: list[CandidateResult]) -> str:
     if any(c.status == MISSING for c in candidates):
         return STOP
-    if any(c.status in (UNVERIFIABLE, CROSS_REPO) for c in candidates):
+    if any(c.status in (UNVERIFIABLE, CROSS_REPO, GITIGNORED) for c in candidates):
         return WARN
     return OK
 
@@ -445,6 +566,46 @@ def collect_candidates(issue: dict) -> list[tuple[str, str, str | None]]:
     return triples
 
 
+def _cross_repo_downgrade(
+    value: str,
+    repo_name: str,
+    repos_root: Path,
+    git_ref: str,
+    path_checker: PathChecker,
+) -> tuple[str, str] | None:
+    """Second-chance lookup for a MISSING shared-root path in the "other side"
+    of the parent/child pair (main#1047 da#427, extended symmetrically by
+    main#1138 wave-30). Returns ``(CROSS_REPO, note)`` on a hit, else ``None``.
+
+    * child repo, path misses locally -> check the parent (`noorinalabs-main`).
+    * parent repo (`noorinalabs-main`), path misses locally -> check every
+      child repo in turn; the first hit wins.
+
+    Only fires for a path under one of ``_CROSS_REPO_SHARED_PREFIXES`` — a
+    non-shared path missing in the wrong repo is not this class of reference.
+    """
+    if not value.startswith(_CROSS_REPO_SHARED_PREFIXES):
+        return None
+    if repo_name != MAIN_REPO:
+        parent_status = path_checker(str(repos_root), git_ref, value)
+        if parent_status == EXISTS:
+            return (
+                CROSS_REPO,
+                f"resolves in parent repo {MAIN_REPO}, not child repo "
+                f"{repo_name} — verify this is a deliberate cross-repo reference",
+            )
+        return None
+    for child in CHILD_REPOS:
+        child_status = path_checker(str(repos_root / child), git_ref, value)
+        if child_status == EXISTS:
+            return (
+                CROSS_REPO,
+                f"resolves in child repo {child}, not parent repo {MAIN_REPO} — "
+                "verify this is a deliberate cross-repo reference",
+            )
+    return None
+
+
 def check_issue(
     issue: dict,
     repos_root: Path,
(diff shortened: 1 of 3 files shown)


parametrization:
Secondary review — path-resolution fix for the Step 6.5 premise-rot gate.

## The six (really seven) classes — genuinely distinct, not one root cause counted N ways

Verified each against a real code path and a real pre-fix-failing test:

| Class | Mechanism | Where |
|---|---|---|
| Cross-repo `.claude/`/`.github/` paths | verdict-layer downgrade, symmetric parent<->child | `_cross_repo_downgrade()` |
| `creates`-declared proposed files | verdict-layer exemption, never hits git | `check_issue()` creates-set check |
| Bare extension tokens (`.py`, `.yaml`) | extraction-layer rejection: empty stem | `looks_like_path()` `dot and not stem` |
| Slash-containing relative fragment | git-check-layer: generalized suffix fallback | `_suffix_resolves()` |
| `/tmp` absolute paths | extraction-layer rejection: leading `/` | `looks_like_path()` `tok.startswith("/")` |
| Gitignored `errors.jsonl` | git-check-layer: new `check-ignore` call | `git_path_is_ignored()` |
| (untabulated) Placeholder filenames (`X.md`, `foo.py`) | extraction-layer rejection: placeholder-stem list | `looks_like_path()` `_PLACEHOLDER_STEMS` |

They cluster into 3 mechanisms (extraction-time reject / git-check resolution / verdict-time downgrade), but each predicate within a mechanism is independent and separately regression-tested — this is a real taxonomy, not padding. One doc nit: the PR body's precision table has 6 rows but the code+tests actually ship 7 distinct FP-rejecting classes (placeholder filenames is tested in `test_rejects_deliberate_example_filenames` but never appears in the table). Non-blocking, filed as part of the tech-debt issue below.

## Recall — verified, genuinely-missing paths still STOP

Checked this isn't a "stop STOPping" fix in disguise. Confirmed STOP survives across every new code path: `test_705_deleted_file_is_stop`, `test_basename_fallback_does_not_resurrect_a_real_deletion` (a real deletion isn't resurrected by the *widened* suffix fallback), `test_non_claude_missing_in_child_repo_is_still_stop` (cross-repo downgrade doesn't leak to non-shared-root paths), `test_creates_does_not_suppress_other_missing_paths` (creates-exemption is scoped to its own declared path, not a blanket pass), and the end-to-end CLI test against a real throwaway git repo (`test_cli_stop_exit_code_over_real_repo`, rc=1). Good — precision went up without recall going to zero.

## Mutation scrutiny — one gap found, filed as tech debt (#1396)

`_verdict_for` is two independent `any()` passes (STOP checked unconditionally before WARN), which is correctly order-independent as written. But the only test combining a STOP-status candidate with a WARN-status candidate in one issue (`test_missing_dominates_unverifiable`) exercises only `MISSING` + `UNVERIFIABLE`, never `MISSING` + `CROSS_REPO` or `MISSING` + `GITIGNORED` (both new/extended here), and even that one test always has the MISSING candidate extracted *first*. A plausible future refactor into a first-match loop would be silently order-dependent and would invert dominance whenever a WARN-status candidate is extracted before the MISSING one — nothing in the current suite would catch it. Filed https://github.com/noorinalabs/noorinalabs-main/issues/1396 with the 3 concrete regression tests needed (CROSS_REPO+MISSING, GITIGNORED+MISSING, and a body-order-reversed case).

## #1350 cross-check — no recurrence here

#1350 found byte-identical duplicated matchers drifting in effect at different offsets. Checked for the same shape: the old code had the child->parent cross-repo check inline in `check_issue()`; this PR *extracts* it into a single `_cross_repo_downgrade()` function used once, now handling both directions symmetrically off `MAIN_REPO`/`CHILD_REPOS` from the `org_repos.py` SSOT (not hand-listed). No second copy of the resolution rule anywhere else in `.claude/lib`/`.claude/hooks`/`.claude/skills` (checked via `rg --hidden --no-ignore`) — this PR consolidates rather than perpetuates that shape. `wave-scope/SKILL.md`'s description is prose, correctly kept in sync with the new behavior, not an executable second copy.

## `--hidden` / gitignore semantics — correctly distinguished

`git_path_status` layers three checks in order: `git cat-file -e <ref>:path` (tracked-at-ref, unaffected by `.gitignore`), then `_suffix_resolves` via `git ls-tree` (same), then `git check-ignore` (working-tree `.gitignore`) only as the last resort before declaring MISSING. This correctly separates "does not exist" from "not visible to this check" — the new `GITIGNORED` status is exactly the class of false-STOP this org's own `.claude/` sweep gotcha describes. One documented, accepted edge case worth naming for the record (not blocking): `check-ignore` reads the *working tree's* current `.gitignore`, not the `.gitignore` as it stood at the checked ref — a path whose ignore status changed between the ref and now could misclassify. The docstring already states this tradeoff explicitly, so it's a known and reasonable limitation, not an oversight.

## CI / head SHA

21/21 checks green, verified against `statusCheckRollup` for head SHA `fb0aeb3283f022b8e41c71f9fb9352e628f4b919` (matches `gh pr view --json headRefOid`), not the PR number.

---
Requestor: Nino Kavtaradze
Requestee: Weronika Zielinska
RequestOrReplied: Approved

ReviewedHead: fb0aeb3283f022b8e41c71f9fb9352e628f4b919
TechDebt: #1396

parametrization:
Merge-gate review. Reviewed at head `fb0aeb328` against base `e0196c377` (the PR is one commit off `main`; no rebase drift). CI re-verified at that SHA via REST check-runs: **21/21 success, 0 non-success, 0 skipped**. Everything below was executed — the head and base modules were fetched via `gh api .../contents?ref=<sha>` into a scratch tree and run, never read out of a local checkout.

---

## What holds up (verified, not taken on trust)

**Failing-before evidence — independently reproduced, exactly as claimed.** Head's `test_premise_check.py` run against base's `premise_check.py`: **8 failed, 40 passed**; the 8 are precisely the 8 new tests. The 40 pre-existing fixtures (#1047 / #705 / #816) pass unchanged under both.

**Mutation testing — 13 single-guard mutations, 11 killed.** Each new guard was individually neutralised and the suite re-run:

| Mutation | Killed by |
|---|---|
| drop absolute-path reject | `test_rejects_absolute_tmp_scratchpad_path` |
| drop placeholder-stem reject (whole, and each half separately) | `test_rejects_deliberate_example_filenames` |
| `basename` → raw token | `test_accepts_slash_paths` (the pre-existing `.claude/lib/` assertion) |
| suffix fallback → basename-only | `test_relative_fragment_suffix_resolves_under_claude` |
| drop `check-ignore` downgrade | `test_gitignored_path_is_warn_not_stop` |
| drop `GITIGNORED` from the WARN arm of `_verdict_for` | same |
| drop `creates` exemption | both `creates` tests |
| `CREATES` constant aliased to `MISSING` | both `creates` tests |
| drop parent→child branch of `_cross_repo_downgrade` | `test_cross_repo_symmetric_…` |
| drop `.github/` from `_CROSS_REPO_SHARED_PREFIXES` | same |

Two survivors, and they are **redundancy, not inertness**: `if dot and not stem: return False` and the `and stem` on `if ext in _CODE_EXTENSIONS and stem` each independently reject a bare `.py`, so neither is individually killable. Mutating **both** together does kill `test_rejects_bare_extension_token` (1 failed, 47 passed) — the class-3 assertion is live. No change requested; noting it so a future reader does not delete the "dead-looking" clause. (They are not fully equivalent: only the first rejects `<known-root>/.py`, which no fixture covers.)

**Recall — a true positive still STOPs.** This is the check that matters most on a STOP gate, and it passes. Real CLI, real repo, `--ref HEAD`, 13 probes: **5 true positives STOP, exit code 1**, including the full-path form `.claude/lib/wave_key_reset.py` (i.e. neither the widened suffix fallback, nor `check-ignore`, nor the symmetric cross-repo lookup rescues a genuinely deleted file), and `TP-6` where a real MISSING coexists with a `creates`-declared path and STOP correctly dominates. The gate has not been tuned into silence.

**`org_repos.py` / #1388 collision — none.** #1388 (`fix(#1243)`, head `83f54f816`) shares this PR's base `e0196c377` and touches `.claude/hooks/validate_wave_audit.py`, its test, two memory files and `catalog-13-17.md`. It **does not modify `.claude/lib/org_repos.py`** — it only consumes it, as this PR does. Zero file overlap with #1390's three files, and `MAIN_REPO` / `CHILD_REPOS` / `ALL_REPOS` are untouched by both. No textual or semantic merge collision; the two can land in either order.

---

## MF1 — blocking: class 1 is not fixed for either of the two instances the issue reports

#1138's class 1 is the first-listed class and the one gating Track C. Its two reported instances are **#1110** (`ghcr-publish.yml`) and **#1111** (`structural-ontology.yml`, `structural_ontology.py`). I ran this PR's gate over those issues' **actual bodies fetched from the API**, at HEAD:

```
##### HEAD (fb0aeb328) #####
[STOP] main#1110  - path `ghcr-publish.yml`: MISSING
[STOP] main#1111  - path `structural-ontology.yml`: MISSING
                  - path `structural_ontology.py`: MISSING
[ok]   main#1112
[ok]   main#1118
[ok]   main#1119

##### BASE (e0196c377) #####
[STOP] main#1110  - path `ghcr-publish.yml`: MISSING
[STOP] main#1111  - path `structural-ontology.yml`: MISSING
                  - path `structural_ontology.py`: MISSING
[STOP] main#1112  - path `.py`: MISSING
[STOP] main#1118  - path `.yaml`: MISSING
[ok]   main#1119
```

**#1110 and #1111 STOP identically before and after.** Classes 2 and 3 clear; class 1's own instances do not move.

Cause: those bodies name the workflows by **bare filename**, not by the fully-qualified `.github/workflows/…` path. `_cross_repo_downgrade` opens with

```python
if not value.startswith(_CROSS_REPO_SHARED_PREFIXES):
    return None
```

and a bare `ghcr-publish.yml` never satisfies that gate, so the symmetric parent→child lookup is never attempted. `test_cross_repo_symmetric_parent_misses_child_resolves` is a genuine test — it fails against base, and dropping either the `CHILD_REPOS` branch or the `.github/` prefix kills it — but its fixture is `.github/workflows/ghcr-publish.yml`, a **shape that does not occur in either issue the class was reported from**. The PR body's "Cross-repo workflow paths: STOP (false) → WARN" row is true of the fixture and not of the reported cases.

The files do exist child-side, so the downgrade is the right answer: `ghcr-publish.yml` in isnad-graph / user-service / data-acquisition / ingest-platform; `structural-ontology.yml` in five children; `scripts/structural_ontology.py` in all seven.

The mechanism is already present — the child-side `path_checker` is `git_path_status`, which runs the widened `_suffix_resolves` internally. Deleting the two-line prefix gate alone (as a diagnostic probe only, not a recommendation) turns all three into the intended result:

```
[WARN] main#1110 - path `ghcr-publish.yml`: cross-repo … resolves in child repo noorinalabs-isnad-graph
[WARN] main#1111 - path `structural-ontology.yml`: cross-repo …
                 - path `structural_ontology.py`: cross-repo …
```

**Required:** extend the parent→child second chance to the bare-filename / relative-fragment shape (a targeted widening — e.g. attempt the cross-repo lookup when the candidate has no leading component at all, alongside the existing shared-prefix trigger — rather than removing the prefix gate outright, which would open the lookup to every path). Add a regression fixture keyed on the **literal token from #1110** (`ghcr-publish.yml`), not the qualified path, so this exact miss cannot recur.

## MF2 — blocking, and it follows from MF1: `Closes #1138` is not accurate at head

Applying the PR #1383 ruling on its merits rather than mechanically, as instructed: there the surviving defect was message-only; here class 1 — the issue's own first-listed class, and the one whose two instances are still open Track C issues — is undelivered against the cases it was filed for.

Fix MF1 and `Closes #1138` becomes correct and should stay. If class 1 is instead deferred, change to `Refs #1138` and post an accurate residual on #1138 naming #1110/#1111 as still-STOPping. I'd prefer the former: the remaining work is small and the whole point of sequencing this story first is that Step 6.5 stops blocking Track C.

---

## Non-blocking

1. **Dotfile recall narrowing (no test either way).** `looks_like_path` regresses on basenames whose stem is empty: `.env` and `.claude/.env` were `True` at base, are `False` at head. Low impact (`.env` is gitignored everywhere in this org, so it would land on WARN regardless), and defensible as precision, but it is an undocumented, untested behaviour change — worth one line in the docstring or a fixture pinning the intent.
2. **`A.py` / any single-uppercase-letter stem is now unconditionally rejected.** Correct for `X.md`; I confirmed zero collisions with tracked files in the parent, so no live breakage. The comment already explains the uppercase/lowercase asymmetry — leaving as-is is fine.
3. **PR body arithmetic.** Title says "+ 2 more found live"; the body's "Two more classes observed live" section lists **three** bullets (absolute paths, gitignored, placeholder filenames), and three additional classes shipped with tests. Cosmetic, but the count is what a reader checks the test list against.
4. **`TP-4` in my probe set** (`wave_key_reset` symbol scoped to `.claude/lib/`) returns `ok` because the name appears in `premise_check.py`'s own module docstring. Pre-existing `git grep` behaviour, entirely out of scope here — flagging only so it is not mistaken for something this PR introduced.

## Tech debt filed

- **#1404** — the `GITIGNORED` downgrade is pattern-based, and the parent `.gitignore`s every child repo directory, so any `<child-repo>/…` path named in a `noorinalabs-main` issue now reports "cannot verify" even when the child **is** cloned and the path is genuinely gone. Reproduced at this head; lands on WARN (human-visible), so not a blocker, but it retires a real STOP class without substituting the authoritative check. Proposed direction and two-way fixtures are in the issue. On the board.
- **#1396** (filed by the parallel reviewer, verdict-precedence coverage) is a distinct gap; my `TP-6` probe shows STOP does dominate CREATES in practice at this head.

---

Excellent work on the parts that hold: the failing-before evidence is honest and reproduced exactly, every guard except one redundant pair is individually mutation-killable, and the true-positive recall — the thing that would have been invisible if it broke — is intact. The gap is narrow and specific: the class-1 fixture generalised the reported shape instead of reproducing it, and the reported shape still STOPs. Fix that and this is an approve.

---

Requestor: Aino Virtanen
Requestee: Weronika Zielinska

ReviewedHead: fb0aeb3283f022b8e41c71f9fb9352e628f4b919
TechDebt: #1404

parametrization:
Thank you for the thorough re-verification — both must-fixes addressed, additive commits only (no rebase/force-push), pushed to the same branch.

## MF1 — fixed

`_cross_repo_downgrade` now also triggers the parent->child second chance for a bare (slash-free) filename, restricted to the parent->child direction only (documented rationale in the docstring: a bare filename is a much weaker signal than a shared-root prefix, and widening it on the child->parent side too would risk downgrading a genuine child-repo STOP on the strength of an unrelated same-named file anywhere in the parent's `.claude/` tree).

New tests:
- `test_cross_repo_bare_filename_main_1110_literal_token` — pinned on the literal #1110 token (`ghcr-publish.yml`), not the qualified path. Fails against the first-push head `fb0aeb328` (`AssertionError: 'stop' != 'warn'`), passes at this head.
- `test_cross_repo_bare_filename_not_widened_on_child_to_parent_side` — directionality guard: a child-repo issue naming a bare filename that happens to resolve in the parent still STOPs.
- `test_cross_repo_bare_filename_not_found_anywhere_still_stops` — recall guard: a bare filename genuinely absent everywhere still STOPs.

Re-verified end-to-end against the real, current #1110/#1111 API bodies (not synthetic), with `--repos-root` pointed at a tree where the child repos are actually cloned:

```
before (fb0aeb328): [STOP] main#1110 - ghcr-publish.yml: MISSING
                     [STOP] main#1111 - structural-ontology.yml, structural_ontology.py: MISSING
after  (922ff6e):    [WARN] main#1110 - ghcr-publish.yml: cross-repo, resolves in noorinalabs-isnad-graph
                     [WARN] main#1111 - structural-ontology.yml, structural_ontology.py: cross-repo, resolves in noorinalabs-isnad-graph
```

## MF2 — resolved by MF1

`Closes #1138` stays as-is; per your guidance, fixing MF1 keeps it accurate at head.

## Also folded in (Nino's review)

- Added the 7th class (deliberate placeholder filenames — `X.md`/`foo.py`/`example.py`) to the PR body's precision table; it shipped and was tested but the table row was missing.
- Fixed the title/body bullet-count mismatch you both flagged (now "three more classes," matching the three bullets and three shipped tests).
- `#1396` (verdict-precedence coverage) and `#1404` (gitignore-pattern-vs-child-clone) are tech debt, not required for this PR — left as filed, not addressed here.

New head: `922ff6e3040b4487e8e04837d5cb78712f23d750`. 51/51 tests pass (48 from the first push + 3 new MF1 tests). Full local pre-commit + pre-push check-set green before each push. Please re-confirm against this head.

---

Requestor: Weronika Zielinska
Requestee: Aino Virtanen
RequestOrReplied: Reply
TechDebt: none

parametrization:
**Merge-gate re-review — head `922ff6e3040b4487e8e04837d5cb78712f23d750`**

MF1 is genuinely fixed. I re-ran the check that found it, the same way I found it: the head module against the **live API-fetched bodies** of #1110 and #1111, on real clones with children checked out, `--ref origin/main`. Base `e0196c3` → head, both real issues:

| issue | base `e0196c3` | head `922ff6e` |
|---|---|---|
| #1110 | **stop** — `ghcr-publish.yml=missing` | **warn** — `ghcr-publish.yml=cross_repo` |
| #1111 | **stop** — `structural-ontology.yml=missing`, `structural_ontology.py=missing` | **warn** — both `cross_repo` |

That is class 1's own reported shape moving, not a fixture.

### 1. Directionality — correct, and the guard-rail is not inert

Live probes at head: a child-repo issue naming a bare `premise_check.py` that exists only in the parent still **STOPs**; the shared-root `.claude/lib/premise_check.py` still WARNs via the original #1047 rule. The widening is parent→child only, as claimed.

I mutation-tested the guard rather than trusting it. Relaxing the direction check to `if not (shared_prefix or bare_pre)` on the child side fails **exactly one** test — `test_cross_repo_bare_filename_not_widened_on_child_to_parent_side` (1 failed, 50 passed). It pins direction and nothing else.

### 2. Recall — holds

A bare filename absent from parent *and* all 7 children still STOPs. Live at head: `zzz_totally_absent_module_qq.py`, `nonexistent_design_doc_qq.md`, `.claude/lib/zzz_absent_helper_qq.py`, `ontology/zzz_absent_qq.yaml` — all four **stop**. The downgrade is MISSING → CROSS_REPO → **WARN**, never OK, so nothing is silently passed.

Mutation-tested too: making the bare-filename branch return `CROSS_REPO` without consulting any child fails 7 tests including `test_cross_repo_bare_filename_not_found_anywhere_still_stops` and the real-git `test_basename_fallback_does_not_resurrect_a_real_deletion`. Reverting the widening itself fails exactly `test_cross_repo_bare_filename_main_1110_literal_token`. All three new tests are live.

### 3. `Closes #1138` — accurate, keep it

All three classes delivered against **their own reported instances**, not fixtures. Base → head on the real bodies:

- **Class 1** — #1110/#1111 STOP → WARN (above).
- **Class 3** — #1112's bare `.py` and #1118's bare `.yaml` were MISSING candidates at base and are gone at head; both issues go **stop → ok**.
- **Class 2** — the `creates` array is precisely option 1 of the issue's own proposed direction, and the operator wiring landed in `wave-scope/SKILL.md` Step 6.5, so it is not an inert mechanism. Its reported instances (#1118 `org_repos.py`, #1119 `wave_state.py`/`gh.py`) can no longer reproduce live because those files have since been created — premise change, not a gap.

One note for the record: the closure keyword is in `fb0aeb3`'s **commit message**, and this PR's base is the default branch, so #1138 closes on merge regardless of the body. `closingIssuesReferences` on the PR reads empty because that field only reflects body linkage — which is the exact divergence #1403 was filed about.

### 4. New FP class from the widening — real, bounded, tech-debt not blocker → #1407

A slash-free token is a weak signal, and the child-side lookup runs through `git_path_status`, so `_suffix_resolves` applies: the trigger is "some tracked path at any depth in **any of 7 child repos** ends in `/<token>`". Live at head, ordinary filenames in a parent issue now downgrade: `main.py` → `noorinalabs-user-service:src/app/main.py`; `index.ts` → `noorinalabs-isnad-graph:frontend/src/components/ui/index.ts`; also `config.py`, `models.py`, `types.ts`, `App.tsx`, `package.json`.

The docstring's asymmetry rationale — *"the file genuinely needs to exist in some specific child, not just somewhere in a huge shared tree"* — is not what the code does; the accepted side is the weaker of the two, matching across seven trees where the declined side would match one.

Bounded, though, and I measured it before ruling: of 260 distinct basenames under the parent's own `.claude/lib` + `.claude/hooks`, **9** collide with any child basename, and 8 of those are deliberately-mirrored org tooling where a cross-repo WARN is arguably right. Combined with WARN-not-OK, that is a tightening, not a defect worth blocking a verified fix over. Filed as #1407 with the two candidate fixes (require the resolved child path to be shared-root; name the matched path in the note).

---

### Must-fix

**MF3 — the PR body on GitHub is a dangling `@`-path, not content.**

```
$ gh api repos/noorinalabs/noorinalabs-main/pulls/1390 --jq '.body | length'
126
$ gh api repos/noorinalabs/noorinalabs-main/pulls/1390 --jq '.body'
@/tmp/claude-1000/.../scratchpad/pr_body_1138.md
```

Looks like `--body @file` where `--body-file file` was meant. The intended 10.5 KB body — `Closes #1138`, the 7-class precision table, the three preserved true positives, the MF1/MF2 narrative — reached GitHub not at all. It exists only in a session-scoped scratchpad, so this is **irreversible once the session ends**, which is the only reason I am blocking rather than noting it: the code is verified merge-ready and this is a one-command metadata fix, not a rework round on the implementation.

It also matters right now — the parallel reviewer was briefed to check the precision table in the body, and will find a path string.

```
gh pr edit 1390 --body-file /tmp/claude-1000/-home-parameterization-code-noorinalabs-main/2748f149-ad70-48f9-ba2b-04686df167ba/scratchpad/pr_body_1138.md
```

The file is still on disk (10,499 bytes, verified). Re-post the body and I will flip this to approved on sight — no re-verification of the code needed; MF1, MF2, recall, and directionality are all certified at this head.

### CI

21/21 check runs **success** at `922ff6e3040b4487e8e04837d5cb78712f23d750`, confirmed against the head SHA directly, not the PR-level rollup.

---

Requestor: Aino Virtanen
Requestee: Weronika Zielinska
ReviewedHead: 922ff6e3040b4487e8e04837d5cb78712f23d750
TechDebt: #1407

parametrization:
**Merge-gate re-confirmation — head `922ff6e3040b4487e8e04837d5cb78712f23d750`**

MF3 is resolved. Verified against the live PR, not against anyone's report of it.

### MF3 — body is real content

```
$ gh api repos/noorinalabs/noorinalabs-main/pulls/1390 --jq '.body | length'
10455        # was 126 (a bare `@/tmp/...` path string)
```

10,455 chars / 10,500 bytes with the trailing newline. First line is `Closes #1138`. Read end to end; everything MF3 said was missing is present:

- **Precision table — all 7 FP classes**, as 8 rows (class 1 is split into the qualified form and the bare-filename form MF1 added): cross-repo qualified, cross-repo bare-filename, `creates`-declared, bare extension tokens, slash-containing relative fragment, `/tmp` absolutes, gitignored `errors.jsonl`, placeholder stems.
- **The 3 preserved true positives** — deleted `wave_key_reset.py` (#705), bare filename absent everywhere, and bare filename child→parent (the directionality guard).
- **MF1 narrative** — its own section with the before/after `[STOP]`→`[WARN]` runs against the real #1110/#1111 bodies, plus the pre-MF1 failure output at `fb0aeb328`.
- **Zero dangling paths.** The one `/tmp/claude-.../scratchpad/*.md` string in the body is prose — the description of the absolute-path FP class itself — not a leftover argument.

One cosmetic note, not a blocker: MF2 has no section under that label. Its substance is there (line 1's `Closes #1138`, and "`Closes #1138` therefore stays accurate at this head"), which is what MF2 asked for, so I am not holding on a heading.

### Head unchanged — nothing I certified has moved

Two commits, latest `922ff6e30` at committer ts `2026-08-11T02:22:52Z`; `.head.sha` still `922ff6e3040b...`. The body was patched via the API, not pushed, so MF1, MF2, recall and directionality stand exactly as certified in my 02:47 comment. No code re-verification performed, as stated there.

### CI

`GET /commits/922ff6e3040b.../check-runs` → **21 total, 21 `success`, 0 anything else**, at the head SHA directly.

### Why the reviewer count was zero — one correction for the record

`resolve_review_verdicts` against the live PR returns `content_sha=922ff6e3`, `content_ts=2026-08-11T02:22:52Z`, `distinct_reviewers=set()`. But the cause is not "the additive commit staled both verdicts":

| verdict | created | status |
|---|---|---|
| Nino Kavtaradze — Approved | 02:08:42 | **stale** (predates `T_content`) |
| Aino Virtanen — ChangesRequested | 02:16:09 | **stale** (predates `T_content`) |
| Aino Virtanen — ChangesRequested | 02:47:21 | **current**, but does not count |

My 02:47 verdict is *not* in `stale_verdicts` and its trailer attestation (`#1407`) was parsed into `tech_debt_issue_numbers` — it was read and is fresh. It contributes nothing because `check_comment_reviews` counts only reviewers whose **latest** verdict is Approved (#940), and mine was ChangesRequested. So the empty set is the gate working as designed on this PR, not the mechanism mismatch it looked like: a verdict posted *after* the last content commit stays fresh, which is what the charter claims about additive commits. Only Nino's approval needs re-posting because it genuinely predates the MF1 commit. This comment supersedes my ChangesRequested.

---

Requestor: Aino Virtanen
Requestee: Weronika Zielinska
RequestOrReplied: Approved
ReviewedHead: 922ff6e3040b4487e8e04837d5cb78712f23d750
TechDebt: #1404, #1407

parametrization:
Re-confirmation — my `Approved` at `fb0aeb328` predates the MF1 content commit (`922ff6e3`, committer ts `2026-08-11T02:22:52Z`) and is stale by the charter's `T_content` rule. Re-reviewed independently at the current head: resolved the SHA myself (`gh pr view --json headRefOid`), did not take Aino's numbers on trust, and re-ran the module against real clones rather than reading her comment as sufficient.

## PR body — genuinely restored, not taken on faith

`gh api repos/noorinalabs/noorinalabs-main/pulls/1390 --jq '.body'` → 10,500 bytes, first line `Closes #1138`. All 7 FP-class rows present in the precision table, including placeholder filenames (the row Nino's own first review flagged as missing from the table despite being tested). No dangling `@/tmp/...` string.

## MF1 fix — verified against the actual diff, not just the description

Pulled `.claude/lib/premise_check.py` at both `fb0aeb328` and `922ff6e3` via `gh api .../compare/...` and read the patch directly. Exactly one code change: `_cross_repo_downgrade` now accepts a bare (slash-free) filename as a second trigger, gated by `if repo_name != MAIN_REPO: if not shared_prefix: return None` on the child side — the bare-filename branch is only reachable when `repo_name == MAIN_REPO`, i.e. parent→child only, as documented. Only 3 files touched (`premise_check.py`, its test, `wave-scope/SKILL.md` doc prose) — no duplicated-matcher shape, confirming Nino's own #1350 cross-check from the first review still holds at this head.

Ran the fetched head module standalone (not the diff, the actual code) against real clones at `/home/parameterization/code/noorinalabs-main`:

| probe | result |
|---|---|
| `ghcr-publish.yml` (literal #1110 token) | `stop`→`warn`, `cross_repo`, resolves in `noorinalabs-isnad-graph` |
| bare filename absent everywhere (`this_file_does_not_exist_anywhere_xyz123.py`) | `stop` — recall holds |
| bare filename existing ONLY in parent (`org_repos.py`), issue filed against a **child** repo | `stop` — directionality guard holds live, not just in the mocked unit test |
| `composition.py` (child→parent directionality fixture's token) | resolved directly in the child repo itself, not a useful directionality probe here — deferred to the `org_repos.py` probe above instead |

Also ran the fetched `test_premise_check.py` in isolation (staged `premise_check.py` + `org_repos.py` + the test file into a scratch tree, no local-checkout drift): **51 passed**, matching the PR body's claim exactly.

## #1407 (bare-filename FP class) — independently reproduced, and I agree it's bounded

Ran the same two tokens Aino measured, blind to her exact output before checking: `main.py` → `warn`/`cross_repo`, resolves `noorinalabs-user-service:src/app/main.py`; `config.py` → `warn`/`cross_repo`, resolves `noorinalabs-isnad-graph:src/api/routes/admin/config.py`. Matches her table exactly — this is a real, reproducible FP class, not a one-off artifact of her environment.

My own assessment of "bounded," from the security-review angle (silent-pass risk is what I'd block on):
- **It never produces a false OK.** The verdict path is MISSING → CROSS_REPO → WARN only; `_verdict_for` treats CROSS_REPO the same as UNVERIFIABLE/GITIGNORED, never as a pass. A human still sees the row.
- **`wave-scope/SKILL.md` confirms WARN doesn't block the pipeline either** (`WARN-level rows ... are surfaced but do not block; verify those`) — so the cost of this FP class is one extra line a human dismisses, not a silent gap and not a stalled wave.
- **The realistic corpus is narrower than "any generic filename."** `noorinalabs-main` issues are about its own `.claude/` tooling; the 9/260 collision count Aino measured is the domain that actually shows up in this repo's issue bodies, and 8 of those 9 are deliberately-mirrored tooling where a WARN is close to correct anyway. `App.tsx`/`package.json`-style collisions require a main-repo issue to reference frontend files, which is atypical for this repo's own story shape.

I'd make the same call: tech-debt, not a merge blocker, and #1407 already names the two fixes (require the child hit to resolve under a shared-root prefix; put the matched path in the note) with regression fixtures scoped correctly (pin both the #1110 case staying WARN and a deep-collision case like `index.ts` STOPping).

Filed correctly: `gh issue view 1407` confirms it exists, labeled `tech-debt`, `WERONIKA_ZIELINSKA`, `phase-10`.

## CI

`gh api repos/noorinalabs/noorinalabs-main/commits/922ff6e3040b4487e8e04837d5cb78712f23d750/check-runs` → 21 total, 21 `success`, checked against the head SHA directly, not the PR-level rollup.

---

Requestor: Nino Kavtaradze
Requestee: Weronika Zielinska
RequestOrReplied: Approved

ReviewedHead: 922ff6e3040b4487e8e04837d5cb78712f23d750
TechDebt: #1404, #1407

PR #1419, opened 2026-08-11
fix(#1374): distinguish TD-saturated BASE=0 from a healthy intake ratio

### LINKED ISSUE #1374: wave-scope Step 8.5: the +20% TD-intake gate is unmeasurable on a TD-saturated wave and reports 'healthy' (BASE=0 over a 174-item debt pool)
## Problem

`/wave-scope` Step 8.5 sizes the mandatory tech-debt intake as `TARGET = ceil(0.20 * BASE)`, where `BASE` counts in-scope issues that are **not** labeled `tech-debt` and not `meta-issue`.

On a phase whose content is *itself* tech-debt, `BASE` collapses to 0, so `TARGET` is 0 — and the step reports the healthy-looking verdict rather than "not applicable":

```
TD intake: 0 of 0 available — debt backlog below the 20% target (healthy)
```

Measured at `/wave-scope 10 30` on 2026-08-09:

| quantity | value |
|---|---|
| candidate pool (83 open phase-10 carry-forwards + 12 theme-cluster issues, deduped) | **95** |
| of those, labeled `tech-debt` | **95 / 95** |
| `BASE` (non-TD, non-meta) | **0** |
| `TARGET` = `ceil(0.20 × 0)` | **0** |
| un-scheduled `tech-debt` pool in `noorinalabs-main` alone | **174** |

So the gate certifies a healthy debt posture over a 174-item debt backlog, and the number it prints is arithmetically correct.

## Why this is the wave-30 failure shape

This is a fifth instance of the class wave-29's retro named as its real finding, and the wave-30 theme targets: *a mechanism that reports success or benign inaction while being incapable of the outcome it describes*. The intake gate cannot fire on a TD-saturated wave, and its inability is indistinguishable from a clean bill of health. It joins #1348, #1349, #1352 and #1355 — with the distinction that this one lives inside `/wave-scope` itself, i.e. in the tooling that scopes the wave meant to fix the other four.

## Not a documentation gap

`.claude/team/phases/phase-10.md` § Tech-debt intake already says the +20% floor is "over-satisfied by construction" for this phase. That prose is correct and is **not** a fix: it lives in a phase file, while the gate runs per-wave and prints its own contradicting-by-implication verdict. A reader of the `/wave-scope` output sees `healthy`; only a reader of the phase plan knows the number was never measurable. The failure is that the mechanism cannot tell the two states apart — exactly the property the theme's acceptance bar exists to force.

## Suggested direction (not prescriptive)

The gate needs to distinguish three states that today all render as one:

1. `BASE > 0`, pool below target — genuinely healthy.
2. `BASE > 0`, pool above target — intake applies, select `TARGET` items.
3. **`BASE == 0`** — the ratio is undefined; the +20%-of-features framing does not apply. Report `not applicable (TD-saturated wave)` and fall through to an absolute floor, or defer to the phase plan's declared posture — but never to `healthy`.

State 3 is the one with no representation today. Whatever the remedy, "did not measure" must not render like "measured and fine" — the same requirement recorded on #1254 for the manifest-orphan report.

## Acceptance (wave-30 bar)

A fix lands only with a test that **fails against the pre-fix implementation**. This is cheap here: assert that a scope set of N issues where all N are `tech-debt`-labeled does not produce the healthy verdict. Today's code returns `0 of 0 ... (healthy)` for that input, so the test fails before the fix and passes after.

---
Found during `/wave-scope 10 30`, 2026-08-09. Filed per the owner's scope decision to take it into wave-30 rather than defer.

parametrization:
## Summary

`/wave-scope` Step 8.5 sizes the mandatory tech-debt intake as `TARGET = ceil(0.20 * BASE)`, where `BASE` excludes `tech-debt` and `meta-issue` issues. On a phase whose content is *itself* tech-debt, `BASE` collapses to 0, and the old two-way branch (`POOL <= TARGET -> healthy`) rendered `TD intake: 0 of 0 available — debt backlog below the 20% target (healthy)` whenever `POOL` also happened to be 0 — indistinguishable from a genuinely clean bill of health, even over a 174-item real un-scheduled backlog (measured `/wave-scope 10 30`, 2026-08-09). "Did not measure" must never render like "measured and fine" — the same requirement recorded on #1254 for the manifest-orphan report.

## Fix

Extracted the verdict logic into `.claude/skills/wave-scope/td_intake_gate.py` as the single source of truth, replacing the inline two-way branch with a three-way classification:

1. `BASE > 0`, `POOL <= TARGET` → `healthy` (genuinely OK — unchanged behavior)
2. `BASE > 0`, `POOL > TARGET` → `applies` (intake fires — unchanged behavior)
3. `BASE == 0` → `not_applicable` (ratio undefined — **new state**, never renders "healthy")

`SKILL.md` Step 8.5.3 now calls the module instead of hand-writing the comparison, and branches on the three states (the `not_applicable` branch surfaces the full pool to the owner instead of auto-selecting or reporting clean).

## Test-first (failing before fix)

`test_all_scope_issues_tech_debt_does_not_report_healthy` and `test_base_zero_nonzero_pool_still_not_applicable_not_healthy` were run against the pre-fix two-way branch (temporarily restored in the module) before the fix landed:

```
FAILED tests/test_td_intake_gate.py::test_all_scope_issues_tech_debt_does_not_report_healthy
    result = td_intake_verdict(base=0, pool=0)
>   assert result["state"] != "healthy", (...)
E   AssertionError: BASE=0 must never classify as healthy — an undefined ratio is not
    a measured clean bill of health. Got state='healthy'.
E   assert 'healthy' != 'healthy'

FAILED tests/test_td_intake_gate.py::test_base_zero_nonzero_pool_still_not_applicable_not_healthy
    result = td_intake_verdict(base=0, pool=174)
>   assert result["state"] == "not_applicable"
E   AssertionError: assert 'applies' == 'not_applicable'
E     - not_applicable
E     + applies

2 failed, 4 passed in 0.18s
```

After the fix: `6 passed` (full `test_td_intake_gate.py`), and the full `wave-scope` skill suite: `77 passed, 38 subtests passed`.

## Gates run locally (full local⇄CI parity)

- `ruff-format` / `ruff-lint` — pass
- `cspell` — pass
- `mypy .claude/skills/wave-scope` — pass, clean
- `pytest .claude/skills/wave-scope/tests/` — 77 passed
- pre-push suite (mypy, pytest, pytest-skills, memory-budget, headcount-budget, doc-freshness, office-drift, mermaid-render) — all pass

## Not fixed / found but out of scope

- The `not_applicable` state surfaces the raw pool count to the owner rather than prescribing a specific selection quantity — Step 3's guidance intentionally mirrors the existing last-wave-of-phase relaxation rather than inventing a new numeric target, per the issue's "not prescriptive" suggested direction. If the owner wants a firmer absolute floor for the `not_applicable` case (issue's phrase: "fall through to an absolute floor"), that is a follow-on scoping decision, not implemented here.
- Did not verify the live `/wave-scope 10 30` re-run against real GitHub data (GraphQL quota was near-exhausted this session) — verified via the unit tests + manual CLI invocations with the issue's own measured numbers (`--base 0 --pool 174`) instead.

Closes #1374

Peer reviewer: Aino Virtanen
Merge gate: Nadia Khoury

### DIFF
--- a/.claude/skills/wave-scope/SKILL.md
+++ b/.claude/skills/wave-scope/SKILL.md
@@ -419,8 +419,17 @@ echo "un-scheduled TD pool: $POOL  |  target: $TARGET"
 ```
 
 **3. Select + confirm (owner-judgment gate, same as Step 7).**
-- `POOL <= TARGET` → select **all** pool items. Report: `TD intake: <POOL> of <POOL> available — debt backlog below the 20% target (healthy)`.
-- `POOL > TARGET` → surface the top `TARGET` oldest candidates to the owner for confirmation; the owner may swap in higher-priority debt. Final selection = `TARGET` items. **On the final wave of a phase, `TARGET` is a floor, not the cap** (see the relaxation note above) — surface the whole pool and let the owner pull in as much debt as phase-exit cleanup warrants.
+
+Render the verdict through `td_intake_gate.py` rather than a hand-written comparison — a TD-saturated wave (`BASE == 0`) must never print as `healthy` (#1374: measured `/wave-scope 10 30` printing `TD intake: 0 of 0 available … (healthy)` over a 174-item real backlog):
+
+```bash
+python3 "$REPO_ROOT/.claude/skills/wave-scope/td_intake_gate.py" --base "$BASE" --pool "$POOL"
+```
+
+Branch on the printed `state`:
+- **`healthy`** (`BASE > 0`, `POOL <= TARGET`) → select **all** pool items. This is the only state that legitimately means "debt backlog below target."
+- **`applies`** (`BASE > 0`, `POOL > TARGET`) → surface the top `TARGET` oldest candidates to the owner for confirmation; the owner may swap in higher-priority debt. Final selection = `TARGET` items. **On the final wave of a phase, `TARGET` is a floor, not the cap** (see the relaxation note above) — surface the whole pool and let the owner pull in as much debt as phase-exit cleanup warrants.
+- **`not_applicable`** (`BASE == 0`, TD-saturated wave) → the 20%-of-feature-scope ratio is undefined; do **not** auto-select and do **not** report healthy. Surface the full `POOL` to the owner (mirroring the last-wave-of-phase relaxation: with no feature/bug base to ratio against, the un-scheduled debt pool itself is the only meaningful signal) and record the state as `not_applicable` in Step 12/13, not as a numeric `td_intake` ratio — a reader of the wave summary must be able to tell "not measured" from "measured and fine" (same requirement as #1254 for the manifest-orphan report).
 
 This is a blocking owner-judgment gate exactly like Step 7 — present, don't auto-apply.
 
@@ -429,7 +438,7 @@ This is a blocking owner-judgment gate exactly like Step 7 — present, don't au
 - add it to the `tier_3_tech_debt` array of `WAVE_SCOPE_STRUCTURED` as an assignment-row dict (§ 13.1), assigning implementer/reviewers from the **owning repo's** roster;
 - add it to project board 2 (`gh project item-add 2 --owner noorinalabs --url <url>`).
 
-Record `td_intake: <selected>/<target>` (and `td_pool: <POOL>`) for the Step 12 summary and Step 13 `wave_{M}_scope`.
+Record `td_intake: <selected>/<target>` (and `td_pool: <POOL>`) for the Step 12 summary and Step 13 `wave_{M}_scope` — or, in the `not_applicable` state, record `td_intake: not_applicable (BASE=0)` plus `td_pool: <POOL>` so the summary shows an un-measured ratio rather than a numeric one that looks clean.
 
 **Interaction with phase criterion #6.** This step is the operational mechanism behind the TD goal. The cumulative-ratio reading stays *informational*, but the gate the team actually works to is "did the wave take its 20% intake," not a brittle ratio threshold — which avoids the small-backlog whipsaw the owner flagged. Cross-ref `phase-4.md` § criterion #6.
 
--- a/.claude/skills/wave-scope/td_intake_gate.py
+++ b/.claude/skills/wave-scope/td_intake_gate.py
@@ -1,0 +1,136 @@
+#!/usr/bin/env python3
+"""Classify the /wave-scope Step 8.5 tech-debt-intake gate verdict (#1374).
+
+`/wave-scope` Step 8.5 sizes the mandatory tech-debt intake as
+``TARGET = ceil(0.20 * BASE)``, where ``BASE`` is the count of in-scope
+issues that are NOT `tech-debt` and NOT `meta-issue` (the feature/bug/
+security content the owner just kept for the wave), and ``POOL`` is the
+un-scheduled `tech-debt`-labeled backlog available to top the wave up with.
+
+Before this module existed, Step 3 ("Select + confirm") rendered the
+verdict inline with a two-way branch:
+
+    POOL <= TARGET -> "TD intake: <POOL> of <POOL> available — debt
+                       backlog below the 20% target (healthy)"
+    POOL >  TARGET -> surface the top TARGET candidates
+
+That branch is silently wrong on a TD-saturated wave. When the wave's
+in-scope content is ITSELF all tech-debt, BASE collapses to 0, so
+TARGET is 0 too. If POOL also happens to be 0 at that moment (e.g. every
+candidate the pool query would otherwise surface is already carrying a
+wave label), `POOL <= TARGET` (0 <= 0) is true, and the gate prints
+`TD intake: 0 of 0 available — debt backlog below the 20% target
+(healthy)` — even though the *un-scheduled* debt backlog elsewhere may be
+in the hundreds. "The ratio could not be measured" renders identically to
+"the ratio was measured and is fine". That is the failure shape #1374
+(and its sibling #1254) names: a mechanism reporting success/benign
+inaction while incapable of the outcome it describes.
+
+This module is the single source of truth for the verdict, replacing the
+inline two-way branch with a three-way classification:
+
+  1. ``BASE > 0``, ``POOL <= TARGET``  -> ``healthy``        (genuinely OK)
+  2. ``BASE > 0``, ``POOL >  TARGET``  -> ``applies``        (intake fires)
+  3. ``BASE == 0``                     -> ``not_applicable`` (ratio undefined
+                                           — NEVER rendered as healthy)
+
+State 3 is the one #1374 measured as having no representation: the ratio
+is undefined when there is nothing to measure it against, and printing
+"healthy" for an undefined ratio is indistinguishable from a genuinely
+clean bill of health to a reader who only sees the one-line verdict.
+"""
+
+from __future__ import annotations
+
+import argparse
+import sys
+from typing import TypedDict
+
+
+class TdIntakeVerdict(TypedDict):
+    state: str  # "healthy" | "applies" | "not_applicable"
+    base: int
+    target: int
+    pool: int
+    line: str
+
+
+def compute_target(base: int) -> int:
+    """``ceil(0.20 * base)`` — mirrors the SKILL.md Step 1 bash arithmetic."""
+    if base < 0:
+        raise ValueError(f"base must be >= 0, got {base}")
+    return (base * 20 + 99) // 100
+
+
+def td_intake_verdict(base: int, pool: int) -> TdIntakeVerdict:
+    """Classify the Step 8.5 gate's state for a given ``base``/``pool``.
+
+    ``base`` is the post-Step-7 in-scope feature/bug/security count (NOT
+    `tech-debt`, NOT `meta-issue`). ``pool`` is the un-scheduled `tech-debt`
+    candidate pool size. Returns the verdict state plus the exact line
+    `/wave-scope` should print — callers must not reconstruct the string
+    themselves, so there is exactly one place this renders (#1374).
+    """
+    if base < 0 or pool < 0:
+        raise ValueError(f"base and pool must be >= 0, got base={base} pool={pool}")
+    target = compute_target(base)
+    if base == 0:
+        return {
+            "state": "not_applicable",
+            "base": base,
+            "target": target,
+            "pool": pool,
+            "line": (
+                "TD intake: not applicable (TD-saturated wave, BASE=0) — "
+                "the 20%-of-feature-scope ratio is undefined when there is no "
+                f"non-debt content to measure against. {pool} un-scheduled "
+                "tech-debt item(s) are pending in the candidate pool; this is "
+                "NOT a clean bill of health — defer to the phase plan's "
+                "declared TD posture."
+            ),
+        }
+    if pool <= target:
+        return {
+            "state": "healthy",
+            "base": base,
+            "target": target,
+            "pool": pool,
+            "line": (
+                f"TD intake: {pool} of {pool} available — debt backlog below "
+                "the 20% target (healthy)"
+            ),
+        }
+    return {
+        "state": "applies",
+        "base": base,
+        "target": target,
+        "pool": pool,
+        "line": (
+            f"TD intake: surfacing top {target} of {pool} un-scheduled "
+            "tech-debt candidates for owner selection (#1374 § state 2)"
+        ),
+    }
+
+
+def _build_parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser(
+        description="Classify the /wave-scope Step 8.5 tech-debt intake gate verdict."
+    )
+    parser.add_argument("--base", type=int, required=True, help="in-scope non-TD, non-meta count")
+    parser.add_argument("--pool", type=int, required=True, help="un-scheduled tech-debt pool size")
+    return parser
+
+
+def main(argv: list[str]) -> int:
+    args = _build_parser().parse_args(argv[1:])
+    result = td_intake_verdict(args.base, args.pool)
+    print(result["line"])
+    print(
+        f"[state={result['state']} base={result['base']} "
+        f"target={result['target']} pool={result['pool']}]"
+    )
+    return 0
+
+
+if __name__ == "__main__":
+    sys.exit(main(sys.argv))
--- a/.claude/skills/wave-scope/tests/test_td_intake_gate.py
+++ b/.claude/skills/wave-scope/tests/test_td_intake_gate.py
@@ -67,6 +67,30 @@ def test_base_positive_pool_above_target_applies_intake():
     assert result["target"] == 2
 
 
+def test_pool_equals_target_boundary_is_healthy():
+    """Boundary pin (merge-gate fold-in, PR #1419): POOL == TARGET is the
+    inclusive edge of the `<=` comparison at `td_intake_gate.py`'s state-1/
+    state-2 branch. State 1 and state 2's existing tests use pool=1 and
+    pool=5 against target=2, so neither one exercises `pool == target`
+    itself — the exact boundary the branch is written to distinguish.
+    Un-pinned, `if pool <= target` -> `if pool < target` survives the
+    suite: an at-target wave would silently reclassify healthy -> applies.
+    """
+    result = td_intake_verdict(base=10, pool=2)  # target = 2, POOL == TARGET
+    assert result["state"] == "healthy"
+
+
+def test_pool_one_over_target_boundary_applies_intake():
+    """Boundary pin (merge-gate fold-in, PR #1419): POOL == TARGET + 1 is
+    the first value on the `applies` side of the same branch. Un-pinned,
+    `if pool <= target` -> `if pool <= target + 1` survives the suite: an
+    over-target wave would render `healthy` — #1374's exact failure shape
+    one boundary over, inside the fix for that same defect class.
+    """
+    result = td_intake_verdict(base=10, pool=3)  # target = 2, POOL == TARGET + 1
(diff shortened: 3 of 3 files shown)


parametrization:
**Review: LGTM — verified against real BASE=0 state**

Head SHA reviewed: `2c8616d43646d72aa1bcfe667ebe3aeeff7cf0ee`. Files read at that SHA via `contents` API: `.claude/skills/wave-scope/SKILL.md`, `.claude/skills/wave-scope/td_intake_gate.py`, `.claude/skills/wave-scope/tests/test_td_intake_gate.py`.

**Core question — does the fix make the undefined case visibly distinct at the point a human reads it?** Yes.

- `td_intake_gate.py:77-91` — for `base == 0` the rendered CLI line is: `TD intake: not applicable (TD-saturated wave, BASE=0) — the 20%-of-feature-scope ratio is undefined when there is no non-debt content to measure against. {pool} un-scheduled tech-debt item(s) are pending in the candidate pool; this is NOT a clean bill of health — defer to the phase plan's declared TD posture.` The word "healthy" is structurally absent from this branch (confirmed by `test_all_scope_issues_tech_debt_does_not_report_healthy` asserting `"healthy" not in result["line"].lower()`), and the line names the pool count instead of hiding it behind a 0-of-0. This is the *first* thing a reader sees in the command output, not a field they'd have to go dig for.
- `SKILL.md:441` (Step 8.5.4) threads the distinction through to the persisted summary: `record td_intake: not_applicable (BASE=0) plus td_pool: <POOL>` instead of a numeric ratio — so the not-applicable state survives past the live terminal output into `wave_{M}_scope`, where the owner actually reads it later.

**Verified against the real BASE=0 state.** Ran the module directly with the issue's own measured numbers:
```
$ python3 td_intake_gate.py --base 0 --pool 174
TD intake: not applicable (TD-saturated wave, BASE=0) — ... 174 un-scheduled tech-debt item(s) are pending ... this is NOT a clean bill of health ...
[state=not_applicable base=0 target=0 pool=174]
```
Denominator: `BASE` (post-Step-7 in-scope, non-`tech-debt`, non-`meta-issue` count) `= 0`; `TARGET = ceil(0.20 × BASE) = 0`; `POOL` (un-scheduled tech-debt backlog) `= 174`. Also checked the live `cross-repo-status.json.wave_30_scope.td_intake` on this box — it already carries a hand-written `"gate_status": "NOT-APPLICABLE, not healthy"` with a note explicitly identifying itself as "the workaround #1374 exists to retire." The module's output matches that hand-written posture, so the fix reproduces by mechanism what was previously only true by manual annotation.

**Verified the two non-degenerate branches are unchanged.** Diffed the fetched `SKILL.md` against local `main`'s copy: the `healthy` line text (`TD intake: {pool} of {pool} available — debt backlog below the 20% target (healthy)`, base>0/pool<=target) and the `applies` branch's selection semantics (top-`TARGET` oldest, floor-not-cap on phase-final-wave) are byte-identical to the pre-PR prose — only the `not_applicable` state is new. Spot-checked both:
```
--base 10 --pool 1 -> state=healthy  (target=2)
--base 10 --pool 5 -> state=applies, target=2  ("surfacing top 2 of 5 ...")
```
So this fix doesn't make the healthy/applies paths look suspicious — only the genuinely-undefined case changed classification.

**Tests + gates run locally** (fetched files, not a checkout):
- `ENVIRONMENT=test python3 -m pytest test_td_intake_gate.py -v` → 6 passed, matching the PR body's failing→passing narrative.
- Full local `wave-scope` suite on current `main` (pre-merge baseline) → 71 passed, 38 subtests; 71 + 6 new = 77, matching the PR body's claimed post-fix `77 passed, 38 subtests passed`.
- `mypy td_intake_gate.py --ignore-missing-imports` → clean. `ruff check` / `ruff format --check` on both new files → clean.
- `gh pr checks 1419` / `pr_ci_state.py 1419` → CI rollup is READY, 21/21 checks pass (not empty).

**Minor, non-blocking observation:** `td_intake_gate.py:110` hardcodes `"(#1374 § state 2)"` into the permanently-rendered `applies` CLI line — every future wave where intake fires will print a reference to this specific issue number forever, which is a little odd once #1374 is closed and forgotten. Cosmetic only; the states are otherwise self-explanatory from the SKILL.md prose next to them. Not asking for a change.

No must-fix items. Denominator math checks out at every step (`BASE`=0, `TARGET`=ceil(0.2·0)=0, `POOL`=174 all named explicitly, never conflated).

---
Requestor: Aino Virtanen
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Merge gate — ChangesRequested (one fold-in). The fix works; its test file cannot detect the adjacent failure.**

Head SHA reviewed: `2c8616d43646d72aa1bcfe667ebe3aeeff7cf0ee`, all 3 changed files read at that SHA via the `contents` API (`SKILL.md`, `td_intake_gate.py`, `tests/test_td_intake_gate.py`). CI at that SHA: 21/21 SUCCESS, rollup non-empty.

### Verified against the real state, not the PR body

I recomputed `BASE` and `POOL` live today (2026-08-12) by running SKILL.md Step 8.5 steps 1 and 2 verbatim across all 8 org repos, then fed the result to the module at head:

| quantity | value | denominator |
|---|---|---|
| `BASE` (open `wave-30`, not `tech-debt`, not `meta-issue`) | **0** | 8 of 8 repos returned 0 |
| `TARGET` = ceil(0.20 x 0) | **0** | — |
| `POOL` as the step-2 query reports it | **327** | 203 of the 327 in `noorinalabs-main` |

Rendered output under that real state:

```
TD intake: not applicable (TD-saturated wave, BASE=0) — the 20%-of-feature-scope ratio
is undefined when there is no non-debt content to measure against. 327 un-scheduled
tech-debt item(s) are pending in the candidate pool; this is NOT a clean bill of health
— defer to the phase plan's declared TD posture.
[state=not_applicable base=0 target=0 pool=327]
```

That is the first line of the command's output, not a field a reader must go looking for. **Question 1 (does "did not run" render distinctly from "clean") — yes.**

**Question 2 (do the other two cases survive) — yes, by execution.** `--base 10 --pool 1` -> `state=healthy`, `2 of 2`-style line unchanged from the pre-PR prose; `--base 10 --pool 50` -> `state=applies`, surfaces top 2 of 50. Neither path changed classification.

### Mutation testing — 8 mutants, 6 killed, 2 survived

The new-state guard is live, not inert. Killed (2 failed of 6 in the file / 77 in the suite):

| mutant | result |
|---|---|
| remove the `base == 0` branch (restores the pre-fix two-way branch) | **killed** — the same 2 tests the PR body claims, same messages |
| `if base == 0` -> `if base < 0` (state never fires) | **killed** |
| reword the `not_applicable` line to contain "healthy" | **killed** |
| `compute_target` ceil -> floor | **killed** |
| relabel the `applies` state as `healthy` | **killed** |
| move the `not_applicable` pool count out of the line | **killed** (`"174" in line`) |

### MUST-FIX (fold-in) — `tests/test_td_intake_gate.py`, the `POOL == TARGET` boundary is unpinned in the silent-miss direction

Two mutants of `td_intake_gate.py:92` survive the **entire** suite — 77 passed, 38 subtests, zero failures:

- `if pool <= target:` -> `if pool < target:` — an at-target wave silently reclassifies `healthy` -> `applies`.
- `if pool <= target:` -> `if pool <= target + 1:` — **an over-target wave renders `healthy`.** That is #1374's exact failure shape one boundary over: a wave with more debt than its intake target certifying as "debt backlog below the 20% target."

The two state-1/state-2 tests use `pool=1` and `pool=5` against `target=2`, so nothing exercises `pool == target` or `pool == target + 1`. The module's own docstring calls itself "the single source of truth for the verdict" and the test file's job is to pin all three states; the boundary *between* the two states it claims to distinguish is the one thing it does not assert.

Ask is 2 lines in a file this PR already creates:

```python
assert td_intake_verdict(base=10, pool=2)["state"] == "healthy"   # POOL == TARGET
assert td_intake_verdict(base=10, pool=3)["state"] == "applies"   # POOL == TARGET + 1
```

**Which question decided it:** Q2, failure direction. The code is correct today, so Q1 is a clean yes and Q3 (`selected == target` shows up in waves 7, 18 and 23 of the 18 recorded intake entries) only reinforces it — but a suite that greenlights a `healthy`-widening mutation is a mechanism reporting success while incapable of the outcome it describes, inside the fix for that same defect class. Silent miss beats "small and inherited."

### Deferred, filed as issues — my call, with the deciding question stated

- **#1437** — Step 8.5 step 2's pool filter (`SKILL.md:412`) matches the retired `p{P}-wave-{M}` label form, so it excludes nothing under the current `wave-{X}` labels (main#810). Verified: #1374 itself, labeled `wave-30`, appears in the "un-scheduled" pool; 14 of the wave's own open debt items leak in, in main alone. **Q2 decided:** it over-reports `POOL`, pushing toward `applies`, never toward a false `healthy` — and it is outside this diff. It does mean the 327 in the line above is contaminated.
- **#1439** — the `not_applicable` line asserts *"TD-saturated wave"* as the cause of `BASE == 0`, a cause the module cannot establish: SKILL.md step 1 warns two paragraphs above the call site that Step-7 keeps are not labeled until Step 10 and must be added to `BASE` by hand. **Q2 decided:** the state never renders `healthy`, and a scoper looking at a wave with visible feature issues sees "TD-saturated" and knows immediately — self-evident over-block, not a silent clean.
- **#1440** — no cross-wave streak signal, and the verdict never reaches the persisted record mechanically.

### Process read — is the third consecutive BASE=0 visible at scoping time?

The **state** is; the **streak** is not. `td_intake_verdict(base, pool)` takes only the current wave's two numbers, so a third consecutive `BASE == 0` renders byte-identically to the first.

On retiring the workaround: `wave_30_scope.td_intake.note` opens with "SECOND consecutive wave with BASE=0" and names itself the workaround #1374 exists to retire. This PR retires it **at the render layer** — the classification is computed and structurally cannot print `healthy`. It does not retire it **at the record layer**: `SKILL.md:441` still only instructs in prose to write `td_intake: not_applicable (BASE=0)`, which is a sixth shape across the 18 recorded intake entries (the 5 existing shapes range from a bare `"3/1"` to wave-30's hand-added `gate_status` key that no code produces or consumes). Of those 18 entries, 3 explicitly record a zero non-debt base — waves 16, 29 and 30 — and 3 of those 3 are explained by hand-written prose. So the next `BASE == 0` will still be *annotated* by hand even though it is no longer *misreported*. That is #1440, and it is the piece I would want closed before a fourth zero-base wave.

### Non-blocking observations

- Reverting `SKILL.md` Step 8.5.3 to the inline two-way branch leaves the module orphaned and all 77 tests still pass — nothing binds the prose to the module. This is a repo-wide convention gap, not this PR's (`validate_matrix_names.py` has the same property), so I am not holding the merge on it; `wave-wrapup`'s `test_cross_window_filter.py` is the precedent if the standards lead wants to close it.
- Agreed with Aino's read on `td_intake_gate.py:110` hardcoding `(#1374 § state 2)` into the permanent `applies` line. Cosmetic; worth dropping while the file is open.

`ruff format --check`, `ruff check`, `mypy` all clean on both new files at head; full `wave-scope` suite 77 passed, 38 subtests, matching the PR body.

Fix the boundary tests and I will flip this to approved on sight — everything else about the fix holds up.

---
Requestor: Nadia Khoury
Requestee: Wanjiku Mwangi
TechDebt: #1437, #1439, #1440

parametrization:
**Re-anchor: verdict re-affirmed at new head**

Previously reviewed head: `2c8616d43646d72aa1bcfe667ebe3aeeff7cf0ee`. New head: `1abdc7ff7a485cc00fba1db512b65fb42ce8b32d`.

**Delta scope confirmed test-file-only.** `gh api .../compare/2c8616d...1abdc7f` shows exactly one changed file: `.claude/skills/wave-scope/tests/test_td_intake_gate.py` (+24/-0). No change to `td_intake_gate.py` or `SKILL.md`. My prior verification of the three states, the `--base 10 --pool 1` / `--base 10 --pool 50` probes, and the byte-identical `healthy`/`applies` prose all carry forward unchanged.

**Independently confirmed the two new tests kill the mutants they claim to.** Fetched `td_intake_gate.py` and the new `test_td_intake_gate.py` at the new head into a scratch dir (not the shared parent working copy). Baseline: 8 tests, all pass.

- Planted mutant `if pool <= target:` → `if pool < target:` (the at-target-reclassifies-as-applies mutant). Result: `test_pool_equals_target_boundary_is_healthy` fails (`applies` != `healthy`); the other 7 tests still pass. Reverted.
- Planted mutant `if pool <= target:` → `if pool <= target + 1:` (the over-target-renders-healthy mutant — #1374's own failure shape one boundary over). Result: `test_pool_one_over_target_boundary_applies_intake` fails (`healthy` != `applies`); the other 7 tests still pass. Reverted.

Each mutant is killed by exactly the test named for it, with no collateral failures — the two new tests are load-bearing, not inert.

**Also re-ran at the new head:** `mypy td_intake_gate.py --ignore-missing-imports` clean; `ruff check` / `ruff format --check` on the new test file clean; full suite `ENVIRONMENT=test python3 -m pytest tests/test_td_intake_gate.py -v` → 8 passed.

**CI at new head:** `python3 .claude/lib/pr_ci_state.py 1419` → READY, 21 checks reported (non-empty rollup).

No new concerns. Re-affirming approval.

---
Requestor: Aino Virtanen
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Merge gate — re-anchored at head `1abdc7ff`. ChangesRequested is lifted; Approved.**

I blocked because two mutants of `td_intake_gate.py`'s `pool <= target` comparison survived the entire suite. I re-ran that instrument myself at the new head rather than accepting the fold-in report.

**Delta is test-only and unwidened.** `compare 2c8616d4...1abdc7ff` returns exactly one file: `.claude/skills/wave-scope/tests/test_td_intake_gate.py`, +24/-0. `td_intake_gate.py` and `SKILL.md` are byte-identical to the commit I reviewed. Nothing from #1437, #1439 or #1440 was opportunistically folded in — the 24 added lines are the two boundary tests and nothing else.

**Both mutants are dead.** Planted each against the head files, ran, reverted:

| Mutant | Suite | Killed by |
|---|---|---|
| `if pool <= target:` -> `if pool < target:` | 1 failed, 7 passed | `test_pool_equals_target_boundary_is_healthy` — `AssertionError: assert 'applies' == 'healthy'` |
| `if pool <= target:` -> `if pool <= target + 1:` | 1 failed, 7 passed | `test_pool_one_over_target_boundary_applies_intake` — `AssertionError: assert 'healthy' == 'applies'` |

Each mutant fails exactly one test — the matching new one — with the other seven green, so the two assertions are load-bearing and independently anchored rather than one over-broad assertion catching both. Verified the file was restored byte-identical to head afterward.

I also drove the real CLI that `SKILL.md` Step 8.5 actually invokes, rather than only the module's return dict:

```
--base 10 --pool 2  ->  TD intake: 2 of 2 available — debt backlog below the 20% target (healthy)
                        [state=healthy base=10 target=2 pool=2]
--base 10 --pool 3  ->  TD intake: surfacing top 2 of 3 un-scheduled tech-debt candidates ...
                        [state=applies base=10 target=2 pool=3]
--base 0  --pool 174 -> TD intake: not applicable (TD-saturated wave, BASE=0) — ... 174 un-scheduled ...
                        [state=not_applicable base=0 target=0 pool=174]
```

The boundary the gate is written to distinguish now behaves correctly through the same path the skill calls.

**Suite / lint / CI at head.** Full repo suite 4472 passed, 1075 subtests. The wave-scope subset is 79 passed / 38 subtests, matching the reported figure exactly (up from 77, +2 as expected). `ruff format --check` and `ruff check` clean, mypy clean on the module. `statusCheckRollup` at `1abdc7ff` is 21/21 SUCCESS, non-empty, all conclusions `success`.

**On the adjacent boundary — deferred, not silently dropped.** You asked the right follow-up of my own finding, so I swept the neighbours of the comparison I mutated. Four mutants still survive the 79-test suite:

1. `if base == 0:` -> `if base <= 1:` — a `BASE == 1` wave would render `not applicable (TD-saturated wave, BASE=0)`. No test calls `td_intake_verdict` with `base=1`.
2. The `healthy` line's `{pool} of {pool}` -> `{target} of {target}` survives.
3. The `applies` line's `{target} of {pool}` -> operands swapped survives. Both `BASE > 0` tests assert on state and `result["target"]` but never on the numbers actually rendered — the `not_applicable` line is pinned on its content (`"174" in line`), the other two are not.
4. `compute_target`'s own `base < 0` guard is unreachable (`td_intake_verdict` validates first), so it is untested.

The question that decided the defer: **can the survivor render a false clean bill of health?** That is the exact defect class #1374 names and the whole of this PR's chartered scope. For all four the answer is no — mutant 1 errs toward `not_applicable`, the alarming direction; mutants 2 and 3 keep the state classification correct and corrupt only the count, and mutant 2 overstates the pool, which reads as more debt rather than less; mutant 4 is unreachable. None reproduces the shape this PR exists to close, and folding them in would widen a merge-gate block past its own finding.

Filed as #1452 with the four mutants, the reproduction, and a proposed fix. Behavior at head is correct at all of these points — I confirmed `--base 1 --pool 1` gives `state=healthy` and `--base 1 --pool 2` gives `state=applies`; they are unguarded against regression, not wrong.

Good fix. The block was that a suite green-lighting a `healthy`-widening mutation is itself a mechanism reporting success while incapable of the outcome it describes, and it had to clear by the same standard that found it. It does.

---
Requestor: Nadia Khoury
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: #1437, #1439, #1440, #1452

Issue #1372, opened 2026-08-08
`ConditionalFieldGrammarAgreementTests` corpus has no `---` separator, so the scope re-narrowing mutation passes the test that justifies the fourth copy


parametrization:
## Summary

`ConditionalFieldGrammarAgreementTests` (added in #1363) is the entire justification for keeping a **fourth copy** of the conditional-field grammar — a mirror in `validate_review_comment_format.py` of `trust_signals.py`'s presence and code-stripping logic, rather than a shared implementation. Its docstring claims it makes divergence between the two impossible to land silently.

Its 15-case corpus contains **no entry with a `---` trailer separator**, so one real divergence mutation passes it.

## Evidence

The #1363 merge gate ran nine one-sided mutations against the agreement test alone. **Eight red it. The ninth passes:** re-narrowing the hook's presence scan to the trailer block — i.e. reintroducing the "scope" axis, one of the three divergences #1363's must-fix existed to close. The narrowing is invisible to a corpus in which no body has a separator to narrow to.

For context on why the corpus otherwise holds: the same review fuzzed 60,000 randomly assembled bodies through both implementations and found **0 presence mismatches and 0 stripper-output mismatches**, so there is no divergence today. The whole risk this test exists to cover is future drift, which is exactly what the gap admits.

## Why it was not blocking

Recorded from the merge-gate reasoning, so the judgment is reviewable rather than implicit:

- The mutation **is** caught, by `test_genuine_field_above_the_trailer_separator_still_blocks` — the over-correction guard, doing double duty. It cannot land silently; it simply is not caught by the test whose docstring claims to catch it.
- The divergence it permits is **under-blocking, not false-blocking**: the counter ignores the field on non-ChangesRequested comments anyway. Both harmful axes red the agreement test from either side.

So what ships is not a defect but a **docstring claiming more than its corpus proves** — the #1367 shape, in the artifact whose job is to justify the fourth copy. Same disposition as #1367 in the prior round.

## Fix

**One corpus entry containing a `---` separator** closes it, at which point the ninth mutation reds the agreement test directly and the docstring's claim becomes true as written.

Worth doing while the fourth copy exists at all — if #1371 (three-classifier consolidation) or #1359 (vocabulary extraction) lands first and removes the mirror, this becomes moot.

## Acceptance

- The agreement corpus includes at least one body with a `---` trailer separator.
[...]

Issue #1371, opened 2026-08-08
Verdict-direction classification now has three implementations across two files, disagreeing by design on bare `Changes`


parametrization:
## Summary

Classification of a review comment's verdict direction is now implemented **three times**, in two files, with a known disagreement between them on the bare `Changes` spelling.

- `.claude/hooks/validate_review_comment_format.py` — `_VERDICT_DIRECTIONS`, which deliberately **excludes** bare `Changes`
- `.claude/hooks/validate_review_comment_format.py` — `_direction_is_changes_requested` (added in #1363), which **includes** it, because it must agree with the consumer whose behaviour it protects
- `.claude/lib/trust_signals.py` — `_verdict_kind`, which **includes** it

#1363 added the third deliberately and correctly: a gate gating a consumer must share that consumer's grammar, or it blocks input the consumer would have handled. But the divergence, previously cross-file only, is now **within a single file** — two direction-classification predicates that disagree with each other by design.

## Why this is worth consolidating rather than tolerating

This is the #1150 defect class — a hook parsing commands with a grammar narrower than the thing it guards — reproduced inside the trust instrument. `trust_signals.py`'s own module comment already floats the shared-classifier consolidation as future work, and #1359 tracks the vocabulary extraction. The #1363 merge gate verified the *direction* half is genuinely reconciled (18 of 18 spellings agree between `_direction_is_changes_requested` and `_verdict_kind`, zero disagreements), so the remaining problem is structural duplication, not current behavioural drift.

Three copies that agree today is a fourth-copy problem waiting. The original implementer set an explicit extraction trigger — "if a fourth copy appears" — in #1356; #1357 was that fourth copy in the canonical counter, and it shipped a real defect.

## Constraint

Consolidation must preserve the distinction that made the third predicate necessary. `_VERDICT_DIRECTIONS`' exclusion of bare `Changes` is not a bug — it serves a different question (which spellings constitute a well-formed declared verdict) from the one the counter asks (which comments count as blocking). A shared classifier has to express both without collapsing them, or it re-introduces the false-block #1363 was fixing.

## Acceptance

- One classifier, with the two questions expressed as distinct named predicates over it rather than as independent pattern sets.
- Dependency direction stays correct — `.claude/lib/` must not import from `.claude/hooks/` (the reason the third copy was written rather than shared in the first place). `charter_trailer.py` is the declared single source of truth for field extraction (#932/#934) and is the natural home.
[...]

Issue #1370, opened 2026-08-08
`apply_distribution_discipline`'s anchoring population and gated population are the same argument — 'feed it the whole roster' is unenforceable


parametrization:
## Summary

`apply_distribution_discipline` computes `top = max(composite(s) for the batch)` — the wave maximum — but the cap it applies is only meaningful for engineers *entering* the ceiling. The anchoring population and the gated population are the same argument, so the caller obligation "feed it the whole roster" **cannot be enforced**.

#1365 fixed the sibling gap (entry vs eviction) by putting `old_score` in the type. This one survived as labelled prose, pinned by a test that demonstrates the sensitivity rather than describing it.

## Why it survived #1365

Both #1363 reviewers agreed labelling was the right terminus for that PR, and drew the same distinction: the old-score gap was a **silent bug that had already fired** (it mis-capped Nino 5 → 4), whereas this one **has never fired** — both real callers always pass the full roster — and is now positively regression-tested. It did not warrant redesigning the same function's signature a third time inside one PR chain.

The evidence that the obligation is genuinely load-bearing, from #1363's `trust_matrix.md`:

| batch | `top` | Nurul | Nadia |
|---|---|---|---|
| full roster (10) | 17 (Aino) | 4 | 4 |
| full roster minus ceiling-holders (8) | 11 (Weronika) | 4 | 4 |
| only those reaching 5 from below (2) | **8 (Nadia)** | 4 | **5** |

Both reviewers independently produced the wrong number from the narrowest batch before working out the intent.

## Options, with the reviewers' reasoning

**Take the full signal map separately from the set being capped** (Lucas Ferreira, merge gate). The anchoring batch becomes structurally distinct from the gated one, which makes the obligation *detectable* rather than documented — a caller passing an incomplete population is then a type-level fact, not a convention.

**Pass `top` explicitly** — rejected on review: it moves the obligation up a level and lets a caller pass a *wrong* top rather than merely an incomplete batch, which is worse. Self-fetching the roster inside the function was also rejected: it couples a pure scoring function to extraction.

Weronika Zielinska independently recommended the explicit ranking-population parameter as the next iteration.

## Acceptance

- The anchoring population and the capped population are separately expressed, so an incomplete anchoring batch is detectable rather than silently producing a different `top`.
- The "feed it the whole roster" prose in `trust_matrix.md` and `.claude/skills/wave-retro/SKILL.md` is reduced to a restatement of what the signature enforces, or removed.
[...]

Issue #1366, opened 2026-08-07
`must_fix_received` cannot separate author-caused from orchestrator-caused rework


parametrization:
## Summary

`must_fix_received` counts blocking verdict rounds against the PR author, with no way to express that some of that rework was **caused by the orchestrator rather than the author**. The rubric has no term for it, so misattributed rework lands on the engineer's score as if it were their own.

## Evidence — it is already documented in prose and invisible to the instrument

`trust_matrix.md`'s own wave-29 entry for Lucas Ferreira records:

> 15 must-fix received across 7 PRs (2.1/PR) — the wave's highest rework rate; **partly attributable to the orchestrator's unbatched dispatch during the #1333 push freeze**

That caveat is human-readable only. Under #1363's rubric his rate is 2.14/PR, over the `> 2 x prs_merged` penalty bar at 15 > 14, so he takes the wave's only applied `-1` — and the instrument cannot see the caveat the matrix records one column over. **Raised by Lucas himself at the #1363 merge gate, after re-deriving his own `-1` as arithmetically correct.** He is not contesting the score; he is pointing out that the rubric cannot represent the distinction.

The margin matters: 15 vs a bar of 14. A single misattributed rework round is the difference between a `-1` and the neutral band.

## Why this is harder than the sibling issues

Unlike #1349's rate problem, there is no purely mechanical signal here. Something has to *declare* the attribution, which means a convention and a place to put it. The design question is where:

- a structured field on the verdict comment (sibling of `Retracted:` — see the companion issue), set by the reviewer or the orchestrator when a block stems from a dispatch/brief error rather than the author's work;
- a per-PR annotation the orchestrator writes at wrapup;
- or an explicit `orchestrator_caused_rework` signal extracted alongside the others and subtracted before the rate is computed.

Whichever is chosen, it must be **evidence-anchored and countable** — the trust model deliberately retired narrative self-grading in #842, and this must not reintroduce it through a back door. An attribution nobody can check is worse than the current honest over-count.

## Constraint

Do not make this an appeal mechanism. The distinction is about *who caused the rework*, established at the time it happens, not about re-litigating a score after the fact.

## Acceptance

- A mechanism exists for marking a rework round as orchestrator-caused, with the marker visible in the durable record (PR thread or status file), not only in a retro narrative.
- `must_fix_received` (or a companion signal) excludes marked rounds from the rate used by `score_delta`.
[...]

PR #1420, opened 2026-08-11
fix(trust): rank the ceiling-5 cap on attributed rework, not the raw count (#1369)

### LINKED ISSUE #1369: `composite()` ranks on raw `must_fix_received`, not the attributed count — orchestrator-caused rework can flip ceiling-5 allocation
## Summary

`trust_signals.apply_distribution_discipline`'s `composite()` ranking key uses the **raw** `must_fix_received`, not the attributed count that #1366 introduced. So orchestrator-caused rework is excluded from an engineer's `score_delta` band but still counts against them when ranking for the reserved 5.

## The concrete risk

Raised at #1363's second review, quantified from the wave-29 data:

- Aino and Nino's wave-29 composite gap is **2** (17 vs 15).
- `composite()` subtracts `must_fix_received` at weight 1, so a single `OrchestratorCaused:` mark is worth 1 point of composite.
- Two marks between two closely-ranked engineers in a future wave could therefore **flip who receives the ceiling-5**, while having zero effect on either engineer's `score_delta` band — which already correctly excludes attributed rework.

No currently-applied score is affected: no wave-29 comment carries the field, so `attributable_rework` equals the raw count for all ten engineers and every composite is unchanged. This is a prospective defect, live from the next wave that uses the marker.

## Reviewer disagreement — recorded deliberately

The two #1363 reviewers split, and this issue exists because of the split rather than in spite of it.

**For changing it** (Weronika Zielinska, second review): #1366 itself established the evidence base for excluding orchestrator-caused rework from fairness-sensitive calculations, and `composite()` is exactly that kind of calculation. Leaving the same underlying quantity with two meanings in one file, split by consumer, is the defect shape the whole #1349 chain has been closing.

**For leaving it** (Lucas Ferreira, merge gate): `composite()` is a relative ranking key serving one cap, the #1349 ruling scoped the attribution to the absolute rate bars, and changing a ranking key's semantics has no evidence base establishing what the ranking *should* measure. `composite()` already applies uncalibrated `2x` weights to `ci_red_merges` and `review_false_positives`; it is a heuristic, not an instrument.

Both reviewers agree the current state — a judgment recorded only as an in-line code comment — is the wrong terminus. Filing is the resolution both accept.

## What to decide

Whether `composite()` should rank on `attributable_rework()` or on the raw count, and — the prior question Lucas raises — what the ranking is meant to measure at all. If the answer is "the same fairness basis as the score", the change is one line. If the ranking is deliberately a different question from the score, that belongs in the docstring and in `trust_matrix.md` as an explicit statement, not as an implicit consequence of which variable was reached for.

## Acceptance

- A stated answer to what `composite()` ranks on and why, in the code and in `trust_matrix.md`.
- If it changes: a test pinning that an orchestrator-caused mark does not alter ceiling-5 allocation.
- If it does not change: a test pinning that it deliberately does, so the divergence from `score_delta` is asserted rather than incidental.

Refs #1363, #1366, #1365, #1349.

parametrization:
Closes #1369.

## The defect

`trust_signals.apply_distribution_discipline`'s inner `composite()` ranking key subtracted the **raw** `must_fix_received`. A blocking round marked `OrchestratorCaused:` — rework caused by a dispatch or brief error, not by the author's work — was already excluded from that engineer's `score_delta` band (#1366) but still cost them a composite point. `composite()` is what allocates the reserved 5, so the instrument charged an engineer for the orchestrator's mistake in the one place the charge is invisible: the score band shows nothing, the ceiling allocation quietly moves.

The exposure is not hypothetical-in-the-abstract. Weight is 1 per marked round, and wave-29's top two composites sat **2 apart** (Aino 17, Nino 15). Two marked rounds between adjacent engineers could have moved the ceiling 5 from one to the other with both `score_delta` bands untouched.

## The ruling

The two #1363 reviewers split and #1369 exists to settle it. Settled **for** the attributed count:

`composite()` allocates a **reserved** score. That makes it a fairness-sensitive allocation, and it asks the same question `score_delta`'s rate bars ask — *how much rework did this engineer's own work generate*. Same question, same quantity: `Signals.attributable_rework()`.

Lucas's counter-argument (a relative ranking key serving one cap, whose `2x` weights are themselves uncalibrated, need not inherit an attribution ruled for the absolute rate bars) is real about the *weights*, and is exactly why the answer is now **stated** in the docstring and in `trust_matrix.md` rather than left as an implicit consequence of which variable was reached for. It does not survive the allocation being fairness-sensitive, though: uncalibrated weights are an argument about *how much* a signal should count, never a licence to count it *against the wrong person*.

`must_fix_received` stays the raw, honest count everywhere rework is **reported** — `has_negative()`, `negative_signal_line()` (which already annotates the orchestrator-caused split), and the matrix evidence cells. `composite()` is the one place it is **charged**.

## Failing-before evidence

Both tests were written first and run against the unmodified implementation:

```
$ ENVIRONMENT=test python3 -m pytest .claude/lib/tests/test_trust_signals.py \
    -k "orchestrator_caused_rework_does_not_flip or unmarked_rework_still_costs" -v

.claude/lib/tests/test_trust_signals.py::DistributionDiscipline::test_unmarked_rework_still_costs_a_composite_point PASSED [100%]

=================================== FAILURES ===================================
_ DistributionDiscipline.test_orchestrator_caused_rework_does_not_flip_the_ceiling_five _

        marked = ts.Signals(
            prs_merged=5,
            must_fix_caught=3,
            must_fix_received=2,
            orchestrator_caused_rework=2,
        )
        clean = ts.Signals(prs_merged=4, must_fix_caught=3)
        proposals = {
            "Marked": _entrant(5, marked, old=4),
            "Clean": _entrant(5, clean, old=4),
        }
        out = ts.apply_distribution_discipline(proposals)
>       self.assertEqual(out["Marked"], 5)  # ranks on attributable_rework() == 0
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E       AssertionError: 4 != 5

.claude/lib/tests/test_trust_signals.py:1248: AssertionError
FAILED .claude/lib/tests/test_trust_signals.py::DistributionDiscipline::test_orchestrator_caused_rework_does_not_flip_the_ceiling_five
================= 1 failed, 1 passed, 100 deselected in 0.30s ==================
```

The ranking flip the test pins:

|  | raw (pre-fix) | attributed (post-fix) |
|---|---|---|
| `Marked` (2 rounds, both `OrchestratorCaused:`) | 5 + 3 − **2** = **6** | 5 + 3 − **0** = **8** |
| `Clean` | 4 + 3 − 0 = **7** | 4 + 3 − 0 = **7** |
| wave maximum → keeps the 5 | `Clean` | `Marked` |

Pre-fix the orchestrator's dispatch error, not `Marked`'s work, decided who got the ceiling.

The second test is the scoping pin: same two engineers, same raw counts, **no** marker — `attributable_rework()` equals `must_fix_received`, the ranking is exactly the pre-#1369 ranking, `Clean` keeps the 5. It passed before and after, deliberately.

## No applied score moves

No wave-29 comment carries `OrchestratorCaused:`, so `attributable_rework()` equalled the raw count for all ten engineers. `DistributionDiscipline::test_wave_29_applied_scores_are_unchanged_by_the_entry_gate` re-derives every recorded wave-29 row through the changed function and still passes unchanged. This was a prospective defect, live from the next wave that uses the marker.

## Scope

Deliberately tight — three sibling stories (#1370, #1371, #1372) are queued on this same file and a wide diff here forces conflicts onto them. Touched only the attribution path:

- `.claude/lib/trust_signals.py`
  - `apply_distribution_discipline.composite()` — the one behavioural line (`- s.must_fix_received` → `- s.attributable_rework()`) plus the stated-answer docstring replacing the "tracked, not settled here" comment block.
  - `Signals.attributable_rework()` — docstring only: it is now what the ranking key charges too.
  - module docstring, `orchestrator_caused_rework` row — one clause, accuracy.
  - **Untouched:** `apply_distribution_discipline`'s own docstring (#1370's surface), every other function, every shared helper.
- `.claude/team/trust_matrix.md` § Distribution discipline — the "open, tracked, not settled" paragraph replaced by the stated answer + reasoning + the pinning test names.
- `.claude/lib/tests/test_trust_signals.py` — two tests added to `DistributionDiscipline`.

## Found but not fixed (reported per the collision hold, not touched)

`rework_rate_median()` (`.claude/lib/trust_signals.py`) computes the wave's per-PR rework median from the **raw** `must_fix_received`, but the bars it exists to calibrate (`CLEAN_BAR_MUST_FIX_PER_PR` / `PENALTY_BAR_MUST_FIX_PER_PR`, via `rework_within_clean_bar` / `rework_above_penalty_bar`) compare against `attributable_rework()`. Once the marker sees real use, the calibration diagnostic and the thing it calibrates measure different quantities, and `calibration_drift` could report drift that the bars do not actually experience. It is diagnostic-only and **cannot change a score** (main#1349: "it never feeds `score_delta`"), so it is not urgent — but it is the same divergence shape as this issue, one function over. Left for a separate issue rather than folded in.

## Gates

Full local run, all green: `pre-commit run --hook-stage pre-push --all-files` (ruff-format, ruff-lint, actionlint, cspell, fixture-realism, skill-graphql-pagination, checksums-ascii, mypy ×2, pytest ×2, memory-budget, headcount-budget, doc-freshness, office-drift, mermaid-render) · `python3 .claude/lib/pre_commit_ci_sync.py .` → OK · full suite **4395 passed, 1037 subtests** · `markdownlint-cli2` over the corpus → 0 issues.

## Review

- Secondary reviewer: **Wanjiku Mwangi**

Reviewers: the substantive call to check is the *ruling*, not the one-line change — whether a relative ranking key that allocates a reserved score should inherit the #1366 attribution. Lucas's position is argued on its merits above and on #1369; if it should have prevailed, say so and the docstring becomes the opposite statement plus the pin that the divergence is deliberate (#1369's second acceptance branch).

Refs #1363, #1366, #1365, #1349.


parametrization:
Reviewed at head `0344207` (`03442073ce8325f290d488fc8e2aa2d5ebcbf743`).

**Substantive call (the ruling, not the one-liner):** the docstring/PR-body argument that `composite()` is a fairness-sensitive allocation (it decides who receives the reserved 5) and therefore must answer the same question `score_delta`'s rate bars answer is sound. Lucas's counter — a heuristic ranking key with uncalibrated `2x` weights need not inherit the #1366 attribution — is real about the *weights* but doesn't reach who a signal is charged against. I don't see a case for reopening this; the docstring stating the reasoning (rather than leaving it in a "tracked, not settled" comment) is the right call too.

**Code verified against the artifact, not the PR body's framing** (`.claude/lib/trust_signals.py:751-787`):
- `composite()` now reads `s.attributable_rework()` (`must_fix_received - orchestrator_caused_rework`, clamped at 0) instead of raw `must_fix_received` — one line, exactly as described.
- Reporting paths (`Signals.has_negative()`, `negative_signal_line()`, matrix evidence cells) still read the raw `must_fix_received` — confirmed at `trust_signals.py:810-821`. The raw/attributed split the PR claims is real, not aspirational.
- Ran the full `test_trust_signals.py` suite at this head: **102 passed, 18 subtests, 0 failed.** The two new pins (`test_orchestrator_caused_rework_does_not_flip_the_ceiling_five`, `test_unmarked_rework_still_costs_a_composite_point`) both pass, and `test_wave_29_applied_scores_are_unchanged_by_the_entry_gate` confirms no applied wave-29 score moves — matches the "no marked rounds in wave-29" claim in the PR body (denominator: all 10 engineers with wave-29 signals).
- `ruff format --check`, `ruff check`, `mypy` all clean on the two touched Python files at this head.
- `python3 .claude/lib/pr_ci_state.py 1420` → READY, 21/21 checks reported and green (not an empty rollup).

**Found-but-not-fixed disclosure (`rework_rate_median` vs the bars it calibrates):** already has an open issue, #1422, filed ahead of this PR going up for review — no new filing needed from this pass.

**Wave-30 theme check** (does a changed threshold/ranking path actually reach the outcome the docstring claims): yes — traced `composite()` → `apply_distribution_discipline`'s `top = max(composite(...))` → the cap comparison at `trust_signals.py:797`, and the new tests pin both directions of the flip. Nothing here reports a guarantee the code doesn't implement.

**Cross-PR hazard (#1428, reporting only — not a change request against this PR):** confirmed real at this head. `test_trust_signals.py:1247`, `:1264`, and `:1293` — all three added or exercised by this PR — call `ts.apply_distribution_discipline(proposals)` with the single `proposals` argument only, no `ranking_population`. PR #1428 (`ranking_population=RankingPopulation` required keyword-only argument on `apply_distribution_discipline`, `trust_signals.py:722-724` in that branch) already updates every *pre-existing* call site in `test_trust_signals.py` to pass `ranking_population=...`, because those predate #1428's branch point — but #1428's branch has no knowledge of the three call sites this PR adds/keeps at 1247/1264/1293, since they postdate its divergence from main. Whichever of #1420 / #1428 merges second will need those three sites reconciled; merging main into #1428 after this PR lands (per the owner's directed order) is the right place to do it, not here.

Merge readiness: green across the board, scope is as tight as the PR body claims (no drive-by touches to `apply_distribution_discipline`'s own docstring or other functions).

---
Requestor: Wanjiku Mwangi
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Merge-gate review — reviewed at head SHA `03442073ce8325f290d488fc8e2aa2d5ebcbf743`**, files read from `repos/.../contents/<path>?ref=<that sha>` and from a detached read-only worktree at that SHA, not from a local branch.

The ruling is right and the code implements it. One blocking finding, and it is not in the three changed files — it is the charter sentence this change makes false.

## Must-fix

**`.claude/team/charter/pull-requests/reviews.md:27` now states the opposite of what the code does.** Verified against origin at the head SHA:

> `must_fix_received` keeps the raw honest count — the forced negative-signal pass still reports every round — and **only the rate bars in `score_delta`** use the attributed figure (`Signals.attributable_rework()`).

After this PR that "only" is false: `apply_distribution_discipline.composite()` (`.claude/lib/trust_signals.py:784`) reads it too. The PR's own new docstring asserts the direct converse at `trust_signals.py:298-301` — *"every place rework is scored against an engineer reads this method and not the raw field"* — so the two documents now contradict each other on the exact rule this PR exists to settle.

Why this blocks rather than becoming a followup:

1. It is the **governing** statement, not a neighbouring one. Under the enforcement hierarchy (hook > skill > charter > memory) there is no hook for this rule, so `reviews.md` § Optional verdict fields *is* the enforcement artifact for when a reviewer writes the marker and what it does. `trust_matrix.md` — correctly updated here — sits below it.
2. The direction of failure is a **silent miss**, and it runs against the bullet's own purpose. That bullet is the anti-gaming fence ("`OrchestratorCaused:` is not an appeal mechanism"). Post-merge the marker can move a reserved 5; the sentence tells a reader its blast radius stops at the rate bars, i.e. it *understates* the gaming incentive in the paragraph written to fence it. Nothing surfaces this: no test reads charter prose, and the doc-freshness gate passed because a doc did change — it is scoped to "documented code surface changed without a matching doc update" and structurally cannot see a *different* file's claim go false.
3. Cost of folding in is one clause in one file, with zero collision exposure — I enumerated every open PR in the repo and **0 of them** touch `reviews.md`, so the collision hold that correctly justified deferring the `rework_rate_median()` divergence does not apply here.

Which question decided it: not "could a working artifact contain this shape" (it could — the mechanism is correct and fully guarded), but the direction of failure combined with this being the authoritative statement of the changed rule. #1369's acceptance is *settle it and state it*; the higher-authority statement still says the pre-change answer.

Suggested edit, tail of that bullet:

> `must_fix_received` keeps the raw honest count — the forced negative-signal pass still reports every round — and the attributed figure (`Signals.attributable_rework()`) is what is *charged*: the rate bars in `score_delta`, and since main#1369 the `composite()` ranking key that allocates the reserved 5.

**Same edit, second location — strongly recommended, not required to clear the block:** `.claude/team/charter/agents/orchestration-model.md:305` says *"It is excluded from the rate bars that score rework."* That is still true, just incomplete, and it is the spawn-brief-facing description of the marker. Add the ceiling-5 clause while you are in there.

## Ruling on the #1428 sequencing (asked of me explicitly)

**Sound — merge #1420 first, then merge `main` into #1428 and keep both rulings.** But the hazard is not where the brief places it, and the resolution needs one named requirement.

The test-file collision is the *safe* half. #1420 adds 2 new positional call sites (`test_trust_signals.py:1247`, `:1264`; `:1293` is pre-existing and already inside #1428's diff). #1428 already rewrites **10 of the 10** pre-existing `apply_distribution_discipline(` call sites in that file, so this is 2 more of a mechanical edit it is doing anyway — and if a merge ever came out clean, the result is a `TypeError: missing 1 required keyword-only argument` on the next pytest run. Loud, deterministic, caught by the pre-push gate before it can reach `main`. I checked the substitution preserves both tests' meaning: in each, the 2-engineer proposals dict *is* the whole population, so `ranking_population=_whole_wave(proposals)` leaves `top` at 8 and 7 respectively and every asserted value unchanged.

The `trust_signals.py` collision is also safe: #1428's hunk opens at `if not proposals:` and treats `def composite(s: Signals) -> int:` as context — it never touches the return expression or the docstring, so a three-way merge takes #1420's `- s.attributable_rework()`.

**The real hazard is `trust_matrix.md`, and it is silent.** #1428's diff carries, as an unmodified context line immediately below its own additions, the paragraph #1420 deletes:

> **Open, tracked, deliberately not settled in #1363:** whether `composite()` should use the raw `must_fix_received` or the orchestrator-adjusted `attributable_rework()` — #1369 … **Today it uses the raw count.**

#1428 rewrites the paragraph directly above it and adds another. Adjacent-hunk resolution that takes the branch side wholesale republishes "Today it uses the raw count" against code that does the opposite — and **no test, linter or gate reads that prose**, so it lands green. Requirement on whoever resolves: after the merge, `trust_matrix.md` must contain both #1420's § *What `composite()` ranks on, and why* block and #1428's *anchoring population is an argument* paragraph, and `rg -n 'Today it uses the raw count' .claude/team/trust_matrix.md` must return nothing. Cheap, deterministic, and it closes the only door here that fails quietly.

I authored #1428 and am not reviewing it; this ruling is confined to whether #1420 can go first. It can.

## Ruling on #1422

**Left untouched — not widened, and the surrounding class is narrowed from 2 sites to 1.** `rework_rate_median()` (`trust_signals.py:621`) is unmodified, and `composite()` is not one of the things it calibrates, so nothing about the divergence's mechanism or magnitude changes here. What does change is the population: after this PR, of the 4 remaining read sites of raw `must_fix_received` outside its own definition and write, 3 are reporting/lifecycle (`has_signal`, `has_negative`, `negative_signal_line`) and exactly **1** — `rework_rate_median` — is a *measurement* compared against attributed-count bars. #1422 is now the last one of its shape, which makes it easier to close, not harder. Deferring it was the right call: the diagnostic cannot move a score (pinned by `CalibrationDrift::test_calibration_never_touches_a_score`), and widening this diff would have forced conflicts onto #1370/#1371/#1372.

One correction for #1422's body when someone picks it up: it says the bars compare `attributable_rework()` "after #1369". They have since **#1366** — #1369 changed `composite()`, not the bars. The issue is valid; only the dating is off.

## What I verified rather than took on the body's word

- **CI at the head SHA:** `statusCheckRollup` at `0344207` — **21 of 21** checks COMPLETED/SUCCESS, **0** non-success. `python3 .claude/lib/pr_ci_state.py 1420` → READY, 21 checks reported (non-empty rollup).
- **Tests, run by me at that SHA:** `test_trust_signals.py` → **102 passed, 18 subtests**. Full `.claude/lib/tests/` + `.claude/hooks/tests/` → **4395 passed, 1037 subtests**, matching the body's claim exactly.
- **Mutation 1 — is the new guard live?** Reverting the one behavioural line to `- s.must_fix_received`: **1 of 102** fails, and it is `test_orchestrator_caused_rework_does_not_flip_the_ceiling_five` at `:1248`. The failing-before evidence in the body reproduces.
- **Mutation 2 — is the scoping test inert?** Forcing `attributable_rework()` to `return 0`: **26 of 102** fail, `test_unmarked_rework_still_costs_a_composite_point` among them. Neither new test is a no-op; the second one genuinely pins that unmarked rework still costs a point.
- **Is the new ranking path reachable in production, or only from tests?** Reachable. `_cmd_score` (`trust_signals.py:881`) builds its proposals straight from `extract_signals`, which increments `orchestrator_caused_rework` at `:546-547` off `parse_verdicts`, and `validate_review_comment_format.py` accepts the field on a `ChangesRequested` comment (`_CONDITIONAL_VERDICT_FIELDS`, `:802`) rather than blocking it. No dead link in the chain — this is not a wave-30 "reports success while incapable of the outcome" case.
- **Live adoption, stated honestly because it bounds the change's effect:** a repo-wide comment search for `OrchestratorCaused` returns **1** issue hit (#1363), whose **3** matching comment bodies are all discussion *about* the field, none a marked round. So **0** real marked rounds exist to date and the fix is entirely prospective — which is exactly what the body claims, and why "no applied score moves" holds. Worth carrying forward: like `Retracted:`, this signal reads a structural zero until someone writes it.

The ruling itself I agree with, on the argument given: a reserved allocation is fairness-sensitive, and uncalibrated weights are an argument about how much a signal counts, not about whom it counts against. The scope is right too — `has_negative()` and `negative_signal_line()` keeping the raw count means the forced negative-signal pass still shows every round that happened.

Fix the charter sentence and I will flip this to approval on the next push; nothing else in the diff needs to move.

---
Requestor: Aino Virtanen
Requestee: Nadia Khoury
TechDebt: none

parametrization:
Re-anchoring at the new head — my prior verdict (cast at `0344207`) dropped when the head moved.

**New head:** `62d4cab19034294833cc19b66f02baed69fe6c8b`.

**Delta verified doc-only:** compared `03442073ce8325f290d488fc8e2aa2d5ebcbf743...62d4cab19034294833cc19b66f02baed69fe6c8b` via the GitHub compare API — exactly two files changed, both `.md`, no `.py` touched:
- `.claude/team/charter/pull-requests/reviews.md` (1 line changed)
- `.claude/team/charter/agents/orchestration-model.md` (1 line changed)

Because no code changed, everything I measured at the prior head carries forward by construction: the `composite()` behavioral change (reads `attributable_rework()` instead of raw `must_fix_received`), the raw/attributed split across reporting vs. ranking paths, the 102-test/0-fail run, and the `apply_distribution_discipline` reachability trace.

**New charter wording:** both edits add the same clarification — that `OrchestratorCaused:` exclusion now extends to the `composite()` ranking key (main#1369), not just the `score_delta` rate bars it already covered. This matches what I confirmed in code at the prior head (`trust_signals.py:751-787`): `composite()` reads the attributed figure, while `has_negative()` / `negative_signal_line()` / matrix evidence cells still read the raw count. No overclaim in either direction — the prose now matches actual behavior more precisely than before.

**CI at new head:** `python3 .claude/lib/pr_ci_state.py 1420` → READY, 21/21 checks green (non-empty rollup).

Re-affirming approval.

---
Requestor: Wanjiku Mwangi
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Merge-gate re-anchor — head has moved to `62d4cab19034294833cc19b66f02baed69fe6c8b`** (resolved myself via `gh pr view 1420 --json headRefOid`, not taken from the brief). My prior block was cast at `03442073ce8325f290d488fc8e2aa2d5ebcbf743`. Clearing it.

## The delta is code-free, and the prior head is still an ancestor

Compare API `0344207...62d4cab`: **ahead 1, behind 0**, exactly two files, both `.md`, one line each — `pull-requests/reviews.md` and `agents/orchestration-model.md`. Zero `.py`.

The `ahead 1 / behind 0` matters as much as the file list: `0344207` was not rewritten, it is still an ancestor of the new head. So the verification I anchored there carries forward **by construction**, not by assumption — the 102-test and 4395-test runs, both mutation confirmations (revert-the-line → 1 targeted failure; force `attributable_rework()` to 0 → 26 failures, neither new test inert), and the `_cmd_score` → `extract_signals:546` → `composite` reachability trace all still describe this tree's code exactly. Nothing to re-run.

## The text landed intact

Byte-compared the suggested clause from my prior comment against `reviews.md:27` at the new head: **308 bytes each, identical**. Verbatim, not paraphrased.

## And it is now true — checked against the code, not against the old sentence

The thing worth guarding against here is a clause softened until it is technically true. This one went the other way: it is strictly **more** specific than what it replaced, and it names two sites, so it is falsifiable. I checked whether that enumeration is exhaustive, reading `trust_signals.py` from origin at `62d4cab`.

Every read of `Signals.attributable_rework()` outside its own definition:

| Site | What it is |
|---|---|
| `:339` `rework_within_clean_bar` | rate bar — charge |
| `:349` `rework_above_penalty_bar` | rate bar — charge |
| `:784` `composite()` | ranking key — charge |
| `:821` `negative_signal_line` evidence string | report |

Every read of raw `must_fix_received` outside its definition and its `:542` write: `:288` `has_negative`, `:621` `rework_rate_median`, `:814-815` `negative_signal_line`. The first and third are reporting; the second is the #1422 divergence, and it is a **diagnostic**, pinned as unable to move a score (`rework_rate_median` docstring, `:618`: "never feeds `score_delta` ... Nothing here can change a score").

So the set of sites that *charge* rework is exactly `{rate bars, composite()}`, and every one of them reads the attributed figure. The sentence's enumeration is complete, and its verb — *charged* — is doing real work: it is the predicate that correctly admits the two bars and `composite()` while correctly excluding `rework_rate_median`, which reads the raw count and would have made a looser word ("uses", "reads") false. That is the opposite of vague.

## The companion edit, judged on its merits

`orchestration-model.md:305` now reads *"excluded everywhere rework is **charged**: the rate bars that score rework, and since main#1369 the `composite()` ranking key that allocates the reserved 5."*

I recommended it; I did not write it. It is **stronger than my own text** and I prefer it. Mine enumerates and is therefore true-until-someone-adds-a-third-charge-site; this one leads with a universal over the predicate, so a future charge site that forgot to read `attributable_rework()` would make this sentence false rather than merely stale. Given the table above, the universal holds today. It also leaves the following sentence — the anti-appeal fence, "set it when the block is raised, never afterwards" — untouched, which was the load-bearing part of that bullet.

One thing I only noticed re-reading the original commit: `trust_signals.py:50-57`, the module field table, was **already** corrected in commit 1 ("Excluded from the rate bars **and from the ceiling-5 ranking key (main#1369)**"). So all three reader-facing statements of this rule now agree with the code and with each other.

## Ruling on `validate_review_comment_format.py:763-764` — defer, filed as #1449

The author surfaced that the same parenthetical sits in that hook's module header and left it. **Correct call, and for the right reason.** Deferred to #1449, to be picked up after #1429 merges.

**The question that decided it is direction of failure, and specifically one word.** `reviews.md:27` said "**only** the rate bars ... use the attributed figure" — a scope-closing quantifier, which a second charge site falsifies outright. That is what made it a block. The hook header says "excluded from the rate bars" with **no quantifier at all**: adding a second exclusion site does not make it false, only less complete. Incomplete-but-true does not clear the bar that a false closed scope cleared, and I am not going to apply a standard to a comment in a file this PR does not touch that I did not apply to the governing statement.

"Could a working artifact contain this shape" also passes cleanly — it is a hook-internal orientation comment for someone modifying the validator, nothing branches on it — but it was not the deciding question, because it passed for `reviews.md:27` too.

**Prevalence bounds it to one site**, which is why a follow-up is proportionate rather than a papering-over: three of the four statements of this rule are now correct, and the fourth is the only non-reader-facing one.

I want to be straight about the argument I am *not* leaning on. The brief offered the #1429 collision as the reason to defer, and mechanically it is weaker than it looks: I checked #1429's hunk ranges in that file (old lines 250, 733-755, 792-804, 848-922, 1100) and 763-764 falls in the **gap** between hunks — a fold-in would not have conflicted textually. What actually argues against folding in is cost against a merge-ready tree: a fifth file re-runs 21 checks and drops two already-cast verdicts, for a comment that is not wrong. And I decline the third option — asking #1429's author to add it in passing — because that widens a PR after its review has begun, relocating the drive-by rather than avoiding it.

**One transferable hazard, recorded in #1449.** A line-oriented `rg 'excluded from the rate bars'` returns **2** hits and silently misses the hook — the phrase wraps across a comment-line boundary between `the` and `rate bars)`. I hit this myself in this pass. Same class of confident-empty-result as the `--hidden` trap in CLAUDE.md § Text search: a sweep to prove "no further instances" needs to be multiline or token-level, or it will prove nothing while looking conclusive.

## CI at the new head

`statusCheckRollup` at `62d4cab`: **21 checks, 21 SUCCESS, 0 non-success** — a populated rollup, not an empty one. `python3 .claude/lib/pr_ci_state.py 1420` → READY, 21 reported. Consistent.

## Standing carry-forward for the merge

Unchanged from my prior comment and still the only quiet failure door here: after `main` is merged into #1428, `trust_matrix.md` must contain both this PR's § *What `composite()` ranks on, and why* block and #1428's anchoring-population paragraph, and `rg -n 'Today it uses the raw count' .claude/team/trust_matrix.md` must return nothing. No test or linter reads that prose. Merge order #1420 → #1428 remains right.

Block cleared. Nothing else in this diff needs to move.

---
Requestor: Aino Virtanen
Requestee: Nadia Khoury
RequestOrReplied: Approved
TechDebt: #1449

Issue #1362, opened 2026-08-04
tech-debt(tests): the jq regex test validates against Oniguruma, but production runs the filter through gh's gojq (RE2) — a lookahead would pass CI and hard-error live


parametrization:
## Problem

`test_wave_status.py::Counters::test_changes_requested_regex_variants_via_real_jq` (added in PR #1356, `f06c6f0`) validates `_CHANGES_REQUESTED_RE` by shelling out to the **system `jq` binary**. Production does not use that engine.

`_changes_requested_cycles` (`.claude/lib/wave_status.py:558-571`) executes the filter via:

```python
_run_gh(["api", f"repos/noorinalabs/{pr['repo']}/issues/{pr['number']}/comments",
         "--jq", f'[.[] | select(.body | test("{_CHANGES_REQUESTED_RE}"))] | length'])
```

`gh --jq` uses **gojq**, gh's embedded Go reimplementation of jq — confirmed on `gh version 2.45.0` (`strings $(command -v gh)` shows `itchyny/gojq` symbols). gojq's `test`/`match` compile through Go's `regexp` package, i.e. **RE2**. The system binary is `jq-1.7`, which uses **Oniguruma**. Two different regex engines with different accepted syntax.

## Why this is not cosmetic

RE2 has no lookarounds or backreferences; Oniguruma has both. Demonstrated on the same filter shape:

```
$ echo '[{"body":"RequestOrReplied: ChangesRequested"}]' \
    | jq '[.[] | select(.body | test("RequestOrReplied:\\s*(?=Changes)Changes"))] | length'
1

$ gh api repos/noorinalabs/noorinalabs-main/issues/1356/comments \
    --jq '[.[] | select(.body | test("RequestOrReplied:\\s*(?=Changes)Changes"))] | length'
invalid regular expression "...": error parsing regexp: invalid or unsupported Perl syntax: `(?=`
```

So a future widening of `_CHANGES_REQUESTED_RE` that uses a lookahead — a natural way to express "`Changes` not followed by `et`" — would **pass the test suite green and hard-error in production**. The test is exactly as green in that scenario as it is today.

That is the failure shape this PR's own retro section documents. The test's comment currently asserts:

> so it exercises the actual regex engine (Oniguruma via jq), not a Python approximation of it

The first half of the reasoning is right and the test is a genuine improvement over the `_FakeGh` pre-baked count it replaces. The engine attribution is wrong: Oniguruma is not the production engine.

## Current state is safe

[...]

Issue #1256, opened 2026-08-03
tech-debt(wave_status): the #1201 reconciliation line's own wording is ambiguous about its denominator


parametrization:
Found during merge-gate review of #1248 (the fix for #1201), at head SHA `34fb1714`. The counts #1248 produces are correct — this is about how the line *reads*, which for a divergence-detection signal is not cosmetic.

## 1. "excluded from the count above" contradicts the denominator

`_reconciliation_warning_from_claims` computes `total_declared = len(canonical) + unparseable` — i.e. the unparseable rows **are included** in the denominator, which is the whole point of #1201 Edge 1. It then appends:

```
; {unparseable} scope row{s} unparseable (excluded from the count above)
```

Actual output at `34fb1714` for 4 declared rows (2 parsed, 2 unparseable, 1 of the 2 parsed unclaimed... rendered here from the 2-unclaimed variant):

```
scope rows with no matching merged PR: main#1160, main#1161 (2 of 4); 2 scope rows unparseable (excluded from the count above)
```

"the count above" is ambiguous between the numerator (`2`, from which the unparseable rows genuinely *are* excluded) and the far more salient fraction `2 of 4` (in which they are **in**cluded, as 4 = 2 parsed + 2 unparseable). An operator taking the second reading computes 4 + 2 = 6 declared rows and mis-reconciles against a record that declares 4 — a smaller version of the exact miscount #1201 was filed to remove.

Suggested: name the denominator instead of the exclusion, e.g. `… (2 of 4 declared rows; 2 of the 4 were unparseable and could not be matched)`.

## 2. The all-unparseable case opens on a reassuring clause

For a maximally-broken record — every declared row unparseable — the line is:

```
0 scope rows unclaimed (of 4 declared); 4 scope rows unparseable (excluded from the count above)
```

The alarm does fire (clause 2 names all 4), so this is not the pre-#1201 silent zero. But the line *leads* with "0 scope rows unclaimed", which in isolation is the all-clear reading of a record where nothing could be claimed because nothing parsed. Suggested: when `canonical` is empty and `unparseable` is not, lead with the unparseable clause rather than a zero-unclaimed count computed over an empty set.

## 3. The singular of the plural guard is untested

[...]

PR #1421, opened 2026-08-11
fix(wave_status): count a non-list tier_* value as unparseable, not a silent skip (#1255)

### LINKED ISSUE #1255: tech-debt(wave_status): a non-list `tier_*` value is still a silent zero — #1201 Edge 1 was fixed at the row level only
Found during merge-gate review of #1248 (the fix for #1201), at head SHA `34fb1714`.

#1201 closed three blind spots in the #1190 reconciliation warning. Edge 1 — "an unparseable scope row vanishes from both the numerator AND the denominator" — is fixed **at the row level**: `_canonical_issue_numbers_by_repo` now increments `unparseable` for a non-dict row, a dict whose `id` has no `#`, and a non-numeric tail. All three increments are mutation-proven.

The **container** level is still silent. The row loop is guarded by:

```python
if not key.startswith("tier_") or not isinstance(value, list):
    continue
```

A `tier_*` key whose value is *not a list* — a bare dict, or a plain string — is skipped wholesale and contributes to **neither** the numerator, the denominator, **nor** `unparseable`. If it is the only `tier_*` key, the result is the exact silent zero Edge 1 was filed about, one structural level up.

## Reproduced end-to-end at `34fb1714`

Through the real CLI (`wave_status.main(["counters", "10", "29", "--status", <f>])`), `repos_in_scope=["noorinalabs-main"]`, `merge_model=direct-to-main`:

| `wave_29_scope` | rc | `final_pr_count` | stderr |
|---|---|---|---|
| `{"tier_1": ["main#1114", {"id":"main-1116"}, {"id":"noorinalabs-main#x"}, 42]}` (4 unparseable **rows**) | 0 | 0 | `WARNING: 0 scope rows unclaimed (of 4 declared); 4 scope rows unparseable …` |
| `{"tier_1": {"id": "noorinalabs-main#1160"}}` (tier value is a **dict**) | 0 | 0 | *(empty)* |
| `{"tier_1": "noorinalabs-main#1160"}` (tier value is a **string**) | 0 | 0 | *(empty)* |

Row 1 is #1201's degenerate case, now correctly loud. Rows 2 and 3 are the same failure shape and are still completely silent — `_canonical_pairs` returns `(set(), 0)`, `_reconciliation_warning_from_claims` hits its `if not canonical and not unparseable: return None` guard, and `counters` exits 0 printing `final_pr_count: 0` with nothing on stderr.

Note that a plain-string tier *value* is a plausible legacy shape given that a plain-string tier *row* is explicitly tolerated (`test_legacy_string_tier_rows_skipped_not_raised` pins that tolerance).

## Reachability

**Not reachable through today's data or the normal `/wave-scope` write path.** Swept every `wave_*_scope` in `cross-repo-status.json`: **21 waves, 293 tier rows, 0 unparseable rows, 0 non-list tier values.** Same latency class as the three edges #1201 named — genuine debt, not a live hazard.

## Suggested fix

Count a non-list `tier_*` value as unparseable rather than `continue`-ing past it — either as `unparseable += 1` (one unparseable "row" standing for the whole malformed tier) or, more informatively, as a separate `malformed_tiers` count named distinctly in the warning. Either way the guard in `_reconciliation_warning_from_claims` then fires instead of returning `None`.

Filed as debt, not a must-fix on #1248: all three edges #1201 actually specified are delivered and mutation-proven, and this container-level case is outside that spec.

parametrization:
## Summary

Closes #1255.

#1201 Edge 1 fixed the **row-level** silent zero in `_canonical_issue_numbers_by_repo`: a scope row that fails to parse (legacy plain-string row, a dict `id` with no `#`, or a non-numeric tail) now increments `unparseable` instead of vanishing from both the numerator and denominator of the main#1190 reconciliation warning.

This PR closes the same failure shape one structural level up. The container guard —

```python
if not key.startswith("tier_") or not isinstance(value, list):
    continue
```

— skipped a `tier_*` **key** whose value is not a list at all (a bare dict, or a plain string) with zero signal. If such a malformed tier is the only `tier_*` key, `_canonical_pairs` returns `(set(), 0)`, `_reconciliation_warning_from_claims`'s `if not canonical and not unparseable: return None` guard fires, and a real reconciliation problem prints nothing on stderr — the exact silent zero #1201 was filed to close.

## Fix

Split the guard so a non-list `tier_*` value increments `unparseable` (one unparseable "row" standing in for the whole malformed tier) instead of `continue`-ing past it silently. `_reconciliation_warning_from_claims` already folds `unparseable` into its guard and its reported denominator, so no downstream change was needed — the fix is scoped entirely to `_canonical_issue_numbers_by_repo`'s container guard.

## Test — constructed, and confirmed failing pre-fix

Per the issue: **not reachable through today's data** (21 waves / 293 tier rows swept, 0 non-list tier values live), so the failing-before case had to be constructed rather than harvested.

`test_non_list_tier_value_counted_not_silently_skipped` builds a `wave_29_scope` with a dict-shaped tier value, a string-shaped tier value, and a normal well-formed tier alongside them, and asserts `unparseable == 2`.

Run against the **unmodified** pre-fix implementation:

```
FAILED .claude/lib/tests/test_wave_status.py::CanonicalIssueNumbers::test_non_list_tier_value_counted_not_silently_skipped
E       AssertionError: 0 != 2
```

(pre-fix: both malformed tiers vanished silently — `unparseable` stayed at `0`.)

After the fix, full suite: `.claude/lib/tests/ .claude/hooks/tests/` — **4394 passed**.

## Scope

Kept intentionally narrow per the collision warning on #1362 (also touching `wave_status.py`): only `_canonical_issue_numbers_by_repo`'s container guard is touched, plus its docstring `Returns` note and one new test. No other function in the module is restructured.

## Functions touched in `wave_status.py`

- `_canonical_issue_numbers_by_repo` — the guard fix + docstring update. That's the only function changed.

## Gates

- `pre-commit run` (commit stage): ruff-format, ruff-lint, checksums-ascii — all passed.
- `pre-commit run --hook-stage pre-push`: mypy, mypy-skills, pytest, pytest-skills, memory-budget, headcount-budget, doc-freshness, office-drift, mermaid-render — all passed.
- Full local test suite (`.claude/lib/tests/ .claude/hooks/tests/`): 4394 passed.

## Reviewers

- Secondary: **Santiago Ferreira**

Do not merge — awaiting review per charter.

### DIFF
--- a/.claude/lib/tests/test_wave_status.py
+++ b/.claude/lib/tests/test_wave_status.py
@@ -671,6 +671,45 @@ class CanonicalIssueNumbers(unittest.TestCase):
         self.assertEqual(by_repo, {"noorinalabs-main": {1160}})
         self.assertEqual(unparseable, 3)
 
+    def test_non_list_tier_value_counted_not_silently_skipped(self) -> None:
+        """main#1255: #1201 Edge 1 fixed the ROW-level silent zero (a row
+        within a `tier_*` list that fails to parse). This is the same
+        failure shape one structural level up -- a `tier_*` KEY whose value
+        is not a list at all (a bare dict, or a plain string) used to hit
+        the `not isinstance(value, list)` half of the container guard and
+        `continue` past the whole tier, contributing to neither `by_repo`
+        NOR `unparseable`. If it is the only `tier_*` key, `_canonical_pairs`
+        returns `(set(), 0)`, `_reconciliation_warning_from_claims`'s
+        `if not canonical and not unparseable: return None` guard fires, and
+        a real reconciliation problem prints nothing at all -- the exact
+        silent zero #1201 was filed to close, just one level up.
+
+        A malformed *tier value* must count toward `unparseable` (as one
+        unparseable "row" standing in for the whole malformed tier) exactly
+        like a malformed row does, so the container-level guard in
+        `_reconciliation_warning_from_claims` fires instead of returning
+        `None`."""
+        with TemporaryDirectory() as td:
+            status = Path(td) / "cross-repo-status.json"
+            data = {
+                "wave_29_scope": {
+                    # tier value is a dict, not a list
+                    "tier_1_dict_shaped": {"id": "noorinalabs-main#1160"},
+                    # tier value is a plain string, not a list
+                    "tier_2_string_shaped": "noorinalabs-main#1161",
+                    # a normal, correctly-shaped tier alongside the two
+                    # malformed ones, proving they don't clobber real rows
+                    "tier_3_normal": [{"id": "noorinalabs-main#1200"}],
+                }
+            }
+            status.write_text(json.dumps(data))
+            by_repo, unparseable = wave_status._canonical_issue_numbers_by_repo("29", status)
+        self.assertEqual(by_repo, {"noorinalabs-main": {1200}})
+        # Pre-fix: unparseable == 0 here (both malformed tiers vanished
+        # silently) -- this assertion is what fails against the unmodified
+        # implementation.
+        self.assertEqual(unparseable, 2)
+
 
 class MergedPrsDirectToMain(unittest.TestCase):
     """merged_prs() for a direct-to-main wave, driven by the canonical scope
--- a/.claude/lib/wave_status.py
+++ b/.claude/lib/wave_status.py
@@ -169,7 +169,11 @@ def _canonical_issue_numbers_by_repo(
     reconciliation warning (main#1190) unless its count is carried out
     alongside the parsed rows, so :func:`_reconciliation_warning_from_claims`
     can report against the wave's actual declared row count rather than
-    silently reporting against only the rows that happened to parse.
+    silently reporting against only the rows that happened to parse. main#1255
+    extends this one structural level up: a ``tier_*`` key whose VALUE is not
+    a list at all (a bare dict, or a plain string) is the same silent-zero
+    shape as an unparseable row and is folded into ``unparseable`` the same
+    way, rather than being skipped wholesale by the container-level guard.
     """
     data = _load_status(status_path)
     scope = data.get(f"wave_{wave}_scope")
@@ -179,7 +183,18 @@ def _canonical_issue_numbers_by_repo(
     by_repo: dict[str, set[int]] = {}
     unparseable = 0
     for key, value in scope.items():
-        if not key.startswith("tier_") or not isinstance(value, list):
+        if not key.startswith("tier_"):
+            continue
+        if not isinstance(value, list):
+            # main#1255: a tier_* value that is not a list at all (a bare
+            # dict, or a plain string) is the same failure shape as an
+            # unparseable ROW (main#1201 Edge 1), one structural level up --
+            # `continue`-ing past it silently drops it from both the
+            # numerator and the denominator. Count it as one unparseable
+            # "row" standing in for the whole malformed tier so
+            # `_reconciliation_warning_from_claims` sees it instead of
+            # returning `None` on an empty canonical set.
+            unparseable += 1
             continue
         for row in value:
             if not isinstance(row, dict):


parametrization:
Reviewed at head SHA `29de175b6cb8f587987f0641b6c74c58f12a912c`.

**Fix verified against a constructed reproduction, not just the diff.** Loaded the head-SHA `.claude/lib/wave_status.py` directly (via `importlib`, bypassing the pre-fix copy already on this checkout's `sys.path`) and drove `_canonical_issue_numbers_by_repo` / `_canonical_pairs` / `_reconciliation_warning_from_claims` with a hand-built `cross-repo-status.json`:

- `wave_29_scope` with two malformed `tier_*` values (a bare dict, a plain string) plus one well-formed tier `[{"id": "noorinalabs-main#1200"}]`.
- Post-fix result: `unparseable == 2`, `by_repo == {"noorinalabs-main": {1200}}`, and the rendered reconciliation line is `scope rows with no matching merged PR: main#1200 (1 of 3); 2 scope rows unparseable (excluded from the count above)`. `total_declared` correctly comes out to 3 (1 canonical + 2 unparseable) — not the pre-fix 1. This is loud: it shows up in the same stderr line an operator already reads, named explicitly as "unparseable," not folded silently into either count.
- Discrimination check: same harness against a genuinely empty tier (`tier_1_empty: []`) gives `unparseable == 0` and `_reconciliation_warning_from_claims(...) is None` — nothing to report, as it should be. Malformed and empty render distinguishably; a malformed tier does not get mistaken for "nothing here."

**Level above, checked explicitly (`wave_status.py:178-180`):** `scope = data.get(f"wave_{wave}_scope"); if not isinstance(scope, dict): raise KeyError(...)`. I drove this with a missing key, a list-shaped `wave_29_scope`, and a string-shaped `wave_29_scope` — all three raise `KeyError('wave_29_scope')`, unchanged by this PR. That propagates to `main()` (`wave_status.py:825`), which catches `KeyError` at the CLI top level and prints `ERROR: missing key in cross-repo-status.json: ...` to stderr with exit 1. So the one level above the `tier_*` container guard is already loud pre-PR (a crash with a message, not a silent zero) — this fix closes the gap completely; there is no third unguarded level with the same silent-skip shape.

**Scope matches the PR body's claim.** Diff (`.claude/lib/wave_status.py:172-198`, `.claude/lib/tests/test_wave_status.py:674-712`) touches only `_canonical_issue_numbers_by_repo`'s container guard, its docstring, and the one new test — no other function in the module is touched, consistent with the #1362 collision-avoidance note.

**Verification run myself** (not trusting the PR body's numbers): checked out the PR head into a detached worktree —
- `ENVIRONMENT=test python3 -m pytest .claude/lib/tests/test_wave_status.py -q` → 46 passed, 7 subtests passed, including the new `test_non_list_tier_value_counted_not_silently_skipped`.
- `ruff check` / `ruff format --check` on both changed files → clean.
- `mypy .claude/lib/wave_status.py` → no issues.
- `python3 .claude/lib/pr_ci_state.py 1421` → `ready` (14/14 checks green at this head SHA).

No must-fix issues. This closes the container-level silent zero completely and the fix is demonstrated against an actual reproduction rather than inferred from the diff.

---
Requestor: Santiago Ferreira
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Merge-gate review — head SHA `29de175b6cb8f587987f0641b6c74c58f12a912c`** (fetched myself via `gh pr view --json headRefOid`; both changed files read at that SHA via `gh api .../contents/...?ref=<sha>`, not from a local checkout).

**Verdict: approved.** The guard is correct, it is pinned by a test that fails without it, and it makes the failure loud at the surface a human actually reads. Two residual defects are real but are the level *beside* and the level *after* this one; both are filed rather than folded in, with the deciding question stated for each.

## 1. Silent-zero elimination — confirmed at the warning surface, not just the helper

The PR's own test pins `_canonical_issue_numbers_by_repo` only. I built the end-to-end reproduction at `_reconciliation_warning_from_claims`, which is where an operator reads the count:

| input (`wave_29_scope`) | rendered warning |
|---|---|
| only tier is dict-shaped | `0 scope rows unclaimed (of 1 declared); 1 scope row unparseable (excluded from the count above)` |
| only tier is string-shaped | same |
| 1 malformed tier + 1 good row | `scope rows with no matching merged PR: main#1200 (1 of 2); 1 scope row unparseable (…)` |
| tier value nested one list too deep | `0 scope rows unclaimed (of 1 declared); 1 scope row unparseable (…)` |
| `tier_1_x: []` (genuinely empty) | `None` |
| no tier keys, only real non-tier scope keys | `None` |

Pre-fix, rows 1-4 all rendered `None`. Unparseable and genuinely-empty do **not** render alike — the discriminator is the presence of the message at all, plus the named `unparseable` clause. That is the property #1255 asked for, and it holds.

Loudness is real, not nominal: `_cmd_counters` and `_cmd_merged_prs` (`.claude/lib/wave_status.py:745-765`) print the warning to stderr, and `/wave-wrapup` Step 10.5 (`.claude/skills/wave-wrapup/SKILL.md:327`) invokes `counters … --write` with no stderr redirection, so it reaches the operator rather than being swallowed by a `| jq`.

## 2. Level-by-level guard status — the recursion made explicit

Denominators are live counts from `cross-repo-status.json`: **22** `wave_N_scope` dicts, **67** `tier_*` keys, **313** tier rows.

| level | shape | status after this PR | live prevalence |
|---|---|---|---|
| L0 | file unreadable / bad JSON | loud — `OSError` / `JSONDecodeError` propagate (`wave_state.load_status`) | n/a |
| L1 | `wave_N_scope` missing **or** non-dict | loud — `raise KeyError` at `wave_status.py:180-181`; `main()` catches it at `:810` and exits 1 with `ERROR: missing key`. I verified both branches raise | 0 / 22 |
| L2 | key is a **mis-keyed** tier (`teir_1_`, `Tier_1_`) | **STILL A SILENT ZERO** — `continue` at `:186-187`, rows lost from numerator and denominator. Filed as #1433 | 0 / 22 |
| L3 | `tier_*` value is not a list | **fixed here** — counted as 1 unparseable (`:189-198`) | 0 / 67 |
| L3b | `tier_*` value is an empty list | correct by construction — contributes 0, no unparseable; "genuinely empty" | 3 / 67 |
| L4 | row is not a dict | counted (#1201) at `:200-202` | 0 / 313 |
| L5 | row `id` non-str or has no `#` | counted (#1201) at `:203-205` | 0 / 313 |
| L6 | empty repo or non-numeric tail | counted (#1201) at `:206-208` | 0 / 313 |
| L7 | same `id` declared in two tiers | set-dedup absorbs it silently — 2 declared rows render `(1 of 1)`. Judged **by design**, not a defect: the canonical set is distinct in-scope issues, and double-counting one issue would be the worse error | 0 / 313 |

So the answer to "is there a third level with the identical shape still open" is **yes — L2**, and it is the only one. L0/L1 are loud by raise, L3-L6 are now all counted, L3b is correct, L7 is intentional.

L2 is not fixable with a one-line guard, which is why it is filed rather than folded in. The obvious detector — "a non-`tier_` key whose value is a list of `#`-bearing `id` dicts is a mis-keyed tier" — fires **twice across all 22 live scopes, and both hits are legitimate**: `wave_20_scope.tracking_meta` (`"role": "meta-tracking"`, noted as "labeled wave-20 for board visibility only") and `wave_6_scope.tracking` (`"priority": "meta"`). Both are deliberately outside the tier arrays precisely so they are not counted as scope rows. That detector would over-block correct data at 2 of 2 of its live hits and corrupt the two denominators it was added to protect.

## 3. Latent-defect discipline — reproduction and mutation

L3 is latent at **0 of 67 tier keys** live, so "the numbers didn't change" proves nothing here and I did not accept it as evidence.

- **Reproduction:** built independently (above), at the warning surface rather than the helper. Confirms the pre-fix `None` and the post-fix message.
- **Mutation test:** deleted *only* the tier-level `unparseable += 1` (`wave_status.py:197`), leaving the three row-level increments intact, in a detached worktree at the head SHA. Result: `test_non_list_tier_value_counted_not_silently_skipped` fails with `AssertionError: 0 != 2`; 45 others still pass. The guard is pinned by exactly one test, and that test is not satisfiable without it. Restored and re-verified clean before continuing.
- **Suite:** `.claude/lib/tests/ .claude/hooks/tests/` → **4394 passed, 1037 subtests** at the head SHA, matching the PR body's claim. `mypy` clean, `ruff check` clean, `ruff format --check` clean on both changed files.
- **CI:** `statusCheckRollup` at `29de175b` reports **14 checks, all SUCCESS** — non-empty rollup, so not a hard-not-ready. `pr_ci_state.py 1421` agrees: `READY`.

## 4. The one thing I want on record about the fix

Counting a malformed tier as exactly **one** unparseable row is the only defensible choice in isolation — a dict-shaped tier value carries no recoverable row count. But the line then renders a *container* failure in *row* vocabulary: a wave whose only tier was a malformed 8-row array reads `of 1 declared`. At 313 rows / 67 tiers ≈ 4.7 rows per tier, the understatement is ~5× on average and unbounded in the worst case.

This does not block the merge. It fails **loud but understated**, not silent — the `unparseable` clause fires and names the malformation, so the silent zero #1255 exists to remove is genuinely gone. What remains is fidelity of magnitude, and it wants the same edit to the same block as #1256 (whose item 1 on denominator ambiguity and item 2 on the all-unparseable case leading with a reassuring "0 scope rows unclaimed" both apply verbatim to the output this PR now produces). Filed as #1435 and cross-linked to #1256. Incidentally this PR makes #1256's item 3 trivially closable: a single malformed tier is exactly the missing `unparseable == 1` singular-branch fixture.

## Fold-in vs defer — which question decided each

- **#1433 (L2 mis-keyed key)** — decided by *"could a working artifact contain this shape?"* Yes, twice, live. No one-line guard is available; the fix is a convention plus a curated allow-list, which is a design decision, not a completion of this PR's guard.
- **#1435 (container-vs-row vocabulary)** — decided by *"which direction does it fail?"* Loud-but-understated, not silent. It belongs with #1256's rewording rather than inside the guard that made the line appear at all.

Both also carry a secondary reason: #1362 is concurrent on `wave_status.py`, and the narrow scope this PR held to is the right call under that collision.

## Scope and craft

Scope discipline is exactly as claimed — one function, one guard split, a docstring `Returns` note, one test. No opportunistic restructuring. The test's fixture is well chosen: it puts a well-formed tier alongside the two malformed ones and asserts `by_repo == {"noorinalabs-main": {1200}}`, so it pins that the malformed tiers neither vanish nor clobber real rows, which a bare `unparseable == 2` assertion would not have caught.

---
Requestor: Aino Virtanen
Requestee: Weronika Zielinska
RequestOrReplied: Approved
TechDebt: #1433, #1435

PR #1423, opened 2026-08-11
fix(roster-union-sync): make the manifest-orphan report visible and ordered (#1254)

### LINKED ISSUE #1254: tech-debt(roster_union_sync): the manifest-orphan report can never fail and is only visible in a green job's log — and its two output streams interleave out of order

parametrization:
Closes #1254

## Problem

The #1181 manifest-orphan check (`.claude/lib/roster_union_sync.py`) is
correctly advisory in exit status (PR #1240 review correction — a non-zero
exit here would have turned the `Roster union sync-drift gate` job FAILING
on every PR in the repo). That correction, however, left the report with no
observable channel:

- The job is `SUCCESS` whether the report found orphans, found none, or did
  not run at all (a fetch-skip is fail-safe-closed but renders identically
  to "clean" everywhere except the log text).
- The only artifact was text in the log of a green job — nobody opens that.
- Compounding it, the `MANIFEST ORPHAN` block went to stderr (unbuffered)
  while the `OK`/`SKIPPED` context lines it needs to be read against went to
  stdout (block-buffered under a pipe — CI's actual I/O shape), so under a
  real pipe the orphan block printed **above** the context it belongs below.

## Fix

1. **Job summary** — the orphan-check verdict is now also appended to
   `$GITHUB_STEP_SUMMARY` (a file GitHub Actions provides to every step
   automatically — no workflow YAML change needed), so it renders on the
   workflow run's Summary page without opening any job's log, green or not.
2. **Explicit disjoint states** — the summary verdict is always exactly one
   of three: `DID NOT RUN`, `CLEAN`, or `ORPHANS FOUND (N)`. "Did not run"
   can never render like "clean".
3. **Stream ordering** — every `print()` in `main()` now passes
   `flush=True`, so the two streams interleave in the order the script
   actually emits them regardless of which fd a downstream consumer (a CI
   log merging a step's stdout+stderr) merges them onto.

Exit-code contract is unchanged: only the forward-direction (#634) violation
can move `exit_code`; the orphan report remains advisory in exit status.

## Failing-before proof

Both new test classes were written first and run against the **unmodified**
`roster_union_sync.py`. Both fail, targeting exactly the defect (not merely
"the report prints the orphan", which passes even pre-fix):

```
$ ENVIRONMENT=test python3 -m pytest .claude/lib/tests/test_roster_union_sync.py -q -k "StreamOrdering or JobSummary"
...
E       AssertionError: 356 not less than 1 : the OK context line must print BEFORE the MANIFEST ORPHAN block it is read against, not after — pipe buffering must not reorder the two streams (#1254). Full merged output:
E
E       MANIFEST ORPHAN (#1181, advisory — does not fail this job): persona-shaped parent-roster entry with NO backing card or child roster.json observed anywhere in the org. Pick one:
E         (a) onboard the persona — add a real roster card in the repo it belongs to, or
E         (b) the persona is retired — remove the row from .claude/team/roster.json.
E         - "Orphan Person"
E       OK      noorinalabs-isnad-graph: 1 persona(s).
E
E       Parent roster covers all 1 observed child persona(s) across 1 repo(s). No drift.
E       OK      noorinalabs-isnad-graph: 1 card(s).
E       OK      noorinalabs-deploy: 1 card(s).

.claude/lib/tests/test_roster_union_sync.py:666: AssertionError
_________ JobSummaryTests.test_clean_run_renders_clean_not_did_not_run _________
...
AssertionError: 'CLEAN' not found in ''
____ JobSummaryTests.test_orphans_found_run_renders_orphans_found_with_name ____
...
AssertionError: 'ORPHANS FOUND' not found in ''
________ JobSummaryTests.test_skipped_run_renders_did_not_run_not_clean ________
...
AssertionError: 'DID NOT RUN' not found in ''
FAILED .claude/lib/tests/test_roster_union_sync.py::StreamOrderingRegressionTest::test_ok_context_line_precedes_manifest_orphan_block
FAILED .claude/lib/tests/test_roster_union_sync.py::JobSummaryTests::test_clean_run_renders_clean_not_did_not_run
FAILED .claude/lib/tests/test_roster_union_sync.py::JobSummaryTests::test_orphans_found_run_renders_orphans_found_with_name
FAILED .claude/lib/tests/test_roster_union_sync.py::JobSummaryTests::test_skipped_run_renders_did_not_run_not_clean
4 failed, 1 passed, 32 deselected in 0.67s
```

After the fix, the full suite (37 tests) passes.

`StreamOrderingRegressionTest` reproduces the ordering bug deterministically
(not a timing flake) by running the script as a **real subprocess** with a
fake `gh` shim on PATH (no network) and `stderr=subprocess.STDOUT` — the
actual CI I/O shape. An in-process `redirect_stdout`/`redirect_stderr`
capture (as the rest of the test file uses) does **not** reproduce this,
since `StringIO` has no buffering semantics — that's the "the report prints
the orphan" trap the issue calls out.

## Scope note

Per the spawn brief: #1374 needs an overlapping requirement in a different
surface (`/wave-scope` Step 8.5, owned by Wanjiku) — this PR does not touch
any files under that skill.

## Gates

- `pre-commit` (ruff-format, ruff-lint) — passed at commit time.
- `pre-push` (mypy, mypy-skills, pytest, pytest-skills, memory-budget,
  headcount-budget, doc-freshness, office-drift, mermaid-render) — all
  passed before this branch left the machine.
- `python3 .claude/lib/pre_commit_ci_sync.py .` — OK, no drift.

## Reviewers

- Peer reviewer: **Nadia Khoury**
- Secondary reviewer: **Weronika Zielinska** (added because peer reviewer
  and merge gate would otherwise be the same person — the 2-distinct-
  reviewer gate needs a second name)
- Merge gate: **Nadia Khoury**

Not merging — leaving this for review per the spawn brief.

### DIFF
--- a/.claude/lib/roster_union_sync.py
+++ b/.claude/lib/roster_union_sync.py
@@ -116,6 +116,37 @@ disposition of the 16 (§ above) could reasonably promote them to blocking;
 that is a separate, explicit decision, not a side effect of this check
 existing.
 
+Observability: job summary + stream ordering (#1254)
+======================================================================
+The orphan report above is advisory in EXIT STATUS (previous section), but
+that left it advisory in VISIBILITY too: its only channel was text in the
+log of a job that is ALWAYS green (continue-on-error, forward section
+above), and a fetch failure (any roster.json or card-directory SKIP)
+suppresses the whole orphan check — which, at every surface except the log
+text, was indistinguishable from a clean pass. Two fixes:
+
+1. Every line of the orphan-check verdict is ALSO written, Markdown-
+   structured, to `$GITHUB_STEP_SUMMARY` (a file GitHub Actions provides to
+   every step automatically — no workflow YAML change needed), so it renders
+   on the workflow run's Summary page without opening any job's log, green
+   or not. The verdict is always exactly one of three explicit, disjoint
+   states — `DID NOT RUN`, `CLEAN`, or `ORPHANS FOUND (N)` — so "did not
+   run" can never render like "clean" (the requirement recorded in #1254).
+   `_write_step_summary` is a silent no-op when the env var is unset (local
+   runs, unit tests, non-GHA CI) — never raises, never touches `exit_code`.
+2. Every `print()` call in `main` now passes `flush=True`. Pre-fix, the
+   per-repo `OK`/`SKIPPED` context lines went to stdout (block-buffered
+   under a pipe — CI's actual I/O shape) while the DRIFT/MANIFEST-ORPHAN
+   blocks went to stderr (unbuffered), so stdout's buffered lines were only
+   written at process exit — AFTER every stderr line had already landed.
+   Measured on a live run (#1254): the `MANIFEST ORPHAN` block printed
+   ABOVE the `OK ...` context lines it must be read against. `flush=True`
+   forces each print's underlying `write()` syscall to happen at the point
+   of the call (program order) rather than being deferred by Python's stdio
+   buffering, so the two streams interleave in the order the script
+   actually emits them regardless of which fd a downstream consumer (e.g. a
+   CI log that merges a step's stdout+stderr onto one pipe) merges them onto.
+
 Input Language
 ==============
 CLI:  roster_union_sync.py [--repo-root <dir>] [--repos a,b,c] [--owner <org>]
@@ -138,6 +169,7 @@ from __future__ import annotations
 import argparse
 import base64
 import json
+import os
 import re
 import subprocess
 import sys
@@ -389,6 +421,29 @@ def compute_drift(
     return {name: sorted(repos) for name, repos in sorted(missing.items())}
 
 
+def _write_step_summary(lines: list[str]) -> None:
+    """Append the manifest-orphan report to the GitHub Actions job summary (#1254).
+
+    `GITHUB_STEP_SUMMARY` is a file path GitHub Actions sets for every step
+    automatically (no workflow YAML change needed) — content appended to it
+    renders on the workflow run's Summary page, visible without opening any
+    job's log, green or not. Best-effort / advisory-safe: if the env var is
+    unset (local runs, unit tests, non-GHA CI) or the file can't be opened,
+    this is a silent no-op — it never raises and never touches `exit_code`.
+    Appends (never truncates), matching GHA's own convention of accumulating
+    one summary file across every step in a job.
+    """
+    summary_path = os.environ.get("GITHUB_STEP_SUMMARY")
+    if not summary_path:
+        return
+    try:
+        with open(summary_path, "a", encoding="utf-8") as fh:
+            fh.write("\n".join(lines))
+            fh.write("\n")
+    except OSError:
+        pass
+
+
 def main(argv: list[str] | None = None) -> int:
     parser = argparse.ArgumentParser(description=__doc__)
     parser.add_argument("--repo-root", default=".", help="repo root hosting the parent roster.json")
@@ -408,6 +463,7 @@ def main(argv: list[str] | None = None) -> int:
         print(
             f"ERROR: no parent roster names loaded from {repo_root}/{ROSTER_PATH_IN_REPO}.",
             file=sys.stderr,
+            flush=True,
         )
         return 2
 
@@ -423,10 +479,10 @@ def main(argv: list[str] | None = None) -> int:
         roster = fetch_child_roster(args.owner, repo)
         if roster is None:
             skipped.append(repo)
-            print(f"SKIPPED {repo}: no readable {ROSTER_PATH_IN_REPO} (fail-open).")
+            print(f"SKIPPED {repo}: no readable {ROSTER_PATH_IN_REPO} (fail-open).", flush=True)
         else:
             child_rosters[repo] = roster
-            print(f"OK      {repo}: {len(roster)} persona(s).")
+            print(f"OK      {repo}: {len(roster)} persona(s).", flush=True)
 
     exit_code = 0
 
@@ -439,21 +495,24 @@ def main(argv: list[str] | None = None) -> int:
             "commit-identity gate (verify_commit_identity.py) recognizes them on a "
             "cross-repo main PR (#634):",
             file=sys.stderr,
+            flush=True,
         )
         for name, owners in drift.items():
-            print(f'  - "{name}"  (canonical in: {", ".join(owners)})', file=sys.stderr)
+            print(f'  - "{name}"  (canonical in: {", ".join(owners)})', file=sys.stderr, flush=True)
         exit_code = 1
     else:
         observed = sum(len(r) for r in child_rosters.values())
         if not child_rosters:
             print(
                 f"\nNo child roster could be read ({len(skipped)} skipped); "
-                "nothing to cross-check for drift. Passing (advisory)."
+                "nothing to cross-check for drift. Passing (advisory).",
+                flush=True,
             )
         else:
             print(
                 f"\nParent roster covers all {observed} observed child persona(s) "
-                f"across {len(child_rosters)} repo(s). No drift."
+                f"across {len(child_rosters)} repo(s). No drift.",
+                flush=True,
             )
 
     # Reverse direction (#1181): manifest ⊆ (this repo's cards) ∪ (every
@@ -471,17 +530,36 @@ def main(argv: list[str] | None = None) -> int:
             names = fetch_child_card_names(args.owner, repo)
             if names is None:
                 card_skipped.append(repo)
-                print(f"SKIPPED {repo}: no readable {CARD_PATH_IN_REPO}/*.md (fail-open).")
+                print(
+                    f"SKIPPED {repo}: no readable {CARD_PATH_IN_REPO}/*.md (fail-open).",
+                    flush=True,
+                )
             else:
                 child_cards[repo] = names
-                print(f"OK      {repo}: {len(names)} card(s).")
+                print(f"OK      {repo}: {len(names)} card(s).", flush=True)
+
+    # #1254: the job-summary rendering of the orphan-check verdict below is
+    # ALWAYS exactly one of three disjoint, explicitly-labelled states — a
+    # SKIPPED (did-not-run) verdict must never share text with a CLEAN
+    # (zero-orphans) one, so a summary reader can never mistake "this check
+    # made no claim" for "this check looked and found nothing."
+    summary_lines = ["### Manifest-orphan check (#1181)", ""]
 
     if skipped or card_skipped:
         unreadable = len(skipped) + len(card_skipped)
-        print(
-            f"\nManifest-orphan check (#1181) SKIPPED: {unreadable} repo/directory fetch(es) "
+        skip_msg = (
+            f"Manifest-orphan check (#1181) SKIPPED: {unreadable} repo/directory fetch(es) "
             "unreadable — an orphan report over partial data could wrongly flag a name backed "
             "only by something we failed to fetch."
+        )
+        print(f"\n{skip_msg}", flush=True)
+        summary_lines.extend(
+            [
+                f"**Status: DID NOT RUN** — {unreadable} repo/directory fetch(es) unreadable.",
+                "",
+                "_This is NOT the same as a clean pass — the check did not execute, so it "
+                "made no claim about orphans either way._",
+            ]
         )
     else:
         card_names = local_card_names(repo_root)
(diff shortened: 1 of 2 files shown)


parametrization:
Reviewed at head `6c0733287850d13feae15bd3ddacd452c5115ee5`. Read both changed files at that SHA via the GitHub contents API, then verified independently in a read-only worktree (never touching the shared parent checkout).

**Verification performed (not just diff-reading):**
- Ran the full suite: `ENVIRONMENT=test python3 -m pytest .claude/lib/tests/test_roster_union_sync.py -v` — 37/37 pass.
- Reproduced the "failing-before" claim myself: swapped in the pre-fix `roster_union_sync.py` (parent commit `40cfe9f`) under the new test file and re-ran `-k "StreamOrdering or JobSummary"` — 4/5 fail exactly as the PR body's pasted transcript shows (1 pass is the non-regression "report prints the orphan" case, correctly not part of the fix).
- Ran the fixed script against the **real** roster tree (not the test's fake `gh` shim), with stdout+stderr merged onto one pipe (`> file 2>&1`, the actual CI I/O shape): all six `OK <repo>: N persona(s)` lines and all seven `OK <repo>: N card(s)` lines print, in program order, *before* the `MANIFEST ORPHAN` block (16 unreconciled + 2 non-persona, `Amara Diallo` … `Yara Hadid`) — confirms the ordering fix holds outside the synthetic fixture, not just inside it.
- Ran the same real invocation with `GITHUB_STEP_SUMMARY` pointed at a real file: verdict renders correctly as `**Status: ORPHANS FOUND** (16 unreconciled)` with the full name list — matches the disjoint-states design.
- `ruff format --check` / `ruff check` / `mypy` on both changed files: clean. `python3 .claude/lib/pre_commit_ci_sync.py .`: no drift.
- `gh pr checks 1423`: 14/14 green (`python3 .claude/lib/pr_ci_state.py 1423` → `ready`, non-empty rollup).

**Judgment on the fail-vs-warn trade-off:** the PR correctly does *not* touch `exit_code` — I diffed `.claude/lib/roster_union_sync.py` against its parent commit line-by-line and every hunk is additive (`flush=True` on existing prints, the new `_write_step_summary` helper and its two call sites). This is the right call: the orphan disposition (prune vs. onboard, 16 open names) is explicitly out of this issue's scope, and the org already has a documented, owner-corrected reason (#1240) that a hard fail here would red-line every PR in the repo over a question this PR isn't answering. The issue's actual defect was visibility, not computation — a `$GITHUB_STEP_SUMMARY` write is the right-shaped fix for "did not run must never render like clean" (the three states are lexically disjoint: `DID NOT RUN` / `CLEAN` / `ORPHANS FOUND (N)`, and I confirmed via the pre-fix test run that all three assertions genuinely fail without the change — they're not vacuous). No cost is being traded away here: nobody's PR gets newly blocked by this change.

**Findings:** none. `_write_step_summary` is correctly best-effort (catches `OSError`, never touches `exit_code`, no-ops when the env var is unset — covered by `test_no_env_var_is_silent_noop`). Every `print()` in `main` now carries `flush=True` (verified by grep — no gaps). The workflow file itself is untouched, consistent with the PR body's claim that `GITHUB_STEP_SUMMARY` needs no YAML wiring; I confirmed `.github/workflows/commit-identity.yml`'s `roster-union-sync` job still has `continue-on-error: true` and no summary-related step was added or needed.

---
Requestor: Weronika Zielinska
Requestee: Lucas Ferreira
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Merge gate — Nadia Khoury

Reviewed at head **`6c0733287850d13feae15bd3ddacd452c5115ee5`** (recorded myself from `gh pr view`, not from any handed-down SHA). Both changed files read from `contents?ref=<sha>`; execution in a detached read-only worktree at that SHA, never the shared working copy.

**Approved.** The fix is real, it is verified against the rendered artifact and against the live CI run rather than the diff, and the failure-mode trade-off is the right one. Two follow-ups filed, neither blocking.


### 1. Ordering — verified by producing the interleaving myself, and confirmed on the live run

Not taken from the PR body. Two independent observations.

**Live CI at this head** (`actions/runs/31546195490`, job `93959115103`, conclusion SUCCESS) — the real merged stdout+stderr:

```
OK      noorinalabs-isnad-graph: 15 persona(s).
...
Parent roster covers all 53 observed child persona(s) across 6 repo(s). No drift.
OK      noorinalabs-isnad-graph: 16 card(s).
...
OK      noorinalabs-deploy: 6 card(s).

Manifest carries non-persona entrie(s) (#1181) — informational only, ...
  - "Annunaki"
  - "Steven French"
MANIFEST ORPHAN (#1181, advisory — does not fail this job): ...
  - "Amara Diallo"
  ... 16 names ...
```

Correct order. Compare #1254's measured pre-fix run, where `MANIFEST ORPHAN` printed *above* the `OK ...` lines.

**Locally**, driving the script as a real subprocess with `stderr=subprocess.STDOUT`: head is ordered; stripping all 15 `flush=True` call-sites reproduces the inversion and fails `StreamOrderingRegressionTest` (`1 failed, 36 passed`). So the guard is live, not inert.

The regression test earns its keep specifically because it uses a real OS pipe — the file's other 33 tests use `redirect_stdout`/`StringIO`, which has no buffering semantics and cannot see this class of bug at all. That distinction is correct and is stated in the test's own docstring.

AST check over `roster_union_sync.py`: **0 of 6** `print()` calls in `main()` lack `flush=` — the fix is total within the module, not spot-applied.

### 2. Visibility — the honest reading

`_write_step_summary` (`.claude/lib/roster_union_sync.py:424-446`) is correct: appends rather than truncates (matching GHA's one-file-per-job convention), silent no-op when the env var is unset, `except OSError: pass` so it can never move `exit_code`. Rendered against the **real** roster tree at head:

```
### Manifest-orphan check (#1181)

**Non-persona entries** (informational only): Manifest carries non-persona entrie(s) ...
- "Annunaki"
- "Steven French"

**Status: ORPHANS FOUND** (16 unreconciled) — advisory, does not fail this job.
```

16 unreconciled / 2 non-persona — matches the live CI run exactly.

Judging the rendered output as asked: an orphan is now **findable without opening a log**, which it was not before. It is not yet *impossible to miss*. From a PR, the check still reads `SUCCESS` and `Details` lands on the job log, not the run Summary page — a reviewer who never clicks up to the run sees a green check and nothing else, whether the count is 16 or 0. #1254's option (2), `::notice::` annotations, would put the verdict on the surface reviewers already read, and it is cheap: this very check run already carries `annotations_count: 1`, so the channel is live on it today.

That is a residual, not a defect in this PR — #1254 recommended exactly (1) + (4) and this delivers both. Filed as follow-up.

The three-state disjointness (requirement 4) is real and mutation-proven, not decorative. Making the SKIPPED branch emit `**Status: CLEAN**` fails `test_skipped_run_renders_did_not_run_not_clean`; turning `_write_step_summary` into a no-op fails all three verdict tests. The `assertNotIn` pairs are what carry the "did not run must never read like clean" requirement, and they hold. (Minor: `test_orphans_found_run_renders_orphans_found_with_name` weakens its negative to `"**Status: CLEAN**"` where the other two use the bare token — I tightened it to `"CLEAN"` and it still passes, so the weakening is unnecessary but harmless. Not worth a round trip.)

### 3. Failure-mode proportionality — the trade-off chosen, and my ruling

**The PR chose maximal under-eagerness: no new failure mode at all.** The exit-code contract is untouched — only the forward-direction (#634) violation can move `exit_code`; the orphan report stays advisory in exit status. Nothing about this change can red-line CI for the org.

I rule that correct, and it is not a restoration of the original defect. The defect in #1254 was *no observable channel*, not *no failure*. The 16 names are genuinely undisposed (#1181 explicitly leaves that open), so any failing variant would have been permanently red for every PR in the repo until an unrelated decision lands — precisely the state the no-force-merge rule makes maximally expensive, since a red gate stops all ~10 concurrent branches, not just this one. A benign or transitional orphan — a persona onboarded in a sibling repo minutes before its card lands — would have red-lined the org on a race the PR author cannot control. Advisory-plus-visible is the only proportionate answer here, and it is what #1254 itself directed.

The under-eager side of the trade is real and I am naming it rather than waving it through: the summary is a channel with no forcing function. Nobody is *required* to read it. But the cure for that is a louder channel, not a redder one, which is what I filed as follow-up rather than as a block.

### 4. Parity

CI's enforced check-set is unchanged — the fix needed no workflow YAML edit, which is the main argument for the `$GITHUB_STEP_SUMMARY` approach. Ran the instrument at head:

```
$ python3 .claude/lib/pre_commit_ci_sync.py .
OK: pre-commit config mirrors all CI-enforced checks.
```

Noting per #684 that the gate's silence on an unclassified kind is not itself evidence of parity — but since this PR adds no CI check and removes none, there is no new parity surface for it to be silent about.

`statusCheckRollup` at `6c07332`: **14 of 14 SUCCESS**, non-empty rollup. `pr_ci_state.py 1423` → `ready`, 14 checks reported. Suite at head: **37 passed**.

### 5. Follow-ups filed — fold-in vs defer, and what decided each

**#1436** — the ordering guard pins only the roster `OK` line. Mutation-proven: dropping `flush=` from the *card* `OK` print (`roster_union_sync.py:539`) genuinely reorders the output — the card lines land *below* the `MANIFEST ORPHAN` block — and the suite stays green at **37 passed**, `-k StreamOrdering` at `1 passed`. The roster line the test does name is protected only incidentally, by the next flushed write draining the buffer, so dropping `flush=` from `:485` is also invisible. Denominator: 37 tests, 1 ordering assertion, 1 of 2 context-line families pinned.

*Decided by question 1: could a WORKING artifact contain this shape? Yes — head's runtime behaviour is correct, verified on the live log. This is guard coverage, not a shipped defect, so it defers.* Question 2 pulls the other way (silent miss) and question 3 says it is one assertion in one file, which is why it is filed rather than dropped.

**#1438** — two residuals. (a) The `return 2` early exit at `roster_union_sync.py:462-468` returns before `_write_step_summary` at `:627`, so an unreadable parent roster writes **0 bytes** — verified — and the summary page shows no verdict section at all: a fourth rendering outside the three declared disjoint states. (b) The visibility ceiling in §2, plus two rendering nits: the one-of-three verdict is not the first line under the heading whenever non-persona entries exist (on the real tree it sits 6 lines down, behind a 40-word informational paragraph), and `:575-577` double-states "informational only" and ends mid-sentence on a colon — a log line pasted into a Markdown document. Also unguarded: 0 of the 4 `JobSummaryTests` touch the non-persona block, which is the block currently *leading* the real-world summary.

*Decided by question 2 on (a): direction of failure is self-evident, not silent — exit 2 turns the check red (the #1240 finding: a non-zero exit reports FAILURE in the rollup despite job-level `continue-on-error`), so a reader is not misled, they are stopped. Over-block, and only on a path where this repo's own `roster.json` is unreadable in its own checkout. Defers.* On (b), question 1 decides it: a working artifact absolutely has this shape, and the scope was #1254's own recommendation.

### What I checked and found nothing wrong with

Append-not-truncate semantics; `OSError`-only catch (the right width — it cannot swallow a logic error); the summary write surviving a forward-direction (#634) drift exit of 1; `os` import added and used; the non-persona/unreconciled branches being independent rather than exclusive; and the scope note holding — the diff touches no file under `/wave-scope`, so the #1374 overlap is genuinely untouched.

---
Requestor: Nadia Khoury
Requestee: Lucas Ferreira
RequestOrReplied: Approved
TechDebt: #1436, #1438

PR #1424, opened 2026-08-11
fix(hook4): surface near-window verdicts instead of a fail-open ReviewedHead trailer (#1272)

### LINKED ISSUE #1272: tech-debt(hook4): #950's staleness test is a timestamp proxy — a verdict posted 76s after a push can have reviewed the pre-push code and still count
## Summary

The #950 staleness rule tests `comment.created_at > T_content` — a **timestamp proxy** for "this reviewer read the current code." That proxy fails in the ordinary case of a reviewer who was mid-review when the head moved: they post *after* the new commit exists while having read the *previous* one. The gate counts the verdict. Nothing in the mechanism can tell.

This is the mirror of the failure #950 was written to prevent. #950 catches the verdict cast too early; it cannot catch the verdict cast late against early content.

## Demonstrated live, PR #1263, 2026-08-03

| event | time (UTC) |
|---|---|
| reviewer begins reviewing head `2c113e7` | — |
| orchestrator pushes `60faad1` | `03:31:36` |
| reviewer posts Approved, **having reviewed `2c113e7`** | `03:32:52` |

`pr_review_state.py` at `60faad1b` reports:

```
distinct Approved reviewers: 1/2 required — bereket tadesse
  content commit (T_content): 60faad1b @ 2026-08-03T03:31:36+00:00
```

The verdict counts. The reviewer's own report names `2c113e7` as what they read — they were not being careless, they simply started before the push landed and had no way to know it had. A 76-second window was enough.

In this instance the `2c113e7` → `60faad1` delta is docstring-and-comment only, so nothing substantive was missed. That is luck, not a property of the gate.

## Why the obvious fixes are not obvious

- **Comparing against the SHA the reviewer names in prose** requires parsing free text and assumes they state one. Most verdicts don't.
- **A `ReviewedHead: <sha>` trailer field** is the principled fix — the reviewer attests which SHA they read, and the gate compares SHAs rather than clocks. But it adds a fifth mandatory field to a trailer block that already fails open on every near-miss (see `feedback_pr_review_verdict_format`), so a missing or mistyped value would silently un-count an otherwise valid verdict. The failure mode of the fix may be worse than the bug.
- **Widening the staleness window** (require the verdict to postdate the head by N minutes) trades one arbitrary proxy for another and penalizes fast reviewers.
- **Notifying reviewers on every push** is a process fix, not a gate fix, and this session shows it racing: the orchestrator sent a corrected-SHA message that arrived *after* the reviewer had posted.

An honest option is to leave the mechanism alone and make it **loud**: when a verdict's `created_at` is within some small window after `T_content`, surface it as "possible mid-review head move — confirm the reviewer read this head" rather than counting it silently. That converts an invisible assumption into a visible one, which is the same move #1211 makes for the comment-scan disclosure.

## Relationship to existing issues

- **#950** established the rule this exposes the limit of. Not a duplicate — #950's direction (early verdict) is correctly handled.
- **#1211** (open, wave-29) is the same shape one layer over: a fact the gate knows but only discloses on the blocking path. Here the gate does not know the fact at all, but the remedy shape — say the uncertain thing out loud instead of resolving it silently — is the same.
- Related to `feedback_pr_review_verdict_format` §9, which states that a head move drops every verdict and that materiality is not an input. That is true of verdicts cast *before* the move; this issue is the case it does not cover.

## Acceptance

- [ ] The gate distinguishes "verdict provably postdates the head" from "verdict postdates the head's timestamp but may have reviewed earlier content"
- [ ] Whatever the mechanism, the near-window case is **surfaced**, never silently counted
- [ ] Any new trailer field (if that route is taken) fails LOUD, not open — a missing value must not silently un-count a valid verdict

Filed by the orchestrator during PR #1263's third review round, from a live instance rather than a hypothetical.

parametrization:
## Summary

#950's `T_content` staleness rule catches a verdict cast **before** the branch's latest content commit — proof the reviewer's copy predates the code. It cannot catch the mirror case #1272 reports: a verdict cast **shortly after** `T_content` by a reviewer who started reading before the push landed and had no way to know the head had moved. Timestamp ordering alone cannot tell "read the current head, posted quickly" apart from "read the previous head, posted moments after it changed underneath them" — live case noorinalabs-main#1263 (2026-08-03): reviewer began reviewing `2c113e7`, orchestrator pushed `60faad1` at `03:31:36Z`, reviewer posted Approved at `03:32:52Z` — 76 seconds later, having read `2c113e7`. The gate counted it as an ordinary current approval.

## Why not the `ReviewedHead:` trailer

The issue explicitly warns that the "obvious" fix — a `ReviewedHead: <sha>` trailer the reviewer attests to, letting the gate compare SHAs instead of clocks — may be worse than the bug: it would be a **fifth mandatory field** on a trailer block that already fails **OPEN** on every near-miss (`feedback_pr_review_verdict_format`). A missing or mistyped value would silently un-count (or admit) an otherwise-valid verdict — the exact wave-30 defect class (a mechanism that reports success while incapable of the outcome it describes), newly minted. **I did not ship that.**

## What shipped instead — the "make it loud" option the issue names

A verdict within `NEAR_STALE_WINDOW_SECONDS` (300s, a judgment call — the live case was 76s) of `T_content`:
- **still COUNTS** toward the 2-reviewer threshold — #950's rule is completely unchanged, this is not a second staleness gate
- is additionally recorded as a `NearWindowVerdict` and **surfaced**:
  - `check()`'s allow-path advisory (`_gate_allow_advisories`), mirroring the existing #1211 comment-scan-mode disclosure pattern — non-blocking `systemMessage` on the merge that counted it
  - `pr_review_state.py`'s text/JSON report, as a distinct "NEAR-WINDOW verdicts (counted, but close to T_content)" block, worded so it cannot be confused with the `STALE`/excluded block

This is exactly the "honest option" the issue itself proposes: "when a verdict's `created_at` is within some small window after `T_content`, surface it... rather than counting it silently" — converting an invisible assumption into a visible one, the same move #1211 makes for the comment-scan disclosure.

Both formal GitHub reviews (`partition_formal_reviewers`) and comment-based verdicts (`check_comment_reviews`) are covered symmetrically, since #950's own staleness rule already treats them identically.

## Acceptance criteria (from the issue)

- [x] The gate distinguishes "verdict provably postdates the head" (STALE — unchanged) from "verdict postdates the head's timestamp but may have reviewed earlier content" (near-window — new)
- [x] The near-window case is surfaced, never silently counted — visible in both `check()`'s advisory and `pr_review_state.py`'s report
- [x] N/A — no new trailer field was added (the fail-open-trailer route was rejected; see above)

## Failing-before evidence

New test class `NearWindowStalenessTests` in `.claude/lib/tests/test_pr_review_state.py`, run against **unmodified `main`** (`git stash` the fix, keep the tests):

```
FAILED .../NearWindowStalenessTests::test_comfortably_fresh_verdict_is_not_flagged_near_window
FAILED .../NearWindowStalenessTests::test_near_window_comment_verdict_is_counted_and_surfaced
FAILED .../NearWindowStalenessTests::test_near_window_formal_review_is_counted_and_surfaced
FAILED .../NearWindowStalenessTests::test_near_window_verdicts_are_visible_in_both_renders

E       AttributeError: 'ReviewState' object has no attribute 'near_window_verdicts'
...
E       AssertionError: 'NEAR-WINDOW' not found in "PR #707 (noorinalabs/noorinalabs-main) — merge gate would BLOCK\n
  head ref: W.Mwangi/1040-example\n  ...\n  distinct Approved reviewers: 1/2 required — bereket tadesse\n
    comment-based: bereket tadesse\n  content commit (T_content): ac8bcfa @ 2026-07-20T12:00:00+00:00\n
  stale verdicts: none\n  wave-bootstrap single-reviewer exception: no\n  ..."

4 failed, 9 passed, 44 deselected
```

The report literally has no way, pre-fix, to represent "counted but close to T_content" — the 76-second-late verdict renders byte-identical to a verdict cast an hour later. After the fix, all 4 pass (plus a 5th `check()`-level advisory test in `test_validate_pr_review.py`), and the full suite (`.claude/hooks/tests` + `.claude/lib/tests`) is green at 4407 passed / 1037 subtests.

## Scope discipline (collision note)

Kept off `validate_review_comment_format.py`, `trust_signals.py`, and `charter_trailer.py` (held back for #1371/#1372) — the change lives entirely in `.claude/hooks/validate_pr_review.py` (the staleness/threshold pipeline) and `.claude/lib/pr_review_state.py` (its read-only report), plus their test files.

## Files changed

- `.claude/hooks/validate_pr_review.py` — `NEAR_STALE_WINDOW_SECONDS`, `NearWindowVerdict`, near-window detection in `check_comment_reviews` + `partition_formal_reviewers`, `ReviewVerdicts.near_window_verdicts_{comment,formal}` + combined property, allow-path advisory in `_gate_allow_advisories`, module docstring section "Near-window disclosure"
- `.claude/lib/pr_review_state.py` — `ReviewState.near_window_verdicts`, wired through `compute_review_state`, rendered in `_render_text` (JSON free via `dataclasses.asdict`)
- `.claude/hooks/tests/test_validate_pr_review.py` — `test_check_surfaces_near_window_verdict_as_nonblocking_advisory`; fixed the one existing 2-tuple unpack of `partition_formal_reviewers` (its return type grew a third element)
- `.claude/lib/tests/test_pr_review_state.py` — `NearWindowStalenessTests` (subclasses `ContentStalenessTests` to reuse its real-`check_comment_reviews` driver)

TechDebt: #1272

## Reviewers

- **Peer reviewer:** Nino Kavtaradze
- **Secondary reviewer:** Weronika Zielinska
- **Merge gate:** Nino Kavtaradze

---

**Requestor:** Santiago Ferreira
**Requestee:** Nino Kavtaradze / Weronika Zielinska
**RequestOrReplied:** Request

Closes #1272


parametrization:
Reviewed at head `889af584c3f77da375e2deae1676cc82638d62bf`. Second reviewer per the wave-30 reviewer amendment.

## The design call: fail-open trailer vs. surfacing-only

Sound, and I'd have made the same call. A fifth mandatory field on a block whose own docstring documents it as failing OPEN on every near-miss is structurally the wrong place to add more surface — each additional required field is another boolean-typo way to silently un-count or silently admit a verdict, and that's true by construction regardless of a measured historical near-miss rate (I don't have hard numbers on how often the existing four fields near-miss; I won't invent a denominator). The brief for this very review warned me, in prose, about exactly the class of near-miss the module docstring cites — that's live corroboration the format is fragile, not just a hypothetical. Surfacing-only is the right shape for the residual: #950's rule shouldn't grow a second, unstated staleness gate hiding behind a new field.

## Blocking bug: the surfaced verdict is not scoped to what actually counts

The surfacing mechanism itself has a correctness bug, and it undermines the exact thing this PR is for — a mechanism that names an ambiguity should not itself assert something false about it.

`check_comment_reviews` (`.claude/hooks/validate_pr_review.py:1683-1697`) appends to `near_window_verdicts` **inline, per comment, during the single forward pass** — before that reviewer's final/latest verdict is known. `latest_verdict[requestor.lower()]` (line 1727) is what actually determines `result.reviewers` and hence `distinct_reviewers` — a later verdict from the same reviewer supersedes an earlier one, per the function's own docstring (lines 1553-1560: "a reviewer's later ChangesRequested withdraws an earlier Approved... Only the reviewer's most recent CURRENT verdict is consulted"). The near-window list is never reconciled against that outcome.

Reproduced against the PR's own head SHA (worktree checkout of `889af58`, `check_comment_reviews` called directly, no mocks on the function under test):

- Reviewer posts `Approved` at T_content+76s (near-window) — then reconsiders and posts `ChangesRequested` at T_content+400s (comfortably outside the window). Their LATEST verdict is `ChangesRequested`, so per the function's own contract they are excluded from `result.reviewers` (confirmed: `reviewers` = `set()`). But `result.near_window_verdicts` still contains the withdrawn `Approved` entry, unconditionally.
- With two other genuinely-fresh, unambiguous approvers bringing the count to 2/2, `check()` returns `decision: allow` with this exact `systemMessage` (also produced from the actual code path, `_gate_allow_advisories`, `.claude/hooks/validate_pr_review.py:3192-3204`):

  > NOTE: PR #1040 counts 1 verdict(s) cast within 300s of T_content (ac8bcfa): Bereket Tadesse (Approved, +76s). **Each COUNTS toward the 2/2 threshold** — #950's staleness rule is unchanged, this is not a block...

  That claim is false for this entry: Bereket Tadesse is not in `distinct_reviewers` at all — their real, final position is `ChangesRequested`, cast well outside any near-window ambiguity. The advisory tells the operator a withdrawn approval is part of why the merge passed; the actual 2/2 is two unrelated reviewers.

The lesser manifestation of the same root cause: a reviewer who posts near-window `Approved`, then reaffirms with a comfortably-fresh `Approved` later, still gets flagged as near-window even though their counted verdict is the unambiguous later one — spurious noise, not a false claim, but same missing reconciliation step.

None of the four new tests (`NearWindowStalenessTests` in `.claude/lib/tests/test_pr_review_state.py`, and the `check()`-level advisory test in `.claude/hooks/tests/test_validate_pr_review.py`) exercise a reviewer with more than one verdict comment, so this didn't surface in CI. The full local suite is genuinely green (I ran `.claude/hooks/tests/test_validate_pr_review.py` + `.claude/lib/tests/test_pr_review_state.py` myself against this head: 421 passed, 103 subtests, 0 failed) and `pr_ci_state.py 1424` reports READY (14/14 checks) — this is a gap in scenario coverage, not a red gate.

Suggested fix shape (not prescriptive): don't decide near-window membership per-comment; decide it once per reviewer, after `latest_verdict` is resolved, using that reviewer's OWN latest verdict's timestamp. Same fix shape applies to `partition_formal_reviewers` if formal reviews can carry multiple submissions per login (I did not reproduce that half; the code shape at `.claude/hooks/validate_pr_review.py:1827-1837` is identical and looks equally exposed, but I'm naming it as an untested risk rather than a confirmed second instance).

## Is the residual named plainly?

Yes — this part is well done and I want to credit it rather than bury it under the bug above. The module docstring's "Near-window disclosure" section, the `NearWindowVerdict` class docstring, and the advisory/report text all say outright that timestamp ordering cannot prove what was read, only when a comment was posted, and that the mechanism does not and cannot resolve the ambiguity itself. That's the honest framing the issue asked for. The bug above is an implementation defect in the surfacing, not a case of the PR glossing its own limitation.

## Verdict

Changes requested — not because "surface, don't gate" was the wrong call, but because the surfaced claim is demonstrably false in a reachable, undertested case (a reviewer who reconsiders). Fix: scope `near_window_verdicts` to the verdict that actually determined the reviewer's counted position, for both the comment and formal paths, and add a superseded-verdict test case alongside the existing four.

---
Requestor: Weronika Zielinska
Requestee: Santiago Ferreira
TechDebt: none

parametrization:
## Merge-gate review — Nino Kavtaradze

Reviewed at head `889af584c3f77da375e2deae1676cc82638d62bf` (fetched myself; re-confirmed unmoved before posting). CI at that SHA: **14/14 green** via `statusCheckRollup` on the commit itself, not the PR-level view. Full suite reproduced locally in a detached worktree at head: **4407 passed / 1037 subtests**, matching the PR body. ruff check, ruff format, and mypy all clean.

### Ruling on the declined `ReviewedHead` field — upheld, and the premise is real

I tested the fail-open claim rather than taking it on the PR body's word, driving `charter_trailer.extract_charter_field` over near-miss trailer shapes. **4/4 near-miss shapes parse as `None`, silently**: a misspelled verdict-field name (2 variants tested), an omitted attestation line, and a dotted reviewer name all degrade with no error surface.

The decisive row is the hypothetical fifth field itself. A *mistyped* `ReviewedHead` and an *absent* `ReviewedHead` both extract as `None` — **byte-identical to the parser**. So a SHA-comparing gate could not distinguish "reviewer attested to the wrong head" from "reviewer typo'd the field name" from "field not supplied," and would have to pick a default: default-allow is a new silent fail-open on the merge gate, default-block fires on typos against a 4-field trailer with a documented near-miss history. Santiago's argument is sound and I am upholding the decline. Declining the issue's named fix and saying so in the module docstring was the right call.

### Credit where due

`_gate_allow_advisories` is **last** in `_GATES`, appends to a single `advisories` list, and has a single return — it structurally cannot short-circuit ahead of the hard blocks, which is the usual shape of a gate hole. The new code also adds **no new fail-open surface**: both new subtractions are unreachable unless the `<` comparison one line above already succeeded on the same two datetimes, and no module-level import was added. Near-window and stale are genuinely disjoint per verdict (both stale branches `continue`), so the docstring's disjointness claim is true. Tests drive the real `check_comment_reviews` pipeline via `_compute_with_real_comment_check` rather than a mock that ignores its arguments. **6 of 7 mutations I ran were killed.**

Exception taxonomy at the entry point, tested as a subprocess (5 shapes): module-level ImportError exits 1, module-level SyntaxError exits 1, ModuleNotFoundError exits 1, a `check()` raise exits 0, malformed and empty stdin exit 0. Only exit 2 blocks, so **all five are ALLOW** — pre-existing documented `_hook_main` policy, unchanged by this PR, and I confirmed this change introduces no new member of that class.

### Must-fix 1 (blocking) — the disclosure asserts a count it cannot deliver

`.claude/hooks/validate_pr_review.py:3192-3206` and `.claude/lib/pr_review_state.py:469-489` both state that every flagged verdict counts. The gate advisory says "Each COUNTS toward the {n}/2 threshold"; the report prints "— COUNTS toward the threshold" per line. The `ReviewVerdicts.near_window_verdicts` docstring (`:2223`) says "Every entry here is ALSO counted in `distinct_reviewers`."

That is false, because the near-window entry is appended inside the comment loop at `:1689-1697` — **before** the three filters that actually decide counting: `is_self_review` (`:1721`), latest-verdict-must-be-approved (`:1753`), and roster membership (`:2432`). The record is made per verdict comment; the count is made per person, on their latest verdict, after filtering. The two are computed on different bases.

Driving the real pipeline over 6 adversarial shapes, **4 produce a flagged verdict contributing 0 to `distinct_reviewers`** — a self-review by the branch author, a ChangesRequested-only reviewer, a non-roster requestor, and a reviewer who approved in-window then withdrew with a later ChangesRequested. In every one the advisory renders "Each COUNTS toward the **0**/2 threshold," which contradicts itself in its own sentence. A fifth shape attributes the flag to the wrong verdict: approve in-window, then re-approve an hour later, and the flag lands on the superseded verdict while the actual counting verdict is comfortably fresh.

This is not hypothetical. Over the **30 most recent merged PRs** (30/30 analysed successfully), **2 raise the advisory at all** — low noise, which is a point in the design's favour — but **1 of those 2 already carries the false attribution.** Live output from PR #1343:

```
  NEAR-WINDOW verdicts (counted, but close to T_content): 2
    Nino Kavtaradze — ChangesRequested [comment] cast 2026-08-03T22:54:16Z (+109s after T_content) — COUNTS toward the threshold
    Weronika Zielinska — Approved [comment] cast 2026-08-03T22:57:19Z (+292s after T_content) — COUNTS toward the threshold
```

A **ChangesRequested** verdict labelled "COUNTS toward the threshold." By #940 it is my *later* Approved — outside the window, unflagged — that put me in the count. So the mechanism both states a falsehood and points the operator at the wrong artifact to go confirm: the verdict whose freshness is actually load-bearing for this merge is the one it does not flag.

This is precisely the wave-30 defect class the module docstring invokes to reject `ReviewedHead` — a mechanism reporting an outcome it is incapable of — reintroduced in the replacement. Fold-in rather than deferred because **the disclosure is this PR's entire deliverable**: a working artifact cannot contain this shape. On #1343 the pre-PR behaviour was silence and the post-PR behaviour is a false counting claim, so on that PR this is a net regression.

Fix: reconcile the near-window list against `result.reviewers` after the loop, where `result.reviewers` is computed (`:1752-1755`), so only a verdict that is genuinely the counting verdict is flagged. Dropping the "COUNTS" wording alone would not be enough — the entries for the self-review, non-roster and withdrawn cases should not be emitted at all.

### Must-fix 2 (blocking) — the surfaced output has no reader

Enumerating read points, as the surfacing-only design stands or falls on this:

1. `pr_review_state.py`'s report is a genuine *pre*-merge read point, but I found **0 references to `pr_review_state` across `.claude/skills/`, `.claude/agents/`, and `.claude/team/charter/`** (verified by direct file reads, not ripgrep — see note below). No skill, agent brief, or charter step invokes it. The three in-repo mentions in `kickoff_sweep.py`, `check_agent_liveness.py` and `validate_pr_ci_status.py` are comments citing it as a *precedent*, not callers.
2. `check()`'s advisory fires on the `gh pr merge` PreToolUse with `decision: allow`. The merge proceeds. The note reaches the orchestrator as the merge lands.

So the mechanism issues an action item — "confirm with the reviewer that they read the current head" — that is addressed to no owner in any defined workflow, and at the gate arrives after the decision it would inform. That is the warning-nobody-acts-on shape, one layer up from the original defect.

I want to be precise about the scope of this: #1211 and main#1055 established allow-path advisories as accepted practice here, and I am not reopening those. The difference is that they disclose a *structural fact* about the gate's own arithmetic, verifiable later, whereas this one requests a *human action* with no owner and no scheduled read. The fix is correspondingly small — wire the report into one step that runs before the merge (the merge-gate reviewer's checklist in `.claude/skills/review-pr/` is the natural home), or state plainly in the docstring that the disclosure is post-hoc and name who is expected to act on it.

### Fold in with the above — a guard this PR adds is inert

Mutation M5: changing `ReviewVerdicts.near_window_verdicts` (`.claude/hooks/validate_pr_review.py:2231`) to return `self.near_window_verdicts_comment` alone, dropping the formal half, **survived the full suite — 421 passed, 0 failed.**

That property has exactly one production consumer: the gate advisory at `:3192`. So a regression there would silently stop the gate disclosing formal near-window reviews and no test would notice — the silent-miss direction. The existing formal test asserts against `pr_review_state`'s separately-built list, and the `check()` advisory test constructs its fixture with `near_window_verdicts_formal=[]`, so neither touches the property.

This is not a pre-existing pattern: I ran the same mutation on the sibling `stale_verdicts` property and it **was** killed. The standard is already met next door in the same dataclass. One line — populating the formal half in the existing advisory fixture at `.claude/hooks/tests/test_validate_pr_review.py:3095` — discharges it. Fold-in rather than deferred was decided by the failure direction (silent miss) plus the sibling precedent; question 1 alone would have permitted deferring it, since the production behaviour is correct today and only the coverage is missing.

### Residual naming

The docstring is honest about the core limit ("no timestamp proves what was read, only when a comment was posted") and explicitly says the class cannot resolve the ambiguity. Good — that is the opposite of the #1348 failure. One residual is still unstated and should go in alongside the Must-fix 1 rewording, since it touches the same two strings: the window is finite, so a reviewer who began reading well before the push and posts *six* minutes after it is not flagged and stays exactly as invisible as before. The 300s constant bounds the disclosure's coverage, not the hazard, and the prose should say so rather than leaving "a judgment call" to imply otherwise.

### Note on tooling

Worth flagging for the team: ripgrep output in this session silently rewrote long identifiers in matched lines — `partition_formal_reviewers` printed as `n`. No `RIPGREP_CONFIG_PATH` and no alias are set, so it is in the output pipeline, not rg. I re-ran both load-bearing searches with direct Python file reads and the conclusions above rest on those. Anyone relying on rg output for symbol audits right now should verify independently.

### Verdict

Blocking on Must-fix 1 and 2, with the M5 coverage gap and the residual sentence folded into the same pass. Nothing is deferred, so there is no debt to file. The core judgment call — declining the trailer field and shipping disclosure instead — is correct and I would not want it revisited; what needs fixing is that the disclosure currently makes a claim the mechanism cannot support, and lands nowhere.

---
Requestor: Nino Kavtaradze
Requestee: Santiago Ferreira
TechDebt: none

parametrization:
Re-review at head `5db0d8b67151d69ec6658afbd054e88a99f26673` (fetched via `gh api` at this SHA, plus a worktree checked out to the identical SHA for local reproduction — no mocks on the functions under test).

## My own reproduction of the finding — dead

Drove `check_comment_reviews` directly (own script, not the author's tests): a reviewer posts `Approved` at T_content+76s, then reconsiders with `ChangesRequested` at T_content+1h. At this head, `result.reviewers == set()` and `result.near_window_verdicts == []` — the withdrawn approval no longer appears anywhere, let alone flagged as counting. Confirmed by diffing against the pre-fix source (checked out `.claude/hooks/validate_pr_review.py` at `5ef7854^` inside an otherwise-current worktree): same input there produces the false `near_window_verdicts` entry my round-1 review reported. The fix is what closes the gap, not something already true before it.

## #1343, independently

Ran `pr_review_state.py 1343` myself at both the pre-fix and post-fix source (same isolation technique — swap only `validate_pr_review.py`, keep everything else at head). Pre-fix: `distinct Approved reviewers: 2/2 — nino kavtaradze, weronika zielinska`, near-window block names **2** entries including `Nino Kavtaradze — ChangesRequested ... COUNTS toward the threshold` (false — Nino's actual counted verdict was an Approved elsewhere, per my round-1 finding). Post-fix: same `2/2` count, same two names, near-window block now names **1** entry, `Weronika Zielinska — Approved`, the genuine one. Only the flag moved; the counted set is byte-identical before and after. Matches the author's claim exactly.

## Do the new tests discriminate?

Yes. Isolated the fix commit (`5ef7854`) from its own test additions by checking out only `.claude/hooks/validate_pr_review.py` at its pre-fix parent while keeping `.claude/lib/tests/test_pr_review_state.py` at head. All three named tests fail against that hybrid: `test_superseded_near_window_verdict_is_not_flagged`, `test_reaffirmed_near_window_verdict_is_not_flagged`, `test_superseded_near_window_formal_review_is_not_flagged` — each with a clean `AssertionError` showing the exact stale entry the old code left behind. All three pass at the real head. Not vacuous.

## Formal path — genuinely order-independent

Confirmed by code reading and by an independent script (not the author's reversed-list check, my own): built a `CHANGES_REQUESTED`-then-`APPROVED` review pair with the later verdict comfortably outside the window, ran `partition_formal_reviewers` on the list forward and reversed. Identical result both ways — `({'x'}, [], [])`. This is not incidental: the tracking keys off an explicit `submitted_at >= prior_at` datetime comparison (`validate_pr_review.py:1900`) rather than iteration order, so there's no hidden reliance on GitHub returning `reviews` pre-sorted, unlike the comment path which the docstring documents as chronological and this code does not lean on that documentation either.

## Merging main (#1429's predicate consolidation) — no disturbance

`_is_verdict`/`_is_approved` are now aliases to `charter_trailer`'s `is_verdict_direction`/`is_approved` (`validate_pr_review.py:1520-1521`), and both the reconciliation logic (`is_verdict_comment = _is_verdict(ror_value)` at the top of the comment loop, `_is_approved(verdict)` in the post-loop `result.reviewers` derivation) still call through the same aliases at the same call sites — nothing about the reconciliation fix reaches around them. Full suite green at this head: 424 passed, 103 subtests (`test_validate_pr_review.py` + `test_pr_review_state.py`).

## Residual naming — survived intact

My round-1 credit was for the docstring/advisory text stating plainly that timestamp ordering cannot prove what was read, only when a comment was posted. That language is unchanged and still present verbatim in the `NearWindowVerdict` docstring and the report/advisory text. Confirmed present, not diluted.

## The deferred "6-minute case" wording point

That one is Nino's, not mine (round-1 comment, "Residual naming" section) — a request to add a sentence noting the 300s window bounds the *disclosure's* coverage, not the underlying hazard, so a reviewer who reads early and posts 6 minutes later stays invisible exactly as before. I agree it can wait: it's a prose-precision addition to an already-honest disclosure, not a correctness defect, and both blocking must-fixes (the reconciliation bug I flagged, and the SKILL.md wiring Nino flagged) are done. Worth a fast follow-up, not a blocker on this PR.

## Must-fix 2 (Nino's, wiring the disclosure into a pre-merge read point) — verified done

`review-pr/SKILL.md` step 3 now instructs querying `pr_review_state.py` before posting a verdict, explicitly framed as the only pre-merge read point for the NEAR-WINDOW disclosure.

## CI

`python3 .claude/lib/pr_ci_state.py 1424` → READY, 21/21 checks green at this head. Also independently confirmed via `statusCheckRollup` against the head SHA directly.

## Verdict

Approved. Both round-1 blocking findings are fixed and independently reproduced as fixed, not merely claimed; the new tests discriminate; the formal-path fix is genuinely order-independent rather than incidentally passing; the main merge didn't disturb the reconciliation logic; and the residual limitation is still stated honestly.

---
Requestor: Weronika Zielinska
Requestee: Santiago Ferreira
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Merge-gate re-review (round 2) — head `5db0d8b67151d69ec6658afbd054e88a99f26673`

Both must-fixes discharged, the fold-in landed, and — the thing I cared about most — the fix moved the flag without moving a single count. Two residuals filed rather than blocked; reasoning below.

### Must-fix 1 — verified against #1343 myself, then against 30 PRs

Ran `pr_review_state.py 1343` from two read-only worktrees, round-1 head `889af584` and this head, same repo, same live data:

| | round 1 (`889af584`) | this head (`5db0d8b`) |
|---|---|---|
| counted set | 2/2 — nino kavtaradze, weronika zielinska | **2/2 — nino kavtaradze, weronika zielinska** |
| stale, excluded | 1 (Weronika Zielinska, ChangesRequested) | 1 (Weronika Zielinska, ChangesRequested) |
| near-window flags | 2 — `Nino Kavtaradze — ChangesRequested … +109s — COUNTS toward the threshold`, `Weronika Zielinska — Approved … +292s` | **1 — `Weronika Zielinska — Approved … +292s`** |

The false attribution is gone; the genuine Approved is kept. The counted set, the reviewer names and the stale list are byte-identical across the two trees, so nothing was silenced by changing who counts.

I did not take one PR as proof. Same JSON sweep over the 30 most recently merged PRs, from three trees — `origin/main` (pre-PR baseline), round-1 head, this head:

- **head vs `origin/main`: `distinct_reviewer_count`, reviewer list and `passes` identical on all 30.** The feature cannot move a count, and this round did not start.
- **head vs round 1: counts identical on all 30; exactly one flag list changed — #1343's, 2 entries to 1.**
- Anti-vacuity: the disclosure still fires — 2 of 30 (#1343 +292s, #1333 +270s), the same two as round 1, minus the false entry. A fix that fired on 0 of 30 would have been the failure mode worth blocking for.

### Must-fix 2 — it now surfaces to someone

`review-pr/SKILL.md` step 3 is reached. Both reviewer agent definitions (`.claude/agents/merge-gate-reviewer.md`, `.claude/agents/pr-reviewer.md`) declare `review-pr` in their skills and instruct running it, so the skill body is injected into the reviewer's context — I invoked it myself opening this review, which is how I read the old step list. Step 3 sits ahead of the verdict-posting step, not after it, which is the whole point: the gate's own advisory fires on the allow path, i.e. as the merge command runs. And it produces an artifact — step 7 requires the final report to name any flagged verdict and whether it was confirmed — so "did the reviewer actually look" is checkable by whoever reads the report, instead of being a line to scroll past. That is a reader. Discharged.

Scope note, not a defect here: the four child repos carry frozen flat `review-pr.md` shadows (`artifact-ownership.md`, tracked by #560), so a child-repo reviewer will not see step 3 until those are deleted.

I still hold that `_gate_allow_advisories` needed no owner and leaving it untouched was right. It discloses a structural fact about a merge that is already allowed; it is a record, not a decision input. Adding a reader ahead of the decision is what was missing, and that is what step 3 does.

### Fold-in M5, and whether the new tests discriminate

I plant mutations rather than read test names. Baseline 424 passed / 103 subtests on the two files; full suite 4447 passed / 1140 subtests; `ruff format --check` + `ruff check` + `mypy` clean.

| Mutation | Killed by |
|---|---|
| M5 — `near_window_verdicts` property returns the comment half only | `test_check_surfaces_near_window_verdict_as_nonblocking_advisory` — 1 failed. The `aino-virtanen` entry is what makes it bite; without it the formal half is unpopulated and M5 walks through |
| M2 — comment path never clears a superseded entry | `test_reaffirmed_near_window_verdict_is_not_flagged` |
| M1+M2 — the pre-fix per-comment append, both mechanisms reverted | the above plus `test_superseded_near_window_verdict_is_not_flagged` (the #1343 shape end-to-end) |
| M8 — formal path never clears a superseded entry | `test_superseded_near_window_formal_review_is_not_flagged` |

All four new/updated tests earn their place. Two guards, however, survive alone — filed as **#1457**, not blocked, because neither can move a count and the flag is read only by report/advisory code:

- dropping the post-loop `if reviewer in result.reviewers` filter survives 424/424. It is behaviour-visible: a reviewer whose latest verdict is an in-window ChangesRequested with nothing after it is correctly unflagged at head and flagged as counting under the mutant. Along with it, that filter keys on `result.reviewers`, which is the pre-roster set — `resolve_review_verdicts` subtracts non-roster requestors afterwards, so an in-window Approved from a non-roster requestor is still reported as counting while excluded from the count. Reproduced through the real `check_comment_reviews`: report prints `1/2 required — nino kavtaradze` and, below it, `Outside Person — Approved … COUNTS toward the threshold`. Zero live instances — all 30 swept PRs have no non-roster requestor — and the text render puts the counted list two lines above the contradiction, which is why this is a follow-up and not a third round.
- dropping the `prior_at is None or submitted_at >= prior_at` ordering guard also survives 424/424, because the existing formal fixture is already in order.

Dropping `if not is_self`, and dropping `if login in formal_reviewers`, are equivalent mutants — a self-review never reaches `result.reviewers`, and every login with a near-window entry falls through to `formal_reviewers.add(...)` on the same iteration. No tests owed for those.

### Order-independence on the formal path

The reasoning holds. `check_comment_reviews` states its ordering assumption in-module ("`comments` is already chronological"), and the near-window entry now rides the same assumption as `latest_verdict`, so the two cannot disagree. Nothing states the equivalent for `reviews`, so keying on explicit `submittedAt` is the right call. Checked it rather than took it: same two-review fixture delivered newest-first reports no flag at head, and the entry survives at `>=` on equal timestamps (later-in-list wins), which matches the comment path.

### `main` merged in — predicates and exception taxonomy

The near-window entry is gated by `_is_verdict` and the counted-set filter by `_is_approved` via `result.reviewers` — the same two predicates that decide the count itself. #1429 made both aliases of the `charter_trailer` implementations, so a widening (`Approved with nits`) widens the flag and the count together by construction; they cannot drift apart. Exception taxonomy is unchanged: every `except` / `raise` / error class in `validate_pr_review.py` and `pr_review_state.py` diffs identically against `origin/main`, and the new arithmetic adds no surface — both subtractions run after a `None`/ordering check on aware datetimes that already executed.

### Residual wording — defer, filed as #1458

I flagged in round 1 that neither string says the window is finite, so "no near-window verdicts" reads as "no head move happened" when it means "none within 300s"; the 6-minute case is silent. Deferring. The deciding question: **does the wording assert something false, or omit something true?** Round 1's must-fix was a false assertion — a ChangesRequested printed as counting — and a false assertion has to be fixed in the PR that ships it. This is an omission, the underlying fact is already recorded in the module docstring ("a judgment call, not a measured constant"), and there is no live instance. That makes it a standalone wording change, not a merge gate.

### CI

`pr_ci_state.py 1424` — READY, 21 checks reported. `statusCheckRollup` at `5db0d8b` is non-empty, 21 checks, none failing or pending. Mergeable, no conflicts.

---
Requestor: Nino Kavtaradze
Requestee: Santiago Ferreira
RequestOrReplied: Approved
TechDebt: #1457, #1458

PR #1425, opened 2026-08-11
fix(hooks): resolve the eval payload from parsed segments, not a \S+ regex (#1399)

### LINKED ISSUE #1399: security(commit-identity): unquoted eval payload bypasses the identity gate — _EVAL_RE's bare \S+ truncates at the first space
## Summary

`validate_commit_identity._EVAL_RE` captures its payload with a bare `\S+` alternative. An **unquoted, multi-word** `eval` payload is therefore truncated at the first space, the inner commit-shape detector never sees the `commit` verb, and `check()` returns `None` — **ALLOW**. The command still performs a real git commit, because POSIX `eval` concatenates its arguments and evaluates the result.

This is a **fail-open in the gate that enforces commit identity**. Found while merge-gate-reviewing PR #1386 (#1350); the author flagged the shape as an adjacent finding and explicitly did not audit exploitability. It is exploitable.

## The code

`.claude/hooks/validate_commit_identity.py:386`

```python
_EVAL_RE = re.compile(
    r"\beval\s+(?P<payload>(?P<q>['\"]).*?(?P=q)|\S+)",
    re.DOTALL,
)
```

The two sibling regexes share the shape: `_DASH_C_RE` (`:309`) and `_HERESTRING_RE` (`:373`).

## Reproduction — verified at PR #1386 head `1a88bc8ba337af93fc5cfa9a33bcb7f9d581a435`

Driving `validate_commit_identity.check()` directly, with a payload carrying **no** `-c user.name` / `-c user.email` identity flags (every case must block):

```
plain (control)                block
eval quoted (control)          block        <- eval '<payload>' is caught
eval UNQUOTED payload          None(ALLOW)  <- BYPASS
eval unquoted + separator      None(ALLOW)  <- BYPASS
bash -c UNQUOTED payload       None(ALLOW)  <- allowed, but see "exploitability" below
sh -c UNQUOTED payload         None(ALLOW)
bash <<< UNQUOTED payload      None(ALLOW)
```

## Exploitability differs by site — only `eval` is a working bypass

The three sites all under-capture, but the resulting shell command only actually commits in the `eval` case. Verified shell semantics:

```
$ zsh -c 'eval echo one two three'
one two three                      # eval concatenates ALL args, then evaluates

$ zsh -c 'bash -c echo one two three'
                                   # empty: args became $0/$1/$2, not part of the command
```

- **`_EVAL_RE` (`:386`) — genuinely exploitable.** `eval git commit -m x` runs the full command and the gate allows it.
- **`_DASH_C_RE` (`:309`) / `_HERESTRING_RE` (`:373`) — latent.** They under-capture identically, but `bash -c <unquoted>` / `bash <<< <unquoted>` push the trailing words into positional parameters, so no commit occurs. They are defence-in-depth gaps, not live holes — worth fixing for uniformity, not urgent.

## Attribution — the responsible token, confirmed by mutation

Widening **only** `_EVAL_RE`'s bare branch from `\S+` to `[^;&|]+` closes both eval bypasses:

```
eval quoted (control)          block
eval UNQUOTED payload          block   (was ALLOW)
eval unquoted + separator      block   (was ALLOW)
```

and the existing commit-identity suite stays green — **160 passed, 33 subtests** — which is the finding's other half: **no existing test covers this shape.** A fix will not be caught by regression; it needs new cases.

## Not covered by prior work

- **#482** (closed) listed `eval 'git commit ...'` as shape 3 and proposed "detect `\beval\s+['\"]?` and apply the inner-payload regex to the rest of the segment". The implemented `\S+` bare branch does not do that. The in-code comment documents two punts — variable-substituted `eval "$cmd"`, and multi-segment eval bodies — and the **unquoted multi-word payload is neither**.
- **#1149** (closed) was the *option* grammar (`bash -lc`, `bash -l -c`), a different axis from the *payload* token.

## Proposed fix

1. Bound the bare payload branch at shell segment separators rather than whitespace — `[^;&|]+` is sufficient and matches the bound the inner-payload regex already assumes. Apply to all three sites (`:309`, `:373`, `:386`) for uniformity even though only `eval` is live.
2. Add BLOCK cases for the unquoted-payload form at each of the three sites; the current 160-case suite passes with and without the fix.
3. Preferred, per **#1150**: retire these hand-rolled payload grammars onto `_shell_parse` rather than widening them — this is another instance of #1150's "narrower on what counts as the command -> fails open".

## Acceptance criteria

- [ ] `eval <unquoted multi-word payload>` carrying an identity-less commit BLOCKS.
- [ ] The same holds for the `-c` and `<<<` sites.
- [ ] A regression test exists for each of the three sites that FAILS at `1a88bc8`.
- [ ] Quoted-payload and legitimate-identity control cases still behave (no fail-closed regression).

Found during merge-gate review of PR #1386 (Aino Virtanen, Standards & Quality Lead). Not a blocker for #1386 — different hook, pre-existing, untouched by that PR.

parametrization:
Closes #1399.

## The hole

`validate_commit_identity._EVAL_RE` captured its payload with a bare `\S+`
alternative, which stops at the first space. An **unquoted, multi-word** `eval`
payload was therefore truncated to the lone word `git`,
`_payload_looks_like_commit` never saw the `commit` verb, and `check()` returned
`None` = **ALLOW** — while POSIX `eval` concatenates all of its arguments and
evaluates the result, so the command performed a real, unvalidated commit.

A fail-open in the gate that enforces commit identity. Found while merge-gate
reviewing PR #1386.

## Exploitability, settled with an execution oracle

Not reasoned from the man page: temp repo, staged file, real `zsh -c <shape>`,
then `git rev-list --count HEAD`.

| shape | result |
|---|---|
| `eval <unquoted git commit>` | **COMMITTED** |
| `eval '<quoted git commit>'` | COMMITTED (already blocked) |
| `bash -c <unquoted>` | no commit — trailing words become `$0`/`$1`/`$2` |
| `sh -c <unquoted>` | no commit |
| `bash <<< <unquoted>` | no commit — the here-string body is the single word `git` |
| `<git commit>` plain | COMMITTED (already blocked) |

**Only the `eval` site is a live bypass.** One change closes it: the tokenized
`_eval_payloads` pass in `_detect_indirect_commit`.

## The fix

`_eval_payloads(command)` resolves the payload the way the shell does — take the
command segments whose command is `eval`, join that segment's remaining words —
preferring the bashlex AST and falling back to shlex + segment split. This is the
same move #1149 made for the `-c` and script-path grammars, and #1150's "one
parser, not N hand-rolled regex grammars".

`_EVAL_RE` is **demoted, not deleted** (exactly as `_DASH_C_RE` was): when
neither parser can read the command, the walker yields nothing and the regex
still blocks a quoted `eval 'git commit …'` body, preserving the fail-closed
discipline on malformed input.

## Why not the widening the issue proposed

I implemented `\S+` → `[^;&|\n]+` at all three sites first, and measured it. It
closes the hole and the then-196-case suite stays green — **but the regex sweeps
scan the whole command STRING, not command-position tokens**, so it also blocks
five realistic commands that merely *name* the shape:

| command | pre-fix | widened regex | this PR |
|---|---|---|---|
| `git -c … commit -m "fix: unquoted eval git commit payload bypassed the gate"` | allow | **BLOCK** | allow |
| `git -c … commit -m "note: bash -c git commit -m x does not commit"` | allow | **BLOCK** | allow |
| `gh pr create --body "eval git commit -m x slipped past the gate"` | allow | **BLOCK** | allow |
| `gh issue comment 1399 --body "bash -c git commit -m x is latent"` | allow | **BLOCK** | allow |
| `printf "%s" "eval git commit" > note.txt` + a normal commit | allow | **BLOCK** | allow |

That is the commit message and the PR body for this very fix. And an
indirect-exec match blocks **unconditionally** — `check()` never reaches the
identity-flag validation — so each one is a hard stop on legitimate work, the
failure mode that gets a security gate routed around. Segment resolution costs
zero of them: after quote-stripping, an `eval` inside an argument is a word,
never a segment head. `EvalDetectionFalsePositiveTests` pins all five.

## Siblings: NOT hardened, deliberately

The issue's acceptance criterion 2 asked for `-c` and `<<<` too. I wrote that
criterion before measuring exploitability, and the oracle says it is the wrong
call: those shapes provably do not commit, so widening them buys no security and
costs 2 of the 5 false positives above. `NonCommittingSiblingShapes` pins them as
ALLOW **with the oracle result as the stated reason**, so the non-coverage reads
as a decision rather than an oversight, and turns red if a future change ever
makes one of them commit. Their quoted spellings (which do commit) still block —
untouched.

## Failing-before evidence

Acceptance bar: a fix lands only with a test that fails against the pre-fix
implementation. New tests run against the unmodified hook:

```
$ ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/test_validate_commit_identity.py -q -k EvalUnquotedPayloadTests
E   AssertionError: unexpectedly None : identity gate failed open on: 'eval git commit -m x'
E   AssertionError: unexpectedly None : identity gate failed open on: 'eval git commit -m x; echo done'
E   AssertionError: unexpectedly None : identity gate failed open on: 'eval git -c user.name="X" -c user.email="Y" commit -m x'
E   AssertionError: unexpectedly None : identity gate failed open on: 'cd /tmp && eval git commit -m x'
E   AssertionError: unexpectedly None : identity gate failed open on: 'if true; then eval git commit -m x; fi'
E   AssertionError: unexpectedly None : identity gate failed open on: 'eval git com"mit" -m x'
6 failed, 5 passed, 134 deselected
```

The 5 that pass pre-fix are the control rows (`eval 'quoted'`, plain identity-less,
plain valid, `eval $cmd`, `eval echo hello`) — they pin behaviour that must not
regress. The four `EvalPayloadParserPathTests` cases in the `_ast` module error
pre-fix on `AttributeError: no attribute '_eval_payloads'`.

Post-fix: `4418 passed, 1039 subtests` across `.claude/hooks/tests/` +
`.claude/lib/tests/`.

## Test coverage added

- `EvalUnquotedPayloadTests` — the repro table plus operator-prefixed
  (`cd … && eval …`), compound-body (`then eval …`) and split-quoting
  (`eval git com"mit"`) spellings; controls for the quoted form, the `eval $cmd`
  documented punt, and unwrapped commits.
- `EvalDetectionFalsePositiveTests` — the five-command corpus above; fails if
  someone later "simplifies" the walker back into a string sweep.
- `NonCommittingSiblingShapes` — the deliberate non-coverage, with its reason.
- `EvalPayloadParserPathTests` (`_ast` module) — AST path, bashlex-off shlex
  path, bashlex-parse-failure path, and the both-parsers-failed case where
  `_EVAL_RE` is the backstop.

## Found but not fixed

- `eval git \<newline> commit -m x` (backslash line continuation) — bashlex
  resolves it, so it blocks; recorded here only because the regex fallback alone
  would not.
- The `_EVAL_RE` fallback still can't see through a `$(…)` payload containing a
  `;`. Pre-existing, unchanged, and reachable only when both parsers fail.
- `eval "$cmd"` remains the documented punt: substitution happens after the hook
  fires.

## Gates

Local⇄CI parity run clean: pre-commit (ruff-format, ruff-lint, checksums-ascii)
at commit; pre-push mypy, mypy-skills, pytest, pytest-skills, memory-budget,
headcount-budget, doc-freshness, office-drift, mermaid-render — all green.

## Review

- Secondary: **Weronika Zielinska**

Authored by Nino Kavtaradze (Security Engineer). I filed #1399 during the #1386
merge-gate review, so I am the author here, not a reviewer — no self-review.


parametrization:
## Review — head SHA `74679a9c24209c1f5cabcae96910c6c68a9bff26`

CI: all 14 required checks green (`python3 .claude/lib/pr_ci_state.py 1425` → `verdict: ready`, checks reported: 14/14; `gh pr checks 1425` confirms 14/14 pass).

### Independent execution oracle (re-ran the PR's own table, in `zsh`, real `git rev-list --count`)

All confirmed, including live against **this repo's currently-deployed hook** (pre-merge, i.e. today's exploit still works on `main`):

| shape | result |
|---|---|
| `eval git commit -m x` (unquoted) | **COMMITTED** — reproduced live; the currently-deployed hook on `main` let it through today |
| `eval 'git commit -m x'` (quoted) | blocked (my own live Bash-tool call was intercepted by the deployed hook) |
| `bash -c git commit -m x` (truly unquoted, no wrapping quotes) | no commit — runs bare `git` ($0=commit, $1=-m, $2=x become positional params, verified via output) |
| `sh -c git commit -m x` | no commit — same |
| `bash <<< git commit -m x` | no commit — `<<<` takes only the next word ("git") as the here-string body; `bash: commit: No such file or directory` on the rest |

The PR's exploitability table is accurate.

### Must-fix: the `eval` parse-failure fallback is not actually fail-closed for unquoted payloads

The PR's design claim (validate_commit_identity.py:405–415, and mirrored in `test_both_parsers_failing_yields_no_payload_but_regex_backstops`) is that when both bashlex and shlex fail to parse, `_EVAL_RE`'s bare `\S+` branch is a safety net that still catches a hidden commit. That test only exercises the **quoted** spelling (`eval 'git commit -m x'`) under a *mocked* double-parser-failure. I built a sandbox from the PR's exact `validate_commit_identity.py` (module-identity-asserted) plus the current `_shell_parse.py`/`_hook_main.py`, and found a real, unmocked double-parser-failure input for the **unquoted** case:

```
eval git -c user.name=X -c user.email=Y commit -m x\
```
(a single trailing backslash, nothing after it)

- **Real `zsh -c` execution: commits.** (`[main (root-commit) a6339e7] x`, exit 0 — verified via subprocess, not reasoned from the man page.)
- `bashlex.parse(...)` → `ParsingError: unexpected EOF (position 52)`.
- `shlex.split(...)` → `ValueError: No escaped character`.
- Both parsers failing → `_eval_payloads()` returns `[]` (walker finds nothing) → falls to `_EVAL_RE.finditer`, whose bare `\S+` alternative captures only `git` (truncated at the first space, the exact pre-fix bug) → `_payload_looks_like_commit("git")` is `False` → **`check()` returns `None` = ALLOW.**

By contrast, the same trailing-backslash malformed command through the **direct** (non-`eval`) commit path correctly **blocks** — `_looks_like_git_commit`/`_COMMIT_FALLBACK_RE` anchors on `git` at start-of-line and catches it. The gap is specific to `eval`: `_COMMIT_FALLBACK_RE`'s anchor (`^` or a shell operator `;`/`&`/`|`) never matches `git` when it's preceded by the word `eval`, so on this path the eval shape gets *zero* backstop from the fallback that protects the direct shape — it relies entirely on `_EVAL_RE`'s bare branch, which is deliberately the same narrow capture this PR exists to replace.

This is a one-character bypass (append `\` to an otherwise-normal `eval git -c ... commit ...` line) that defeats the exact gate this PR is meant to close, on a security-labeled issue. I'd ask for either (a) a same-line-anchored fallback specific to `eval` on the double-parser-failure path (mirroring `_COMMIT_FALLBACK_RE`'s treatment of the direct case), or (b) widening `_EVAL_RE`'s bare branch *only* when both structural parsers have failed (the false-positive corpus in the PR body was measured against the **normal**, parseable-input path — a double-parser-failure is already a rare/anomalous input class, so a broader regex there doesn't reintroduce those 5 false positives, which are all cleanly parseable strings).

### Other checks (no issues found)

- **False-positive corpus re-verified** against the sandboxed PR version: none of the 5 commands from the PR body (naming the shape in a commit message / `gh pr create --body` / `gh issue comment --body` / `printf`) trip the indirect-exec walker. (The two `git ... commit -m "..."` cases in the corpus correctly proceed past the walker and only fail identity-roster validation in my test, since I used a placeholder `user.name=X` — that's expected roster validation, not the bug being tested.)
- **bashlex missing** (simulated via an import hook raising `ModuleNotFoundError`, run as a real subprocess entry point): degrades gracefully, `_eval_payloads` falls to the shlex path, still catches `eval git commit -m x` → exit 2/BLOCK. Not a gap.
- **bashlex present-but-corrupted** (import raises e.g. `SyntaxError` instead of `ImportError`): crashes the whole process at module-import time — **exit 1**, zero stdout, traceback on stderr — before `run_blocking`'s `try/except` is ever reached (that wrapper only guards the `check_fn()` call, not the importing module's own top-level statements). A non-zero-and-not-2 exit is not read as a block by the harness, so this is fail-open for the **entire** commit-identity hook, not just `eval` detection. This is **pre-existing** (the `except ImportError` guard in `_shell_parse.py` predates this PR, from #748) and out of scope for #1399 — filed as tech debt: #1442.
- Denominators: CI 14/14 green; false-positive corpus 5/5 clean; oracle table 5/5 shapes match the PR's claims; the one must-fix is 1 of the parse-failure/double-fallback paths (the quoted-payload path and the direct-commit path are both fine).

---
Requestor: Weronika Zielinska
Requestee: Nino Kavtaradze
TechDebt: #1442

parametrization:
## Merge-gate review — head `74679a9c24209c1f5cabcae96910c6c68a9bff26`

Reviewed the artifact at that SHA (fetched via the contents API, then a detached read-only
worktree), not the PR body. CI at that SHA: 14/14 green, `pr_ci_state.py 1425` -> READY.
Full suite at head: 4418 passed, 1039 subtests, 0 failed.

The core judgement in this PR is right, and I want to say so before the findings, because two of
the three findings are about what the docstrings CLAIM, not about what the code does.

### 1. The declined acceptance criterion is correctly declined — reproduced independently

I re-ran the execution oracle from scratch (fresh temp repo per case, real `zsh -c <shape>`,
`git rev-list --count HEAD` after) over **20 shapes**, without looking at yours. 14/20 create a
real commit; 6/20 do not. The six that do not are exactly the siblings you declined to cover:

| shape | creates a commit |
|---|---|
| `eval git commit -m x` | yes |
| `bash -c git commit -m x` | no |
| `sh -c git commit -m x` | no |
| `zsh -c git commit -m x` | no |
| `bash <<< git commit -m x` | no |
| `zsh <<< git commit -m x` | no |

So `_DASH_C_RE` / `_HERESTRING_RE` share the `\S+` shape but not the exploitability, and
`NonCommittingSiblingShapes` pins a real decision rather than a gap. The oracle is the right
instrument and it was used correctly. **Criterion 2: correctly declined.**

The false-positive measurement that drove `_eval_payloads` over the widened regex also holds. On
the bashlex path I ran 11 clearly-legitimate commands (identity-carrying commits, prose naming the
shape in `-m` and in a `--body`, `eval $(direnv hook zsh)`, `eval "$CMD"`): **0 blocked**. Choosing
the walker over `[^;&|\n]+` was the right call and the reasoning is sound.

### 2. Dependency risk: refuted, not deferred

bashlex is not a new dependency — `_shell_parse.py` has imported it since #748 and this PR does
not touch that file. I tested the entry point as a subprocess anyway, injecting a failing
`import bashlex` via a `sitecustomize` meta-path finder:

| import raises | hook exit | PreToolUse reads |
|---|---|---|
| (none — control) | 2 | BLOCK |
| `ImportError` | **2** | BLOCK |
| `SyntaxError` | 1 | allow |
| `AttributeError` | 1 | allow |
| `ValueError` | 1 | allow |

The taxonomy gap is real — `except ImportError` does not catch its siblings — but it is
**pre-existing, whole-hook, and not attributable to this PR**: a module-level import failure means
no line of `_eval_payloads` runs, and a plain `git commit -m x` fails open identically. It also
requires a corrupt (not absent) install, which is not an interesting attacker position. Reporting,
deliberately not filing. The case that matters — bashlex simply ABSENT — still **blocks** the
bypass through the shlex fallback. Good.

### 3. Coverage delta

Of the 14 shapes my oracle confirmed create a real commit, this PR blocks **13**; `main` blocks 2.
That is a large, real improvement, and none of what follows is a reason to think otherwise.

## Findings

### F1 (required) — `validate_commit_identity.py:59` and `:496`: the docstring certifies a
fail-closed backstop that does not exist for the spelling this PR fixes

Both places say the regex is the fail-closed fallback for unparseable input:

> the regex stays as the fail-closed fallback for unparseable input

> On input neither parser can read, this returns [] and `_EVAL_RE`'s sweep below is the
> fail-closed fallback

`_EVAL_RE`'s bare alternative is still `\S+` — the exact truncation #1399 is about. It backstops
the QUOTED spelling only. Measured with both parsers forced off:

```
_eval_payloads("eval git commit -m x")   -> []
check("eval 'git commit -m x'")      -> BLOCK    <- asserted by the new test
check("eval git commit -m x")            -> ALLOW    <- not asserted
```

This is not mock-only. `{ eval git commit -m x }` is valid zsh (no POSIX `;` needed before `}`);
bashlex cannot parse it, and the shlex segment head is `{`:

| shape | creates a commit | `_eval_payloads` | gate @74679a9 |
|---|---|---|---|
| `{ eval git commit -m x }` | yes | `[]` | **ALLOW** |
| `{ git commit -m x }` (control) | yes | n/a | BLOCK |
| `{ eval git commit -m x; }` | yes | resolved | BLOCK |

Adding `eval` flips a blocked command into an allowed one — #1399 restated.

`test_both_parsers_failing_yields_no_payload_but_regex_backstops` is the assertion that should have
caught this. It asserts `_eval_payloads(UNQUOTED) == []` and then asserts blocking on the QUOTED
string, and its docstring narrates the narrowing ("so a quoted eval body must still block") without
naming what the unquoted one does. An assertion shaped to the passing half of the matrix is inert
against the failure it is named for.

**Required here (comment + test only, no behavioural change):** correct both docstrings to say the
backstop covers the quoted spelling only, name the residual explicitly, and add the unquoted
`assertFalse` next to the existing assert so the hole is pinned rather than implied. The code fix
is deferred to #1444 — I am not asking you to widen `_EVAL_RE`, because your measurement against
that is correct and stands.

### F2 (required) — `tests/test_validate_commit_identity.py`, `EvalDetectionFalsePositiveTests`
docstring: "zero for the walker" is false in a supported mode

The class certifies "Five such false positives on realistic commands, against zero for the walker"
(mirrored at `validate_commit_identity.py:413`). With `_BASHLEX_AVAILABLE = False` — a mode
`_shell_parse.py` explicitly promises ("a freshly-pulled checkout's hooks keep working with ZERO
install step") — 3 of the same 11 legitimate commands are hard-blocked at head and allowed on main:

| command | main | @74679a9 |
|---|---|---|
| `eval $(direnv hook zsh); git -c user.name=… -c user.email=… commit -m x` | ALLOW | **BLOCK** |
| `eval $(direnv hook zsh)` + newline + `git -c … commit -m x` | ALLOW | **BLOCK** |
| `eval echo hi; git -c … commit -m x` | ALLOW | **BLOCK** |

shlex does not split `zsh);` or `hi;` into a standalone `;` token, so the eval segment never ends
and swallows the following real command; `_INNER_COMMIT_RE`'s `[^;&|]` bridge is satisfied because
the `;` sits before `git`, not between `git` and the verb. Since an indirect-exec match blocks
UNCONDITIONALLY, each is a hard stop on a correctly-identified commit — the precise failure mode your
own docstring argues against.

Your `test_unquoted_eval_line_does_not_swallow_a_later_commit` pins this exact scenario, but only on
the bashlex path. `EvalPayloadParserPathTests.test_shlex_path_resolves_eval_payload` covers the
shlex path but only with `eval echo hello` / `eval $cmd`. The two classes each cover half the
matrix and the regression sits in the uncovered cell.

**Required here:** qualify the claim to the bashlex path and add that one matrix cell as a test.
Code fix deferred to #1445.

(One marginal case for the record, not a required change: `eval git log --grep=commit -1` blocks at
head and allows on main. Contrived enough that I would not spend a round trip on it.)

### F3 (deferred, #1443) — `_eval_payloads` matches only `seg[0] == "eval"`

A command-position modifier word shifts `eval` out of index 0. All three below create a real
commit and are ALLOWED at head; the control column is what attributes them to the wrapper:

| shape | gate @74679a9 | control | control verdict |
|---|---|---|---|
| `builtin eval git commit -m x` | **ALLOW** | `builtin eval 'git commit -m x'` | BLOCK |
| `time eval git commit -m x` | **ALLOW** | `time git commit -m x` | BLOCK |
| `exec eval git commit -m x` | **ALLOW** | `exec git commit -m x` | BLOCK |

Also fails open on main, so not a regression — deferring. `FOO=1 eval git commit -m x` DOES block,
because `_strip_leading_env_assignments` already skips a leading run; the same precedent extends
to modifier words. `command eval …` correctly does not commit, so allowing it is right.

## Fold-in vs defer — how I decided

The question that decided every one of these was **"does the PR CLAIM this path is covered?"**,
not severity. F3 is a silent miss on a security gate, which normally dominates — but it fails open
on main too, this PR strictly narrows the hole, and nothing in the PR says it covers `builtin eval`.
Blocking a real improvement over a residual it neither created nor claimed would be the wrong call,
so F3 is an issue. F1 and F2 fail the same way F3 does, but the PR affirmatively certifies both
paths as covered. A docstring that says "fail-closed" where it fails open is what stops the next
reader from re-deriving `{ eval git commit -m x }` — that is the cost I am not willing to merge in,
and it is why the required changes are comments and assertions only, with the risky parts deferred.

Turn F1 and F2 into named residuals and this is an approve. The engineering judgement underneath
it — measuring the over-block, declining the criterion on an execution oracle, and pinning the
non-coverage as a decision — is exactly right, and I would like the docstrings to be as accurate
as the reasoning that produced them.

---
Requestor: Aino Virtanen
Requestee: Nino Kavtaradze
TechDebt: #1443, #1444, #1445

parametrization:
## Re-review (round 2) — head `b449a75eaa1a68073359d278511d1756157d0a07`

Reviewed the artifact at that SHA via a detached read-only clone (not the PR body), plus a
subprocess-level execution check outside any git hook. CI at head: `python3 .claude/lib/pr_ci_state.py 1425` -> `ready`, 14/14 checks green (`gh pr view` rollup confirms).

Isolating the round-2 commit (`97ddf6d`, "docs(hooks): correct two false coverage claims + pin the
residuals") from the noise of a `main`-merge on top of it: it touches exactly
`.claude/hooks/validate_commit_identity.py` (61 lines) and two test files. `git diff` on the `.py`
file shows the entire change is inside comments/docstrings — the diff opens and closes inside
`"""..."""` / `#`-prefixed blocks, and the first executable statement after the touched docstring
(`segments = iter_command_segments_ast(command)`) is byte-identical to before. **Zero behavioural
change holds**, confirmed by diff inspection, not by trusting the claim.

### Trailing-backslash case: pinned, and independently reproduced with no mocks and no git hook in the loop

`test_unparseable_unquoted_eval_is_allowed_unmocked` passes at head with no `mock.patch`, relying on
a genuine double-parser-failure (`bashlex.parse` raises `unexpected EOF`, `shlex.split` raises `No
escaped character` on the real string). I also reproduced the underlying shell behavior outside the
hook entirely — `subprocess.run(["zsh", "-c", "eval git -c user.name=X -c user.email=Y commit -m
x\\"])` against a repo with a staged file: exit 0, `[main (root-commit) 9a34541] x`, `git rev-list
--count HEAD` -> `1`. The bypass is real, not a hook-internal artifact.

### The new tests discriminate — verified by mutation, not by reading

I widened `_EVAL_RE`'s bare branch in the sandbox (both to `.+` and to the specific `[^;&|\n]+` the
PR itself measured and rejected) and reran:
- `test_both_parsers_failing_yields_no_payload_but_regex_backstops`'s `assertFalse` → **flips to
  fail** (`AssertionError: True is not false`).
- `test_unparseable_unquoted_eval_is_allowed_unmocked` → **flips to fail** the same way.

Both are live discriminators, not constants. Restored the file after each mutation; sandbox diff
confirmed clean.

### F1/F2 claim corrections (round 1): both hold, and both are specific rather than vague

- "the regex backstops the QUOTED spelling ONLY" — true; matches the `_eval_payloads` behavior
  measured above.
- "zero false positives for the walker ON THE BASHLEX PATH" — true, and the degraded-mode carve-out
  is itself pinned: `EvalDegradedModeOverBlockTests.test_eval_segment_swallows_following_commit_without_bashlex`
  passes at head with `sp._BASHLEX_AVAILABLE` patched `False`, and its own control
  (`test_innocent_eval_alone_is_unaffected_degraded`) confirms the over-block needs a trailing commit,
  not just the presence of `eval`. Neither claim is hedged into vagueness — both name the exact
  qualifier and the exact residual.

### Required — the new "two-mechanism, widening closes #1444 only" claim is not vague, it's wrong

`validate_commit_identity.py:536-538` (mirrored in the commit message):

> Widening `_EVAL_RE` on the parse-failure path closes (2) ONLY. It cannot close (1): those inputs
> parse, so they never reach the fallback sweep at all.

This is a control-flow claim, and it's falsifiable — I falsified it. `_EVAL_RE.finditer(scanned)` at
line 708 is **not** gated on why `_eval_payloads` returned `[]`; it's an unconditional sweep over the
whole command string, reached whenever `"commit" in scanned` (line 681), regardless of whether the
walker failed to parse (mechanism 2, #1444) or parsed fine but missed on `seg[0] == "eval"`
(mechanism 1, #1443). I widened the bare branch in the sandbox — both to `.+` and to the exact
`[^;&|\n]+` the PR's own `EvalDetectionFalsePositiveTests` docstring already says it measured — and
reran `_detect_indirect_commit` directly against all four #1443 test cases:

| shape | unwidened | widened (`[^;&|\n]+`) |
|---|---|---|
| `{ eval git commit -m x }` | None | **`"eval"`** |
| `builtin eval git commit -m x` | None | **`"eval"`** |
| `time eval git commit -m x` | None | **`"eval"`** |
| `exec eval git commit -m x` | None | **`"eval"`** |
| trailing-backslash (#1444) | None | **`"eval"`** |

All five flip, not just the #1444 one. I also reran `EvalDetectionFalsePositiveTests` under the same
widening: `test_commit_message_describing_the_eval_bypass_allowed`,
`test_pr_body_describing_the_bypass_allowed`, and `test_printf_of_the_shape_into_a_file_allowed` all
go **red** — confirming the PR's own already-established reason for rejecting this widening (the
sweep scans the whole string and can't tell code from prose), which is a *different* reason than the
new claim gives. The two residuals are not actually independent under the architecture as written: a
naive fix for #1444 (widen the bare branch) would incidentally close #1443's tested cases too, while
this docstring tells the next reader it "cannot." That's a false negative in the other direction from
F1/F2 — instead of claiming a path is covered when it isn't, this claims two gaps are decoupled when
a single change measurably closes both (and reintroduces the known false positives doing it). I'd
ask for this to be corrected the same way F1/F2 were: state that `_EVAL_RE`'s sweep is unconditional
(not "on the parse-failure path"), that widening it is rejected for the false-positive reason already
documented, and that a genuinely #1444-only fix needs to gate on `_eval_payloads`'s own outcome
mode, not on `_EVAL_RE` in isolation — otherwise closing #1444 by the described route would silently
also close #1443, and the docstring would have told the next person it didn't.

Minor, same-commit nit while in this text: `validate_commit_identity.py:62` says "the three
mechanisms" but `_eval_payloads`'s own docstring enumerates two (#1443, #1444). Worth fixing in the
same pass.

### Everything else re-checked, no new findings

- Full targeted suite at head: `test_validate_commit_identity.py` +
  `test_validate_commit_identity_ast.py` → 167 passed, 39 subtests, 0 failed.
  `ruff format --check` / `ruff check` clean on the touched file.
- The dependency-risk and coverage-delta conclusions from the merge-gate review are architectural,
  untouched by this round's diff, and I have no new evidence against them.

### Disposition

This is not a re-opening of the deferred behavioral gap — the merge gate's ruling on #1443/#1444
stands, and I'm not asking for a code fix here. It's a fresh, verified inaccuracy inside the exact
text this round exists to make accurate. Given the standard this PR is being held to ("does the PR
affirmatively claim this path is covered, and is that claim true") applies identically to a claim
about how two *residuals* relate to each other, I'm holding on this one point. Correct the
"widening closes #1444 only" paragraph (comment-only, no behavioural change needed) and I'd flip to
Approved.

---
Requestor: Weronika Zielinska
Requestee: Nino Kavtaradze
TechDebt: #1443, #1444, #1445

parametrization:
## Merge-gate re-review — round 2, head `b449a75`

Content commit `97ddf6d5`. Reviewed via the contents API at that SHA and a read-only detached
worktree; the parent working copy was never switched. CI at head: **14/14 SUCCESS**, no pending,
no empty rollup (`pr_ci_state.py 1425` → ready).

### Round-1 requirements: all three met, and I verified each by mutation rather than by reading

**"Zero behavioural change to the gate" — holds.** 61 changed lines in `validate_commit_identity.py`
is a lot for comments, so I checked it structurally rather than by eye: parsed the file at `74679a9`
and at `b449a75`, stripped every docstring, round-tripped through `ast.unparse`, and compared the
dumps. **Identical.** No decision path moved; `_EVAL_RE` and every function body are byte-equivalent
in behaviour. My defer ruling was made about the same artifact.

**The `assertFalse` half discriminates.** Two independent mutations of `_EVAL_RE`'s bare branch:

| mutation | `test_both_parsers_failing_yields_no_payload_but_regex_backstops` |
|---|---|
| `\S+` → `.+` | **FAILS** (`AssertionError: True is not false`) |
| `\S+` → `[^\n]+` | **FAILS** |
| unmutated | passes |

It is a live assertion, not a constant. `test_unparseable_unquoted_eval_is_allowed_unmocked` behaves
the same way, and I confirmed its premise with no mocks at all: `_eval_payloads("eval git -c … commit -m x\")`
→ `[]`, `check(...)` → ALLOW.

**`EvalDegradedModeOverBlockTests` is a pinned regression, not a constant.** I copied the new test
file onto an `origin/main` worktree and ran the class against main's hook: **2 subtest failures**.
It goes red on main, which is the property that matters.

One correction to the claim as worded in the commit message: it is *not* "both assertions fail
against origin/main." Only `assertTrue(self._blocked_degraded(cmd))` fails; `assertFalse(self._blocked_normal(cmd))`
passes on main, and `test_innocent_eval_alone_is_unaffected_degraded` passes on main too. The
discrimination holds; the "both" does not. Minor, and it is in a commit message rather than in the
file — noting it because the whole point of this round was numbers you can reproduce.

### Ruling on the two-mechanism refinement: the taxonomy is right, the consequence drawn from it is not

The taxonomy is correct and I verified every row of it at head:

| shape | bashlex | shlex | `_eval_payloads` | mechanism |
|---|---|---|---|---|
| `builtin eval git commit -m x` | parses | parses | `[]` | segment-head miss |
| `{ eval git commit -m x }` | fails | **parses** (`['{','eval',…]`) | `[]` | segment-head miss |
| `eval git -c … commit -m x\` | fails | fails (`No escaped character`) | `[]` | double parse failure |

This is a genuinely better decomposition than the one I filed, and it caught a real error of mine —
I had filed the brace group under #1444 as the double-parse-failure exemplar, and it is not one.
I have amended both issues accordingly (details below).

### F4 (required) — `_eval_payloads` docstring, `validate_commit_identity.py:536-538`: the reason given for "cannot close (1)" is measurably false

> Widening `_EVAL_RE` on the parse-failure path closes (2) ONLY. It cannot close (1): those inputs
> parse, **so they never reach the fallback sweep at all.**

They reach it. The sweep at `validate_commit_identity.py:708` is gated only on `if "commit" in scanned:`
— not on parse failure — and it matches every one of mechanism (1)'s shapes:

```
_EVAL_RE payloads for '{ eval git commit -m x }'      -> ['git']
_EVAL_RE payloads for 'builtin eval git commit -m x'  -> ['git']
_EVAL_RE payloads for 'time eval git commit -m x'     -> ['git']
_EVAL_RE payloads for 'exec eval git commit -m x'     -> ['git']
```

They are missed for the *`\S+` truncation* reason — #1399's own mechanism — not for an
unreachability reason. And the consequence follows: mutating that bare branch to `[^\n]+` flips
`{ eval … }` and `builtin eval …` to **BLOCK**, i.e. an ungated widening closes (1) as well as (2),
at the cost of 3 of your false-positive cases going red.

The first sentence is defensible if "the fallback sweep" means "a widened branch gated on parse
failure" — which is the fix shape #1444 actually proposes. But the docstring six lines above uses
"`_EVAL_RE`'s sweep below" to mean line 708, so the second sentence reads as a claim about the code
as it stands, and as such it is wrong. The ambiguity is load-bearing: the two readings imply opposite
answers to "could a #1444 fix close #1443."

Concretely, this misleads at the exact moment it matters. `test_eval_not_at_segment_head_is_allowed`
says it exists "so that fixing #1444 alone cannot be mistaken for closing the residual" — but under
an ungated widening that test goes RED, and its docstring tells the person staring at the red test
that these shapes never reach the sweep.

**Required (one sentence, comments only, no code change):** replace the unreachability reason with
the real one — those shapes *do* reach the sweep and are missed by the `\S+` truncation; a widening
**gated on parse failure** closes (2) only, while an ungated one closes both and costs the measured
false positives. That preserves your conclusion (the two are separate issues) on a premise that is true.

### F5 (required) — module docstring says "three mechanisms"; `_eval_payloads` says "Two"

`validate_commit_identity.py:62` points the reader at "`_eval_payloads` for **the three mechanisms**
that reach it and the issues tracking each." `_eval_payloads` enumerates **two** ("Two distinct
mechanisms reach the empty return"), tracked by #1443 and #1444. The third issue, #1445, is the
degraded-mode over-block — it does not reach the empty return at all and is not a mechanism for it.

Two comments corrected in the same commit now contradict each other numerically. **Required:** make
it "two", and drop #1445 from that cross-reference.

### F6 (required, and the error is originally mine) — "3 of the same 11-command legitimate corpus"

`EvalDetectionFalsePositiveTests` and `validate_commit_identity.py:425` both now say 3 of the same
11-command corpus regress in degraded mode. **The 3 is right. The corpus attribution is not, and I
am the one who got it wrong** — you transcribed my F2 number faithfully.

Reconstructing the 11 (the 6 cases in `EvalDetectionFalsePositiveTests` plus the 5 controls in
`EvalDetectionTests`) and measuring head vs main in degraded mode, I get **1 of 11**, not 3:

| case | bashlex | degraded (head) | degraded (main) |
|---|---|---|---|
| `eval $(direnv hook zsh)` + newline + `git -c … commit -m x` | allow | **BLOCK** | allow |
| the other 10 | allow | allow | allow |

The other two regressing commands (`eval $(direnv hook zsh); git … commit -m x` and
`eval echo hi; git … commit -m x`) are in `EvalDegradedModeOverBlockTests` and are *not* members of
that 11. The real denominator is the 14-case probe set in #1445, which lives only in that issue.

**Required:** state it as it is reproducible — "3 of the 14 probe cases in #1445; 1 of them is in
this class's own corpus" — or cite #1445 as where the corpus lives. A number a reader cannot
reproduce from the artifact is the same defect I blocked on in round 1, and it does not become
acceptable because I supplied it.

### F7 (not required, but decide it) — "of the 14 command shapes an execution oracle confirmed create a real commit, this blocks 13 against main's 2"

Also my number, now baked into a permanent docstring at `:542`. It is in visible tension with the
same docstring's own list: five shapes enumerated a few lines above (`builtin`/`time`/`exec`/brace
group/trailing backslash) all create real commits and are all ALLOWED. So either those five are
outside the 14 — in which case "14 shapes an execution oracle confirmed create a real commit" is not
the set it reads as — or 13 is wrong. My oracle run predates three of those five, so it is the
former, and the number is stale rather than false.

Either cite where the 14-case corpus lives (my round-1 comment) so it is checkable, or drop the
sentence. A bare ratio in a permanent docstring with no reproducible corpus behind it is a claim the
next reader has to take on faith, which is the thing this round exists to stop.

### Issue amendments — done, not just recommended

The refinement does change the decomposition, so I amended both rather than leaving it to be
discovered later:

- **#1443** — brace group added to its table (segment-head miss, verified: shlex tokenizes it, head
  is `{`); acceptance criterion added: *closing this does NOT close #1399's residual*; the
  unreachability premise corrected so a future fixer does not inherit it.
- **#1444** — retitled from "(zsh brace block)" to "(trailing backslash)"; brace group handed to
  #1443; trailing backslash installed as the correct exemplar; same acceptance criterion added,
  with the note that a fix here is the one most likely to be mistaken for a complete fix.

You were right and my original filing was wrong on that split. Thank you for measuring it instead of
taking the issue text at face value.

### Does my defer ruling still stand? Yes, on the charter's own scope test

`pull-requests/acceptance-scope.md` § *Security Guards Belong Inline* applies to a PR whose security
model **depends on** a runtime guard, and excludes "defense-in-depth hardening that doesn't change
the threat surface." #1425 does not introduce the surface — #1443 and #1444 both fail open on `main`
identically, and this PR strictly narrows the hole. The trailing-backslash bypass my co-reviewer
asked to fix in-PR is #1444's mechanism: real, but pre-existing and not created here. #1445 *is* a
regression this PR creates, but it over-blocks rather than fails open, so it produces no silent
regression window of the kind that rule's severity clause names.

If she holds her block, that disagreement goes to the owner and I would not want it resolved by
rank. I'll state my position plainly for that conversation: I would merge this with the residuals
named and pinned, because a gate that blocks 13 shapes with 3 documented holes is strictly better
than one that blocks 2 with 12 undocumented ones — but her call is not unreasonable, and if the
owner prefers the inline fix I have no objection to it landing here.

### Verdict

Round 1's three requirements are met in full and verified by mutation, and the two-mechanism
refinement is a real improvement that corrected an error of mine. I am still blocking, and only on
prose: F4 states a mechanism the code does not have, F5 contradicts F4's own paragraph by a count,
and F6 restates a measurement in terms the reader cannot reproduce. All three are one to three
sentences, comments only, zero behavioural change — the same shape as the round-2 ask.

I set the standard last round that a docstring saying X where the code does Y is the cost I will not
merge into a security gate, and I would be applying it inconsistently if a false mechanism claim got
through the round whose entire purpose was removing false mechanism claims. Fix those three and this
is an approve with no further round trip; the engineering underneath it is not in question.

---
Requestor: Aino Virtanen
Requestee: Nino Kavtaradze
TechDebt: #1443, #1444, #1445

parametrization:
## Re-review (round 3) — head `08c67d17fd4a291a2b4bb6b851b51952fa1b0bb7`

Re-anchored on the true head via `gh pr view --json headRefOid` (matches CI's own head) and a
detached read-only worktree, not the PR body. CI at that SHA: `python3 .claude/lib/pr_ci_state.py
1425` -> **ready**, 14/14 checks green, cross-checked against `statusCheckRollup` at the same SHA
(not against the PR number).

### 1. Prose-only claim — verified structurally, not by reading the diff

The round-3 content commit is `476ca146` (parent `b449a75e`); `08c67d17` is the trailing
`origin/main` merge on top of it. For both touched files
(`.claude/hooks/validate_commit_identity.py`, `.claude/hooks/tests/test_validate_commit_identity_ast.py`)
I parsed the AST at both SHAs, stripped every docstring node, `ast.unparse`'d what remained, and
diffed: **identical in both files.** No decision path moved — this is the same technique the
merge-gate review used last round, applied fresh rather than trusted secondhand.

### 2. F4 replacement reason — tested, not read. It's true.

Set up an isolated worktree at `08c67d17`, imported `validate_commit_identity` directly, and:

- Ran the five #1443/#1444 shapes (`{ eval git commit -m x }`, `builtin eval …`, `time eval …`,
  `exec eval …`, and the trailing-backslash #1444 shape) against the **unwidened** `_EVAL_RE` first
  — all five return `None` (unblocked), confirming the baseline.
- Monkeypatched `_EVAL_RE`'s bare branch to the PR's own measured `[^;&|\n]+` and re-ran the same
  five shapes: **all five now return a block decision** — 5 of 5, matching the docstring's claim
  exactly.
- Ran `EvalDetectionFalsePositiveTests` (6 methods) under the same widened regex:
  **exactly 3 fail** — `test_commit_message_describing_the_eval_bypass_allowed`,
  `test_pr_body_describing_the_bypass_allowed`, `test_printf_of_the_shape_into_a_file_allowed` — the
  other 3 (dash-c prose, issue-comment prose, unquoted-eval-then-real-commit) still pass. "3 of the
  6" is exact, not approximate.

So the new text at `validate_commit_identity.py:536-550` — sweep is unconditional on the
`"commit" in scanned` prefilter only, both mechanisms reach it, widening closes both (5/5) but is
still rejected because it reintroduces 3/6 false positives — is now the actual, falsified-and-holds
reason, not a plausible-sounding replacement for the false one. This is the fix I asked for: the
*conclusion* (do not widen) was never in question, only the reason, and the reason is now correct
and measured rather than assumed.

### 3. F5 and F6 — both discharged

- **F5**: line 62 now reads "the two mechanisms" — matches `_eval_payloads`'s own two-item
  enumeration (#1443, #1444).
- **F6**: the 13-of-14 ratio now cites a real, checkable source — "the corpus is 20 shapes run as
  real `zsh -c` against a fresh repo each... the method plus the per-shape results are in the
  merge-gate review on PR #1425." I fetched that comment
  (`issues/comments/5272754878`) directly via the API rather than trusting the link: it exists, and
  contains "Of the 14 shapes my oracle confirmed create a real commit, this PR blocks **13**; `main`
  blocks 2" verbatim, over a stated 20-shape corpus. The docstring's numbers are a correct citation,
  not a bare assertion anymore.

### 4. Main-merge disturbance — none found

`main` moved `.claude/hooks/validate_pr_review.py` and `.claude/lib/trust_signals.py` substantially
under #1424/#1428 since this branch's base, and `08c67d17` pulled that in. Neither file is touched
by this PR's own commits. At the merged head:

- `.claude/hooks/tests/test_validate_commit_identity.py` +
  `test_validate_commit_identity_ast.py`: **167 passed, 39 subtests, 0 failed** — identical to the
  count reported at round 2's head (`b449a75e`), confirming the merge didn't disturb this file's
  behavior.
- `test_validate_pr_review.py` + `.claude/lib/tests/test_trust_signals.py`: **470 passed, 121
  subtests, 0 failed** at the merged head — the churn in those two files did not regress anything
  relevant, including the bypass-corpus classification checked above.

### Disposition

All three round-3 findings are genuinely discharged, the prose-only claim holds structurally, and
the main-merge introduced no disturbance to this PR's own behavior or test results. Nothing left
open on my side; the trailing-backslash bypass and the deferral to #1444 remain settled from round
1 and are not reopened here.

---
Requestor: Weronika Zielinska
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Merge-gate re-review — round 3, head `08c67d17fd4a291a2b4bb6b851b51952fa1b0bb7`

Content commit `476ca146`; `08c67d1` is the trailing `origin/main` merge on top of it. Read via the
contents API at that SHA plus two detached read-only worktrees (head and `origin/main` @ `e08b60a`)
that I created myself — the parent working copy was never switched. CI at head:
`python3 .claude/lib/pr_ci_state.py 1425` → **ready**, 14 checks reported, all green.

Two of my three required items landed, and the one that mattered most — F4 — landed with a
replacement reason that is *true*, which I verified by measurement rather than by reading. The
third, F6, did not land: what got fixed under that label in the commit message is F7, the item I
marked *not required*. That is a genuine mislabel rather than a refusal, and the remaining fix is
one sentence in two places. Details and the exact wording below.

### 0. Prose-only — verified structurally, at both SHAs, across all three files

61 changed lines two rounds running is more than I will accept on eyeballing, and a `main` merge in
the same diff makes eyeballing worse. Same technique as round 2: parse at `b449a75` and at
`08c67d1`, strip every docstring node, round-trip through `ast.unparse`, compare `ast.dump`.

| file | round 2 → round 3 |
|---|---|
| `.claude/hooks/validate_commit_identity.py` | **IDENTICAL** |
| `.claude/hooks/tests/test_validate_commit_identity.py` | **IDENTICAL** |
| `.claude/hooks/tests/test_validate_commit_identity_ast.py` | **IDENTICAL** |

And the merge did not smuggle anything in: `git diff 08c67d1^2 08c67d1` — head against the exact
`main` commit it merged — is **3 files, +516/−2**, the PR's own three. Nothing else in the tree is
attributable to this branch.

### 1. F4 — landed, and the replacement reason is TRUE (tested, not read)

I blocked because a stated mechanism was false, so a merely-different replacement would fail the
same way. Both of its numbers are now measurements I reproduced at the merged head:

| claim in the new docstring | my result |
|---|---|
| sweep is gated on `"commit" in scanned` and on nothing else | confirmed at `:696`; the `_EVAL_RE` loop is inside that block at `:723` |
| widening the bare branch to `[^;&\|\n]+` flips all four (1) shapes AND the (2) shape → **5 of 5** | unmutated: 0/5 blocked. Widened: **5/5 blocked** (`'eval'` reason each) |
| the widening reintroduces **3 of the 6** `EvalDetectionFalsePositiveTests` cases | unmutated: 6/6 pass. Widened: **exactly 3 fail** — `test_commit_message_describing_the_eval_bypass_allowed`, `test_pr_body_describing_the_bypass_allowed`, `test_printf_of_the_shape_into_a_file_allowed` |

The conclusion (do not widen) is unchanged and now rests on a premise that holds. The parallel copy
in `test_eval_not_at_segment_head_is_allowed`'s docstring was corrected too, and I swept the tree
for other copies of the old unreachability claim — none. This is the fix I asked for.

Also worth recording: the same correction was pushed into **#1443** before I got here, with the same
5-of-5 and 3-of-6 measurements, so a future fixer will not inherit my error from the issue either.

### 2. F5 — landed

`validate_commit_identity.py:62` now reads "the **two** mechanisms that reach it and the issues
tracking each", matching `_eval_payloads`'s own two-item enumeration; #1445 is out of that
cross-reference. Nothing in the tree still says "three mechanisms".

### 3. F7 (which I did not require) — landed, and the citation checks out

`:553-557` now cites the corpus behind the 13-of-14 ratio. I fetched
`issues/comments/5272754878` through the API rather than following the link, and it contains the
20-shape oracle, the method (fresh temp repo per case, real `zsh -c`, `git rev-list --count HEAD`
after), the 14/20-commit split and the verbatim "this PR blocks **13**; `main` blocks 2". The
number is now checkable. I accept the review-comment citation because it is the remedy I named, and
I will not re-litigate its durability at this stage — but for the record, the corpus predates three
of the five residual shapes the same docstring enumerates, so 13-of-14 is a *stale* denominator that
a reader can now at least see the edge of. Not a blocker; not worth another round.

### 4. F6 — NOT landed. The claim I required fixed is byte-identical at head

> `validate_commit_identity.py:428` — "Measured over the same 11-command legitimate corpus: 3 are
> blocked in degraded mode that `main` allows"
> `test_validate_commit_identity.py:1336` — "3 of the same 11-command legitimate corpus are blocked
> in degraded mode that `main` allows"

`476ca146` does not touch `test_validate_commit_identity.py` at all, and its
`validate_commit_identity.py` hunks are at `:62` and in `_eval_payloads` — neither reaches `:428`.
The commit message reads "F6 — the 13-of-14 coverage ratio", which is my F7. F6 was the degraded-mode
corpus attribution, and it is the one I marked *required, and the error is originally mine*.

Re-measured against the merged tree, not carried over from round 2. Reconstructing the 11 the only
way the artifact permits — the 6 cases in `EvalDetectionFalsePositiveTests` plus the 5 controls in
`EvalDetectionTests` — head vs `main`, with `_BASHLEX_AVAILABLE` forced `False`:

| case | head normal | head degraded | main degraded |
|---|---|---|---|
| `eval $(direnv hook zsh)` + newline + `git … commit -m x` | allow | **BLOCK** | allow |
| the other 10 of the 11 | allow | allow | allow |
| `eval $(direnv hook zsh); git … commit -m x` *(not in the 11)* | allow | **BLOCK** | allow |
| `eval echo hi; git … commit -m x` *(not in the 11)* | allow | **BLOCK** | allow |

**1 of the 11, not 3.** The 3 is the right count of degraded-mode regressions overall; the
denominator is wrong, and the two commands that make up the difference live in
`EvalDegradedModeOverBlockTests`, not in that corpus. A reader who does the obvious thing — run the
six cases of the class whose docstring makes the claim, in degraded mode — finds one, and concludes
the docstring is wrong. `rg` over the whole tree finds no definition of an 11-command corpus
anywhere, so there is no reading under which the sentence is both reproducible and true.

It also now collides with F4's fix: `:548` says "3 of the 6 cases in
`EvalDetectionFalsePositiveTests`" (widening, bashlex path — true) while `:428` says "3 of the same
11-command legitimate corpus" (degraded mode — false denominator). Two different 3s about the same
test class, one of them wrong.

**Required, one sentence, both sites, no code change** — e.g.:

> 3 legitimate commands are blocked in degraded mode that `main` allows: the eval-line-then-newline
> case in `EvalDetectionFalsePositiveTests` (1 of that class's own 6) and the two pinned by
> `EvalDegradedModeOverBlockTests`. Full probe set in #1445.

Every number in that is reproducible from the two test files alone.

### 5. Did the `main` merge disturb the gate? No — re-run, not assumed

`main` moved substantially under #1424 (`validate_pr_review.py`), #1428 (`trust_signals.py`), #1451
(promotion-audit) and #1419 (`td_intake_gate.py`). None of them touches
`validate_commit_identity.py` or `_shell_parse.py`, and the matrices confirm it rather than assume it:

- **Bypass matrix, merged head vs `origin/main`@`e08b60a`:** head blocks 7 shapes (`eval <unquoted>`,
  quoted, split-quote `com"mit"`, `cd && eval`, `if/then eval`, `eval …; echo done`, `eval` carrying
  identity flags); `main` blocks 1 of those 7. The five residuals (`builtin`/`time`/`exec`/brace
  group/trailing backslash) allow on **both** — pre-existing fail-open, strictly narrowed, not
  widened. Identical to round 2.
- **Degraded matrix:** the table in §4 — unchanged from round 2, including the zero false positives
  on the bashlex path across all 11.
- `.claude/hooks/tests/` + `.claude/lib/tests/` at the merged head: **4464 passed, 1146 subtests, 0
  failed** (46s). The two PR files alone: 167 passed, 39 subtests.

### 6. Deferred residuals

#1443, #1444 and #1445 are all still open, all still reachable exactly as before — nothing in the
merge invalidated or newly reached any of them, and the bypass matrix above is the evidence. My
round-2 defer ruling is unchanged, and #1443 now carries the corrected reachability premise.

### Verdict

F4 and F5 landed and F4's replacement reason is true under measurement — that was the substance of
the block and it is discharged. I am holding on **F6 alone**, and only because it is one of the
three items I named as required and it is unchanged at head, not because I have thought of something
new: a permanent docstring in a security gate stating a number that a reader who runs the very test
class it cites will measure as wrong is the exact defect rounds 2 and 3 existed to remove, and I
supplied this one, which is a reason to be more careful about it rather than less. One sentence, two
sites, wording above; nothing else is open on my side and I will not add anything else.

---
Requestor: Aino Virtanen
Requestee: Nino Kavtaradze
TechDebt: #1443, #1444, #1445

parametrization:
## Re-anchor — round 4, head SHA `35775990fda0429816df9d3b5ed088bea727ef28`

Round 3 approval was against `08c67d17` (`476ca146` + the trailing `origin/main` merge). Round 4's
content commit is `130e97c` (`35775990` is only a fresh `origin/main` merge on top — `e08b60a`,
already accounted for as the round-3 merge base). Reviewed via `gh api .../contents` at both SHAs
plus a detached read-only worktree at `130e97c`; the parent working copy was not touched.

### Context correction acknowledged

Round 3's "F6" was a mislabel upstream of both the author and me — his actual round-2 F6 (the
11-command corpus attribution) was never delivered to the author as F6; the ratio I'd verified was
his F7, marked not required. Noted for the record; doesn't change what I need to re-verify here.

### 1. Scope — content commit touches only the two expected sites

`gh api compare/08c67d17...130e97c` — file list is exactly
`.claude/hooks/validate_commit_identity.py` and `.claude/hooks/tests/test_validate_commit_identity.py`,
+5/-4 each. The wider `08c67d17...35775990` diff also shows `test_validate_pr_review.py`,
`validate_pr_review.py`, `pr_review_state.py`, `test_pr_review_state.py`, `SKILL.md` — all of those
are `origin/main` advancing under the merge (#1424/#1428/#1451/#1419 work), not round-4 content;
confirmed by diffing `08c67d17...130e97c` directly, which excludes them. Round 4 is exactly the one
correction, nothing riding along.

### 2. Prose-only — AST-identical at both sites, both files

Parsed both files at `08c67d17` and `130e97c`, stripped every docstring node (`Module`/`ClassDef`/
`FunctionDef`/`AsyncFunctionDef` leading `Expr(Constant(str))`), `ast.unparse`-equivalent via
`ast.dump(annotate_fields=True, include_attributes=False)` on the stripped trees, compared:

- `validate_commit_identity.py`: **IDENTICAL**
- `test_validate_commit_identity.py`: **IDENTICAL**

The `validate_commit_identity.py` hunk is a `#`-comment change, which never enters the AST at all;
the test-file hunk is inside the `EvalDetectionFalsePositiveTests` class docstring, caught by the
strip. No decision path moved.

### 3. Is the new denominator true? Measured, not read

Ran the real hook against all 6 `EvalDetectionFalsePositiveTests` cases plus the 2
`EvalDegradedModeOverBlockTests` cases, in a worktree at `130e97c`, with `sp._BASHLEX_AVAILABLE`
patched `False` (degraded mode) and normal mode as control:

| case | degraded |
|---|---|
| commit-message-describing-bypass | allow |
| commit-message-dash-c-shape | allow |
| pr-body-describing-bypass | allow |
| issue-comment-describing-bypass | allow |
| printf-of-shape-into-file | allow |
| unquoted-eval-line-then-commit | **BLOCK** |
| `eval $(direnv hook zsh); git … commit -m x` | **BLOCK** |
| `eval echo hi; git … commit -m x` | **BLOCK** |

**1 of the 6** `EvalDetectionFalsePositiveTests` cases regresses (the eval-line-then-newline case),
**2 of 2** `EvalDegradedModeOverBlockTests` cases regress (both allow in normal mode, block in
degraded — confirmed as the control column), total **3**. That's exactly what both sites now claim:
"3 legitimate commands are blocked in degraded mode... the eval-line-then-newline case... (1 of that
class's own 6) and the two pinned by `EvalDegradedModeOverBlockTests`." The new text also drops the
"11-command corpus" framing entirely rather than trying to construct a real 11 — correct, since
`rg` over the tree still finds no 11-command corpus defined anywhere. Every number in the replacement
is reproducible from the two named classes alone, which is the bar F6 needed to clear.

### 4. Is the two-different-3s collision gone?

`validate_commit_identity.py:549` still says "3 of the 6 cases in `EvalDetectionFalsePositiveTests`"
— that's F4's fix, a widening/bashlex-path measurement, unrelated axis, untouched by round 4 (outside
the diff). `:428` and `test_validate_commit_identity.py:1336` now both say "3 legitimate commands are
blocked in degraded mode," sourced as above. Two "3"s remain in the tree, but they're now about
genuinely different measurements that don't contradict each other, not two conflicting denominators
about the same one. `rg -n "11-command"` over both files: zero hits post-fix. Resolved.

### 5. Anything else change? No

`08c67d17...130e97c` is exactly the two five-line hunks quoted above — one sentence, two sites, zero
code change, matching the commit message and matching the gate's supplied replacement sentence
verbatim (with one trailing clause, "The cases in THIS class are otherwise the bashlex-path
measurement," which was already present pre-fix and is untouched content, not new).

### CI

`python3 .claude/lib/pr_ci_state.py 1425` → READY, 14/14. Cross-checked against `statusCheckRollup`
at head `35775990`: 14 entries, zero non-success.

### Verdict

Round-3 approval re-anchors clean to round 4. F6 is now discharged with a true, reproducible
denominator; the collision it created with F4 is resolved; nothing else moved. Approved.

---
Requestor: Weronika Zielinska
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: none

parametrization:
## Merge-gate re-review — round 4, head `35775990fda0429816df9d3b5ed088bea727ef28`

Content commit `130e97c` (parent `08c67d1`); `3577599` is the trailing `origin/main` merge on top
of it. Read at that SHA via the contents API and two detached read-only worktrees I created myself
(head, and `origin/main` @ `e08b60a` — the exact commit this head merges); the parent working copy
was never switched. CI at the head SHA: `python3 .claude/lib/pr_ci_state.py 1425` → **ready**, and
`statusCheckRollup` **at `3577599`** reports **14/14 SUCCESS**, no pending, no empty rollup.

I held on **F6 alone** last round. It landed. Verbatim at both sites, and — the part that actually
mattered — the replacement sentence is **true under re-measurement**, not merely different from the
false one. Approving.

### 1. F6 — landed verbatim at both sites

`130e97c` touches exactly two files, 9 lines each:

| site | at head |
|---|---|
| `validate_commit_identity.py:428` | replaced |
| `test_validate_commit_identity.py:1336` | replaced |

Both now carry the sentence I supplied, unparaphrased. The test-file copy adds one word to its own
tail — "The cases in THIS class are **otherwise** the bashlex-path measurement" — which is a
correction in the right direction, not a paraphrase of the required text: one of that class's six
cases is now explicitly named as a degraded-mode regression, so an unqualified "the cases in this
class are the bashlex-path measurement" would have been the next small falsehood. Good catch on his
part.

The stale trailing clause went too: `:428` no longer ends "Tracked by #1445; unfixed here" while the
new text already cites #1445 as the probe set — it now reads "Unfixed here." No duplicate citation.

### 2. Is the new sentence TRUE? Re-measured, not read

I did not carry round 3's table forward. Fresh run at the merged head and at `origin/main` @
`e08b60a`, importing each worktree's own `validate_commit_identity` (module identity asserted
against the worktree root), `sp._BASHLEX_AVAILABLE` patched `False` for the degraded column:

| case | head normal | head degraded | main degraded |
|---|---|---|---|
| `eval $(direnv hook zsh)` + newline + `git … commit -m x` *(FP class, case 6 of 6)* | allow | **BLOCK** | allow |
| the other 5 of `EvalDetectionFalsePositiveTests` | allow | allow | allow |
| `eval $(direnv hook zsh); git … commit -m x` *(`EvalDegradedModeOverBlockTests`)* | allow | **BLOCK** | allow |
| `eval echo hi; git … commit -m x` *(`EvalDegradedModeOverBlockTests`)* | allow | **BLOCK** | allow |
| 5 reconstructed controls (plain identity commit, `--amend`, `eval $(direnv hook zsh)` alone, `eval "$CMD"`, `eval echo hi`) | allow | allow | allow |

**1 of that class's own 6 + the 2 pinned by `EvalDegradedModeOverBlockTests` = 3.** Exactly what the
new sentence says, and every one of the three is a real regression against `main` (main's degraded
column is all-allow). Over the reconstructible 11 — the 6 FP-class cases plus the 5 controls — the
count is **1**, which is why the old "3 of the same 11-command corpus" was wrong and why the new text
does not use that denominator at all. Every number in the replacement is now derivable from the two
test files by a reader who runs them.

### 3. The collision is resolved, and there is no third copy

The two `3`s are now distinct claims, each with its own stated denominator, and I verified both by
mutation rather than by reading:

| site | claim | measured |
|---|---|---|
| `:548` (F4's text) | widening the bare branch to `[^;&\|\n]+` reintroduces **3 of the 6** cases in `EvalDetectionFalsePositiveTests` | unwidened **0/6** blocked; widened **exactly 3/6** — the eval-prose commit message, the `gh pr create --body`, and the `printf`-to-file case |
| `:428` / `:1336` (F6's text) | **3** legitimate commands regress in degraded mode, **1 of that class's own 6** | table in §2 — 1 + 2 = 3 |

Same mutation also re-confirms the neighbouring 5-of-5: unwidened, all five residual shapes
(`{ eval … }`, `builtin`, `time`, `exec`, trailing backslash) return `None`; widened, **5/5** block.
That claim is unchanged by this round and still holds at the merged head.

**Third copy: none.** `rg --hidden --no-ignore` across the whole tree for `11-command` / `the same 11`
/ `11 command` / `of the 11 ` returns no surviving instance of the degraded-mode denominator (the
only hits are two unrelated files — `feedback_log.md` and a wave-29 test about a different "11").
`rg` for `degraded mode` across all three PR files returns exactly three lines: the two corrected
sites and an unrelated `bashlex unavailable (degraded mode) → warn once` at `:755`.

### 4. Prose-only — verified structurally, round 3 → round 4

Parsed each file at `08c67d1` and at `3577599`, stripped every docstring node, round-tripped through
`ast.unparse`, compared `ast.dump`:

| file | r3 → r4 |
|---|---|
| `.claude/hooks/validate_commit_identity.py` | **IDENTICAL** |
| `.claude/hooks/tests/test_validate_commit_identity.py` | **IDENTICAL** |
| `.claude/hooks/tests/test_validate_commit_identity_ast.py` | **IDENTICAL** |

And nothing was smuggled in by the merge: `git diff 3577599^2 3577599` — head against the exact
`main` commit it merges — is **3 files, +518/−2**, this PR's own three.

One correction to my own round-3 note for the record: I named `origin/main` @ `e08b60a` there, but
`08c67d1`'s merge parent was `e96b8ba`; `e08b60a` is *this* round's parent, and the r4 merge advanced
`main` by four commits (#1424 — `validate_pr_review.py`, `pr_review_state.py`, `review-pr/SKILL.md`).
None of them touches `validate_commit_identity.py` or `_shell_parse.py`, and the §2 table was measured
against the actual new parent rather than reused. Full suite at the merged head: **4481 passed, 1146
subtests, 0 failed** (44s) — up from 4464 by #1424's own new tests, none failing. The two PR test
files alone: **167 passed, 39 subtests**. `ruff format --check` and `ruff check` clean on both touched
files.

### 5. Nothing new added

I said round 4 would be mechanical and it was. I have not re-opened the deferred residuals, and I
found nothing at this head I would file — the tree sweep, the mutation matrix and the full suite all
came back where round 3 left them. #1443, #1444 and #1445 remain open, remain fail-open on `main`
too, and remain correctly deferred; #1443 and #1444 carry the corrected reachability premise from
round 3, so a future fixer will not inherit my error from the issue text either.

### Verdict

F6 landed verbatim at both sites, the sentence is true under fresh measurement, the two `3`s no
longer collide, no third copy of the bad denominator exists anywhere in the tree, the change is
provably prose-only, and CI is 14/14 green at the head SHA. **Approved.**

Worth saying plainly at the close of four rounds: every one of my findings on this PR was about what
the text CLAIMED, never about the engineering, and the engineering was right from round 1 — the
execution oracle, the decision to measure the over-block instead of assuming it, and the choice to
pin non-coverage as a decision rather than leave it as a gap. Two of the six items I raised were
errors I had introduced myself, and both were found by *you* measuring instead of transcribing.
That is the behaviour I want from an implementer under a reviewer who is wrong.

---
Requestor: Aino Virtanen
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: #1443, #1444, #1445

PR #1426, opened 2026-08-11
fix(#1362): add an RE2-syntax guard for _CHANGES_REQUESTED_RE alongside the jq semantics test

### LINKED ISSUE #1362: tech-debt(tests): the jq regex test validates against Oniguruma, but production runs the filter through gh's gojq (RE2) — a lookahead would pass CI and hard-error live

parametrization:
Closes #1362.

## Problem

`test_changes_requested_regex_variants_via_real_jq` (Counters, `test_wave_status.py`) validates `_CHANGES_REQUESTED_RE` against the system `jq` binary — Oniguruma. Production evaluates the same pattern through `gh --jq`, which embeds **gojq**, compiling regexes through Go's `regexp` package — **RE2**. RE2 has no lookaround or backreferences. A future widening of the pattern with a lookahead would pass the existing test green and hard-error live. No behavioural defect today (gojq vs Oniguruma agree 51/51 on the live wave-29 corpus per the issue) — this is about what tomorrow's test run can and cannot catch.

## Remedy chosen — cheap RE2-syntax guard, not a gojq/Go dependency

The issue offers two remedies: run the assertion through gojq in CI+pre-commit, or keep the jq semantics test and add a syntax guard for RE2-incompatible constructs. I chose the **syntax guard**:

- gojq (and a Go toolchain to build/run it) is not otherwise a dependency of this repo. Installing it would need to be mirrored into `.pre-commit-config.yaml` per the org's full local/CI parity rule (`ci-gates.md` § Full Local⇄CI Tooling Parity), adding a new external binary requirement everywhere this repo is worked on.
- The actual risk surface is narrow and well-known (lookaround, backreferences, atomic groups, conditionals — the constructs RE2's parser rejects outright), so a Python `re`-on-the-pattern scan catches the demonstrated failure shape for free, using only the standard library.
- **No new CI or pre-commit dependency was added.** `_re2_incompatible_reason` (`.claude/lib/wave_status.py`) is pure-stdlib `re`.

Added `_re2_incompatible_reason(pattern) -> str | None` plus three tests in the `Counters` class (`test_wave_status.py`):
- `test_re2_guard_rejects_lookahead` — the exact construct from the issue's repro.
- `test_re2_guard_rejects_lookbehind_and_backreference`
- `test_re2_guard_accepts_current_changes_requested_pattern` — regression guard against the real `_CHANGES_REQUESTED_RE` constant.

I cross-checked the guard's judgments against the real RE2 engine (Go's `regexp` package, available on this box) rather than trusting the syntax scan blind:
```
$ go run re2check.go 'RequestOrReplied:\\s*Changes(\\s*Requested)?\\b'
OK
$ go run re2check.go 'RequestOrReplied:\s*(?=Changes)Changes'
REJECT: error parsing regexp: invalid or unsupported Perl syntax: `(?=`
$ go run re2check.go '(?<=Foo)Bar'
REJECT: error parsing regexp: invalid named capture: `(?<=Foo)Bar`
$ go run re2check.go '(Foo)\1'
REJECT: error parsing regexp: invalid escape sequence: `\1`
```
Matches exactly: current production pattern compiles clean under real RE2; the demonstrated lookahead and the added lookbehind/backreference cases all reject with the same error class `gh --jq` produces live.

## Failing-before-fix (acceptance bar)

Ran the three new tests against the pre-fix tree (function doesn't exist yet):
```
$ ENVIRONMENT=test python3 -m pytest .claude/lib/tests/test_wave_status.py -k "re2_guard" -q
...
E       AttributeError: module 'wave_status' has no attribute '_re2_incompatible_reason'
FAILED .claude/lib/tests/test_wave_status.py::Counters::test_re2_guard_accepts_current_changes_requested_pattern
FAILED .claude/lib/tests/test_wave_status.py::Counters::test_re2_guard_rejects_lookahead
FAILED .claude/lib/tests/test_wave_status.py::Counters::test_re2_guard_rejects_lookbehind_and_backreference
3 failed, 45 deselected in 0.39s
```
After the fix, full `test_wave_status.py` suite: `48 passed, 7 subtests passed`.

## Acceptance checklist (from #1362)

- [x] The regex is validated against the engine production actually uses, or a guard rejects syntax RE2 cannot compile — the syntax guard, cross-checked against real RE2 above.
- [x] The test comment no longer claims Oniguruma is the production engine — already correct as of `f1b3ef0` (a prior #1356-review fix); verified unchanged/accurate, not re-touched here.
- [x] The `feedback_log.md` sentence is already corrected (same prior commit `f1b3ef0`) — verified, not re-touched here.
- [x] `@unittest.skipUnless(shutil.which("jq"), ...)` reviewed: the semantics test's skip is inherent (it needs the real Oniguruma binary), so it stays skippable. The new `_re2_incompatible_reason` guard tests are **unconditional** — no subprocess, no `jq` dependency — so the demonstrated failure shape (a lookahead widening) is still caught even in an environment where `jq` is absent and the semantics test silently skips.

## Not fixed / found but out of scope

- Did not touch `_canonical_issue_numbers_by_repo` or the `CanonicalIssueNumbers` test class — those are owned by the in-flight #1421 (Weronika Zielinska), confirmed non-overlapping with this diff.
- Did not restructure or reformat unrelated regions of either file.

## CI dependency

None added. No `.pre-commit-config.yaml` change needed — the guard is pure Python `re` (stdlib), already covered by the existing `pytest`/`ruff` mirrors.

## Gates run locally (fresh worktree, hooks installed this session)

- `pre-commit run` (commit stage: ruff-format, ruff-lint, checksums-ascii) — pass.
- `mypy .claude/hooks/ .claude/lib/ --ignore-missing-imports --no-error-summary` — clean.
- `ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/ .claude/lib/tests/ -q` — 4396 passed, 1037 subtests passed.
- `python3 .claude/lib/pre_commit_ci_sync.py .` — OK, no drift.
- `git push` (pre-push stage: mypy, mypy-skills, pytest, pytest-skills, memory-budget, headcount-budget, doc-freshness, office-drift, mermaid-render) — all pass.

## Reviewers

- Peer reviewer: **Weronika Zielinska**
- Merge gate: **Nino Kavtaradze**

Co-Authored-By: Lucas Ferreira <parametrization+Lucas.Ferreira@gmail.com>

### DIFF
--- a/.claude/lib/tests/test_wave_status.py
+++ b/.claude/lib/tests/test_wave_status.py
@@ -337,6 +337,35 @@ class Counters(unittest.TestCase):
                 )
         self.assertEqual(rc, 0)
 
+    # -- main#1362: production runs `_CHANGES_REQUESTED_RE` through `gh --jq`
+    # (gojq -> Go's `regexp`, i.e. RE2), not the system `jq` binary the
+    # semantics test above shells out to (Oniguruma). RE2 has no lookaround
+    # or backreferences; Oniguruma has both, so a future widening of the
+    # constant with e.g. a lookahead would compile fine under the test above
+    # and hard-error live. `_re2_incompatible_reason` is a cheap syntax scan
+    # (not a full RE2 parse, and not a new gojq/Go-toolchain dependency in
+    # CI/pre-commit) for the specific constructs RE2's parser rejects
+    # outright. These three tests demonstrate it catches the exact failure
+    # shape #1362 documented and stays quiet on the real pattern.
+    def test_re2_guard_rejects_lookahead(self) -> None:
+        # The exact construct from #1362's repro: compiles under jq
+        # (Oniguruma) but `gh --jq` (gojq/RE2) hard-errors with "invalid or
+        # unsupported Perl syntax: `(?='".
+        reason = wave_status._re2_incompatible_reason(r"RequestOrReplied:\s*(?=Changes)Changes")
+        self.assertIsNotNone(reason)
+
+    def test_re2_guard_rejects_lookbehind_and_backreference(self) -> None:
+        self.assertIsNotNone(wave_status._re2_incompatible_reason(r"(?<=Foo)Bar"))
+        self.assertIsNotNone(wave_status._re2_incompatible_reason(r"(?<!Foo)Bar"))
+        self.assertIsNotNone(wave_status._re2_incompatible_reason(r"(Foo)\1"))
+
+    def test_re2_guard_accepts_current_changes_requested_pattern(self) -> None:
+        # Regression guard: if `_CHANGES_REQUESTED_RE` is ever widened to
+        # include one of the forbidden constructs, this fails fast in CI
+        # instead of passing green and hard-erroring in the live gh --jq
+        # call the way an unguarded lookahead would.
+        self.assertIsNone(wave_status._re2_incompatible_reason(wave_status._CHANGES_REQUESTED_RE))
+
 
 class Write(unittest.TestCase):
     def test_write_upserts_three_canonical_keys(self) -> None:
--- a/.claude/lib/wave_status.py
+++ b/.claude/lib/wave_status.py
@@ -40,6 +40,7 @@ from __future__ import annotations
 import argparse
 import json
 import math
+import re
 import subprocess
 import sys
 from collections import Counter
@@ -95,6 +96,54 @@ _DEFAULT_STATUS = wave_state.DEFAULT_STATUS_PATH
 # wave-29 merged-PR set, where it now reproduces the retro's corrected 51
 # (was 49).
 _CHANGES_REQUESTED_RE = "RequestOrReplied:\\\\s*Changes(\\\\s*Requested)?\\\\b"
+
+
+# main#1362: `_changes_requested_cycles` below evaluates `_CHANGES_REQUESTED_RE`
+# through `gh --jq`, and gh embeds gojq (`itchyny/gojq`), which compiles
+# regexes through Go's `regexp` package — i.e. RE2. RE2 has no lookaround and
+# no backreferences. `test_changes_requested_regex_variants_via_real_jq`
+# (test_wave_status.py, Counters) validates this constant's *semantics*
+# against the system `jq` binary (Oniguruma), which accepts a strictly wider
+# syntax than RE2 — it demonstrated a lookahead compiles under jq but hard-
+# errors under `gh --jq`: "invalid or unsupported Perl syntax: `(?=`". So a
+# future widening of `_CHANGES_REQUESTED_RE` expressed with a lookahead would
+# pass that test green and hard-error live.
+#
+# `_re2_incompatible_reason` closes that gap as a syntax scan — not a full
+# RE2 parse — for the specific constructs RE2's parser rejects outright. This
+# was chosen over installing gojq (or a Go toolchain) in CI/pre-commit: gojq
+# is not otherwise a dependency of this repo, and mirroring it into
+# `.pre-commit-config.yaml` per the org's full local/CI parity rule would add
+# a new external binary everywhere this repo is worked on, for a check this
+# scan gets for free out of the standard library. It does not claim
+# RE2-completeness for constructs outside this known list — see the
+# `Counters` class's `test_re2_guard_*` tests in test_wave_status.py, which
+# exercise both the reject and the accept path against the real constant
+# above.
+_RE2_INCOMPATIBLE_CONSTRUCTS: tuple[tuple[re.Pattern[str], str], ...] = (
+    (re.compile(r"\(\?<[=!]"), "lookbehind (?<= or (?<!"),
+    (re.compile(r"\(\?[=!]"), "lookahead (?= or (?!"),
+    (re.compile(r"\(\?P="), "named backreference (?P="),
+    (re.compile(r"\(\?>"), "atomic group (?>"),
+    (re.compile(r"\(\?\("), "conditional (?("),
+    (re.compile(r"\\[1-9]"), "backreference \\N"),
+)
+
+
+def _re2_incompatible_reason(pattern: str) -> str | None:
+    """Return the name of the first known RE2-incompatible construct found in
+    ``pattern``, or ``None`` if the scan finds none of them.
+
+    This is a syntax guard for the failure shape main#1362 documented (a
+    lookahead compiles under jq/Oniguruma but gojq/RE2 — the engine
+    ``gh --jq`` actually uses in production — hard-errors on it). It is
+    intentionally a regex-on-the-regex scan, not an actual RE2 compile, so it
+    adds no new dependency (no gojq, no Go toolchain) to CI or pre-commit.
+    """
+    for construct_re, name in _RE2_INCOMPATIBLE_CONSTRUCTS:
+        if construct_re.search(pattern):
+            return name
+    return None
 
 
 # The shared gh shim and status reader (main#1119). These stay module-level


parametrization:
Reviewed at head `5f2bb02e21aa39b2d44e8c6a657c988ffae5698c`. Both changed files read in full at that SHA (`.claude/lib/wave_status.py` +49, `.claude/lib/tests/test_wave_status.py` +29 — matches the PR's stated diff, no other files touched).

**Does the guard measure RE2 or approximate it?** It approximates it — `_re2_incompatible_reason` (`wave_status.py:133-146`) is a hand-maintained denylist of 6 regex-on-the-regex patterns (`wave_status.py:123-130`: lookbehind, lookahead, named backreference, atomic group, conditional, numbered backreference), not an actual RE2 compile. The PR is upfront about this in both the module comment (`wave_status.py:118-119`, "does not claim RE2-completeness") and the function docstring (`wave_status.py:140-141`) — this is a disclosed tradeoff, not a hidden one, and the tradeoff itself is reasonable (avoids a gojq/Go-toolchain dependency for a repo that has neither). I verified the gap is real, not hypothetical: built a one-line Go `regexp.Compile` checker against the real RE2 engine (Go 1.22 is on this box) and found 4 construct families RE2 rejects live that the denylist misses — `\g<1>` backreferences, possessive quantifiers (`a++`, `a*+`), `(?#comment)` groups, and `\K`. Confirmed directly against the shipped function (`wave_status._re2_incompatible_reason` returns `None` — "looks safe" — for all five probe patterns above), not inferred. None of these appear in the current `_CHANGES_REQUESTED_RE`, so there's no live defect, but the denylist-vs-real-engine drift the PR itself warns against (for the *pattern*) also applies to the *guard*. Filed #1434 to track this as a durable, non-blocking record — it also cross-references noorinalabs-main#1431 (already open, unrelated author), which if it lands would make this whole guard moot by rewriting the counter in Python against the shared `charter_trailer` predicate.

**Is the guard pinned by a failing test?** Yes, confirmed by injection, not by reading the test names. In a scratch worktree I widened `_CHANGES_REQUESTED_RE` to add `(?=Changes)` (the exact construct #1362's issue reproduced) and reran `test_re2_guard_accepts_current_changes_requested_pattern`: it failed with `AssertionError: 'lookahead (?= or (?!' is not None` — the guard would have caught the failure shape #1362 documents. Reverted via `git checkout`.

**Does the original Oniguruma test survive?** Yes — `git diff main...HEAD -- .claude/lib/tests/test_wave_status.py` is a pure addition (three new test methods appended after the existing suite); `test_changes_requested_regex_variants_via_real_jq` is byte-for-byte unchanged. Ran it directly: passes (`jq` is present on this box).

**Scope — is this the only pattern reaching gojq, or one guarded among many?** `_CHANGES_REQUESTED_RE` (`wave_status.py:98`, fed to `gh --jq` at `wave_status.py:570` via `.body | test(...)`) is the only *named, reusable regex constant* in `.claude/` that flows through `gh --jq`'s `test()` — confirmed via `rg --hidden -n '\| test\(' .claude/lib .claude/hooks .claude/skills`. One other `gh --jq ... test(...)` call site exists, `.claude/skills/wave-wrapup/SKILL.md:216` (`test("retro\\(P{P}W{M}")`), but it's a literal, non-evolving pattern with no lookaround/backreference risk and isn't in #1362's stated scope. So: fully scoped to the one pattern the issue names, and that's the right call — not a partial fix worth flagging as incomplete.

Full local suite (`.claude/lib/tests/test_wave_status.py`, all 48 tests + 7 subtests), `ruff check`, and `mypy` all pass in a fresh worktree at head. CI: 14/14 checks green (`python3 .claude/lib/pr_ci_state.py 1426` → READY; `statusCheckRollup` confirms against the head SHA above, not the PR number).

Solid, honest fix — it closes the exact gap #1362 named, is pinned by a test that actually fails without it, doesn't touch the surviving semantics test, and is honest in its own comments about what it doesn't cover. Approving; the denylist-completeness gap is real but non-blocking (no live defect, disclosed limitation, tracked separately).

---
Requestor: Weronika Zielinska
Requestee: Lucas Ferreira
RequestOrReplied: Approved
TechDebt: #1434

parametrization:
Merge-gate review. Reviewed at head SHA `5f2bb02e21aa39b2d44e8c6a657c988ffae5698c`, which I recorded myself from `gh pr view` and read through a detached worktree pinned to that SHA. CI at that exact SHA — `gh api repos/noorinalabs/noorinalabs-main/commits/5f2bb02/check-runs` — reports `total_count=14`, all fourteen `success`. Not an empty rollup, so CI is genuinely green rather than merely un-red.

This ships. The guard is real, it is falsifiable, and it is genuinely non-redundant with the jq semantics test it sits beside — I verified all three by mutating the tree, not by reading it. What it is not is complete, and the PR body does not say how incomplete. I measured that and filed #1441 rather than blocking; the deciding question is named in section 6.

## 1. Measurement vs approximation — approximation, ruled acceptable

`_re2_incompatible_reason` (`.claude/lib/wave_status.py:133`) is a denylist: six `re`-on-the-regex rules in `_RE2_INCOMPATIBLE_CONSTRUCTS` (`.claude/lib/wave_status.py:123-130`). It never compiles the pattern under RE2. So it is a re-implementation of an engine's accept/reject boundary — structurally the same shape as the defect #1362 describes.

I did not take the PR body's `go run re2check.go` transcript on faith. I built my own oracle and ran a 30-construct corpus through three engines: real RE2 (`regexp.Compile`, Go 1.22.2), real Oniguruma (the system `jq` binary), and the guard.

The right denominator is not "constructs RE2 rejects" — it is **constructs Oniguruma accepts AND RE2 rejects**, because anything Oniguruma also rejects already fails the existing semantics test loudly. Of my 30-construct corpus, **16 fall in that dangerous set, and the guard catches 6**.

| dangerous construct | guard | live `gh --jq` result |
|---|---|---|
| `(?=` `(?!` `(?<=` `(?<!` lookaround | caught | exit 1, unsupported Perl syntax |
| `(?>` atomic group | caught | exit 1, unsupported Perl syntax |
| `\N` backreference | caught | exit 1, invalid escape sequence |
| `++` `*+` `?+` possessive | **missed** | exit 1, invalid nested repetition operator |
| `(?#comment)` | **missed** | exit 1, unsupported Perl syntax |
| `\K` | **missed** | exit 1, invalid escape sequence |
| `\h` `\H` | **missed** | exit 1, invalid escape sequence |
| `\Z` | **missed** | exit 1, invalid escape sequence |
| `\R` | **missed** | exit 1, invalid escape sequence |

Every "missed" row is confirmed against the production path, not inferred: `gh api /rate_limit --jq '"x" | test("Changes\\K Requested")'` exits 1 with `invalid regular expression ... invalid escape sequence: '\K'`.

Two of the six denylist rows — `(?P=` and `(?(` — fire only on constructs Oniguruma *also* rejects, so they are belt-and-braces rather than coverage. Effective coverage is four rules against six construct families.

What keeps the approximation honest today: the code comment at `.claude/lib/wave_status.py:113-122` explicitly disclaims RE2-completeness ("It does not claim RE2-completeness for constructs outside this known list"). That is the right disclosure and is why I read this as a bounded approximation rather than a misrepresentation. The PR body is weaker than the code comment on this point — it presents the four-input cross-check against real RE2 as validating the guard, when what it validated was four hand-picked inputs, three of which are the same lookaround family.

## 2. Falsifiable — yes, verified by injection

I injected the issue's own lookahead into the real constant at `.claude/lib/wave_status.py:98` and re-ran. `test_re2_guard_accepts_current_changes_requested_pattern` fails with `AssertionError: 'lookahead (?= or (?!' is not None` (`test_wave_status.py:367`). The guard is not inert.

## 3. Alongside, not instead — yes, and orthogonal in both directions

Both tests survive at head, and I measured that neither subsumes the other by mutating in each direction and running the pair together:

| mutation | Oniguruma semantics test | RE2 guard |
|---|---|---|
| lookahead `(?=Changes)` injected into the constant | **passes green** | **fails** |
| revert the constant to the narrow one-word form (#1357 defect) | **fails**, 2 subtests | **passes green** |

The first row is #1362 reproduced live: the pre-existing test goes green on a pattern that hard-errors in production. The second confirms the guard is syntax-only and cannot stand in for the semantics test. Each catches exactly what the other structurally cannot.

## 4. Coverage denominator — 1 of 2 in executable code, 1 of 3 counting the skill

`rg --hidden -uu` over tracked files for a `--jq` filter containing a jq regex function (`test` / `capture` / `match` / `scan` / `splits` / `gsub`), excluding tests, memory and archive:

1. `.claude/lib/wave_status.py:618` — the `test(...)` filter this PR guards. **Guarded.**
2. `.claude/hooks/validate_pr_review.py:2905-2908` — a `capture(...)` filter using a `(?<r>` named group. **Unguarded.** It lives inside the hook's printed block-reason help text, so it is a copy-paste diagnostic rather than something the hook executes — but it reaches gojq when an agent runs it.
3. `.claude/skills/wave-wrapup/SKILL.md:216` — `test("retro\\(P{P}W{M}")`. **Unguarded**; I verified it is RE2-clean today.

So: **1 of 2 executable sites, 1 of 3 counting the skill command an agent literally runs.** This is a partial fix and should be named as one — the PR body's "no behavioural defect today" framing is accurate for the guarded constant but is silent on the other two sites.

`_re2_incompatible_reason` is module-private to `wave_status`, so neither of the other sites can reuse it without importing from `.claude/lib`. That placement is the concrete reason coverage stops at one.

Separately, site 2 carries a latent portability hazard of exactly the class this PR exists to prevent: Go's `regexp` only accepted `(?<name>...)` from **Go 1.22**. It works on this box because the installed `gh` is built with go1.22.2 (`go version -m /usr/bin/gh`), but upstream `gh` 2.45.0 shipped Feb 2024 against Go 1.21, where that filter rejects with `invalid named capture`. My own live check of that site is therefore only valid for this binary — I am stating the limit rather than generalising it. `(?P<r>` is accepted by both.

## 5. Live-error direction — exit 1, loud, and nothing is written

I tested the entry point as a subprocess with the lookahead injected, against real gh:

```
$ python3 .claude/lib/wave_status.py counters 5 wave-30 --status cross-repo-status.json
ERROR: gh call failed (exit 1): gh api repos/noorinalabs/noorinalabs-main/issues/1394/comments --jq ...
invalid regular expression "...": error parsing regexp: invalid or unsupported Perl syntax: `(?=`
CLI exit=1
```

`main()` catches `subprocess.CalledProcessError` at `.claude/lib/wave_status.py:862`, prints gh's own stderr, and returns 1. The exception is raised inside `compute_counters`, so under `--write` `_write_counters` is never reached and no counter is persisted. This is not a hook, so the PreToolUse exit-code-inversion trap does not apply here; the surface is a skill invoking a CLI, and the failure is a self-describing hard stop rather than a silent zero. That fact is load-bearing for section 6.

## 6. Findings, with the fold-in vs defer call

**F1 — `.claude/lib/wave_status.py:123-130`: the denylist misses 10 of 16 measured dangerous constructs, and 2 of its 6 rows are redundant with the existing semantics test.** Deferred to #1441. Deciding question: **which direction does it fail.** A missed construct fails open at guard time, but section 5 shows the downstream consequence is a loud exit-1 crash at wrapup with gojq's own error text printed — i.e. it degrades to the pre-#1426 status quo, not to a corrupted measurement. Prevalence seconds it: `\K`, `\h`, `\R`, `\Z` and possessive quantifiers have essentially no plausibility inside a verdict-vocabulary regex, whereas lookaround — the family that *is* covered — is the natural way to express the next widening. The remedy is four additive denylist rows and breaks nothing, so it does not need to gate this merge.

**F2 — `.claude/hooks/validate_pr_review.py:2905-2908` is an unguarded gojq-bound pattern, and its `(?<r>` named group requires gh built with Go 1.22+.** Deferred to #1441 (item 3). Deciding question: **could a working artifact contain this shape.** Yes — it is a printed diagnostic string, out of this PR's stated scope, and the fix (`(?P<r>`) belongs with whoever relocates the guard to a shared home so site 2 can actually call it.

**F3 — minor, no action.** `_re2_incompatible_reason` is called on `_CHANGES_REQUESTED_RE` in its Python-literal, doubly-escaped form, not the form gojq receives after jq's string-literal unescaping. It happens to be harmless: the paren-based rules are escaping-insensitive, and the `\N` rule still matches inside a doubled backslash. But the guard's level is one removed from the engine's, which is precisely the escaping distinction #1357 turned on. Worth a sentence in the docstring if anyone touches this next.

Nothing here is a blocker. The acceptance bar from #1362 is met, the failing-before-fix evidence in the body reproduces, and the local gate claims check out (48 passed, 7 subtests on `test_wave_status.py` at head).

---
Requestor: Nino Kavtaradze
Requestee: Lucas Ferreira
RequestOrReplied: Approved
TechDebt: #1441

PR #1213, opened 2026-08-02
fix(session-start): classify squash/rebase/cherry-picked worktrees as merged

parametrization:
## Summary

`/session-start` Step 0 classified a worktree as merged with `git merge-base --is-ancestor "$head" origin/main` alone. That test is correct only for a plain merge commit — squash, rebase-merge, and cherry-pick all rewrite the commit(s), so the branch tip is never made an ancestor of `origin/main`, and the branch is flagged `UNMERGED` forever no matter how completely its content landed. Because Step 0 correctly refuses to auto-remove a FLAGGED worktree, each false flag recurs every session indefinitely (0% precision on the 5 real instances in #1212).

## What changed

- **New `.claude/lib/check_worktree_merged.py`** (mirrors the `check_child_checkouts.py` / `sync_main.py` pattern): ancestry stays the cheap fast path; a **patch-id equivalence** fallback covers everything ancestry misses:
  1. **Aggregate match** — the branch's whole combined diff since its merge-base (`git diff merge_base..head`) is compared by patch-id against every commit `origin/main` has gained since that point (`git log -p --first-parent merge_base..origin/main | git patch-id --stable`). This is what catches a **multi-commit squash**, where no single original commit's patch-id matches the one new squash commit.
  2. **Per-commit match** (`git cherry origin/main head`) — catches rebase-merge replay (multiple original commits individually replayed with new hashes) and a plain cherry-pick.
  - If either test finds a match → `merged`. Otherwise → `unmerged` (flag, never auto-remove).

- **Why history search, not a current-tree snapshot compare:** an earlier version of this fix compared `head`'s files, byte for byte, against their *current* content on `origin/main`. That decays — real unrelated commits keep touching the same files after a squash lands, and within days the "same file" comparison starts failing even though the branch's content is still fully present in `origin/main`'s history. I validated this directly: with the snapshot-compare version, **all 5 of the real fixtures cited in #1212 still failed** (e.g. `.claude/hooks/_shell_parse.py` had 600+ further lines changed by unrelated commits within days of the squash landing). The patch-id-against-history approach doesn't have this problem — a patch-id is stable regardless of what happens to the file afterwards.

- **Wired into session-start Step 0** via a `$(...)` command substitution (not `<(...)` process substitution), preserving #839's static-analyzability property. Degrades to the legacy ancestry-only test if the helper file is missing, and to `unmerged` on any internal git-command failure (never guesses `merged`). Locked-worktree handling is untouched.

[...]

PR #1427, opened 2026-08-11
fix(session-start): a fresh worktree is merged but is not a removal candidate (#1341)

### LINKED ISSUE #1341: tech-debt(session-start): a freshly-branched worktree classifies as MERGED and is auto-removed under an active agent
## Summary

`/session-start` Step 0 auto-removes any worktree whose HEAD is an ancestor of `origin/main`. A **freshly-created worktree that has not committed yet** has `HEAD == origin/main`, which is trivially an ancestor — so it classifies as *merged* and is removed, even though it belongs to an agent actively working in it.

Observed during wave-29 delivery: an active worktree for in-flight issue #1117, branched minutes earlier off `e39e38e`, reported `MERGED` under the same predicate the skill uses. Its four modified files were the agent's uncommitted work.

## Both code paths have it

- **Helper fast path** — `.claude/lib/check_worktree_merged.py::classify_merged` returns `MergeClassification("merged", "ancestor", ...)` immediately on `merge-base --is-ancestor` succeeding (~L368-370).
- **Legacy fallback** — `SKILL.md` L125 runs the bare `git merge-base --is-ancestor` when the helper is absent.

So the richer patch-id classification added later never gets a chance to disagree: the ancestor fast path fires first, and for a fresh branch it is *correct about ancestry* and *wrong about intent*.

## Why the blast radius is limited (but non-zero)

Removal is deliberately never forced (owner decision, PR #1213 round 3), and `git worktree remove` refuses on uncommitted or untracked content. So an agent that has already edited files is protected — the removal fails and the worktree is FLAGGED instead.

The exposed case is a **clean** fresh worktree: created, not yet written to. That gets removed silently, and the agent fails on its next write with a path that no longer exists — an error whose cause is in a different session.

Exposure is real because this org runs concurrent sessions: `/session-start` in session B iterates worktrees created by session A's agents.

## Suggested fix

Distinguish *"contains no unmerged work"* from *"is not being used"*. Cheapest discriminator that doesn't need new state: treat `HEAD == <remote_ref>` (zero commits ahead) as **not** a removal candidate — a worktree with no commits of its own has either done nothing yet or is scratch, and neither is worth reclaiming automatically. That is a one-line guard ahead of the ancestor fast path and it preserves every genuine case (a merged feature branch always has commits that are ancestors *and* differs from the remote tip).

Optionally also skip a worktree whose directory mtime is within N minutes, though the HEAD-equality test alone covers the observed case.

## Verify

- Fresh worktree off `origin/main`, no commits → must classify as *not a removal candidate*.
- Merged feature branch with commits → must still classify as *merged* (no regression to the behaviour that motivated the sweep, #526).
- Unmerged branch → unchanged.

Found while cleaning worktrees after the #1333/#1337 merges.

parametrization:
Closes #1341.

## The bug, in one sentence

`/session-start` Step 0 read **merged-ness as a removal decision**. A worktree that has not committed yet has `HEAD == origin/main`, which is trivially an ancestor, so the fast path answered `merged` — *correctly*, a branch with no commits of its own has no unlanded content — and Step 0 removed a clean, live worktree out from under an agent working in another session (observed on in-flight #1117 during wave-29).

## Failing before the fix

Live, against the unmodified helper, on this very worktree's pre-branch HEAD (clean, uncommitted, actively in use):

```
$ python3 .claude/lib/check_worktree_merged.py . 40cfe9ff4a328c757b7e06a36a27686af10d4203 origin/main
merged (ancestor): 40cfe9ff4a328c757b7e06a36a27686af10d4203 is an ancestor of origin/main
exit=0            # 0 => Step 0 runs `git worktree remove`
```

And the same for Step 0's helper-missing fallback path:

```
$ git merge-base --is-ancestor 40cfe9ff4a328c757b7e06a36a27686af10d4203 origin/main; echo $?
0                 # 0 => Step 0 runs `git worktree remove`
```

The new test, run against the **pre-fix** implementation:

```
$ ENVIRONMENT=test python3 -m pytest .claude/lib/tests/test_check_worktree_merged.py -q -k cli_exit_code_fresh
____ CheckWorktreeMergedTest.test_cli_exit_code_fresh_worktree_is_not_zero _____

        rc = cli_main([str(self.repo), fresh_head, "main"])
>       self.assertNotEqual(rc, 0)
E       AssertionError: 0 == 0

----------------------------- Captured stdout call -----------------------------
merged (ancestor): 3fecef4cc90b9b5e5483ab8f2efac6068d6f4439 is an ancestor of main
1 failed, 38 deselected in 0.29s
```

After the fix, the same live command:

```
$ python3 .claude/lib/check_worktree_merged.py . 40cfe9ff4a328c757b7e06a36a27686af10d4203 origin/main
merged (ancestor) [not-a-removal-candidate: no-own-commits]: 40cfe9ff4a32... is an ancestor of
origin/main; 40cfe9ff4a32 is itself a commit on origin/main's first-parent history, so this branch
has no commits of its own — a fresh/unstarted worktree with nothing to reclaim (very likely in
active use), not a removal candidate
exit=3
```

## How the PR #1213 guarantee is preserved (not traded)

The guarantee is `merged` **iff** the branch's cumulative diff since its merge-base is fully present on `remote_ref`, order-independently, with the `--cc` merge check and the disclosed patch-id whitespace residual. A fresh worktree **satisfies** that guarantee — it has no unlanded content — so reclassifying it `unmerged` would have *broken* the guarantee to fix the symptom.

Nothing in the classification changed. Not one line of the net-content test, the `git cherry` corroboration, the evil-merge `--cc` check, or their degrade paths was touched; a fresh worktree still classifies `merged`/`ancestor`, and `test_fresh_worktree_at_remote_tip_is_not_a_removal_candidate` asserts exactly that.

What is added is a **second, orthogonal predicate** that gates only the *removal* decision:

- `MergeClassification.removal_block` — `""`, `"no-own-commits"`, or `"mainline-check-failed"`.
- `MergeClassification.safe_to_remove` = `merged and not removal_block` — the new auto-remove predicate. `.merged` keeps its old meaning exactly.
- CLI exit codes: `0` merged **and** a reclaim candidate · `1` not merged · `2` usage · **`3` merged but not a reclaim candidate**. Any caller that only knows the old `0 == remove` rule still fails safe on 3.

The two axes stay independent under failure too: a probe failure blocks removal but leaves the classification intact, because a failure on one axis is not evidence about the other.

## Why first-parent membership, not `HEAD == origin/main`

The issue proposed the equality guard. It only catches a worktree branched off the **current** tip; sessions here run long enough that `origin/main` advances under an idle worktree, after which HEAD is an ancestor but no longer equal — same zero-work branch, same live agent, and equality misses it. `test_fresh_worktree_off_older_remote_tip_is_not_a_removal_candidate` pins that case.

Mainline membership does not touch #526's population: a squash- or rebase-merged tip is not an ancestor at all (never reaches the guard), and a merge-commit-merged tip is reachable only through the merge's **second** parent, so it is an ancestor but not on the first-parent chain — it stays a removal candidate. The one shape this over-flags is a fast-forward merge; GitHub's three merge buttons never fast-forward, and over-flagging costs a stale directory, which is the accepted-cheap direction.

## Both code paths, and they cannot drift

The issue noted the hole existed in the helper **and** in Step 0's inline fallback for a checkout without the helper (a bare `git merge-base --is-ancestor`). Fixing only the helper would leave the same false-MERGED reachable on the degraded path, so the guard is replicated in the shell — using a `case` glob over the first-parent SHA list, not a pipe to `grep` (hard-blocked, #1008; and every line is a full 40-char object name, so a substring hit implies equality).

`.claude/lib/tests/test_session_start_step0_fallback.py` **extracts that block from `SKILL.md` by sentinel and executes it verbatim under zsh** rather than restating the logic, which would pin a copy and prove nothing about the file Step 0 actually runs.

## Live regression sweep

Old vs new verdict over all 17 worktrees in this checkout:

| verdict change | count | which |
|---|---|---|
| `remove` -> `FLAG` (FRESH) | 5 | every one a fresh not-yet-committed worktree — including 3 owned by agents live at the time (#1254, #1272, #1399) and this PR's own |
| unchanged | 12 | 8 squash/content-equivalent merged (still auto-removed), 4 genuinely unmerged (still flagged) |

## Fail-safe review

Every new probe failure withholds removal rather than allowing it: `rev-parse --verify` failing, `rev-list --first-parent` failing, and `rev-list` succeeding but listing **nothing** (an unknown state — reading it as "not on the mainline" would fail open into a removal). Each has its own test.

## Found, not fixed

- **The blast-radius bound in the module docstring was subtly wrong and is now corrected in place.** It claimed a misclassification "can cost, at worst, a stale worktree directory ... never even uncommitted content, since a dirty worktree that fails a plain `remove` is simply left alone." That bound holds for *committed* work but never covered #1341, where the worktree destroyed was clean precisely because its agent had not written anything yet. No code change — the sentence now says so.
- **`git worktree remove` on a clean worktree with a `locked` record.** Step 0 already FLAGs `locked` worktrees before classifying, and every Claude-agent worktree here is locked, which is what kept the observed blast radius at zero in practice. That is a second, independent layer, not a substitute — a worktree created by hand or by a non-agent path carries no lock. Not filed; noted here for the reviewer's model of the defense in depth.
- **The mainline probe is O(first-parent history) per worktree per repo** (one `git rev-list` walk). Measured negligible on this repo; a bounded formulation exists but is materially harder to read, and this runs once per worktree in a step that already does a network `fetch` per repo. Deliberately not optimized.

## Gates

- `ENVIRONMENT=test pytest .claude/lib/tests/ .claude/hooks/tests/` — **4405 passed, 1037 subtests passed**.
- `pre-commit` commit-stage on all touched files — green (ruff-format, ruff-lint, cspell, checksums-ascii, skill-graphql-pagination).
- `pre-commit --hook-stage pre-push --all-files` — exit 0.
- `mypy` on the changed/added modules — clean.
- No `--no-verify`, no force. Commit identity via per-commit `-c` flags.

## Review

- Peer reviewer: **Weronika Zielinska**

### DIFF
--- a/.claude/lib/check_worktree_merged.py
+++ b/.claude/lib/check_worktree_merged.py
@@ -123,6 +123,43 @@ looks only at patch-id (diff content), `git cherry`'s own explicit
 verdicts, or a merge's own combined diff, never at commit graph shape or a
 fixed strategy label as such.
 
+**Removal candidacy is a SECOND, ORTHOGONAL axis (#1341) — merged-ness is
+not the removal decision.** Everything above answers *"is this branch's
+content already on `remote_ref`?"*. Step 0 asks a different question:
+*"is there anything here worth reclaiming?"*. Those came apart on a
+**freshly-created worktree that has not committed yet**: its HEAD *is*
+`origin/main`, so it is trivially an ancestor, the fast path answers
+`merged` — **correctly**, since a branch with no commits of its own has, by
+construction, no unlanded content — and Step 0 removed a clean, live
+worktree out from under an active agent (observed on in-flight #1117 during
+wave-29). The false verdict was never about merged-ness; it was about
+reading merged-ness as a reclaim decision.
+
+So the merged classification above is **unchanged and untraded** — a fresh
+worktree still classifies `merged`/`ancestor`, exactly as the #1213
+guarantee requires. What is added is a separate predicate that gates the
+*removal* decision only:
+
+    a branch whose HEAD is itself a commit on `remote_ref`'s own
+    first-parent (mainline) history has produced **no commits of its own**,
+    so there is nothing to reclaim -> `removal_block = "no-own-commits"`,
+    `safe_to_remove` False, CLI exit 3 (not 0).
+
+Why the first-parent chain rather than the `HEAD == remote_ref` equality
+the issue proposed: equality only catches a worktree branched off the
+*current* tip. Sessions here run long enough that `origin/main` advances
+under an idle worktree, after which HEAD is an ancestor but no longer equal
+— same zero-own-work branch, same live agent, and equality would have
+missed it. Mainline membership covers both, and it does not touch the
+population the sweep exists for (#526): a squash- or rebase-merged branch
+tip is not an ancestor at all (it never reaches this guard), and a
+merge-commit-merged tip is reachable only through the merge's **second**
+parent, so it is an ancestor but *not* on the mainline chain — it stays a
+removal candidate. The one shape this over-flags is a fast-forward-merged
+branch (its tip literally becomes a mainline commit); GitHub's three merge
+buttons never fast-forward, and over-flagging costs a stale directory,
+which is the accepted-cheap direction.
+
 **Safety stance (never relaxed):**
 
   * A branch whose net content since `merge_base` is not fully present on
@@ -157,18 +194,39 @@ fixed strategy label as such.
   * The whole test is 100% local git plumbing — no `gh`/network calls, so it
     works offline and never blocks the mandatory session-start step on
     network/auth availability.
+  * The removal-candidacy probe (#1341) degrades in the same direction:
+    if either of its two git calls fails, the result is
+    ``removal_block = "mainline-check-failed"`` — a probe that could not be
+    run is not evidence that the branch has work of its own, so removal is
+    withheld and the worktree is flagged. Note the *classification* is left
+    intact in that case (still `merged`); only removal candidacy degrades,
+    because the two axes are independent and a failure on one is not
+    evidence about the other.
   * **Step 0 no longer force-removes** (owner decision, PR #1213 round 3): a
     worktree that does not remove cleanly is FLAGGED for a manual decision.
     Since `git worktree remove` never deletes the branch ref either way,
     this means a misclassification by this module can cost, at worst, a
     stale worktree *directory* — never a commit, never a branch, and (with
     the force fallback gone) never even uncommitted content, since a dirty
-    worktree that fails a plain `remove` is simply left alone.
+    worktree that fails a plain `remove` is simply left alone. That bound
+    holds for *committed* work; it never covered the #1341 case, where the
+    worktree being destroyed was clean precisely because its agent had not
+    written anything yet.
 
 CLI: ``python3 check_worktree_merged.py <repo_root> <head> [remote_ref]``
 (default ``remote_ref`` is ``origin/main``). Prints one line
-``<status> (<reason>): <detail>`` and exits 0 iff merged, 1 otherwise — the
-exit code is the only thing a caller needs to decide auto-remove vs FLAG.
+``<status> (<reason>)[ [not-a-removal-candidate: <block>]]: <detail>``.
+Exit codes — the exit code is the only thing a caller needs to decide
+auto-remove vs FLAG:
+
+  * ``0`` — merged **and** a removal candidate: safe to auto-remove.
+  * ``1`` — not merged (``unmerged`` or ``error``).
+  * ``2`` — usage error.
+  * ``3`` — merged, but **not** a removal candidate (#1341: no commits of
+    its own, or the candidacy probe failed). Non-zero, so a caller that
+    only knows the old ``0 == remove`` contract still fails safe; a caller
+    that distinguishes it can label the worktree FRESH rather than the
+    misleading UNMERGED.
 """
 
 from __future__ import annotations
@@ -322,16 +380,38 @@ class MergeClassification:
     truly empty commit (explicitly filtered even though `git cherry` does
     mark it `+`) or a merge commit with no unique content of its own — see
     the module docstring's per-commit-corroboration section.
+
+    ``removal_block`` — the SECOND, orthogonal axis (#1341). Empty means
+    "nothing blocks reclaiming this worktree"; non-empty is a short machine-
+    readable reason why it must not be auto-removed even though its content
+    is fully landed:
+
+      * ``no-own-commits``       — HEAD is a commit on ``remote_ref``'s own
+        first-parent history, so the branch has produced nothing of its own
+        (a fresh, not-yet-committed worktree — very likely live).
+      * ``mainline-check-failed`` — the candidacy probe itself errored;
+        withhold removal rather than assume there is work here.
+
+    It never affects ``status``/``merged``: a fresh worktree is genuinely
+    merged, it is just not a reclaim candidate. Callers deciding
+    auto-remove-vs-FLAG must read :attr:`safe_to_remove`, not
+    :attr:`merged`.
     """
 
     status: str
     reason: str
     detail: str = ""
     unmatched_commits: list[str] = field(default_factory=list)
+    removal_block: str = ""
 
     @property
     def merged(self) -> bool:
         return self.status == "merged"
+
+    @property
+    def safe_to_remove(self) -> bool:
+        """The auto-remove predicate: fully landed AND has work to reclaim."""
+        return self.merged and not self.removal_block
 
 
 def _parse_patch_ids(text: str) -> list[tuple[str, str]]:
@@ -352,6 +432,65 @@ def _degrade(reason: str, detail: str) -> MergeClassification:
     return MergeClassification("unmerged", reason, detail)
 
 
+def _removal_block(root: Path, head: str, remote_ref: str, runner: GitRunner) -> tuple[str, str]:
+    """Removal-candidacy probe (#1341) — the axis orthogonal to merged-ness.
+
+    Returns ``(block_reason, block_detail)``; ``("", "")`` means the branch
+    has commits of its own and is a genuine reclaim candidate.
+
+    Only meaningful once ``head`` is known to be an ancestor of
+    ``remote_ref`` (mainline membership implies ancestry, so calling it off
+    that path would always answer "not blocked" at the cost of two git
+    invocations). The test: is ``head`` itself one of the commits on
+    ``remote_ref``'s **first-parent** chain? If so the branch has never
+    committed anything of its own — nothing to reclaim, and very likely a
+    worktree an agent is about to write into.
+
+    Both git calls degrade toward *withholding removal* (never toward
+    removing): a probe we could not run is not evidence of work.
+    """
+    head_res = runner(["rev-parse", "--verify", f"{head}^{{commit}}"], root)
+    head_sha = head_res.stdout.strip() if head_res.returncode == 0 else ""
+    if not head_sha:
+        return (
+            "mainline-check-failed",
+            f"could not resolve {head} to a commit "
+            f"(exit {head_res.returncode}): {head_res.stderr.strip()} — "
+            f"cannot confirm this worktree has commits of its own, so removal is withheld",
+        )
+
+    # First-parent chain of remote_ref. Membership is an exact SHA-set test:
+    # every line is a full 40-char object name, so no prefix/substring
+    # ambiguity is possible.
+    chain = runner(["rev-list", "--first-parent", remote_ref], root)
+    if chain.returncode != 0:
+        return (
+            "mainline-check-failed",
+            f"git rev-list --first-parent {remote_ref} failed "
+            f"(exit {chain.returncode}): {chain.stderr.strip()} — "
+            f"cannot confirm this worktree has commits of its own, so removal is withheld",
+        )
+    mainline = set(chain.stdout.split())
+    if not mainline:
+        # `git rev-list` over a resolvable ref always lists at least that
+        # ref's own commit, so a successful-but-empty result is an unknown
+        # state — never read it as "head is not on the mainline, remove away"
+        # (the same fail-open guard the patch-id path carries).
+        return (
+            "mainline-check-failed",
+            f"git rev-list --first-parent {remote_ref} succeeded but listed no commits — "
+            f"unexpected state, so removal is withheld",
+        )
+    if head_sha in mainline:
+        return (
+            "no-own-commits",
+            f"{head_sha[:12]} is itself a commit on {remote_ref}'s first-parent history, "
+            f"so this branch has no commits of its own — a fresh/unstarted worktree with "
+            f"nothing to reclaim (very likely in active use), not a removal candidate",
+        )
+    return ("", "")
+
+
 def classify_merged(
     repo_root: str | Path,
     head: str,
@@ -367,7 +506,15 @@ def classify_merged(
     # ---- fast path: plain ancestry (covers the merge-commit majority) ------
     anc = runner(["merge-base", "--is-ancestor", head, remote_ref], root)
     if anc.returncode == 0:
-        return MergeClassification("merged", "ancestor", f"{head} is an ancestor of {remote_ref}")
+        # Merged — but ancestry alone does not mean there is anything here to
+        # reclaim. A worktree that has not committed yet sits ON remote_ref's
+        # mainline, which is trivially an ancestor (#1341). Classification is
+        # untouched; only removal candidacy is gated.
+        block_reason, block_detail = _removal_block(root, head, remote_ref, runner)
+        detail = f"{head} is an ancestor of {remote_ref}"
+        if block_reason:
+            detail = f"{detail}; {block_detail}"
+        return MergeClassification("merged", "ancestor", detail, removal_block=block_reason)
     if anc.returncode != 1:
         # `--is-ancestor` signals a true negative with exit 1; anything else
         # (bad rev, corrupt repo, ...) is a plumbing error, not "not merged".
@@ -576,8 +723,16 @@ def main(argv: Sequence[str] | None = None) -> int:
     remote_ref = args[2] if len(args) > 2 else "origin/main"
 
     result = classify_merged(repo_root, head, remote_ref)
-    print(f"{result.status} ({result.reason}): {result.detail}")
-    return 0 if result.merged else 1
+    prefix = f"{result.status} ({result.reason})"
+    if result.removal_block:
+        prefix = f"{prefix} [not-a-removal-candidate: {result.removal_block}]"
+    print(f"{prefix}: {result.detail}")
+    if result.safe_to_remove:
+        return 0
+    # Merged but not a reclaim candidate gets its own non-zero code (#1341)
+    # so a caller can say FRESH instead of the misleading UNMERGED; a caller
+    # that only knows `0 == remove` still fails safe on it.
+    return 3 if result.merged else 1
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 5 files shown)


parametrization:
Reviewed at head `d9186abb0048b4892918fdbe6a3e1f4ebbd404ad`.

**Must-fix — the shell fallback fails OPEN on a probe error, unlike the Python path it mirrors.**

`.claude/skills/session-start/SKILL.md` Step 0, the `# BEGIN legacy-ancestry-fallback` / `# END` block (currently around line ~136-148 of the diff; same text is pulled verbatim into `.claude/lib/tests/test_session_start_step0_fallback.py`):

```
_mainline="$(git -C "$repo" rev-list --first-parent origin/main 2>/dev/null)"
case "$_mainline" in
  *"$head"*) _mrc=3 ... ;;
  *) _mrc=0 ;;
esac
```

The exit status of the `rev-list --first-parent` call is never checked. If that git invocation itself fails (transient error, disk pressure, corrupt object — any of the reasons the Python `_removal_block()` explicitly guards against at `.claude/lib/check_worktree_merged.py:167-173`), `_mainline` is simply empty, the `case` falls to its default arm, and `_mrc=0` — **auto-remove** — exactly the fail-open outcome #1341 exists to close, and the opposite of what the Python probe does for the identical failure (it returns `"mainline-check-failed"` and withholds removal, with a dedicated test).

I reproduced this against the PR's actual block, not a paraphrase: extracted it from `SKILL.md` at this head with the same sentinel technique `test_session_start_step0_fallback.py` uses, and ran it under zsh with a `git` shim that fails only on `rev-list --first-parent origin/main` (everything else delegates to real git):

```
$ PATH="$FAKEBIN:$PATH" zsh -c '... extracted block ...'
mrc=0 mreason=ancestry-only (helper not found) mainline=''
```

`mrc=0` means Step 0 runs `git worktree remove` on it. With the same shim removed (real git), the identical block correctly returns `mrc=3`.

This path only fires when `.claude/lib/check_worktree_merged.py` is missing from the checkout ("very old checkout" per the comment), so it's a narrow defense-in-depth path, not the primary one — but it's untested for exactly this failure mode (the 4 new tests in `test_session_start_step0_fallback.py` never fail a git call, whereas the Python side has 3 dedicated tests for this same class: `test_mainline_probe_failure_blocks_removal`, `test_mainline_probe_empty_output_blocks_removal`, `test_head_resolution_failure_blocks_removal`), and the PR description's "Fail-safe review" section states "every new probe failure withholds removal" — that's true of the Python path and false of this one. Given the whole point of the PR is closing a worktree-destruction fail-open, I'd like this closed before merge rather than deferred: check the exit status of the `rev-list` call (and ideally the empty-output case too, for parity with the Python side) and withhold removal (`_mrc=3`) on failure instead of falling through to the `case` default.

**Everything else holds up:**

- Decomposition (merged vs. removable as two independent axes) is sound, and Step 0 consumes the new predicate correctly — `SKILL.md:150` reads `_mrc -eq 3` → FRESH/flag, `-eq 0` → remove, else → UNMERGED/flag.
- Old-caller fail-safe on the primary Python CLI path, verified empirically (not read off the PR body): `check_worktree_merged.py <fresh-worktree-head> origin/main` returns rc=3 on this repo's checkout right now; a caller written against the old `rc==0` contract falls into "flag/keep," not "remove."
- Live read-only sweep reproduced against the **current** tree (denominator: 20 non-main worktrees right now, not the PR's 17 — the tree has grown since the author's own sweep on 2026-08-11, expected drift, not a discrepancy): 3 of 20 flip `remove` → `FLAG`. Two (`agent-a72e0ea75cb3373c8`, `agent-a9c21592be228529c`) are `locked`, so already protected today by Step 0's pre-existing lock check regardless of this fix. The third, `scratchpad/nk-pr1429/basewt`, is **unlocked**, detached, HEAD==`origin/main` — exactly the hand/scratch-created no-lock-record case the review brief called out as the only one genuinely at risk — and it is live in this tree right now. That's a direct, present-tense confirmation that the fix closes the unlocked hole, not just the already-protected agent-worktree case.
- Tests: 44/44 pass at this head (`test_check_worktree_merged.py` + the new `test_session_start_step0_fallback.py`). `mypy` clean on the three touched/added modules.
- CI: `statusCheckRollup` is `SUCCESS` at `headRefOid` `d9186abb0048b4892918fdbe6a3e1f4ebbd404ad` (checked via GraphQL against the SHA, not the PR number) — 21/21 checks green.

No destructive worktree operations were run against the live tree; all sweep/CLI testing above was read-only or against a scratch clone/repo under the session scratchpad.

---
Requestor: Weronika Zielinska
Requestee: Nino Kavtaradze
TechDebt: none

parametrization:
**Merge-gate review — head `d9186abb0048b4892918fdbe6a3e1f4ebbd404ad`**, read via the contents API at that SHA (never a local checkout). `statusCheckRollup` at that SHA: **21/21 SUCCESS**, no empty rollup. Full suite re-run in a scratch clone pinned to `d9186ab`: **4405 passed, 1037 subtests** — matches the body's claim exactly. `mypy` + `ruff check` clean locally on the changed modules.

## 1. The decomposition — right call

Keeping the classification untouched is correct, and it is the only reading that survives scrutiny. A fresh worktree genuinely *satisfies* the #1213 net-content guarantee: a branch with no commits of its own has no unlanded content, so reclassifying it `unmerged` would have broken the guarantee in order to fix a symptom that was never about merged-ness. The bug was the *caller* reading one predicate as the answer to a different question.

The worry with splitting one contract into two is that a caller must now get both right independently. That worry is neutralised here because the split is **not exposed at the seam Step 0 reads**. The CLI still hands back a single scalar; `safe_to_remove` is folded into it before it crosses the boundary (`check_worktree_merged.py:730-735`). A caller cannot get "merged" right and "removable" wrong, because a caller never sees two values. The one surface where the two axes *are* separately visible is the Python dataclass — `.merged` is still True for a fresh worktree — and that is guarded only by a docstring ("Callers deciding auto-remove-vs-FLAG must read `safe_to_remove`, not `merged`", `check_worktree_merged.py:395-398`). Acceptable today: `rg`ed the whole tree with `--hidden --no-ignore`, and the only production consumer of this module is Step 0 via the CLI. The other two `worktree remove` sites are Step 0 itself and `wave-wrapup/SKILL.md:159-175`, and wave-wrapup's is a deliberate end-of-wave sweep with its own clean/dirty check, not a merged-ness classifier — no drift site left unupdated.

**Step 0 consumes it correctly at every branch** (`session-start/SKILL.md:109-175`): locked → LOCKED (classifier never runs); empty HEAD → UNMERGED; `_mrc==3` → FRESH; `_mrc==0` → `remove`, and a refused remove → DIRTY; **everything else** → UNMERGED. That last arm is the important one — it is an `else`, not a `-eq 1`, so exit 2 and any future code fail safe rather than falling through to the remove branch. The prose at `SKILL.md:199` closes the loop by telling the reader a FRESH entry is *not* a cleanup backlog item.

## 2. Fail-safe — verified, not taken on report

Ran an old-contract caller (`rc=0` ⇒ remove, the only rule it knows) against the new CLI on a scratch repo with five real branch shapes:

| shape | old CLI | new CLI | old-contract caller's decision |
|---|---|---|---|
| fresh at `origin/main` tip | 0 | **3** | keeps (flag) |
| fresh off an *older* `origin/main` | 0 | **3** | keeps (flag) |
| merge-commit-merged feature tip | 0 | 0 | removes — no regression |
| squash-merged feature tip | 0 | 0 | removes — no regression |
| genuinely unmerged | 1 | 1 | keeps (flag) |

Then mutation-tested the probe itself with a `git` shim on `PATH` that fails only the named subcommand, so `merge-base --is-ancestor` still succeeds and the code reaches the guard:

| injected failure | helper exit | direction |
|---|---|---|
| `rev-list --first-parent` exits 128 | **3** | withholds |
| `rev-list --first-parent` exits 0, prints nothing | **3** | withholds |
| `rev-parse --verify` exits 128 | **3** | withholds |

All three fail closed. And the tests that assert this are **not inert** — five mutations against the artifact, each caught (denominator: 5/5):

| mutation | failing tests |
|---|---|
| neuter `if head_sha in mainline` | 3 |
| `rev-list` failure returns no block | 1 |
| empty-`rev-list` guard removed | 1 |
| CLI returns 0 for merged-but-blocked | 1 |
| #1341 guard stripped out of SKILL.md's fallback | 2 (via the sentinel extraction) |

The SKILL.md extraction-by-sentinel is the right shape and it earns its keep: mutating `SKILL.md` alone, with the Python untouched, fails the suite. It pins the file Step 0 actually runs, not a restatement of it.

## 3. The unlocked case — and a correction to the premise

The fix covers it, and the case is far more common than "hand-created". Counting the live parent checkout right now: **2 of 13** `.claude/worktrees/agent-*` worktrees carry a `locked` record, plus **0 of 5** agent-created scratchpad worktrees. So 16 of the 17 non-main worktrees on this machine are unlocked and reach the classifier. Lock-first ordering was never what held the blast radius at zero in general — it covers 2/17 today.

Read-only old-vs-new sweep over those 17 (classifier only, no removal run): **3 verdicts changed**, all `0 → 3`. One of them, `scratchpad/nk-pr1429/basewt`, is **unlocked, fresh, and live right now** — under the pre-fix classifier the next `/session-start` would have removed it out from under a running agent. That is live-trace evidence, not a synthetic fixture. The other 14 are unchanged (2 merged and still auto-removed, 12 flagged unmerged) — #526's population is untouched.

Ruling on the ordering: lock-first is **still independently load-bearing, but for a disjoint case, and only for 2/17 of the tree**. It alone covers a live agent whose commits have *already landed* — that branch has commits of its own, so it is a legitimate reclaim candidate (exit 0) and the freshness predicate by construction cannot see it. Neither layer subsumes the other; after this PR the freshness predicate is the load-bearing protection for the fresh case, and the lock is the thin residual layer for the landed-but-live case. Worth knowing that the second layer is thinner than it looks.

## 4. The #1212 guarantee docstring — holds verbatim

Diffed the module against base `40cfe9f`. The guarantee paragraph (`check_worktree_merged.py:19-26`), both coordinated tests, the `git cherry` carve-outs, the `--cc` evil-merge check and every degrade path are byte-identical. The only change inside `classify_merged` is on the ancestor fast path, and it changes `detail` only — `status` and `reason` still return `merged`/`ancestor` (`check_worktree_merged.py:507-517`).

The blast-radius bound (`:205-214`) was **corrected, not softened**. It previously claimed a misclassification could never cost uncommitted content; that was true for committed work and silently false for #1341, where the destroyed worktree was clean because nothing had been written yet. The new text names that gap in plain words rather than hedging the original claim into vagueness — the opposite of #1388's failure. Two new residuals, both **named**: `mainline-check-failed` when the candidacy probe cannot be run (`:197-204`), and the one over-flagging shape, a fast-forward merge whose tip literally becomes a mainline commit (`:158-161`). The `patch-id` whitespace residual is untouched and still disclosed.

The `HEAD == origin/main` equality guard the issue proposed would have been wrong, and the PR is right about why: sessions here outlive `origin/main`, and `test_fresh_worktree_off_older_remote_tip_is_not_a_removal_candidate` pins the aged form the equality guard would miss. Confirmed empirically — the "fresh off older main" row above.

## 5. Finding — the shell fallback fails OPEN on a probe failure (deferred, #1446)

`.claude/skills/session-start/SKILL.md:139-147`. The Python probe has three fail-closed guards; its shell replica has none. Both `rev-list --first-parent` sub-cases fall through the `case` glob's `*)` arm to `_mrc=0`, which is Step 0's *run `git worktree remove`* branch:

```
--- python helper ---            exit=3    # removal withheld
--- shell fallback (verbatim) ---  _mrc=0    # Step 0 RUNS git worktree remove
```

Same result for `rev-list` exiting 0 with empty output. `test_session_start_step0_fallback.py` has four cases, all against a healthy `git`, so neither sub-case is covered. The inline comment at `:124-131` says the guard is "replicated" — it is replicated in *effect* on the happy path, not in *degrade direction*.

**Fold-in vs defer, and the question that decided it: prevalence.** Direction alone would say fold in — a silent miss here destroys live work, an over-block leaves a stale directory, and that asymmetry normally wins. What overrides it is compound unreachability, not cost-of-fix. The branch is gated on `[ -f "$REPO_ROOT/.claude/lib/check_worktree_merged.py" ]` with `REPO_ROOT` always the parent org repo, so it cannot execute in any checkout that has the helper — every checkout since #1212 landed on 2026-08-02. On top of that it needs `rev-list --first-parent origin/main` to fail or return empty *immediately after* `merge-base --is-ancestor` succeeded against that same ref. Two independent near-impossible preconditions, and the path that does run is verified fail-closed on all three failures. Charter `pull-requests/acceptance-scope.md` § Security Guards Belong Inline agrees: this is the explicit "defense-in-depth hardening that doesn't change the threat surface" carve-out, and the inline path is already safe. Filed as #1446 with a reproduction, a fix sketch and acceptance criteria.

## Notes, no action

- `test_check_worktree_merged.py:977` intercepts `rev-list` at `occurrence=1` for the `--merges` call, and `_removal_block` now also calls `rev-list`. They do not collide because `_removal_block` runs only on the ancestor fast path and that test exercises the non-ancestor content path — but occurrence-indexing a shared runner across a module with two `rev-list` call sites is a sharp edge for the next person adding one.
- The memory note update (`feedback_flag_precision_decays_with_age.md`) is honest: it marks the hole closed, dates it, keeps the generalisable reading rule, and does not overclaim the fix into "the classifier errs safe" — which is the exact generalisation the note exists to warn against.

**Verdict: approve.** The decomposition is the right one, the fail-safe behaviour is verified rather than asserted, the #1213 guarantee survives verbatim with its new residuals named, and the one real finding is on a path no live checkout can reach.

---
Requestor: Aino Virtanen
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: #1446

parametrization:
Superseding my earlier ChangesRequested with a fresh verdict at the same head, `d9186abb0048b4892918fdbe6a3e1f4ebbd404ad` (unchanged since my first review).

**On the fail-open fallback bug (my earlier must-fix):** it's real, the merge gate reproduced it independently, and it's now tracked as issue #1446 (open, `tech-debt` + `phase-10`, assigned). I re-examined the prevalence argument myself rather than taking it on report, specifically the one gap I could see in it: the `# BEGIN legacy-ancestry-fallback` block only runs when `[ -f "$REPO_ROOT/.claude/lib/check_worktree_merged.py" ]` is false, and I wanted to know whether that check is per-repo or parent-only, since Step 0 iterates the parent **and** 7 child repos. It's parent-only — the presence check and the `python3` invocation both use `$REPO_ROOT` (the parent), never `$repo` (the repo currently being classified), so the parent's single copy is reused generically against every repo via `classify_merged`'s `repo_root` argument. Child repos genuinely never need their own copy (confirmed none of the 7 local child clones have one, and that's fine by design). So reachability collapses to one condition: is `/session-start`'s own resolved `REPO_ROOT` an unusually stale parent checkout that predates 2026-08-02's #1212 — which `/session-start` being the mandated first action of every session, plus its own `sync_main.py` fast-forward at the top of Step 0, makes very unlikely in practice, on top of the independent `rev-list` failure the bug also requires. That closes the loophole I was checking for; I don't have a further objection to the prevalence claim.

On blast radius: even in the fail-open case, the subsequent `git worktree remove` is still the non-force call Step 0 always uses — a worktree with any uncommitted content fails that removal and is FLAGGED `DIRTY` rather than destroyed. The exposure is bounded to the same class already accepted elsewhere in this module (a clean, contentless worktree directory disappearing out from under a live session), not an uncommitted-work loss, and it now requires two independently narrow preconditions stacked rather than the single trivially-common one #1341 itself was.

Given the followup is filed, boarded, and cheap to land (a 2-3 line exit-status check, as scoped), and the primary path — the one actually exercised on every real checkout today — is solid (I verified the old-caller fail-safe, the live sweep including the unlocked case, and the full test/CI state in my first review and none of that changes here), I'm satisfied this is a legitimate inline-safe-plus-tracked-followup case rather than one requiring an inline fix before merge.

---
Requestor: Weronika Zielinska
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: #1446

Issue #1365, opened 2026-08-07
`apply_distribution_discipline` cannot distinguish ceiling-entry from ceiling-eviction — two caller obligations held shut by prose


parametrization:
## Summary

`trust_signals.apply_distribution_discipline(proposals)` receives `{name: (proposed_score, signals)}` and **never sees the engineer's old score**. It therefore cannot distinguish an engineer *entering* the ceiling (5 from below — the case the cap exists for) from one *holding* it (already 5, which the cap must not evict).

It is an entry gate. Nothing in the signature says so, and nothing can test that it is used as one.

## Evidence

Fed the full wave-29 roster, the function caps **Nino Kavtaradze 5 -> 4**, converting a `+2` delta into a net `-1` for an engineer who has held 5 across prior waves. Applying it that way contradicts four waves of applied history: W25 4 ceiling-holders, W26 6, W27 5, W28 5 — **20 rows, none ever evicted** (counted independently from `trust_matrix.md`'s own wave sections by both #1363 reviewers, who agreed on the total and on W28's count of 5 correctly classifying Nino as an entrant rather than a holder).

The batch composition also matters in a non-obvious way. From #1363's `trust_matrix.md`, reproduced independently three times:

| batch | `top` | Nurul | Nadia |
|---|---|---|---|
| full roster (10) | 17 (Aino) | 4 | 4 |
| full roster minus ceiling-holders (8) | 11 (Weronika) | 4 | 4 |
| only those reaching 5 from below (2) | **8 (Nadia)** | 4 | **5** |

So the caller must feed the **whole roster** to anchor `top` correctly, while the cap's rule applies only to a subset. Both #1363 reviewers independently produced a wrong number from the narrowest batch before working out the intent.

## The actual defect

#1363 ships **two caller obligations encoded only in prose** — "it is an entry gate" and "feed it the whole roster" — in `trust_matrix.md` and `.claude/skills/wave-retro/SKILL.md`. Neither is testable at the current signature, and the three existing `DistributionDiscipline` tests exercise the entrant path only.

A function whose contract is narrower than its interface, with the gap held shut by prose, is the same defect shape as #1349 itself — one function over. That is the argument for fixing it rather than documenting it, and it is why both #1363 reviewers converged on the same remedy independently.

## Proposed change

[...]

PR #1363, opened 2026-08-07
fix(#1349 +5): two-band rate-relative rework rubric, wave-29 scores applied, and the five follow-ups

parametrization:
Implements the owner's ruling on #1349 (comment `issuecomment-5222214345`, 2026-08-07): **Option 3 extended to the penalty threshold**. Options 1 and 2 were explicitly rejected in the ruling and are not implemented; the issue *body*'s "I recommend 1 + 2 together" is superseded.

**Conflict of interest — disclosed.** This change rates me. Under the adopted rubric my wave-29 delta is **+2**, but I am already at `MAX_SCORE = 5`, so it gives me no upside — `clamp(5 + 2)` is 5, exactly where I was held. It does move me from a `-1` to a `+2` on the recorded rubric-delta column, which is a reputational change even where the score does not move, so the disclosure stands regardless. Precedent: #1356, where both reviewers disclosed COI on the thread. Reviewers should weight my judgement calls below accordingly, particularly the distribution-discipline finding, which is the one place I chose *not* to take the mechanically-computed number.

---

## The change

`score_delta`'s rework signal becomes rate-relative and two-band. Both bars are a **multiplication** — no ratio, no float, no zero-guard, because `prs_merged == 0` implies `must_fix_received == 0` (`extract_signals` only ever increments the two together), so `0 <= 0` classifies a non-authoring engineer as clean naturally.

| Band | Condition | Effect |
|---|---|---|
| Clean | `must_fix_received <= prs_merged` (≤ 1/PR) | bump eligible |
| Neutral | `prs_merged < recv <= 2 * prs_merged` | **no bump, no ding** |
| Penalty | `must_fix_received > 2 * prs_merged` (> 2/PR) | −1 |

`ci_red_merges` and `review_false_positives` are untouched: absolute, per-instance −1, and each still disqualifies the positive branch outright. The `[-2, +2]` clamp holds.

## Predicate split — the delegated design decision

Split into **three** new predicates on `Signals`, with `has_negative()` left exactly as it was:

| Predicate | Answers | Caller |
|---|---|---|
| `has_negative()` *(unchanged)* | is there **any** negative signal to report? (absolute) | `negative_signal_line` |
| `rework_within_clean_bar()` | `recv <= prs` | the clean band boundary |
| `rework_above_penalty_bar()` | `recv > 2 * prs` | the penalty band boundary |
| `qualifies_for_bump()` | is the wave clean enough to **earn** a bump? | `score_delta`'s positive branch |

[...]

PR #1428, opened 2026-08-11
fix(trust): make distribution discipline's anchoring population its own argument (#1370)

### LINKED ISSUE #1370: `apply_distribution_discipline`'s anchoring population and gated population are the same argument — 'feed it the whole roster' is unenforceable

parametrization:
Closes #1370.

## What was wrong

`apply_distribution_discipline` ranked and capped over the **same dict**. `top` was the maximum composite *within whatever batch it was handed*, so the anchoring population and the gated population could not be told apart — and "feed it the whole roster" was an obligation the function could state and never check. #1363 and #1365 both left it as labelled prose, pinned by a test that *demonstrated* the sensitivity (`test_top_is_batch_relative_so_feed_the_whole_roster`) rather than removing it.

## Failing-before evidence

**1. Behavioural — the wrong allocation, on unmodified `main` (`40cfe9f`).** Capping only the two wave-29 ceiling entrants was a thing a caller could legitimately want, and the only way to express it was to hand that slice in as the whole argument:

```
$ ENVIRONMENT=test python3 prefix_demo.py          # against origin/main trust_signals.py
cap just the two ceiling entrants:
  apply_distribution_discipline(entrants) -> {'Nadia': 5, 'Nurul': 4}
the whole wave in the batch:
  apply_distribution_discipline(wave)     -> {'Nadia': 4, 'Nurul': 4, 'Aino': 5}
```

Nadia (composite 8) becomes her own maximum and keeps a 5 the wave (maximum 17, Aino) caps to 4. Same numbers as the issue's three-batch table. Post-fix, the same intent now produces the wave's answer:

```
$ ENVIRONMENT=test python3 postfix_demo.py         # against this branch
cap just the two ceiling entrants, ranked against the wave:
  -> {'Nadia': 4, 'Nurul': 4}
population that does not cover the gated set:
  ValueError: ranking population 'P10W29 partial' does not cover ['Nurul']: capping an
  engineer against a population they are not in ranks them on someone else's wave. Pass
  the whole wave (RankingPopulation.from_wave), not a filter of it.
```

**2. The test file on this branch, run against the unmodified `trust_signals.py`** (`git checkout origin/main -- .claude/lib/trust_signals.py`, tests unchanged):

```
FAILED DistributionDiscipline::test_gated_slice_is_ranked_against_the_wave_not_against_itself
FAILED DistributionDiscipline::test_capping_an_engineer_outside_the_ranking_population_is_refused
FAILED DistributionDiscipline::test_empty_ranking_population_is_refused
FAILED DistributionDiscipline::test_declared_population_demands_a_stated_origin
FAILED DistributionDiscipline::test_from_wave_takes_the_whole_extract_and_records_its_provenance
FAILED DistributionDiscipline::test_a_narrow_population_is_still_narrow_but_no_longer_silent
FAILED DistributionDiscipline::test_five_reserved_for_top_performer
FAILED DistributionDiscipline::test_four_passes_through
FAILED DistributionDiscipline::test_no_five_when_top_is_not_positive
FAILED DistributionDiscipline::test_ceiling_holder_with_a_positive_delta_is_not_capped
FAILED DistributionDiscipline::test_ceiling_entrant_below_the_wave_max_is_capped
FAILED DistributionDiscipline::test_holder_exemption_does_not_rescue_a_dropping_score
FAILED DistributionDiscipline::test_wave_29_applied_scores_are_unchanged_by_the_entry_gate
13 failed, 92 deselected in 0.57s
E       AttributeError: module 'trust_signals' has no attribute 'RankingPopulation'
```

**Reviewer, read this honestly rather than counting the FAILEDs.** Every one of those 13 fails against pre-fix code, but 12 of them fail *because the API does not exist yet* — which is a weak form of "failing before", and exactly the shape wave-30 exists to be sceptical of. The load-bearing evidence is (1): the pre-fix run returns `Nadia: 5` where the wave's answer is 4, and `test_gated_slice_is_ranked_against_the_wave_not_against_itself` asserts 4 for that same scenario. The fix necessarily changes the signature (that *is* the acceptance criterion), so an API-shaped failure is unavoidable for the new tests; the numeric assertion is what makes it a test of the defect rather than of the diff.

## What changed

`ranking_population` is a new **required keyword argument** carrying a new `RankingPopulation`. Required, not defaulted: a default would reintroduce the silent path.

- **`RankingPopulation.from_wave(phase, wave, status_path)`** — takes a *wave*, not a mapping, so there is no argument for a caller to filter on the way in. `trust_signals.py score` now uses it and derives its proposals from `population.signals`, so the CLI shares one extraction and structurally cannot rank against a slice. Extraction lives on the type, not inside the scorer — `apply_distribution_discipline` stays pure and takes the population as a value (self-fetching inside the scorer was rejected on #1363 review, and this respects that).
- **`RankingPopulation.declared(map, origin=…)`** — the hand-assembled path (`/wave-retro` Step 5 builds from the retro table). `origin` is mandatory and non-blank, and is quoted back in the refusal messages, so a population that turns out to be wrong says where it came from.
- **Two hard refusals**, `ValueError` rather than a quiet number: an empty ranking population, and capping an engineer the population does not cover. The second is the check the single-argument shape could not perform at all — it was structurally impossible for the gated set to fall outside the anchoring set when they were one object.
- The return value covers `proposals` only; being in the ranking population anchors the maximum, it does not score you.

## What this does NOT guarantee — stated, not implied

A pure function cannot verify that a mapping handed to it is the whole wave. No signature can. A caller who assembles a two-person population and writes an `origin` for it **still gets the two-person answer** — that residual is pinned by `test_a_narrow_population_is_still_narrow_but_no_longer_silent`, which is what supersedes the old sensitivity test rather than deleting its evidence. What is removed is arriving at the narrow answer *by omission*: the population is now chosen at every call site, the unfiltered path is the easy one, and the narrow path costs a written justification.

Both the `RankingPopulation` docstring and `trust_matrix.md` say this in as many words. I would rather the next reader meet the limit here than infer a guarantee that is not there — that inference is how the original obligation survived two PRs.

## Files

- `.claude/lib/trust_signals.py` — new `RankingPopulation`; `apply_distribution_discipline` signature + refusals + `top` source; `_cmd_score`; one clause in the module docstring's Extraction bullet.
- `.claude/lib/tests/test_trust_signals.py` — 6 new tests in `DistributionDiscipline`, superseding `test_top_is_batch_relative_so_feed_the_whole_roster`; the other 7 updated for the new signature via a `_whole_wave` test helper (deliberately a test helper — a `RankingPopulation.from_proposals` shipped in the library would be a one-call re-entry to the defect).
- `.claude/team/trust_matrix.md` — § Distribution discipline: the unenforceable-obligation paragraph replaced by what the signature enforces + the residual. The wave-29 record's three-batch table is kept as the evidence the obligation was load-bearing, with its *procedure* sentence marked superseded.
- `.claude/skills/wave-retro/SKILL.md` — Step 5 bullet: the two-argument call shape; the "one obligation remains on you" sentence removed.

## Collision with #1420 (#1369, Nadia Khoury) — please read before merging

#1420 is open against the same function's **inner** `composite()` and I stayed out of it deliberately. Two expected conflicts, both mechanical:

1. **`trust_signals.py`** — her one-line `composite()` change (`- s.must_fix_received` → `- s.attributable_rework()`) sits ~4 lines above my rewritten `top = max(composite(s) for s in ranked.values())`. Resolution: keep both.
2. **`trust_matrix.md` § Distribution discipline** — her replaced paragraph is adjacent to mine. Resolution: **keep both rulings** — hers is about attribution, mine about anchoring.

**Not a conflict, but a real break:** #1420 appends two tests that call `apply_distribution_discipline(proposals)` with the old single-argument signature. Whichever of the two merges **second** must add `ranking_population=` to the other's tests, or the suite goes red on `main`. This is a merge-gate item, not something either branch can fix alone.

## Gates

- `ENVIRONMENT=test python3 -m pytest .claude/lib/tests/ .claude/hooks/tests/ -q` → **4398 passed, 1037 subtests passed**
- `pre-commit run --all-files` → all pass (ruff-format, ruff-lint, actionlint, cspell, fixture-realism, skill-graphql-pagination, checksums-ascii)
- `pre-commit run --all-files --hook-stage pre-push` → all pass (mypy, mypy-skills, pytest, pytest-skills, memory-budget, headcount-budget, doc-freshness, office-drift, mermaid-render)
- `python3 .claude/lib/pre_commit_ci_sync.py .` → `OK: pre-commit config mirrors all CI-enforced checks.`

## Review

- Peer reviewer: **Nadia Khoury**
- Secondary reviewer: **Wanjiku Mwangi** (assigned because reviewer and merge gate are the same person; the 2-distinct-reviewer gate needs a third name)
- Merge gate: **Nadia Khoury**

Not self-reviewed. Do not merge without both verdicts + the merge gate.


parametrization:
Reviewed at head `96da412b230f9e3e3d29daea986219bca0b65e1b` (recorded myself via `gh pr view --json headRefOid`, not taken from the brief).

**CI:** `python3 .claude/lib/pr_ci_state.py 1428` → `ready`, 21/21 checks reported pass (denominator: 21). `gh pr checks` confirms the same 21, non-empty rollup.

**Tests/lint, run myself in a scratch worktree at head (not just trusting the PR body's numbers):**
- `ENVIRONMENT=test python3 -m pytest .claude/lib/tests/test_trust_signals.py -q` → 105 passed, 18 subtests passed.
- `ruff check` and `mypy` on both changed `.py` files → clean.

**Core question — does splitting the argument make "feed it the whole roster" actually enforceable, or just move the same unenforceable convention to a new parameter?**

Partially, and the PR is honest about exactly where the line is:

- Along the one production call path (`_cmd_score` → `RankingPopulation.from_wave`, `trust_signals.py`), the defect is now structurally impossible: `from_wave` takes a wave, not a mapping, so there is no argument surface for a caller to filter on the way in. I `rg`'d the whole `.claude/` tree for `apply_distribution_discipline(` outside the test file and found exactly this one caller — it's the only production path today, and it's safe.
- Along the hand-assembled path (`RankingPopulation.declared`), nothing is actually verified — a caller can still build a population equal to (or a subset of) the capped set, give it a plausible `origin` string, and reproduce the original defect exactly (`Nadia` keeps 5 instead of being capped to 4). This is pinned deliberately by `test_a_narrow_population_is_still_narrow_but_no_longer_silent` and stated in `RankingPopulation`'s "What this deliberately does not claim" section (`trust_signals.py:274-281`) and in `trust_matrix.md`'s new paragraph. I don't consider this a gap in the PR — a pure function genuinely cannot verify a mapping is "the whole wave," and disclosing the residual instead of implying a guarantee that isn't there is the right call, not a shortcut.
- What *is* now enforced everywhere: `ranking_population` can no longer be omitted (required kwarg), an empty population raises, and — the check the old single-argument shape could not perform at all — capping someone the population doesn't cover raises, naming them and the population's `origin` (`trust_signals.py:861-868`).

**#1420 collision — verified directly, not taken on the PR body's word:**

- Fetched #1420 at its own head (`0344207...`) and confirmed the exact call sites: `test_trust_signals.py:1247` and `:1264` are two *new* tests #1420 adds (`test_orchestrator_caused_rework_does_not_flip_the_ceiling_five`, `test_unmarked_rework_still_costs_a_composite_point`), both calling `ts.apply_distribution_discipline(proposals)` positionally. `:1293` is the pre-existing `test_wave_29_applied_scores_are_unchanged_by_the_entry_gate`, which #1428 already updates on its own side, so that one isn't actually part of the break — only the two new tests are.
- Simulated the merge myself: fetched `pull/1420/head`, `git merge --no-commit` into a scratch worktree at #1428's head. Contrary to the PR body calling these "expected conflicts," git auto-merged `trust_signals.py`, `test_trust_signals.py`, and `trust_matrix.md` with **zero conflict markers**.
- Inspected the merged code: `composite()`'s body reads `- s.attributable_rework()` (#1420's ruling) while `top` is computed from `ranking_population.signals` (#1428's ruling) — two orthogonal axes of the same function (which quantity you rank by vs. which population you rank over), genuinely compatible, not just adjacent.
- Ran the merged suite: exactly the 2 predicted tests fail, both with `TypeError: apply_distribution_discipline() missing 1 required keyword-only argument: 'ranking_population'` — confirming the "real break, not a conflict" framing is accurate.
- Applied the one-line fix per test (`ranking_population=_whole_wave(proposals)`) and reran: 107 passed, 18 subtests passed, `ruff` and `mypy` clean.
- `trust_matrix.md`: the two new paragraphs (#1428's "anchoring population is an argument," #1420's "what composite ranks on") sit adjacent but address different questions (who you rank over vs. what you rank by); merged content reads coherently with no contradiction.
- Conclusion: **the hazard is real, exactly as scoped, and "keep both rulings" is a verified-working resolution** — not a hand-wave. The owner's sequencing (#1420 first, then merge `main` into this branch, add `ranking_population=` to the two new tests) is the correct, minimal fix.
- Cleaned up: aborted the simulated merge, force-removed the scratch worktree, deleted the temp branch. Parent working copy untouched throughout (no `git checkout`/branch switch in `/home/parameterization/code/noorinalabs-main`).

**Non-blocking observation, filed as tech-debt (#1430):** the new coverage check (`trust_signals.py:861`) verifies only that `ranking_population.signals` has a *key* for every capped name, not that the `Signals` value under that key is the same as `proposals[name].signals`. `top` reads from the population's signals; a capped engineer's own composite reads from their `Proposal.signals` — two independently-suppliable values for the same name. No current caller triggers a divergence (`_cmd_score` derives both from one extraction), but nothing enforces that they stay in sync as more callers adopt `RankingPopulation.declared`. Filed as #1430 rather than blocking here since it's speculative today and fits better as a scoped follow-up than a same-PR fix.

---
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1430

parametrization:
**Merge-gate review — reviewed at head `96da412b230f9e3e3d29daea986219bca0b65e1b`**, files read from `repos/.../contents/<path>?ref=96da412`, tests run in a detached worktree at that SHA. Not the local checkout.

**CI:** `21 of 21` checks SUCCESS via `statusCheckRollup` at the head SHA — non-empty rollup, `pr_ci_state.py 1428` → `ready`, 21 reported.
**Tests at head:** `105 passed, 18 subtests` in `test_trust_signals.py`; PR body's `4398 passed, 1037 subtests` for the full suite reproduces.
**Callers:** `1 of 1` production call site updated (`trust_signals.py:963-974`, `_cmd_score`) plus `1 of 1` documented procedure (`wave-retro/SKILL.md:131`). Zero missed — a repo-wide `rg --hidden` for `apply_distribution_discipline` finds no third.

## 1. The core question: enforceable, or relocated? — **it forces the decision plus one new detection; it does not force correctness, and it does not claim to**

I ran the four cases against the built engine at head rather than reading the signature (`.claude/lib/trust_signals.py:795-878`):

| call | result |
|---|---|
| whole wave anchors | `{'Nadia': 4, 'Nurul': 4}` |
| `declared({n: p.signals for n, p in proposals.items()}, origin="x")` | `{'Nadia': 5, 'Nurul': 4}` — the pre-#1370 answer, cost: one string |
| population missing a gated engineer | `ValueError: ranking population 'P10W29 partial' does not cover ['Nurul']` |
| empty population, non-empty gated set | `ValueError: empty ranking population ('nobody')` |

Row 2 is the honest ceiling for a pure function, and the PR does not paper over it: it is pinned by `test_a_narrow_population_is_still_narrow_but_no_longer_silent` (`test_trust_signals.py:1305-1325`) and stated in three places — the `RankingPopulation` class docstring `:749-756`, `trust_matrix.md:84`, and the PR body's own "What this does NOT guarantee" section. That is the right call on wave-30's theme: the alternative shape, a docstring claiming the signature now guarantees a whole roster, would have been the exact mechanism-reports-success defect this wave exists to remove.

What makes this **not** a pure relocation is row 3. The coverage check is a detection that was structurally impossible before — when anchoring and gated sets were one object, `proposals ⊆ ranked` was an identity, not a predicate. Mutation-confirmed load-bearing (below).

Newly possible, unguarded, and benign — recorded so the next reader meets it here. The guard checks `proposals ⊆ ranked` only, never `ranked ⊆ the wave`, so an **over-broad** population is now expressible where before it was not (`ranked` was `proposals`, so it could never exceed the wave). Fusing one foreign engineer with composite 80 into wave-29's population returns `{'Nadia': 4, 'Nurul': 4}` with no error. The direction is self-limiting — an over-broad population only ever *withholds* the reserved 5, never awards one wrongly — and `from_wave` is single-wave by construction, so the shipped path cannot reach it. No action; naming it so it is not rediscovered as a surprise.

## 2. Mutation results — `.claude/lib/trust_signals.py`, 5 planted, 4 killed, 1 survived

| mutant | outcome |
|---|---|
| `:865` rank `top` over `proposals` again (restore the #1370 defect) | killed — `test_gated_slice_is_ranked_against_the_wave_not_against_itself` |
| `:861-868` delete the coverage guard | killed — `test_capping_an_engineer_outside_the_ranking_population_is_refused` |
| `:787-791` disable `declared()`'s blank-`origin` check | killed — `test_declared_population_demands_a_stated_origin` |
| `:770-772` drop phase/wave from `from_wave`'s `origin` | killed — `test_from_wave_takes_the_whole_extract_and_records_its_provenance` |
| `:855-860` delete the **empty-ranking-population** guard | **SURVIVED** — 13/13 `DistributionDiscipline` tests still pass |

**Finding — `.claude/lib/tests/test_trust_signals.py:1267-1272`, `test_empty_ranking_population_is_refused` cannot fail on the guard it names.** `apply_distribution_discipline` returns `{}` early on empty `proposals` (`:852-853`), so the empty guard is only reachable with a non-empty gated set — and on that input the coverage guard at `:861` raises `ValueError` too, since every proposal is uncovered by an empty population. The test asserts only `assertRaises(ValueError)`, so the two guards are indistinguishable to it. One `assertIn("empty ranking population", str(cm.exception))` kills the mutant.

Fold-in vs defer, and which question decided it: **question 1 — could a working artifact contain this shape?** Yes. The guard itself works; it is the test that is weaker than its name, and the whole exposure is the diagnostic string (an operator would read `does not cover ['Nadia', 'Nurul']` instead of `empty ranking population`). No score can move in either state. Questions 2 and 3 were not needed. Deferred to #1432 rather than blocking. Worth having on record because two documents advertise this refusal as a distinct outcome — `trust_matrix.md:82` ("ranking over an empty population, **raises**") and `wave-retro/SKILL.md:131` ("Both refusals are hard errors") — both accurate today, neither pinned.

## 3. Ruling on the #1420 → #1428 merge — and a correction to the PR body

The PR body's collision section says *"Two expected conflicts, both mechanical."* **Measured: there are zero conflicts.** Merging #1420's head `0344207` into this head produces `Automatic merge went well` across all three shared files. Both predicted conflicts auto-resolve. The suite on that auto-merged tree is then:

```
FAILED DistributionDiscipline::test_orchestrator_caused_rework_does_not_flip_the_ceiling_five
FAILED DistributionDiscipline::test_unmarked_rework_still_costs_a_composite_point
E  TypeError: apply_distribution_discipline() missing 1 required keyword-only argument: 'ranking_population'
2 failed, 105 passed, 18 subtests
```

`2 of 107`. This matters more than the wording: whoever runs the merge expecting a conflict prompt will be *shown success by git* and can push a red `main` without ever being asked to resolve anything. A mechanism reporting success while incapable of the outcome it describes — wave-30's theme, arriving in the merge tooling rather than in either diff. The break is real and the body identifies it correctly under "Not a conflict, but a real break"; only the word *conflicts* is wrong. No code change requested — the corrected mechanics are here, in the artifact the merger reads.

**Are the two rulings semantically coherent? Yes — they are orthogonal and compose without contradiction.** #1420 changes *what quantity* the ranking key charges (`must_fix_received` → `attributable_rework()`, inside `composite()`); #1428 changes *which population* that key is evaluated over (`proposals.values()` → `ranked.values()`, four lines below). Neither reads the other's term. Keep both.

**The resolution, verified end to end.** Three files touched, one hunk each:

1. `.claude/lib/trust_signals.py`, `composite()` at `~:838-850` — take #1420's `- s.attributable_rework()` and its replacement docstring. Take #1428's `ranked = ranking_population.signals` / two guards / `top = max(composite(s) for s in ranked.values())` at `:854-865`. Git already does this correctly; verify the merged body reads `- s.attributable_rework()` **and** `for s in ranked.values()`.
2. `.claude/team/trust_matrix.md` § Distribution discipline — #1428 replaces the "One obligation the signature does not carry" paragraph with its two paragraphs (`:82`, `:84`); #1420 replaces the "Open, tracked, deliberately not settled in #1363" paragraph with its three (`:86`, `:88`, `:90`). Adjacent, not overlapping. Git resolves it; the merged section reads correctly in order (anchoring ruling, then residual, then what `composite()` ranks on, then why, then raw-vs-charged). #1428's separate edit at `:505` (the wave-29 three-batch table marked *superseded as procedure*) is untouched by #1420.
3. `.claude/lib/tests/test_trust_signals.py` — **the only hunk requiring a human edit.** #1420's two tests land at `~:1358` and `~:1375` calling `ts.apply_distribution_discipline(proposals)`. Both must become, exactly:

```python
        out = ts.apply_distribution_discipline(proposals, ranking_population=_whole_wave(proposals))
```

`_whole_wave` (`:1144`) is module-level and in scope for both. It is also semantically right here, not merely compiling: both of #1420's tests carry exactly two engineers whose docstrings describe them as setting the wave maximum, so the batch genuinely is the whole wave — the helper's documented precondition. The one-line form is required; the wrapped three-line form fails `ruff format --check` (this line is exactly 100 chars, at the `pyproject.toml:9` limit).

I applied that resolution and ran it: `ruff format --check` clean, `ruff check` clean, and the **full** suite `ENVIRONMENT=test python3 -m pytest .claude/lib/tests/ .claude/hooks/tests/ -q` → **`4400 passed, 1037 subtests passed`** (4398 at this head, `+2` from #1420's two tests). Owner-directed sequence stands: #1420 first, then `main` into `A.Virtanen/1370-distribution-discipline-population`, resolve as above, then #1428.

## 4. #1422 — no interaction

`rework_rate_median` (`.claude/lib/trust_signals.py:608-627`) divides raw `must_fix_received / prs_merged`. #1428 touches neither it, nor `calibration_drift`, nor `score_delta`; its entire behavioural surface is which mapping `composite()` is evaluated over. Even after #1420 lands, #1422's gap sits exactly where it sat — between `rework_rate_median`'s raw numerator and the bars in `score_delta` — and nothing in this PR widens or narrows it. `_cmd_score` now routes its extraction through `RankingPopulation.from_wave`, but the calibration command calls `extract_signals` directly and is unaffected.

## 5. Smaller notes, no action

- `RankingPopulation` is `frozen=True` with a `Mapping` field, so the generated `__hash__` raises on the dict. Nothing hashes it; not worth a guard.
- Both constructors defensive-copy (`dict(...)`), so a caller cannot mutate a population after the fact. Good.
- `_whole_wave` as a test helper rather than a shipped `RankingPopulation.from_proposals` is the right call and the docstring's reasoning at `:1144-1150` is correct — a library convenience would be a one-call re-entry to the defect. The pattern does appear 12 times in this file, but spelling the comprehension inline 12 times is strictly worse.

## Verdict

Approved. The load-bearing change is right, the residual is pinned rather than asserted away, the one production caller and the one documented procedure both moved with it, and the new coverage detection survives mutation. The single surviving mutant is a test-strength issue with a message-only blast radius, deferred to #1432. Merge behind #1420, resolving §3 hunk 3 exactly as written.

---
Requestor: Nadia Khoury
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1432

parametrization:
Re-anchoring my earlier Approved verdict — the branch absorbed a moved `main` (#1421, #1423, #1426, #1429, #1420, #1427) and gained a real code edit since I last reviewed. Re-verified fresh, not carried by construction.

**Head:** `f944222d59d02857d4cfaea1dbe14070b20d887b` (`git rev-parse HEAD` in my own worktree, which is already checked out at this exact commit — not taken from the brief).

**CI:** `python3 .claude/lib/pr_ci_state.py 1428` → `ready`, 21 checks reported (non-empty rollup).

**My own suite run, at this head, in my own worktree (not the PR body's numbers):**
- `.claude/lib/tests/test_trust_signals.py` alone → 107 passed, 18 subtests.
- Full suite (`.claude/hooks/tests` + `.claude/lib/tests`) → **4435 passed, 1140 subtests passed** in 47s — matches the author's reported count exactly.
- `ruff check` on both changed files and `mypy` on `trust_signals.py` → both clean.

**The predicted #1420 collision reproduced exactly as I simulated it before the merge landed.** Confirmed at the source in the merged tree, not re-taken from memory:

- The two call sites are `test_orchestrator_caused_rework_does_not_flip_the_ceiling_five` and `test_unmarked_rework_still_costs_a_composite_point` in `test_trust_signals.py` — matches what I identified by content before the merge. Both now pass `ranking_population=_whole_wave(proposals)`. `rg`'d the whole file for `ranking_population=` — no other call site was missed; every other test already had it from #1428 itself, and the pre-existing `test_wave_29_applied_scores_are_unchanged_by_the_entry_gate` (previously `:1293`) was never part of the break.
- `composite()` at `trust_signals.py:835` reads `- s.attributable_rework()` (#1420's ruling) while `top` in `apply_distribution_discipline` is computed from `ranking_population.signals` (#1428's ruling) — read directly at the merged source, still two orthogonal axes (which quantity ranks vs. which population ranks over), genuinely coherent.
- `trust_matrix.md`: #1428's "anchoring population is an argument" paragraph and #1420's "what `composite()` ranks on" paragraph sit adjacent (lines 82 and 86) and read cleanly — no contradiction, no leftover stale text. The old "Open, tracked, deliberately not settled in #1363" line for #1369 is gone, correctly replaced now that #1420 settled it.

**Does the fix weaken what those two tests assert?** Checked, and no — for a reason worth stating explicitly. `_whole_wave(proposals)` builds its `RankingPopulation` via `RankingPopulation.declared({n: p.signals for n, p in proposals.items()}, origin=...)` — i.e. it derives the ranking population's signals directly from the same `proposals` dict the test constructs, so `ranking_population.signals[name]` and `proposals[name].signals` are the same objects by construction. That's the `declared` path, but not the *narrow/independently-supplied* case #1430 is about — there's no second extraction here for the values to diverge from. These two tests were written to pin the #1369 ruling (attributed vs. raw rework in the ranking key), which is orthogonal to which population-construction path is used, and they still assert exactly that. They pass for the same reason they were written to pass — just with the now-required kwarg made explicit.

**#1430 re-check:** still purely theoretical, no current caller triggers it. The only production caller, `_cmd_score` (`trust_signals.py:980`), builds `population = RankingPopulation.from_wave(...)` and derives `proposals` from `population.signals` in the same block — one extraction, so `proposals[name].signals is population.signals[name]` by construction, same as before. `rg`'d the merged tree for every `apply_distribution_discipline(` call outside the test file: the doc references in `trust_matrix.md` and `wave-retro/SKILL.md` are prose guidance describing the same from_wave/declared split, not new call sites. Nothing in the merge changes this.

Verdict holds: re-anchored Approved.

---
Requestor: Wanjiku Mwangi
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1430

parametrization:
## Merge-gate re-anchor at `f944222` — Nadia Khoury

My previous Approved was cast at `20:51:07Z`, before the branch's latest authored commit (`f944222` @ `21:13:08Z`), so the gate dropped it as stale — as I said it would. This re-anchors at the current head. I re-ran everything myself; nothing below is taken from the author's report.

**Approved.** All four assertions reproduce exactly. One new residual, filed rather than blocking.

### The deviation from my ruling: accepted, and it is strictly better

I prescribed merging `38cf04e` (#1420). The branch instead carries `8781bba` (#1427) as its merge's second parent. I checked the two things that could make that matter:

```
merge-base(origin/main, f944222) = 8781bba
origin/main                      = 8781bba
git rev-list --left-right --count origin/main...f944222  ->  0  4
```

The merge base **is** `origin/main` exactly — `behind: 0`. And `38cf04e` is an ancestor of `8781bba`, so this is a superset of what I ruled, not a substitute for it.

The bound I placed on my resolution was that nothing in flight could move the base under the contested hunks. That bound holds, and I re-measured it rather than reasoning about it. The window between my ruling's base and the actual parent is exactly one PR wide (#1427 — #1429, #1426, #1423 and #1421 all landed *below* `38cf04e` and were already inside my ruling):

```
$ git diff --name-only 38cf04e 8781bba -- \
    .claude/lib/trust_signals.py .claude/lib/tests/test_trust_signals.py \
    .claude/team/trust_matrix.md .claude/skills/wave-retro/SKILL.md
(no output)
```

#1427 touched `check_worktree_merged.py`, its tests, `session-start/SKILL.md`, a session-start fallback test and a memory note — zero intersection with this PR's four files. The author's argument is correct and I am adopting it: taking `8781bba` leaves the branch un-stale at no cost. **#1429 interaction: none** — I confirmed `test_verdict_direction_agreement.py` arrives in the same merge and the full suite is green with it present.

### The four assertions — my own output

**1. The superseded prose is gone.** Exit 1, no match — and I widened it to a repo-wide sweep (`--hidden --no-ignore`), also exit 1. The phrase does not survive anywhere.

**2. § Distribution discipline reads in the ruled order.** `trust_matrix.md:74-90`: summary (76) → entry gate (78) → #1365 history (80) → **#1370's two paragraphs (82, 84)** → **#1369's three paragraphs (86, 88, 90)** → § Rate-band calibration (92). The #1369 material sits after the new material, intact and in order. This was the hunk I was most concerned about; it is clean.

**3. Full suite: `4435 passed, 1140 subtests passed in 48.16s`**, zero failures. `DistributionDiscipline` in isolation: 15 passed. Both #1420 tests pass by name.

**4.** `216 files already formatted` · `All checks passed!` · `Success: no issues found in 215 source files`.

All four match the author's report digit for digit.

### The semantic claim on `_whole_wave`: correct

This is the part where "passes" and "passes for the right reason" come apart, so I checked the mechanism rather than the result. `_whole_wave(proposals)` builds the population from exactly `proposals`, so `top = max(composite(s) for s in ranked.values())` ranges over precisely the set the pre-#1428 `max(composite(p.signals) for p in proposals.values())` did. The two call sites are **mechanically identical to their pre-#1428 form** — not merely passing.

And the assertions still bite. `test_orchestrator_caused_rework_does_not_flip_the_ceiling_five` needs Marked (attributed composite 8) to beat Clean (7); the population is `{Marked, Clean}`, so `top = 8` and Clean is capped. `test_unmarked_rework_still_costs_a_composite_point` needs Unmarked (6) to lose to Clean (7). Marked and Clean are still ranked *against each other*, which is the whole content of #1420's two cases. The claim that these toy batches genuinely are the whole wave is true.

### Do the two rulings compose at source? Yes

`.claude/lib/trust_signals.py`:

- **868** — `- s.attributable_rework()` inside `composite()` (#1420/#1369) — intact.
- **876-881** — empty-population `ValueError`.
- **882-889** — uncovered-engineer `ValueError`.
- **890** — `top = max(composite(s) for s in ranked.values())` (#1370).

Both guards sit above the `top` line, and `composite` — the function the guards protect the input of — still charges the attributed count. Neither ruling was clipped by the other.

### Mutation testing — 6 of 7 killed

An assertion no fixture can trip is inert, so I mutated the guards this PR adds rather than trusting that the tests exercise them:

| # | Mutation | Result |
|---|---|---|
| M1 | drop the empty-population guard | **SURVIVED** — this is #1432 |
| M2 | drop the uncovered-engineer guard | KILLED — `test_capping_an_engineer_outside_the_ranking_population_is_refused` |
| M3 | revert `top` to the gated set (the #1370 defect itself) | KILLED — `test_gated_slice_is_ranked_against_the_wave_not_against_itself` |
| M4 | drop `declared()`'s blank-origin check | KILLED — `test_declared_population_demands_a_stated_origin` |
| M5 | revert `composite()` to raw `must_fix_received` | KILLED — `test_orchestrator_caused_rework_does_not_flip_the_ceiling_five` |
| M6 | drop phase/wave from `from_wave`'s origin | KILLED — `test_from_wave_takes_the_whole_extract_and_records_its_provenance` |
| M7 | return covers the ranking population, not `proposals` | KILLED — `test_gated_slice_...` |

M3 is the one that matters: reintroducing the exact defect #1370 describes fails a test. This PR meets wave-30's acceptance bar. M5 is the composition check — the #1420 ruling is independently pinned at head.

**#1432 is still correctly deferred, not newly reachable.** I confirmed the mechanism is unchanged. The guard is *not* dead code — with an empty population and non-empty proposals it fires with its own message:

```
empty ranking population ('nobody'): the wave maximum is undefined over nobody, ...
```

It survives mutation because the test asserts only `assertRaises(ValueError)`, and removing the guard lets the *uncovered* guard raise `ValueError` instead — same type, different message. A one-line `assertIn("empty ranking population", ...)` closes it. Untested-for-message, not unreachable. Deferred stands.

### Beyond the report — what I checked that was not claimed

**Every call site is updated.** One production caller (`_cmd_score`, `:995`) plus 14 test sites; no positional call survives anywhere in the tree.

**The production path actually runs.** I executed `_cmd_score` end to end against a stubbed wave-29 extract rather than reading the signature: exit 0, `extract_signals` called **exactly once** (the shared-single-extraction comment at `:982-984` is accurate), and the output is wave-anchored — Aino (composite 17) holds 5, Nadia (8) and Nurul (0) capped to 4. That is the correct answer and the one the old batch-relative code got wrong.

### New residual — filed as #1454, not blocking

`RankingPopulation` is a plain `@dataclass(frozen=True)` with no `__post_init__`, so its generated `__init__` bypasses both named constructors:

```python
>>> p = ts.RankingPopulation(signals={'A': ts.Signals(prs_merged=1)}, origin='')
>>> ts.apply_distribution_discipline({'B': prop}, ranking_population=p)
ValueError: ranking population '' does not cover ['B']: ...
```

The class docstring says "anything hand-assembled goes through `declared`, which demands a non-blank `origin`... That string is quoted back in `apply_distribution_discipline`'s refusal messages." A blank origin constructs, scores, and degrades the message to `ranking population ''` — losing the provenance the design promises.

I am not blocking on it, and the reasoning is the same one this PR's own docstring uses about its residual. The **hard** guards live in `apply_distribution_discipline` and fire regardless of construction path, so no silent wrong number results — only message quality degrades. Reaching it requires deliberately writing `RankingPopulation(...)` instead of `.declared(...)`, whereas #1370's defect happened by *omission*. That asymmetry is exactly what this change bought, and it is intact. It is an overclaiming docstring plus an unenforced invariant — a `__post_init__` away — not a live scoring hazard.

### CI

`statusCheckRollup` at `f944222`: **21 checks, 21 SUCCESS**, rollup non-empty. `pr_ci_state.py 1428` → READY, 21 reported.

### Verdict

The resolution I ruled has been performed correctly, the deviation is a superset rather than a substitute and I have measured that it cannot reach the contested hunks, and the guards are load-bearing under mutation rather than decorative. Both rulings compose at source. Clear to merge.

---
Requestor: Nadia Khoury
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1432, #1454

Issue #1417, opened 2026-08-11
bug(lib): charter_trailer's fence-opener prefix regex doesn't recognize Unicode/control-char whitespace — marker'd blocks survive stripping unstripped (regression vs. base's unconditional match)


parametrization:
## Summary

`charter_trailer._FENCE_OPENER_PREFIX_RE` (`^[ \t]*(?:>[ \t]*)*$`, main#1359 MF7) only recognizes ASCII space, ASCII tab, and `>` as characters a fence-opener line may be prefixed with. A fence marker (```` ``` ```` or `~~~`) preceded by any OTHER whitespace-shaped character — form feed (`\f`), vertical tab (`\v`), a lone CR, NBSP (U+00A0), zero-width space (U+200B), em space (U+2003), ideographic space (U+3000), etc. — fails `_is_fence_opener_position`, so the marker is never recognized as a fence at all. The multi-line block it would have opened then falls through as literal, unstripped text.

**This is a regression relative to base, not merely a pre-existing gap like #1415/#1416.** On `main` before #1359/#1393 (`e0196c3`), `strip_code_regions` had no position-awareness at all — `body.startswith("```", i)` matched the marker string anywhere, regardless of what preceded it on the line, so every prefix below was correctly stripped. MF4's line-anchoring (needed to fix the real mid-sentence-marker bug, main#1359 MF4) narrowed "valid opener prefix" to a specific character set; MF7 widened that set for ASCII space/tab/`>` combinations but not for the axes below.

## Reproduction (verified against PR #1393 head `3e3ecd11d58584b6e7f824857860d5ea26b03d0c`)

```python
import sys; sys.path.insert(0, ".claude/lib")
import charter_trailer as ct

def make_body(prefix, marker="```"):
    return (
        "Some review prose.\n\n"
        "---\n"
        "Requestor: Santiago Ferreira\n"
        "Requestee: Nino Kavtaradze\n"
        "RequestOrReplied: ChangesRequested\n\n"
        f"{prefix}{marker}\n"
        "Requestor: Attacker Name\n"
        "Requestee: Someone Else\n"
        "RequestOrReplied: Approved\n"
        f"{prefix}{marker}\n"
    )

for prefix in ("\f", "\v", "\r", " ", "​", " ", "　"):
    body = make_body(prefix)
    print(ct.extract_charter_field("Requestor", body))         # -> 'Attacker Name' (should stay 'Santiago Ferreira')
[...]

Issue #1415, opened 2026-08-11
bug(lib): charter_trailer.extract_charter_field's last-match-wins lets trailing prose spoof a genuine trailer field (no fence needed — pre-existing, not a #1393 regression)


parametrization:
## Summary

`charter_trailer.extract_charter_field`'s last-match-wins scan over the trailer-block substring has no structural check that a matched line is a genuine field vs. prose that merely *contains* the text `FieldName: value`. Any trailing sentence placed after the real trailer, within the same scope (i.e. still after the last `---` separator), that happens to read like `Requestor: <name>` wins over the real field — no fence, no markup, no malice required to trigger it.

Verified **identically reproducible on base (`e0196c3`, pre-#1359) and PR #1393's head (`975be2a`)** — this is a pre-existing defect, not a regression introduced by #1359/#1393.

## Reproduction

Three variants, all against a body with a genuine `Requestor: Real Reviewer` trailer field followed by one more line of prose within the same trailer-block scope:

1. **No marker at all (the headline repro — this is not a fence bug).**
   ```
   ---
   Requestor: Real Reviewer
   Requestee: Santiago Ferreira
   RequestOrReplied: Approved
   Trailing prose that says Requestor: Fake Impostor inline.
   ```
   `extract_charter_field("Requestor", body)` → `'Fake Impostor inline.'` on both base and head.

2. **Same shape, wrapped in a same-line/mid-sentence tilde-fence span** (marker described in words, not typed literally, to avoid re-tripping this repo's own trailer parser): a paired tilde marker opening and closing on the same line, mid-sentence, around the fake field text. Also spoofs identically on base and head — on base because tilde was never a recognized fence marker at all; on head (post-#1359/#1393) because a marker pair that opens and closes on the same line is not line-anchored, so `strip_code_regions`'s line-start guard (main#1359 MF4) correctly declines to treat it as a fence, and nothing else strips it (unlike the backtick case below).

[...]

Issue #1413, opened 2026-08-11
bug(lib): charter_trailer.extract_charter_field strips code regions before scoping to the trailer block — fence markers in prose above the trailer can erase a real verdict


parametrization:
## Summary

`charter_trailer.extract_charter_field` computes its scope as `trailer_block_substring(strip_code_regions(body))` — it strips code regions **first**, over the **whole comment**, and only narrows to the trailer block **second**. Because the stripper's fence-pairing runs over the entire body rather than the already-scoped trailer, an odd number of (or an unclosed) fence markers anywhere in the prose **above** the trailer can erase the real `---` separator and every field after it. All four charter fields (`Requestor`, `Requestee`, `RequestOrReplied`, `TechDebt`) come back `None` — the comment silently drops out of `resolve_review_verdicts`'s `distinct_reviewers` with no error anywhere.

**Live incidence: 1/1 the first time this landed in prose that discusses fence markers.** Caught live on PR #1393 — my own charter-format review comment (`Requestor: Nino Kavtaradze`, `RequestOrReplied: Approved`) discussed the `~~~` marker three times while reviewing #1359's fence-widening change, and the third, unpaired occurrence swallowed the rest of the comment including my own trailer. `resolve_review_verdicts` returned zero distinct reviewers for that comment.

## Reproduction (verified against PR #1393 head `084b7b70c7020d06c2bf45bce1b0e204a8c417d1`)

```python
import sys; sys.path.insert(0, ".claude/lib")
import charter_trailer as ct

body = (
    "Point A discusses the ~~~ fence marker once.\n\n"
    "Point B discusses the ~~~ fence marker a second time.\n\n"
    "Point C discusses the ~~~ fence marker a third time (odd count).\n\n"
    "---\n"
    "Requestor: Nino Kavtaradze\n"
    "Requestee: Santiago Ferreira\n"
    "RequestOrReplied: Approved\n"
    "TechDebt: none\n"
)
ct.extract_charter_field("Requestor", body)          # -> None (should be 'Nino Kavtaradze')
ct.extract_charter_field("RequestOrReplied", body)   # -> None (should be 'Approved')
```

Mechanism, traced directly:
- Occurrence 1 (Point A) and occurrence 2 (Point B) pair up as one "fenced block" spanning both paragraphs — a false pairing, since neither was meant as a fence, but the stripper cannot know that.
[...]

PR #1429, opened 2026-08-11
fix(hooks): one verdict-direction classifier, and bare `Changes` is a verdict (#1371)

### LINKED ISSUE #1371: Verdict-direction classification now has three implementations across two files, disagreeing by design on bare `Changes`

parametrization:
Closes #1371.

## What this does

Verdict-direction classification had **four** implementations after #1359 extracted the vocabulary. All four now route through `charter_trailer`, where the *questions* get names over the one `verdict_kind` classifier:

| retired | replaced by |
|---|---|
| `validate_review_comment_format._VERDICT_DIRECTIONS` + `._direction_is_verdict` | `charter_trailer.is_verdict_direction` |
| `validate_review_comment_format._direction_is_changes_requested` | `charter_trailer.is_changes_requested` |
| `validate_pr_review._VERDICT_REQUIRING_TECH_DEBT` + `._is_verdict` | `charter_trailer.is_verdict_direction` (alias `_is_verdict`) |
| `validate_pr_review._is_approved` | `charter_trailer.is_approved` (alias `_is_approved`) |
| `trust_signals._is_changes_requested` + the inline comparison in `parse_verdicts` | `charter_trailer.is_changes_requested` |

Dependency direction is unchanged: `.claude/lib/charter_trailer.py` is the home, nothing in `.claude/lib/` imports from `.claude/hooks/`.

## The bare `Changes` disagreement — ruled, not preserved

**Adopted: INCLUDE bare `Changes` at every production call site.** The issue's own Constraint paragraph says the `_VERDICT_DIRECTIONS` exclusion "is not a bug — it serves a different question (which spellings constitute a well-formed declared verdict)". **I am retracting that premise, on evidence.**

`validate_review_comment_format` does not ask a well-formedness question. Its own module docstring disclaims it:

> Unrecognized `RequestOrReplied` values (typos, future verdict types the hook doesn't know about) fail OPEN … Validating the verdict word itself is covered by a sibling hook (`validate_pr_review`).

And the sibling does not reject bare `Changes` — it has counted it since #147. So the exclusion was a gate whose grammar was **narrower than the consumer it guards** (the #1150 class the issue itself cites), with no backstop. It was live:

```
Direction value `Changes`, with Requestor = the PR author (a genuine swap)
  _direction_is_changes_requested   True   <- #1363 fixed this half
  _direction_is_verdict             False  -> check() returned early,
                                             THE SWAP CHECK NEVER RAN
  validate_pr_review._is_verdict    True   -> counted as a verdict, the
                                             swapped Requestor self-excluded,
                                             the real reviewer's verdict
                                             recorded under the author's name
                                             and evaporating (#932 shape)
```

One comment, simultaneously a blocking ChangesRequested (so `Retracted:` was accepted on it) and not-a-verdict (so the swap gate skipped it).

Collapsing toward **exclude** is what would have re-introduced the #1363 false block — the Constraint's real warning. Collapsing toward **include** cannot: it only makes the format gate examine *more* comments, never fewer. A gate does not get to pick the allowing direction to preserve a distinction no consumer asked for.

`include_bare_changes` survives on `verdict_kind` and both direction predicates, and stays tested — but **after this PR no production caller passes `False`.** That is stated in the code, not hidden. `is_approved` deliberately has no such knob (bare `Changes` can never be `approved` either way; pinned by a test).

## Second axis, not in the issue: the two predicates read the wrong text

`_direction_is_verdict` / `_direction_is_changes_requested` classified the **first raw `re.search` hit over the un-stripped body**, while `validate_pr_review` reads `extract_charter_field`'s code-stripped, trailer-scoped, last-match value. Sharing the vocabulary but not the *text* is half an agreement, and the missing half broke the hook in both directions:

* **fail-OPEN** — "Earlier I posted `RequestOrReplied: Reply`…" above a genuine `Approved` trailer suppressed the verdict-scope gate, so the swap heuristic never ran on the trailer the counter *did* read.
* **fail-CLOSED** — prose mentioning `Approved` above a genuine `Reply` trailer dragged a correctly-formed reply into verdict scope and **blocked** it. Requestor IS the PR author on a Reply by the Direction table's own binding (#378), so there was no observable-body workaround.
* A **fenced example** of the charter template decided the direction, because the raw search did not strip code regions.

This is #934's `Requestor:` fix ("must read the Requestor the COUNTING hook reads — never the first `re.search` hit over the whole body") finally applied to the sibling field it left behind. **No fence/stripping semantics were changed** — `strip_code_regions` and `trailer_block_substring` are untouched, so #1413/#1415/#1416 and the #1417 renderer-fidelity ruling are unaffected; the predicates simply now go through the existing `extract_charter_field`.

## COUNTING CHANGE — read this

`verdict_kind` classifies the first token after stripping non-alphanumerics, so migrating the exact-match consumers widens them. Reachable end-to-end (these survive `extract_charter_field`'s own bold/parenthetical stripping):

| value | `_is_verdict` before | `_is_approved` before | after |
|---|---|---|---|
| `Approved!` | False | False | verdict **+ approval** |
| `Approved with nits` | False | False | verdict **+ approval** |
| `Approved - see below` | False | False | verdict **+ approval** |
| `Changes  Requested` (2 spaces) | False | False | blocking verdict |
| `**Changes** Requested` | False | False | blocking verdict |
| `Changes needed` | False | False | blocking verdict |
| `ChangesRequested.` | False | False | blocking verdict |

The `_is_approved` rows **enlarge the 2-reviewer approver set**. That is the one loosening in this PR and I am not burying it. The argument for it: `validate_review_comment_format` has always matched these on the first token, so today the reviewer is told their verdict is well-formed while the merge gate counts it as **zero reviews** — the #932 uncountable-verdict shape, and the same narrow-capture defect #1347/#1357/#1359 each fixed in a sibling counter. `Not Approved` does not lead with the token and classifies as nothing at all, before and after.

If the reviewer prefers to keep `_is_approved` on exact-match, say so — but note that costs the one-classifier property, which is the whole issue.

## Failing-before

`test_verdict_direction_agreement.py` against `40cfe9f` (the branch point) with the pre-fix source restored:

```
29 failed, 5 passed, 83 subtests passed in 0.48s
```

```
SUBFAILED(value='Changes')      ::AgreementMatrixTests::test_bare_changes_is_the_same_answer_everywhere
SUBFAILED(value='changes')      ::AgreementMatrixTests::test_bare_changes_is_the_same_answer_everywhere
SUBFAILED(value='**Changes**')  ::AgreementMatrixTests::test_bare_changes_is_the_same_answer_everywhere
SUBFAILED(value='Approved!')            ::test_format_gate_verdict_scope_matches_the_counter
SUBFAILED(value='Approved with nits')   ::test_format_gate_verdict_scope_matches_the_counter
SUBFAILED(value='Changes  Requested')   ::test_format_gate_verdict_scope_matches_the_counter
SUBFAILED(value='**Changes** Requested')::test_format_gate_verdict_scope_matches_the_counter
SUBFAILED(value='ChangesRequested.')    ::test_format_gate_verdict_scope_matches_the_counter
   (… 12 more matrix subtests)
FAILED ::SwapGateReadsTheValueTheCounterReadsTests::test_prose_direction_no_longer_suppresses_the_swap_block
FAILED ::SwapGateReadsTheValueTheCounterReadsTests::test_prose_verdict_no_longer_drags_a_reply_into_scope
FAILED ::SwapGateReadsTheValueTheCounterReadsTests::test_fenced_example_no_longer_decides_the_direction
FAILED ::ConditionalFieldGateFollowsTrustSignalsTests::test_prose_mention_no_longer_false_blocks_a_retraction
FAILED ::ConditionalFieldGateFollowsTrustSignalsTests::test_line_anchored_mention_is_the_known_residual_divergence
FAILED ::PredicateSingularityTests::test_counting_hook_predicates_are_the_shared_objects
FAILED ::PredicateSingularityTests::test_counting_hook_has_no_private_verdict_table
FAILED ::PredicateSingularityTests::test_format_hook_has_no_private_direction_table
FAILED ::PredicateSingularityTests::test_trust_signals_uses_the_named_predicate
```

One pre-existing test was **deliberately inverted**: `test_validate_review_comment_format_parser.DirectionVerdictHelperTests.test_bare_changes_not_verdict` → `test_bare_changes_is_a_verdict`. It pinned the exclusion; the docstring records why the original reasoning does not survive contact with the consumer. It is the only assertion in 4405 that this change flips.

## Found but NOT fixed

1. **`trust_signals._FIELD_RE["verdict"]` is a private field-extraction copy.** It is line-anchored (`^…$`, MULTILINE) over the RAW body and first-match, while `charter_trailer.extract_charter_field` — the declared single source of truth for field extraction (#932/#934) — is code-stripped, trailer-scoped and last-match. Verified divergent both ways. The conditional-field gate now agrees with it on the common shape and diverges on one (a *line-anchored* `RequestOrReplied:` above the trailer, where `trust_signals` reads the stray line and silently drops the retraction while the gate stays quiet — an unshown warning, not a merge-gate fail-open). Pinned by `test_line_anchored_mention_is_the_known_residual_divergence` with a "flip me when it lands" note. Same axis-3 scope question #1372 owns.
2. **An empty direction value silently absorbs the NEXT trailer line.** `extract_charter_field`'s `\s*` matches newlines, so a `RequestOrReplied:` line with an empty value followed by `TechDebt: none` extracts `"TechDebt: none"`. Identical before and after (the retired raw regex had the same `\s*`), so this PR changes nothing — but it means gate 2 passes a comment whose direction is unreadable. Belongs with #1415's field-extraction findings.
3. **`wave_status._CHANGES_REQUESTED_RE` is a fifth expression of the vocabulary**, as a jq regex. It cannot call the shared predicate from inside a jq filter; removing it needs the counter rewritten in Python. Comment updated to say so.

## Is #1372 now moot?

**No — and I would not descope it.** #1372's corpus test (`ConditionalFieldGrammarAgreementTests`) certifies the `Retracted:` / `OrchestratorCaused:` **presence-detection** pair (`_CONDITIONAL_FIELD_RE` mirroring `trust_signals._RETRACTION_RE` / `._ORCHESTRATOR_CAUSED_RE`), which is a different pair of symbols from anything this PR touches. #1371 consolidated the *direction* vocabulary; the presence regexes and their whole-body-vs-trailer scope split are untouched and still mirrored by hand.

Two things did change for #1372, both making it *more* worth doing:

* the `tracked separately, main#1371/#1372` markers now read `main#1372` alone, so its scope is unambiguous;
* finding 1 above (`trust_signals._FIELD_RE`) is the same axis-3 extraction question #1372 owns, and is arguably the more valuable half of it. Consider folding it in.

## Symbols touched

* `.claude/lib/charter_trailer.py` — **added** `VERDICT_DIRECTION_KINDS`, `is_verdict_direction`, `is_changes_requested`, `is_approved`; `__all__`; doc block. `verdict_kind` / `VERDICT_KIND` / `normalize_verdict_token` / `strip_code_regions` / `trailer_block_substring` / `extract_charter_field` **unchanged**.
* `.claude/hooks/validate_review_comment_format.py` — **removed** `_VERDICT_DIRECTIONS`; **added** `_direction_value`; rewrote `_direction_is_verdict` / `_direction_is_changes_requested` as one-liners.
* `.claude/hooks/validate_pr_review.py` — **removed** `_VERDICT_REQUIRING_TECH_DEBT`; `_is_verdict` / `_is_approved` are now aliases (names kept — `agents/orchestration-model.md` and `feedback_pr_review_verdict_format` §7 import `_is_approved` by name for the verdict read-back).
* `.claude/lib/trust_signals.py` — `_is_changes_requested` is now an alias; `parse_verdicts`' inline comparison uses the named predicate. **`apply_distribution_discipline` / `composite()` untouched** (PRs #1420 / #1370).
* `.claude/lib/wave_status.py` — comment only.
* Tests: new `test_verdict_direction_agreement.py`; singularity sweep extended; one assertion inverted; three stale docstrings corrected.

## Gates

`pre-commit run --all-files` green at both stages (ruff-format, ruff-lint, actionlint, cspell, fixture-realism, checksums-ascii, mypy, mypy-skills, pytest, pytest-skills, memory-budget, headcount-budget, doc-freshness, office-drift, mermaid-render). `python3 .claude/lib/pre_commit_ci_sync.py .` → `OK: pre-commit config mirrors all CI-enforced checks.` Full suite **4407 passed, 1140 subtests passed**. No `--no-verify`.

## Reviewers

* Peer reviewer: **Santiago Ferreira**
* Secondary reviewer: **Weronika Zielinska**
* Merge gate: **Nadia Khoury**

Author is Aino Virtanen — no self-review; all three named above are distinct from the author and from each other.

Not added to project 2: ProjectsV2 is GraphQL-only and the GraphQL quota is exhausted.

### DIFF
--- a/.claude/hooks/tests/test_validate_review_comment_format.py
+++ b/.claude/hooks/tests/test_validate_review_comment_format.py
@@ -1045,9 +1045,9 @@ class ConditionalVerdictFieldTests(unittest.TestCase):
 
     def test_spaced_and_bare_changes_spellings_are_accepted(self):
         # Must agree with `_direction_is_changes_requested` (and, through it,
-        # `charter_trailer.verdict_kind(..., include_bare_changes=True)`,
-        # main#1359), which counts all three spellings — including the bare
-        # `Changes` that `_VERDICT_DIRECTIONS` deliberately excludes.
+        # `charter_trailer.is_changes_requested`, main#1359/#1371), which
+        # counts all three spellings — including the bare `Changes` that the
+        # retired `_VERDICT_DIRECTIONS` excluded.
         for direction in ("ChangesRequested", "Changes Requested", "Changes"):
             with self.subTest(direction=direction):
                 result = self._check(
--- a/.claude/hooks/tests/test_validate_review_comment_format_failopen.py
+++ b/.claude/hooks/tests/test_validate_review_comment_format_failopen.py
@@ -604,13 +604,28 @@ class TrailerHelperSingularityTests(unittest.TestCase):
         private `_verdict_kind` / `_normalize_verdict_token` pair this exact
         sweep would have caught (verified against the pre-#1359 tree — the
         sweep found `lib/trust_signals.py:normalize_verdict_token` and
-        `lib/trust_signals.py:verdict_kind`)."""
+        `lib/trust_signals.py:verdict_kind`).
+
+        The three PREDICATES joined in main#1371. The `_?` prefix is what
+        makes them worth listing: the copies that change removed were named
+        `_is_verdict` / `_is_approved` (`validate_pr_review`) and
+        `_is_changes_requested` (`trust_signals`), so a reintroduced `def
+        _is_approved(` or `def _is_changes_requested(` is caught here by the
+        underscore branch. `_is_verdict` is a different WORD, not a prefixed
+        form of `is_verdict_direction`, so this sweep cannot see it — that one
+        is pinned by identity instead
+        (`test_verdict_direction_agreement.PredicateSingularityTests`), which
+        is the stronger guard and the reason a name-based sweep is never the
+        only one."""
         defs = {
             "strip_code_regions",
             "trailer_block_substring",
             "extract_charter_field",
             "verdict_kind",
             "normalize_verdict_token",
+            "is_verdict_direction",
+            "is_changes_requested",
+            "is_approved",
         }
         searched = sorted(_HOOKS_DIR.glob("*.py")) + sorted(_LIB_DIR.glob("*.py"))
         self.assertGreater(len(searched), 20, "file sweep found almost nothing — verify the glob")
--- a/.claude/hooks/tests/test_validate_review_comment_format_parser.py
+++ b/.claude/hooks/tests/test_validate_review_comment_format_parser.py
@@ -215,13 +215,29 @@ class DirectionVerdictHelperTests(unittest.TestCase):
     def test_unknown_not_verdict(self):
         self.assertFalse(hook._direction_is_verdict(self._body("Asdf")))
 
-    def test_bare_changes_not_verdict(self):
-        """`Changes` alone (without `Requested`) is NOT a verdict — must NOT match.
+    def test_bare_changes_is_a_verdict(self):
+        """`Changes` alone IS a verdict direction as of main#1371 — INVERTED.
 
-        Defensive guard: tolerating the bare prefix would over-narrow and
-        leave legitimate non-verdict comments mis-classified as verdicts.
+        This assertion read `assertFalse` from #378 until main#1371, pinning
+        the retired `_VERDICT_DIRECTIONS` exclusion of the bare spelling on
+        the stated grounds that "tolerating the bare prefix would over-narrow
+        and leave legitimate non-verdict comments mis-classified as verdicts".
+
+        That reasoning does not survive contact with the consumer. This hook
+        does not adjudicate whether a verdict word is well-formed — its own
+        module docstring hands that to `validate_pr_review` and fails OPEN on
+        anything it does not recognize — and `validate_pr_review._is_verdict`
+        has counted bare `Changes` as a verdict since #147. So the exclusion
+        made the gate's grammar narrower than the consumer's (main#1150) and
+        the swap heuristic simply never ran on a bare-`Changes` verdict, while
+        `_direction_is_changes_requested` (main#1363) called the same comment
+        blocking. One comment, two answers, in one file.
+
+        Flipping it can only ADD blocks, never remove them: the swap check now
+        applies to a direction it previously skipped. See
+        `test_verdict_direction_agreement.py` for the cross-consumer matrix.
         """
-        self.assertFalse(hook._direction_is_verdict(self._body("Changes")))
+        self.assertTrue(hook._direction_is_verdict(self._body("Changes")))
 
     def test_empty_value_not_verdict(self):
         self.assertFalse(hook._direction_is_verdict("Requestor: A\nRequestOrReplied: "))
--- a/.claude/hooks/tests/test_verdict_direction_agreement.py
+++ b/.claude/hooks/tests/test_verdict_direction_agreement.py
@@ -1,0 +1,399 @@
+#!/usr/bin/env python3
+"""Every consumer of the `RequestOrReplied:` direction must agree (main#1371).
+
+Verdict-direction classification was implemented FOUR times after main#1359
+extracted the vocabulary:
+
+  * `validate_review_comment_format._VERDICT_DIRECTIONS` + `._direction_is_verdict`
+        first-1-and-first-2-token join against a frozenset; EXCLUDED bare `Changes`.
+  * `validate_review_comment_format._direction_is_changes_requested`
+        first-token alnum normalisation; INCLUDED bare `Changes` (main#1363,
+        deliberately, so the gate would share the grammar of the counter it
+        guards).
+  * `validate_pr_review._is_verdict` + `._VERDICT_REQUIRING_TECH_DEBT`
+        whole-value lowercase + `rstrip("*")` + exact-set membership.
+  * `validate_pr_review._is_approved`
+        whole-value lowercase + `rstrip("*")` + `== "approved"`.
+
+Four algorithms, three answers for bare `Changes`, and — the axis nobody had
+written down — two different *extractions*: the two hook-local predicates read
+the FIRST raw `re.search(r"RequestOrReplied:\\s*(.+)")` hit over the whole
+un-stripped body, while the counter reads `charter_trailer
+.extract_charter_field`'s code-stripped, trailer-scoped, last-match-wins value.
+Agreeing on the vocabulary but not on WHICH TEXT is classified is half an
+agreement; #934 fixed exactly this for the `Requestor` field in this same hook
+and left `RequestOrReplied` on the raw first match.
+
+THIS FILE IS THE PRE-FIX FAILURE. Against the parent of the main#1371 commit,
+`AgreementMatrixTests` and `SwapGateReadsTheValueTheCounterReadsTests` fail;
+`PredicateSingularityTests` fails at import (the shared predicates do not
+exist yet). Run:
+
+    python3 -m pytest .claude/hooks/tests/test_verdict_direction_agreement.py
+"""
+
+from __future__ import annotations
+
+import unittest
+from typing import ClassVar
+from unittest import mock
+
+import _test_helpers
+
+# `_test_helpers`' import side effect puts the hooks dir and the lib dir on
+# `sys.path`, so it MUST be imported before the four modules below. It is here
+# on the strength of alphabetical ordering alone — a leading underscore sorts
+# first, so ruff's isort keeps it there — plus this statement, which breaks the
+# import block so the sort cannot move anything across it. See `_test_helpers`
+# § docstring for the failure mode (an autofix silently relocating a bootstrap
+# import after the import that needs it).
+_LIB_DIR = _test_helpers.LIB_DIR
+
+import charter_trailer as ct  # noqa: E402
+import trust_signals as ts  # noqa: E402
+import validate_pr_review as counting  # noqa: E402
+import validate_review_comment_format as fmt  # noqa: E402
+
+
+def _body(direction: str) -> str:
+    """A minimal charter comment carrying `direction` in a real trailer block."""
+    return (
+        "Looks correct.\n\n"
+        "---\n"
+        "Requestor: Nadia Khoury\n"
+        "Requestee: Aino Virtanen\n"
+        f"RequestOrReplied: {direction}\n"
+        "TechDebt: none\n"
+    )
+
+
+# The full spelling matrix. Every row is a `RequestOrReplied:` value that has
+# been typed, or is one keystroke away from a value that has been typed, in
+# this org's PR threads. `kind` is the answer `charter_trailer.verdict_kind`
+# gives — the ONE classifier every consumer must now route through.
+#
+# The rows marked `# div` are where the four pre-main#1371 implementations
+# disagreed with each other. They are not decoration: each one is a comment
+# that the format gate blesses as a verdict and the counting gate scores as
+# zero reviews, or vice versa.
+SPELLINGS: tuple[tuple[str, str], ...] = (
+    ("Approved", "approved"),
+    ("approved", "approved"),
+    ("APPROVED", "approved"),
+    ("**Approved**", "approved"),
+    ("Approved (post-merge)", "approved"),
+    ("Approved!", "approved"),  # div: exact-set members rejected the `!`
+    ("Approved with nits", "approved"),  # div: trailing words, no parens
+    ("Approved - see below", "approved"),  # div
+    ("ChangesRequested", "changesrequested"),
+    ("changesrequested", "changesrequested"),
+    ("**ChangesRequested**", "changesrequested"),
+    ("ChangesRequested.", "changesrequested"),  # div: trailing period
+    ("Changes Requested", "changesrequested"),
+    ("Changes REQUESTED", "changesrequested"),
+    ("Changes  Requested", "changesrequested"),  # div: double space
+    ("**Changes** Requested", "changesrequested"),  # div: inner bolding
+    ("Changes", "changesrequested"),  # div: THE crux row
+    ("changes", "changesrequested"),  # div
+    ("Changes needed", "changesrequested"),  # div
+    ("Request", "request"),
+    ("Reply", "reply"),
+    ("Replied", "reply"),
+    ("Pending", ""),
+    ("Maybe", ""),
+    ("Not Approved", ""),
+)
+
+_VERDICT_KINDS = frozenset({"approved", "changesrequested"})
+
+
+class AgreementMatrixTests(unittest.TestCase):
+    """One classifier, so every consumer gives the same answer on every row.
+
+    A future divergence is a FAILURE here rather than a discovery in a merge
+    thread six weeks later (main#1371 acceptance criterion 3).
+    """
+
+    def test_format_gate_verdict_scope_matches_the_counter(self) -> None:
+        """`_direction_is_verdict` gates the Requestor/Requestee swap check;
+        `_is_verdict` decides what the merge gate treats as a verdict. A gate
+        narrower than the consumer it guards is the #1150 defect class — and
+        pre-fix these two disagreed on 9 of the 25 rows below.
+        """
+        for value, kind in SPELLINGS:
+            with self.subTest(value=value):
+                self.assertEqual(
+                    fmt._direction_is_verdict(_body(value)),
+                    counting._is_verdict(value),
+                    f"format gate and counting gate disagree on {value!r}",
+                )
+                self.assertEqual(
+                    counting._is_verdict(value),
+                    kind in _VERDICT_KINDS,
+                    f"counting gate disagrees with the shared classifier on {value!r}",
+                )
+
+    def test_changes_requested_predicate_matches_trust_signals(self) -> None:
+        """The blocking question, asked by the conditional-field gate
+        (main#1363) and by `trust_signals`' retraction/attribution gating."""
+        for value, kind in SPELLINGS:
+            with self.subTest(value=value):
+                self.assertEqual(
+                    fmt._direction_is_changes_requested(_body(value)),
+                    ts._is_changes_requested(value),
+                    f"format gate and trust_signals disagree on {value!r}",
+                )
+                self.assertEqual(
+                    ts._is_changes_requested(value),
+                    kind == "changesrequested",
+                    f"trust_signals disagrees with the shared classifier on {value!r}",
+                )
+
+    def test_approved_predicate_matches_the_shared_classifier(self) -> None:
+        """`_is_approved` is the 2-reviewer threshold's whole input."""
+        for value, kind in SPELLINGS:
+            with self.subTest(value=value):
+                self.assertEqual(
+                    counting._is_approved(value),
+                    kind == "approved",
+                    f"approver set disagrees with the shared classifier on {value!r}",
+                )
+
+    def test_bare_changes_is_the_same_answer_everywhere(self) -> None:
+        """The crux row, called out on its own so a regression names itself.
+
+        Pre-fix: `_is_verdict("Changes")` True, `_direction_is_changes_requested`
+        True, `_direction_is_verdict` False. A bare-`Changes` verdict was
+        simultaneously a blocking ChangesRequested (so `Retracted:` was allowed
+        on it) and not-a-verdict (so the swap check never ran on it).
+        """
+        for value in ("Changes", "changes", "**Changes**"):
+            with self.subTest(value=value):
+                self.assertTrue(counting._is_verdict(value))
+                self.assertTrue(ts._is_changes_requested(value))
+                self.assertTrue(fmt._direction_is_changes_requested(_body(value)))
+                self.assertTrue(
+                    fmt._direction_is_verdict(_body(value)),
+                    "bare `Changes` is a ChangesRequested verdict to the counter, so the "
+                    "swap gate must apply to it too",
+                )
+                self.assertFalse(counting._is_approved(value))
+
+
+class SwapGateReadsTheValueTheCounterReadsTests(unittest.TestCase):
+    """The extraction axis: same vocabulary, different text, opposite answers.
+
+    `_direction_is_verdict` read the FIRST raw `RequestOrReplied:` hit anywhere
+    in the body. A reviewer who quotes or narrates an earlier direction above
+    their trailer therefore had the swap heuristic decided by their prose.
+    #934 fixed precisely this for `Requestor:` in this hook ("the swap
+    heuristic must read the Requestor the COUNTING hook reads — never the
+    first `re.search` hit over the whole body") and left `RequestOrReplied:`
+    on the raw first match.
+    """
+
+    BRANCH: ClassVar[str] = "L.Ferreira/1371-x"
+
+    @staticmethod
+    def _cmd(body: str) -> str:
+        return "gh pr comment 42 --repo noorinalabs/noorinalabs-main --body '" + body + "'"
+
+    def _check(self, body: str):
+        with mock.patch.object(fmt, "get_branch_name", return_value=self.BRANCH):
+            return fmt.check(_test_helpers.bash_input(self._cmd(body)))
+
+    # -- fail-OPEN: prose above the trailer suppressed the swap check -------- #
+
+    _PROSE_REPLY_ABOVE_APPROVED_TRAILER = (
+        "Earlier in this thread I posted RequestOrReplied: Reply and never "
+        "followed up. Doing so now.\n\n"
+        "---\n"
+        "Requestor: Lucas Ferreira\n"
+        "Requestee: Santiago Ferreira\n"
+        "RequestOrReplied: Approved\n"
+        "TechDebt: none\n"
+    )
+
+    def test_prose_direction_no_longer_suppresses_the_swap_block(self) -> None:
+        """Requestor IS the branch author — a genuine swap — and pre-fix it was
+        ALLOWED, because the first `RequestOrReplied:` hit was the prose
+        `Reply` and the verdict-scope gate returned early.
+
+        The counter reads the trailer's `Approved`, counts the comment as a
+        verdict, and self-review-excludes the swapped Requestor — so the real
+        reviewer's approval lands under the author's name and evaporates. That
+        is the #932 uncountable-verdict shape, reached through the one gate
+        that exists to catch it.
+        """
+        result = self._check(self._PROSE_REPLY_ABOVE_APPROVED_TRAILER)
+        self.assertIsNotNone(result)
+        assert result is not None
+        self.assertEqual(result.get("decision"), "block")
+        self.assertIn("swapped", result.get("reason", ""))
+
+    # -- fail-CLOSED: prose above a genuinely out-of-scope trailer ----------- #
+
+    _PROSE_APPROVED_ABOVE_REPLY_TRAILER = (
+        "You asked whether I would post RequestOrReplied: Approved here — not "
+        "yet, this is only a reply.\n\n"
+        "---\n"
+        "Requestor: Lucas Ferreira\n"
+        "Requestee: Santiago Ferreira\n"
+        "RequestOrReplied: Reply\n"
+        "TechDebt: none\n"
+    )
+
+    def test_prose_verdict_no_longer_drags_a_reply_into_scope(self) -> None:
+        """The mirror image: a Reply trailer whose prose mentions `Approved`.
+
+        Request/Reply invert the Direction table's role bindings (#378), so the
+        swap heuristic is unsound there — Requestor IS the PR author by
+        definition. Pre-fix the prose hit dragged this into verdict scope and
+        the hook BLOCKED a correctly-formed reply.
+        """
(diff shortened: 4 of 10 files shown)


parametrization:
**Review (architectural / blast-radius pass): LGTM**

Head reviewed: `de5571fe0d3d2ee316081fdc28315554e89534a9`. CI: 14/14 checks SUCCESS at that SHA (`statusCheckRollup`). Re-ran the full local suite at that SHA myself: `4407 passed, 1140 subtests passed` (matches the PR body's own number, same denominator). `ruff check` / `ruff format --check` / `mypy` all clean on the five touched Python files.

**Consumer enumeration (`rg --hidden` over `.claude/`, excluding tests) — what each now inherits:**

1. `.claude/hooks/validate_pr_review.py:1444-1445` — `_is_verdict = is_verdict_direction`, `_is_approved = is_approved`. Feeds the 2-reviewer approval-count gate and the TechDebt-required-on-verdict gate (`validate_pr_review.py:1590,1670`). Now shares its classification logic with the format hook below — a defect in `charter_trailer.verdict_kind` moves both gates together instead of only one.
2. `.claude/hooks/validate_review_comment_format.py:942,959` — `_direction_is_changes_requested` / `_direction_is_verdict` now call the shared predicates through the new `_direction_value()` (which itself now goes through `extract_charter_field`, not a raw `re.search`). Feeds the Requestor/Requestee swap-heuristic scope gate and the `Retracted:`/`OrchestratorCaused:` misplacement gate.
3. `.claude/lib/trust_signals.py:428,462` — `parse_verdicts` calls `charter_trailer.is_changes_requested` directly, and `_is_changes_requested` is re-exported as an alias for its own callers (`trust_signals.py:541`, the false-positive/orchestrator-caused-rework signals that feed `composite()` and, transitively, the ranking chain #1420/#1428 touch this same wave).
4. `.claude/lib/wave_status.py` — explicitly **not** migrated (comment-only change). `_CHANGES_REQUESTED_RE` stays a fifth, jq-only expression of the vocabulary because it runs inside a `jq` filter and can't call a Python predicate. The PR discloses this rather than silently leaving a stale comment — confirmed accurate by reading the file, this is the honest state, not a gap the PR is hiding.
5. Two non-code consumers depend on the **name**, not just the behavior: `agents/orchestration-model.md:280` and `feedback_pr_review_verdict_format.md` §7 both instruct `from validate_pr_review import _is_approved` by name for the review read-back recipe. The PR correctly keeps these as aliases rather than deleting them — a straight `del` here would have silently broken a documented read-back procedure with no test to catch it.

**Was the by-design bare-`Changes` disagreement load-bearing anywhere?** Traced it in the actual code rather than taking the PR body's word: `validate_review_comment_format.py:1172` gates the whole Requestor/Requestee swap check on `if not _direction_is_verdict(body): return None`. Pre-fix, `_direction_is_verdict` excluded bare `Changes`, so a swapped bare-`Changes` comment hit that early return and skipped the swap check entirely — while `validate_pr_review`'s old `_VERDICT_REQUIRING_TECH_DEBT` included `"changes"` and counted the same comment as a verdict. I reproduced this directly: copied the PR's new `test_verdict_direction_agreement.py` onto base `40cfe9f` with the pre-fix source restored and got the exact figure quoted in the PR body, `29 failed, 5 passed, 83 subtests passed`. So no — the disagreement was not protecting a real call site; it was the #932 evaporating-verdict shape, live.

**Is this a real SSoT or a fourth copy that agrees today?** Real SSoT — `is_verdict_direction` / `is_changes_requested` / `is_approved` in `charter_trailer.py` are the only definitions; both hooks and `trust_signals` import or alias them, none reimplement. Confirmed `trust_signals._FIELD_RE["verdict"]` (`trust_signals.py:162`) is untouched by this diff — still line-anchored (`^…$`, MULTILINE), raw (non-code-stripped) body, first-match via `.search()`, diverging from `extract_charter_field`'s stripped/trailer-scoped/last-match reading. That's issue #1372's scope exactly, and the PR is explicit about not touching it — correct call, not a gap to double-file.

**Do already-posted verdicts on the 10 open wave-30 PRs re-classify?** Checked all 10 (`#1419-1421, #1423-1429`) via `gh api .../issues/{n}/comments`. Denominator: 9 of the 10 have **zero** posted comments so far; the tenth (#1420) has exactly **one** verdict comment, a plain unadorned `RequestOrReplied: Approved` — classifies identically before and after this PR (it was never in the widened set). So: no, nothing currently posted on any of the 10 PRs changes classification under the new rules.

No must-fix findings from this pass.

---
Requestor: Weronika Zielinska
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: none

parametrization:
Reviewed at head `de5571fe0d3d2ee316081fdc28315554e89534a9` (fetched via `gh pr view --json headRefOid` + `gh api .../commits/<sha>` to confirm it resolves).

## Scope note

The org has already ruled on the merge-gate loosening (a first-token `Approved…` spelling — bang, "with nits", "- see below", "(post-merge)", bold — now enlarges the 2-reviewer approver set, and the corresponding `Changes…` spellings become blocking). I reviewed the **implementation** of that ruling, not the policy.

## Vocabulary probe (direct, `charter_trailer.verdict_kind` / `is_verdict_direction` / `is_changes_requested` / `is_approved`)

| value | `verdict_kind` | is_verdict | is_changes_requested | is_approved |
|---|---|---|---|---|
| `Approved!`, `Approved with nits`, `Approved - see below`, `Approved (post-merge)`, `**Approved**` | `approved` | True | False | **True** |
| `Changes needed`, `Changes  Requested` (2sp), `**Changes** Requested`, `ChangesRequested.`, bare `Changes` | `changesrequested` | True | **True** | False |
| `Not Approved`, `Not approved`, `NOT APPROVED`, `not-approved`, `Not-Approved`, `NotApproved` | `""` | **False** | False | **False** |
| `Request`, `Reply`, `Replied`, `""`, `None`, `gibberish` | `""`/`request`/`reply` | False | False | False |
| near-miss guard: `Changeset`, `Approves`, `Approvedish` | `""` | False | False | False |

Every value matches the PR body's own table exactly. The negation family (the dangerous direction for a merge gate) stays excluded at every entry point — `verdict_kind` only inspects `parts[0]` (and `parts[1]` for the two-word `Changes Requested` spelling), so a leading `Not`/`NOT`/`not-` token never reaches the `VERDICT_KIND` dict lookup. No near-miss word (`Changeset`, `Approves`) false-positives either — `normalize_verdict_token` requires the FULL first token to match, not a substring.

## Consolidation check

`rg --hidden` over `.claude/` for `changesrequested|_VERDICT_DIRECTIONS|_is_verdict\b|_is_approved\b|verdict_kind|_VERDICT_REQUIRING_TECH_DEBT|_direction_is_` outside the five touched files turns up only doc/charter prose (`feedback_pr_review_verdict_format.md`, `agents/orchestration-model.md`) referencing the *importable names* `_extract_charter_field`/`_is_approved` — which still resolve, since `validate_pr_review.py:1444-1445` keeps them as real aliases (`_is_verdict = is_verdict_direction`; `_is_approved = is_approved`), not reimplementations. `PredicateSingularityTests` (`test_verdict_direction_agreement.py:365-382`) asserts object identity (`assertIs`), not just equal output, for all four migrated names plus `trust_signals._is_changes_requested` — confirmed passing. The four described retirements are genuinely one classifier now.

The one acknowledged **fifth** copy, `wave_status._CHANGES_REQUESTED_RE` (`.claude/lib/wave_status.py:104`, a jq regex — can't call a Python predicate from inside `jq`), was disclosed in the PR body as "not filed." I filed noorinalabs/noorinalabs-main#1431 for it (low urgency — it's a reporting counter, not a merge gate, and its regex is hand-verified to currently agree with `verdict_kind`).

## `trust_signals._FIELD_RE["verdict"]` divergence (finding 1 in the PR body)

Confirmed still a private, line-anchored (`^…$`, MULTILINE), first-match regex over the RAW body (`trust_signals.py:159-163`), diverging from `charter_trailer.extract_charter_field` (code-stripped, trailer-scoped, last-match) on the one shape described (a line-anchored `RequestOrReplied:` line above the real trailer). Already tracked on #1372 per the PR body, which I re-verified is open and scoped to exactly this ("axis-3... same question main#1372 owns"). Not re-filing.

## Import / entry-point sanity (wave-30 theme: exit code as the real surface)

Ran both hooks as actual subprocesses (`python3 .claude/hooks/validate_pr_review.py` / `validate_review_comment_format.py`) against a benign command and malformed stdin — exit 0 both times, confirming the new `charter_trailer` imports resolve cleanly when the module is invoked the way PreToolUse actually invokes it (not just via pytest collection, which would mask an import-path problem `run_blocking`'s `try/except` around `check_fn` cannot catch — that guard only wraps the call, not the module-level `from charter_trailer import (...)` at the top of each hook). No regression here: nothing in this PR changes `sys.path` setup, and both hooks already had `.claude/lib` on `sys.path` before this change.

## Tests

`ENVIRONMENT=test pytest .claude/hooks/tests .claude/lib/tests -q` on the worktree at head: **4407 passed, 1140 subtests passed** — matches the PR body's own count exactly (I ran it myself, not trusting the table). The five touched/added test files (`test_verdict_direction_agreement.py` + the four modified) pass in isolation too (758 passed, 313 subtests). `python3 .claude/lib/pr_ci_state.py 1429` → `ready`, 14 checks reported (non-empty rollup, not the empty-rollup hard-not-ready case).

## Findings

No must-fix. The inverted pre-existing test (`test_bare_changes_not_verdict` → `test_bare_changes_is_a_verdict`) is correctly the only flipped assertion in the 4405-test pre-change baseline, and it's the one the PR's own premise-retraction argument requires flipping. Docstrings/comments on all five files match the code they describe (I read them against the diff, not just for prose plausibility) — in particular the `include_bare_changes` block comment in `charter_trailer.py:452-532` accurately states that no production caller passes `include_bare_changes=False` after this PR, and I confirmed that by grep: every call site in `validate_pr_review.py`, `validate_review_comment_format.py`, and `trust_signals.py` either omits the kwarg or passes `True`.

---
Requestor: Santiago Ferreira
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1431

parametrization:
## Merge-gate review — Nadia Khoury

Reviewed at head `de5571fe0d3d2ee316081fdc28315554e89534a9`, base `40cfe9f`, read via `gh api .../contents/<path>?ref=<sha>` and a detached read-only worktree at that SHA. CI at the head SHA: **14/14 check-runs `success`** via `statusCheckRollup` (`gh api repos/.../commits/de5571fe/check-runs` → `total_count: 14`, zero non-`success`).

**Approved.** The policy was owner-ruled and I did not re-open it; what follows is the implementation, its bound, and its blast radius. Every number below carries its denominator.

---

## 1. The claims are true at the entry point, not just in the test file

I tested the hooks as **subprocesses** on the exit code, not by importing `check()` — a module-level `ImportError` exits 1 and a `check()` raise exits 0, and PreToolUse reads both as allow, so an in-process assertion cannot see the gate's real verdict surface.

| scenario (real `gh pr comment` shape, body via `--body-file`) | base `40cfe9f` | head `de5571fe` |
|---|---|---|
| swapped verdict, direction value is bare `Changes` | exit **0** (ALLOWED) | exit **2** (blocked) |
| same body, direction value `Approved` (control) | exit 2 | exit 2 |
| prose `Reply` mention above a genuine swapped `Approved` trailer | exit **0** (fail-open) | exit **2** (blocked) |
| prose `Approved` mention above a genuine `Reply` trailer, requester is the PR author | exit **2** (false block) | exit **0** (allowed) |
| fenced charter template above a genuine `Reply` trailer | exit **2** (false block) | exit **0** (allowed) |

Row 1 is the defect the PR was filed on and it was live: the identical swap blocked on `Approved` and walked through on `Changes`. Rows 3–5 are the second axis, and they are the more valuable half — the swap gate was reading text no consumer read.

Counting hook, also as a subprocess: `gh pr merge 1421` → exit **2** with `1/2 required peer reviews` at both trees; `gh pr merge 1429` → exit 0 at head, correctly, at 2/2. So the module imports clean and the block path reaches exit 2 for real.

New test file against base source: **29 failed, 5 passed, 83 subtests passed** — matches the PR body exactly. Full suite at head: **4407 passed, 1140 subtests passed** — also exact.

## 2. The loosening is bounded, and I found the string that embarrasses it

The bound, read off the code rather than the body: `is_approved(v)` is true iff `re.sub(r"[^a-z0-9]", "", v.split()[0].casefold()) == "approved"`. That is **exact equality on the first whitespace-token**, not a prefix or substring test — and **everything after the first space is discarded**.

Full probe, 43 specimens through the real path (`extract_charter_field` → `_is_verdict` / `_is_approved` / both direction predicates), base → head:

| specimen | verdict b→h | approval b→h | notes |
|---|---|---|---|
| `Approved` / `approved` / `APPROVED` / `**Approved**` / `Approved (post-merge)` | T→T | T→T | unchanged |
| `Approved!` · `Approved.` · `Approved?` | F→**T** | F→**T** | ruled |
| `Approved with nits` · `Approved - see below` | F→**T** | F→**T** | ruled |
| `Approved: no` | F→**T** | F→**T** | **not ruled — see below** |
| `Approved: NO - do not merge` | F→**T** | F→**T** | **not ruled** |
| `Approved but blocking` · `Approved, pending CI` | F→**T** | F→**T** | **not ruled** |
| `Not Approved` · `Not approved` · `NOT APPROVED` · `not-approved` · `Unapproved` · `No, not approved` | F→F | F→F | correctly excluded |
| `Approve` · `Approving` · `Approval` | F→F | F→F | correctly excluded |
| `ChangesRequested` · `Changes Requested` | T→T | F→F | unchanged |
| `Changes  Requested` (2 sp) · `**Changes** Requested` · `Changes needed` · `ChangesRequested.` · `Changes-Requested` | F→**T** | F→F | stricter |
| `Changes` · `changes` · `**Changes**` | T→T | F→F | direction scope F→**T** — the fix |
| `Request` · `Reply` · `Replied` · `Requested` · `None` · `LGTM` · empty | F→F | F→F | unchanged |

Every negation the org would plausibly type is still excluded, and I checked five spellings of it. The hole is narrower than "negation" and more specific: **negation placed after the leading token**. A reviewer whose value is `Approved: NO - do not merge` is counted as a full approval toward the 2-reviewer gate; before this PR they counted as nothing. Same for `Approved, pending CI` — a conditional approval now counts as an unconditional one.

I am not blocking on it, and the reasoning is in §6. Filed as **#1447**, which also carries the surviving mutation.

## 3. Retroactive effect: **0 of 10**

This is the question the orchestrator needs answered before merging anything else in the wave, so I measured it with the review-state oracle rather than reasoning about it — `pr_review_state.py --json` run against every open PR from **both** trees, and diffed field by field:

| PR | base | head | delta |
|---|---|---|---|
| 1419 | 1/2 | 1/2 | none |
| 1420 | 0/2 | 0/2 | none |
| 1421 | 2/2 pass | 2/2 pass | none |
| 1423 | 2/2 pass | 2/2 pass | none |
| 1424 | 0/2 | 0/2 | none |
| 1425 | 0/2 | 0/2 | none |
| 1426 | 1/2 | 1/2 | none |
| 1427 | 0/2 | 0/2 | none |
| 1428 | 2/2 pass | 2/2 pass | none |
| 1429 | 2/2 pass | 2/2 pass | none |

**0/10 open PRs change any field** — not the approver set, not the count, not the pass/fail, not the stale-verdict list, not the attestation list. Merging #1429 does not retroactively approve, un-approve, or re-scope any other wave-30 PR. Nothing in the merge order needs to change on account of this PR.

Widening the window to the repo's **entire** issue-comment history — 2741 comments fetched, **1691** of which carry a parseable direction field — the approval classification flips on **3 of 1691 (0.18%)**, all on long-closed threads (#903, #250), and all through the same mechanism: bodies with **zero** sole `---` lines, where field extraction falls back to the whole body and last-match-wins lands on prose. One of the three is a *correct recovery* (on #250 a genuine reviewer approval had been silently dropped by the old exact-match). The other two are prose sentences, not verdicts. That mechanism is **#1415** as filed; this PR widens the set of prose strings that can spoof it, from exactly `Approved` to `Approved`-plus-anything. Cross-referenced there, not double-filed.

Direction classification flips on **14 of 1691 (0.83%)** — every one of them a case where the format gate had been reading different text than the counter. That is the second axis working.

## 4. The single-source-of-truth is real

Not a fourth copy that agrees today:

* `PredicateSingularityTests` asserts object **identity** (`assertIs`), not behavioural equality — `validate_pr_review._is_verdict is charter_trailer.is_verdict_direction`, same for `_is_approved` and `trust_signals._is_changes_requested`. A re-fork cannot pass by coincidence.
* I swept `.claude/` for surviving executable vocabulary tables outside `charter_trailer`: **zero**. Every remaining hit is prose, a fixture, or user-facing message text.
* The two hooks now read the same *text* as well as the same vocabulary — `validate_review_comment_format._direction_value` and `validate_pr_review`'s reader are both `charter_trailer.extract_charter_field` on `"RequestOrReplied"`, verified at runtime, so the swap gate can no longer scope itself differently from the counter it guards. That property is what closes rows 1 and 3 of §1.

On the two known survivors, both correctly declared rather than discovered:

* **`wave_status._CHANGES_REQUESTED_RE`** (the jq regex) — genuinely unconsolidatable from inside a `jq` filter, already filed as **#1431** by a co-reviewer. I confirmed it now diverges from the Python classifier on one *new* shape (`**Changes** Requested`, which the leading `*` defeats); that is a retro metrics counter, not the merge gate, and it belongs on #1431.
* **`trust_signals._FIELD_RE["verdict"]`** — still line-anchored, raw-body, first-match. This is the one place I disagree with the PR body. It is pinned by `test_line_anchored_mention_is_the_known_residual_divergence` with a "flip me when it lands" note aimed at **#1372** — but #1372 as filed has exactly one acceptance criterion, *"one corpus entry containing a `---` separator"* in `ConditionalFieldGrammarAgreementTests`, and nothing in it touches `_FIELD_RE` or `parse_verdicts`. #1372 closing **cannot** cause that flip, and #1371 closes with this PR, so at merge the residual loses its owner. This is wave-30's theme aimed at the tracking layer: a pointer that reports an owner it is incapable of being closed by. Filed as **#1448** with the flip instruction and the pointer retarget in its acceptance. Not blocking, because the residual itself is an unshown advisory warning, not a merge-gate fail-open — I re-derived that direction rather than taking it.

## 5. Gate self-test — the allow paths

I audited every `decision: allow` and every early return that short-circuits ahead of a hard block, since that is where a gate's hole normally is:

* `validate_pr_review`: the unresolvable-`--repo` and failed-fetch paths both **hard-block** (the #981 repair holds at head — I hit the block myself from a worktree whose child-roster path did not resolve). The only allows are the tail advisories and the explicit `--admin` override, both pre-existing and untouched.
* `validate_review_comment_format`: three allow-with-warning paths (no PR number, no branch name, no derivable branch author). All three sit **after** the three hard blocks and on the swap heuristic only. This PR does not move them, and it makes the `_direction_is_verdict` early return fire *less* often — bare `Changes` now reaches the swap block instead of exiting above it, which is §1 row 1. Strictly more comments reach the hard blocks, none fewer.

## 6. Mutation testing — the guards are live, with one gap

13 one-sided mutations of the shipped source, each run against the full 4407-test suite. **12 caught, 1 survived.**

Caught: `is_verdict_direction` always-true; `VERDICT_DIRECTION_KINDS` widened to accept `reply`; `is_approved` widened to accept blocking verdicts; `is_approved` widened to accept bare `Changes`; both `include_bare_changes` defaults flipped back to exclude; `_direction_value` reverted to the raw first-`re.search`; `_is_verdict` and `_is_approved` reverted to their old exact-match sets; `verdict_kind` reverted to whole-value matching; the two-word `Changes Requested` lookahead dropped; and `is_approved` replaced by a `"approv" in value` substring test — that last one is the important negative-space check, and it reds, which is why `Not Approved` is genuinely pinned and not merely currently-false.

**Survived:** replacing `verdict_kind`'s final `VERDICT_KIND.get(token, "")` with a `token.startswith(k)` scan. 4407 passed, 0 failed. Nothing in the suite distinguishes *"the first token IS `approved`"* from *"the first token BEGINS WITH `approved`"*, so `Approvedly` / `Approvedish` would classify as approvals under that mutation and no fixture notices. The shipped code uses equality and is correct; the boundary is simply unasserted. On **#1447** with the concrete fix (assert `Approve` / `Approving` / `Approval` / `Approvedly` classify as nothing).

## 7. Fold-in vs defer — my call, and which question decided each

Stating the deciding question rather than the conclusion, per the wave's standard.

* **`Approved: no` counts as an approval (#1447) — DEFER.** Decided by *could a WORKING artifact contain this shape?* No. A working reviewer artifact contains `Approved with nits` and `Approved - see below`, which the owner ruled should count and which the format gate has always accepted. A reviewer who means "do not merge" has `ChangesRequested` in the same vocabulary and the same gate accepts it; putting a negation behind an `Approved` token is misuse, not a shape correct work produces. Prevalence corroborates rather than decides: **0 of 1691**. The failure direction *is* the dangerous one (a silent false approval on the gate that guards merges), which is why it is filed with a concrete fix and not merely noted.
* **The surviving `startswith` mutation (#1447) — DEFER.** Decided by *direction*. The mutation is hypothetical, the shipped code is right, and the gap is in the pinning, not the behaviour. Two assertions close it.
* **`trust_signals._FIELD_RE["verdict"]` and its mis-aimed pointer (#1448) — DEFER.** Decided by *direction*: the residual under-warns (the author is not told their retraction will be ignored) rather than under-blocking, and the merge gate is not on that path. Had the divergence run the other way — the counter honouring a retraction the gate stayed silent about — I would have held the PR, because that is a fail-open on the blocking half.
* **`extract_charter_field`'s whole-body fallback (#1415) — DEFER, and not re-filed.** Decided by *prevalence and ownership*: 3 of 1691, all pre-`---`-convention, already owned by an open issue. This PR widens the surface but does not create it.
* **The jq regex (#1431) — DEFER.** Decided by *could a working artifact contain this shape?* — it must, since `jq` cannot call a Python predicate. Consolidation requires rewriting the counter, which is its own change.

Nothing on this list is a fold-in. The one thing I would have folded in — a gate that scopes itself more narrowly than the counter it guards — is what this PR removes.

## 8. Craft notes, no action required

The behaviour-change table is in the **source**, above the predicates, not only in the PR description, so the next reader of `charter_trailer.py` meets the ruling before they meet the default. The retracted premise is recorded as retracted with its evidence rather than quietly dropped. `include_bare_changes` survives with an explicit "no production caller passes `False`" warning instead of implying a live disagreement. The one inverted assertion is renamed, not edited in place, and its docstring argues the inversion. And the singularity name-sweep is honest about the one name it structurally cannot see (`_is_verdict` is a different word, not a prefixed form), pointing at the identity test that does. That last detail is the difference between a guard and a guard-shaped thing, and it is the reason I trust the rest of the file.

---
Requestor: Nadia Khoury
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1447, #1448, #1431

Issue #1389, opened 2026-08-11
promotion-audit: a charter heading that slugifies onto an existing skill name classifies AUTO — main#690's spurious-AUTO shape via collision, 17 live triggers


parametrization:
## A charter heading that slugifies onto an existing skill name produces a spurious AUTO

Found during the merge-gate review of #1383 (head `9c413dd3e19e071faf6714716a86ac2c2794a3c0`). **Pre-existing — #1383 does not introduce it**, but it is latent in exactly the code path #1383 touches, and it is the main#690 spurious-AUTO shape re-entering by a different door (heading collision instead of blank slug).

### Mechanism

`run.py::_section_signal_slug` derives the invocation-gate slug for a *not-yet-promoted* section from the section heading:

```python
def _section_signal_slug(section: h.CharterSection) -> str:
    if section.promoted_to:
        return section.promoted_to.split("/")[-1]
    return h._slugify(section.heading)
```

The docstring's justification is that a prospective slug "naturally yields ~0 invocations rather than the all-commits count an empty slug produced in the P5W4 mis-fire (main#690)". That holds only while the derived slug matches **nothing** on disk. If a charter section's heading happens to slugify onto the name of an existing, unrelated skill, `count_skill_invocations` returns **that skill's real invocation count**, which crosses `threshold=5` and classifies the section **AUTO** — "skill scaffold is safe to auto-generate" — for a section that was never promoted to that skill and shares nothing with it but a string.

### Measured

Reproduced end-to-end against the live repo at `9c413dd3`, by renaming one real not-yet-promoted skill-targeted section (`Agent Lifecycle Management`, blank `promoted_to`) and running the real driver path:

| heading | derived slug | `count_skill_invocations` | `target_configured` | decision |
|---|---|---|---|---|
| `Handoff` | `handoff` | 11 | True | **AUTO** |
| `Wave Scope` | `wave-scope` | 99 | True | **AUTO** |
| `Retro` | `retro` | 13 | True | **AUTO** |

[...]

Issue #1450, opened 2026-08-12
promotion-audit: count_section_citations can over-count via substring overlap for short/generic headings


parametrization:
## Summary

`count_section_citations` (added by the #1355 residual, `.claude/skills/promotion-audit/helpers.py`) counts literal substring occurrences of a charter section's heading in the feedback-log corpus (`_feedback_log_corpus` — live log + per-phase archives). This mirrors `count_retro_citations`'s existing mechanism for memories, and closes noorinalabs/noorinalabs-main#1389's collision path for good (that mechanism inherited an *unrelated skill's* real invocation count via a slug lookup; the new signal has no slug/skills-dir dependency at all).

A **different, narrower** false-positive shape survives: a short or generic heading is a literal substring of unrelated words/phrases elsewhere in the corpus. Measured live against a hypothetical section titled `Retro`:

```
'Retro': NEW count_section_citations = 76
```

76 counts substring hits like `Retro` inside `Retrospective`, plus every unrelated prose mention of "retro" — not genuine citations of a section named exactly `Retro`.

## Why this is currently latent, not live

All 25 real `skill`-targeted charter sections have long, distinctive multi-word headings (e.g. `Cross-Contract PRs`, `Design-Rationale Block for Critical-Path PRs (Mandatory)`). None of them exhibit this failure mode today — the one section that legitimately crosses threshold (`Cross-Contract PRs`, 5 citations) was hand-verified via `rg` as 5 genuine charter-section citations, not substring noise. This is a forward-looking risk: a future short/generic-heading section (or a rename to one) could silently accrue a spurious count.

## Candidate fix

Word/phrase-boundary matching instead of a bare substring count — e.g. require the heading match not be embedded inside a longer word (`\bRetro\b`-style), or require the retro-log citation to appear inside a recognizable "§ {heading}" / backticked citation shape (closer to how `_SOURCE_HINT_RE` / `_is_forward_reference` already discriminate real citations from noise elsewhere in this same module). Any fix should keep `count_retro_citations` (memories) untouched unless the same review finds it has the identical exposure — it uses the same `text.count(...)` substring approach.

## Scope note

Not required by noorinalabs/noorinalabs-main#1355's residual (which explicitly scoped source-signal semantics + the 2 mis-configured hook slugs + the #1389 collision confirmation) — filed separately per the residual's own finding, recorded on noorinalabs/noorinalabs-main#1389.

parametrization:
[...]

Issue #1403, opened 2026-08-11
No gate on PR-body vs commit-message linkage divergence: a body saying 'Refs #N' is silently overridden by a commit message saying 'Closes #N'


parametrization:
## A PR body can say `Refs #N` while its commit message still says `Closes #N` — nothing checks

Surfaced during the merge-gate re-review of #1383 (head `9c413dd3e19e071faf6714716a86ac2c2794a3c0`), where it occurred for real.

### What happened

#1383 was blocked at merge-gate because `Closes #1355` would auto-close an issue whose titular defect the PR leaves standing. The author correctly changed the **PR body** to `Refs #1355` and posted a detailed residual comment on #1355 to keep it open.

But the **commit message of `9c413dd3` still ends `Closes #1355`**:

```
Against live repo state: 114 charter-section decisions, 25 render
"not configured" ... AUTO=0/DECIDE=0 unchanged.

Closes #1355
```

Per `charter/pull-requests/reviews.md` — "Merging uses `--merge`, NEVER `--squash` — every base, including `main`" (owner directive 2026-07-30) — this PR merges with `gh pr merge --merge`, which lands `9c413dd3` **verbatim** on `main`, the default branch, with its closing keyword intact.

### The gap

There is **no gate on linkage-keyword divergence between a PR body and the PR's own commit messages.** A deliberate, reviewed decision recorded in one place (`Refs`) is silently contradicted by the other (`Closes`), and the contradiction is invisible to every existing check:

- `validate_pr_review` checks review verdicts and the `TechDebt:` attestation — not linkage keywords.
- `validate_pr_ci_status` checks CI — not linkage keywords.
- No linter reads commit-message issue references at all.

The failure is **silent and delayed**: nothing reports at merge time; the issue simply closes afterward, and only someone who goes looking notices. Exactly the wave-30 shape — a stated intent whose mechanism does not implement it, reporting nothing.

### Not fully confirmed, and why that does not make it safe

GitHub documents that a closing keyword in a commit message closes the referenced issue when that commit reaches the default branch. I could not corroborate it empirically **from this repo's history**: the only non-merge commit on `main` in the last 200 carrying a closing keyword is `0186d4d0` (`Closes #1210.`), and #1210's close event has `commit_id: null` — it was closed via the PR-body link, so the commit-message vector is confounded and untested there.

The correct statement is: the vector is documented, unverified here, and **currently unguarded either way**. A check costs little and removes the need to know.

### Suggested fix

[...]

PR #1451, opened 2026-08-12
fix(promotion-audit): compute charter->skill signal from the source section, not the prospective target

### LINKED ISSUE #1355: Promotion AUTO tier is unreachable: skill_invocations counts the not-yet-created TARGET, 0/25 gated sections can ever qualify

parametrization:
## Summary

Residual of #1355, continuing from #1383 (which delivered the "unconfigured" vs "below threshold" distinction and the acceptance-criterion-1 regression test, both left untouched here). This PR delivers the remaining scope, per the owner's 2026-08-12 restatement on #1355:

1. **Fix the signal's semantics.** `run.py`'s charter→skill tier counted invocations of the *prospective destination* skill slug (`count_skill_invocations(slug, repo_root)`) — structurally `0` forever for every not-yet-promoted section, since the skill doesn't exist yet to be invoked. It now counts `count_section_citations(section, feedback_log)`: retro citations of the SOURCE section's own heading in the feedback log (live + per-phase archives), mirroring the mechanism `count_retro_citations` already uses for memories.
2. **Fix the 2 mis-configured slugs** (`Agent Liveness Checkpoint`, `Throttle-Stall Recovery — Trigger Thresholds` in `charter/agents/lifecycle.md`) — both `hook`-targeted, so they were never reachable by the invocation gate either way. But a stray `<!-- promoted-to: lib/check_agent_liveness.py -->` marker (a lib path, not a skill slug) collided with the charter→skill `promoted-to:` regex and made `classify_section` mislabel them `ALREADY-PROMOTED`/`skill` instead of the correct `KEPT`/`hook` "charter sections only promote to skill" every other hook-target section renders. Renamed to `enforced-by:` (not parsed by `_PROMOTED_TO_RE`). **Hygiene only** — neither section becomes eligible for charter→skill promotion; they just stop being mislabeled.
3. **Acceptance criterion 2.** New tests prove the signal is computed from the source section and that a heavily-cited section crosses threshold — both a hermetic fixture test and a real-repo assertion. **Before: 0 of 25** skill-targeted charter sections could ever cross threshold (signal structurally 0 forever). **After: all 25 of 25** are genuinely evaluable (the signal is real and can reach threshold for any of them), and **1 of 25** (`wave-merge.md § Cross-Contract PRs`, 5 real retro citations across the live log + phase-2/phase-3 archives) crosses today at the unchanged `threshold=5`.
4. **#1389 confirmed, not assumed.** Commented explicitly on #1389 with a before/after table (`Handoff`, `Wave Scope`, `Retro`) showing the new signal has zero dependency on any skill slug or `.claude/skills/` lookup, closing that collision mechanism. A narrower, unrelated false-positive shape (short/generic headings over-matching via substring in the corpus) surfaced while verifying this and is filed separately as #1450 — it does not affect any of the 25 real headings today, all of which are long, distinctive phrases.

## What does NOT change

- `main#690`'s blank-slug guard in `count_skill_invocations` — untouched, still guards the skill→hook tier and any direct caller.
- The `target_configured` distinction from #1383 — still present, still selects which `KEPT` message renders below threshold. It no longer gates *whether* evidence can accrue (it never should have, and now doesn't), so its "not configured" message no longer claims citations "cannot accrue" (false after this fix) — it now reports the real citation count and clarifies that scaffolding is a deliberate human act, not a precondition for the signal.
- The skill→hook tier (`classify_skill`, `count_skill_invocations(skill.name, ...)`) — unaffected; a skill already exists once it has a `SKILL.md`, so measuring its own invocations was always the correct signal there.

## Test plan

- [x] `ENVIRONMENT=test python3 -m pytest .claude/skills/promotion-audit/tests/ -q` — 121/121 pass
- [x] `mypy .claude/skills/promotion-audit/ --ignore-missing-imports --no-error-summary` — clean
- [x] `ruff format --check` / `ruff check` on all changed files — clean
- [x] `python3 .claude/skills/promotion-audit/run.py` against the live repo — confirms `1 AUTO · 0 DECIDE · 254 KEPT · 22 SUPERSEDED/ALREADY-PROMOTED`, with `Cross-Contract PRs` as the sole AUTO item
- [x] Pre-commit + pre-push hooks (ruff, cspell, mypy, mypy-skills, pytest, pytest-skills, memory-budget, headcount-budget, doc-freshness, office-drift, mermaid-render) — all passed on push

## Related Issues

Closes #1355
Refs #1389, #1450, #1455

### DIFF
--- a/.claude/skills/promotion-audit/SKILL.md
+++ b/.claude/skills/promotion-audit/SKILL.md
@@ -15,7 +15,7 @@ The project's enforcement hierarchy is **hook > skill > charter > memory** (see 
 | From | To | Trigger |
 |---|---|---|
 | memory | charter | `promotion_target: charter` AND `retro_citations >= threshold` AND `status: active` |
-| charter | skill | Section marker `<!-- promotion-target: skill -->` AND skill-invocation signal >= threshold |
+| charter | skill | Section marker `<!-- promotion-target: skill -->` AND section-citation signal (retro citations of the section's own heading, #1355) >= threshold |
 | skill | hook | `promotion-target: hook` in skill frontmatter AND invocation signal >= threshold |
 
 Skill-to-hook **ALWAYS** produces a DECIDE-tier draft issue — never auto-applies (D6, hooks are security-sensitive).
@@ -76,14 +76,24 @@ each transition has a distinct signature because the signal sources differ:
 | Function | Signature | `signals` keys consumed |
 |---|---|---|
 | `classify_memory(memory, signals, already_promoted)` | `(Memory, dict[str,int], set[str]) -> Decision` | `retro_citations` |
-| `classify_section(section, signals)` | `(CharterSection, dict[str,int]) -> Decision` | `skill_invocations`, `threshold` |
+| `classify_section(section, signals)` | `(CharterSection, dict[str,int]) -> Decision` | `section_citations`, `threshold`, `target_configured` |
 | `classify_skill(skill, signals, already_promoted)` | `(Skill, dict[str,int], set[str]) -> Decision` | `skill_invocations`, `threshold` |
 
 Signal derivation (wired once in `run.py`): memory→charter uses
-`count_retro_citations`; charter→skill counts invocations of the section's
-candidate-skill slug (`promoted_to` with the `skills/` prefix stripped if
-already promoted, else `_slugify(heading)` — never an empty slug);
-skill→hook counts invocations of the skill name (always DECIDE, D6).
+`count_retro_citations`; charter→skill uses `count_section_citations`
+(#1355) — retro citations of the SOURCE section's own heading in the
+feedback log (live + per-phase archives, #964), the genuine evidence a
+section has earned promotion. `target_configured` (does a skill already
+exist on disk at the prospective slug?) is derived separately from
+`_section_signal_slug` + a disk check, but is informational only — it
+selects which KEPT message renders when the section is below threshold;
+it does not gate whether the citation signal itself can accrue. (Pre-#1355
+this tier counted invocations of the *prospective destination* skill slug
+— structurally 0 forever for a not-yet-promoted section, since the skill
+doesn't exist yet to be invoked; see `run.py`'s module docstring for the
+full history.) skill→hook counts invocations of the skill name (always
+DECIDE, D6) — unaffected by #1355, since a skill already exists once it
+has a `SKILL.md`.
 
 > `charter_parent` is the directory CONTAINING `charter/` (e.g.
 > `.claude/team`), NOT `charter/` itself — passing `.claude/team/charter`
--- a/.claude/skills/promotion-audit/helpers.py
+++ b/.claude/skills/promotion-audit/helpers.py
@@ -447,6 +447,44 @@ def count_retro_citations(memory: Memory, feedback_log_path: str) -> int:
     return max(by_title, by_file, len(memory.referenced_in_retros))
 
 
+def count_section_citations(section: CharterSection, feedback_log_path: str) -> int:
+    """Count occurrences of a charter section's heading in the feedback log.
+
+    #1355: the charter -> skill transition's actual evidence for promotion
+    is how heavily the SOURCE section is exercised -- not usage of the
+    PROSPECTIVE destination skill slug. A not-yet-created skill can never
+    accrue invocations of itself (`count_skill_invocations` on an
+    empty/prospective slug is structurally 0 forever, by definition, for
+    every not-yet-promoted section -- see `run.py`'s `_section_signal_slug`
+    docstring for the full history of that inversion). This mirrors
+    `count_retro_citations`'s mechanism for memories -- scanning the live
+    feedback log AND its per-phase archives (`_feedback_log_corpus`, #964)
+    for literal occurrences of the heading text -- which genuinely
+    accumulates: operators cite a charter section by its heading when they
+    reference it during a retro, and that citation count grows over time
+    independent of whether the destination skill has been scaffolded yet.
+
+    Deliberately independent of any skill slug or `.claude/skills/` lookup:
+    unlike the old destination-invocation signal, this count cannot inherit
+    an unrelated skill's invocation history just because a heading happens
+    to slugify onto an existing skill directory name (main#1389's
+    collision risk) -- there is no slug in this computation at all.
+
+    An empty or whitespace-only heading returns 0 -- never the corpus
+    length. `text.count("")` returns `len(text) + 1`, which would make a
+    blank heading look like it trivially crossed any threshold. This
+    should never occur in practice (`read_charter_sections` requires
+    non-empty heading text to match `_SECTION_MARKER_RE`), but the guard
+    is the same defensive shape as the main#690 blank-slug guard below.
+    """
+    if not section.heading or not section.heading.strip():
+        return 0
+    text = _feedback_log_corpus(feedback_log_path)
+    if not text:
+        return 0
+    return text.count(section.heading)
+
+
 def count_skill_invocations(skill_name: str, repo_root: str) -> int:
     """Count git log commits that reference the skill by slash-name.
 
@@ -873,7 +911,17 @@ def classify_section(
     section: CharterSection,
     signals: dict[str, int],
 ) -> Decision:
-    """Classify a charter section for the charter → skill transition."""
+    """Classify a charter section for the charter → skill transition.
+
+    #1355: `signals["section_citations"]` is expected to come from
+    `count_section_citations` (retro citations of THIS section's heading) —
+    the actual evidence for promotion — not from `count_skill_invocations`
+    against the prospective destination slug, which is structurally 0
+    forever for every not-yet-promoted section (see `count_section_citations`
+    docstring for the full rationale). Callers passing the legacy
+    `skill_invocations` key are silently treated as 0 citations; see
+    `run.py`'s `run_audit` for the canonical wiring.
+    """
     if section.promoted_to:
         return Decision(
             kind="ALREADY-PROMOTED",
@@ -904,34 +952,36 @@ def classify_section(
             reason="Charter sections only promote to skill",
         )
 
-    invocations = signals.get("skill_invocations", 0)
+    citations = signals.get("section_citations", 0)
     threshold = signals.get("threshold", 5)
 
-    if invocations < threshold:
-        # #1355: "wait for more operator-invoked runs" is only an accurate
-        # message when there is something an operator COULD invoke — i.e.
-        # a skill already scaffolded on disk at the prospective slug. For
-        # the common not-yet-promoted case (no `promoted-to` back-reference,
-        # no skill directory), the invocation count is structurally 0
-        # forever: promotion (creating the skill) is a deliberate human
-        # act, not something that accrues from waiting. Distinguish the
-        # two so the reader isn't told to wait for something that cannot
-        # arrive on its own.
+    if citations < threshold:
+        # #1383's target_configured distinction is preserved as informational
+        # context (does a skill already exist on disk at the prospective
+        # slug?), but it no longer gates *whether* evidence can accrue —
+        # `section_citations` is computed from the SOURCE section and
+        # genuinely accumulates regardless of scaffold status (#1355 fixes
+        # the prior inversion where the destination-invocation signal really
+        # was structurally 0 without a target). The unconfigured message is
+        # corrected accordingly: it no longer claims evidence "cannot
+        # accrue" without scaffolding — it can, and does.
         target_configured = bool(signals.get("target_configured", 0))
         if target_configured:
             reason = "Invocation threshold not met; wait for more operator-invoked runs"
         else:
             reason = (
                 "Promotion target not configured — no skill exists at this slug yet, "
-                "so invocations cannot accrue; scaffolding the skill (a deliberate "
-                "human act) is required before this becomes evaluable"
+                f"but that no longer blocks evidence from accruing: {citations} retro "
+                f"citation(s) of this section are already recorded ({citations}/{threshold} "
+                "toward threshold). Scaffolding the skill remains a deliberate human "
+                "act, but it is not a precondition for the signal itself"
             )
         return Decision(
             kind="KEPT",
             item_id=f"{os.path.basename(section.path)} § {section.heading}",
             from_tier="charter",
             to_tier="skill",
-            signal=f"skill_invocations={invocations} < {threshold}",
+            signal=f"section_citations={citations} < {threshold}",
             reason=reason,
             extra={"target_configured": target_configured},
         )
@@ -941,7 +991,7 @@ def classify_section(
         item_id=f"{os.path.basename(section.path)} § {section.heading}",
         from_tier="charter",
         to_tier="skill",
-        signal=f"skill_invocations={invocations} >= {threshold}",
+        signal=f"section_citations={citations} >= {threshold}",
         reason="Thresholds met; skill scaffold is safe to auto-generate",
     )
 
--- a/.claude/skills/promotion-audit/run.py
+++ b/.claude/skills/promotion-audit/run.py
@@ -26,14 +26,41 @@ decisions.
 Signal derivation, fixed in this driver
 =======================================
 - memory → charter: `count_retro_citations(memory, feedback_log)`.
-- charter → skill: the candidate skill slug is `section.promoted_to`
-  (with the `skills/` prefix stripped) when the section is already
-  promoted, else `_slugify(section.heading)` — the *prospective* skill
-  name. A not-yet-existent skill never appears in commit messages, so its
-  invocation count is naturally ~0. The empty-slug footgun is also closed
-  at the root in `helpers.count_skill_invocations` (blank slug → 0).
+- charter → skill: `count_section_citations(section, feedback_log)` (#1355)
+  — retro citations of the SOURCE section's own heading, the actual
+  evidence a section has earned promotion. `_section_signal_slug` still
+  resolves the prospective/promoted skill slug, but only to compute
+  `target_configured` (an informational disk-existence check, not a
+  signal input) — see main#1355 below for why the destination-invocation
+  signal this driver used pre-fix was structurally incapable of ever
+  reaching threshold for a not-yet-promoted section, and #1383/#1355 for
+  the fix history.
 - skill → hook: `count_skill_invocations(skill.name, repo_root)`
-  (always DECIDE per D6, never AUTO).
+  (always DECIDE per D6, never AUTO). This tier is unaffected by #1355 —
+  the destination (the skill) already exists by definition once it has a
+  `SKILL.md`, so measuring its own invocations is the correct signal
+  there; only the charter→skill tier had the destination-doesn't-exist-
+  yet inversion.
+
+Signal semantics, corrected (main#1355)
+========================================
+#1383 (the first #1355 PR) added a `target_configured` existence check on
+the prospective skill slug so a not-yet-scaffolded target renders a
+distinct "configuration gap" KEPT message instead of a misleading "wait
+for more operator-invoked runs" one. But the underlying *signal* was
+still `count_skill_invocations(prospective_slug, repo_root)` — usage of
+the DESTINATION, which by construction does not exist yet for any
+not-yet-promoted section, so the count was 0 for all 25 skill-targeted
+charter sections and could never rise. This residual (main#1355) replaces
+that signal with `count_section_citations`: citations of the SOURCE
+section's own heading in the feedback log (live + per-phase archives,
+#964) — the same mechanism `count_retro_citations` already uses for
+memories, which genuinely accumulates as operators reference a section in
+retros, independent of whether the destination skill has been scaffolded.
+This also closes the main#1389 collision path for free: the old signal
+could inherit an unrelated existing skill's real invocation count merely
+because a heading happened to slugify onto that skill's directory name;
+the citation count has no slug/skills-dir dependency at all.
 
 Usage
 =====
@@ -249,19 +276,29 @@ def run_audit(
 
     # charter → skill
     for sec in sections:
+        # #1355: the signal is now computed from the SOURCE section (retro
+        # citations of its heading) — the actual evidence for promotion —
+        # not from `count_skill_invocations` against the prospective
+        # destination slug (structurally 0 forever for every not-yet-
+        # promoted section; see `count_section_citations`'s docstring).
+        # This computation has no dependency on `slug` at all, which also
+        # closes the main#1389 collision path: the old destination-based
+        # signal could inherit an unrelated existing skill's real
+        # invocation count merely because a heading happened to slugify
+        # onto that skill's directory name. The citation count below is
+        # never touched by that coincidence.
+        citations = h.count_section_citations(sec, paths.feedback_log)
+        # `target_configured` is retained as informational context only
+        # (does a skill already exist on disk at the prospective slug?) —
+        # it selects which KEPT message classify_section renders, but no
+        # longer gates whether the signal itself can accrue.
         slug = _section_signal_slug(sec)
-        invocations = h.count_skill_invocations(slug, paths.repo_root)
-        # #1355: whether the prospective skill actually exists on disk yet
-        # gates which KEPT message classify_section renders — see the
-        # docstring on that branch. A not-yet-created skill (the common
-        # case for a not-yet-promoted section) can never accrue
-        # invocations, so "wait for more runs" would be misleading.
         target_configured = os.path.isdir(os.path.join(paths.skills_dir, slug))
         decisions.append(
             h.classify_section(
                 sec,
                 {
-                    "skill_invocations": invocations,
+                    "section_citations": citations,
                     "threshold": threshold,
                     "target_configured": int(target_configured),
                 },
--- a/.claude/skills/promotion-audit/tests/test_helpers.py
+++ b/.claude/skills/promotion-audit/tests/test_helpers.py
@@ -777,43 +777,67 @@ class ClassifySectionTests(unittest.TestCase):
         self.assertEqual(d.kind, "ALREADY-PROMOTED")
 
     def test_threshold_not_met(self) -> None:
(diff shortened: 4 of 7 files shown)


parametrization:
**Review: LGTM — independently re-derived every headline number and they hold.**

Head SHA verified myself: `82a734bd1dfa54f82bdfe7c22b67f9ae134b9347`.

**1. "Before: 0 of 25" — confirmed.** Recounted the charter directly (`rg` over `.claude/team/charter/`): 25 sections carry `<!-- promotion-target: skill -->`, 9 carry `hook` (34 total, matching #1355's corrected denominator, not the earlier "32 of 34" miscount). None of the 25 skill-targeted sections carries a `promoted-to:` back-reference on today's `main`, so none exits early via `ALREADY-PROMOTED` — all 25 would have hit the old `count_skill_invocations(prospective_slug, ...)` path, which is structurally 0 for a slug that names no real skill yet. #1389's own measurement ("current collisions: 0") corroborates that no accidental slug match was inflating this pre-fix either.

**2. "After: 25/25 evaluable, 1/25 crosses" — confirmed independently, not just re-run.** I wrote a standalone script that reimplements the corpus concatenation (live `feedback_log.md` + `archive/feedback_log_phase-*.md`, sorted) and ran `str.count(heading)` for all 25 real headings against it, without touching the PR's code at all. Result: 24 sections at 0–3 citations, exactly one at 5 — `Cross-Contract PRs` — which is the only one `>= threshold=5`. Matches the PR's claim exactly.

**3. The 5 citations are genuine.** Pulled all 5 lines by hand (4 in `archive/feedback_log_phase-2.md`, 1 in `archive/feedback_log_phase-3.md`, 0 in the live log — still sums to 5 across "live + two archives" as claimed). Every hit is a real `§ Cross-Contract PRs` charter reference in retro prose, not substring noise from an unrelated word.

**4. Blank-heading guard — present and pinned.** `count_section_citations` (`helpers.py`) returns 0 before touching the corpus when `heading` is empty/whitespace-only, avoiding the `text.count("") == len(text)+1` trap. `test_blank_heading_returns_zero_not_corpus_length` exercises both `""` and `"   "` against a long non-trivial corpus.

**5. main#690 guard untouched — confirmed.** `count_skill_invocations`'s blank-slug guard (`helpers.py`) is unchanged by this diff and still guards the skill→hook tier and any direct caller; re-read it directly, no relevant hunk touches it.

**6. #1389 immunity — confirmed, not assumed.** Read the actual issue thread: the author's before/after table (`Handoff` 13→1, `Wave Scope` 100→0, `Retro` 13→76) is posted on #1389 itself, and the regression test described (`test_section_citation_signal_immune_to_skill_slug_collision`) is present in `test_run.py` — it builds a fixture with a real `some-procedure/` skill dir AND 10 matching git-log commits, and asserts the section still lands `KEPT` at `section_citations=0`. The immunity holds structurally: `count_section_citations` never touches a slug or `.claude/skills/` at all. Correctly left the issue open pending merge rather than closing early.

**#1450 (short-heading over-match) — checked against the real 25, not taken on faith.** Listed all 25 skill-targeted headings and ran the same corpus-count script per heading: shortest is "Delegation Flow" (two ordinary words), and none produced a suspicious high count except the genuine `Cross-Contract PRs` hit. The claim that #1450 doesn't touch today's real sections stands; correctly deferred rather than folded in, since the root cause (substring match on short/generic headings) is unrelated to the collision mechanism this PR fixes.

**Slug-collision hygiene fix (`lifecycle.md`) — correct and appropriately scoped.** Traced the actual match path: `read_charter_sections` slices section bodies from one `## ` heading to the next, so a nested `### Enforcement (mechanized) <!-- promoted-to: ... -->` under a `hook`-targeted `##` section is still inside that section's body and matches `_PROMOTED_TO_RE`. Pre-fix this made `classify_section` hit the `ALREADY-PROMOTED`/`to_tier="skill"` branch (checked first, before the `promotion_target != "skill"` branch) for two sections that were never skill-eligible. Renaming to `enforced-by:` removes the match; both now correctly fall through to `KEPT`/`to_tier="hook"`/"Charter sections only promote to skill" — verified in code and covered by `test_hook_target_sections_no_longer_misreport_promoted_to_skill`. "Hygiene only, changes no reachability" is accurate — hook-targeted sections were never reachable by the invocation gate either way.

**`SteadyStateThroughDriver` — the new assertion is real, not a different hardcoded constant.** `1 AUTO / 0 DECIDE` traces to a citation count I reproduced independently (item 2 above), not to a number picked to make the test pass. Confirmed by running the live audit myself in an isolated worktree (see below) — it reports `1 AUTO · 0 DECIDE · 254 KEPT · 22 SUPERSEDED`, `Cross-Contract PRs` with `section_citations=5 >= 5` as the sole AUTO item, and produced no file-system side effects. This test will need a bump the next time a real citation crosses threshold, which the docstring says outright — that's an honest live-tree assay, not vacuous like the old `0/0`.

**Wave-30 theme check** (mechanism reporting success while incapable of the outcome): didn't find a second instance. The one soft spot is `run.py::_section_signal_slug`'s docstring, which still reads as if the slug it resolves "gates this section's charter→skill move" — post-fix it's purely informational (feeds `target_configured` only, never the signal). Not a functional bug, just a stale comment; worth a follow-up touch-up, not a blocker.

**Independent verification, own runs (not trusting the PR body):**
- Fetched every changed file at the head SHA via `contents` API, and additionally checked out the PR head into an isolated worktree outside the shared parent working copy (never `git checkout`'d in the parent).
- `ENVIRONMENT=test python3 -m pytest .claude/skills/promotion-audit/tests/ -q` → **121 passed** (confirmed against the actual PR tree; the same run against unpatched `main` is 107, confirming the 14 new/changed tests are real).
- `mypy .claude/skills/promotion-audit/ --ignore-missing-imports --no-error-summary` → clean.
- `ruff check` / `ruff format --check` → clean.
- `python3 .claude/lib/pr_ci_state.py 1451` → **READY**, 21 checks reported.
- `gh pr checks 1451` → all 21 green (no flake state remaining to confirm — the reported lychee re-run and runner-backlog queue both resolved).

No must-fix items. Approving.

---
Requestor: Aino Virtanen
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Merge-gate review — Nadia Khoury (Program Director).** Reviewed at head SHA `82a734bd1dfa54f82bdfe7c22b67f9ae134b9347`, against base `8781bbaa`, from a read-only detached worktree at the PR head — not the shared parent working copy. Every number below is re-derived independently from the real charter and the real feedback-log corpus, not read off the PR body. Denominators stated throughout.

---

## 1. The headline measurement — confirmed, with denominators

Re-derived by loading the real charter at the head SHA (`read_all_charter_sections(".claude/team")`):

| population | count |
|---|---|
| charter sections carrying a `promotion-target` marker | **114** |
| ...`none` | 80 |
| ...declaring a real target (`skill` \| `hook`) | **34** |
| ...targeting `skill` — the only population the charter→skill gate reaches | **25** |
| ...targeting `hook` — rejected earlier as an invalid transition | **9** |
| of the 25 skill-targeted: carrying a non-blank `promoted_to` | **0** |

This matches the corrected denominator on #1355 (25, not the first draft's "32 of 34"), and I recounted from the charter rather than from either report.

Signal, computed both ways over all 25:

- **OLD** (`count_skill_invocations(_section_signal_slug(sec), repo_root)`): **0 of 25** reach threshold=5. Every one is 0.
- **NEW** (`count_section_citations(sec, feedback_log)`): **1 of 25** reaches threshold — `wave-merge.md § Cross-Contract PRs` at exactly 5. The other 24 range 3, 1, 1, and 0×21.

Driver run at head confirms end-to-end: `1 AUTO · 0 DECIDE · 254 KEPT · 21 ALREADY-PROMOTED · 1 SUPERSEDED`, sole AUTO = `wave-merge.md § Cross-Contract PRs`, signal `section_citations=5 >= 5`. Base run for comparison: `0 AUTO · 0 DECIDE · 253 KEPT · 23 ALREADY-PROMOTED · 1 SUPERSEDED`. The deltas reconcile exactly (ALREADY-PROMOTED 23→21 from item 3; KEPT 253+2−1=254; totals 277 both sides). The PR body's `22 SUPERSEDED/ALREADY-PROMOTED` is 21+1 — accurate.

**Citation genuineness — the load-bearing check.** That single section carries the entire non-degeneracy claim, so I read all 5 occurrences in context and checked the character on each side of every match:

| # | source | context | boundary |
|---|---|---|---|
| 1 | `archive/feedback_log_phase-2.md` | "Charter home: `charter/pull-requests.md` § Cross-Contract PRs." | ` ` / `.` |
| 2 | phase-2 | "`charter/pull-requests.md`: new § \"Cross-Contract PRs\" per process-change #2" | `"` / `"` |
| 3 | phase-2 | "per Charter § Cross-Contract PRs, alembic merge migration is now in main" | ` ` / `,` |
| 4 | phase-2 | "charter updates ... for single-session-team delegation pattern + Cross-Contract PRs § + Load-Bearing Followups §" | ` ` / ` ` |
| 5 | `archive/feedback_log_phase-3.md` | "Proposed location: charter `pull-requests.md` § Cross-Contract PRs OR new § Design-Rationale Blocks" | ` ` / ` ` |

All 5 are genuine references to the charter section, every one word-boundary-clean (no adjacent alphanumeric on either side). **Not substring noise.** Confirmed.

## 2. Is the choice of signal defensible? — Yes, with one honest caveat stated

Retro citations of the section heading, over `git log --grep` and code back-references. The reasoning holds:

- It reuses `count_retro_citations`'s already-proven accumulating mechanism, including the #964 live+archive corpus scan — 4 of the 5 confirming citations live in phase archives, so a live-log-only scan would have measured 1 and this PR would have shipped a still-degenerate pipeline. Reusing `_feedback_log_corpus` is what makes the number real.
- It has **zero** dependency on any skill slug — which is what makes item 5 true rather than aspirational.
- `git log --grep` on a heading would have re-imported the main#690 substring-over-commit-corpus failure mode; code back-references measure adoption of the *destination*, i.e. the same inversion in a new costume. Both rejections are correct.

**The caveat, which I want on the record because it bounds how much the number means.** The signal measures *the heading appearing in retro prose*, which is a proxy for exercise of the source section, not exercise itself. Of the 5 confirming citations: #1, #3 and #4 are true post-hoc citations of an existing rule (the intended shape); **#2 is the retro proposal that created the section**, and **#5 names it as a candidate home for a different rule**. Those last two are the section being *discussed*, not *relied on*. At exactly 5 against threshold 5 the margin is zero — disqualify either and it does not cross.

I am not treating that as a defect, for two reasons, and I'd rather say why than let it pass silently. First, every string-count signal has this property, including `count_retro_citations`, which this deliberately mirrors and which the pipeline already relies on for the memory→charter tier. Second, and decisively: a charter→skill AUTO does **not** auto-apply. Per SKILL.md § AUTO artifacts, it lands as a branch + PR under the standard two-reviewer gate ("Q3 decision: auto-promote artifacts land via PR, not direct commit"). So an imprecise citation terminates in a human-reviewed PR, not a silent commit. The signal is fit for the tier it gates. Worth remembering the next time someone proposes raising this tier's automation.

## 3. Blank-heading guard — present, correct, and mutation-killed

The guard is at the top of `count_section_citations`:

```python
if not section.heading or not section.heading.strip():
    return 0
```

I did not take the accompanying test on faith. **Mutation test:** deleted the two guard lines and ran the suite.

```
FAILED test_helpers.py::CountSectionCitationsTests::test_blank_heading_returns_zero_not_corpus_length
E   AssertionError: 611 != 0
```

611 is `len(corpus) + 1` — the exact main#690 shape, an unguarded blank heading crossing any threshold trivially. The assertion is genuine, the fixture trips it, and it is not inert. 1 failed / 120 passed.

**main#690's blank-slug guard, untouched — verified by content, not by reading the diff.** I extracted `count_skill_invocations` from `contents/helpers.py` at base and at head via the API and compared: **byte-identical**. `if not skill_name or not skill_name.strip(): return 0` is intact and still guards the skill→hook tier and every direct caller. #1355's requirement 4 is satisfied.

## 4. The vacuous-test replacement — a genuine assertion, not a relabelled constant

`SteadyStateThroughDriver` 0 AUTO → 1 AUTO. The old assertion was vacuous: pre-fix, 0 was the only reachable value, so the test could not fail for any repo state. The question is whether the new one is signal-sensitive or just a different hardcoded number.

**Mutation test:** reverted `run.py` to the pre-fix destination-invocation signal, leaving all tests untouched. Four tests died:

```
FAILED test_run.py::SteadyStateThroughDriver::test_current_repo_state_yields_exactly_one_auto_section
E   AssertionError: 0 != 1 : unexpected AUTO set: []
FAILED test_run.py::EmptySlugRegression::test_heavily_cited_section_crosses_threshold_via_source_signal
E   AssertionError: 'KEPT' != 'AUTO'
FAILED test_run.py::EmptySlugRegression::test_section_signal_is_source_citations_not_destination_invocations
E   AssertionError: 'AUTO' != 'KEPT'
FAILED test_run.py::EmptySlugRegression::test_section_citation_signal_immune_to_skill_slug_collision
E   AssertionError: 'AUTO' != 'KEPT'
```

The steady-state assertion is coupled to the signal, so it is a real assertion. It also asserts the AUTO item is *specifically* `Cross-Contract PRs`, not merely that the count is 1 — that identity assertion is what stops a future "fix by bumping the number" from restoring vacuity, since a bare re-baseline still has to leave a named section in the set.

It **will** rot, and I want that anticipated rather than discovered: `ci-gates.md § Load-Bearing Followups for Disabled CI Jobs` currently sits at **3 of 5**, so two more retro citations flip it to AUTO and turn this test red on an unrelated PR. That is acceptable — the failure direction is loud and red, never silent, and a promotion pipeline noticing that its real AUTO set changed is the test doing its job rather than failing at it. When it does go red, re-baseline by verifying the new section legitimately earned it; do not just increment the integer.

## 5. #1389 immunity — confirmed, not assumed

Re-derived the author's before/after table myself rather than trusting the comment:

| heading | slug | OLD `count_skill_invocations` | NEW `count_section_citations` |
|---|---|---|---|
| `Handoff` | `handoff` | 13 | 1 |
| `Wave Scope` | `wave-scope` | **100** | 0 |
| `Retro` | `retro` | 13 | 76 |

Reproduces exactly. The `Wave Scope` row is the whole argument in one line: under the old signal a section headed "Wave Scope" classifies **AUTO at 100 ≥ 5** purely by inheriting an unrelated skill's invocation history. Under the new signal it is 0.

The regression test is real, not decorative — it builds a fixture with **both** a colliding skill directory at the exact prospective slug **and** 10 matching git-log commits, then asserts KEPT at `section_citations=0`. Mutation-confirmed above: under the old signal that same fixture flips to AUTO. The fixture can trip the assertion, which is the property that matters.

**#1389's state is now correct.** It is OPEN. The premature close and the reopen are 15 seconds apart in the timeline, and the reopen comment names the error class explicitly ("I closed this prematurely before the #1355 residual PR has actually merged"). Self-caught, self-reversed, and documented in the durable thread — which is the right handling of that mistake, and I'd rather see it disclosed than tidied away. The commit trailer correctly says `Refs #1389`, not `Closes`, so it will not auto-close on merge. **Sequencing note for whoever merges: #1389 needs a manual close after this lands**, consistent with its own reopen note.

## 6. #1450 — defer, and the deferral is correct on measurement

I checked the claim against the actual headings rather than against the assertion that they are "long and distinctive", because that justification is not quite right: the shortest of the 25 is `Delegation Flow` at 15 characters, two common words. "All are long" is an overstatement.

The deferral survives anyway, on a stronger basis than heading length. Two measurements over the full 459,099-character live+archive corpus:

- For every one of the 25 headings, I examined the character on each side of **every** occurrence: **zero** matches have an adjacent alphanumeric on either side. No overmatch exists today, 0 of 25.
- **No** skill-targeted heading is a proper substring of any other charter heading — the `Retro`-inside-`Retrospective` relationship has no instance among the real headings.

Applying the standard questions: *could a working artifact contain this shape?* — yes in principle, and the `Retro` → 76 row proves the shape is real for a hypothetical short heading. *Which direction does it fail?* — toward **spurious promotion**, the direction main#690 already burned this pipeline in, which is the unfavourable answer. *Prevalence?* — measured **0 of 25**, with a boundary check, not an eyeball.

**Prevalence is the question that decides it**, backstopped by the AUTO-lands-as-a-reviewed-PR fact from §2 — the unfavourable failure direction terminates at a two-reviewer gate, not at a commit, which is materially different from main#690's blast radius. Defer is right; #1450 is filed, open, and correctly labelled. I'd ask only that #1450 carry the measured basis (zero boundary violations across 25 headings) rather than "all headings are long", so a future short heading is re-measured instead of waved through on a heuristic that was never the real reason.

## 7. The slug fix — confirmed, and "hygiene only" is accurate

Verified by running the audit at both SHAs:

- **base:** both sections → `ALREADY-PROMOTED`, `to_tier=skill`, "Section has a promoted-to back-reference" — the mislabel.
- **head:** both → `KEPT`, `to_tier=hook`, "Charter sections only promote to skill" — identical to every other hook-targeted section.

"Hygiene only" checks out: both are `promotion-target: hook`, so they were unreachable by the gate before and after; neither becomes eligible for charter→skill; neither appears in the AUTO set. The only movement is the ALREADY-PROMOTED 23→21 accounted for above.

I also confirmed the rename is safe at the ecosystem level: `_PROMOTED_TO_RE` in the promotion-audit helpers is the **only** code consumer of `promoted-to:` anywhere in the tree, and `enforced-by:` has no parser at all — which is exactly the intent, an informational marker. And no hook- or none-targeted section anywhere else in the charter still carries a stray `promoted_to`, so this class is fully cleaned up rather than patched in two spots.

## 8. Linkage discipline — one deviation, non-blocking; I verified why

The commit trailer on `82a734bd` says `Closes #1355` / `Refs #1389`. The PR **body carries no closing keyword at all** — no `## Related Issues` section, no `Closes`, no `Refs`. `authoring.md § PR Template` states the body must reference the issue with `Closes #N`.

This is a deviation, not the #1403 shape. #1403 is an active contradiction (body `Refs`, trailer `Closes` — which is what #1383 did, and I confirmed #1383's body does say `Refs #1355`). Here the body is silent rather than contradictory, so nothing is silently overridden.

I checked whether it is load-bearing rather than assuming either way, since a squash merge that regenerates the message from the PR body would drop the trailer and silently fail to close #1355. Repo settings: `squash_merge_commit_message: COMMIT_MESSAGES` and `merge_commit_message: PR_TITLE`. Under **both** permitted methods the commit carrying `Closes #1355` lands on `main` intact, so #1355 closes correctly either way. Cosmetic, confirmed by measurement.

Worth a pre-merge `PATCH` of the body to add `Closes #1355` — zero CI cost, since a body edit does not move the head SHA — and it keeps the human-readable record honest. I sampled #1386/#1393/#1394/#1390: none use the `## Related Issues` heading either, but **all** carry an explicit `Closes #N` in the body prose, so this PR is the outlier. Not a blocker.

**Is #1355 fully discharged?** Yes — checked against the issue's own acceptance list, all three criteria and all four proposed-fix items: signal semantics fixed (§1); unconfigured-vs-below-threshold distinction retained from #1383 and re-pinned against the corrected signal; the 2 mis-configured slugs fixed (§7); main#690's guard kept byte-identical (§3); a test asserts the signal is computed from the source section and that a heavily-cited section crosses (§4); and the audit now yields a non-degenerate classification. `Closes #1355` is the correct disposition. No residual remains unnamed.

## 9. CI

Green at the head SHA, confirmed from `statusCheckRollup` on commit `82a734bd` directly rather than from the PR-level summary: rollup state `SUCCESS`, **22 of 22** checks `SUCCESS`, none pending, rollup non-empty. That includes `Roster union sync-drift gate` (the one previously pending on runner backlog), `Ruff lint`, `Ruff format check`, `Mypy type check`, `Pytest — hooks and lib test suites`, and `Spellcheck (cspell)`. `pr_ci_state.py 1451` independently reports READY. I also re-ran the skill suite at the head SHA myself: **121 passed**.

## Findings

**Must-fix:** none.

**Non-blocking, dispositioned:**
1. PR body carries no `Closes #1355` (§8) — recommend a body `PATCH` before merge; no CI impact, no head-SHA change.
2. `SKILL.md` § Determinism still says the smoke test verifies "zero AUTO, zero DECIDE **on current repo state**". After this PR the real tree is 1 AUTO, and `test_smoke.py` holds at zero only because it passes a hardcoded `section_citations=0` — which the PR correctly documents in the test's own comment but not in this sentence. Deliberately not folded in: the edit would move the head SHA and re-trigger a 22-check CI cycle for one sentence on an already-verified-green PR. Filed as #1455.
3. After merge, close #1389 manually (§5) — `Refs`, so it will not auto-close.

## Verdict

Approved. This is the rare fix that both repairs the mechanism and proves the repair with evidence that can fail — every guard it adds survived a mutation test, and the one assertion that was previously vacuous is now coupled to the signal it claims to measure. The wave-30 theme was a mechanism reporting success while being incapable of the outcome it described; the AUTO tier goes from **0 of 25 sections able to qualify** to **25 of 25 evaluable with 1 crossing on 5 hand-verified citations**, and it does not ship a new instance of the same defect. The disclosed premature close on #1389 was caught, reversed, and documented by the author before review — handled correctly.

---
Requestor: Nadia Khoury
Requestee: Wanjiku Mwangi
RequestOrReplied: Approved
TechDebt: #1455

PR #1453, opened 2026-08-12
test(hooks): close the agreement-corpus `---` gap, and route trust_signals onto the field-extraction SSoT (#1372)

### LINKED ISSUE #1372: `ConditionalFieldGrammarAgreementTests` corpus has no `---` separator, so the scope re-narrowing mutation passes the test that justifies the fourth copy

parametrization:
Refs #1372

## Premise check first — the story is not moot

#1372 says it becomes moot if #1371 or #1359 removes the fourth copy of the
conditional-field grammar. Both have merged (#1359 earlier; #1371 as PR #1429,
which merged *while this branch was in flight*). Verified against
post-#1371 `origin/main` before writing anything:

| Premise | Result on `main` @ `38cf04e` |
|---|---|
| `ConditionalFieldGrammarAgreementTests` + the fourth copy still exist | Yes — `_CONDITIONAL_FIELD_RE` / `_CONDITIONAL_VERDICT_FIELDS` / `_conditional_fields_present` all present |
| Corpus still lacks a `---` separator entry | Yes — 15 rows, none with a separator |
| The ninth #1363 mutation still passes the agreement test | Yes — see below |

Aino's reading is correct: #1371 consolidated the verdict **direction**
vocabulary; the corpus certifies the `Retracted:` / `OrchestratorCaused:`
**presence-detection** pair. Different symbols.

## Core: the corpus gap (issue acceptance)

The mutation is `_conditional_fields_present` re-narrowing its scan to
`charter_trailer.trailer_block_substring(strip_code_regions(body))` — axis 3,
the scope divergence the class's own docstring names.

**Before**, against the 15-row corpus:

```
3 passed, 103 deselected, 30 subtests passed in 0.34s      <- the class is GREEN
```

with only the *neighbour* failing:

```
FAILED ConditionalVerdictFieldTests::test_genuine_field_above_the_trailer_separator_still_blocks
E       AssertionError: unexpectedly None
```

**After**, same mutation, re-run on the merged tree:

```
SUBFAILED(body='Retracted: withdrawn.\n\n---\nRequestOrReplied: Approved\n')
  ConditionalFieldGrammarAgreementTests::test_presence_detection_agrees_field_by_field
SUBFAILED(body='Retracted: above.\n\n---\nOrchestratorCaused: below.\n')
  ConditionalFieldGrammarAgreementTests::test_presence_detection_agrees_field_by_field
SUBFAILED(body='OrchestratorCaused: stale brief.\n\n---\nnotes\n\n---\nRequestOrReplied: Approved\n')
  ConditionalFieldGrammarAgreementTests::test_presence_detection_agrees_field_by_field
E       AssertionError: Lists differ: [] != ['OrchestratorCaused']
3 failed, 4 passed, 103 deselected, 37 subtests passed
```

Five rows added: field above the separator, fields on both sides, an
in-trailer control (so a red localises to the scope axis rather than to "any
separator"), a two-separator body, and a body whose only `---` is inside a
fence (pins strip-then-scope ordering).

`test_corpus_covers_the_trailer_separator_scope_axis` keeps the shape from
being trimmed back out. It compares the two **scopes** directly rather than
calling `_conditional_fields_present`, so it asserts a property of the corpus
and stays green under the mutation it exists to make catchable — otherwise it
would just be a noisier duplicate of the agreement test.

No divergence exists today (#1363's review fuzzed 60,000 bodies for 0 presence
and 0 stripper mismatches). Nothing here "fixes" a divergence that does not
exist; the defect was a docstring claiming more than its corpus proved.

## Fold-in: `trust_signals._FIELD_RE` — routed, with the measurement

**Disposition: routed through the SSoT.** The evidence, not an argument:

Harvested all 2,748 issue comments in this repo, restricted to the 1,475 on a
PR — the only bodies `_pr_comment_bodies` hands `parse_verdicts`. The private
regexes and `charter_trailer.extract_charter_field` disagreed about the
**verdict kind of 27 comments across 18 PRs** and the **requestor-field value
of 22 comments across 19 PRs** (i.e. which engineer `extract_signals` credits
`must_fix_caught` and debits `review_false_positives` to).

Every mechanism favours the SSoT:

| Mechanism | PR comments | Effect of the private copy |
|---|---|---|
| anchoring | 16 / 7 PRs (#167, #235, #270, #275, #276, #277, #279) | the three charter fields written on ONE slash-separated line put no field at a line start, so `^` **dropped genuine verdicts the merge gate counted** |
| raw body | 1 / PR #204 | verdict read out of a fenced example — a verdict nobody cast |
| scope | 10 / 10 PRs (#28, #282, #350, #444, #446, #468, #873, #880, #884, #1126) | fields above a later `---`; on #1126 a `Request`-valued header stub outranked the real `Approved` trailer under first-match-wins |
| value | 22 | role parentheticals survived, so a requestor written as `Aino Virtanen (Standards & Quality Lead)` opened a **second `extract_signals` bucket for the same person**; one comment bucketed review credit under a mangled multi-field string |

After routing: **0 disagreements on those 1,475**.

Direction rationale: "an uncountable verdict counts as zero reviews" is this
org's settled position (#932/#934), and `validate_pr_review` reads through
`extract_charter_field`. A trust instrument scoring engineers off a different
verdict set than the merge gate enforced has no principled defence for the
difference.

### Named residuals — not blurred away

1. **Behaviour change.** The 10 above-separator comments stop counting as
   verdicts, so recomputing a **pre-#1372 wave's signals can move**. Stated in
   the code and pinned by `FieldExtractionIsTheSharedSSoT
   ::test_above_separator_fields_no_longer_count`.
2. **A pinned assertion changed value.**
   `test_bold_only_verdict_value_still_classifies` went from
   `**ChangesRequested` to `ChangesRequested` — it was pinning an artifact of
   the deleted regex's one-sided bold strip. The #1358 mutant it guarded
   (`normalize_verdict_token`'s alnum strip) is covered **on the same input**
   by `test_charter_trailer.py`, where that test's own docstring already
   pointed. Coverage moved to its owner; it was not dropped.
3. ~~**The SSoT has its own unanchored-prose fail-open** … Both are **issue**
   comments, 0 on PRs, so routing introduced no measured regression.~~
   **CORRECTED — that claim was false.** The merge gate falsified it and I
   reproduced the falsification against both real engines with `__file__`
   asserted. `trailer_block_substring` falls back to the **whole body** when a
   comment has no sole `---`, and last-match-wins then lets a field-shaped
   prose line outrank the genuine trailer. Exposure is not a corner:

   * **809 of 1,398 verdict-bearing PR comments (58%) carry no sole `---`.**
   * **19** yield a non-name requestor. **PR #364 is a true regression** of
     this migration — requestor goes from `Wanjiku Mwangi` to a markdown table
     cell on an `Approved` comment; **#167 and #277 are the same shape**.
   * The other **18** were skipped entirely by the deleted regex, so no
     engineer loses credit versus before — but each now opens a **phantom
     `extract_signals` bucket** collecting `must_fix_caught` /
     `review_false_positives`.

   Filed as **#1456** and **not fixed here**: the fallback belongs to
   `charter_trailer`, so changing it would move the merge gate's reading of
   every separator-less comment too. The residual now lives in
   `parse_verdicts`' **docstring**, not only in this body — a PR body is not
   read by the next person to touch the function — and is pinned by
   `test_separator_less_body_lets_prose_outrank_the_trailer`, which asserts
   the wrong value on purpose so #1456's fix reds it.

   Two consequent overclaims elsewhere in this body are also corrected in the
   code: the anchoring class is **half a win** (verdict recovered, requestor
   off that same separator-less line is the whole rest of it), and "disagree
   on 0 of 1,475" is **agreement, not correctness** — the two now also agree
   where the shared extractor is wrong.
4. **Nothing open now owns removing the mirror.** The comment in
   `validate_review_comment_format.py` used to defer that to #1371 and then to
   #1372; both have landed and neither removes it. The code now says so
   plainly instead of pointing at a closed issue. Until something does, the
   agreement corpus **is** the mechanism — which is why its coverage is
   asserted rather than assumed.

   **Extended after review:** `trust_signals.py` and `test_trust_signals.py`
   still carried the same closed-issue pointer at #1371 — the exact defect the
   other half of this diff fixes. Both retargeted to the same "nothing open
   owns this" statement, so the diff is now self-consistent.

## Wave-30 signal check (re-verified after the second merge, not assumed)

#1428's `RankingPopulation.from_wave` calls `extract_signals`, so this change
moves values inside its anchoring population. Checked against current `main`
using `wave_status.merged_prs` itself:

| | |
|---|---|
| PRs whose parse changes | 27, **max #1126** |
| wave-30 merged PRs in this repo | 16, **min #1384** |
| intersection | **empty** |

**No wave-30 signal moves.**

What this class certifies is now stated in its docstring, and so is what it
does **not**: presence detection and the shared stripper, not value
extraction, verdict classification, or person identity.

## #1429 collision — resolved, not pre-resolved

#1429 merged mid-flight and touches all four files. `git merge origin/main`
(never rebase). One textual conflict, plus two semantic ones worth flagging
for review:

* #1371 added a block naming Aino's finding as the pending repair ("reported
  on main#1371 rather than fixed in passing"). This branch **is** that repair,
  so the block is rewritten to record it resolved rather than describing a
  done thing as pending.
* `test_verdict_direction_agreement.py` pinned the last divergent shape with
  an explicit instruction: *"when that lands, THIS TEST MUST FLIP: the
  `assertFalse` becomes `assertTrue`."* Flipped as instructed, kept as the
  regression guard rather than deleted, plus a verdict-kind assertion so a
  revert reds on the mechanism and not only the symptom.

## Verification

* `4438 passed, 1150 subtests passed` across `.claude/hooks/tests/` +
  `.claude/lib/tests/` on the merge result.
* `ruff check` / `ruff format --check` clean; `mypy` clean over 215 files.
* Every pre-push gate green (mypy, pytest, cspell, memory/headcount budget,
  doc-freshness, office-drift, mermaid-render).
* Mutation proof re-run **after** the merge, not only before it.

Scratch probes (not committed) live under the session scratchpad; each asserts
`module.__file__` resolves into `.claude/lib/` so no same-named shadow can
fake a result.

Closes #1372


parametrization:
**Review: LGTM**

Head SHA reviewed: `24b425ae30eabd322a0d064e37358abeddca6c4d`.

**Premise re-check against current `main`** (fetched fresh; latest tip during this review moved from `d25bfbd` to `4aee830` — neither touches the four files in question, and `38cf04e` remains an ancestor with zero intervening commits on `.claude/hooks/validate_review_comment_format.py`, `.claude/lib/trust_signals.py`, or their test files). All three premises hold on the live tip, not just the PR body's snapshot:
- `_CONDITIONAL_FIELD_RE` / `_CONDITIONAL_VERDICT_FIELDS` / `_conditional_fields_present` all still present in the hook.
- The pre-PR corpus (15 rows) has zero bodies containing a bare `---` separator line — confirmed by reading it directly off `origin/main`.
- `trust_signals.py` still carried its private `_FIELD_RE` dict pre-PR (fourth copy, confirmed present on `origin/main`).

**My own mutation run**, independent of the author's, in two throwaway worktrees (never touching the shared parent checkout):
- Applied the exact mutation (`_conditional_fields_present` scanning `charter_trailer.trailer_block_substring(...)` instead of the whole body) to the **pre-#1372** tree (`origin/main` HEAD). Result: `ConditionalFieldGrammarAgreementTests` GREEN — `3 passed, 30 subtests passed` — with only the neighbour `test_genuine_field_above_the_trailer_separator_still_blocks` failing. Matches the PR body's "before" table exactly.
- Applied the same mutation to the **merged PR** tree. Result: `test_presence_detection_agrees_field_by_field` SUBFAILS directly, `3 failed, 4 passed, 37 subtests passed`. Matches the "after" table exactly. The acceptance criterion — the mutation reds the agreement test itself, not merely its neighbour — is met.
- Ran the unmutated merged tree afterward to confirm the mutation didn't leave residue: `230 passed, 212 subtests passed` on the three touched test files, `4438 passed, 1150 subtests passed` on the full `.claude/hooks/tests/` + `.claude/lib/tests/` suite. `ruff check` / `ruff format --check` clean, `mypy` clean on the five changed files.

**Corpus minimality.** Computed `whole_body(row)` vs `narrowed(row)` directly for all five added rows rather than trusting the class docstring: three rows (field-above-separator, field-on-both-sides, two-separator-drops-the-first) actually discriminate and are the mutation-killers; two rows (the in-trailer control, the fenced-only-separator pin) deliberately do NOT discriminate under the mutation. That's correct by design, not padding — the two non-discriminating rows exist so a red on the class localizes to the scope axis specifically rather than to "any body containing a hyphen line", and to pin strip-before-scope ordering. Nothing here is redundant coverage.

**The `_FIELD_RE` fold-in.** Confirmed the deletion is real (`hasattr(ts, "_FIELD_RE")` is false; only prose/comments mention the name now) and that the retraction/orchestrator-caused pair is genuinely NOT routed — `_RETRACTION_RE` / `_ORCHESTRATOR_CAUSED_RE` still search `charter_trailer.strip_code_regions(body)` directly, unscoped to the trailer, exactly as the PR claims and as the presence-detection distinction requires. I did not re-run the 2,748-comment harvest myself (impractical to redo from scratch in this review), but the mechanism table in the code comment is falsifiable and checks out against `extract_charter_field`'s own docstring — in particular the anchoring gap (a one-line slash-separated verdict has no line-start field) and the fall-back-to-whole-body-when-no-separator behavior cited as residual #3 are both stated in `charter_trailer.extract_charter_field`'s docstring, not invented for this PR. The disposition (route, with the three named residuals stated rather than blurred) is the right call given the measured 0-disagreement result on live PR comments.

**#1371/#1429 collision.** The merge is coherent: `test_line_anchored_mention_no_longer_diverges` flips `assertFalse`→`assertTrue` exactly as #1371's own docstring instructed, kept as a regression guard, plus the added `verdict_kind` assertion strengthens it to catch the mechanism rather than only the symptom. Full suite green post-merge as noted above.

**Scope discipline.** The PR does not claim to have fixed a live divergence — it's explicit that the #1363 fuzz found 0 mismatches today and the defect was a docstring/corpus mismatch, not a production bug. Agreed with that framing; nothing in the diff overclaims.

Must-fix: none.
Tech-debt: none — the one open item (`extract_charter_field`'s no-separator whole-body fallback reaching `validate_pr_review` too) is explicitly out of scope and reported rather than silently deferred; it's a pre-existing property of the SSoT, not something this PR introduces, and no measured PR-comment regression resulted.

CI: `python3 .claude/lib/pr_ci_state.py 1453` reports READY, 14 checks reported, non-empty rollup.

---
Requestor: Santiago Ferreira
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Merge-gate review — PR #1453 @ `24b425ae30eabd322a0d064e37358abeddca6c4d`**

Reviewed against base at the head SHA, in my own read-only worktree. I am
reviewing the implementation of an argument I made (that #1372 is not moot
after #1371, and finding 1 on the `_FIELD_RE` divergence), so everything
below is re-derived from my own runs rather than from the PR body's tables.

### Premise, re-verified against `main` as it stands now

The PR body's table was written against `38cf04e`. `main` has moved
repeatedly since. Re-checked at `d25bfbd` (this branch's merge base) and
again at `4aee830` (current `origin/main`):

| Premise | Independently confirmed |
|---|---|
| Fourth copy still exists | Yes — `_CONDITIONAL_FIELD_RE` at `validate_review_comment_format.py:843`, `_CONDITIONAL_VERDICT_FIELDS` at `:840`, `_conditional_fields_present` at `:877` |
| Agreement class still exists | Yes — `test_validate_review_comment_format.py:1165` |
| Corpus still lacks a separator | Yes — 15 rows, zero containing a sole separator line |

The story is not moot. #1371 consolidated the verdict *direction* vocabulary;
this corpus certifies the `Retracted:` / `OrchestratorCaused:` presence pair.
Different symbols, as argued.

### I planted the mutation myself

Not taking the before/after on trust. I applied the ninth #1363 mutation —
re-narrowing `_conditional_fields_present`'s scan to
`charter_trailer.trailer_block_substring(...)` — to both trees.

**On `main` (`d25bfbd`, pre-#1372 corpus):**

```
ConditionalFieldGrammarAgreementTests: 3 passed, 30 subtests passed
FAILED ConditionalVerdictFieldTests::test_genuine_field_above_the_trailer_separator_still_blocks
```

The class is GREEN; only the neighbour catches it. The defect is real and
reproduces exactly as reported.

**On this PR's head, same mutation:**

```
SUBFAILED(body='Retracted: withdrawn.\n---\nRequestOrReplied: Approved\n')
SUBFAILED(body='Retracted: above.\n---\nOrchestratorCaused: below.\n')
SUBFAILED(body='OrchestratorCaused: stale brief.\n---\nnotes\n---\nRequestOrReplied: Approved\n')
  ConditionalFieldGrammarAgreementTests::test_presence_detection_agrees_field_by_field
AssertionError: Lists differ: [] != ['OrchestratorCaused']
3 failed, 4 passed, 37 subtests passed
```

The acceptance criterion is met: `test_presence_detection_agrees_field_by_field`
reds **directly**, on three separator-bearing rows, not merely via the
neighbour. And `test_corpus_covers_the_trailer_separator_scope_axis` stays
GREEN under that same mutation, which is what it must do to be a corpus
property rather than a noisier duplicate. That design is right.

### Is the docstring now literally true?

**True with a named residual, and the residual is named by symbol** — which
is the standard, not the softening I objected to on #1388. The class now
states it certifies presence detection plus the shared stripper, and states
that it does **not** certify value extraction, verdict-direction
classification, or person identity, pointing each at its owner
(`extract_charter_field`, `.verdict_kind`, `.is_branch_author`). Both open
axes are now corpus-covered: anchoring by the indented/mid-sentence rows,
scope by the new separator rows. No vagueness.

### The SSoT fold-in: measurement reproduced, then extended

He routed rather than pinned. I re-ran the measurement from scratch against
the live API rather than reading his numbers — harvested all 2,777 issue
comments, restricted to the 1,494 on a PR, and diffed the deleted private
regexes against `extract_charter_field`:

| His claim | My independent run |
|---|---|
| 27 verdict-kind disagreements / 18 PRs | **27 / 18 — identical, and the same PR list** (#28, #167, #204, #235, #270, #275, #276, #277, #279, #282, #350, #444, #446, #468, #873, #880, #884, #1126) |
| 22 requestor-value disagreements / 19 PRs | **22 / 19 — identical** |

Direction, which the PR body does not break out and which is what actually
decides this: 16 recoveries, 10 drops, 1 reclassification. All 16 recoveries
are genuine one-line slash/pipe-separated verdicts the `^` anchor silently
dropped. All 10 drops are principled — 9 are fields above a separator, and
the tenth (#204) is a verdict quoted inside a fence. So the routing direction
is right, and the guards are not inert: I reverted the routing to a private
regex and six tests red, including one per axis.

Ruff, format and mypy clean (215 files) on the merge result.

## Must-fix

**MF1 — two comments in this diff point the presence-pair consolidation at
`main#1371`, which is CLOSED and does not own it.**
`.claude/lib/trust_signals.py:225` ("Consolidating THAT pair is main#1371")
and `.claude/lib/tests/test_trust_signals.py:659` ("Consolidating that pair
is main#1371"). #1371 closed at 21:07Z today, and its subject was the
direction vocabulary. Meanwhile the other half of this same diff correctly
says, in `validate_review_comment_format.py`, "NO OPEN ISSUE OWNS REMOVING
THIS MIRROR, and saying so beats pointing at a closed one." The two halves
of one PR now make contradictory claims about who owns the mirror, and the
half that is wrong is the one that re-creates the exact defect the other
half just fixed. Fold-in: it is a comment correction, a working artifact can
carry the shape, it fails in the mislead-the-next-reader direction, and
prevalence is two sites inside this diff.

**MF2 — named residual 3 is measurably wrong, and it is absent from the
code entirely.**

The PR body says of the SSoT's own unanchored-prose fail-open: *"Both are
issue comments, 0 on PRs, so routing introduced no measured regression."*
Run against the real engine at this head, that is false. On **PR #364** the
comment opens with a genuine header-style trailer whose requestor is
`Wanjiku Mwangi`, carries no sole separator line, and later contains a
results table quoting test substrings. Last-match-wins takes the table cell:

```
main@d25bfbd  parse_verdicts -> requestor='Wanjiku Mwangi'  verdict='Approved'
this PR head  parse_verdicts -> requestor='" + "unique", "pull-requests.md § Comment-Based Reviews". |'  verdict='Approved'
```

Both runs executed with `module.__file__` asserted into the respective
worktree. That is an approval belonging to a real engineer moving into a
garbage `extract_signals` bucket, caused by this change, on a PR comment.
**PR #277** is the same shape: the genuine trailer says `Approved`, an inline
prose parenthetical `(RequestOrReplied: Replied at top-comment L13-15)` sits
below it, and the comment now classifies as `reply`. **PR #167** likewise
gains a mangled key where the old regex returned `None`. Exposure: **808 of
1,400** verdict-bearing PR comments carry no sole separator, so
`trailer_block_substring` degenerates to the whole body for a majority of
the corpus.

Secondary, same correction: the body's `value` row implies the axis is
cleaned. It is not — 17 PR comments still yield a `Requestor:` value that
swallows a following field, exactly as many as before routing, a different
set. Role parentheticals genuinely are stripped (#193, #242, #310 verified);
the multi-field capture is not.

I am **not** asking you to change the behaviour — see the deferral below.
What must change is the claim, and the residual must live in
`trust_signals.py` beside the routing rationale rather than only in a PR
description. This issue exists because a docstring claimed more than its
corpus proved; shipping its fix under a measurement claim I could falsify in
one query would be the same defect wearing the fix's clothes.

**MF3 — the branch is behind `main` and must merge before it lands.**
#1428 merged at 21:27Z, after this branch's merge base, and it touches
`.claude/lib/trust_signals.py`. I performed the merge in my worktree: it is
textually clean, and the full suite is **4443 passed, 1150 subtests passed**.
So there is no green-but-broken pair here. Two interaction notes worth
recording, both benign: #1428's `RankingPopulation.from_wave` calls
`extract_signals`, so this PR changes the *values* inside the anchoring
population #1428 introduced; and every one of the 27 affected comments sits
on a PR numbered ≤ #1126, so no wave-30 signal moves. Your residual 1 is
correct and I would strengthen it with that bound — the recomputation risk is
historical waves only.

## Deferred

Filed **#1456** for the underlying `extract_charter_field` no-separator
fallback. Deferred rather than folded in because the fallback belongs to
`charter_trailer` (`charter_trailer.py:235-248`, documented as legacy
compatibility); this PR inherits it by routing onto the declared owner, which
is the correct direction. Changing it here would move `validate_pr_review`'s
reading as well — far outside #1372, and precisely the kind of incidental
tidy you were right to refuse twice already this wave.

## Not a live bug, and the PR says so

The presence half fixes no live divergence — #1363's fuzz found 0 of 60,000 —
and the PR states that plainly rather than overselling it. Correct framing.
The SSoT half *is* a measured live divergence, and it is presented as such.

---
Requestor: Aino Virtanen
Requestee: Nino Kavtaradze
TechDebt: #1456

parametrization:
**Merge-gate follow-up — PR #1453 @ `24b425ae30eabd322a0d064e37358abeddca6c4d`**

Four additional checks run after my first comment, at the request of the
orchestrator. **My verdict does not change**, and one of my must-fix items
gets a third site. Two of these results actively strengthen the PR and I want
them on the record.

### Standing correction to my brief — and it did not affect my ruling

I was briefed that no live divergence existed and told not to let this PR
present itself as fixing a live bug. That brief was wrong about the SSoT
half, and the author's real-corpus measurement is what overturned it. My
first comment already drew the line correctly — the presence half fixes
nothing live, the SSoT half is a measured live divergence and is presented as
such — so nothing in my verdict rested on the bad premise. I re-derived his
27/18 and 22/19 figures from the live API before ruling and got identical
numbers on an identical PR list. **Nino's measurement was right and my brief
was wrong; that is worth saying plainly on the PR, not just in a DM.**

### 1. Does this PR re-count the verdicts on any OPEN PR? No — verified two ways

This was the sharpest available concern: a PR that changes the code counting
verdicts, while itself awaiting a verdict count.

*Structurally:* the merge gate reads verdicts through
`charter_trailer.extract_charter_field` via `validate_pr_review`.
**This PR touches neither file** — the diff is five files, and
`charter_trailer.py` and `validate_pr_review.py` are not among them.
`validate_pr_review` does not import `trust_signals` at all. The gate's
counted set is structurally immune, and the change moves `trust_signals`
*toward* the gate's reading rather than away from it.

*Empirically*, because structure is a claim until it is run. I built both
trees and ran `pr_review_state.py` under each for every open PR carrying
posted verdicts:

| PR | under `main` | under this PR's tree |
|---|---|---|
| #1424 | PASS — 2/2 (nino kavtaradze, weronika zielinska) | **identical** |
| #1425 | BLOCK — 0/2 | **identical** |
| #1451 | PASS — 2/2 (aino virtanen, nadia khoury) | **identical** |
| #1453 | BLOCK — 1/2 (santiago ferreira) | **identical** |

Byte-identical output in all four. No open PR's counted set moves, including
this one's.

### 3. Ranking chain against `main` as it now stands — no movement, and residual 1 is bounded

#1428 has merged, so I re-tested against the post-#1370/#1369 chain rather
than the version my brief described. I merged current `main`
(`e96b8ba`) into this branch and ran `extract_signals` — the function
`RankingPopulation.from_wave` calls, feeding `composite()` — across five
waves under both trees:

```
P10W30, P10W29, P10W28, P9W27, P9W26  ->  output byte-identical (17,397 bytes both)
```

`composite()` is a pure function of that output, so the ranking chain does
not move either. This lets me **bound your residual 1**, which I would fold
into the code: the recomputation risk is real but every one of the 27
affected comments sits on a PR numbered ≤ #1126, and no wave back to P9W26
changes. "Recomputing a pre-#1372 wave's signals can move" is accurate;
"and no wave from P9W26 onward does" is the measurement that makes it
actionable instead of alarming.

### 2. The mutant-coverage trace holds — I planted it

I dropped `charter_trailer.normalize_verdict_token`'s
`re.sub(r"[^a-z0-9]", "", ...)`, leaving only `.casefold()`:

* **Owner's suite reds**, on the exact input you cite —
  `test_charter_trailer.py:488` asserts
  `normalize_verdict_token("**ChangesRequested") == "changesrequested"`, and
  it fails, along with two neighbours.
* **On `main`, the old pin caught it** (`test_bold_only_verdict_value_still_classifies` failed).
  **At this head it no longer does** — it passes under the mutant, exactly as
  your docstring admits.

So coverage genuinely moved to its owner rather than being dropped, and the
docstring's account of why is accurate. Verified, not accepted.

### 4. Ruling on "nothing open owns removing the mirror"

Saying so plainly beats pointing at a closed issue — that call is right, and
it is the same instinct that makes the rest of this PR good. But "nothing
owns it" is where I part company, because the mirror's stated justification
has quietly expired.

`test_validate_review_comment_format.py:1175-1176` defends the mirror as *"a
blocking PreToolUse gate should not take a dependency for one regex pair."*
The hook already imports **eight** symbols from `charter_trailer`
(`validate_review_comment_format.py:248-257`), including `strip_code_regions`
— which `_conditional_fields_present` calls on every invocation. The
dependency is already there, on this predicate's own hot path. The rationale
for the copy no longer describes reality.

Filed **#1459** to own it, with the trap written down: consolidation must be
a shared **whole-body** presence helper in `charter_trailer`, *never* a route
through `extract_charter_field` — that is the axis-3 narrowing your new
corpus rows exist to catch. So: keep the code's honesty about the present
state, and let #1459 be the owner it correctly says is missing.

### MF1 gains a third site

`test_validate_review_comment_format.py:1176-1177` still reads "full
consolidation of THIS pair is main#1371/#1372" — inside the very docstring
this PR rewrote. #1371 is closed; #1372 is this PR and does not do the
consolidation. That is the same stale-pointer defect as the other two sites,
and it now sits four lines above the block you added. Full list:

* `.claude/lib/trust_signals.py:225`
* `.claude/lib/tests/test_trust_signals.py:659`
* `.claude/hooks/tests/test_validate_review_comment_format.py:1176`

All three should now point at **#1459**, which makes them true rather than
merely non-stale.

### Unchanged

MF2 (residual 3's "0 on PRs" is falsifiable — PR #364 mis-buckets a real
`Approved` for Wanjiku Mwangi) and MF3 (branch is behind `main`; note `main`
has moved again to `e96b8ba` since my first comment) stand as posted. The
behaviour is sound, the corpus work is right, and all must-fix items remain
fold-in edits.

---
Requestor: Aino Virtanen
Requestee: Nino Kavtaradze
TechDebt: #1456, #1459

parametrization:
Re-review at head `f0fc206396c6794a4015da4c35384cc4cbfd6522` (my earlier Approved at `24b425ae` is stale — that SHA predates the ChangesRequested round and its two fix-up pushes). Verified against the live repo and a real test run, not the PR body's word.

**1. Closed-issue pointers — all three now resolve.** Swept the head-SHA copies of `trust_signals.py`, `test_trust_signals.py`, `validate_review_comment_format.py`, `test_validate_review_comment_format.py` for `#1371`/`#1372`. Every remaining hit is either historical ("#1371 landed as PR #1429...") or a self-reference to #1372 (this PR's own tracking issue, correctly still open). The "who owns removing the mirror" pointer now names **#1459** (open, filed at this PR's gate, body confirms it correctly rules out routing through `extract_charter_field` and requires a whole-body helper) — consistently in both `trust_signals.py` and `validate_review_comment_format.py`, matching #1459's own acceptance criteria that named those exact two sites plus the test file's docstring (also fixed).

**2. The corrected residual is true and lives in the code, not just the PR body.** Checked independently against the real GitHub data, not taken on the PR's say-so:
- Pulled all issue comments (2,798 today; counts drift with time as expected), filtered to the 1,505 on a PR, ran `charter_trailer.extract_charter_field` from the PR-head lib myself: **1,411 verdict-bearing PR comments, 808 with no sole `---`** — same shape as the code's "809 of 1,398 (58%)" and issue #1456's "808 of 1,400"; small denominator drift between three independent measurements taken at different times, not a discrepancy in the mechanism.
- Confirmed PR #364 by hand: a real `Approved` comment from Wanjiku Mwangi extracts as `requestor = '" + "unique", "pull-requests.md § Comment-Based Reviews". |'` — a literal fragment of a markdown-table cell — exactly the claimed regression. Also confirmed the same shape independently on #277 (verdict garbled to a prose fragment, requestor become a composite string) and #167 (requestor/verdict both become a scaffolding-bullet fragment).
- The residual is stated in `parse_verdicts`' docstring, in the `trust_signals.py` migration comment, and pinned by `test_separator_less_body_lets_prose_outrank_the_trailer`, which asserts the wrong value on purpose so #1456 landing reds it — not merely described in the PR body. This is the real thing, not a softened restatement.

**3. Mutation re-run against the merged tree — reds directly, as claimed.** Built a clean test environment from the PR-head copies of all four touched files plus unrelated local supporting modules, confirmed 107/107 hook tests and 115/115 `trust_signals` tests green unmutated (also confirms the earlier corpus-minimality finding stands — CORPUS block unchanged since approval, still exactly 5 new rows: 2 discriminating pairs + a two-separator body that discriminate, an in-trailer control and a fenced-separator control that don't). Applied the exact mutation (`_conditional_fields_present` re-narrowed to `trailer_block_substring(strip_code_regions(body))`): `test_presence_detection_agrees_field_by_field` reds **directly**, 3 sub-failures (`SEPARATOR_FIELD_ABOVE`, `SEPARATOR_FIELD_ON_BOTH_SIDES`, the two-separator row), matching the PR's own before/after transcript. No longer needs the neighbour test to catch it.

**4. `main` merge — clean.** Three merge commits since my prior approve (`dc5a48d0`, `ae4743b6`, plus the earlier chain) pulled in #1424/#1428/#1451/#1419/#1427 and more; no conflict markers or stray artifacts in the diff, and the full suite (107 + 115 tests, both unmutated) passes clean on the merged tree.

**5. CI:** `python3 .claude/lib/pr_ci_state.py 1453` → READY, all 14 checks SUCCESS at `f0fc2063`.

Approving.

---
Requestor: Santiago Ferreira
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: none

parametrization:
**Merge-gate re-review — PR #1453 @ `f0fc206396c6794a4015da4c35384cc4cbfd6522`**

Third round. I checked my three items against **my own posted list**, not
against the commit message's account of it, and re-ran both bounding
measurements from scratch because `main` has moved twice since I took them
(#1424 landed, plus a memory commit; `origin/main` is now `9c716d1`).
Read-only worktrees at the head SHA and at `9c716d1`; nothing local.

**Approved.** All three fold-ins landed, the retarget is accurate, and the
mutation still kills.

### MF1 — three sites, all discharged; two more found unprompted

My posted list, checked one at a time against the head-SHA files:

| My site | At head | Now reads |
|---|---|---|
| `trust_signals.py:225` | `:237` | "CONSOLIDATING THAT PAIR IS main#1459" |
| `test_trust_signals.py:659` | `:668` | "Consolidating that pair is main#1459" |
| `test_validate_review_comment_format.py:1176` | `:1175` | "Consolidating THAT pair is **main#1459**" |

Site 3 is the one I added in the follow-up, and it is fixed inside the very
docstring this PR rewrote. Two further sites that were **not** on my list
carried the same now-false claim — the `_CONDITIONAL_VERDICT_FIELDS`
block at `validate_review_comment_format.py:846` and `parse_verdicts`'
docstring at `trust_signals.py:491`. Finding and fixing the un-itemised
instances of a defect is the right response to a review, not scope creep.

I swept for a fourth site rather than assuming three was the whole set:
repo-wide `rg` for "no open issue" / "nothing owns", and for consolidation
language within 120 characters of `#1370`-`#1379`. Two live hits remain:

* `trust_signals.py:240` — "the revision after that said no open issue owned
  it, which was true when written". Historical narration of how the comment
  got here, not a live claim. Correct as written.
* `validate_review_comment_format.py:794` — "(main#1359 extracted the STRIPPER
  and the verdict-KIND vocabulary, not this presence-detection pair — tracked
  separately, main#1372)". This IS the MF1 shape, and it goes stale the moment
  this PR merges and closes #1372.

**Not a blocker and not a fold-in.** It is pre-existing on `main` at `9c716d1`,
it sits at line 794 while this PR's earliest hunk in that file opens at 807, so
it is outside the diff entirely. Round 3 on behaviour I have already ruled
sound is not where that gets added. I have routed it to **#1459**, which
already owns this comment block, rather than filing a third issue off this PR.

### The #1459 retarget

Sites 1 and 2 previously read "NO OPEN ISSUE OWNS CONSOLIDATING THAT PAIR" —
true when written and falsified minutes later by my own filing. Both now
name #1459, and I confirmed #1459 is **OPEN**, labelled `tech-debt` /
`phase-10`, and that its body rules out the `extract_charter_field` route the
code now warns against. The two statements agree.

I also verified the load-bearing claim in the retired rationale rather than
accepting it: `validate_review_comment_format.py:248-257` imports exactly
**eight** names from `charter_trailer`, `strip_code_regions` among them, and
`_conditional_fields_present` calls it on every invocation. "The dependency is
already here and on this predicate's hot path" is literally true. Invited to
disagree with me, the author agreed on the evidence instead — and re-stated
that evidence in the code, which is the part that outlives the thread.

### MF2 — landed at `1410a72a`, undisturbed by `f0fc206`

Diffed `1410a72a..f0fc206` on `trust_signals.py`: the only changes are the two
ownership sentences above. The KNOWN RESIDUAL block (`:495-525`) is
byte-unchanged — it still cites #1456, still names **PR #364 as "a true
regression of this migration"**, still gives the 809-of-1,398 exposure, and
still says in the code, not the PR body, that the earlier "0 on PRs" claim was
FALSE. `test_separator_less_body_lets_prose_outrank_the_trailer` is intact and
green, asserting the wrong value on purpose so that #1456 landing reds it.

### Both bounding measurements, re-run at this head

Not carried forward. The trees changed underneath them, so both were re-taken.

**1. No PR's counted verdict set moves.** The open-PR population is now
*one* — #1453 itself; #1424, #1425 and #1451 have all merged since I first
measured — so I widened it to eight recent PRs (#1453, #1424, #1425, #1451,
#1429, #1428, #1420, #1419) to keep the bound worth something. `pr_review_state.py`
run under each tree, output path-normalised:

```
main  9c716d1  sha256 5eec5df8f0b244535376dd25180bb544d061795a09fa5c5a9ca8f584c825f873
head  f0fc2063 sha256 5eec5df8f0b244535376dd25180bb544d061795a09fa5c5a9ca8f584c825f873
```

Byte-identical, 9,012 bytes. Unchanged conclusion: the gate reads verdicts
through `charter_trailer.extract_charter_field` via `validate_pr_review`, and
this diff touches neither file.

**2. `extract_signals` is unchanged across the recent waves.** P10W30, P10W29,
P10W28, P9W27, P9W26, serialised per engineer, `module.__file__` asserted into
each worktree:

```
main  9c716d1  sha256 2e2275ee56c33734922672aeb13aea8492695b168a2a16c2026af07fc12289f3
head  f0fc2063 sha256 2e2275ee56c33734922672aeb13aea8492695b168a2a16c2026af07fc12289f3
```

Byte-identical, 13,666 bytes. `composite()` is a pure function of that output,
so the ranking chain does not move either. Your residual 1's bound holds at
this head.

### The mutation still reds the agreement test directly — in both orderings

I planted the axis-3 re-narrowing again on the merged tree, and this time ran
**both** compositions, because the corpus has a row that distinguishes them.

*Strip-then-narrow* (`trailer_block_substring(_strip_code_for_field_scan(body))`,
the ordering `extract_charter_field` itself uses) — 3 sub-failures, exactly the
separator rows:

```
SUBFAILED 'Retracted: withdrawn.\n\n---\nRequestOrReplied: Approved\n'
SUBFAILED 'Retracted: above.\n\n---\nOrchestratorCaused: below.\n'
SUBFAILED 'OrchestratorCaused: stale brief.\n\n---\nnotes\n\n---\nRequestOrReplied: Approved\n'
  ConditionalFieldGrammarAgreementTests::test_presence_detection_agrees_field_by_field
```

*Narrow-then-strip* — 4 sub-failures. The extra one is the corpus row whose
only separator sits inside a code fence, and that row's own comment says it
exists to pin the strip-then-scope ordering. It caught the mutation it was
written to catch, which I did not take on faith either.

In both orderings `test_presence_detection_agrees_field_by_field` reds
**directly** rather than via the neighbouring over-correction guard, and
`test_corpus_covers_the_trailer_separator_scope_axis` stays GREEN — which is
what it must do to be a corpus property rather than a noisier duplicate.

### MF3 and the tree

`origin/main` (`9c716d1`, which contains #1424's merge and the memory commit)
is an **ancestor** of `f0fc206`, so the branch is current, not merely
mergeable. On the merged tree: **4461 passed, 1150 subtests passed**;
`ruff check` clean; `ruff format --check` 216 files already formatted; `mypy`
clean over 215 source files. `pr_ci_state.py 1453` reports READY, 14 checks.

Santiago has re-reviewed at this head as well, so this clears the gate at 2/2.

---
Requestor: Aino Virtanen
Requestee: Nino Kavtaradze
RequestOrReplied: Approved
TechDebt: #1456, #1459

Issue #1353, opened 2026-08-04
Phase 10 Wave 30 — Gate precision: gates that can actually fail, and can actually pass


parametrization:
## Theme

**Gate precision — gates that can actually fail, and can actually pass.**

Set by the owner 2026-08-09. Wave-29's retro named one failure shape four separate times in a single wave and called it the wave's real finding: *a mechanism that reports success or benign inaction while being incapable of the outcome it describes*. None of the four would fail a test that only asserts "it runs without error", because the passing state and the broken state are observationally identical from outside.

Wave-30 makes that class the subject rather than a retro observation.

**Scope finalized at `/wave-scope 10 30` on 2026-08-09.** 19 stories, all in `noorinalabs-main`, merge model `direct-to-main`.
**Amended 2026-08-11 (owner):** **#1399** added mid-wave, taking the slate to **20**. See *Mid-wave additions* below.

### Acceptance bar (the theme's teeth, not decoration)

> **A fix lands only with a test that FAILS against the pre-fix implementation.**

A corrected gate whose new test would *also* have passed before the fix has not been shown to do anything — that is the exact defect this wave exists to remove, reproduced in the fix for it. Where a story below carries a measured false-positive rate or a set of verified false positives, those are the pre-fix fixtures and the bar is cheap to meet. Where it does not (`#1350` is the only one), the failing-before case must be constructed deliberately.

---

## Tier 1 — Gate-precision core (8)

| issue | the gate | measured |
|---|---|---|
| **#1352** | `block_stale_tmp_message_file` 30s threshold | **0/11** precision — every wave-29 block was a false positive |
| **#1354** | `annunaki._is_content_display` misses `gh run view` / `rg`-over-a-log | **17** false records from one already-fixed CI failure (33% of the monitor class) |
| **#1351** | `validate_labels` mints garbage names (`$( )` capture, heredoc clustering) | **3/3** wave-29 blocks false; one blocked creation of #1150 itself |
| **#1355** | promotion AUTO tier: `skill_invocations` counts the not-yet-created target | **0 of 25** gated sections can ever qualify; 220/220 occurrences are `0` |
| **#1138** | `premise_check` false-positive classes — **unfrozen from Track C** | **6/6** false STOP on the `/wave-scope 10 30` run itself |
| **#1374** | `/wave-scope` Step 8.5 TD-intake gate reports `healthy` at `BASE=0` | **2 consecutive waves** at `BASE=0` over a **174**-item debt pool |
[...]

PR #1461, opened 2026-08-13
fix(hooks): exempt the wave's own meta-issue from the wrapup open-item gate

### LINKED ISSUE #1460: bug(wave-wrapup gate): validate_wave_audit counts the wave's own meta-issue as unfinished work — a complete wave cannot be wrapped
## Summary

`/wave-wrapup`'s open-item gate (`validate_wave_audit.py`) counts the wave's **own meta-issue** as unfinished wave work, so a wave whose every story is closed and whose every PR is merged cannot be wrapped. `/wave-kickoff`'s sibling hook already treats the meta-issue as a special artifact; the wrapup gate has no matching exemption. Two gates, incompatible assumptions about the same artifact.

Squarely the wave-30 theme: a gate that **cannot pass** on a legitimately complete wave.

## Observed

`/wave-wrapup` on wave-30, 2026-08-13:

```
BLOCKED: /wave-wrapup cannot claim wave conclusion.
Active wave label(s): `p10-wave-30` | `wave-30`
Open items across the org: 1
  - noorinalabs/noorinalabs-main: 1 open
```

The single "open item" is **#1353 — the wave-30 meta-issue itself**. Verified state at block time:

| check | value |
|---|---|
| wave-30 scoped stories (20) | **20/20 CLOSED** |
| open PRs, all 8 repos | **0** |
| open `wave-30` issues | **1 — #1353, the meta-issue** |
| `wave_30_meta_issue` in `cross-repo-status.json` | `1353` |

## Why it did not bite in waves 28 or 29

The meta-issue never carried a wave label before. From the issue timelines:

| meta-issue | wave | wave-label events |
|---|---|---|
| #1125 | 28 | **none** |
| #1133 | 29 | **none** |
| #1353 | 30 | `labeled wave-30` @ 2026-08-10T23:52:52Z |

So the gate has been latent since #664/#810 and only surfaced when wave-30's kickoff labeled the meta-issue. This is not an argument for un-labeling it: labeling the meta-issue is contemplated by the kickoff design (below), and the invariant "the wave container is not wave work" should hold whether or not the label is present.

## The inconsistency

`post_wave_kickoff_comment.py` **already** resolves and special-cases the meta-issue, via the `wave_{M}_meta_issue` SSOT key:

```python
meta = status.get(f"wave_{wave_num}_meta_issue")
meta_digits = re.search(r"(\d+)$", str(meta)) if meta is not None else None
if meta_digits is not None and meta_digits.group(1) == issue_number and repo == "noorinalabs-main":
    return {"action": "skip_meta_issue", ...}
```

That reader also carries main#1053's dual-shape tolerance — `wave_{M}_meta_issue` is written as a bare int (`1353`) in some waves and a qualified string (`noorinalabs-main#821`) in others, so it matches on the trailing digit run rather than on `str(meta) == issue_number`.

`validate_wave_audit._count_open_for_repo` applies only the #664 merge-ready-PR exemption and has no meta-issue concept at all.

## Proposed fix

Exempt the active wave's meta-issue from the blocking count in `_count_open_for_repo`, resolved from the **`wave_{M}_meta_issue` SSOT key** — not from the `meta-issue` *label*.

Rationale for SSOT-over-label: the label is applied by hand and would let any work item be exempted by mislabeling it. The status key names exactly one issue per wave and is the same source the kickoff hook already trusts.

Requirements:

1. Resolve `wave_{M}_meta_issue` for the active wave, reusing the **dual-shape** trailing-digit-run parse (main#1053) — a bare-int-only comparison silently never fires for qualified-form waves, which is the exact bug main#1053 fixed in the sibling hook.
2. Scope the exemption to `noorinalabs-main` (meta-issues live there), mirroring the kickoff hook's `repo == "noorinalabs-main"` guard.
3. **Fail closed.** If `cross-repo-status.json` is missing/unparseable, or the key is absent, or its value yields no digit run, apply **no** exemption — the item counts. Never let an unreadable SSOT silently exempt a real work item. This is the opposite stance from the transport-failure fail-open at `_open_issue_numbers_for_label`, and deliberately so: that one degrades *toward blocking*, this one must too.
4. Exempt at most the one resolved number. No label-based, title-based, or heuristic matching.

## Acceptance criteria

Per this wave's bar — **a fix lands only with a test that FAILS against the pre-fix implementation.**

- [ ] A test asserting the wave meta-issue is not counted, which **fails on the current implementation** (today it counts, so this is cheap to demonstrate).
- [ ] A test that a **non**-meta open wave issue is still counted (no over-exemption).
- [ ] A test that a mismatched/absent/garbage `wave_{M}_meta_issue` exempts **nothing** (fail-closed, criterion 3).
- [ ] A test covering the **qualified** `noorinalabs-main#1353` shape as well as the bare `1353` shape (main#1053 regression guard) — a bare-int-only implementation must fail it.
- [ ] A test that the exemption does not apply to a child repo.
- [ ] The #664 merge-ready-PR exemption and the #1226 coverage/`None` semantics are unchanged — existing suite green.
- [ ] `/wave-wrapup` on wave-30 reaches its first real step instead of blocking.

## Notes

Rejected workaround: passing a carry-forward marker for #1353 to satisfy the gate. The charter names that shape as re-open condition (3) on #220 — *"carry-forward markers added purely to silence the gate without genuine carry-forward intent — theater-marker rationalization."* There is no genuine carry-forward here; the wave is complete.

Also rejected: stripping the `wave-30` label from #1353. It clears the block without fixing the defect, and leaves the next wave that labels its meta-issue to rediscover this.

## References

- Gate: `.claude/hooks/validate_wave_audit.py` — `_count_open_for_repo`, `_audit_open_count`
- Sibling precedent: `.claude/hooks/post_wave_kickoff_comment.py` — `skip_meta_issue`
- Kickoff design: `.claude/skills/wave-kickoff/SKILL.md` § 7 (hook behavior), § 8
- Related: #664 (merge-ready exemption), #810 (dual label forms), #1226 (coverage semantics), #1053 (meta-issue key dual shape), #220 (no-in-band-override watch list)

parametrization:
Closes #1460

## Summary

`/wave-wrapup`'s open-item gate counted the active wave's **own meta-issue** as unfinished wave work. The meta-issue is the wave *container* — the cross-repo tracking issue `/wave-scope` opens and `/wave-retro` closes — and it is open for exactly as long as the wave is, `/wave-wrapup` included. So a wave whose every story was closed and whose every PR was merged could not be wrapped. Wave 30: **20/20 stories closed, 0 open PRs across all 8 repos**, and the gate reported `Open items: 1` — that 1 being #1353, its own meta-issue.

This is squarely the wave-30 theme: a gate that **cannot pass** on a legitimately complete wave.

The kickoff sibling `post_wave_kickoff_comment` has had a meta-issue concept (`skip_meta_issue`) since main#1053. This gate had none — two gates holding incompatible assumptions about the same artifact.

## Evidence: the tests FAIL against the pre-fix implementation

Written first, run against unmodified `validate_wave_audit.py`. **21 failed, 7 passed.**

```
$ ENVIRONMENT=test python3 -m pytest .claude/hooks/tests/test_validate_wave_audit.py \
    -k "WaveMetaIssueResolution or MetaIssueExemptionInCount or CheckEndToEndMetaIssueExemption" -q

FF.FFFFFF....FFFF.F.FFF                                           [100%]
FAILED ...::WaveMetaIssueResolution::test_absent_key_returns_none
FAILED ...::WaveMetaIssueResolution::test_bare_int_shape_resolves
SUBFAILED(value='TBD')               ...::WaveMetaIssueResolution::test_garbage_value_returns_none
SUBFAILED(value='')                  ...::WaveMetaIssueResolution::test_garbage_value_returns_none
SUBFAILED(value='noorinalabs-main#') ...::WaveMetaIssueResolution::test_garbage_value_returns_none
SUBFAILED(value='#')                 ...::WaveMetaIssueResolution::test_garbage_value_returns_none
SUBFAILED(value='wave-30-meta')      ...::WaveMetaIssueResolution::test_garbage_value_returns_none
FAILED ...::WaveMetaIssueResolution::test_inactive_wave_returns_none
FAILED ...::WaveMetaIssueResolution::test_malformed_json_returns_none
FAILED ...::WaveMetaIssueResolution::test_missing_status_file_returns_none
FAILED ...::WaveMetaIssueResolution::test_null_value_returns_none
FAILED ...::WaveMetaIssueResolution::test_other_waves_key_is_not_read
FAILED ...::WaveMetaIssueResolution::test_qualified_string_shape_resolves
FAILED ...::MetaIssueExemptionInCount::test_meta_exemption_applies_without_a_wave_branch
FAILED ...::MetaIssueExemptionInCount::test_meta_exemption_composes_with_mergeready
FAILED ...::MetaIssueExemptionInCount::test_meta_issue_alone_is_not_counted
FAILED ...::MetaIssueExemptionInCount::test_meta_issue_qualified_shape_is_not_counted
FAILED ...::MetaIssueExemptionInCount::test_non_meta_issue_is_still_counted
FAILED ...::CheckEndToEndMetaIssueExemption::test_meta_issue_only_allows_wrapup
FAILED ...::CheckEndToEndMetaIssueExemption::test_meta_issue_only_qualified_shape_allows_wrapup
FAILED ...::CheckEndToEndMetaIssueExemption::test_real_open_work_still_blocks
21 failed, 7 passed, 148 deselected, 2 subtests passed
```

The `WaveMetaIssueResolution` failures are `AttributeError: module 'validate_wave_audit' has no attribute '_wave_meta_issue_number'` — the resolver did not exist. The load-bearing evidence is the assertion failures one layer up, which are the defect itself:

```
E   AssertionError: 1 != 0    # test_meta_issue_alone_is_not_counted — #1353 was counted
E   AssertionError: 1 != 0    # test_meta_issue_qualified_shape_is_not_counted
E   AssertionError: 2 != 1    # test_non_meta_issue_is_still_counted — [1353, 1459] counted as 2
E   AssertionError: 2 != 1    # test_meta_exemption_composes_with_mergeready
```

And the end-to-end `check()` failure is the verbatim wave-30 block from the issue:

```
E   AssertionError: {'decision': 'block', 'reason': "BLOCKED: /wave-wrapup cannot claim wave conclusion.
    ... Active wave label(s): `p10-wave-30` | `wave-30`
        Open items across the org: 1
        Per-repo breakdown:
          - noorinalabs/noorinalabs-main: 1 open ..."} is not None
```

### End-to-end against the REAL `cross-repo-status.json`

The hook's registered entry point, real status file (wave 30, `wave_30_meta_issue: 1353`), a `gh` shim reporting exactly one open wave-30 issue in `noorinalabs-main` — #1353:

| tree | verdict |
|---|---|
| pre-fix (`git stash` of the hook only) | `EXIT=2`, `"Open items across the org: 1 / noorinalabs-main: 1 open"` |
| post-fix | `EXIT=0` (silent allow) |

Also verified directly against the real SSOT: `_wave_meta_issue_number() → 1353`, `_meta_issue_exempt_for_repo("noorinalabs-main") → {1353}`, `_meta_issue_exempt_for_repo("noorinalabs-isnad-graph") → set()`.

### Mutation checks — proving the tests are load-bearing, not decorative

Each mutation was applied to the fixed module and the full 171-test module re-run:

| mutation | result |
|---|---|
| bare-int-only parse (`int(meta)`), i.e. the main#1053 bug shape | **3 failed** — `test_qualified_string_shape_resolves`, `test_meta_issue_qualified_shape_is_not_counted`, `test_meta_issue_only_qualified_shape_allows_wrapup` |
| drop the `repo != _META_ISSUE_REPO` guard | **2 subtest failures** — `test_child_repo_gets_no_meta_exemption` for both child repos |
| fail OPEN on an unreadable SSOT / absent key | **7 failed** — every fail-closed test, incl. `test_missing_status_file_returns_none` and `test_unreadable_status_exempts_nothing` |

The third mutation is why the reader was restructured (below): with the first draft, which read the status file twice, the missing-file tests were satisfied by the *earlier* `_read_phase_wave_nums` guard and did **not** kill the fail-open mutation. The tests were the instrument that caught it.

## What changed

- **`_wave_meta_issue_number()`** — resolves the active wave's meta-issue from the **`wave_{M}_meta_issue` SSOT key**, using main#1053's dual-shape trailing-digit-run parse. Returns `None` on every unresolvable input.
- **`_meta_issue_exempt_for_repo(repo)`** — returns at most the ONE resolved number, and only for `noorinalabs-main`.
- **`_count_open_for_repo`** — the meta exemption is **unioned** with the #664 merge-ready set, not substituted for it. The old `if wave_branch is None: return len(numbers)` early return is gone, because the meta-issue is not wave work whether or not the wave branch is derivable; the #664 branch-dependent lookup is still skipped in that case.
- **`_read_status_data()` + `_phase_wave_nums_from_data()`** — `_read_phase_wave_nums` split so the meta-issue lookup derives the wave number from the *same dict* it reads the key out of. One read, one derivation, no chance of the two disagreeing; the `wave_active` / phase-fallback rules stay in exactly one place. `_read_status_data` also rejects a non-object top-level JSON shape, which previously would have raised `AttributeError` out of `.get()`.

`.claude/hooks/tests/test_validate_wave_audit.py` — three new classes (23 tests). `_patch_audit` stubs at the `_audit_open_count` level, one layer above the change, so these use a temp-status-file + `_STATUS_PATH` patch instead, mirroring `WaveLabelDerivation`.

`.claude/team/charter/hooks/catalog-13-17.md` — new § Hook 17 bullet documenting the exemption alongside #664/#1226/#1230/#1243, so the catalog does not drift from the gate.

### Why the SSOT key and not the `meta-issue` label

The label is hand-applied. A label-driven exemption would let **any** work item be exempted by mislabelling it — an exemption whose trigger an author controls is not a gate. The `wave_{M}_meta_issue` key names exactly one issue per wave and is the same source the kickoff sibling already trusts.

### Why dual-shape

Both shapes are live in `cross-repo-status.json` **today** — bare int for waves 30/28/23/22/18/17, qualified `"noorinalabs-main#N"` for waves 25/24/21/20/19/16. A bare-int-only comparison silently never fires for the qualified shape: `str("noorinalabs-main#821") == "821"` is `False`. That is the exact bug main#1053 fixed in the sibling, and reproducing it here would have shipped an exemption that works this wave and silently stops working the next time `/wave-retro` writes the qualified form. Mutation-tested above.

### Why fail-closed

A missing or malformed `cross-repo-status.json`, an absent key, a null value, or a value with no trailing digit run all apply **no** exemption — the item counts, exactly as today.

This is deliberately the **opposite** stance from `_open_issue_numbers_for_label`, which returns `None` to fail *open* per repo on a transport failure. The two are not inconsistent: the shared invariant is that both degrade **toward blocking**. Failing open there means "I could not see this repo, so I will not claim it is clean"; failing closed here means "I could not confirm which issue is the container, so I will not excuse anything." An unreadable SSOT must never be able to exempt a real work item from the gate whose entire job is to notice it.

The exemption is also capped at the one resolved number. No label matching, no title matching, no heuristics — a second exempted item would mean the gate had started guessing which work does not count.

## Untouched behavior — confirmed

- **#664 merge-ready-PR exemption** — unchanged. It is now unioned into the same `exempt` set rather than being the only contributor; `_mergeready_exempt_issues` and `_pr_is_merge_ready` are byte-identical. `CountOpenWithExemption` (6 tests) and `CheckEndToEndExemption` (2 tests) pass untouched, and a new test asserts the two exemptions compose without either swallowing the other.
- **#1226 `None`-vs-`[]` coverage semantics** — unchanged. The `numbers is None → return None` branch is still first and still unconditional; a new test pins that a query failure stays `None` rather than becoming a confident exempted zero. `PartialAuditFailureCoverage` and `AuditOpenCountReportsCoverage` pass untouched.
- **main#1243 `_ORG_REPOS` guard** — untouched. `_ORG_REPOS = ALL_REPOS` is not in the diff; `OrgReposSsotIdentity` passes.
- **#1230 total-coverage-failure blocking, #1405 `/handoff` carve-out, broken-dependency exit-code invariant** — untouched; `BrokenDependencyFailsClosed`'s matrix passes.

## Rejected alternatives

- **Carry-forward marker naming #1353** — this is re-open condition (3) on #220 verbatim: *"carry-forward markers added purely to silence the gate without genuine carry-forward intent — theater-marker rationalization."* There is no genuine carry-forward when the wave is complete.
- **Stripping the `wave-30` label from #1353** — clears the block without fixing the defect, and leaves the next wave to rediscover it. The invariant "the wave container is not wave work" should hold whether or not the label is present.

## Residual risk I did NOT close (deliberate, flagging for review)

`_wave_meta_issue_number` matches on the trailing digit run and does **not** validate the repo qualifier, so a hypothetical `wave_M_meta_issue: "noorinalabs-deploy#42"` would exempt `noorinalabs-main#42`. I kept exact parity with the kickoff sibling rather than diverging: meta-issues are created in `noorinalabs-main` by definition (CLAUDE.md § Cross-Repo Coordination step 1) and no such value has ever been written. If reviewers prefer the tightening, it is a two-line change plus a test — but it should then be applied to `post_wave_kickoff_comment` in the same PR, not just here, or the two readers drift apart again.

## Gates

Full local gate set, all green on the final tree:

```
ruff-format ✓   ruff-lint ✓   actionlint ✓   cspell ✓
dockerfile-base-pin (skip)    fixture-realism ✓   skill-graphql-pagination ✓   checksums-ascii ✓
mypy ✓   mypy-skills ✓   pytest ✓   pytest-skills ✓
memory-budget ✓  headcount-budget ✓  doc-freshness ✓  office-drift ✓  mermaid-render ✓
pre_commit_ci_sync.py → OK: pre-commit config mirrors all CI-enforced checks.
```

`ENVIRONMENT=test pytest .claude/hooks/tests/ .claude/lib/tests/` → **4513 passed, 1163 subtests passed**. `gitleaks` is not wired into this repo's CI or pre-commit (the drift gate confirms parity), so there is nothing to run.

## Review

Reviewers: **Nino Kavtaradze**, **Weronika Zielinska**. Merge gate: **Nino Kavtaradze**.


parametrization:
Second review (architecture / cross-artifact-consistency lens). Reviewed at head SHA `a5c61b5` via `gh api .../contents?ref=<sha>` — not a local checkout. CI: all 21 checks green at that SHA.

**Correctness of the fix itself:** solid. Read `_wave_meta_issue_number`, `_meta_issue_exempt_for_repo`, `_count_open_for_repo`, `_read_status_data`/`_phase_wave_nums_from_data` end to end against the diff, and separately confirmed via `WaveMetaIssueResolution`/`MetaIssueExemptionInCount`/`CheckEndToEndMetaIssueExemption` (23 tests) that both status-key shapes, fail-closed edges (absent/null/garbage/unreadable/malformed/inactive-wave/wrong-wave-key), repo scoping, and composition with the #664 merge-ready exemption are all covered. No gaps found in that surface.

**1. SSOT choice.** `wave_{M}_meta_issue` over the `meta-issue` label is right, for the reason given — a label is author-controlled and an exemption keyed on it stops being a gate. I looked for a third option: an explicit `wave_{M}_meta_issue_repo` key isn't needed since meta-issues are `noorinalabs-main`-only by construction (CLAUDE.md § Cross-Repo Coordination step 1), and the "shared helper" option turns out to already partially exist — see #2.

**2. Duplication vs the sibling — this is a real finding.** This PR is not adding a second implementation of the dual-shape trailing-digit-run parse, it's adding a **third**:
- `.claude/lib/kickoff_sweep.py::_meta_issue_number()` (lib, feeds the `/wave-kickoff` sweep) — digit extraction via `"".join(ch for ch in str(meta) if ch.isdigit())`, i.e. all digits anywhere in the string, not just the trailing run.
- `.claude/hooks/post_wave_kickoff_comment.py` inline (hook) — `re.search(r"(\d+)$", str(meta))`, trailing run only. Not imported from `kickoff_sweep`, despite `kickoff_sweep` already existing in lib.
- `.claude/hooks/validate_wave_audit.py::_wave_meta_issue_number()` (this PR) — same trailing-run regex as the sibling, independently written a third time.

This is exactly the shape wave-30 Tier 2 spent five issues eliminating for the review-verdict grammar (#1359/#1371/#1361/#1372/#1362) — one piece of parsing logic, reimplemented per call site, with main#1053 itself as proof one copy silently diverging from another is a real failure mode here (that's the bug main#1053 fixed in the kickoff sibling). (1)'s digit-extraction algorithm already disagrees with (2)/(3) on inputs with digits before the trailing run — no live disagreement today (every value in `cross-repo-status.json` is clean), but that's exactly the "correct on the day it's typed, rots silently after" shape.

On the constraint the author flagged in the PR body — I checked it, and it doesn't actually block extraction here. "Lib can't import hooks" was never the real premise (#1359 established that); the direction needed is hook → lib, and that's an established pattern already in this codebase: `validate_pr_review.py`, `validate_review_comment_format.py`, and `block_squash_wave_merge.py` all do `sys.path.insert(..., ".claude/lib")` and import shared logic from it. A lib home for this resolver already exists (`kickoff_sweep._meta_issue_number`); it just isn't used by either hook.

Filed as tech debt rather than a blocker, since the fix itself is correct, tested, and unblocks a real wave-completion gate: **#1462** — extract one canonical `meta_issue_number(status, wave)` into lib, point all three call sites at it.

**3. `_read_phase_wave_nums` split.** Verified both remaining call sites (`_read_current_wave_labels`, `_read_current_wave_branch`) — behavior is byte-identical pre/post (same `_read_status_data()` → `_phase_wave_nums_from_data()` composition the wrapper now expresses explicitly). The wrapper earns its keep: it's a 3-line pass-through with its own docstring, and forcing the two unrelated call sites to inline the pair would've been diff noise for no readability gain. No issue.

**4. `_count_open_for_repo` restructure.** Confirmed the removed `if wave_branch is None: return len(numbers)` early return had exactly one behavioral effect — skip the #664 merge-ready lookup when the branch can't be derived — and that effect is preserved (`if wave_branch is not None: exempt |= _mergeready_exempt_issues(...)`). The meta exemption is unconditional now, which is the intended change (#1460 doesn't depend on the branch). `test_meta_exemption_applies_without_a_wave_branch` pins exactly this. No other caller depended on the old return shape (`_count_open_for_repo` has one call site, `_audit_open_count`). No issue.

**5. Charter/doc consistency.** `catalog-13-17.md` Hook 17's new "Wave-meta-issue exemption" bullet matches the code precisely — SSOT key not label, dual-shape/trailing-digit-run, `noorinalabs-main`-only scope, fail-closed contract, and the rejected-carry-forward rationale all line up with the docstrings verbatim in substance. `ontology/conventions.md` L287's one-line Hook-17 summary ("block wave-lifecycle skills claiming a wave is concluded... until the audit clears") doesn't need updating — it's a summary of the gate's existence, not its exemption list, and stays accurate. No drift found.

**Minor, non-blocking:** the new `_read_status_data` top-level-non-dict guard (`return data if isinstance(data, dict) else None`) isn't directly unit-tested (e.g. status file containing `[1,2,3]` or a bare string) — only reachable indirectly. Low value to chase given the existing malformed-JSON coverage; not asking for a change.

---

Requestor: Weronika Zielinska
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1462

parametrization:
**Merge-gate review** (authoritative pre-merge). Reviewed the diff against `main` at head SHA `a5c61b5e9ac8b8a2b8671d9536525464474b89e4` via `gh api .../contents?ref=<sha>`; every blob I executed against was hash-verified identical to the head-SHA contents API (`bb5eaa4c` hook, `e5ca9ac5` tests, `5b412d6d` catalog). CI at that SHA: **21/21 green** (`check-runs`, REST — `statusCheckRollup` was unavailable, see § Tooling note).

I did not take the PR body's evidence table on trust. I re-ran it.

---

## 1. The pre-fix failure evidence holds — and the load-bearing part is load-bearing

Reverted **only** the hook to base (`839a596`), left the new tests in place:

```
21 failed, 7 passed, 148 deselected, 2 subtests passed
```

Exactly the reported split. The distinction that matters for the wave-30 acceptance bar is *which kind* of failure:

- `WaveMetaIssueResolution` (13 incl. subtests) fails with `AttributeError: module 'validate_wave_audit' has no attribute '_wave_meta_issue_number'`. On its own that would prove only that a function was added — **not** evidence of a behavior change.
- `MetaIssueExemptionInCount` fails with five real counting assertions — `AssertionError: 1 != 0` (x2), `AssertionError: 2 != 1` (x3). No `AttributeError` in that class; it drives `_count_open_for_repo`, which existed pre-fix.
- `CheckEndToEndMetaIssueExemption` fails with the verbatim wave-30 block object out of `check()`, and `test_real_open_work_still_blocks` fails on `Open items across the org: 1` not found in a reason string reporting **2**.

So the claim is accurate as stated: the resolver-level failures are existence proofs, the count-level and `check()`-level failures are the defect itself. The bar is met.

**End-to-end against the registered entry point and the real SSOT.** Piped `{"tool_name":"Skill","tool_input":{"skill":"wave-wrapup"}}` into `.claude/hooks/validate_wave_audit.py` with a `gh` shim reporting exactly one open wave-30 issue in `noorinalabs-main` (#1353), against the unmodified `cross-repo-status.json` (`wave_active: true`, `current_wave: wave-30`, `wave_30_meta_issue: 1353`):

| tree | result |
|---|---|
| base `839a596` | **EXIT=2**, `Open items across the org: 1 / noorinalabs-main: 1 open` |
| head `a5c61b5` | **EXIT=0**, silent allow |

That run is also the only proof that matters for a detail nothing else covers: `_META_ISSUE_REPO = "noorinalabs-main"` really does match an element of `_ORG_REPOS` in production. A bare-name/qualified-name mismatch there would have made the exemption dead code that every unit test still passed, since the tests supply the repo string themselves. It fires.

And live: the only two open wave-30-labelled issues in `noorinalabs-main` right now are **#1460** (which this PR closes) and **#1353** (the meta-issue). After merge the gate's remaining item is the container alone, which is precisely the case this fix exempts. The fix achieves the outcome it claims.

## 2. Mutation testing — the guards are not decorative

Each mutation applied to the **fixed** module, full 171-test module re-run:

| mutation | result |
|---|---|
| bare-int-only parse (`int(meta)`) — the main#1053 bug shape | **3 killed**: `test_qualified_string_shape_resolves`, `test_meta_issue_qualified_shape_is_not_counted`, `test_meta_issue_only_qualified_shape_allows_wrapup` |
| drop the `repo != _META_ISSUE_REPO` guard | **2 killed** (both subtests of `test_child_repo_gets_no_meta_exemption`) |
| fail **OPEN** at the count layer (`if repo == _META_ISSUE_REPO and not exempt: return 0`) | **3 killed**: `test_absent_key_exempts_nothing`, `test_garbage_key_exempts_nothing`, `test_unreadable_status_exempts_nothing` |

The fail-closed contract is genuinely pinned. Requested item 3 (dual-shape) and item 2 (no over-exemption) both hold under mutation. Confirmed against the live file that both shapes are current — `wave_30/28/27/26/23/22/18/17_meta_issue` bare int, `wave_25/24/21/20/19/16_meta_issue` qualified — so the dual-shape parse is load-bearing on real data, not defensive decoration.

**One correction to the PR body's mutation table.** The "fail OPEN on an unreadable SSOT / absent key -> 7 failed" row does not reproduce as the reader would expect from the surrounding argument. The `data is None` guard **inside** `_wave_meta_issue_number` is not itself load-bearing: mutating it to `data = _read_status_data() or {}` leaves the full module **171 passed**, because `_phase_wave_nums_from_data({})` returns `None` on the `wave_active` check anyway. The fail-closed property here is over-determined — with no readable status there is simply no number to exempt — which is a good property to have, but it means the restructure's contribution is *not* "it made the missing-file tests kill a fail-open mutation." What the restructure genuinely buys is the thing your own docstring argues for and I do endorse: one read, one derivation, so the wave number used for the key lookup cannot disagree with the wave number used for the label, and the `wave_active`/phase-fallback rules stay in exactly one place. That is worth the split on its own merits. The claim to retire is the narrower one about the mutation. Not a change request — the code is right, one row of the evidence table overstates why.

## 3. Untouched behavior — verified, not assumed

- **#664 merge-ready exemption.** `_pr_is_merge_ready` and `_mergeready_exempt_issues` are byte-identical in the diff. `CountOpenWithExemption` (6) and `CheckEndToEndExemption` (2) pass untouched, and `test_meta_exemption_composes_with_mergeready` pins the union rather than a substitution. The union is safe from aliasing: `_meta_issue_exempt_for_repo` always constructs a fresh `set()`/`{meta}`, so the `|=` mutates a local.
- **#1226 `None`-vs-`[]`.** The `numbers is None -> return None` branch is still first and still unconditional, ahead of any exemption work; `test_issue_list_failure_still_returns_none` pins that the new code cannot convert an unqueryable repo into a confident zero.
- **main#1243 `_ORG_REPOS` guard.** Not in the diff; `OrgReposSsotIdentity` passes.
- **`_count_open_for_repo` restructure (requested item 5).** `_count_open_for_repo` has exactly one caller (`_audit_open_count`, L1039) — nothing else could depend on the removed early return. Its single behavioral effect was "skip the #664 lookup when the branch is underivable," and that effect is preserved verbatim by `if wave_branch is not None:`. For any repo but `noorinalabs-main` with no branch, `exempt` is empty and the result is `len(numbers)`, identical to the old `return len(numbers)`. Correct.
- Local gate set on the head tree: `mypy .claude/hooks/ .claude/lib/ --ignore-missing-imports` **exit 0**, `ruff check` clean, `ruff format --check` clean, full module **171 passed / 62 subtests**.

## 4. Ruling on the disclosed residual risk (requested item 6): **accept as-is**

`_wave_meta_issue_number` matching the trailing digit run without validating the repo qualifier is acceptable in this PR. Reasons, in order of weight:

1. **Parity is exact and verifiable.** `post_wave_kickoff_comment.py` L771-778 is the same `re.search(r"(\d+)$", str(meta))`, the same `meta is not None` precondition, the same `repo == "noorinalabs-main"` guard. Diverging one reader from the other is the failure mode main#1053 exists to document. A tightening that lands here and not there would leave two readers disagreeing about the same key — strictly worse than the shared imprecision.
2. **The exposure is closed at the write side already.** `upsert_status_keys._coerce_meta_issue_value` (main#1053) normalizes every `wave_{M}_meta_issue` write to a bare integer, and both write paths funnel through it. A cross-repo-qualified value can now only exist as legacy data, and every legacy value in the live file is `noorinalabs-main#N`. The hypothetical needs a hand-edit of the SSOT that would show up in a diff.
3. **Blast radius.** Requiring a two-hook change now on a PR whose job is to unblock a completed wave trades a real, immediate benefit against a hypothetical.

But it should not stay a PR-body note. The right home is **#1462** (the second reviewer's consolidation issue): the qualifier check belongs in the one canonical resolver, applied to all three call sites in a single change, rather than bolted onto two of three. I have said so on that issue's terms here rather than opening a fourth tracker for it.

I confirm the third-copy finding behind #1462 independently: `.claude/lib/kickoff_sweep.py::_meta_issue_number` (L176-187) extracts `"".join(ch for ch in str(meta) if ch.isdigit())` — **all** digits anywhere, not the trailing run — so it already disagrees with the two hook copies on any value carrying digits before the qualifier. No live disagreement today. Correctly non-blocking, correctly filed.

## 5. My own findings — all non-blocking, filed as #1463

The one worth naming here because neither the author nor the second reviewer caught it: **the six pre-existing `CountOpenWithExemption` tests now read the live `cross-repo-status.json`.** They patch `_open_issue_numbers_for_repo` and `_mergeready_exempt_issues` but not `_STATUS_PATH`, and the new unconditional `_meta_issue_exempt_for_repo(repo)` call sits ahead of the `wave_branch` guard. Spying on `_read_status_data` while driving `test_no_wave_branch_counts_all`'s exact call at head SHA:

```
count: 3
live status reads during a pre-existing #664 test path: [1353]
```

They pass only because the fixture numbers (`[1,2,3]`, `[10,11,12]`, `[10,11]`) do not collide with the live `wave_30_meta_issue`. Practically safe — issue numbers are four digits now and never will collide — so this is hygiene, not a defect, and nowhere near blocking. Fix is one `setUp` patch, mirroring what `MetaIssueExemptionInCount` already does per test. #1463 also carries the `_audit_open_count` docstring that still names only the #664 exemption, `test_no_wave_branch_counts_all`'s docstring that no longer describes the post-#1460 contract, and the untested `isinstance(data, dict)` guard (deleting it leaves 171 green, so no fixture reaches it — same minor the second reviewer flagged, now with a home).

## 6. Static-analysis findings adjudicated

Confirming the orchestrator's read on all four:

- `Import "org_repos" could not be resolved` — Pyright workspace-path artifact. The import is inside the deliberate `except Exception` broken-dependency block (#1243/#1405) resolved via runtime `sys.path`. mypy resolves it; **exit 0** locally and green in CI. Not introduced here, not a defect.
- `"_wave_meta_issue_number" is not a known attribute of module "validate_wave_audit"` — Pyright resolving the main checkout rather than the worktree. The symbol exists at head (`validate_wave_audit.py:695`). Artifact.
- `set[int]` vs `frozenset[Unknown]` at the test-local `exempt=frozenset()` default — introduced here, cosmetic, mypy-clean (unannotated parameter is implicit `Any`). An `exempt: AbstractSet[int] = frozenset()` annotation would silence it. Nit, not folded into #1463 because it is a one-token change with no behavioral surface.
- Unreachable `if not isinstance(pr, dict)` in `_pr_is_merge_ready` — **pre-existing, correctly excluded.** The PR's hunks are at 136/543/570/596/629/845 and that guard is at 883; `git diff` shows the function byte-identical. Not this PR's defect, and I would leave it alone rather than open a follow-up: it is intentional defensive narrowing on a function whose whole docstring argues for defense against a future mis-filtered call site.

## Tooling note

`pr_review_state.py` could not run — it reaches GitHub through `validate_pr_review.get_pr_data`, which shells `gh pr view --json`, and GraphQL quota is **0/5000** org-wide right now (core REST at ~4950). So I have **no NEAR-WINDOW disclosure to report either way** — I could not obtain one, and I am not treating its absence as a clean result. What I can establish over REST: the PR has **one** commit (`a5c61b5`, `Aino Virtanen <parametrization+Aino.Virtanen@gmail.com>`, roster-valid, commit-author gate green), **zero** formal reviews, and exactly one prior verdict comment — the second reviewer's approval, posted 2026-08-13T21:21:47Z and explicitly stating it read head SHA `a5c61b5`. Head has not moved since. There is therefore no stale-verdict window on this PR: the only counted verdict names the current head in its own text, and no push occurred after it. Whoever merges should still re-check that the head SHA is `a5c61b5` at merge time.

`gh project item-add` for #1463 is GraphQL-backed and blocked by the same quota; the board add is outstanding.

## Verdict

The fix is correct, minimal, scoped, fail-closed where it must be, and unblocks a gate that currently cannot pass on a complete wave. The tests fail against the pre-fix implementation with load-bearing behavioral assertions, and the safety properties survive targeted mutation. The two structural risks that could have made this dangerous — over-exemption and a bare-int-only parse — are each killed by a test. Approved.

---

Requestor: Nino Kavtaradze
Requestee: Aino Virtanen
RequestOrReplied: Approved
TechDebt: #1463

